Issue #71244, opened 2022-12-15
eCommerce Flow - design tweaks


blackjackkent:
Notes from Saxon in Slack regarding tweaks for the eCommerce flow: 
* [x] Like Newsletter we can use loader bar at top to signify progress (#71341)
* [x] I noticed labels on forms inputs differ from newsletter, should look like screens (#71435)
* [x] The old address form was also shown for me post checkout now, intentional? (To be removed as part of #71472)
* [x] Domains aren't pre-populated with domain suggestions based on store name like Newsletter flow _(Can't reproduce)_.
* [x] [Adding](https://github.com/Automattic/wp-calypso/issues/71244#issuecomment-1354405313) "WooCommerce Powered" to welcome screen , loading and profiler (see Newsletter for an example and existing styles) (#71391)
* [x] To be consistent with other tailored flows it looks like we should re-use the initial background image for our loader (#71476, **needs review**)

SaxonF:
One more little thing is domains aren't pre-populated with domain suggestions based on store name like Newsletter flow.

https://wordpress.com/start/newsletter/domains?new=My%20newsletter%20test&search=yes&hide_initial_query=yes&siteAccentColor=%23113AF5

SaxonF:
One more. Adding "WooCommerce Powered" to welcome screen , loading and profiler (see Newsletter for an example and existing styles)

![image](https://user-images.githubusercontent.com/1072756/208060162-a620a629-9588-4ad7-b091-3e8e26fd927b.png)

![image](https://user-images.githubusercontent.com/1072756/208060180-3d60a207-3014-4168-add8-48725e39c4ff.png)

![image](https://user-images.githubusercontent.com/1072756/208060222-b677a54c-7c26-46d9-97b2-ae18323bfe27.png)

blackjackkent:
> One more little thing is domains aren't pre-populated with domain suggestions based on store name like Newsletter flow.
> 
> https://wordpress.com/start/newsletter/domains?new=My%20newsletter%20test&search=yes&hide_initial_query=yes&siteAccentColor=%23113AF5

@SaxonF I can't reproduce your issue with this. When I use (for instance) the site title "Test Test ABC", I get these initial results prepopulated on the domain step:

<img width="1218" alt="image" src="https://user-images.githubusercontent.com/13437011/208437062-1565abcc-fed3-4fdb-aabf-c08ccd08a9ef.png">

Can you let me know what you're seeing on your end?

[...]

PR #71341, opened 2022-12-19
Ecommerce Tailored Flow: Fixing ecommerce flow progress bar

blackjackkent:
#### Proposed Changes

* This PR fixes a missing set of config values for the ecommerce flow, allowing it to properly show a progress bar across the top of the screen as steps progress.

Before:
image

After: 
image

#### Testing Instructions

* Check out this PR to testing environment and navigate to `/setup/ecommerce`
* Navigate through the flow and verify that the progress bar appears and progresses appropriately through each step.

Related to #71244

### DIFF
--- a/packages/onboarding/src/flow-progress/use-flow-progress.ts
+++ b/packages/onboarding/src/flow-progress/use-flow-progress.ts
@@ -1,4 +1,4 @@
-import { LINK_IN_BIO_FLOW, LINK_IN_BIO_TLD_FLOW, ECOMMERCE_FLOW } from '../utils/flows';
+import { ECOMMERCE_FLOW, LINK_IN_BIO_FLOW, LINK_IN_BIO_TLD_FLOW } from '../utils/flows';
 
 /* eslint-disable no-restricted-imports */
 interface FlowProgress {
--- a/packages/onboarding/src/flow-progress/use-flow-progress.ts
+++ b/packages/onboarding/src/flow-progress/use-flow-progress.ts
@@ -53,13 +53,13 @@ const flows: Record< string, { [ step: string ]: number } > = {
 	[ ECOMMERCE_FLOW ]: {
 		intro: 0,
 		storeProfiler: 1,
-		storeAddress: 2,
+		designCarousel: 2,
 		domains: 3,
-		designCarousel: 4,
-		siteCreationStep: 5,
-		processing: 5,
-		waitForAtomic: 5,
-		checkPlan: 5,
+		siteCreationStep: 4,
+		processing: 4,
+		waitForAtomic: 4,
+		checkPlan: 4,
+		storeAddress: 5,
 	},
 };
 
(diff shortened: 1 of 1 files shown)


packages/onboarding/src/flow-progress/use-flow-progress.ts
```diff
@@ -1,4 +1,4 @@
-import { LINK_IN_BIO_FLOW, LINK_IN_BIO_TLD_FLOW } from '../utils/flows';
+import { LINK_IN_BIO_FLOW, LINK_IN_BIO_TLD_FLOW, ECOMMERCE_FLOW } from '../utils/flows';
```
daledupreez: Extreme nit: I'd prefer this to be in alphabetical order.
```suggestion
import { ECOMMERCE_FLOW, LINK_IN_BIO_FLOW, LINK_IN_BIO_TLD_FLOW } from '../utils/flows';
```
And this is _definitely_ not blocking!

packages/onboarding/src/flow-progress/use-flow-progress.ts
```diff
@@ -50,6 +50,17 @@ const flows: Record< string, { [ step: string ]: number } > = {
...
+		intro: 0,
+		storeProfiler: 1,
+		storeAddress: 2,
+		domains: 3,
+		designCarousel: 4,
+		siteCreationStep: 5,
+		processing: 5,
+		waitForAtomic: 5,
+		checkPlan: 5,
+	},
```
daledupreez: I believe these steps are out of order, which results in the progress bar jumping around unexpectedly.

When I work through the flow, I see the following order:
1. intro
2. storeProfiler
3. designCarousel
4. domains
5. processing (site creation)
6. checkout (not in our list)
7. checkPlan
8. processing (waitForAtomic)
9. storeAddress
```suggestion
	[ ECOMMERCE_FLOW ]: {
		intro: 0,
		storeProfiler: 1,
		storeAddress: 7,
		domains: 4,
		designCarousel: 3,
		siteCreationStep: 5,
		processing: 5,
		waitForAtomic: 6,
		checkPlan: 5,
	},
```

daledupreez (changes requested):
The steps here don't seem to be in the right order, so the progress bar jumped around unexpectedly relative to the step I was on.

blackjackkent:
> I believe these steps are out of order, which results in the progress bar jumping around unexpectedly.

@daledupreez Yep you're super correct. (The designs view in my local environment started erroring out at the end of the day, so I pushed this up with the intention of fixing that and doublechecking this morning, and yes, it is definitely out of order. 😛 That'll teach me to make assumptions.) I've pushed a fix to the step ordering that I think is correct.

daledupreez (approved):
Re-tested with the updates and everything looks good to me. The only minor gotcha is that we show the `processing` step before and after checkout, so there is no reflection of progress there, but I think that's something we should iterate on separately.

PR #71159, opened 2022-12-13
Stats: Add a tooltip for annual highlights

grzegorz-cp:
#### Proposed Changes

* adding a tooltip to display detailed numbers when hovering the number section of a highlight card
* ~~enhancing annual highlights by adding arrows to navigate between years~~ moved to #71343

Before:
Screenshot 2022-12-14 at 1 52 25 PM

After:
Screenshot 2022-12-20 at 2 17 55 PM

#### Testing Instructions

* navigate to the `Insights` section of Stats
* hover over the card's numbers and verify that a tooltip shows up. When hovering a shortened number, the tooltip should show an accurate number, not the shortened version

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #70206

Issue #70206, opened 2022-11-19
Stats: Improve annual highlight section in the insights page


jsnmoon:
Tasks originally from #69227.

- [x] Add tooltip support to the highlights component.
- [x] Add arrow navigation support for showing data from prior or next years.

# Design

- Metrics for the current year by default, with the ability to switch to previous years and back.
![Image](https://user-images.githubusercontent.com/390760/196414578-9422edfd-2bf0-4e75-bb6c-e0c9744196a0.png)

![Image](https://user-images.githubusercontent.com/390760/196414649-3a1d5b29-4021-4867-afb3-5c335d5cb7e8.png)

- When these truncated numbers are hovered/focused we can display the exact number via a tooltip (https://github.com/Automattic/wp-calypso/issues/69178). For screenreaders we can simply use the full number.

![Image](https://user-images.githubusercontent.com/390760/196414615-def22d6e-3c23-4b98-970d-cd04a523cab9.png)

PR #71343, opened 2022-12-19
Stats: Add year navigation to annual highlights

grzegorz-cp:
#### Proposed Changes
Split from #71159.

* enhancing annual highlights by adding arrows to navigate between years

#### Testing Instructions

* navigate to the `Insights` section of Stats
* verify that there are arrows present on the right side of `2022 in review` and clicking them changes visible year in review (annual stat cards below should also change)
* make sure that you can't navigate to the years not available on `View all annual insights` page - years earlier to the first data point or the future
* check that you can navigate to the arrows using a keyboard
* `Followers` count should only be visible for the current year. Past years should show an empty value represented by a hyphen.

Before:
Screenshot 2022-12-14 at 1 52 25 PM

After:
Screenshot 2022-12-20 at 2 10 59 PM

#### Pre-merge Checklist

Related to #70206


client/my-sites/stats/annual-highlights-section/index.tsx
```diff
@@ -58,6 +71,21 @@ export default function AnnualHighlightsSection( { siteId }: { siteId: number }
 	const siteSlug = useSelector( ( state ) => getSiteSlug( state, siteId ) );
 	const viewMoreHref = siteSlug ? `/stats/annualstats/${ siteSlug }` : null;
 
+	const previousYear = insights?.years?.find( ( y ) => y.year === ( year - 1 ).toString() );
```
grzegorz-cp: There's an issue [described here](https://github.com/Automattic/wp-calypso/pull/71159/files#r1051278483) about skipping years.

client/my-sites/stats/controller.jsx
```diff
@@ -167,7 +167,34 @@ export function redirectToDefaultModulePage( context ) {
 }
 
 export function insights( context, next ) {
-	context.primary = <StatsInsights />;
+	const {
```
grzegorz-cp: This has to be simplified. More details [here](https://github.com/Automattic/wp-calypso/pull/71159/files#r1051280601).

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -66,28 +66,39 @@ class StatsPeriodNavigation extends PureComponent {
...
-					className={ classNames( 'stats-period-navigation__next', {
-						'is-disabled': hideNextArrow || isToday,
-					} ) }
-					href={ `${ url }${ nextDayQuery }` }
-					onClick={ this.handleClickNext }
-				>
-					<Icon className="gridicon" icon={ arrowRight } />
-				</a>
+
+				{ ! hidePreviousArrow ? (
```
grzegorz-cp: Discussion about the HTML structure [started here](https://github.com/Automattic/wp-calypso/pull/71159/files#r1051282421).

grzegorz-cp: The comment doesn't apply anymore after I simplified the controller and used a new navigation component. So I'll revert this file to its original state.

packages/components/src/highlight-cards/annual-highlight-cards.tsx
```diff
@@ -42,33 +44,41 @@ export default function AnnualHighlightCards( {
...
+				{ header }
+				{ navigation && navigation }
+			</div>
 
 			<div className="highlight-cards-list">
 				<HighlightCard
 					heading={ translate( 'Posts' ) }
 					icon={ <Icon icon={ postContent } /> }
 					count={ counts?.posts ?? null }
+					showValueTooltip
```
kangzj: Guess the changes of `showValueTooltip` belongs to https://github.com/Automattic/wp-calypso/pull/71159

kangzj: And a few others below too

client/my-sites/stats/annual-highlights-section/index.tsx
```diff
@@ -1,9 +1,10 @@
 import { AnnualHighlightCards } from '@automattic/components';
-import { useMemo } from 'react';
+import { useMemo, useState } from 'react';
 import { useSelector } from 'react-redux';
 import QuerySiteStats from 'calypso/components/data/query-site-stats';
 import { getSiteSlug } from 'calypso/state/sites/selectors';
 import { getSiteStatsNormalizedData } from 'calypso/state/stats/lists/selectors';
+import YearNavigation from './yearNavigation';
```
dognose24: Could we use the Kebab Case naming `year-navigation` for the file name?

packages/components/src/highlight-cards/annual-highlight-cards.tsx
```diff
@@ -15,13 +15,15 @@ export type AnnualHighlightCardsProps = {
 	};
 	titleHref?: string | null;
 	year?: string | number | null;
+	navigation?: object | null;
```
dognose24:
```suggestion
	navigation?: React.ReactNode;
```

Is the type `React.ReactNode` more accurate?

client/my-sites/stats/annual-highlights-section/yearNavigation.tsx
```diff
@@ -0,0 +1,59 @@
...
+	};
+
+	const onKeyDown = ( event: React.KeyboardEvent< HTMLButtonElement >, next?: boolean ) => {
+		if ( event.key === 'Enter' || event.key === 'Space' ) {
+			handleClickArrow( next );
+		}
+	};
+
+	return (
+		<div className={ classNames( 'stats-period-navigation', 'stats-year-navigation' ) }>
```
dognose24: Maybe we can polish the component `StatsPeriodNavigation` to reuse it in the future.

packages/components/src/highlight-cards/highlight-card.tsx
```diff
@@ -27,24 +30,37 @@ export function percentCalculator( part: number | null, whole: number | null ) {
 	return Math.abs( answer ) === Infinity ? 100 : Math.round( answer );
 }
 
+const FORMATTER = new Intl.NumberFormat();
+function formatNumber( number: number | null ) {
```
dognose24: We could modify and export the [existing formatNumber](https://github.com/Automattic/wp-calypso/blob/0ef4f96bd3b1c0cea1e3effb407fca54c3adba8e/packages/components/src/shortened-number/index.tsx#L12-L14) to reuse here, which also benefits the [usage inside AllTimeHighlightsSection](https://github.com/Automattic/wp-calypso/blob/0ef4f96bd3b1c0cea1e3effb407fca54c3adba8e/client/my-sites/stats/all-time-highlights-section/index.tsx#L35-L38) in the future.

packages/components/src/highlight-cards/annual-highlight-cards.tsx
```diff
@@ -42,33 +44,41 @@ export default function AnnualHighlightCards( {
 
 	return (
 		<div className={ classNames( 'highlight-cards', className ?? null ) }>
-			{ header }
+			<div className="highlight-year-navigation">
+				{ header }
+				{ navigation && navigation }
```
dognose24:
```suggestion
				{ navigation }
```

Could the `undefined` rendering check be ignored?

dognose24:
The navigation arrows on mobile seem to need the gap to the window's right edge.

截圖 2022-12-21 上午12 45 06

packages/components/src/highlight-cards/annual-highlight-cards.tsx
grzegorz-cp: That's right. I was hoping to merge #71159 first so once I update this PR it will be already on `trunk` and won't show up as changes here. Sorry for not mentioning it earlier.

packages/components/src/highlight-cards/annual-highlight-cards.tsx
grzegorz-cp: Yes, of course, thanks for catching this.

packages/components/src/highlight-cards/highlight-card.tsx
grzegorz-cp: The first one you mentioned outputs a specific short version; if we extend it, its name won't be accurate anymore. 
We could create a utility file inside the stats directory and import the function from there. Do you think this would improve the usage here? If you think so, I could add it in another PR.

dognose24: > We could create a utility file inside the stats directory and import the function from there.

Sounds like a great plan. We could have more discussion on that PR later.

client/my-sites/stats/annual-highlights-section/index.tsx
```diff
@@ -34,7 +35,7 @@ const FOLLOWERS_QUERY = { type: 'wpcom', max: 0 };
 
 // Meant to replace annual-site-stats section
 export default function AnnualHighlightsSection( { siteId }: { siteId: number } ) {
-	const year = new Date().getFullYear();
+	const [ year, setYear ] = useState( new Date().getFullYear() );
```
jsnmoon, line 38: Could we hide the followers count when viewing a non-current year? This is what I have in mind:

#### This year:
image

#### Any past years:
image

We can probably accomplish this by tweaking the count determination on line 56:
```js
        followers: year === currentYear ? ( followers?.total_wpcom ?? null ) : null,
```

client/my-sites/stats/annual-highlights-section/yearNavigation.tsx
jsnmoon: I suspect Dognose is referring to the back/forward arrow pattern being present throughout multiple stats pages. I agree it would be nice to refactor the back/forward arrows into a reusable component.

jsnmoon (changes requested):
The overall approach looks solid.

My only blocking concern is that we're technically showing incorrect follower counts when we page into past data. Since we don't currently have access to historical follower counts, let's omit the count altogether when we've paged into past years.

client/my-sites/stats/annual-highlights-section/index.tsx
grzegorz-cp: Good point. What do you think about hiding the card (visually) instead of showing a hyphen?

client/my-sites/stats/annual-highlights-section/yearNavigation.tsx
grzegorz-cp: That would be handy. I'll add a task to our maintenance board if you don't mind.

jsnmoon: Sounds good to me!

client/my-sites/stats/annual-highlights-section/index.tsx
jsnmoon: I think it'll be better to keep the followers card there to avoid reflowing the other four cards. I'll defer to our designers if they say otherwise, but blanking the count with a hyphen is probably a conservative approach here.

grzegorz-cp: Sounds good. I also updated the testing instructions.

jsnmoon (approved):
Looks good and ready to ship!

One bug to fix in a future PR: It's possible to select a year less than the `oldestYear` by doing the following:

1) Open `/stats/insights/site-with-lots-of-history.com`.
2) Page the annual highlights into the past (e.g., 2012).
3) Use the top left sites selector control to select a new site with less annual history.
image
4) Annual highlights section remains paged to the year selected in step 2, regardless of what `oldestYear` is for the site selected in step 3.

Issue #71310, opened 2022-12-19
Checkout: hide the presales Zendesk chat when purchases are blocked


yoavf:
### Details

We should not show the ZD presales chat widget to users when they're blocked from purchasing. 
See pdm0RZ-nC-p2#comment-782

Related: #70118 where the presales chat was added to checkout

yoavf:
@sirbrillig 
I thought I could do something like
```
	const cartKey = useCartKey();
	const { responseCart } = useShoppingCart( cartKey );
	// Check if responseCart.message.errors include an object with `code=blocked`
```
But it seems like the cart clears the messages once they're displayed. 
Any suggestions on how to access the error messages (from the cart or elsewhere)?

sirbrillig:
> But it seems like the cart clears the messages once they're displayed.

Hm. I'm not sure what would do that. CartMessages displays messages but AFAIK doesn't mutate the cart object so they should still be present. 

Although it does make me wonder: cart action promises are rejected if there are errors in the messages. Could that be related to what you are seeing? I'll look more closely at your code.

yoavf:
> Hm. I'm not sure what would do that

I understood this to do that (once the messages are displayed), but maybe I'm wrong:
https://github.com/Automattic/wp-calypso/blob/37a991c492e216043578841b4a69ac1d55bc97cc/packages/wpcom-checkout/src/use-display-cart-messages.tsx#L76

sirbrillig:
Oh! Hah, yes, I guess it does. It probably shouldn't do that, but I guess there was never a need to keep those error messages around before. We could remove that behavior. It should be the display component's responsibility not to show the same message more than once. I'll make a PR... 

Update: https://github.com/Automattic/wp-calypso/pull/71346

PR #71346, opened 2022-12-19
Shopping Cart: stop clearing cart messages after displaying them

sirbrillig:
#### Proposed Changes

The `CartMessages` component is in charge of displaying any error or success messages returned by the shopping-cart endpoint any time that it is fetched. It includes code to clear the messages from the cart after they've been displayed, but this means that other components which may want to react to errors in the cart may not have time to do so.

In this PR, we remove the code that clears the cart messages.

To do this safely, we need to fix some issues with the shopping cart manager which clearing the errors was hiding:

1. Prior to this PR, if adding initial products to the cart fails during checkout, the cart will appear to be loading indefinitely. Now the cart will finish loading even if the products fail to be added (either way the user will receive an error message).
2. Prior to this PR, if a cart action caused an error (rejecting its Promise), any subsequent noop actions (eg: updating the cart's tax location if it is unchanged, which often happens upon loading checkout) will appear to fail (rejecting their Promises). Now such noop actions will appear successful since by that time the rejection will already have happened (and the cart will still contain an error).

Noticed in https://github.com/Automattic/wp-calypso/issues/71310

#### Testing Instructions

- Cause a shopping cart error. One way to do this is to try and add an invalid product to the cart via URL like `/checkout/foobar`.
- Verify that the error is displayed.
- Verify that checkout still loads, even with the error displayed.
- Without changing the cart, navigate to other pages in calypso and make sure that they do not show any additional versions of the error message (it's ok if the original message remains, although it should stay gone if dismissed).
- Next, we need to create a shopping cart success message and repeat the steps. One way to do this is to add a coupon to the cart.

### DIFF
--- a/client/my-sites/checkout/cart/cart-messages.tsx
+++ b/client/my-sites/checkout/cart/cart-messages.tsx
@@ -25,11 +25,7 @@ function CartMessage( { message }: { message: ResponseCartMessage } ) {
 
 export default function CartMessages(): null {
 	const cartKey = useCartKey();
-	const {
-		responseCart: cart,
-		isLoading: isLoadingCart,
-		clearMessages,
-	} = useShoppingCart( cartKey );
+	const { responseCart: cart, isLoading: isLoadingCart } = useShoppingCart( cartKey );
 	const reduxDispatch = useDispatch();
 
 	const showErrorMessages = useCallback(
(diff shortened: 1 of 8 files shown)


yoavf (approved):
Reviewed and tested from the POV of #71310 - this indeed solves the problem mentioned there: errors stay in the cart and can be reused by other componets. 

Didn't notice any unwanted side effects, and the everything seems to work as exepected and described.

Thanks for taking care of this @sirbrillig!

Issue #71165, opened 2022-12-13
Activity Log: Managed plugin updates look buggy


danielbachhuber:
When Atomic pushes a managed plugin update, it looks somewhat suspicious:

<img width="1109" alt="image" src="https://user-images.githubusercontent.com/36432/207448763-c1f9e5ed-fe80-46a2-9316-e605c2b99466.png">

See some conversation in p1670966377513079-slack-C7YPW6K40

**Done is:**
* Managed plugin updates look less suspicious.

danielbachhuber:
Seeking opinions on the best way to go about this: pdWQjU-b6-p2

danielbachhuber:
Landed the first part D96022-code

danielbachhuber:
🚢

PR #71355, opened 2022-12-19
Activity Logo: Use custom design for 'Server' actor

danielbachhuber:
See https://github.com/Automattic/wp-calypso/issues/71165
See conversation in pdWQjU-b6-p2

## Proposed Changes

Adds a custom design for the 'Server' actor:

image

## Testing Instructions

1. Navigate to the Activity Log for an Atomic site that's been around for a little bit.
2. Verify the Server Actor appears as expected.


simison:
Was half-expecting the icon be with grey background instead of so high contrast dark. Did you check yet with a designer?

client/my-sites/activity/activity-log-item/style.scss
```diff
@@ -234,6 +235,21 @@
 		}
 	}
 
+	.server-actor-icon {
+		/* stylelint-disable-next-line declaration-property-unit-allowed-list */
+		border-radius: 50%;
+		background-color: #23292e;
+		color: #fff;
```
simison: Would be better use CSS variables so that Calypso stays themeable and easier to maintain.

client/my-sites/activity/activity-log-item/activity-actor.jsx
```diff
@@ -27,6 +27,15 @@ const HAPPINESS_ACTOR = (
 	</div>
 );
 
+const SERVER_ACTOR = (
+	<div className="activity-log-item__actor">
+		<div class="server-actor-icon"></div>
+		<div className="activity-log-item__actor-info">
+			<div className="activity-log-item__actor-name">Server</div>
```
simison: Needs translation function

danielbachhuber:
> Was half-expecting the icon be with grey background instead of so high contrast dark. Did you check yet with a designer?

It was my "prototype for demonstration purposes" version 😁

client/my-sites/activity/activity-log-item/style.scss
```diff
@@ -234,6 +235,21 @@
 		}
 	}
 
+	.server-actor-icon {
+		/* stylelint-disable-next-line declaration-property-unit-allowed-list */
```
simison: Not in this PR but `50%` should really just be on the list of allowed radii.

simison:
>It was my "prototype for demonstration purposes" version 😁

No worries, I saw I was commenting a draft PR but wanted to drop thoughts anyway in case it helps!

client/my-sites/activity/activity-log-item/style.scss
danielbachhuber: 6d180f97b4be7ca7cea439b762487b69f218241d

client/my-sites/activity/activity-log-item/activity-actor.jsx
danielbachhuber: I was originally thinking 'Server' should be hard-coded similarly to 'Jetpack' and 'Happiness Engineer'. I suppose it makes sense to translate, though ce21a232ec891e8ace2cd3148f3db9fe1b8b1c61

client/my-sites/activity/activity-log-item/style.scss
danielbachhuber: Agree https://github.com/Automattic/wp-calypso/pull/71379

danielbachhuber:
> > Was half-expecting the icon be with grey background instead of so high contrast dark. Did you check yet with a designer?
> 
> It was my "prototype for demonstration purposes" version 😁

Asking for design feedback p1671545425735879-slack-C9EJ7KSGH

client/my-sites/activity/activity-log-item/activity-actor.jsx
simison: Jetpack is a brand thus not translated but "Happiness Engineer" should probably also be translated. Seems just an oversight.

wojtekn (commented):
Works as expected for me. As regards the visual, would it make sense to use an icon that shows a computer (e.g. `gridicons-computer`), server, or something similar instead of the "/."?

danielbachhuber:
> As regards the visual, would it make sense to use an icon that shows a computer (e.g. `gridicons-computer`), server, or something similar instead of the "/."?

@wojtekn Fair enough, the `./` is somewhat cryptic.

Here are a couple of gridicons:

**Computer** (would need to fix the clipping)

image

**Cloud**

image

I think I prefer "Cloud". What do you think?

wojtekn:
@danielbachhuber, thanks for preparing those two versions. I think the computer is a better fit. The cloud may be confusing, and it is already used in the column that shows the activity type:

208704468-489a1226-9fd3-4b27-ad28-fd886bb24db2

We may see if we get more feedback from design team, though.

danielbachhuber:
> The cloud may be confusing, and it is already used in the column that shows the activity type:

@wojtekn Oh, good point.



> I think the computer is a better fit.

What I don't love about the computer is that it looks like a _desktop computer_. Slightly different paradigm than what we're going for...

simison:
Tool, Code or Cog icons from [WP icons](https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library) might work. 

image
image

image

Computer display doesn't visually look that great. Cloud seems to work fine, too. I'd check for designer input tho.

simison:
Oh and worth checking with mobile apps so that these work there!

danielbachhuber:
Saygun gave us a fancy server icon:

image

> Oh and worth checking with mobile apps so that these work there!

Seems to work fine already!

IMG_5AF6472BFAA5-1

wojtekn:
@danielbachhuber, this one looks perfect!

simison:
Looks fine on mobile, thanks for checking. Though ideally the new actor and icon would be added there too:
- https://github.com/wordpress-mobile/WordPress-Android/search?q=actor
- https://github.com/wordpress-mobile/WordPress-ios/search?q=actor

Maybe at least a github issue as a todo?

danielbachhuber:
> Though ideally the new actor and icon would be added there too:

@simison I'm not quite sure. It appears there's deliberate logic for when an icon should appear vs. not:

https://github.com/wordpress-mobile/WordPress-iOS/blob/1910f8ace8e942b12ba85be2b800225b47529470/WordPress/Classes/ViewRelated/Activity/ActivityDetailViewController.swift#L132-L141

In any case, we can take the conversation over there: https://github.com/wordpress-mobile/WordPress-iOS/issues/19800

Issue #71326, opened 2022-12-19
LaunchPad - Implement circular progress bar


Addison-Stavlo:
Develop a new progress bar with circular style design as shared in the p2 below, and use this as the progress bar in the launchpad checklist.

New designs for a 'cohesive launchpad experience' - paYE8P-2kx-p2 - use a circular progress bar for the launchpad. This helps create a cohesive and understandable experience as we will use a similar icon to represent the launchpad in other areas.

![Screen Shot 2022-12-19 at 9 35 15 AM](https://user-images.githubusercontent.com/28742426/208449694-6da5024e-81c6-4b9e-b8c4-0a4d9ce3ca62.png)

nuriapenya:
@Sharpirate I prepared this [Figma file](9wpoYJrMlYKAKUYvVBMN74-fi-1760%3A43986) with some design guidelines. Let me know if you need something more specific :)

PR #71363, opened 2022-12-20
Components: Create Circular Progress Bar for Launchpad

sharpirate:
#### Proposed Changes
Create a new `circular-progress-bar` component and display it on the Launchpad screen

image

#### Testing Instructions
You can test this with any Launchpad flow.

1. Checkout this branch
2. Go to `http://calypso.localhost:3000/setup/link-in-bio/intro`
3. Complete the flow until you arrive at Launchpad
4. You should see the new circular progress bar showing the number of completed tasks and the number of total tasks
5. Complete a task
6. You should see both the numbers and the progress circle on the progress bar update accordingly

https://user-images.githubusercontent.com/20927667/208854642-03151d2d-5bd0-41c1-a584-b460ecd3e6b4.mov

#### Pre-merge Checklist

Related to #71326

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
@@ -1,4 +1,4 @@
-import { Gridicon, ProgressBar } from '@automattic/components';
+import { Gridicon } from '@automattic/components';
 import { useRef, useState } from '@wordpress/element';
 import { useTranslate } from 'i18n-calypso';
 import { StepNavigationLink } from 'calypso/../packages/onboarding/src';
@@ -12,6 +12,7 @@ import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { ResponseDomain } from 'calypso/lib/domains/types';
 import { usePremiumGlobalStyles } from 'calypso/state/sites/hooks/use-premium-global-styles';
 import Checklist from './checklist';
+import CircularProgressBar from './circular-progress-bar';
 import { getArrayOfFilteredTasks, getEnhancedTasks } from './task-helper';
 import { tasks } from './tasks';
 import { getLaunchpadTranslations } from './translations';
@@ -37,16 +38,19 @@ function getUrlInfo( url: string ) {
 	return [ siteName ? siteName[ 0 ] : '', topLevelDomain ? topLevelDomain[ 0 ] : '' ];
 }
 
-function getChecklistCompletionProgress( tasks: Task[] | null ) {
+function getTasksProgress( tasks: Task[] | null ) {
 	if ( ! tasks ) {
-		return;
+		// TODO: How do we want to handle that?
+		return [ null, null ];
 	}
 
-	const totalCompletedTasks = tasks.reduce( ( total, currentTask ) => {
+	const completedTasks = tasks.reduce( ( total, currentTask ) => {
 		return currentTask.completed ? total + 1 : total;
 	}, 0 );
 
-	return Math.round( ( totalCompletedTasks / tasks.length ) * 100 );
+	const totalTasks = tasks.length;
+
+	return [ completedTasks, totalTasks ];
 }
 
 const Sidebar = ( { sidebarDomain, siteSlug, submit, goNext, goToStep, flow }: SidebarProps ) => {
@@ -75,7 +79,7 @@ const Sidebar = ( { sidebarDomain, siteSlug, submit, goNext, goToStep, flow }: S
 			flow
 		);
 
-	const taskCompletionProgress = site && getChecklistCompletionProgress( enhancedTasks );
+	const [ currentTask, numberOfTasks ] = getTasksProgress( enhancedTasks );
 	const launchTask = enhancedTasks?.find( ( task ) => task.isLaunchTask === true );
 	const showLaunchTitle = launchTask && ! launchTask.disabled;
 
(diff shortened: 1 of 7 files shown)


simison:
Any possibility to add this to `@automattic/components` package if this isn't too specific to launchpad?

sharpirate:
@simison Thanks, I was thinking I should do that. I just updated the PR to move the component to `@automattic/components`

sharpirate:
### Notes
The current implementation is not 100% matching the design. Here are the differences:
1. In the design the font weight is 500 but it doesn't look like we have all of the different font weights available to us right now. I tried using the 600 font weight but it looks too thick so for now I am sticking with the 100. Let me know if that looks good.

font-weights

2. In the design the stroke width for the gray circle and the blue progress circle are different: 3px and 5px respectively. However, this is really tricky to implement in the code because that means both circles should have different radiuses and we start getting issues with the svg being cropped and scaled unevenly. For now, I kept both circles at 5px. Let me know if that looks good.

3. In the design we have a few different sizes for the progress bar. We have `40px`, `48px` and `52px`. Are we planning to have these three different versions so we can choose the appropriate size depending on where we render the progress bar (Launchpad, Site-Editor e.g.) or is the idea that the size should adjust automatically on smaller screens? Because if it's the latter this will be a problem too, because it's not as simple as changing CSS rules based on screen size. We are actually using the size in the code to calculate the length of the arc of the progress circle. So we need to know the size of the element before hand. For now, I have set the size to a static `48px`. Let me know if that looks good.

@nuriapenya Curious what your thoughts are on all of these

simison:
> [...] or is the idea that the size should adjust automatically on smaller screens? 
>
> [...] because it's not as simple as changing CSS rules based on screen size. We are actually using the size in the code to calculate the length of the arc of the progress circle.

Does `transform: scale(#)` solve above problem? You might still want to optimize how pixels look (e.g. narrower lines when circle is small) and stick to specific fixed sizes, though.

nuriapenya:
Thanks, @Sharpirate for building this so fast! 💎 

>I am sticking with the 100

Sounds good. Initially, I was using `Medium`, which I think is the equivalent of 300-400. But if that's not possible, I'd tend to go on thin rather than thick.

> For now, I kept both circles at 5px. 

Also fine. The main objective was to create a nice distinction between the two levels, but we can ditch it for now if this is too complex.

>or is the idea that the size should adjust automatically on smaller screens?

That would be my preference. Especially since we plan to use this element in different areas, and they might require a bit of scaling up or down. But we can have it fixed for now. Could we do at least two sizes? One for desktop and another for mobile?

Desktop: 48px (font-size: 15px)
Mobile: 40px (font-size: 12px)

nuriapenya:
@Sharpirate One small correction from my side: 

Let's make the stroke of all circles 4px. This way looks more balanced both on desktop & mobile. I updated [all the designs](9wpoYJrMlYKAKUYvVBMN74-fi-1760%3A43986).

Screenshot 2022-12-21 at 12 35 27

sharpirate:
> Does transform: scale(#) solve above problem?

@simison Thanks for the suggestion, that will definitely work!

sharpirate:
Thanks, @nuriapenya! I've updated the PR to have the stroke width at `4px` and added a mobile size of `40px` (and desktop at `48px`):

Circular Progress Bar Mobile Size

Actually with Mikael's suggestion of using CSS `transform: scale` we can have different sizes easily. Let me know if we should add more sizes and for what screen sizes or if we should stick with `40px` mobile and `48px` desktop for now

nuriapenya:
>Actually with Mikael's suggestion of using CSS transform: scale we can have different sizes easily. Let me know if we should add more sizes and for what screen sizes or if we should stick with 40px mobile and 48px desktop for now

Looking cool! I'd say, for now, let's leave the two fixed sizes. Once (and if) we start to use this element in more often, we can always add the `transform: scale` property.

The desktop font should be smaller. Can we try with 13px? 🙏🏽  

Thanks a lot @Sharpirate

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
```diff
@@ -37,16 +37,19 @@ function getUrlInfo( url: string ) {
...
+		return [ null, null ];
 	}
 
-	const totalCompletedTasks = tasks.reduce( ( total, currentTask ) => {
+	const completedTasks = tasks.reduce( ( total, currentTask ) => {
 		return currentTask.completed ? total + 1 : total;
 	}, 0 );
 
-	return Math.round( ( totalCompletedTasks / tasks.length ) * 100 );
+	const totalTasks = tasks.length;
```
edanzer: Worth noting that calculating and returning totalTasks here might be a bit redundant. You could already get that directly from from enhancedTasks.length without this method. That said, it's also tidy doing it this way, so works for me.

edanzer (approved):

edanzer:
This tests well for me! 
- [x] Icon looks like it should based on screenshots/gifs above.
- [x] Responsive behavior matches gif above.
- [x] Code changes look good, too.
- [x] Nice to see the documentation, and the documentation is good.
- [x] I also checked our launchpad tests. We have one for `displays a progress bar based off of task completion` that I thought might fail now, but looks like that also still passes since you've added the role=progressBar to the new icon. 

Approving with a these notes: 
   - I think we still need to make desktop font changes requested by @nuriapenya above. 
   - There are some failing tests. And given amount of time since this PR was initiated, it might be worth rebasing. If you do that and want another test, let me know.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
sharpirate: @edanzer That is a great suggestion and simplifies the function a lot (returning a single value instead of an array)

sharpirate:
@nuriapenya 
>The desktop font should be smaller. Can we try with 13px?

Looks like the stylelint is forcing me to use `rem` units for the font size and the closest it's allowing me to do is `0.75rem` which is `12px`, would that work?

packages/components/src/circular-progress-bar/README.md
```diff
@@ -0,0 +1,22 @@
+# Circular Progress Bar
+
+This component is used to display a circular progress bar as part of the cohesive Launchpad Experience
+See https://cylonp2.wordpress.com/2022/12/14/a-cohesive-launchpad-experience/
+It displays a gray circle and another blue circle on top of it showing the progress.
+It renders text in the center in the center: X/Y where X is the current step and Y is the total number of steps
+The currentStep and numberOfSteps props are used to calculate represent the current progress as a fraction
+
```
Addison-Stavlo: I wonder if we should name this something like "LaunchPad Circular Progress Bar", or put a big warning that this is intended for launchpad checklists (in their various places) and not necessarily for other unrelated progress bar solutions. I say that since our launchpad/checklist icons used for distinguishing and returning will all use the similar distinguishable circular progress icon, and if other interfaces start using the circular progress bar format as well that could get confusing.

I see you already noted "This component is used to display a circular progress bar as part of the cohesive Launchpad Experience", but maybe good to add that its not intended for non-launchpad related interfaces.

Addison-Stavlo:
This looks great and works well in my testing.

Other than whats noted above, I just wanted to say I think this is a really cool progress bar. Kudos to both Nuria and Bobby here!

PR #71371, opened 2022-12-20
Add/ Feature flag for Email Stats

JuanLucha:
#### Proposed Changes

* This PR just adds the feature flag for the email stats feature.


TimBroddin (approved):
The flag is disabled in all environments, right? In that case, we'll use the feature flag 👍

PR #71379, opened 2022-12-20
stylelintrc: Allow `%` for `border-radius`

danielbachhuber:
From https://github.com/Automattic/wp-calypso/pull/71355#discussion_r1053348700

## Proposed Changes

Adds `%` to allowed list of units for `border-radius`. Percentages are really helpful for creating rounded elements.

## Testing Instructions

1. Verify nothing breaks.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -28,7 +28,7 @@
 			{
 				"/^font-size$|^font$/": [ "rem" ],
 				"line-height": [ "px" ],
-				"/radius$/": [ "px" ]
+				"/radius$/": [ "px", "%" ]
 			},
 			{ "ignore": [ "inside-function" ] }
 		],
--- a/apps/editing-toolkit/editing-toolkit-plugin/event-countdown-block/blocks/src/style.scss
+++ b/apps/editing-toolkit/editing-toolkit-plugin/event-countdown-block/blocks/src/style.scss
@@ -117,7 +117,6 @@ $ecf_color6: #b82806;
 		top: 0;
 		width: 4px;
 		height: 4px;
-		/* stylelint-disable-next-line declaration-property-unit-allowed-list */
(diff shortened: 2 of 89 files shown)


simison:
Probably not possible to allow `50%` specifically? That's really the only value that should be allowed, but `%` works if that's not possible.

tyxla (approved):
Looks good! A percentage for border-radius is perfectly valid, especially for the case when we want to create a circle by providing `border-radius: 50%`.

tyxla:
> Probably not possible to allow `50%` specifically? That's really the only value that should be allowed, but `%` works if that's not possible.

I don't see why we should be that strict. An ellipse can be achieved with a different percentage value, and I don't see why we'd want to prevent that.

danielbachhuber:
Currently on `trunk`:

```
$ ag 'stylelint-disable-line declaration-property-unit-allowed-list' --stats-only
105 matches
60 files contained matches
$ ag 'stylelint-disable-next-line declaration-property-unit-allowed-list' --stats-only
416 matches
213 files contained matches
```

It looks like there are 521 total disables.

With the changes in this pull request:

```
$ ag 'stylelint-disable-line declaration-property-unit-allowed-list' --stats-only
70 matches
37 files contained matches
$ ag 'stylelint-disable-next-line declaration-property-unit-allowed-list' --stats-only
337 matches
166 files contained matches
```

It looks like this brings us down to 407 disables, or a ~22% reduction in use of `stylelint-disable-next-line` and `stylelint-disable-line`.

tyxla:
Thank you @danielbachhuber! Successes like this are also great p2 material FWIW - no pressure 😅

danielbachhuber:
@tyxla Good call! Posted with p4TIVU-ama-p2

PR #71380, opened 2022-12-20
Allow self-service Atomic site deletion.

### LINKED ISSUE #71233: Unable to Delete a Site Due to Active Subscription
### Quick summary

Atomic site owners who wish to delete their site see a message to contact support, instead of being redirected to Purchases to remove their subscriptions. We should guide customers to remove their subscriptions first, so they can delete their Atomic sites without HE intervention.

### Steps to reproduce

On an atomic site with an active Business/ eCommerce / Pro plan:

1. Go to Settings → General 
2. Scroll down to the 'delete your site permanently' option
3. Observe the message to contact the support team to delete the site.

I expected to be able to delete my site without contacting support.

### What actually happened

I see this message:
image

### Context

The Close Account flow was addressed by @mpkelly in #57399. The Delete Site flow could be addressed the same way.

Encountered with a customer in #36077448-hc

Atomic

### Theme-specific issue?

_No response_

### Browser, operating system and other notes

_No response_

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

Yes, difficult to implement

### Workaround details

Users must contact support to delete their site. We can lessen the burden on support by guiding users to the self-serve flow that already exists.

mpkelly:
Closes https://github.com/Automattic/wp-calypso/issues/71233

#### Proposed Changes

* You can now delete an Atomic site after removing any plans

#### Testing Instructions

* Create a new site with a monthly Business plan
* Note the site ID and use that to make a WoA site 
* Install any plugin to trigger migration to Atomic infrastructure
* Try to delete the site at `/settings/general/:site`, and you should be told to remove the plan
* Remove the plan
* Return back to `/settings/general/:site` immediately and try again to delete the site
* Confirm that the delete button is disabled, and you're shown the warning as shown below
* After a minute, return to the page, and it should delete the site

image

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/state/sites/actions.js
+++ b/client/state/sites/actions.js
@@ -231,6 +231,20 @@ export function deleteSite( siteId ) {
 					return;
 				}
 
+				if ( error.error === 'unauthorized' ) {
+					// this error is used when you try to delete an Atomic site
+					// right after removing the plan(s) from it. Normally the
+					// message will be 'User cannot delete jetpack site via API.'
+					// The user needs to wait a minute before self-service deletion is available.
+					dispatch(
+						errorNotice(
+							'Unable to delete site. If you recently removed the plan(s) then please wait a couple of minutes and retry.',
+							siteDeletionNoticeOptions
+						)
+					);
+					return;
+				}
+
 				dispatch( errorNotice( error.message, siteDeletionNoticeOptions ) );
 			} );
 	};
--- a/client/state/sites/actions.js
+++ b/client/state/sites/actions.js
@@ -230,21 +230,6 @@ export function deleteSite( siteId ) {
 					);
 					return;
 				}
-
-				if ( error.error === 'unauthorized' ) {
-					// this error is used when you try to delete an Atomic site
-					// right after removing the plan(s) from it. Normally the
-					// message will be 'User cannot delete jetpack site via API.'
-					// The user needs to wait a minute before self-service deletion is available.
-					dispatch(
-						errorNotice(
-							'Unable to delete site. If you recently removed the plan(s) then please wait a couple of minutes and retry.',
-							siteDeletionNoticeOptions
-						)
-					);
-					return;
-				}
-
 				dispatch( errorNotice( error.message, siteDeletionNoticeOptions ) );
 			} );
 	};
--- a/client/my-sites/site-settings/delete-site/index.jsx
+++ b/client/my-sites/site-settings/delete-site/index.jsx
@@ -17,7 +17,6 @@ import Notice from 'calypso/components/notice';
 import DeleteSiteWarningDialog from 'calypso/my-sites/site-settings/delete-site-warning-dialog';
 import { hasLoadedSitePurchasesFromServer } from 'calypso/state/purchases/selectors';
 import hasCancelableSitePurchases from 'calypso/state/selectors/has-cancelable-site-purchases';
-import isSiteAutomatedTransfer from 'calypso/state/selectors/is-site-automated-transfer';
 import { deleteSite } from 'calypso/state/sites/actions';
 import { getSite, getSiteDomain } from 'calypso/state/sites/selectors';
 import { setSelectedSiteId } from 'calypso/state/ui/actions';
@@ -121,7 +120,7 @@ class DeleteSite extends Component {
 	};
 
 	render() {
-		const { isAtomic, siteDomain, siteId, siteSlug, translate } = this.props;
+		const { siteDomain, siteId, siteSlug, translate } = this.props;
 		const exportLink = '/export/' + siteSlug;
 		const deleteDisabled =
 			typeof this.state.confirmDomain !== 'string' ||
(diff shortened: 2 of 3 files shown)


mpkelly:
@donalirl you can give this a test if you want using the Calypso Live link. After removing the plan, you need to wait a minute before you can delete the site - I added a notice for this which tells the user to retry. It's not the best UX for them but hopefully they shouldn't need support.

client/state/sites/actions.js
```diff
@@ -231,6 +231,20 @@ export function deleteSite( siteId ) {
 					return;
 				}
 
+				if ( error.error === 'unauthorized' ) {
+					// this error is used when you try to delete an Atomic site
+					// right after removing the plan(s) from it. Normally the
+					// message will be 'User cannot delete jetpack site via API.'
+					// The user needs to wait a minute before self-service deletion is available.
```
zaguiini: How could we prevent this from happening? Maybe after removing the plans, we could force-sync the permissions or something?

zaguiini (commented):
Great job!

As a follow-up, I wonder if we could ask the user for more information on why they are deleting the site. Do you think it'd be valuable? Why would someone delete a site in the first place? (Nothing to act now, I'd just like to understand it better)

client/state/sites/actions.js
vykes-mac: is there a way to determine if delete is ready? we could disable the button and add some info about it being ready in 5 mins or so.

mpkelly: @zaguiini We have a feedback flow for removing the plan, which might capture the same feedback. If they remove the plan just before removing the site, it could be a bit much to ask for two lots. 

@zaguiini @vykes-mac, I agree this is not optimal. From memory, I think there is an async job that runs when you remove an Atomic plan, and this needs some time to complete. I didn't check if it sets some flag somewhere to say "in progress" which we could expose via the API and use on the frontend. That could be a nice follow-up PR.

wojtekn: @mpkelly I tested this flow to see what differences are in the site's object in those three states: site before removal, site after removing the plan, and site eligible for removal after the async job runs, but I've got stuck. I waited 15 minutes, and I'm still getting the "Unable to delete site. If you recently removed the plan(s) then please wait a couple of minutes and retry." error message. Should this process work for a site which is still Atomic?

mpkelly: @wojtekn, thanks for testing this. 

>Should this process work for a site which is still Atomic?

On trying to delete a site that is still Atomic, i.e. a site with a Business plan, it should tell you to remove the plan. This was the existing logic. After removing the plan, if you retry deleting the site, it should call the API to try and do that. Depending on how long you waited, it should either be successful or show you the notice that I added. 

What is the current state of the site you were trying to delete? Does it still have an Atomic domain with "staging" in it? Does it still have plans shown? It sounds like the async job might have not been completed.

wojtekn: Thanks @mpkelly . I could test this flow now by triggering plan removal from Calypso.

I think I see a cleaner way of warning the user that their site is not ready for removal yet. Now, when they click the "Delete site" button for the site that still has paid upgrade, we display this message in popup:

Screen Shot 2022-12-22 at 10 32 33

When they remove the paid upgrades, the site still passes the `isAtomic` checks, but has a `free_plan` set. For that case, we could introduce another popup saying something like _"Your site is not ready for removal yet - we are currently reverting it. If you recently removed the plan(s) then please wait a couple of minutes before you try removing the site."_

That way, we would show the final confirmation popup with the "Confirm delete site" heading and the input field only if the site is really applicable for removal.

mpkelly: Thanks, @wojtekn. That sounds like an improvement. I will investigate and update the PR.

sejas: I've tested this flow twice, and it's looking good. It will save many support interactions.
The first time I didn't need to wait, and I deleted the site using the wpcomstaging subdomain. Very smooth process, great job 👍 .


https://user-images.githubusercontent.com/779993/209147840-dff7551c-bd64-41ce-9de2-9ce2063803e0.mp4


The second time I waited a bit longer, I got an error loading the dashboard due to the subdomain migration from wpcomstaging.com to wordpress.com. This PR does not produce this issue, but it is something to consider.

Screenshot 2022-12-22 at 13 47 33


My only suggestion that we can address in a separate PR is adding a link to the site settings in the Removed plan notice.
It will be easier for the user to return to the settings and continue deleting the site with fewer clicks.
Screenshot 2022-12-22 at 13 22 12

mpkelly:
I reworked this solution based on @wojtekn's feedback. A warning is shown now, and the user is told to check back. This is an improvement over the API call and error notice, I think. 

image

zaguiini:
@mpkelly what do you think of delivering an optimistic message to the user, then scheduling a delete job whenever it's possible?

Like: once the user removes the plan, immediately make the delete site button available, without the red copy. Then, let them click the delete button with a message saying "this action might take a few minutes to complete", letting the user indicate the intention of deleting the site.

On the back-end, we'll have this information and we might schedule a callback for when the plan is removed and the site is ready to be deleted, scheduling a delete job ourselves with the aval from the user. How does that sound?

client/state/sites/actions.js
wojtekn: Thanks, @mpkelly. It's better now, as we inform the user upfront instead of relying on the API error.

One thing still bothers me - we block delete action in two cases: when the site still has a paid upgrade and when it is not eligible for removal because it's still Atomic. In both cases, we use different ways of displaying the message: in the first, we use the popup. In the second, we display that new text above the button, and we disable the button. Would it be better to use the same way in both cases?

mpkelly:
@zaguiini, that would be a better solution. Ideally, the user shouldn't have to come back, and we'd accept their request to delete the site right away. But it sounds more involved, too. I think it's a different PR. In this PR, I aimed to get self-service deletion working in the simplest way possible. We can iterate to a better solution if we feel it's worth the effort.

mpkelly:
@donalirl, how do you feel about this PR in its current form?

zaguiini:
>Ideally, the user shouldn't have to come back, and we'd accept their request to delete the site right away.

Created https://github.com/Automattic/dotcom-forge/issues/1428.

zaguiini (approved):
Works for me 👍 

Design-wise, I think that adding the notice right below the button would be a better option. That'd be a little bit more inline with our design system, where we display explanations before the action button. Non-blocking, though!

mpkelly:
I just went ahead with this as it was. We still add value and can do a follow-up PR later if we think it's necessary.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7696523

Thank you @mpkelly for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7696523).

Issue #71368, opened 2022-12-20
[Bug]: Theme Showcase: The "My Themes" filter sometimes can't be clicked


fushar:
### Quick summary

The filter can't be clicked when the current filter is not "All".

@Automattic/lego 

### Steps to reproduce

1. Create a site with Business plan
2. Open `/themes/<siteSlug>`
3. Click on a filter other than "All" (e.g. "Blog")
4. Now click on "My Themes" filter

### What you expected to happen

The My Themes filter is selected

### What actually happened

The My Themes filter is NOT selected

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Click on the My Themes filter twice

arthur791004:
Close it via https://github.com/Automattic/wp-calypso/pull/71381

PR #71381, opened 2022-12-20
Theme Showcase: Fix My Themes tab might not able to be clicked

arthur791004:
#### Proposed Changes

* When you click "My Themes" filter from another filter except "All", the URL will be changed and theme showcase component will try to resolve the current filter from the URL. However, the resolved filter will be "All" as we don't have any information in the URL to indicate the current filter is "My Themes". As a result, this PR avoids resolving the filter from the URL if the current filter is the static filter.

#### Testing Instructions

* Open /themes/<your_ecommerce_site>
* Click on a filter other than "All" (e.g. "Blog")
* Now click on the "My Themes" filter and verify it works!

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/71368

### DIFF
--- a/client/my-sites/themes/theme-showcase.jsx
+++ b/client/my-sites/themes/theme-showcase.jsx
@@ -58,6 +58,9 @@ const optionShape = PropTypes.shape( {
 	action: PropTypes.func,
 } );
 
+const findTabFilter = ( tabFilters, filterKey, defaultValue ) =>
+	Object.values( tabFilters ).find( ( filter ) => filter.key === filterKey ) || defaultValue;
+
 class ThemeShowcase extends Component {
 	constructor( props ) {
 		super( props );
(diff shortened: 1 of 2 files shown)


client/my-sites/themes/theme-showcase.jsx
```diff
@@ -343,10 +338,11 @@ class ThemeShowcase extends Component {
 				.filter( ( key ) => ! subjectFilters.includes( key ) )
 				.join( '+' );
 
-			const newFilter =
-				tabFilter.key !== this.staticFilters.ALL.key
-					? [ filterWithoutSubjects, subjectTerm ].join( '+' )
-					: filterWithoutSubjects;
+			const newFilter = ! [ this.staticFilters.ALL.key, this.staticFilters.MYTHEMES.key ].includes(
```
fushar: What if we replace this with something like `this.staticFilters.includes(tabFilter)` ? I find that the intention is clearer this way.

(I think we can do it because for static filters we always return the same object, CMIIW)

client/my-sites/themes/theme-showcase.jsx
```diff
@@ -58,6 +58,9 @@ const optionShape = PropTypes.shape( {
 	action: PropTypes.func,
 } );
 
+const findTabFilter = ( tabFilters, filterKey, defaultValue ) =>
```
fushar: `defaultValue` is always the same (`ALL`). Can we for now put this as member function inside `ThemeShowcase` class (above `getTabFilterFromUrl`)? So that we can just hardcode `this.staticFilters.ALL` there.

client/my-sites/themes/theme-showcase.jsx
arthur791004: Done. I made a `isStaticFilter` member function to check whether a filter is a static filter!

client/my-sites/themes/theme-showcase.jsx
arthur791004: Done.

fushar (approved):
Works nicely!

arthur791004:
@fushar I found there is an issue below and the reason is the page is not reset, so I committed a change to reset the page once the site id is changed.

* When you are in "My Themes" filter and switch the site to another eCommerce site or self-hosted site, it will show themes are not found.

fushar:
@arthur791004 Ah, nice, thanks! I've just tested it and it works well :)

fushar (approved):

Issue #46205, opened 2020-10-06
Checkout: add identifier for USD in non-US geos


michaeldcain:
We already include non-USD identifiers for supported currencies that use the $ as a currency symbol (i.e. "C$" for Canadian dollars), but we don't include a USD identifier when a user is in a location that uses the $ and we don't currently support their currency (i.e. Argentinian Pesos).

When a user is outside of the US, we should add an idicator to the amount (i.e. "US$").

DavidRothstein:
See also c/8yG0JNLf-tr (internal link)

abemos:
Hi there, just wanted to follow up and ask about the ETA for this request? @michaeldcain  @DavidRothstein . Thanks!

kylemcph:
Just noting that there is still confusion about this. Another example is #25056033-hc.

jartes:
I also had another example of confusion here 23979794-hc about Argentinian Pesos.

michaeldcain:
Work in progress in D52530-code

sirbrillig:
Requested changes on the diff.

carladoria:
@michaeldcain  or @sirbrillig  are there any updates on this? we still get several requests related to this per day. 
I personally get around 2 per day.

michaeldcain:
@carladoria sorry for the delayed response. The issue should now be fixed in the popover shopping cart and checkout, but additional work will be required to update plans lists and subscription management screens. I'll open another issue to track those changes.

carladoria:
Great, thanks a lot!

marianearchimbaud:
Hi there!! I am not sure if this is the exact same issue, but I wanted to know if it was possible to add this request:
In the pricing page, we find this: https://wordpress.com/pricing/

And some uses are confused about these symbol/currency, as it also means "pesos".
Are we able to be a bit more specif with the currencies displayed here?
As this example? https://d.pr/i/HIrxW9
This issue was reported by a user on this chat, they were confused and asked for a refund as they were not expecting the charge was going to be made in US$

hc-1569496

sirbrillig:
FWIW I _think_ the plans page is managed by @Automattic/martech-decepticons so perhaps they can weigh in here as well. 

> but additional work will be required to update plans lists and subscription management screens. I'll open another issue to track those changes.

@michaeldcain did we ever create a new issue to track the plans page?

michaeldcain:
[...]

PR #71387, opened 2022-12-20
format-currency: Change formatCurrency to use locale

sirbrillig:
#### Proposed Changes

In calypso when we need to format prices for display, we often use the `@automattic/format-currency` library. The library contains a hard-coded list of currencies and the data needed to format them appropriately for display. It exports a `formatCurrency()` function to perform this formatting. However, the data this system uses is entirely based on the currency code, and does not take into account the user's locale. Many currencies are formatted differently depending on the locale, which makes this system inaccurate.

For example, sometimes EUR 1000 is formatted `€1,000.00` but other times it is formatted `1.000,00 €`.

In this PR we modify `formatCurrency()` and its sibling `getCurrencyObject()` to ignore the hard-coded list of data and instead use the browser's built-in library that formats currency for locales: [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat). It retrieves the locale from the browser (although it can also be passed-in as an option if needed). 

Before this PR. The variant picker prices are formatted by `formatCurrency` and the line item price is formatted by the backend. Note that in the IDR example the thousands group separators of the variant picker do not match those of the line item.

before-en_US-IDR

before-de_DE-EUR

After this PR. Note that the thousands separators now are correct in both the frontend and backend versions of the price.

Screenshot 2023-01-05 at 1 11 51 PM

Screenshot 2022-12-28 at 5 09 34 PM

**NOTE:** a possible improvement would be to pass the locale from WordPress to the library instead of relying on the browser. This PR adds a `setDefaultLocale()` function which could be used to consume the `getCurrentLocaleSlug()` Redux selector, but I'm not sure where that side effect would best be placed.

**NOTE:** some of the official currency symbols returned by using `Intl.NumberFormat` might be confusing for users in their own locale. For example, if a user has their locale set to `en_CA` but their currency set to `CAD`, they would see prices displayed like `$100` instead of `C$100` as they do before this PR and that could be misinterpreted as USD (although if their locale is `en_US` they would see `CA$100`). To make this change less disruptive, the PR retains the hard-coded list of currency _symbols_ and uses those as overrides for each currency, _regardless of locale_ (except for USD because of https://github.com/Automattic/wp-calypso/issues/46205).

This PR removes the hard-coded formatting data from the package, and the unnecessary `CURRENCIES` and `getCurrencyDefaults()` exports. It also removes the `decimal`, `grouping`, and `symbol` options from `formatCurrency()` and `getCurrencyObject()` because those are now totally under the control of the formatter.

Part of https://github.com/Automattic/payments-shilling/issues/1300

#### Testing Instructions

Automated tests should mostly cover this, but it's worth viewing a few significant places in calypso that display currencies to make sure there are no obvious problems. Please do this with different currencies and different locales.

- The plans page.
- Checkout.

As for the changes in the API, you can verify that there are no uses of the removed options or exports with the following commands. Make sure that you check any results to be sure they are not using the removed options or exports:

- `rg --context 10 -w formatCurrency |rg -w 'grouping|decimal|symbol'`
- `rg --context 10 -w getCurrencyObject |rg -w 'grouping|decimal|symbol'`
- `rg -w CURRENCIES`
- `rg -w getCurrencyDefaults`


sirbrillig:
After some discussion in https://github.com/Automattic/payments-shilling/issues/1300, I retained the currency symbol configuration which will be used for every currency (except for USD which I removed because of https://github.com/Automattic/wp-calypso/issues/46205), regardless of locale (but will still use the locale for positioning, grouping, and decimals). This should make this PR less disruptive to the current UI and protect against symbol ambiguity. The hard-coded list could be reduced over time if we want to rely more on the browser.

dlind1 (commented):
Potentially found a few formatting inconsistencies:

1. Browser locale set to `he`, user currency set to IDR, calypso.localhost
Screenshot 2023-01-02 at 02 29 41

2. Minus sign is displayed incorrectly for RTL UI, but that seems to be a back-end issue
Screenshot 2023-01-02 at 01 59 14
Screenshot 2023-01-02 at 01 49 51

sirbrillig:
Thank you, @dlind1!

> Browser locale set to he, user currency set to IDR, calypso.localhost [image shows currency symbol as "IDR" instead of "Rp"].

Interesting! I guess the currency symbol override isn't always working, which is a big deal, but I'm having trouble reproducing this. If I use the Chrome Sensors panel to set my locale to `he` and have `IDR` as my currency, I see this:

Screenshot 2023-01-02 at 12 56 19 PM

But clearly you were able to get a different display. A few things I can think of: is it possible you had a checkout of this branch before [3ea1c2e](https://github.com/Automattic/wp-calypso/pull/71387/commits/3ea1c2e478cfdbed60b9c7e7cb3d92b1a5b3ba3d)? Or are you using a different browser (I wasn't able to test this in Firefox because I can't find a way to change the locale yet so that could be related)? Maybe there's something else I'm not thinking of... 

> Minus sign is displayed incorrectly for RTL UI, but that seems to be a back-end issue

Could you check to see if that's fixed by https://github.com/Automattic/wp-calypso/pull/71537?

dlind1:
> But clearly you were able to get a different display. A few things I can think of: is it possible you had a checkout of this branch before [3ea1c2e](https://github.com/Automattic/wp-calypso/pull/71387/commits/3ea1c2e478cfdbed60b9c7e7cb3d92b1a5b3ba3d)? Or are you using a different browser (I wasn't able to test this in Firefox because I can't find a way to change the locale yet so that could be related)? Maybe there's something else I'm not thinking of...

That was it, my locale branch didn't have that commit. The dropdown does display the correct symbol now.

> > Minus sign is displayed incorrectly for RTL UI, but that seems to be a back-end issue
> 
> Could you check to see if that's fixed by #71537?

Just  checked, and unfortunately it's not fixed there.
Screenshot 2023-01-03 at 00 45 05

Screenshot 2023-01-03 at 00 48 59

sirbrillig:
> Minus sign is displayed incorrectly for RTL UI, but that seems to be a back-end issue

Thanks for double-checking that! I am far from an expert on RTL, but it seems like this might actually be a browser issue?

If I display a negative price, then change `document.dir` to `'rtl'`, the negation sign moves from the left to the right side on its own. (I think this is how we do RTL localization on WordPress.com.)

Screenshot 2023-01-03 at 2 34 28 PM

The browser's price formatting itself seems to only be LTR.

```
const defaultFormatter = new Intl.NumberFormat( 'he', { style: 'currency', currency: 'ILS' } );
defaultFormatter.format(-100103.53)
'‏‎-100,103.53 ₪'
```

In any case, this appears to be an existing issue and is not made any worse by this PR.

dlind1:
> Thanks for double-checking that! I am far from an expert on RTL, but it seems like this might actually be a browser issue?

This works correctly from the BIDI algorithm point of view since the minus sign has weak directionality and with base RTL direction it'd be incorrectly put to the right of the numbers. The easiest way to fix it would be to explicitly wrap the minus sign and all the numbers in an LTR run, like `<bdo dir="ltr">-140.00</bdo> €`, or use some other approach from https://www.w3.org/International/questions/qa-bidi-unicode-controls.

Since it's an existing issue, it shouldn't prevent this PR from being merged, but it'd be nice to fix that at some point.

sirbrillig:
I've learned so much. Thanks! I created https://github.com/Automattic/wp-calypso/issues/71657 to track that issue.

michaeldcain:
I'm not sure which is correct, but with my Chrome location set to Mumbai (`mr-IN`) and INR as my currency, I'm seeing differences in the /plans price display and the shopping cart:

| Production | PR |
| ----------- | ----------- |
| Screenshot 2023-01-05 at 11 35 17 AM | Screenshot 2023-01-05 at 11 34 28 AM |

I think this could be related to my WordPress locale not matching my browser locale, but the difference between the cart, /plans, and /purchases views is still strange.

sirbrillig:
Thanks for noticing that, @michaeldcain. The issue was that this PR improves the client-side formatting to a point that it becomes better than the server-side formatting, so any parts of the UI that use server-side formatting suddenly will seem wrong. The one you saw (the mini-cart) isn't the only place. 

I decided to refactor https://github.com/Automattic/wp-calypso/pull/71722 so it could be merged before this PR and rebased this work on top of it. That way this PR should now change all price formatting equally.

sirbrillig:
Here's how the plans page and the mini-cart look in INR with `mr-IN` now:

Screenshot 2023-01-05 at 1 19 55 PM

The purchases page does not use `formatCurrency`, so its price formatting is going to be different from the mini-cart, but I _think_ that's a relatively minor issue that can be corrected later. I created https://github.com/Automattic/wp-calypso/issues/71723 to handle that.

Screenshot 2023-01-05 at 1 21 44 PM

dlind1:
Should the [Devanagari numbers](https://en.wikipedia.org/wiki/Devanagari_numerals) (or any other non-western numerals) be used here? While correct from a localization standpoint, they would make it more difficult to support and the users are usually familiar with [Western Arabic Numerals](https://en.wikipedia.org/wiki/Arabic_numerals) anyway. I remember that there was a similar issue with numbers in mobile apps using Eastern Arabic Numerals and the decision was to always use the Western ones.

sirbrillig:
Oh! Good question! Those numerals are new to me (and thank you for linking to what they are) so I just assumed they were "right", but I guess we might want to not use them in this case.

The JS module has a `numberingSystem` [option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options) which can be set to `'latn'` to force latin numbers to be used instead of the locale-specific numbers, but for some reason I don't seem to be allowed to use that option according to TypeScript. I'm asking about this in Slack but in the meantime I've updated the library to use [a Unicode extension](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/numberingSystem#adding_a_numbering_system_via_the_locale_string) to force latin numerals. (Update: it looks like this is [a TypeScript bug](https://github.com/microsoft/TypeScript/issues/52072).)

dlind1 (approved):
LGTM!

DavidRothstein:
>NOTE: a possible improvement would be to pass the locale from WordPress to the library instead of relying on the browser.

That seems somewhat important, right?  Isn't almost everything else in the user interface based on the WordPress account's language rather than the browser language?

To take an example, if I use the IDR currency and then switch my WordPress account language from English to Indonesian, prices on the purchase management page (which still come from the server, if I understand correctly) switch from `536,400` to `536.400` (comma to period) and that's the correct behavior.  But prices in checkout don't switch, even though it seems like they should.  They don't switch without this pull request either; without this pull request they're always `536.400` and with this pull request they're always `536,400` (in my web browser) -- although neither behavior is correct, the first seems slightly better since it matches the common usage of the currency.

I'm not sure if this is extremely important (it depends on how often people actually have the correct language configured in their browser) but for consistency with everything else I think prioritizing the WordPress account language would be good.

Overall this pull request looks like it's going to be a great change, though!

sirbrillig:
> That seems somewhat important, right? Isn't almost everything else in the user interface based on the WordPress account's language rather than the browser language?

I agree. I think it should be done, and I think I know how to do it, but I don't know _where_ to do it. It's a very high-level side effect so probably should live somewhere high up the render chain, but we don't have a single top-level component. Calypso is built with many entry points.

My thought was that a side effect like that probably deserves its own PR so its details could be debated separately from this (already fairly large) change.

Two other things have kept me from trying to implement that effect in this change: 

- I think that in practice most people's browser locale probably will match their WordPress locale and neither is likely to change often.
- I'm working on migrating all the display prices I can to use `formatCurrency` so that there will be less places which risk a mismatched format. All of checkout should be using it since https://github.com/Automattic/wp-calypso/pull/71722 and I'm working on making the purchases pages use it starting with https://github.com/Automattic/wp-calypso/pull/71737 and https://github.com/Automattic/payments-shilling/issues/1322.

DavidRothstein:
>I think that in practice most people's browser locale probably will match their WordPress locale

I'm not sure if that's true -- what I can gather from an internet search is that there are a whole lot of people out there whose browsers are pre-configured to English (even though that's not their native language) and most people don't know how to change that setting.  Whereas if I start at https://wordpress.com/fr/ (for example) and create a new WordPress.com account, the account winds up with French as its language regardless of browser locale.

That said, I agree with you that waiting until a followup pull request to implement that is fine -- this pull request seems to overall make things better (especially in light of pNPgK-6iv-p2).  I haven't reviewed the code really, but I didn't notice any problems while doing some brief testing, and I have no objection to this being merged without the WordPress.com account language issue being addressed.

sirbrillig:
Rebased over https://github.com/Automattic/wp-calypso/pull/71804

sirbrillig:
Rebased over https://github.com/Automattic/wp-calypso/pull/71737 which added some more tests that this changes.

sirbrillig:
I created https://github.com/Automattic/wp-calypso/issues/71816 to track setting the currency locale from the user's locale.

daledupreez:
@sirbrillig, how does this functionality behave for cases where we're running on the server like Server Side Rendering?

daledupreez:
I am also unclear on how this is expected to behave for `USD` if I am in a country that uses `$` as the currency, but I have `en` or `en-US` set up for my browser locale.

sirbrillig:
> how does this functionality behave for cases where we're running on the server like Server Side Rendering?

@daledupreez I don't know; I've never actually interacted with the parts of calypso that use SSR. But to guess: if there's no browser locale available (from the `window.navigator` API) and there's no WordPress UI locale available (from Redux), it [will default](https://github.com/Automattic/wp-calypso/blob/5b559da005378cccfd1b79e052c559a4f2c7b9bc/packages/format-currency/src/index.ts#L8) to `en` as the locale. That [could be customized](https://github.com/Automattic/wp-calypso/blob/5b559da005378cccfd1b79e052c559a4f2c7b9bc/packages/format-currency/src/index.ts#L17) by calling `setDefaultLocale()` if there's some other way to discover the locale during SSR.

> I am also unclear on how this is expected to behave for USD if I am in a country that uses $ as the currency, but I have en or en-US set up for my browser locale.

If you have CAD and you're in, say, Canada (which also uses `$`) and your browser is set to `en-US`, but your calypso UI interface language is set to `en-CA`, you'll see prices rendered as `C$`. That's for two reasons: one is that calypso basically ignores the browser locale in favor of the UI interface locale since https://github.com/Automattic/wp-calypso/pull/71895, and also because `formatCurrency()` always overrides the localized currency symbol with hard-coded ones in this list:

https://github.com/Automattic/wp-calypso/blob/5b559da005378cccfd1b79e052c559a4f2c7b9bc/packages/format-currency/src/currencies.ts#L84-L86

If you're in the same situation but using USD, you'll see `US $` because we do still localize the currency symbol for USD.

To quote the PR description above:

> NOTE: some of the official currency symbols returned by using Intl.NumberFormat might be confusing for users in their own locale. For example, if a user has their locale set to en_CA but their currency set to CAD, they would see prices displayed like $100 instead of C$100 as they do before this PR and that could be misinterpreted as USD (although if their locale is en_US they would see CA$100). To make this change less disruptive, the PR retains the hard-coded list of currency symbols and uses those as overrides for each currency, regardless of locale (except for USD because of https://github.com/Automattic/wp-calypso/issues/46205).

daledupreez:
Thanks for clarifying, @sirbrillig!

I think my biggest concern is that the Calypso UI only offers `en` and `en-gb` as language options, so the "what happens when I am in Argentina with my currency set to USD and my language set to English?" is relevant, especially because [this code](https://github.com/Automattic/wp-calypso/blob/ff249a3840e2697b3095e4453b6f6e53d5209206/client/state/ui/language/actions.ts#L14) in the `setLocale()` function sets the default to the Calypso language.

That particular example is important, because I am worried that with `en` as the locale, we'd show USD prices using just `$`, which would be ambiguous with prices in ARS that would also use `$`.

sirbrillig:
That's a good concern! It's the same one that https://github.com/Automattic/wp-calypso/issues/46205 and D52530-code were intended to solve, but it's possible that this has reintroduced the issue since D52530-code relied on IP address. I created https://github.com/Automattic/wp-calypso/issues/73496 to see what we can do.

PR #71403, opened 2022-12-21
Add/ Skeleton for Email Stats page

JuanLucha:
#### Proposed Changes

This PR adds the skeleton and the route for the Email Stats page. It's empty at the moment, but it's hidden behind the `newsletter/stats` feature flag.

#### Testing Instructions

1. Checkout this branch on your local repository and start the application.
2. Go to a route in this format `calypso.localhost:3000/stats/email/open/<SITE_DOMAIN>/day/<POST_ID>?flags=newsletter/stats` (for example: `calypso.localhost:3000/stats/email/open/testingrightnow.blog/day/37?flags=newsletter/stats`).
3. You should see the placeholder text on the screen like this:

Screenshot 2022-12-21 at 12 09 02

### DIFF
--- a/client/my-sites/stats/controller.jsx
+++ b/client/my-sites/stats/controller.jsx
@@ -11,6 +11,7 @@ import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import StatsCommentFollows from './comment-follows';
 import StatsOverview from './overview';
 import StatsSite from './site';
+import StatsEmailOpenDetail from './stats-email-open-detail';
 import StatsInsights from './stats-insights';
 import StatsPostDetail from './stats-post-detail';
 import StatsSummary from './summary';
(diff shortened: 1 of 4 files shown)


client/my-sites/stats/controller.jsx
```diff
@@ -437,3 +438,9 @@ export function wordAds( context, next ) {
 
 	next();
 }
+
+export function emailOpen( context, next ) {
```
TimBroddin, line 442: I see other routes doing 

`	if ( 0 === siteId ) {
		window.location = '/stats';
		return next();
	}
`
Is this something that can be added here as well? Or is this something you plan on adding later?

TimBroddin (approved):
LGTM! 🚢

I added one small comment, but this is something than can be added later. (Or you can add it now and I'll reapprove.)

client/my-sites/stats/controller.jsx
JuanLucha: Indeed, that will be there when the siteId is required by the component. Thanks for the review! 🏅

PR #71405, opened 2022-12-21
HappyChat: Add New Year's closure times

renancarvalho:
#### Proposed Changes

Add message for New Year's holiday limited availability.
More context [here](https://github.com/Automattic/wp-calypso/pull/p1671504597022099-slack-C029S5LLB7D)

#### Testing Instructions

- Pull this branch or use Calypso's live link
- Open HappyChat and and click on "still need help button "
- You should see the December dates message
- Change your computer time to a date after December 25th 
- You should see the NY's  message like bellow

image

### DIFF
--- a/packages/help-center/src/components/help-center-contact-page.tsx
+++ b/packages/help-center/src/components/help-center-contact-page.tsx
@@ -116,9 +116,9 @@ export const HelpCenterContactPage: React.FC = () => {
 				<h3>{ __( 'Contact our WordPress.com experts', __i18n_text_domain__ ) }</h3>
 				<HelpCenterActiveTicketNotice tickets={ tickets } />
 				<GMClosureNotice
-					displayAt="2022-12-17 00:00Z"
-					closesAt="2022-12-24 00:00Z"
-					reopensAt="2022-12-26 07:00Z"
+					displayAt="2022-12-26 07:00Z"
+					closesAt="2022-12-31 00:00Z"
+					reopensAt="2023-01-02 07:00Z"
 					enabled={ hasAccessToLivechat }
 				/>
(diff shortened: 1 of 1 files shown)


escapemanuele:
Hey @renancarvalho, a couple of considerations!

What about we add another `GMClosureNotice` like they are doing [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/me/help/help-contact/index.jsx#L580-L612)? They will show one after the another without overlap and without us having to remember to publish this PR at the right time.

What about, like the link I provided before, we use `ChatHolidayClosureNotice` if it is doable?
It also seems strange to me to use a GM component for a Holiday! 

I think it would be nice to extend the [GMClosureNotice](https://github.com/Automattic/wp-calypso/blob/trunk/packages/components/src/gm-closure-notice/gm-closure-notice.tsx) component that we have, maybe change its name, and have some flag to identify if it is the GM or a Holiday :) 

Let me know what you think!


(those are just different approaches we can have, not asking to do all of those :D )

cc @arcangelini

renancarvalho:
> What about we add another `GMClosureNotice` like they are doing [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/me/help/help-contact/index.jsx#L580-L612)? They will show one after the another without overlap and without us having to remember to publish this PR at the right time.

Thank you for checking @escapemanuele, this is indeed a good idea, have applied this. I think given the urgency of it, this solution is fine for now but of course we are always open for improvements.

escapemanuele (approved):
Thank you for applying the changes! 
If you are sure those 2 do not overlap, I think this is good to go! 👍

PR #71406, opened 2022-12-21
Sensei Tailored Flow

dadish:
#### Proposed Changes

* Adds a Sensei Tailored flow.

p6rkRX-4VM-p2

#### Testing Instructions

* Go to `calypso.localhost:3000/setup/sensei` to kickoff the Sensei Tailored flow.
* Confirm everything works as expected.

You can also test this out with the calypso live link here: https://container-ecstatic-khayyam.calypso.live/setup/sensei

#### Pre-merge Checklist


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7683343

Hi @dadish, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `Make your course site shine with a custom domain. Not sure yet ?`
- `Course Creator`
- `Hosting with the <a>WordPress.com</a>`
- `Course creation and LMS tools powered by SenseiLMS`

Thank you in advance!

dadish:
> - `Make your course site shine with a custom domain. Not sure yet ?`
> - `Course Creator`

Screen Shot 2022-12-22 at 11 37 12

> - `Hosting with the <a>WordPress.com</a>`
> - `Course creation and LMS tools powered by SenseiLMS`

Screen Shot 2022-12-22 at 11 39 05

client/landing/stepper/declarative-flow/internals/sensei.scss
```diff
@@ -0,0 +1,5 @@
+.sensei.launchpad {
+	.badge {
+		height: auto;
+	}
+}
```
simison, line 6: Finding the placement of this file a bit odd in framework's "internal" folder. Not a blocker for merging but something to consider.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/styles.scss
```diff
@@ -0,0 +1,127 @@
+@import "@automattic/onboarding/styles/mixins";
+
+$sensei-color: #43af99;
```
simison: Since this variable gets defined multiple times in multiple places with the same value, did you consider creating a new shared file that you would import?

Not a blocker for merging.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/sensei-step-error/components.ts
```diff
@@ -0,0 +1,38 @@
...
+export const ErrorIcon = styled( UnstyledErrorIcon )`
+	width: 40px;
+	height: 40px;
+	path {
+		fill: var( --studio-orange-40 );
+	}
+`;
+
+export const Title = styled.h1`
+	font-family: 'Recoleta';
```
simison: Recommend including fallback fonts after "Recoleta" in case it doesn't load for some reason, see https://github.com/Automattic/wp-calypso/blob/b74abfa38d482735dfde4270f5312a732c58b783/packages/components/src/styles/_typography.scss#L1-L3

Not a blocker for merging.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/styles.scss
```diff
@@ -0,0 +1,127 @@
...
+
+.sensei .progress-bar__progress {
+	background-color: $sensei-color;
+}
```
simison: Similarly I see some of the component overrides multiple times for each step (like this `.progress-bar__progress` override). There should be a more centralized way to do the same I'm sure.

Not a blocker for merging as this is functional.

packages/data-stores/src/site/types.ts
```diff
@@ -458,6 +458,7 @@ export interface LaunchPadCheckListTasksStatuses {
 	site_launched?: boolean;
 	site_edited?: boolean;
 	video_uploaded?: boolean;
+	publish_first_course: boolean;
```
simison: Might benefit from making it clear this is for Sensei more specifically, either by comment or prefix. Not a biggie though.

packages/onboarding/src/flow-progress/use-flow-progress.ts
```diff
@@ -50,6 +50,12 @@ const flows: Record< string, { [ step: string ]: number } > = {
 		processing: 6,
 		launchpad: 7,
 	},
+	sensei: {
```
simison, line 59: Surprised not to see `user` step here.

packages/onboarding/src/sensei-step-container/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+				<Container>
+					<TitleContainer>
+						<Icon icon={ wordpress } />
+						<Title>{ __( 'Course Creator' ) }</Title>
+					</TitleContainer>
+					{ children }
+					<Footer>
+						<FooterText>
+							{ createInterpolateElement( __( 'Hosting with the <a>WordPress.com</a>' ), {
+								a: <ExternalLink href="https://wordpress.com" target="_blank" />,
```
simison: `ExternalLink` probably adds target=_blank already from the component itself?

packages/onboarding/src/sensei-step-container/styles.scss
```diff
@@ -0,0 +1,27 @@
+.sensei {
```
simison: Note that this style bundle might persist through their session in Calypso. So if you have anything else named "sensei" now or in future, you might have CSS collision. Something more unique like `sensei-onboarding` might make sense.

I've recently noticed this kind of problems with signup form where we used a lot of classes like `.woo`, `.videopress` or `.jetpack`.

Not a blocker!

packages/onboarding/src/sensei-step-container/styles.scss
```diff
@@ -0,0 +1,27 @@
...
+
+		&__progress {
+			border-radius: 0;
+		}
+	}
+
+	* {
+		box-sizing: border-box;
+	}
+}
```
simison: Overall I'd be careful about dramatic spacing modifications to `StepContainer` as these seem fragile; someone changing the `StepContainer` in future for sure won't test Sensei flow.

Worth cloning to different HTML structure then even?

Not a blocker!

packages/onboarding/src/sensei-step-container/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+			stepContent={
+				<Container>
+					<TitleContainer>
+						<Icon icon={ wordpress } />
+						<Title>{ __( 'Course Creator' ) }</Title>
+					</TitleContainer>
+					{ children }
+					<Footer>
+						<FooterText>
+							{ createInterpolateElement( __( 'Hosting with the <a>WordPress.com</a>' ), {
```
dlind1: We've received a comment from a translator that `Hosting with the WordPress.com` sounds awkward. Should it be `Hosted by WordPress.com` or `Hosting with WordPress.com`?

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7683343

Hi @dadish, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include this string: `Hosted by <a>WordPress.com</a>`

Thank you in advance!

client/landing/stepper/declarative-flow/internals/sensei.scss
gikaragia: Moved this to launchpad directory in 280af0fde890863542d855924e60ef3ea6c353a3

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/styles.scss
gikaragia: Moved to a common scss file in 2b105cd55e551c7b35b817c6e06caa967c148645

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/sensei-step-error/components.ts
gikaragia: Added fallback fonts in 1783b4ca14565edeb9765acb74c44e250e515a8a

packages/data-stores/src/site/types.ts
gikaragia: Good point. As we changed the requirements we will probably remove this altogether so I didn't refactor.

packages/onboarding/src/flow-progress/use-flow-progress.ts
gikaragia: I tried adding user but as we have too many steps it seems like the senseiSetup has almost half the bar completed which didn't look very nice.

packages/onboarding/src/sensei-step-container/index.tsx
gikaragia: Fixed in 270644442f3b97fc30fd72dc66b64a1930628a28

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/styles.scss
gikaragia: Moved to a common file here: 2b105cd55e551c7b35b817c6e06caa967c148645

packages/onboarding/src/sensei-step-container/styles.scss
gikaragia: I agree however this class comes from the stepper API and I think that we will need to rename the whole flow if we don't want to add an extra class. Do you think that it's ok to leave it as it is?

packages/onboarding/src/sensei-step-container/styles.scss
gikaragia: I moved the selector inside our own components in d31a0e0fd3631924d3f29358fca4a65d4e937d5e to mitigate the risk.

packages/onboarding/src/sensei-step-container/index.tsx
gikaragia: Good point fixed in a965d60e28bb449a8e097aa332967171e47e3256

gikaragia:
Hey @simison this is ready for review again. I have also added a feature flag in order to release this only in pre-production environments to further test it. Thanks!

gikaragia:
Hey @dlind1 I updated the string to 'Hosted by WordPress.com' and I received a ping from the Translation Bot saying that the string is outside the 'watched paths'. This seems a bit weird as the translators tried to translate the original string. Should I just ignore the ping?

dlind1:
> Hey @dlind1 I updated the string to 'Hosted by WordPress.com' and I received a ping from the Translation Bot saying that the string is outside the 'watched paths'. This seems a bit weird as the translators tried to translate the original string. Should I just ignore the ping?

The strings are translatable because the `String Freeze` label was applied, but the path doesn't exist yet (neither in trunk or our watched paths). I've added it to the Weekly Kit watched paths and it should be resolved now.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7683343).

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-domain/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+	} );
+	const { setDomain } = useDispatch( ONBOARD_STORE );
+
+	const onSkip = () => {
+		setDomain( domain );
+		submit?.( { domain: domain } );
+	};
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	const onAddDomain = ( selectedDomain: any ) => {
```
alshakero:
```suggestion
	const onAddDomain = ( selectedDomain: typeof domain ) => {
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-domain/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+		setDomain( selectedDomain );
+		submit?.( { domain: selectedDomain } );
+	};
+
+	const domainSuggestion = domain ? domain.domain_name : siteTitle;
```
alshakero:
```suggestion
	const domainSuggestion = domain?.domain_name ?? siteTitle;
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+				const userGlobalStylesLink: string =
+					theme?._links?.[ 'wp:user-global-styles' ]?.[ 0 ]?.href || '';
+				const userGlobalStylesId = parseInt( userGlobalStylesLink.split( '/' ).pop() || '', 10 );
+				const styleVariation = styleVariations.find(
+					( variation ) => variation.title === selectedStyleVariationTitle
+				);
+				if ( styleVariation && userGlobalStylesId && ! isNaN( userGlobalStylesId ) ) {
+					await updateGlobalStyles( siteId, userGlobalStylesId, styleVariation );
+				}
+			}
```
alshakero, line 211: I think you need an `else` here for error handling. Low hanging fruit.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+				languageSlug: locale,
+				visibility,
+			} );
+
+			setProgress( {
+				percentage: 50,
+				title: cartProgressTitle,
+			} );
+
+			const newSite = getNewSite();
```
alshakero, line 183: This selector is a bit of code smell and I wish I saw the PR that added it long ago tbh. Which new site? This is a source of bugs in async scenarios.

I recommend using the return value of `createSenseiSite` it gives you the same thing.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+			const siteId = newSite?.blogid;
+			setSelectedSite( newSite?.blogid );
+			await setIntentOnSite( newSite?.site_slug as string, SENSEI_FLOW );
+			await saveSiteSettings( newSite?.blogid as number, { launchpad_screen: 'full' } );
```
alshakero: Since not handling errors, let's run them in parallel for better perf.

```suggestion
			await Promise.all([
				setIntentOnSite( newSite?.site_slug as string, SENSEI_FLOW ),
				saveSiteSettings( newSite?.blogid as number, { launchpad_screen: 'full' })
			]
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+			setProgress( {
+				percentage: 75,
+				title: styleProgressTitle,
+			} );
+
+			if ( siteId ) {
+				const selectedStyleVariationTitle = getSelectedStyleVariation()?.title;
+				const styleVariations: StyleVariation[] = await getStyleVariations( siteId, 'pub/course' );
+				const theme: Theme = await getSiteTheme( siteId, 'pub/course' );
```
alshakero: I'd parallelize this as well.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+				const userGlobalStylesLink: string =
+					theme?._links?.[ 'wp:user-global-styles' ]?.[ 0 ]?.href || '';
+				const userGlobalStylesId = parseInt( userGlobalStylesLink.split( '/' ).pop() || '', 10 );
```
alshakero, line 204: It's NaN in either case. 

```suggestion
				const userGlobalStylesId = parseInt( userGlobalStylesLink.split( '/' ).pop(), 10 );
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+				const styleVariation = styleVariations.find(
+					( variation ) => variation.title === selectedStyleVariationTitle
+				);
+				if ( styleVariation && userGlobalStylesId && ! isNaN( userGlobalStylesId ) ) {
```
alshakero: I think `NaN` is falsey. 

```suggestion
				if ( styleVariation && userGlobalStylesId && ) {
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+			}
+
+			await cartManagerClient.forCartKey( cartKey ).actions.addProductsToCart( productsToAdd );
+			const redirectTo = encodeURIComponent(
+				`/setup/sensei/senseiLaunch?siteSlug=${ newSite?.site_slug }&siteId=${ newSite?.blogid }`
+			);
+
+			window.location.replace(
+				`/checkout/${ newSite?.site_slug }?signup=1&redirect_to=${ redirectTo }`
+			);
```
alshakero: I'd return just to stop execution.
```suggestion
			return window.location.replace(
				`/checkout/${ newSite?.site_slug }?signup=1&redirect_to=${ redirectTo }`
			);
```

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+		},
+		{
+			name: __( 'Best-in-class hosting' ),
+			requiresAnnuallyBilledPlan: false,
+		},
+		{
+			name: __( 'Advanced Jetpack features' ),
+			requiresAnnuallyBilledPlan: false,
+		},
+	];
```
alshakero: Should be in its own file, this file is already too large IMO.

client/landing/stepper/declarative-flow/sensei.ts
```diff
@@ -0,0 +1,101 @@
...
+	useStepNavigation( _currentStep, navigate ) {
+		const flowName = this.name;
+		const { setStepProgress } = useDispatch( ONBOARD_STORE );
+		const flowProgress = useFlowProgress( { stepName: _currentStep, flowName } );
+		const siteSlug = useSiteSlug();
+		setStepProgress( flowProgress );
+
+		const goBack = () => {
+			return;
+		};
```
alshakero:
```suggestion

```

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/tasks.tsx
```diff
@@ -68,6 +68,18 @@ export const tasks: Task[] = [
 		taskType: 'blog',
 		disabled: true,
 	},
+	{
+		id: 'sensei_setup',
+		completed: true,
+		taskType: 'blog',
+		disabled: false,
```
agrullon95: You can change this property to `disable: true` to ensure the `Set up Course Site` does not appear as if it's clickable on the launchpad screen.

client/landing/stepper/declarative-flow/sensei.ts
```diff
@@ -0,0 +1,101 @@
...
+				case 'senseiSetup':
+					return navigate( 'senseiDomain' );
+				case 'senseiDomain':
+					return navigate( 'senseiPlan' );
+				case 'launchpad':
+					return window.location.assign( `/view/${ siteSlug }` );
+				default:
+					return navigate( 'launchpad' );
+			}
+		};
```
alshakero: It seems none of your steps use this, you can delete it. Assuming you switch to `submit` from the other comment.

client/landing/stepper/declarative-flow/sensei.ts
```diff
@@ -0,0 +1,101 @@
...
+				case 'launchpad':
+					return navigate( 'processing' );
+				default:
+					return navigate( 'launchpad' );
+			}
+		};
+
+		const goToStep = ( step: string ) => {
+			navigate( step );
+		};
```
alshakero:
```suggestion

```

client/landing/stepper/declarative-flow/sensei.ts
```diff
@@ -0,0 +1,101 @@
...
+
+		return { goNext, goBack, goToStep, submit };
```
alshakero:
```suggestion
		return { submit };
```

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -210,14 +210,15 @@ export function getEnhancedTasks(
 						actionDispatch: () => {
 							if ( site?.ID ) {
 								const { setPendingAction, setProgressTitle } = dispatch( ONBOARD_STORE );
-								const { launchSite } = dispatch( SITE_STORE );
+								const { launchSite, saveSiteSettings } = dispatch( SITE_STORE );
 
 								setPendingAction( async () => {
 									setProgressTitle( __( 'Launching Website' ) );
 									await launchSite( site.ID );
+									await saveSiteSettings( site.ID, { launchpad_screen: 'off' } );
```
agrullon95: Adding `await saveSiteSettings( site.ID, { launchpad_screen: 'off'...` here will be redundant and affect all the flows.

We handle updating the `launchpad_screen` option to `off` on WPCOM (launchpad.php) to ensure that this option only updates under certain conditions. The sensei flow will have to be added into the WPCOM (launchpad.php) `track_site_launch_task()` function to ensure the `launchpad_screen` option is updated correctly.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+	const { goNext } = navigation;
+	const dispatch = useDispatch( ONBOARD_STORE );
+	const handleSubmit = useCallback( () => {
+		dispatch.setSiteTitle( siteTitle );
+		const variation = styles.find( ( style ) => style.name === checked ) || styles[ 0 ];
+		dispatch.setSelectedStyleVariation( {
+			slug: variation.name,
+			title: variation.title,
+		} as StyleVariation );
+		goNext?.();
```
alshakero: You can call submit instead. `goNext` is deprecated.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -270,6 +271,23 @@ export function getEnhancedTasks(
...
+					taskData = {
+						title: translate( 'Set up Course Site' ),
+						completed: true,
+					};
+					break;
+				case 'sensei_publish_first_course':
+					taskData = {
+						title: translate( 'Publish your first Course' ),
+						completed:
+							site?.options?.launchpad_checklist_tasks_statuses?.publish_first_course || false,
```
agrullon95, line 293: The `publish_first_course` launchpad_checklist_tasks_statuses property can also be updated in WPCOM (launchpad.php), so the correct value will be reflected here.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/index.tsx
agrullon95: The `goNext()` function is deprecated. You can use `submit()` instead: https://github.com/Automattic/wp-calypso/pull/68172

alshakero (changes requested):
Reviewed the code and left some comments. Tested the product and it looks and feels great.

One general feedback I have is that this flow is highly specialized (as in the opposite of reusable). It adds its own non-reusable steps, its own non-reusable dispatchers, etc. This really limits the value of the code you added to this flow. Non-blocker, but I recommend writing or reusing generic steps and customizing them on top of the existing code.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-launch/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+						window.location.replace( launchpadUrl );
+					} );
+				} else {
+					window.location.replace( launchpadUrl );
+				}
+			}, 800 );
+		}, 3000 );
+
+		return () => clearInterval( intervalId );
+	}, [ plugins, launchpadUrl, dispatch, siteId, saveSiteSettings, launchpadScreen ] );
```
agrullon95: You may need to add `retries` to this useEffect dependencies array.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
```diff
@@ -0,0 +1,398 @@
...
+	Bundling,
+	Error,
+}
+
+const siteProgressTitle: string = __( 'Laying out the foundations' );
+const cartProgressTitle: string = __( 'Preparing Your Bundle' );
+const styleProgressTitle: string = __( 'Applying your site styles' );
+
+const getStyleVariations = ( siteId: number, stylesheet: string ): Promise< StyleVariation[] > =>
+	wpcomProxyRequest( {
```
agrullon95: I'm not too familiar with the usage of `wpcomProxyRequest`, but I did find a P2 recommending we should not use wpcomProxyRequest directly (if I'm understanding correctly): p4TIVU-9GP-p2#comment-10997.

Im wondering if we should use wpcom.req.get / wpcom.req.post or React Query instead?

@alshakero - Do you have knowledge regarding these api requests?

edanzer:
UPDATE: This PR has been updated, and is now running updated Node. 

ORIGINAL: Just starting to test, but one quick note is that Calypso was [updated yesterday to node v18](https://github.com/Automattic/wp-calypso/pull/72245). Given the nature of a node version change and the size of this PR, it may be worth rebasing and retesting against the updated version of node to confirm no issues.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/index.tsx
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-setup/index.tsx
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
alshakero: Yes I'd totally move this to either a store on a React query.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
yscik: 👋 To clarify, is it okay to turn the option off from Calypso in a step that's only run for the Sensei flow? (When we are setting up other options like the site intent.) Or would that interfere with something? I'd prefer not spreading around setting the option and a flow that relies on it to different parts of the system since that adds a potential source of issue.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/tasks.tsx
gikaragia: Fixed in 3bd4789eafd176d21af3091af84787085a67df85

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: Good idea but `createSite` doesn't return the same thing unfortunately. It returns true on success, false on failure. Updated the return type to be correct in 7e64b31da912fc1a522908f1ae0139b9406ff481

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Fixed in f2da75ffa9b20b20abf3b1f4b8c14da042610ddb

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: Done b9dce195869c24f0eed172a9f71e162af98a778a

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: Yep but parseInt requires strictly a string but pop() can return undefined.

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: Fixed in 3d1308e836631c0c811b5fe37d622b1a137fb3e7

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: Fixed in 98c14ca687b2d4bfc32d1eb4a57d516ca3d031e9

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-launch/index.tsx
gikaragia: Fixed in 56d52543c7a98f99cc12740c3f58a2b5b3ee981c

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: I switched to using wpcom.req for now. We'll consider creating a store in a future PR 095f8dcec971f31d505997d3a488f88e1643aad7

client/landing/stepper/declarative-flow/internals/steps-repository/sensei-plan/index.tsx
gikaragia: The `else` case here is a normal scenario. It can happen when the user hasn't selected a style in the first step. In case the `updateGlobalStyles` fails, then we handle the error normally (although we could generally improve error handling, we have a separate card for it)

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
gikaragia: We are going to drop the launchpad screen from our flow so we will handle this as part of another card.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
gikaragia: As we are going to eventually drop the launchpad, I have removed the code that affects other flows. Any more changes to our launchpad screen are going to be handled as part of another PR 667560ba720104391e8ec8f3f903644020c8e66f

edanzer:
I just ran through an extra test of this. For the most part, fluid and nice flow. I did notice a few things on the Launchpad screen. Sounds like you may be skipping this screen in the future, so you may or may not fix these. They could also easily be resolved in follow up PRs. 
 - The Go to Admin link (bottom of sidebar) does not work. We actually recently ran into this when introducing some other flows. I had it myself. The fix (for now) is to reintroduce the deprecated goNext method you were asked above to remove. You can see my exact commit fixing it for another flow here: https://github.com/Automattic/wp-calypso/pull/72520/commits/9b0278e3e4fe2febf26143a847d52936b36dbee8 We need to update launchpad to not rely on that, be we can handle that in a separate PR, and will circular around to remove goNext in the sensei flow when done. 
 - There are some small styling issues you would never notice if you weren't working on Launchpad. You need to add an extra class (your .sensei class) to ensure all styles apply. Again, you can see a commit to fix the same thing here: https://github.com/Automattic/wp-calypso/pull/72520/commits/0afac59d2e1d4293c30be97a5690d75d540017fc 
- Clicking the Publish first course button should turn off the launchpad screen off, which should mean a) My Home doesn't not redirect to launchpad and b) if you try to load launchpad, it redirects back to My Home. For some reason (b) is not applying. Small issue because users could only get back to that screen by entering the url manually or via the back button. This is something we can also troubleshoot and resolve may be something with launchpad itself.

packages/data-stores/src/site/types.ts
```diff
@@ -463,6 +463,7 @@ export interface LaunchPadCheckListTasksStatuses {
...
```
agrullon95: Minor change, can you make this property optional? `publish_first_course?: boolean`

packages/data-stores/src/site/types.ts
edanzer: I wonder if publish_first_course here needs to be made optional with ? to avoid typescript errors elsewhere? Also worth noting you might be able to leverage first_post_published - depends on if you're needed to to do something specialized in terms of how you manage the logic for this.

client/landing/stepper/declarative-flow/sensei.scss
```diff
@@ -0,0 +1,6 @@
+$sensei-mobile-layout-width: 1200px;
```
edanzer: As a small issue of code organization (possibly personal preference on this), part of me would like to find a better place to put this scss file. It's kind of by itself here alongside all the flow defintiations. Most obvious place would be stepper/declarative-flow/internals/global.scss. There is already some flow specific scss in that file that's intended to apply to all steps in a particular flow.

client/landing/stepper/declarative-flow/sensei.ts
```diff
@@ -0,0 +1,80 @@
...
+import SenseiDomain from './internals/steps-repository/sensei-domain';
+import SenseiLaunch from './internals/steps-repository/sensei-launch';
+import SenseiPlan from './internals/steps-repository/sensei-plan';
+import SenseiSetup from './internals/steps-repository/sensei-setup';
+import { AssertConditionState, Flow } from './internals/types';
+import './sensei.scss';
+
+const sensei: Flow = {
+	name: SENSEI_FLOW,
+	title: 'Sensei',
```
agrullon95: You should translate this title

packages/data-stores/src/site/types.ts
gikaragia: Fixed in b656077c7c6d2d0b909ae5d2099b75930d19c640

client/landing/stepper/declarative-flow/sensei.scss
gikaragia: > As a small issue of code organization (possibly personal preference on this), part of me would like to find a better place to put this scss file

Indeed this was initially inside the `internals` directory but we moved it outside it as part of another review 🙂  As the file is sourced by both JS and CSS files it would be more complex to move the CSS inside `global.scss`. For this reason, I just moved it inside the internals directory, which is what other flows are doing too. a03be43838f1646a592501d144d5b5f1e564ac29

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Fixed in f3631ed58cff0e73f3f9f714baea43b958b49a1c. Not sure how this will work though. As Sensei is our brand maybe we don't want it translated?

gikaragia:
>  Sounds like you may be skipping this screen in the future, so you may or may not fix these. They could also easily be resolved in follow up PRs.

Indeed for now we will not fix this screen as we intend to remove it in the future. As discussed I also moved `SenseiStepContainer` outside of `packages` in 25f5d954eb3886b1013d2dd1cdcc83570f2cd8fc. I guess that the reasoning that we placed it there is that it closely relates to `StepContainer` however putting it inside `components` is fine too.

client/landing/stepper/declarative-flow/sensei.ts
gikaragia: Hmm I now see that we also translate Sensei in other files so it should be translated everywhere for consistency.

agrullon95:
The styling looks great !
I went through the flow and successfully landed on the launchpad. The launchpad still has some issues like the `Go to admin` link missing on mobile view port, not working on desktop view, and the site not launching after publishing my first course. Since the flow is behind a feature flag and the launchpad screen is going to be removed in a follow up PR, I dont think its a big deal.

> Indeed for now we will not fix this screen as we intend to remove it in the future. As discussed I also moved SenseiStepContainer outside of packages in https://github.com/Automattic/wp-calypso/commit/25f5d954eb3886b1013d2dd1cdcc83570f2cd8fc. I guess that the reasoning that we placed it there is that it closely relates to StepContainer however putting it inside components is fine too.

Moving the SenseiStepContainer to this location makes sense to me. Lets double check with @alshakero to make sure its makes sense to move it into the stepper directory.

edanzer (approved):
Approving this! This is a BIG PR. I have focused on: 
 - Confirming that testing works - I can go through the flow without issues as a user.
 - Confirming that no harm is done - ie, no unanticipated (negative) effects or regressions outside of the feature flag. The feature flag provides a measure of safety for this PR.
 - Confirm the areas of code I know well look good.

I'm comfortable with all that. There are couple caveats: 
  - I didn't dive too deeply into the logic on specific new Sensei screens. Some have a lot of complex logic. I haven't gone through the logic on those screens in detail. There is confirmation others have.
  - There are few issues with the Launchpad screen, but that screen will be removed in a subsequent PR. 
  - There are some failing tests. Authors checked with Calypso team and it sounds like those can be ignored in this case.

PR #69797, opened 2022-11-04
Implement the tld-given variation of Link in Bio flow

southp:
## Proposed Changes
_This PR is based on https://github.com/Automattic/wp-calypso/pull/69887_

This PR implements the dotlink variation of Link in Bio flow required for the dotlink partnership project. The detailed project info can be found in pau2Xa-4tC-p2. 

The new flow interweaves a new signup flow on the classic signup framework, `/start/link-in-bio-tld`, and the Link in Bio Stepper flow, `/setup/link-in-bio`. Stricyly speaking it's a technical compromise that's complicated more than it should have. However, before we [move everything to Stepper](https://github.com/Automattic/wp-calypso/pull/68722), we will need to interweave the both frameworks in some ways. In this case, the flow goes as: 

domains(signup) -> user(signup) -> pattern(Stepper) -> link in bio site info(Stepper) -> plans(signup) -> processing screen(signup) -> *checkout(directed from the signup framework if the cart is not empty) -> launchpad(Stepper)

To make this work, this PR includes:

1. Add a new "middle destination" mechanism to the classic signup flow that redirects to other URLs after a certain step is completed.
2. A new flow on the classic signup framework, `link-in-bio-tld`, that provides `tld` dependency and transit to `link-in-bio` Stepper flow after the user step using the above "middle destination" mechanism.
3. Add this new `link-in-bio-tld` to various places like `link-in-bio` does, so they acts as close as possible. It's a clear indication that we should move away from determining these by the flow name, but it's not the right timing to do it. 
4. Update the `link-in-bio` Stepper flow so the step order varied when `tld` query parameter is given.
5. The domain step component now populates `link` as the `managedSubdomains` prop to the underlying domain registration subcomponent, so the `w.link` option will suppress `wordpress.com` free subomaind when the .link tld is given.
6. Hide the step UI for the `link-in-bio-tld` flow, since the following Stepper flow has its own progress indicator. It's not perfect, but at least it would be less confusing. See pau2Xa-4tC-p2#comment-12834.

Demonstration:

https://user-images.githubusercontent.com/1842898/201882048-8ac2e101-cf0a-4635-8397-5aad7b5d9888.mp4

### Test the dotlink variation of Link in Bio flow
[...]

PR #71409, opened 2022-12-21
Fix the availabilty check for FQDNs in the signup flow

hambai:
It seems that we've stopped doing availability checks in the signup flow. The issue is that in certain cases the domain suggestion engine doesn't return an exact match if you're searching for FQDN (I suppose it might be cached domain availability). In those cases a customer would search for a domain that is available but they won't get it in the signup domain suggestion step but it would show up if you search in the regular calypso add a domain flow though.

I did some investigations and it seems the reason is that `promoTlds` is an empty array which apparently doesn't work with the check we do in `RegisterDomainStep::checkDomainAvailability` method. This was introduced in #69797 about a month ago. I think a quick fix would be to just replace the empty array with `null`

#### Proposed Changes

* Replace the empty array value of `promoTlds` with `null`

#### Testing Instructions

* Enable browser dev tools network monitor
* Go to /start/domain and search for a FQDN
* Verify that `is-available` network request is triggered

#### Pre-merge Checklist

Related to #69797

see: p2MSmN-aKw-p2

### DIFF
--- a/client/signup/steps/domains/index.jsx
+++ b/client/signup/steps/domains/index.jsx
@@ -525,7 +525,7 @@ class DomainsStep extends Component {
 		}
 
 		const includeWordPressDotCom = this.props.includeWordPressDotCom ?? ! this.props.isDomainOnly;
-		const promoTlds = this.props?.queryObject?.tld?.split( ',' ) ?? [];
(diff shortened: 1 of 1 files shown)


delputnam (approved):
Really good catch @hambai !  Thank you for noticing and fixing this!!

Issue #70975, opened 2022-12-09
Stats: Improve card hover state improvements


jeffgolenski:
The current hover / active states of each visual line item, isn't great for a couple of reasons:

The blue icons on top of a light blue bar are difficult to see
![Image](https://user-images.githubusercontent.com/214813/206597498-3db7391d-6b18-431c-ab65-a666222dcca0.png)

The text is difficult to read when the bar ends underneath it and is hovered.
![Image](https://user-images.githubusercontent.com/214813/206597517-02f6b3a6-2e51-4180-912b-c76416bcd779.png)

### Proposed

![Image](https://user-images.githubusercontent.com/214813/206598135-8f5c2011-55ae-4e76-909f-231ba7cc41f2.png)

To fix this issue, we're proposed some deeper contrast and consistency with a new hover / active state. With this:

- [ ] All text and icons will be the same color
- [ ] The length of the bar when hovered is more clear
- [ ] Deeper contrast with a deeper color background and white text / icons (theme dependent, will need to test)

**See prototypes**
[Hover state](https://codesandbox.io/s/stats-module-v2-09uyhm), [Nested hover state](https://codesandbox.io/s/stats-module-v2-1-w-groupings-h6qq9v) 

(@grzegorz-cp feel free to alter the final product as you mentioned in slack. you had some great ideas!)

jsnmoon:
Closed by #71437.

PR #71437, opened 2022-12-21
Stats: Updating hover CSS for horizontal bars

grzegorz-cp:
#### Proposed Changes

* Adding new hover style for horizontal bars to improve visibility

#### Testing Instructions

* open live branch URL
* navigate to Stats -> Traffic
* hover one of the rows containing horizontal bars and verify that it looks properly (label, value and action icons are visible) across all breakpoints

Before:
Screenshot 2022-12-22 at 2 25 58 PM

After:
Screenshot 2022-12-22 at 7 31 17 PM
Screenshot 2022-12-22 at 7 30 40 PM

#### Pre-merge Checklist

Related to #70975


grzegorz-cp:
@jeffgolenski @keoshi Is the colour used for hover correct? 
I used  `--theme-highlight-color` which is the same colour but only for one theme.

packages/components/src/horizontal-bar-list/style.scss
```diff
@@ -28,17 +28,32 @@
...
+
+		}
+	}
+
+	.horizontal-bar-list-group {
+		.horizontal-bar-list-item {
+			&:hover,
+			&:focus,
+			&:focus-within {
+				margin-left: $stats-child-hover-indentation;
```
grzegorz-cp: ~~I didn't implement any indentation for nested lists. The hover is indented, but the actual rows are not. I did it to keep the same reference point for the bar's length and relation.
The downside of this approach is that it's harder to see which items are expanded.
Which version do you think is better @keoshi @jeffgolenski?~~

If an indentation is needed for the normal view (not only hover), this margin has to be moved 3 lines up. That should fix the issue.

dognose24: Regarding the indentation, there could be a background-color transition issue with setting the margin and padding on the `list item` directly, which would not appear with putting the padding on the `list`.


https://user-images.githubusercontent.com/6869813/209045953-50ad87cc-4d7e-4c11-bee8-a6a9ffd31c61.mov

grzegorz-cp: That's a good point. I can't put the margin on an element other than the list, which is also a list item, because the highlight background won't take the whole width of the parent items.

This made me think that only one level of indentation for hover and no indentation for regular (not hover) view looks very strange for elements with more than two levels of nesting (e.g. Referrers for some pages).
I added margin to a group which added indentation to all of the children and applied it to all child groups independent of the indentation level.

If this doesn't look right, we can revert the latest commit and fix the transition issue.

grzegorz-cp: The PR description has updated images of nested indentation (implemented in commit [e8234a0](https://github.com/Automattic/wp-calypso/pull/71437/commits/e8234a0fa129ad7816acacc82e8e13ba627aec20)). Before the change, it looked like this:

Screenshot 2022-12-22 at 2 22 21 PM
Screenshot 2022-12-22 at 2 22 40 PM

jeffgolenski:
Nice @grzegorz-cp. From a design standpoint, I think these are looking good. I ran through most of the themes and it's very nice. Works well on keyboard nav, too—thanks for that.

https://user-images.githubusercontent.com/214813/209231818-5a28a62b-8e24-4455-862e-f77eec2b3627.mov

Screen Shot 2022-12-22 at 4 47 03 PM
Screen Shot 2022-12-22 at 4 45 02 PM

RE: indented children on hover - Just want a sanity check from @keoshi here... on hover, should the highlight background color occupy the full width? Feels weird with the hover bg color not occupying the full width of the container. What do you all think?

Screen Shot 2022-12-22 at 4 49 11 PM

jsnmoon (approved):
Looks good to me! We can iron out the hover margin detail in a follow-up PR.

(I personally prefer having the hover highlight occupy the entire width.)

PR #71468, opened 2022-12-22
Marketplace Themes: Add loading state to the banner

### LINKED ISSUE #71319: Marketplace Themes: Disable banner when loading to prevent double click
We should disable the banner on the theme page when the user first clicks it to prevent double clicks while adding the products to the cart.

We can use the `isLoading` prop to achieve it:
https://github.com/Automattic/wp-calypso/blob/45096ef79969be766ffea022429acbdc5514452c/client/my-sites/theme/main.jsx#L681

valterlorran:
#### Proposed Changes

*  Disables the banner on the theme page while adding the items to cart for marketplace themes.

https://user-images.githubusercontent.com/1234758/209195941-35ca112f-d7c7-40e9-8d45-c4c3d0531703.mov

#### Testing Instructions

* Navigate to `/theme/makoney/{SITE}` on a free site
* Click on the banner
* The banner should be disabled while loading the cart

Closes #71319

### DIFF
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -771,6 +771,7 @@ class ThemeSheet extends Component {
 			isExternallyManagedTheme,
 			isSiteEligibleForManagedExternalThemes,
 			isMarketplaceThemeSubscribed,
+			isLoading,
 		} = this.props;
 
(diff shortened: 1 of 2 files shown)


markbiek (approved):
LGTM!

PR #47349, opened 2020-11-12
CRO i5: Remove jpcom masterbar fixed behavior on desktop.

elliottprogrammer:
### Changes proposed in this Pull Request

This PR removes the jpcom masterbar fixed on scroll behavior, making it static full-time - But only for jetpackConversionRateOptimization A/B test variant: `i5 - Saas table design`.

#### Additional implementation notes:

- In the `JetpackComMasterbar` component, I removed the `useDetectWindowBoundary()` hook and `ref` to go along with it.  This hook was detecting when the masterbar reaches the top of the viewport (which is immediately on page load) and makes the element "sticky", however the same behavior was already being accomplished with css so there was no need for this hook here, or the overhead to go along with it.
- I also fixed an issue with previous iterations (`v1 - 3 cols layout` & `v2 - slide outs`) where the plans grid UI was "jumping up" every time the plans-filter-bar goes to 'sticky' position. Scrolling the plans page on these iterations is much smoother now.

### Testing instructions

- Download the PR
- Run both Calypso and Jetpack cloud concurrently.
- Select the `i5 - Saas table design` variant of the a/b test `jetpackConversionRateOptimization`
- Visit `jetpack.cloud.localhost:3001/pricing` page.
- Verify that the top jpcom Masterbar is no longer "sticky" (fixed to the top) on scroll.
- Select previous iterations, and verify that the Masterbar **_is_** fixed on scroll on these iterations.  (all iterations except `i5`)
- While you're viewing previous iterations, notice that the plans page is no longer "_Jumping up_" when the plans-filter-bar goes sticky/fixed. You can visit wordpress.com to view the undesired jumping behavior.

Fixes 1196341175636977-as-1199168581500326

PR #71470, opened 2022-12-22
update: Draft on how sticky would look for header and banner

andrii-lysenko:
#### Proposed Changes

This is a Dratf to look like the sticky header will look like alongside the banner on pricing page. This is to agree if we should proceed this way. Only available on desktop (screen >900px wide)

#### Testing Instructions

*

#### Pre-merge Checklist

Related to #


elliottprogrammer:
I think it's a great @andrii-lysenko! 💯  Although still my personal opinion is that we continue on with this and proceed with the "one bar stuck at a time" suggestion. ( p1671486957716509/1671476803.542619-slack-C03RCHNVC0H ).

To me, it looks a bit weird having the masterbar left-aligned, and then having the intro-pricing-banner centered right below it.

But it's just my personal opinion, I'd love to know what any others think. The main thing is that the cart is always in view now, and this does it. 🙂

I know this is just a POC but I noticed there is a small gap between the two bar when they are both in fixed position. Is this intentional? (See image)

Markup 2022-12-22 at 15 01 55

Also, I remember about a year ago, the masterbar used to be "fixed" when scrolling like this. Doing a quick search, I see here is when/where we removed the sticky/fixed behavior from the masterbar in the first place:  https://github.com/Automattic/wp-calypso/pull/47349 . I don't know if seeing the code from when it was fixed back then is helpful or not, probably not, but I thought I'd reference it, just in case.

andrii-lysenko:
@elliottprogrammer Hmm, I don't see this gap and they are stick to each other. However, I do agree that it doesn't look ideal. To stick these one after one would be really easy if we could use `sticky` property, however, we can't because of `overflow: hidden`. I will try to see what we can to make it look good.

PR #71471, opened 2022-12-22
Ecommerce Tailored Flow: Remove the intro step

valterlorran:
#### Proposed Changes

* Removes the `intro` step and makes the `storeProfiler` as the first step.

Related: 
* paYKcK-2xV-p2#comment-1926
* peapX7-1go-p2

#### Testing Instructions

* Navigate to `/setup/ecommerce` with an authenticated user and make sure the flow works.
* Navigate to `/setup/ecommerce` without being logged in and make sure you are directed to the signup page. You can also add some flags to the URL to ensure that it's passing all the way after the signup.


markbiek (approved):
LGTM!

Issue #69947, opened 2022-11-11
Users: Privacy Information link is broken in Followers section


zdenys:
### Quick summary

When you click the (i) icon in Users → Followers you'll see the link Privacy Information that points to `https://wordpress.com/people/followers/{SITENAME}#privacy` and hence leads to the same page instead of supposedly to an external support page.

![LQXUDR.png](https://user-images.githubusercontent.com/39308239/201315232-c862acb3-4c58-4a96-97a8-6d1e579ff5ab.png)

### Steps to reproduce

1. On a site where you have at least one follower go to Users → Followers
2. Click the (i) icon to expand the modal window
3. Click Privacy Information

### What you expected to happen

An external support page to be opened (maybe https://wordpress.com/support/followers/ but it lacks the Privacy Information section, so maybe it should be another page)

### What actually happened

The same page reloads

### Context

Noticed while working on 5665433-zd-woothemes

### Platform (Simple, Atomic, or both?)

Simple

### Theme-specific issue?

No

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

No but the platform is still usable

jamiepalatnik:
📌 SCRUBBING : RESULT
- Tested on Simple - Replicated

📌 FINDINGS/SCREENSHOTS/VIDEO
- I was able to replicate this. 
- Clicking Privacy Information link reloads the page with #privacy at the end of the URL. 
- I would also expect this link to go to a more useful page, like relevant support documentation.

📌 ACTIONS
- Marked as Triaged for Quality Squad review

sejas:
- @zdenys , I've fixed this issue on https://github.com/Automattic/wp-calypso/pull/71536
I removed the broken `Privacy Information` and added a `Learn more` link to https://wordpress.com/support/followers/

Thanks for reporting the issue! 🙌 

| **Before** | **After** |
|---|---|
| <img width="614" alt="Screenshot 2022-12-29 at 12 48 52" src="https://user-images.githubusercontent.com/779993/209947272-a53d6bbb-188c-4cea-bc35-db0e0ec1ecd7.png"> |<img width="641" alt="Screenshot 2022-12-29 at 12 25 38" src="https://user-images.githubusercontent.com/779993/209944789-5bc6eedb-b105-41d8-bd73-c7621eb84c5b.png"> |

PR #71474, opened 2022-12-22
Hide incorrect privacy link

vykes-mac:
#### Proposed Changes
This [issue](https://github.com/Automattic/wp-calypso/issues/69947) relating to this PR has a greater impact and requires a bigger refactor than this immediate fix that's done here.

What's happening is that the popover for _Users>Followers_ is rendering a privacy link when It should not have. 
image
All the link does is reloads the followers page. This happens because of how the `<SupportInfo/>` component works by making `privacyLink` defaults to `true` so that even if an empty link is passed it still generates a privacyLink with `#privacy`. This design was call into question [here](https://github.com/Automattic/wp-calypso/pull/61260#pullrequestreview-887775837).

- We could make `privacyLink=false` that way the link doesn't render but that would require modifying components `PeopleListSectionHeader > SectionHeader ` to pass this prop down to  `<SupportInfo/>`. 
- We could also modify the `<SupportInfo/>` component to accept as props a privacy link or a general link instead of the component trying to be smart.

Both approaches will have a larger impact on the codebase and would require more refactoring. 

* For this fix I'm using css to remove the incorrect privacy link. 

#### Testing Instructions

1. Visit Users>Followers and ensure the popover doesn't show "Privacy Information" link.
image

*

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/69947

### DIFF
--- a/client/my-sites/people/followers-list/style.scss
+++ b/client/my-sites/people/followers-list/style.scss
@@ -2,3 +2,7 @@
 	font-size: $font-title-small;
 	margin-bottom: 16px;
 }
+
+.is-section-people .info-popover__tooltip .popover__inner .support-info__privacy {
+	display: none;
+}


vykes-mac:
Do we invest the time to do the larger refactor and fix the ```<SupportInfo/>``` or ship this as is?

mpkelly:
Is there any reason not to do this instead? Of the 44 uses that WebStorm finds, 43 of them pass a `link` prop and should continue working. The only place I could find that doesn't pass `link` is on the users page. 

```js
function makePrivacyLink( privacyLink, link ) {
	if ( privacyLink ) {
		if ( typeof privacyLink === 'string' ) {
			return privacyLink;
		}
		// proposal: make this part conditional
		if ( link ) {
			return link + '#privacy';
		}
	}

	return null;
}
```

sejas (commented):
@vykes-mac , thanks for writing down all the options you've considered.
 
The CSS is working as described 👌 . But I think we can change the react code to fix the issue.

After reviewing the code, I've found that `PeopleListSectionHeader` is the only component that calls `SectionHeader` with the `popoverText` prop. So we could add `privacyLink={false}` on `SectionHeader`: https://github.com/Automattic/wp-calypso/blob/f2e6f88ae65c8571bbac21b284c29b48ec987598/client/components/section-header/index.jsx#L33

However if `PeopleListSectionHeader` is the only component that uses the popover, then we can move the `SupportInfo` to the parent. Here is my suggested PR: 

- [SectionHeader: refactor SupportInfo to call it in the parent PeopleListSectionHeader #71536](https://github.com/Automattic/wp-calypso/pull/71536)

It is a bigger refactor, but I think it's safe. It was weird that the SectionHeader had a popover with the PrivacyLink "hardcoded".

vykes-mac:
Thanks @mpkelly @sejas , I'm closing this in favour of @sejas [PR](https://github.com/Automattic/wp-calypso/pull/71536)

PR #69710, opened 2022-11-02
Remove featured benefits experiment code

niranjan-uma-shankar:
#### Proposed Changes

* Removes the experiment implemented in https://github.com/Automattic/wp-calypso/pull/68804.

#### Testing Instructions

* Go through the signup flow and verify that the plans step works without issues.

PR #71487, opened 2022-12-23
Stepper: Remove unused folder

escapemanuele:
https://github.com/Automattic/wp-calypso/pull/68722  re-added the https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plan-comparison-with-highlights folder which was previously removed in https://github.com/Automattic/wp-calypso/pull/69710. 

It would be good to remove the folder since we don't use it anywhere.


niranjan-uma-shankar (approved):
LGTM, thanks!

PR #71496, opened 2022-12-23
Purchases: Change PurchaseMetaPrice to use billPeriodDays instead of label

### LINKED ISSUE #70549: PurchaseMetaPrice assumes english language terms
`PurchaseMetaPrice` contains a switch statement that looks at the contents of the `billPeriodLabel` on the purchase to decide what term to show.

https://github.com/Automattic/wp-calypso/blob/7f68bb845d543a8f1ac8ea8602300a7d8be63c1a/client/me/purchases/manage-purchase/purchase-meta.jsx#L300-L311

However, `billPeriodLabel` is localized for the user's language, so the switch statement will fail for non-EN languages. It must be replaced. 

In theory we could change it to compare the strings against translated versions of the cases, but that might not be reliable since the translations are coming from two different places (`billPeriodLabel` comes from the backend). 

Better would be to use some other property, either a number or a code, to determine the term length. 

This code was added by https://github.com/Automattic/wp-calypso/pull/65006 for Marketplace products specifically although it is used for all kinds of products.

sirbrillig:
#### Proposed Changes

`PurchaseMetaPrice` contains a switch statement that looks at the contents of the `billPeriodLabel` on the purchase to decide what term to show. However, `billPeriodLabel` is localized for the user's language, so the switch statement will fail for non-EN languages. It must be replaced.

This PR replaces it (and all the other logic that was being used) with `billPeriodDays` (aka `bill_period_days` on the data returned by the server), which, as far as I can tell, is set for every product returned by the purchases endpoint. In case a product does not match any of the options, it will be rendered with no label instead. This might be more accurate for one-time-purchases than the previous default of "yearly" (although I think that those may actually handled by a separate code path).

Fixes https://github.com/Automattic/wp-calypso/issues/70549

#### Testing Instructions

First, verify that the billing term is displayed correctly for various products:

* Use a site that has a monthly plan.
* Visit the purchases page (`/me/purchases`) and select the plan subscription.
* Verify that the period shown next to the price is `X / month`.
* Repeat the above steps with an annual plan and verify that the price is shown as `X / year`.
* Repeat the above steps for an email product and verify that the price is shown with the correct term.

Next, we must cover the testing instructions from https://github.com/Automattic/wp-calypso/pull/65006:

* Go to a marketplace Plugin page and buy the monthly subscription. Ex: `/plugins/woocommerce-subscriptions`
* Go to the purchases page (`/me/purchases`) and select the plugin subscription.
* Verify that the period shown next to the price is `X / month`.

Screenshot 2022-12-23 at 1 01 52 PM

### DIFF
--- a/client/me/purchases/manage-purchase/purchase-meta-price.tsx
+++ b/client/me/purchases/manage-purchase/purchase-meta-price.tsx
@@ -4,6 +4,8 @@ import {
 	isDIFMProduct,
 	isDomainTransfer,
 	isEmailMonthly,
+	PLAN_ANNUAL_PERIOD,
+	PLAN_MONTHLY_PERIOD,
 	TERM_BIENNIALLY,
 	TERM_MONTHLY,
 } from '@automattic/calypso-products';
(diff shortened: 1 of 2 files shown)


michaeldcain (approved):
This LGTM and works well. I tested with a monthly plan, 1yr plan, 2yr plan, marketplace product, and jetpack product.

PR #71526, opened 2022-12-27
Checkpoint Redux

JuanLucha:
#### Proposed Changes

This PR modifies the redux setup for email stats in the following way:

- Creating actions, reducers and selectors for the "requesting" state of the endpoint, so we can use it to trigger transformations of data and other things when the endpoint gives back the response.
- Creating a new transformation function to accommodate the new format of the email stats response.

#### Testing Instructions

In order to test the changes you can checkout this branch, start the application and use Redux Tools in the browser to test the new actions.


TimBroddin (commented):
Hey @JuanLucha ,

To test this PR I (temporary) added the `QueryEmailStats` component to `StatsEmailOpenDetail` skeleton, and mapped the state to a prop for inspection. This triggers the whole Redux flow.

```js
import { Component } from 'react';
import { connect } from 'react-redux';
import QueryEmailStats from 'calypso/components/data/query-email-stats';
class StatsEmailOpenDetail extends Component {
	render() {
		console.log(this.props.stats);
		window.emailStats = this.props.stats;
		return (
			<div>
				<QueryEmailStats postId={ 313 } siteId={ 210703284 } period="day" quantity={ 16 } /> Email
				Stats placeholder
			</div>
		);
	}
}

export default connect( ( state ) => ( {
	stats: state.stats.emails,
} ) )( StatsEmailOpenDetail );
````

While inspecting the data flowing into Redux, I noticed that the shape for both `items` and `requests` is `[siteId][postId]`. When looking at how the graph on the main overview of Jetpack stats is rendered, you'll notice that the shape is `[siteId][period]`. When you change the tab to Weeks, Months or Years, a new action is dispatched containing the `period`which loads the data into the right `period` entry of the store. By doing so, consequent switching between the tabs is fast.

| Email Open Stats shape | Jetpack Main stats page |
|-------------------------|-------------------------|
| CleanShot 2022-12-27 at 11 46 38@2x | CleanShot 2022-12-27 at 11 49 13@2x |

An alternative approach would be to handle this like the post detail stats page does. If you take a look at the stats of a post, you'll see that the action that's dispatched here already contains all the data to populate the weeks, months & years views of the chart:

CleanShot 2022-12-27 at 11 50 57@2x

I'm also confused by the two stores of data (`state/stats/email-chart-tabs` & `state/stats/emails`). It seems to me that email-chart-tabs won't be used, since `<QueryEmailStats />` calls `requestEmailStats` inside `state/stats/emails/actions`. I see that period is already being taken into account inside `state/stats/email-chart-tabs`.

Please ping me when anything isn't clear 👍

JuanLucha:
Thanks for the great comment, @TimBroddin !

I have made the changes to do so, and now the state have that period after the postId, both in storing the items data and for showing the request status:
Screenshot 2022-12-27 at 14 22 07

TimBroddin:
Very nice work @JuanLucha ! 🥇 

Would it be possible to add the type as well? So we can use this store to store the data for all the e-mail stats. ✋

TimBroddin (approved):
The shape LGTM! 

CleanShot 2022-12-27 at 16 49 39@2x

JuanLucha:
> Very nice work @JuanLucha ! 🥇
> 
> Would it be possible to add the type as well? So we can use this store to store the data for all the e-mail stats. ✋

Done! 🎸

PR #71537, opened 2022-12-28
Shopping Cart: Use formatCurrency on all prices meant for display

sirbrillig:
#### Proposed Changes

There are two ways that prices are formatted for display in calypso. One way uses the backend to perform the formatting, providing the price to the frontend as a string. The other uses the `@automattic/format-currency` package. For example, the checkout variant picker uses `formatCurrency()` but the checkout line item uses the backend price.

checkout-line-item-where-formatting-orignates

This PR changes checkout's display logic to use `formatCurrency()` for all of its prices.

Depends on https://github.com/Automattic/wp-calypso/pull/71387

Before:

after-de_DE-EUR

After:

Screenshot 2022-12-28 at 5 09 34 PM

#### Testing Instructions

To prove that this replaces all instances of the backend-formatted prices, this PR removes those prices from the shopping-cart package types. This would cause the TypeScript compiler to throw an error if something was still using those properties. Of course, there are many non-TypeScript files in calypso, so you can also use grep to search for the removed properties and verify that any remaining ones are part of tests or are in packages that do not make use of them: `rg 'total_tax_display|total_cost_display|coupon_savings_total_display|sub_total_with_taxes_display|sub_total_display|credits_display|item_original_cost_display|item_original_cost_for_quantity_one_display|item_subtotal_display|product_cost_display|item_subtotal_monthly_cost_display|item_original_subtotal_display'`

- Try using a user whose currency is set to something with formatting quite different from USD like `IDR` or `EUR`.
- Set your browser's locale to match the currency's most common locale. For `IDR` try setting the locale to `in` and for `EUR` try setting the locale to `de-DE`. In Chrome's devtools you can change your browser's locale setting [by using the "Sensors" panel](https://developer.chrome.com/docs/devtools/device-mode/geolocation/).
- Add a product to your cart and visit checkout in two windows: one with this branch and one with production calypso.
- Verify that all the prices displayed on the page look correct and more-or-less identical to the production version (there may be slight variations but they should match the prices that already use `formatCurrency` which is the whole point of this PR).


spraveenitpro:
LGTM 

<details>
<summary>In the PR Branch</summary>



</details>

<details>
<summary>Production Branch</summary>


</details>

sirbrillig:
Closing this in favor of https://github.com/Automattic/wp-calypso/pull/71722 which does the same thing, but without the dependency on https://github.com/Automattic/wp-calypso/pull/71387, which will allow this PR to be merged first.

PR #71563, opened 2022-12-29
Theme Showcase: Implement Theme Preview modal component

taipeicoder:
#### Proposed Changes

This PR implements a new theme preview modal for the Theme Showcase, building on top of the `@automattic/design-preview` components.

Screen Shot 2023-01-03 at 11 18 18 AM

The scope of this PR is only the theme preview modal in isolation, which means that the end-to-end interaction flow is not implemented here. Specifically: 

1. Carrying over selected style variation from the theme card to the theme preview.
2. Post activation flow in the theme preview (it's kept the same as the one in production).
3. The selected style variation will not be applied to the activated theme.

#### Testing Instructions

* Select any theme except the active theme, and click on the "Open live demo" button in the theme details page.
* Ensure that the new preview modal shows up, and is as per design spec in 0EogtxvlmxMGcnEEaMVECv-fi-1401%3A53070&t=w0yGbxu7F2y0vApp-0.
* Ensure that the new preview modal also works in the logged-out Theme Showcase (https://wordpress.com/themes).
* Also test the new preview modal in an Atomic site.

#### Pre-merge Checklist


client/components/theme-preview-modal/style.scss
```diff
@@ -0,0 +1,192 @@
+@import "@wordpress/base-styles/breakpoints";
+@import "@wordpress/base-styles/mixins";
+@import "@wordpress/base-styles/variables";
+
+$break-design-preview: 1024px;
```
taipeicoder, line 5: It would be better to use one of the breakpoints defined [here](https://github.com/WordPress/gutenberg/blob/4ccdee3b0004ca9aa464151b93dfdced8502dcac/packages/base-styles/_breakpoints.scss) and avoid declaring this custom breakpoint over and over in different places.

I'll create a follow-up PR to consolidate this breakpoint into `break-xlarge` (1080px) cc: @SaxonF

client/state/sites/hooks/use-premium-global-styles.ts
```diff
@@ -1,45 +1,14 @@
...
-};
-
-// While we are loading the Global Styles Info we can't assume that we should limit global styles, or we would be
-// showing notices for paid sites until we fetch the data from the server.
-const DEFAULT_GLOBAL_STYLES_INFO: GlobalStylesStatus = {
-	shouldLimitGlobalStyles: false,
-	globalStylesInUse: false,
-};
-
-const getGlobalStylesInfoForSite = ( siteId: number | null ): GlobalStylesStatus => {
```
taipeicoder, line 22: The reason for moving these functions in a new file `use-site-global-styles-status.ts` is so that we can reuse this query hook without introducing dependency of the `ONBOARD_STORE` which we don't need for the Theme Showcase.

arthur791004: I think `usePremiumGlobalStyles` can return `useSiteGlobalStylesStatus( siteId );` directly. However, as the file is still in `client/state/sites/hooks`, I'd prefer to modify `usePremiumGlobalStyles` to accept the `siteId` as input parameter instead of getting the value from `ONBOARD_STORE` internally. Otherwise, this hook should be moved to `@automattic/onboard` package or `landing/stepper` folder.

It's not a blocker, but I think we have to fix [SiteSettingsFormGeneralWithGlobalStylesNotice](https://github.com/Automattic/wp-calypso/blob/4ceb8e33c74dabea7ed2d70fa4df96dab6ac6e9b/client/my-sites/site-settings/form-general.jsx#L913) and replace `usePremiumGlobalStyles` with new one if possible as I don't think it should rely on the `ONBOARD_STORE`

client/components/theme-preview-modal/style.scss
```diff
@@ -0,0 +1,192 @@
...
+	align-items: center;
+	bottom: 0;
+	display: flex;
+	inset-inline-end: 0;
+	inset-inline-start: 0;
+	height: 60px;
+	gap: 24px;
+	margin: 0;
+	padding: 0 20px;
+	pointer-events: none;
```
arthur791004, line 41: I think we can set it to `auto` here and we don't need `> * { ... }` below. What do you think?

client/components/theme-preview-modal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+		setSelectedStyleVariation( variation );
+	}
+
+	function showGlobalStylesPremiumBadge() {
+		if ( ! shouldLimitGlobalStyles ) {
+			return null;
+		}
+
+		return (
+			<PremiumBadge
```
arthur791004: Could we move this into `@automattic/design-preview` and make `showGlobalStylesPremiumBadge` a boolean prop?

client/components/theme-preview-modal/style.scss
taipeicoder: Unfortunately, we need `pointer-events: none` so that the device switcher can be clicked:
Screen Shot 2023-01-04 at 9 47 17 AM

However, we can change `> * { ... }` to ` > a, button { ... }` to be less aggressive. What do you think?

client/components/theme-preview-modal/index.tsx
taipeicoder: Sure! This will impact the unified design picker, so please test for regression bugs 🙂

client/state/sites/hooks/use-premium-global-styles.ts
taipeicoder: +1, feels a bit odd to have a strong dependency to the `ONBOARD_STORE within the hook. Let's have a separate issue for this!  #71653

client/components/theme-preview-modal/style.scss
taipeicoder: Follow up issue: #71655

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -18,7 +18,6 @@ import { useDispatch as useReduxDispatch, useSelector } from 'react-redux';
 import AsyncLoad from 'calypso/components/async-load';
 import { useQuerySitePurchases } from 'calypso/components/data/query-site-purchases';
 import FormattedHeader from 'calypso/components/formatted-header';
-import PremiumBadge from 'calypso/components/premium-badge';
```
fushar, line 21: Noone is using this component anymore after this, so let's remove it? 😄

client/components/theme-preview-modal/style.scss
arthur791004: Okay, I didn't notice that. Let's leave it as it is 🙂

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
taipeicoder: With pleasure 🙂

arthur791004 (approved):
Tested on the below scenario and it looks good 👍
* simple site
* atomic site
* logged-out theme showcase
* build flow

bangank36 (approved):
Working great, detail screenshot as following
|      Simple sites     | Atomic | Logged out |
|-----------|--------------|--------|
|  |               |     |

fushar (approved):
For me I tested purchasing Premium / WooCommerce themes, still working great :)

PR #71572, opened 2022-12-29
E2E: Fix tracking specs against GB v14.8.x

WunderBart:
> **Warning**
> Do not merge before Gutenberg v14.8 is in production.

Context:
- p4TIVU-am0-p2
- p4TIVU-amj-p2

#### Proposed Changes

Fix tracking specs against the [GB v14.8.x upgrade](https://github.com/Automattic/wp-calypso/issues/71272):
- specs/editor-tracking/editor-tracking__block-inserted-event.ts
- specs/editor-tracking/editor-tracking__other-block-events.ts
- specs/editor-tracking/editor-tracking__toolbar-events.ts
- specs/editor-tracking/editor-tracking__global-styles-events.ts
- specs/editor-tracking/editor-tracking__fse-template-events.ts
- specs/editor-tracking/editor-tracking__document-actions-events.ts
- specs/editor-tracking/editor-tracking__pattern-events.ts

#### Testing Instructions

The above suites should be passing [against edge sites in the CI](https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_editor_tracking_simple_edge_desktop?branch=fix%2Fe2e-tracking-specs&mode=builds).

### DIFF
--- a/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
+++ b/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
@@ -43,6 +43,10 @@ describe(
 				editorPage = new EditorPage( page, { target: features.siteType } );
 			} );
 
+			afterAll( async () => {
+				await page.close();
+			} );
+
 			it( 'Start a new post', async function () {
 				await editorPage.visit( 'post' );
 				await editorPage.waitUntilLoaded();
@@ -146,6 +150,10 @@ describe(
 				editorPage = new EditorPage( page, { target: features.siteType } );
 			} );
 
+			afterAll( async () => {
+				await page.close();
+			} );
+
 			it( 'Start a new page', async function () {
 				await editorPage.visit( 'page' );
 				await editorPage.waitUntilLoaded();
@@ -202,9 +210,22 @@ describe(
 				fullSiteEditorPage = new FullSiteEditorPage( page, { target: features.siteType } );
 			} );
 
+			afterAll( async () => {
+				// Always try to delete the created template part.
+				if ( templatePartName ) {
+					await fullSiteEditorPage.deleteTemplatePart( templatePartName );
+				}
+
+				await page.close();
+			} );
+
 			it( 'Visit the site editor', async function () {
 				await fullSiteEditorPage.visit( testAccount.getSiteURL( { protocol: false } ) );
 				await fullSiteEditorPage.prepareForInteraction( { leaveWithoutSaving: true } );
+			} );
+
+			it( 'Close the navigation sidebar', async function () {
+				await fullSiteEditorPage.closeNavSidebar();
 			} );
 
 			// A lot of block insertions sometimes happen on load
@@ -299,13 +320,6 @@ describe(
 					);
 					expect( eventDidFire ).toBe( false );
 				} );
-			} );
-
-			// Always try to delete the created template part.
-			afterAll( async function () {
-				if ( templatePartName ) {
-					await fullSiteEditorPage.deleteTemplatePart( templatePartName );
-				}
 			} );
 		} );
 	}
--- a/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
+++ b/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
@@ -280,7 +280,7 @@ describe(
 							matchingProperties: {
 								block_name: 'core/page-list',
 								entity_context: 'core/template-part',
-								template_part_id: `pub/blockbase//${ templatePartName.toLowerCase() }`,
+								template_part_id: `pub/twentytwentytwo//${ templatePartName.toLowerCase() }`,
 							},
 						}
 					);
--- a/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
+++ b/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
@@ -288,7 +288,7 @@ describe(
 				} );
 			} );
 
-			describe( 'Adding blocks from existing template parts', function () {
+			describe.skip( 'Adding blocks from existing template parts', function () {
 				it( 'Add a Header block', async function () {
 					const block = await fullSiteEditorPage.addBlockFromSidebar(
 						HeaderBlock.blockName,
@@ -312,16 +312,12 @@ describe(
 				// arguably a reasonable outcome. We need to decide whether
 				// to adjust the test to match the tracking behavior or adjust
 				// the underlyting tracking behavior.
-				it.skip( '"wpcom_block_instered" event does NOT fire', async function () {
+				it( '"wpcom_block_instered" event does NOT fire', async function () {
 					const eventDidFire = await editorTracksEventManager.didEventFire(
 						'wpcom_block_inserted'
 					);
 					expect( eventDidFire ).toBe( false );
 				} );
-			} );
-
-			it( 'Close the page', async function () {
-				await page.close();
 			} );
 		} );
 	}
--- a/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
+++ b/test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
@@ -43,10 +43,6 @@ describe(
 				editorPage = new EditorPage( page, { target: features.siteType } );
 			} );
 
-			afterAll( async () => {
-				await page.close();
-			} );
-
 			it( 'Start a new post', async function () {
 				await editorPage.visit( 'post' );
 				await editorPage.waitUntilLoaded();
@@ -132,6 +128,10 @@ describe(
 					} );
 				} );
 			} );
+
+			it( 'Close the page', async function () {
+				await page.close();
+			} );
 		} );
 
 		describe( 'In the page editor', function () {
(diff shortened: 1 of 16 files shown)


WunderBart:
@dpasque @Addison-Stavlo 

@noahtallen noticed (p1672270550575539/1672268819.313079-slack-CBTN58FTJ) that the tracking specs are failing for the Gutenberg v14.8 upgrade, but I suspect they've been broken for a while now. Failures from this PR seem to have been piling up since the last few upgrades. For example, the [`Details` feature was (re)moved in v14.5](https://github.com/WordPress/gutenberg/pull/44788), which would mean [those specs](https://github.com/Automattic/wp-calypso/pull/71572/commits/5cdab594256fae7f2b97346a7c8720cf4918ea3f) should be failing since ~Nov 18th.

What's the policy on running tracking specs? Are they considered blocking/required for the Gutenberg upgrade? I wasn't sure about that, so we've postponed the v14.8 upgrade (p4TIVU-amj-p2) mostly because of those failures.


/cc @worldomonation @fullofcaffeine

WunderBart:
There's still one failure that I didn't know how to address because it seems to actually point to a missing event:
- https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_editor_tracking_simple_edge_desktop/9236189?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A9236189%29%2Cid%3A1119. 

Could you take a look @Addison-Stavlo / @dpasque? 🙏

WunderBart:
The [`specs/editor/editor__schedule.ts: Editor: Schedule: Schedule: future`](https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_gutenberg_simple_edge_desktop/9236235?buildTab=tests&status=failed) spec is failing, but it's unrelated to this PR. I will address it in a separate one.

dpasque:
@WunderBart  -- thank you so much for working on these! 🙇 

> What's the policy on running tracking specs? Are they considered blocking/required for the Gutenberg upgrade? I wasn't sure about that, so we've postponed the v14.8 upgrade (p4TIVU-amj-p2) mostly because of those failures. 

I can help weigh in here!

Two helpful threads that I think should give a lot of context:
p1654549794215299-slack-C1A1EKDGQ
pdqkMK-mx-p2

It's been hard for a while to dial in the right relevant priority for these Tracks metric! They're valuable, but still require a decent amount of maintenance, and placing that among the many other priorities we have on dotcom is hard! I know @Addison-Stavlo's team has done a lot of work keeping up on the changes, but there will likely continue to be breakages over time. 

As discussed in the Slack thread, we opted to move them to their own build and make them non-blocking for Gutenberg upgrades. Since they usually only point to failures that don't affect the user experience, we didn't want to hold up those releases. The idea was that Gutenberg could still be deployed with a failing test, and then we would ping Cylon/KitKat to notify of the failure and we would go back and make updates. However, looking at the volume of what needed to be fixed here, I'm wondering if that will be sustainable. 😬 

I've been thinking a lot about Eric's comment here: pdqkMK-mx-p2#comment-382. @Addison-Stavlo -- do you think it would make sense to pare down the E2E tests to only the Tracks metrics that are actively being used? Is that too dramatic? I'm just trying to consider other ways to minimize the maintenance here!

test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
```diff
@@ -43,6 +43,10 @@ describe(
 				editorPage = new EditorPage( page, { target: features.siteType } );
 			} );
 
+			afterAll( async () => {
+				await page.close();
+			} );
+
```
dpasque: **Question**: Do we need this here? Aren't all the pages closed up by our custom jest environment? AFAIK, we usually don't need to close up pages manually during tests.

dpasque (commented):
Code changes look good btw! 👍 One question for you about the page closures.

test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
dpasque: And actually, I think that may mess with capturing screenshots and videos, because we iterate through open pages on teardown and grab the screenshots and videos. But if the page is already closed, the teardown logic won't find it!

dpasque:
> There's still one failure that I didn't know how to address because it seems to actually point to a missing event:
https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_editor_tracking_simple_edge_desktop/9236189?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A9236189%29%2Cid%3A1119.
Could you take a look @Addison-Stavlo / @dpasque? 🙏

I fixed this in [7d4b937](https://github.com/Automattic/wp-calypso/pull/71572/commits/7d4b937d8112ae097f54cec6892ea95921aea94c) -- it looks like we are now tracking the page list item block as part of the event now. 👍 

Also, it looks like the new sidebar design didn't play well with our pattern for deleting template parts, and we weren't deleting more than the first template part! I fixed that in [d90d6ca](https://github.com/Automattic/wp-calypso/pull/71572/commits/d90d6ca12f4ef478e0908dc53e6f5bfaac5da4d0).

With that, it looks like the [CI Edge build is green](https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_editor_tracking_simple_edge_desktop/9236725?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true)! 🙌 

cc @WunderBart  and @Addison-Stavlo (fyi that this no longer needs investigation! 👍 )

test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
WunderBart: Perfect catch! I've just realized that the `after` hook is fired even if the block fails, making it the wrong spot to put the page closing at because it indeed would prevent the artifacts from being created. 💥 In the case of suites like these, though, where there are multiple `describe` blocks, I'd still close the page at the end of every block (only in a dedicated step instead of the `after` hook). Otherwise, the recordings will not be stopped even for successful blocks and will be uploaded along with failed ones. For example, this will upload three sets of artifacts, of which only the last one is relevant:

```js
// foo.spec.js

describe({
  // opens page #1, succeeds
})
describe({
  // page #1 is open
  // opens page #2, succeeds
})
describe({
  // page #1 is open
  // page #2 is open
  // opens page #3, fails
})

// Artifacts for page #1, #2 and #3 are uploaded.
```


This is true for our current env setup, so alternatively, maybe we could make some changes not to create artifacts for successful blocks? Let me know what you think, @dpasque! 

---

Accidentally, while testing if what I've described above is true, I've uncovered a bug in our env setup that causes each artifact (trace/recording/image/etc.) to overwrite the previous one. It happens because we're creating the artifact name [outside of the context→page iteration blocks](https://github.com/Automattic/wp-calypso/blob/a70ebd434c737c8c8fb53aea94b53101f7f7f07f/packages/calypso-e2e/src/jest-playwright-config/environment.ts#L155-L180), which causes each context & page have the `0` index. I'll make a fix in another PR!

WunderBart:
Thank you for all the [context](https://github.com/Automattic/wp-calypso/pull/71572#issuecomment-1369055846) and [fixing that last bit](https://github.com/Automattic/wp-calypso/pull/71572#issuecomment-1369106863), @dpasque – much appreciated! 🙌
TBH, I suspected that these might not be required, but the fact that there were so many failures made me think that we need to take a closer look regardless before we continue upgrading. Having said that, I'll move on with the upgrade now that it's all clear and green. Thanks again! 🙇

test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
WunderBart: PR here: https://github.com/Automattic/wp-calypso/pull/71623

dpasque: > This is true for our current env setup, so alternatively, maybe we could make some changes not to create artifacts for successful blocks? Let me know what you think, @dpasque!

Yes for sure, I've long wanted to do this, but never just set aside to knock it out! I think that would be a nice improvement over our current setup 👍

dpasque (approved):
Approving pending the removal of the `afterAll` hooks! 👍 🚀

test/e2e/specs/editor-tracking/editor-tracking__block-inserted-event.ts
worldomonation: **comment**: A counterpoint is that we may have a situation where the previous top-level `describe` blocks may have passed but it inadvertently put the webapp in a condition that leads to an immediate failure in the next top-level `describe` block.

In the hypothetical spec with three `describe` blocks above, it is possible that an overlay appears between the last step of the second block and the beginning of the third block on the same page.

Granted, it is little bit of a cherry-picked scenario, I will admit that!

worldomonation (approved):

WunderBart:
@dpasque, I need some more help with this one if you have a spare cycle! 🙏 Looks like the Template Modal has been refactored with one of the recent upgrades, and [some specs started failing](https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_editor_tracking_simple_edge_desktop/9323119?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true) because of it. I'm having a hard time figuring out the right fix for this. Would you be able to take a look?

WunderBart:
Thanks for taking this one over the finish line @dpasque @fullofcaffeine! 🙇

PR #71600, opened 2022-12-30
Add ability to render custom list items in Banner component

### LINKED ISSUE #71597: Banner: allow custom list items
### Details

Part of pdKhl6-1th-p2.

Right now, the `<Banner />` component allows plain-text list rendering:

image

It's simple enough, but in some cases, you need more flexibility. We should make it possible to render any arbitrary list items, along with allowing styling the list through a custom class name.

### Done is

- [x] It's possible to render custom list items;
- [x] It's possible to pass a custom class name to the list element, or to style an existing class name.

vykes-mac:
#### Proposed Changes

*Add function to allow rendering of custom list items

#### Testing Instructions

1. Go to `/devdocs/design`
2. Find the banner component and verify banner with custom body
image

3. Verify the other banners in the devdocs is not affected and renders correctly

*

#### Pre-merge Checklist

Related to #71597

### DIFF
--- a/client/components/banner/README.md
+++ b/client/components/banner/README.md
@@ -57,6 +57,7 @@ function render() {
 | `jetpack`                    | `bool`   | false   | When true, a Jetpack logo will be rendered and the border color will be set to Jetpack green.                                |
 | `icon`                       | `string` | null    | The component icon.                                                                                                          |
 | `list`                       | `string` | null    | A list of the upgrade features.                                                                                              |
+| `body`                       | `node`   | null    | A custom component to render as the main content.                                                                            |
 | `onClick`                    | `string` | null    | A function associated to the click on the whole banner or just the CTA or dismiss button.                                    |
 | `plan`                       | `string` | null    | PlanSlug of the plan that upgrade leads to.                                                                                  |
 | `price`                      | `string` | null    | One or two (original/discounted) upgrade prices.                                                                             |
--- a/client/components/banner/README.md
+++ b/client/components/banner/README.md
@@ -57,7 +57,7 @@ function render() {
 | `jetpack`                    | `bool`   | false   | When true, a Jetpack logo will be rendered and the border color will be set to Jetpack green.                                |
 | `icon`                       | `string` | null    | The component icon.                                                                                                          |
 | `list`                       | `string` | null    | A list of the upgrade features.                                                                                              |
-| `body`                       | `node`   | null    | A custom component to render as the main content.                                                                            |
+| `renderListItem`             | `string` | null    | A function that allows rendering of custom list items.                                                                       |
 | `onClick`                    | `string` | null    | A function associated to the click on the whole banner or just the CTA or dismiss button.                                    |
 | `plan`                       | `string` | null    | PlanSlug of the plan that upgrade leads to.                                                                                  |
 | `price`                      | `string` | null    | One or two (original/discounted) upgrade prices.                                                                             |
--- a/client/components/banner/README.md
+++ b/client/components/banner/README.md
@@ -41,34 +41,34 @@ function render() {
 
 ### Props
 
-| Name                         | Type     | Default | Description                                                                                                                  |
-| ---------------------------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------- |
-| `callToAction`               | `string` | null    | Shows a CTA text.                                                                                                            |
-| `className`                  | `string` | null    | Any additional CSS classes.                                                                                                  |
-| `compact`                    | `bool`   | false   | Display a compact version of the banner.                                                                                     |
-| `description`                | `string` | null    | The banner description.                                                                                                      |
-| `disableHref`                | `bool`   | false   | When true, prevent the Banner to be linked either via the `href` props or as a side effect of the `siteSlug` connected prop. |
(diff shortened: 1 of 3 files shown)


zaguiini:
Thanks for putting this up!

I believe that this implementation would be better if we could maintain the list rendering on the `Banner` component and come up with a `renderListItem` prop function so that we can hook into the rendering process of the list instead of replacing them with our approach. This way we could make it flexible enough to render custom items, without repeating the loop. We're still rendering a list, so semantically it's the same thing -- just visually different.

It'd be something like this:

 ```diff
diff --git a/client/components/banner/index.jsx b/client/components/banner/index.jsx
index d390c7f2c0..fd1fcee05c 100644
--- a/client/components/banner/index.jsx
+++ b/client/components/banner/index.jsx
@@ -183,6 +183,7 @@ export class Banner extends Component {
 			feature,
 			compact,
 			list,
+			renderListItem,
 			price,
 			primaryButton,
 			title,
@@ -213,8 +214,12 @@ export class Banner extends Component {
 						<ul className="banner__list">
 							{ list.map( ( item, key ) => (
 								<li key={ key }>
-									<Gridicon icon="checkmark" size={ 18 } />
-									{ item }
+									{ renderListItem?.( item ) ?? (
+										<>
+											<Gridicon icon="checkmark" size={ 18 } />
+											{ item }
+										</>
+									) }
 								</li>
 							) ) }
 						</ul>

```

And then usage would go like this:

```diff
diff --git a/client/components/banner/docs/example.jsx b/client/components/banner/docs/example.jsx
index 4a7fa3b58a..7df3b9741f 100644
--- a/client/components/banner/docs/example.jsx
+++ b/client/components/banner/docs/example.jsx
@@ -1,5 +1,8 @@
+import { ListTile, Gridicon } from '@automattic/components';
 import Banner from 'calypso/components/banner';
 
+const icons = [ 'custom-post-type', 'cloud', 'layout' ];
+
 const BannerExample = () => (
 	<div>
 		<Banner disableHref title="A simple banner" />
@@ -10,6 +13,42 @@ const BannerExample = () => (
 			showIcon
 			title="Simple banner with description and call to action"
 		/>
+		<Banner
+			callToAction="Update"
+			disableHref
+			showIcon={ false }
+			title="Simple banner with custom list items"
+			list={ icons }
+			renderListItem={ ( icon ) => (
+				<ListTile
+					title={ <h5>Feature name</h5> }
+					leading={
+						<div
+							style={ {
+								display: 'flex',
+								marginRight: '10px',
+								alignItems: 'center',
+								borderRadius: '50%',
+								justifyContent: 'center',
+								flexShrink: 0,
+								width: '24px',
+								height: '24px',
+								padding: '3px 4px 4px 3px',
+								backgroundColor: 'var(--color-accent)',
+								color: 'var(--color-text-inverted)',
+							} }
+						>
+							<Gridicon icon={ icon } size={ 18 } />
+						</div>
+					}
+					trailing={
+						<div style={ { marginLeft: '8px', marginTop: '5px' } }>
+							<Gridicon color="#C6C6C6" icon="info-outline" size={ 18 } />
+						</div>
+					}
+				/>
+			) }
+		/>
 		<Banner showIcon={ false } title="Banner with showIcon set to false" />
 		<Banner
 			callToAction="Backup"

```

Instead of wrapping the items with a `div`, we could style the `banner__list` class to wrap the items horizontally 👍

vykes-mac:
@zaguiini  I wasn't thinking from the perspective of rendering a list even though that's what we are doing in this PR, I was thinking more generally since the component doesn't render children it means we will keep modifying if we want to render something else in another style apart from a list. It's not really customisable although that might have been the original intention.

zaguiini:
I see @vykes-mac. I'm a big fan of only including things only when we need them, and so far we did not need to render anything apart from a list 🤔 

I think that when there is a solid case where we want to render something else, we can go and add a `body` property (or even a `children`) instead. But, for now, I'd go with customizing list rendering. It's up to you, though: no strong opinions.

If we're going with the body version, I think it'd be a better idea to use `children` instead of `body` -- it's more React-ish.

vykes-mac:
> If we're going with the body version, I think it'd be a better idea to use children instead of body -- it's more React-ish.

This make sense. we could change the `renderListItem?` to maybe `onRenderListItem?` to show we are hooking into the list render process?

zaguiini:
>we could change the `renderListItem?` to maybe `onRenderListItem?` to show we are hooking into the list render process?

I don't think this is a good idea. The `on` prefix is generally used as an event callback (e.g. `onClick`, `onChange`), not as part of the rendering process. The `render` prefix is adequate in this case.

vykes-mac:
> Instead of wrapping the items with a div, we could style the banner__list class to wrap the items horizontally

@zaguiini  Wouldn't this affect the default list which is vertical? or do you mean override it

zaguiini:
>Wouldn't this affect the default list which is vertical?

Yes, the idea is to override the styles from `.banner__list` to align the items horizontally -- `display:flex;` might do the trick.

zaguiini (approved):
LGTM 🚢

p-jackson (approved):
> The render prefix is adequate in this case.

Agree, [render props](https://reactjs.org/docs/render-props.html) usually start with `render*`

I'm not a fan of how the banner controls on the right get centred like this, but we'll address it if it actually comes up in real use cases.
CleanShot 2023-01-06 at 16 29 32@2x

I've updated the docs a little and will merge now.

PR #71609, opened 2023-01-02
Plugin Details: Add a track event when the premium version available USP link is clicked

WBerredo:
#### Proposed Changes

Add a track event when a premium version available USP link is clicked

#### Testing Instructions

*On the browser console terminal enter localStorage.setItem( 'debug', 'calypso:analytics*' )
* Go to a plugin page associated with a premium plugin. Ex: `/plugins/wordpress-seo` or `/plugins/mailpoet`
* Click on the `Premium version available` USP link
* Check if the event `calypso_plugin_details_premium_plugin_link_click` was logged

Screen Shot 2022-12-21 at 12 03 12

### DIFF
--- a/client/my-sites/plugins/plugin-details-sidebar/index.jsx
+++ b/client/my-sites/plugins/plugin-details-sidebar/index.jsx
@@ -4,8 +4,10 @@ import {
 	WPCOM_FEATURES_INSTALL_PURCHASED_PLUGINS,
 } from '@automattic/calypso-products';
 import { useTranslate } from 'i18n-calypso';
+import { useCallback } from 'react';
 import { useSelector } from 'react-redux';
 import './style.scss';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { formatNumberMetric } from 'calypso/lib/format-number-compact';
 import { PlanUSPS, USPS } from 'calypso/my-sites/plugins/plugin-details-CTA/usps';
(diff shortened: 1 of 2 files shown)


client/my-sites/plugins/plugin-details-sidebar/index.jsx
```diff
@@ -64,6 +66,14 @@ const PluginDetailsSidebar = ( {
 
 	const isPremiumVersionAvailable = !! premium_slug;
 	const premiumVersionLink = `/plugins/${ premium_slug }/${ selectedSite?.slug || '' }`;
+	const premiumVersionLinkOnClik = useCallback( () => {
+		recordTracksEvent( 'calypso_plugin_details_premium_plugin_link_click', {
+			current_plugin_slug: slug,
+			premium_plugin_slug: premium_slug,
+			premium_plugin_link: premiumVersionLink,
```
cpapazoglou: What's the rationale of having the `premiumVersionLink` in the event properties?

cpapazoglou (approved):
Works as advertised 
CleanShot 2023-01-03 at 14 53 25@2x

I just left a minor comment.

client/my-sites/plugins/plugin-details-sidebar/index.jsx
WBerredo: I don't have a rationale 😅, I basically added it because I have it. Do you suggest to remove?

cpapazoglou: Yes we can remove it, anyway we already have the parts that build into it tracked.

PR #71611, opened 2023-01-02
WooExpress Trial: Initial flow framework

markbiek:
#### Proposed Changes

This adds the basic WooExpress trial flow framework. We also added product constants for use in future PRs.

Accessing the flow creates a new site with a default domain name and free plan and redirects the user to the dashboard.

#### Testing Instructions

**Logged-in user**
1. Go to http://calypso.localhost:3000/setup/wooexpress/
2. You should see the `processing` step.
3. After the `processing` step, you should be redirected to the dashboard.

**Logged-out user**
1. Go to http://calypso.localhost:3000/setup/wooexpress/
2. You should be redirected to the account creation page.

The new site should have a domain name like `[wpcom username][NN].wordpress.com` (eg `markbiek94.wordpress.com`) and be on the Free plan.

#### Pre-merge Checklist

- [NA] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [NA] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [NA] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [NA] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [NA] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #

### DIFF
--- a/client/landing/stepper/declarative-flow/registered-flows.ts
+++ b/client/landing/stepper/declarative-flow/registered-flows.ts
@@ -40,7 +40,7 @@ const availableFlows: Record< string, () => Promise< { default: Flow } > > = {
 			/* webpackChunkName: "tailored-ecommerce-flow" */ '../declarative-flow/tailored-ecommerce-flow'
 		),
 
-	wooExpressTrial: () =>
+	wooexpress: () =>
 		import(
 			/* webpackChunkName: "trial-wooexpress-flow" */ '../declarative-flow/trial-wooexpress-flow'
 		),
--- a/client/landing/stepper/declarative-flow/registered-flows.ts
+++ b/client/landing/stepper/declarative-flow/registered-flows.ts
@@ -40,9 +40,9 @@ const availableFlows: Record< string, () => Promise< { default: Flow } > > = {
 			/* webpackChunkName: "tailored-ecommerce-flow" */ '../declarative-flow/tailored-ecommerce-flow'
 		),
 
-	ecommerceTrial: () =>
+	wooExpressTrial: () =>
 		import(
-			/* webpackChunkName: "trial-ecommerce-flow" */ '../declarative-flow/trial-ecommerce-flow'
+			/* webpackChunkName: "trial-wooexpress-flow" */ '../declarative-flow/trial-wooexpress-flow'
 		),
 
 	free: () => import( /* webpackChunkName: "free-flow" */ '../declarative-flow/free' ),
--- a/client/landing/stepper/declarative-flow/registered-flows.ts
+++ b/client/landing/stepper/declarative-flow/registered-flows.ts
@@ -40,6 +40,11 @@ const availableFlows: Record< string, () => Promise< { default: Flow } > > = {
 			/* webpackChunkName: "tailored-ecommerce-flow" */ '../declarative-flow/tailored-ecommerce-flow'
 		),
 
+	ecommerceTrial: () =>
+		import(
+			/* webpackChunkName: "trial-ecommerce-flow" */ '../declarative-flow/trial-ecommerce-flow'
+		),
+
 	free: () => import( /* webpackChunkName: "free-flow" */ '../declarative-flow/free' ),
 
 	'free-post-setup': () =>
@@ -50,4 +55,5 @@ if ( config.isEnabled( 'themes/plugin-bundling' ) ) {
 	availableFlows[ 'plugin-bundle' ] = () =>
 		import( /* webpackChunkName: "plugin-bundle-flow" */ '../declarative-flow/plugin-bundle-flow' );
 }
+
 export default availableFlows;
--- a/client/landing/stepper/declarative-flow/trial-ecommerce-flow.ts
+++ b/client/landing/stepper/declarative-flow/trial-ecommerce-flow.ts
@@ -1,0 +1,62 @@
+import { useSelect, useDispatch } from '@wordpress/data';
+import { useSiteSetupFlowProgress } from '../hooks/use-site-setup-flow-progress';
+import { useSiteSlugParam } from '../hooks/use-site-slug-param';
+import { ONBOARD_STORE } from '../stores';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import ProcessingStep, { ProcessingResult } from './internals/steps-repository/processing-step';
+import SiteCreationStep from './internals/steps-repository/site-creation-step';
+import { Flow, ProvidedDependencies } from './internals/types';
+
+const ecommerceTrialFlow: Flow = {
+	name: 'ecommerce-trial',
+
+	useSteps() {
+		return [
+			{ slug: 'siteCreationStep', component: SiteCreationStep },
+			{ slug: 'processing', component: ProcessingStep },
+		];
+	},
+	useStepNavigation( currentStep, navigate ) {
+		const flowName = this.name;
+		const intent = useSelect( ( select ) => select( ONBOARD_STORE ).getIntent() );
+		const siteSlugParam = useSiteSlugParam();
+
+		const { setStepProgress } = useDispatch( ONBOARD_STORE );
+		const storeType = useSelect( ( select ) => select( ONBOARD_STORE ).getStoreType() );
+
+		const flowProgress = useSiteSetupFlowProgress( currentStep, intent, storeType );
+
+		if ( flowProgress ) {
+			setStepProgress( flowProgress );
+		}
+
+		const exitFlow = ( to: string ) => {
+			window.location.assign( to );
+		};
+
+		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
+			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
+			const siteSlug = ( providedDependencies?.siteSlug as string ) || siteSlugParam || '';
+
+			switch ( currentStep ) {
+				case 'siteCreationStep': {
+					return navigate( 'processing' );
+				}
+
+				case 'processing': {
+					const processingResult = params[ 0 ] as ProcessingResult;
+
+					if ( processingResult === ProcessingResult.FAILURE ) {
+						return navigate( 'error' );
+					}
+
+					return exitFlow( `/home/${ siteSlug }` );
+				}
+			}
+		}
+
+		return { submit, exitFlow };
+	},
+};
+
+export default ecommerceTrialFlow;
--- a/client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
+++ b/client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
@@ -10,8 +10,8 @@ import SiteCreationStep from './internals/steps-repository/site-creation-step';
 import { AssertConditionState } from './internals/types';
 import type { AssertConditionResult, Flow, ProvidedDependencies } from './internals/types';
 
-const wooExpressTrialFlow: Flow = {
-	name: 'wooexpress-trial',
+const wooexpress: Flow = {
+	name: 'wooexpress',
 
 	useSteps() {
 		return [
@@ -37,7 +37,7 @@ const wooExpressTrialFlow: Flow = {
 			}
 
 			const redirectTarget =
-				`/setup/wooExpressTrial` +
+				`/setup/wooexpress` +
 				( hasFlowParams ? encodeURIComponent( '?' + flowParams.toString() ) : '' );
 			const url =
 				locale && locale !== 'en'
@@ -101,4 +101,4 @@ const wooExpressTrialFlow: Flow = {
 	},
 };
 
-export default wooExpressTrialFlow;
+export default wooexpress;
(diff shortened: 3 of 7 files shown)


client/landing/stepper/declarative-flow/trial-ecommerce-flow.ts
```diff
@@ -0,0 +1,62 @@
...
+const ecommerceTrialFlow: Flow = {
+	name: 'ecommerce-trial',
+
+	useSteps() {
+		return [
+			{ slug: 'siteCreationStep', component: SiteCreationStep },
+			{ slug: 'processing', component: ProcessingStep },
+		];
+	},
+	useStepNavigation( currentStep, navigate ) {
```
daledupreez: I think we also need a safeguard to make sure users are actually logged in already. For reference, we do that [here](https://github.com/Automattic/wp-calypso/blob/4af60e6e499d81ac9e9a49e29f817ad594311fb5/client/landing/stepper/declarative-flow/tailored-ecommerce-flow.ts#L63-L94) in the tailored eCommerce flow. We can definitely fold that in later, but I want to make sure we handle that case.

packages/calypso-products/src/constants/wpcom.ts
```diff
@@ -26,6 +26,7 @@ export const PLAN_BLOGGER_2_YEARS = 'blogger-bundle-2y';
 export const PLAN_ECOMMERCE_MONTHLY = 'ecommerce-bundle-monthly';
 export const PLAN_ECOMMERCE = 'ecommerce-bundle';
 export const PLAN_ECOMMERCE_2_YEARS = 'ecommerce-bundle-2y';
+export const PLAN_ECOMMERCE_TRIAL_MONTHLY = 'wp-bundle-ecommerce-trial';
```
daledupreez: As per my post for the creation of the new product (paYKcK-2zK-p2), this is the product slug and should be `ecommerce-trial-bundle-monthly`.
```suggestion
export const PLAN_ECOMMERCE_TRIAL_MONTHLY = 'ecommerce-trial-bundle-monthly';
```

packages/onboarding/src/utils/flows.ts
```diff
@@ -6,6 +6,7 @@ export const LINK_IN_BIO_POST_SETUP_FLOW = 'link-in-bio-post-setup';
 export const VIDEOPRESS_FLOW = 'videopress';
 export const IMPORT_FOCUSED_FLOW = 'import-focused';
 export const ECOMMERCE_FLOW = 'ecommerce';
+export const ECOMMERCE_TRIAL_FLOW = 'ecommerce-trial';
```
daledupreez: I think we should more clearly differentiate this flow as one specific to WooExpress.

client/landing/stepper/declarative-flow/registered-flows.ts
```diff
@@ -40,6 +40,11 @@ const availableFlows: Record< string, () => Promise< { default: Flow } > > = {
 			/* webpackChunkName: "tailored-ecommerce-flow" */ '../declarative-flow/tailored-ecommerce-flow'
 		),
 
+	ecommerceTrial: () =>
```
daledupreez: I think we should call this something more like `wooExpress`, as we're really building a WooExpress-specific signup flow.

daledupreez (commented):
Overall, this is working as described, though there seemed to be almost no styling on the site creation and processing steps. It would be worth calling out what the expectations on that front ought to be -- I am 100% happy with handling that in a follow-up PR, as long as we're consciously deciding to do so.

I also think we should explicitly note that this is a WooExpress flow, in terms of naming and URL path.

daledupreez:
Also, it looks like we're no longer using the `signup/trial-ecommerce` feature flag -- please update the description to reflect that.

markbiek:
I've addressed all of the above feedback. I think the styles may be a simple tweak. If it ends up being complicated, I'll make a note that we're addressing it later.

markbiek:
The styles were an easy fix so I've taken care of those too.

client/landing/stepper/declarative-flow/registered-flows.ts
```diff
@@ -40,6 +40,11 @@ const availableFlows: Record< string, () => Promise< { default: Flow } > > = {
 			/* webpackChunkName: "tailored-ecommerce-flow" */ '../declarative-flow/tailored-ecommerce-flow'
 		),
 
+	wooExpressTrial: () =>
```
daledupreez: We might simply want this to be `wooExpress` or `wooexpress` - the trial is not really something we need to call out in the URL.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,107 @@
...
+import { USER_STORE, ONBOARD_STORE } from '../stores';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import ProcessingStep, { ProcessingResult } from './internals/steps-repository/processing-step';
+import SiteCreationStep from './internals/steps-repository/site-creation-step';
+import {
+	AssertConditionResult,
+	AssertConditionState,
+	Flow,
+	ProvidedDependencies,
+} from './internals/types';
```
daledupreez: I think we should split out the type imports so we have a clear declaration of what we're importing.
```suggestion
import { AssertConditionState } from './internals/types';
import type {
	AssertConditionResult,
	Flow,
	ProvidedDependencies,
} from './internals/types';
```

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,107 @@
...
+
+			const redirectTarget =
+				`/setup/wooexpress-trial/processing` +
+				( hasFlowParams ? encodeURIComponent( '?' + flowParams.toString() ) : '' );
+			const url =
+				locale && locale !== 'en'
+					? `/start/account/user/${ locale }?variationName=${ flowName }&redirect_to=${ redirectTarget }`
+					: `/start/account/user?variationName=${ flowName }&redirect_to=${ redirectTarget }`;
+
+			return url + ( flags ? `&flags=${ flags }` : '' );
```
daledupreez: I am not sure that we need to be including any `flags` here - given that we're not using any feature flags, I think we can drop this to keep things simpler.
```suggestion
			return url; // and remove the declaration of const flags above
```

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,107 @@
...
+		};
+
+		if ( ! userIsLoggedIn ) {
+			const logInUrl = getStartUrl();
+			window.location.assign( logInUrl );
+			result = {
+				state: AssertConditionState.FAILURE,
+				message: 'store-setup requires a logged in user',
```
daledupreez: This message should be updated to reflect the current flow/step.
```suggestion
				message: 'wooexpress-trial flow requires a logged in user',
```

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,107 @@
...
+			switch ( currentStep ) {
+				case 'siteCreationStep': {
+					return navigate( 'processing' );
+				}
+
+				case 'processing': {
+					const processingResult = params[ 0 ] as ProcessingResult;
+
+					if ( processingResult === ProcessingResult.FAILURE ) {
+						return navigate( 'error' );
```
daledupreez, line 97: Does this work if we don't have an `error` step defined in the flow?

packages/calypso-products/src/constants/wpcom.ts
```diff
@@ -26,6 +26,7 @@ export const PLAN_BLOGGER_2_YEARS = 'blogger-bundle-2y';
 export const PLAN_ECOMMERCE_MONTHLY = 'ecommerce-bundle-monthly';
 export const PLAN_ECOMMERCE = 'ecommerce-bundle';
 export const PLAN_ECOMMERCE_2_YEARS = 'ecommerce-bundle-2y';
+export const PLAN_WOOEXPRESS_TRIAL_MONTHLY = 'ecommerce-trial-bundle-monthly';
```
daledupreez: Let's keep this as `PLAN_ECOMMERCE_TRIAL_MONTHLY` - while we're building a WooExpress flow to offer the plan for now, the underlying plan is fairly generic in terms of what it represents.
```suggestion
export const PLAN_ECOMMERCE_TRIAL_MONTHLY = 'ecommerce-trial-bundle-monthly';
```

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,107 @@
...
+			let hasFlowParams = false;
+			const flowParams = new URLSearchParams();
+
+			if ( locale && locale !== 'en' ) {
+				flowParams.set( 'locale', locale );
+				hasFlowParams = true;
+			}
+
+			const redirectTarget =
+				`/setup/wooexpress-trial/processing` +
```
daledupreez: There are two problems here:
1. Is this the wrong flow slug? Should it not be `wooExpressTrial` to match the flow start URL?
2. We need to redirect the user back to the start of the flow or `siteCreationStep` here, otherwise they won't get a site created, and end up at `/read`. That also makes me realise that it may be better to call this step something simpler like `createSite` - the "Step" seems redundant. (I couldn't get this to work when I tried to hit the `siteCreationStep` directly.)
```suggestion
				`/setup/wooexpressTrial/` +
```

daledupreez (changes requested):
I hit some snags with the new user flow that we need to address, and I didn't see the CSS fixes you mentioned in [this comment](https://github.com/Automattic/wp-calypso/pull/71611#issuecomment-1371397408) - I am guessing you didn't push up the changes.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
markbiek: Regarding item 1, the stepper framework changes the url from `/setup/wooExpressTrial` to `/setup/wooexpress-trial` once you get to the first step. So, if you're redirecting to a specific step, that's the url format you need to use.

That said, give item 2, it probably makes more sense to start the flow over completely by redirecting to `/setup/wooExpressTrial`.

I'm hesitant to change `siteCreationStep` as part of this PR since it's used in a few other places. I've created #71707 so we'll remember to come back to it.

markbiek:
> I didn't see the CSS fixes you mentioned in https://github.com/Automattic/wp-calypso/pull/71611#issuecomment-1371397408 - I am guessing you didn't push up the changes.

@daledupreez  Correct 🤦 . The CSS fixes are there now.

I'm a little curious about your comment about the styles also not appearing on the `siteCreationStep`. I never actually see that step because it finishes so quickly. I think the CSS changes that fixed the `process` step will also fix `siteCreationStep` but I may need help verifying that.

markbiek:
@daledupreez Latest feedback has been addressed.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -0,0 +1,104 @@
...
+
+			const redirectTarget =
+				`/setup/wooexpress` +
+				( hasFlowParams ? encodeURIComponent( '?' + flowParams.toString() ) : '' );
+			const url =
+				locale && locale !== 'en'
+					? `/start/account/user/${ locale }?variationName=${ flowName }&redirect_to=${ redirectTarget }`
+					: `/start/account/user?variationName=${ flowName }&redirect_to=${ redirectTarget }`;
+
+			return url;
```
daledupreez: This assignment makes my head hurt. I think it would be easier to understand if we either add an early return or if we substitute only the start of the URL.
```suggestion
			// Early return approach
			if ( locale || locale === 'en' ) {
				return `/start/account/user/${ locale }?variationName=${ flowName }&redirect_to=${ redirectTarget }`;
			}
			
			return `/start/account/user?variationName=${ flowName }&redirect_to=${ redirectTarget }`;

			// Initial URL approach
			const baseUrl = '/start/account/user' + ( locale && locale !== 'en' ? `/${ locale }` : '' );

			return baseUrl + `?variationName=${ flowName }&redirect_to=${ redirectTarget }`;
```

daledupreez (approved):
Overall this is looking good to me, and tests well as an existing user and a new user signing up.

The only two gotchas I noticed are as follows:
1. We're defaulting the theme to `pub/lettre` - [code](https://github.com/Automattic/wp-calypso/blob/756ca1a8404bf407c1158d774d4f9ed86e2657b9/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx#L40) - this feels like some logic we need to pull out of the default step and have it explicitly injected in some way
2. We're creating the site as public/launched, when we want either Coming Soon or Private - [code](https://github.com/Automattic/wp-calypso/blob/756ca1a8404bf407c1158d774d4f9ed86e2657b9/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx#L43-L54) - this also feels like it's a situation where the data should be injected somehow.

I think we can tackle these items as follow-up PRs, however, as I feel other folks may be able to guide us as to how we can handle this kind of data within the framework.

markbiek:
I created issues #71730 and #71729 to tackle setting the theme and site launch status.

Issue #333, opened 2015-11-20
Editor: Media: gallery order is wrong in the edit details view


designsimply:
When I drag several images in order into the media library modal, the numbering is correct but the image order is all jumbled:

<img width="1280" alt="e6c5d6da-862f-11e5-90b6-caae789b3499" src="https://cloud.githubusercontent.com/assets/1119271/11311464/09342d62-8f9e-11e5-954f-2bbf058122ee.png">
<sup>Tested with Chrome 46.0.2490.80 on Mac OS X 10.10.5</sup>

In the edit details view, the images are displayed in the order they appear in the media library and not in the numbered order:

<img width="1280" alt="369f66da-8630-11e5-8952-5edbf05c8350" src="https://cloud.githubusercontent.com/assets/1119271/11311479/0e6a52f2-8f9e-11e5-89a1-d6b08d769af1.png">
<sup>This image should appear 6th (see screenshot above), but appears first here. Tested with Chrome 46.0.2490.80 on Mac OS X 10.10.5</sup>

Outside of these two views (i.e. in the preview/reorder tabs and in the post after the gallery is inserted), the order goes back to the expected numbered ordering.

aduth:
Upon further investigation, I've found that this is the expected behavior, albeit with some confusion around the shifting of the media order in the list view. The detail view order does not depend on the selected order of the items and is instead ordered by the upload date of the item. You can verify this with existing items by selecting an older item (further down the list), then a newer item, clicking Edit, and observing that the newer item is shown first despite being the second selected item. The shifting of the items in the list view during the upload occurs because the uploads may finish in an order not consistent with the temporary copy. While this may be visually disruptive, the original selected order shown should be preserved as the media items finish uploading.

ryanboren:
I accidentally cancel the modal often. I lose selection order when this happens. Since the images are uploaded in a jumble (instead of in the order they are dropped), I can't easily resurrect selection order like I can in wp-admin, which retains drop order. I reupload entire galleries to Calypso's modal rather than scrutinize images trying to restore the selection order by hand. A very frustrating and anxiety inducing exercise that happens to me with regularity.

Since the gallery flow doesn't offer full-width images during gallery edit, I try to caption using the Edit button from the media lib view so that I can get at bigger images and wider text inputs while captioning. Since images are shown in display (random) order instead of selection/upload order, captioning this way is pain. I've given up trying to caption and order galleries in Calypso.

ryanboren:
[...]

Issue #111, opened 2015-11-18
New post button changes when searching in page for "p" or "n"


hoverduck:
Steps to reproduce:

1. Go to https://wordpress.com/stats
2. Click CTRL+F to search for a keyword. Start typing "plugins" (I assume it's the 'p').
3. Close the search field (esc).
4. "new post" button now appears as seen below:

![image](https://cloud.githubusercontent.com/assets/7233112/11248533/949e5c84-8def-11e5-8b68-9cab187b4922.png)

This happens in both Chrome 44.0.2403.107 and Safari 8.0.7 on Yosemite. Firefox 39 is unaffected.

lancewillett:
I couldn't reproduce this, testing on https://wordpress.com/stats with a test user `lancetest35` in Mac-n-Chrome. No trouble in Safari 9.0.3 either (on OS X 10.11.3).

PR #71622, opened 2023-01-03
VideoPress Onboarding: Add style to choose a domain step filters

thedebian:
#### Proposed Changes

This PR add style to choose a domain step filters.
@evilluendas don't hesitate to give your opinion on this one, I can update accordingly, here is a screenshot: 
Capture d’écran 2023-01-03 à 11 29 48

#### Testing Instructions

* Apply PR
* Go to /start/videopress
* Follow the onboarding flow until choose a domain
* Click on "filter"
* ✅ The popup should be stylized

Fixes Automattic/greenhouse#1430

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/choose-a-domain/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/choose-a-domain/style.scss
@@ -299,14 +299,14 @@ $videopress-background: #010101;
 			.token-field__token-text {
 				background: none;
 				color: #fff;
-				border: 1px solid #ffffff70;
+				border: 1px solid rgba(255, 255, 255, 0.2);
 				border-right: none;
 			}
 
 			.token-field__remove-token {
 				background: none;
 				fill: #fff;
-				border: 1px solid #ffffff70;
+				border: 1px solid rgba(255, 255, 255, 0.2);
 				border-left: none;
 			}
 		}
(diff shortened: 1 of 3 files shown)


jgcaruso (approved):
Css changes look good and the filter component still works as expected. Filter look match the screenshot, but the mix of black and grey just looks a little "off" to my non-designer eye. Not sure what can be done though, it can always be tweaked later if necessary I suppose now that this base work to style it has been done.

evilluendas:
Nice! I think I'd make the background slightly darker than the boxes below but not as dark as the background (I'm using `#161616` in the screenshot), and then make the background of the list not as dark (`#111` here) with a lighter border (`#333`).

image

I'd also make the border of the token and the X inside a bit more transparent, and the text itself (`.com`) completely white. The border of the "Clear" button is also lighter, like the secondary buttons in the background.

It's a bit hard to see where the box ends when seen in a static screenshot, but I think it's more discernible when you open it yourself clicking the button. Otherwise, we can make the background brighter than the boxes and lift all the grays a bit so there's more contrast, or make the shadow behind brighter (although I don't like that, generally it makes UIs feel a bit muddy)... or even just leave the white background as it is by default.

Here's how it looks making the shadow brighter:

image

thedebian:
@evilluendas Here is how it looks like now
Capture d’écran 2023-01-16 à 09 58 24

thedebian:
@evilluendas Here is how it looks like now :

Capture d’écran 2023-01-16 à 10 09 31

evilluendas (approved):

PR #71623, opened 2023-01-03
E2E: Fix artifacts overwriting

WunderBart:
#### Proposed Changes

- Fix our E2E env setup teardown where artifacts are not being indexed correctly, ending up overriding each other,
- Change the indexing from 0- to 1-based for convenience (easier to tell how many contexts/pages were captured).

#### Testing Instructions

1. Create `test/e2e/specs/teardown-test.ts` with the following content:
   
   ```
   import { Browser } from 'playwright';
   
   declare const browser: Browser;
   
   describe( 'Teardown', () => {
   	it( 'creates artifacts for all contexts and', async () => {
   		const c1 = await browser.newContext();
   
   		await c1.newPage();
   		await c1.newPage();
   		await c1.newPage();
   
   		const c2 = await browser.newContext();
   
   		await c2.newPage();
   		await c2.newPage();
   		await c2.newPage();
   
   		expect( false ).toBe( true );
   	} );
   } );
   ```
2. `cd test/e2e`
3. `yarn jest specs/teardown-test.ts`
4. Open `test/e2e/results` folder and ensure that the created `teardown-test-*` folder contains artifacts for both contexts and their pages.
5. `git checkout trunk`
6. `yarn jest specs/teardown-test.ts`
7. Again, go to `test/e2e/results` and confirm that the last created `teardown-test-*` folder contains artifacts for only 1 context and 1 page.

### DIFF
--- a/packages/calypso-e2e/src/jest-playwright-config/environment.ts
+++ b/packages/calypso-e2e/src/jest-playwright-config/environment.ts
@@ -148,16 +148,17 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
 		}
 		const contexts = this.global.browser.contexts();
 		if ( this.failure ) {
-			let contextIndex = 0;
-			let pageIndex = 0;
+			let contextIndex = 1;
 
 			const artifactFilename = `${ this.testFilename }__${ sanitizeString( this.failure.name ) }`;
 
 			for await ( const context of contexts ) {
+				let pageIndex = 1;
 				const traceFilePath = path.join(
 					this.testArtifactsPath,
 					`${ artifactFilename }__${ contextIndex }.zip`
 				);
+
 				// Traces are saved per context.
 				await context.tracing.stop( { path: traceFilePath } );
 
--- a/packages/calypso-e2e/src/jest-playwright-config/environment.ts
+++ b/packages/calypso-e2e/src/jest-playwright-config/environment.ts
@@ -151,27 +151,29 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
 			let contextIndex = 0;
 			let pageIndex = 0;
 
-			// Define artifact filename templates.
 			const artifactFilename = `${ this.testFilename }__${ sanitizeString( this.failure.name ) }`;
-			const traceFilePath = path.join(
-				this.testArtifactsPath,
-				`${ artifactFilename }__${ contextIndex }.zip`
-			);
-			const mediaFilePath = path.join(
-				this.testArtifactsPath,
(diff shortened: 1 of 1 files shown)


dpasque (approved):
❤️ 🙇 🚢

packages/calypso-e2e/src/jest-playwright-config/environment.ts
```diff
@@ -148,30 +148,33 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
...
-				`${ artifactFilename }__${ contextIndex }.zip`
-			);
-			const mediaFilePath = path.join(
-				this.testArtifactsPath,
-				`${ artifactFilename }__${ contextIndex }-${ pageIndex }`
-			);
 
 			for await ( const context of contexts ) {
+				let pageIndex = 1;
+				const traceFilePath = path.join(
```
worldomonation, line 157: **comment**: IIRC I moved the declaration for these path templates out of the `for` loop to consolidate such declarations in one place for easier tracking. Is there a particular reason to move them back into separate locations?

Addendum: One reason I moved the filename template declarations out is because otherwise the `const` is declared ver and over again.

worldomonation (approved):
Looks fine from my end, just a comment/question I had.

packages/calypso-e2e/src/jest-playwright-config/environment.ts
WunderBart: > IIRC I moved the declaration for these path templates out of the for loop to consolidate such declarations in one place for easier tracking. Is there a particular reason to move them back into separate locations?

Taking those out of the `for` block was actually the cause of the bug that is fixed in this PR (artifacts overwriting)! The `traceFilePath` and `mediaFilePath` values need to be updated with each iteration because they depend on the `context` and `page` indexes, so this needs to happen **inside** the `for` block.

> One reason I moved the filename template declarations out is because otherwise the const is declared ver and over again.

Each iteration of the `for` loop is a separate block, so the constant declared inside is scoped only to that iteration. When redeclaring a constant, you'd definitely see the `Cannot redeclare block-scoped variable 'xxx'. ts(2451)` error!

worldomonation:
@WunderBart Do you plan to merge this soon? I ran into a situation where this fix would have been helpful 😛

WunderBart:
> @WunderBart Do you plan to merge this soon? I ran into a situation where this fix would have been helpful 😛

✅ Done!

Sorry for the delay - I was on a Support Rotation last week. 😅

PR #71625, opened 2023-01-03
BBE: Add initial store flow

aneeshd16:
#### Proposed Changes

PT: pdh1Xd-1Cz-p2

This PR adds a new flow `do-it-for-me-store`, which can be used to purchase BBE for store/eCommerce sites.
This is the first part of several changes that will follow soon and adds the barebones flow. Since this is a new flow, it's safe to merge.

#### Testing Instructions

* Go to `/start/do-it-for-me-store`.
* Confirm that the subheader copy mentions that the Business plan would be required.
image

* Select "New Site". Confirm that the next screen look like this:
image

- Go through the flow steps till you reach the page picker step.
- Confirm that the cart contains the BBE product and the Business plan.
- Go to checkout and confirm that the cart contains the BBE product and the Business plan.
- Complete the purchase. You'll be taken to the website content form.

#### Pre-merge Checklist

Fixes Automattic/martech#1373

### DIFF
--- a/client/lib/signup/step-actions/index.js
+++ b/client/lib/signup/step-actions/index.js
@@ -326,6 +326,9 @@ export function setThemeOnSite( callback, { siteSlug, themeSlugWithRepo } ) {
 
 function addDIFMLiteProductToCart( callback, dependencies, step, reduxStore ) {
 	const { selectedDesign, selectedSiteCategory, isLetUsChooseSelected, siteSlug } = dependencies;
+	if ( step.lastKnownFlow === 'do-it-for-me-store' ) {
+		dependencies.isStoreFlow = true;
+	}
 	const extra = buildDIFMCartExtrasObject( dependencies );
 	const cartItem = {
 		product_slug: WPCOM_DIFM_LITE,
--- a/client/signup/config/flows-pure.js
+++ b/client/signup/config/flows-pure.js
@@ -480,6 +480,23 @@ export function generateFlows( {
 			enableBranchSteps: true,
 		},
 		{
+			name: 'do-it-for-me-store',
+			steps: [
+				'user',
+				'new-or-existing-site',
+				'difm-site-picker',
+				'difm-store-options',
+				'social-profiles',
+				'difm-design-setup-site',
+				'difm-page-picker',
+			],
+			destination: getDIFMSignupDestination,
+			description: 'The BBE store flow',
+			excludeFromManageSiteFlows: true,
+			lastModified: '2023-03-01',
+			enableBranchSteps: true,
+		},
+		{
 			name: 'website-design-services',
 			steps: [ 'difm-options', 'social-profiles', 'difm-design-setup-site', 'difm-page-picker' ],
 			destination: getDIFMSignupDestination,
--- a/client/signup/config/step-components.js
+++ b/client/signup/config/step-components.js
@@ -78,6 +78,7 @@ const stepNameToModuleName = {
 	'difm-site-picker': 'difm-site-picker',
 	'difm-design-setup-site': 'design-picker',
 	'difm-options': 'site-options',
+	'difm-store-options': 'site-options',
 	'difm-page-picker': 'page-picker',
 	'website-content': 'website-content',
 	intent: 'intent',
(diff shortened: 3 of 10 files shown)


southp:
Going through the flow, the Premium plan is added for me, instead of the Business plan:

https://user-images.githubusercontent.com/1842898/210949940-65f38263-94bc-4495-bf8b-ad15bbfd740d.mov

@aneeshd16 Could you help me confirm if it's expected or it will be in one of the follow-up PRs? I don't see the code changing the bundled plan into the Business plan here, hence asking :)

client/lib/signup/step-actions/index.js
```diff
@@ -326,6 +326,9 @@ export function setThemeOnSite( callback, { siteSlug, themeSlugWithRepo } ) {
 
 function addDIFMLiteProductToCart( callback, dependencies, step, reduxStore ) {
 	const { selectedDesign, selectedSiteCategory, isLetUsChooseSelected, siteSlug } = dependencies;
+	if ( step.lastKnownFlow === 'do-it-for-me-store' ) {
```
southp, line 329: I know we have been using flow names as a way to introduce conditional logics like this quite pervasively, but in my opinion it's actually in general not a good idea, since it'd basically defeat the purpose of modularity: a flow is meant to be an assembly of the underlying components, like step components, actions, etc. By deriving condtions by the flow name, we are introducing cyclic dependency between the assembly and its components, which means the "component" won't act the same across all "assemblies". Imagine a LEGO brick doesn't act the same depending on being used on a castle or a boat :)  

Can we move away from depending on the flow name here? There are various options, e.g. a new `difm-design-setup-site-store` step tied with a new step action that populates the `isStoreFlow` dependency, or we can use the `prop` mechanism([example](https://github.com/Automattic/wp-calypso/blob/trunk/client/signup/config/steps-pure.js#L146)) to inject `isStoreFlow` which then get passed to this step action somehow, etc. I'll defer to your best judgement :)

aneeshd16:
> Going through the flow, the Premium plan is added for me, instead of the Business plan:
> @aneeshd16 Could you help me confirm if it's expected or it will be in one of the follow-up PRs? I don't see the code changing the bundled plan into the Business plan here, hence asking :)

Thanks for the review @southp! There are two different carts here:
* Cart on page picker step - this is handled in a follow-up PR: #71627 
* Cart on the checkout page - This was caused due to a backend bug. The fix for that is in D97209-code.

client/lib/signup/step-actions/index.js
aneeshd16: I'm glad that you brought this up! I completely agree with you on this, and this makes sense in the context of this signup framework. In fact, in the previous PR I had started with separate steps for the store flow, which I later removed via https://github.com/Automattic/wp-calypso/pull/71320/commits/f056abc63e7a7f84b9a6f16cb485aad5ef485ff7.

The reasoning behind this was that these steps would eventually have to be migrated to the Stepper framework for various reasons - eliminating loaders, reusing existing steps in Stepper, etc. I noticed that the pattern in the Stepper framework was to change behaviour based on the flow name, for example:
https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/intro/index.tsx
I could be wrong, and there could be a better way to do this in Stepper, so I appreciate your inputs on that as well. :)

southp: Thanks for the context. I personally think it's one major flaw of Stepper that needs to be fixed, and I should stop postponing writing my proposal 😓 

In that case, please feel free to keep it this way before a better way is implemented :)

southp (approved):
Thanks for the explanation. No concern from me then :) Let's 🚢  this.

client/lib/signup/step-actions/index.js
ddc22: Agreeing aggressively 😬 . And looking at this from a big picture standpoint, The concept of dependency injection was one of the key things that made the signup framework so elegant. That concept should have been a hard requirement when implementing the stepper framework. It will be hard to implement though with branching. typescript interfaces could possibly be used to hide some complexity here.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7725466

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7725466).

PR #71644, opened 2023-01-03
Fix warning when running e2e decrypt

noahtallen:
#### Proposed Changes

When running the e2e decrypt command, a message is printed when the secrets file doesn't exist. (Just from `rm $file` saying there's an issue when $file doesn't exist.) It doesn't impact functionality, but it made me second guess whether the command worked! An easy fix is `rm -f $file`, which doesn't show an error if the file doesn't exist.

#### Testing Instructions
1. If it exists already, delete `./src/secrets/decrypted-secrets.json`
2. Run `yarn workspace wp-e2e-tests decrypt-secrets`.
3. There shouldn't be an error about the secrets file missing. (Though if you don't have the secret set up, the next command might fail. That's unimportant for this PR) On trunk, an error is printed. (Which you could quickly check by checking on trunk and running the command again.)


worldomonation (approved):
Thanks for the change - in the initial implementation I had it as `rm -f` but decided I didn't need the `-f` flag as I always have the file on disk.

Issue #70609, opened 2022-11-30
Stats polish: Posting activity UI


jeffgolenski:
# Development

### Must have:
- [x] Restyle container and header.
  - [x] Remove card borders.
  - [x] Enlarge the title and change font to match the headers on the Traffic page.

- [x] Restyle "day" squares within each month. 
  - [x] Add 2px border-radius to each square.
  - [x] Increase size of each day square to be 12x12px. Padding between each should be 4px.

- [x] Restyle the chart legend ("Fewer ■■■■■ More Posts").
  - [x] Should be centered and enlarged.
  - [x] Remove padding between squares in this area.
  - [x] Add 5px rounded corners.

- [x] Orient each month square horizontally instead of vertically. [See Filipe's comment](https://github.com/Automattic/wp-calypso/issues/70609#issuecomment-1333620075).

- [x] Make the interface more responsive.
  - [x] Show the posting activity section in a mobile viewport.
  - [x] Reduce the number of months shown depending on the viewport width.

### Nice to have:

- [ ] Improve styling when hovering over a day square.
  - [x] Display the date, day of the week, and amount of posts published in the hover tooltip.
  - [x] Increase the size of the hovered "day" square to 14x14 px, if possible.
  - [ ] Drop all other day squares' opacity to 50% when a day square is being hovered.

- [x] ~~Add arrow navigation so users can scan back past 12 months.~~ (This is no longer a valid task -@jsnmoon)

Must-have items must be completed before closing this issue. Please spin off any unfinished tasks into separate issues.

# Design

### Details

Just making note of a few refinements to make on the insights graphs. Nothing major, just some visual polish. Breaking down each section for easier consumption

### **Posting activity**

**Current**
![Screen Shot 2022-11-30 at 5 22 49 PM](https://user-images.githubusercontent.com/214813/204921058-2a41e69d-d150-403d-aa73-884ff73e83bf.png)

**After:**
<img width="1792" alt="Post activity" src="https://user-images.githubusercontent.com/214813/204922717-c9f1e167-1282-40f4-8ac0-8efde24e1d45.png">

![Screen Shot 2022-11-30 at 5 40 00 PM](https://user-images.githubusercontent.com/214813/204923706-7b259bbc-057d-411e-a8f6-b394002ce125.png)

In context:
[...]

PR #71676, opened 2023-01-04
Stats: fix some details on the Posting Activity section

wilmersondasilva:
#### Proposed Changes

* Fix the horizontal margins on mobile
* Refactor the CSS by moving the tooltip CSS rules to a more appropriate CSS file

#### Testing Instructions

* Go to the Calypso live link
* Navigate to the Insights tab
* Ensure the Posting Activity section has a horizontal margin on mobile
* Ensure the tooltip on the Posting Activity looks the same as before (see the screenshot below)

#### Screenshots

image
image

#### Pre-merge Checklist

Related to #70609


client/components/tooltip/style.scss
```diff
@@ -1,3 +1,7 @@
+@import "@automattic/typography/styles/variables";
+
+$post-trends-box-shadow-color: #0000000d;
```
grzegorz-cp: Do you think that using rgba value here would be easier to understand?

client/components/tooltip/style.scss
```diff
@@ -196,4 +200,45 @@
 			background: var(--studio-gray-100);
 		}
 	}
+
+	&.post-trends-day__tooltip {
```
grzegorz-cp: Do you think this is a better place? We are putting a specific CSS class (for one component) in a generic component. 
I had a similar problem with the annual highlight popover. If we need the same styles, maybe it would be better to create another variant for the tooltip (and avoid components specific classes in favour of something like a tooltip--dark)?

client/components/tooltip/style.scss
wilmersondasilva: Yes, I do. I'll change it

client/components/tooltip/style.scss
wilmersondasilva: What did you do in the situation you mentioned? Can I take a look at your PR?

grzegorz-cp: I kept most of the CSS specific to my component in that component's SCSS file. That's what you did initially here.
Here's the PR: https://github.com/Automattic/wp-calypso/pull/71159/files 
I think we share some of the styles, so maybe we should extract them into yet another component.
If you've got a different idea, we can talk about it, but if you'd like to follow this one, it might be better to do it in another PR. What do you think?

wilmersondasilva: @grzegorz-cp got it. I like this idea. I'll create the variant for the tooltip on this PR, and regarding extracting them into another component, I agree that we can do it in another PR.

client/components/tooltip/style.scss
```diff
@@ -1,3 +1,7 @@
+@import "@automattic/typography/styles/variables";
+
+$post-trends-box-shadow-color: rgba(0, 0, 0, 0.05);
```
jsnmoon:
```suggestion
$large-tooltip-box-shadow-color: rgba(0, 0, 0, 0.05);
```

client/components/tooltip/style.scss
```diff
@@ -1,3 +1,7 @@
+@import "@automattic/typography/styles/variables";
```
jsnmoon: Where is this import being used within this file? If we're not using this, then let's omit this change from the PR.

jsnmoon (approved):
Changes look reasonable; I left two nitty pieces of feedback inline.

client/components/tooltip/style.scss
wilmersondasilva: Thanks, Jason!

client/components/tooltip/style.scss
wilmersondasilva: Well observed! Thanks!

grzegorz-cp:
Looks good. Thanks for making the changes.

PR #71682, opened 2023-01-04
Adding new Twitter tracking pixel conversion events

donlair:
#### Proposed Changes

This PR adds support for two new Twitter Ads conversion events (`New Signup Complete` (ID: tw-nvzbs-ode0f) and `New Purchase` (ID: tw-nvzbs-ode0u)). P2 discussion at pau2Xa-4Im-2.

Here's the generated tracking code for one of the events:

```
<!-- Twitter conversion tracking event code -->
<script type="text/javascript">
  // Insert Twitter Event ID
  twq('event', 'tw-nvzbs-ode0f', {
    value: null, // use this to pass the value of the conversion (e.g. 5.00)
    currency: null, // use this to pass the currency of the conversion with an ISO 4217 code (e.g. ‘USD’)
    conversion_id: null // use this to pass a unique ID for the conversion event for deduplication (e.g. order id '1a2b3c')
  });
</script>
<!-- End Twitter conversion tracking event code -->
```

I created new tracking events for:
- Page views (event id: `tw-nvzbs-odfz9`)
- Registration complete (user step) (event id: `tw-nvzbs-odfz8`)
- Signup complete (event id: `tw-nvzbs-ode0f`)
- Purchase (event id: `tw-nvzbs-ode0u`)

I changed the retargeting event to use this new event so that everything is consistently using the new events.

#### Testing Instructions
* Check out this PR and set `ad-tracking` to true in your development.json config file
* Sandbox store so that you can place a credit card order
* Download the Twitter Pixel Helper Chrome extension
* Start Calypso locally
* Navigate through signup and confirm that the events are firing as expected as you progress through the signup flow.
* Confirm in the Twitter admin that the conversion events were captured.

Related to #

### DIFF
--- a/client/lib/analytics/ad-tracking/ad-track-registration.js
+++ b/client/lib/analytics/ad-tracking/ad-track-registration.js
@@ -61,5 +61,13 @@ export async function adTrackRegistration() {
 		window.pintrk( ...params );
 	}
 
+	// Twitter
+
+	if ( mayWeTrackByTracker( 'pinterest' ) ) {
+		const params = [ 'event', 'tw-nvzbs-odfz8' ];
+		debug( 'adTrackRegistration: [Twitter]', params );
+		window.twq( ...params );
+	}
+
 	debug( 'adTrackRegistration: dataLayer:', JSON.stringify( window.dataLayer, null, 2 ) );
 }
--- a/client/lib/analytics/ad-tracking/ad-track-registration.js
+++ b/client/lib/analytics/ad-tracking/ad-track-registration.js
@@ -63,7 +63,7 @@ export async function adTrackRegistration() {
 
 	// Twitter
 
-	if ( mayWeTrackByTracker( 'pinterest' ) ) {
+	if ( mayWeTrackByTracker( 'twitter' ) ) {
 		const params = [ 'event', 'tw-nvzbs-odfz8' ];
 		debug( 'adTrackRegistration: [Twitter]', params );
 		window.twq( ...params );
--- a/client/lib/analytics/ad-tracking/ad-track-signup-complete.js
+++ b/client/lib/analytics/ad-tracking/ad-track-signup-complete.js
@@ -159,15 +159,7 @@ export async function adTrackSignupComplete( { isNewUserSite } ) {
 	// Twitter
 
 	if ( mayWeTrackByTracker( 'twitter' ) ) {
-		const params = [
-			'event',
-			'tw-nvzbs-ode0f',
-			{
-				value: syntheticCart.total_cost,
-				currency: syntheticCart.currency,
-				conversion_id: syntheticOrderId,
-			},
-		];
+		const params = [ 'event', 'tw-nvzbs-ode0f' ];
 		debug( 'recordSignup: [Twitter]', params );
 		window.twq( ...params );
 	}
--- a/client/lib/analytics/ad-tracking/ad-track-signup-complete.js
+++ b/client/lib/analytics/ad-tracking/ad-track-signup-complete.js
@@ -159,7 +159,15 @@ export async function adTrackSignupComplete( { isNewUserSite } ) {
 	// Twitter
 
 	if ( mayWeTrackByTracker( 'twitter' ) ) {
-		const params = [ 'track', 'Signup', {} ];
+		const params = [
+			'event',
+			'tw-nvzbs-ode0f',
+			{
+				value: syntheticCart.total_cost,
+				currency: syntheticCart.currency,
+				conversion_id: syntheticOrderId,
+			},
+		];
 		debug( 'recordSignup: [Twitter]', params );
 		window.twq( ...params );
 	}
--- a/client/lib/analytics/ad-tracking/record-order.js
+++ b/client/lib/analytics/ad-tracking/record-order.js
@@ -96,15 +96,7 @@ export async function recordOrder( cart, orderId ) {
 
 	// Twitter
 	if ( mayWeTrackByTracker( 'twitter' ) ) {
-		const params = [
-			'event',
-			'tw-nvzbs-ode0u',
-			{
-				value: cart.total_cost.toString(),
-				currency: cart.currency,
-				order_id: orderId,
-			},
-		];
+		const params = [ 'event', 'tw-nvzbs-ode0u', { value: usdTotalCost } ];
 		debug( 'recordOrder: [Twitter]', params );
 		window.twq( ...params );
 	}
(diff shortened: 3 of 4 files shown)


gmovr:
As for testing, when working on this locally you can set the `ad-tracking` flag to `true` in development.json. It can be set via URL as well, but for me it's more reliable to set it in the config file. Once that's done, you should start seeing calls to t.co for various events.

If you go through the signup/purchase process, you'll be able to see if the URLs called are of the expected format. I'll take a look at the docs for the new pixel, I'm not too familiar with the format.

donlair:
@gmovr I updated this and confirmed that all but the new purchase events were firing. I'm guessing the purchase events are fired in the backend but I haven't had a chance to explore that yet.

donlair:
@gmovr I played with this more and it turns out the purchase events weren't firing because I was paying with credits. They fired as expected when I switched to the store sandbox:

CleanShot 2023-01-12 at 07 45 06@2x

All test conversions were detected by Twitter:
CleanShot 2023-01-12 at 07 46 53@2x

client/lib/analytics/ad-tracking/ad-track-registration.js
```diff
@@ -61,5 +61,13 @@ export async function adTrackRegistration() {
 		window.pintrk( ...params );
 	}
 
+	// Twitter
+
+	if ( mayWeTrackByTracker( 'pinterest' ) ) {
```
gmovr:
```suggestion
	if ( mayWeTrackByTracker( 'twitter' ) ) {
```

client/lib/analytics/ad-tracking/ad-track-signup-complete.js
```diff
@@ -159,7 +159,7 @@ export async function adTrackSignupComplete( { isNewUserSite } ) {
 	// Twitter
 
 	if ( mayWeTrackByTracker( 'twitter' ) ) {
-		const params = [ 'track', 'Signup', {} ];
+		const params = [ 'event', 'tw-nvzbs-ode0f' ];
```
gmovr, line 162: If we expect this to change, maybe we can move it out to a constant, like we do with Google. If not, this is fine.

gmovr (changes requested):
A minor change reqeusted (a typo I think :) ).

And just to clarify, we're not interested in the "old" twitter tracking anymore? I see it takes in more data, e.g. the cart, while the new one is simpler and only takes in the total cart value. Just to confirm.

I'm also not familiar with the new Twitter pixel, if it requires any changes on the actual script loaded. For reference, we load the twitter script here https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/analytics/ad-tracking/load-tracking-scripts.js#L79

And then initialize the queue here: https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/analytics/ad-tracking/load-tracking-scripts.js#L121

I'll read the docs on the new pixel, and take this for a spin as well.

client/lib/analytics/ad-tracking/ad-track-signup-complete.js
donlair: The docs didn't mention any upcoming changes, so I think this should be stable

donlair:
Re: the pixel, there is one change I wasn't sure about. [Here are the docs](https://business.twitter.com/en/help/campaign-measurement-and-analytics/conversion-tracking-for-websites.html).

To save you some clicks, here's the full pixel code:

```
<!-- Twitter conversion tracking base code -->
<script>
!function(e,t,n,s,u,a){e.twq||(s=e.twq=function(){s.exe?s.exe.apply(s,arguments):s.queue.push(arguments);
},s.version='1.1',s.queue=[],u=t.createElement(n),u.async=!0,u.src='https://static.ads-twitter.com/uwt.js',
a=t.getElementsByTagName(n)[0],a.parentNode.insertBefore(u,a))}(window,document,'script');
twq('config','nvzbs');
</script>
<!-- End Twitter conversion tracking base code -->
```

Looks like the param has changed to `config` from `init`. I played with that at one point while debugging a separate issue, but the Twitter Helper extension seemed happy either way. The only thing it was really unhappy about was using the older `track` param for conversions.

One other potential issue I noticed is that the pageview event doesn't get recorded on pages where there's another conversion as well. For example, on registration success, the registration gets recorded but the pageview event gets dropped. Seems like it only wants one conversion event fired on page load. Not a big deal afaict since the more important conversion was recorded in all of my test cases. It's potentially flaky though.

donlair:
Oof, I screwed up that rebase

gmovr (approved):
I'm seeing the events firing, and I can't see any errors. It's a bit hard to debug these tracking events without having some data, so I'd say we merge it, and we can follow up with changes if there are any needed for the base pixel.

```js
localStorage.setItem( 'debug', 'calypso:analytics:*' );
undefined

16909.e3bcdad3f610c1eb9c8f.min.js:2 calypso:analytics:ad-tracking loadTrackingScripts: [Loaded] +92ms https://static.ads-twitter.com/uwt.js
16909.e3bcdad3f610c1eb9c8f.min.js:2 calypso:analytics:ad-tracking retarget: [Twitter] [rate limited] +4ms (2) ['event', 'tw-nvzbs-odfz9']0: "event"1: "tw-nvzbs-odfz9"length: 2[[Prototype]]: Array(0)
16909.e3bcdad3f610c1eb9c8f.min.js:2 calypso:analytics:ad-tracking recordOrder: [Twitter] +0ms (3) ['event', 'tw-nvzbs-ode0u', {…}]0: "event"1: "tw-nvzbs-ode0u"2: {value: '58.080'}length: 3[[Prototype]]: Array(0)
```

I'll approve this, we can defer any follow-ups to a new PR, so we can get conversion tracking live and unblock that.

==== Other thoughts ====

I imagine the reason why you're not seeing all page views, is because the Twitter pixel is placed in the rate limited block in `retarget.js`. 

It's wrapped in a conditional here:
```
if ( nowTimestamp >= lastRetargetTime + retargetingPeriod ) {
		lastRetargetTime = nowTimestamp;
                // Tracking callbacks here incl. Twitter.
```
While Google Ads, Facebook, etc. are placed outside this block, and will fire on all page views. I'm not sure if there's a historical reason for this, but we can test, and possibly follow up with another PR to move the retargeting/page view to not be rate limited. 

This retargeting function is triggered from `recordPageView` in `page-view.js`, so it's triggered on each page view, but then we only enter that block whenever the `retargetingPeriod` has lapsed.  

The Twitter code here for page views also uses the old `track` argument, but it seems to work fine.

PR #71688, opened 2023-01-04
Launchpad: Enable Site Editor redirect from Launchpad iframe

### LINKED ISSUE #71329: LaunchPad Preview - "On Click" to enter site editor
Lets make the launchpad preview load into the site editor on click:

Screen Shot 2022-12-19 at 10 17 56 AM

Open questions to raise in team meeting today:
* Should newsletters also have this functionality?
* Should there be a confirmation modal or anything else before redirecting?

sharpirate:
#### Proposed Changes
This PR adds an overlay with a button over the Launchpad's web preview which is displayed on hover to allow going directly to the Site Editor.

#### Testing Instructions
1. Checkout this branch or use Calypso Live
2. Go to Launchpad
3. Verify that once you hover over the web preview, the overlay is displayed and clicking the "Edit" button takes you to the Site Editor and you can go back to Launchpad. Verify you can interact with the overlay using the keyboard
4. Verify that the styling looks good. The CSS changes in this PR are significant so I'd appreciate a thorough test:
• Verify that everything looks good when switching between the phone and the desktop preview using the buttons above the web preview
• Verify that shrinking the browser to mobile size looks good both when using the desktop and phone previews
• In general pay attention to the styling
5. We should also test another instance of the WebPreview component to make sure these changes won't affect it. You can go through the Free Flow and test at the design picker step.

https://user-images.githubusercontent.com/20927667/212164158-baf7de7d-5fbb-4077-a18d-b0f8cc34bd66.mov

#### Pre-merge Checklist

Fixes #71329

### DIFF
--- a/client/components/web-preview/component.jsx
+++ b/client/components/web-preview/component.jsx
@@ -63,6 +63,8 @@ export class WebPreviewModal extends Component {
 		fixedViewportWidth: PropTypes.number,
 		// Prevents tabbing into the iframe.
 		disableTabbing: PropTypes.bool,
+		// Allow specific functionality for Launchpad
+		isLaunchpad: PropTypes.bool,
 	};
 
 	static defaultProps = {
--- a/client/components/web-preview/component.jsx
+++ b/client/components/web-preview/component.jsx
@@ -63,8 +63,8 @@ export class WebPreviewModal extends Component {
 		fixedViewportWidth: PropTypes.number,
 		// Prevents tabbing into the iframe.
 		disableTabbing: PropTypes.bool,
-		// Allow specific functionality for Launchpad
-		isLaunchpad: PropTypes.bool,
+		// Edit overlay that redirects to the Site Editor
+		enableEditOverlay: PropTypes.bool,
 	};
 
 	static defaultProps = {
@@ -83,7 +83,7 @@ export class WebPreviewModal extends Component {
 		hasSidebar: false,
 		overridePost: null,
 		autoHeight: false,
-		isLaunchpad: false,
+		enableEditOverlay: false,
 	};
 
 	constructor( props ) {
--- a/client/components/web-preview/component.jsx
+++ b/client/components/web-preview/component.jsx
@@ -83,6 +83,7 @@ export class WebPreviewModal extends Component {
 		hasSidebar: false,
 		overridePost: null,
 		autoHeight: false,
+		isLaunchpad: false,
 	};
 
 	constructor( props ) {
--- a/client/components/web-preview/content.jsx
+++ b/client/components/web-preview/content.jsx
@@ -31,6 +31,7 @@ export default class WebPreviewContent extends Component {
 		loaded: false,
 		isLoadingSubpage: false,
 		isMobile: isWithinBreakpoint( '<660px' ),
+		showIFrameOverlay: false,
 	};
 
 	setIframeInstance = ( ref ) => {
@@ -410,6 +411,7 @@ export default class WebPreviewContent extends Component {
 								'is-resizable': ! this.props.isModalWindow,
 							} ) }
 						>
+							{ /* eslint-disable jsx-a11y/mouse-events-have-key-events */ }
 							<iframe
 								ref={ this.setIframeInstance }
 								className="web-preview__frame"
@@ -420,7 +422,23 @@ export default class WebPreviewContent extends Component {
 								fetchpriority={ fetchpriority ? fetchpriority : undefined }
 								scrolling={ autoHeight ? 'no' : undefined }
 								tabIndex={ disableTabbing ? -1 : 0 }
+								onMouseOver={ () => this.setState( { showIFrameOverlay: true } ) }
+								onMouseOut={ () =>
+									setTimeout( () => this.setState( { showIFrameOverlay: false } ), 150 )
+								}
 							/>
+							{ this.state.showIFrameOverlay && (
+								<div className="web-preview__frame-edit-overlay">
+									<button
+										className="web-preview__frame-edit-button"
+										onClick={ () => {
+											window.location.assign( `/site-editor/${ this.props.externalUrl }` );
+										} }
+									>
+										Edit
+									</button>
+								</div>
+							) }
 						</div>
 					) }
 					{ 'seo' === this.state.device && (
--- a/client/components/web-preview/content.jsx
+++ b/client/components/web-preview/content.jsx
@@ -427,7 +427,7 @@ export default class WebPreviewContent extends Component {
 								style={ {
 									...this.state.iframeStyle,
 									height: this.state.viewport?.height,
-									pointerEvents: this.props.isLaunchpad ? 'none' : 'all',
+									pointerEvents: this.props.isLaunchpad ? 'auto' : 'all',
 								} }
 								src="about:blank"
 								onLoad={ () => this.setLoaded( 'iframe-onload' ) }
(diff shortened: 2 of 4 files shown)


jeyip:
@Sharpirate What kind of behavior do we expect with keyboard navigation? Should the user be able to see the overlay by using the tab key? Currently it skips the overlay and edit button.

sharpirate:
@jeyip That's a great catch! Tbh I'm not quite sure, maybe @nuriapenya can provide additional details here

jeyip:
## Testing
### Requirements
- [x] Hovering over the web preview shows an overlay with an "Edit" button 
- [x] Clicking the "Edit" button takes you to the Site Editor
- [x] You can navigate back to the Launchpad using the WordPress icon in the editor ( Top left hand corner of screen )
- [x] Shrinking the browser to mobile size looks good both when using the desktop and phone previews
- [ ] Styling ( See below )

### Browsers
- [x] Edge
- [x] Firefox
- [x] Chrome

**Note:** Seeing some wonky loading issues in Firefox in calypso live, but I think it has to do with wonky caching of my credentials. I'll try and check again later.

Edit button isn't vertically centered in the _desktop_ iframe preview overlay.

Screen Shot 2023-01-06 at 2 15 02 PM

Non-blocking, but wondering if we want the "Edit" button to fade in from below like in the themes selection overlays. Do we wanna chat with design to get a sense of how to make things consistent with existing conventions?

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,40 @@ export default class WebPreviewContent extends Component {
...
 						>
+							{ /* eslint-disable jsx-a11y/mouse-events-have-key-events */ }
 							<iframe
 								ref={ this.setIframeInstance }
 								className="web-preview__frame"
-								style={ { ...this.state.iframeStyle, height: this.state.viewport?.height } }
+								style={ {
+									...this.state.iframeStyle,
+									height: this.state.viewport?.height,
+									pointerEvents: 'none',
```
jeyip: Setting `pointerEvents` to `none` might be a pretty big change. Off the top of my head, I can't see why this would be a problem, but do we need to cross check some other implementations of the web preview to make sure nothing is relying on this functionality?

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,40 @@ export default class WebPreviewContent extends Component {
 					) }
 					{ 'seo' !== this.state.device && (
 						<div
+							onMouseEnter={ () => this.setState( { showIFrameOverlay: true } ) }
```
jeyip: Since this isn't gated behind any conditionals, does that mean that all implementations of web preview will now have an iframe overlay?

jeyip: 9wpoYJrMlYKAKUYvVBMN74-fi?t=LSr3Run0HrsvQdqa-0

client/components/web-preview/content.jsx
sharpirate: That's a good point, maybe we can set the pointer-events conditionally only for Launchpad just to be safe?

client/components/web-preview/content.jsx
sharpirate: That's true, I think I should do a check for Launchpad for all of these changes

sharpirate:
@jeyip 
>Edit button isn't vertically centered in the desktop iframe preview overlay.

I wasn't able to reproduce that one, would you be able to record a video / GIF?

sharpirate:
I've addressed the comments about adding a check for Launchpad

nuriapenya:
@Sharpirate @jeyip 

>What kind of behavior do we expect with keyboard navigation? Should the user be able to see the overlay by using the tab key? Currently it skips the overlay and edit button.

Ideally, yes. Keyboard users should be able to toggle between Desktop <> Mobile and have a way to activate the overlay, so they can access the `Edit` button.

I'm unsure how we can do that technically (maybe by adding a state or an aria role? that indicates there's something inside the button).

>Non-blocking, but wondering if we want the "Edit" button to fade in from below like in the themes selection overlays. Do we wanna chat with design to get a sense of how to make things consistent with existing conventions?

Agree 💯 I would suggest keeping the opacity of the overlay as you @Sharpirate already built it, darker than in themes, but copying the exact animation/transition if possible.

sharpirate:
Thanks, @nuriapenya, I've added keyboard navigation support (using the Tab key) and added the transition effect.

jeyip:
Seeing a coverage issue with the desktop overlay 🤔

client/components/web-preview/component.jsx
```diff
@@ -63,6 +63,8 @@ export class WebPreviewModal extends Component {
 		fixedViewportWidth: PropTypes.number,
 		// Prevents tabbing into the iframe.
 		disableTabbing: PropTypes.bool,
+		// Allow specific functionality for Launchpad
+		isLaunchpad: PropTypes.bool,
```
edanzer: I see we're adding the isLaunchpad prop to ensure changes in this PR only apply to Launchpad. 

Should we also set the default to false in the defaultProps below?

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,61 @@ export default class WebPreviewContent extends Component {
...
 							} ) }
 						>
 							<iframe
 								ref={ this.setIframeInstance }
 								className="web-preview__frame"
-								style={ { ...this.state.iframeStyle, height: this.state.viewport?.height } }
+								style={ {
+									...this.state.iframeStyle,
+									height: this.state.viewport?.height,
+									pointerEvents: this.props.isLaunchpad ? 'none' : 'all',
```
edanzer: For pointerEvents, should the 'all' here be 'auto'?

client/components/web-preview/content.jsx
```diff
@@ -507,6 +548,10 @@ WebPreviewContent.propTypes = {
 	inlineCss: PropTypes.string,
 	// Uses the CSS selector to scroll to it
 	scrollToSelector: PropTypes.string,
+	// Allow specific functionality for Launchpad
+	isLaunchpad: PropTypes.bool,
+	// Show Edit button on hover
+	showIFrameOverlay: PropTypes.bool,
```
edanzer: Wondering if we should provide default values for these in defaultProps below?

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,61 @@ export default class WebPreviewContent extends Component {
...
+									pointerEvents: this.props.isLaunchpad ? 'none' : 'all',
+								} }
 								src="about:blank"
 								onLoad={ () => this.setLoaded( 'iframe-onload' ) }
 								title={ this.props.iframeTitle || translate( 'Preview' ) }
 								fetchpriority={ fetchpriority ? fetchpriority : undefined }
 								scrolling={ autoHeight ? 'no' : undefined }
 								tabIndex={ disableTabbing ? -1 : 0 }
 							/>
+							<div
```
edanzer: The way we're doing this seems like it should be fine, but I wonder if it would be worth considering making the render of this entire div/overlay conditional? Along the lines of { isLaunchpad && <div.... /> }. 

It seems like for non launchpad situations, we don't need the extra markup on the page. Seems also relevant because all CSS related to this div is actually in the Launchpad scss and not the web preview scss.

sharpirate:
I've addressed Erick's comments and the styling bug we found should be fixed now

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,63 @@ export default class WebPreviewContent extends Component {
...
+									pointerEvents: this.props.isLaunchpad ? 'auto' : 'all',
+								} }
 								src="about:blank"
 								onLoad={ () => this.setLoaded( 'iframe-onload' ) }
 								title={ this.props.iframeTitle || translate( 'Preview' ) }
 								fetchpriority={ fetchpriority ? fetchpriority : undefined }
 								scrolling={ autoHeight ? 'no' : undefined }
 								tabIndex={ disableTabbing ? -1 : 0 }
 							/>
+							{ this.props.isLaunchpad && (
```
jeyip: `NB:` Giving this more thought, it seems like other features beside Launchpad might eventually want access to this functionality. Rather then require toggling by feature, which would lead to consistently needing to update this file, we probably just want a prop like `enableOverlay` which is passed in through the consumer

client/components/web-preview/content.jsx
```diff
@@ -406,21 +407,63 @@ export default class WebPreviewContent extends Component {
...
+											transition: 'all 0.2s ease',
+										} }
+										className="web-preview__frame-edit-button"
+										onClick={ () => {
+											window.location.assign( `/site-editor/${ this.props.externalUrl }` );
+										} }
+										onFocus={ () => this.setState( { showIFrameOverlay: true } ) }
+										onBlur={ () => this.setState( { showIFrameOverlay: false } ) }
+									>
+										{ translate( 'Edit' ) }
```
jeyip, line 464: Currently, my mac screen reader only dictates this button as the "Edit" button which might not be the best experience for sight challenged users.

Maybe we could consider changing the text of the button itself to "Edit Site"? Otherwise, we could add something like `aria-label="Edit your new site"`

jeyip: Feel free to handle this in a follow-up if you prefer

jeyip:
## Testing
### Requirements
- [x] Hovering over the web preview shows an overlay with an "Edit" button 
- [x] Clicking the "Edit" button takes you to the Site Editor
- [x] You can navigate back to the Launchpad using the WordPress icon in the editor ( Top left hand corner of screen )
- [x] Shrinking the browser to mobile size looks good both when using the desktop and phone previews 


### Browsers
- [x] Edge
- [x] Firefox
- [x] Chrome

**Note:** The site editor keeps breaking out of the iframe into `wp-admin` in Firefox. Unsure why, but it doesn't appear related to this PR

jeyip (approved):

client/components/web-preview/content.jsx
jeyip: For reference https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8

PR #48762, opened 2021-01-11
update: account password data

flootr:
#### Changes proposed in this Pull Request

* Remove the `AccountPasswordData` event emitter
* Split up `AccountPasswordData` and create a general `generatePassword` utility function
* Move password validation to `me/account-password`
* Make `AccountPassword` a class and remove `createReactClass` 

#### Testing instructions

* Go to `/me/security/password` or `/me/security` and choose _Password_
* Make sure proper validation messages show up in the following cases:
  * Type in a weak password (i.e. doesn't match the password rules)
  * Type in a password you used previously
  * Type in a strong password (i.e. does match the password rules)
* Use the _Generate strong password_ button and make sure it works as expected

fixes #48404

PR #71704, opened 2023-01-05
lib: make `generatePassword` cryptographically secure

flootr:
#### Proposed Changes

See p9oQ9f-1vT for context.

* Make `generatePassword` cryptographically secure

#### Testing Instructions

* Follow testing instructions from #48762 and make sure those steps work as described


packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Forked from https://github.com/WordPress/wporg-two-factor/blob/5241afeac49739ba58d8162f468434291dbc0fb6/settings/src/components/password.js#L132-L160
```
iandunn:
```suggestion
 * Forked from https://github.com/WordPress/wporg-two-factor/blob/3e19f8a11e9a18633a77a3f7a36f0ef9a984b3d3/settings/src/components/password.js#L136-L164
```

I just merged https://github.com/WordPress/wporg-two-factor/pull/33/, so `5241afea` is now on a deleted branch that GH will probably clean up eventually. 

`3e19f8a1` is in `trunk` and should be permanent.

flootr: Thanks! I updated the link in https://github.com/Automattic/wp-calypso/pull/71704/commits/29e2ad21caa190be3aa7ec67f08ef834d61e0b03

packages/generate-password/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"author": "Automattic Inc.",
+	"sideEffects": false,
+	"main": "dist/cjs/index.js",
+	"module": "dist/esm/index.js",
+	"calypso:src": "src/index.ts",
+	"private": true,
+	"repository": {
+		"type": "git",
+		"url": "git+https://github.com/Automattic/wp-calypso.git",
+		"directory": "packages/js-utils"
```
tyxla:
```suggestion
		"directory": "packages/generate-password"
```

packages/generate-password/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"version": "0.1.0",
+	"description": "Generate cryptographically secure random passwords",
+	"homepage": "https://github.com/Automattic/wp-calypso",
+	"license": "GPL-2.0-or-later",
+	"author": "Automattic Inc.",
+	"sideEffects": false,
+	"main": "dist/cjs/index.js",
+	"module": "dist/esm/index.js",
+	"calypso:src": "src/index.ts",
+	"private": true,
```
tyxla: If we're aiming to publish it and later consume it by `@wordpress` packages, should this be omitted?

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Forked from https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/password.js#L139-L167
+ *
+ * Generate a cryptographically secure random password.
+ *
+ * This is just as secure as using the `generate-password` AJAX endpoint, but faster, and avoids introducing the
+ * possibility of XHR failures, etc.
+ */
+export function generatePassword() {
+	const characterPool = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()';
+	const passwordLength = 24;
```
tyxla: Given that we're now packaging and generalizing this function, should `length` be an argument inside the function instead of hardcoded? We could default to `24` characters, of course.

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Forked from https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/password.js#L139-L167
+ *
+ * Generate a cryptographically secure random password.
+ *
+ * This is just as secure as using the `generate-password` AJAX endpoint, but faster, and avoids introducing the
+ * possibility of XHR failures, etc.
+ */
+export function generatePassword() {
```
tyxla: Should we add some basic tests?

tyxla (commented):
Thanks for working on it @flootr, it looks pretty solid!

I've left a few questions, let me know what you think.

Other than that, this tests great!

Also, I think the failing e2e tests are unrelated, but a rebase should likely address them.

packages/generate-password/src/index.ts
iandunn: It looks like that references `trunk`, which isn't a permalink. The line number refs will point to different lines soon. Referencing the specific commit also makes it obvious which commit this code was forked from.

`trunk` works too though if you prefer, but in that case I'd recommend removing the line numbers.

packages/generate-password/src/index.ts
iandunn: Another potential arg could be for the character pool. The  `wporg-two-factor` version was based on `wp_generate_password()`, so I hardcoded `$chars` and `$special_chars`, but it also provides `$extra_special_chars`.

https://github.com/WordPress/WordPress/blob/01f1448f7370234069f1a12c239a7b7169868b55/wp-includes/pluggable.php#L2595-L2602

To use the package, I'd definitely want `$special_chars` to be available, but I don't personally feel strongly about `$extra_special_chars`.

packages/generate-password/package.json
iandunn: ➕ i'd use it for `wporg-two-factor` if it were published

iandunn (approved):
This looks good to me from a package-consumer point of view 👍🏻

packages/generate-password/src/index.ts
tyxla: Agreed. The defaults are useful, but it would be worth being able to change them.

flootr: Totally agree on making these configurable 👍🏻

packages/generate-password/package.json
flootr: Copy pasta 😬

packages/tsconfig.json
```diff
@@ -40,10 +41,10 @@
 		{ "path": "./plans-grid" },
 		{ "path": "./privacy-toolset" },
 		{ "path": "./search" },
-		{ "path": "./sites" },
 		{ "path": "./shopping-cart" },
-		{ "path": "./subscriber" },
+		{ "path": "./sites" },
 		{ "path": "./state-utils" },
+		{ "path": "./subscriber" },
```
tyxla, line 47: Thanks for sorting those 👍

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,62 @@
+type CharacterPoolOptions = {
```
tyxla: Maybe we could export the type? Could be useful externally.

tyxla: If all the settings are optional, perhaps it would make more sense to make the type [`Partial`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) rather than declare each property as  optional. WDYT?

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	useSpecialChars?: boolean;
+	/**
+	 * Use characters -_ []{}<>~`+=,.;:/?| in character pool
+	 */
+	useExtraSpecialChars?: boolean;
+};
+
+/**
+ * Forked from https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/password.js#L139-L167
```
tyxla: Noting that if we include `trunk` in the URL, it might become obsolete. I'd always recommend using the full hash there, because it will (almost) always lead to what we intended.

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Define the length of the generated random password
+	 */
+	length?: number;
+	/**
+	 * Use digits 0-9 in character pool
+	 */
+	useNumbers?: boolean;
+	/**
+	 * Use characters !@#$%^&*() in character pool
```
tyxla:
```suggestion
	 * Use characters !@#$%^&*() in the character pool
```

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	useSpecialChars?: boolean;
+	/**
+	 * Use characters -_ []{}<>~`+=,.;:/?| in character pool
```
tyxla:
```suggestion
	 * Use characters -_ []{}<>~`+=,.;:/?| in the character pool
```

packages/generate-password/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	const randomNumber = new Uint8Array( 1 );
+	let password = '';
+
+	// JS doesn't provide a way to generate cryptographically secure random number within a range, so instead
+	// we just throw out values that don't correspond to a character. This is a little bit slower than using a
+	// modulo operation, but it avoids introducing bias in the distribution. Realistically, it's easily performant
+	// in this context.
+	// @link https://dimitri.xyz/random-ints-from-random-bits/
```
tyxla, line 57: Did you mean to use the `{@link}` JSdoc syntax:

https://jsdoc.app/tags-inline-link.html

packages/generate-password/src/test/index.ts
```diff
@@ -0,0 +1,28 @@
...
+	describe( `.length`, () => {
+		test( 'should use default length of 24', () => {
+			expect( generatePassword() ).toHaveLength( 24 );
+		} );
+
+		test( 'should respect `length` param', () => {
+			expect( generatePassword( { length: 12 } ) ).toHaveLength( 12 );
+			expect( generatePassword( { length: 1337 } ) ).toHaveLength( 1337 );
+			expect( generatePassword( { length: 99 } ) ).toHaveLength( 99 );
+		} );
```
tyxla, line 26: Optional tests to add here could be negative tests for the different character pool settings - for example, when `useSpecialChars` is disabled, a test that verifies that a special character does not exist in the generated password, and so on.

packages/generate-password/src/test/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { generatePassword } from '../';
+
+describe( 'generatePassword', () => {
+	const mockRandomValues = jest.fn( ( array ) => webcrypto.getRandomValues( array ) );
+
+	Object.defineProperty( window, 'crypto', {
+		value: {
+			getRandomValues: mockRandomValues,
+		},
+	} );
```
tyxla, line 15: You did it again, jsdom 😅 Bummer that we have to use that.

I wonder if it makes sense to use `node:crypto` as a fallback, and that way make the package SSR-compatible? Maybe that'd be over-engineering, so I'm just thinking out loud.

tyxla (approved):
Tests well for me 👍 

Left a little bit more feedback, I think nothing is blocking.

Nice work 🚀

packages/generate-password/src/test/index.ts
flootr: I thought about that and wasn't sure whether such a test would be that useful. Using the provided options to extend the character pool make it likely those characters appear in the password but don't guarantee that, it's totally possible for a password to not contain special characters if the option is set to `true`.

I added some tests in https://github.com/Automattic/wp-calypso/pull/71704/commits/c2241c41fb589b4dac357114eb7caf895c26a2f0 but I feel like that had to be noted. Given the used password length and the amount of times those tests run it's unlikely we run into problems though.

packages/generate-password/src/index.ts
flootr: I'm by no means an expert in that area. I added `Partial` earlier but changed it because I wasn't sure about what the exact benefit of using `Partial` is in this context. https://github.com/Automattic/wp-calypso/pull/71704/commits/83a7401f17bd597ac69fc06df5ffb74102f76524 brings it back.

packages/generate-password/src/index.ts
flootr: Good point 👍🏻 fixed via https://github.com/Automattic/wp-calypso/pull/71704/commits/07173bc49726d7a96dd4678095a9469ac0005050

packages/generate-password/src/index.ts
tyxla: It basically does the job of setting all fields to optional, rather than having you to do that yourself manually. It also sets the expectations to us about what to expect from that type (that each of its fields are optional) without having to parse the entire type.

flootr: Alright. I just saw it from another angle but it works well so it's fine by me 👍🏻

iandunn:
🎉 great work @flootr !

Will the npm package be published automatically at some point, or is there some manual process needed for that?

tyxla:
Hey @iandunn,

Packages are manually published at this time. 

I've just published it for you, `@automattic/generate-password@0.1.0`, see [tag](https://github.com/Automattic/wp-calypso/releases/tag/%40automattic%2Fgenerate-password%400.1.0) and [npm](https://www.npmjs.com/package/@automattic/generate-password).

If you need to publish it in the future, we have information and step-by-step instructions in the [monorepo docs](https://github.com/Automattic/wp-calypso/blob/trunk/docs/monorepo.md#publishing). If you run into trouble or need anything, please don't hesitate to ping us or me personally.

Thanks!

iandunn:
Thanks!

PR #71735, opened 2023-01-05
Remove unused slack notification build

noahtallen:
#### Proposed Changes
Per @nsakaimbo, this isn't used any more, so we can remove it.

#### Testing Instructions
CircleCI shouldn't complain about broken file syntax

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,6 @@
 version: 2.1
 
 orbs:
-  slack: circleci/slack@3.4.2
   win: circleci/windows@5.0.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,6 @@
 version: 2.1
 
 orbs:
-  slack: circleci/slack@3.4.2
```
noahtallen, line 4: Slack isn't otherwise used in the file

nsakaimbo (approved):
👍🏽

PR #71478, opened 2022-12-23
Fix IDR price display decimal

niranjan-uma-shankar:
#### Proposed Changes

* In IDR currency, the signup plans step and Calypso `/plans` display a `,00` fraction value for the price. This is unnecessary since we don't want to display zero value fractions.
* The reason is due to an incorrect subtraction of the raw price of a plan and integer value of the price. IDR and a few other currencies use dot as a group-of-three separator and the JS subtraction operator considers this to be a decimal point. 

**BEFORE**
Screenshot 2022-12-23 at 10 05 42 AM

Screenshot 2022-12-23 at 10 06 21 AM

**AFTER**
Screenshot 2022-12-23 at 10 05 23 AM
Screenshot 2022-12-23 at 10 06 33 AM

#### Testing Instructions

* IN IDR currency, go through the signup flow and check the plans step currency. Verify that it does not have the `,00` fractional value.
* Verify the same in Calypso `/plans` step too that there's no `,00` fractional value.
* There should be no difference in USD, EUR, INR etc. Verify for a few currencies. 

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #

PR #63828, opened 2022-05-19
Experiment: Upsell screen in Jetpack purchase flow

monsieur-z:
### Changes proposed in this Pull Request

This PR implements a new screen to upsell Jetpack products, as an A/B test. This new screen will appear:
- Between the pricing page and the checkout page
- Only if a user selects Backup, Scan, or Antispam (the upsold product is Security)

Context: pbNhbs-2pq-p2

This PR implements all the logic for this experiment and new page. The design of the page is handled in #64329.

### Implementation notes

- This new page has a dedicated URL, depending on where the user is seeing the landing page from:
  - `cloud.jetpack.com/pricing/upsell/:product/:site?`
  - `wordpress.com/pricing/upsell/:product/:site?`
  - `wordpress.com/jetpack/connect/store/upsell/:product`
  - `wordpress.com/jetpack/connect/plans/upsell/:product/:site`
- Name of the experiment: `calypso_jetpack_upsell_page_2022_06 `

### Testing instructions

_Note: clicking the close button in the checkout page should redirect you to the pricing page (passed in the `checkoutBackUrl` query param). This doesn't work as expected, and should be addressed in a separate PR._

#### Prerequisites

- Make sure you have a site connected to Jetpack with no subscriptions
- Check for JS errors in the console

#### Regression testing

- If you've tested this PR before, make sure to delete the `explat-experiment--jetpack_upsell_page_2022_05` key in your local storage, or assign yourself to the control variation in the experiment tool.
- In Calypso blue, visit `/plans/:site`
- Select a product, and make sure you're directed to the checkout page with the product in the cart
- Repeat the same operation with a couple more products
- Visit `/plans/upsell/jetpack_backup_t1_yearly/:site`: you should be redirected to the pricing page (you might briefly see the upsell page flash, but that's acceptable since we'll eventually keep or get rid of the page definitively)

#### Upsell screen

- Assign yourself to the treatment variation
- Calypso pricing page: 
  - Visit `/plans/:site`
  - Check that selecting Backup, Scan, or Anti-spam in the pricing page leads you to the upsell page
  - In the upsell page:
    - Check that the copy makes sense
    - Check that the price difference is accurate
    - Check that both links do what they advertise
    - Refresh the page, and verify that you see the same information
  - Check that selecting another product (than the 3 mentioned above) in the pricing page leads you directly to the checkout page
  - Visit `/plans/upsell/jetpack_complete/:site`, and check that you're redirected to the checkout page
- Calypso connect page: 
[...]

PR #71378, opened 2022-12-20
New Pricing Grid: Integration PR

niranjan-uma-shankar:
#### Proposed Changes

* This is the integration PR for the new pricing grid. The sub-PRs are listed in pdvytD-6M-p2.
* The design for the pricing grid can be found in 2e991-pb.
* The new pricing grid is launched to `en` only, in any signup flow which has a plans step. 

**Desktop**

npg

**Tablet**

Check [this](https://user-images.githubusercontent.com/1269602/210099964-b1450f58-2cea-4050-9834-3deb3450b69c.jpg).

**Mobile**
Check [this](https://user-images.githubusercontent.com/1269602/210100273-372603a3-8b4e-421c-b46e-0f2727e40ae9.png).

#### Testing Instructions

**IN EN LOCALE**

* Signup as a new user from `/start` and verify that the plans step design matches the Figma in desktop, tablet, and mobile viewport sizes. 
    * Note that the Business plan is highlighted as popular, the Enterprise price aligns with the "per month" line, and the plan tagline has a 1px bigger font than shown in Figma, as per feedback shared in p1672275676517789?thread_ts=1672225348.890109-slack-C040UAN6TPS and p1672277398019369?thread_ts=1672225348.890109-slack-C040UAN6TPS.

* Verify the UI looks okay in a few different currencies, that is, nothing breaks in the UI for USD, EUR, JPY, IDR etc.
    * In INR, MXN, and PHP currencies the legacy plans UI shows crossed-out price for the 20% discount. The new pricing grid retains that design, even though it's not explicitly discussed in Figma. 
* Verify that you see the free plan both with a free domain and a custom domain selected in the domains step.
* Verify that the Tracks events on the plans step all work appropriately. 
* Confirm that Calypso `/plans` UI is unaffected in all viewport sizes.
* Verify that the new pricing grid looks and works fine in other flows which have a plans step, such as the domain-only flow at `/start/domain/domain-only` when you select the option to create a site.
* Verify the new pricing grid works fine in the site launch flow. As an FYI, the launch flow will show the plans step only for free plan sites.

**IN NON-EN LOCALE**

* You should see the legacy plans step in all contexts and all viewport sizes.

PR #71737, opened 2023-01-05
PlanPrice: convert to TypeScript and improve strip-zeros behavior

sirbrillig:
#### Proposed Changes

This PR converts the `PlanPrice` component to TypeScript. It also improves one significant behavior of the component: displaying numbers in "strip-zeros" style.

PlanPrice renders most of its prices using the "strip-zeros" approach: if a price is an integer number, the number will be rendered without its decimal section. However, the way it does this is by trying to perform math on the _formatted_ version of the non-decimal section of a price (eg: `10.000,50` for 10,000.50 EUR); this version is a formatted string including spaces, grouping separators, etc. and will therefore often be challenging to use for math.

This change modifies the code to examine the _number_ version of the price instead to see if it contains non-zero decimal places, which should always be accurate regardless of currency or locale.

**Note:** Due to the added types, the TS conversion surfaced a subtle bug in `JetpackUpsellPage` (added in https://github.com/Automattic/wp-calypso/pull/63828) where the currency was always ignored. The bug is fixed by this PR.

#### Testing Instructions

Automated tests already cover this code and the component's output should be unchanged. To be sure, this PR adds additional tests that work with or without this PR (you can check yourself by reverting the changes to the file before running them).

Also, because there was previously no test for it (though this PR adds one), make sure that this doesn't break the fix to IDR applied in https://github.com/Automattic/wp-calypso/pull/71478, ie: that the Plans page does not display zeros for integer prices in currencies with a period (`.`) as the grouping separator:

Screenshot 2023-01-05 at 6 23 34 PM

### DIFF
--- a/client/blocks/importer/wordpress/import-everything/confirm-upgrade-plan.tsx
+++ b/client/blocks/importer/wordpress/import-everything/confirm-upgrade-plan.tsx
@@ -83,7 +83,7 @@ export const ConfirmUpgradePlan: FunctionComponent< Props > = ( props ) => {
 			<div className={ classnames( 'import__upgrade-plan-container' ) }>
 				<div className={ classnames( 'import__upgrade-plan-price' ) }>
 					<h3 className={ classnames( 'plan-title' ) }>{ plan?.getTitle() }</h3>
-					<PlanPrice rawPrice={ rawPrice } currencyCode={ currencyCode } />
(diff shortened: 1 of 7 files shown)


sirbrillig:
Rebased over https://github.com/Automattic/wp-calypso/pull/71378 which added a new prop.

PR #71738, opened 2023-01-05
desktop: rotate secrets

nsakaimbo:
#### Proposed Changes

This PR uses updated secrets to re-encrypt select files that support the desktop application - Mac and Windows code signing certificates, as well as the `secrets.json` file.

- `CALYPSO_SECRETS_ENCRYPTION_KEY`
- `CSC_KEY_PASSWORD`
- `WIN_CSC_KEY_PASSWORD`
- `WP_DESKTOP_SECRET`
- `WPDESKTOP_NOTARIZATION_PWD`

The variable `GH_TOKEN` is also required by the desktp app (for deployment). While the existing value has been revoked, it's unclear which account generated the token to begin with.


noahtallen (approved):

nsakaimbo:
Marking as [HOLD] as we need to update the credentials used by TeamCity.

PR #71747, opened 2023-01-06
Block Renderer: Support rendering customized site info

arthur791004:
#### Proposed Changes

* Referring to pbxlJb-39t-p2, we want to support the customized site info for the preview of the patterns. So, this PR proposes to send the customized site title and tagline to the `/block-renderer/patterns/render` endpoint to control the site title and tagline in the preview of patterns

#### Testing Instructions

* Apply D97204-code to your sandbox
* Go to `/setup?siteSlug=<your_site>&flags=pattern-assembler/client-side-render`
* When you're in the Pattern Assembler step, verify the rendered site title and tagline are correct.
  * The default site title for the newly created site is "Site Title"
  * The default site tagline is "Site Tagline"

#### Pre-merge Checklist

Related to pbxlJb-39t-p2

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -2,7 +2,7 @@ import { isEnabled } from '@automattic/calypso-config';
 import { StepContainer } from '@automattic/onboarding';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { useTranslate } from 'i18n-calypso';
-import { useState, useRef, useEffect } from 'react';
+import { useState, useRef, useEffect, useMemo } from 'react';
 import { useDispatch as useReduxDispatch } from 'react-redux';
 import AsyncLoad from 'calypso/components/async-load';
 import DocumentHead from 'calypso/components/data/document-head';
@@ -14,6 +14,7 @@ import { useSiteSlugParam } from '../../../../hooks/use-site-slug-param';
 import { useThemeDetails } from '../../../../hooks/use-theme-details';
 import { SITE_STORE, ONBOARD_STORE } from '../../../../stores';
 import { recordSelectedDesign } from '../../analytics/record-design';
+import { SITE_TAGLINE } from './constants';
 import PatternLayout from './pattern-layout';
 import PatternSelectorLoader from './pattern-selector-loader';
 import { useAllPatterns } from './patterns-data';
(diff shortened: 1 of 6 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -55,6 +56,14 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 		activePosition,
 	};
 
+	const siteInfo = useMemo(
```
fushar: Curious, why do we need `useMemo` here? There is no heavy computation at all

arthur791004: I just want to ensure the `siteInfo` is always the same object after every re-renders if the data is never changed. But let me check whether there is a performance issue if we create a new object after every re-renders.

bangank36: site?.title is always `undefined`, tested on default site info & customized info
image
Pattern 5601-174455321 should have both title & tagline

arthur791004: Oops...I use the wrong variable 🤦‍♂️

arthur791004: Fixed the site title issue and removed the `useMemo` 😉

bangank36: Work great

fushar:
Hmm, I cannot get the site logo to work:

|image|image|
|-|-|

When I finish the flow, somehow other logo is there? I don't understand, maybe I'm missing something here

image

fushar:
Wait sorry ah so the site logo is not supported intentionally in this PR? 😂

fushar:
Sorry, after the weekend I totally forgot that we only support placeholder image there :)

fushar:
But still is it the correct behavior if the site does actually have site logo?

arthur791004:
> But still is it the correct behavior if the site does actually have site logo?

It's the expected behavior for the onboarding flow as we never have a step to upload the site logo. However, as we support the placeholder image, we can use the current site instead of the theme demo site to render the site-related blocks. But my concern is the posts of the query blocks. I think for the newly created site, the posts should be from the theme demo site (blank canvas site). So, I think we can keep the current behavior until we have a step to upload the site logo or we want to get the posts from another site. What do you think?

fushar:
> So, I think we can keep the current behavior until we have a step to upload the site logo or we want to get the posts from another site.

I see now. I agree.

fushar (approved):

Issue #71598, opened 2022-12-30
Hosting Configuration: Describe premium features in the upsell nudge


zaguiini:
### Details

Part of pdKhl6-1th-p2.

The current upsell nudge on the Hosting Configuration page is not attractive at all:

![image](https://user-images.githubusercontent.com/26530524/210105371-547ec45f-888f-49bb-8a05-621bf4aa97b0.png)

It tells me to upgrade but doesn't say what I'll get if I do 🤔 😅 

### Done is

- [x] I see what features I'll get before upgrading to the Business plan
- [x] There's detailed information for every feature so I'm sure of what I'm getting

p-jackson:
Planning on using this design from pdKhl6-1th-p2#comment-1811, but without the price info.

![image](https://user-images.githubusercontent.com/1500769/210931599-16a6d20a-3ae1-43b6-a942-ddbdf18c03cb.png)

Explains the value proposition well.

p-jackson:
I'm unsure about including the pricing info (see pdKhl6-1th-p2#comment-1836), so I'm going to submit my current WIP for review without adding it. Because it's gated behind an experiment no one will see the updated banner yet and I can add the price info in a follow up PR.

- [x] Add new banner variation behind A/B test #71750
- [ ] ~Add price information~ https://github.com/Automattic/wp-calypso/issues/71833
- [ ] ~Finalise copy~ https://github.com/Automattic/wp-calypso/issues/71832

PR #71750, opened 2023-01-06
Hosting Config: Describe premium features in the upsell nudge

p-jackson:
#### Proposed Changes

Users in the `treatment` cohort of the `calypso_hosting_configuration_upsell_list_features` experiment see an updated upsell nudge on the `/hosting-config` page.

CleanShot 2023-01-09 at 18 47 15@2x

* Uses `<Experiment>` component to switch between the two banner variations (read more pbxNRc-2f8-p2)
* Displays spinner if experiment cohort is still being loaded/assigned
* Custom icon for each feature we're listing: pdKhl6-1th-p2#comment-1839
* Columns resize and wrap as the screen width changes
* CTA links to the checkout, just like the previous nudge
* Banner uses the same tracks events so that the experiment tools can make the comparison

#### Implementation Discussion

The `<UpsellBanner>` and `<Banner>` components feel like god-classes. They have a lot of properties to toggle between all their different modes. Ideally these components would be more composable rather than have lots of modes e.g rather than exposing a `primaryButton` prop it would allow users to provide their own button.

Rather than adding yet more flags to the shared components I've been able to achieve most of the re-styling in CSS. However the CSS selectors need to be very specific (especially overloading specific media queries) which is not ideal. I suppose it's arguable which is better for long-term maintainability: god-components vs. fragile CSS overrides. I'd be interested in your thoughts.

#### Out of scope for this PR

This PR doesn't:
- represent the final copy. We'll have time to adjust that.
- include the price as scene in the figma design. I'll do that in a follow up PR.

#### Testing Instructions

* Go to experiment page 20995-explat-experiment
* Add the `treatment` bookmarklet to your bookmark bar
* Click bookmarklet and add your user to the `treatment` cohort (tutorial video p1673249013170409/1673241333.083479-slack-C04H4NY6STW)
* `calypso.localhost:3000/hosting-config` in a site with no business plan
* You should see the new nudge
* Test being in the other cohort too to confirm nothing has changed.

#### Pre-merge Checklist

Related to #71598


mpkelly:
The code is nice and clean, as is the UI! I couldn't test it, though (I sent a message on Slack). My only comment is about the icons - ideally, we'd have one of each, but some of the icons used on the banner are slightly different to those used on the hosting features. 

Regarding CSS specificity: very specific CSS is a good thing in this code base IMO. 😄 

A nice start to 2023, @p-jackson!

p-jackson:
> My only comment is about the icons - ideally, we'd have one of each, but some of the icons used on the banner are slightly different to those used on the hosting features.

Yeah good point I hadn't noticed that. The problem is that the cards don't line up 1-to-1 with features. The credentials card at the top is needed to use SFTP, SSH or CLI; so which icon should it match with?

CleanShot 2023-01-10 at 12 56 13@2x

And sometimes the card refers to SSH and sometimes it doesn't.

CleanShot 2023-01-10 at 12 57 09@2x

So it's not just a matter of lining them up.

Since the first card is the only situation when this clash of icons is apparent (since the overlay fades out the rest of the cards) I think I can just come up with a fix for this one. Since it's the credentials card I'll try using a key icon or something.

p-jackson:
This PR changes the icon to a key https://github.com/Automattic/wp-calypso/pull/71828

mpkelly:
Looks good to me and can be merged. If I were being _ultra_ picky I would say some icons could be flipped in RTL mode:

image

Using `transform: scale(-1, 1);`

image

(I am curious to know how the PHP would even look to a person that reads RTL). 

I tested with Firefox and Chrome on mobile, tablet and desktop. 

You should consider updating the Calypso examples to include a banner in this style. It was updated in https://github.com/Automattic/wp-calypso/issues/71597 but only to show simple list items. I am a big fan of rich content like this and think we should promote it.

mpkelly (approved):

p-jackson:
Thanks for testing RTL!
English words and letters don't get mirror flipped when written in Arabic, so I don't think it makes sense to flip that icon (we're moving to a Database icon anyway). The CLI icon is interesting though! I'm googling-image-searching for what localised terminals look like.

p-jackson:
I was able to see what the standard terminal application looks like in RTL

CleanShot 2023-01-12 at 11 30 36@2x

The prompt still renders LTR, so I don't think the icon needs to change.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7727672

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7727672).

PR #71766, opened 2023-01-06
Refactor getSiteFragment to add new exception for plan upsell path

### LINKED ISSUE #71669: Post purchase upsell: fix conversion tracking
The path prop on the `calypso_page_view` event is returning `checkout/thank-you/:site/pending/:order_id` for the upsell page, instead of `/checkout/:site/offer-plan-upgrade/business/:order_id`. We'll need to fix that or add another event to track upsell page impressions.

Acceptance criteria:
- [ ] We can track upsell page impressions.

DustyReagan:
Context: The `calypso_page_view event` Tracks event is not running on the Premium-Business upsell page. e.g. `/checkout/example.wordpress.com/offer-plan-upgrade/business-monthly/65846531` This is because `getSiteFragment` is returning `:order_id` instead of `:site` for the the upsell page.

Issue here: https://github.com/Automattic/wp-calypso/issues/71669

#### Proposed Changes

* By default `getSiteFragment` looks for site slug or site ID in the last 2 URL positions. However, there are exceptions to that generality.
* This PR refactors `getSiteFragment` so it can better wrangle those exceptions by reducing and discouraging code duplication.
* This PR adds a new exception for the Premium to Business upsell page path. e.g. `/checkout/example.wordpress.com/offer-plan-upgrade/business-monthly/65846531`
  * Before this PR `getSiteFragment` incorrectly returned 65846531 for this path. Now it correctly returns `example.wordpress.com`.
* This PR also adds a unit test for the above exception.

#### Testing Instructions

Code:
* Before this PR I added more unit tests to cover `getSiteFragment` here: https://github.com/Automattic/wp-calypso/pull/71736 These should continue to pass as well as the new unit test included in this PR.
* Review the PR diff to be double check nothing was missed in the refactor.

Functionality:
* Follow this guide so you can view Tracks events fire in real-time: PCYsg-cae-p2
* In production load this page http://wordpress.com/checkout/example.wordpress.com/offer-plan-upgrade/business-monthly/65846531 and notice that the `calypso_page_view` event is NOT called.
* Load this PR and go to the same page locally: http://calypso.localhost:3000/checkout/example.wordpress.com/offer-plan-upgrade/business-monthly/65846531. Notice that the `calypso_page_view` event IS called and the `path` property is correct.

#### Pre-merge Checklist

Fixes https://github.com/Automattic/wp-calypso/issues/71669

### DIFF
--- a/client/lib/route/path.ts
+++ b/client/lib/route/path.ts
@@ -1,4 +1,4 @@
-import { URL as URLString, SiteSlug, SiteId } from 'calypso/types';
+import { SiteId, SiteSlug, URL as URLString } from 'calypso/types';
 import { trailingslashit, untrailingslashit } from './index';
 
 /**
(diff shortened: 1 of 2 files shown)


allilevine:
> Review the PR diff to be double check nothing was missed in the refactor.

@DustyReagan This seems like a case where adding Github comments to each change could make reviewing easier. Like this code was deleted here but moved here, this is just a comment update, etc.

client/lib/route/path.ts
```diff
@@ -17,46 +17,47 @@ export function getSiteFragment( path: URLString ): SiteSlug | SiteId | false {
...
 		0 === basePath.indexOf( '/checkout/thank-you/' )
 	) {
-		const piece = pieces[ 3 ]; // 0 is the empty string before the first `/`
-		if ( piece && -1 !== piece.indexOf( '.' ) ) {
-			return piece;
-		}
-		const numericPiece = parseInt( piece, 10 );
-		if ( Number.isSafeInteger( numericPiece ) ) {
-			return numericPiece;
-		}
```
DustyReagan, line 46: The code here is a duplication of the code in the `for` loops below, but missing some recent updates. To prevent duplicating this code again for our "upsell path" exception looking in the 2nd URL position, I've refactored the `getSiteFragment` function to define the `searchPositions` we'd like to look in. Then at the bottom of the function we loop through `searchPositions` looking for the slug or ID.

client/lib/route/path.ts
```diff
@@ -17,46 +17,47 @@ export function getSiteFragment( path: URLString ): SiteSlug | SiteId | false {
...
+		// Avoid confusing the subscription ID for the site ID in gifting checkouts.
+		( basePath.includes( '/gift/' ) && basePath.startsWith( '/checkout/' ) )
+	) {
 		return false;
 	}
 
-	// Avoid confusing the subscription ID for the site ID in gifting checkouts.
-	if ( basePath.includes( '/gift/' ) && basePath.startsWith( '/checkout/' ) ) {
-		return false;
-	}
```
DustyReagan, line 31: I combined these two exceptions that always return `false` in the refactor.

client/lib/route/path.ts
```diff
@@ -1,4 +1,4 @@
-import { URL as URLString, SiteSlug, SiteId } from 'calypso/types';
+import { SiteId, SiteSlug, URL as URLString } from 'calypso/types';
```
DustyReagan, line 1: I guess WebStorm decided these needed to be in alphabetical order. 😄

DustyReagan:
> This seems like a case where adding Github comments to each change could make reviewing easier. Like this code was deleted here but moved here, this is just a comment update, etc.

Good call, @allilevine! I added some GitHub comments to help. I think in this case it might be easiest to review looking at both functions in whole side-by-side, in lieu of a "diff view", or possibly using the 'split' diff view.

client/lib/route/path.ts
allilevine: @DustyReagan Thanks for clarifying! 🙌 Is there any functionality we should test, like that other Tracks events are recording the correct path?

allilevine (commented):
This is working well for me, I see `/checkout/:site/offer-plan-upgrade/:upgrade_item/:receipt_id` tracked as the path. It even showed up in Tracks Live from calypso.live. 🎉

client/lib/route/path.ts
DustyReagan: I'm leaning pretty heavy on the unit tests for regression testing. Eyeballing those for any missing permutations would be good. 🤔 I suppose we could also cruise WordPress.com watching `calypso_page_view` in the console for the correct `path`.

allilevine: > Eyeballing those for any missing permutations would be good.

I'm not seeing any issues but more 👀 seem like a good idea. @tyxla @Automattic/kitkat Any chance you can take a look at the change here and whether we have sufficient test coverage?

client/lib/route/test/index.js
```diff
@@ -168,7 +175,7 @@ describe( 'route', function () {
 					'example.wpcomstaging.com'
 				);
 			} );
-			test( 'should return false when plugin name ends a number', function () {
+			test( 'should return false when plugin name starts with a number', function () {
```
tyxla, line 178: Nice find 👍

client/lib/route/path.ts
tyxla: IMHO test coverage of that function is quite satisfactory. I've taken a look and more or less all cases and exceptions are covered. I think we can rely on those tests 👍

candy02058912 (approved):
Checked that the tracks event is showing after the fix.
Reading through all the tests (phew) and test coverage seems to give us the confidence that we won't break anything to the best of our knowledge.

Great work!

client/lib/route/path.ts
allilevine: Thank you! 🙌

allilevine (approved):

PR #71769, opened 2023-01-06
Chart: Fix a bug in the detection of empty charts

wilmersondasilva:
The value `isEmptyChart` was true even if the `data` prop was an empty array. So, instead of one `empty state` section, the user only saw an empty chart (see screenshot below).

#### Proposed Changes

* Fixes detection of empty charts

#### Testing Instructions

* It's probably easier to test this locally, so `git checkout` to this branch locally
* Go to some file that is using the Chart component and pass an empty array to the `data` prop
* You should see an `empty state` section when you run this locally 
* If you do the same on the branch `trunk`, you will only see something similar to the screenshot below

#### Screenshot

Before
image

After
image

#### Pre-merge Checklist

### DIFF
--- a/client/components/chart/index.jsx
+++ b/client/components/chart/index.jsx
@@ -114,7 +114,7 @@ function Chart( {
 
 		return {
 			chartData: nextData,
-			isEmptyChart: Boolean( nextVals.length && ! nextVals.some( ( a ) => a > 0 ) ),
+			isEmptyChart: Boolean( ! nextVals.some( ( a ) => a > 0 ) ),
 			yMax: getYAxisMax( nextVals ),
 		};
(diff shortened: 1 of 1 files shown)


jsnmoon (approved):
This makes sense and works as intended. 🚢 

However, I don't think this currently affects real customers in production since we appear to always pass along lengthy arrays into the chart component. I tested this PR using a test site created three days ago and found that `nextVals` was an array of 30 zeros, for example.

wilmersondasilva:
> I tested this PR using a test site created three days ago and found that `nextVals` was an array of 30 zeros, for example.

For the chart on the Traffic page, this is true. But on others places, we're passing an empty array, like on the Store page.

PR #71417, opened 2022-12-21
Introduce block-renderer to site assembler to improve performance

arthur791004:
#### Proposed Changes

This PR is focusing on the block-renderer package and uses it to improve the performance of the site assembler. See the simple [README.md](https://github.com/Automattic/wp-calypso/blob/a58ad832102e59f4aa8b856bfb063390d7df3c6c/packages/block-renderer/README.md) to understand how to use the block-renderer package.

pbxlJb-39t-p2 describes the idea of the block-renderer, and pbxlJb-3gc-p2 does the performance analysis

Here is a follow-up PR to handle the scroll behavior: https://github.com/Automattic/wp-calypso/pull/71527

https://user-images.githubusercontent.com/13596067/209108222-d3114764-04b4-4d73-838f-9c5f35a75288.mov

#### Testing Instructions

* Apply D95663-code to your sandbox
* Go to `/setup?siteSlug=<your_site>&flags=pattern-assembler/client-side-render`
* Click the "Continue" button until you land on the Design Picker step
* Scroll to the bottom and select the Blank Canvas CTA
* When you're in the Pattern Assembler step
  * Verify every pattern in the pattern list looks good as before
  * Verify the large preview looks good after you select any header, sections, and footers

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [x] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

[...]

PR #70833, opened 2022-12-06
Update Cookie Banner (with hotfix for card and button styles)

robertsreberski:
#### Proposed Changes

- Add @automattic/privacy-toolset to wp-calypso
- Remove GdprBanner and replace it with CookieBanner (SSR and async)
- Include checking both GDPR
- Support new cookie structure server-side
- Ensure all styles related to card and button components are still displaying correctly

#### Testing Instructions

> Disclaimer: this PR changes env flag from `gdpr-banner` to `cookie-banner`. Please consider that in the testing plan.

* Checkout changes
* Modify the config for your test environment to ensure that cookie-banner is set to true

##### Testing client-side Cookie Banner
* Run the calypso environment (`yarn start`)
* Go to any authenticated calypso page e.g., `http://calypso.localhost:3000/me/privacy?flags=cookie-banner`
* Change your cookies to GDPR region `country_code=NL;`
* Ensure you don't have `sensitive_pixel_option` and `sensitive_pixel_options` cookies
* Reload the page
* You should see the cookie banner floating at the bottom of the viewport

CleanShot 2022-11-30 at 19 17 03@2x

* Click "Customize"
* You should see the cookie banner with bucket options

CleanShot 2022-12-02 at 08 26 28@2x

* Click "Accept selection"
* Reload the page
* Cookie Banner should not load again

##### Testing server-side Cookie Banner

* (if you don't have VPN)
* Run the calypso environment (`yarn start`)
* Load `http://calypso.localhost:3000/log-in` as a logged-out user
* Change your cookies to GDPR region `country_code=NL`
* (if you have VPN)
* Set VPN to any of the EU countries, e.g., France.
* (continue)
* Run calypso environment with server-side `yarn build-server && yarn start`)
* Ensure you don't have `sensitive_pixel_option` and `sensitive_pixel_options` cookies
* Load `http://calypso.localhost:3000/log-in` as a logged-out user
* Ensure the cookie bar gets displayed as part of the first render
* Click the Accept button
* Ensure the cookie bar animates away
* Reload the page
* Ensure the cookie bar does not get displayed as part of the first render

##### Testing that `<Card />` and `<Button />` (`@automatttic/components`) styles are still valid
* Go to `http://calypso.localhost:3000/themes` (you don't need to trigger cookie banner)
* Ensure the design (especially padding) looks as below:
 
image

* Go to `http://calypso.localhost:3000/start/account/user` as logged out
* Ensure the design looks as below:

image

[...]

PR #71772, opened 2023-01-07
Fix missing TypeScript dependencies for privacy-toolkit and block-renderer

sirbrillig:
#### Proposed Changes

Currently when running the TypeScript compiler on calypso's `client` directory, you get a lot of errors like:

```
client/blocks/do-not-sell-dialog/index.tsx:1:33 - error TS2307: Cannot find module '@automattic/privacy-toolset' or its corresponding type declarations.
client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-list-renderer.tsx:1:33 - error TS2307: Cannot find module '@automattic/block-renderer' or its
corresponding type declarations.
```

These errors are also visible if you use a TypeScript-aware editor when editing the mentioned files. However, the missing packages do exist, they just had not been listed as dependencies or added to the TS config file. 

This PR adds both packages, `@automattic/privacy-toolkit` (added by https://github.com/Automattic/wp-calypso/pull/70833) and `@automattic/block-renderer` (added by https://github.com/Automattic/wp-calypso/pull/71417) as dependencies to the `client` package and also to the TS config file.

**Note:** Yarn's build process now reports the following error, but I'm not sure how to fix it.

```
➤ YN0060: │ wp-calypso@workspace:. provides react (p4692f) with version 17.0.2, which doesn't satisfy what @automattic/block-renderer and some of its descendants request
➤ YN0060: │ wp-calypso@workspace:. provides react-dom (p0b524) with version 17.0.2, which doesn't satisfy what @automattic/block-renderer and some of its descendants request
...
➤ YN0000: wp-calypso@workspace:. provides react@npm:17.0.2 with version 17.0.2, which doesn't satisfy the following requirements:
➤ YN0000: react-autosize-textarea@npm:7.1.0 [f63de]                            → ^0.14.0 || ^15.0.0 || ^16.0.0 ✘
```

So the problem is that `@automattic/block-renderer` depends on `@wordpress/block-editor`, which does not support React 17 because of its dependency on `react-autosize-textarea`. The issue would be resolved by this PR but I think the repo might be abandoned: https://github.com/buildo/react-autosize-textarea/pull/148 

#### Testing Instructions

Before this PR, running `yarn tsc -p client/tsconfig.json --noEmit` would result in 8 errors.

After this PR, the command will result in no errors.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -50,6 +50,7 @@
 		"@automattic/onboarding": "workspace:^",
 		"@automattic/pattern-picker": "workspace:^",
 		"@automattic/plans-grid": "workspace:^",
+		"@automattic/privacy-toolset": "workspace:^",
 		"@automattic/react-virtualized": "^9.22.3",
 		"@automattic/request-external-access": "workspace:^",
 		"@automattic/search": "workspace:^",
--- a/package.json
+++ b/package.json
@@ -134,6 +134,7 @@
 		"whybundled:server": "NODE_ENV=production CALYPSO_ENV=production EMIT_STATS=withreasons CONCATENATE_MODULES=false yarn run build-server && whybundled client/stats-server.json"
 	},
 	"dependencies": {
+		"@automattic/block-renderer": "workspace:^",
 		"@automattic/calypso-babel-config": "workspace:^",
 		"@automattic/calypso-build": "workspace:^",
 		"@automattic/calypso-color-schemes": "workspace:^",
--- a/packages/tsconfig.json
+++ b/packages/tsconfig.json
@@ -7,6 +7,7 @@
 	// Reference all TS packages
 	"references": [
 		{ "path": "./accessible-focus" },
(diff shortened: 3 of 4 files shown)


sirbrillig:
Nevermind, since the dependencies cannot be safely added due to React 17, I just fixed the TS errors only in https://github.com/Automattic/wp-calypso/pull/71773

noahtallen:
By the way, we do have a lot of stuff in our `.yarnrc.yml` file which lets yarn use these dependencies: https://github.com/Automattic/wp-calypso/blob/trunk/.yarnrc.yml

Essentially, the "solution" is to ignore the warning if we test that it'll work correctly. Now that dependency management like yarn is strict about peer dependencies, this is our main tool for working with older/abandoned packages. Ideally, though, Gutenberg would move away from things which don't support React 17! (But at this point, Gutenberg is on React 18, so it'll have to go even further :))

PR #70118, opened 2022-11-17
Enable Zendesk pre-sales chat in Checkout

yoavf:
#### Proposed Changes

* This enables a Zendesk pre-sales chat widget on checkout (availability of the chat is controlled by the wpcom endpoint)
* It replaces the HappyChat-based pre-sales chat. However, HappyChat will still be available to users who typically have access to it when the ZD isn't available, via the `CheckoutSummaryHelpButton` (as before).

Known issue: Once the ZD loaded widget is loaded, it will stay visible, even if we navigate away from the checkout section. That's currently acceptable.

#### Screenshots
**Zendesk Chat widget active:**
Screenshot 2022-11-17 at 13 44 15

**Zendesk pre-sales chat not available, and user does **not** have chat support level:**
Screenshot 2022-11-17 at 13 43 49

**Zendesk pre-sales chat not available, and user does have access to chat support level:**
Screenshot 2022-11-17 at 13 43 21

#### Testing Instructions

Reproduce the states from the above screenshots in checkout:
 - ZD chat available (by overwriting `presalesZendeskChatAvailable` if it returns false)
 - ZD chat not available, while user does have access to chat support (i,e with an existing plan)
 - ZD chat not available, while user does not have access to chat support

#### Pre-merge Checklist

- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

This PR partially replaces #69706 - ZD chat in signup will be added separately.

PR #71789, opened 2023-01-09
Checkout: don't display pre-sales chat if purchases are blocked

### LINKED ISSUE #71310: Checkout: hide the presales Zendesk chat when purchases are blocked

yoavf:
#### Proposed Changes

* This checks the cart for purchases-blocked error messages and prevents the display of pre-sales chat if purchases are blocked.

#### Testing Instructions

* To block purchase for a user on wpcom, run `update_user_attribute( $_user_id, 'block_purchases', true )` in wpsh;
* See #70118 for general testing instructions (note that by default, presales chat availability is controlled by wpcom endpoint). You can overwrite `presalesZendeskChatAvailable` to check just this new condition.

#### Pre-merge Checklist

Fixes #71310

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-help-link.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-help-link.tsx
@@ -147,7 +147,7 @@ export default function CheckoutHelpLink() {
 	const cartKey = useCartKey();
 	const { responseCart } = useShoppingCart( cartKey );
 
-	const cartErrors = responseCart.messages?.errors || [];
+	const cartErrors = responseCart.messages?.persistent_errors || [];
 	const purchasesAreBlocked = cartErrors.some(
 		( error: ResponseCartMessage ) => error.code === 'blocked'
 	);
--- a/client/my-sites/checkout/composite-checkout/components/checkout-help-link.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-help-link.tsx
@@ -3,6 +3,7 @@ import { Gridicon } from '@automattic/components';
 import { HelpCenter } from '@automattic/data-stores';
 import { SUPPORT_FORUM, shouldShowHelpCenterToUser } from '@automattic/help-center';
 import { useIsEnglishLocale } from '@automattic/i18n-utils';
+import { ResponseCartMessage, useShoppingCart } from '@automattic/shopping-cart';
 import { keyframes } from '@emotion/react';
 import styled from '@emotion/styled';
 import { useDispatch as useDataStoreDispatch } from '@wordpress/data';
@@ -11,6 +12,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import QuerySupportTypes from 'calypso/blocks/inline-help/inline-help-query-support-types';
 import AsyncLoad from 'calypso/components/async-load';
 import HappychatButtonUnstyled from 'calypso/components/happychat/button';
+import useCartKey from 'calypso/my-sites/checkout/use-cart-key';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
 import { getCurrentUserId } from 'calypso/state/current-user/selectors';
 import getSupportLevel from 'calypso/state/happychat/selectors/get-support-level';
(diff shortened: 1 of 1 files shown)


client/my-sites/checkout/composite-checkout/components/checkout-help-link.tsx
```diff
@@ -142,6 +144,13 @@ export default function CheckoutHelpLink() {
 	const { setShowHelpCenter } = useDataStoreDispatch( HELP_CENTER_STORE );
 	const isEnglishLocale = useIsEnglishLocale();
 
+	const cartKey = useCartKey();
+	const { responseCart } = useShoppingCart( cartKey );
+
+	const cartErrors = responseCart.messages?.errors || [];
+	const purchasesAreBlocked =
+		cartErrors.filter( ( error: ResponseCartMessage ) => error.code === 'blocked' ).length > 0;
```
sirbrillig: Really minor but you could also use `some` instead of `filter` and `length`.

```suggestion
		cartErrors.some( ( error: ResponseCartMessage ) => error.code === 'blocked' );
```

nbloomf:
This is not currently working; I suspect it's because something changed about the way cart errors are collected on the backend in the time since this was last tested. Here's what I saw.

I created a new user and manually applied a fraudster decision from store admin. On entering checkout (logged in with a personal plan in the cart) live chat was available, though the actual purchase was blocked correctly. The shopping cart (where I think cartErrors is coming from) returned an empty error array:

Screen Shot 2023-02-07 at 11 00 30 AM

I'm at a meetup right now and unable to dig into this at the moment, but I think the fix will involve making sure the shopping cart endpoint (/v1.1/me/shopping-cart) detects when the user has a purchase block and adding an appropriate error message. The has_new_purchase_block function in fraud-risk.php should be useful for that.

klimeryk:
Thanks for having a look, @nbloomf 🙇‍♂️ I've retested and can confirm your findings. However, it seems the issue goes deeper:
 - If I just set the `block_purchases` user attribute as per the testing instructions, everything works as expected. There's a cart error, there's a notice in checkout, widget is not loaded. There's a notice in SA that the purchases for this user are blocked.
 - Setting the user as fraudster from SA failed for me because of a fatal error (quickfix for it in D100843-code).
 - After applying the fix for that, I was able to mark them as fraudster. That seems to set the `block_purchases` attribute as well. But indeed, for some reason that does not cause the `blocked` error to be returned from the backend.
 - Interestingly, setting them as fraudster (`Fraud 💩`) did _not_ cause a notice in shopping cart that their purchases are blocked.

I've verified that the `block_purchases` attribute should cause the `blocked` error to be added in `Default_Shopping_Cart_Validator`. And it does - but then the error is at some point discarded.

I'd be for merging this PR as-is and figuring out in a separate (backend, probably) patch what's broken. I'd ask @Automattic/shilling for help with that, as I'm sure y'all know much better those systems and how to not break them 🙏

sirbrillig:
I have a feeling that I know what's happening. Is it possible that in your testing you might only see the errors when you've made a POST to the shopping-cart endpoint (note that this happens when adding an item to the cart or when setting the cart's tax location or coupon) but not when you make a GET?

> I've verified that the block_purchases attribute should cause the blocked error to be added in Default_Shopping_Cart_Validator. And it does - but then the error is at some point discarded.

You're right that the logic inside the shopping-cart endpoint says: if the `block_purchases` user attribute is set, add an error to the cart. The only thing inside the endpoint I can see that would clear errors is when a saved cart is loaded: we clear all errors in that case. This was added in D79621-code to hide certain types of errors but then the cart was re-validated and persistent errors like this one would have remained. However, this second validation was removed in D80124-code so that the masterbar cart won't cause those errors to show up on every page in calypso.

It seems like there might be a better solution to this. Rather than clearing all errors when the cart loads we might want to clear only certain types of errors? I created https://github.com/Automattic/payments-shilling/issues/1384 to look into this.

klimeryk:
@sirbrillig, thanks for digging into it 🙇 I trust you're right about the root cause and I imagine you/Shilling is the best team to tackle it. I know you're probably at the DM or soon coming back from it - but when would you be able to create a fix for this? It's driving additional traffic for our sales efforts (pdm0RZ-nC-p2#comment-782).
Alternatively, do you think this PR needs any adjustments or the fix will be on the backend and the approach here is correct? Seems that it works in some cases, so it'd be an improvement already, right?

sirbrillig (approved):
This looks good to me. There are some issues with how the `blocked` message gets returned from the endpoint, but those are backend issues and should not be a blocker for this PR.

sirbrillig:
...I'm going to change where the `'blocked'` error message lives in D101050-code, so it may be a good idea to check both places in this PR, depending on which merges first.

sirbrillig:
Those changes are merged, so if you rebase this you should be able to check for `'blocked'` in `cart.messages.persistent_errors` and it should work for all situations.

sirbrillig:
Thanks!

klimeryk:
@sirbrillig, sorry for the late follow-up, this got lost from my radar in the post-meetup recovery 😅 Thank you so much for the backend work on this and ensuring we have a reliable source of these errors 🥇

PR #70588, opened 2022-11-30
Cart on Jetpack cloud Pricing page

andrii-lysenko:
#### Proposed Changes

We are enabling the purchase of multiple products on Jetpack Cloud pricing page and adding Shopping cart to achieve this. We are re-using existing Shopping Cart functionality enhancing the styles to match expected designs. 

#### Testing Instructions

* Create a new JN site
* click on Jetpack Cloud live link below
* or boot up this PR 
    * Run `git fetch && git checkout fix/cloud-cart-lightbox-cta`
    * Run `yarn start-jetpack-cloud`
    * Goto http://jetpack.cloud.localhost:3000
* Modify URL adding `/pricing/:'[JN site]?flags=jetpack/pricing-page-cart` in the end.
* Make sure the cart shows up in the header
* Test cart functionality by adding and deleting items, navigating to checkout, etc.
* Make sure Cart represented correctly on all screen sizes.

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #
p6TEKc-6FI-p2

PR #71803, opened 2023-01-09
Jetpack Cloud Cart integration - split changes in layout and mini-cart folder 

vikneshwar:
#### Proposed Changes

* This PR includes the common changes made to client/layout/ and packages/mini-cart in PR #70588 
* This adds the following props to `masterbar-cart-button.tsx`
  * `forceShow` : by default the cart icon shows up only if one or more items are added. This props make sure to show the cart all the time
  * `showCount`: by default count is not shown next to the cart icon. This enables it
  * `cartIcon`: Used to override the default cart icon
* This also adds following props to `mini-cart.tsx`:
  * `checkoutLabel`: Overrides the default checkout button label inside cart 
  * `emptyCart`: By default when cart is empty nothing is shown, this property allows us to show any custom component on empty cart 

#### Testing Instructions

* Merge this branch with the branch in PR #70588 and follow the testing instructions there.

##### Cart with count

Screenshot 2023-01-09 at 11 58 49 PM

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
+++ b/client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
@@ -2,11 +2,13 @@ import { Popover } from '@automattic/components';
 import { CheckoutErrorBoundary } from '@automattic/composite-checkout';
 import { MiniCart } from '@automattic/mini-cart';
 import { useShoppingCart } from '@automattic/shopping-cart';
+import { useMobileBreakpoint } from '@automattic/viewport-react';
 import { useTranslate } from 'i18n-calypso';
 import { useEffect, useRef, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
 import MasterbarItem from '../item';
+import { MasterBarCartCount } from './masterbar-cart-count';
 import { CartIcon } from './masterbar-cart-icon';
 
 import './masterbar-cart-button-style.scss';
@@ -17,6 +19,11 @@ export type MasterbarCartButtonProps = {
 	goToCheckout: ( siteSlug: string ) => void;
 	onRemoveProduct?: ( uuid: string ) => void;
 	onRemoveCoupon?: () => void;
+	forceShow?: boolean;
+	showCount?: boolean;
+	checkoutLabel?: string;
+	cartIcon?: React.ReactNode;
+	emptyCart?: React.ReactNode;
 };
 
 export function MasterbarCartButton( {
@@ -25,6 +32,11 @@ export function MasterbarCartButton( {
 	goToCheckout,
 	onRemoveProduct,
 	onRemoveCoupon,
+	forceShow = false,
+	showCount = false,
+	checkoutLabel,
+	cartIcon,
+	emptyCart,
 }: MasterbarCartButtonProps ) {
 	const { responseCart, reloadFromServer } = useShoppingCart(
 		selectedSiteId ? selectedSiteId : undefined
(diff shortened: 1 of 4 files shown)


andrii-lysenko (approved):
LGTM! Thank you for doing this. It is providing needed additional functionality and don't disturb current one on wordpress.com

client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
```diff
@@ -43,14 +58,26 @@ export function MasterbarCartButton( {
 		}
 	}, [ shouldShowCart, reduxDispatch ] );
 
+	useEffect( () => {
+		const isPricingPage = document.body.classList.contains( 'is-section-jetpack-cloud-pricing' );
+
+		//add the className for body, only if the page is in mobile view and in pricing page
+		if ( isActive && isMobile && isPricingPage ) {
+			document.body.classList.add( 'is-mobile-cart' );
```
sirbrillig: This feels a little awkward to me for three reasons: it's effectively using the body tag's class as a way to hold state, the class `is-mobile-cart` is very broadly named, and that the page this is designed to affect is in a very different part of the codebase from the masterbar.

I guess that you need to know on the pricing page if the masterbar cart is active for some reason. Is there a better way to do that?

What if we changed this `useEffect` to add a class which is specific to the masterbar cart, like `body--masterbar-cart-visible` (following BEM style), any time that the cart is visible; then the pricing page could respond to that class as needed without any page-specific logic here? If it also needs to know when the viewport is at mobile width it could use a media query (if inside CSS) or `useMobileBreakpoint` (if inside JS), but that can be done on the pricing page itself rather than here.

So something like:
```js
useEffect(() => {
	if (isActive) {
		document.body.classList.add( 'body--masterbar-cart-visible' );
	} else {
		document.body.classList.remove( 'body--masterbar-cart-visible' );
	}
}, [ isActive ]);

```

client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
```diff
@@ -59,17 +86,21 @@ export function MasterbarCartButton( {
...
 				className="masterbar-cart-button"
 				alwaysShowContent
-				icon={ <CartIcon newItems={ !! responseCart.products } active={ isActive } /> }
+				icon={ icon }
 				tooltip={ tooltip }
 				onClick={ onClick }
 				isActive={ isActive }
 				ref={ cartButtonRef }
 			/>
+			{ showCount && <MasterBarCartCount cartCount={ cartCount } /> }
```
sirbrillig, line 98: Can you add a screenshot of the cart count to the PR description?

sirbrillig (commented):
Overall I think this is a great addition. I had one suggestion to improve its DX, since this is a general change to the masterbar and could be useful for many parts of calypso.

It might also be useful to explain very briefly in the PR description what this PR actually does on its own (adds some specific props that do a few things and a CSS class). That way anyone reading it in the history after it is merged can quickly understand its purpose without having to click through to a different (and very complex) PR.

client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
vikneshwar: @sirbrillig, thanks for this valuable comment. I have updated accordingly

client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
vikneshwar: Added a screenshot.

sirbrillig (approved):
Looks great! Thank you!

PR #71805, opened 2023-01-09
WooExpress Trial: Add the ecommerce-trial-bundle-monthly plan to a new trial site

### LINKED ISSUE #71764: WooExpress Trial: Add trial plan as part of the new stepper flow
Once the endpoint for adding the trial plan is complete (D97045-code), we need to update the wooexpress flow so that the created site has the trial plan instead of a free plan.

markbiek:
#### Proposed Changes

When a new WooExpress trial site is added, ensure the site is created with the new ecommerce-trial-bundle-monthly plan.

image

#### Testing Instructions

* Patch your sandbox with D97045-code and sandbox the public-api.
* Visit http://calypso.localhost:3000/setup/wooexpress/.
* You should go through the `siteCreationStep`, `processing`, and `assignTrial` steps.
* You should finish on the dashboard for the new site.
* Click Upgrades -> Purchases and verify that there the "Active Upgrade" is "WordPress.com eCommerce Trial".

**Note: The admin sidebar will still say "Upgrades: Free" at this point. This will be addressed in future PRs.**.

#### Pre-merge Checklist

- [NA] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [NA] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [NA] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [NA] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [NA] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Closes #71764

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
@@ -22,6 +22,10 @@ const AssignTrialPlanStep: Step = function AssignTrialPlanStep( { navigation, da
 		if ( submit ) {
 			const assignTrialPlan = async () => {
 				try {
+					if ( ! data?.siteSlug ) {
+						throw new Error( 'Invalid site slug' );
+					}
+
 					await wpcom.req.post(
 						`/sites/${ data?.siteSlug }/ecommerce-trial/add/ecommerce-trial-bundle-monthly`,
 						{
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
@@ -1,4 +1,9 @@
+import { StepContainer } from '@automattic/onboarding';
+import { useI18n } from '@wordpress/react-i18n';
 import { useEffect } from 'react';
+import DocumentHead from 'calypso/components/data/document-head';
+import { LoadingEllipsis } from 'calypso/components/loading-ellipsis';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import wpcom from 'calypso/lib/wp';
 import type { Step } from '../../types';
 
@@ -11,6 +16,7 @@ export enum AssignTrialResult {
 
 const AssignTrialPlanStep: Step = function AssignTrialPlanStep( { navigation, data } ) {
 	const { submit } = navigation;
+	const { __ } = useI18n();
 
 	useEffect( () => {
 		if ( submit ) {
@@ -34,7 +40,29 @@ const AssignTrialPlanStep: Step = function AssignTrialPlanStep( { navigation, da
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [] );
 
-	return null;
+	const getCurrentMessage = () => {
+		return __( 'Setting up your trial' );
+	};
+
+	return (
+		<>
+			<DocumentHead title={ getCurrentMessage() } />
+			<StepContainer
+				shouldHideNavButtons={ true }
+				hideFormattedHeader={ true }
+				stepName="assign-trial-step"
+				isHorizontalLayout={ true }
+				recordTracksEvent={ recordTracksEvent }
+				stepContent={
+					<>
+						<h1>{ getCurrentMessage() }</h1>
+						<LoadingEllipsis />
+					</>
+				}
+				showFooterWooCommercePowered={ false }
+			/>
+		</>
+	);
 };
 
 export default AssignTrialPlanStep;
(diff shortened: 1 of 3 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/styles.scss
```diff
@@ -0,0 +1,25 @@
+@import "../style";
+
+$font-family: "SF Pro Text", $sans;
```
daledupreez: This feels like it should be scoped so it doesn't bleed out into Calypso in any unexpected ways.

client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/styles.scss
```diff
@@ -0,0 +1,25 @@
...
+
+		h1.formatted-header__title {
+			font-size: $font-title-medium;
+
+			@include break-medium {
+				font-size: $font-headline-medium;
+			}
+		}
+	}
+}
```
daledupreez: Was this a mistake? It looks like none of these styles should apply to the new step.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -17,6 +20,7 @@ const wooexpress: Flow = {
 		return [
 			{ slug: 'siteCreationStep', component: SiteCreationStep },
 			{ slug: 'processing', component: ProcessingStep },
+			{ slug: 'assignTrialPlan', component: AssignTrialPlanStep },
```
daledupreez, line 23: Question: what are we waiting for in the processing step?

client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const AssignTrialPlanStep: Step = function AssignTrialPlanStep( { navigation, data } ) {
+	const { submit } = navigation;
+
+	useEffect( () => {
+		if ( submit ) {
+			const assignTrialPlan = async () => {
+				try {
+					await wpcom.req.post(
+						`/sites/${ data?.siteSlug }/ecommerce-trial/add/ecommerce-trial-bundle-monthly`,
```
daledupreez, line 30: Should we not return an error when `! data.siteSlug`? It feels like various things could break if we don't have a site slug available.

daledupreez (commented):
Overall, this is working as expected, but I don't see anything rendered on the `assignTrialPlan` step, meaning I just see a blank page while that request is occurring. Given that I have longer latency, I am seeing this as a longer wait that we should fill with some sort of content.

Also, I want to make sure I understand the intent around how we want to handle the logic when the logic to add the free trial also triggers an Atomic transfer -- are we planning on handling that in this step? Or handling it in a new step?

client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/styles.scss
markbiek: Yes, this was a mistake. I copied this step from somewhere else and forget to adjust the styles.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
markbiek: The actual site creation happens in the processing step. Basically the site creation step queues up all the stuff that needs to happen, then it actually gets processed in the processing step.

I initially attempted to make the trial plan assignment happen the same way and have processing happen last but there didn't seem to be a way to make the queued actions happen in a particular order.

client/landing/stepper/declarative-flow/internals/steps-repository/assign-trial-plan/index.tsx
markbiek: My initial thought was that the API call would fail if the site slug was missing and we'd return the error that way.

But it probably makes more sense to explicitly check for a missing site slug.

markbiek:
> Also, I want to make sure I understand the intent around how we want to handle the logic when the logic to add the free trial also triggers an Atomic transfer -- are we planning on handling that in this step? Or handling it in a new step?

I think it makes the most sense to handle the Atomic transfer in this step in a follow-up PR.

sixhours:
Tested and this is working for me!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7723783

Thank you @markbiek for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7723783).

PR #71826, opened 2023-01-10
E2E: refactor methods in `RestAPIClient` used to delete a site.

worldomonation:
#### Proposed Changes

This PR is a performance refactor of the `RestAPIClient` methods `getAllSites` and `deleteSite` in light of the experience detailed in p1672955277744179/1672953171.452809-slack-C04J1601X97.

Key changes:
- use the less expensive call to `/all-domains/` which returns the list of domains registered to the user's account, to obtain a list of sites the user owns.
- simplify the parameters to `deleteSite` to require just the `siteID` and `siteURL`.
- use `filter` instead of a reverse `for` loop to search for a match of site to be deleted against the list of sites the account owns.

#### Testing Instructions

  - [x] Calypso E2E (desktop)
  - [ ] Pre-Release Tests

Additional verification if required:
1. pull the branch.
2. edit a spec with the `calypso-release` tag with `throw new Error()` somewhere in the steps.
3. run the spec and verify that sites are deleted as expected.

#### Pre-merge Checklist

Related to #

### DIFF
--- a/packages/calypso-e2e/src/rest-api-client.ts
+++ b/packages/calypso-e2e/src/rest-api-client.ts
@@ -13,7 +13,6 @@ import {
 import type { Roles } from './lib';
 import type {
 	AccountDetails,
-	SiteDetails,
 	BearerTokenResponse,
 	MyAccountInformationResponse,
 	AccountClosureResponse,
@@ -262,7 +261,7 @@ export class RestAPIClient {
 	 * Otherwise the active subscription must be first cancelled
 	 * or else the REST API will throw a HTTP 403 status.
 	 *
-	 * @param {SiteDetails} expectedSiteDetails Expected details for the site to be deleted.
+	 * @param { {id: number, domain: string}} targetSite Details for the target site to be deleted.
 	 * @returns {SiteDeletionResponse | null} Null if deletion was unsuccessful or not performed. SiteDeletionResponse otherwise.
 	 */
 	async deleteSite( targetSite: {
--- a/packages/calypso-e2e/src/rest-api-client.ts
+++ b/packages/calypso-e2e/src/rest-api-client.ts
@@ -180,21 +180,18 @@ export class RestAPIClient {
 	/* Sites */
 
 	/**
-	 * Gets the list of users's sites.
+	 * Gets the list of domains belonging to the user.
 	 *
-	 * This method returns an array of Site objects, where
-	 * each Site object exposes a few key pieces of data from
+	 * This method returns an array of DomainData objects, where
+	 * each object exposes a few key pieces of data from
 	 * the response JSON:
-	 * 	- ID
-	 * 	- name
-	 * 	- site description
-	 * 	- URL
-	 * 	- site owner
+	 * 	- domain
+	 * 	- blog id
 	 *
 	 * @returns {Promise<AllDomainsResponse>} JSON array of sites.
 	 * @throws {Error} If API responded with an error.
 	 */
-	async getAllSites(): Promise< AllDomainsResponse > {
+	async getAllDomains(): Promise< AllDomainsResponse > {
 		const params: RequestParams = {
 			method: 'get',
 			headers: {
@@ -281,7 +278,7 @@ export class RestAPIClient {
 			? targetSite.domain.replace( 'https://', '' )
 			: targetSite.domain;
 
-		const mySites: AllDomainsResponse = await this.getAllSites();
+		const mySites: AllDomainsResponse = await this.getAllDomains();
 
 		const match = mySites.domains.filter( ( site: DomainsMetadata ) => {
 			site.blog_id === targetSite.id && site.domain === targetDomain;
--- a/packages/calypso-e2e/src/rest-api-client.ts
+++ b/packages/calypso-e2e/src/rest-api-client.ts
@@ -273,8 +273,9 @@ export class RestAPIClient {
 			return null;
 		}
 
-		const targetDomain = targetSite.domain.startsWith( 'http' )
-			? targetSite.domain.replace( 'https://', '' )
+		const scheme = 'http://';
+		const targetDomain = targetSite.domain.startsWith( scheme )
+			? targetSite.domain.replace( scheme, '' )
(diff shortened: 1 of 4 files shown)


packages/calypso-e2e/src/rest-api-client.ts
```diff
@@ -185,10 +191,10 @@ export class RestAPIClient {
 	 * 	- URL
 	 * 	- site owner
 	 *
-	 * @returns {Promise<AllSitesResponse>} JSON array of sites.
+	 * @returns {Promise<AllDomainsResponse>} JSON array of sites.
```
worldomonation, line 190: The interface `AllDomainsResponse` is just a renaming of the original interface.

packages/calypso-e2e/src/types/rest-api-client.types.ts
```diff
@@ -43,16 +43,13 @@ export interface BearerTokenResponse {
...
 
-interface SiteMetadata {
-	ID: number;
-	name: string;
-	description: string;
-	URL: string;
-	site_owner: number;
+export interface DomainsMetadata {
+	blog_id: number;
+	domain: string;
```
worldomonation: The `/all-domains/` endpoint returns more data, but for E2E purposes the only important one are these.

packages/calypso-e2e/src/rest-api-client.ts
```diff
@@ -247,48 +253,64 @@ export class RestAPIClient {
...
+		// Every new testing site since 2021 has the prefix `e2e` of some sort.
+		if ( ! targetSite.domain.includes( 'e2e' ) ) {
 			return null;
 		}
 
-		const mySites = await this.getAllSites();
-		const myAccountInformation = await this.getMyAccountInformation();
+		const targetDomain = targetSite.domain.startsWith( 'http' )
+			? targetSite.domain.replace( 'https://', '' )
+			: targetSite.domain;
```
worldomonation: This function strips the `https://` prefix from the site to create a slug.

This is because the `/all-domains/` endpoint only returns the site slug, while other endpoints that we normally work with `eg. /me/sites` return the fully qualified URL.

packages/calypso-e2e/src/rest-api-client.ts
```diff
@@ -247,48 +253,64 @@ export class RestAPIClient {
...
-	 * If the target site has an upgrade purchased within the
+	 * Additionally, if the target site has an upgrade purchased within the
 	 * sandboxed environment, it can be deleted without first
 	 * cancelling the subscription.
 	 *
 	 * Otherwise the active subscription must be first cancelled
-	 * otherwise the REST API will throw a HTTP 403 status.
+	 * or else the REST API will throw a HTTP 403 status.
 	 *
 	 * @param {SiteDetails} expectedSiteDetails Expected details for the site to be deleted.
```
dpasque: **Nit (optional)**: I would update the JSDOC comment here to match the new signature.

packages/calypso-e2e/src/rest-api-client.ts
```diff
@@ -247,48 +253,64 @@ export class RestAPIClient {
...
+	} ): Promise< SiteDeletionResponse | null > {
+		// Every new testing site since 2021 has the prefix `e2e` of some sort.
+		if ( ! targetSite.domain.includes( 'e2e' ) ) {
 			return null;
 		}
 
-		const mySites = await this.getAllSites();
-		const myAccountInformation = await this.getMyAccountInformation();
+		const targetDomain = targetSite.domain.startsWith( 'http' )
+			? targetSite.domain.replace( 'https://', '' )
```
dpasque: **Question (optional)**: Any reason to not have the `startsWith` match exactly what you are replacing?

packages/calypso-e2e/src/rest-api-client.ts
```diff
@@ -185,10 +191,10 @@ export class RestAPIClient {
 	 * 	- URL
 	 * 	- site owner
 	 *
-	 * @returns {Promise<AllSitesResponse>} JSON array of sites.
+	 * @returns {Promise<AllDomainsResponse>} JSON array of sites.
 	 * @throws {Error} If API responded with an error.
 	 */
-	async getAllSites(): Promise< AllSitesResponse > {
+	async getAllSites(): Promise< AllDomainsResponse > {
```
dpasque: **Suggestion (non-blocking)**: Can we rename this function (and maybe it's docs) to reference that it's getting all the domains?

Domains and sites aren't technically interchangeable entities, as users could purchase domains unattached to a site, and could have multiple domains for a single site.

I 10000% agree this is the right API call to use for this validation, it's like 10x faster!! 😮  But, I think we should just be explicit about the API we're using and add a comment saying that we're validating with domains because it's significantly faster 👍

dpasque (approved):
Few optional bits, otherwise LGTM! 🥳

packages/calypso-e2e/src/rest-api-client.ts
worldomonation: > Can we rename this function (and maybe it's docs) to reference that it's getting all the domains?

Valid point. While thankfully we don't have scenarios such as that in our testing environment, you are right that domains are not necessarily sites.

packages/calypso-e2e/src/rest-api-client.ts
worldomonation: Thanks, done in 5e1aaadf55006cb8a4f58a8d5a3570a08c306b3c.

packages/calypso-e2e/src/rest-api-client.ts
worldomonation: > Any reason to not have the startsWith match exactly

No particular reason, just wanted to save some keystrokes.

worldomonation: Updated in 47fc4e765e7979e4ad005d2dd0c80e6b0acfd859.

worldomonation:
> Few optional bits, otherwise LGTM! 🥳

Thanks! Implemented changes for all items you mentioned.

PR #71829, opened 2023-01-10
[GB 14.8+] Misc Calypso GB E2E test fixes

fullofcaffeine:
#### Proposed Changes

Fix misc issues with the WPCOM (Calypso) Gutenberg E2E tests. Some issues have not been introduced by 14.9.1, some might. This is being done as part of the release process for Gutenberg 14.9.1 on WPCOM.

#### Testing Instructions

* Calypso Gutenberg WPCOM simple site's E2E tests should pass.

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/71630

### DIFF
--- a/packages/calypso-e2e/src/lib/components/editor-settings-sidebar-component.ts
+++ b/packages/calypso-e2e/src/lib/components/editor-settings-sidebar-component.ts
@@ -31,9 +31,8 @@ const selectors = {
 
 	// Schedule
 	scheduleButton: `button.edit-post-post-schedule__toggle`,
-	scheduleInput: ( attribute: string ) => `role=spinbutton[name="${ attribute }"i]`,
+	scheduleInput: ( name: string ) => `.edit-post-post-schedule__dialog label:has-text("${ name }")`,
 	scheduleMeridianButton: ( meridian: 'am' | 'pm' ) => `role=button[name="${ meridian }"i]`,
-	scheduleMonthSelect: `role=combobox[name="month"i]`,
 
 	// Category
 	categoryCheckbox: ( categoryName: string ) =>
@@ -270,8 +269,8 @@ export class EditorSettingsSidebarComponent {
 			if ( key === 'month' ) {
 				// For month numbers less than 10, pad the digit to be
 				// 2 digits as required by the select.
-				const monthSelectLocator = this.editor.locator( selectors.scheduleMonthSelect );
-				await monthSelectLocator.selectOption( date[ key ].toString().padStart( 2, '0' ) );
+				const monthSelectLocator = this.editor.locator( selectors.scheduleInput( 'month' ) );
+				await monthSelectLocator.selectOption( ( date[ key ] + 1 ).toString().padStart( 2, '0' ) );
 				continue;
 			}
 			if ( key === 'date' ) {


packages/calypso-e2e/src/lib/components/editor-settings-sidebar-component.ts
```diff
@@ -270,8 +269,8 @@ export class EditorSettingsSidebarComponent {
 			if ( key === 'month' ) {
 				// For month numbers less than 10, pad the digit to be
 				// 2 digits as required by the select.
-				const monthSelectLocator = this.editor.locator( selectors.scheduleMonthSelect );
-				await monthSelectLocator.selectOption( date[ key ].toString().padStart( 2, '0' ) );
+				const monthSelectLocator = this.editor.locator( selectors.scheduleInput( 'month' ) );
+				await monthSelectLocator.selectOption( ( date[ key ] + 1 ).toString().padStart( 2, '0' ) );
```
fullofcaffeine, line 273: The select's values start with 01, and for some reason, `date['month']` was `00` for January, while in the select the value was `01`. Didn't dig too deeply into this, but the `+1` fixed the issue.

worldomonation: Weird, must have changed with this release? Anyway, as long as it's passing.

packages/calypso-e2e/src/lib/components/editor-settings-sidebar-component.ts
```diff
@@ -31,9 +31,8 @@ const selectors = {
 
 	// Schedule
 	scheduleButton: `button.edit-post-post-schedule__toggle`,
-	scheduleInput: ( attribute: string ) => `role=spinbutton[name="${ attribute }"i]`,
+	scheduleInput: ( name: string ) => `.edit-post-post-schedule__dialog label:has-text("${ name }")`,
```
worldomonation, line 34: **comment:non-blocking**: this may be slightly out of scope for this change and it is up to you whether you'd want to make the change.

@WunderBart has been replacing many selector templates like this with [role-based locators](https://playwright.dev/docs/locators) and I am a big fan of the change.

For instance, instead of having this CSS selector we can remove `scheduleInput` entirely, replacing its usage as follows:

```
await page.getByLabel('Day')

or

await page.getByLabel('Month')
```

right in the function call itself. It's one way to remove the need to constantly maintain selectors.

worldomonation (approved):
Approving to unblock merge progress, one non-blocking comment.

packages/calypso-e2e/src/lib/components/editor-settings-sidebar-component.ts
fullofcaffeine: That sounds good! The only downside is that they are not scoped to the post schedule dialog, so there's a small risk that it could select the wrong element if the page ever has two inputs with the same Label (which I think is unlikely in this context, in practice).

fullofcaffeine: I'll take note of that and do it in a small follow-up soon.

PR #71850, opened 2023-01-10
Theme Showcase: Theme preview limit global styles

taipeicoder:
#### Proposed Changes

This PR implements the limitation of global styles. New sites will have global styles limited to Premium plans (see p7DVsv-fY6-p2 for more context). As a result, whenever a user selects a style variation in the theme preview, the primary button will change to "Unlock this style". Clicking the button will open an overlay, where users can select to try the style or head to checkout. See screenshots below for reference:

**Selecting a style variation will change the primary button to "Unlock this style"**
Screen Shot 2023-01-10 at 9 38 01 PM

**Clicking the "Unlock this style" button will open the upgrade modal**
Screen Shot 2023-01-10 at 9 38 09 PM

#### Testing Instructions

* Open the preview of any theme with global style selection.
* Ensure that selecting any style other than the default will change the primary button to "Unlock this button".
* Ensure that clicking the "Unlock this button" will open the upgrade modal.
* Ensure that clicking the "Try it out first" button will open the theme switch modal.
* Ensure that clicking the "Upgrade plan" button will take users to the checkout page.
* Since this PR modifies the onboarding design picker, ensure that the design preview flow for limiting global styles works as before. 

#### Pre-merge Checklist

### DIFF
--- a/client/components/theme-preview-modal/index.tsx
+++ b/client/components/theme-preview-modal/index.tsx
@@ -4,7 +4,6 @@ import { useEffect, useMemo, useState } from '@wordpress/element';
 import classnames from 'classnames';
 import { useSelector } from 'react-redux';
 import AsyncLoad from 'calypso/components/async-load';
-import { useSiteGlobalStylesStatus } from 'calypso/state/sites/hooks/use-site-global-styles-status';
 import {
 	doesThemeBundleSoftwareSet as getDoesThemeBundleSoftwareSet,
 	isExternallyManagedTheme as getIsExternallyManagedTheme,
@@ -26,6 +25,7 @@ interface ThemePreviewModalProps {
 	theme: ThemeWithStyleVariations;
 	previewUrl: string;
 	actionButtons: React.ReactNode;
+	shouldLimitGlobalStyles: boolean;
 	selectedVariation?: StyleVariation;
 	onSelectVariation: ( variation: StyleVariation ) => void;
 	onClickCategory: ( category: Category ) => void;
@@ -36,6 +36,7 @@ const ThemePreviewModal: React.FC< ThemePreviewModalProps > = ( {
 	theme,
 	previewUrl,
 	actionButtons,
+	shouldLimitGlobalStyles,
 	selectedVariation,
 	onSelectVariation,
 	onClickCategory,
@@ -52,7 +53,6 @@ const ThemePreviewModal: React.FC< ThemePreviewModalProps > = ( {
 	const doesThemeBundleSoftwareSet = useSelector( ( state ) =>
 		getDoesThemeBundleSoftwareSet( state, theme.id )
 	);
-	const { shouldLimitGlobalStyles } = useSiteGlobalStylesStatus( siteId );
 	const [ selectedStyleVariation, setSelectedStyleVariation ] = useState< StyleVariation | null >(
 		selectedVariation || null
 	);
--- a/client/my-sites/themes/theme-preview.jsx
+++ b/client/my-sites/themes/theme-preview.jsx
@@ -1,16 +1,20 @@
 import { isEnabled } from '@automattic/calypso-config';
 import { Button } from '@automattic/components';
 import { getDesignPreviewUrl } from '@automattic/design-picker';
+import { createHigherOrderComponent } from '@wordpress/compose';
 import { localize } from 'i18n-calypso';
+import page from 'page';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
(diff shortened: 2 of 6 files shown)


client/components/theme-preview-modal/index.tsx
```diff
@@ -96,6 +98,10 @@ const ThemePreviewModal: React.FC< ThemePreviewModalProps > = ( {
 		};
 	}, [] );
 
+	useEffect( () => {
+		onShouldLimitGlobalStyles?.( shouldLimitGlobalStyles );
```
fushar: Wondering what is it for? I tried removing this `useEffect` block locally and I've found no issues.

taipeicoder: This is to notify the parent component wherever the `shouldLimitGlobalStyles` value changes. By default, it's set to `false` since it's pre API fetch. If the API returns `true`, we call the callback `onShouldLimitGlobalStyles` to let the parent component know that we should limit global styles. In this use case, we change the action button to "Unlock this style" when limiting global styles.

Let me know if that answers your question, and suggestions welcomed 🙂

fushar: I see, I find it a bit hard to understand because a parent component's behavior is dictated from API call done from a child component.

I was going to suggest that we pass the `shouldLimitGlobalStyles` prop from `ThemePreview` to `ThemePreviewModal`, but we can't since `ThemePreview` cannot use the `useSiteGlobalStylesStatus` hook because it's a class component 🤦 

Either we leave it as is for now, or refactor `ThemePreview` to be functional component 😄 your decision!

taipeicoder: > but we can't since ThemePreview cannot use the useSiteGlobalStylesStatus hook because it's a class component

I've run into this scenario many times already 🥲
Personally, I think it's a bit high-risk to refactor `ThemePreview`. Would be good to spend some time after the build stage of the initiative to rewrite a lot of these files to be functional components. What do you think if we leave a comment with an explanation for now? But to your point, I agree, components should be separate from data.

fushar: > I think it's a bit high-risk to refactor ThemePreview.

Makes sense due to this reason 😄

arthur791004: I think we don't need to refactor `ThemePreivew` component. Instead, we can replace its HOC such as connect with a functional component and inject the props what we want.

For example

```js
const ThemePreviewContainer = ( props ) => {
  const themeId = useSelector( themePreviewVisibility );
  const siteId = useSelector( getSelectedSiteId );

  ...

  return (
    <ThemePreview
      {...props}
      themeId={ themeId }
      siteId={ siteId }
      ...
      hideThemePreview={ () => dispatch( hideThemePreview() ) }
      ...
    />
  );
}
```

What do you think?

taipeicoder: I've refactored the component by using a HOC to use `useSiteGlobalStylesStatus()` in `<ThemePreview/>` 🙂

client/my-sites/themes/theme-preview.jsx
```diff
@@ -69,17 +113,31 @@ class ThemePreview extends Component {
...
-	getPrimaryOption = () => {
-		return this.props.themeOptions.primary;
+	onUnlockStyleButtonClick = () => {
+		this.setState( { showUnlockStyleUpgradeModal: true } );
 	};
 
-	getSecondaryOption = () => {
-		const { isActive } = this.props;
-		return isActive ? null : this.props.themeOptions.secondary;
+	onShouldLimitGlobalStyles = ( shouldLimitGlobalStyles ) => {
```
arthur791004: We can also remove this

client/my-sites/themes/theme-preview.jsx
```diff
@@ -169,8 +224,12 @@ class ThemePreview extends Component {
...
 							theme={ theme }
 							previewUrl={ this.getPreviewUrl() }
-							selectedVariation={ this.getStyleVariationOption() }
-							actionButtons={ this.renderPrimaryButton() }
+							selectedVariation={ selectedVariation }
+							actionButtons={
+								showUnlockStyleButton ? this.renderUnlockStyleButton() : this.renderPrimaryButton()
+							}
+							shouldLimitGlobalStyles={ shouldLimitGlobalStyles }
+							onShouldLimitGlobalStyles={ this.onShouldLimitGlobalStyles }
```
arthur791004: Same here

arthur791004 (approved):
LGTM 👍

Issue #71706, opened 2023-01-05
☂ Pattern Assembler flow in logged-out Theme Showcase


bangank36:
# Description
The logged-out theme showcase, which is access from `wordpress.com/themes` currently did not showing blank-canvas-3 ( BC) theme. This issue will bring the BC to theme showcase and navigate user to /patternAssembler step after picking BC and sign up

- POC video
<video src="https://user-images.githubusercontent.com/10071857/211517009-93e7d9e5-a09f-4df9-b40f-9bc48beb9840.mov"></video>

## Dev tasks
- [x] Display blank-canvas-3 theme in the returned result on theme showcase
       - Introduce feature flag for patternAssembler on theme showcase https://github.com/Automattic/wp-calypso/pull/71708
       - Update `query_themes` on wpcom to return blank-canvas-3 to logged-out showcase D97205-code
- [x]  Collect the query parameter to change the destination of `/with-theme` [flow](https://github.com/Automattic/wp-calypso/blob/ac3f356eeaa0dfd2d79d615942ef250e61f98fa3/client/signup/config/flows-pure.js#L87) https://github.com/Automattic/wp-calypso/pull/71708
- [x] Introduce new Assembler flow #72747 
- [x] [ Jan, 10 2023 ] Should display blank-canvas-3 theme as Blank Canvas CTA #72752 
- ~~[x]  Pre-select blank-canvas-3 design on /patternAssembler step to navigate user on [PA step after sign up](https://github.com/Automattic/wp-calypso/blob/8d8bbeb0fbcecd8e40ed640d9dc35d81fb2b03da/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx#L47) #71852~~
- ~~[x] Make sure user is landed on /site-editor after complete logged-out theme showcase flow #71852~~
- [x] [ Jan, 10 2023 ] Metrics for logged-out theme showcase flow

## Testing
* Log out from wp account and visit the `/themes` showcase 
* Choosing BC theme ( in a form of the blank canvas CTA )
* Finish sign up
* Then you should land on /patternAssembler step
* Finish adding patterns & land on /site-editor

## Reference
* Spike issue can be found [here](https://github.com/Automattic/dotcom-forge/issues/1383)

@Automattic/lego

autumnfjeld:
[...]

PR #71852, opened 2023-01-10
Make sure user lands on PatternAssembler after select blank-canvas

bangank36:
#### Proposed Changes

* This PR will address the solution to pre-select blank-canvas design for users that come from the logged-out theme showcase ( after they selected blank-canvas-3 theme )
* Also make sure user will be landed on /site-editor after PA step completed

#### Testing Instructions

* Checkout this branch, if you are using calypso live link, open the browser console and run this to make sure the feature flag is enabled `document.cookie = 'flags=pattern-assembler/logged-out-showcase;max-age=1209600;path=/'`
* Log out from your wp account
* Access theme showcase via /themes
* Select blank-canvas-3 theme and confirm that /patternAssembler step is shown after finishing site creation
* Select other theme and confirm that the end of flow is /home
* Recording
<video src="https://user-images.githubusercontent.com/10071857/212860158-f8c88972-438a-472a-8d08-b1cfb933fa50.mp4"></video>

#### Pre-merge Checklist

Related to #71706

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -7,6 +7,7 @@ import { useDispatch as useReduxDispatch } from 'react-redux';
 import AsyncLoad from 'calypso/components/async-load';
 import DocumentHead from 'calypso/components/data/document-head';
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
+import { getSignupSelectedThemeSlug } from 'calypso/signup/storageUtils';
 import { requestActiveTheme } from 'calypso/state/themes/actions';
 import { useSite } from '../../../../hooks/use-site';
 import { useSiteIdParam } from '../../../../hooks/use-site-id-param';
@@ -24,24 +25,16 @@ import type { DesignRecipe, Design } from '@automattic/design-picker/src/types';
 import './style.scss';
 
 const PatternAssembler: Step = ( { navigation, flow } ) => {
-	const hardcodeDesign = {
+	const blankCanvasDesign = {
 		slug: 'blank-canvas-3',
 		title: 'Blank Canvas',
-		description:
-			'Blank Canvas is a barebones starter theme, stripped off of content templates but only a footer and a header.',
 		recipe: {
 			stylesheet: 'pub/blank-canvas-3',
 		},
 		verticalizable: false,
-		categories: [],
 		is_premium: false,
 		is_bundled_with_woo_commerce: false,
-		software_sets: [],
 		design_type: 'assembler',
-		style_variations: [],
-		features: [],
-		template: '',
-		theme: '',
 	};
 
 	const translate = useTranslate();
@@ -52,11 +45,12 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	const [ sectionPosition, setSectionPosition ] = useState< number | null >( null );
 	const incrementIndexRef = useRef( 0 );
 	const [ activePosition, setActivePosition ] = useState( -1 );
-	const { goBack, goNext, submit } = navigation;
+	const { goBack, goNext, goToStep, submit } = navigation;
 	const { setThemeOnSite, runThemeSetupOnSite, createCustomTemplate } = useDispatch( SITE_STORE );
 	const reduxDispatch = useReduxDispatch();
 	const { setPendingAction } = useDispatch( ONBOARD_STORE );
 	const selectedDesign = useSelect( ( select ) => select( ONBOARD_STORE ).getSelectedDesign() );
+	const signupSelectedThemeSlug = getSignupSelectedThemeSlug();
 	const { setSelectedDesign } = useDispatch( ONBOARD_STORE );
 	const intent = useSelect( ( select ) => select( ONBOARD_STORE ).getIntent() );
 	const site = useSite();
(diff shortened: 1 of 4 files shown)


bangank36:
Update: for some reason, the change only work on localhost but not Calypso Live, will need to figuring why...

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -64,7 +79,11 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	useEffect( () => {
 		// Require to start the flow from the first step
 		if ( ! selectedDesign ) {
-			goToStep?.( 'goals' );
+			if ( signupSelectedThemeSlug !== 'blank-canvas-3' ) {
```
miksansegundo, line 66: We can use `isBlankCanvasDesign( signupSelectedThemeSlug )` from `@automattic/design-picker`.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -25,6 +26,18 @@ import type { DesignRecipe, Design } from '@automattic/design-picker/src/types';
 import './style.scss';
 
 const PatternAssembler: Step = ( { navigation, flow } ) => {
+	const blankCanvasDesign = {
+		slug: 'blank-canvas-3',
+		title: 'Blank Canvas',
+		recipe: {
+			stylesheet: 'pub/blank-canvas-3',
+		},
+		verticalizable: false,
```
miksansegundo: We can move `blankCanvasDesign` to the file `./constants`. Also, I think we can remove all the `false` props.

bangank36:
Update: the Calypso Live URL require the flag to be persisted on URL, but since the theme showcase requires a redirect, we need to enable the feature via cookie p1673950877830519/1673944921.451349-slack-CRWCHQGUB

`document.cookie = 'flags=pattern-assembler/logged-out-showcase;max-age=1209600;path=/'`


https://user-images.githubusercontent.com/10071857/212875769-d7c7778d-268e-4a3f-aa79-f7b58ea3c9f2.mp4

client/signup/controller.js
```diff
@@ -295,6 +296,11 @@ export default {
 		if ( ! isEmpty( additionalDependencies ) ) {
 			context.store.dispatch( updateDependencies( additionalDependencies ) );
 		}
+		if ( themeParameter ) {
+			setSignupSelectedThemeSlug( query.theme );
```
miksansegundo: I'd use an empty string `''` instead of `null` because the `sessionStorage` saves a string but in your case that works too.

Screenshot 2566-01-17 at 17 27 05

To save some characters:

```
setSignupSelectedThemeSlug( themeParameter || '' )
```

miksansegundo (commented):
Maybe it's my environment, but It didn't work on the first attempt. I'll try again.

client/signup/controller.js
arthur791004: I think it would be better to set the query string via the `getChecklistThemeDestination` function because that's why we store the `themeParameter` in `additionalDependencies` (L290 - L294)

See https://github.com/Automattic/wp-calypso/blob/bac42aae3c2c623f5028c7498416267dd83551cb/client/signup/config/flows.js#L115

arthur791004: Besides, I think it would be better to create a new flow and include only the PA step. It's a bit weird to redirect the user to a certain step of the flow after the signup. What do you think?

For example, [podcast flow](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/podcasts.ts#L8), [free flow](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/free.ts#L28), etc

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -64,7 +79,11 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	useEffect( () => {
 		// Require to start the flow from the first step
 		if ( ! selectedDesign ) {
-			goToStep?.( 'goals' );
+			if ( signupSelectedThemeSlug !== 'blank-canvas-3' ) {
+				goToStep?.( 'goals' );
+			}
+			// User has selected blank-canvas-3 theme from theme showcase
+			setSelectedDesign( blankCanvasDesign as Design );
```
arthur791004: this has to be wrapped inside the `else` condition to ensure we won't save the blank canvas design after `goToStep` executed

client/signup/controller.js
mmtr: > create a new flow and include only the PA step

That would be extremely helpful for https://github.com/Automattic/wp-calypso/pull/72286 too!

miksansegundo:
Just noting here that team T-Rex is working on adding a direct link from the theme research results to access the assembler directly. See https://github.com/Automattic/dotcom-forge/issues/1453

Screenshot 2566-01-23 at 10 07 40

Let's keep in mind that the assembler flow won't be not only a _site creation flow_ but also a _theme creation flow_ to design a new homepage on existing sites and creating a new flow with only the assembler step will help to skip tweaks that are only for the site creation use case like the redirection to ensure users have an `intent` saved in the `goals` step. cc: @autumnfjeld

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
bangank36: @miksansegundo `signupSelectedThemeSlug` is a string while `isBlankCanvasDesign` expects a Design object, so I don't think we can use it like suggested

bangank36:
> Besides, I think it would be better to create a new flow and include only the PA step. It's a bit weird to redirect the user to a certain step of the flow after the signup. What do you think?

> For example, [podcast flow](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/podcasts.ts#L8), [free flow](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/free.ts#L28), etc

This is brilliant idea, let's see what I can do

bangank36:
The issues of this PR will be addressed on #72747

PR #68566, opened 2022-10-04
Plugin Details: removes plugin details layout feature flag and unused legacy code.

cpapazoglou:
After the new Plugin Details being in production more than 1 month, we can safely do a clean up.

#### Proposed Changes

* removes `plugins/plugin-details-layout` feature flag
* removes unused legacy code

#### Testing Instructions

Test the following:
- Plugin Details page still looks the same for installed / not installed and wp.org / wp.com plugins.
- No console errors

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

PR #71864, opened 2023-01-10
Plugins: Remove unused legacy components

tyxla:
#### Proposed Changes

This PR is a follow-up cleanup after #68566 that removes a bunch of unused old plugins components.

#### Testing Instructions

Verify removed code is not in use.


epeicher (approved):
I have gone through all the files and checked if they are used elsewhere and I have not found any occurrences. I have also not found unexpected errors or warnings in the Dev Console when opening some plugin details pages.

I will leave others who know more in detail about those components to review the changes.

Given the above, changes Look Good to Me!

tyxla:
Thanks, @epeicher! I'll proceed with merging, and we can always grab a component from the repo history if we end up needing it. History shows that that happens rarely!

PR #71877, opened 2023-01-10
update Newspack Blocks to v1.60.0

### LINKED ISSUE #71842: [Bug]: [Block] Blog Posts Block: More posts button missing styles - incorrect syntax
### Quick summary

When the Blog Post block is in use: https://wordpress.com/support/wordpress-editor/blocks/blog-posts-block/ and the More Posts button is active, this button is missing styles.

Looking at inspector tools the console shows: 

`Uncaught SyntaxError: Unexpected token '<'
`
Looking at the HTML, it looks like the code is indeed missing a `/> ` making the closing button tag problematic. 

```
<button type="button" data-next="https://public-api.wordpress.com/wpcom/v2/sites/213069933/articles?className=is-style-default&amp;showReadMore=1&amp;imageShape=square&amp;moreButton=1&amp;moreButtonText=MORE%20POSTS&amp;showAuthor=0&amp;showCategory=1&amp;postLayout=grid&amp;postsToShow=9&amp;typeScale=3&amp;postType%5B0%5D=post&amp;align=wide&amp;showExcerpt=1&amp;excerptLength=55&amp;readMoreLabel=Keep%20reading&amp;showDate=1&amp;showImage=1&amp;showCaption=0&amp;minHeight=0&amp;showAvatar=1&amp;columns=3&amp;mediaPosition=top&amp;&amp;&amp;&amp;&amp;&amp;&amp;imageScale=3&amp;mobileStack=0&amp;sectionHeader&amp;specificMode=0&amp;textColor&amp;customTextColor&amp;singleMode=0&amp;showSubtitle=0&amp;page=2">
				MORE POSTS				</button>

```

### Steps to reproduce

1. Go to page, add new 
2. Add Blog Post block, if there are 6 blog posts in the site, set to show three posts 
3. Enable the "Show Load More Posts button" 
4. Publish page

To see the More posts button with styles as seen in the editor on the live site. 

### What actually happened

More Posts button is missing styles on the live site. 

### Browser

Google Chrome/Chromium, Mozilla Firefox, Microsoft Edge, Apple Safari

### Context

https://wordpress.com/forums/topic/display-not-matching-the-editor/?view=all

Simple, Atomic

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

All

### Available workarounds?

None

### Workaround details

_No response_

dkoo:
### Proposed Changes

* Updates Newspack Blocks in Editing Toolkit from v1.37.1 to v1.60.0. This should resolve #71842 as well as bring the fix from https://github.com/Automattic/newspack-blocks/pull/1173 to ETK.

With props to @noahtallen for helping us debug and update the sync script now that Newspack Blocks uses TypeScript in its source.

### Calypso updates to get it working
- Update peer dependencies, and remove a bunch of old ones we don't need any more.
- Remove swiper patch since it has no effect any more. Also remove prettier resolution which does nothing.
- Pin the stripe-js version to avoid updating it and breaking types.
- Add missing swiper dependency to a handful of packages.
- Force resolution of `wordpress` deps for newspack blocks. Otherwise, they'll get out of sync with the repo and pull in updates which break Calypso.
- Sync index.d.ts file from newspack-blocks, and adjust it to work with Calypso.
- Pin new swiper version in ETK to match newspack-blocks. Otherwise, the ETK build will resolve the repo-wide version, which is 4.5. This is because we don't consume newspack blocks as an npm dependency, so yarn/webpack doesn't resolve its dependencies normally. Instead, when linking newspack blocks code, the dependency version from ETK is used if it exists.

### Testing Instructions

* Confirm that the Blog Posts block continues to work, and that #71842 and https://github.com/Automattic/newspack-blocks/pull/1173 are resolved.
* Verify that wpcom phpunit tests pass before merging

#### Pre-merge Checklist

Related to #


.yarnrc.yml
```diff
@@ -132,48 +128,9 @@ packageExtensions:
   "@types/wordpress__plugins@3.0.0":
     peerDependencies:
       react: "*"
-  "@wordpress/block-library@6.0.18":
```
noahtallen, line 135: Nice! I remember looking forward to this update knowing we could get rid of a ton of these extra WordPress rules for older packages. A while ago, I updated gutenberg to correctly specify peer dependencies, and the main consumer of the older versions was newspack-blocks -- until now!

noahtallen:
I think yarn is upgrading `@wordpress/` deps within the allowed semver ranges. This is bringing along (among other things) the react 18 upgrade in some `@wordpress/` packages. We need to make sure yarn is keeping those versions the same as they already were.

package.json
```diff
@@ -292,8 +292,6 @@
 	},
 	"resolutions": {
 		"react-dates": "17.1.1",
-		"@automattic/newspack-blocks/swiper@4.5.1": "patch:swiper@4.5.1#.patches/swiper@4.5.1.diff",
-		"@automattic/newspack-blocks/prettier": "npm:wp-prettier@2.6.2-beta-1",
```
noahtallen, line 296: Prettier isn't included in the dependency tree here anymore.

package.json
```diff
@@ -292,8 +292,6 @@
 	},
 	"resolutions": {
 		"react-dates": "17.1.1",
-		"@automattic/newspack-blocks/swiper@4.5.1": "patch:swiper@4.5.1#.patches/swiper@4.5.1.diff",
```
noahtallen, line 295: This diff was removing a self-promotional message from the package which doesn't exist in the new version, so the patch isn't needed.

package.json
```diff
@@ -292,9 +292,11 @@
...
-		"@automattic/newspack-blocks/prettier": "npm:wp-prettier@2.6.2-beta-1",
 		"@types/react": "^17.0.39",
+		"newspack-components/@wordpress/base-styles": "4.5.0",
+		"newspack-components/@wordpress/components": "19.15.0",
+		"newspack-components/@wordpress/element": "4.7.0",
+		"newspack-components/@wordpress/i18n": "4.9.0",
 		"keytar@npm:7.7.0/node-addon-api": "3.1.0",
```
noahtallen, line 300: This is an interesting change.

In wp-calypso, we currently use `@wordpress/components` v19.15.0 (which is what's resolved in the yarn.lock file).

newspack-components (a dependency of newspack-blocks) uses `@wordpress/components` v19.0.1. Instead of using our _existing_ resolution -- 19.15.0 -- yarn instead decides to get a more recent version of `@wordpress/components` which depends on other newer versions of WordPress packages. At least one of these WordPress packages isn't considered a breaking change, so it gets updated to a _very_ recent version of WordPress components.

Unfortunately, `@wordpress/*` packages don't work very well when versions are somewhat misaligned. In this case, we end up pulling in the react 18 dependency, which also ends up completely upgrading some `@type/` packages with poorly specified version constraints (e.g. >= 16.0), and we end up with a fair number of dependencies changing in the tree.

All of which means, without this, the dependency upgrades would be incompatible with Calypso.

noahtallen:
@dkoo, if you run `yarn tsc --build apps/editing-toolkit/tsconfig.json` in calypso (in apps/editing-toolkit), there are a lot of type issues in the synced newspack blocks. Any thoughts on how to fix them?

package.json
noahtallen: Note: I want to double test the assumption that this broke things once we get everything working. Edit: yep, it seems we can't remove this.

packages/pattern-picker/package.json
```diff
@@ -33,6 +33,7 @@
 		"@automattic/onboarding": "workspace:^",
 		"classnames": "^2.3.1",
 		"react-query": "^3.32.1",
+		"swiper": "^8.4.5",
```
noahtallen: This package (and design-carosel) were relying on the fact that `swiper` was in the dependency tree, without actually depending on them.

packages/calypso-stripe/package.json
```diff
@@ -36,7 +36,7 @@
 	"homepage": "https://github.com/Automattic/wp-calypso/tree/HEAD/packages/calypso-stripe#readme",
 	"dependencies": {
 		"@stripe/react-stripe-js": "^1.4.1",
-		"@stripe/stripe-js": "^1.17.1",
+		"@stripe/stripe-js": "1.17.1",
```
noahtallen, line 39: This needed to be pinned, because the semver range makes it compatible with Stripe 1.46. Unfortunately, our typescript types are _not_ compatible with Stripe 1.46. By pinning it, we make Calypso resolve Stripe 1.17 as it should.

package.json
dkoo: It sounds like upgrading `@wordpress/components` in Newspack Components to v19.15.0 would solve this in the future?

Issue #71862, opened 2023-01-10
WooExpress Trial: Hosting Configuration


sixhours:
The screen at `/hosting-config/[siteSlug]` includes an upsell to the Business plan that should be updated for the trial context:

![screen-shot-2023-01-09-at-9 37 02-am](https://user-images.githubusercontent.com/2124984/211576471-0322b702-3bb5-478d-9ab3-5281ab77b975.png)

sixhours:
This has been resolved in #71881

PR #71881, opened 2023-01-10
Hosting Config: Update upsell message to be less specific about plan name

sixhours:
#### Proposed Changes

* Show a different upsell for users on the eCommerce Trial, bypassing the new experiment introduced in #71750 
* Update copy for upsell to *not* specify a plan name, since there are multiple upgrade paths for this feature
* Instead of linking the upsell directly to checkout with the Business plan in the cart, link to the `/plans` page

**eCommerce Trial**
Screen Shot 2023-01-16 at 10 11 48 AM

**eCommerce/Business Plan**

Screen Shot 2023-01-16 at 10 11 13 AM

**Experiment variation for everyone else**

(This may differ depending on which experiment id you're assigned to)

Screen Shot 2023-01-16 at 10 10 54 AM

#### Testing Instructions

* Switch to this PR, visit `/hosting-config/[siteSlug]`
* Confirm the upsell only shows for sites on the Free, Pro, Premium, and eCommerce Trial plans.
* Confirm the wording has changed for eCommerce Trial plans.
* Confirm the upsell links to the `/plans` page for eCommerce Trial plans.

#### Pre-merge Checklist

- N/A [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- N/A Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- N/A For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #71862

### DIFF
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -1,6 +1,10 @@
+<<<<<<< HEAD
 import { PLAN_BUSINESS, FEATURE_SFTP } from '@automattic/calypso-products';
 import { englishLocales } from '@automattic/i18n-utils';
 import { Spinner } from '@wordpress/components';
+=======
+import { FEATURE_SFTP } from '@automattic/calypso-products';
+>>>>>>> ea1876e34c (Update Hosting Config upsell message to be less specific about plan name)
 import { localize } from 'i18n-calypso';
 import { Component, Fragment } from 'react';
 import wrapWithClickOutside from 'react-click-outside';
@@ -80,24 +84,33 @@ class Hosting extends Component {
 		} = this.props;
 
 		const getUpgradeBanner = () => (
-			<Experiment
-				name="calypso_hosting_configuration_upsell_list_features"
-				defaultExperience={
-					<UpsellNudge
-						title={ translate( 'Upgrade to the Business plan to access all hosting features' ) }
-						event="calypso_hosting_configuration_upgrade_click"
-						href={ `/checkout/${ siteId }/business` }
-						plan={ PLAN_BUSINESS }
-						feature={ FEATURE_SFTP }
-						showIcon={ true }
-					/>
-				}
-				treatmentExperience={ <HostingUpsellNudge siteId={ siteId } /> }
-				loadingExperience={ <Spinner className="hosting__upsell-experiment-spinner" /> }
-				options={ {
-					isEligible: englishLocales.includes( locale ),
-				} }
-			/>
+			<>
+				<Experiment
+					name="calypso_hosting_configuration_upsell_list_features"
+					defaultExperience={
+						<UpsellNudge
+							title={ translate( 'Upgrade to the Business plan to access all hosting features' ) }
+							event="calypso_hosting_configuration_upgrade_click"
+							href={ `/checkout/${ siteId }/business` }
+							plan={ PLAN_BUSINESS }
+							feature={ FEATURE_SFTP }
+							showIcon={ true }
+						/>
+					}
+					treatmentExperience={ <HostingUpsellNudge siteId={ siteId } /> }
+					loadingExperience={ <Spinner className="hosting__upsell-experiment-spinner" /> }
+					options={ {
+						isEligible: englishLocales.includes( locale ),
+					} }
+				/>
+				<UpsellNudge
+					title={ translate( 'Upgrade your plan to access hosting features' ) }
+					event="calypso_hosting_configuration_upgrade_click"
+					href={ `/plans/${ siteSlug }/` }
+					feature={ FEATURE_SFTP }
+					showIcon={ true }
+				/>
+			</>
 		);
 
 		const getAtomicActivationNotice = () => {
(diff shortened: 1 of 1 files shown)


client/my-sites/hosting/main.js
```diff
@@ -76,10 +76,9 @@ class Hosting extends Component {
 
 		const getUpgradeBanner = () => (
 			<UpsellNudge
-				title={ translate( 'Upgrade to the Business plan to access all hosting features' ) }
+				title={ translate( 'Upgrade your plan to access all hosting features' ) }
 				event="calypso_hosting_configuration_upgrade_click"
-				href={ `/checkout/${ siteId }/business` }
-				plan={ PLAN_BUSINESS }
+				href={ `/plans/${ siteSlug }/` }
```
daledupreez: I think we should leverage the `feature` URL parameter to ensure that the correct feature and plan(s) are highlighted in the plans page.
Also, I think the URL used in the left nav does not include a trailing `/`.
```suggestion
				// import { FEATURE_SFTP_DATABASE } from '@automattic/calypso-products';
				href={ `/plans/${ siteSlug }?feature=${ encodeURIComponent( FEATURE_SFTP_DATABASE ) }` }
```

client/my-sites/hosting/main.js
```diff
@@ -76,10 +76,9 @@ class Hosting extends Component {
...
 				feature={ FEATURE_SFTP }
```
daledupreez: As above, I think we should use `FEATURE_SFTP_DATABASE` to identify the higher-level set of features rather than just SFTP.

daledupreez (commented):
I think the wording here is sufficiently generic for that to work for all use cases, and I think it is better to push users to the plan selection page as a starting point.

I do think we should highlight the features that need to be enabled, especially for users who are on the Free or Personal plans, which requires a minor change to the URL we create and the feature we're tracking.

client/my-sites/hosting/main.js
sixhours: I updated the import to use the broader feature set and added it as a query param. I thought there was supposed to be a visible highlight on the Plans page based on that query param but I'm not seeing one. Maybe I'm misremembering!

sixhours: @daledupreez Would it make sense to add the `&plan=${ encodeURIComponent( PLAN_BUSINESS ) }` query so we're only showing Business + eComm (plans that have the feature) when users land on the Plans page?

Screen Shot 2023-01-11 at 9 45 41 AM

sixhours:
i18n bot suggested this string which we already have translations for, which would make this faster to ship:

`Upgrade your plan to access this feature and more`

Thoughts?

client/my-sites/hosting/main.js
daledupreez: This feels like it might cause us problems if we go down this path, especially if we're not going to show the Business plan to users within the free eCommerce trial.

client/my-sites/hosting/main.js
```diff
@@ -76,11 +76,10 @@ class Hosting extends Component {
 
 		const getUpgradeBanner = () => (
 			<UpsellNudge
-				title={ translate( 'Upgrade to the Business plan to access all hosting features' ) }
+				title={ translate( 'Upgrade your plan to access this feature and more' ) }
```
daledupreez: I prefer the more specific `Upgrade your plan to access all hosting features` or even `Upgrade your plan to access advanced hosting features`.

In this context, we show the banner above what may be one or more features, so "this feature and more" doesn't make it clear what "this feature" is.

daledupreez (approved):
Re-tested and things look good to me, with the exception of my proposed wording change (which falls back on text you had earlier). Let's get the String Freeze label applied so we can get this translated next week.

daledupreez:
Hmmm... it looks like there is an A/B test kicking off around this component - the merge conflicts stem from #71750. We're going to need to ensure we exclude eCommerce trial users from that experiment, and see whether we can either update the control copy or add a completely separate implementation for the eCommerce trial (which is what we were trying to avoid! 🙃 ).

client/my-sites/hosting/main.js
```diff
@@ -76,11 +76,10 @@ class Hosting extends Component {
 
 		const getUpgradeBanner = () => (
 			<UpsellNudge
```
daledupreez: With the changes from the A/B test, I think we may want to add an early return to this function that returns the generic language when we're in the eCommerce trial, and let the experiment continue for all other plans. Flagged for the experimenting team here: pbxNRc-2f8-p2#comment-4030

p-jackson: I agree with doing an early return if the user is on an eCommerce trial. I'm going to launch the experiment as soon as I'm back on Monday so I'd like to keep the experiment as designed, which includes linking directly to the checkout.

Once we know which variation works better then I think we can work to make it make sense for eCommerce trial users.

markbiek (approved):
Works as expected. LGTM!

PR #71886, opened 2023-01-10
WooExpress Trial: Add selector to check if the site is the trial period

valterlorran:
#### Proposed Changes

* Add a selector to check if the site is in the trial period. We will use this selector to adapt the texts and behaviors of the actions in the admin.

#### Testing Instructions

* The selector is not being used yet, but I added a few unit tests. You can run it with the following command:
```bash
 yarn test-client client/state/sites/plans/test/selectors.js
```


client/state/sites/selectors/index.js
```diff
@@ -60,6 +60,7 @@ export { default as isRequestingSites } from './is-requesting-sites';
 export { default as isSearchPluginActive } from './is-search-plugin-active';
 export { default as isSingleUserSite } from './is-single-user-site';
 export { default as isSiteConflicting } from './is-site-conflicting';
+export { default as isSiteOnWooExpressTrailPeriod } from './is-site-on-woo-express-trail-period';
```
sixhours: Minor but `Trail` and `trail` should be `Trial` and `trial`

client/state/sites/selectors/is-site-on-woo-express-trial-period.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getPurchases } from 'calypso/state/purchases/selectors';
+import type { AppState } from 'calypso/types';
+
+/**
+ * Checks whether the current site is on the Woo Trial period.
+ *
+ * @param {AppState} state Global state tree
+ * @returns {?boolean} Returns true if the site is on the trial period
+ */
+export default function isSiteOnWooExpressTrialPeriod( state: AppState ) {
```
daledupreez: Some general thoughts here:
1. It may be simpler to call this `isSiteOnWooExpressTrial()` - the use of `Period` feels like it implies that we should be checking the expiration of the plan as well
2. It may be worth borrowing logic similar to the code in [the isSiteOnPaidPlan() selector](https://github.com/Automattic/wp-calypso/blob/c8fb9b13f62074a61430f6fc84076e688ef5c0a4/client/state/selectors/is-site-on-paid-plan.js#L13), which relies on the `getCurrentPlan()` selector and the `isFreePlan()` function from `@automattic/calypso-products` -- we'd need an `isEcommerceTrialPlan()` helper function for that approach.
3. Might this selector make more sense in `client/state/sites/plans/selectors/`?

client/state/sites/test/selectors.js
```diff
@@ -3642,4 +3643,42 @@ describe( 'selectors', () => {
...
+					isFetchingUserPurchases: false,
+					hasLoadedSitePurchasesFromServer: false,
+					hasLoadedUserPurchasesFromServer: true,
+				},
+			};
+
+			expect( isSiteOnWooExpressTrialPeriod( state ) ).toBeTruthy();
+		} );
+
+		test( 'Should return false when the e-commerce trial is not in the purchases list', () => {
```
daledupreez: It might be worth adding test cases where we have a non-plan purchase and a plan purchase that isn't the eCommerce trial -- the main eCommerce plan might be the best way to test that.

client/state/sites/plans/test/selectors.js
```diff
@@ -695,4 +700,90 @@ describe( 'selectors', () => {
 			).toEqual( PLAN_PREMIUM );
 		} );
 	} );
+
+	describe( 'isSiteOnECommerceTrial()', () => {
```
sixhours: Minor, looks like we tend to use a `#` before each test description for the other tests here.

sixhours (approved):
Looks good!

PR #71892, opened 2023-01-10
Revert "Refactor getSiteFragment to add new exception for plan upsell path"

leonardost:
Reverts Automattic/wp-calypso#71766

This was breaking domain renewals - see p1673375464862939-slack-C03TY6J1A


leonardost:
@DustyReagan when you have a chance, could you please take a look at this? It seems #71766 broke domain renewals and possibly other places.

`client/my-sites/controller.js` line 529 was breaking because `siteFragment` is now a site ID instead of a string.

DustyReagan:
Having a look now, @leonardost. Thanks for the revert and calling it to my attention!

leonardost:
No worries, thanks for taking a look!

PR #71560, opened 2022-12-29
Theme showcase: disallow tier filtering on My Themes tab

fushar:
#### Proposed Changes

The launch of `themes/showcase-i4/search-and-filter` feature flag has a regression where we allow tier filtering on the My Themes tab. This PR fixes the regression by disallowing such case.

#### Testing Instructions

1. Create an Atomic site.
2. Go to `/themes`.
3. Verify that you CANNOT get the `My Themes` filter and `Free` / `Premium` tier go together. In such cases, the filter/tier (or both) will just go back to `All`:
   - [x] On My Themes filter, if we click on any other tier, the filter will go to All.
   - [x] On Free/Premium tier, if we click on My Themes filter, the tier will go to All.

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [x] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to:

- #71382

PR #71893, opened 2023-01-10
i18n: Fix loading of filter translations on the logged-out themes page

jornp:
#### Proposed Changes
This PR fixes the loading of the translations of the tier filters (All / Free / Premium) and static filters (All / My Themes) in the logged-out theme showcase. 

The translations were loaded in the `ThemeShowcase` component's constructor, during the assignment of `this.tiers` and `this.staticFilters`. At that point, the translations weren't loaded yet. Loading them through functions (`getTiers` and `getStaticFilters`) makes sure that the translations are returned correctly and will update in case they change asynchronously.

**EDIT:** We recently pushed a PR (#71924) that makes the translations from preloaded chunks load earlier, which (accidentally) fixes the issue in this PR too. Still, it would be good to move the `translate` calls out of the constructor.

| Before #71924 | After |
| ------- | ----- |
| image | image |

#### Testing Instructions

* The issue is only visible when using translation chunks, so run Calypso locally using `BUILD_TRANSLATION_CHUNKS=true ENABLE_FEATURES=use-translation-chunks yarn start`.
* Visit /themes while logged out in a non-English locale and verify that the translations are working.
* Visit /themes while logged in and verify that the translations are still working.
* Make sure the testing instructions in #71560 still work.

#### Pre-merge Checklist

Related to 544-gh-Automattic/i18n-issues.

### DIFF
--- a/client/my-sites/themes/theme-showcase.jsx
+++ b/client/my-sites/themes/theme-showcase.jsx
@@ -153,8 +153,6 @@ class ThemeShowcase extends Component {
 	};
 
 	getStaticFilters = () => {
-		// As the values of these filter is static and won't be changed, we use it to check a filter
-		// is a static filter directly in the `isStaticFilter` function.
 		return {
 			MYTHEMES: {
 				key: 'my-themes',
--- a/client/my-sites/themes/theme-showcase.jsx
+++ b/client/my-sites/themes/theme-showcase.jsx
@@ -120,7 +120,11 @@ class ThemeShowcase extends Component {
 		}
 	}
 
-	isStaticFilter = ( tabFilter ) => Object.values( this.getStaticFilters() ).includes( tabFilter );
+	isStaticFilter = ( tabFilter ) => {
+		return Object.values( this.getStaticFilters() ).some(
+			( staticFilter ) => tabFilter.key === staticFilter.key
+		);
+	};
 
 	getSubjectFilters = ( props ) => {
 		const { subjects } = props;
--- a/client/my-sites/themes/theme-showcase.jsx
+++ b/client/my-sites/themes/theme-showcase.jsx
@@ -58,29 +58,8 @@ class ThemeShowcase extends Component {
 		this.scrollRef = createRef();
 		this.bookmarkRef = createRef();
 
-		// As the values of these filter is static and won't be changed, we use it to check a filter
-		// is a static filter directly in the `isStaticFilter` function.
-		this.staticFilters = {
-			MYTHEMES: {
-				key: 'my-themes',
-				text: props.translate( 'My Themes' ),
-				order: 3,
-			},
-			ALL: {
-				key: 'all',
-				text: props.translate( 'All' ),
-				order: 4,
-			},
-		};
-
 		this.subjectFilters = this.getSubjectFilters( props );
 		this.subjectTermTable = getSubjectsFromTermTable( props.filterToTermTable );
-
-		this.tiers = [
-			{ value: 'all', label: props.translate( 'All' ) },
-			{ value: 'free', label: props.translate( 'Free' ) },
-			{ value: 'premium', label: props.translate( 'Premium' ) },
-		];
 
 		this.state = {
 			tabFilter: this.getTabFilterFromUrl( props.filter ),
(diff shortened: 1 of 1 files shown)


yuliyan (approved):
LGTM!

client/my-sites/themes/theme-showcase.jsx
```diff
@@ -137,7 +116,7 @@ class ThemeShowcase extends Component {
 		}
 	}
 
-	isStaticFilter = ( tabFilter ) => Object.values( this.staticFilters ).includes( tabFilter );
+	isStaticFilter = ( tabFilter ) => Object.values( this.getStaticFilters() ).includes( tabFilter );
```
fushar: Thanks for looking into it! I just realized that `translate()` in a constructor might not work.

However, unfortunately, this change breaks this `isStaticFilter()` function, as this function assumes that the static filters are constant. After this change, it always returns a new object, so `includes()` here always returns false.

More concretely, this change breaks the fix in this PR: https://github.com/Automattic/wp-calypso/pull/71560.

So, I have another idea:

In the constructor, we only construct the keys, e.g.:

```
this.staticFilterKeys = {
  MYTHEMES: 'my-themes',
  ALL: 'all',
};
```

Then, we construct the actual filter object in `getTabFilters()`:

```
return {
	...( shouldShowMyThemesFilter && {
		MYTHEMES: {
           key: this.staticFilters.MYTHEMES,
           text: props.translate(...),
	} ),
	ALL: {
       key: this.staticFilters.ALL,
       text: props.translate(...),
	} 
	...this.subjectFilters,
};
```

What do you think?

fushar (changes requested):

client/my-sites/themes/theme-showcase.jsx
jornp: Hi @fushar, thank you for spotting that!!

I like your suggestion! Am I understanding correctly that we'd then change `isStaticFilter` to check by key? In that case, could it be simpler to fix the functionality of #71560 by only changing `includes` to `some`, like this?

```
isStaticFilter = ( tabFilter ) => Object.values( this.getStaticFilters() ).some( ( staticFilter ) => tabFilter.key === staticFilter.key );
```

That way, we can keep the definition of the static filters together in `getStaticFilters`, while also leaving the functionality of #71560 intact. Let me know what you think! 🙂

fushar: I see... we can do that 😄

client/my-sites/themes/theme-showcase.jsx
```diff
@@ -155,15 +134,41 @@ class ThemeShowcase extends Component {
...
+				MYTHEMES: staticFilters.MYTHEMES,
 			} ),
-			ALL: this.staticFilters.ALL,
+			ALL: staticFilters.ALL,
 			...this.subjectFilters,
 		};
 	};
 
+	getStaticFilters = () => {
+		// As the values of these filter is static and won't be changed, we use it to check a filter
```
fushar: We can then remove this comment; it's not correct anymore

client/my-sites/themes/theme-showcase.jsx
jornp: Great, done! 🙂

client/my-sites/themes/theme-showcase.jsx
jornp: That makes sense, thanks, I've removed the comment!

fushar (approved):
👍 thanks for fixing!

arthur791004 (approved):
LGTM 👍

jornp:
@fushar Thanks for your input and help! @arthur791004 Thanks for the review!

noahtallen:
EDIT! Never mind, it was unrelated!

I think this corresponds to a severe performance issue in the `/themes` route. Logged out requests are performing about 10x worse now vs before the change was deployed. ([See here](https://mc.a8c.com/grafana/d/000000133/calypso-node-response-times-production?orgId=1&from=now-2d&to=now&viewPanel=9&var-aggregator=mean))

image

I'm not quite sure why, but digging in. In our logs, the slowest part is the bit which fetches theme data. 

I'm also not 100% sure it's this PR, but it lines up pretty closely with the performance issue and is related

noahtallen:
Disregard that comment -- I discovered that it corresponded to a new theme being added. Ultimately fixed here: https://github.com/Automattic/wp-calypso/pull/72248

PR #58634, opened 2021-11-30
i18n: eliminate the setRawLocaleData action

jsnajdr:
Less ambitious version of #58343, which tried to do too many things at once. Eliminates just the `setLocaleRawData` Redux action, which is not really a pure action, because it calls a `i18n.setLocale` side-effect. After this patch, all places that used to call `setLocaleRawData` (there were two), do the `i18n.setLocale` and `store.dispatch( LOCALE_SET )` statements separately.

This is a part of a series of PRs where I aim to clean up the i18n code a bit, so that I can eventually merge #58345 which moves locale-setting code for localized routes (ones that have locale prefix or suffix) from the `<LocaleSuggestions>` component to the page.js handlers.

PR #18297, opened 2017-09-26
Improve locale state in calypso

Tug:
This PR seeks to improve how locale is set or updated by moving the value from the context to the redux state.

#### Testing instructions
  
1. Run `git checkout update/lang` and start your server, or open a [live branch](https://calypso.live/?branch=update/lang)
2. Open the [`Login` page](http://calypso.localhost:3000/log-in/fr)
3. Assert that you see the correct language & no CSS is loaded twice
4. Assert that correct (RTL/LTR) CSS tags are received from the server ( view-source )

| User logged in | User language | URL Language Param | Result language |
|----------------|---------------|--------------------|-----------------|
| ✓              | HE            | FR                 | FR              |
| ✓              | HE            | HE                 | HE              |
| ✓              | FR            | FR                 | FR              |
| ✓              | HE            | N/A                | HE              |
| ✓              | FR            | N/A                | FR              |
| ✗              | N/A           | HE                 | HE              |
| ✗              | N/A           | FR                 | FR              |
| ✗              | N/A           | N/A                | EN              |
 
5. Assert that the Support User feature still works as intended and that you are switched to the user's locale.

* Note: Each test should be performed with and without `"wpcom-user-bootstrap": true`. See #18264 
#### Additional notes

* Commit [d54d453](https://github.com/Automattic/wp-calypso/pull/18297/commits/d54d4533df989969df678132195dbae6c1e5e8fd) is only for testing and should be removed before merging.
* SSR is only used for the default locale ( EN ).

#### Reviews
  
- [x] Code
- [x] Product
- [ ] Tests

PR #71895, opened 2023-01-10
format-currency: Set currency locale when changing translation locale

### LINKED ISSUE #71816: formatCurrency does not use WordPress.com user interface locale
https://github.com/Automattic/wp-calypso/pull/71387 modifies the `formatCurrency()` function to respect the browser's locale when deciding how to format a price in calypso. However, it's possible that the browser's locale might differ from the user's WordPress locale, which could cause situations where prices formatted by the server appear differently than prices formatted by the client.

For users, this means that, if their web browser is set to an English locale but they change their WordPress.com user interface language to Canadian French, prices will still be localized for English, displaying as something like `C$ 1,000.53` instead of the expected `1 000,53 C$` (this problem would be worse if we weren't [overriding currency symbols](https://github.com/Automattic/wp-calypso/pull/71387/commits/d44637f905f2b64b3014ce5726c510806df0c0ff)).

We should respect the WordPress.com user interface language for price formatting.

https://github.com/Automattic/wp-calypso/pull/71387 added a function, `setDefaultLocale()`, which can be used to set a global locale override for price formatting. There is also a Redux selector, `getCurrentLocaleSlug()` which seems like it would contain the WordPress locale of the user. We should be able to create a side effect sometime during calypso's rendering process to use these two functions to synchronize the locale. I'm just not sure where to put such an effect.

sirbrillig:
#### Proposed Changes

When changing the translation language of calypso, this PR adds a secondary side effect which also sets the locale used by the `formatCurrency()` function.

This side effect is added to the `setLocale()` redux action creator for the UI state, which is awkward because action creators should not have side effects. However, this places it alongside the other side effect (which sets the translation locale) which has been in this location since https://github.com/Automattic/wp-calypso/pull/18297... although reading https://github.com/Automattic/wp-calypso/pull/58634 I wonder if maybe I should put this someplace else? cc @jsnajdr 

Before (with en browser locale, WordPress.com user interface set to Canadian French, and currency set to CAD):

before-setting-currency-locale

After (with en browser locale, WordPress.com user interface set to Canadian French, and currency set to CAD):

after-setting-currency-locale

Fixes https://github.com/Automattic/wp-calypso/issues/71816

#### Testing Instructions

- Use a web browser with its locale set to `en`. (If you need to change your web browser's locale, you can use the devtool "Sensors" panel in Google Chrome.)
- Set your user's currency to CAD.
- Set your user's user interface language to "Canadian French" using the form at `/me/account`.

Screenshot 2023-01-10 at 4 07 48 PM

- Now add a product to your cart and visit checkout.
- Compare checkout with or without this PR. Verify that without this PR, the price is rendered with the currency symbol (eg: `C$123` which is the `en` formatting) before the number and that after this PR the currency symbol is rendered after the number (eg: `123 C$` which is the `fr-CA` formatting).


client/state/ui/language/actions.ts
```diff
@@ -0,0 +1,28 @@
...
+	localeVariant: string | null | undefined = null
+) => {
+	/**
+	 * Also change the translation and currency locale.
+	 *
+	 * This is not a great place to put a side effect because it's an action
+	 * creator, and should simply return an object, but it's been here for a long
+	 * time since https://github.com/Automattic/wp-calypso/pull/18297
+	 */
+	setCurrencyLocale( localeVariant || localeSlug );
```
sirbrillig, line 14: It's hard to see because of the TypeScript conversion, but this line is the only change to this function.

packages/format-currency/src/index.ts
```diff
@@ -65,17 +67,26 @@ function getCachedFormatter( {
...
-		// in FF so we have to strip zeros manually.
-		...( noDecimals ? { maximumFractionDigits: 0, minimumFractionDigits: 0 } : {} ),
-	} );
+	try {
+		const formatter = new Intl.NumberFormat( addNumberingSystemToLocale( locale ), {
+			style: 'currency',
+			currency,
+			// There's an option called `trailingZeroDisplay` but it does not yet work
+			// in FF so we have to strip zeros manually.
+			...( noDecimals ? { maximumFractionDigits: 0, minimumFractionDigits: 0 } : {} ),
```
tyxla, line 87: I wonder if it makes sense for the `Intl.NumberFormat` settings to be customizable when calling `formatCurrency()`.

packages/format-currency/src/index.ts
```diff
@@ -194,7 +205,11 @@ export function formatCurrency(
 ): string {
 	const locale = options.locale ?? defaultLocale ?? getLocaleFromBrowser();
 	if ( ! doesCurrencyExist( code ) ) {
-		code = 'USD';
+		// eslint-disable-next-line no-console
+		console.warn(
+			`formatCurrency was called with a non-existent currency "${ code }"; falling back to ${ fallbackCurrency }`
+		);
+		code = fallbackCurrency;
```
tyxla: This part seems a bit repetitive, does it make sense to abstract it to a helper function?

tyxla (approved):
LGTM 👍 🚀 

I've left a couple of minor questions/suggestions, but nothing is blocking IMHO.

Nice work!

packages/format-currency/src/index.ts
sirbrillig: By allowing passing in arbitrary options to `NumberFormat`? 🤔 I'm not sure. I think that makes for a bit of a leaky abstraction. If you want to use `NumberFormat`, a caller can just use it directly without needing this function. I _do_ think that there's the potential for better abstractions in this library in the future, like functions that perform some of these options by default but for now I'd opt to keep the options already present in `formatCurrency` unchanged in case we ever want to do things differently inside.

packages/format-currency/src/index.ts
sirbrillig: Good idea.

dlind1 (approved):
Tested with EUR/DE, LGTM!

Before:
Screenshot 2023-01-11 at 19 05 21

After:
Screenshot 2023-01-11 at 19 05 32

PR #71899, opened 2023-01-10
Preload Post Purchase Upsell experiment assignment

razvanpapadopol:
Related to pbxNRc-2eP-p2#comment-4003

#### Proposed Changes

* When adding a premium plan to cart, preload the 20989-explat-experiment assignment.

#### Testing Instructions

* Add a Premium plan to cart
* When the checkout page is loaded, a network request to the URL: `public-api.wordpress.com/wpcom/v2/experiments/0.1.0/assignments/calypso?_envelope=1&experiment_name=calypso_plans_postpurchasepage_newdesign_202301_v1` should be made. It returns an object with the correct experiment variation.

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -23,7 +23,7 @@ import { styled } from '@automattic/wpcom-checkout';
 import { useSelect, useDispatch } from '@wordpress/data';
 import debugFactory from 'debug';
 import { useTranslate } from 'i18n-calypso';
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useEffect } from 'react';
 import { useDispatch as useReduxDispatch, useSelector } from 'react-redux';
 import MaterialIcon from 'calypso/components/material-icon';
 import {
@@ -253,6 +253,12 @@ export default function WPCheckout( {
 
 	const isWcMobile = isWcMobileApp();
 
+	useEffect( () => {
+		if ( hasPremiumPlan( responseCart ) ) {
+			loadExperimentAssignment( 'calypso_plans_postpurchasepage_newdesign_202301_v1' );
+		}
+	}, [ responseCart ] );
+
 	if ( transactionStatus === TransactionStatus.COMPLETE ) {
 		debug( 'rendering post-checkout redirecting page' );
 		return (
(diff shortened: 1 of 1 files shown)


client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
```diff
@@ -297,6 +299,10 @@ export default function WPCheckout( {
 
 	const isDIFMInCart = hasDIFMProduct( responseCart );
 
+	if ( hasPremiumPlan( responseCart ) ) {
+		loadExperimentAssignment( 'calypso_plans_postpurchasepage_newdesign_202301_v1' );
```
sirbrillig: I don't know enough to be sure; is it safe to call `loadExperimentAssignment()` on every render?

sirbrillig (commented):
I didn't run through the tests (I leave that to your team) but from a checkout perspective this should be fine as long as loadExperimentAssignment is idempotent.

paulopmt1 (approved):
It's working great @razvanpapadopol! I tested the experiment page load with and without this code (coming from the /checkout page) and saw an average of 400ms difference between when we click the checkout button and when we see the upsell screen.


We're storing the data properly:

image

Profile with cache:

experiment with-cache


Profile without cache:

image

paulopmt1:
Sorry for the notification bomb folks, apparently I still didn't get the correct rebase workflow. Next time will use `--force-with-lease ` with the `push` command (I though it was only necessary when using the `force` parameter)

client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
razvanpapadopol: Good point, @sirbrillig. I've investigated a bit and it looks ok. As it's mentioned also in the [package readme](https://github.com/Automattic/wp-calypso/tree/trunk/packages/explat-client#usage):
> Call as many times and as much as you like, we manage the state and requests.

So even if the preloading function was designed to be used outside of the React application (signup controller was the first usage), it should be safe to call it on multiple renders.

But in a React Component any side effect should be explicitly defined, together with its dependencies. So I've wrapped the call in a `useEffect` and placed it before any of the component returns (https://github.com/Automattic/wp-calypso/pull/71899/commits/d7209c2721d4456bec6f3c2589a048f36b94f26e). There shouldn't be any noticeable change in behavior. The function will still get called multiple times when a new `responseCart` object is passed.

DustyReagan (approved):
I see the network request as well. Seems to be working! 👍

paulopmt1:
Before merging this, we'll need to configure the ongoing experiment to use triggers. Let me know when you feel it's good to go @razvanpapadopol so I'll go ahead and change the experiment + deploying this improvement.

Thanks!

paulopmt1:
I'll close this now, so we'll use it in follow-up experiments. Here's more context: p1673541619155519/1673289198.965569-slack-CKZHG0QCR

Thanks for all your work on it!

PR #71907, opened 2023-01-11
WooExpress Trial: Add selectors to handle the expiration date

valterlorran:
#### Proposed Changes

This PR adds three selectors we will use in the WooExpress Trial project.

* getWooExpressTrialExpiration - To get the expiration date of the WooExpress Trial
* getWooExpressTrialDaysLeft - Gets the number of days left of the trial
* isWooExpressTrialExpired - Checks whether the trial is expired.

#### Testing Instructions

* The selectors are not being used yet, but I added a few unit tests. You can run it with the following command:
```bash
 yarn test-client client/state/sites/test/selectors.js
```


client/state/sites/selectors/get-woo-express-trial-days-left.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @returns {null|number} Number of days left in the trial, or null if the trial is not active.
+ */
+export default function getWooExpressTrialDaysLeft( state: AppState ): number | null {
+	const trialExpirationDate = getWooExpressTrialExpiration( state );
+	if ( ! trialExpirationDate ) {
+		return null;
+	}
+
+	const expiryDate = +new Date( trialExpirationDate );
+	const diffTime = expiryDate - +new Date();
```
daledupreez: My main concern here is that the expiry date is almost certainly in UTC, while the user's time zone could be anything. We should handle that offset here.
Also, please explain the `+new` usage - it's not obvious to me what that is doing.

client/state/sites/selectors/get-woo-express-trial-expiration.ts
```diff
@@ -0,0 +1,21 @@
...
+export default function getWooExpressTrialExpiration( state: AppState ) {
+	const trialPurchase = getPurchases( state ).find(
+		( purchase ) => purchase.productSlug === PLAN_ECOMMERCE_TRIAL_MONTHLY
+	);
+
+	if ( ! trialPurchase ) {
+		return null;
+	}
+
+	return trialPurchase.expiryDate;
```
daledupreez: Thought: might it make sense for this function to return a `Date` object? That way callers don't need to take into account time zone offsets etc, and they have a much better return type to work with.

client/state/sites/selectors/get-woo-express-trial-expiration.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getPurchases } from 'calypso/state/purchases/selectors';
+import type { AppState } from 'calypso/types';
+
+/**
+ * Returns the expiration date of the Woo Express trial. If the trial is not active, returns null.
+ *
+ * @param {AppState} state - Global state tree
+ * @returns {string|null}
+ */
+export default function getWooExpressTrialExpiration( state: AppState ) {
```
daledupreez: Do we want to explicitly document the return type in the function definition rather than the JSDoc?
```suggestion
export default function getWooExpressTrialExpiration( state: AppState ): string | null {
```

client/state/sites/selectors/get-woo-express-trial-expiration.ts
```diff
@@ -0,0 +1,21 @@
...
+	const trialPurchase = getPurchases( state ).find(
+		( purchase ) => purchase.productSlug === PLAN_ECOMMERCE_TRIAL_MONTHLY
+	);
+
+	if ( ! trialPurchase ) {
+		return null;
+	}
```
daledupreez: Can we use the `getCurrentPlan()` selector here and then check whether the current plan is the eCommerce trial?
```suggestion
	// import { getCurrentPlan } from 'client/state/sites/plans/selectors';
	const currentPlan = getCurrentPlan( state, siteId );
	if ( ! currentPlan || ! isECommerceTrialPlan( currentPlan ) ) {
		return null;
	);

	return currentPlan.expiryDate;
```

client/state/sites/selectors/get-woo-express-trial-expiration.ts
daledupreez: Logic question: don't we need to filter the purchases by the current site?

client/state/sites/selectors/get-woo-express-trial-days-left.ts
```diff
@@ -0,0 +1,20 @@
+import { AppState } from 'calypso/types';
+import getWooExpressTrialExpiration from './get-woo-express-trial-expiration';
+
+/**
+ * Get the number of days left in the Woo Express trial. If the trial is not active, returns null.
+ *
+ * @param {AppState} state - Global state tree
+ * @returns {null|number} Number of days left in the trial, or null if the trial is not active.
+ */
+export default function getWooExpressTrialDaysLeft( state: AppState ): number | null {
```
daledupreez: General naming thought: it may be more future-proof to use `EcommerceTrial` instead of `WooExpressTrial` in all of these functions, as that gets us away from the branding and closer to the underlying entities. That would also align more clearly with the `ECOMMERCE_TRIAL` naming we are using elsewhere.

client/state/sites/selectors/get-woo-express-trial-days-left.ts
valterlorran: Thanks for catching it! I started using `moment` to facilitate it. The `+new` is basically to convert it to a timestamp.

client/state/sites/selectors/get-woo-express-trial-expiration.ts
valterlorran: As far as I can see, the current site scopes the purchases state, but I see some places filtering by the siteId. I will add it as well to make sure.

daledupreez:
I will take a look tomorrow if no-one else from Serenity beats me to it.

sixhours:
@valterlorran It looks like these need the same treatment as in #71886; replacing `WooExpress` with `ECommerce` and moving them under `/client/state/sites/plans/selectors`

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
 			expect( isSiteOnECommerceTrial( state, siteId ) ).toBeFalsy();
 		} );
 	} );
+
+	describe( '#getECommerceTrialExpiration()', () => {
+		const siteId = 1337;
+		test( 'Returns the expiracy date', () => {
```
markbiek:
```suggestion
		test( 'Returns the expiration date', () => {
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+			const expiracyDate = '2022-02-10T00:00:00+00:00';
```
markbiek:
```suggestion
			const expiryDate = '2022-02-10T00:00:00+00:00';
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+
+			const plan = {
+				ID: 1,
+				productSlug: PLAN_ECOMMERCE_TRIAL_MONTHLY,
+				blogId: siteId,
+				expiryDate: expiracyDate,
```
markbiek:
```suggestion
				expiryDate,
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+						},
+					},
+				},
+				siteSettings: {
+					items: {},
+				},
+			} );
+
+			expect(
+				getECommerceTrialExpiration( state, siteId ).isSame( moment( expiracyDate ) )
```
markbiek:
```suggestion
				getECommerceTrialExpiration( state, siteId ).isSame( moment( expiryDate ) )
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+			expect( getECommerceTrialExpiration( state, siteId ) ).toBeNull();
+		} );
+	} );
+
+	describe( '#getECommerceTrialDaysLeft()', () => {
+		const siteId = 1337;
+		jest.useFakeTimers().setSystemTime( new Date( '2022-01-10T00:00:00+00:00' ) );
+
+		test( 'Should return the correct number of days left before the trial expires', () => {
+			const expiracyDate = '2022-02-10T00:00:00+00:00';
```
markbiek:
```suggestion
			const expiryDate = '2022-02-10T00:00:00+00:00';
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+
+			const plan = {
+				ID: 1,
+				productSlug: PLAN_ECOMMERCE_TRIAL_MONTHLY,
+				blogId: siteId,
+				expiryDate: expiracyDate,
```
markbiek:
```suggestion
				expiryDate,
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+			expect( getECommerceTrialDaysLeft( state, siteId ) ).toBe( 31 );
+		} );
+	} );
+
+	describe( '#isECommerceTrialExpired()', () => {
+		const siteId = 1337;
+		jest.useFakeTimers().setSystemTime( new Date( '2022-01-10T00:00:00+00:00' ) );
+
+		test( 'The trial period should be expired', () => {
+			const expiracyDate = '2022-01-09T00:00:00+00:00';
```
markbiek:
```suggestion
			const expiryDate = '2022-01-09T00:00:00+00:00';
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+
+			const plan = {
+				ID: 1,
+				productSlug: PLAN_ECOMMERCE_TRIAL_MONTHLY,
+				blogId: siteId,
+				expiryDate: expiracyDate,
```
markbiek:
```suggestion
				expiryDate,
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+				siteSettings: {
+					items: {},
+				},
+			} );
+
+			expect( isECommerceTrialExpired( state, siteId ) ).toBeTruthy();
+		} );
+
+		test( 'The trial period should not be expired if is the same day', () => {
+			const expiracyDate = '2022-01-10T23:59:59+00:00';
```
markbiek:
```suggestion
			const expiryDate = '2022-01-10T23:59:59+00:00';
```

client/state/sites/plans/test/selectors.js
```diff
@@ -786,4 +790,175 @@ describe( 'selectors', () => {
...
+			const plan = {
+				ID: 1,
+				productSlug: PLAN_ECOMMERCE_TRIAL_MONTHLY,
+				blogId: siteId,
+				expiryDate: expiracyDate,
```
markbiek:
```suggestion
				expiryDate,
```

markbiek (approved):
LGTM!

PR #68631, opened 2022-10-05
Fix Jetpack Social discounted price by hard-coding it

manzoorwanijk:
#### Proposed Changes

This PR fixes the Jetpack Social discounted price which was considered to be a yearly discount. It hard-codes it to be a monthly discount.

#### Testing Instructions

- Do any one of these
    - Click on Jetpack Cloud live link below and wait for the redirect to complete, then go to `/pricing`
    - or boot up this PR 
        - Run `git fetch && git checkout fix/jetpack-social-price`
        - Run `yarn start-jetpack-cloud`
        - Goto [http://jetpack.cloud.localhost:3000/pricing](http://jetpack.cloud.localhost:3000/pricing)
- Confirm that the price for Jetpack Social is **$1 <s>$10</s>** instead of **$0.08 <s>$10</s>**

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Related to p1664891686656539-slack-C03RCHNVC0H

Screenshot 2022-10-05 at 11 48 01 AM

PR #71936, opened 2023-01-11
Fix display price if no intro offer is found

robertsreberski:
#### Proposed Changes

* Fix displaying subtle text when no intro offer applied

#### Testing Instructions

* Ensure that you use the production store
* Run calypso green env `yarn start-jetpack-cloud`
* Enter the pricing page `http://jetpack.cloud.localhost:3000/pricing`
* VaultPress Backup should have "/month, billed yearly" subtle text, instead "for the first month, billed yearly":
CleanShot 2023-01-11 at 16 57 30@2x

* Jetpack Social should be without a change:
CleanShot 2023-01-11 at 17 04 40@2x

Related to #68631

### DIFF
--- a/client/my-sites/plans/jetpack-plans/use-item-price.ts
+++ b/client/my-sites/plans/jetpack-plans/use-item-price.ts
@@ -176,7 +176,10 @@ const useItemPrice = (
 				)
 			) {
 				discountedPrice = introductoryOfferPrices.introOfferCost || undefined;
-				discountedPriceDuration = 1;
+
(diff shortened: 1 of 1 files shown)


IanRamosC (approved):
> VaultPress Backup should have "/month, billed yearly" subtle text, instead "for the first month, billed yearly":

The change looks good and work as expected. I'd recommend editing the description, as the main fix affects pretty much when using the production API.

I also took a quick look at the code to see if `discountedPriceDuration` would impact other places but it doesn't seem like it.

PR #71949, opened 2023-01-11
Add setting for writing prompt email notification

### LINKED ISSUE #71947: Writing Prompts - Add Email notification setting
### Details

We need to add a new email notification setting to a sites notifications settings.
This setting should be set to disabled by default, similar to the bell notification.

Screenshot 2023-01-11 at 17 53 18

The writing prompt guide will check this setting to determine if a user should receive an email notification for a particular site.

### Checklist

- [ ] Add new email notification setting
- [ ] Update guides to use setting

### Related

_No response_

eoigal:
#### Proposed Changes

* This adds a new setting for the writing prompts email notification.

The setting was not added in the V1 and was filtered out. This PR removes that filter and allows the setting to render.

Screenshot 2023-01-12 at 08 59 33

#### Testing Instructions
* Go to `/me/notifications` and check that the setting is visible
* The settings should be disabled by default
* Update 0-sandbox.php to include the following;
```
function sandbox_jobs( $sandboxed, $type ) {
	return in_array( $type, [ 'notification_settings_persistance' ] ); ;
}

add_filter( 'sandbox_async_job', 'sandbox_jobs', 10, 2 );
```
* Run the watcher `php /home/wpcom/public_html/async-jobs/watcher.php`
* Check the new setting and save
* Refresh page and that setting should be saved

### DIFF
--- a/client/me/notification-settings/settings-form/constants.js
+++ b/client/me/notification-settings/settings-form/constants.js
@@ -1,3 +1,3 @@
 export const NOTIFICATIONS_EXCEPTIONS = {
-	email: [ 'achievement', 'scheduled_publicize', 'store_order', 'blogging_prompt' ],
+	email: [ 'achievement', 'scheduled_publicize', 'store_order' ],
 };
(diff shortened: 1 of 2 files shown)


creativecoder (commented):
It looks like this adds email checkboxes for Site achievements and Post Publicized, as well--I don't think that's intended?

| **Before** | **After** |
| - | - |
| image | image |

eoigal:
> It looks like this adds email checkboxes for Site achievements and Post Publicized, as well--I don't think that's intended?

Good catch! that's not intended. I'll fix that up.

PR #71822, opened 2023-01-09
Stats: Remove showNewModules and apply horizontal bars to other components

grzegorz-cp:
#### Proposed Changes

* applying horizontal bars to all pages needing it besides Traffic (mainly summary detail pages - visible when clicking on "View details/all")
* adding a feature flag to control the above
* adding a temporary header to the summary details page (behind a feature flag) to be updated in a following pull request
* extending Horizontal bar component to accommodate `Tags` that use an array for labels instead of a string (work in progress while the design is finalised)

This change focuses on removing `showNewModules` introduced with the horizontal bar component.
The `Insights` page and `summary detail` page are a work in progress, hidden behind a feature flag to be updated incrementally.

#### Testing Instructions

* go to the live branch
* without applying a feature flag, nothing should change visually anywhere (potential areas of focus - Traffic, Insights and summary details pages)
* apply the feature flag by adding `?flags=stats/horizontal-bars-everywhere` to the URL
* verify that Traffic, Insights and summary details pages are not broken but keep in mind that the last two places are under development

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #71638

Issue #71638, opened 2023-01-03
Stats: Apply `showNewModules` property to all instances of `StatsModule`


jsnmoon:
Apply the new horizontal bars to all instances of the `StatsModule` component:

- [x] On the traffic page
- ~~On the insights page. This page may require a new grid layout to accommodate the horizontal bars.~~ moved to #72038
- [x] On the individual stats modules pages (part of #71018).
    - [x] `/stats/day/posts/example.com`
    - [x] `/stats/day/referrers/example.com`
    - [x] `/stats/day/countryviews/example.com`
    - [x] `/stats/day/authors/example.com`
    - [x] `/stats/day/searchterms/example.com`
    - [x] `/stats/day/clicks/example.com`
    - ~~`/stats/day/videoplays/example.com`~~ moved to #71018
    - ~~`/stats/day/filedownloads/example.com`~~ moved to #71018

Identify and note any additional pages requiring migration in the above to-do list.

Once all of the `StatsModule` components use horizontal bars, we can remove the `showNewModule` property. It was used to prevent horizontal bars from showing on the `Insights` page.

cc @grzegorz-cp.

grzegorz-cp:
I opened a [draft PR](https://github.com/Automattic/wp-calypso/pull/71822) for removing `showNewModule` property. Besides this, I added a feature flag controlling the bars - without the FF, it should only be visible on the Traffics page, but with the FF, it is visible everywhere.

This highlights more work that needs to be done, but I didn't want to inflate the above PR so it wouldn't stay open for a week.

The summary details page will be updated in another PR. First, I need to replace the card's title and update the page header (below breadcrumbs, where you switch periods). Existing "tabs" have to look like buttons next to the chart.

I'm holding off on applying the grid layout on the Insights page while investigating why specific tags (each tag, not a category) don't have any view value. So this change will come in yet another PR.

cc @jsnmoon @jeffgolenski

grzegorz-cp:
I noticed that this touches on [#71018](https://github.com/Automattic/wp-calypso/issues/71018).
When I removed `showNewModule` in the PR mentioned above, I had to touch on the header for the summary details page. It shares styles with other headers, so I'll open another PR to add an updated header as part of this issue.
I can also add a grid, but I propose handling any tweaks to the header and summary pages as part of the Modernization project. The same would apply to adjusting modules placed on the grid on the Insights page.

grzegorz-cp:
[...]

Issue #71018, opened 2022-12-09
Stats: Detail Page(s) - Summary Pages


jeffgolenski:
# Development

- [x] Apply horizontal bars to the stats list.
- [x] Restyle the period type selector.
- [x] Ensure that all detail pages have been updated.
- [x] Verify that all additional items are present on the page, for example, CSV export option
- [ ] Align `Videos` to the other pages and apply horizontal bars

Additional tasks:
- #72131
- #72132

# Design

Each summary page we have needs a slight redesign. These pages will mostly utilize components we've already built elsewhere in stats, so we shouldn't need dev specs... but if we do, let me know!

These pages are accessed when a user clicks "View all" or "View details" at the bottom of cards on the Traffic and Insights pages.

_Note: There are many summary pages, so we may have to adjust designs if we find any special use-cases_

### Two before / after examples

**Country summary**

![Image](https://user-images.githubusercontent.com/214813/206773895-0a51a63a-2efe-46cb-817c-7dbceb4ce491.png)

**Posts & Pages summary**

![Image](https://user-images.githubusercontent.com/214813/206774216-2a412de1-b0f0-4fa1-b41a-9c8602f24f11.png)

grzegorz-cp:
As part of modernising the Insights page, we'd like to create a summary detail page for `Tags & categories`.

grzegorz-cp:
While working on #71638, I discovered three special cases for the summary page - [details](https://github.com/Automattic/wp-calypso/issues/71638#issuecomment-1381180070).

`Authors` and `File downloads` need investigation of the API to check if we can update it to apply the standard header available on other summary pages.

`Videos` need more guidance or a new design because the table exposes multiple values, not only views.
An alternative approach could drop all of the additional values. While this could be useful for the `watch time`, which is often `>1h`, it would be a shame to remove `impressions`.
If we keep all the values, we may improve `watch time` by providing an exact time.

It's also a shame that the users can't sort the table by `impressions`. This sorting should be easy to implement on the client side, but it would beat the purpose of the horizontal bars, which are affected by views. A toggle option could help here (similar to some of the components on the new Insights page designs), but this has to be discussed further.

jsnmoon:
> - Create a summary page for Tags & categories
> - Extend API for Authors and File downloads to accept different periods and restore the header visible on other summary pages (propose an alternative if the task is too time-consuming)

[...]

PR #71957, opened 2023-01-11
Stats: Updating page header for summary details

grzegorz-cp:
#### Proposed Changes

* Updating the header for a summary details page (behind a feature flag)
* Adding a mixin file combining CSS needed by the summary details page and:
    * Section heading (used above the chart on the Traffic page)
    * Navigation segments (used above the chart on the Traffic page)

This PR continues work started in #71822 

#### Testing Instructions

* Navigate to `Stats` and make sure that the header above the chart on the Traffic page didn't change (consider multiple breakpoints)
* Navigate to `View all`/`View deatils` for stats modules below the chart (summary detail page) and make sure that horizontal bars are NOT visible and the original header is visible
* apply the feature flag by adding `?flags=stats/horizontal-bars-everywhere` to the URL
* confirm that the summary detail pages now show horizontal bars and a new header, similar to the one above the chart on the main page (period buttons should not have an equal width here)
* confirm that the Insights page loads without problems (The tags component is a work in progress)

Before applying a feature flag
Screenshot 2023-01-12 at 11 36 17 AM

Header with a feature flag not including changes in this PR
Screenshot 2023-01-12 at 11 37 23 AM

Header with a feature flag and proposed changes
Screenshot 2023-01-12 at 11 37 38 AM

#### Pre-merge Checklist

Related to #71638


client/my-sites/stats/_modernized-mixins.scss
```diff
@@ -0,0 +1,87 @@
+@import "@wordpress/base-styles/breakpoints";
+@import "@automattic/typography/styles/variables";
+
+$font-recoleta: Recoleta, $sans;
+
+/**
+ * CSS used to transform existing Segment component to the styles used in the modernized Stats.
+ */
+@mixin navigation-segment-control-buttons( $segment-width: 100px ) {
```
grzegorz-cp, line 9: I combined the styles in a new file. Do you think this is helpful?

client/my-sites/stats/_modernized-mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+	padding: 0 24px 0 0;
+	margin: 0;
+
+	@media ( max-width: $break-mobile ) {
+		margin: 16px 0;
+		min-width: unset;
+	}
+}
+
+// TODO: would it be better to have a mixin file in Calypso components?
```
grzegorz-cp, line 76: There is a potential for improvement, including heading styles used by components we import (especially on the Insights page).

client/my-sites/stats/stats-module/index.jsx
```diff
@@ -173,12 +173,13 @@ class StatsModule extends Component {
 				) }
 				{ shouldShowNewModule && (
 					<>
+						{ /* TODO: Move this header to the summary page when modernising it */ }
```
grzegorz-cp, line 184: Moving this header to the summary page will be beneficial because it will unify all of the summary pages. The only thing left is to handle `path` for all cases, so it could be a part of modernising the Summary page.

grzegorz-cp:
I noticed that `Videos` uses a different table, and `File downloads` is missing the header. I'll investigate.

jsnmoon (approved):
Changes look reasonable, though the videos metric page lacks the horizontal bars and the downloads page lacks the header, as already noted. Great work getting us on the right path to unifying all metric summary pages.

I'd like to see what the rest of the team thinks about our mixin usage here. (I like how it makes the intent of our styles more explicit.)

grzegorz-cp:
I found the component for the video page [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/videopress-stats-module/index.jsx).
It duplicates a lot of the code used for the other modules, so there's a lot of potential to tidy it up, but it differs from all the other pages because it displays more values. Not only views but also impressions and hours watched. It doesn't have any period navigation.

grzegorz-cp:
The header is missing for `File downloads` and `Authors` because those modules are excluded from displaying the data by different periods. I'm not sure why. 
It makes it very difficult to check authors for different days. So at the moment, you've got to return to the traffic page, select a different day, go back to `authors` details, and so on if you'd like to check other days.

a8ck3n:
In my testing, it works as described in the no-feature-flag mode. I'm seeing the date header printed twice (matching your screenshot).

With the flag enabled, the header gets clipped on smaller screens. I'm seeing this:

Snap2

If we do the same thing as on the Traffic page, we probably want the period control to drop down below the header as the screen size gets smaller.

grzegorz-cp:
Thanks for finding this @a8ck3n. I made several updates to fix the display:
- fixed title for various screen sizes
- added padding for CSV at the bottom of the page
- added a header variation for `Authors` and `Downloads` by keeping the new title font style but hiding the period navigation and replacing it with a CSV download button
- excluded `Tags` and `Publicize` from using new horizontal bars (they will be updated when working on the bottom part of the insights page)

More changes should probably be done as part of #71018 because it requires shifting more HTML, but if anything looks broken, I can create an issue and fix it before enabling the feature flag for production.

a8ck3n:
This is looking pretty good. I'm still seeing some issues with padding & clipping though.

ONE. In between around 600 & 780 pixels, the padding between the title and the card doubles up from 24 to 48px.

DoubleGap

TWO. The period selector is still getting clipped at mobile sizes. By contrast, the old control changed to a dropdown menu at mobile sizes.

ClippedPeriod

kangzj (approved):
It works as described and LGTM 👍 

PS: I can see the two issues pointed out by @a8ck3n are both addressed.

PR #71991, opened 2023-01-12
Themes: Hide Business plan upsell if there are results

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1400

#### Proposed Changes

Hides the Business plan upsell from the themes search if we're displaying results.

This upsell was intended to show on searches for Astra as an experiment when we were not displaying WP.org themes in the results.

Now that WP.org themes are listed in the search, we shouldn’t show the upsell if there are matching themes.

Before | After
--- | ---
Screenshot 2023-01-12 at 13 55 36 | Screenshot 2023-01-12 at 13 57 15

#### Testing Instructions

- Go to Appearance > Themes
- Search for "Astra" or "Hotel"
- Make sure the upsell doesn't show up
- Search for something that doesn't produce any result
- Make sure the upsell shows up

### DIFF
--- a/client/components/themes-list/index.jsx
+++ b/client/components/themes-list/index.jsx
@@ -171,16 +171,18 @@ function Empty( props ) {
 					{ ...props }
 				/>
 			) : (
-				<div className="themes-list__not-found-text">
-					{ translate( 'No themes match your search' ) }
-				</div>
+				<>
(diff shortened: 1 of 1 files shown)


BogdanUngureanu (approved):
LGTM! 🚢

PR #72001, opened 2023-01-12
Jetpack cloud cart integration - fix license activation banner link

vikneshwar:
#### Proposed Changes
- Make the header sticky using IntersectionObserver instead of CSS
- This makes sure the link below (like the license activation banner is clickable)

#### Testing Instructions
**Prerequisites:** 
- You'll need a un-activated product license key.
- If you don't have one:
    - Go to https://jetpack.com/
    - Scroll down the page a bit until you see the 3 product cards (Backup, Security, Complete). Click on "**Get Backup**" and purchase it on the checkout page.
    - On the 'Thank you' page, it's asking you to select a site you want to apply the license to. Do Not select a site. Rather select the dropdown and choose, "**I don't see my site, let me configure it manually**".
    - You now should have an unattached product license key.

----
**Test**
- Do any one of these
    * Click on Jetpack Cloud live link below and wait for the redirect to complete, then goto `/pricing/:site` (`:site` is the site-slug of a jetpack site you own, or a Jurassic Ninja site you created **and connected**.) (This is probably the quickest and easiest way to test this PR).
    * or run this PR 
        * Run `git fetch && git checkout fix/cloud-sticky-header`
        * Run `yarn start-jetpack-cloud` (it takes a little while to build the application)
        * Goto http://jetpack.cloud.localhost:3000/pricing/:site  (where `:site` is the site-slug of a jetpack site you own, or a Jurassic Ninja site you created **and connected**.)
- You may be prompted with a dialog/modal notifying you that you have an available license key. Close (X) the dialog/modal. Then Notice/verify you see a black horizontal banner, notifying you the same, that you have an available license key. It will have an underlined link that says, "**Activate it now**".
- Hover over the **"Activate it now**" link and verify you cursor turns into a pointer. Verify you can click the link and it works! 
Markup 2023-01-11 at 23 43 14

- At various other viewport widths, all the way to mobile, verify that the 2 sticky banners are working correctly as before and verify that the **"Activate it now**" link is working properly and the cursor turns into a pointer when hovering the link

#### Pre-merge Checklist

Related to # 1203495437860512-as-1203698368270970/f

### DIFF
--- a/client/components/jetpack/intro-pricing-banner/index.tsx
+++ b/client/components/jetpack/intro-pricing-banner/index.tsx
@@ -1,9 +1,10 @@
 import { recordTracksEvent } from '@automattic/calypso-analytics';
 import { useBreakpoint } from '@automattic/viewport-react';
 import { useTranslate } from 'i18n-calypso';
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import CloudCart from 'calypso/jetpack-cloud/sections/pricing/jpcom-masterbar/cloud-cart';
+import { usePrevious } from 'calypso/landing/gutenboarding/hooks/use-previous';
 import useDetectWindowBoundary from 'calypso/lib/detect-window-boundary';
 import { preventWidows } from 'calypso/lib/formatting';
 import isJetpackCloud from 'calypso/lib/jetpack/is-jetpack-cloud';
@@ -16,7 +17,8 @@ import useBoundingClientRect from './hooks/use-bounding-client-rect';
 import people from './people.svg';
 
 const CALYPSO_MASTERBAR_HEIGHT = 47;
-const CLOUD_MASTERBAR_HEIGHT = 0;
+const CLOUD_MASTERBAR_HEIGHT = 47;
+const CONNECT_STORE_HEIGHT = 0;
 
 const IntroPricingBanner: React.FC = () => {
 	const translate = useTranslate();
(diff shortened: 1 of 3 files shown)


client/components/jetpack/intro-pricing-banner/index.tsx
```diff
@@ -25,14 +27,34 @@ const IntroPricingBanner: React.FC = () => {
...
+		if ( ! shouldShowCart ) {
+			return;
+		}
+		const navHeaderEle = document.getElementsByClassName( 'header__content-background-wrapper' );
+		if ( ! navHeaderEle || ! navHeaderEle.length ) {
+			return;
+		}
+
+		if ( hasCrossed ) {
+			navHeaderEle[ 0 ].classList.remove( 'header__content-background-wrapper--sticky' );
```
andrii-lysenko, line 52: This way we would have very touch coupling between elements, that actually used separately across the codebase. If we would show banner on some other page, or change the classname, this will break the behaviour. We need to think of a way that classnames would be passed or something.

vikneshwar: Hi Andrii, yeah, i agree with you on this. I tried to figure out a better, less coupled way to do it but couldn't find a quick solution. So, I decided to do this way to ship it quickly and also considering `<IntroPricingBanner/>` is rendered only on the pricing page, and it shouldn't impact other pages.

vikneshwar: And also, to me the `<IntroPricingBanner/>` should be refactored better to just render the banner and other logic like attaching the cart and making it sticky should be handled by the parent component rendering it, since the pricing page might need the sticky banner but not the other pages rendering it.

andrii-lysenko (approved):
This approach also looks good to me. As mentioned in the Slack thread we could go either this way, or still use sticky with `pointer-events` updated. @elliottprogrammer what do you think?

If all ok, and design team agrees this is good to go.

elliottprogrammer (approved):
Yes, I like it, It works well and delivers it's purpose well (keeping the cart visible/accessible to the user at all times, while scrolling the page. 🙂 
LGTM! Lets's ship it! 🚀

Perhaps it is not as extensible/reusable, or code perfect as we would like it (using classnames, static elem heights, etc,- but for our purposes now, it works very well for the time being, and we can choose to refactor another time to make it super extensible and reusable should we decide in the future. But from a users perspective, the way it is now, it works very good and fine! 😉

PR #72040, opened 2023-01-13
Stats: Avoid redirecting back to the views tab

### LINKED ISSUE #71788: Stats: Avoid redirecting back to the "Views" tab after clicking the navigation arrows on other tabs
Navigating on the `Visitors` tab should stay at the same tab when clicking on the navigation arrows.

https://user-images.githubusercontent.com/6869813/211269937-dfcf0b38-8c57-4026-b8ca-9a104dcef21e.mov

annacmc:
#### Proposed Changes

This PR fixes #71788

This adds the `tab` query string onto the URL to persist the current tab instead of redirecting it back to views. This implementation fixes the issue. However, I feel like there is a better/cleaner way to implement this. So, I'm looking for some feedback/suggestions on a better approach

#### Testing Instructions

* Open the Calypso live branch
* Navigate to the Stats page
* Switch to one of the tabs aside from `views` (eg. to `visitors`, `likes`, or `comments`)
* Use the left and right arrows to navigate between different dates
* Confirm that while switching through date ranges, that the active tab remains the same

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/my-sites/stats/stats-period-navigation/index.jsx
+++ b/client/my-sites/stats/stats-period-navigation/index.jsx
@@ -52,8 +52,7 @@ class StatsPeriodNavigation extends PureComponent {
 
 	handleArrowNext = () => {
 		const { date, moment, period, url, queryParams } = this.props;
-		const getQueryParams = new URLSearchParams( window.location.search );
-		const tabValue = getQueryParams.get( 'tab' );
+		const tabValue = queryParams.tab;
 		const nextDay = moment( date ).add( 1, period ).format( 'YYYY-MM-DD' );
 		const nextDayQuery = qs.stringify(
 			Object.assign(
--- a/client/my-sites/stats/stats-period-navigation/index.jsx
+++ b/client/my-sites/stats/stats-period-navigation/index.jsx
@@ -71,9 +71,8 @@ class StatsPeriodNavigation extends PureComponent {
 
 	handleArrowPrevious = () => {
 		const { date, moment, period, url, queryParams } = this.props;
+		const tabValue = queryParams.tab;
 		const previousDay = moment( date ).subtract( 1, period ).format( 'YYYY-MM-DD' );
-		const getQueryParams = new URLSearchParams( window.location.search );
-		const tabValue = getQueryParams.get( 'tab' );
 		const previousDayQuery = qs.stringify(
 			Object.assign(
 				{},
--- a/client/my-sites/stats/stats-period-navigation/index.jsx
+++ b/client/my-sites/stats/stats-period-navigation/index.jsx
@@ -52,8 +52,9 @@ class StatsPeriodNavigation extends PureComponent {
 
 	handleArrowNext = () => {
 		const { date, moment, period, url, queryParams } = this.props;
+		const usedPeriod = 'hour' === period ? 'day' : period;
 		const tabValue = queryParams.tab;
-		const nextDay = moment( date ).add( 1, period ).format( 'YYYY-MM-DD' );
+		const nextDay = moment( date ).add( 1, usedPeriod ).format( 'YYYY-MM-DD' );
 		const nextDayQuery = qs.stringify(
 			Object.assign( {}, queryParams, { startDate: nextDay }, tabValue && { tab: tabValue } ),
 			{
@@ -66,8 +67,9 @@ class StatsPeriodNavigation extends PureComponent {
 
 	handleArrowPrevious = () => {
 		const { date, moment, period, url, queryParams } = this.props;
+		const usedPeriod = 'hour' === period ? 'day' : period;
 		const tabValue = queryParams.tab;
-		const previousDay = moment( date ).subtract( 1, period ).format( 'YYYY-MM-DD' );
+		const previousDay = moment( date ).subtract( 1, usedPeriod ).format( 'YYYY-MM-DD' );
 		const previousDayQuery = qs.stringify(
 			Object.assign( {}, queryParams, { startDate: previousDay }, tabValue && { tab: tabValue } ),
 			{ addQueryPrefix: true }
--- a/client/my-sites/stats/stats-period-navigation/index.jsx
+++ b/client/my-sites/stats/stats-period-navigation/index.jsx
@@ -53,14 +53,10 @@ class StatsPeriodNavigation extends PureComponent {
 	handleArrowNext = () => {
 		const { date, moment, period, url, queryParams } = this.props;
 		const usedPeriod = 'hour' === period ? 'day' : period;
-		const tabValue = queryParams.tab;
 		const nextDay = moment( date ).add( 1, usedPeriod ).format( 'YYYY-MM-DD' );
-		const nextDayQuery = qs.stringify(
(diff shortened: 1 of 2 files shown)


annacmc:
Closed this up as it's been resolved in a different PR. https://github.com/Automattic/wp-calypso/pull/72049

dognose24:
#72049 is related to the period changing instead of the navigation arrows. :slightly_smiling_face: cc @annacmc

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -52,19 +52,36 @@ class StatsPeriodNavigation extends PureComponent {
...
+			}
+		);
 		const href = `${ url }${ nextDayQuery }`;
 		this.handleArrowEvent( 'next', href );
 	};
 
 	handleArrowPrevious = () => {
 		const { date, moment, period, url, queryParams } = this.props;
 		const previousDay = moment( date ).subtract( 1, period ).format( 'YYYY-MM-DD' );
+		const getQueryParams = new URLSearchParams( window.location.search );
```
kangzj: I would recommend to assign `content.query`  injected by a `page.js` middleware to `queryParams` and use it here, otherwise it probably wouldn't work for Odyssey, because Odyssey uses hashes as for queries: pejTkB-2s#2-2-routing

annacmc: Great, thanks so much Jasper this is really helpful advice!

annacmc:
> #72049 is related to the period changing instead of the navigation arrows. 🙂 cc @annacmc

Thanks @dognose24 I looked at this right before I AFK'd sick and missed that it wasn't quite the same issue! 🙈  Thank you for reopening 🙌

client/my-sites/stats/stats-period-navigation/index.jsx
kangzj: Related code:

**Odyssey**
https://github.com/Automattic/wp-calypso/blob/b37ccf43981ced060e523616039b58834ea66188/apps/odyssey-stats/src/page-middleware/setup-context.jsx#L13

**Calypso**
https://github.com/Automattic/wp-calypso/blob/b37ccf43981ced060e523616039b58834ea66188/client/boot/common.js#L65

Pass it to top level components and then pass down to child component I suppose:
https://github.com/Automattic/wp-calypso/blob/b37ccf43981ced060e523616039b58834ea66188/client/my-sites/stats/controller.jsx#L368

kangzj (approved):
It works as described and I tested Odyssey, which works well too 👍 

A note here, not a blocker:

`StatsPeriodNavigation` is used in some other places too, so we would possibly want to look if they need changes too.

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -52,21 +52,34 @@ class StatsPeriodNavigation extends PureComponent {
 
 	handleArrowNext = () => {
 		const { date, moment, period, url, queryParams } = this.props;
-
-		const usedPeriod = 'hour' === period ? 'day' : period;
```
dognose24, line 56: I think this checked period for the `hour` is [intended](https://github.com/Automattic/wp-calypso/commit/c54598ec16df83717c75a55c52c5a8895d28a1a4#diff-fc24058576fcc9d1fe4a8859e1d0a7ea6cb6c027624e0f1bc5e209b96ddb13a1R56) for the [Email Stats page](https://github.com/Automattic/wp-calypso/pull/72165).

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -52,21 +52,34 @@ class StatsPeriodNavigation extends PureComponent {
...
-			addQueryPrefix: true,
-		} );
+		const tabValue = queryParams.tab;
+		const nextDay = moment( date ).add( 1, period ).format( 'YYYY-MM-DD' );
+		const nextDayQuery = qs.stringify(
+			Object.assign(
+				{},
+				queryParams,
+				{ startDate: nextDay },
+				{ tab: tabValue ? tabValue : 'views' }
```
dognose24:
```suggestion

```
Since there could be any page using this `StatsPeriodNavigation` component, indiscriminately appending `tab=views` to another page's URL looks unreasonable. Could we directly determine the query string `tab` via `queryParams`?

dognose24 (dismissed):
As @kangzj mentioned, the `StatsPeriodNavigation` is used elsewhere, especially on the Email Stats page. We have to avoid breaking it about the period checking.

client/my-sites/stats/stats-period-navigation/index.jsx
annacmc: Thanks @dognose24 - I've updated it so that the tab is only appended to the URL conditionally if we have a queryParam tab to append. Is this kind of in line with what you were thinking?

client/my-sites/stats/stats-period-navigation/index.jsx
annacmc: Yes, I believe so too - AFAIK this isn't breaking anything anywhere, and my implementation isn't likely to break the Email Stats page. However, I'm having trouble accessing the email stats page to test. Did you see any indication that this was breaking anything?

dognose24: Yes. I have tried this PR with the Email Stats page. And if I press the button to navigate next on the period `hour`, the operation is stuck. We can keep the original code since it would not break our pages.

##### Before
https://user-images.githubusercontent.com/6869813/213317227-17a28696-6a86-46eb-9ce8-9e051fc5773c.mov

##### This PR
https://user-images.githubusercontent.com/6869813/213317291-741a85fc-0da6-473a-8539-bcaafbb1d113.mov

client/my-sites/stats/stats-period-navigation/index.jsx
dognose24: Hey @annacmc, you could just also keep the original code rather than append the `tabValue` checking for the tab. It worked as what you changed in [22d9722](https://github.com/Automattic/wp-calypso/pull/72040/commits/22d9722287da2a091749993b8328e21b53949348) since the `tab` is already in `queryParams`.

client/my-sites/stats/stats-period-navigation/index.jsx
kangzj:
```
const usedPeriod = 'hour' === period ? 'day' : period;
```
Any particular reason to remove the line? @annacmc

annacmc: Ah, I see - sorry I misunderstood! Yes, you're right - it is needed for that page :)

annacmc: > Any particular reason to remove the line? @annacmc

Nope, sorry. Definitely a mistake!

client/my-sites/stats/stats-period-navigation/index.jsx
annacmc: You are very right! Thank you

kangzj (approved):
LGTM ❤️

Issue #69899, opened 2022-11-09
Free Onboarding Flow Tracking Issue


jeyip:
### Details

This is a tracking issues for task related to the new Free Flow, which will be based on stepper and use architecture similar to our existing tailored flow cases like newsletters and link in bio websites. 

### On Deck ###

- [ ] https://github.com/Automattic/wp-calypso/issues/70845
- [x] https://github.com/Automattic/wp-calypso/issues/72268
- [x] https://github.com/Automattic/wp-calypso/issues/72269

### Done
- [x] https://github.com/Automattic/wp-calypso/pull/70350
- [x] https://github.com/Automattic/wp-calypso/issues/70843
- [x] Ensure My Home redirects to launchpad.
- [x] https://github.com/Automattic/wp-calypso/issues/70556
- [x] https://github.com/Automattic/wp-calypso/issues/70558
- [x] https://github.com/Automattic/wp-calypso/issues/69921 
- [x] https://github.com/Automattic/wp-calypso/issues/71280
- [x] https://github.com/Automattic/wp-calypso/issues/70559
- [x] https://github.com/Automattic/wp-calypso/issues/70844
- [x] https://github.com/Automattic/wp-calypso/issues/69920 
- [x] https://github.com/Automattic/wp-calypso/issues/71353
- [x] https://github.com/Automattic/wp-calypso/issues/71354

Abandoned
- [ ] https://github.com/Automattic/wp-calypso/issues/70842

### Relevant Links
- Project Thread: paYE8P-2bD-p2
- Marketing Plan: p5uIfZ-cw9-p2
- P2: p9Jlb4-5ST-p2

agrullon95:
@Addison-Stavlo / @ebinnion - Is the [optional domain name generator button](https://github.com/Automattic/wp-calypso/issues/69922) supposed to be displayed on the existing sign up flow?

jeyip:
> Is the https://github.com/Automattic/wp-calypso/issues/69922 supposed to be displayed on the existing sign up flow?

Designers are in the process of creating mocks. Eric says "I would focus efforts right now on the backend, such as getting the library in place, endpoint, etc." ( [link](p1668451042654359/1668450288.931879-slack-CHN6J22MP) )

edanzer:
Following a lot of discussion, we have decided to try creating a new Free Flow based on Stepper. This is inline with all our other recent work on tailored flows as well as efforts to move generic onboarding screens like Domains and Plans to stepper.

[...]

PR #68766, opened 2022-10-07
Redirect Calypso Home to WooCommerce Home in Ecommerce plans

xristos3490:
#### Proposed Changes

This PR aims to redirect the default home URL `wordpress.com/home/[site]` to `[site].wpcomstaging.com/wp-admin/admin.php?page=wc-admin` only on Ecommerce plans. This should be considered a temporary redirection until we create **a dedicated Home for Ecommerce** in Calypso.

As you might see, the redirection happens with a bit of delay, thus causing a flash-of-content. Could we implement this in a way that it's instant? 🤔 

#### Testing Instructions

* Apply this patch
* Select and visit an Ecommerce site from "My Sites"
* Notice the redirection to the WooCommerce Home page.

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Related to #68765

PR #72044, opened 2023-01-13
Onboarding bug crash patch - Error protection when getting the slug from a missing product object

miksansegundo:
#### Proposed Changes
To fix a production bug that breaks the onboarding in the `/start/domain` step when creating a new site from `Dashboard > My sites > Switch site > Add new site`

* Check if the product object exists before getting the slug in `isFreePlanProduct`

#### Testing Instructions

* From WordPress admin, go to My Sites
* Click on `Add new site` at the bottom of the list
Screenshot 2566-01-13 at 12 17 22
* Confirm that you land on the domains step and it loads as expected. It should NOT be a blank page.

| Before | After |
|------|-----|
|Screenshot 2566-01-13 at 12 22 03|Screenshot 2566-01-13 at 12 19 52|

#### Pre-merge Checklist

- [X] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [X] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [X] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related https://github.com/Automattic/wp-calypso/pull/68766
Maybe related to https://github.com/Automattic/wp-calypso/issues/69899

### DIFF
--- a/packages/calypso-products/src/is-free-plan.ts
+++ b/packages/calypso-products/src/is-free-plan.ts
@@ -3,5 +3,5 @@ import { PLAN_FREE } from './constants';
 import type { WithSnakeCaseSlug, WithCamelCaseSlug } from './types';
 
 export function isFreePlanProduct( product: WithSnakeCaseSlug | WithCamelCaseSlug ): boolean {
-	return camelOrSnakeSlug( product ) === PLAN_FREE;
+	return product && camelOrSnakeSlug( product ) === PLAN_FREE;
 }


bangank36:
That's weird, I can not reproduce it on my side!

arthur791004 (approved):

bangank36 (approved):
LGTM

arthur791004:
> That's weird, I can not reproduce it on my side!

This issue happens only when the site plan is not accessible. So, sometimes it might not be able to reproduce

packages/calypso-products/src/is-free-plan.ts
```diff
@@ -3,5 +3,5 @@ import { PLAN_FREE } from './constants';
 import type { WithSnakeCaseSlug, WithCamelCaseSlug } from './types';
 
 export function isFreePlanProduct( product: WithSnakeCaseSlug | WithCamelCaseSlug ): boolean {
-	return camelOrSnakeSlug( product ) === PLAN_FREE;
+	return product && camelOrSnakeSlug( product ) === PLAN_FREE;
```
sirbrillig, line 6: Since this changes the behavior of the function, I think it might be a good idea to also change the types to match. Specifically this function could be changed to support an undefined or null product.

```ts
export function isFreePlanProduct( product: WithSnakeCaseSlug | WithCamelCaseSlug | undefined | null ): boolean {
```

You could argue that the type as-is is better because it's what we _want_ to be processing in this function, but clearly because of the bug that was reported, something was passing a falsy value in as an argument when it shouldn't have been. That means that either the calling function was not in TypeScript or it has incorrect types itself. 

A better solution to this bug therefore might be to find the calling function and prevent calling `isFreePlanProduct` at all with a falsy value, but since that didn't happen (perhaps finding the caller was too difficult?) then I suggest we change the types to cover what this function actually does.

PR #72051, opened 2023-01-13
Social: Landing page pricing tweaks. 

spsiddarthan:
This PR enables you to select multiple plans of Jetpack Social.  The variant selection for checking out the advanced plan isn't working currently on calypso blue correctly. It doesn't show the monthly variant when you go to the checkout page
#### Proposed Changes

* The changes would have been a lot simpler if we decided to do the same thing that backup does, but since we decided to not show the radio buttons, the product-lightbox changes needed to be a little bit more social specific, as opposed to being generic. The changes are mainly around the multiplechoice question component and adding some additional props to the PaymentPlan component.
* I have introduced the `jetpack-social/advanced-plan` feature flag, so the changes are picked up only on local dev environment.

How Social looks:

Screenshot 2023-01-13 at 18 49 09

How Backup looks:

Screenshot 2023-01-13 at 18 50 03

#### Testing Instructions

* Checkout to this branch and do `yarn start-jetpack-cloud` and go to http://jetpack.cloud.localhost:3000/pricing?view=products
*  Clicking on Social's GET button and more about Social button should open the modal.
* After opening the modal, select the plan of your choice and finish checkout.

To test plan's variant selections b/w the monthly and yearly plans for the advanced plan
*  Do `yarn start` to boot calypso
* Go to http://calypso.localhost:3000/checkout/jetpack/jetpack_social_advanced_yearly to make sure the monthly variant is shown as well. 
Screenshot 2023-01-13 at 18 56 16

#### Pre-merge Checklist


spsiddarthan:
Some unit tests are failing, I need to figure out what they're but will appreciate a review, so I can address them as well as I figure out the unit tests.

client/my-sites/plans/jetpack-plans/product-lightbox/constants.ts
```diff
@@ -19,11 +21,19 @@ export const PRODUCT_OPTIONS: Record< string, TranslateResult > = {
 	[ PLAN_JETPACK_SECURITY_T2_YEARLY ]: translate( '1TB (1000GB)', {
 		comment: 'GB here is the abbreviation for Gigabyte, and TB is abbreviation for Terabyte',
 	} ),
+	[ PRODUCT_JETPACK_SOCIAL_BASIC ]: translate( 'Social Basic', {
+		comment: 'Get upto 1000 shares',
```
danielpost: "upto" -> "up to".

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -71,6 +78,24 @@ const ProductLightbox: React.FC< Props > = ( {
...
+			// Tracking when variant selected inside the lightbox
+			dispatch(
+				recordTracksEvent( 'calypso_product_lightbox_variant_select', {
+					site_id: siteId,
+					product_slug: productSlug,
+				} )
+			);
+		},
+		[ onChangeProduct, dispatch, siteId ]
+	);
```
danielpost: Is it necessary to add this function separately for Social products? It seems identical to the `onChangeOption` function that's declared right above it.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -108,7 +133,14 @@ const ProductLightbox: React.FC< Props > = ( {
...
 
-	const shouldShowOptions = variantOptions.length > 1;
+	const isSocialProduct = [
+		PRODUCT_JETPACK_SOCIAL_ADVANCED,
+		PRODUCT_JETPACK_SOCIAL_BASIC,
+	].includes( product.productSlug );
+
+	const shouldShowOptions = isSocialProduct
+		? variantOptions.length > 1 && config.isEnabled( 'jetpack-social/advanced-plan' )
+		: variantOptions.length > 1;
```
danielpost, line 125: I'm guessing this is temporary because the Advanced plan isn't available yet, right? If so, maybe we can add a comment here and add an Asana task to remove it once the plan is launched.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -176,7 +208,7 @@ const ProductLightbox: React.FC< Props > = ( {
 				<div className="product-lightbox__sidebar">
 					<div className="product-lightbox__variants">
 						<div className="product-lightbox__variants-content">
-							{ shouldShowOptions && (
+							{ shouldShowOptions && ! isSocialProduct && (
```
danielpost: Doesn't this extra `! isSocialProduct` check render the changes to `shouldShowOptions` made earlier obsolete?

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -187,11 +219,52 @@ const ProductLightbox: React.FC< Props > = ( {
 									/>
 								</div>
 							) }
-							<PaymentPlan
-								isMultiSiteIncompatible={ isMultiSiteIncompatible }
-								siteId={ siteId }
-								product={ product }
-							/>
+							{ isSocialProduct ? (
```
danielpost: I'm wondering if we want to turn this into something more generic, since we might add more products with multiple plan options in the future.

What do you think about adding some kind of check that checks if the product has multiple plan options, rather than checking if it's a Social product?

client/my-sites/plans/jetpack-plans/product-lightbox/payment-plan.tsx
```diff
@@ -13,11 +13,15 @@ type PaymentPlanProps = {
...
 	product: SelectorProduct;
+	hideTitle?: boolean;
+	showBorder?: boolean;
 };
 const PaymentPlan: React.FC< PaymentPlanProps > = ( {
 	isMultiSiteIncompatible,
 	siteId,
 	product,
+	hideTitle = false,
+	showBorder = false,
```
danielpost: Maybe we can describe the state here rather than the UI—something like `isActive`?

client/my-sites/plans/jetpack-plans/product-lightbox/payment-plan.tsx
```diff
@@ -61,9 +65,15 @@ const PaymentPlan: React.FC< PaymentPlanProps > = ( {
...
+					{ ! hideTitle && <p>{ translate( 'Payment plan:' ) }</p> }
 
-					<div className="product-lightbox__variants-plan-card">
+					<div
+						className={
+							showBorder
+								? 'product-lightbox__variants-plan-card'
+								: 'product-lightbox__variants-plan-card noborder'
+						}
+					>
```
danielpost: You could use the `classNames` helper here instead (untested):
```js
<div className={ classNames(
  'product-lightbox__variants-plan-card',
  ! showBorder && 'product-lightbox__variants-plan-card noborder'
) }>
```

I think here it's also better to describe the state rather than just the UI change, so I'd recommend something like this:
```js
<div className={ classNames(
  'product-lightbox__variants-plan-card',
  isActive && 'product-lightbox__variants-plan-card--active'
) }>
```

(and then change the class in the CSS file as well of course)

client/my-sites/plans/jetpack-plans/product-store/items-list/all-items.tsx
```diff
@@ -95,14 +96,24 @@ export const AllItems: React.FC< AllItemsProps > = ( {
 						<li key={ item.productSlug }>
 							<SimpleItemCard
 								ctaAsPrimary={ ctaAsPrimary }
-								ctaHref={ getCheckoutURL( item ) }
+								ctaHref={
+									item.productSlug === PRODUCT_JETPACK_SOCIAL_BASIC &&
+									config.isEnabled( 'jetpack-social/advanced-plan' )
+										? '#'
+										: getCheckoutURL( item )
+								}
```
danielpost: I think it would be good to move this logic to a separate `const` declaration above the return statement, with a comment explaining what it's doing.

client/my-sites/plans/jetpack-plans/product-store/items-list/all-items.tsx
```diff
@@ -95,14 +96,24 @@ export const AllItems: React.FC< AllItemsProps > = ( {
...
 								icon={ <img alt="" src={ getProductIcon( { productSlug: item.productSlug } ) } /> }
 								isCtaDisabled={ isCtaDisabled }
 								isCtaExternal={ isExternal }
-								onClickCta={ getOnClickPurchase( item ) }
+								onClickCta={
+									item.productSlug === PRODUCT_JETPACK_SOCIAL_BASIC &&
+									config.isEnabled( 'jetpack-social/advanced-plan' )
+										? onClickMoreInfoFactory( item )
+										: getOnClickPurchase( item )
+								}
```
danielpost: Same here as last comment :)

packages/calypso-products/src/constants/jetpack.ts
```diff
@@ -102,12 +104,18 @@ export const JETPACK_SCAN_PRODUCTS = <const>[
...
 	PRODUCT_JETPACK_SOCIAL_BASIC,
 	PRODUCT_JETPACK_SOCIAL_BASIC_MONTHLY,
 ];
 
+// Social Advanced
+export const JETPACK_SOCIAL_ADVANCED_PLANS = <const>[
+	PRODUCT_JETPACK_SOCIAL_ADVANCED,
+	PRODUCT_JETPACK_SOCIAL_ADVANCED_MONTHLY,
+];
+
```
danielpost, line 118: `JETPACK_SOCIAL_PRODUCTS` contains the Basic products and `JETPACK_SOCIAL_ADVANCED_PLANS` contains the Advanced products. It would probably be good to make these names consistent.

packages/calypso-products/src/constants/jetpack.ts
```diff
@@ -144,6 +152,7 @@ export const JETPACK_PRODUCTS_LIST = <const>[
 	...JETPACK_VIDEOPRESS_PRODUCTS,
 	...JETPACK_SOCIAL_PRODUCTS,
 	...JETPACK_BACKUP_ADDON_PRODUCTS,
+	...JETPACK_SOCIAL_ADVANCED_PLANS,
```
danielpost: Very nitpicky, but let's put the lines with the Social products right below each other.

danielpost (commented):
Nice work! I've added some initial thoughts from reading the code, will give it a functional test next.

danielpost:
Pressing GET on the pricing page takes me directly to the checkout page (using the live link for testing), rather than opening the modal. Is that a bug?

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
spsiddarthan: You're right, we don't need this function

spsiddarthan:
> Pressing GET on the pricing page takes me directly to the checkout page (using the live link for testing), rather than opening the modal. Is that a bug?

It's because of the env variable `jetpack-social/advanced-plan` which is enabled only for the development mode, meaning it will work only local. I will find out how to enable it for the live url sites, it should be a matter of changing one of the config files.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
spsiddarthan: Thanks, good point, the whole change is around how we currently display products with multiple plans. 

The existing code does support displaying multiple products, backup works like that, for instance. 

But the way we display the multiple products for Social and backup is different, for the latter, we use a radio button. For Social, we are displaying them, one below the other. So, maybe we have two types of modals - one for the products where we should have the radio button and the other for products like social. I will make this have a more generic name.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
spsiddarthan: I changed the variable name here and it looks like this now, let me know if that makes sense. 

```
{ shouldShowOptions && ! showVariantsWithoutRadioButtons && (
								<div className="product-lightbox__variants-options">
									<MultipleChoiceQuestion
										question={ PRODUCT_OPTIONS_HEADER[ product?.productSlug ] }
										answers={ variantOptions }
										selectedAnswerId={ product?.productSlug }
										onAnswerChange={ onChangeOption }
										shouldShuffleAnswers={ false }
									/>
								</div>
							) }
							{ shouldShowOptions && showVariantsWithoutRadioButtons ? ( .....
```

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
danielpost: The typecheck `: string` got removed, is that intentional?

packages/calypso-products/src/constants/jetpack.ts
danielpost: Should we also change `JETPACK_SOCIAL_PRODUCTS` to `JETPACK_SOCIAL_BASIC_PRODUCTS`?

danielpost (commented):
Nice, just some tiny things!

I noticed this part is empty: 
Screenshot 2023-01-17 at 14 53 10@2x

Is that something we need to add in this PR as well?

packages/calypso-products/src/constants/jetpack.ts
spsiddarthan: Sure thing, it is being used in many files though, and the last time we tried tweaking that, we broke the pricing page. Due to that reason, I wonder if we can do that renaming in a different PR as a follow-up and test that separately, to reduce risk.

spsiddarthan:
@danielpost, I have fixed the empty "Great for" and changed the name to "Advanced Plan Beta", per Kristina's recommendation. We might tweak the name later, but I think we can always do that - for now, the display value is "Advanced Plan Beta".

danielpost (approved):
Looks good to me, nice work! Maybe @pablinos or @gmjuhasz can take a look as well just to be safe.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -57,7 +64,7 @@ const ProductLightbox: React.FC< Props > = ( {
 	const translate = useTranslate();
 
 	const onChangeOption = useCallback(
-		( productSlug: string ) => {
+		( productSlug ) => {
```
pablinos: I think @danielpost asked this, but is `productSlug` no longer a string? If so, can we update the type rather than removing it?

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -108,7 +118,15 @@ const ProductLightbox: React.FC< Props > = ( {
 		} ) );
 	}, [ product.productSlug ] );
 
-	const shouldShowOptions = variantOptions.length > 1;
+	const showVariantsWithoutRadioButtons = [
```
pablinos: This is a bit of a nitpick (and I'm struggling to think of a better option) but this name and the `showPlansOneBelowTheOther` seem a bit too tied to the design.

Maybe it's something for a follow-up but all the conditional checks might mean that we could break this into multiple components and have two styles for the product variants.

pablinos (commented):
I've only given this a quick look so far. I wonder if it's worth getting a review from Origin if we can?

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -187,11 +205,48 @@ const ProductLightbox: React.FC< Props > = ( {
...
+											isMultiSiteIncompatible={ isMultiSiteIncompatible }
+											siteId={ siteId }
+											product={ JETPACK_SOCIAL_ADVANCED_PRODUCT as SelectorProduct }
+											showPlansOneBelowTheOther={ true }
+											isActive={
+												product?.productSlug === JETPACK_SOCIAL_ADVANCED_PRODUCT?.productSlug
+											}
+										/>
+									</div>
+								</div>
```
manzoorwanijk: Can we extract this to a separate component and then conditionally render it here?

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
manzoorwanijk: I agree with the suggestion.

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
```diff
@@ -187,11 +205,48 @@ const ProductLightbox: React.FC< Props > = ( {
...
+										<PaymentPlan
+											isMultiSiteIncompatible={ isMultiSiteIncompatible }
+											siteId={ siteId }
+											product={ JETPACK_SOCIAL_BASIC_PRODUCT as SelectorProduct }
+											showPlansOneBelowTheOther={ true }
+											isActive={
+												product?.productSlug === JETPACK_SOCIAL_BASIC_PRODUCT?.productSlug
+											}
+										/>
+									</div>
```
manzoorwanijk: For a11y and better keyboard navigation support, these should be real radio buttons that can be styled the way we want.

client/my-sites/plans/jetpack-plans/product-store/items-list/all-items.tsx
```diff
@@ -91,18 +92,30 @@ export const AllItems: React.FC< AllItemsProps > = ( {
 						isSuperseded
 					);
 
+					// Go to the checkout page for all products when they click on the 'GET' CTA, except for Jetpack Social where we open a modal.
+					const ctaHref =
+						item.productSlug === PRODUCT_JETPACK_SOCIAL_BASIC &&
+						config.isEnabled( 'jetpack-social/advanced-plan' )
+							? '#'
```
manzoorwanijk: To make sure that the URL gets updated correctly and the page doesn't scroll to the top.

```suggestion
							? `#${ item.productSlug }`
```

client/my-sites/plans/jetpack-plans/product-store/items-list/all-items.tsx
```diff
@@ -91,18 +92,30 @@ export const AllItems: React.FC< AllItemsProps > = ( {
...
+							: getCheckoutURL( item );
+					const onClickCtaUrl =
```
manzoorwanijk: This is not a URL (string), it's rather a callback (function), thus `onClickCta` sounds a better name than `onClickCtaUrl`

manzoorwanijk (commented):
I see the a11y and keyboard navigation issue in payment plans for Social - https://github.com/Automattic/wp-calypso/pull/72051#discussion_r1073109650

client/my-sites/plans/jetpack-plans/product-lightbox/index.tsx
spsiddarthan: I have made it a component, but the variable name I am not convinced, but I can definitely think of a better name and update it in the future @pablinos.

manzoorwanijk (approved):
This looks good to me now.

danielpost (approved):

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7754399

Thank you @spsiddarthan for including a screenshot in the description! This is really helpful for our translators.

PR #70428, opened 2022-11-25
Pricing page rework | Remove Feature Flag and Clean up unused code

manzoorwanijk:
Since we have reached a point of stability for the Pricing Page Rework v1 project. It's the time to revisit and clean up the code and remove the feature flag and its usage.

#### Proposed Changes

* Remove feature flag usage from the code
* Remove feature flag from config JSON files
* Remove unused code from old pricing page

#### Testing Instructions

- Do any one of these
    * Click on Jetpack Cloud live link below and wait for the redirect to complete, then goto `/pricing`
    * or boot up this PR 
        * Run `git fetch && git checkout pricing-page-rework/clean-up`
        * Run `yarn start-jetpack-cloud`
        * Goto [http://jetpack.cloud.localhost:3000/pricing](http://jetpack.cloud.localhost:3000/pricing)
- Confirm that everything is the same as before
- Confirm the same for Calypso Blue - `/plans/:site`

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Completes 1202858161751496-as-1202835600809438/f

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202835600809438

PR #72074, opened 2023-01-13
Jetpack Plans: Remove old unused components

tyxla:
#### Proposed Changes

This PR performs some follow-up cleanup after #70428, where old Jetpack plan components that are no longer in use are suggested to be removed.

#### Testing Instructions

* Verify that the removed components are no longer in use.
* Verify Jetpack Cloud and Calypso blue plans/pricing pages still work well as per the test instructions of #70428.


manzoorwanijk (approved):
Works well

flootr (approved):
🚀

PR #72075, opened 2023-01-13
BBE: Save website content to server

aneeshd16:
#### Proposed Changes

UI Changes:
* Adds a new button on the website content form to save website content to the server. Clicking on the button will make a network request that saves all form data to the server.
image

* The "Save Changes" button is enabled only if there are unsaved changes.
* The "Next" button also saves content first and then opens the next section.
* Clicking on any of the section titles opens the relevant section and saves content in the background.
* For the above two actions, the network request to save changes is only sent if there are unsaved changes.
* Refreshing the page or opening the page in a different browser/tab/device should pre-populate the form with the saved data.

Code changes:
* While the quantity of changes seems to be large, they are mostly due to refactors, considering that the source of truth for this data is the server and not browser-based storage. I have left inline comments to explain major changes.

#### Testing Instructions

* Go to `/start/do-it-for-me`, proceed through the flow steps and complete the purchase.
* Once you reach the website content form, type in any text content and click on the "Save Changes" button.
* Confirm that the button has a loading state and form interaction is blocked during this time.
* Reload the page and confirm that the typed data is prepopulated.
* Repeat the above three steps with the site logo section and the feedback section. 
* Open the form link in an incognito window or in a different browser and confirm that the form is prepopulated with saved data.

#### Pre-merge Checklist

Fixes Automattic/martech#1405


client/state/signup/steps/website-content/reducer.ts
```diff
@@ -13,234 +13,224 @@ import {
...
-import {
-	schema,
-	initialState,
-	WebsiteContentCollection,
-	MEDIA_UPLOAD_STATES,
-	LOGO_SECTION_ID,
-} from './schema';
+import { initialState, LOGO_SECTION_ID, MEDIA_UPLOAD_STATES } from './constants';
+import { WebsiteContentCollectionState } from './types';
 import type { AnyAction } from 'redux';
```
aneeshd16, line 20: Github shows this is as a large diff, but it is majorly an indentation change due to the removal of `withSchemaValidation`.

The code for only `SIGNUP_STEPS_WEBSITE_CONTENT_INITIALIZE_PAGES` action type has been changed. This the redux state is no longer the source of truth, I have removed the `siteId` checks that caused issues with persisted state data before.

client/state/signup/steps/website-content/selectors.ts
```diff
@@ -1,10 +1,11 @@
 import 'calypso/state/signup/init';
-import { initialState, WebsiteContentCollection, MEDIA_UPLOAD_STATES } from './schema';
+import { initialState, MEDIA_UPLOAD_STATES } from './constants';
+import type { WebsiteContentCollectionState } from './types';
 
 export interface WebsiteContentStateModel {
 	signup: {
 		steps: {
-			websiteContentCollection: WebsiteContentCollection;
+			websiteContentCollection: WebsiteContentCollectionState;
```
aneeshd16, line 8: Renamed the `WebsiteContentCollection` type to `WebsiteContentCollectionState` to better reflect its meaning.

client/state/signup/steps/website-content/schema.ts
```diff
@@ -1,140 +0,0 @@
-import type { PageId } from 'calypso/signup/difm/constants';
```
aneeshd16, line 1: Since browser storage is no longer the source of truth, we removed the `withSchemaValidation` function. The schema is also no longer necessary since the entire state already has Typescript definitions.

client/state/signup/steps/website-content/types.ts
```diff
@@ -0,0 +1,84 @@
+import type { Design } from '@automattic/design-picker';
```
aneeshd16, line 1: Consolidated all types in this file.

client/signup/difm/translation-hooks.tsx
```diff
@@ -24,22 +25,24 @@ import type { TranslateResult } from 'i18n-calypso';
...
-		[ PORTFOLIO_PAGE ]: translate( 'Portfolio' ),
-		[ FAQ_PAGE ]: translate( 'FAQ' ),
-		[ SERVICES_PAGE ]: translate( 'Services' ),
-		[ TESTIMONIALS_PAGE ]: translate( 'Testimonials' ),
-		[ PRICING_PAGE ]: translate( 'Pricing' ),
-		[ TEAM_PAGE ]: translate( 'Team' ),
-		[ SHOP_PAGE ]: translate( 'Shop' ),
-	};
-	return pages;
+	return useMemo( () => {
```
aneeshd16, line 28: Since this hook is invoked multiple times when the component re-renders, we can return memoized values.

client/signup/steps/website-content/index.tsx
```diff
@@ -232,38 +253,17 @@ export default function WrapperWebsiteContent(
 		  );
 
 	useEffect( () => {
-		async function fetchSelectedPageTitles() {
-			try {
-				const response: {
```
aneeshd16, line 237: Moved this to a react-query hook:
https://github.com/Automattic/wp-calypso/pull/72075/files#diff-c5b55802aa84610eedd9ab19f63a1457b1e04a9257577d1f435cbadea67aa27c

This has a few advantages:
* Removes redundant local state (`pageTitles`, `isLoading`, `isStoreFlow`).
* Separates fetching logic from component logic.
* Mapping of server response to local state is separate.

client/signup/steps/website-content/wordpress-media-upload.tsx
```diff
@@ -156,6 +156,13 @@ export function WordpressMediaUpload( {
 	const allowedFileTypes = mediaType === 'VIDEO' ? allowedVideoExtensions : allowedImageExtensions;
 	const allowedFileTypesString = allowedFileTypes.map( ( type ) => `.${ type }` ).join();
 
+	// Initialize uploadState if media has already been uploaded.
```
aneeshd16, line 159: In L150, we initialize the `uploadState`, but the `media` prop is updated after the initial render. So the `uploadState` remains as `UPLOAD_STATES.NOT_SELECTED` even though there is a valid `url`.

This hook can be thought of as the equivalent of a `getDerivedStateFromProps` method.

southp (approved):
It works as described and the code also LGTM. Fascinating effort on pulling this off while making the code cleaner :)

Ideally we'd want to decouple a major change like this into smaller and focus-on-one-thing chunks. For example, this PR roughly consists of 3 things:

1. Interacting with the content saving endpoint through `react-query` and its accompanied data structures.
2. Integrating the app state with the content saving endpoint, which in turn makes the reducer no longer the source of truth.
3. Exposing the content saving functionalities onto the UI.

We _might_ be able to split these three things into dependent PRs. However, I can only say _might_ because it might be unnecessarily hard to make each teastable in isolation due to `react-query` 🤔

aneeshd16:
Thanks for the review, @southp!

I was also hesitant about creating a big PR since it is pretty difficult to review! Since we developed this new feature without a design reference, the full functionality needed to be testable. This was also necessary to ensure the endpoint (D98112-code) worked correctly.

> Integrating the app state with the content saving endpoint, which in turn makes the reducer no longer the source of truth.
Exposing the content saving functionalities onto the UI.

I think these two cannot be separated. Shipping one before the other would have broken the experience for our users. 🙂 

Having said that, I'll keep this in mind so that we ship small diffs in the future!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7805552

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7805552).

PR #72080, opened 2023-01-13
i18n: Remove unused lib functions

tyxla:
#### Proposed Changes

This PR removes a couple of functions that were moved from `/lib/i18n-utils` to `@automattic/i18n-utils`, but were not removed afterwards.

#### Testing Instructions

Verify all tests pass and the removed code is not in use.


client/state/concierge/signup-form/test/reducer.js
```diff
@@ -18,7 +17,7 @@ describe( 'concierge/signupForm/reducer', () => {
 	const mockSignupForm = {
 		firstname: 'Foo',
 		lastname: 'Bar',
-		timezone: 'UTC',
+		timezone: 'Asia/Kolkata',
```
tyxla, line 20: We're providing a different time zone because previously this was setting the same one as the default one, making the test useless for the timezone.

client/state/concierge/signup-form/test/reducer.js
```diff
@@ -1,4 +1,3 @@
-import guessTimezone from 'calypso/lib/i18n-utils/guess-timezone';
```
tyxla, line 1: This was the sole use for `guessTimezone`, I've replaced it with hardcoded versions since that's fine for this test IMHO.

yuliyan (approved):
I've verified that these lib functions are not currently not being used and neither were referenced anywhere recently. LGTM! 🚢

PR #49408, opened 2021-01-28
Nav Unification: address color schemes Beta feedback

frontdevde:
### Changes proposed in this Pull Request

This PR aims to address a collection of issues related to color schemes reported in the Beta feedback.

In various color schemes this PR:
- fixes the selected state of menu items in Calypso /me/account
- fixes the selected state of menu items in the Reader
- fixes the unread normal/hover/selected states in the Reader
- fixes the hover color of submenu items in Calypso for Blue and Sunrise color schemes
- adds a visual divider between sidebar and content in the Contrast color scheme

Related issue https://github.com/Automattic/wp-calypso/issues/48009

#### General testing instructions

The before/after changes should be tested in specific themes as outlined below. That said, please also double check that the rest of the color schemes and interface still looks as expected.

When testing locally there's a know issue where the /me/account screen doesn't work in Chrome. You can use Firefox to switch your color scheme in /me/account locally.

#### Shared testing instructions

- Check out and run branch locally or use the calypso.live link
- Log in as an A11n and ensure you're seeing the Nav Unification interface

#### Testing fixed selected state of menu items in Calypso /me/account

|Before|After|
|-|-|
|101065707-77a0e680-3563-11eb-9b4b-9ee0cdcb6b3d|Screenshot 2021-01-28 at 14 59 31|

- Navigate to /me/account
- Ensure that the After state compares to the screenshot above 
- Confirm the result is as expected for the following color schemes: 
  - [ ] Light
  - [ ] Classic Blue
  - [ ] Contrast
  - [ ] Aquatic
 
#### Testing fixes in the Reader

|Before|After|
|-|-|
|Screenshot 2021-01-28 at 15 15 26|Screenshot 2021-01-28 at 15 10 36|

- Navigate to /read
- Ensure that the After states compare to the screenshots above for:
  - [ ] the selected state of menu items in the Reader
  - [ ] the unread normal/hover/selected states in the Reader
  - [ ] the hover color of submenu items in the Reader
- Confirm the results are as expected for the following color schemes: 
  - [ ] Aquatic
  - [ ] Contrast
  - [ ] Blue
  - [ ] Sunrise

#### Testing the addition of a visual divider in the Contrast color scheme

|Before|After|
|-|-|
|Screenshot 2021-01-28 at 15 08 05|Screenshot 2021-01-28 at 15 07 34|

- select the Contrast color scheme
- confirm there is now a visual divider between the sidebar and the content

PR #72104, opened 2023-01-13
Fix solid white circle when reader "Automattic" section is selected

davemart-in:
## Description

Minor bug fix for "Automattic" reader menu. Once selected, it's showing a solid white circle. This PR fixes the issue with no regressions that I can find.

### Before
before

### After
after

### DIFF
--- a/client/my-sites/sidebar/style.scss
+++ b/client/my-sites/sidebar/style.scss
@@ -590,10 +590,6 @@ $font-size: rem(14px);
 				&::after {
 					display: block;
 				}
-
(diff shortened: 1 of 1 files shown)


ollierozdarz:
@davemart-in visually looks great to me!

obenland (approved):
Introduced in #49408. 

Initially I was worried that certain color schemes might break without it but that does not seem to be the case. Tested in Reader and Me section.

davemart-in:
Awesome. Thanks for the review @obenland!

Issue #54421, opened 2021-07-09
CoBlock: publishing a page with Click to Tweet and Logos block auto-dismisses the post-publish panel.


worldomonation:
### Steps to reproduce the behavior

1. Start a new post.
2. Set title of post.
3. Insert the `Click to Tweet` coblock.
4. Fill in the tweet body.
5. Insert the `Logos` coblock.
6. Click on `Publish` and follow through with the process until the post is published.

#### What I expected to happen
7. Post-publish panel remains on screen.

#### What actually happened
7. Post-publish panel auto-dismisses.

### Context
While migrating Selenium E2E tests to Playwright as part of Quality Squad project.

#### Browser / OS version
OS: macOS Big Sur 11.4
Browser: Firefox 91.0a1 (2021-07-09) (64-bit)
Browser: Firefox Nightly 92.0a1 (2021-07-23) (64-bit)

#### Is this specific to the applied theme? Which one?
Reproducible on at least two themes, `Twenty Twenty-One` and `Twenty Nineteen`.

#### Does this happen on simple or atomic sites or both?
Both Simple and Atomic.

#### Is there any console output or error text?
No.

#### Level of impact (Does it block purchases? Does it affect more than just one site?)
Customer: low (minor annoyance)
Automated E2E testing; high (breaks test flow in #54337).

#### Reproducibility (Consistent, Intermittent) Leave empty for consistent.
Consistent

#### Screenshot / Video: If applicable, add screenshots to help explain your problem.
https://user-images.githubusercontent.com/6549265/125141302-7a4f5580-e0c9-11eb-900f-972830016008.mov

### Workaround

Save the post as draft first.

worldomonation:
Appears to be no longer an issue on deployed version `b532f5ca135f184e8359e8eb0514caf3fa25b678`.

Closing for now.

worldomonation:
Reopening the issue as it is reproducible again.

Browser: Firefox Nightly 92.0a1 (2021-07-23) (64-bit)
OS: macOS Big Sur 11.4 (20F71)

worldomonation:
Adding prioritization `Normal` as it impacts all users that use the Click to Tweet coblock, but there is a workaround available.

worldomonation:
Changing priority to Low because CoBlocks has been deprecated on newer sites.

PR #72112, opened 2023-01-13
E2E: discontinue the use of Editor frames where possible.

worldomonation:
#### Proposed Changes

This PR refactors out the usage of the `getEditorHandle` method from specs where possible and feasible.

Context for this change: p1673294397010529-slack-CBTN58FTJ and p1673397395255099-slack-C1A1EKDGQ

In short:
Gutenberg 14.9.1 introduces a nested iframe for the editor canvas if a block-based theme is used (eg. Twenty-Twenty Two). This means that our current way of interacting with the editor using `EditorPage` is no longer workable if such themes are used.

We are currently bypassing the issue entirely by using an older, non-block-based theme (eg. Twenty-Twenty One) but this is unsustainable in the longer term.

This PR is part of a series in an attempt to update the `EditorPage` to work with both non-block-based and block-based themes. 

Key changes:
- rename the locator method to `getEditorWindowLocator`.
- discontinue usage of editor frame in favor of locators.

#### Testing Instructions

  - [x] Gutenberg E2E (desktop)
  - [x] Calypso E2E (desktop)
  - [x] i18n E2E

#### Pre-merge Checklist

Related to #

### DIFF
--- a/packages/calypso-e2e/src/lib/components/editor-toolbar-component.ts
+++ b/packages/calypso-e2e/src/lib/components/editor-toolbar-component.ts
@@ -213,6 +213,17 @@ export class EditorToolbarComponent {
 	/* Publish and unpublish */
 
 	/**
+	 * Returns the text present for the save/publish button.
+	 *
+	 * @returns {Promise<string>} String found on the button.
+	 */
+	async getPublishButtonText(): Promise< string > {
+		const publishButtonLocator = this.editor.locator( selectors.publishButton( 'enabled' ) );
+
+		return await publishButtonLocator.innerText();
+	}
+
+	/**
 	 * Clicks on the primary button to publish the article.
 	 *
 	 * This is applicable for the following scenarios:
--- a/packages/calypso-e2e/src/lib/pages/editor-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/editor-page.ts
@@ -621,13 +621,10 @@ export class EditorPage {
 	async unpublish(): Promise< void > {
 		await this.editorToolbarComponent.switchToDraft();
 
-		const frame = await this.getEditorHandle();
 		// @TODO: eventually refactor this out to a ConfirmationDialogComponent.
-		await frame.click( `div[role="dialog"] button:has-text("OK")` );
+		await this.editor.getByRole( 'button' ).getByText( 'OK' ).click();
 		// @TODO: eventually refactor this out to a EditorToastNotificationComponent.
-		await frame.waitForSelector(
-			'.components-editor-notices__snackbar :has-text("Post reverted to draft.")'
-		);
+		await this.editor.getByRole( 'button', { name: 'Dismiss this notice' } ).waitFor();
 	}
 
 	/**
--- a/packages/calypso-e2e/src/lib/pages/editor-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/editor-page.ts
@@ -159,7 +159,7 @@ export class EditorPage {
 	 *
 	 * @returns A pointer to frame-safe, top-level locator within the editor.
 	 */
-	getEditorLocator(): Locator {
+	getEditorWindowLocator(): Locator {
 		return this.editor;
 	}
 
--- a/packages/calypso-e2e/src/lib/pages/editor-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/editor-page.ts
@@ -2,6 +2,7 @@ import { Page, Frame, ElementHandle, Response, Locator } from 'playwright';
 import { getCalypsoURL } from '../../data-helper';
 import { reloadAndRetry } from '../../element-helper';
 import envVariables from '../../env-variables';
+import { NewPostResponse } from '../../types';
 import {
 	EditorPublishPanelComponent,
 	EditorNavSidebarComponent,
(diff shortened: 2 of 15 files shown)


worldomonation:
Unrelated failure in I18N.

packages/calypso-e2e/src/lib/components/editor-toolbar-component.ts
```diff
@@ -212,6 +212,17 @@ export class EditorToolbarComponent {
 
 	/* Publish and unpublish */
 
+	/**
+	 * Returns the text present for the save/publish button.
+	 *
+	 * @returns {Promise<string>} String found on the button.
+	 */
+	async getPublishButtonText(): Promise< string > {
```
worldomonation, line 220: This method is part of the refactor to the `publish` method to eliminate relying on flaky UI elements.

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -563,39 +564,51 @@ export class EditorPage {
...
-		if ( await this.editorPublishPanelComponent.panelIsOpen() ) {
-			// Invoke the second stage of the publish step which handles the
-			// publish checklist panel if it is present.
-			await this.editorPublishPanelComponent.publish();
+		// Determine whether the post/page is yet to be published or the post/page
+		// is merely being updated.
+		// If not yet published, a second click on the EditorPublishPanelComponent
+		// is added to the array of actions.
+		if ( publishButtonText.toLowerCase() === 'publish' ) {
+			actionsArray.push( this.editorPublishPanelComponent.publish() );
```
worldomonation: This change, while seemingly unrelated to the task of eliminating the use of Editor handles, is actually quite important.

In its current form, the `publish` method uses entirely UI-driven methods to publish then obtain the published post/page's URL.

Not only is this possibly flaky as the post/page publish toast fades after some time, the post-publish confirmation panel can also be auto-dismissed due to issues with some blocks (eg. https://github.com/Automattic/wp-calypso/issues/54421).

By intercepting the response to the publishing endpoint, the process to obtain the post/page's URL is made much more reliable.

test/e2e/specs/blocks/blocks__coblocks__extensions__cover-styles.ts
```diff
@@ -57,17 +58,16 @@ skipDescribeIf( features.siteType === 'atomic' )(
 			await coverBlock.upload( imageFile.fullpath );
 			// After uploading the image the focus is switched to the inner
 			// paragraph block (Cover title), so we need to switch it back outside.
-			const editorFrame = await editorPage.getEditorHandle();
-			await editorFrame.click( '.wp-block-cover', { position: { x: 1, y: 1 } } );
+			editorWindowLocator = editorPage.getEditorWindowLocator();
```
worldomonation, line 61: Everything below here is the replacement of editor frames with editor locators.

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -39,7 +39,7 @@ const selectors = {
 	// Welcome tour
 	welcomeTourCloseButton: 'button[aria-label="Close Tour"]',
 };
-const EXTENDED_TIMEOUT = 90 * 1000;
+const EXTENDED_TIMEOUT = 30 * 1000;
```
worldomonation, line 42: This is not strictly related to the core fix proposed by this PR. However, I feel reasonably safe that we can reduce the overly generous 90s timeout now that we are not performing a prolonged wait for the editor to fully load (see below).

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -120,46 +120,23 @@ export class EditorPage {
 	 * @returns {Promise<Frame>} iframe holding the editor.
 	 */
 	async waitUntilLoaded(): Promise< void > {
-		// Once https://github.com/Automattic/wp-calypso/issues/57660 is resolved,
-		// the next line should be removed.
-		const editor = await this.getEditorHandle();
-
-		const titleLocator = editor.locator( selectors.editorTitle );
-		await titleLocator.waitFor( { timeout: EXTENDED_TIMEOUT } );
```
worldomonation, line 128: This is a moderately large change.

Instead of waiting for the Title block in the editor to appear, I have made the calculated decision to wait for a request to finish loading.

I feel confident that this will not lead to intermittent failures:
1. if the request and response to the`posts` endpoint has not completed within 30s, it is unlikely that waiting additional time for the editor title to appear would have been any safer.
2. waiting for the editor title causes issues with Gutenberg 14.9.1 where an iframe now wraps the editor canvas. 

For the second point, while there is a follow-up PR to address the inner iframe, I decided that relying less on UI elements and more on the network requests would ultimately result in a more reliable loading pattern.

I'm open to hearing opinions on this change. If required, it is possible to abandon this change and look forward to a fix in the follow-up PR.

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -563,39 +540,65 @@ export class EditorPage {
...
+		if ( this.target === 'atomic' ) {
+			actionsArray.push( this.page.waitForResponse( /v2\/(posts|pages)\/[\d]+/ ) );
+		} else {
+			actionsArray.push( this.page.waitForResponse( /sites\/[\d]+\/(posts|pages)\/[\d]+.*/ ) );
+		}
 
-		if ( await this.editorPublishPanelComponent.panelIsOpen() ) {
-			// Invoke the second stage of the publish step which handles the
-			// publish checklist panel if it is present.
-			await this.editorPublishPanelComponent.publish();
```
worldomonation, line 578: The refactor of `publish` method is little bit of a scope creep that was prompted by the removal of the `getEditorHandle` method entirely. 

As the original `getPublishedURLFromToast` method used the now-removed `getEditorHandle` method, I figured I'll take this opportunity to refactor how the published article's URL is obtained. My primary reasoning is that using UI elements only is flaky.

With that said, if this should be in a separate PR, I am all for it.

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -621,28 +624,23 @@ export class EditorPage {
...
+	 * post-publish toast is still visible on the page.
 	 *
+	 * @deprecated Please use the return value from `EditorPage.publish` where possible.
 	 * @returns {URL} Published article's URL.
 	 */
 	async getPublishedURLFromToast(): Promise< URL > {
-		const frame = await this.getEditorHandle();
-
-		const toastLocator = frame.locator( selectors.toastViewPostLink );
+		const toastLocator = this.editor.locator( selectors.toastViewPostLink );
```
worldomonation, line 643: This method was originally removed, but I had to restore it to maintain support for the `privacy-testing` suite, in particular the `private page/post` spec.

worldomonation:
@dpasque This can be a confusing PR to review and I've tried to annotate it as much as possible, but please do leave me a ping if you're unsure about anything be it my reasoning or the changes.

dpasque:
@worldomonation Thank you so much for making this change! 🎉  I think the changes here look good (and will approve), but wanted to ask more of a philosophical question!

In the two "extra" changes here, you opt to rely on network requests rather than visual indicators in the UI to make things more stable. I was wondering if you could share more about your perspective there. Do you think that's a general strategy we should adopt going forward in these E2E tests?

My gut reaction leans the other way -- favoring UI indicators over network requests, but it's not something I feel particularly strongly or confidently about!

dpasque (approved):
🙇

worldomonation:
>  I was wondering if you could share more about your perspective there. Do you think that's a general strategy we should adopt going forward in these E2E tests?

Good question @dpasque. This is definitely a philosophical question and there is no right or wrong answer here.

In this case, my biggest reason for using API and network requests is the possible flakiness of relying solely on UI elements. I cannot seem to find it now, but earlier on in the migration project I encountered issues when relying solely on UI when extracting the published article URL. 

One such issue is https://github.com/Automattic/wp-calypso/issues/54421, where the post-publish panel was dismissed immediately upon publishing. In fact, this is the sole reason that the `getPublishedURLFromToast` exists - it was a workaround for the above issue.

Broadening the conversation to more generally - my informal criteria for deciding on when to use network/API requests typically goes as follows:

1. is the UI interaction flaky? 
2. does it take an inordinate amount of time if solely relying on UI interactions?
3. is this a "core" flow?

Let me elaborate a bit more on each point.

1 covers the situation described above with the post-published panel so I won't rehash.

2 covers the situation where an action can be performed quicker programmatically but may take many clicks and screens to complete via UI. Flows such as closing an account or deleting a site falls under this scenario.

3 supersedes the other two points in importance, however. Here, I consider whether the action or flow forms the core of what we are looking to test. For instance, if user creation is part of the tested flow, each step is performed via the UI. However, if user creation is part of a setup and thus is not part of the core flow, then I would prefer to use API/network actions to get this done as quickly as possible.

I hope this clarifies my reasoning behind the "additional" changes.

dpasque:
@worldomonation -- that's all very helpful, thank you for the detailed answer! 😄 Sounds good to me! 👍

fullofcaffeine:
Late to the party here, sorry for the delay, and thanks for working on this! Kudos @dpasque, for the prompt review, too!

@worldomonation A few questions for when you have time:
1. Regarding the approach you're using here, we're basically skipping the iframe and selecting the elements directly? Why wasn't this approach considered before?
2. Do you know why selecting an inner iframe doesn't work well? (not suggesting we go down this path, just curious)

Thanks!

worldomonation:
>     1. Regarding the approach you're using here, we're basically skipping the iframe and selecting the elements directly? Why wasn't this approach considered before?

We aren't selecting the elements directly - the iframe is still involved. The main goal of this PR was to stop using the raw Frame or Page object in order to perform actions or wait on selectors. 

In order to do that, I had to refactor out all usages of `getElementHandle` and replace them with Locators referencing the overall editor window. The editor window Locator is already obtained and assigned as part of the constructor process, so it was a short hop to replace all instances of `getEditorHandle` throughout the specs and POMs within the Locator equivalent calls.

>     2. Do you know why selecting an inner iframe doesn't work well? (not suggesting we go down this path, just curious)
> 
> 
> Thanks!

Selecting the inner iframe by default is one way to go, but it would run into issues if/when we have specs that test against a non-block-based theme site (eg. 2021). 

The challenge I'm facing is how best to support the following permutations without having to create a specific POM per permutation:

- simple and block-based theme
- simple and non-block-based theme
- atomic and block-based theme
- atomic and non-block-based theme

PR #72116, opened 2023-01-14
Pricing: Add conditional logic to format symbol position correctly in PlanPrice

### LINKED ISSUE #72115: PlanPrice always displays its currency symbol before the price regardless of localization
The `PlanPrice` component is used for a variety of purposes and has a complex API. One of its biggest features is the ability to split up a price, rendering the currency symbol, the integer section, and the decimal section of the string in different HTML tags.

I've done some recent refactoring to clean it up, but in doing so I noticed something awkward; the currency symbol is always rendered _before_ the number of the price, even if the localized currency format says to render it afterward (eg: Canadian French).

This can create awkward UI inconsistencies where spacing between the currency symbol and the number is not correct or multiple representations of the currency on the same page put the symbol on different sides.

Screenshot 2023-01-13 at 2 35 33 PM

PlanPrice should probably be updated to put the symbol on the localized side.

jjchrisdiehl:
#### Proposed Changes

This PR adds logic to use the `symbolPosition` provided by `getCurrencyObject()` to left or right align the currency symbol correctly for `<PlanPrice>`.

#### Testing Instructions
**Left aligned symbols**
* Go to /me/account and change **Interface Language** to English
* Go to any page that uses the <PlanPrice> component, i.e. plans, purchase management etc
* Currency symbol should be left aligned 

**Right aligned symbols**
* Go to /me/account and change **Interface Language** to French Canadian
* Go to any page that uses the <PlanPrice> component, i.e. plans, purchase management etc
* Currency symbol should be right aligned 

Fixes https://github.com/Automattic/wp-calypso/issues/72115

### DIFF
--- a/client/my-sites/plan-price/index.tsx
+++ b/client/my-sites/plan-price/index.tsx
@@ -231,7 +231,10 @@ function FlatPriceDisplay( {
 	className?: string;
 	isSmallestUnit?: boolean;
 } ) {
-	const { symbol: currencySymbol } = getCurrencyObject( smallerPrice, currencyCode );
+	const { symbol: currencySymbol, symbolPosition } = getCurrencyObject(
+		smallerPrice,
+		currencyCode
+	);
 	const translate = useTranslate();
 
 	// TODO: render currencySymbol on the localized side (before or after) of
@@ -239,12 +242,13 @@ function FlatPriceDisplay( {
 	if ( ! higherPrice ) {
 		return (
 			<span className={ className }>
-				{ currencySymbol }
+				{ symbolPosition === 'before' ? currencySymbol : null }
 				<PriceWithoutHtml
 					price={ smallerPrice }
 					currencyCode={ currencyCode }
 					isSmallestUnit={ isSmallestUnit }
 				/>
+				{ symbolPosition === 'after' ? currencySymbol : null }
 			</span>
 		);
 	}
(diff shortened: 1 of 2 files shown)


jjchrisdiehl:
image

The two figures in the image above currently show the same format, but they achieve this using two different formatter functions and I believe there might be a small formatting issue with `PurchaseMetaPrice`.

While the code in this PR will fix instances of `<PlanPrice>`, the `<PurchaseMetaPrice>` uses `formatCurrency` instead of `getCurrencyObject.symbolPosition` as seen in `PlanPrice`. 

I may be wrong, but I believe formatCurrency doesn't take into account the symbolPosition, hence using `getCurrencyObject`, so it might be useful to update `PurchaseMetaPrice` to use `PlanPrice` as well? 

Maybe @sirbrillig has more insight here.

sirbrillig:
`formatCurrency` and `getCurrencyObject` are both part of the `@automattic/format-currency` package and operate with the same information. They both put the currency symbol in the correct place, but you have to position it manually with `getCurrencyObject`. `formatCurrency` is much easier to use and does the positioning itself. You should only need `getCurrencyObject` (and therefore `PlanPrice`) if you want to manually format a price using custom HTML.

I'm curious, though, how you came to the conclusion that `formatCurrency` was not correctly placing the currency symbol. In the screenshot above it looks correct.

jjchrisdiehl:
> I'm curious, though, how you came to the conclusion that `formatCurrency` was not correctly placing the currency symbol. In the screenshot above it looks correct.

I agree, the screenshot is correct, I guess I am not exactly sure where in the formatCurrency function that the positioning is done.

sirbrillig:
> I guess I am not exactly sure where in the formatCurrency function that the positioning is done.

JavaScript does it, actually! Using the `Intl.NumberFormat.formatToParts` function. It happens [here](https://github.com/Automattic/wp-calypso/blob/5653520cb4dbe516ad7a5829cf089705cbf69609/packages/format-currency/src/index.ts#L103). You can read more about the decisions and steps that led to this API in https://github.com/Automattic/wp-calypso/pull/71387. You can also see where we started using `formatCurrency` for receipts in https://github.com/Automattic/wp-calypso/pull/71952.

sirbrillig (approved):
Works as described!

Screenshot 2023-03-17 at 5 09 58 PM

Screenshot 2023-03-17 at 5 09 41 PM

spraveenitpro (approved):
Static analysis looks good! LGTM


<details>
<summary>Before Patch</summary>


</details>


<details>
<summary>After patch:</summary>


</details>

PR #72129, opened 2023-01-15
Add site_intent to Sensei onboarding flow

mikeyarce:
#### Proposed Changes

* Add the `site_intent` option to the site. This will let us run scripts after the site has been created that specifically target sites created in this flow.

#### Testing Instructions

* Go through the Sensei flow (http://calypso.localhost:3000/setup/sensei/senseiSetup)
* After the flow has finished, you can check the site options page `wp-admin/options.php` and check for the `site_intent` option and make sure it says `sensei`.


gikaragia (approved):
This seems reasonable!

PR #72155, opened 2023-01-16
Copy Site: Generate destination site slug based on source slug

kozer:
#### Proposed Changes

When I copied `test.wpcomstaging.com`, the new site was `<my username><random number>.wpcomstaging.com`. I would've expected it to be `test12.wpcomstaging.com` or similar (original slug with a random hash appended).

#### Testing Instructions

* Select an atomic site under `/sites`
* Click on the ellipsis menu and select `Copy Site`
* Continue with the steps till you navigate to the `/checkout` screen.
* Navigate back to `/sites`
* Ensure that there is a new site created named similar to your previous atomic site. (Eg: If you had `test.wpcomstaging.com`, look for something like `test12.wordpress.com`)

copy_site_fix_slug_1

copy_site_fix_slug_2

copy_site_fix_slug_3

#### Pre-merge Checklist

Related to https://github.com/Automattic/dotcom-forge/issues/1496


danielbachhuber (commented):
>  (original slug with a random hash appended)

Is it possible to append a random hash instead of a number?

Also, maybe we should store the original site name in an option so that a copy of a copy follows a similar pattern (instead of two random hashes appended)?

kozer:
> > (original slug with a random hash appended)
> 
> Is it possible to append a random hash instead of a number?
> 
> Also, maybe we should store the original site name in an option so that a copy of a copy follows a similar pattern (instead of two random hashes appended)?
Thanks @danielbachhuber for your comment!
The name "randomization" is happening in the backend. This is core functionality and it tries to keep this randomness by having as few characters as possible. (eg it is mentioned that `Subdomains can have at most 63 characters. Let's also limit the length of the random numbers part to 8 to try to keep it short`)
I don't know what the impact of this change is gonna be, and I think that the value is not worth the possibility to break something.

danielbachhuber:
> The name "randomization" is happening in the backend. This is core functionality and it tries to keep this randomness by having as few characters as possible.

Ok, fair enough.

sejas (commented):
I've tested it and it's looking good with `wordpress` subdomains.
My concern is if we copy a site with a custom domain.
- `example.com` would generate a good random variation using the word `example`
- `www.exmaple.com` will produce a weird variation using `www` as a base for the new subdomain.

@kozer What do you think? Maybe is an edge case. As a workaround we could use the "original" slug instead of the mapped domain.

danielbachhuber:
@sejas @kozer  @automattic/yolo Let’s try incorporating the domain picker into the flow and hold off on additional algorithmic generation for now. At this point, it seems like we might need to go the domain picker route.

kozer:
> I've tested it and it's looking good with `wordpress` subdomains.
> My concern is if we copy a site with a custom domain.
> - `example.com` would generate a good random variation using the word `example`
> - `www.exmaple.com` will produce a weird variation using `www` as a base for the new subdomain.
> 
> @kozer What do you think? Maybe is an edge case. As a workaround we could use the "original" slug instead of the mapped domain.

You are right @sejas . This will be strange. Let me try to add the domain picker.
@danielbachhuber Do you want to close the pr,and create a new issue?

wojtekn:
> Let’s try incorporating the domain picker into the flow and hold off on additional algorithmic generation for now. At this point, it seems like we might need to go the domain picker route.

I like that we are going with an explicit domain picker route for the Copy Site feature. Auto-generation may be better for staging/sandbox cloning site purposes.

However, if we work on auto-generation in the future, another idea to consider is using three random words for the new site slug, similar to what we do on Calypso Live:
 
https://container-goofy-cannon.calypso.live/sites

danielbachhuber:
> However, if we work on auto-generation in the future, another idea to consider is using three random words for the new site slug, similar to what we do on Calypso Live:

@wojtekn Ah, this is a great idea. I think we should use this for staging sites.

danielbachhuber:
> Do you want to close the pr,and create a new issue?

@kozer We can close this PR, yes. The new pull request can be submitted against https://github.com/Automattic/dotcom-forge/issues/1478

PR #72165, opened 2023-01-16
Fix arrow buttons in the Email Stats page

### LINKED ISSUE #71932: Fix problem with left and right arrow when `hours` is selected in Email Stats Details page

JuanLucha:
#### Proposed Changes

This PR fixes the left and right arrows of the Email Stats chart when the `hours` period is selected.

Screenshot 2023-01-16 at 19 15 54

#### Testing Instructions

2. Insert manually in the browser address bar an URL with this format: https://calypso.localhost:3000/stats/email/open/[the-doomain-of-your-blog]/day/[post-id]?flags=newsletter/stats. Note that you can change day for week and month.
3.  Click on the previous or the next arrow (as shown in the above picture). Wait for the data to come in. You should see the right day showing in the chart.

#### Known things to fix in next PRs
- We need to fix the selected column (the light blue column in the chart). Shouldn't be there as per design.
- The chart loads from the endpoint every time, so if we click the next arrow and then then previous arrow, the endpoint is still called. That doesn't happen with `days` is selected.

Fixes https://github.com/Automattic/wp-calypso/issues/71932

### DIFF
--- a/client/my-sites/stats/stats-email-chart-tabs/index.jsx
+++ b/client/my-sites/stats/stats-email-chart-tabs/index.jsx
@@ -47,6 +47,7 @@ class StatModuleChartTabs extends Component {
 		chartTab: PropTypes.string,
 		switchTab: PropTypes.func,
 		queryDate: PropTypes.string,
+		queryHour: PropTypes.string,
 		recordGoogleEvent: PropTypes.func,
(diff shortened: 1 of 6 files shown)


TimBroddin (approved):
Nice one!

Issue #72106, opened 2023-01-13
Stats: Extend Year in Review display period


jeffgolenski:
During the first month of a new year, the Year in Review section at the top of Insights can look quite bleak. See below

![Image](https://user-images.githubusercontent.com/214813/212425341-7afc97c7-7560-42cd-bc7c-0fa0203737d5.png)

Let's change the logic, so the year in review shows the previous year until February 1. So a user should see the "2022 in review" until February 1, 2023. This will allow a bit more impactful data to accumulate before we make the switch, and a customer won't have to stare at a bunch of 0s after the ball drops on new years eve!

cc @keoshi @jsnmoon

jsnmoon:
This is a solid idea; moved the card to the todo column.

PR #72181, opened 2023-01-16
Stats: Avoid showing low scores in January

grzegorz-cp:
#### Proposed Changes

* show the previous year's annual stats in January

#### Testing Instructions

* open a live branch
* navigate to the Insights page and verify that 2022 is visible in the annual highlight component
* verify that you can still see 2023 when using the arrows on the right-hand side
* go to another blog and check that it loads 2022 instead of 2023

#### Pre-merge Checklist

Related to #72106


annacmc (approved):
Nice one! Tested out & works as expected with the correct 2022/2023 years on multiple different blogs 🙌  LGTM 👍

kangzj:
I have mixed feelings over the approach - it does provide some kind of convenience for user to review the previous year in Jan but it could also be confusing as well 🤔 Wondering what the rest of the team think @Automattic/red

wilmersondasilva:
> I have mixed feelings over the approach - it does provide some kind of convenience for user to review the previous year in Jan but it could also be confusing as well 🤔 Wondering what the rest of the team think @Automattic/red

Maybe we can decide if we go with 2022 or 2023 based on the 2023 data? For instance, if there are at least three properties with a value greater than 0 for 2023, we show 2023 or something similar to this.

grzegorz-cp:
It's not like we are cutting off the people from the data. We are only "automatically clicking the arrow" in January.
It's also a simple logic. In January, we show the previous year for people to reflect on.

I'm afraid that it will be much more arbitrary if we start checking specific stats because we either check for several 0 stats or a number of posts. Even one new post can still result in many "0"s. This feels like a balancing game which will always be bad in some cases.

kangzj (approved):
LGTM

PR #72237, opened 2023-01-17
WooExpress Trial: Users should not be able to launch sites on WooExpress Trial plan

### LINKED ISSUE #71859: WooExpress Trial: Site launch needs to be disabled on trial plan
At `/settings/general`: Sites on the trial are created in Coming Soon mode and should not be able to launch/go public until they've paid for a plan. When a site is first created with the trial plan, the privacy settings look like this:

Screen Shot 2023-01-10 at 3 12 53 PM

Clicking "Launch" launches the site; this should be disabled for folks on the trial plan. Instead, there should be a nudge noting the user has to have a paid plan to launch, and we should direct them to the `/plans/[siteSlug]` page to upgrade.

A launched site's settings look like this:

screen-shot-2023-01-09-at-9 43 35-am

I'm not sure if folks will be able to have a site that is already launched on the eCommerce trial? If so, we also need to account for that by blocking the ability to set the site to Public.

See this post for more information/mockups: peapX7-1fL-p2

valterlorran:
#### Proposed Changes

* Adds a nudge on the "Launch site" card to let the users know that to launch the site, they need to upgrade to an eCommerce plan.

Screenshot 2023-01-24 at 17 01 14

#### Testing Instructions

* On your site with the WooExpress trial plan, navigate to `/settings/general/{SITE}`
* Scroll down to the "Launch site" card.
* You should see the nudge inviting the user to upgrade to an eCommerce plan.
* You can also click on the nudge, it should take you to the plans page.

Closes #71859

### DIFF
--- a/client/my-sites/site-settings/form-general.jsx
+++ b/client/my-sites/site-settings/form-general.jsx
@@ -585,6 +585,7 @@ export class SiteSettingsFormGeneral extends Component {
 		const btnText = translate( 'Launch site' );
 		let querySiteDomainsComponent;
 		let btnComponent;
+		let eCommerceTrailUpsell;
 		const isLaunchable = ! isSiteOnECommerceTrial;
 
 		if ( 0 === siteDomains.length ) {
@@ -616,10 +617,26 @@ export class SiteSettingsFormGeneral extends Component {
 
 		const LaunchCard = showPreviewLink ? CompactCard : Card;
 
+		if ( isSiteOnECommerceTrial ) {
+			eCommerceTrailUpsell = (
+				<a href={ `/plans/${ siteSlug }?plan=business-bundle-monthly` }>
+					<Notice
+						icon="info"
+						showDismiss={ false }
+						text={ translate(
+							'In order for everyone to see your store, you just need to subscribe to a eCommerce plan. Subscribe'
+						) }
+						isCompact={ false }
+					></Notice>
+				</a>
+			);
+		}
+
 		return (
 			<>
 				<SettingsSectionHeader title={ translate( 'Launch site' ) } />
 				<LaunchCard>
+					{ eCommerceTrailUpsell }
 					<div className="site-settings__general-settings-launch-site">
 						<div className="site-settings__general-settings-launch-site-text">
 							<p>
--- a/client/my-sites/site-settings/form-general.jsx
+++ b/client/my-sites/site-settings/form-general.jsx
@@ -618,17 +618,24 @@ export class SiteSettingsFormGeneral extends Component {
 		const LaunchCard = showPreviewLink ? CompactCard : Card;
 
 		if ( isSiteOnECommerceTrial ) {
+			const eCommerceTrialUpsellText = translate(
+				'Before you can share your store with the world, you need to {{NoticeAction}}pick a plan{{/NoticeAction}}.',
+				{
+					components: {
+						NoticeAction: (
+							<NoticeAction href={ `/plans/${ siteSlug }?plan=ecommerce-bundle-monthly` } />
+						),
+					},
+				}
+			);
 			eCommerceTrialUpsell = (
-				<a href={ `/plans/${ siteSlug }?plan=business-bundle-monthly` }>
-					<Notice
-						icon="info"
-						showDismiss={ false }
-						text={ translate(
-							'In order for everyone to see your store, you just need to subscribe to a eCommerce plan. Subscribe'
-						) }
-						isCompact={ false }
-					></Notice>
-				</a>
+				<Notice
+					className="site-settings__ecommerce-trial-notice"
+					icon="info"
+					showDismiss={ false }
+					text={ eCommerceTrialUpsellText }
+					isCompact={ false }
+				></Notice>
 			);
 		}
 
--- a/client/my-sites/site-settings/form-general.jsx
+++ b/client/my-sites/site-settings/form-general.jsx
@@ -44,6 +44,7 @@ import siteHasFeature from 'calypso/state/selectors/site-has-feature';
 import { getDomainsBySiteId } from 'calypso/state/sites/domains/selectors';
 import { usePremiumGlobalStyles } from 'calypso/state/sites/hooks/use-premium-global-styles';
 import { launchSite } from 'calypso/state/sites/launch/actions';
+import { isSiteOnECommerceTrial as getIsSiteOnECommerceTrial } from 'calypso/state/sites/plans/selectors';
 import {
 	getSiteOption,
 	isJetpackSite,
@@ -575,6 +576,7 @@ export class SiteSettingsFormGeneral extends Component {
 			fields,
 			hasSitePreviewLink,
 			site,
+			isSiteOnECommerceTrial,
 		} = this.props;
 
 		const launchSiteClasses = classNames( 'site-settings__general-settings-launch-site-button', {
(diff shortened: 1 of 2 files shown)


client/my-sites/site-settings/form-general.jsx
```diff
@@ -583,12 +585,18 @@ export class SiteSettingsFormGeneral extends Component {
 		const btnText = translate( 'Launch site' );
 		let querySiteDomainsComponent;
 		let btnComponent;
+		let eCommerceTrailUpsell;
```
markbiek:
```suggestion
		let eCommerceTrailUpsell;
```

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,26 @@ export class SiteSettingsFormGeneral extends Component {
 
 		const LaunchCard = showPreviewLink ? CompactCard : Card;
 
+		if ( isSiteOnECommerceTrial ) {
+			eCommerceTrailUpsell = (
```
markbiek:
```suggestion
			eCommerceTrialUpsell = (
```

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,26 @@ export class SiteSettingsFormGeneral extends Component {
...
+					></Notice>
+				</a>
+			);
+		}
+
 		return (
 			<>
 				<SettingsSectionHeader title={ translate( 'Launch site' ) } />
 				<LaunchCard>
+					{ eCommerceTrailUpsell }
```
markbiek:
```suggestion
					{ eCommerceTrialUpsell }
```

markbiek (commented):
Functionality works as expected. Just needs a couple of typos fixed.

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,26 @@ export class SiteSettingsFormGeneral extends Component {
...
 		const LaunchCard = showPreviewLink ? CompactCard : Card;
 
+		if ( isSiteOnECommerceTrial ) {
+			eCommerceTrialUpsell = (
+				<a href={ `/plans/${ siteSlug }?plan=business-bundle-monthly` }>
+					<Notice
+						icon="info"
+						showDismiss={ false }
+						text={ translate(
+							'In order for everyone to see your store, you just need to subscribe to a eCommerce plan. Subscribe'
```
sixhours: I think "Subscribe" would make sense as a `<NoticeAction />` linked to the plans page, to make it more clear this notice is click-able, rather than wrapping the whole component in an anchor.

I know this is following the given design, but the wording here feels off to me. It would be good to get another set of eyes on the copy before we add the "String Freeze" label to get it ready for translation.

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,26 @@ export class SiteSettingsFormGeneral extends Component {
 
 		const LaunchCard = showPreviewLink ? CompactCard : Card;
 
+		if ( isSiteOnECommerceTrial ) {
+			eCommerceTrialUpsell = (
+				<a href={ `/plans/${ siteSlug }?plan=business-bundle-monthly` }>
```
daledupreez: Shouldn't this be the eCommerce bundle?

client/my-sites/site-settings/form-general.jsx
daledupreez: Agreed on both counts: we should use the `NoticeAction` component for the CTA, and the wording here is a bit awkward.

@verofasulo, could you take a first pass at the wording for us? My thinking is that something like the following would be better:
> You can launch your store by subscribing to a paid eCommerce plan [Subscribe]

or 

> To launch your store, you need to subscribe to a paid eCommerce plan [Subscribe]

verofasulo:
Hey @daledupreez 👋

> could you take a first pass at the wording for us? 

@FlorianBru submitted a UX copy request to the marketing team. The copy review will be finalized by Jan 23 - Does this work for you? 

In the meantime, this is the suggestion that @FlorianBru made:

image

client/my-sites/site-settings/form-general.jsx
valterlorran: I fixed it in my last commit. Thanks for noticing that :)

valterlorran:
Thanks, @verofasulo. I applied your suggestion in my last commit.

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,33 @@ export class SiteSettingsFormGeneral extends Component {
...
+			const eCommerceTrialUpsellText = translate(
+				'Before you can share your store with the world, you need to {{NoticeAction}}pick a plan{{/NoticeAction}}.',
+				{
+					components: {
+						NoticeAction: (
+							<NoticeAction href={ `/plans/${ siteSlug }?plan=ecommerce-bundle-monthly` } />
+						),
+					},
+				}
+			);
```
daledupreez: Given that we've switched to inline text for the CTA, we should use a more standard `a` tag, and not a `NoticeAction`. If you take a look at the [Notice dev docs](https://wpcalypso.wordpress.com/devdocs/design/notice), the `NoticeAction` is intended for a separate CTA (which we had before), and not an inline link.
```suggestion
			// TODO: flesh this out and wire it up, but we want to be able to understand how often people click on the CTA
			// We should also trigger an event when the notice is rendered - can we do that via a functional wrapper component that uses `useEffect()` to trigger a render based on the incoming props?
			const recordTracksEventForClick = recordTracksEvent( 'calypso_ecommerce_trial_launch_banner_click' );
			const eCommerceTrialUpsellText = translate(
				'Before you can share your store with the world, you need to {{a}}pick a plan{{/a}}.',
				{
					components: {
						a: (
							<a href={ `/plans/${ siteSlug }?plan=ecommerce-bundle-monthly` } onClick={ recordTracksEventForClick }  />
						),
					},
				}
			);
```

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +617,33 @@ export class SiteSettingsFormGeneral extends Component {
...
+			eCommerceTrialUpsell = (
+				<Notice
+					className="site-settings__ecommerce-trial-notice"
+					icon="info"
+					showDismiss={ false }
+					text={ eCommerceTrialUpsellText }
+					isCompact={ false }
+				></Notice>
```
daledupreez: Minor stylistic note: You can also use `/>` to close off the element.
```suggestion
				/>
```

daledupreez (commented):
@valterlorran, can you refactor the `NoticeAction` into a simpler component and update the CSS to match?

And can you consider whether you'd like to add Tracks events in this PR or a follow-up? I think we should also handle that work in this PR so we can safely wait for translations to come in.

client/my-sites/site-settings/form-general.jsx
```diff
@@ -609,10 +618,38 @@ export class SiteSettingsFormGeneral extends Component {
...
+		if ( isSiteOnECommerceTrial ) {
+			const recordTracksEventForClick = () =>
+				recordTracksEvent( 'calypso_ecommerce_trial_launch_banner_click' );
+			const eCommerceTrialUpsellText = translate(
+				'Before you can share your store with the world, you need to {{a}}pick a plan{{/a}}.',
+				{
+					components: {
+						a: (
+							<a
+								href={ `/plans/${ siteSlug }?plan=ecommerce-bundle-monthly` }
```
sixhours: We should probably use the plan constant (although I recognize this will probably change!) rather than a hard-coded value.

valterlorran: Fixed it :)

sixhours (approved):
Working well! Once the copy is finalized/cleared let's add the String Freeze label to get translations started. :)

daledupreez:
@valterlorran and @sixhours, I think we can get this merged and deployed without waiting for the strings to be translated, as no users should see this upsell yet.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7771747

Thank you @valterlorran for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7771747).

FlorianBru:
Just a quick update, because the copy was changed. So the new design and content look like this:
image

PR #72240, opened 2023-01-17
Only log react render duration stat if uncached

noahtallen:
#### Proposed Changes
This stat (`ssr.react_render.duration`) is always logged no matter what. It measures the time it takes to render the react markup to an HTML string. Since we always log it, though, the stat includes times the markup is pulled from cache.

This makes the stat less useful, since we want to use it to understand how the actual render mechanism performs without cache.

#### Testing Instructions
TC, verify logic


tyxla (commented):
The change makes sense, thanks 👍

PR #72243, opened 2023-01-17
Removed storeFeatures step from stepper

### LINKED ISSUE #72157: Seller Experience: Remove the storeFeatures step
The `storeFeatures` step is no longer in use since #66867. This issue is to remove the step entirely.

nelsonec87:
#### Proposed Changes

* This PR removed the storeFeatures step from stepper, since it not used anymore.

#### Testing Instructions

- Go through the site-setup flow (start at `/start`), especially the `sell` intent,  and confirm nothing is broken.
- There shouldn't be any changes from the user perspective!

Closes https://github.com/Automattic/wp-calypso/issues/72157


sixhours (commented):
Signup still works! Looks like a unit test is failing as a result of these changes:

Screen Shot 2023-01-18 at 9 18 24 AM

I think there are other instances of `useSiteSetupFlowProgress` that still have a third `storeType` arg.

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -192,16 +190,7 @@ const siteSetupFlow: Flow = {
 			switch ( currentStep ) {
 				case 'options': {
 					if ( intent === 'sell' ) {
-						/**
```
sixhours: Might be good to leave this comment in for context.

nelsonec87: Agreed! :)

nelsonec87:
> I think there are other instances of useSiteSetupFlowProgress that still have a third storeType arg.

Yes, it was it 👍

sixhours (approved):
Looks good!

PR #72248, opened 2023-01-18
Allow theme demo URIs to be optional

noahtallen:
#### Proposed Changes
Problem: sometimes a theme's demo URI is set to `false` when the demo site can't be found. This is usually an accident when a new theme is deployed. When this happens, the theme fails validation against the schema, because the schema expects the demo URI to be a string.

The result is that the theme response data is **never** cached in this scenario. Or rather, it's written to the cache, but when it's pulled from the cache, it fails validation and the data is **always** fetched again. (And the cycle continues -- garbage in, garbage out.)

The result of _that_ is terrible performance of the `/themes` route. About 10x worse than normal. (2000ms vs 200ms).

So to avoid tanking performance when this happens, we need to do two things:
1. Allow demo_uri's to be optional. This way, the theme can be cached, but the demo site button won't show up.
2. Discard any theme which doesn't have a demo_uri (e.g. which fails schema validation.) This way, the other themes can continue to be cached.

The first option is a lot easier to implement because every component I could find already handles the case that demo uri is falsey. So I just updated the other types related to that.

The second option is harder, but probably ideal in the long term. We shouldn't be persisting stuff which will already fail schema validation! (But the schema validation reducers only support validation at the time of de-serialization, meaning it'll be hard to validate when new data comes in.)

#### Testing Instructions
**Note: these won't work once Makoney's demo site is correctly set up.** The instructions do work and test correctly at the time of writing.

1. Load calypso.localhost with these changes
2. It should take ~2s to load. See "long request duration" messages in the logs.
3. Force-refresh the page.
4. It should load very quickly (a few ms). There is no "long request duration" in the logs.
5. Check the "makoney" theme. It won't have an "open live demo" button.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -45,8 +45,7 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	const siteSlugOrId = siteSlug ? siteSlug : siteId;
 	const allPatterns = useAllPatterns();
 	const { data: theme } = useThemeDetails( selectedDesign?.slug );
-	const themeDemoSiteSlug =
-		theme && theme.demo_uri.replace( /^https?:\/\//, '' ).replace( '/', '' );
+	const themeDemoSiteSlug = theme?.demo_uri?.replace( /^https?:\/\//, '' ).replace( '/', '' );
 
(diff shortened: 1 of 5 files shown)


noahtallen:
I am going to deploy this immediately, since I've tested it quite a bit. The performance issue is big enough I think it's fine to proceed without review, but I can absolutely address follow-up tasks!

tyxla (commented):
Nice find and the code looks great 👍

Issue #72038, opened 2023-01-13
Stats: Apply CSS grid to the Insights page


grzegorz-cp:
Tasks:
* [x] apply CSS grid from the Traffic page to the Insights page
* [x] add `hideNewModule` to the `Tags & categories` and `Publicize` components (they will be extended in #72133 and #72134)
* [x] place correct modules in the new grid following the below design

While modernising the Traffic page, we created a CSS grid to place each module with horizontal bars in the correct place.
The same grid has to be applied to the insights page.

Order of modules

![Image](https://user-images.githubusercontent.com/112354940/212216981-998e107d-0e32-4cc0-8610-4d8d80107d3c.png)

wilmersondasilva:
Will the `Latest post summary` be removed? I don't see it on the image you attached.

![image](https://user-images.githubusercontent.com/6406071/212399603-65ca8b42-41b3-4c2b-bbd6-ed55c8a4b0c8.png)

jsnmoon:
> * add `hideNewModule` to the `Tags & categories` and `Publicize` components (they will be extended here and here) (@grzegorz-cp)

It looks like "here and here" didn't get their respective href links 🙂

grzegorz-cp:
> Will the `Latest post summary` be removed? I don't see it on the image you attached.

`Latest post summary` will be replaced in #70852 so for the grid task we'll need to keep it in place or keep a feature flag until #70852 is implemented.

grzegorz-cp:
I applied `hideNewModule` where needed as part of #71957 because otherwise, I won't be able to enable that feature flag in production.

wilmersondasilva:
> Latest post summary will be replaced in https://github.com/Automattic/wp-calypso/issues/70852 so for the grid task we'll need to keep it in place or keep a feature flag until https://github.com/Automattic/wp-calypso/issues/70852 is implemented.


But what position should it be in for now? The same for `Publicize`, where should it be in the new grid?

grzegorz-cp:
I think that we'll have to push the `Latest post summary` down into a third row that will be deleted when the component is replaced.

`Publicize` will merge with `Followers total` into `Social followers`, so let's put them where `Social followers` will be. The remaining component could go to the third row with `Latest post summary`. Could you check if this looks correct? Did you plan to put it behind a feature flag?

wilmersondasilva:
Thanks for the answer, @grzegorz-cp! I think it will look correct. I'll implement it and let you know if something else comes up. 

[...]

Issue #72134, opened 2023-01-16
Stats: Modernize bottom modules on the Insights page


grzegorz-cp:
After the CSS grid was applied in #72038, modernise the components to use horizontal bars and combine data when needed.

# Developmnet
- [x] verify `Tags & categories`
- [x] update `Comments`
- [x] update `Shares`
- [x] update `Followers`
- [x] combine `Publicize` and `Follower totals` (skip `Social`)
- [x] verify if users that didn't install `Publicize` are affected by the changes

Depends on:
- #72038 to ass a CSS grid
- #70852 to implement `Latest post summary`. When it's done, the old `Latest post summary` should already be hidden.
- #72133 to extend horizontal bars for `Tags & categories`
- #72135 to extend the horizontal bar card and allow a toggle in the header

# Designs
![Image](https://user-images.githubusercontent.com/112354940/212594980-4efc3cac-21fd-4eb2-af14-dd2011921a2d.png)

jsnmoon:
@grzegorz-cp: Feel free to pare down the scope of this PR to meet our deadline for this Friday!

PR #72249, opened 2023-01-18
Stats: Apply CSS grid to the Insights page

wilmersondasilva:
While modernizing the Traffic page, we created a CSS grid to place each module with horizontal bars in the correct place.
The same grid has to be applied to the insights page.

#### Proposed Changes

* Create a new flag called `stats/insights-page-grid`
* Apply the CSS grid to the `Insights` page when the new flag is enabled

#### Testing Instructions

* Go to the Calypso Live link below
* Go to the `Insights` page and ensure it looks the same as before
* Apply the feature flag by adding `?flags=stats/insights-page-grid` to the URL
* Ensure the modules with horizontal bars are following what is described on #72038 (the UI of each module remains the same and it will be changed on follow-up PRs) 

#### Screenshots

| Before | After |
| ----- | ----- |
| image | image |

#### Pre-merge Checklist

Related to #72038

### DIFF
--- a/client/my-sites/stats/grid-layout.scss
+++ b/client/my-sites/stats/grid-layout.scss
@@ -3,27 +3,27 @@
 $grid-horizontal-gutters: 24px;
 $grid-vertical-gutters: 32px;
 
-@mixin stats-desktop-grid($grid-template-areas ) {
+@mixin stats-desktop-grid($grid-template-areas, $grid-template-rows) {
 	align-items: stretch;
 	display: grid;
 	grid-template-columns: repeat(12, 1fr);
-	grid-template-rows: repeat(5, min-content);
+	grid-template-rows: repeat($grid-template-rows, min-content);
 	gap: $grid-vertical-gutters $grid-horizontal-gutters;
 	grid-template-areas: $grid-template-areas;
 }
 
-@mixin stats-tablet-grid($grid-template-areas ) {
+@mixin stats-tablet-grid($grid-template-areas, $grid-template-rows) {
 	@media (max-width: $break-large ) {
 		grid-template-columns: repeat(8, 1fr);
-		grid-template-rows: repeat(6, min-content);
+		grid-template-rows: repeat($grid-template-rows, min-content);
 		grid-template-areas: $grid-template-areas;
 	}
 }
 
-@mixin stats-mobile-grid($grid-template-areas ) {
+@mixin stats-mobile-grid($grid-template-areas, $grid-template-rows) {
 	@media (max-width: $break-medium ) {
 		grid-template-columns: repeat(4, 1fr);
-		grid-template-rows: repeat(9, min-content);
+		grid-template-rows: repeat($grid-template-rows, min-content);
 		gap: 0;
 		grid-template-areas: $grid-template-areas;
 	}
@@ -34,14 +34,16 @@ $grid-vertical-gutters: 32px;
 	@include stats-desktop-grid(
 		"tags-categories tags-categories tags-categories tags-categories tags-categories tags-categories comments comments comments comments comments comments"
 		"shares shares shares shares followers followers followers followers publicize publicize publicize publicize"
-		"latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary social-followers social-followers social-followers social-followers social-followers social-followers"
+		"latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary social-followers social-followers social-followers social-followers social-followers social-followers",
+		3
 	);
 
 	@include stats-tablet-grid(
 		"tags-categories tags-categories tags-categories tags-categories comments comments comments comments"
 		"shares shares shares shares followers followers followers followers"
 		"publicize publicize publicize publicize latest-post-summary latest-post-summary latest-post-summary latest-post-summary"
-		"social-followers social-followers social-followers social-followers social-followers social-followers social-followers social-followers"
+		"social-followers social-followers social-followers social-followers social-followers social-followers social-followers social-followers",
+		4
 	);
 
 	@include stats-mobile-grid(
@@ -51,7 +53,8 @@ $grid-vertical-gutters: 32px;
 		"followers followers followers followers"
 		"publicize publicize publicize publicize"
 		"latest-post-summary latest-post-summary latest-post-summary latest-post-summary"
-		"social-followers social-followers social-followers social-followers"
+		"social-followers social-followers social-followers social-followers",
+		7
 	);
 
 	@media (max-width: 660px) {
@@ -100,7 +103,8 @@ $grid-vertical-gutters: 32px;
 		"country country country country country country country country country country country country"
 		"authors authors authors authors search search search search clicks clicks clicks clicks"
 		"videos videos videos videos videos videos downloads downloads downloads downloads downloads downloads"
-		"emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open"
+		"emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open",
+		5
 	);
 
 	@include stats-tablet-grid(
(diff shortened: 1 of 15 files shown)


kangzj (changes requested):
I followed the test instructions, but the layout doesn't look quite right to me.

It seems to me some class names are missing from the blocks, for example `list-publicize`.

image

client/my-sites/stats/stats-module/style.scss
```diff
@@ -28,6 +33,50 @@
...
+	grid-template-areas:
+		"tags-categories tags-categories tags-categories tags-categories tags-categories tags-categories comments comments comments comments comments comments"
+		"shares shares shares shares followers followers followers followers publicize publicize publicize publicize"
+		"latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary latest-post-summary social-followers social-followers social-followers social-followers social-followers social-followers";
+
+	@media (max-width: 660px) {
+		padding: 0 16px;
+	}
+
+	.list-tags-categories {
```
grzegorz-cp: Looks like this class is not applied and this breaks the grid.

client/my-sites/stats/stats-module/style.scss
```diff
@@ -28,6 +33,50 @@
 	@include clear-fix;
 }
 
+.stats__module-list.stats__module--unified.is-insights-page-enabled {
+	align-items: stretch;
```
grzegorz-cp: Those properties are shared with the ones in `grid-layout.scss`. Could you reuse as much as possible? One way would be to put shared styles into a mixin, import them in this scss and keep the specific `grid-template-areas` here. Another idea would be to create a generic grid, but I'm not sure that's possible because the countries component is quite specific to the other page.

client/my-sites/stats/stats-module/style.scss
```diff
@@ -1,3 +1,8 @@
+@import "@wordpress/base-styles/breakpoints";
+
+$grid-horizontal-gutters: 24px;
```
grzegorz-cp: Those variables could be shared between the grids if we used a mixin with a default parameter value.

grzegorz-cp (changes requested):
This is the right track, but some minor changes could improve it. I left a few comments and also noticed that this grid is missing `$break-large` and `$break-small`. Could you apply similar breakpoints as seen in [this place](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/grid-layout.scss), and we'll see how it works?

Since the file linked above is in the root directory, we could put this grid there. A mixin, even in the same file, could reduce the size and duplicated code. What do you think?

wilmersondasilva:
> It seems to me some class names are missing from the blocks, for example `list-publicize`.

I have no idea why this happened in this environment, @kangzj. It worked properly while I was developing. I'll take a look at it. Thanks for pointing it out.

client/my-sites/stats/stats-module/style.scss
wilmersondasilva: I thought of doing something like this. I was thinking about what was the better way. I think I'll implement the mixin approach. It should work fine for this.

wilmersondasilva:
> this grid is missing $break-large and $break-small.

I thought that it would be better to add this later along with the modules UI modernization. But I'll add it now. 

> A mixin, even in the same file, could reduce the size and duplicated code. What do you think?

I totally agree with you, @grzegorz-cp. I'll implement the mixin.

wilmersondasilva:
Just updated the PR with the changes suggested @grzegorz-cp! 👍

client/my-sites/stats/grid-layout.scss
```diff
@@ -3,51 +3,127 @@
 $grid-horizontal-gutters: 24px;
 $grid-vertical-gutters: 32px;
 
-.is-section-stats .stats__module-list--traffic {
+@mixin stats-desktop-grid($grid-template-areas ) {
 	align-items: stretch;
 	display: grid;
 	grid-template-columns: repeat(12, 1fr);
 	grid-template-rows: repeat(5, min-content);
```
dognose24: The repeated `grid-template-rows` count needs to be updated with the `$grid-template-areas` argument.

client/my-sites/stats/grid-layout.scss
```diff
@@ -3,51 +3,127 @@
...
-		"emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open emails-open";
+	grid-template-areas: $grid-template-areas;
+}
 
-	// Tablets
-	@media ( max-width: $break-large ) {
+@mixin stats-tablet-grid($grid-template-areas ) {
+	@media (max-width: $break-large ) {
 		grid-template-columns: repeat(8, 1fr);
 		grid-template-rows: repeat(6, min-content);
```
dognose24: The repeated `grid-template-rows` count needs to be updated with the `$grid-template-areas` argument.

client/my-sites/stats/grid-layout.scss
```diff
@@ -3,51 +3,127 @@
...
+		grid-template-areas: $grid-template-areas;
 	}
+}
 
-	// Mobile
-	@media ( max-width: $break-medium ) {
+@mixin stats-mobile-grid($grid-template-areas ) {
+	@media (max-width: $break-medium ) {
 		grid-template-columns: repeat(4, 1fr);
 		grid-template-rows: repeat(9, min-content);
```
dognose24: The repeated `grid-template-rows` count needs to be passed along with the `$grid-template-areas` argument.

client/my-sites/stats/grid-layout.scss
```diff
@@ -3,51 +3,127 @@
...
+	.list-comments {
+		grid-area: comments;
 	}
 
+	.stats__module-wrapper--shares,
+	.list-shares {
+		grid-area: shares;
+	}
+
+	.stats__module-wrapper--followers,
```
dognose24, line 79: This class name seems redundant since the `Followers` component is not made of `StatsModule`.

client/my-sites/stats/grid-layout.scss
```diff
@@ -3,51 +3,127 @@
...
+		"latest-post-summary latest-post-summary latest-post-summary latest-post-summary"
+		"social-followers social-followers social-followers social-followers"
+	);
+
+	@media (max-width: 660px) {
+		padding: 0 16px;
+	}
+
+	.stats__module-wrapper--tags-categories,
+	.list-tags-categories {
```
dognose24, line 65: This class name seems unused.

dognose24 (commented):
I noticed the redundant gap along with those shared mixins and left a few nitty comments for class names that could be refactored in the future.

client/my-sites/stats/grid-layout.scss
wilmersondasilva: It'll be refactored to be a `StatsModule` soon, so I think I'll keep the class to prevent the layout breaks when the change happens. Do you think I should remove it for now?

client/my-sites/stats/grid-layout.scss
wilmersondasilva: In the `StatsListCard` component we have this dynamic className

image

client/my-sites/stats/grid-layout.scss
wilmersondasilva: Done!

client/my-sites/stats/grid-layout.scss
wilmersondasilva: Done!

client/my-sites/stats/grid-layout.scss
wilmersondasilva: Done!

grzegorz-cp (approved):
It looks good from the code perspective. The grid is applied, which was the purpose of the ticket. Thanks for doing this.

The components don't expand to take the full height, but they are not using new horizontal bar card components (this is part of the future tickets). Until those cards are converted as part of #72134 we shouldn't enable the FF in prod.

wilmersondasilva:
Thanks for reviewing it, @grzegorz-cp! I'll merge and ship it then.

jeffgolenski:
@wilmersondasilva 👋  thanks for getting to this my friend. My environment looks as expected. Just making notes about the breakpoints. Not sure if you're tackling that in this PR though!

Screen Shot 2023-01-20 at 11 20 32 AM

PR #72254, opened 2023-01-18
Site Assembler: Render patterns from the user site and disable cache of rendered patterns

arthur791004:
#### Proposed Changes

* Referring to p1673966406507379-slack-C048CUFRGFQ, non-a8c users cannot get the settings from the theme demo site. So, I change back to get the settings from the current user site to get settings and rendered patterns. The downside is the posts of the query blocks will be from the current user site, and it might only have 1 or 2 posts. But I think we can deal with it later.
* Referring to p1673926768597069-slack-CRWCHQGUB, there is an issue related to the cache of the rendered patterns. The last page might not be fetched while the cache is expired at that time. The new user might not encounter this issue because the cache won't be expired in the onboarding flow, but it's annoyed and make developers feel weird. So, I think it'd be better to disable it.

#### Testing Instructions

* Use non-a8c account and go to /setup?siteSlug=<your_site>
* When you're in the Pattern Assembler step
  * Verify the styles of the patterns are correct
  * Verify you won't see any "blank" pattern

#### Pre-merge Checklist

Related to p1673966406507379-slack-C048CUFRGFQ, p1673926768597069-slack-CRWCHQGUB

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -362,7 +362,7 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 		</div>
 	);
 
-	if ( ! selectedDesign ) {
+	if ( ! site?.ID || ! selectedDesign ) {
 		return null;
 	}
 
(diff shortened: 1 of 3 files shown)


taipeicoder:
> non-a8c users cannot get the settings from the theme demo site

Tested and it's fixed.

> there is an issue related to the cache of the rendered patterns

Tested and I can still see it on my end. See screenshot:
Screen Shot 2023-01-18 at 2 53 01 PM

I was able to reproduce it by going back all the way to the Goals screen by editing the URL, then accessing the PA step again. I was using Firefox.

arthur791004:
> Tested and I can still see it on my end. See screenshot:

@taipeicoder Hmmm...looks like there is an issue with this implementation. Hence, I change to fetch rendered patterns by setting the page manually. Could you try it again? Thanks!

taipeicoder (approved):
Works nicely  👍

PR #72267, opened 2023-01-18
Changed the arrows behavior in Email Stats chart

### LINKED ISSUE #72229: Email Stats: with `days` period selected, we need the arrows to behave differently
## The actual behavior

The arrows in the Email Stats chart, when the period selected is `days`, select the next or the previous day. Now days selection is not needed anymore as the rest of the stats doesn't not depend on the selected date anymore.

## Desired behavior

When the `days` period is selected and the arrows are clicked, we want the chart to jump forward or backward 30 days, effectively showing the next or the previous "page" in the chart. This behavior can be seen in action at the moment selecting `hours` and clicking the arrows.

The discussion can be seen here: pdDOJh-1mZ-p2

JuanLucha:
#### Proposed Changes

The previous behavior of the left and right arrows in the Email Stats chart was to advance or move back one day. In the case of the period selected being `days` we want that to change to 30 days. That way we are moving onward or backward a "page" int he stats.

#### Testing Instructions

2. Insert manually in the browser address bar an URL with this format: https://calypso.localhost:3000/stats/email/open/[the-domain-of-your-blog]/day/[post-id]?flags=newsletter/stats. Note that you can change day for week and month.
3. Click on the right arrow. You should see the new chart data to begin 30 days after your initial date. For example, if the first column of your chart is 9th Jan and you click the right arrow, you should see a chart with 8th Feb as the first one. 
4. Click on the left arrow and check that you are in the initial date again (in my example, the 9th Jan again). 
5. Click one more time in left arrow and check that you should be 30 days before the initial date (in my case 10th Dec).
6. Go back and forth and check that everything is ok.
7. Go to /stats/day/[blog-domain] and check that the arrows still works as intended: moving one day only.

Fixes https://github.com/Automattic/wp-calypso/issues/72229

### DIFF
--- a/client/components/chart/index.jsx
+++ b/client/components/chart/index.jsx
@@ -1,7 +1,7 @@
 import classNames from 'classnames';
 import { localize, withRtl } from 'i18n-calypso';
 import PropTypes from 'prop-types';
-import { useState, useCallback, useMemo } from 'react';
+import { useState, useCallback, useMemo, useEffect } from 'react';
 import Notice from 'calypso/components/notice';
 import Tooltip from 'calypso/components/tooltip';
 import { hasTouch } from 'calypso/lib/touch-detect';
@@ -55,6 +55,7 @@ function Chart( {
 	translate,
 	chartXPadding,
 	sliceFromBeginning,
+	onChangeMaxBars,
 } ) {
 	const [ tooltip, setTooltip ] = useState( { isTooltipVisible: false } );
 	const [ sizing, setSizing ] = useState( { clientWidth: 650, hasResized: false } );
@@ -107,6 +108,13 @@ function Chart( {
 
 	const width = isTouch && sizing.clientWidth <= 0 ? 350 : sizing.clientWidth - chartXPadding; // mobile safari bug with zero width
 	const maxBars = Math.floor( width / minWidth );
+
+	useEffect( () => {
+		if ( onChangeMaxBars ) {
+			onChangeMaxBars( maxBars );
+		}
+	}, [ maxBars, onChangeMaxBars ] );
+
 	// Memoize data calculations to avoid performing them too often.
 	const { chartData, isEmptyChart, yMax } = useMemo( () => {
 		const nextData = ( () => {
--- a/client/my-sites/stats/stats-email-open-detail/index.jsx
+++ b/client/my-sites/stats/stats-email-open-detail/index.jsx
@@ -110,6 +110,7 @@ class StatsEmailOpenDetail extends Component {
 		showPreview: false,
 		activeTab: null,
 		activeLegend: null,
+		maxBars: 0,
 	};
 
 	static getDerivedStateFromProps( props, state ) {
@@ -173,6 +174,8 @@ class StatsEmailOpenDetail extends Component {
 
 	onChangeLegend = ( activeLegend ) => this.setState( { activeLegend } );
 
+	onChangeMaxBars = ( maxBars ) => this.setState( { maxBars } );
+
 	switchChart = ( tab ) => {
 		if ( ! tab.loading && tab.attr !== this.props.chartTab ) {
 			this.props.recordGoogleEvent( 'Stats', 'Clicked ' + titlecase( tab.attr ) + ' Tab' );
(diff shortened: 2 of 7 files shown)


TimBroddin:
Hey Juan,

I think instead of 30 days, this chart should jump the amount of bars shown.

This is my initial view. Last day is January 22nd.

CleanShot 2023-01-18 at 14 19 58@2x

When I click right the chart jumps to January 27th.

CleanShot 2023-01-18 at 14 21 54@2x

I know this is probably a tough one to fix, since the number of bars shown is decided in the chart component.

client/components/chart/index.jsx
```diff
@@ -107,8 +108,12 @@ function Chart( {
 
 	const width = isTouch && sizing.clientWidth <= 0 ? 350 : sizing.clientWidth - chartXPadding; // mobile safari bug with zero width
 	const maxBars = Math.floor( width / minWidth );
+
 	// Memoize data calculations to avoid performing them too often.
 	const { chartData, isEmptyChart, yMax } = useMemo( () => {
+		// We communicate the new value for maximum bars in the chart to the parent component
+		onChangeMaxBars( maxBars );
```
TimBroddin: Very nice solution!

client/my-sites/stats/stats-email-open-detail/index.jsx
```diff
@@ -335,6 +340,8 @@ const connectComponent = connect(
 			hasValidDate,
 		} = getPeriodWithFallback( ownProps.period, ownProps.date, isValidStartDate, post?.date );
 
+		const maxBars = 0;
```
TimBroddin: Why is this?

client/components/chart/index.jsx
TimBroddin: You might want to check if it's defined, though. And also pass it to the deps array (to keep ESLint from complaining).

JuanLucha: Indeed 👍

client/my-sites/stats/stats-email-open-detail/index.jsx
JuanLucha: Initializing the state. It's part of the state that is passed to the component in line 363.

client/my-sites/stats/stats-email-open-detail/index.jsx
```diff
@@ -215,6 +219,8 @@ class StatsEmailOpenDetail extends Component {
 						date={ query.date }
 						period={ query.period }
 						hasValidDate={ hasValidDate }
+						quantity={ maxBars }
```
TimBroddin, line 222: Nice one! Passing the quantity here as well.

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -62,9 +67,9 @@ class StatsPeriodNavigation extends PureComponent {
 	};
 
 	handleArrowPrevious = () => {
-		const { date, moment, period, url, queryParams } = this.props;
-		const usedPeriod = 'hour' === period ? 'day' : period;
-		const previousDay = moment( date ).subtract( 1, usedPeriod ).format( 'YYYY-MM-DD' );
+		const { date, moment, period, url, queryParams, isEmailStats } = this.props;
+		const numberOfDAys = isEmailStats ? 30 : 1;
```
TimBroddin: The back button still goes back 30 days 🤪

TimBroddin (changes requested):
Nice work! I've left one remark for the back button.

client/my-sites/stats/stats-period-navigation/index.jsx
```diff
@@ -50,10 +50,18 @@ class StatsPeriodNavigation extends PureComponent {
 		}
 	};
 
+	isHoursPeriod = ( period ) => 'hour' === period;
+
+	getNumberOfDays = ( isEmailStats, period, maxBars ) =>
+		isEmailStats && ! this.isHoursPeriod( period ) ? maxBars : 1;
```
TimBroddin, line 56: I don't think it's necesarry to pass `maxBars `around, since it's already in `this.props`. We might want to refactor this later.

TimBroddin (approved):

client/my-sites/stats/stats-period-navigation/index.jsx
JuanLucha: I like pure functions 😀 But if it's hurting your eyes, I'll change it. But... I really think the function is not going to be as pretty if it has to grab the value of maxBars from the props itself.

TimBroddin:
I found a small regression:

CleanShot 2023-01-20 at 13 40 13@2x

On load, it triggers a request with` quantity=0` (when maxBars = 0). This is ignored by the API, and it returns its default (=30). We might want to make a follow-up PR that addresses this.

PR #72286, opened 2023-01-18
Themes Showcase: Replace empty search card with more theme options

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1453

#### Proposed Changes

Replaces the previous upsell nudge displayed on empty searches, with a new card that offer users to perform the following actions:
- Design your own theme / homepage: Allow users to change their own homepage in the site editor (only in the logged-in showcase, and only if the site has a block theme active).
- Do it for me: Upsells the DIFM service.
- Upload a theme: Upsells the Business plan and/or nudges users to upload a third-party/custom theme missing.

**Screenshots**
<table>
<thead><tr><th>Conditions</th><th>Before</th><th>After</th></tr></thead>
<tbody>
<tr>
<td>Showcase: Logged-in<br><br>Plan: Free<br><br>Current theme: Block theme</td>
<td>
Screenshot 2023-02-02 at 14 49 25
</td>
<td>
Screenshot 2023-02-03 at 11 15 23
</td>
</tr>
<tr>
<td>Showcase: Logged-in<br><br>Plan: Business<br><br>Current theme: Block theme</td>
<td>
Screenshot 2023-02-02 at 14 52 14
</td>
<td>
Screenshot 2023-02-03 at 11 16 00
</td>
</tr>
<tr>
<td>Showcase: Logged-in<br><br>Plan: Business<br><br>Current theme: Legacy theme</td>
<td>
Screenshot 2023-02-02 at 14 52 14
</td>
<td>
Screenshot 2023-02-03 at 11 28 35
</td>
</tr>
<tr>
<td>Showcase: Logged-out</td>
<td>
Screenshot 2023-02-02 at 14 56 13
</td>
<td>
Screenshot 2023-02-02 at 14 56 59
</td>
</tr>
</tbody>
</table>

**Events**

The following events a tracked when a user clicks on any of the options:
- Design your own theme / homepage: 
  - Event: `calypso_themeshowcase_more_options_design_homepage_click`
  - Props:
    - `site_plan`: The site plan.
    - `search_term`: The term searched.
    - `destination`: Where we are sending the user to. The only possible value right now is 'site-editor', but it can be extended in the future once we can start the Pattern Assembler.
- Do it for me:
  - Event: `calypso_themeshowcase_more_options_difm_click`
  - Props:
    - `site_plan`: The site plan.
    - `search_term`: The term searched.
- Upload a theme:
  - Event: `calypso_themeshowcase_more_options_upload_theme_click`
  - Props:
    - `site_plan`: The site plan.
    - `search_term`: The term searched.
    - `destination`: Where we are sending the user to. Possible values are `signup` (for the logged-out showcase), `upload-theme` if the site has an eligible plan, or `checkout` if the site doesn't have an eligible plan.

#### Testing Instructions

- Switch to a Free site with a block theme active
- Go to Appearance > Themes
- Enter a term that doesn't produce any result
- Observe the new card
- Make sure it includes 3 options: Design your own homepage, Do it for me, Upload theme
- Make sure it looks good and the copy is clear
- Make sure all actions redirect to the appropriate place
- Make sure it is responsive on mobile viewports
- Switch to a Simple site with a Business plan and a block theme active
- Make sure the "Upload theme" action no longer upsells the Business plan
- Make sure the "Upload theme" action sends users to `/theme/upload/:site`
- Activate a legacy theme
- Make sure the "Design your own homepage" action is no longer visible
- Make sure the "Upload theme" action sends users to `/wp-admin/theme-install.php`
- Switch to an Atomic site with a Business plan
- Open the `/themes` path on an incognito window
- Enter a term that doesn't produce any result
- Make sure it includes 2 options: Do it for me, Upload theme
- Make sure the "Upload theme" action initiates the Business plan signup flow


BogdanUngureanu:
Nice work, Miguel! I've tested it for both logged-in and logged-out users and I noticed some small issues.

**Logged in users, on an Atomic site, the upload theme button is not working.**
Steps to reproduce:
- Go to WordPress.com/themes/your-atomic-site
- Search for "foo"
- Go to the new section and click on the "Upload theme" button
- Notice that nothing happens, you aren't redirected to the upload screen

**Logged out users: The previous upgrade nudge is no longer displayed**
- Open an incognito window and go to /themes
- Search for a random string that doesn't have any results
- Notice that we no longer display any upgrade CTA banner

Is this on purpose? :)

BogdanUngureanu:
Two other remarks:
- The previous CTA banner had a tracking event. Maybe we should add tracking events for each of these buttons? Since on mobile we have a different button for the pattern assembler, maybe we should pass a is_mobile parameter (or can we determine it based on the global variables) to differentiate between them?
-  Since on mobile we have a fallback to the Site Editor - which IMHO doesn't bring much value in the context of theme selection since the user is interested in selecting a theme for their site, maybe we should re-order the buttons? Would it make more sense to display the `Upload a theme` first?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -61,9 +62,18 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
...
+				slug: 'blank-canvas-3',
+				title: 'Blank Canvas',
+				recipe: {
+					stylesheet: 'pub/blank-canvas-3',
+				},
+				verticalizable: false,
+				is_premium: false,
+				is_bundled_with_woo_commerce: false,
+				design_type: 'assembler',
+			} as Design );
```
mmtr: @Automattic/lego this might conflict with your work done in https://github.com/Automattic/wp-calypso/pull/71852. This was needed to allow users coming from the themes showcase to start the pattern assembler flow while skipping the rest of the site setup flow. Do you see any issues with this approach?

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -479,6 +479,13 @@ const siteSetupFlow: Flow = {
 					return navigate( 'vertical' );
 
 				case 'patternAssembler':
+					// eslint-disable-next-line no-case-declarations
+					const backTo = urlQueryParams.get( 'backTo' );
+
+					if ( backTo ) {
+						return redirect( backTo );
+					}
+
```
mmtr: @Automattic/lego same here, this might also conflict with your work done in https://github.com/Automattic/wp-calypso/pull/71852. This was needed to allow users to go back to the themes showcase from the pattern assembler flow if they click the back button. Do you see any issues with this approach?

mmtr:
> Logged in users, on an Atomic site, the upload theme button is not working.

Nice catch! Fixed in https://github.com/Automattic/wp-calypso/pull/72286/commits/901f84bd73aa04fa5ffb9ce3d3550d0e99c3a6c9

> Logged out users: The previous upgrade nudge is no longer displayed

Yeah, this is in purpose because there isn't a equivalent nudge / action in the new footer, so I opted for removing it for logged out users. 

I'm open to bring it back or to add and alternative nudge / action in the new footer, but it probably needs a discussion with @Copons, @ianstewart or @SaxonF. 

> The previous CTA banner had a tracking event. Maybe we should add tracking events for each of these buttons?

Yup, I also noted that, and ultimately considered that we can make that in a separate PR since we cannot keep using the existing tracking event from the previous banner. But happy to handle it now if folks think otherwise.

> Since on mobile we have a fallback to the Site Editor - which IMHO doesn't bring much value in the context of theme selection since the user is interested in selecting a theme for their site, maybe we should re-order the buttons? Would it make more sense to display the Upload a theme first?

While I agree that the site editor fallback is not great, I'm not sure that the upload theme is a better option for mobile users, since they likely won't have a .zip file at hand.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
miksansegundo: Thanks for the ping, Miguel!

We added that redirection to the `goals` step to ensure users _creating a new site_ don't skip the goals step if they access the pattern assembler URL directly. Now that there are other use cases like yours to access the assembler step directly on existing sites, we'll have to add an exception to that check and create a new flow with only the assembler.

For now, rather than remove the redirection you could use a URL param to identify the new use case and skip the redirection. In the future, we could use the new assembler flow name to skip the redirection. I'll keep you updated.

client/landing/stepper/declarative-flow/site-setup-flow.ts
miksansegundo: I don't see any issues. Thanks for sharing it! We could use it too.

dsas:
How much use does the existing 'no results' footer see on the logged out showcase? 🤔 . My intuition is that this is probably a low priority problem that doesn't have to block this PR and we can file in a separate issue and prioritise against the rest of the backlog.

----

Once we will include dotorg results in the search, zero search results will make the existing footer nonsensical - no matching dotorg themes will exist so upgrading won't help. Definitely the right decision to remove it.

On the new footer actions:
 * We can't link to the pattern assembler without having a site. Maybe we could add a url param that onboarding uses to highlight / pre-select pattern assembler.
 * We could link to DIFM
 * We could mention that you can upload your own theme you get from a 3rd party marketplace / own designer by buying the business plan, linking to https://wordpress.com/start/business/ 

Perhaps for now we could replace the pattern assembler nudge with a reminder that you can find a "close enough" theme and extensively customise it in the site-editor / customizer 🤔 

Maybe it would be interesting to link to a logged out version so people can have a play with it, but that would be a whole different project and Marvel has higher priorities.

BogdanUngureanu:
Since it drove conversions, I've re-added the Upgrade banner for logged-out themes page so it's RfR again. p1674553640183219-slack-C04DZ8M0GHW

@dsas I think we should leave it as it is for now.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
bangank36: > URL param to identify the new use case

I think this would be a better idea for now

client/components/themes-list/index.jsx
```diff
@@ -130,70 +139,148 @@ function ThemeBlock( props ) {
...
-				<>
-					<div className="themes-list__not-found-text">
-						{ translate( 'No themes match your search' ) }
+	return (
+		<div className="themes-list__footer">
+			<div className="themes-list__footer-heading">
+				{ translate( "Can't find what you're looking for?" ) }
+			</div>
+			<div className="themes-list__footer-subheading">
+				{ translate( 'Here are a few more options' ) }
```
dsas: Non-blocking nitpick. Should this have a colon? 🤔

client/components/themes-list/index.jsx
```diff
@@ -130,70 +139,148 @@ function ThemeBlock( props ) {
...
+							: translate( 'Open the editor' ) }
+					</Button>
+				</div>
+			</div>
+			<div className="themes-list__footer-action">
+				<Icon className="themes-list__footer-action-icon" icon={ brush } size={ 28 } />
+				<div className="themes-list__footer-action-content">
+					<div className="themes-list__footer-action-text">
+						<div className="themes-list__footer-action-title">
+							{ translate( 'Hire our team of experts to design one for you' ) }
```
dsas: All of the other titles explicitly say 'theme': "Create your own theme from scratch", "Upload your theme". Should this say "Hire our team of experts to design a theme for you"?

dsas: Actually maybe this is intentionally vague, lets leave it.

DIFM doesn't design a theme for you. You have to choose a theme from the even more limited list offered, or let us choose one of them for you. DIFM will then make custom page layouts for you.

client/components/themes-list/index.jsx
mmtr: Yup, I wondered the same when I saw the proposed copy in https://github.com/Automattic/dotcom-forge/issues/1453. I think a colon it's more consistent with the rest of our writing style, so I'll add it.

client/components/themes-list/index.jsx
mmtr: > Actually maybe this is intentionally vague, lets leave it.

You made a good point about being vague though, since translators for example might have trouble understanding what "one" means in this context, so we can probably add a translator comment to clarify.

Copons: +1 for being vague and +1 for adding a translator comment! 👍

client/components/themes-list/index.jsx
Copons: That's not a colon though! 😛

mmtr: Oh gosh, I was completely lost in translation and genuinely thought for a moment that a colon was the same than a dot 🤦

autumnfjeld:
See comment here on path names:  https://github.com/Automattic/wp-calypso/pull/72747#discussion_r1090458726.

So we stay in sync on best path name.  Welcome input. :)

autumnfjeld:
To be clear, this flow is a theme switch and we have decided that it is ok to NOT give a warning about what will happen with user's existing site and content?  (Let me know if I am misunderstanding.)  cc: @Automattic/lego  @ianstewart 


https://user-images.githubusercontent.com/5665959/215635026-c3996896-c160-4ff1-b4eb-1da07173479a.mp4

autumnfjeld:
One test:
Starting site:
image


What i select in PA


image


Landing in the Site Editor
image


This test worked: Classic -> BCPA

A previous test I did had wonky results, what I saw in the site editor didn't match up to what I selected in the PA.  Just an indication of thorough theme switch testing needed.

Copons:
As discussed internally (p1675128979469429-slack-C048CUFRGFQ), the PA flow as a theme switch is not ready yet, and we should remove it from these site actions for the time being.

cc @mmtr 

I think it's not a big deal, design-wise. The site actions work regardless. Maybe we should consider if moving the primary to another action, or just keep all of them secondary for now.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
mmtr: This has been already addressed in https://github.com/Automattic/wp-calypso/pull/72747, so I'll revert these changes.

mmtr:
Alright, I added the changes below, so this should be ready for review again:

- The footer is now visible in the logged-out showcase as well, effectively removing the upgrade nudge that @BogdanUngureanu re-added in https://github.com/Automattic/wp-calypso/pull/72286/commits/486508b388a5b717faf6eee508ce2472165d5819.
- The "Create your own theme" action no longer start the pattern assembler flow. Instead, it redirects logged-in users to the site editor (since the pattern assembler is still not ready to be used within the logged-in showcase), while logged-out users are redirected to the `with-theme` signup flow to create a new site with the Blank Canvas theme (which includes the Pattern Assembler step if the `pattern-assembler/logged-out-showcase` feature flag is enabled).
- The "Upload your theme" action redirects logged-in users to either the checkout (with the Business plan automatically added to the cart) or to the install theme page if the site already has the Business plan. Logged-out users are redirected to the signup flow that creates a new Business site.

BogdanUngureanu:
Works great for me! 🎉 One thing I noticed though (non-blocking): the `Get started` button (the third one) is slightly smaller than the other ones. Should we make it with the the same width?

mmtr:
> the `Get started` button (the third one) is slightly smaller than the other ones. Should we make it with the the same width?

Yeah, the width is adapted to its content, as seen in the original design available in https://github.com/Automattic/dotcom-forge/issues/1453, so I think we can leave it as-is.

mmtr:
I just realized the new footer added here conflicts with the new Pattern Assembler CTA introduced in https://github.com/Automattic/wp-calypso/pull/72752, so I asked in https://github.com/Automattic/dotcom-forge/issues/1453#issuecomment-1411951568 if we should drop the CTA in favor of the new footer.

autumnfjeld:
> drop the CTA in favor of the new footer.

No.  😄    Per discussion p1675291431107489/1675248776.600559-slack-C048CUFRGFQ

Thanks guys.

Copons:
> > drop the CTA in favor of the new footer.
> 
> No. 😄 Per discussion p1675291431107489/1675248776.600559-slack-C048CUFRGFQ
> 
> Thanks guys.

The big Lego vs T-Rex fight IS ON!

mmtr:
Alright, I just pushed an update that hides the new multi-options card when the Pattern Assembler CTA is visible. I also removed any option from the card that was starting the Pattern Assembler.

mmtr:
As per feedback in p1675348333564989/1675248776.600559-slack-C048CUFRGFQ, I pushed another update to exclude the new multi-options card from searches with results, since the removal of the pattern assembler entry point doesn't make it worthwhile.

I also added a small fix to prevent upsell banners from appearing if the card shows up.

client/components/themes-list/index.jsx
```diff
@@ -148,70 +159,165 @@ function ThemeBlock( props ) {
 	);
 }
 
-function Empty( props ) {
-	const { wpOrgThemes, emptyContent, searchTerm, upsellCardDisplayed, translate } = props;
+function Options( { isFSEActive, recordTracksEvent, searchTerm, translate, upsellCardDisplayed } ) {
```
Copons, line 162: This is a pretty big component; what do you think of moving it to its own file?

Copons: We might even simplify the props by importing them into the end-consumer component rather than dragging them across multiple depth levels, but let's not overcomplicate this PR any further.

client/components/themes-list/index.jsx
```diff
@@ -148,70 +159,165 @@ function ThemeBlock( props ) {
...
+		// This should also start the Pattern Assembler, which is currently in development for
+		// the logged-out showcase. Since there isn't any proper fallback for the meantime, we
+		// just don't include this option.
+	}
+
+	// Do it for me.
+	options.push( {
+		title: translate( 'Hire our team of experts to design one for you', {
+			comment:
+				'"One" means a theme in this context (i.e. "Hire our team of experts to design a theme for you")',
```
Copons: IIRC, we chose to keep "one" vague as it can mean both "theme" and "site".
Probably not a big deal for translators, but we might want to mention this in the comment.
E.g.
```suggestion
				'"One" could mean "theme" or "site" in this context (i.e. "Hire our team of experts to design a theme for you")',
```

dsas:
~~I'm still getting a 'Design your own' when I have empty search results. I'm on a simple site with a business plan~~
image

edit: ah wait, that's linking to the editor not to the bcpa, sorry!

Copons (approved):
This LGTM.

The only feedback I have is fundamentally about past code decisions; we can ignore them in this PR.

I also have one request: let's record all the new events somewhere!
Even just adding them to the PR description is fine (I can take care of reporting them to the PT).

And, of course, don't forget to register them once they have been used! 🙂

dsas (approved):

mmtr:
> I also have one request: let's record all the new events somewhere!
> Even just adding them to the PR description is fine (I can take care of reporting them to the PT).

Done! I've updated the PR description with them.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7793677

Thank you @mmtr for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7793677).

Issue #72169, opened 2023-01-16
WooExpress Flow: Making loading/progress indicators consistent.


markbiek:
The flow currently uses a mix of the loading ellipses and the blue progress bar.

We want to standardize on the purple progress bar.

![image](https://user-images.githubusercontent.com/917632/212746039-2d839376-76a4-4097-bf50-4a9aceecc963.png)

PR #72294, opened 2023-01-18
WooExpress Trial: Extract loading bar into a separate component

sixhours:
#### Proposed Changes

* Extract loading bar from processing step into its own component so we can use it on multiple steps
* For consistency throughout the flow, always show the LoadingBar for WooExpress steps
* Adjust flow `progress` state to account for the whole flow
* The progress bar doesn't currently alter the actual progress state so it doesn't save between step changes. I'd suggest we tackle this in a separate PR since this one is already pretty hefty.

**Demo**

https://user-images.githubusercontent.com/2124984/213499828-9228704b-efa5-4ca5-8177-3157690dd133.mov

#### Testing Instructions

* Switch to this PR, navigate to `/setup/wooexpress`
* The same purple progress bar should be seen under the main title throughout the flow
* Other `setup` flows should stay the same as production; the two I'm aware of are `/start` and `/setup/link-in-bio`, but there may be more

#### Pre-merge Checklist

- NA Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- NA Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- NA For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #72169

### DIFF
--- a/client/components/loading-bar/index.tsx
+++ b/client/components/loading-bar/index.tsx
@@ -15,7 +15,7 @@ export function LoadingBar( { className, progress }: LoadingBar ) {
 		if ( progress >= 0 ) {
 			timeoutReference = setTimeout( () => {
 				if ( progress > simulatedProgress || progress === 1 ) {
-					setSimulatedProgress( () => progress );
+					setSimulatedProgress( progress );
 				} else if ( simulatedProgress < 1 ) {
 					setSimulatedProgress( ( previousProgress: number ) => {
 						let newProgress = previousProgress + Math.random() * 0.04;
--- a/client/components/loading-bar/index.tsx
+++ b/client/components/loading-bar/index.tsx
@@ -1,0 +1,47 @@
+import { useEffect, useState } from 'react';
+import './style.scss';
+
+type LoadingBar = {
+	className?: string;
+	progress: number;
+};
+
+export function LoadingBar( { className, progress }: LoadingBar ) {
+	// Progress smoothing, works out to be around 40seconds unless step polling dictates otherwise
+	const [ simulatedProgress, setSimulatedProgress ] = useState( progress );
+
+	useEffect( () => {
+		let timeoutReference: NodeJS.Timeout;
+		if ( progress >= 0 ) {
+			timeoutReference = setTimeout( () => {
+				if ( progress > simulatedProgress || progress === 1 ) {
+					setSimulatedProgress( () => progress );
+				} else if ( simulatedProgress < 1 ) {
+					setSimulatedProgress( ( previousProgress: number ) => {
+						let newProgress = previousProgress + Math.random() * 0.04;
+						// Stall at 95%, allow complete to finish up
+						if ( newProgress >= 0.95 ) {
+							newProgress = 0.95;
+						}
+						return newProgress;
+					} );
+				}
+			}, 1000 );
+		}
+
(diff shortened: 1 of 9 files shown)


markbiek:
Looks like the styles are off on the plugin-bundle flow.

image

You can't go directly to that screen anymore (since #70478) so you'll need to go through `/start`, pick the "Sell" intent, and then select a WooCommerce theme.

daledupreez:
> Adjust flow progress state to account for the whole flow

At a high level, this concerns me a little bit, as it looks like we could easily end up with situations where the progress is auto-incremented as well as manually updated, so we could end up with the non-automatic `setProgress()` calls possibly reducing or resetting the progress bar. The conflation of whole-flow progress and stepwise progress feels confusing to me, but that may also be because I am not getting some key detail.

sixhours:
> At a high level, this concerns me a little bit, as it looks like we could easily end up with situations where the progress is auto-incremented as well as manually updated, so we could end up with the non-automatic setProgress() calls possibly reducing or resetting the progress bar. The conflation of whole-flow progress and stepwise progress feels confusing to me, but that may also be because I am not getting some key detail.

Yeah, I completely understand that concern. That's effectively what's happening right now, because the progress bar isn't showing real-time updates to `progress`, it's showing a simulated animation based on a static `progress` value we pass to it at runtime. So when we change steps in the flow, that simulated "progress" gets wiped out/reset. This particular loading bar was designed to show individual step progress, not flow progress.

My thought was to refactor it to reflect flow progress for the purposes of the Woo flow, and step progress for everything else. But we also have a flow-level progress indicator at the top of the screen, so we could re-introduce that for Woo, and use the loading bar as a step-level progress indicator.

LMK if that makes sense!

sixhours:
> Looks like the styles are off on the plugin-bundle flow.

Aha, I think `isHorizontalLayout` was set to `true` for the processing step. Since it's a centered-text step, I think that should be false. That seems to fix it, at least. :D

daledupreez:
I am still a bit confused as to how we decide which mode we're in.

Might it make sense to specify both a starting progress value and a ceiling for a specific step? That way the loading bar knows what the total range it should work on is, and it's much clearer as to what kind of loading state is being displayed?

I am thinking of something like the following:
```
interface LoadingBarProps {
  progressFloor?: number,
  progressCeiling?: number,
};

const LoadingBar = ( { progressStart =  0, progressCeiling = 1 }: LoadingBarProps )  => {
  ...
}
```

sixhours:
Dale and I chatted about this to clear up confusion. This PR will handle extracting the loading bar into a separate component; I'll work on making it state-aware with progress minimums and maximums in a future PR.

markbiek (approved):
The WooExpress flow looks good as do `/start`, `/setup/link-in-bio`, the plugin-bundle flow, and `/setup/newsletter`

PR #72300, opened 2023-01-18
E2E: add support to interact with editors with Gutenberg 14.9.1 inner iframes.

worldomonation:
#### Proposed Changes

This PR adds support for interacting with editors that have the inner iframes, introduced with Gutenberg 14.9.1.

Previous PR: https://github.com/Automattic/wp-calypso/pull/72112
Context: p1673294397010529-slack-CBTN58FTJ and p1673397395255099-slack-C1A1EKDGQ

In short:
Gutenberg 14.9.1 introduces a nested iframe at the editor canvas level if a block-based theme is used. 
If an iframe wraps the editor canvas the existing POMs are no longer able to interact with editor elements (eg. blocks).

In this PR, by using a new optional parameter to tell the EditorPage about the usage of block-based themes we can add support for both non-block-based and block-based themes. 

Key changes:
- add optional parameter `blockTheme` which defaults to `false`.
- add new class property `editorCanvas` that refers to jus the editor canvas. 

#### Testing Instructions

  - [ ] Gutenberg E2E (desktop)
  - [x] Gutenberg E2E (mobile)
  - [x] Calypso E2E (desktop)
  - [x] Calypso E2E (mobile)
  - [x] Pre-Release E2E
  - [x] i18n E2E

Additionally, change the theme for a user used in the Gutenberg suite of tests to a block-based theme and ensure the spec continues to pass.

#### Pre-merge Checklist

Related to #

### DIFF
--- a/packages/calypso-e2e/src/lib/pages/editor-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/editor-page.ts
@@ -94,10 +94,8 @@ export class EditorPage {
 		// If the parameter `blockTheme` is true, the editor canvas is hidden inside a new
 		// iframe and must be pierced to be visible.
 		if ( blockTheme ) {
-			console.log( 'klwejr' );
 			this.editorCanvas = this.editor.frameLocator( selectors.editorCanvasFrame ).locator( 'body' );
 		} else {
-			console.log( 'wkljre' );
 			this.editorCanvas = this.editor;
 		}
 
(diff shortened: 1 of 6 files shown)


packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -65,37 +68,65 @@ export class EditorPage {
...
+		// iframe and must be pierced to be visible.
+		if ( blockTheme ) {
+			this.editorCanvas = this.editorWindow
+				.frameLocator( selectors.editorCanvasFrame )
+				.locator( 'body' );
 		} else {
-			// For Simple editors, the editor is located within an iFrame
-			// and thus it must first be extracted.
-			this.editor = page.frameLocator( selectors.editorFrame ).locator( selectors.editor );
+			this.editorCanvas = this.editorWindow;
```
worldomonation, line 101: This is the core of the PR.

I explored several ways for `EditorPage` to perform this task without any intervention from the user, not even with parameters. Such attempts were not successful because the constructor is synchronous and thus we do not have insight into whether the editor features an inner iframe or not.

Alternatives I considered were to force the spec writer to call a series of initialization methods, but I decided against that because of the issues it would cause when the spec writers forget to call such actions.

```
const editorPage = new EditorPage();
// What if the spec writer forgets the next line?
// Or what if `EditorPage`changes? It would require changing this across all specs.
await editorPage.initialize();
```

One path I had not explored yet is to have a private async method that is called when anything needs to interact with the editor canvas. The method can silently look for an inner iframe and update the class property if it finds the frame.

test/e2e/specs/editor-tracking/editor-tracking__pattern-events.ts
```diff
@@ -73,26 +72,25 @@ describe( DataHelper.createSuiteTitle( 'Editor tracking: Pattern-related events'
...
+			const patternName = 'About Me Card';
+			const patternNameInEventProperty = 'a8c/about-me-card';
 
 			it( 'Clear event stack for clean slate', async function () {
 				await eventManager.clearEvents();
 			} );
 
 			it( 'Add pattern from inline inserter', async function () {
-				const openInlineInserter: OpenInlineInserter = async ( editor ) => {
-					const editorGutenbergComponent = new EditorGutenbergComponent( page, editor );
```
worldomonation, line 85: With the changes proposed in this PR, the `EditorGutenbergComponent` POM now has two parameters instead of one. 

Instead of updating the initialization of `EditorGutenbergComponent` in this line to accept two parameters, I felt that a minor refactor to the inline block inserter was preferable.

According to p1673990937312019-slack-CQD1HH4MA, the inner function and "raw" `EditorGutenbergComponent` is used because there are many inline inserters (at the editor canvas top-level and nested within blocks) and there needs to be a way to specify which inline inserter to use.

Instead, I chose to refactor the inline pattern inserter to accept the Locator of the inline inserter itself. The `addPatternInline` accepts a locator to the intended inline block inserter instead of a function, and the rest is the same.

test/e2e/specs/onboarding/ftme__write.ts
```diff
@@ -107,7 +107,7 @@ describe( DataHelper.createSuiteTitle( 'FTME: Write' ), function () {
 		} );
 
 		it( 'Editor loads', async function () {
-			editorPage = new EditorPage( page, { target: 'simple' } );
+			editorPage = new EditorPage( page, { target: 'simple', blockTheme: true } );
```
worldomonation, line 110: Because all new sites use block-based themes, we must set this flag to `true`.

dpasque (approved):
Failures are unrelated, and code changes LGTM -- thanks for making this update! 😄

PR #72308, opened 2023-01-19
Update jsdoc eslint plugin

noahtallen:
#### Proposed Changes
Separating this from #72245 to simplify things. Updates eslint-plugin-jsdoc across the repo to properly support Node 18. cc @anomiex 

The initial update introduced two problems:
1. Firstly, JSDoc wanted to replace `Object` with `object`. 
2. Secondly, there were a lot of issues with require-returns check. It seemed to have gotten a lot smarter :)

The first fix is easy -- just add a setting which allows both Object and object. I don't see a big reason to enforce one over the other at this point, especially since we'd rather use Typescript.

The second one just meant going through several files and fixing the return type. Almost every error was caused by functions like this:

```js
/**
 * @returns {string}
 */
function foo( x ) {
  if ( x ) {
    return "bar";
  }
}
```

This is obviously problematic, because the real return type is `string|undefined`. So I just made that adjustment in several files.

There are a few more issues reported by the main linter build below, but not enough to fail the build. I did find weird behavior in the eslint jsdoc lib, which I reported here: https://github.com/gajus/eslint-plugin-jsdoc/issues/949

#### Testing Instructions
CI, and verify that changes make sense. The full eslint build is running here: https://teamcity.a8c.com/buildConfiguration/calypso_CheckCodeStyle/9329626

### DIFF
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -109,7 +109,7 @@ function waitForPreferredEditorView( context ) {
  *
  * @param {object} context  Shared context in the route.
  * @param {Function} next   Next registered callback for the route.
- * @returns {*}             Whatever the next callback returns.
+ * @returns {undefined}             Whatever the next callback returns.
  */
 export const authenticate = ( context, next ) => {
 	const state = context.store.getState();
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -107,7 +107,7 @@ function waitForPreferredEditorView( context ) {
  * tracking), so we redirect the user to the WP Admin login page in order to store the auth cookie. Users will be
  * redirected back to Calypso when they are authenticated in WP Admin.
  *
- * @param {Object} context Shared context in the route.
+ * @param {object} context Shared context in the route.
  * @param {Function} next  Next registered callback for the route.
  * @returns {*|undefined}  Whatever the next callback returns.
  */
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -107,9 +107,9 @@ function waitForPreferredEditorView( context ) {
  * tracking), so we redirect the user to the WP Admin login page in order to store the auth cookie. Users will be
  * redirected back to Calypso when they are authenticated in WP Admin.
  *
- * @param {object} context  Shared context in the route.
- * @param {Function} next   Next registered callback for the route.
- * @returns {undefined}             Whatever the next callback returns.
+ * @param {Object} context Shared context in the route.
+ * @param {Function} next  Next registered callback for the route.
+ * @returns {*|undefined}  Whatever the next callback returns.
  */
 export const authenticate = ( context, next ) => {
 	const state = context.store.getState();
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -109,7 +109,7 @@ function waitForPreferredEditorView( context ) {
  *
  * @param {object} context Shared context in the route.
  * @param {Function} next  Next registered callback for the route.
- * @returns {*|undefined}  Whatever the next callback returns.
+ * @returns {*}            Whatever the next callback returns.
  */
 export const authenticate = ( context, next ) => {
 	const state = context.store.getState();
--- a/client/lib/media/utils/validate-media-item.js
+++ b/client/lib/media/utils/validate-media-item.js
@@ -8,7 +8,7 @@ import { isSupportedFileTypeInPremium } from 'calypso/lib/media/utils/is-support
  *
  * @param  {object}      site Site object
  * @param  {object}      item Media item
- * @returns {Array|null}      Validation errors, or null if no site.
+ * @returns {Array|undefined} Validation errors, or null if no site.
  */
 export function validateMediaItem( site, item ) {
 	const itemErrors = [];
--- a/client/lib/media/utils/validate-media-item.js
+++ b/client/lib/media/utils/validate-media-item.js
@@ -6,8 +6,8 @@ import { isSupportedFileTypeInPremium } from 'calypso/lib/media/utils/is-support
 /**
  * Validates a media item for a site, and returns validation errors (if any).
  *
- * @param  {Object}      site Site object
- * @param  {Object}      item Media item
+ * @param  {object}      site Site object
+ * @param  {object}      item Media item
  * @returns {Array|undefined} Validation errors, or undefined if no site.
  */
 export function validateMediaItem( site, item ) {
--- a/client/lib/media/utils/validate-media-item.js
+++ b/client/lib/media/utils/validate-media-item.js
@@ -6,9 +6,9 @@ import { isSupportedFileTypeInPremium } from 'calypso/lib/media/utils/is-support
 /**
  * Validates a media item for a site, and returns validation errors (if any).
  *
- * @param  {object}      site Site object
- * @param  {object}      item Media item
- * @returns {Array|undefined} Validation errors, or null if no site.
+ * @param  {Object}      site Site object
+ * @param  {Object}      item Media item
+ * @returns {Array|undefined} Validation errors, or undefined if no site.
  */
 export function validateMediaItem( site, item ) {
 	const itemErrors = [];
--- a/client/lib/post-normalizer/rule-content-make-images-safe.js
+++ b/client/lib/post-normalizer/rule-content-make-images-safe.js
@@ -31,7 +31,7 @@ const removeUnwantedAttributes = ( node ) => {
  */
 const imageShouldBeRemovedFromContent = ( imageUrl ) => {
 	if ( ! imageUrl ) {
-		return;
+		return false;
 	}
 
 	const bannedUrlParts = [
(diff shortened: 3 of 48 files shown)


client/gutenberg/editor/controller.js
```diff
@@ -109,7 +109,7 @@ function waitForPreferredEditorView( context ) {
  *
  * @param {object} context  Shared context in the route.
  * @param {Function} next   Next registered callback for the route.
- * @returns {*}             Whatever the next callback returns.
+ * @returns {undefined}             Whatever the next callback returns.
```
tyxla: According to its TS definition, the callback can return any value:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/page/index.d.ts#L257

Therefore changing this to `undefined` could cause an inconsistency. 

We can enable the any type in JSDoc by delcaring `*` in `settings.jsdoc.preferredTypes`.

client/lib/media/utils/validate-media-item.js
```diff
@@ -8,7 +8,7 @@ import { isSupportedFileTypeInPremium } from 'calypso/lib/media/utils/is-support
  *
  * @param  {object}      site Site object
  * @param  {object}      item Media item
- * @returns {Array|null}      Validation errors, or null if no site.
+ * @returns {Array|undefined} Validation errors, or null if no site.
```
tyxla: We should alter the description here, as well. 
```suggestion
 * @returns {Array|undefined} Validation errors, or `undefined` if no site.
```

client/lib/post-normalizer/rule-content-make-images-safe.js
```diff
@@ -31,7 +31,7 @@ const removeUnwantedAttributes = ( node ) => {
  */
 const imageShouldBeRemovedFromContent = ( imageUrl ) => {
 	if ( ! imageUrl ) {
-		return;
+		return false;
```
tyxla, line 34: This is a safe change, since it's used only once in an `if()` and the result is implicitly converted to a boolean anyway.

client/my-sites/plans/legacy-plan-notice.jsx
```diff
@@ -25,6 +25,7 @@ const maybeRenderLegacyPlanNotice = ( eligibleForProPlan, { plan } ) => {
 			></Notice>
 		);
 	}
+	return null;
```
tyxla, line 28: Change looks good, but it seems like this function's JSDoc is missing a return type?

client/post-editor/media-modal/detail/detail-preview-media-file.tsx
```diff
@@ -35,10 +35,14 @@ const EditorMediaModalDetailPreviewMediaFile: React.FC< Props & LocalizeProps >
...
 
+	const mediaUrl = url( item, {} );
+	if ( ! mediaUrl ) {
+		return null;
+	}
 	return (
 		<MediaFile
 			component={ component }
-			src={ url( item, {} ) }
+			src={ mediaUrl }
```
tyxla, line 45: Ideally, I'd recommend doing this in another PR.

client/server/render/index.js
```diff
@@ -67,6 +67,8 @@ export function renderJsx( view, props ) {
 	return doctype + ReactDomServer.renderToStaticMarkup( createElement( component, props ) );
 }
 
+// JSdoc rule can't handle the function throwing an Error instead of returning.
```
tyxla: Can't we just use `@throws`?

packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
```diff
@@ -10,6 +10,15 @@ module.exports = {
 	},
 	extends: [ 'eslint:recommended', 'plugin:jsdoc/recommended' ],
 	plugins: [ 'jsdoc', 'wpcalypso', 'inclusive-language' ],
+	settings: {
+		jsdoc: {
+			preferredTypes: {
+				// Allow both variations of object types.
```
tyxla: Any good reason for supporting both variations? Perhaps we should unify around a single one (I think `Object` is the consensus in the Gutenberg codebase).

packages/social-previews/src/twitter-preview/tweet.jsx
```diff
@@ -94,7 +94,7 @@ export class Tweet extends PureComponent {
 	 */
 	renderMedia( media ) {
 		if ( ! media ) {
-			return;
+			return null;
```
tyxla, line 97: Should we include `| null` in the returned type above?

tyxla (commented):
Thanks for working on this 🙌 

Left a few suggestions and questions, let me know what you think.

FWIW I think it looks great already!

packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
noahtallen: If that's the case, we can go with `Object`. The main reason was to avoid changing a bunch of code to fix it

client/post-editor/media-modal/detail/detail-preview-media-file.tsx
noahtallen: We unfortunately have to do this here since the types of `url` are changing -- we'd either have to not fix the types in the other file, or accept a failing typescript build

client/gutenberg/editor/controller.js
noahtallen: 👍

noahtallen: It looks like this triggers a bug in eslint jsdoc, where `*` by itself doesn't handle a void/undefined return type. So we have to write `*|undefined`, which I guess is fine for now

noahtallen: It also looks like `*` is allowed, so we don't need to add it to `preferredTypes`

noahtallen: Bug report: https://github.com/gajus/eslint-plugin-jsdoc/issues/950

client/my-sites/plans/legacy-plan-notice.jsx
noahtallen: True. Interestingly, this is still valid, because the rule we're fixing `require-returns-check` just means that "if `@returns` is in jsdoc, there must be a return statement in the code," not necessarily anything about the types.

I did change this to `import('react').Element`, which we seem to use elsewhere

packages/social-previews/src/twitter-preview/tweet.jsx
noahtallen: I think React allows null as an element, so IMO we don't need to (since this is a react component after all)

client/server/render/index.js
noahtallen: Hm, so that didn't work when I originally tried it, but thanks for asking me to take a second look. It appears that JSDoc was correct -- this function can return `undefined sometimes, because this is the catch statement:

https://github.com/Automattic/wp-calypso/blob/beef4ae07d0b4d825bd79f6edf73fdc55e378048/client/server/render/index.js#L137-L147

So we need to update the return type instead

noahtallen:
BTW, the reason there are so many oddities in the jsdoc eslint plugin is because they actually fixed a bug with the `@returns` rule! Before the update, the rule would pass as long as a single `return` statement existed. [Now, they've fixed it so that it checks every branch of the code](https://github.com/gajus/eslint-plugin-jsdoc/issues/932#issuecomment-1372091142). This is definitely better overall, because it forces us to use correct types in more places. But there do seem to be one or two bugs with this new behavior :)

Bug reports:

https://github.com/gajus/eslint-plugin-jsdoc/issues/950

https://github.com/gajus/eslint-plugin-jsdoc/issues/949

packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
noahtallen: I split this off into another PR, because the change and autofix was quite large: https://github.com/Automattic/wp-calypso/pull/72348

noahtallen: As a result, I removed the rule change from this PR.

client/gutenberg/editor/controller.js
```diff
@@ -107,9 +107,9 @@ function waitForPreferredEditorView( context ) {
  * tracking), so we redirect the user to the WP Admin login page in order to store the auth cookie. Users will be
  * redirected back to Calypso when they are authenticated in WP Admin.
  *
- * @param {object} context  Shared context in the route.
- * @param {Function} next   Next registered callback for the route.
- * @returns {*}             Whatever the next callback returns.
+ * @param {Object} context Shared context in the route.
+ * @param {Function} next  Next registered callback for the route.
+ * @returns {*|undefined}  Whatever the next callback returns.
```
tyxla: Doesn't `undefined` qualify as any type too?

tyxla (approved):
This looks great 👍 Thanks @noahtallen! 🚀 

I've committed a couple of last-minute fixes if you don't mind.

tyxla:
Oh, I just saw #72348. Feel free to delete that commit!

noahtallen:
Wow, they already released a fix for one of my bug reports! I updated to include that version, which let us remove the weird `*|undefined` definitions we had to add originally.

Issue #72045, opened 2023-01-13
Stats: Add totals section to the Ads page


grzegorz-cp:
# Development

Use the component showing highlights on the Traffic page (7-day highlight) and the Insights page (annual highlights) to:
- [x] display 3 cards for `Totals`
- [x] extend the component to display (i) tooltip
- [x] place any missing icons as SVG in the shared place where we currently store the `eye` icon SVG
- [x] verify that the mobile view works as expected
- [x] replace the old component and move it above the chart
- [x] populate the tooltip with existing content

Bonus:
- [x] reuse tooltip styles from the Traffic and Insights pages

# Designs

- Wide screen view
![Image](https://user-images.githubusercontent.com/4044428/211943915-f0184879-4cb6-4828-a6b8-ffa1955ace6e.png)

- Small screen view
![Image](https://user-images.githubusercontent.com/112354940/212245593-31a96c3e-634e-4ce6-8a48-d3419d3f7e7c.png)

grzegorz-cp:
@jeffgolenski Tooltip should look the same as on the other pages (dark with bigger padding), shouldn't it?

a8ck3n:
Hello @jeffgolenski — 👋 

Do you have the SVG data for the three required icons? I tried exporting them from Figma but have run into some issues with that. The first two exported fine but they don't draw correctly for me. For the third one (the lightbulb?), I was unable to get it to export. I'll use some existing icon data in the meantime.

Thanks!

jeffgolenski:
@a8ck3n The three icons in the cards above are part of the WordPress Design Library set, which we're loading in calypso. We should only use manual SVGs when an icon isn't present in that set! The icon names are:

* `payment`
* `receipt`
* `tip`

@grzegorz-cp and I discussed the icon system in the past. I'm not exactly sure how they're implemented, but I know the icon set is good to go, we're using it on other pages! I believe you can follow the paper trail[ in this PR](https://github.com/Automattic/wp-calypso/pull/70625), Kevin.

grzegorz-cp:
Thanks for the clarification Jeff.
@a8ck3n you can use the icons Jeff mentioned in a similar way it's [done here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/site.jsx#L4) but include the icon names mentioned above. Use the `Icon` component like [done here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/site.jsx#L81).

[...]

PR #72320, opened 2023-01-19
Stats: Add highlights to Ads page

a8ck3n:
#### Proposed Changes

Add highlights section to top of Ads page.

The first pass at getting the highlights working. Uses the existing data from the store and formats it for display.

- Adds new highlights section at the top of the page
- Tooltip should show correct payout/status notices
- Hides old Totals section (below chart)

AdsHighlightsSnap

Does not yet:

* use correct icon data
* remove the previous implementation

I'd like to address these in follow up PRs.

#### Testing Instructions

* Visit the Calypso Live link.
* View a site with WordAds enabled.
* Visit Stats → Ads and confirm highlights are shown correctly.
* Confirm things work correctly on smaller screen sizes.
* Confirm hover works correctly with the tooltip.

If you need to compare with the old Totals section, you can enable it via the React dev tools. Select the `WordAdsEarnings` component and toggle the `showTotalsSection` property.

DevTools

The new code is only active on non-production domains so it can safely be merged as is.

#### Pre-merge Checklist

Related to #72045.

### DIFF
--- a/client/my-sites/stats/wordads/highlight-card-simple.jsx
+++ b/client/my-sites/stats/wordads/highlight-card-simple.jsx
@@ -6,7 +6,9 @@ function HighlightCardSimple( { heading, icon, value } ) {
 		<Card className="highlight-card">
 			<div className="highlight-card-icon">{ icon }</div>
 			<div className="highlight-card-heading">{ heading }</div>
-			<div className="highlight-card-count">{ displayValue }</div>
+			<div className="highlight-card-count">
+				<span className="highlight-card-count-value">{ displayValue }</span>
+			</div>
 		</Card>
 	);
 }
--- a/client/my-sites/stats/wordads/highlight-card-simple.jsx
+++ b/client/my-sites/stats/wordads/highlight-card-simple.jsx
@@ -1,7 +1,6 @@
 import { Card } from '@automattic/components';
 
 function HighlightCardSimple( { heading, icon, value } ) {
-	// typeof myVar === 'string'
 	const displayValue = typeof value === 'string' ? value : '-';
 	return (
 		<Card className="highlight-card">
--- a/client/my-sites/stats/wordads/highlight-card-simple.jsx
+++ b/client/my-sites/stats/wordads/highlight-card-simple.jsx
@@ -1,0 +1,15 @@
+import { Card } from '@automattic/components';
+
+function HighlightCardSimple( { heading, icon, value } ) {
+	// typeof myVar === 'string'
+	const displayValue = typeof value === 'string' ? value : '-';
+	return (
+		<Card className="highlight-card">
+			<div className="highlight-card-icon">{ icon }</div>
+			<div className="highlight-card-heading">{ heading }</div>
+			<div className="highlight-card-count">{ displayValue }</div>
+		</Card>
+	);
+}
+
+export default HighlightCardSimple;
--- a/client/my-sites/stats/wordads/highlights-section.jsx
+++ b/client/my-sites/stats/wordads/highlights-section.jsx
@@ -1,0 +1,37 @@
+import { Gridicon, HighlightCard } from '@automattic/components';
+import { Icon, starEmpty } from '@wordpress/icons';
+import { useTranslate } from 'i18n-calypso';
+
+// TODO: HighlightCard does not accept string values.
+// Should refactor to accept strings and move the business logic into the callers.
+
+export default function HighlightsSection( props ) {
+	const translate = useTranslate();
+	if ( ! props.isVisible ) {
+		return null;
+	}
+	return (
+		<div className="highlight-cards">
+			<h1 className="highlight-cards-heading">
+				{ translate( 'Totals' ) } <Gridicon icon="info-outline" />
+			</h1>
+			<div className="highlight-cards-list">
+				<HighlightCard
+					heading={ translate( 'Earnings' ) }
+					icon={ <Icon icon={ starEmpty } /> }
+					count={ 0 }
+				/>
+				<HighlightCard
+					heading={ translate( 'Paid' ) }
+					icon={ <Icon icon={ starEmpty } /> }
+					count={ 0 }
+				/>
+				<HighlightCard
+					heading={ translate( 'Outstanding amount' ) }
+					icon={ <Icon icon={ starEmpty } /> }
+					count={ 0 }
+				/>
+			</div>
+		</div>
+	);
+}
--- a/client/my-sites/stats/wordads/highlights-section.jsx
+++ b/client/my-sites/stats/wordads/highlights-section.jsx
@@ -8,6 +8,7 @@ import HighlightCardSimple from './highlight-card-simple';
 // Then refactor this Comp to use HighlightCard again.
 
 function HighlightsSectionHeader( props ) {
+	// TODO: Add support for popup.
 	const translate = useTranslate();
 	const localizedTitle = translate( 'Totals' );
 	return props.showInfoIcon ? (
--- a/client/my-sites/stats/wordads/highlights-section.jsx
+++ b/client/my-sites/stats/wordads/highlights-section.jsx
@@ -56,15 +56,15 @@ function getHighlights( earnings ) {
 }
 
 function payoutNotices( earnings ) {
-	const owed =
-		earnings && earnings.total_amount_owed ? numberFormat( earnings.total_amount_owed, 2 ) : '0.00';
+	const amountOwed = earnings?.total_amount_owed || 0;
+	const amountOwedFormatted = getAmountAsFormattedString( amountOwed );
 	const notice = {
 		id: 'notice',
 		value: translate(
-			'Outstanding amount of $%(amountOwed)s does not exceed the minimum $100 needed to make the payment.',
+			'Outstanding amount of %(amountOwed)s does not exceed the minimum $100 needed to make the payment.',
 			{
 				comment: 'WordAds: Insufficient balance for payout.',
-				args: { amountOwed: owed },
+				args: { amountOwed: amountOwedFormatted },
 			}
 		),
 	};
@@ -80,14 +80,14 @@ function payoutNotices( earnings ) {
 	const payout = {
 		id: 'payout',
 		value: translate(
-			'Outstanding amount of $%(amountOwed)s will be paid approximately 45 days following the end of the month in which it was earned.',
+			'Outstanding amount of %(amountOwed)s will be paid approximately 45 days following the end of the month in which it was earned.',
 			{
 				comment: 'WordAds: Payout will proceed.',
-				args: { amountOwed: owed },
+				args: { amountOwed: amountOwedFormatted },
 			}
 		),
 	};
-	return owed < 100 ? [ notice, limit ] : [ payout ];
+	return amountOwed < 100 ? [ notice, limit ] : [ payout ];
 }
 
 function HighlightsSectionHeader( props ) {
--- a/client/my-sites/stats/wordads/highlights-section.jsx
+++ b/client/my-sites/stats/wordads/highlights-section.jsx
@@ -1,6 +1,7 @@
-import { Gridicon } from '@automattic/components';
+import { Gridicon, Popover } from '@automattic/components';
 import { Icon, starEmpty } from '@wordpress/icons';
 import { numberFormat, translate } from 'i18n-calypso';
+import { useRef, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { getWordAdsEarnings } from 'calypso/state/wordads/earnings/selectors';
 import HighlightCardSimple from './highlight-card-simple';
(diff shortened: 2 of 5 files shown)


wilmersondasilva:
I noticed the `font-family` for the amounts is different from what we're using on the other Traffic and Insights pages and also from the screenshots on https://github.com/Automattic/wp-calypso/issues/72045. Are they different on purpose?

image

wilmersondasilva (approved):
Aside from the `font-family` difference I mentioned above, it worked properly, and the code is fine too.

client/my-sites/stats/wordads/index.jsx
```diff
@@ -136,6 +137,10 @@ class WordAds extends Component {
 		);
 	};
 
+	isProduction() {
```
grzegorz-cp: Could we use feature flags instead of a URL check? It could be done in a similar way to [this](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/stats-insights/index.jsx#L31).

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,153 @@
...
+	const localizedTitle = translate( 'Totals', {
+		comment: 'Heading for WordAds earnings highlights section',
+	} );
+	if ( ! props.notices || props.notices.length === 0 ) {
+		return <h1 className="highlight-cards-heading">{ localizedTitle }</h1>;
+	}
+	return (
+		<h1 className="highlight-cards-heading">
+			{ localizedTitle }{ ' ' }
+			<Gridicon
```
grzegorz-cp: Let's try to use `<Icon />` instead of a `<Gridicon />`. We are migrating as many icons as possible to help with loading them for Odyssey Stats.

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+// TODO: HighlightCard does not accept string values.
+// Should refactor to accept strings and move the business logic into the callers.
+// Then refactor this Comp to use HighlightCard again.
+
+function getAmountAsFormattedString( amount ) {
+	// Takes a Number, formats it to 2 decimal places, and prepends a "$".
+	// This mimics the existing behaviour. I'm assuming we only view/report
+	// on earnings in USD.
+	const formattedAmount = '$' + numberFormat( amount, 2 );
```
grzegorz-cp, line 17: If we need a new number formatter, would you be able to [extend those](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/number-formatters) so we'd have them centralised?

client/my-sites/stats/wordads/index.jsx
```diff
@@ -245,7 +256,7 @@ class WordAds extends Component {
 								</>
 
 								<div className="stats__module-list stats__module-headerless--unified">
-									<WordAdsEarnings site={ site } />
+									<WordAdsEarnings site={ site } showTotalsSection={ ! showNewHighlights } />
```
grzegorz-cp: If you add a condition around this component, you won't have to pass the property. For example:
```
{ !isFeatureFlagEnabled && <WordAdsEarnings site={ site } /> }
```
It might also be better to add the condition around the whole `div` if it doesn't break the layout.

client/my-sites/stats/wordads/highlight-card-simple.jsx
```diff
@@ -0,0 +1,14 @@
+import { Card } from '@automattic/components';
+
+function HighlightCardSimple( { heading, icon, value } ) {
```
grzegorz-cp, line 3: For another variant of a highlight card, could you place it along with others [here](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/highlight-cards)? This way, we have all cards centralised making it easier if we decide to unify them or combine them somehow. We could also add a storybook story for it.

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,153 @@
...
+					heading={ highlight.heading }
+					icon={ highlight.icon }
+					value={ highlight.formattedAmount }
+				/>
+			) ) }
+		</div>
+	);
+}
+
+export default function HighlightsSection( props ) {
```
grzegorz-cp, line 146: Similar to the cards, would it be possible to put another variant of the header [here](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/highlight-cards)?

grzegorz-cp (changes requested):
Thanks for doing this. Before enhancing the solution in future PRs could you use a feature flag and move some of the files to the other place where we store highlight sections?

a8ck3n:
> I noticed the font-family for the amounts is different from what we're using on the other Traffic and Insights pages and also from the screenshots on https://github.com/Automattic/wp-calypso/issues/72045. Are they different on purpose?

That was not on purpose. I didn't notice that at first but now that you've pointed it out, it's pretty obvious! I'll get that fixed up!

client/my-sites/stats/wordads/index.jsx
a8ck3n: I purposefully avoided a feature flag in this case as this is a very localized change. If I didn't anticipate a follow up PR for the icons, I wouldn't need a check at all. Now that I've added in the icons, I think we can just ship this as production-ready though.

I'll make those changes first, then revisit this.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: Thanks for the heads up. I wasn't aware of this. I'll get this fixed up!

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: I'm unsure about this suggestion. The `numberFormat()` used here is part of `i18n-calypso` (the same place our `translate()` call lives) and it's used extensively around Calypso. Does it make sense to introduce a new version of this or maybe to move our new formatters into `i18n-calypso`?

I'd lean towards using the existing version as it will safely replicate the behaviour of the code it's replacing.

Let me know if I'm missing something here that makes the number-formatters component the better choice though. I'm happy to move it if that's the better option.

client/my-sites/stats/wordads/index.jsx
a8ck3n: We can't conditionally render the WordAdsEarnings component because it also handles the render of up to three tables in addition to the totals section. If we used a feature flag we could do that check inside the component but, as noted above, I can remove this completely now that the new icons are part of the PR (incoming).

client/my-sites/stats/wordads/highlight-card-simple.jsx
a8ck3n: My thinking was that I'd open up a maintenance task to refactor the various highlight cards and then move them into the component. We also have a variant here:

https://github.com/Automattic/wp-calypso/tree/trunk/client/my-sites/stats/stats-email-top-row

The new HighlightCardSimple component and the TopCard component are mostly logic free so they should factor together well. Any objection to addressing this in a follow-up PR?

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: I hadn't considered this. I was thinking of the Storybook components as components that might be reusable outside Stats. For example, I can see other teams needing to make use of simple highlight cards. The section components are a bit more data-coupled though, so it made sense they would live inside Stats.

Are we thinking we should develop entire sections or layouts inside the components package, even if we don't see them as reusable?

client/my-sites/stats/wordads/index.jsx
a8ck3n: This check has been removed now!

client/my-sites/stats/wordads/index.jsx
a8ck3n: The old code path has been removed so the property is no longer needed.

a8ck3n:
I've updated the PR to add in the proper icons, update the fonts, and to remove the old code paths. There are still a few bits of feedback unresolved though. Would appreciate your thoughts on those @grzegorz-cp. Thanks!

client/my-sites/stats/wordads/highlight-card-simple.jsx
grzegorz-cp: Sounds good. It will be great to refactor the cards as a maintenance task.

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,154 @@
...
+	// on earnings in USD.
+	const formattedAmount = '$' + numberFormat( amount, 2 );
+	// Differs from previous behaviour in that we don't want "$0.00" as a result.
+	// Per design spec we'll return "$0" in this scenario.
+	// https://github.com/Automattic/wp-calypso/issues/72045
+	return formattedAmount === '$0.00' ? '$0' : formattedAmount;
+}
+
+function getHighlights( earnings ) {
+	const total = earnings && earnings.total_earnings ? Number( earnings.total_earnings ) : 0;
```
kangzj: You could use optional chaining to shorten this as `earnings?.total_earnings ? .... ...`.

And several places in the following code could be shortened in the same way.

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,154 @@
...
+		id: 'payout',
+		value: translate(
+			'Outstanding amount of $%(amountOwed)s will be paid approximately 45 days following the end of the month in which it was earned.',
+			{
+				comment: 'WordAds: Payout will proceed.',
+				args: { amountOwed: owed },
+			}
+		),
+	};
+	return owed < 100 ? [ notice, limit ] : [ payout ];
```
kangzj: `owed` is a string and `100` is a number. It's generally not safe to compare strings to numbers - in this case it runs okay but I'd still recommend not doing this.

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,154 @@
...
+	const [ isTooltipVisible, setTooltipVisible ] = useState( false );
+	const infoReferenceElement = useRef( null );
+	const localizedTitle = translate( 'Totals', {
+		comment: 'Heading for WordAds earnings highlights section',
+	} );
+	if ( ! props.notices || props.notices.length === 0 ) {
+		return <h1 className="highlight-cards-heading">{ localizedTitle }</h1>;
+	}
+	return (
+		<h1 className="highlight-cards-heading">
```
kangzj, line 101: Sorry for being picky. The code here are mostly duplication of the early return above. Would it be better only to conditional show the differences?

client/my-sites/stats/wordads/highlights-section.jsx
grzegorz-cp: That's a very good point. From what I can see, `numberFormat` with two parameters only formats a number but doesn't take locale into account, which means that for countries using a comma instead of a period before fractions, this won't be swapped. On the other hand, we hard-coded the dollar sign. Can this page show a different currency, or is it always USD?
Does showing a dollar value mean we use the US number format?
All the other highlight sections take locale into account (they use browser locale), so I lean towards using a localised number even with a foreign currency.
The table values make the same mistake, so if we swap, this wouldn't be constant until table numbers are also addressed.
This is not a blocking issue, so I can accept both outcomes, but we should probably create a task to look at this if we don't do it here.

kangzj (commented):
Good job! It works as described.

A tiny styling feedback: On smaller screens the width of the cards are not aligned with the ones on other pages.

image

image

client/my-sites/stats/wordads/highlights-section.jsx
```diff
@@ -0,0 +1,154 @@
...
+	);
+}
+
+export default function HighlightsSection( props ) {
+	const earningsData = useSelector( ( state ) => getWordAdsEarnings( state, props.siteId ) );
+	const highlights = getHighlights( earningsData );
+	const notices = payoutNotices( earningsData );
+	return (
+		<div className="highlight-cards wordads">
+			<HighlightsSectionHeader notices={ notices } />
```
grzegorz-cp, line 152: Since those two lines are other functional components, they should be placed in their own files, or they should stay in this one, not other functions.
Were you planning to do it later?

client/my-sites/stats/wordads/highlights-section.jsx
grzegorz-cp: I see your point now, and this file has many functions specific to ads. So let's keep it here.

grzegorz-cp (approved):
I added some suggestions, but nothing is blocking.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: My assumption was that WordAds only deals in USD amounts and my initial focus was to replicate the existing behaviour so as not to confuse things. The "$" symbol is hard-coded in the previous implementation as well.

My preference would be to address changes in the currency handling in a separate PR though, once we've had a chance to investigate and confirm that results should always be shown in USD (or not).

I'll spin up an issue to track this and I'll make a note that the behaviour should be consistent between the highlight cards and the values in the tables.

a8ck3n: Opened an issues for this [here](https://github.com/Automattic/wp-calypso/issues/72492).

client/my-sites/stats/wordads/highlight-card-simple.jsx
a8ck3n: Opened a task for this [here](https://github.com/Automattic/wp-calypso/issues/72493).

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: Thanks for the reminder. I was meaning to update the logic here. I'll get it updated.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: I'll fix this up to. I didn't realize I was doing this.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: I initially had that but it generated an empty DIV in the DOM for some reason. If you clicked on it, it would disappear. Coding it this way resulted in cleaner DOM output. I'll check on this again to see if I was overlooking something.

client/my-sites/stats/wordads/highlights-section.jsx
grzegorz-cp: Great, thanks @a8ck3n.

a8ck3n:
> A tiny styling feedback: On smaller screens the width of the cards are not aligned with the ones on other pages.

Hey @kangzj, thanks for the review!

This was intentional as the strings we are displaying can be much longer. Likewise for the "Outstanding amount" label. I bumped up the minimum width to account for those. See [this commit](https://github.com/Automattic/wp-calypso/pull/72320/commits/2ef43cbde8f0dc92707ffdda471c51d3d79f3104) for my initial thoughts.

Maybe we could introduce a second style (with a smaller `min-width`) and apply the correct class based on the string length. I'll test against the longer labels first though as I'm pretty sure it didn't fit within the previous min.

— Updated with testing notes.

So here's what it looks like if we keep the existing `min-width` value of 180px. Larger amounts don't fit comfortably and the longer label in the 3rd box causes wrapping.

MinSize

On the plus side, now that I've fixed the font issue mentioned by @wilmersondasilva those larger amounts take less space (monospaced vs proportional font) so I can update this to have a smaller `min-width`. I'll drop this down from the current 280px to 240px.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: I'm still seeing this behaviour in Chrome but not in Firefox. Maybe it's just a quirk with how Chrome is reformatting it's output when browsing the DOM.

client/my-sites/stats/wordads/highlights-section.jsx
a8ck3n: > Were you planning to do it later?

I wasn't planning to break these out. I was thinking of them more like helper components that didn't necessarily need to be exposed (or exported). I'm of the mind that smaller functions and components are easier to understand when it comes to reading new code (or revisiting old code) and figuring out what's happening.

If it's a best practice we are looking to follow, I guess I'd favour inlining the code but I do feel that hurts readability.

a8ck3n:
Rebasing in the hopes that fixes the (newly) failing E2E tests.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7770919

Thank you @a8ck3n for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7770919).

Issue #72191, opened 2023-01-17
Pattern Assembler - Pattern Renderer fails to render height 100vh


miksansegundo:
## Description

There are [Hero patterns](https://dotcompatterns.wordpress.com/2022/10/11/hero/) that use `min-height:100vh` which is limited to `2000px` in the thumbnail because there is a [MAX_HEIGHT](https://github.com/Automattic/wp-calypso/pull/69573#issuecomment-1357449351) constant with that value.

However, in the large preview, it works differently because the `100vh` is limited to `330px`, not sure for what reason but anyways if it works with the `maxHeight` as `2000px` won't be good either because it should be `100vh`. The editor does render this pattern well, I think that's because it doesn't use the same component. 

## Acceptance criteria

- [ ] Patterns with `height: 100vh` should be rendered with that height in the large preview 

## Screenshots

![Image](https://user-images.githubusercontent.com/1881481/212863041-285b0918-0ac6-438e-b14c-9e612474fb7f.png)
![Image](https://user-images.githubusercontent.com/1881481/212863065-8484647d-324d-4194-bf84-2f6a5e8130ad.png)
![Image](https://user-images.githubusercontent.com/1881481/212863094-95d2774d-d088-4791-930f-987461778446.png)
![Image](https://user-images.githubusercontent.com/1881481/212863113-aa92f1b5-7722-479d-a8a1-104cfd1ca4e4.png)

Related https://github.com/Automattic/wp-calypso/pull/70063

PR #72321, opened 2023-01-19
Block Renderer: Support viewportHeight to render 100vh correctly

arthur791004:
#### Proposed Changes

* Make `block-renderer-container` support the `viewportHeight` and pass it to the height of iframe before the content is loaded. So, the `100vh` of the content will render correctly.

Note that this is **not a perfect solution** and we still need to resolve the following issues
* If you add the pattern too fast, the `100vh` might not be rendered correctly as the `viewportHeight` might be `undefined` at the beginning
* ~~If you resize the window height (including switching the device to mobile), the rendered `100vh` won't be changed.~~

**Screenshots**

image

#### Testing Instructions

* Add the following pattern to [patterns-data.tsx](https://github.com/Automattic/wp-calypso/blob/b576934649d8cba228c05d9066067d36b1a66462/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/patterns-data.ts#L188)
  ```js
  {
    id: 3747,
    name: 'Hero',
    category: 'posts',
  },
  ``` 
* Go to /setup?siteSlug=<your_site>
* When you're in the Pattern Assembler step, select the newly added section and verify its height is correct

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/72191

### DIFF
--- a/packages/block-renderer/src/components/block-renderer-container.tsx
+++ b/packages/block-renderer/src/components/block-renderer-container.tsx
@@ -107,7 +107,7 @@ const ScaledBlockRendererContainer = ( {
 				style={ {
 					position: 'absolute',
 					width: viewportWidth,
-					height: ( contentHeight as number ) || viewportHeight,
+					height: viewportHeight || ( contentHeight as number ),
 					pointerEvents: 'none',
 					// This is a catch-all max-height for patterns.
 					// See: https://github.com/WordPress/gutenberg/pull/38175.
--- a/packages/block-renderer/src/components/block-renderer-container.tsx
+++ b/packages/block-renderer/src/components/block-renderer-container.tsx
@@ -17,6 +17,7 @@ interface BlockRendererContainerProps {
 	styles?: RenderedStyle[];
 	inlineCss?: string;
 	viewportWidth?: number;
+	viewportHeight?: number;
 	maxHeight?: number;
 	minHeight?: number;
 }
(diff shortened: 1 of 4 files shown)


packages/block-renderer/src/components/block-renderer-container.tsx
```diff
@@ -105,7 +107,7 @@ const ScaledBlockRendererContainer = ( {
 				style={ {
 					position: 'absolute',
 					width: viewportWidth,
-					height: contentHeight as number,
+					height: ( contentHeight as number ) || viewportHeight,
```
miksansegundo: I found that `viewportHeight` should be first.

miksansegundo:
I've pushed commits to update the viewport size on resize and the device switch.

miksansegundo (approved):
It's working well.

https://user-images.githubusercontent.com/1881481/213618776-aa0ca471-f66b-4e0e-bb0f-4c9ba63c32c5.mov

There is still an issue with the pattern height in the thumbnail preview that seems not related to the recent changes. 

It seems that the loading time of a pattern affects the calculation of its height because I see different heights applied to the Hero pattern randomly each time I go back to the design picker to enter the assembler again.


This is the expected height because of the `maxHeight: 2000px`.
Screenshot 2566-01-20 at 11 33 58

These are not expected, but it's not bad because I think 2000px is too tall.

Screenshot 2566-01-20 at 11 35 42

Screenshot 2566-01-20 at 11 34 16

We'll follow up with another issue.

autumnfjeld:
>  because I think 2000px is too tall.
Agree.

autumnfjeld:
All groovy. Good to merge. :)

PR #72327, opened 2023-01-19
Copy Site: Add back functionality on domains step

kozer:
#### Proposed Changes

Add back functionality on domains step

#### Testing Instructions

1. Navigate to `/sites`
2. Copy a site.
3. You will be navigated to the domains selection screen
4. Make sure that a back button exists in the top right corner.
5. Click the back button
6. Make sure you'll be navigated back to `/sites`

back_button_domains_step

#### Pre-merge Checklist

Fixes https://github.com/Automattic/dotcom-forge/issues/1519


client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
```diff
@@ -471,6 +473,7 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
 				</div>
 			}
 			recordTracksEvent={ recordTracksEvent }
+			goBack={ () => submit?.( { previousStep: true } ) }
```
danielbachhuber: This seems a little hacky.

Given we have access to `isCopySiteFlow( flow )` in this context, can we call `window.location.assign( '/sites' )` directly here?

packages/onboarding/src/step-navigation-link/index.tsx
```diff
@@ -59,7 +59,9 @@ const StepNavigationLink: React.FC< Props > = ( {
 			primary={ primary }
 			borderless={ borderless }
 			className={ buttonClasses }
-			onClick={ onClick }
+			// Use onMouseDown here to prevent onClick blocked by onBlur
+			// if any exists in adjacent elements
+			onMouseDown={ onClick }
```
danielbachhuber: Out of curiosity, which browsers did you test this in?

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
```diff
@@ -461,7 +462,8 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
 		<StepContainer
 			stepName="domains"
 			isWideLayout={ true }
-			hideBack={ true }
+			shouldStickyNavButtons={ isCopySiteFlow( flow ) }
```
danielbachhuber, line 465: What other flows should we test to review these changes? The pull request description doesn't seem to indicate any.

packages/onboarding/src/step-navigation-link/index.tsx
kozer: On Chrome. Do you want me to test it on others like Firefox for example? (I'm on linux, so Safari is not an option for me.)

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
kozer: It's a bit hacky, but I don't think that adding `window.location.assign` here is the best way to go. All the navigation decisions are happening inside `useStepNavigation` which is why the function exists. Adding it here, we scatter the navigation decision. There is a `goBack` function we can use, but the linter marks it as it's deprecated.

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
kozer: I don't think there are other flows that we need to test. All those changes are happening only if it's a copy site flow, otherwise, they are the same as they were before.

packages/onboarding/src/step-navigation-link/index.tsx
danielbachhuber: What about mobile?

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
danielbachhuber: If this code is used by other signup flows, it'd be good to test at least one of them to make sure we haven't accidentally broken something.

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
```diff
@@ -461,7 +462,8 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
...
+			hideBack={ ! isCopySiteFlow( flow ) }
 			hideSkip={ true }
 			flowName="linkInBio"
```
sejas: We probably need to change this hardcoded `flowName`.

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
sejas: @kozer is right and `goBack` is deprecated. What about using `exitFlow` ?
The logic would live in `useStepNavigation` but it would be more clear that we are leaving the flow 👌 .  Everybody wins 😎 .

```suggestion
			goBack={ () => exitFlow?.( '/sites' ) }
```

We would also need to declare the `exitFlow` function in `declarative-flow/copy-site.tsx`:
```diff
diff --git a/client/landing/stepper/declarative-flow/copy-site.tsx b/client/landing/stepper/declarative-flow/copy-site.tsx
index 0be6d8983b..d279612cb9 100644
--- a/client/landing/stepper/declarative-flow/copy-site.tsx
+++ b/client/landing/stepper/declarative-flow/copy-site.tsx
@@ -80,3 +79,0 @@ const copySite: Flow = {
-					if ( providedDependencies.previousStep === true ) {
-						return window.location.assign( `/sites` );
-					}
@@ -136 +133,5 @@ const copySite: Flow = {
-		return { goNext, goBack, goToStep, submit };
+		const exitFlow = ( location: string ) => {
+			window.location.assign( location );
+		};
+
+		return { goNext, goBack, goToStep, submit, exitFlow };
```

sejas (changes requested):
Great work @kozer 🙌 ,
- ~I noticed that the back button is not present on mobile version.~ , It's present in the footer and it works well.
- Would be great if the back button is on the left instead of the right. What do you think? Not a strong opinion, though.
- I added a couple of suggestions and created a PR https://github.com/Automattic/wp-calypso/pull/72450 related to the `submit` call.

packages/onboarding/src/step-navigation-link/index.tsx
danielbachhuber: To expand on my thought process: semantically, it doesn't feel correct to change this from `onClick` to `onMouseDown`. It seems like a bug that some `onBlur` preventing `onClick`...

wojtekn:
> Make sure that a back button exists in the top right corner.

It would make sense to keep naming, position, and behavior consistent with the "Back to Sites" we have in other flows. To see an example, see the import flow:

1. Navigate to `/sites`
2. Click the arrow on the "Add new site" button
3. Click "Import an existing site"

packages/onboarding/src/step-navigation-link/index.tsx
kozer: There is indeed an `onBlur` in a neighbor element. Leaving it as is, means that we can't use an onBlur that triggers changes in state, and even if we fix that, there is a possibility that this bug will reappear. I'll check mobile and if doesn't work, I'll try to change the `onBlur` function. However, this onBlur is deep inside the StepContainer and I'm not sure what ll be affected by doing that.

kozer:
> I noticed that the back button is not present on mobile version.

I tried to use what already existed in `StepContainer` component. I guess it didn't work well initially. Thanks for that! I'll revisit it.

> Would be great if the back button is on the left instead of the right. What do you think? Not a strong opinion though.

I can manage to make it appear on the left, however, it seems that it messes up with the `Copy Site` title.

> I added a couple of suggestions and created a PR https://github.com/Automattic/wp-calypso/pull/72450 related to the submit call.

Thanks! Left some comments!

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
kozer: This is like we reinvented the `goBack` functionality.
Docs mention the following, so I'm a bit skeptical about this solution:

	 * @deprecated Avoid this method. Use submit() instead.
	 * If you need to navigate back and forth between
	 * stepper screens, consider adding screen(s) to a new
	 * stepper flow and linking directly between flows/screens.

kozer:
@danielbachhuber @sejas 

> I noticed that the back button is not present on mobile version.
> What about mobile?

Tested this on my android phone using `calypso.live` and it worked perfectly (Chrome and firefox)

Here is the link if you want to try it: https://hash-e730f1c799326ef392e49d60396883ecc5130c89.calypso.live/sites

Screenshot_2023-01-24-12-40-37-751-edit_org mozilla firefox

packages/onboarding/src/step-navigation-link/index.tsx
kozer: Tested it on Android on Chrome and Firefox. Seems that works as expected. If you or @sejas can give this a try on Safari, it would be great.

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
sejas: resolving this conversation since we decided to merge the changes and use `exitFlow`

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
sejas: Done!

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
sejas: - This is done, after merging https://github.com/Automattic/wp-calypso/pull/72450

packages/onboarding/src/step-navigation-link/index.tsx
sejas: It works on mobile 👌 .
It would be great if we could use the `onClick` and move the button to the left like in the importer flow, following @wojtekn's [suggestion](https://github.com/Automattic/wp-calypso/pull/72327#issuecomment-1401555490).
I know it's not an easy task because our stepper and the import flow use different components, so I'm fine creating a new issue to improve this.

| Screenshot 2023-01-24 at 13 40 13 | Screenshot 2023-01-24 at 13 30 05 |
|---|---|
| Copy Site Flow | [Import Flow](https://wordpress.com/start/import/domains) |

danielbachhuber: @kozer Want to share the bug you've found on calypsop2 so we can get broader input on how we might solve it?

sejas: For future reference: p4TIVU-aoH-p2

kozer: @danielbachhuber , @sejas I'll wait for the outcome of this p2 post to see how I'll proceed from here.
In the meantime, I pushed some changes, changing what I had pushed originally with the second solution I propose in this post. 
It restores the functionality as it was + the `onMouseDown` with `preventDefault`, so we won't have any unexpected behavior as we could have with `onMouseDown`

p-jackson:
I think I've gotten to the bottom of how to prevent the blur event from breaking the click event p4TIVU-aoH-p2#comment-11701

Issue #72019, opened 2023-01-12
Site Editor - Open in edit mode


Addison-Stavlo:
Recently, Gutenberg changes have forced the sidebar in the site editor to be open by default. This exposes complexities to our users at the wrong time and makes it easy for them to get lost with no current way to return to the default homepage template+content view that the editor loads in.  Some context and discussion in p9Jlb4-6ad-p2

![Screen Shot 2023-01-12 at 12 45 54 PM](https://user-images.githubusercontent.com/28742426/212140809-c6bd0b25-7176-4b8e-b8a5-564638ccdfe5.png)

It seems the [edit mode can be persisted in the URL of the editor](https://github.com/WordPress/gutenberg/issues/36667#issuecomment-1377221779). 

To keep things simple and streamlined for our users and their flows, we should update the URLs we use for loading the site editor to open this in edit mode.

agrullon95:
Added `canvas=edit` query parameter to Calypso site-editor to always default to edit mode (hides sidebar on load) with:  #72339

PR #72339, opened 2023-01-19
Launchpad: Update tasks to redirect into site-editor edit mode

agrullon95:
#### Estimated Time to Test / Review:

Test -> Short
Review -> Short

#### Proposed Changes

The latest Gutenburg site editor changes will open the sidebar by default when users navigate to `/site-editor/{siteSlug}`. This may be confusing for users as we try to direct them towards the template/content editing view.

* These changes add "canvas=edit" query parameter to the site editor route to navigate directly into "edit" mode (Prevents the sidebar navigation from displaying first)

#### Testing Instructions

* Checkout this branch
* Create new free / link-in-bio site (`/setup/free/` or `/setup/link-in-bio/`) or use existing site
* Navigate to free flow launchpad screen and select the `Edit site design` task. Confirm it navigates directly into site-editor edit mode.
* Navigate to link-in-bio flow launchpad screen (you can use the same site slug) and select the `Add links` task. Confirm it navigates directly into site-editor edit mode.
* Create a videopress site (`/setup/videopress`) and click the `Upload your first video` task. Confirm that it navigates into site-editor edit mode.

#### Pre-merge Checklist

Related to #72019

### DIFF
--- a/client/gutenberg/editor/calypsoify-iframe.tsx
+++ b/client/gutenberg/editor/calypsoify-iframe.tsx
@@ -791,6 +791,7 @@ const mapStateToProps = (
 	const currentRoute = getCurrentRoute( state );
 	const postTypeTrashUrl = getPostTypeTrashUrl( state, postType );
 	const siteOption = isJetpackSite( state, siteId ) ? 'jetpack_frame_nonce' : 'frame_nonce';
+	const canvasEditMode = getQueryArg( window.location.href, 'canvas' ) === 'edit' ? 'edit' : null;
 
 	let queryArgs = pickBy( {
 		post: postId,
@@ -811,6 +812,7 @@ const mapStateToProps = (
 		...pressThisData,
 		answer_prompt: getQueryArg( window.location.href, 'answer_prompt' ),
 		completedFlow,
+		...( canvasEditMode && { canvas: canvasEditMode } ),
 	} );
 
 	// needed for loading the editor in SU sessions
--- a/client/gutenberg/editor/calypsoify-iframe.tsx
+++ b/client/gutenberg/editor/calypsoify-iframe.tsx
@@ -791,7 +791,6 @@ const mapStateToProps = (
 	const currentRoute = getCurrentRoute( state );
 	const postTypeTrashUrl = getPostTypeTrashUrl( state, postType );
 	const siteOption = isJetpackSite( state, siteId ) ? 'jetpack_frame_nonce' : 'frame_nonce';
-	const canvasEditMode = getQueryArg( window.location.href, 'canvas' ) === 'edit' ? 'edit' : null;
 
 	let queryArgs = pickBy( {
 		post: postId,
@@ -812,7 +811,7 @@ const mapStateToProps = (
 		...pressThisData,
 		answer_prompt: getQueryArg( window.location.href, 'answer_prompt' ),
 		completedFlow,
-		...( canvasEditMode && { canvas: canvasEditMode } ),
+		canvas: 'edit', // Load side editor in edit mode instead of displaying sidebar by default (Gutenberg v15.0.0)
 	} );
 
 	// needed for loading the editor in SU sessions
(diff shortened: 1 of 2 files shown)


agrullon95:
cc: @roundhill & @jgcaruso - I added a minor change related to the videopress flow. Can you review these changes?

client/gutenberg/editor/calypsoify-iframe.tsx
```diff
@@ -811,6 +812,7 @@ const mapStateToProps = (
 		...pressThisData,
 		answer_prompt: getQueryArg( window.location.href, 'answer_prompt' ),
 		completedFlow,
+		...( canvasEditMode && { canvas: canvasEditMode } ),
```
mattwiebe: I think we should simply assign this at all times. We don't want to be sending people into the template selection mode at all, ever.

edanzer:
This tests well for me. 
 - [x] Free Flow > Edit Site goes to full screen editor (and query arg is present in url)
 - [x] Link in Bio > Add Links goes to full screen editor (and query arg is present in url)
 - [x] VideoPress - Best I can tell this behavior *has not changed.* The Upload Video link goes to the page editor. The query param is not present there. I see we have a fallback to go to the side editor with query param, but that is not triggered. I assume this is the expected behavior. 
 
Code changes also look good and straightforward.

client/gutenberg/editor/calypsoify-iframe.tsx
edanzer: This seems sensible to me. But one concern is that changes goes well beyond our current use case and seems to change the calypso site editor handling/display for all cases. That seems like a big change, and I I wonder if that warrants wider discussion beyond our team?

edanzer (approved):

client/gutenberg/editor/calypsoify-iframe.tsx
Addison-Stavlo: I'm in agreement with Matt here. We don't want to surface this to our users by default in any entry the way it is right now. But we may be able to in the future after core has iterated on its development.

With that it's important to leave a clear comment that we are doing this.

Addison-Stavlo: > This seems sensible to me. But one concern is that changes goes well beyond our current use case and seems to change the calypso site editor handling/display for all cases. That seems like a big change, and I I wonder if that warrants wider discussion beyond our team?

To me the big change was in Gutenberg that started exposing this to our users by default. It is still being iterated on, but is currently not in a place that is good for us to expose by default to our users on dotcom.

We relayed some of this feedback to core in this p2: p9Jlb4-6ad-p2#comment-6657, expressed desire to be able to bypass this, and were soon after notified that a change had been implemented that would allow us to do so.

We also reached out to team-calypso in slack to make them aware of this change, and present them the option to handle it on their end since they generally handle things around the gutenberg upgrade cycles. They responded that they would be happy for us to handle this, which we are happy to do 😁  p1673663392570019/1673546040.780839-slack-C7YPUHBB2

jgcaruso:
> * VideoPress - Best I can tell this behavior _has not changed._ The Upload Video link goes to the page editor. The query param is not present there. I see we have a fallback to go to the side editor with query param, but that is not triggered. I assume this is the expected behavior.

Correct, the site editor link is just a fallback for an extreme case where the Home page id can't be found. Not actually testable without modifying the code to force it into the fallback condition... which I did to test this and the new query param whether there or not seems fine to me.

agrullon95:
@jgcaruso - I reverted the changes since we will just default to edit mode when navigating to the site editor. Thanks for testing.

sharpirate:
This tests well for me:
✔️ For Free Flow -> Goes directly to the edit screen of the Site Editor
✔️ For Link in Bio -> Goes directly to the edit screen of the Site Editor
✔️ For VideoPress -> Goes to the page editor

mattwiebe (approved):

PR #69643, opened 2022-10-31
Checkout: Remove edit button and show only longer term variants

sirbrillig:
#### Proposed Changes

Monthly plan variants are being selected too frequently in checkout. To help with this, the term variant picker (for WordPress.com, not Jetpack) was hidden behind an "Edit" button in https://github.com/Automattic/wp-calypso/pull/68606. However, now longer term variants are not being selected often enough.

This PR removes the "Edit" button and alters the variant picker (again, for WordPress.com and not Jetpack since Jetpack has no other way to select a monthly plan) so that it will not display variants which have a shorter term than the item that was in the cart when checkout loaded. 

Fixes https://github.com/Automattic/wp-calypso/issues/69633

Note: there's also a plan to consider switching to radio buttons from the designs in https://github.com/Automattic/wp-calypso/pull/69494#issuecomment-1295315829 (we can reuse some of the code in https://github.com/Automattic/wp-calypso/pull/66027) but that will be done separately.

#### Screenshots

Before:
Screen Shot 2022-11-02 at 6 21 35 PM

After:
Screen Shot 2022-11-02 at 6 22 11 PM

Before:
Screen Shot 2022-11-02 at 6 23 01 PM

After:
Screen Shot 2022-11-02 at 6 23 08 PM

#### Testing Instructions

Unit tests are included but you can test manually by doing the following.

Test annual/biennial plans:

- Add a WordPress.com annual plan to your cart and visit checkout.
- Verify that underneath the plan is a variant picker (and no "Edit" button).
- Click on the variant picker and verify that you can see the annual option and the biennial option, but not the monthly option.
- Click to select the biennial option and wait for the cart to finish updating.
- Click on the variant picker and verify that you can _still_ see the annual option and the biennial option, but not the monthly option.
- Reload the checkout page.
- Verify that there is no variant picker visible.

Test monthly plans:

- Add a WordPress.com monthly plan to your cart and visit checkout.
- Verify that underneath the plan is a variant picker (and no "Edit" button).
- Click on the variant picker and verify that you can see the monthly, annual, and biennial options.
- Click to select the annual option and wait for the cart to finish updating.
- Click on the variant picker and verify that you can _still_ see the monthly, annual, and biennial options.
- Reload the checkout page.
- Click on the variant picker and verify that you can only the annual and biennial options, but not the monthly option.

Test Jetpack plans:

- Add a Jetpack annual plan to your cart (for a Jetpack site) and visit checkout.
[...]

PR #72344, opened 2023-01-19
Improve the nudge for dotorg themes

dsas:
#### Proposed Changes

Reworded nudge
Before | After
-------|------
image | image

New dialog

Custom domain | wpcom subdomain
-------|------
image | image

#### Testing Instructions

 1. Use live link go to /themes on a < business site
 2. Search for a dot org only theme (not a TTn theme)
 3. Click on the theme
 4. You should see the theme detail page, including a nudge
 'Upgrade your plan to install third-party themes'
 5. Click the nudge and a dialog should appear
 6. The dialog should explain your domain will change (if on a wordpress.com subdomain) along with some enticing wording about upgrading.
 7. The dialog should have a button that takes you through the purchase process.
 8. Reasonable tracks events should also be emitted

You should test with a site with a custom domain as well as with a subdomain to ensure the dialog is worded correctly.

You should also test with zaino (woo specific theme) and Makoney (marketplace theme). They should not use the new wording or trigger the new dialog

##### Tracks events
Event name | Description | Properties
-----------|-------------|-----------------
calypso_theme_upsell_plan_click | Click on the upsell nudge | `cta_name:theme_upsell_plan_click, cta_feature:upload-themes`
calypso_automated_transfer_eligibility_ineligible | Display of the dialog | `holds: NO_BUSINESS_PLAN`
calypso_automated_transfer_eligibilty_click_proceed | Click proceed on the dialog | `context:themes`

 
#### Pre-merge Checklist

  - There are none 

Related to https://github.com/Automattic/dotcom-forge/issues/1299

### DIFF
--- a/client/blocks/eligibility-warnings/hold-list.tsx
+++ b/client/blocks/eligibility-warnings/hold-list.tsx
@@ -223,7 +223,7 @@ export const HoldList = ( { context, holds, isMarketplace, isPlaceholder, transl
 
 	return (
 		<>
-			{ ! isPlaceholder && context !== 'plugin-details' && (
+			{ ! isPlaceholder && context !== 'plugin-details' && context !== 'themes' && (
 				<HardBlockingNotice
 					holds={ holds }
 					translate={ translate }
@@ -258,9 +258,11 @@ export const HoldList = ( { context, holds, isMarketplace, isPlaceholder, transl
 						! isKnownHoldType( hold, holdMessages ) ? null : (
 							<div className="eligibility-warnings__hold" key={ hold }>
 								<div className="eligibility-warnings__message">
-									<div className="eligibility-warnings__message-title">
-										{ holdMessages[ hold ].title }
-									</div>
+									{ holds.length > 1 && (
+										<div className="eligibility-warnings__message-title">
+											{ holdMessages[ hold ].title }
+										</div>
+									) }
 									<div className="eligibility-warnings__message-description">
 										{ holdMessages[ hold ].description }
 									</div>
--- a/client/blocks/eligibility-warnings/index.tsx
+++ b/client/blocks/eligibility-warnings/index.tsx
@@ -96,7 +96,7 @@ export const EligibilityWarnings = ( {
 			'eligibility-warnings__placeholder': isPlaceholder,
 			'eligibility-warnings--with-indent': showWarnings,
 			'eligibility-warnings--blocking-hold': hasBlockingHold( listHolds ),
-			'eligibility-warnings--without-title': context !== 'plugin-details',
+			'eligibility-warnings--without-title': context !== 'plugin-details' && context !== 'themes',
 		},
 		className
 	);
@@ -135,7 +135,7 @@ export const EligibilityWarnings = ( {
 	const blockingMessages = getBlockingMessages( translate );
 
 	let filteredHolds = listHolds;
-	if ( context === 'plugin-details' ) {
+	if ( context === 'plugin-details' || context === 'themes' ) {
 		filteredHolds = listHolds.filter( ( hold ) => hold !== 'NO_BUSINESS_PLAN' );
 	}
 
@@ -150,15 +150,17 @@ export const EligibilityWarnings = ( {
 				eventName="calypso_automated_transfer_eligibility_show_warnings"
 				eventProperties={ { context } }
 			/>
-			{ ! isPlaceholder && context === 'plugin-details' && hasBlockingHold( listHolds ) && (
-				<CompactCard>
-					<HardBlockingNotice
-						holds={ listHolds }
-						translate={ translate }
-						blockingMessages={ blockingMessages }
-					/>
-				</CompactCard>
-			) }
+			{ ! isPlaceholder &&
+				( context === 'plugin-details' || context === 'themes' ) &&
+				hasBlockingHold( listHolds ) && (
+					<CompactCard>
+						<HardBlockingNotice
+							holds={ listHolds }
+							translate={ translate }
+							blockingMessages={ blockingMessages }
+						/>
+					</CompactCard>
+				) }
 			{ ! isPlaceholder && context === 'plugin-details' && (
 				<CompactCard>
 					<div className="eligibility-warnings__header">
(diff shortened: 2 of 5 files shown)


BogdanUngureanu:
Great work, Dean! I made a few tests and while it looks and feels great for logged-in users, the modal is also displayed for logged-out users. :(

I think for logged out users we should keep the existing behavior.

BogdanUngureanu:
Another thing I noticed is the messaging: In the upgrade nudge we are saying that the Business price is 40$/month, but when the user clicks on the upgrade button they are sent to the checkout where we ask them to pay 300$/year.

There are two things here:
- The monthly price is incorrect
- We are saying that they'll have to pay 40$ but in fact, we asking them to pay 300$, without the option to switch to a monthly plan from the checkout page.

client/blocks/eligibility-warnings/hold-list.tsx
```diff
@@ -223,7 +223,7 @@ export const HoldList = ( { context, holds, isMarketplace, isPlaceholder, transl
 
 	return (
 		<>
-			{ ! isPlaceholder && context !== 'plugin-details' && (
+			{ ! isPlaceholder && context !== 'plugin-details' && context !== 'themes' && (
```
BogdanUngureanu, line 226: Non-blocking: Looks like ``context !== 'plugin-details' && context !== 'themes' ` ` condition is used in a few places. Would it make sense to extract `context !== 'plugin-details' && context !== 'themes' ` into a function so that we don't duplicate this condition in a lot of places? This way, it would give some semantic meaning?

e.g. something like `isContextEligibleForWarnings(context)` would be easier to understand (at least for me) than using the condition everywhere; plus we can add some docblocks about it's purpose.

dsas: The semantic meaning isn't "is context eligible for warnings" - warnings do get displayed still, they just get displayed differently (see [the issue](https://github.com/Automattic/dotcom-forge/issues/1299#issuecomment-1396767198)). HardBlockingNotice.

I'm not keen on most of the 'context' uses inside the component, it's a bit of a kludge. Perhaps ideally it would be broken down differently and we'd have a ThemeEligiblityWarnings and a PluginEligibilityWarnings etc components.

dsas:
> The monthly price is incorrect
> We are saying that they'll have to pay 40$ but in fact, we asking them to pay 300$, without the option to switch to a monthly plan from the checkout page.

The price displayed on the dialog is the per month, billed monthly price. Whereas what is in the basket is the annual product. This is the exact same approach as is used for plugins. It might have made more sense when that was implemented - until October (see https://github.com/Automattic/wp-calypso/pull/69643) users were able to decrease the subscription term in the basket, so you'd get the most expensive monthly price and were able to change it by increasing the term length.

I've implemented what @SaxonF designed, but I'm going to do some research to see what we should be doing here, I think the options are:
 1. Display monthly price for annual payment: "£nn per month, billed annually"
 2. Display annual price for annual payment
 3. Display monthly price for monthly payment, and add monthly product to basket. Possibly amend copy to say 'up to £xx'.
 4. Do something ridiculously complicated so the user can choose in the dialog.

I think 2. is the most accurate and we prefer customers to be on longer term plans, but perhaps saying "£20 per month, billed annually" is more enticing than £240 per year.

BogdanUngureanu:
> > The monthly price is incorrect
> > We are saying that they'll have to pay 40$ but in fact, we asking them to pay 300$, without the option to switch to a monthly plan from the checkout page.
> 
> The price displayed on the dialog is the per month, billed monthly price. Whereas what is in the basket is the annual product. This is the exact same approach as is used for plugins. It might have made more sense when that was implemented - until October (see #69643) users were able to decrease the subscription term in the basket, so you'd get the most expensive monthly price and were able to change it by increasing the term length.
> 
> I've implemented what @SaxonF designed, but I'm going to do some research to see what we should be doing here, I think the options are:
> 
> 1. Display monthly price for annual payment: "£nn per month, billed annually"
> 2. Display annual price for annual payment
> 3. Display monthly price for monthly payment, and add monthly product to basket. Possibly amend copy to say 'up to £xx'.
> 4. Do something ridiculously complicated so the user can choose in the dialog.
> 
> I think 2. is the most accurate and we prefer customers to be on longer term plans, but perhaps saying "£20 per month, billed annually" is more enticing than £240 per year.

Oh, yeah, maybe "incorrect" is not the proper word. I just wanted to highlight that the messaging might be confusing and create a different expectation.

> is the most accurate and we prefer customers to be on longer term plans, but perhaps saying "£20 per month, billed annually" is more enticing than £240 per year.

Yes, that's true, but at the same time, we should still allow them to use a monthly subscription from these upgrade nudges. 

Just to be clear, I don't think we should implement this right now, but it should be something to look into it in the near future, especially after we implement dotcom-forge/issues/1455 (Themes: Simplify the Upgrade Flow). 

I think updating the text to something like "£20 per month, billed annually" as you said should be enough for now. :)

dsas:
> I've implemented what @SaxonF designed, but I'm going to do some research to see what we should be doing here, I think the options are:

For plugins this was a deliberate design decision from @vinimotaa: p1664497052980909/1664477943.507869-slack-CKZHG0QCR Though that predates the changes to prevent term downgrades in checkout.

 The only code that displays prices prior to plans/checkout pages are:
 * Dotorg Plugins - behaves the same as implemented here (re-uses same code)
 * Premium plugins - allows you to choose between monthly and annual 
 * Premium themes - displays price but that's a one-off purchase
 * Domains - always annual.

----

I'm minded to leave this pr as it is and not block on this request, though I fully agree it's not a great experience.

I think we should either have a consistent approach, or clear guidelines on
 * How best to display pricing
 * Whether to link to the checkout or to /plans page.

The best place to achieve that is on P2, not on a pr.

client/blocks/eligibility-warnings/hold-list.tsx
dsas: Or maybe we'd just push the theme / plugin specific texts down into a child component and have EligibilityWarnings just pass the specific texts through 🤔 

Either way I think this is deserving of some refactoring and thinking in future.

mmtr (commented):
According to the discussion in pdqkMK-Su-p2, I think we want the nudge to behave like this (at least until we have an intermediary modal for the plan selection):
- It should mention the Business plan
- It should go to the checkout with the yearly Business plan automatically added
- It shouldn't mention any price information
- After checkout, user should be redirected back to this page

mmtr:
Uhm, I just realized this upgrade nudge is not an usual nudge, since it opens a new dialog.

The dialog seems redundant for sites with a custom domain – can we just redirect users to the checkout after they click on the nudge?

For sites without a custom domain, users might think that the domain will change after upgrading, but that won't happen. The domain will change only when they install a third party theme, and the same domain change warning will show up, so I wonder why we need to show the custom dialog before they upgrade. If we cannot upgrade and install the theme within the same flow, I think we should just redirects users to the checkout without showing the domain dialog.

dsas:
Closing this in favour of https://github.com/Automattic/wp-calypso/pull/72823 as discussed in paYJgx-2Uu-p2

PR #72348, opened 2023-01-19
Capitalize jsdoc array and object types

noahtallen:
#### Proposed Changes
Split from #72308, this PR would enforce capitalized types for `Object` in jsdoc. E.g.:

```js
/*
 * @param {Array} foo Some list
 * @returns {Object} Some object
/*
function arrToObj( foo ) {
  return {};
}
``` 
_Note that `Array` is capitalized by default._

Why? This is what WordPress coding style uses, because it's what the JSdoc standard uses. Thankfully, with a minimal rule change, this is completely autofixable. And if you accidentally write a lower-case variant, it should be auto-fixed in the editor if you lint on save :)

The counterpoint to this is the documentation here: https://github.com/gajus/eslint-plugin-jsdoc/tree/v39.0.0#why-not-capital-case-everything.

#### Testing Instructions
Verify code and rule changes make sense.


WunderBart (approved):
Fastest ~3k lines review ever 🚢 😎

tyxla (approved):
🚢 Thank you! 

I almost missed this as I was reviewing #72308 first. 

I appreciate it that we maintain consistency with Gutenberg's codebase. 

🚀

noahtallen:
I had to ignore a couple of existing problems in order to get everything passing. But I added `@todo` comments and links to relevant info in those cases. (There were only two or three of these.) I added all that to 05403f1, and kept the auto fix separate in 60f70ce

noahtallen:
I'll look into why translate is timing out next :)

worldomonation (approved):

noahtallen:
> I'll look into why translate is timing out next :)

I think it's just because the number of files changing is very large. The script loops through every changed file, and then every changed line. But I'll keep an eye on it after merging

Issue #72016, opened 2023-01-12
Writing Prompts - Visually show blockquote in reader feed


davemart-in:
## Description

Lets show the blockquote in the feed if possible vs. stripping the tags and converting it to text.

### Before

![Image](https://user-images.githubusercontent.com/5634774/212942114-513fc875-82bc-437d-935e-0b899eafa077.png)

### After

![Image](https://user-images.githubusercontent.com/5634774/212122781-1678485c-2802-4232-b9b1-e933f7bc17d8.png)

eoigal:
Think this can be closed. Nice job Eru.

PR #72361, opened 2023-01-20
Add/dailyprompt blockquote in reader

roo2:
#### Proposed Changes
 implements https://github.com/Automattic/wp-calypso/issues/72016

##### Design
Screen Shot 2023-02-02 at 1 38 02 pm
brief discussion in slack p1674180489954289-slack-C03NLNTPZ2T

##### Implementation
I added another case to the `chooseExcerpt` function that checks to see if the post has a `dailyprompt` tag.
If the post has a dailyprompt tag, it searches the content for starting with a blockquote.
If the content starts with a blockquote, it replaces the start of the excerpt with the html markup for a `blockquote` with the prompt content.
I also had to apply a class to the excerpt which makes it taller to acomodate the space used by the blockquote.

#### Testing Instructions

Go to /tag/dailyprompt and view the prompts. There is quite a variety of random different content being generated to test with.

### DIFF
--- a/client/blocks/reader-excerpt/index.jsx
+++ b/client/blocks/reader-excerpt/index.jsx
@@ -43,7 +43,7 @@ const chooseExcerpt = ( post ) => {
 	// Need to figure out if custom excerpt is different to better_excerpt
 	if ( post.excerpt.length > 0 ) {
 		// If the post is a dailyprompt, attempt to replace the prompt text with a pullquote.
-		if ( post.tags && post.hasOwnPromperty( 'dailyprompt' ) ) {
+		if ( post.tags && post.tags.hasOwnProperty( 'dailyprompt' ) ) {
 			const dom = domForHtml( post.content );
 			const promptQuote = dom.querySelector( '.is-reader > .wp-block-pullquote:first-child' );
 
--- a/client/blocks/reader-excerpt/index.jsx
+++ b/client/blocks/reader-excerpt/index.jsx
@@ -1,6 +1,8 @@
-import { trim } from 'lodash';
+import { trim, escapeRegExp } from 'lodash';
 import PropTypes from 'prop-types';
 import AutoDirection from 'calypso/components/auto-direction';
+import { domForHtml } from 'calypso/lib/post-normalizer/utils';
+
 import './style.scss';
 
 // Excerpt is set to use webkit-line-clamp to limit number of lines of text to show inside container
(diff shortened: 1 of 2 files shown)


roo2:
Updated to better implement the designs, this is ready for testing

eoigal:
I found a couple of cases where the response to the answer prompt is in the page title. I don't think we handle this too well in reader.

Screenshot 2023-02-02 at 14 20 17


From [here](https://aqsaayub542.wordpress.com/2023/02/02/im-afraid-of-getting-attached-to-people-right-this-time-i-think-i-need-to-more-social/).

client/blocks/reader-excerpt/index.jsx
```diff
@@ -37,9 +40,49 @@ const convertExcerptNewlinesToBreaks = ( excerpt ) => {
...
+
+/**
+ * Gets the writing prompt text which was inserted as a pullquote at the begining of the post's content.
+ *
+ * @param {Object} post the post object
+ * @returns writing prompt text
+ */
+const getDailyPromptText = ( post ) => {
+	const dom = domForHtml( post.content );
+	const promptQuote = dom.querySelector( '.wp-block-pullquote:first-child' );
```
eoigal, line 61:
```suggestion
	const promptQuote = dom.querySelector( '.wp-block-pullquote:first-of-type' );
```
I found this more reliable when parsing out the prompt quote

roo2: actually, I do want to assert that it's the first block in the post content as well, because someone may have deleted the writing prompt blockquote but have one of their own unrelated blockquotes embedded in the post

roo2:
> I found a couple of cases where the response to the answer prompt is in the page title. I don't think we handle this too well in reader.
> 
> Screenshot 2023-02-02 at 14 20 17
> 
> From [here](pesEpl-1x-p2).

I think that we shouldn't do something specific for writing prompts, because it's not how people are "supposed" to be using the feature anyway and it seems too niche to add a special case for this. 

But maybe there is a better way that we can display very short posts in reader, it seems that sometimes people write posts which are more like tweets than traditional blog posts. But maybe we can better present these "tweet posts", but without that being tied specifically to writing prompts?

client/blocks/reader-excerpt/index.jsx
```diff
@@ -1,6 +1,9 @@
-import { trim } from 'lodash';
+import classNames from 'classnames';
+import { trim, escapeRegExp } from 'lodash';
```
creativecoder, line 2: I believe we're trying to deprecate lodash in Calypso, so I'd avoid adding new imports, if you can. (p4TIVU-9GZ-p2)

creativecoder (commented):
Works as described!

image

This seem very specific, and I worry about the maintainability of code like this over time.

Do you think making this more generic and applying to any post with a blockquote or pullquote at the very beginning would make sense?

client/blocks/reader-excerpt/index.jsx
roo2: I think the alternative would be escaping it manually with `.replace` like https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript. I see there are a few other places in the code we still use `escapeRegExp` from `lodash` so I think that it's probably ok to add this and then when we get around to it, to replace them all with a helper function

roo2:
> Do you think making this more generic and applying to any post with a blockquote or pullquote at the very beginning would make sense?

hmmm yes I think that makes sense. It did end up being a bit fiddly code.

I'll 🚢 this first and then have a think/discussion in slack about if it should be generic.

PR #72363, opened 2023-01-20
Plugins: Improve performance of plugin selectors and convert to TypeScript

robertf4:
#### Proposed Changes

This fixes some performance issues with the plugins selectors by reducing runtime complexity, indexing, and memoizing where possible. As I was making this change I encountered a lot of confusion regarding the types of various objects which led to bugs, so I also switched the selectors and much of the code that calls them to TypeScript which cleared all of that up.

#### Testing Instructions

For every page be sure to also visit the production page and verify that everything looks identical. This PR is not supposed to change any functionality.

Check that the following works:
Plugin details page (/plugins/:plugin_slug)
* Activating/deactivating
* Enabling/disabling autoupdate
* Updating

Site details page (/plugins/manage/:site_slug)
* Activating/deactivating via toggle
* Enabling/disabling autoupdate via toggle
* Updating all plugins via primary update button
* Updating one plugin via update link
* Activating/deactivating via "edit all"
* Enabling/disabling autoupdate via "edit all"
* Updating via "edit all"
* Removing via "edit all"
* Removing via ellipsis

Plugins manage page
Site details page (/plugins/manag)
* Activating/deactivating via toggle
* Enabling/disabling autoupdate via toggle
* Updating all plugins via primary update button
* Updating one plugin via update link
* Activating/deactivating via "edit all"
* Enabling/disabling autoupdate via "edit all"
* Updating via "edit all"
* Removing via "edit all"
* Removing via ellipsis

Regression test the following:
* Jetpack plugin updates guided tour
* Marketplace
  * Checkout thank you
  * Plugin install
* Site settings (google analytics, cloudflare analytics, seo

#### Pre-merge Checklist

Related to 1202619025189113-as-1203496553274728

### DIFF
--- a/client/my-sites/plugins/main.jsx
+++ b/client/my-sites/plugins/main.jsx
@@ -30,7 +30,7 @@ import { recordGoogleEvent, recordTracksEvent } from 'calypso/state/analytics/ac
 import { appendBreadcrumb, updateBreadcrumbs } from 'calypso/state/breadcrumb/actions';
 import { getBreadcrumbs } from 'calypso/state/breadcrumb/selectors';
 import {
-	getPlugins,
+	getFilteredAndSortedPlugins,
 	isRequestingForSites,
 	isRequestingForAllSites,
 	requestPluginsError,
@@ -560,8 +560,8 @@ export default flow(
 			const selectedSiteId = getSelectedSiteId( state );
 			const visibleSiteIds = siteObjectsToSiteIds( getVisibleSites( sites ) ) ?? [];
 			const siteIds = siteObjectsToSiteIds( sites ) ?? [];
-			const pluginsWithUpdates = getPlugins( state, siteIds, 'updates' );
-			const allPlugins = getPlugins( state, siteIds, 'all' );
+			const pluginsWithUpdates = getFilteredAndSortedPlugins( state, siteIds, 'updates' );
+			const allPlugins = getFilteredAndSortedPlugins( state, siteIds, 'all' );
 			const jetpackNonAtomic =
 				isJetpackSite( state, selectedSiteId ) && ! isAtomicSite( state, selectedSiteId );
 			const hasManagePlugins =
@@ -586,8 +586,8 @@ export default flow(
 					selectedSite && canJetpackSiteUpdateFiles( state, selectedSiteId ),
 				wporgPlugins: getAllWporgPlugins( state ),
 				isRequestingSites: isRequestingSites( state ),
-				currentPlugins: getPlugins( state, siteIds, filter ),
-				currentPluginsOnVisibleSites: getPlugins( state, visibleSiteIds, filter ),
+				currentPlugins: getFilteredAndSortedPlugins( state, siteIds, filter ),
+				currentPluginsOnVisibleSites: getFilteredAndSortedPlugins( state, visibleSiteIds, filter ),
 				pluginUpdateCount: pluginsWithUpdates && pluginsWithUpdates.length,
 				pluginsWithUpdates,
 				allPluginsCount: allPlugins && allPlugins.length,
--- a/client/my-sites/plugins/plugin-details-header/index.jsx
+++ b/client/my-sites/plugins/plugin-details-header/index.jsx
@@ -17,6 +17,21 @@ const PluginDetailsHeader = ( { plugin, isPlaceholder, isJetpackCloud } ) => {
 
 	const selectedSite = useSelector( getSelectedSite );
 
+	// This gets the version of the plugin on the last site or
+	// the wporg version if that doesn't exist. This is preserving
+	// some pre-existing behavior during a refactor and doesn't
+	// necessarily have to remain this way in the future.
+	let version = plugin?.version;
+	if ( plugin.sites ) {
+		const sites = Object.values( plugin.sites );
+		for ( let i = sites.length - 1; i >= 0; i-- ) {
+			if ( sites[ i ].version ) {
+				version = sites[ i ].version;
+				break;
+			}
+		}
+	}
+
 	if ( isPlaceholder ) {
 		return <PluginDetailsHeaderPlaceholder />;
 	}
@@ -74,7 +89,7 @@ const PluginDetailsHeader = ( { plugin, isPlaceholder, isJetpackCloud } ) => {
 				</div>
 				<div className="plugin-details-header__info">
 					<div className="plugin-details-header__info-title">{ translate( 'Version' ) }</div>
-					<div className="plugin-details-header__info-value">{ plugin.version }</div>
+					<div className="plugin-details-header__info-value">{ version }</div>
 				</div>
 				{ Boolean( plugin.active_installs ) && (
 					<div className="plugin-details-header__info">
--- a/client/my-sites/plugins/plugin-management-v2/test/plugin-row-formatter.test.tsx
+++ b/client/my-sites/plugins/plugin-management-v2/test/plugin-row-formatter.test.tsx
@@ -18,7 +18,7 @@ const initialReduxState = {
 	siteConnection: { items: { [ site.ID ]: true } },
 	sites: { items: { [ site.ID ]: site } },
 	currentUser: {
-		capabilities: {},
+		capabilities: { [ site.ID ]: { manage_options: true } },
 	},
 	plugins: {
 		installed: {
@@ -106,6 +106,7 @@ describe( '<PluginRowFormatter>', () => {
 
 	test( 'should render correctly and show site count(plural) on small screen', () => {
 		props.columnKey = 'sites';
+		const sitesOriginal = plugin.sites;
 		plugin.sites = {
 			[ site.ID ]: { ID: site.ID, canUpdateFiles: true },
 			[ site.ID + 1 ]: { ID: site.ID + 1, canUpdateFiles: true },
@@ -116,9 +117,15 @@ describe( '<PluginRowFormatter>', () => {
 			`Installed on ${ Object.keys( plugin.sites ).length } sites`
 		);
 		expect( autoManagedSite ).toBeInTheDocument();
+		plugin.sites = sitesOriginal;
 	} );
 
-	test( 'should render correctly and show activate and toggle checked value', async () => {
+	// These skipped test nevers actually worked, it just tests that the HTML toggle actually toggled.
+	// To verify comment out the entirety of the toggle callbacks in the relevant components and observe
+	// that the test still pass.
+	// They were passing because their toggle code didn't execute. I changed some test data which caused
+	// the toggle code to actually execute, and now it needs to be rewritten to use nock to simulate the server calls.
+	test.skip( 'should render correctly and show activate and toggle checked value', async () => {
 		props.columnKey = 'activate';
 		const { container } = render( <PluginRowFormatter { ...props } /> );
 
@@ -131,7 +138,7 @@ describe( '<PluginRowFormatter>', () => {
 		expect( toggleButton ).toHaveProperty( 'checked', true );
 	} );
 
-	test( 'should render correctly and show auto-update and toggle checked value', async () => {
+	test.skip( 'should render correctly and show auto-update and toggle checked value', async () => {
 		props.columnKey = 'autoupdate';
 		const { container } = render( <PluginRowFormatter { ...props } /> );
 
--- a/client/my-sites/plugins/plugin-management-v2/test/remove-plugin.test.tsx
+++ b/client/my-sites/plugins/plugin-management-v2/test/remove-plugin.test.tsx
@@ -22,7 +22,7 @@ const props = {
 const initialState = {
 	sites: { items: { [ site.ID ]: site } },
 	currentUser: {
-		capabilities: {},
+		capabilities: { [ site.ID ]: { manage_options: true } },
 	},
 	plugins: {
 		installed: {
@@ -53,8 +53,13 @@ describe( '<RemovePlugin>', () => {
 	beforeAll( () => {
 		nock( 'https://public-api.wordpress.com:443' )
 			.persist()
+			.post( `/rest/v1.1/sites/${ site.ID }/plugins/${ plugin.id }`, { active: false } )
+			.reply( 200, plugin );
+
+		nock( 'https://public-api.wordpress.com:443' )
+			.persist()
 			.post( `/rest/v1.1/sites/${ site.ID }/plugins/${ plugin.id }/delete` )
-			.reply( 200 );
+			.reply( 200, plugin );
 	} );
 
 	afterAll( () => {
--- a/client/my-sites/plugins/plugin-management-v2/test/update-plugin.test.tsx
+++ b/client/my-sites/plugins/plugin-management-v2/test/update-plugin.test.tsx
@@ -13,7 +13,7 @@ import { site, plugin } from './utils/constants';
 const initialState = {
 	sites: { items: { [ site.ID ]: site } },
 	currentUser: {
-		capabilities: {},
+		capabilities: { [ site.ID ]: { manage_options: true } },
 	},
 	plugins: {
 		installed: {
@@ -54,10 +54,12 @@ describe( '<UpdatePlugin>', () => {
 
 		expect(
 			container.getElementsByClassName( 'update-plugin__current-version' )[ 0 ].textContent
-		).toEqual( plugin.version );
+		).toEqual( plugin.sites[ site.ID ].version );
 
 		const [ updateButton ] = container.getElementsByClassName( 'update-plugin__new-version' );
-		expect( updateButton.textContent ).toEqual( `Update to ${ plugin.update.new_version }` );
+		expect( updateButton.textContent ).toEqual(
+			`Update to ${ plugin.sites[ site.ID ].update.new_version }`
+		);
 
 		await userEvent.click( updateButton );
 		expect( props.updatePlugin ).toHaveBeenCalledTimes( 1 );
--- a/client/my-sites/plugins/plugin-management-v2/update-plugin/index.tsx
+++ b/client/my-sites/plugins/plugin-management-v2/update-plugin/index.tsx
@@ -5,8 +5,6 @@ import { useTranslate } from 'i18n-calypso';
 import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { UPDATE_PLUGIN } from 'calypso/lib/plugins/constants';
-import { siteObjectsToSiteIds } from 'calypso/my-sites/plugins/utils';
-import { getPluginOnSites } from 'calypso/state/plugins/installed/selectors';
 import getSites from 'calypso/state/selectors/get-sites';
 import PluginActionStatus from '../plugin-action-status';
 import { getAllowedPluginActions } from '../utils/get-allowed-plugin-actions';
@@ -28,37 +26,17 @@ export default function UpdatePlugin( { plugin, selectedSite, className, updateP
 	const allSites = useSelector( getSites );
 	const state = useSelector( ( state ) => state );
 
-	const getPluginSites = ( plugin: PluginComponentProps ) => {
-		return Object.keys( plugin.sites ).map( ( siteId ) => {
-			const site = allSites.find( ( s ) => s?.ID === parseInt( siteId ) );
-			return {
-				...site,
-				...plugin.sites[ siteId ],
-			} as any; // This must be cast as any until this file is updated to work with the selectors in state/plugins/installed/selectors
-		} );
-	};
-
-	const sites = getPluginSites( plugin );
-	const siteIds = siteObjectsToSiteIds( sites );
-	const pluginsOnSites: any = getPluginOnSites( state, siteIds, plugin?.slug );
-
-	const currentVersions = sites
-		.map( ( site ) => {
-			const siteId = selectedSite ? selectedSite.ID : site.ID;
-			const sitePlugin = pluginsOnSites?.sites[ siteId ];
-			return sitePlugin?.version;
-		} )
-		.filter( ( version ) => version );
-
-	const updatedVersions = sites
-		.map( ( site ) => {
-			const siteId = selectedSite ? selectedSite.ID : site.ID;
-			const sitePlugin = pluginsOnSites?.sites[ siteId ];
-			return sitePlugin?.update?.new_version;
-		} )
-		.filter( ( version ) => version );
+	const updatedVersions = selectedSite
+		? [ plugin.sites[ selectedSite.ID ].update?.new_version ]
+		: Object.values( plugin.sites )
+				.map( ( site ) => site.update?.new_version )
+				.filter( Boolean );
 
 	const currentVersionsRange = useMemo( () => {
+		const currentVersions = selectedSite
+			? [ plugin.sites[ selectedSite.ID ]?.version ]
+			: Object.values( plugin.sites ).map( ( site ) => site.version );
+
 		const versions = [
 			// We want to remove the duplicated versions in the array, because if multiple sites have
 			// the same plugin version, we don't want to display the range.
@@ -78,13 +56,16 @@ export default function UpdatePlugin( { plugin, selectedSite, className, updateP
 			min: versions[ 0 ],
 			max: versions.length > 1 ? versions[ versions.length - 1 ] : null,
 		};
-	}, [ currentVersions ] );
+	}, [ plugin, selectedSite ] );
 
-	const hasUpdate = sites.some( ( site ) => {
-		const siteId = selectedSite ? selectedSite.ID : site.ID;
-		const sitePlugin = pluginsOnSites?.sites[ siteId ];
-		return sitePlugin?.update?.new_version && site.canUpdateFiles;
-	} );
+	const hasUpdate = selectedSite
+		? plugin.sites[ selectedSite.ID ]?.update?.new_version &&
+		  allSites.find( ( site ) => site && site.ID === selectedSite.ID )?.canUpdateFiles
+		: Object.entries( plugin.sites ).some(
+				( [ siteId, site ] ) =>
+					site?.update?.new_version &&
+					allSites.find( ( site ) => site && site.ID === Number( siteId ) )?.canUpdateFiles
+		  );
 
 	const allowedActions = getAllowedPluginActions( plugin, state, selectedSite );
 
@@ -97,7 +78,7 @@ export default function UpdatePlugin( { plugin, selectedSite, className, updateP
 			status.pluginId === plugin.id &&
 			status.action === UPDATE_PLUGIN &&
 			// Filter out status based on selected site if any
-			( selectedSite ? parseInt( status.siteId ) === selectedSite.ID : true )
+			( selectedSite ? Number( status.siteId ) === selectedSite.ID : true )
 	);
 
 	if ( ! allowedActions?.autoupdate ) {
--- a/client/my-sites/plugins/plugins-list/index.tsx
+++ b/client/my-sites/plugins/plugins-list/index.tsx
@@ -1,8 +1,7 @@
 import { recordTracksEvent } from '@automattic/calypso-analytics';
 import { WPCOM_FEATURES_MANAGE_PLUGINS } from '@automattic/calypso-products';
-import { localize } from 'i18n-calypso';
+import { localize, translate } from 'i18n-calypso';
 import { isEqual, reduce } from 'lodash';
-import PropTypes from 'prop-types';
 import { Component } from 'react';
 import { connect } from 'react-redux';
 import QueryProductsList from 'calypso/components/data/query-products-list';
@@ -30,48 +29,85 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
 import { getSelectedSite, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 import PluginManagementV2 from '../plugin-management-v2';
 import { getPluginActionDailogMessage, getSitePlugin, handleUpdatePlugins } from '../utils';
+import type { SiteDetails } from '@automattic/data-stores';
+import type { Plugin, PluginStatus } from 'calypso/state/plugins/installed/types';
 
 import './style.scss';
 
-function checkPropsChange( nextProps, propArr ) {
-	let i;
-	let prop;
-
-	for ( i = 0; i < propArr.length; i++ ) {
-		prop = propArr[ i ];
-
-		if ( ! isEqual( nextProps[ prop ], this.props[ prop ] ) ) {
-			return true;
-		}
-	}
-	return false;
+interface Props {
+	hasManagePlugins?: boolean;
+	header: string;
+	inProgressStatuses: PluginStatus[];
+	isPlaceholder: boolean;
+	plugins: Plugin[];
+	selectedSite?: SiteDetails;
+	selectedSiteSlug?: string;
+	siteIsAtomic?: boolean;
+	siteIsJetpack?: boolean;
+	searchTerm: string;
+	recordGoogleEvent: typeof recordGoogleEvent;
+	allSites: ReturnType< typeof getSites >;
+	pluginsOnSites: ReturnType< typeof getPluginsOnSites >;
+	removePluginStatuses: typeof removePluginStatuses;
+	translate: typeof translate;
+	removePlugin: typeof removePlugin;
+	updatePlugin: typeof updatePlugin;
+	activatePlugin: typeof activatePlugin;
+	deactivatePlugin: typeof deactivatePlugin;
+	enableAutoupdatePlugin: typeof enableAutoupdatePlugin;
+	warningNotice: typeof warningNotice;
+	disableAutoupdatePlugin: typeof disableAutoupdatePlugin;
+	isJetpackCloud: boolean;
+	isLoading: boolean;
+	requestPluginsError: boolean;
 }
 
-export class PluginsList extends Component {
-	static propTypes = {
-		plugins: PropTypes.arrayOf(
-			PropTypes.shape( {
-				sites: PropTypes.object,
-				slug: PropTypes.string,
-				name: PropTypes.string,
-			} )
-		).isRequired,
-		hasManagePlugins: PropTypes.bool,
-		header: PropTypes.string.isRequired,
-		isPlaceholder: PropTypes.bool.isRequired,
-		selectedSite: PropTypes.object,
-		selectedSiteSlug: PropTypes.string,
-		siteIsAtomic: PropTypes.bool,
-		siteIsJetpack: PropTypes.bool,
-	};
+interface State {
+	disconnectJetpackNotice: boolean;
+	removeJetpackNotice: boolean;
+	bulkManagementActive: boolean;
+	selectedPlugins: { [ pluginSlug: string ]: boolean };
+}
 
+type Action =
+	| 'updating'
+	| 'activating'
+	| 'deactivating'
+	| 'enablingAutoupdates'
+	| 'disablingAutoupdates'
+	| 'removing';
+
+type BulkAction =
+	| 'activate'
+	| 'deactivate'
+	| 'enableAutoupdates'
+	| 'disableAutoupdates'
+	| 'update'
+	| 'remove';
+
+export class PluginsList extends Component< Props, State > {
 	static defaultProps = {
+		// eslint-disable-next-line @typescript-eslint/no-empty-function
 		recordGoogleEvent: () => {},
 	};
 
-	shouldComponentUpdate( nextProps, nextState ) {
-		const propsToCheck = [ 'plugins', 'sites', 'selectedSite' ];
-		if ( checkPropsChange.call( this, nextProps, propsToCheck ) ) {
+	checkPropsChange( nextProps: Props, propArr: ( keyof Props )[] ) {
+		let i;
+		let prop;
+
+		for ( i = 0; i < propArr.length; i++ ) {
+			prop = propArr[ i ];
+
+			if ( ! isEqual( nextProps[ prop ], this.props[ prop ] ) ) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	shouldComponentUpdate( nextProps: Props, nextState: State ) {
+		const propsToCheck = [ 'plugins', 'sites', 'selectedSite' ] as ( keyof Props )[];
+		if ( this.checkPropsChange( nextProps, propsToCheck ) ) {
 			return true;
 		}
 
@@ -115,18 +151,18 @@ export class PluginsList extends Component {
 		this.maybeShowRemoveNotice();
 	}
 
-	state = {
+	state: State = {
 		disconnectJetpackNotice: false,
 		removeJetpackNotice: false,
 		bulkManagementActive: false,
 		selectedPlugins: {},
 	};
 
-	isSelected = ( { slug } ) => {
+	isSelected = ( { slug }: { slug: string } ) => {
 		return !! this.state.selectedPlugins[ slug ];
 	};
 
-	togglePlugin = ( plugin ) => {
+	togglePlugin = ( plugin: Plugin ) => {
 		const { slug } = plugin;
 		const { selectedPlugins } = this.state;
 		const oldValue = selectedPlugins[ slug ];
@@ -138,20 +174,20 @@ export class PluginsList extends Component {
 		this.props.recordGoogleEvent( 'Plugins', eventAction, 'Plugin Name', slug );
 	};
 
-	canBulkSelect( plugin ) {
+	canBulkSelect( plugin: Plugin ) {
 		const { autoupdate: canAutoupdate, activation: canActivate } =
 			this.getAllowedPluginActions( plugin );
 		return canAutoupdate || canActivate;
 	}
 
-	setBulkSelectionState = ( plugins, selectionState ) => {
+	setBulkSelectionState = ( plugins: Plugin[], selectionState: boolean ) => {
 		const slugsToBeUpdated = reduce(
 			plugins,
 			( slugs, plugin ) => {
-				slugs[ plugin.slug ] = this.canBulkSelect( plugin ) && selectionState;
+				slugs[ plugin.slug ] = !! ( this.canBulkSelect( plugin ) && selectionState );
 				return slugs;
 			},
-			{}
+			{} as { [ pluginSlug: string ]: boolean }
 		);
 
 		this.setState( {
@@ -160,7 +196,7 @@ export class PluginsList extends Component {
 	};
 
 	filterSelection = {
-		active( plugin ) {
+		active: ( plugin: Plugin ) => {
 			if ( this.isSelected( plugin ) ) {
 				return Object.keys( plugin.sites ).some( ( siteId ) => {
 					const sitePlugin = getSitePlugin( plugin, siteId, this.props.pluginsOnSites );
@@ -169,7 +205,7 @@ export class PluginsList extends Component {
 			}
 			return false;
 		},
-		inactive( plugin ) {
+		inactive: ( plugin: Plugin ) => {
 			if ( this.isSelected( plugin ) && plugin.slug !== 'jetpack' ) {
 				return Object.keys( plugin.sites ).some( ( siteId ) => {
 					const sitePlugin = getSitePlugin( plugin, siteId, this.props.pluginsOnSites );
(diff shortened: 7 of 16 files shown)


client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -60,8 +62,8 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 		wpComPluginData ? wpComPluginData.software_slug || wpComPluginData.org_slug : productSlugs[ i ]
 	);
 
-	const pluginsOnSite: [] = useSelector( ( state ) =>
-		getPluginsOnSite( state, siteId, softwareSlugs )
+	const pluginsOnSite = useSelector( ( state ) =>
+		siteId ? getPluginsOnSite( state, siteId, softwareSlugs ) : []
```
tyxla: Won't `getPluginsOnSite()` always return an array, given that internally it returns the result of an `Array.prototype.map()` on `softwareSlugs`, which is also a result of `Array.prototype.map()` on its own?

client/my-sites/plugins/main.jsx
```diff
@@ -79,12 +83,14 @@ export class PluginsMain extends Component {
...
-			}
-		} );
+		if ( currentPlugins.length <= WPORG_PLUGIN_COUNT_CUTOFF ) {
+			currentPlugins.forEach( ( plugin ) => {
+				const pluginData = this.props.wporgPlugins?.[ plugin.slug ];
+				if ( ! pluginData ) {
+					this.props.wporgFetchPluginData( plugin.slug );
+				}
+			} );
+		}
```
tyxla: This feels like a regression to me. There are visible regressions that result from this, and I think they're not acceptable. Here's a comparison:

Site with <= 20 plugins:

Screenshot 2023-01-24 at 10 39 44

Site with more than 20 plugins:

Screenshot 2023-01-24 at 10 39 20

You can see that there are differences in the plugin title (Akismet Anti-Spam vs Akismet Spam Protection for example), icons are missing, etc. I think this is not acceptable.

Furthermore, how did we decide at what number of WP.org plugin requests the performance degrades? For some machines, it could be 20, but for slower configurations, it could be 10. The 20 number feels arbitrary and I'm in doubt it will work well in all scenarios.

There are various ways we could optimize for this scenario. I'll offer some solutions below, and I believe we could use any combination of them to approach addressing this issue.

One way to solve this on the client side would be to defer the loading of WP.org data, rather than totally not loading it. That would mean, not sending requests immediately, but debouncing them, and in the meantime, displaying an indicator that the plugin data is still loading.

Another approach we could consider on the client side would be leveraging caching. It seems like the WP.org data is not cached locally, and that feels off. I believe we should be caching it, and that would make the slower loading run way less frequently. 

I can think of ways to solve this on the server as well: we could work on a single endpoint that returns that information for plugins with provided slugs. To optimize that endpoint, we could cache the plugin info for some longer time, including the version of the plugin in the cache key so it would be invalidated when a new version is released. A single request would really boost the performance IMHO, so it would be worth the time.

To summarize, I believe this should be part of a separate PR since it feels like an arbitrary fix that many would disagree with. I'm happy to aid in getting that resolved with whatever tools and leverages I can.

Let me know what you think!

client/my-sites/plugins/plugin-details-header/index.jsx
```diff
@@ -16,6 +16,21 @@ const PluginDetailsHeader = ( { plugin, isPlaceholder, isJetpackCloud } ) => {
...
+	if ( plugin.sites ) {
+		const sites = Object.values( plugin.sites );
+		for ( let i = sites.length - 1; i >= 0; i-- ) {
+			if ( sites[ i ].version ) {
+				version = sites[ i ].version;
+				break;
+			}
+		}
+	}
+
```
tyxla, line 34: This needs to be in a dedicated selector IMHO.

client/my-sites/plugins/plugin-details-header/index.jsx
```diff
@@ -16,6 +16,21 @@ const PluginDetailsHeader = ( { plugin, isPlaceholder, isJetpackCloud } ) => {
...
 	const selectedSite = useSelector( getSelectedSite );
 
+	// This gets the version of the plugin on the last site or
+	// the wporg version if that doesn't exist. This is preserving
+	// some pre-existing behavior during a refactor and doesn't
+	// necessarily have to remain this way in the future.
+	let version = plugin?.version;
+	if ( plugin.sites ) {
+		const sites = Object.values( plugin.sites );
+		for ( let i = sites.length - 1; i >= 0; i-- ) {
```
tyxla, line 27: We could also `reverse()` the array and do a regular loop with incrementing key.

client/my-sites/plugins/plugin-management-v2/types.ts
```diff
@@ -81,3 +67,10 @@ export type PluginActionStatusMessage = {
 		[ PLUGIN_INSTALLATION_UP_TO_DATE ]?: ReactNode;
 	};
 };
+
+export type ExtendedPlugin = Plugin & {
```
tyxla: I believe the intention of this type is worth documenting. Also, any type that extends another type is an extended type, so naming kind of feels redundant. Do you think we can figure out a more descriptive name?

client/my-sites/plugins/plugin-management-v2/plugin-details-v2/index.tsx
```diff
@@ -76,7 +77,11 @@ export default function PluginDetailsV2( {
 
 	return (
 		<div className="plugin-details-v2">
-			<QueryJetpackPlugins siteIds={ siteIds } />
+			{ siteIds.length === 1 ? (
+				<QueryJetpackPlugins siteIds={ siteIds } />
+			) : (
+				<QueryAllJetpackSitesPlugins />
+			) }
```
tyxla: This feels like work for another PR. How does it relate to this PR? Seems like we could move it out.

client/my-sites/plugins/plugin-management-v2/plugin-manage-subscription/index.tsx
```diff
@@ -25,7 +25,7 @@ export default function PluginManageSubcription( { site, plugin }: Props ): Reac
 	const currentPurchase: Purchase = getPluginPurchased(
 		plugin,
 		purchases,
-		plugin.isMarketplaceProduct
+		!! plugin.isMarketplaceProduct
```
tyxla: Won't this be boolean already?

client/my-sites/plugins/plugin-management-v2/test/remove-plugin.test.tsx
```diff
@@ -51,10 +51,15 @@ describe( '<RemovePlugin>', () => {
...
+		nock( 'https://public-api.wordpress.com:443' )
+			.persist()
+			.post( `/rest/v1.1/sites/${ site.ID }/plugins/${ plugin.id }`, { active: false } )
+			.reply( 200, plugin );
+
 		nock( 'https://public-api.wordpress.com:443' )
 			.persist()
 			.post( `/rest/v1.1/sites/${ site.ID }/plugins/${ plugin.id }/delete` )
-			.reply( 200 );
+			.reply( 200, plugin );
```
tyxla, line 62: Is it intentional that the plugin deletion endpoint returns the plugin object after successful deletion?

client/my-sites/plugins/plugin-management-v2/test/update-plugin.test.tsx
```diff
@@ -54,10 +54,12 @@ describe( '<UpdatePlugin>', () => {
 
 		expect(
 			container.getElementsByClassName( 'update-plugin__current-version' )[ 0 ].textContent
-		).toEqual( plugin.version );
+		).toEqual( plugin.sites[ site.ID ].version );
 
 		const [ updateButton ] = container.getElementsByClassName( 'update-plugin__new-version' );
-		expect( updateButton.textContent ).toEqual( `Update to  ${ plugin.update.new_version }` );
+		expect( updateButton.textContent ).toEqual(
+			`Update to  ${ plugin.sites[ site.ID ].update.new_version }`
```
tyxla: I know this was here before, but I find it odd that there's a double space after "Update to". Feels like we something we can follow-up with.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
+import { createSelector } from '@automattic/state-utils';
```
tyxla, line 1: It would have been useful to `git mv` this file so we preserve the history. That way it will be marked as a new file and any prior history will be lost unless traced to this PR.

client/state/plugins/installed/test/selectors.ts
```diff
@@ -1,5 +1,5 @@
 import deepFreeze from 'deep-freeze';
-import { pick } from 'lodash';
+import { omit, pick } from 'lodash';
```
tyxla: Ideally, we should be getting rid of Lodash and not introducing new usage. Could we do something alternative here, like `Object.entries()` + a `.filter()` to remove the properties we want to be omitted?

client/state/plugins/installed/types.ts
```diff
@@ -0,0 +1,65 @@
...
+	update?: PluginUpdate;
+	wporg?: boolean;
+};
+
+export type PluginSites = {
+	[ key: string ]: {
+		active: boolean;
+		autoupdate: boolean;
+		update?: PluginUpdate;
+		version: string;
```
tyxla: All those fields can be found in the `InstalledPluginData` type. Is this an opportunity to create a new type out of this and make `InstalledPluginData` an intersection type?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+	( state: AppState, siteIds: number[], pluginFilter?: PluginFilter ) => {
+		const allPluginsIndexedBySiteId = getAllPluginsIndexedBySiteId( state );
+
+		// Properties on the objects in allPluginsIndexedBySiteId will be modified and the
+		// selector memoization always returns the same object, so use cloneDeep to avoid
+		// altering it for everyone.
+		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = cloneDeep(
+			siteIds
+				.map( ( siteId: number ) => allPluginsIndexedBySiteId[ siteId ] )
+				.filter( ( plugin ) => !! plugin )
```
tyxla: This might as well be:

```suggestion
				.filter( Boolean )
```

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+);
+
+export const getFilteredAndSortedPlugins = createSelector(
+	( state: AppState, siteIds: number[], pluginFilter?: PluginFilter ) => {
+		const allPluginsIndexedBySiteId = getAllPluginsIndexedBySiteId( state );
+
+		// Properties on the objects in allPluginsIndexedBySiteId will be modified and the
+		// selector memoization always returns the same object, so use cloneDeep to avoid
+		// altering it for everyone.
+		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = cloneDeep(
```
tyxla, line 165: Could you please elaborate on the `cloneDeep` change? From what you're describing it feels suboptimal that we're using memoization but still we're overriding it here. Perhaps we could fix this by avoiding mutation in the first place?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+			siteIds
+				.map( ( siteId: number ) => allPluginsIndexedBySiteId[ siteId ] )
+				.filter( ( plugin ) => !! plugin )
+				.reduce( ( accumulator, current ) => ( { ...accumulator, ...current } ), {} )
```
tyxla, line 169: Won't that already create a new version of the plugin objects, making the `cloneDeep()` change unnecessary?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
+import { createSelector } from '@automattic/state-utils';
+import { cloneDeep, filter, get, omit, pick, some, sortBy } from 'lodash';
```
tyxla: Same as above for introducing `_.omit()` usage.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+	]
+);
+
+/**
+ * The plugins here differ from the plugin objects found on state.plugins.installed.plugins in that each plugin
+ * object has data from all the different sites on it, whereas on state.plugins.installed.plugins they only have
+ * the state for the site which index they are under. The objects here are the same as those returned from
+ * getAllPluginsIndexedByPluginSlug, except they are indexed by siteId.
+ */
+export const getAllPluginsIndexedBySiteId = createSelector(
```
tyxla, line 132: I'd suggest trying to move all the selector changes outside of this PR. Creating new selectors and refactoring existing selectors to work with the new selectors, alongside a TS migration and related changes just feels too much for a single PR.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+
+const _filters: { [ key in PluginFilter ]: ( plugin: Plugin ) => boolean } = {
+	none: function () {
+		return false;
+	},
+	all: function () {
+		return true;
+	},
+	active: function ( plugin: Plugin ) {
+		return !! (
```
tyxla: `_.some()` will always return a boolean, right? The need for explicit casting to boolean comes from the fact that `statusRecentlyChanged` could be missing, right? So it might make sense to move the explicit boolean cast there.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+			( plugins: { [ key: string ]: Plugin }, siteId ) => {
+				if ( isRequesting( state, siteId ) ) {
+					return plugins;
+				}
+
+				const installedPlugins = state.plugins.installed.plugins as InstalledPlugins;
+
+				const pluginsForSite = installedPlugins[ siteId ] || [];
+				pluginsForSite.forEach( ( plugin: InstalledPluginData ) => {
+					const sitePluginProperties = [ 'active', 'autoupdate', 'update', 'version' ];
```
tyxla: Since this array is declared multiple times, it might make sense to declare it outside of this context and reuse it universally.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = cloneDeep(
+			siteIds
+				.map( ( siteId: number ) => allPluginsIndexedBySiteId[ siteId ] )
+				.filter( ( plugin ) => !! plugin )
+				.reduce( ( accumulator, current ) => ( { ...accumulator, ...current } ), {} )
+		);
+
+		// Filter the sites object on the plugins so that only data for the requested siteIds is present
+		for ( const pluginSlug of Object.keys( allPluginsForSites ) ) {
+			allPluginsForSites[ pluginSlug ].sites = pick(
```
tyxla, line 174: This is the place we're doing the mutation I was talking about above. Ideally, we could seek ways to do this without mutating the objects, for example by returning new objects, and that would allow us to avoid the clone call above.

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+	( state: AppState ) => [ getAllPluginsIndexedBySiteId( state ) ],
+	( state: AppState, siteIds: number[], pluginFilter?: PluginFilter ) => {
+		return [ siteIds, pluginFilter ].flat().join( '-' );
+	}
+);
+
+export function getPluginsWithUpdates( state: AppState, siteIds: number[] ) {
+	return filter( getFilteredAndSortedPlugins( state, siteIds, undefined ), _filters.updates ).map(
+		( plugin ) => ( {
+			...plugin,
```
tyxla, line 196: We used to provide `version` here, but we're no longer doing that. Is that intentional?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+			type: 'plugin',
+		} )
+	);
+}
+
+export function getPluginsOnSites( state: AppState, plugins: Plugin[] ) {
+	return Object.values( plugins ).reduce(
+		( acc: { [ pluginSlug: string ]: Plugin }, plugin: Plugin ) => {
+			const siteIds = Object.keys( plugin.sites ).map( ( x ) => Number( x ) );
```
tyxla, line 205: Converting site IDs to numbers seems like something we're doing a few times, should we abstract a function that manages it?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+
+	if ( ! plugin ) {
+		return undefined;
+	}
+
+	for ( const siteId of siteIds ) {
+		if ( plugin.sites[ siteId ] ) {
+			return plugin;
+		}
+	}
```
tyxla, line 227: What was wrong with the `Array.prototype.find()` implementation here?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: string ) {
+	const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
+
+	return plugin && plugin.sites[ siteId ]
+		? // Because this is a plugin for a single site context only the data for the selected
+		  // site is included. When I changed this file to TypeScript I found that some code
+		  // assumes it will be on the plugin object, while other code assumes it will be on the
+		  // sites object, so I included both.
+		  {
+				...omit( plugin, [ 'sites' ] ),
```
tyxla: The omission of a single key is extremely simple with vanilla JS, could we do that instead of calling `_.omit()`:

```
const { sites, ...pluginWithoutSites } = plugin;

File: client/state/plugins/installed/README.md [hunk 25]
```diff
@@ -30,9 +30,10 @@ A module for managing installed plugins on connected sites.
 
 Returns the state of the network request for fetching all available plugins on all Jetpack sites of the current user.
 
-### `getPlugins( state: Object, siteIds: Array, pluginFilter: Object )`
+### `getFilteredAndSortedPlugins( state: Object, siteIds: Array, pluginFilter: Object )`
```
tyxla, line 33: Since we introduced a bunch of additional selectors and changes to existing ones, do we need to work on the README as well?

client/state/plugins/installed/selectors.ts
```diff
@@ -0,0 +1,382 @@
...
+
+				return pluginsIndexedBySiteId;
+			},
+			{}
+		);
+	},
+	( state ) => [ getAllPluginsIndexedByPluginSlug( state ), getSiteIdsThatHavePlugins( state ) ]
+);
+
+export const getFilteredAndSortedPlugins = createSelector(
```
tyxla, line 158: I think it makes sense that the complex new selectors need unit tests. This sounds like work for another PR, but also it provides food for thought as to whether it's not a good idea to just work on these selectors in separate PR(s).

client/my-sites/plugins/utils.js
```diff
@@ -128,20 +128,18 @@ export function handleUpdatePlugins( plugins, updateAction, pluginsOnSites ) {
...
+			if ( ! site.update || ! site.update?.new_version ) {
+				continue;
+			}
+
+			updatedPlugins.add( plugin.slug );
+			updatedSites.add( siteId );
+			const sitePlugin = getSitePlugin( plugin, siteId, pluginsOnSites );
+			updateAction( siteId, sitePlugin );
+		}
+	}
```
tyxla, line 143: Could you elaborate on this change? Have you done any performance measurements on the impact here?

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -30,48 +29,85 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
...
+	| 'deactivate'
+	| 'enableAutoupdates'
+	| 'disableAutoupdates'
+	| 'update'
+	| 'remove';
+
+export class PluginsList extends Component< Props, State > {
 	static defaultProps = {
+		// eslint-disable-next-line @typescript-eslint/no-empty-function
 		recordGoogleEvent: () => {},
```
tyxla, line 91: Do we need the default props here? Why is this even necessary - if it's for testing purposes, let's make sure to provide a mock there.

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -30,48 +29,85 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
...
 	};
 
-	shouldComponentUpdate( nextProps, nextState ) {
-		const propsToCheck = [ 'plugins', 'sites', 'selectedSite' ];
-		if ( checkPropsChange.call( this, nextProps, propsToCheck ) ) {
+	checkPropsChange( nextProps: Props, propArr: ( keyof Props )[] ) {
```
tyxla, line 94: Is there any benefit of moving this inside the component?

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -619,14 +665,18 @@ export class PluginsList extends Component {
...
 		const selectedSite = getSelectedSite( state );
 		return {
 			allSites: getSites( state ),
 			pluginsOnSites: getPluginsOnSites( state, plugins ),
 			selectedSite,
 			selectedSiteSlug: getSelectedSiteSlug( state ),
-			hasManagePlugins: siteHasFeature( state, selectedSite?.ID, WPCOM_FEATURES_MANAGE_PLUGINS ),
+			hasManagePlugins: siteHasFeature(
+				state,
+				selectedSite?.ID ?? null,
```
tyxla, line 682: Does it make sense to call `siteHasFeature()` with a `null`? Should that be `false` already? Work for another PR obviously, but it seems odd that `siteHasFeature()` even supports a `null` siteId.

client/my-sites/plugins/plugin-management-v2/test/plugin-row-formatter.test.tsx
```diff
@@ -107,9 +108,15 @@ describe( '<PluginRowFormatter>', () => {
...
+		plugin.sites = sitesOriginal;
 	} );
 
-	test( 'should render correctly and show activate and toggle checked value', async () => {
+	// These skipped test nevers actually worked, it just tests that the HTML toggle actually toggled.
+	// To verify comment out the entirety of the toggle callbacks in the relevant components and observe
+	// that the test still pass.
+	// They were passing because their toggle code didn't execute. I changed some test data which caused
+	// the toggle code to actually execute, and now it needs to be rewritten to use nock to simulate the server calls.
+	test.skip( 'should render correctly and show activate and toggle checked value', async () => {
```
tyxla, line 128: Ideally, we should follow-up and fix those skipped tests.

tyxla (commented):
Thanks for your hard work on this, @robertf4!

I appreciate the throrough testing instructions, too!

I've taken the time to leave some feedback and suggestions, as well as comments and questions.

However, I'm a bit concerned that this PR introduces far too many changes, and far too many different groups of changes, making it hard to follow and potentially creating comprehensibility issues for our future selves. I'm concerned also because the plugins area is one of the most complicated and difficult to test areas of Calypso, and we have no guarantee that we won't intrduce subtle issues like we've done before.

I know it might delay the work substantially, but I'd recommend considering to split this work to smaller pieces.

At a first glance, I can see these threads of work in this PR, where each of them can be its own PR, or even multiple separate PRs:

- Fixes (like using `QueryAllJetpackSitesPlugins` instead of `QueryJetpackPlugins` for more than 1 site).
- Additional site / siteId existence checks
- New selectors + tests
- Updating old code to use new selectors, removing unused selectors
- Fixing skipped tests
- Performance optimizations - one PR per optimization, including declarative-to-imperative refactors
- Refactoring to TypeScript

What do you think?

client/my-sites/plugins/main.jsx
robertf4: This was intended to be a temporary fix, you can read some of the discussion on it here: p1671076313369439-slack-C039K7CD119. The 20 number is definitely arbitrary.

I am partial to the method of caching these results on the server so that they are all retrieved in a single request but I also recognize that this is a lot of dev effort.

I moved these changes to https://github.com/Automattic/wp-calypso/pull/72621 so that they aren't in this PR.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
robertf4: Yes, but it would be an array of undefined because `getPluginOnSite` returns undefined in that case. I think in this case most code would handle an empty array better than an array of undefined.

client/my-sites/plugins/plugin-details-header/index.jsx
robertf4: I agree, but I think the behavior needs to be better defined first before a selector that could be re-used is created. Previously the main selector returned a `version` property on the plugin. However, in the data returned from the server the version is given per site. A plugin can be installed on several different sites with a different version on each site. So what does it mean for a plugin itself to have a version?

Previously it was just the version of whatever site was last in the list because the version of each site overwrote the version of the last one. There were several properties like this, things specific to a site that were being lifted to the plugin object and the last site determining what the values are. I removed all of those and left them only on the site objects.

There should probably be selectors for getting the lowest version on a site, highest version on a site, and most recent version (i.e. the wporg version), but I feel that warrants its own PR.

client/my-sites/plugins/plugin-management-v2/plugin-details-v2/index.tsx
robertf4: I moved it to https://github.com/Automattic/wp-calypso/pull/72627

client/my-sites/plugins/plugin-management-v2/plugin-manage-subscription/index.tsx
robertf4: It is `boolean | undefined`, and `getPluginPurchased` takes a boolean.

client/my-sites/plugins/plugin-management-v2/test/remove-plugin.test.tsx
robertf4: That is how it works in production and something about the actions that fire with the response necessitated this, but I don't remember exactly what. But the tests fail if this response object isn't there.

client/my-sites/plugins/plugin-management-v2/test/update-plugin.test.tsx
robertf4: I didn't even see that, I'm not sure if it's intentional or not. I created 1202619025189113-as-1203834327086972 for it.

client/my-sites/plugins/plugin-management-v2/types.ts
robertf4: I added a comment but I can't think of a better name because these are just random properties that are stuck on to the plugin objects here and there in the component code, so there's not a new specific name to give it. I consider this an anti-pattern but it would be a lot of work to change it.

client/my-sites/plugins/utils.js
robertf4: I did not measure the performance impact of this particular change in isolation from everything else, but `filter` creates a new array and this way avoid that while being just as clear to read.

client/state/plugins/installed/selectors.ts
robertf4: I did `git mv`, but git doesn't track history like svn so it doesn't help. `git mv` is just shorthand for `mv oldFile newFile; git rm oldFile; git add newFile`. To determine if two files with different names are the same git calculates a number from their diff and if they are similar enough then it calls that a rename. If they are different enough then it calls them different files. It's unnoticeable when it works but in this case it gets it wrong. See here for more info: https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F 

I would probably have to make the changes in the .js file in one PR and change it to .ts in another PR to keep the history, but I used the .ts conversion to find and fix some bugs that were very difficult to track down due to object types being changed in several places.

client/state/plugins/installed/selectors.ts
robertf4: With the change to not use `omit` it is no longer declared.

client/state/plugins/installed/selectors.ts
robertf4: Originally it was just performance improvements for the selectors. `getFilteredAndSortedPlugins` was `getPlugins` and it was very slow algorithmically.

I created `getAllPluginsIndexedBySiteId` and `getAllPluginsIndexedByPluginSlug` as internal helpers to provide some data structures that would reduce the algorithmic complexity in `getPlugins`. Then I noticed that `getPlugins` was being used in many places to find a single plugin by searching the array of all plugins which was part of the performance issues, and my helper selectors were memoized and find a single plugin in constant time so I exposed them just for that use case.

I then spent a lot of time tracking down minor bugs which were very difficult to find because they were happening in code far away from these selectors, usually in a component that had data from the selector passed down through many components and modified by some until finally the bug manifested. In almost every case it was a type error because some javascript code either changed the type of an object or used a type that didn't exist. This took enough time and I found enough of these bugs that I decided to change it to TypeScript which did fix them all at once and will prevent them from happening again.

At this point I don't think that separating the two is possible, or at least not worth the effort. I would have to take the existing TypeScript code and convert it back into JavaScript in order to keep the bug fixes, and then convert it back to TypeScript in another PR, and it just doesn't seem worth it.

client/state/plugins/installed/selectors.ts
robertf4: I created 1202619025189113-as-1203834705849811 to make unit tests. They should be covered by the existing tests for the other selectors though as these new selectors were originally just internal helpers for the other selectors.

client/state/plugins/installed/selectors.ts
robertf4: If I recall correctly some tests were failing because they called this selector with different sets of site Ids, and the mutation done in this selector was changing the objects returned from `getAllPluginsIndexedBySiteId` depending on the site Ids passed. The `cloneDeep` fixed that.

The object has to be mutated because of the behavior under the "Filter the sites object on the plugins so that only data for the requested siteIds is present" comment, which is assumed by a lot of component code. I think ideally that wouldn't be an assumption of the component code because the plugin should still have all its sites regardless of which sites were requested, but I looked into it and this assumption would be very difficult to change. 

The memoization is still useful since both selectors are memoized. `allPluginsIndexedBySiteId` will only be calculated once per fetch of plugins, and `getFilteredAndSortedPlugins` will be calculated once for each set of siteIds and filter passed to it, and only the data for the sites that are requested is getting cloned.

client/state/plugins/installed/selectors.ts
robertf4: I think this makes a new object at the highest level, but the lower level objects retain the same reference so a mutation here is reflected for everything that calls `getAllPluginsIndexedBySiteId`. 

You can observe this with the unit tests. Run them and watch them pass, then remove the `cloneDeep` and watch them fail. They fail because each time `getFilteredAndSortedPlugins` is called with a different set of `siteIds` it mutates the data returned by `getAllPluginsIndexedBySiteId`. 

As an aside, this is horrifying to me because it means every memoized selector relies on all code which calls it to never mutate a returned object. I think it's worth considering making the memoization deep freeze the cached object.

client/state/plugins/installed/selectors.ts
robertf4: Wouldn't returning a new object be equivalent to cloning in this case?

client/state/plugins/installed/selectors.ts
robertf4: Yes, I think I explained a similar situation in another comment. The `version` was coming from `plugin.update.new_version`, and the `update` object is actually a site object and not a plugin object. The `update` object was being lifted onto the plugin from its site objects, and it was whatever the last site in the list was because each one overwrote the previous one. 

This led to some strange behavior and I think it is properly speaking a data normalization error so I eventually decided to leave the site specific data only on the site objects.

client/state/plugins/installed/selectors.ts
robertf4: It was using `getPlugins` (now called `getFilteredAndSortedPlugins`) which was causing this selector to have performance issues. (Most of the performance issues came down to `getPlugins`.) 

After my changes `getFilteredAndSortedPlugins` was still sorting which is unnecessary for this selector so I changed this implementation to call a lower level data structure and avoid that.

client/state/plugins/installed/types.ts
robertf4: It could be done that way technically but I'm not sure that conceptually it works. `InstalledPluginData` is intended to reflect what is in the Redux store, which in turn is what comes from the server. `PluginSites` is the shape of objects computed from that server data. 

So I don't see `InstalledPluginData` as being composed out of any sub-types, it is simply what comes from the server. Even if I made this type a new type I wouldn't want to use it in `InstalledPluginData` as I find it easiest if the data which comes from a server is clearly represented as its own type, which makes it easiest for future developers to update.

client/my-sites/plugins/plugin-details-header/index.jsx
robertf4: I didn't do that because `reverse` creates a new array.

client/state/plugins/installed/selectors.ts
robertf4: I could, but I wouldn't rely on future developers to know about and use such a function when it is so obvious that the sites object is indexed by site ID. There was lots of mismatching between numbers and strings before because not all code was written to handle object keys always being strings, but TypeScript has fixed all of that so such mistakes shouldn't be possible going forward and the conversion to a number is pretty simple.

client/state/plugins/installed/test/selectors.ts
robertf4: Is that true even for unit tests? I understand it for production code since it's one less library to go in the bundle, but for code like this it seems like lodash can be pretty helpful.

client/my-sites/plugins/plugins-list/index.tsx
robertf4: The safe access operator returns `undefined` if it fails, and `siteHasFeature` takes a null, so here I am essentially turning an `undefined` into a `null`. 

I don't know why `siteHasFeature` would take a null siteId, I'm not sure what a function with site in the name should be expected to do if there is no site.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
tyxla: In that case, should we just call `.filter( Boolean )` on the result of `.map()` to ensure we don't have any `undefined` items in the array? That would also result in an empty array if all of the items are undefined.

client/my-sites/plugins/main.jsx
tyxla: Thanks for moving them out - feel free to ping me when this PR is ready for review and we can discuss it separately.

client/my-sites/plugins/plugin-details-header/index.jsx
tyxla: A separate PR for such a selector sounds good to me. We can include the priority of versioning that you described in the inline docs of that selector. Some unit tests would also help.

client/my-sites/plugins/plugin-details-header/index.jsx
tyxla: This is actually not true. `Array.prototype.reverse()` will [reverse the array in place](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse#return_value), and not return a new copy of the array. Also, for the protocol, `Object.values()` already returns a new array, so we're safe from mutating anything, too.

client/my-sites/plugins/plugin-management-v2/types.ts
tyxla: Maybe something around `PluginComponentProps` then?

client/my-sites/plugins/plugin-management-v2/plugin-details-v2/index.tsx
tyxla: Thanks, I'll be glad to review this once it's ready - don't hesitate to ping me.

client/my-sites/plugins/plugin-management-v2/plugin-manage-subscription/index.tsx
tyxla: Sounds good 👍

client/my-sites/plugins/plugin-management-v2/test/remove-plugin.test.tsx
tyxla: Thanks for clarifying 👍

client/state/plugins/installed/selectors.ts
tyxla: For what it's worth, we make it easier for Git to track similarities between files when we introduce fewer changes to a file in a PR. Just another great reason to keep changes small and contained.

client/state/plugins/installed/test/selectors.ts
tyxla: I'd argue it's true for any code. We own all of our code, and a lodash usage that exists in tests can easily sneak into production code if someone is borrowing code from that test as an example. Ideally, we shouldn't be relying on Lodash, regardless of the context. We're aiming to fully get rid of Lodash at some point, see p4TIVU-9Bf-p2.

client/state/plugins/installed/types.ts
tyxla: I don't have particularly strong feelings about it, it just feels like those types are a bit complex and still have some repetition, so it might make sense to try improving them in the future. I do agree that the data returned by the server and the data originally on the client don't always correspond and are historically a bit messy.

client/state/plugins/installed/selectors.ts
tyxla: I see, thanks for clarifying.

Perhaps we could use [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead of introducing yet another Lodash usage?

client/state/plugins/installed/selectors.ts
tyxla: I agree this is suboptimal 😓 Anyway, if we really need cloning, as I mentioned above, perhaps we could use [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone).

client/state/plugins/installed/selectors.ts
tyxla: I understand your rationale, but I'd disagree with the approach taken; I think the approach here to do all refactors in the same PR has created a potentially incomprehensible change that we just should not merge altogether, but rather spend some effort splitting changes out into meaningful, atomic changes. What you described seemed to me like at least a few PRs we could have undertaken one by one. I realize this takes more time, but it's just the more responsible way to work with a huge codebase spanning hundreds of developers, hundreds of projects, almost ten years of work, and thousands of commits. My only advice here is to be responsible and push changes in a way that our future selves will thank our past selves for. I'm happy to provide ideas on how to split the changes if you're struggling. I did provide a rough plan in the primary comment of my previous review, but I'm happy to take it to the lower level if that will help.

client/state/plugins/installed/selectors.ts
tyxla: Nice, thank you 👍

client/state/plugins/installed/selectors.ts
tyxla: The context here is that I'm looking at addressing this as if we've removed the cloning already. If we work around it with `structuredClone`, I believe this should be fine, although, ideally, we should avoid mutation even if we're mutating new objects. The rationale is that in this current context, a mutation might be safe, but if at some point we refactor the code, we might lose a trace that this actually mutates code and inadvertently introduce mutations in the future. Let me know if that makes sense.

client/state/plugins/installed/selectors.ts
tyxla: Sounds like something that could be moved to another PR?

client/state/plugins/installed/selectors.ts
tyxla: Thank you. Noting that ideally, unit tests should be an inseparable part of the code they're testing. I understand that it gets complex when there's a 1K line PR without writing new tests, but this is just another reason why I'm advocating for splitting to smaller PRs as much as possible.

client/my-sites/plugins/utils.js
tyxla: It makes it a bit less readable in my mind. How about we split it out into a separate PR and do some benchmarking? Perhaps the impact is negligible and the change is just not required? Let's find out.

client/my-sites/plugins/plugins-list/index.tsx
tyxla: Worth exploring and diving deeper to find out, but as I mentioned earlier, it can be done in a separate PR if you're interested.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
robertf4: What makes that method preferable to this one? The ternary operator results in less computation, although it's all negligible overall.

client/state/plugins/installed/selectors.ts
robertf4: `structuredClone` is added in node 17, but Calypso still supports 16 according to package.json and the tests seem to be running on 16. So if I switch to `structuredClone` it will crash in some supported environments.

client/state/plugins/installed/selectors.ts
robertf4: I agree that mutation like this shouldn't be done, but there is component code which assumes this mutation happens. In my original refactor I removed the mutation and it caused bugs.

client/my-sites/plugins/plugins-list/index.tsx
robertf4: It was present in the js version of the code and I was changing as little as possible.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
tyxla: It's about simplicity mostly. Which I prefer, unless you can demonstrate that it makes things slower 😉

client/state/plugins/installed/selectors.ts
tyxla: I'd love to hear more about this - could you elaborate?

client/state/plugins/installed/selectors.ts
tyxla: We could add a polyfill for the server. There are a few modules that can help with that (https://www.npmjs.com/package/realistic-structured-clone). It's definitely worth the effort, given that we're getting rid of Lodash and we do our best not to introduce additional usages. It's also worth double-checking if on the server we really need this module's functionality, or it's necessary just to compile and will never execute on the server. If we have the latter case, we could just add a workaround to noop that module for the server.

client/my-sites/plugins/plugin-details-header/index.jsx
WunderBart: You should also be able to do it using `Array.prototype.findLast`:

```js
let version = plugin?.version;
if ( plugin.sites ) {
  const sites = Object.values( plugin.sites );
  version = sites.findLast( site => site.version ).version;
}
```

client/my-sites/plugins/plugin-management-v2/update-plugin/index.tsx
```diff
@@ -5,60 +5,38 @@ import { useTranslate } from 'i18n-calypso';
...
-			const siteId = selectedSite ? selectedSite.ID : site.ID;
-			const sitePlugin = pluginsOnSites?.sites[ siteId ];
-			return sitePlugin?.update?.new_version;
-		} )
-		.filter( ( version ) => version );
+	const updatedVersions = selectedSite
+		? [ plugin.sites[ selectedSite.ID ].update?.new_version ]
+		: Object.values( plugin.sites )
+				.map( ( site ) => site.update?.new_version )
+				.filter( ( x ) => x );
```
WunderBart: This can also be `.filter(Boolean)`

client/state/plugins/installed/selectors.ts
noahtallen: We just updated to Node 18 last week -- the main Calypso server runs Node 18 now, and I don't have any plans to maintain backwards compatibility with Node 16

client/my-sites/plugins/plugin-management-v2/update-plugin/index.tsx
robertf4: Good suggestion, thanks!

client/state/plugins/installed/selectors.ts
robertf4: I did some more investigation, and it seems now the issue is in `jest`. Jest 27 [had a bug](https://github.com/facebook/jest/issues/12628) where `structuredClone` wasn't available and it is fixed in Jest 28.

tyxla: This might be a good opportunity to upgrade Jest - this is something @jsnajdr has been looking to do in Gutenberg as well. 

In the meantime, if tests are the only reason for not using `structuredClone` we could temporarily do something like this in the test:

```
global.structuredClone = jest.fn( ( value ) => {
    return JSON.parse( JSON.stringify( value ) );
} );
```

richtabor:
Closing stale pull requests.

Feel free to re-open and push it through if relevant—but don't let it sit. :)

Issue #72313, opened 2023-01-19
Pattern Assembler - Fast preview component max-height: 2000px cuts off patterns in tablet viewport


miksansegundo:
## Description

The fast preview component used for the large preview in the assembler has a `max-height: 2000px` per pattern, which is not tall enough for some section patterns, and they are being cut off. 

![Image](https://user-images.githubusercontent.com/1881481/213371369-aeaec047-68f2-4339-a9a1-da05b0a058c4.png)

## Testing steps
- Create a new site, choose a domain, and a plan...
- Click continue until the Design picker, and scroll down to click `Start designing`
- Add the pattern with four images in a grid 2x2
![Image](https://user-images.githubusercontent.com/1881481/213375052-7edcc32e-ed6b-46ab-ba16-38328f803b02.png)
- Check the large preview using the `tablet` viewport
![Image](https://user-images.githubusercontent.com/1881481/213375133-8bbfe226-37dd-492b-ae93-71af485cd797.png)

## Acceptance criteria

 - [X] Patterns should not be cut off in the large preview

|Before|After|
|------|-----|
|![Image](https://user-images.githubusercontent.com/1881481/213375133-8bbfe226-37dd-492b-ae93-71af485cd797.png)|![Image](https://user-images.githubusercontent.com/1881481/213375942-f1ae779f-0f7d-487c-8701-c761ea5825bf.png)|

## Dev tasks

- [X] The component `PatternRenderer` should have the option to use or not the `max-height: 2000px`. See the line [here](https://github.com/Automattic/wp-calypso/blob/trunk/packages/block-renderer/src/components/pattern-renderer.tsx#L20).
- [X] The component `PatternRenderer` in `PatternLargePreview` should not use `max-height`. See the line [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx#L28).

## Related
https://github.com/Automattic/wp-calypso/pull/72253

PR #70063, opened 2022-11-16
WIP Pattern Assembler: Render block patterns on calypso side

arthur791004:
#### Proposed Changes

* Use the `pattern-assembler/client-side-render` flag to control the client-side render
* Introduce the `@automattic/blocks-renderer` package to support rendering the block patterns on the calypso side. The main concept is as followed:
  * `BlocksProvider`: Fetch the block settings from API and store it in the store of the `block-editor`
  * `BlocksRenderer`: Render the HTML and generated styles of the block pattern. The HTML of the block pattern is from the `/block-patterns/render` endpoint.
  * `PatternsRendererProvider`: The provider to batch query the rendered patterns from the server side
  * `PatternsRenderer`: Render the patterns with the given pattern ids
* This PR only uses client-side rendering on the sidebar. Will continue to apply the client-side rendering on the large preview.

See pbxlJb-30c-p2 for more context

TODOs

- [ ] Render site logo placeholder
- [ ] Scroll to the specified pattern when the selected pattern set is changed
- [ ] Maybe removing animation

https://user-images.githubusercontent.com/13596067/203761687-0ebe8afc-8174-45ec-ab24-769af15f7534.mov

#### Testing Instructions

* Apply D86835-code to your sandbox
* Go to `/setup?siteSlug=<your_site>`
* Click the "Continue" button until you land on the Design Picker step
* Click the "Get started" button below "Start with a blank canvas" to enter the Pattern Assembler step
* On the Pattern Assembler step, verify all of the patterns shown in the sidebar are rendered correctly

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
[...]

PR #72364, opened 2023-01-20
Pattern Assembler - Fast preview component max-height: 2000px cuts off patterns in tablet viewport

miksansegundo:
#### Proposed Changes

* Allow overwriting the default block renderer `maxHeight`
* Disable `maxHeight` in the large preview
* Limit the height of `100vh` patterns to `500px` instead of `2000px` or use the viewport height if it's passed

This solution fixes the pattern cut-off issue in the large preview while still supporting `viewportHeight` for patterns with `100vh` and sets a height of `500px` for their thumbnail preview, which looks better than `2000px`.

**Thumbnail for 100vh patterns**

|Before|After|
|------|-----|
|Screenshot 2566-01-20 at 11 35 24|Screenshot 2566-01-20 at 14 30 22|

**Large preview for patterns that create viewport scroll**

|Before|After|
|------|-----|
|Screenshot 2566-01-20 at 14 35 57|Screenshot 2566-01-20 at 14 33 35|

#### Testing Instructions

- Create a new site, choose a domain, and a plan...
- Click continue until the Design picker, and scroll down to click `Start designing`
- Add the pattern with four images in a grid 2x2
Image
- In the large preview, use the `tablet` viewport, and the scroll should appear because the pattern height is taller
- Confirm that the whole pattern is visible when you scroll the viewport down and it's not cut off

#### Pre-merge Checklist

- [X] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [X] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [X] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to https://github.com/Automattic/wp-calypso/issues/72313 https://github.com/Automattic/wp-calypso/pull/70063

### DIFF
--- a/packages/block-renderer/src/components/block-renderer-container.tsx
+++ b/packages/block-renderer/src/components/block-renderer-container.tsx
@@ -20,6 +20,8 @@ interface BlockRendererContainerProps {
 	viewportHeight?: number;
 	maxHeight?: string | number;
 	minHeight?: number;
+	isMinHeight100vh?: boolean;
+	maxHeightFor100vh?: number;
 }
 
 interface ScaledBlockRendererContainerProps extends BlockRendererContainerProps {
@@ -35,6 +37,8 @@ const ScaledBlockRendererContainer = ( {
 	containerWidth,
 	maxHeight = BLOCK_MAX_HEIGHT,
 	minHeight,
+	isMinHeight100vh,
+	maxHeightFor100vh,
 }: ScaledBlockRendererContainerProps ) => {
 	const [ contentResizeListener, { height: contentHeight } ] = useResizeObserver();
 	const { styles, assets, duotone } = useSelect( ( select ) => {
@@ -84,9 +88,19 @@ const ScaledBlockRendererContainer = ( {
 	}, [] );
 
 	const scale = containerWidth / viewportWidth;
-	let iframeHeight = contentHeight as number;
-	if ( viewportHeight && ( contentHeight as number ) < viewportHeight ) {
-		iframeHeight = viewportHeight;
+
+	let scaledHeight = ( contentHeight as number ) * scale || minHeight;
+	if ( isMinHeight100vh && maxHeightFor100vh && ! viewportHeight ) {
+		scaledHeight = maxHeightFor100vh * scale;
+	}
+
+	let iframeHeight = ( contentHeight as number ) || minHeight;
+	if ( isMinHeight100vh ) {
+		if ( viewportHeight ) {
+			iframeHeight = viewportHeight;
+		} else if ( maxHeightFor100vh ) {
+			iframeHeight = maxHeightFor100vh;
+		}
 	}
 
 	return (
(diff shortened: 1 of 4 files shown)


miksansegundo:
@autumnfjeld I'm going to merge this PR because the issue exists in production now.

autumnfjeld:
I think your changes are definitely good.  I tested in production and it is much better. :) 
I followed your testing instructions ✅ 
Thank you for your attention to detail on this!

I do wonder if we should allow the thumbnail to be a height greater than 500px.   It’s not really possible to convey that “this pattern will fill your screen”.    But maybe it should be taller than 500px?  The reason I say that is to give a better “impression” that the pattern is “very tall” :slightly_smiling_face:.  

I’m fine with leaving it how it is however.  Just wanted to say my two thoughts.    I mentioned that in a comment.

packages/block-renderer/src/components/block-renderer-container.tsx
```diff
@@ -85,13 +89,30 @@ const ScaledBlockRendererContainer = ( {
...
 			className="scaled-block-renderer"
 			style={ {
 				transform: `scale(${ scale })`,
-				height: ( contentHeight as number ) * scale || minHeight,
-				maxHeight: ( contentHeight as number ) > maxHeight ? maxHeight * scale : undefined,
+				height: scaledHeight,
+				maxHeight:
+					maxHeight !== 'none' && ( contentHeight as number ) > maxHeight
+						? ( maxHeight as number ) * scale
+						: undefined,
```
noahtallen, line 115: Hey @miksansegundo, I noticed this while working on #74540. What are the units for maxHeight here? One issue I noticed is that `maxHeight` can technically be a string (e.g. "200px"), which probably wouldn't work with this logic. (But it doesn't seem like we use it like that anywhere.)

But the main issue is that the types are technically wrong, since we now allow `maxHeight` to be a _string_ or a number. If we only want to allow "none", could we change it so that the prop type is `"none" | number`?

miksansegundo: Thanks for the ping @noahtallen, the units for `maxHeight ` are `pixels` and the type is `number`. I added the type `string` only to support disabling the default value using `none` . 

I'll change the type to use `'none' | number` for now and follow up with a better way to disable the default value via a new prop. See https://github.com/Automattic/wp-calypso/pull/74843

noahtallen: Cool!

PR #69461, opened 2022-10-26
Themes: Show wp.org results when there is no wp.com results to show

WBerredo:
#### Proposed Changes

* Query WP.org themes and show them when there are no WP.com results to show
  * The name or id of the theme should be the same as the searched 
* Add a fallback for screenshot images when it cannot be handled by `photon`

#### Testing Instructions
* Go to Themes page with a free site: `/themes/{free_site}`
* Search for an option that returns no results in WP.com but returns in WP.org. Ex: `astra`
* It should show the wp.org cards above the plan upgrade CTA card

| Before  | After |
| ------------- | ------------- |
|Screen Shot 2022-10-25 at 21 37 34|Screen Shot 2022-10-25 at 16 58 15|

* Search for a theme that doesn't returns anything on WP.com or WP.org. Ex: `abc`
* It should show the message `"No themes match your search"` as the **Before** version above

---

* Go to Themes page with a Business or ecommerce site: `/themes/{business_site}`
* Search for an option that returns no results in WP.com but returns in WP.org. Ex: `astral`
* It should show the wp.org cards 

| Before  | After |
| ------------- | ------------- |
|Screen Shot 2022-10-26 at 11 54 20|Screen Shot 2022-10-26 at 11 43 41|

* Search for a theme that doesn't returns anything on WP.com or WP.org. Ex: `abc`
* It should show the message `"Sorry, no themes found."` as the **Before** version above

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Fixes #69382

PR #72368, opened 2023-01-20
Theme Showcase: Fix Incorrect Search Logging

Copons:
#### Proposed Changes

* Create a new `SearchThemesTracks` component, entirely devoted to recording search events in the Theme Showcase.
* Introduce a new `calypso_themes_search_results` Tracks event with detailed information about theme searches.

#### Logic

The Theme Showcase displays themes in many different ways.

If themes matching the current query are already available in the Redux state, we don't call any external endpoints. Otherwise, we request:
- Dotcom themes.
- Dotorg themes.
- Locally installed themes (for Jetpack sites).

To ensure we record the complete count, we first have to determine if we are retrieving from the Redux state (immediate), or wait until all the fetches are complete.

Furthermore, currently, we only show Dotorg themes when there are no Dotcom results — and only Dotorg themes whose names are exact matches for the search query — despite always fetching Dotorg themes regardless of whether we need them or not.

To record the actual number of themes returned by a search and displayed to the user, we have to perform this additional matching.

#### Why a New Event?

We already have `calypso_themeshowcase_search` and `calypso_themeshowcase_search_empty_results`, but they are part of the remote fetches, meaning:
- They are actually logging the fetches, not what the client (and therefore the user) ends up displaying.
- They record Dotorg and Dotcom fetches for the same search separately: for every search, we'll have multiple events.
- They just don't give us accurate client-side data.

I'll leave it to other people to decide if we should remove those events, or leave them be, since they are recording different things compared to this new event.

#### The New Event: `calypso_themes_search_results`

- `search`: the search query (e.g. `foo`).
- `search_term`: the full search query including filters (e.g. `subject:blog foo`).
- `search_taxonomies`: the search filters (e.g. `subject:blog`).
- `tier`: the theme tier filter (e.g. `premium`).
- `result_count`: the total count of all themes matching the query (all Dotcom + all Dotorg results).
- `actual_count`: the count of themes actually displayed on the search results page (this is what the user is actually seeing, and the data we should use for analyzing the search).
- `wpcom_result_count`: the count of Dotcom results.
- `wporg_result_count`: the count of Dotorg results.

#### Additional Checks

@WBerredo I need to know if I have replicated correctly the "Dotorg exact match" introduced in https://github.com/Automattic/wp-calypso/pull/69461.
My concern at this point is that maybe I should not do it *for every search*, but only in some specific cases (maybe when the `forceWpOrgSearch` prop is true?).

#### Testing Instructions

* In the browser's console, run `localStorage.setItem( 'debug', '*' );` to turn on the debug logging.
* Filter the console by `calypso:analytics Record event "calypso_themes_search_results" called with props` to remove all the noise.
* Navigate to `/themes` and play around with searching themes.
* Ensure that the event recording is consistent with what you are seeing on screen.
  * Especially: the amount of themes displayed should be the same as `actual_count`.
  * The event should only be recorded once per search (when changing search term or filters).
  * It should work for Jetpack sites when searching for their installed themes.

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/my-sites/themes/search-themes-tracks.js
+++ b/client/my-sites/themes/search-themes-tracks.js
@@ -7,9 +7,8 @@ import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 
 function useThemeSearchDetails( query ) {
 	const search = query.search || '';
-	const searchTaxonomies = useSelector( ( state ) =>
-		prependThemeFilterKeys( state, query.filter )
-	);
+	const searchTaxonomies =
+		useSelector( ( state ) => prependThemeFilterKeys( state, query.filter ) ) || '';
 	const searchTerm = searchTaxonomies + search;
 	const prevSearchTerm = usePrevious( searchTerm );
 	const searchTermHasChanged = prevSearchTerm !== searchTerm && !! searchTerm && searchTerm !== '';
--- a/client/my-sites/themes/search-themes-tracks.js
+++ b/client/my-sites/themes/search-themes-tracks.js
@@ -6,7 +6,7 @@ import { isRequestingThemesForQuery, prependThemeFilterKeys } from 'calypso/stat
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 
 function useThemeSearchDetails( query ) {
-	const search = query.search;
+	const search = query.search || '';
 	const searchTaxonomies = useSelector( ( state ) =>
 		prependThemeFilterKeys( state, query.filter )
 	);
@@ -68,8 +68,8 @@ export default function SearchThemesTracks( { query = {}, themes = [], wporgThem
 
 	const recordSearchEvent = recordTracksEvent( 'calypso_themes_search_results', {
 		search,
-		search_term: searchTerm?.trim(),
-		search_taxonomies: searchTaxonomies?.trim(),
+		search_term: searchTerm.trim(),
+		search_taxonomies: searchTaxonomies.trim(),
 		tier: query.tier,
 		result_count: totalCount,
 		actual_count: actualCount,
--- a/client/my-sites/themes/search-themes-tracks.js
+++ b/client/my-sites/themes/search-themes-tracks.js
@@ -1,0 +1,109 @@
+import { usePrevious } from '@wordpress/compose';
+import { useEffect, useMemo } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import { isRequestingThemesForQuery, prependThemeFilterKeys } from 'calypso/state/themes/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+
+function useThemeSearchDetails( query ) {
+	const search = query.search;
+	const searchTaxonomies = useSelector( ( state ) =>
+		prependThemeFilterKeys( state, query.filter )
+	);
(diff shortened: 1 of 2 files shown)


client/my-sites/themes/search-themes-tracks.js
```diff
@@ -0,0 +1,109 @@
...
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+
+function useThemeSearchDetails( query ) {
+	const search = query.search;
+	const searchTaxonomies = useSelector( ( state ) =>
+		prependThemeFilterKeys( state, query.filter )
+	);
+	const searchTerm = searchTaxonomies + search;
+	const prevSearchTerm = usePrevious( searchTerm );
+	const searchTermHasChanged = prevSearchTerm !== searchTerm && !! searchTerm && searchTerm !== '';
```
WBerredo, line 14: The expectation is to pick all changes, except from searching _something_ to _empty_, right?

client/my-sites/themes/search-themes-tracks.js
```diff
@@ -0,0 +1,109 @@
...
+import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import { isRequestingThemesForQuery, prependThemeFilterKeys } from 'calypso/state/themes/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+
+function useThemeSearchDetails( query ) {
+	const search = query.search;
+	const searchTaxonomies = useSelector( ( state ) =>
+		prependThemeFilterKeys( state, query.filter )
+	);
+	const searchTerm = searchTaxonomies + search;
```
WBerredo, line 12: There might be a need to cast `search` to string here, when we load the page the `searchTerm` is being logged as `"undefined"`, which is probably caused by doing something like `'' + undefined`.

Another case is when we select a category we end up with:

```
search: undefined
search_taxonomies: "subject:business"
search_term: "subject:business undefined"
```


When it should be: 

```
search: undefined (or '')
search_taxonomies: "subject:business"
search_term: "subject:business"
```

WBerredo (changes requested):
Thanks for diving into this @Copons!
The changes seem to be good except for the mentioned point below.

For the WP.org search, it will only be triggered when the flag `forceWpOrgSearch` is enabled, but I don't see any problem in logging as is, because I'm assuming it will return an empty list anyways.

PS: Some unit tests are failing.

client/my-sites/themes/search-themes-tracks.js
Copons: Ooh, good catch, I guess this is what's failing the unit tests! 🤔 

I had something to guard against undefined `search` at some point, but then refactored everything and must have totally lost it in the process.

client/my-sites/themes/search-themes-tracks.js
Copons: Yes.
My reasoning is that if the search is empty, it's not really technically a search, is it?

It's basically the default unfiltered view. Logging this doesn't give us any particularly interesting info, although the more I think about it, the weaker my opinion is.

I guess I'm gonna keep excluding empty searches for now; we can revisit this decision in the future if we determine we need to know more about them too.

WBerredo: I tend to agree with you, thanks for the explanation.

WBerredo (approved):
LGTM! 🚢

PR #72373, opened 2023-01-20
Create a separate signup flow for paid media

donlair:
#### Proposed Changes

* Creates a separate onboarding flow to be used with paid media landing pages.
* This flow will be used only for WPcom paid media, so I removed a lot of the conditional logic that's used by other A8C properties.
* This v1 is essentially a duplicate of the current experience. We'll optimize this for paid media in future PRs.
* I removed most of Lodash and other external libraries.
* More context available at pcbrnV-4Kk-p2/

#### Details
- Flow name: `onboarding-pm`
- Steps: [ `user-pm`, `domains-pm`, `plans-pm` ]

#### Testing Instructions

- Checkout this PR and start Calypso locally
- Navigate to: `/start/onboarding-pm`
- Proceed through the flow in both mobile and desktop viewports
- Verify that you can create both Free and Paid sites
- Test the `Use your own domain` flow.
- Test selecting a variety of domain categories (custom, featured, free)
- Inspect the Tracks events and confirm that they are reporting the correct Flow and Step names

#### Pre-merge Checklist

Related to #


client/signup/config/flows-pure.js
```diff
@@ -119,6 +119,15 @@ export function generateFlows( {
 			lastModified: '2020-12-10',
 			showRecaptcha: true,
 		},
+		{
+			name: 'onboarding-pm',
+			steps: [ 'user-pm', 'domains-pm', 'plans-pm' ],
```
gmovr, line 124: I think it would be better to keep the same steps, until they change, and then split them out. Doing so will reduce this PR to just introducing the new flow here, and we can iterate and make these steps as they're needed.

client/signup/steps/domains-pm/utils.js
```diff
@@ -0,0 +1,30 @@
+import validUrl from 'valid-url';
```
gmovr, line 1: These are unlikely to change I think, even if the steps change visually. Can we reference the existing files (with the same content)? As now we're duplicating the utils and their corresponding test files, with the same content.

config/development.json
```diff
@@ -26,7 +26,7 @@
 	"dsp_widget_js_src": "https://widgets.wp.com/promote/widget.js",
 	"zendesk_presales_chat_key": "216bf91d-f10f-4f66-bf65-a0cba220cd38",
 	"features": {
-		"ad-tracking": false,
+		"ad-tracking": true,
```
gmovr, line 29: We shouldn't toggle this flag (it's fine for testing locally though), but now it will be enabled always in dev, which we don't want.

gmovr (commented):
I added some comments, all though I think it would be good to get a final review from @Automattic/martech-decepticons  or maybe @claudiucelfilip  who's a bit more familiar with the signup flow👍 

I think we can avoid a lot of the duplication, which would reduce this PR (for now, until steps change), to just the changes in `flows-pure.js` (making a new flow with existing steps). This will give you a functional signup flow with a distinct URL, that can be used for testing.

This also allows us to only duplicate what's needed, e.g. we probably don't need to duplicate utils and their corresponding tests. If we change the domain step, as an example, we can introduce these changes in a new PR, and only change what's needed.

When changes are done to the steps, it would also be better (I think) to only duplicate what's needed, e.g. importing the utils and logic from the original steps, and weeding out some of the conditional code as the paid ads flow will be more focused.

donlair:
@gmovr thanks for the review! Your suggestions make sense to me, so I spun up a [separate PR](https://github.com/Automattic/wp-calypso/pull/72681) that uses the existing steps for now

PR #72379, opened 2023-01-20
Add a post-purchase redirect parameter for Premium themes upgrade nudge

BogdanUngureanu:
#### Proposed Changes

* Add a redirect_to parameter to logged-in upgrade nudges that are displayed in the Theme details page for Premium themes.

#### Testing Instructions

* Checkout the branch locally on your Calypso environment (the Calypso.live environment has an issue with redirect_to parameters)
* Using a free site, navigate to theme/alonso/your-site
* Click on the upgrade nudge and toggle between Yearly and Monthly tabs
* Notice that the `redirect_to` parameter is present between toggles
* Purchase a Premium subscription
* After you've purchased the subscription, you should be automatically redirected to `theme/alonso/your-site`
* Check that a notice appears on the top right of the page mentioning that the purchase has been made

#### Pre-merge Checklist

Related to https://github.com/Automattic/dotcom-forge/issues/1456


dsas:
This didn't actually work for me, I got the "are you sure you don't want business" page and then the "look what you bought" page. The redirect_to was in the URL though 🤔 


https://user-images.githubusercontent.com/93301/213761686-89551753-30f1-48eb-91c2-55c4454c8f46.mp4

BogdanUngureanu:
Hmm... nice catch! It looks like when you switch between those recurrence tabs the `redirect_to` parameter is not passed. Although, I did re-tested it with the calypso.live link and it seems that it doesn't work even if you don't toggle between those tabs.

Strange, it worked on my local env.

BogdanUngureanu:
I've opened another issue for the [successNotice issue](https://github.com/Automattic/wp-calypso/issues/72566), since it's not related to this PR. 

Since there seems a problem with the redirect_to logic for Calypso.live links, I'll update the testing steps so that we can move forward with PR without being blocked by unrelated implementations.

client/my-sites/theme/main.jsx
```diff
@@ -791,7 +791,9 @@ class ThemeSheet extends Component {
 			plansUrl = localizeUrl( 'https://wordpress.com/pricing' );
 		} else if ( siteSlug ) {
 			const plan = isExternallyManagedTheme || isBundledSoftwareSet ? PLAN_BUSINESS : PLAN_PREMIUM;
-			plansUrl = plansUrl + `/${ siteSlug }/?plan=${ plan }`;
+			plansUrl =
+				plansUrl +
+				`/${ siteSlug }/?plan=${ plan }&redirect_to=${ encodeURIComponent( document.URL ) }`;
```
dsas: Very likely to be just me, and it doesn't matter in the least, but i'm surprised there's not a nicer way of getting the current page. Even `window.location.href.replace( window.location.origin, '' );`. I don't think it's worth worrying over in the slightest though.

dsas (approved):

client/my-sites/theme/main.jsx
mmtr: Cannot `analyticsPath` be re-used here?

```suggestion
				`/${ siteSlug }/?plan=${ plan }&redirect_to=${ analyticsPath }`;
```

BogdanUngureanu: The `analyticsPath` contains a `:site` string instead of the actual site name. I've added a hardcoded `redirectTo` const that builds the redirect URL.

The weird thing is that I've tried before a relative path and it didn't work. Apparently now it works 🪄

client/my-sites/plans-features-main/plan-type-selector.tsx
```diff
@@ -40,6 +40,7 @@ export type PlanTypeSelectorProps = {
 	eligibleForWpcomMonthlyPlans?: boolean;
 	isPlansInsideStepper: boolean;
 	hideDiscountLabel: boolean;
+	redirectTo?: string | null;
```
tyxla, line 43: Is there a good reason to support both `undefined` and `null` for this prop?

client/my-sites/theme/main.jsx
```diff
@@ -790,8 +790,11 @@ class ThemeSheet extends Component {
 		if ( ! isLoggedIn ) {
 			plansUrl = localizeUrl( 'https://wordpress.com/pricing' );
 		} else if ( siteSlug ) {
+			const redirectTo = `/theme/${ themeId }${ section ? '/' + section : '' }/${ siteId }`;
 			const plan = isExternallyManagedTheme || isBundledSoftwareSet ? PLAN_BUSINESS : PLAN_PREMIUM;
-			plansUrl = plansUrl + `/${ siteSlug }/?plan=${ plan }`;
+			plansUrl =
+				plansUrl +
+				`/${ siteSlug }/?plan=${ plan }&redirect_to=${ encodeURIComponent( redirectTo ) }`;
```
mmtr: Nitpick: `encodeURIComponent` replaces `/` with `%2F` (e.g. `?redirect_to=%2Ftheme%2Fastra%2Fexample.wordpress.com`), but slashes don't need to be escaped (`?redirect_to=/theme/astra/example.wordpress.com` works too). Since `redirectTo` is defined a few lines above (i.e. it is not given as param), we can guarantee it doesn't include any character that needs to be escaped and simplify this a little bit.
```suggestion
				`/${ siteSlug }/?plan=${ plan }&redirect_to=${ redirectTo }`;
```

client/my-sites/plans-features-main/plan-type-selector.tsx
```diff
@@ -152,6 +156,7 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > =
 					path={ generatePath( props, {
 						intervalType: 'monthly',
 						get_domain: domainFromHomeUpsellFlow,
+						...redirectTo,
```
mmtr: Is this really needed? Looks like the `SegmentedControl.Item` component does nothing with the `redirect_to` prop.

BogdanUngureanu: I'm not sure I follow. The `redirectTo` property is added to the `generatePath` function that generates a path. We need this so that we pass `redirect_to` parameter between the `Yearly` and `Monthly` toggles.

Otherwise,  when the user switches between plan types, it'll lose the `redirect_to` parameter.

client/my-sites/plans-features-main/plan-type-selector.tsx
```diff
@@ -137,6 +139,8 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > =
 		return null;
 	}
 
+	const redirectTo = props.redirectTo ? { redirect_to: props.redirectTo } : {};
```
tyxla: I'd say the naming of that constant is a bit confusing. `props.redirectTo` comes in as a `string | null | undefined`, and then we name a variable the same way, but it's an object. I think this can be resolved by renaming the constant to something else, like `redirectProps` or `additionalPathProps` for example.

client/my-sites/theme/main.jsx
```diff
@@ -922,6 +927,13 @@ class ThemeSheet extends Component {
 
 		return (
 			<Main className={ className }>
+				{ siteSlug && (
```
tyxla: Why do we display the global notices only when there's a site slug? What's the rationale? How would this make sense in a year from now when we want to render a notice on this page and we don't care about the current site and if there's even a current site?

client/my-sites/plans-features-main/plan-type-selector.tsx
mmtr: > The `redirectTo` property is added to the `generatePath` function that generates a path

Oh, I completely missed that, sorry. I thought it was a property for the `SegmentedControl.Item` component.

client/my-sites/plans-features-main/plan-type-selector.tsx
BogdanUngureanu: The prop was already implemented but not defined in the prop type. Since I didn't knew precisely what value types are accepted, I made it as permissive as possible.

client/my-sites/theme/main.jsx
BogdanUngureanu: I wanted to limit the notices only for logged-out screens, but I guess it makes sense to make it available for all screens.

BogdanUngureanu:
I made some adjustments to the PR and also added the global-notices component. Looks like when the redirect_to parameter is a relative path, the global notices are displayed on the theme/:theme page 🎉 

This means this PR [also fixes this issue](https://github.com/Automattic/wp-calypso/issues/72566).

client/my-sites/theme/main.jsx
```diff
@@ -863,6 +864,8 @@ class ThemeSheet extends Component {
 				( isExternallyManagedTheme &&
 					( ! isMarketplaceThemeSubscribed || ! isSiteEligibleForManagedExternalThemes ) );
 
+			const noop = () => {};
```
tyxla: We could move this out of the component so it would be declared only once, and not on every re-render. Theoretically, this might save us unnecessary updates and re-renders.

client/my-sites/theme/main.jsx
```diff
@@ -922,6 +925,7 @@ class ThemeSheet extends Component {
 
 		return (
 			<Main className={ className }>
+				<AsyncLoad require="calypso/components/global-notices" placeholder={ null } id="notices" />
```
tyxla: I'd suggest moving this below the `Query*`, `DocumentHead`, and `PageViewTracker` components. All of them render nothing but perform some sort of queries or actions under the hood. The first time we render something is where we call `this.renderBar()`, so we might move this line right above it.

tyxla (approved):
This looks good to me! 👍 

Left a couple of optional suggestions, otherwise 🚀

PR #72401, opened 2023-01-21
Add code path for eCommerce trial on My Plan page

### LINKED ISSUE #72102: Task: initial wiring for eCommerce-specific plans content
### Details

While we don't have final designs for the new logged-in plans page or the My Plans page, we can create some initial PRs that identify when the site has the eCommerce trial plan, and sets up separate code paths when that is the case. As a starting point, we can render exactly the same content via the two code paths, but it will reduce our work once we do start implementation of the new designs.

### Checklist

- [x] Ensure we have a separate code path for the My Plans page when the site has the eCommerce trial
- [x] Ensure we have a separate code path for the Plans page when the site has the eCommerce trial

### Related

_No response_

nelsonec87:
#### Proposed Changes

* As part of https://github.com/Automattic/wp-calypso/issues/72102, this PR adds the new code path in the My Plan page for when the selected site has the eCommerce trial plan.

#### Testing Instructions

* Create a site with the new eCommerce Trial plan
* Open the My Plan page (`/plans/my-plan/<site-slug>`)
* Check if the special path was triggered:
image
* Sites with other plans shouldn't be affected.

Fixes  https://github.com/Automattic/wp-calypso/issues/72102

### DIFF
--- a/client/my-sites/plans/current-plan/index.jsx
+++ b/client/my-sites/plans/current-plan/index.jsx
@@ -155,38 +155,45 @@ class CurrentPlan extends Component {
 		return <PaidPlanThankYou />;
 	}
 
-	renderEcommerceTrialPage() {
-		const { path, selectedSiteId, translate } = this.props;
+	renderMain() {
+		const { selectedSiteId, selectedSite, showJetpackChecklist, translate } = this.props;
 		const isLoading = this.isLoading();
+		const currentPlanSlug = selectedSite.plan.product_slug;
+		const planTitle = getPlan( currentPlanSlug ).getTitle();
+		const planFeaturesHeader = translate( '{{planName/}} plan features', {
+			components: { planName: <>{ planTitle }</> },
+		} );
 
 		return (
-			<Main className="current-plan" wideLayout>
-				<DocumentHead title={ translate( 'My Plan' ) } />
-				<FormattedHeader
-					brandFont
-					className="current-plan__page-heading"
-					headerText={ translate( 'Plans' ) }
-					subHeaderText={ translate(
-						'Learn about the features included in your WordPress.com plan.'
-					) }
-					align="left"
-				/>
-				<div className="current-plan__content">
-					<QuerySites siteId={ selectedSiteId } />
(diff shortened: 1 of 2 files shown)


client/my-sites/plans/current-plan/index.jsx
```diff
@@ -154,6 +155,40 @@ class CurrentPlan extends Component {
 		return <PaidPlanThankYou />;
 	}
 
+	renderEcommerceTrialPage() {
```
daledupreez: This implementation effectively removes a _lot_ of logic around prompts and Tracks events. Might it be cleaner to have a conditional that injects our custom content?

@verofasulo, this may also be an area where we need your input, because there are multiple notices and warnings that we currently show on the My Plan page, and we need to work out how we should handle//inject them in the new layout.

daledupreez: In terms of the implementation, @nelsonec87, it may be worth pulling some of the shared code like the `<Query*>` components and the `TrackComponentView` out into helper functions and then inject them in both locations or code paths. Right now there is a _lot_ of duplication that feels like it is really likely to introduce bugs in the long term.

client/my-sites/plans/current-plan/index.jsx
```diff
@@ -154,6 +155,47 @@ class CurrentPlan extends Component {
...
+					placeholder={ null }
+					plan={ currentPlanSlug }
+					isPlaceholder={ isLoading }
+				/>
+			</>
+		);
+	}
+
+	renderEcommerceTrialPage() {
+		return <div className="ecommerce-trial">{ this.renderMain() }</div>;
```
daledupreez: It may be worth updating this to use `ecommerce-trial-wrapper` for the class so it's clearer.
```suggestion
		return <div className="ecommerce-trial-wrapper">{ this.renderMain() }</div>;
```

daledupreez (approved):
Tested and looks good to me. You may want to add a `Fixes` to the PR description to link the task correctly, and it may be worth updating the class name before merging.

PR #72402, opened 2023-01-21
Add code path for eCommerce trial on Plans page

nelsonec87:
#### Proposed Changes

* As part of https://github.com/Automattic/wp-calypso/issues/72102, this PR adds the new code path in the Plans page for when the selected site has the eCommerce trial plan.

#### Testing Instructions

* Create a site with the new eCommerce Trial plan
* Open the My Plan page (`/plans/<site-slug>`)
* Check if the special path was triggered (`div.ecommerce-trial` is added as a wrapper):
image
* Sites with other plans shouldn't be affected.

### DIFF
--- a/client/my-sites/plans/main.jsx
+++ b/client/my-sites/plans/main.jsx
@@ -163,7 +163,7 @@ class Plans extends Component {
 	}
 
 	renderEcommerceTrialPage() {
-		return <div className="ecommerce-trial">{ this.renderPlansMain() }</div>;
+		return <div className="plans__ecommerce-trial-wrapper">{ this.renderPlansMain() }</div>;
 	}
 
 	render() {
--- a/client/my-sites/plans/main.jsx
+++ b/client/my-sites/plans/main.jsx
@@ -163,7 +163,7 @@ class Plans extends Component {
 	}
 
 	renderEcommerceTrialPage() {
-		return this.renderPlansMain();
+		return <div className="ecommerce-trial">{ this.renderPlansMain() }</div>;
 	}
 
 	render() {
--- a/client/my-sites/plans/main.jsx
+++ b/client/my-sites/plans/main.jsx
@@ -163,44 +163,7 @@ class Plans extends Component {
 	}
 
 	renderEcommerceTrialPage() {
-		const { selectedSite, translate, canAccessPlans, domainAndPlanPackage } = this.props;
-		const description = translate(
-			'See and compare the features available on each WordPress.com plan.'
-		);
-
-		return (
-			<div>
-				{ selectedSite.ID && <QuerySitePurchases siteId={ selectedSite.ID } /> }
-				<DocumentHead title={ translate( 'Plans', { textOnly: true } ) } />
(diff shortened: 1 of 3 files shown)


client/my-sites/plans/main.jsx
```diff
@@ -157,16 +162,65 @@ class Plans extends Component {
...
+
 		if ( ! selectedSite || this.isInvalidPlanInterval() || ! currentPlan ) {
 			return this.renderPlaceholder();
 		}
 		const description = translate(
 			'See and compare the features available on each WordPress.com plan.'
 		);
+
+		if ( isEcommerceTrial ) {
+			return this.renderEcommerceTrialPage();
```
daledupreez: It feels like it would be much cleaner to have `this.renderEcommerceTrialPage()` called via conditional code where we currently call `this.renderPlansMain()` so we're keeping all of the existing logic in place.

daledupreez: Also, it may help to have the new function simply call the existing function (i.e. have it call `this.renderPlansMain()`) possibly with a light wrapper div, as we then won't have really ugly placeholder content visible.

nelsonec87: Thanks, I left it in a way that can be merged without impacting any user.

client/my-sites/plans/main.jsx
```diff
@@ -157,16 +162,24 @@ class Plans extends Component {
 		);
 	}
 
+	renderEcommerceTrialPage() {
+		return <div className="ecommerce-trial">{ this.renderPlansMain() }</div>;
```
sixhours: Nit: We might want to make this class name more specific, like `plans__ecommerce-trial-wrapper` or something; just thinking `ecommerce-trial` is short and broad enough to cause conflicts.

sixhours (approved):
Working for me!

client/my-sites/plans/main.jsx
```diff
@@ -157,10 +162,17 @@ class Plans extends Component {
...
 
+	renderEcommerceTrialPage() {
+		return <div className="plans__ecommerce-trial-wrapper">{ this.renderPlansMain() }</div>;
+	}
+
 	render() {
 		const { selectedSite, translate, canAccessPlans, currentPlan, domainAndPlanPackage } =
 			this.props;
 
+		const currentPlanSlug = selectedSite.plan.product_slug;
```
claudiucelfilip, line 173: @nelsonec87 This line seems to be causing an error when you try to create a new site via the sidebar `Add New Site` while you're in the /plans section.

Screenshot 2023-02-03 at 18 28 29

ddc22: I can confirm that this is broken in Production

ddc22: I have raised the issue here
https://github.com/Automattic/wp-calypso/issues/72990

daledupreez: Working on a fix in #72992.

PR #72424, opened 2023-01-23
Marketplace: Decode html in plugin details title

### LINKED ISSUE #70363: Marketplace: Ensure plugin title correctly displays html-encoded characters
Some plugins that contain special characters like `&` do not display correctly the title after installation

For example `Health Check & Troubleshooting`, when the plugin is installed,  `&` in the title changes to `&amp;`. 

Screen Shot on 2022-11-24 at 15:40:37

One suggestion is to use a library like `@wordpress/html-entities` to the decode them.

Related convo: p1669275801456189-slack-C029JEQRVRT

epeicher:
#### Proposed Changes

* Use the utility `decodeEntitites` from `calypso/lib/formatting` to decode any html-encoded characters in the title of the Plugin Details screen

|Before | After|
|-------|------|
|CleanShot 2023-01-23 at 13 46 41@2x|CleanShot 2023-01-23 at 13 46 27@2x|

#### Testing Instructions

- On a site with a Business or eCommerce plan, navigate to `/plugins/:siteId`
- Search for a plugin that contains special HTML characters in the title, e.g. `Health & Check Troubleshooting`
- Click on the plugins to open the details screen
- Check that both the title of the plugin and the breadcrumb navigation on top do not contain html-encoded characters

#### Pre-merge Checklist

~- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?~
~- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)~
~- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?~
~- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~

Fixes #70363

### DIFF
--- a/client/components/breadcrumb/index.tsx
+++ b/client/components/breadcrumb/index.tsx
@@ -3,6 +3,7 @@ import styled from '@emotion/styled';
 import { useTranslate } from 'i18n-calypso';
 import { Key } from 'react';
 import InfoPopover from 'calypso/components/info-popover';
+import { decodeEntities } from 'calypso/lib/formatting';
 
 const flexAligned = {
 	display: 'flex',
@@ -100,7 +101,7 @@ const Breadcrumb: React.FunctionComponent< Props > = ( props ) => {
 		const [ item ] = items;
 		return (
 			<StyledItem>
-				<StyledRootLabel>{ item.label }</StyledRootLabel>
+				<StyledRootLabel>{ decodeEntities( item.label ) }</StyledRootLabel>
 				{ renderHelpBubble( item ) }
 			</StyledItem>
 		);
@@ -112,7 +113,7 @@ const Breadcrumb: React.FunctionComponent< Props > = ( props ) => {
 		return (
 			<StyledBackLink href={ urlBack }>
 				<Gridicon icon="chevron-left" size={ 18 } />
-				{ label }
+				{ decodeEntities( label ) }
 			</StyledBackLink>
 		);
 	}
@@ -126,7 +127,7 @@ const Breadcrumb: React.FunctionComponent< Props > = ( props ) => {
 						{ item.href && index !== items.length - 1 ? (
 							<a href={ item.href }>{ item.label }</a>
 						) : (
-							<span>{ item.label }</span>
+							<span>{ decodeEntities( item.label ) }</span>
 						) }
 						{ renderHelpBubble( item ) }
 					</StyledLi>
--- a/client/my-sites/plugins/plugin-details-header/index.jsx
+++ b/client/my-sites/plugins/plugin-details-header/index.jsx
@@ -3,7 +3,7 @@ import { useSelector } from 'react-redux';
 import Badge from 'calypso/components/badge';
 import { useLocalizedMoment } from 'calypso/components/localized-moment';
 import { formatNumberMetric } from 'calypso/lib/format-number-compact';
-import { preventWidows } from 'calypso/lib/formatting';
+import { preventWidows, decodeEntities } from 'calypso/lib/formatting';
 import PluginRatings from 'calypso/my-sites/plugins/plugin-ratings/';
 import { useLocalizedPlugins } from 'calypso/my-sites/plugins/utils';
 import { getSelectedSite } from 'calypso/state/ui/selectors';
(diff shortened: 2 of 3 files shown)


renancarvalho:
Hi @epeicher 👋 Thank you for working on this issue, really good job! 👍 

While testing, I accidentally saw another place that had the same issue, basically when removing a plugin, there is a modal asking to confirm the removal. I have pushed a commit that will fix it as well. https://github.com/Automattic/wp-calypso/pull/72424/commits/a75b5214b88ef6e5d8a5c1376e7e6bfff6ea476d

image

But this made me think about other places that we don't know that maybe have the same issue. What do you think if we have the html decode on the reducer instead of the react component ? this way we can apply the decode just once and it would work everywhere. 

I also noted that a plugin has, name ( which we are fixing ) and description which is in any human language and could also have special chars not being handled. Perhaps we could threat both on the reducer? What do you think?

We could add something like `reducer.js`

```js
case PLUGINS_RECEIVE: {
    const espapedPluginData = action.data.map( ( pluginData ) => ( {
        ...pluginData,
         // decode at root source just once.
        name: decodeEntities( pluginData.name ),
        // maybe fix description too here.
    } ) );
    return { ...state, [ action.siteId ]: espapedPluginData };
}

```

I am happy to implement and test these changes if you think they make sense.

Thank you! 🙇

epeicher:
Good catch @renancarvalho ! And thank you very much for fixing it!

Your suggestion of decoding in the reducers makes sense to me, and it is actually implemented when normalizing the data for the plugins [here](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/lib/plugins/utils.js#L203). The code that uses that method is the one that reads `.org` plugins [here](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/state/plugins/wporg/actions.js#L49) and the code that reads the `.com` plugins [here](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/data/marketplace/use-wpcom-plugins-query.ts#L100), both initially called when populating plugin data [here](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/my-sites/plugins/plugin-details.jsx#L141).

So it seems that when the plugin is installed, the flow to populate the plugin details is different or some other component is overriding that, I will investigate a bit further.

Additionally, the fix I added to the breadcrumb [d49c4f3](https://github.com/Automattic/wp-calypso/pull/72424/commits/d49c4f3ebf88c8152a9958c069849497f4dded3b) seems safe as that component is shared between different pages and it is good that it ensures the labels being decoded.

epeicher:
Hi @renancarvalho , I have been investigating and from the four possible data sources (`esPlugin`, `wpcomPlugin`, `wporgPlugin` and `plugin` in [the following method](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/my-sites/plugins/plugin-details.jsx#L170-L171)) that populate the plugin details, two of them normalize the plugin data as indicated [in my previous comment](https://github.com/Automattic/wp-calypso/pull/72424#issuecomment-1400439150) **but** the `plugin` data source that initially comes from [this group of reducers](https://github.com/Automattic/wp-calypso/blob/d49c4f3ebf88c8152a9958c069849497f4dded3b/client/state/plugins) does not update or assume any shape of the data.

Although it would be simple to add the `decodeEntities` to `plugin.name` in each reducer, I think is safer to let those reducers as they are and apply the formatting just before displaying the data as we have already done.

We can always create follow-up tasks if we find another occurrence of a bad decoded title.

Regarding decoding the plugin's description, I have not seen problems with them checking a few plugins (so not very thorough testing 😅). But I would prefer to fix that as a follow-up task unless you can come-up with an example of a body badly escaped. That way, we can fix this issue and iterate faster.

What do you think? 😄

renancarvalho:
HI @epeicher, sounds good to me, I think the changes we did are fine to go and perhaps we are trying to fix a problem that we don't have 😄 

Thank you for the deep investigation! 👍

renancarvalho (approved):
Have tested and these changes are good to go! 👍

Issue #72353, opened 2023-01-19
Email Stats: we need to cancel deprecated requests to the backend.


JuanLucha:
When the user clicks on a new period (hours or days) or if they click in the left or right arrows, this triggers a new fetch of the data from the backend.

![Image](https://user-images.githubusercontent.com/3832570/213558843-a04747e8-1e3b-4456-93c7-9dacd5417d75.png)

However, the user can click multiple times, and that will pile up the requests to the backend. It's possible that those responses doesn't come at the same order they were requested, making the data shown to the user wrong and inconsistent.

We need to make the previous request to be canceled when the new one is made.

TimBroddin:
I agree. 

`fetch` has an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) which can be called to abort a request. However, we'd have to go deep into wpcom.js's internals to be able to cancel inflight requests.

TimBroddin:
It seems requests are being made by this package: https://github.com/Automattic/wpcom-xhr-request 
Which in turns uses this package to make XHR requests: https://github.com/ladjs/superagent

JuanLucha:
If there is no way at the moment to cancel the request, we could invalidate the response making a check in the `then` of the promises somehow.

TimBroddin:
I think one possible solution would be to move out our API calling functions from wpcom.js and make them regular `fetch` calls. We wouldn't be the first ones to do that:

<img width="1655" alt="CleanShot 2023-01-20 at 11 29 43@2x" src="https://user-images.githubusercontent.com/528287/213674097-0a6a1a00-ea28-443f-ab2f-8509ee6a6a32.png">

(and we'd also not be the first ones to use an AbortController in the codebase either)

JuanLucha:
I like the idea of using `fetch`

TimBroddin:
Here's a nice example of combining this with Redux: https://spencerfeng.medium.com/create-an-abortable-api-using-abortcontroller-and-abortsignal-d774bed40701

TimBroddin:
It might be a little tricky because we are firing multiple HTTP requests in a method. But nothing tricky is impossible!

yansern:
I have 2 questions in mind (I haven’t look deeper technically so bear with me):

- Does the abort() function exist in wpcom.req.xhr.abort()? And is this somehow exposable through (what I imagine) are layers of abstractions?
[...]

PR #72436, opened 2023-01-23
Email Stats: don't clean state before calling endpoint

### LINKED ISSUE #72375: Email Stats: we are cleaning all the email stats redux data when changing date. We should do that only for `timeline`

JuanLucha:
#### Proposed Changes

**The problem**
Each time the email stats endpoint was called, we were erasing the data in redux. That made the screen to be "loading" every time the user went back and forth, even if they were visiting that date moments ago.

**The solution**
This PR fixes that, but storing all the dates retrieved in redux, but then filtering those at the moment of reading them by the chart component.

#### Testing Instructions

2. Insert manually in the browser address bar an URL with this format: https://calypso.localhost:3000/stats/email/opens/[the-domain-of-your-blog]/day/[post-id]?flags=newsletter/stats. Note that you can change day for week and month.
3. After the chart stops loading, click on "next" arrow. You should see the chart loading circle while the endpoint retrieves the data.
4. After the chart stops loading, click on "previous" arrow. You shouldn't see the chart loading circle, as the chart would use the data stored already on a previous call to that endpoint.
5. Click again on "previous" arrow. You should see the chart loading circle while the endpoint retrieves the data because that date has no data yet on the redux store.

Fixes https://github.com/Automattic/wp-calypso/issues/72375

### DIFF
--- a/client/components/data/query-email-stats/index.jsx
+++ b/client/components/data/query-email-stats/index.jsx
@@ -15,16 +15,7 @@ const requestAlltimeStats = ( siteId, postId, statType, quantity ) => ( dispatch
 	dispatch( requestEmailAlltimeStats( siteId, postId, statType, quantity ) );
 };
 
-function QueryEmailStats( {
-	siteId,
-	postId,
-	period,
-	date,
-	quantity,
-	hasValidDate,
-	statType,
-	isRequesting = false,
-} ) {
+function QueryEmailStats( { siteId, postId, period, date, quantity, hasValidDate, statType } ) {
 	const dispatch = useDispatch();
 
 	useEffect( () => {
@@ -39,7 +30,7 @@ function QueryEmailStats( {
 
 	useEffect( () => {
 		// if hasValidatedDate is false, the date was not set we don't have a post publish date yet
-		if ( siteId && postId > -1 && hasValidDate && ! isRequesting ) {
+		if ( siteId && postId > -1 && hasValidDate ) {
 			dispatch( requestPeriodStats( siteId, postId, period, date, statType, quantity ) );
 		}
 	}, [ dispatch, siteId, postId, hasValidDate, period, date, statType, quantity ] );
--- a/client/my-sites/stats/stats-email-chart-tabs/index.jsx
+++ b/client/my-sites/stats/stats-email-chart-tabs/index.jsx
@@ -130,6 +130,7 @@ const connectComponent = connect(
 			postId,
 			statType,
 			onChangeMaxBars,
+			maxBars,
 		}
 	) => {
 		const siteId = getSelectedSiteId( state );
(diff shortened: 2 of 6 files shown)


TimBroddin (commented):
I'm seeing a weird issue when I'm using the arrow keys and data has not yet come in.

CleanShot 2023-01-23 at 16 11 42

On clicking the right arrow key, I see a chart with only a few days. 

I think this is because the initial action doesn't pass a quantity, so it receives the default number (30). When clicking next the chart tries to display what it has in local state (only those entries). 

CleanShot 2023-01-23 at 16 15 36@2x
_(Note: this screenshot shows a different post than the GIF, but the behavior is the same)_

I can see two approaches:
- Fill the remaining days of the period shown with 0, until we receive the data. This can be confusing to users, since they will see 0 opens, until the data comes in.
- Show a loading state if the period we're trying to show is "incomplete"

JuanLucha:
> I'm seeing a weird issue when I'm using the arrow keys and data has not yet come in.
> 
> CleanShot 2023-01-23 at 16 11 42 [ CleanShot 2023-01-23 at 16 11 42 ](https://user-images.githubusercontent.com/528287/214075462-33dc4078-263a-416e-96fd-bf64851f6cc5.gif) [ ](https://user-images.githubusercontent.com/528287/214075462-33dc4078-263a-416e-96fd-bf64851f6cc5.gif)
> 
> On clicking the right arrow key, I see a chart with only a few days.
> 
> I think this is because the initial action doesn't pass a quantity, so it receives the default number (30). When clicking next the chart tries to display what it has in local state (only those entries).
> 
> CleanShot 2023-01-23 at 16 15 36@2x
> 
> I can see two approaches:
> 
>     * Fill the remaining days of the period shown with 0, until we receive the data. This can be confusing to users, since they will see 0 opens, until the data comes in.
> 
>     * Show a loading state if the period we're trying to show is "incomplete"

Good catch, @TimBroddin ! I like the second approach. I'll go for that 👍

JuanLucha:
@TimBroddin I think we should trigger the endpoint every time the maxBars variable changes. Sadly that will trigger it 3 times on load as we saw the maxBars is calculated every time the size of the chart changes, and it changes from 0 to 18 and then to 26 during the load. 

At the same time, that will be fixed once we implement the mechanism to cancel the previous ones when a new call is done (discussed [here](https://github.com/Automattic/wp-calypso/issues/72353))

TimBroddin:
That means deleting the isRequesting here, right?

```javascript
useEffect( () => {
  // if hasValidatedDate is false, the date was not set we don't have a post publish date yet
  if ( siteId && postId > -1 && hasValidDate && ! isRequesting ) {
	  dispatch( requestPeriodStats( siteId, postId, period, date, statType, quantity ) );
  }
 }, [ dispatch, siteId, postId, hasValidDate, period, date, statType, quantity ] );
```

Maybe throw in a check to make sure `quantity > 0`, so we don't fetch the default 30 days.

JuanLucha:
> That means deleting the isRequesting here, right?
> 
> ```js
> useEffect( () => {
>   // if hasValidatedDate is false, the date was not set we don't have a post publish date yet
>   if ( siteId && postId > -1 && hasValidDate && ! isRequesting ) {
> 	  dispatch( requestPeriodStats( siteId, postId, period, date, statType, quantity ) );
>   }
>  }, [ dispatch, siteId, postId, hasValidDate, period, date, statType, quantity ] );
> ```
> 
> Maybe throw in a check to make sure `quantity > 0`, so we don't fetch the default 30 days.

Good point. I'll implement this and we check.

PR #72383, opened 2023-01-20
Plugins: Update the ratings visualization on the cards

WBerredo:
#### Proposed Changes

Update the ratings display on the Plugins Browser cards.

#### Testing Instructions
* Go to `/plugins`
* Check the ratings on the plugins that contains ratings
* Search for something and check if the .org plugins have ratings

| Before  | After |
| ------------- | ------------- |
|CleanShot 2023-01-20 at 13 07 32@2x|CleanShot 2023-01-20 at 13 12 13@2x|

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Fixes #70906

PR #72441, opened 2023-01-23
Show ratings for paid plugins

### LINKED ISSUE #72382: No reviews are shown for some products

gcsecsey:
#### Proposed Changes

* removes guard to show ratings for marketplace products
* removes scaling of ratings for WPCOM results, as we've already adjusted the scores on the backend

#### Testing Instructions

* Visit `/plugins`
* Verify that ratings are shown for paid products in the featured sections
* Go to `/plugins/browse/paid`
* Verify that ratings are shown for most products (where we have the data)
* Search for a paid plugin (eg. `woocommerce bookings`)
* Verify that both the rating and the number of ratings are shown on the search result (where we have the data)

Page | Before | After
--- | ------ | ----- 
Paid section | CleanShot 2023-01-23 at 16 08 58@2x | CleanShot 2023-01-23 at 16 08 09@2x
Paid page | CleanShot 2023-01-23 at 16 19 30@2x | CleanShot 2023-01-23 at 16 08 33@2x
Search | CleanShot 2023-01-23 at 16 10 03@2x | CleanShot 2023-01-23 at 16 09 48@2x |

#### Pre-merge Checklist

Fixes #72382


client/my-sites/plugins/use-plugins/index.ts
```diff
@@ -107,42 +89,32 @@ const usePlugins = ( {
 			enabled: category === 'featured' && wpcomEnabled,
 		} ) as WPCOMResponse;
 
-	const featuredPlugins = useMemo(
-		() => ( featuredPluginsRaw as Plugin[] ).map( updateWpComRating ),
-		[ featuredPluginsRaw ]
-	);
```
WBerredo: I would rename the `featuredPluginsRaw` as `featuredPlugins`, this way you wouldn't need to change the next occurrences. 

FWIR, I named `featuredPluginsRaw` just because I would recalculate the rating, but it was called `featuredPlugins` before.

WBerredo (commented):
Why don't we have the number of ratings in all views? It's somewhat awkward to show it on search but not on the paid plugins list. 

Other than that, it looks good.

client/my-sites/plugins/use-plugins/index.ts
WBerredo: As you suggested, I've applied a similar change on #72383, but removing the `raw` part as I mentioned.

gcsecsey: > would rename the `featuredPluginsRaw` as `featuredPlugins`, this way you wouldn't need to change the next occurrences.
> 
> FWIR, I named `featuredPluginsRaw` just because I would recalculate the rating, but it was called `featuredPlugins` before.

Thanks, you're right it sounds more straightforward that way! I'll update this PR too soon, this way the branches hopefully won't conflict when merging. 😃

gcsecsey: > Why don't we have the number of ratings in all views? It's somewhat awkward to show it on search but not on the paid plugins list.

I don't know the original cause for this, I went back to check the Figma designs, there the number of ratings are only shown on some cards. I think it's meant to be only missing when we don't have the number:
f84fqtCG4pCnCh2Hq0But5-fi-5669%3A41901
CleanShot 2023-01-24 at 14 37 47@2x

I'll update the PR to show it in all views.

gcsecsey: Actually, the Calypso side looks good, rating numbers aren't hidden on purpose:
https://github.com/Automattic/wp-calypso/blob/7b309f3b6110da1abb41cce4894c5f984443af13/client/my-sites/plugins/plugin-ratings/index.jsx#L119-L121

I think this is caused by some views still being served from the static catalog, which doesn't have the `num_ratings` field and it doesn't store the number of ratings in another field either. For search, we have the data now in the ES index and that's shown correctly. So this is a non-issue, it will be solved by migrating to the dynamic catalog.

WBerredo: Great, thanks for diving into this.

PR #72452, opened 2023-01-23
Marketplace Themes: Adds the Premium badge

### LINKED ISSUE #72201: Marketplace Themes: Missing the indicator that the theme is premium
The star that identifies the theme as premium is not displayed for Marketplace themes.

image (2)

valterlorran:
#### Proposed Changes

* Adds the Premium badge for Marketplace themes on the Theme Showcase.

Screen Shot 2023-01-23 at 17 06 42

#### Testing Instructions

* Navigate to `/themes/{YOU-SITE}?s=Makoney`
* The theme card should contain the Premium badge

Closes #72201

### DIFF
--- a/client/components/theme/index.jsx
+++ b/client/components/theme/index.jsx
@@ -514,6 +514,7 @@ export class Theme extends Component {
 			hasPremiumThemesFeature,
 			isPremiumTheme,
 			didPurchaseTheme,
+			isExternallyManagedTheme,
 		} = this.props;
 		const { name, description, screenshot, style_variations = [] } = theme;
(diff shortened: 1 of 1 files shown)


daledupreez (approved):
Overall this is working as described and is a _big_ improvement over the previous behaviour.

I think we should seriously consider changing the label to be something more like `Marketplace` so there is a clearer distinction between "normal" premium themes and Marketplace themes, but I think that should occur in a follow-up.

There are also some wording issues in some of the related code that I'll tackle in a follow-up PR shortly.

PR #72507, opened 2023-01-24
Checkout: Update all info at once when updating tax info

sirbrillig:
#### Proposed Changes

When updating the tax location fields in checkout, we dispatch two separate actions to the `wpcom-checkout` data store reducer: one for the country code and one for the postal code. However, this is inefficient and if we ever want to include more information here, we will need to add more dispatch calls.

In this PR, we collapse both calls into one call that updates all the fields at once.

#### Testing Instructions

- Add a non-domain product to your cart and visit checkout.
- Edit your billing location and try changing the country and postal code.
- Verify that they change as expected.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/contact-details-container.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/contact-details-container.tsx
@@ -59,13 +59,11 @@ export default function ContactDetailsContainer( {
 		return null;
 	}
 
-	const { updateDomainContactFields, updateCountryCode, updatePostalCode, updateEmail } =
-		checkoutActions;
+	const { updateDomainContactFields, updateSomeFields, updateEmail } = checkoutActions;
 	const contactDetails = prepareDomainContactDetails( contactInfo );
 	const contactDetailsErrors = prepareDomainContactDetailsErrors( contactInfo );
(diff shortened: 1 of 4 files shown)


client/my-sites/checkout/composite-checkout/components/contact-details-container.tsx
```diff
@@ -52,13 +52,12 @@ export default function ContactDetailsContainer( {
 		.filter( ( product ) => isDomainProduct( product ) || isDomainTransfer( product ) )
 		.filter( ( product ) => ! isDomainMapping( product ) )
 		.map( getDomain );
-	const { updateDomainContactFields, updateCountryCode, updatePostalCode, updateEmail } =
+	const { updateDomainContactFields, updateSomeFields, updateEmail } =
```
martian77: This absolutely makes sense to do, I was just wondering if it might be better to name this as something slightly more descriptive like `updateSomeTaxFields`? Throughout, not just here.

sirbrillig: Good idea.

pottedmeat (approved):
- Changed country between postal code and non-postal code countries and saw postal code box added and removed
- Was able to change the value of the postal code when present
- Code looks good and makes sense

PR #72513, opened 2023-01-24
VAT Info: Correct UK country code in VatInfo

sirbrillig:
#### Proposed Changes

At `/me/purchases/vat-details` we provide the option for people to give us their VAT details. We currently use two letter country codes in the dropdown instead of full country names, and we're using `UK` for United Kingdom instead of `GB`, the ISO 2-letter code.

This PR fixes the mistake by changing the form to use `GB`. It also adds a fallback to support countries that were already set to `UK`.

Partially takes care of https://github.com/Automattic/payments-shilling/issues/1276

Before:

Screenshot 2023-01-24 at 2 06 26 PM

After:

Screenshot 2023-01-24 at 2 07 24 PM

#### Testing Instructions

First, set up VAT data with `UK` as the country code. One way to do this is to use trunk's version of calypso and visit `/me/purchases/vat-details`. You can use the VAT number `553557881` if the API is sandboxed.

Next, using this PR's branch:

- Verify that the selected country code reads `GB`.

### DIFF
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -189,9 +189,15 @@ function CountryCodeInput( {
 		'SE',
 		'SI',
 		'SK',
-		'UK',
+		'GB',
 		'XI',
 	];
+
+	// Some historical country codes were set to 'UK', but that is not a valid
+	// country code. It should read 'GB'.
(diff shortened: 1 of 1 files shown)


martian77 (approved):
LGTM!

PR #72378, opened 2023-01-20
Copy site: marketing products to cart for copy site

kozer:
#### Proposed Changes

Cleanup and fix code for including marketing plugins when the user copies his site.

#### Testing Instructions

1. Create a site with `Business plan`
2. Navigate to `Plugins > Add new`. Find `Yoast SEO premium` and buy a subscription for that plugin.
3. Ensure that your site is atomic.
4. Navigate to `/sites` in calypso.
5. Find your site and select `Copy Site` in the right corner (the three dots menu)
5. Make sure that the Yoast subscriptions exist in your cart items

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #

PR #72291, opened 2023-01-18
Copy Site: Include Marketplace Plugins when purchasing a new Copy Site

danielbachhuber:
See https://github.com/Automattic/dotcom-forge/issues/1514

## Proposed Changes

Includes any Marketplace plugins on the origin site in the copy site checkout flow.

## Testing Instructions

1. Create a new Business site and take it Atomic.
2. Purchase WooCommerce Subscriptions.
3. Initiate 'Copy site' flow.
4. Verify WooCommerce Subscriptions is included in the checkout cart.
5. Purchase the site copy.
6. Verify WooCommerce Subscriptions add-on is included on your new site.

PR #70529, opened 2022-11-29
Gifting: Remove all gift products from the cart whenever one is removed

DavidRothstein:
#### Proposed Changes

Subscription gifts need to be all or nothing -- for example, if someone manages to gift a plan renewal but not the corresponding primary domain renewal that was automatically added to the shopping cart, then the site would likely break when the domain expires, so it would have wound up being a confusing and not very useful gift.

Currently, this is mostly enforced on the server, but the user experience is bad -- for example, trying to remove the gift domain renewal from the cart does nothing (pdtkmj-Ks-p2#comment-1269).

This pull request makes it so removing any gift product from the cart automatically removes all the others, and changes the confirmation message for removing a gift (which previously didn't quite make sense for someone gifting a subscription to someone else's site) to match.

Before (removing a plan or domain, and in some cases the removal wouldn't actually work after you press "Continue"):

remove-plan-before
remove-domain-before

After (removing either the plan or the domain):

remove-gift-after

#### Testing Instructions

The easiest way to test this is to have a site with a Personal plan and custom domain, where the custom domain is set to primary, and where subscription gifting is turned on via `/settings/general/`.

Log in as a different user, and visit a URL like `/checkout/personal-bundle/gift/[ID]`, where ID is the subscription ID (obtained from the server, or from the number in the URL when the original user clicks through to manage their subscription from the Purchases page).  This should put the plan and domain renewal in your shopping cart as a gift.

Test that if you go to remove either the plan or domain from the cart, the expected confirmation message is shown, and that going through with the removal causes both products to be removed from the cart and the cart to be emptied.

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
[...]

PR #69837, opened 2022-11-07
Checkout: remove post-checkout polling for site for create_new_blog

sirbrillig:
#### Background

The transactions endpoint supports a property on the shopping cart object called `create_new_blog`. If set, the endpoint will create a new site before the transaction is processed. However, even though [we set this property on the frontend](https://github.com/Automattic/wp-calypso/blob/2c707eaab587ae48cd342fe76d37546eed2a68ef/client/my-sites/checkout/composite-checkout/lib/translate-cart.ts#L89) for any transaction where there is no currently selected site (which should be domain-only checkout and Jetpack siteless checkout, although [the latter may have a special case](https://github.com/Automattic/wp-calypso/blob/2c707eaab587ae48cd342fe76d37546eed2a68ef/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx#L200)), it doesn't appear to be ever used on the backend to create a new site (see https://github.com/Automattic/payments-shilling/issues/759#issuecomment-1306365375). This is because [logged-out and siteless flows](https://github.com/Automattic/wp-calypso/blob/17f26290ffabb9a9d6437d5166a0e8f5265cd70d/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx#L139) that currently go through checkout (except for Jetpack siteless checkout which understandably does not need to create a site) [create the site before the transaction is submitted](https://github.com/Automattic/wp-calypso/blob/17f26290ffabb9a9d6437d5166a0e8f5265cd70d/client/my-sites/checkout/composite-checkout/lib/submit-wpcom-transaction.ts#L32-L34).

While it may not yet be safe to remove the `create_new_blog` property entirely (since it may be used on the backend cart to signal that a site _will_ be created), we can probably remove some of the machinery that uses it.

#### Proposed Changes

In this PR we remove code that causes some post-checkout flows (for non-redirect payment methods which have `create_new_blog` and a domain in the purchase) to poll the sites endpoint after the transactions endpoint returns a successful response. This code was probably intended to prevent issues where the post-checkout page would try to display information about the newly created site and fail, but I think this is unlikely ever to happen since post-checkout for new sites already [causes a hard-redirect](https://github.com/Automattic/wp-calypso/blob/17f26290ffabb9a9d6437d5166a0e8f5265cd70d/client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx#L195-L199) which will probably reload the sites list.

[...]

PR #72523, opened 2023-01-24
Copy Site: hide remove from cart buttons on checkout

sejas:
#### Proposed Changes

I've found that users could remove products from the cart and still create the copied site. 

* In this PR we don't show the `remove from cart` buttons for any product. The user is forced in the UI to buy the site plan and extra products/plugins.

#### Testing Instructions

* Go to `/sites`
* Clone a business site. Better if it includes plugins from the marketplace.
* Observe that in the checkout the `remove from cart` buttons are not visible, not in the DOM either.

#### Screenshots

| **Before** | **After** |
|---|---|
| Screenshot 2023-01-24 at 22 08 59 | Screenshot 2023-01-24 at 22 08 46 |

#### Pre-merge Checklist

Related to :
- https://github.com/Automattic/wp-calypso/pull/72291
- https://github.com/Automattic/wp-calypso/pull/72378
- https://github.com/Automattic/dotcom-forge/issues/1514

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
@@ -13,9 +13,11 @@ import {
 } from '@automattic/wpcom-checkout';
 import styled from '@emotion/styled';
 import { useState } from 'react';
+import { isCopySiteFlow } from 'calypso/../packages/onboarding/src';
 import { useExperiment } from 'calypso/lib/explat';
 import { isWcMobileApp } from 'calypso/lib/mobile-app';
 import { useGetProductVariants } from 'calypso/my-sites/checkout/composite-checkout/hooks/product-variants';
+import { getSignupCompleteFlowName } from 'calypso/signup/storageUtils';
 import { ItemVariationPicker } from './item-variation-picker';
 import type { OnChangeItemVariant } from './item-variation-picker';
 import type { Theme } from '@automattic/composite-checkout';
@@ -177,7 +179,12 @@ function LineItemWrapper( {
 } ) {
 	const isRenewal = isWpComProductRenewal( product );
 	const isWooMobile = isWcMobileApp();
-	const isDeletable = canItemBeRemovedFromCart( product, responseCart ) && ! isWooMobile;
+	let isDeletable = canItemBeRemovedFromCart( product, responseCart ) && ! isWooMobile;
+
+	const signupFlowName = getSignupCompleteFlowName();
+	if ( isCopySiteFlow( signupFlowName ) ) {
+		isDeletable = false;
+	}
 
 	const shouldShowVariantSelector =
 		onChangePlanLength && ! isWooMobile && ! isRenewal && ! hasPartnerCoupon;
--- a/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
@@ -182,7 +182,7 @@ function LineItemWrapper( {
 	let isDeletable = canItemBeRemovedFromCart( product, responseCart ) && ! isWooMobile;
 
 	const signupFlowName = getSignupCompleteFlowName();
-	if ( isCopySiteFlow( signupFlowName ) ) {
+	if ( isCopySiteFlow( signupFlowName ) && ! product.is_domain_registration ) {
 		isDeletable = false;
 	}
 
--- a/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
@@ -1,5 +1,6 @@
 import { isJetpackPurchasableItem } from '@automattic/calypso-products';
 import { FormStatus, useFormStatus } from '@automattic/composite-checkout';
+import { isCopySiteFlow } from '@automattic/onboarding';
 import {
 	canItemBeRemovedFromCart,
 	getCouponLineItemFromCart,
@@ -13,7 +14,6 @@ import {
 } from '@automattic/wpcom-checkout';
 import styled from '@emotion/styled';
 import { useState } from 'react';
-import { isCopySiteFlow } from 'calypso/../packages/onboarding/src';
 import { useExperiment } from 'calypso/lib/explat';
 import { isWcMobileApp } from 'calypso/lib/mobile-app';
 import { useGetProductVariants } from 'calypso/my-sites/checkout/composite-checkout/hooks/product-variants';


kozer (approved):

sejas:
Hi @Automattic/shilling, I'm adding a customization related to the copy site flow. Can I get someone to review it?

sirbrillig (approved):
Seems safe to me! Thanks for the ping.

DavidRothstein:
>I've found that users could remove products from the cart and still create the copied site.

But they can still do that with this pull request, can't they?

First, it looks to me like the site is already copied **before** they enter checkout, so the copied site is created regardless of whether they buy anything anyway.  And second, they can still remove the products simply by exiting the cart:

exit-cart
remove-items

Also, you are giving people a choice of what to buy **before** they enter the cart, so I'm not sure why you would take that choice away after they get there?  For example, when I copy a site, it seems to give me a choice of whether to get a domain or not:

choose-domain

But now with this pull request you're basically preventing them from changing their mind on that and removing the domain once they get to checkout?

checkout

sejas:
@DavidRothstein , Thanks for testing this PR. Here are my comments:

> But they can still do that with this pull request, can't they?

No, they can't.

> First, it looks to me like the site is already copied before they enter checkout, so the copied site is created regardless of whether they buy anything anyway. And second, they can still remove the products simply by exiting the cart:

We create a simple, free and empty site before entering the cart. If they empty the cart, they will end up having only a simple empty site. This is how `Link in bio` and `Videopress` flows are currently working, creating the simple site before going to the checkout page.
If there is already a way to create a simple site after the purchase, that would be something I would love to learn about.

> Also, you are giving people a choice of what to buy before they enter the cart, so I'm not sure why you would take that choice away after they get there? For example, when I copy a site, it seems to give me a choice of whether to get a domain or not:

Great point. We force the user to buy the same exact plan and extra plugins they currently have in the original site. The only part they can decide is to buy a domain or not and Yes!, you are right, it seems fair they have the option to remove the domain from the cart, so I just added this option here https://github.com/Automattic/wp-calypso/pull/72523/commits/ffd9bf101742aa87ec63de7f4c3d979ca7690fe4

@DavidRothstein, do you think that forcing the user to buy the same plans for their destination site copy makes more sense now?

DavidRothstein:
Ah, thanks, I missed that the new site was empty.  In that case this pull request is probably OK, now that you've restored the ability to remove the domain from the cart.

@michaeldcain did point out that a more traditional way to force them to buy the plan/plugin together would be to allow them to remove items from the cart, but just remove all related items (plan/plugins/etc) when one is removed, with a message explaining the consequences.  This is how it works in other cases like domain+email, and a recent example can be found at https://github.com/Automattic/wp-calypso/pull/70529.  This might be a better experience for the user since they won't get confused about items being "stuck" in their cart that can't be removed, but that said, I think you could do it either way.

>If there is already a way to create a simple site after the purchase, that would be something I would love to learn about.

Yes, a couple examples can be found at https://wordpress.com/domains/ (choose the "Just buy a domain" option when it gives you the choice), which uses the `/checkout/no-site` route, and https://cloud.jetpack.com/pricing (buy anything other than Search).  In those cases, the site isn't actually created until after the purchase is completed.  I'm not sure the exact technical details of how those work, but they're ultimately controlled by the `create_new_blog` key on the shopping cart which gets sent to the server.

sirbrillig:
I agree that this PR should be fine now.

> but just remove all related items (plan/plugins/etc) when one is removed, with a message explaining the consequences. This is how it works in other cases like domain+email, and a recent example can be found at https://github.com/Automattic/wp-calypso/pull/70529.

FYI this can also be accomplished on the backend inside the shopping-cart validator, and that is often a better place for it since it's more easily testable. Gifting was a special case, but most of the rules regarding what groups of products to add or remove together are controlled there.

> In those cases, the site isn't actually created until after the purchase is completed. I'm not sure the exact technical details of how those work, but they're ultimately controlled by the create_new_blog key on the shopping cart which gets sent to the server.

Actually, I'm pretty certain that the `create_new_blog` key hasn't been used to do that for some time, although it does have that capability (I've been [working on removing](https://github.com/Automattic/wp-calypso/pull/69837) `create_new_blog` but it's not trivial to do). The logic to create an _account_ and site just before a successful purchase [lives in calypso](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/my-sites/checkout/composite-checkout/lib/submit-wpcom-transaction.ts#L32-L34) but I don't think there's any other existing flow that waits until after checkout to create a new site; signup [creates](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/lib/signup/step-actions/index.js#L276-L277) the site before checkout is reached (also a [few](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/packages/onboarding/src/cart/index.tsx#L125-L126) other [places](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/packages/data-stores/src/site/actions.ts#L73-L74)).

DavidRothstein:
>FYI this can also be accomplished on the backend inside the shopping-cart validator, and that is often a better place for it since it's more easily testable.

One advantage of doing it at least partially on the frontend is that you can display a message explaining what will happen before they go ahead with the removal; however, I agree it is best to enforce on the backend.

>Actually, I'm pretty certain that the `create_new_blog` key hasn't been used to do that for some time, although it does have that capability...The logic to create an _account_ and site just before a successful purchase [lives in calypso](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/my-sites/checkout/composite-checkout/lib/submit-wpcom-transaction.ts#L32-L34) 

Doesn't the code you linked to use `create_new_blog` itself, specifically [here](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts#L51)?

DavidRothstein:
>Doesn't the code you linked to use `create_new_blog` itself, specifically [here](https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts#L51)?

Oh, I see, maybe that code is only a fallback, and the site is usually already created before it runs.

I am confused though because I could have sworn https://wordpress.com/domains/ would still use `create_new_blog` (even if the above code doesn't) but from your logging at https://github.com/Automattic/payments-shilling/issues/759#issuecomment-1306365375 maybe that uses something else now too.

sirbrillig:
> Oh, I see, maybe that code is only a fallback, and the site is usually already created before it runs.

Yeah, `create_new_blog` is just being used as a flag to allow creating a cart without a blog_id _before_ a purchase, but it's redundant because it's set every time there is no blog_id, so we could just change the endpoint to always allow creating siteless carts and then remove the flag.

PR #72525, opened 2023-01-24
Add backup features to checkout

CodeyGuyDylan:
#### Proposed Changes

Add a feature list of VaultPress backup when it is the only item in the cart

P2: p1HpG7-k8M-p2

#### Testing Instructions

1. Check out these changes via `git switch add/backup-features-to-checkout`
2. Start your local environment via `yarn start`
3. Go to https://cloud.jetpack.com/pricing and add VaultPress Backup (10GB) to your cart
4. When you are redirected to the cart, replace `https://wordpress.com` with `http://calypso.localhost:3000/` in the URL
5. The new feature list should now show up above the total
image
6. If you'd like to run the new tests manually, running ` yarn test-client client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx` from the root directory should do the trick

#### Pre-merge Checklist


sirbrillig (approved):
I'll leave a final review to your team, but from a checkout perspective, this looks good to me!

CodeyGuyDylan:
> I'll leave a final review to your team, but from a checkout perspective, this looks good to me!

Thanks!

CodeyGuyDylan:
Rebased ^

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx
```diff
@@ -310,6 +312,12 @@ function CheckoutSummaryFeaturesList( props: {
 	const plans = responseCart.products.filter( ( product ) => isPlan( product ) );
 	const hasPlanInCart = plans.length > 0;
 
+	const jetpackProducts = responseCart.products.filter( ( product ) =>
+		isJetpackProduct( product )
+	);
```
robertsreberski: It just looks nice, but it's totally extra. 

```suggestion
	const jetpackProducts = responseCart.products.filter( isJetpackProduct );
```

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx
```diff
@@ -399,6 +411,24 @@ function CheckoutSummaryFeaturesListDomainItem( { domain }: { domain: ResponseCa
...
 
+function CheckoutSummaryJetpackProductFeatures( { product }: { product: ResponseCartProduct } ) {
+	const translate = useTranslate();
+	const productFeatures = getJetpackProductFeatures( product, translate );
+
+	return (
+		<>
+			{ productFeatures.map( ( feature ) => {
+				return (
+					<CheckoutSummaryFeaturesListItem key={ String( feature ) }>
```
robertsreberski: From what I see, the type of `feature` is already `string`. Do we really need to cast it here?

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
```diff
@@ -0,0 +1,20 @@
...
+import { useTranslate } from 'i18n-calypso';
+import type { WithSnakeCaseSlug, WithCamelCaseSlug } from '@automattic/calypso-products';
+
+export default function getJetpackProductFeatures(
+	product: WithSnakeCaseSlug | WithCamelCaseSlug,
+	translate: ReturnType< typeof useTranslate >
+): string[] {
+	if ( isJetpackBackupT1( product ) ) {
+		return [
+			String( translate( 'Real-time cloud backups' ) ),
```
robertsreberski: Casting here might be redundant:
- you've already set return type in function definition
- `translate()` is typed and returns `string`

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
```diff
@@ -0,0 +1,20 @@
+import { isJetpackBackupT1 } from '@automattic/calypso-products';
+import { useTranslate } from 'i18n-calypso';
+import type { WithSnakeCaseSlug, WithCamelCaseSlug } from '@automattic/calypso-products';
+
+export default function getJetpackProductFeatures(
+	product: WithSnakeCaseSlug | WithCamelCaseSlug,
```
robertsreberski: This function (`getJetpackProductFeatures()`) is currently run only in one place, and the `ResponseCartProduct` type has a product slug defined with snake case `product_slug`. Here, we introduce a generic type just so we can cast it back again in `isJetpackBackupT1()`.

Not sure if it's necessary, especially in typescript code.

packages/calypso-products/src/is-jetpack-backup-t1.ts
```diff
@@ -0,0 +1,7 @@
+import { camelOrSnakeSlug } from './camel-or-snake-slug';
+import { isJetpackBackupT1Slug } from './is-jetpack-backup-t1-slug';
+import type { WithCamelCaseSlug, WithSnakeCaseSlug } from './types';
+
+export function isJetpackBackupT1( product: WithCamelCaseSlug | WithSnakeCaseSlug ): boolean {
```
robertsreberski: It feels a bit redundant to create a drill-down into two separate exported functions to run this one-line check:
`( JETPACK_BACKUP_T1_PRODUCTS as ReadonlyArray< string > ).includes( productSlug )`

If we really want to make a function for it, maybe we could consider a locally defined function in the file:
client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
robertsreberski: Also, at this point, we probably don't need to get access to the whole `product` if we only use product slug.

robertsreberski (changes requested):
The code works well, and I see the approach taken here.

But I just want to make sure we don't add redundant code by adding a couple of global methods just to get product features copy. Especially since we have only one product there.

I would start small by defining it locally and then scale it further if we see the potential of the logic becoming reusable in more parts of the code.

I also see some redundant casting in some places. It was there probably because the code was created in Javascript at first, but we can break from this casting now that we have types.

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
sirbrillig: `translate()` actually returns `TranslateResult`, since the function can sometimes produce a React element if it uses interpolation. Obviously this string is not interpolated, but if you want TypeScript to know this you have to explicitly tell it by passing `{ textOnly: true }` like here:

https://github.com/Automattic/wp-calypso/blob/d05dbbe45e09c112baad817541945124f7e6555f/client/components/jetpack/profile-dropdown/index.tsx#L65

...or, you need to cast it to a string.

CodeyGuyDylan: Good to know the reasoning behind this! I will leave these casting as is then

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx
CodeyGuyDylan: I'll refer back to [this comment](https://github.com/Automattic/wp-calypso/pull/72525#discussion_r1088255273) on this one 😄

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx
CodeyGuyDylan: I totally spaced that we can do this 😅

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx
CodeyGuyDylan: You are correct on this one, nice catch!

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
CodeyGuyDylan: You're right, I was modeling after most of the other types of these functions, but the type does seem overly complicated here. I have simplified it as well as just used the slug to determine if it is VaultPress Backup T1

packages/calypso-products/src/is-jetpack-backup-t1.ts
CodeyGuyDylan: Yes I think you are correct, I have updated it slightly to only check the slug now

CodeyGuyDylan:
>But I just want to make sure we don't add redundant code by adding a couple of global methods just to get product features copy. Especially since we have only one product there.

In regards to this, I wanted to make sure it was set up the same-ish way as other feature lists are in checkout, and in a way that would be very easy to add more if we wanted to. I know it may seem over the top in some aspects but I felt it was worth it due to the complexity of the checkout in general. I am willing to have a conversation about simplifying it if you think it is important 😄

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
robertsreberski: Ah, I see @sirbrillig; I didn't know we can set it like this, thank you for specifying that!

However, I meant that `TranslateResult` is a union type that includes `string`. That way, we don't need to force cast the value to string, but setting a return type for the function or variable (e.g. `const a: string = translate( ... )` or `function foo(): string { return translate( ... ) }`) will already be correct from the Typescript perspective. And `getJetpackProductFeatures()` defines the return type.

Sorry for making my comment confusing about that. I meant we cast here twice instead of only in the function definition.

---

But I'll leave it as a suggestion, not a blocker.

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
```diff
@@ -0,0 +1,20 @@
+import { isJetpackBackupT1Slug } from '@automattic/calypso-products';
+import { useTranslate } from 'i18n-calypso';
+import type { WithSnakeCaseSlug } from '@automattic/calypso-products';
+
+export default function getJetpackProductFeatures(
+	product: WithSnakeCaseSlug,
+	translate: ReturnType< typeof useTranslate >
+): string[] {
+	if ( isJetpackBackupT1Slug( product.product_slug ) ) {
```
robertsreberski, line 9: Would it be more reusable to make a more generic lib function `isProductSlug( JETPACK_BACKUP_T1_PRODUCTS, product.product_slug )` instead of defining the function explicitly for Jetpack Backup?

robertsreberski (approved):
> In regards to this, I wanted to make sure it was set up the same-ish way as other feature lists are in checkout, and in a way that would be very easy to add more if we wanted to.

Ah, I see; then `getJetpackProductFeatures()` function is completely fine by me! I mostly meant the functions checking the Jetpack Backup slug, but you already reduced the drill down by removing one of the functions.

I left one more comment about this below (and the second about casting) - but by any means, I don't want to block you; these are just suggestions. The PR looks good and tests well!

client/my-sites/checkout/composite-checkout/lib/get-jetpack-product-features.ts
CodeyGuyDylan: I think the reason there isn't one of those yet is potentially to avoid having to import the constants directly from the package and keeping it more contained like that.

The idea of this `isProductSlug` wouldn't really have to do anything with the Calypso packages since you'd essentially be giving it a list of products, the product slug, and asking it if the second parameter is in the array of the first. That could just be re-written as something like this
```
if( JETPACK_BACKUP_T1_PRODUCTS.includes( product.product_slug ) ) {
...
}
```

Which is not the end of the world, but I think I like keeping these functions set in the `calypso-products` package and not to break the consistency here

robertsreberski: That's fair; thanks for explaining it! I'm fine with having it like that then

client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx
```diff
@@ -0,0 +1,168 @@
...
+	return mock;
+} );
+
+const jetpackBackupFeatureList = [
+	'Real-time cloud backups',
+	'10GB of backup storage',
+	'30-day archive & activity log',
+	'One-click restores',
+	'Priority support',
+];
```
creativecoder, line 41: Nice work on adding some meaningful tests!

One minor suggestion for future component tests: it would be nice to pull the feature strings used here from the same place as the component being tested, so that tests don't need to be updated here just because a feature description changes in the component.

PR #72535, opened 2023-01-25
2x faster unit tests

noahtallen:
### Proposed Changes
This is an experiment to move the Calypso Unit Tests from individual build steps into a bash script, and then ultimately parallelize those steps.

From what I can tell, the TeamCity set up works like this:
- There are two servers with 64 cores and like 128GB RAM.
- Each runs 4 agents, giving 16 cores and 32GB per agent.
- Each agent can run only ONE build at a time.

As a result, any free CPU resources during a build run are literally wasted, and can't be used by any other agents or builds in the system.

This means we should probably be trying to max out the system as much as we can _per build._ So lightweight standalone builds are actually _not good_, because they essentially take up all 16 cores for the entire duration, despite not needing that much power.

### Problem with Current Build
Every task runs in serial. As a result, any step which doesn't parallelize well (e.g. doesn't use as many cores as it can) isn't performing optimally. After some resarch, I learned that Typescript doesn't parallelize well. In fact, even [the Deno project notes](https://deno.com/blog/v1#tsc-bottleneck) that `tsc` is a bottleneck. It seems to use only a few cores at a time, which in our system leaves about 12 cores unused.

Plus, while the various jest suites are using as many workers as there are cores, there are still parts of it that probably don't parallelize well (like reporting results, booting up, etc). So running some test suites simultaneously is still a good idea.

### Solution
After **a lot** of testing, I've landed on the following workflow:

In Parallel:
- Run the client test suite with 8 cores
- Run the assorted other test suites in serial with 4 cores
- Run typescript with the remaining CPU

The assorted other suites include package tests, server tests, build tool tests, and the storybook check. These run in serial, because they're each rather short. If we started them all simultaneously, it would eat so much CPU that the other tasks would become starved. (This can introduce timeouts and other flakiness)

The typescript stuff first builds all the packages. This takes a long time -- definitely a majority of the build. Once that's done, we can check types for the client, ETK, and e2e tests in parallel. We have to build the packages first because the other typescript configs depend on the package types being compiled.

This results in about a **2x faster build**. (What takes 6-7 minutes on trunk takes less than 3:30min in this branch.)

### Implementation
I originally wrote a solution in Bash. Unfortunately, Bash stopped being a good solution after I needed to run some things in serial, and others in parallel, all at the same time. The main thing which made Bash not work well was the need to capture all output and essentially re-order it, as well as generalizing how to run tasks in parallel. (And then a subtask in serial.)

So I ultimately wrote it in Node in two parts:
1. A function `runTask` which spawns a child process with the command, captures all the output, makes everything look good for TeamCity with service messages, and then resolves/rejects with the result.
2. A file which creates several tasks, then orchestrates calls to `runTask` using async/await. This file is called by TeamCity.

We can reuse this `runTask` concept for any other build we like. It took **a lot of trial and error** to figure out the TeamCity service messages to integrate with the existing tests. (66 commits, in fact...)

This is what it looks like with service messages working:
image

### Testing Instructions
- TeamCity passes.

Various failure conditions work properly:
- [ ] Client unit test
- [ ] Package unit test
- [ ] Package type issue
- [ ] Client type issue

### DIFF
--- a/.teamcity/_self/projects/WebApp.kt
+++ b/.teamcity/_self/projects/WebApp.kt
@@ -329,7 +329,6 @@ object RunAllUnitTests : BuildType({
 	artifactRules = """
 		test_results => test_results
 		artifacts => artifacts
-		checkstyle_results => checkstyle_results
 	""".trimIndent()
 
 	vcs {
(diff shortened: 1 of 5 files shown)


.teamcity/step-scripts/calypso-unit-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+
+#### Run unit tests
+# TODO: parallelize these
+unset NODE_ENV
+unset CALYPSO_ENV
+yarn test-client --maxWorkers=$JEST_MAX_WORKERS --ci --reporters=default --reporters=jest-teamcity --silent
+yarn test-server --maxWorkers=$JEST_MAX_WORKERS --ci --reporters=default --reporters=jest-teamcity --silent
+yarn test-packages --maxWorkers=$JEST_MAX_WORKERS --ci --reporters=default --reporters=jest-teamcity --silent
+yarn test-build-tools --maxWorkers=$JEST_MAX_WORKERS --ci --reporters=default --reporters=jest-teamcity --silent
+yarn workspaces foreach --verbose --parallel run storybook --ci --smoke-test
```
worldomonation: **thought:non-blockingI** - I have mixed thoughts about moving contents inside a script when that script won't be reused anywhere else. 

On one hand, it shortens the main Kotlin file, making it easier to navigate.
But to make changes a developer must follow the "cookie crumbs" to get to where the actual changes need to be made.

worldomonation (approved):
Really cool! Were you able to establish improved resource utilization?

worldomonation:
Looks like a bit of savings.

image

image

Edit: re-ran it, this time it was 6m01s.

noahtallen:
Well, I hadn't yet parallelized anything 😅

.teamcity/step-scripts/calypso-unit-tests.sh
noahtallen: I mainly did it since I needed to write more normal bash, so I wanted to have syntax highlighting and not do weird string interpolation

noahtallen:
One problem with the current approach is all the output will be out of order. I think I saw some tips about how to fix that which should probably be implemented. But I wanted to see what the performance characteristics would be like on CI first.

Here's a typical build: https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Run_All_Unit_Tests/9362095

Note high CPU utilization for a couple things -- mainly tsc and unit tests for the main client, which is much larger than the other things.

noahtallen:
Some stats:

- total duration on trunk: 6m11s
- total duration here: 4m1s (2m faster)
- tsc on trunk: 2m38s
- tsc on this branch: 1m14s , but it's broken

Higher CPU usage during tsc at least.
image

Looking through the log, the part after the spike before the bit at 40% is tsc finishing up. The part at 40% is the client unit tests. The bit after that at around 80% is the other suites running in parallel.

I wonder if it's worth being careful with some of these, or if we can just trigger all of them in parallel and let the CPU sort it out

noahtallen:
image

Now that's what I'm talking about. 😎 2x faster overall

The part towards the end is just anything that takes a bit longer than the other tasks continuing on, while other "queued" tasks finish up. The part at the beginning is yarn install and other setup tasks which probably can't be parallelized

noahtallen:
However, there are two major problems:
1. Failures aren't currently reported
2. Logs are probably not easy to read since everything is logged simultaneously.

I think theoretically, both are solvable problems

worldomonation:
> However, there are two major problems:
> 
>     1. Failures aren't currently reported

Just a stab in the dark, could it be the `--silent` flag? 

>     2. Logs are probably not easy to read since everything is logged simultaneously.

This is a problem we've been having over on the E2E side. But Playwright covers this base somewhat by producing trace files, video replays and the (not-too-useful) logs.

One way to solve this, if it is even possible, is to redirect/tee the standard output of Jest into a log file per spec file.

I casually looked into this many months ago to solve the out-of-order log messages for E2Es but could not arrive a conclusion on whether Jest "knows" about that it's writing to the stdout. If it does, there _must_ be a way to intercept that and tee it into a log file.

noahtallen:
> One way to solve this, if it is even possible, is to redirect/tee the standard output of Jest into a log file per spec file.

I'll try to just redirect the output on a command-by-command basis, and then echo that output in sequence after everything finishes

noahtallen:
Ok, I think this is working now:
image


I'm seeing about 4 minutes on this branch, and 7 minutes on trunk currently. Pretty decent improvement! 

One of the things making this take especially long is `tsc`, which does not multithread very well. Some of the typescript checks also have to run in serial. For example, the packages have to be compiled before the other typechecks can run; otherwise, the correct types won't exist.

noahtallen:
The reason I switched to JS is because I was struggling to get more intricate pipeline working in bash. Which was this:

1. Typecheck. First build packages, then run other tsc in parallel.
2. run all commands in parallel, including the typecheck task

Much easier to do that in JS. 

Either way, I'm currently making sure that it's actually possible to see test failures and fix them. Which is not working very well yet. I'm trying to use service messages, but TeamCity really isn't designed to break down information in a better way here. For example, while it lets you assign messages to different parallel flows, it doesn't have any way to visualize that in the build log.

Previously, the build log was easy to work with because it was organized by build step. We no longer have that, because we can't parallelize build steps. I'm trying to use blocks, but some of the test output (via TeamCity test reporter) doesn't integrate with that super well.

noahtallen:
Alright, got it working! I have simulated some unit test failures, and they worked well, but I'll test different failure cases tomorrow. 

For now, this is ready to review!

.teamcity/_self/projects/WebApp.kt
```diff
@@ -329,7 +329,6 @@ object RunAllUnitTests : BuildType({
 	artifactRules = """
 		test_results => test_results
 		artifacts => artifacts
-		checkstyle_results => checkstyle_results
```
worldomonation, line 332: **question:non-blocking** - is this removed because any failures are now reported properly in the Tests tab for a run?

bin/unit-test-suite.mjs
```diff
@@ -0,0 +1,93 @@
...
+// manually so that there is some amount of margin.
+//
+// After some testing, I've found that 8+4 is a good setup. The largest task
+// runs by itself with 8 cores, and the other tasks run one by one with 4 other
+// cores. This leaves a final 4 cores free for tsc + any other tasks. This seems
+// to result in the fastest overall completion time.
+const testClient = withUnitTestInfo( 'test-client --maxWorkers=8' );
+const testPackages = withUnitTestInfo( 'test-packages --maxWorkers=4' );
+const testServer = withUnitTestInfo( 'test-server --maxWorkers=4' );
+const testBuildTools = withUnitTestInfo( 'test-build-tools --maxWorkers=4' );
```
worldomonation, line 51: **comment** - I see how you pinned tasks to a specific number of cores. Useful to know!

worldomonation (approved):

worldomonation:
Overall I think this reads well! I was able to follow what the individual parts did. Just a non-blocking question.

.teamcity/_self/projects/WebApp.kt
noahtallen: Actually just because eslint wasn't being run in this test suite! It has its own build right now, so this file hasn't been used in a while

bin/unit-test-suite.mjs
noahtallen: Yep! I think this is helpful when we want to run test suites with different complexity at the same time.

PR #72541, opened 2023-01-25
Yarn install as a snapshot dependency

noahtallen:
#### Proposed Changes
The idea here is to have `YarnInstall` be a snapshot dependency so that no other builds need to do it.

Honestly not sure if we'll need to do anything other than declare snapshot dependencies 🤔


worldomonation (approved):
As far as I can tell this only adds a new build, if that was the intention then great!

noahtallen:
yep!

PR #72547, opened 2023-01-25
Hide backup scheduled or delayed cards if the backups are stopped

rustamshaikh:
#### Proposed Changes

- This hides the backup scheduled or delayed cards based on the storage usage level.
- If the storage usage level is Full, these cards will be hidden. The usage level defaults to 'Normal' when not set.
- Uses `:empty` CSS selector to hide the card because of the component hierarchy

#### Testing Instructions

- For a site whose backups are stopped due to storage full
  - The Backup Scheduled card should be hidden
   
   Before the changes, the Backup Scheduled were visible even when backups were stopped
   Screenshot 2023-01-25 at 2 49 32 PM
   
   After the changes, it should be hidden for such cases
   Screenshot 2023-01-25 at 2 50 38 PM

   
  - When selecting a previous date other than today but after which backups were stopped, check that `The backup attempt for <selected_date> was delayed' card should be hidden.
  
  Before the changes, when selecting Yesterday as date
  Screenshot 2023-01-25 at 2 49 54 PM

  After the changes
  Screenshot 2023-01-25 at 2 50 49 PM

- For sites whose backups are not in STOPPED state, it should work as existing
  Screenshot 2023-01-25 at 2 58 23 PM

#### Pre-merge Checklist


client/my-sites/backup/main.jsx
```diff
@@ -202,11 +204,11 @@ function BackupStatus( {
 				) }
 
 				<BackupDatePicker onDateChange={ onDateChange } selectedDate={ selectedDate } />
-				<BackupStorageSpace />
+				<BackupStorageSpace onUsageLevel={ setUsageLevel } />
```
Initsogar: @rustamshaikh I think maybe we can use a hook to manage the usageLevel state instead of passing it to multiple components. It might be useful for the next project as well.

What do you think?

Initsogar (commented):
It looks great! Just left a suggestion. Let me know what do you think.

Initsogar (approved):
Approving the changes given that the suggestion is something we can do in a next iteration if preferred.

client/my-sites/backup/main.jsx
rustamshaikh: Thanks, @Initsogar for the suggestion. I will implement this using redux global state 👍

rustamshaikh:
@Initsogar Done the changes for using the global state to store usage level.

client/state/rewind/storage/reducer.ts
```diff
@@ -0,0 +1,18 @@
+import { AnyAction } from 'redux';
+import { REWIND_STORAGE_USAGE_LEVEL_SET } from 'calypso/state/action-types';
+import { StorageUsageLevels } from './types';
+
+const initialState = {
+	usageLevel: StorageUsageLevels.Normal,
+	usage: null,
```
Initsogar: I think we can omit `usage` for now, as we are not considering it for this PR.

client/state/rewind/storage/types.ts
```diff
@@ -0,0 +1,8 @@
+export type StorageUsageLevelName = 'Full' | 'Critical' | 'Warning' | 'Normal' | 'BackupsDiscarded';
```
Initsogar, line 1: Good idea to move it here :)

Initsogar (commented):
Overall, the code looks good! Thanks for addressing the concern about passing the usage level in multiple components.

Also, I think it would be great to add unit tests for the new reducer and selector for better maintainability. Let me know if you need any help with that 😊

Initsogar (approved):
The last changes and tests looks good. Great work @rustamshaikh!

Initsogar (approved):

PR #72575, opened 2023-01-25
Copy site: Add site copy hook to site tools

kozer:
#### Proposed Changes

Create a hook for initiating a site copy transfer, and move reusable code from Site Tools.

#### Testing Instructions

1. Navigate to `Settings -> General -> Site Tools`
2. Ensure you have a site with a `Business` plan
3. Make sure that you see the "Site tools" section

image

4. Press the copy site, and go through the steps. Ensure that your site is copied successfully.
5. Make sure that you can see the analytics tracking when you click the copy site button

image

#### Pre-merge Checklist

Fixes https://github.com/Automattic/dotcom-forge/issues/1556

### DIFF
--- a/client/landing/stepper/hooks/use-site-copy.tsx
+++ b/client/landing/stepper/hooks/use-site-copy.tsx
@@ -39,7 +39,7 @@ export const useSiteCopy = ( site: SiteExcerptData ) => {
 
 	const { setPlanCartItem } = useDispatch( ONBOARD_STORE );
 
-	const shouldShowSiteCopyItem = useCallback( () => {
+	const shouldShowSiteCopyItem = useMemo( () => {
 		return hasCopySiteFeature && isSiteOwner && plan && ! isLoadingPurchase;
 	}, [ hasCopySiteFeature, isSiteOwner, plan, isLoadingPurchase ] );
 
--- a/client/landing/stepper/hooks/use-site-copy.tsx
+++ b/client/landing/stepper/hooks/use-site-copy.tsx
@@ -1,4 +1,5 @@
 import { WPCOM_FEATURES_COPY_SITE } from '@automattic/calypso-products';
+import { createHigherOrderComponent } from '@wordpress/compose';
 import { useDispatch } from '@wordpress/data';
 import { useEffect, useMemo, useCallback } from 'react';
 import { useDispatch as useReduxDispatch, useSelector } from 'react-redux';
@@ -26,10 +27,7 @@ function useSafeSiteHasFeature( siteId: number, feature: string ) {
 	} );
 }
 
-export const useSiteCopy = (
-	site: SiteExcerptData,
-	recordTracks: ( eventName: string, extraProps?: Record< string, any > ) => void
-) => {
+export const useSiteCopy = ( site: SiteExcerptData ) => {
 	const userId = useSelector( ( state ) => getCurrentUserId( state ) );
 	const hasCopySiteFeature = useSafeSiteHasFeature( site.ID, WPCOM_FEATURES_COPY_SITE );
 	const plan = site?.plan;
@@ -47,19 +45,22 @@ export const useSiteCopy = (
 		return hasCopySiteFeature && isSiteOwner && plan && ! isLoadingPurchase;
 	}, [ hasCopySiteFeature, isSiteOwner, plan, isLoadingPurchase ] );
 
-	const startSiteCopy = useCallback( () => {
-		if ( ! plan || ! purchases ) {
-			return;
-		}
-		clearSignupDestinationCookie();
-		setPlanCartItem( { product_slug: plan.product_slug } );
-		const marketplacePluginProducts = purchases
-			.filter( ( purchase ) => purchase.productType === 'marketplace_plugin' )
-			.map( ( purchase ) => ( { product_slug: purchase.productSlug } ) );
+	const startSiteCopy = useCallback(
+		( recordTracks: () => void ) => {
+			if ( ! plan || ! purchases ) {
+				return;
+			}
+			clearSignupDestinationCookie();
+			setPlanCartItem( { product_slug: plan.product_slug } );
+			const marketplacePluginProducts = purchases
+				.filter( ( purchase ) => purchase.productType === 'marketplace_plugin' )
+				.map( ( purchase ) => ( { product_slug: purchase.productSlug } ) );
 
-		setProductCartItems( marketplacePluginProducts );
-		recordTracks( 'calypso_sites_dashboard_site_action_copy_site_click' );
-	}, [ plan, recordTracks, setPlanCartItem, setProductCartItems, purchases ] );
+			setProductCartItems( marketplacePluginProducts );
+			recordTracks();
+		},
+		[ plan, setPlanCartItem, setProductCartItems, purchases ]
+	);
 
 	return useMemo(
 		() => ( {
@@ -69,3 +70,18 @@ export const useSiteCopy = (
 		[ shouldShowSiteCopyItem, startSiteCopy ]
 	);
 };
+
+export const withSiteCopy = createHigherOrderComponent(
+	( Wrapped ) => ( props ) => {
+		// Limit query to 1 since we are only interested in the totalItems count.
+		const { shouldShowSiteCopyItem, startSiteCopy } = useSiteCopy( props.site );
+		return (
+			<Wrapped
+				{ ...props }
+				shouldShowSiteCopyItem={ shouldShowSiteCopyItem }
+				startSiteCopy={ startSiteCopy }
+			/>
+		);
+	},
+	'withSiteCopy'
+);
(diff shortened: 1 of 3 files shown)


client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -0,0 +1,79 @@
...
+			shouldShowSiteCopyItem,
+			startSiteCopy,
+		} ),
+		[ shouldShowSiteCopyItem, startSiteCopy ]
+	);
+};
+
+export const withSiteCopy = createHigherOrderComponent(
+	( Wrapped ) => ( props ) => {
+		// Limit query to 1 since we are only interested in the totalItems count.
```
wojtekn: Is this comment relevant?

wojtekn (commented):
@kozer nice refactor! 

The Copy Site process works well for me after the refactoring. I noticed that we don't use this new code on the Sites page. Should we refactor that part, too, to fully benefit from extracting this logic to a separate hook?

https://github.com/Automattic/wp-calypso/blob/76bedd060f565be562d8428ec758e5eb0928f2fe/client/sites-dashboard/components/sites-ellipsis-menu.tsx#L186

client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -0,0 +1,79 @@
...
+		},
+		[ plan, setPlanCartItem ]
+	);
+
+	return useMemo(
+		() => ( {
+			shouldShowSiteCopyItem,
+			startSiteCopy,
+		} ),
+		[ shouldShowSiteCopyItem, startSiteCopy ]
```
sejas: Would it make sense if `shouldShowSiteCopyItem` is a value instead of a function?

client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -0,0 +1,79 @@
...
+	}, [ hasCopySiteFeature, isSiteOwner, plan, isLoadingPurchase ] );
+
+	const startSiteCopy = useCallback(
+		( recordTracks: () => void ) => {
+			if ( ! plan ) {
+				return;
+			}
+			clearSignupDestinationCookie();
+			setPlanCartItem( { product_slug: plan.product_slug } );
+			recordTracks();
```
sejas: If the function will always `recordTracks` , then I prefer to receive the eventName as string. Something like:

```suggestion
		( eventName: CopySiteEventName ) => {
			if ( ! plan ) {
				return;
			}
			clearSignupDestinationCookie();
			setPlanCartItem( { product_slug: plan.product_slug } );
			recordTracksEvent(eventName);
```

where `eventName` is a typed variable:

```
type CopySiteEventName = 'calypso_settings_copy_site_option_click' | 'another_event'
```

client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -0,0 +1,79 @@
...
+function useSafeSiteHasFeature( siteId: number, feature: string ) {
+	const dispatch = useReduxDispatch();
+	useEffect( () => {
+		dispatch( fetchSiteFeatures( siteId ) );
+	}, [ dispatch, siteId ] );
+
+	return useSelector( ( state ) => {
+		return siteHasFeature( state, siteId, feature );
+	} );
+}
```
sejas, line 28: It would be great if we could import and reuse this hook. My concern is to avoid maintaining it in two places:

https://github.com/Automattic/wp-calypso/blob/trunk/client/sites-dashboard/components/sites-ellipsis-menu.tsx#L137-L146

The goal of this hook is similar to what `useQuerySitePurchases` does. "Fetching site data".

sejas (commented):
Wow!, @kozer , Great refactor! 👏 
I love the HOC `withSiteCopy`

I agree with @wojtekn , can we use the hooks in SMP?

client/landing/stepper/hooks/use-site-copy.tsx
kozer: Ooopsie, leftover.

kozer:
> @kozer nice refactor!
> 
> The Copy Site process works well for me after the refactoring. I noticed that we don't use this new code on the Sites page. Should we refactor that part, too, to fully benefit from extracting this logic to a separate hook?
> 
> https://github.com/Automattic/wp-calypso/blob/76bedd060f565be562d8428ec758e5eb0928f2fe/client/sites-dashboard/components/sites-ellipsis-menu.tsx#L186

I was thinking to wait, till the "marketplace plugin" pr is merged, and refactoring it then, alongside a minor refactor in the hook.

client/landing/stepper/hooks/use-site-copy.tsx
kozer: I'm waiting for the marketplace pr to merge in and clean it then. I have already done this locally :)

client/landing/stepper/hooks/use-site-copy.tsx
kozer: I'm not sure about the function that is called in both cases (in SiteTools and in EllipsisMenu).
Site tools seem that it is using `recordTracksEvent` and ellipsis menu uses a `recordTrack` which is a prop, and I'm not familiar if it does something else except that records the event.

sejas: Yes!,
`recordTrack` in `sites-ellipsis-menu.tsx` uses `recordTracksEvent` under the hood.

https://github.com/Automattic/wp-calypso/blob/76bedd060f565be562d8428ec758e5eb0928f2fe/client/sites-dashboard/components/sites-ellipsis-menu.tsx#L263-L268

sejas: The case I would like to avoid is that `recordTracks` callback function starts being used for more things than just "tracking events".

kozer: Ok then! I'll refactor it to just pass the event instead of passing a callback! Thanks for clarifying that!

client/landing/stepper/hooks/use-site-copy.tsx
kozer: This is what was discussed originally with Daniel (to be a function). However it makes more sense to me as well, and I was not feeling very well about that being a function. I will change it to a value. Thanks for your comment

client/landing/stepper/hooks/use-site-copy.tsx
kozer: > type CopySiteEventName = 'calypso_settings_copy_site_option_click' | 'another_event'

I left it as as string so someone would not have to update it every time this is used. It could probably become a monstrous type ( not many chances though).
Also, leaving it as a string, matches what we already have in ellipsis menu.

client/my-sites/site-settings/site-tools/index.jsx
```diff
@@ -104,13 +98,11 @@ class SiteTools extends Component {
...
 				) }
-				{ isEnabled( 'sites/copy-site' ) && canSiteBeCopied && (
+				{ isEnabled( 'sites/copy-site' ) && shouldShowSiteCopyItem && (
 					<SiteToolsLink
 						href={ copySiteUrl }
 						onClick={ () => {
-							clearSignupDestinationCookie();
-							setPlanCartItem( { product_slug: planSlug } );
-							recordTracksEvent( 'calypso_settings_copy_site_option_click' );
+							startSiteCopy( 'calypso_settings_copy_site_option_click' );
```
danielbachhuber, line 105: One nit is that this should maybe be an object of options, instead of passing the event name as the first argument.

We could potentially do this later, though.

client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -0,0 +1,80 @@
...
+	return useMemo(
+		() => ( {
+			shouldShowSiteCopyItem,
+			startSiteCopy,
+		} ),
+		[ shouldShowSiteCopyItem, startSiteCopy ]
+	);
+};
+
+export const withSiteCopy = createHigherOrderComponent(
```
danielbachhuber, line 68: Do we need a higher order component because we're using Calypso store data?

danielbachhuber (approved):
Landing this so we can keep things moving along

client/landing/stepper/hooks/use-site-copy.tsx
kozer: we need a HOC because we use a hook.

danielbachhuber: > we need a HOC because we use a hook.

@kozer This hook doesn't seem to need a HOC though?

https://github.com/Automattic/wp-calypso/blob/f31905ccef603a0f15c963177574022ffd7ed915/client/data/sites/use-site-excerpts-query.ts#L27-L40

What's the difference between the two?

kozer: It doesn't seem that we use it inside a class component. Do we? The moment we need to do that, we will need to wrap it inside a functional component (which is what the HOC I created does.). Hooks can't be used inside class components.

danielbachhuber: > Hooks can't be used inside class components.

@kozer Ah, this makes sense. Thanks for satisfying my curiosity!

PR #72585, opened 2023-01-25
Fix random console errors

arcangelini:
#### Proposed Changes

I found 3 console errors that all appear to be coming from Blogging Prompts. Here are the errors and proposed fixes. The final one has no fix, I didn't have a lot of time to dig into that one but I wanted to point it out.

```
Warning: Invalid DOM property `stroke-width`. Did you mean `strokeWidth`?
```

_Changed it to camel case_

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `PromptsNavigation`. See https://reactjs.org/link/warning-keys for more information.
```

_Added the index to use as a key_

```
Warning: Cannot update a component (`Connect(MomentProvider)`) while rendering a different component (`PromptsNavigation`). To locate the bad setState() call inside `PromptsNavigation`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```

_Wasn't sure what to do with this last one_

#### Testing Instructions

1. Pull branch and start local calypso
2. Log in and go to /home
3. Check the console log to see that the errors are gone.
4. I am not sure if these affected the actual component


arcangelini:
All this has been addressed in other PR's

PR #54643, opened 2021-07-16
i18n: Improve canonical and hreflang URLs on themes pages

jornp:
#### Changes proposed in this Pull Request

Theme details pages:
* The canonical URL and OG URL have been localized, which is the correct way of combining canonical URLs and hreflang links. See [here](https://yoast.com/hreflang-ultimate-guide/#hreflang-canonical), for example:
  > Every language should have a rel="canonical" link pointing to itself.

Theme details & themes showcase: 
* The trailing slash has been removed from the canonical, hreflang, and OG URLs, to make them match the page's URL exactly.
* Outputting the canonical URL has been moved from the `DocumentHead` component to a new middleware/controller function, `setLocalizedCanonicalUrl`, right below to `setHrefLangLinks`. This solves the problem of the canonical URL only showing on the first page-load (disappearing after refreshing the page once or twice). 

Implementation notes:
* I’m leaving the OG tags in `DocumentHead`, because those tags are built up very dynamically and would be hard to collect at the middleware stage.
* The middleware functions have been changed to use a hard-coded https://wordpress.com base. This simplifies the code a bit and it has the added benefit that it also works correctly locally.
* In `themes/index.node.js` I had to move `setHrefLangLinks` and `setLocalizedCanonicalUrl` down a bit, until after `ssrSetupLocale`. Otherwise, `getLocaleSlug` would return an incorrect value.

#### Testing instructions

While logged out, visit in several languages (English and non-English):

1. a couple of themes (e.g. http://calypso.localhost:3000/es/theme/blank-canvas) and their /setup and /support subpages,
2. the themes showcase (e.g. http://calypso.localhost:3000/es/themes/) and their filter-subpages like /business, /blog/filter/art, /food/filter/feature:portfolio,

and do the following:

* Verify that the canonical URL and OG URL:
  1. are localized,
  2. have a https://wordpress.com base, 
  3. do not have a trailing slash,
  4. match the URL in the address bar exactly.
* Verify that the canonical URL does not disappear when reloading the page.
* Verify that the hreflang links:
  1. are still there, 
  2. have a https://wordpress.com base, 
  3. are localized correctly,
  4. do not have a trailing slash,
  5. match the URL in the address bar exactly.

Related to 291-gh-Automattic/i18n-issues

PR #59903, opened 2022-01-10
Fix race conditions in localized SSR

jsnajdr:
Fixes a serious bug in localized server-side rendering discovered by @sgomes when testing the `/log-in` routes. It's caused by the fact that there is one global singleton instance of the `i18n` object on the server and all requests from all users share it.

1. Locally, run `yarn build-languages` so that your `public/languages/` folder is populated with language files like `id-v.1.1.json`.
2. Then run local Calypso with `yarn start`.
3. Disable JavaScript in your browser devtools so that you are sure that you always see only the server-rendered markup and that the client-side JavaScript doesn't run.
4. Open the localized `/id/themes` page. You should see Calypso localized in Indonesian.
5. Now open the `/log-in` page. As it doesn't have any locale suffix (like `/log-in/id`) it should be in English. But it's in Indonesian, the language of the previously rendered Themes page!

Screenshot 2022-01-10 at 12 24 15

The reason is that the `ssrSetupLocale` page.js handler, used on the `/:lang/themes` page, will set the global `i18n` instance to Indonesian locale. Then, when the `/log-in` page is rendered, it doesn't modify the `i18n` instance and uses it, assuming it's using the default `en` locale`. But it doesn't. It uses the somewhat random locale that the last `/themes` request happened to request.

The fix has several parts:
- first, I had to clean up `i18n-calypso` TS types so that they declare the `I18N` class. TypeScript files wouldn't compile otherwise.
- then I added support for React context to `i18n-calypso`. The `localize` HOC and the `useTranslate` hook no longer use the global `i18n` instance directly, but instead use the one provided by `I18NContext`. The default global `i18n` instance is the default value of that context, so that unless you mount a provider, the library behaves exactly the same way as it did before.
- another thing I had to modify is the `/theme-filters` REST request issued by the `/themes` SSR. There is a `wpcom` handler that adds the `locale` query parameter to it, reading the locale value from the global `i18n` instance. That's going to always be `en` now. But we need to pass the real locale, because the filters response contains localized labels. I achieved this by adding a custom `locale` parameter to the `requestThemeFilters` action that overrides the `locale` param.
- finally I modified the `ssrSetupLocale` handler and the rendering code to create a local `i18n` instance and pass it in context.

**How to test:**
Verify that the bug described above is indeed fixed: the `/log-in` page is always English, no matter what request you did before.

Also verify that the `/id/themes` page continues to be localized:

Screenshot 2022-01-10 at 11 28 36

[...]

PR #72626, opened 2023-01-26
Adding const langParamSlug to help set href for localized canonicals …

chickee0311:
Canonicals on /themes and /plugins pages are always returning an EN value regardless of their locale. Since in almost all cases canonicals should self-reference, we should fix these to make sure they are pulling in the locale. 

Canonicals for localized /themes pages were previously set in #54643. At some point they stopped working, but we aren't sure exactly why yet. 

This PR introduces one solution to fix that issue. 

#### Proposed Changes

* Adds `const langParamSlug = context.params.lang;`
*  Uses `langParamSlug` for `getLocalizedCanonicalUrl` in href value

#### Testing Instructions

* Checkout branch or test with live link
* Visit localized /plugins pages while logged out or incognito such as `/es/plugins` `/es/plugins/browse/seo` and `/es/plugins/wordpress-seo` 
* View the page source or inspect the page to search for `canonical`
* Verify that the canonical points to itself and not to the default EN page.
* Complete these same steps above while visiting several /themes pages such as `/es/themes` `/es/themes/filter/blog` and `/es/theme/mugistore`

#### Pre-merge Checklist

### DIFF
--- a/client/controller/localized-links.js
+++ b/client/controller/localized-links.js
@@ -1,6 +1,5 @@
 import config from '@automattic/calypso-config';
 import { localizeUrl, getLanguageSlugs } from '@automattic/i18n-utils';
-import { getLocaleSlug } from 'i18n-calypso';
 import performanceMark from 'calypso/server/lib/performance-mark';
 import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 import { setDocumentHeadLink } from 'calypso/state/document-head/actions';
@@ -28,7 +27,8 @@ export const setLocalizedCanonicalUrl = ( context, next ) => {
 		return;
 	}
 
-	const href = getLocalizedCanonicalUrl( context.originalUrl, getLocaleSlug() );
+	const langParamSlug = context.params.lang;
+	const href = getLocalizedCanonicalUrl( context.originalUrl, langParamSlug );
 	const link = {
 		rel: 'canonical',
 		href,
--- a/client/controller/localized-links.js
+++ b/client/controller/localized-links.js
@@ -27,7 +27,7 @@ export const setLocalizedCanonicalUrl = ( context, next ) => {
 		return;
 	}
 
-	const langParamSlug = context.params.lang;
+	const langParamSlug = context.i18n.getLocaleSlug();
 	const href = getLocalizedCanonicalUrl( context.originalUrl, langParamSlug );
 	const link = {
 		rel: 'canonical',


yuliyan (commented):
> Canonicals for localized /themes pages were previously set in https://github.com/Automattic/wp-calypso/pull/54643. At some point they stopped working, but we aren't sure exactly why yet.

Canonical URLs have probably stopped working after https://github.com/Automattic/wp-calypso/pull/59903, as it switched from using the global to a private i18n-calypso instance and thus the global `getLocaleSlug` was no longer returning the actual locale slug of the translations that the page was rendered with.

I've tested your changes and it seems like the problem has been resolved. However, I'd like to propose an alternative approach of using the `getLocaleSlug` of the private i18n-calypso instance (i.e. `context.i18n.getLocaleSlug()`). Let me know what do you think.

client/controller/localized-links.js
```diff
@@ -28,7 +27,8 @@ export const setLocalizedCanonicalUrl = ( context, next ) => {
 		return;
 	}
 
-	const href = getLocalizedCanonicalUrl( context.originalUrl, getLocaleSlug() );
+	const langParamSlug = context.params.lang;
+	const href = getLocalizedCanonicalUrl( context.originalUrl, langParamSlug );
```
yuliyan, line 31: Should we use `context.i18n.getLocaleSlug()` here instead of the language URL parameter?

In theory, the URL parameter and the locale slug from i18n-calypso should always be the same, but if, for some reason, the two values are different, I think it might be a little more accurate to use the locale slug from the translation data used to render the page.

chickee0311: Thanks for the suggestions here @yuliyan. I've swapped out the `context.params.lang` in favor of  `context.i18n.getLocaleSlug()`. Looks like this tests ok on my end with the same test plan previously. Would you mind taking another look?

client/controller/localized-links.js
```diff
@@ -28,7 +27,8 @@ export const setLocalizedCanonicalUrl = ( context, next ) => {
 		return;
 	}
 
-	const href = getLocalizedCanonicalUrl( context.originalUrl, getLocaleSlug() );
+	const langParamSlug = context.i18n.getLocaleSlug();
+	const href = getLocalizedCanonicalUrl( context.originalUrl, langParamSlug );
```
yuliyan, line 31: Thank you for applying the changes!

Optional: It could be simplified a bit more by removing the constant definition above and using  `context.i18n.getLocaleSlug()` right away in the function call:
```js
const href = getLocalizedCanonicalUrl( context.originalUrl, context.i18n.getLocaleSlug() );
```

yuliyan (approved):
Tested again and everything seems to be working as expected.

I left one comment for a small nit, but it can be considered optional and non-blocking for the PR to be merged.

LGTM 🚢

PR #72665, opened 2023-01-26
Marketplace Themes: Makoney couldn't be activated after reverting the site

### LINKED ISSUE #72444: Marketplace Themes: Can't activate Makoney after resetting the site
Users are not able to activate Makoney theme after resetting their site.

Slack thread: p1674361885647269-slack-C0160HSMDQV

valterlorran:
#### Proposed Changes

* Handles the edge case that prevented users from activating a Marketplace theme after the site was reverted. 
* Adds the confirmation modal to inform the users that the site will be transferred to atomic.

#### Testing Instructions

* Create a site and activate Makoney
* Revert it to simple: p1674614956322819/1674361885.647269-slack-C0160HSMDQV
* Go to the `/themes/<SITE-SLUG>`
* Find Makoney and click on the menu, then click on Activate
* The transference modal should pop up
* Confirming it, should transfer your site to atomic and activate Makoney

### Screenshots
Screen Shot 2023-04-12 at 14 28 27
Screen Shot 2023-04-12 at 14 18 35
Screen Shot 2023-04-12 at 14 13 45

Related to #72444


client/state/themes/actions/index.js
```diff
@@ -41,3 +41,6 @@ export { tryAndCustomizeTheme } from 'calypso/state/themes/actions/try-and-custo
 export { uploadTheme } from 'calypso/state/themes/actions/upload-theme';
 export { updateThemes } from 'calypso/state/themes/actions/theme-update';
 export { upsellCardDisplayed } from 'calypso/state/themes/actions/upsell-card-displayed';
+export { acceptEligibilityWarningDialog } from 'calypso/state/themes/actions/accept-eligibility-warning-dialog';
+export { dismissEligibilityWarningDialog } from 'calypso/state/themes/actions/dismiss-eligibility-warning-dialog';
+export { showEligibilityWarningDialog } from 'calypso/state/themes/actions/show-eligibility-warning-dialog';
```
daledupreez: Nit: Can we keep the new exports in alphabetical order like the existing ones?

client/my-sites/themes/eligibility-warning-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+import {
+	getCanonicalTheme,
+	getThemeForEligibilityWarning,
+	shouldShowEligibilityWarning,
+} from 'calypso/state/themes/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+
+import './auto-loading-homepage-modal.scss';
+
+interface Props {
```
daledupreez: Nit: can we call this something more specific like `EligibilityWarningModalProps`?

client/my-sites/themes/eligibility-warning-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+		author_uri: string;
+		id: string;
+		name: string;
+	};
+	dispatchAcceptEligibilityWarningDialog: typeof acceptEligibilityWarningDialog;
+	dispatchDismissEligibilityWarningDialog: typeof dismissEligibilityWarningDialog;
+	dispatchActivateTheme: typeof activateTheme;
+	dispatchRecordTracksEvent: typeof recordTracksEvent;
+}
+class EligibilityWarningModal extends Component< Props > {
```
daledupreez: Thought: it's not obvious to me what `Eligibility` this is about. Might it be clearer to call this something like `AtomicTransferDialog`? It might be more flexible if we also pass in a callback like `onAcceptAtomicTransfer` and any descriptive text as props. (We might also have such a component already.)

client/my-sites/themes/eligibility-warning-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+interface Props {
+	siteId?: number;
+	showEligibility: boolean;
+	theme: {
+		author: string;
+		author_uri: string;
+		id: string;
+		name: string;
+	};
```
daledupreez: I discovered yesterday that there is a `Theme` type in `calypso/types`, which should work here too.

client/my-sites/themes/eligibility-warning-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	handleDismiss() {
+		return this.props.dispatchDismissEligibilityWarningDialog();
+	}
+
+	render() {
+		const { showEligibility } = this.props;
+
+		const isMarketplaceProduct = true;
```
daledupreez: This feels like it's WIP. I think we can use the `isExternallyManagedTheme` selector for this.

client/state/themes/selectors/index.js
```diff
@@ -68,6 +69,9 @@ export { isWporgTheme } from 'calypso/state/themes/selectors/is-wporg-theme';
 export { prependThemeFilterKeys } from 'calypso/state/themes/selectors/prepend-theme-filter-keys';
 export { shouldShowTryAndCustomize } from 'calypso/state/themes/selectors/should-show-try-and-customize';
 export { shouldShowHomepageWarning } from 'calypso/state/themes/selectors/should-show-homepage-warning';
+export { shouldShowEligibilityWarning } from 'calypso/state/themes/selectors/should-show-eligibility-warning';
```
daledupreez: Super-tiny nit: Can we move this one line up? This line should sort before the one above it. 🤓

client/state/themes/selectors/is-atomic-site-required.ts
```diff
@@ -0,0 +1,8 @@
+import { isExternallyManagedTheme } from './is-externally-managed-theme';
+import 'calypso/state/themes/init';
+
+/**
+ */
+export function isAtomicSiteRequired( state = {}, themeId: string ): boolean {
```
daledupreez: Naming suggestion: how about `doesThemeRequireAtomicSite()`?

daledupreez (commented):
I have some comments from inspection. Will try to test tomorrow.

sixhours:
> Handles the edge case that prevented users from activating a Marketplace theme after the site was reverted.

Looking at the original issue, I don't think this solves the problem. I'm still seeing the theme activation error after resetting the site with this PR:

Screen Shot 2023-02-01 at 1 38 10 PM

The issue reported does not involved a fully reverted Atomic site. I can see where that's confusing, because the revert tool page has a secondary utility at the bottom of the page. I'd never really noticed it before. 😅  With this tool, the site contents/plugins are reset, but the site is still Atomic; it's not moved back to the WP.com simple infrastructure.

sixhours (commented):
This is working as described, but only for the main theme showcase page -- if you click "Activate this design" from the individual theme page at `/theme/makoney` nothing happens; you won't see the transfer modal pop up until you navigate back to the main themes page.

sixhours:
I also noticed it's possible to click outside the transfer modal window to close it before the transfer has completed. The transfer UI includes a warning about navigating away before the transfer completes, but it isn't shown because the modal isn't visible. This makes it possible for someone to navigate away from the page and disrupt the transfer; we should probably prevent that.

nelsonec87:
In my tests, activating from the showcased worked, but from the theme page, I got:
image

valterlorran:
@nelsonec87 Did you see the modal when you tested it on the theme page? I tested and couldn't reproduce it :( Could you check if your local branch is updated?

nelsonec87:
@valterlorran ,I've tested in calypso-live :) 
The modal just "blinked" on the screen, but I'll test again shortly ;)

nelsonec87:
I don't know if it's in the scope of this PR, but I got this in my tests:
1. When I fully revert the site, it works as expected
2. When I just "reset" the site
image
Then I can't get the theme installed. I get this:
image

valterlorran:
Oh, right, Got it! The issue **resetting** the site was addressed here: https://github.com/Automattic/wp-calypso/pull/73113
I rebased, and that should work now.

client/blocks/eligibility-warnings/index.tsx
```diff
@@ -79,6 +80,7 @@ export const EligibilityWarnings = ( {
 	launchSite: launch,
 	makeSitePublic,
 	translate,
+	isContinueButtonDisabled,
```
daledupreez: Minor naming thought: might it make sense to call this `disableContinueButton` or `disableProceedButton`? The idea is that the caller is providing an instruction to trigger that effect.

client/my-sites/themes/atomic-transfer-dialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+		dispatchInitiateThemeTransfer( siteId, null, '' );
+	}
+
+	getCurrentPath = () => {
+		const { origin, href } = window.location;
+
+		return href.replace( origin, '' );
+	};
+
+	componentDidUpdate( prevProps: Readonly< AtomicTransferDialogProps > ): void {
```
daledupreez, line 120: The use of this function makes me think this might have been better implemented as a functional component. But I think it's too late to switch that across. It would be nice to iterate on the implementation, but I am not sure when we'll be able to do that. Also, to be 100% explicit, I don't think we should do that in this PR.

client/state/themes/selectors/does-theme-require-atomic-site.ts
```diff
@@ -0,0 +1,13 @@
...
+import 'calypso/state/themes/init';
+
+/**
+ * Checks if a theme requires an Atomic site.
+ *
+ * @param state The app global state
+ * @param themeId The theme ID
+ * @returns True if the theme requires an Atomic site
+ */
+export function doesThemeRequireAtomicSite( state = {}, themeId: string ): boolean {
```
daledupreez: Multiple comments here:
1. Can we use the `AppState` type for `state`?
2. Can we add types to the JSDoc?

client/state/themes/selectors/get-theme-for-atomic-transfer-dialog.ts
```diff
@@ -0,0 +1,12 @@
+import 'calypso/state/themes/init';
+import { AppState } from 'calypso/types';
```
daledupreez: This should be `import type`.
```suggestion
import type { AppState } from 'calypso/types';
```

client/state/themes/selectors/has-atomic-transfer-dialog-accepted.js
```diff
@@ -0,0 +1,15 @@
+import 'calypso/state/themes/init';
+
+/**
+ * Checks if the atomic transfer dialog has been accepted.
+ *
+ * @param {Object} state   Global state tree
+ * @param {string} themeId Theme ID to check if the atomic transfer dialog has been accepted.
+ * @returns True if the atomic transfer dialog has been accepted. Otherwise, False.
+ */
+export function hasAtomicTransferDialogAccepted( state, themeId ) {
```
daledupreez: Main comment: might it be clearer if we call this `wasAtomicTransferDialogAccepted()`? It still has the boolean prefix, but more clearly represents the check for a thing that already happened.

Secondary cleanup comments: Can we also add the `AppState` type here? And maybe enforce/explicitly note that the return is a `boolean`?

client/state/themes/selectors/should-show-atomic-transfer-dialog.ts
```diff
@@ -0,0 +1,16 @@
+import 'calypso/state/themes/init';
+import { AppState } from 'calypso/types';
```
daledupreez: This should be `import type`.
```suggestion
import type { AppState } from 'calypso/types';
```

client/state/themes/selectors/should-show-atomic-transfer-dialog.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @param state The app global state
+ * @param themeId The theme ID
+ * @returns
+ */
+export function shouldShowAtomicTransferDialog( state: AppState, themeId: string ) {
+	return (
+		state.themes.themeHasAtomicTransferDialog?.themeId === themeId &&
+		state.themes.themeHasAtomicTransferDialog?.show
+	);
```
daledupreez: Can we explicitly add and document the `boolean` return type?

client/blocks/eligibility-warnings/index.tsx
```diff
@@ -235,7 +237,8 @@ export const EligibilityWarnings = ( {
 						disabled={
 							isProceedButtonDisabled( isEligible, listHolds ) ||
 							siteIsSavingSettings ||
-							siteIsLaunching
+							siteIsLaunching ||
+							isContinueButtonDisabled
 						}
 						busy={ siteIsLaunching || siteIsSavingSettings }
```
daledupreez: Thought: might it also be worth showing a `busy` state for the Continue button while we're transferring? (Not a blocker.)

client/my-sites/themes/atomic-transfer-dialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+	handleAccept() {
+		const { siteId, dispatchInitiateThemeTransfer } = this.props;
+		if ( ! siteId ) {
+			return;
+		}
+
+		dispatchInitiateThemeTransfer( siteId, null, '' );
+	}
+
+	getCurrentPath = () => {
```
daledupreez: Minor thought: we're only using this in one place, so we could pull the `.wordpress.com` => `.wpcomstaging.com` replacement into the function. If we do that, we could rename the function to `getAtomicSitePath()` or something similar.

client/my-sites/themes/atomic-transfer-dialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+		const { origin, href } = window.location;
+
+		return href.replace( origin, '' );
+	};
+
+	componentDidUpdate( prevProps: Readonly< AtomicTransferDialogProps > ): void {
+		const { siteId, siteSlug, isTransferred } = this.props;
+		if ( siteId && siteSlug && prevProps.isTransferred !== isTransferred && isTransferred ) {
+			window.location.replace(
+				this.getCurrentPath().replace( /\b.wordpress.com/, '.wpcomstaging.com' )
```
daledupreez: While it doesn't block shipping this patch, I think we should also try to trigger a state update. In my case, the navigation went ahead, but the new page didn't detect that I had the marketplace theme active, so the main CTA was still `Activate this design` rather than `Customize`.
Might we be seeing an issue where the site transfer has completed, but the theme hasn't been activated yet? We might need to wait for both to be true, not just that the transfer completed.

client/my-sites/themes/atomic-transfer-dialog.tsx
valterlorran: I created this issue so we can tackle it later.
https://github.com/Automattic/wp-calypso/issues/73731

markbiek:
I'm still getting the "Theme not yet available for this site" after reverting the site to Simple.

image

markbiek:
To recap what we ran into testing together.

I'm now seeing the modal but, after the modal closes, it's not showing the theme as being active until the page is refreshed.

This happens on both `/themes` and `/theme/makoney/:siteSlug`.

andres-blanco:
Did this steps:
1. Created a free site
2. Find Makoney and click "Upgrade to activate"
3. Went to revert tool and chose the `I want to reset the site makoneyrevert.wpcomstaging.com.` option.
4. Went to `/themes`, searched for Makoney and clicked on `Activate`.
5. Got this popup
image

In the test instructions it said that a confirmation popup should appear. I didn't need to confirm anything. Is this OK @valterlorran ?

andres-blanco:
I've retested using the **reset** tool. I get the popup with strange behaviour. In the video below the not-so-good behaviour can be seen at 0:44, where the top spinner disappears but the modal is still there and the `Continue` button is enabled. When I clicked it again nothing happened.

https://user-images.githubusercontent.com/375980/227312166-431bea6e-a68d-49dc-8c5f-e611be7e6acc.mp4

valterlorran:
Thanks for catching it @andres-blanco. The problem was sometimes the reset would take a while to finish, preventing us from starting the transfer right away. Now we will attempt several times before showing the error message.

client/state/themes/selectors/does-theme-require-atomic-site.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { AppState } from 'calypso/types';
+
+/**
+ * Checks if a theme requires an Atomic site.
+ *
+ * @param {AppState} state   The app global state
+ * @param {string}   themeId The theme ID
+ * @returns {boolean} True if the theme requires an Atomic site
+ */
+export function doesThemeRequireAtomicSite( state: AppState, themeId: string ): boolean {
```
markbiek: I think this function isn't needed. I think it's more clear to remove it and just call `isExternallyManagedTheme` in the one place we're calling `doesThemeRequireAtomicSite`.

My concern here is that there are other themes that also require Atomic (like the WooCommerce themes) but we're not checking for that case in this function.

Since the WooCommerce themes have their own process for transferring the site to Atomic, we don't need the check for those themes and can probably get rid of this new function.

markbiek (commented):
Everything is working as expected but I had one concern that I noted above.

client/state/themes/selectors/does-theme-require-atomic-site.ts
valterlorran: Hey @markbiek, the idea of this function was to support multiple types of themes in the future, instead of checking specifically whether it is managed externally. With that in mind, let me know what you think about it.

markbiek: I think the time to add this function would be when we have more than one type of theme we're checking for or if we need to perform that check in more than one place.

Right now, I think it just adds the potential for confusion and it'd be better to remove it.

markbiek (approved):
LGTM!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7921271

Thank you @valterlorran for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7921271).

PR #72680, opened 2023-01-26
Odyssey Stats: Apply Emerald typography styling

### LINKED ISSUE #72107: Odyssey Stats: Change display type throughout
"Display type" is a term used to describe the large heading type throughout a product or design. For stats in WordPress.com we're using Recoleta in many places, as seen here in "2023 in review", "All-time highlights" etc.

Image

However, for the stats we're displaying in Jetpack for wp-admin, we should use a different typeface. Recoleta is only used in the confines of WordPress.com. For any Jetpack plugin in wp-admin we use a different visual language called "Emerald"

Image

Emerald display type is all sans-serif text, while Recoletta is a serif. We want the Headline Small styling [can be found in Storybook here](https://automattic.github.io/jetpack-storybook/?path=/story/js-packages-components-text-heading--default). Let's see how this feels!

- [ ] Replace Recoleta throughout Odyssey stats with Emerald headlines small

cc @ederrengifo

jsnmoon:
Fixes #72107.

#### Proposed Changes

|||
|-|-|
|Before|image|
|After|image|

* Moves a styling mixin from `client/my-sites/stats/` to `packages/components/src/styles/`.
* Introduce a new styling mixin for overriding display type styling in Odyssey Stats.

#### Testing Instructions

* Check the live branch for this PR and ensure that all the type font remains unchanged from production.
* Apply this change to your local Jetpack site's Odyssey Stats (option 1 at PejTkB-3E-p2).
* Ensure that the headings now use a sans-serif font instead of Recoleta. Make sure to check both the traffic and insights pages!


apps/odyssey-stats/src/styles/typography.scss
```diff
@@ -0,0 +1,21 @@
+@import "@automattic/components/src/styles/typography.scss";
+
+$display-type-font: -apple-system,BlinkMacSystemFont,"Segoe UI","Roboto","Oxygen-Sans","Ubuntu","Cantarell","Helvetica Neue",sans-serif;
+$display-type-weight: 700;
+$display-type-line-height: 40px;
+
+// Override display type styling for Jetpack.
+@mixin stats-section-header-for-jetpack {
```
kangzj: I'm not an expert of SCSS, just wondering - are we able to use some kind of env variable (e.g. $is-jetpack ?) so that we only compile the Emrald typography conditionally, so that we don't have to override anymore.

jsnmoon: If our styling codebase were more centralized, we'd be able to leverage [the `!default` flag for variables](https://sass-lang.com/documentation/variables#default-values) and sort out this problem by overriding that default variable. This is how a lot of SCSS frameworks used to provide theming support back in the day.

There's no canonical way to perform environmental variable detection in a built-in way with SCSS. We can absolutely [hack something up like this](https://stackoverflow.com/questions/60058352/pass-webpack-environment-variable-to-scss-file), but that'll make things even harder to maintain than what I'm proposing, in my opinion.

kangzj (approved):
It works as described. Thanks for working on this 👍 

One thing worth noting is that the Jetpack app promo card is still using Recoleta. Should we address these too?

image

jsnmoon:
> It works as described. Thanks for working on this 👍
> 
> One thing worth noting is that the Jetpack app promo card is still using Recoleta. Should we address these too?
>
> image

Good catch. This should be fixed!

image

kangzj (approved):
LGTM

apps/odyssey-stats/src/styles/variables.scss
```diff
@@ -0,0 +1,13 @@
+@import "@automattic/components/src/styles/typography.scss";
+
+$display-type-font: -apple-system,BlinkMacSystemFont,"Segoe UI","Roboto","Oxygen-Sans","Ubuntu","Cantarell","Helvetica Neue",sans-serif;
```
dognose24: Could we use [$sans](https://github.com/Automattic/wp-calypso/blob/trunk/packages/typography/styles/variables.scss#L10) or [$a8c-font-family-sans](https://github.com/Automattic/wp-calypso/blob/trunk/packages/components/src/styles/_typography.scss#L1) for this variable?

dognose24 (approved):
LGTM! I left a nitty comment, which won't be a blocker.

apps/odyssey-stats/src/styles/variables.scss
jsnmoon: Good catch, fixed!

PR #53881, opened 2021-06-22
Opt-in persistent cache for e2e images

scinos:
### Background

See #53867

### Changes proposed in this Pull Request

This PR includes a few changes needed to opt-in persistent cache for e2e images:

**Enable webpack persistent cache in e2e builds**

Adds `PERSISTENT_CACHE=true` env var to e2e builds. This does not affect production, only the images used to power `caypso.live` (used to run e2e tests).

**Make the base docker image configurable** 

This will allow us to experiment with images with different cache content to see the effect in the builds. It defaults to `registry.a8c.com/calypso/base:latest` (same value as before). 

**Added env var `PROFILE` to toggle webpack profiling**

Before we had `WEBPACK_OPTIONS='--progress=profile'` for the webapp and `PROFILE=true` for the Node.js app. This change uses `PROFILE=true` for both.

**Do not prune dependency tree when installing dependencies in production**

In an attempt to reduce the size of the docker image (and speed up its creation) we deleted some `package.json` that we know we don't need to install to bundle the app (eg: `apps/editing-toolkit`). Let's call that the "simplified tree".

This was preventing us to get value from the persistent cache: the cache is created with a "full tree", and when we tried to use it in a build that has a "simplified tree", webpack saw that the content of `node_modules` didn't match: the "full tree" will have more packages that got hoisted to the root `node_modules`. In that scenario webpack things the cache is not good and it won't use it.

This change will actually increase build time when we can't use the existing cache.

**Enable babel-cache even if the persistent cache is enabled**

This is an attempt to reduce the build time when the persistent cache can't be used. In theory it will make image generation slower (it has to generate and write babel cache and then webpack cache), but I wasn't able to find any meaningful difference. I'll keep an eye once merged.

**Added webpack cache logs**

It helps to diagnose when a cache was used and how long it took to serialize and deserialize.

#### Testing instructions

N/A

PR #72687, opened 2023-01-26
Teamcity: improve webpack cache hit rate

noahtallen:
The goal of this PR is to improve performance of the "build docker image" TeamCity build by increasing the number of webpack cache hits.

### The issue with the webpack cache:
The main problem with the webpack cache is that a changing "build dependency" will invalidate the _entire_ cache. This adds at least two minutes to a build. This can also happen frequently (several times a day) for two reasons:
1. The webpack cache is not usually up to date. Before a recent experiment, it was up to a day old, since it was regenerated once per day.
2. A build dependency is anything that the webpack config `requires`, recursively. This covers a lot of files in Calypso and even the client's package.json file, along with several others.

Combining this, if you update client/package.json in a PR _after_ the base image build refreshed the webpack cache, every single build between then and the _next_ base image build (at least several hours) will be two minutes longer. With hundreds of these builds per day, that adds up!

### How to fix it?
Ideally, we'd constantly keep the webpack cache updated, so that every commit can use a very up-to-date cache. One solution is to run the base image build more frequently. However, this comes with a performance overhead: 10 minutes extra per base image build. Running that a lot can easily consume our limited agent space. We could also dynamically trigger the base image build when we notice the webpack cache invalidating. While better than before, we still end up with a lot of extra time in the queue -- maybe too much we happen to invalidate the cache several times in a day!

So I think we should attempt to persist and reuse the cache in a granular way:

 - Detect when the cache invalidates in Webpack
 - If so, copy the .cache directory into the base image
 - Push an updated `latest` tag for the base image.

Some benefits:
- An invalidated cache doesn't "spill over." E.g. this invalidation doesn't become a part of every build afterwards for several hours.
- Avoids running webpack twice -- we already calculated the cache per commit in the first place, we just don't persist it per commit.

### Implementation:
- We save the webpack output to a file
- We add a script which checks that file for the string which indicates the cache was invalidated
- That script then sets a TeamCity build parameter
- If on trunk and if that parameter exists, then trigger a cache update
- Cache update is a separate image in the Dockerfile, which we target directly. This should re-use the docker build cache (e.g. the cached layers) from previously in the build, making it pretty fast.

Looking at the parameters tab, we can see the invalidation flag getting set correctly:
image

### Todo:
- [x] Docker layer cache is not being fully utilized (--cache-from seemed to help this)
- [ ] Measure performance of updating just webpack cache, or the whole .cache directory. (With whole .cache directory, doing a cache update takes 2-3minutes. Probably 1 minute faster without).

Example build: https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/9414013

### DIFF
--- a/.teamcity/_self/projects/WebApp.kt
+++ b/.teamcity/_self/projects/WebApp.kt
@@ -47,6 +47,7 @@ object BuildDockerImage : BuildType({
 
 	params {
 		text("base_image", "registry.a8c.com/calypso/base:latest", label = "Base docker image", description = "Base docker image", allowEmpty = false)
+		text("base_image_publish_tag", "latest", label = "Tag to use for the published base image", description = "Base docker image tag", allowEmpty = false)
 		checkbox(
 			name = "MANUAL_SENTRY_RELEASE",
 			value = "false",
@@ -55,6 +56,15 @@ object BuildDockerImage : BuildType({
 			checked = "true",
 			unchecked = "false"
 		)
+		checkbox(
+			name = "UPDATE_BASE_IMAGE_CACHE",
+			value = "false",
+			label = "Update the base image from the cache.",
+			description = "Updates the base image by copying .cache files from the current build. Runs on trunk by default if the cache invalidates during the build.",
+			checked = "true",
+			unchecked = "false"
+		)
+		param("env.WEBPACK_CACHE_INVALIDATED", "false")
 	}
 
 	vcs {
@@ -63,15 +73,13 @@ object BuildDockerImage : BuildType({
 	}
 
 	steps {
-
 		script {
 			name = "Webhook Start"
+			conditions {
+				equals("teamcity.build.branch.is_default", "true")
+			}
 			scriptContent = """
 				#!/usr/bin/env bash
-
-				if [[ "%teamcity.build.branch.is_default%" != "true" ]]; then
-					exit 0
-				fi
 
 				payload=${'$'}(jq -n \
 					--arg action "start" \
@@ -87,11 +95,11 @@ object BuildDockerImage : BuildType({
 
 		script {
 			name = "Post PR comment"
+			conditions {
+				doesNotEqual("teamcity.build.branch.is_default", "true")
+			}
 			scriptContent = """
 				#!/usr/bin/env bash
-				if [[ "%teamcity.build.branch.is_default%" == "true" ]]; then
-					exit 0
-				fi
 
 				export GH_TOKEN="%matticbot_oauth_token%"
 				chmod +x ./bin/add-pr-comment.sh
@@ -114,6 +122,19 @@ object BuildDockerImage : BuildType({
 			dockerPull = true
 			dockerImagePlatform = ScriptBuildStep.ImagePlatform.Linux
 		}
+		
+		val commonArgs = """
+			--label com.a8c.image-builder=teamcity
+			--label com.a8c.build-id=%teamcity.build.id%
+			--build-arg workers=32
+			--build-arg node_memory=32768
+			--build-arg use_cache=true
+			--build-arg base_image=%base_image%
+			--build-arg commit_sha=${Settings.WpCalypso.paramRefs.buildVcsNumber}
+			--build-arg manual_sentry_release=%MANUAL_SENTRY_RELEASE%
+			--build-arg is_default_branch=%teamcity.build.branch.is_default%
+			--build-arg sentry_auth_token=%SENTRY_AUTH_TOKEN%
+		""".trimIndent().replace("\n"," ")
 
 		dockerCommand {
 			name = "Build docker image"
@@ -126,20 +147,7 @@ object BuildDockerImage : BuildType({
 					registry.a8c.com/calypso/app:commit-${Settings.WpCalypso.paramRefs.buildVcsNumber}
 					registry.a8c.com/calypso/app:latest
 				""".trimIndent()
-				commandArgs = """
-					--pull
-					--label com.a8c.image-builder=teamcity
-					--label com.a8c.target=calypso-live
-					--label com.a8c.build-id=%teamcity.build.id%
-					--build-arg workers=32
-					--build-arg node_memory=32768
-					--build-arg use_cache=true
-					--build-arg base_image=%base_image%
-					--build-arg commit_sha=${Settings.WpCalypso.paramRefs.buildVcsNumber}
-					--build-arg manual_sentry_release=%MANUAL_SENTRY_RELEASE%
-					--build-arg is_default_branch=%teamcity.build.branch.is_default%
-					--build-arg sentry_auth_token=%SENTRY_AUTH_TOKEN%
-				""".trimIndent().replace("\n"," ")
+				commandArgs = "--pull --label com.a8c.target=calypso-live $commonArgs"
 			}
 			param("dockerImage.platform", "linux")
 		}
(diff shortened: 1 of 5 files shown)


worldomonation:
Context: p1674767459661009-slack-C1A1EKDGQ

Dockerfile
```diff
@@ -58,11 +58,6 @@ RUN bash /tmp/env-config.sh
 #
 # This layer is populated with up-to-date files from
 # Calypso development.
-#
-# We remove apps, tests and desktop because they are not needed to
-# build or run calypso, but yarn will still install their
-# dependencies which end up bloating the image.
-# /apps/notifications is not removed because it is required by Calypso
```
noahtallen, line 65: This bit isn't true any more -- the logic it mentions was removed in #53881

Dockerfile
```diff
@@ -76,6 +71,14 @@ RUN yarn run build
 # Delete any sourcemaps which may have been generated to avoid creating a large artifact.
 RUN find /calypso/build /calypso/public -name "*.*.map" -exec rm {} \;
 
+###################
+# A cache-only update can be generated with "docker build --target update-base-cache"
+FROM ${base_image} as update-base-cache
```
noahtallen, line 77: In practice, this image isn't triggered when building the file, just because the final target (`app`) doesn't reference this image. However, we can take advantage of `--target` to tell `docker build` to just run this section. Given that each layer has been cached by docker, a separate `docker build` command which targets the update cache image should take advantage of the previously updated .cache files. (This happened in `yarn build` in the image above.)

worldomonation (approved):
We chatted over Slack about this and I don't see any particular faults with the approach!

Dockerfile
```diff
@@ -58,11 +58,6 @@ RUN bash /tmp/env-config.sh
 #
 # This layer is populated with up-to-date files from
 # Calypso development.
-#
-# We remove apps, tests and desktop because they are not needed to
-# build or run calypso, but yarn will still install their
-# dependencies which end up bloating the image.
-# /apps/notifications is not removed because it is required by Calypso
 COPY . /calypso/
 RUN yarn install --immutable --check-cache
```
WunderBart, line 62: Would it also make sense to separate the yarn install layer? For example:

```dockerfile
COPY package.json yarn.lock ./calypso/
RUN yarn install --immutable --check-cache

COPY . /calypso/
```

Deps change rarely, so we don't need that layer depending on the rest of the files IIUC.

noahtallen: Interesting idea! Do you think this will speed up builds even with `--pull`? I have to say I don't understand layers that well yet.

WunderBart: > Do you think this will speed up builds even with --pull?

Not sure what a "`--pull`" build is, but in general, the `yarn` layer should depend only on the content of `package.json` and `yarn.lock` files – it stays cached as long as those stay unchanged. Maybe we could cache out some other assets like this?

> I have to say I don't understand layers that well yet.

I've only scratched the surface of the layer caching, but this idea actually came from the official docs I read the other day: https://docs.docker.com/build/cache/#order-your-layers!

noahtallen: Thanks for the link! I think `--pull` makes Docker "pull" images from the repository even if a local image matches it already. As a result, I think the docker image is built from scratch for each commit

.teamcity/_self/projects/WebApp.kt
```diff
@@ -114,34 +122,42 @@ object BuildDockerImage : BuildType({
...
+			scriptContent = """
+				#!/usr/bin/env bash
+				set -e
+				
+				docker build -f Dockerfile $commonArgs --pull \
+					--label com.a8c.target=calypso-live \
+					-t registry.a8c.com/calypso/app:build-%teamcity.build.number% \
+					-t registry.a8c.com/calypso/app:commit-${Settings.WpCalypso.paramRefs.buildVcsNumber} \
+					-t registry.a8c.com/calypso/app:latest \
+					. 2>&1 | tee docker_output.txt
```
noahtallen: We have to change this to the direct `docker build` command so that we can observe the output. On top of that, Docker has really weird output, which is why we save it to a file to look at after the command finishes.

.teamcity/_self/projects/WebApp.kt
```diff
@@ -114,34 +122,42 @@ object BuildDockerImage : BuildType({
...
+
+				cache_invalidated=false
+				if grep "resolving of build dependencies is invalid\|will re-resolve build dependencies\|invalidated because hashes differ" docker_output.txt ; then
```
noahtallen: I added a few strings which can indicate when this happens -- I guess this is kind of hacky, because we're relying on weback to continue to use these exact strings. I think this is the easiest approach for now though.

noahtallen: Plus, something else could technically output these strings... Anyways, I think this is something we can tweak moving forward

.teamcity/_self/projects/WebApp.kt
```diff
@@ -55,6 +56,15 @@ object BuildDockerImage : BuildType({
...
 		)
+		checkbox(
+			name = "UPDATE_BASE_IMAGE_CACHE",
+			value = "false",
+			label = "Update the base image from the cache.",
+			description = "Updates the base image by copying .cache files from the current build. Runs on trunk by default if the cache invalidates during the build.",
+			checked = "true",
+			unchecked = "false"
+		)
+		param("env.WEBPACK_CACHE_INVALIDATED", "false")
```
noahtallen, line 67: This flag is ultimately set by a service message. I added a checkbox above so that we can also trigger it on branches.

.teamcity/_self/projects/WebApp.kt
```diff
@@ -222,6 +229,56 @@ object BuildDockerImage : BuildType({
...
+				equals("UPDATE_BASE_IMAGE_CACHE", "true")
+			}
+			commandType = build {
+				source = file {
+					path = "Dockerfile"
+				}
+				namesAndTags = "registry.a8c.com/calypso/base:%base_image_publish_tag%"
+				commandArgs = """
+					--target update-base-cache
+					--cache-from=registry.a8c.com/calypso/app:commit-${Settings.WpCalypso.paramRefs.buildVcsNumber},%base_image%
```
noahtallen, line 266: cache-from seems to help the layer cache work better -- maybe worth trying for other things too!

Dockerfile
```diff
@@ -71,11 +66,19 @@ RUN yarn install --immutable --check-cache
 # This contains built environments of Calypso. It will
 # change any time any of the Calypso source-code changes.
 ENV NODE_ENV production
-RUN yarn run build
+RUN yarn run build 2>&1 | tee build_log.txt && \
```
noahtallen, line 69: I found it easier to do this in the Dockerfile. Doing it in teamcity meant moving away from the built-in docker command type, which has some default settings that seem to make it work better.

client/webpack.config.js
```diff
@@ -114,7 +114,7 @@ function filterEntrypoints( entrypoints ) {
  * Note this is not the same as looking for `__dirname+'/node_modules/'+pkgName`, as the package may be in a parent
  * `node_modules`
  *
- * @param {string} pkgName Name of the package to search for
+ * @param {string} pkgName Name of the package to search for.
```
noahtallen, line 117: This was just triggering the webpack cache invalidation

worldomonation (approved):
I had another careful look and the logic implemented in this PR looks sound to me!

PR #72692, opened 2023-01-27
Publish cache in app docker images

noahtallen:
Publishes the .cache directory in the docker image so that it's available for future builds. Required for #72687 to work.

The main downside here is that the image will be larger, but I'm not sure by how much. Will test.

Pending question: What's the latest build? The most recent one to build, or just trunk? Unsure. I think latest might get tagged locally, but not published.

To test... not quite sure! I don't have the registry set up locally, so that's probably the first step.


worldomonation (approved):

noahtallen:
The size increase (~5GB) ended up being too large, so we won't go with this approach. See #72687

PR #72698, opened 2023-01-27
Disable Ecommerce Home Redirection when SSO is disabled

xristos3490:
#### Proposed Changes

This PR aims to disable the Ecommerce home redirection when a site has not enabled Jetpack's SSO. That's because the Home redirection only makes sense when the menu is the new Ecommerce Atomic menu; thus, the Calypso Home is replaced with the WooCommerce one.

#### Specification

Please note that I'm done my best to tackle this issue, but it's not yet optimal. 

I get a flash of content when navigating through sites, and I cannot bypass this. Check [this screencast](https://d.pr/v/tX1SLH) that navigates to (a) a Simple site, (b) an Ecommerce site with SSO disabled, and (c) to an Ecommerce site with SSO enabled (which results in redirection). 

Note that all Ecommerce sites have a flash of the Home content before showing the loader. How can we bypass this?

#### Testing Instructions

Ensure that you have at least one Ecommerce site with SSO disabled. 

**To disable SSO:**
- Navigate to **Settings > Security** and disable the "_Allow users to log in to this site using WordPress.com accounts_" option.

**Testing instructions:**
- Navigate to the site's home (`wordpress.com/home/[site_suffix]`) that has SSO disabled.
- Notice that you get redirected to WooCommerce Home (or the wp-admin login page, depending on the login cookie).
- Apply this PR.
- Navigate to home again and ensure that the redirection is removed; therefore, you should see the Calypso home.

#### Pre-merge Checklist

Related to https://github.com/Automattic/wc-calypso-bridge/pull/925

### DIFF
--- a/client/my-sites/customer-home/controller.jsx
+++ b/client/my-sites/customer-home/controller.jsx
@@ -1,7 +1,9 @@
 import { isEcommerce } from '@automattic/calypso-products/src';
 import page from 'page';
+import { fetchModuleList } from 'calypso/state/jetpack/modules/actions';
 import { fetchSitePlugins } from 'calypso/state/plugins/installed/actions';
 import { getPluginOnSite } from 'calypso/state/plugins/installed/selectors';
+import isJetpackModuleActive from 'calypso/state/selectors/is-jetpack-module-active';
 import { requestSite } from 'calypso/state/sites/actions';
 import {
 	canCurrentUserUseCustomerHome,
@@ -70,10 +72,14 @@ export async function maybeRedirect( context, next ) {
 			// We need to make sure that sites on the eCommerce plan actually have WooCommerce installed before we redirect to the WooCommerce Home
 			// So we need to trigger a fetch of site plugins
 			await context.store.dispatch( fetchSitePlugins( siteId ) );
+			await context.store.dispatch( fetchModuleList( siteId ) );
 
 			const refetchedState = context.store.getState();
+
 			const installedWooCommercePlugin = getPluginOnSite( refetchedState, siteId, 'woocommerce' );
-			if ( installedWooCommercePlugin && installedWooCommercePlugin.active ) {
+			const isSSOEnabled = !! isJetpackModuleActive( refetchedState, siteId, 'sso' );
+
+			if ( isSSOEnabled && installedWooCommercePlugin && installedWooCommercePlugin.active ) {
 				window.location.replace( siteUrl + '/wp-admin/admin.php?page=wc-admin' );
 			}
 		}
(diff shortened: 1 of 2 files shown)


client/my-sites/customer-home/controller.jsx
```diff
@@ -70,10 +72,14 @@ export async function maybeRedirect( context, next ) {
 			// We need to make sure that sites on the eCommerce plan actually have WooCommerce installed before we redirect to the WooCommerce Home
 			// So we need to trigger a fetch of site plugins
 			await context.store.dispatch( fetchSitePlugins( siteId ) );
+			await context.store.dispatch( fetchModuleList( siteId ) );
```
daledupreez, line 75: Suggestion: as written, we're going to sequentially call and wait for each call. Might it be better to wrap this code in `Promise.allSettled()` so we can perform both fetches in parallel and wait for both to complete.

This pair of `await` statements may also be impacting the initial rendering of the page, as many parts of the client may continue to run/render rather than waiting for this async code to be executed.

```suggestion
			Promise.allSettled( [
				context.store.dispatch( fetchSitePlugins( siteId ) ),
				context.store.dispatch( fetchModuleList( siteId ) ),
			] ).then(
				///...
			);
```

client/my-sites/customer-home/main.jsx
```diff
@@ -152,6 +160,8 @@ Home.propTypes = {
 	hasWooCommerceInstalled: PropTypes.bool.isRequired,
 	isStaticHomePage: PropTypes.bool.isRequired,
 	isRequestingSitePlugins: PropTypes.bool.isRequired,
+	ssoModuleActive: PropTypes.bool.isRequired,
+	fetchingJetpackModules: PropTypes.bool.isRequired,
 	site: PropTypes.object.isRequired,
 	siteId: PropTypes.number.isRequired,
 	trackViewSiteAction: PropTypes.func.isRequired,
```
daledupreez, line 167: Minor (non-blocking) nit: can we add these new types in alphabetical order?
```suggestion
	fetchingJetpackModules: PropTypes.bool.isRequired,
	hasWooCommerceInstalled: PropTypes.bool.isRequired,
	isStaticHomePage: PropTypes.bool.isRequired,
	isRequestingSitePlugins: PropTypes.bool.isRequired,
	site: PropTypes.object.isRequired,
	siteId: PropTypes.number.isRequired,
	ssoModuleActive: PropTypes.bool.isRequired,
	trackViewSiteAction: PropTypes.func.isRequired,
```

daledupreez (commented):
Functionally, this is working as expected for me, but I did drop some comments about switching the checks to be in parallel. That probably won't prevent the flashing, but may reduce the time the content displays.

> Note that all Ecommerce sites have a flash of the Home content before showing the loader. How can we bypass this?

I think we should tackle that issue separately if we confirm that it's still happening after we switch to `Promise.allSettled()`. If we do still have a problem, then I suspect we may need some state to support the delay.

PR #72733, opened 2023-01-27
Plugins: Decode html-encoded title when installing or reading a plugin

### LINKED ISSUE #70273: Thank you page: plugin name is HTML escaped
When installing a plugin that has special characters in the title (eg. &<>%), on the thank you page, the title will be HTML escaped:
Image

epeicher:
#### Proposed Changes

* The following components display the plugin title from the backend. In some scenarios, for example, when the plugin is already installed, that title could contain html-encoded characters. This PR will use the `decodeEntities` to display it correctly 

|Remove modal | After|
|-------|------|
|CleanShot 2023-01-27 at 18 12 34@2x| CleanShot 2023-01-27 at 18 14 57@2x  |

#### Testing Instructions

- In a Business or eCommerce site
* Purchase a plugin that contains an ampersand character, for example `Health Check & Troubleshooting`
* In the thank you page, make sure you don't see any html-encoded characters, i.e. you don't see things like `&amp;`
* Navigate to the plugin details of the installed plugin, e.g. `plugins/health-check/:siteId`
* Near the `Deactivate` button on the sidebar, click on the ellipsis button and click on `Remove`
* In the dialogue that appears, make sure you don't see any html-encoded characters, i.e. you don't see things like `&amp;`

#### Performance considerations

In one of the scenarios, this change iterates over a list of sites, then for each of the sites, it iterates over a list of plugins and then it spreads the properties of each plugin to update the name. That implies multiple nested loops, **but** the assumptions are that in the worst-case scenario, the number of sites for a user is in the 10s or less, the number of plugins would be in the 30s or less, and the number of properties of a plugin is around 30, so the number of iterations in the worst-case scenario should not impact the performance noticeably.

#### Pre-merge Checklist

~- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?~
~- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~

Fixes #70273


client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you-plugin-section.tsx
```diff
@@ -100,13 +102,13 @@ export const ThankYouPluginSection = ( { plugin }: { plugin: any } ) => {
 				alt={
 					translate( "%(plugin)s's icon", {
 						args: {
-							plugin: plugin.name,
+							plugin: pluginName,
```
gavande1: While this is resolving the issue, this is not enough and we should try to find the root cause of the problem. We are already decoding the name and slug fields in `normalizePluginData()`. However, I might be wrong with my thinking process. 
https://github.com/Automattic/wp-calypso/blob/0ab30f791331895435ee69707964511d76b5135c/client/lib/plugins/utils.js#L198-L203

epeicher: Thanks for your comments. I found already the root cause of the problem and that was already discussed in [this other PR](https://github.com/Automattic/wp-calypso/pull/72424#issuecomment-1400515675), but we could re-assess and iterate if we think the other approach is better. 

My main concern is iterating over a collection of plugins in a collection of sites in [here](https://github.com/Automattic/wp-calypso/blob/9701076387f9185c8dce2d6de6d62796fc539b8e/client/state/plugins/installed/actions.js#L626). But I could add some traces and do a quick performance check and then we can decide. 

Do you think we should iterate over that approach?

epeicher:
Most of the issues have been fixed by applying the decoding in the reducer instead of the UI, and it is only done in a centralised area. But I still need to find out one issue
CleanShot 2023-01-31 at 19 25 52@2x

epeicher:
> Most of the issues have been fixed by applying the decoding in the reducer instead of the UI, and it is only done in a centralised area.

This has been fixed with latest commit [c538476](https://github.com/Automattic/wp-calypso/pull/72733/commits/c538476ff72d4436e237e0bb7c7baaa77f628ea2)

gcsecsey (approved):
The latest changes LGTM and work great! 🚢 

CleanShot 2023-02-01 at 16 17 43@2x

epeicher:
Hi @renancarvalho 👋, eventually, as part of this PR, we decided to implement this change in the `reducer` as you suggested. Iteration for the win! 😄 Thanks for your initial suggestion [here](https://github.com/Automattic/wp-calypso/pull/72424#issuecomment-1400379331) 🙌

PR #71285, opened 2022-12-16
Checkout: add VAT fields to tax fields

sirbrillig:
#### Proposed Changes

This PR adds a VAT and an Organization field to the `TaxFields` component which is rendered as the first step of checkout if there is no domain product or GSuite product in the cart.

The fields are only made available if the country code selected in the tax form is one of a set of hard-coded countries that support VAT.

**Currently this new field will not be visible in production as it is behind a feature flag.**

Unlike the other fields in the contact form which pre-populate from and update the cached contact details endpoints, the VAT fields pre-populate from and send their updates to the existing VAT details endpoints used by the calypso VAT pages (see https://github.com/Automattic/wp-calypso/pull/53133). This means that the fields are subject to the same validation and restrictions of that form (eg: it's not possible to edit the VAT ID once it's been set for the user). In effect, these fields are just another interface to the existing VAT form under `/me/purchases/vat-details`. _However_, there are two differences: 

First, the checkout version of the form does not include the "Address" field. I did this intentionally because I felt that it cluttered the form, added friction, and may not actually be required. I also did it because once we add this form to the domain form, which already includes address fields, it's likely to feel redundant and awkward (we can't use those address fields because they are separated into address1/address2/city/state/etc.).

Second, the `/me/purchases/vat-details` form uses the country code `UK` for the United Kingdom (which is incorrect; see https://github.com/Automattic/payments-shilling/issues/1276) and this form uses `GB`. This won't be as much of an issue after https://github.com/Automattic/wp-calypso/pull/72513.

The information collected by this form is sent to three endpoints: the VAT details endpoint, the shopping-cart endpoint (along with other tax data), and the transactions endpoint (as part of the shopping-cart data). **It's worth noting that if the checkbox that displays the form is _not checked_, the data _will not_ be sent to the shopping-cart endpoint or the transactions endpoint even if it's saved to the VAT details endpoint.** This is by design so that users can chose to apply their VAT information on a per-transaction basis.

Part of https://github.com/Automattic/payments-shilling/issues/1278

#### Screenshots

VAT checkbox unchecked:

vat-unchecked

VAT checkbox checked:

vat-unfilled

VAT checkbox checked for UK:

vat-uk-unfilled

VAT checkbox pre-filled:

vat-pre-filled

#### To do

- [x] Show fields for applicable countries.
- [x] Toggle fields with checkbox.
- [x] When fields are hidden (due to checkbox or unsupported country), do not send their values to any endpoints (validate, cache, transactions, cart).
[...]

PR #72734, opened 2023-01-27
Checkout: Add "Address for VAT" field to VatForm

sirbrillig:
#### Proposed Changes

This adds a new field to the `VatForm` component in checkout for the address (sometimes) required by our VAT processors. This field matches the one that already exists on the `/me/purchases/vat-details` page. It was left out of the checkout form intentionally in https://github.com/Automattic/wp-calypso/pull/71285 because, to quote myself,

> First, the checkout version of the form does not include the "Address" field. I did this intentionally because I felt that it cluttered the form, added friction, and may not actually be required. I also did it because once we add this form to the domain form, which already includes address fields, it's likely to feel redundant and awkward (we can't use those address fields because they are separated into address1/address2/city/state/etc.).

However, it's recently been decided that the address field is necessary in checkout as well.

Screenshot 2023-01-27 at 2 07 25 PM

Part of implementing https://github.com/Automattic/payments-shilling/issues/1278

D99532-code is required for this to update the data on the server.

Depends on https://github.com/Automattic/wp-calypso/pull/72740

#### Testing Instructions

- Apply D99532-code if it is not already merged.
- On this branch, add a product to your cart and visit checkout.
- If the checkout contact/billing details step is automatically completed, click the "Edit" button to make it active.
- Select a VAT country like "United Kingdom" in the country field and a valid postal code number like `NW1 4NP`.
- Click the VAT checkbox and verify that the VAT fields appear. Set the company name to `Test company`. Set the VAT Number to a valid one like `553557881` and fill in anything in the "Address for VAT" field.
- Click continue to complete the step. (Note: this is a test number that will only work when the API is sandboxed).
- Verify that the validation is successful and that the step completes.
- Click to edit the step again.
- Verify that you can no longer edit the VAT Number but that you can edit the organization and the address.
- Visit `/me/purchases/vat-details` and verify that the form there has the same information entered in checkout.
- Return to checkout (or reload the page), edit the step again, and verify that the form is pre-populated with the VAT details you entered before.

##### Verifying the data is sent correctly

- Open your browser's devtools panel to monitor network traffic and look for transactions to the `/me/shopping-cart` endpoint.
- Click to complete the step and verify that the shopping-cart endpoint request includes the VAT `address` information in its `tax` property.

PR #72740, opened 2023-01-28
Checkout: Keep VAT country in sync with parent country

sirbrillig:
#### Background

When VAT form support was added to checkout in https://github.com/Automattic/wp-calypso/pull/71285, there was a subtle bug. Checkout only has one country code form field and it has three uses: the shopping cart endpoint tax country, the cached domain contact endpoint country, and (now) the VAT data endpoint country. Normally, all three of these values are the same, but the VAT form has to have its own local version of the data because there are cases (Northern Ireland) where the country code for the VAT form will differ from the other two (although the shopping cart tax data will also use this value, it is still stored in a separate place in the code).

Here's the bug: if the cached contact details endpoint returns one country code, but the VAT data endpoint returns a different country code, we need to make sure that we use the correct one. In this case that's the cached contact details endpoint code because that is the one that the checkout form uses; if we used the VAT country code, which is essentially a hidden field, there would be no way to see or change it and you'd get strange validation errors from the VAT endpoint.

#### Proposed Changes

To prevent this bug, in this PR we make a more concerted effort to keep the "hidden" VAT country code field (it's not actually a field; just a variable) in sync with the `VatForm`'s parent country code field. Any time the parent's country code field (the only one on the form) changes, we update the VAT country code to match. We only break that sync if the Northern Ireland checkbox is checked.

In addition, this fixes a bug where changing the country from a VAT-supported country to a non-VAT-supported country would still send any existing VAT data to the shopping cart. This was happening because while the VAT data is cleared when the "Add VAT details" checkbox is unchecked, that side effect was not happening when the `VatForm` itself was unmounted as a result of the country no longer being applicable. Here we fix that issue by always rendering `VatForm` and having it hide itself when the country does not apply so it can continue to run side effects as needed.

#### Testing Instructions

This PR adds tests to reproduce the bugs mentioned above and prove that they are fixed. This also fixes several other tests that weren't quite testing what they claimed to be testing.

You can test this manually by using checkout to set a cached country code in the EU (like "Spain") and clicking "Continue"; then using the `/me/purchases/vat-details` page to set a VAT country that's different (like `GB`, along with valid VAT details; you can use the Vat ID `553557881` if the store is sandboxed).

After that's set up, visit checkout and you'll see that the country field shows the cached country (eg: "Spain"), but if you try to submit the form and then look at the contact step summary. Before this PR you'll see it reads `GB`; after this PR you'll see `ES`.

Finally, change the country code to "United States" and set a valid postal code like `90210`. Press "Continue" and verify that the summary shows `US`.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/domain-contact-details.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/domain-contact-details.tsx
@@ -11,7 +11,7 @@ import {
 } from 'calypso/lib/cart-values/cart-items';
 import { getTopLevelOfTld } from 'calypso/lib/domains';
 import useCartKey from 'calypso/my-sites/checkout/use-cart-key';
-import { VatForm, isVatSupportedFor } from './vat-form';
+import { VatForm } from './vat-form';
 import type { DomainContactDetails as DomainContactDetailsData } from '@automattic/shopping-cart';
 import type { DomainContactDetailsErrors } from '@automattic/wpcom-checkout';
 
@@ -45,9 +45,7 @@ export default function DomainContactDetails( {
 	const needsAlternateEmailForGSuite = needsOnlyGoogleAppsDetails;
 	const tlds = getAllTopLevelTlds( domainNames );
 
-	const isVatSupported =
-		config.isEnabled( 'checkout/vat-form' ) &&
-		Boolean( contactDetails.countryCode && isVatSupportedFor( contactDetails.countryCode ) );
+	const isVatSupported = config.isEnabled( 'checkout/vat-form' );
 
 	return (
 		<Fragment>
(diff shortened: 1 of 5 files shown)


pottedmeat:
- I cleared the VAT details for my testing account.
- I switched my billing information to Spain on the checkout form.
- I went to the VAT details edit page.
- I set country to GB and filled out VAT details.
- I went back to checkout and reloaded it
- "Add VAT details" is checked
- I see all my VAT details and the country is set to Spain
- I switched it to United Kingdom and gave it a UK postal code
- It saved correctly
- Reloading the page shows the same VAT details

pottedmeat (approved):
Code looks good.

My only feedback is that `setVatDetailsInForm` is updated sometimes with `savedVatDetails` and sometimes with `vatDetailsInForm`. I feel like I understand why, but it would be nice to have additional comments explaining it.

sirbrillig:
Thanks! I refactored the way that `VatForm` pre-fills its fields to fix the issue you described in https://github.com/Automattic/wp-calypso/pull/72734#issuecomment-1407433999 and also added a test for that bug. However, I made the fixes in https://github.com/Automattic/wp-calypso/pull/72734 and tried to backport them here ~and they don't work here. I don't have time to figure out why just now so please ignore this PR for the time being and just test that one~ I figured out why the tests were failing. It was just a typo.

sirbrillig:
Ok, I think I've got this working correctly again.

pottedmeat (approved):
Reran my tests. Differences from first run:

- Billing information was automatically expanded after filling out the form on the VAT details edit page but before changing it from Spain to UK

Everything else passed as before.

PR #72742, opened 2023-01-28
Plugins: Add indexed plugin selectors

robertf4:
#### Proposed Changes

This PR adds two new selectors: `getAllPluginsIndexedByPluginSlug` and `getAllPluginsIndexedBySiteId`. They will allow quick access to the plugins by either plugin slug or site ID, as opposed to the current method which is to call `getPlugins` and search the resulting list.

These are added in a file called `selectors-ts.ts` because the plan is for `selectors.js` to be refactored to TypeScript in later PRs and the selectors combined at that time.

#### Testing Instructions

This code isn't being used yet, so just review the code and check the tests.

#### Pre-merge Checklist


tyxla (commented):
Given that it's a draft, I'm not sure if this is ready to review, but I was just wondering what's the idea behind copying all other selectors here. Did we intend to only add the new selectors to this new `selectors-ts` file?

robertf4:
> Given that it's a draft, I'm not sure if this is ready to review, but I was just wondering what's the idea behind copying all other selectors here. Did we intend to only add the new selectors to this new selectors-ts file?

I did not intend for it to be a draft, oversight on my part.

I created `selectors-ts` to facilitate the conversion of `selectors.js` to TypeScript over time. The plan is for it to eventually end up with only `selectors.ts`.

tyxla:
> I did not intend for it to be a draft, oversight on my part.

Ah, no worries about that. I assumed it might be the case, but just wanted to be sure.

> I created selectors-ts to facilitate the conversion of selectors.js to TypeScript over time. The plan is for it to eventually end up with only selectors.ts.

Hm, I'm not sure I understand the plan though. We seem to be losing track of the history of all pre-existing selectors that way, don't we? Should we just introduce the new selectors and keep the old ones? Or am I misunderstanding your plan about maintaining the history of the pre-existing selectors?

robertf4:
> Hm, I'm not sure I understand the plan though. We seem to be losing track of the history of all pre-existing selectors that way, don't we? Should we just introduce the new selectors and keep the old ones? Or am I misunderstanding your plan about maintaining the history of the pre-existing selectors?

I will do my best to make Git think the files are the same, but at the end of the day this ends in essentially a rewrite and I don't think keeping the history of the selectors outweighs the many issues of the preexisting selectors.

tyxla:
> I will do my best to make Git think the files are the same, but at the end of the day this ends in essentially a rewrite and I don't think keeping the history of the selectors outweighs the many issues of the preexisting selectors.

I understand, but it's not only about Git thinking seeing them as the same file. It's about making atomic changes, atomic enough that a new engineer can understand them a year or two from now. I suggested a rough plan about how to split the changes from the other PR into somewhat atomic changes, and it can further be split into smaller pieces if needed. This is aligned to our policy of [keeping branches small and short-lived](https://github.com/Automattic/wp-calypso/blob/trunk/docs/git-workflow.md#short-branches-merge-early-and-often).

A first step here would be introducing the new selectors alongside the existing selectors, not in a new file. Maybe in this PR we can also add some new tests for the new selectors. The rest of the selectors would be unaffected by this first change. 

WDYT?

robertf4:
> A first step here would be introducing the new selectors alongside the existing selectors, not in a new file. 

I have these selectors in a new file because they are written in TypeScript, not JavaScript. I don't think it is a good use of time to rewrite them to JavaScript in order to re-rewrite them to TypeScript again later.

> Maybe in this PR we can also add some new tests for the new selectors. The rest of the selectors would be unaffected by this first change.

This PR does add new tests for the new selectors and doesn't affect the rest of the selectors, so I'm not sure that I'm following here.

tyxla (commented):
I think we're on the same page here, the only thing that I think we should do is reuse the existing selectors from the existing `selectors` file, instead of redeclaring them here.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * The server returns plugins store at state.plugins.installed.plugins are indexed by site, which means
+ * that the information for a plugin may be spread across multiple site objects. This selector transforms
+ * that structure into one indexed by the plugin slugs and memoizes that structure.
+ */
+export const getAllPluginsIndexedByPluginSlug = createSelector(
+	( state: AppState ) => {
+		if ( isRequestingForAllSites( state ) ) {
+			return {};
```
tyxla: Should this return the same empty object for performance optimization, instead of a new one every time? We could declare it above and reuse it here.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -0,0 +1,96 @@
...
+					} );
+
+					plugins[ plugin.slug ] = {
+						...plugins[ plugin.slug ],
+						...otherPluginInfo,
+						sites: {
+							...plugins[ plugin.slug ]?.sites,
+							[ siteId ]: sitePluginInfo,
+						},
+					};
```
tyxla, line 54: Would have been nice to avoid repetition with `getPluginsSelector()` here, but I guess we can follow up on this in another PR.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -0,0 +1,96 @@
...
+		const allPluginsIndexedByPluginSlug = getAllPluginsIndexedByPluginSlug( state );
+
+		return Object.values( allPluginsIndexedByPluginSlug ).reduce(
+			(
+				pluginsIndexedBySiteId: { [ siteId: number ]: { [ pluginSlug: string ]: Plugin } },
+				plugin
+			) => {
+				Object.keys( plugin.sites )
+					.map( ( siteId ) => Number( siteId ) )
+					.forEach( ( siteId ) => {
```
tyxla: Can we move the casting to a number to the `forEach()` to avoid yet another loop?

client/state/plugins/installed/README.md
```diff
@@ -40,6 +40,14 @@ Get plugins installed on a list of sites (can also be just one site, but it shou
 
 Get the most recent status for a plugin action (including "inProgress" for currently-running actions).
 
+### `getAllPluginsIndexedByPluginSlug( state: Object )`
+
+Get an object for retrieving a plugin by its slug
+
+### `getAllPluginsIndexedBySiteId( state: Object )`
+
+Get an object for retrieving the plugins for a site by a site ID
```
tyxla: This can be a bit confusing since we don't really pass a site ID, rather they're indexed by it, and it's not retrieving the plugins of a site, but rather all plugins indexed by site. Also, is this really about retrieving or not? 

Should we change it to something like:

```suggestion
Get an object of all installed plugins of each site, indexed by site ID.
```

client/state/plugins/installed/README.md
```diff
@@ -40,6 +40,14 @@ Get plugins installed on a list of sites (can also be just one site, but it shou
 
 Get the most recent status for a plugin action (including "inProgress" for currently-running actions).
 
+### `getAllPluginsIndexedByPluginSlug( state: Object )`
+
+Get an object for retrieving a plugin by its slug
```
tyxla:
```suggestion
Get an object of all installed plugins in all sites, indexed by the plugin slug.
```

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const nonExistingSiteId = 0;
+
+const createError = function ( error, message, name = false ) {
+	const errorObj = new Error( message );
+	errorObj.name = name || error;
+	return errorObj;
+};
+
+const state = deepFreeze( {
```
tyxla, line 23: If we're using `deepFreeze` to ensure we're not introducing any unnecessary mutation, should we use it to freeze the rest of the fixture objects?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -0,0 +1,193 @@
...
+		},
+		[ siteTwoId ]: {
+			active: false,
+			version: '1.6',
+			autoupdate: true,
+		},
+	},
+};
+
+describe( 'getAllPluginsIndexedByPluginSlug', () => {
```
tyxla, line 151: The following scenarios appear to be already tested in the unit tests:
* we're requesting plugins of a site
* with the same plugin in more than 1 site.

However they're not represented by their own separate tests. Is this something that makes sense to have separately tested, since it's a substantial part of the selector functionality?

Furthermore, should we also cover the case when we're requesting plugins for all sites?

Finally, should we introduce a few tests that verify that the cache is working fine?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -0,0 +1,193 @@
...
+	test( 'Returns the jetpack plugin', () => {
+		const plugins = getAllPluginsIndexedByPluginSlug( state );
+
+		expect( plugins[ 'jetpack' ] ).toEqual( jetpackWithSites );
```
tyxla, line 155: If this is the common usage of the selector, then we likely didn't create the API correctly. Should we pass a plugin slug to the selector, or have a separate selector that allows that?

tyxla (approved):
Thanks for splitting it out @robertf4 🙌 

I've left some feedback, nothing is critical, let me know what you think.

Feel free to ship once all items have been addressed. Thanks!

client/state/plugins/installed/selectors-ts.ts

client/state/plugins/installed/selectors-ts.ts
robertf4: It will be replaced eventually as I migrate the other selectors, `getPluginsSelector` will call this selector and not have that logic in it.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: Yeah, that's probably a good idea.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I see what you mean. Since these selectors will be used by the other selectors, I think it would be better to make these selectors internal then, and create some new ones like getAllPluginsForPluginSlug( slug ). I'll do that in the next PR.

client/state/plugins/installed/selectors-ts.ts
tyxla: Sounds good 👍

client/state/plugins/installed/test/selectors-ts.ts
tyxla: Sounds like a plan 👍

client/state/plugins/installed/test/selectors-ts.ts
robertf4: That was a mistake on my part, I accidentally put the assertions in a single test. I split them out now, and I added tests for the memoization.

The case for requesting all sites isn't covered by these selectors, so it will be in the next PR.

PR #72747, opened 2023-01-29
Adding new assembler flow

bangank36:
#### Proposed Changes

* Instead of redirect to the PA step of site-setup flow, we should create new Assembler flow with only PA step
* This will help introduce the pattern assembler flow into logged-out and logged-in theme showcase 
* PR #71852 will be retired

#### Testing Instructions

 
* Checkout this branch, if you are using calypso live link, open the browser console and run this to make sure the feature flag is enabled `document.cookie = 'flags=pattern-assembler/logged-out-showcase;max-age=1209600;path=/'`
* Log out from your wp account
* Access theme showcase via /themes
* Select blank-canvas-3 theme and confirm that /patternAssembler step is shown after finishing site creation
* Select other theme and confirm that the end of flow is /home

| Blank Canvas Card  <br> Large screen ( > 960px ) | Blank Canvas Card <br> Small Screen | Other Card |
|---------------------------------------------|--------------------------------|------------|
|   <video src="https://user-images.githubusercontent.com/10071857/215333419-d896a860-0113-45f0-ad6b-92f3be23f500.mp4"></video> |               <video src="https://user-images.githubusercontent.com/10071857/215333504-43a1c1ef-242c-4acc-b7c9-6d0b6c3910cb.mp4"></video>                 |   <video src="https://user-images.githubusercontent.com/10071857/215333705-bb4b69ed-06d3-4ea0-92e8-608951830cf4.mov"></video>  |

#### Pre-merge Checklist

#### Reference
Related to #71706
Detailed post pbxlJb-3oe-p2

### DIFF
--- a/client/landing/stepper/declarative-flow/assembler.ts
+++ b/client/landing/stepper/declarative-flow/assembler.ts
@@ -1,0 +1,53 @@
+import { useFlowProgress, ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { useDispatch } from '@wordpress/data';
+import { useSiteSlug } from '../hooks/use-site-slug';
+import { ONBOARD_STORE } from '../stores';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import PatternAssembler from './internals/steps-repository/pattern-assembler';
+import Processing from './internals/steps-repository/processing-step';
+import { Flow, ProvidedDependencies } from './internals/types';
+
+const assembler: Flow = {
+	name: ASSEMBLER_FLOW,
+	useSteps() {
+		return [
+			{ slug: 'patternAssembler', component: PatternAssembler },
+			{ slug: 'processing', component: Processing },
+		];
+	},
+
+	useStepNavigation( _currentStep, navigate ) {
+		const flowName = this.name;
+		const { setStepProgress, setPendingAction } = useDispatch( ONBOARD_STORE );
+		const flowProgress = useFlowProgress( { stepName: _currentStep, flowName } );
+		setStepProgress( flowProgress );
+		const siteSlug = useSiteSlug();
+
+		const exitFlow = ( to: string ) => {
+			setPendingAction( () => {
+				return new Promise( () => {
+					window.location.assign( to );
+				} );
+			} );
+
+			return navigate( 'processing' );
+		};
+
+		const submit = ( providedDependencies: ProvidedDependencies = {} ) => {
+			recordSubmitStep( providedDependencies, '', flowName, _currentStep );
+
+			switch ( _currentStep ) {
+				case 'processing':
+					return exitFlow( `/site-editor/${ siteSlug }` );
+
+				case 'patternAssembler': {
+					return navigate( 'processing' );
+				}
+			}
+		};
+
+		return { submit };
+	},
+};
+
+export default assembler;
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts
@@ -2,3 +2,12 @@ export const PATTERN_SOURCE_SITE_ID = 174455321; // dotcompatterns
 export const PUBLIC_API_URL = 'https://public-api.wordpress.com';
 export const PREVIEW_PATTERN_URL = PUBLIC_API_URL + '/wpcom/v2/block-previews/pattern';
 export const SITE_TAGLINE = 'Site Tagline';
+
+export const BLANK_CANVAS_DESIGN = {
+	slug: 'blank-canvas-3',
+	title: 'Blank Canvas',
+	recipe: {
+		stylesheet: 'pub/blank-canvas-3',
+	},
+	design_type: 'assembler',
+};
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -1,5 +1,5 @@
 import { isEnabled } from '@automattic/calypso-config';
-import { StepContainer, SITE_SETUP_FLOW, ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { StepContainer, SITE_SETUP_FLOW, SITE_ASSEMBLER_FLOW } from '@automattic/onboarding';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { useTranslate } from 'i18n-calypso';
 import { useState, useRef, useEffect } from 'react';
@@ -13,7 +13,7 @@ import { useSiteIdParam } from '../../../../hooks/use-site-id-param';
 import { useSiteSlugParam } from '../../../../hooks/use-site-slug-param';
 import { SITE_STORE, ONBOARD_STORE } from '../../../../stores';
 import { recordSelectedDesign } from '../../analytics/record-design';
-import { SITE_TAGLINE, BLANK_CANVAS_DESIGN } from './constants';
+import { SITE_TAGLINE } from './constants';
 import PatternLayout from './pattern-layout';
 import PatternSelectorLoader from './pattern-selector-loader';
 import { useAllPatterns } from './patterns-data';
@@ -371,7 +371,7 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 			<DocumentHead title={ translate( 'Design your home' ) } />
 			<StepContainer
 				stepName="pattern-assembler"
-				hideBack={ showPatternSelectorType !== null || flow === ASSEMBLER_FLOW }
+				hideBack={ showPatternSelectorType !== null || flow === SITE_ASSEMBLER_FLOW }
 				goBack={ onBack }
 				goNext={ goNext }
 				isHorizontalLayout={ false }
@@ -401,4 +401,3 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 };
 
 export default PatternAssembler;
-export { BLANK_CANVAS_DESIGN };
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -1,5 +1,5 @@
 import { isEnabled } from '@automattic/calypso-config';
-import { StepContainer } from '@automattic/onboarding';
+import { StepContainer, SITE_SETUP_FLOW, ASSEMBLER_FLOW } from '@automattic/onboarding';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { useTranslate } from 'i18n-calypso';
 import { useState, useRef, useEffect } from 'react';
@@ -8,12 +8,13 @@ import AsyncLoad from 'calypso/components/async-load';
 import DocumentHead from 'calypso/components/data/document-head';
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { requestActiveTheme } from 'calypso/state/themes/actions';
+import { useQuery } from '../../../../hooks/use-query';
 import { useSite } from '../../../../hooks/use-site';
 import { useSiteIdParam } from '../../../../hooks/use-site-id-param';
 import { useSiteSlugParam } from '../../../../hooks/use-site-slug-param';
 import { SITE_STORE, ONBOARD_STORE } from '../../../../stores';
 import { recordSelectedDesign } from '../../analytics/record-design';
-import { SITE_TAGLINE } from './constants';
+import { SITE_TAGLINE, BLANK_CANVAS_DESIGN } from './constants';
 import PatternLayout from './pattern-layout';
 import PatternSelectorLoader from './pattern-selector-loader';
 import { useAllPatterns } from './patterns-data';
@@ -35,7 +36,7 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	const { goBack, goNext, submit, goToStep } = navigation;
 	const { setThemeOnSite, runThemeSetupOnSite, createCustomTemplate } = useDispatch( SITE_STORE );
 	const reduxDispatch = useReduxDispatch();
-	const { setPendingAction } = useDispatch( ONBOARD_STORE );
+	const { setPendingAction, setSelectedDesign } = useDispatch( ONBOARD_STORE );
 	const selectedDesign = useSelect( ( select ) => select( ONBOARD_STORE ).getSelectedDesign() );
 	const intent = useSelect( ( select ) => select( ONBOARD_STORE ).getIntent() );
 	const site = useSite();
@@ -43,6 +44,7 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 	const siteId = useSiteIdParam();
 	const siteSlugOrId = siteSlug ? siteSlug : siteId;
 	const allPatterns = useAllPatterns();
+	const selectedTheme = useQuery().get( 'theme' );
 
 	const largePreviewProps = {
 		placeholder: null,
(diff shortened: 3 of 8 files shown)


client/landing/stepper/declarative-flow/assembler.ts
```diff
@@ -0,0 +1,53 @@
+import { useFlowProgress, ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { useDispatch } from '@wordpress/data';
+import { useSiteSlug } from '../hooks/use-site-slug';
+import { ONBOARD_STORE } from '../stores';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import PatternAssembler from './internals/steps-repository/pattern-assembler';
+import Processing from './internals/steps-repository/processing-step';
+import { Flow, ProvidedDependencies } from './internals/types';
+
+const assembler: Flow = {
```
arthur791004: I'd prefer to call it `siteAssembler` as `assembler` is a bit ambiguous. WDYT?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -59,8 +61,11 @@ const PatternAssembler: Step = ( { navigation, flow } ) => {
 
 	useEffect( () => {
 		// Require to start the flow from the first step
-		if ( ! selectedDesign ) {
+		if ( ! selectedDesign && flow === SITE_SETUP_FLOW ) {
 			goToStep?.( 'goals' );
+		} else if ( selectedTheme === 'blank-canvas-3' && flow === ASSEMBLER_FLOW ) {
+			// User has selected blank-canvas-3 theme from theme showcase and enter assembler flow
+			setSelectedDesign( BLANK_CANVAS_DESIGN as Design );
```
arthur791004: I think we can do it when the user lands on the site assembler flow instead of this step. For example, we can call `useEffect( () => { ... }, [] )` in the `useSteps` function (just like other flow does). But maybe there is a better way to do it.

bangank36: I'll make sure to name it `siteAssemblerFlow`, similar to `siteSetupFlow`, does it make sense?

client/landing/stepper/declarative-flow/assembler.ts
bangank36: I'll make sure to name it `siteAssemblerFlow`, similar to `siteSetupFlow`, does it make sense?

arthur791004: Sure!

bangank36: Update the file name as suggested and change the flow name into `site-assembler`

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
bangank36: Make sense, now the PA step won't have to preselect the design, this will happens on the Site assembler flow

packages/onboarding/src/utils/flows.ts
```diff
@@ -13,6 +13,8 @@ export const MIGRATION_FLOW = 'import-focused';
 export const COPY_SITE_FLOW = 'copy-site';
 export const BUILD_FLOW = 'build';
 export const WRITE_FLOW = 'write';
+export const ASSEMBLER_FLOW = 'site-assembler';
```
arthur791004: [nit] `SITE_ASSEMBLER_FLOW`

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts
```diff
@@ -2,3 +2,12 @@ export const PATTERN_SOURCE_SITE_ID = 174455321; // dotcompatterns
...
 export const PREVIEW_PATTERN_URL = PUBLIC_API_URL + '/wpcom/v2/block-previews/pattern';
 export const SITE_TAGLINE = 'Site Tagline';
+
+export const BLANK_CANVAS_DESIGN = {
+	slug: 'blank-canvas-3',
+	title: 'Blank Canvas',
+	recipe: {
+		stylesheet: 'pub/blank-canvas-3',
+	},
+	design_type: 'assembler',
```
arthur791004, line 12: I think it should be the same as the actual blank canvas design. Why do we need to assign a different value to it?

client/landing/stepper/declarative-flow/site-assembler-flow.ts
```diff
@@ -0,0 +1,68 @@
+import { useFlowProgress, ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { useDispatch } from '@wordpress/data';
+import { useEffect } from 'react';
+import { useQuery } from '../hooks/use-query';
+import { useSiteSlug } from '../hooks/use-site-slug';
+import { ONBOARD_STORE } from '../stores';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import PatternAssembler, {
+	BLANK_CANVAS_DESIGN,
+} from './internals/steps-repository/pattern-assembler';
```
arthur791004: [nit] I'd prefer to import `BLANK_CANVAS_DESIGN` from `pattern-assembler/constants.ts` directly and then we don't need to re-export from `pattern-assembler/index.ts`

client/landing/stepper/declarative-flow/site-assembler-flow.ts
```diff
@@ -0,0 +1,68 @@
...
+				return new Promise( () => {
+					window.location.assign( to );
+				} );
+			} );
+
+			return navigate( 'processing' );
+		};
+
+		const submit = ( providedDependencies: ProvidedDependencies = {} ) => {
+			recordSubmitStep( providedDependencies, '', flowName, _currentStep );
```
arthur791004, line 51: Would it be better to introduce new intent for the site assembler flow? Or is the flow name enough for tracking cc @autumnfjeld

client/landing/stepper/declarative-flow/site-assembler-flow.ts
bangank36: Okay

packages/onboarding/src/utils/flows.ts
bangank36: Good choice

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts
bangank36: @arthur791004 Sorry I don't get your concern, do you mind to explain more? In this flow we don't have the design from API, only got the themeSlug value from the URL

client/landing/stepper/declarative-flow/site-assembler-flow.ts
bangank36: I think the flow name should be enough for now cc: @autumnfjeld

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts
arthur791004: I mean the `design_type` property. I think we use `standard` in the Design Picker

arthur791004: Oh sorry, I might miss something before, and I found we overwrite the `design_type` of the blank canvas design to `assembler` when the user selects it in the Design Picker 😅 So, you can ignore this comment...

client/landing/stepper/declarative-flow/site-assembler-flow.ts
```diff
@@ -0,0 +1,67 @@
...
+
+			switch ( _currentStep ) {
+				case 'processing':
+					return exitFlow( `/site-editor/${ siteSlug }` );
```
arthur791004, line 56: Last thing: we have to store the completed flow in the session storage so that we can show the customized first slide to the people going through the PA step. See https://github.com/Automattic/wp-calypso/blob/76d9feaa83e7cc12c2fad0ab8ed626faf6abb0cf/client/landing/stepper/declarative-flow/site-setup-flow.ts#L223

bangank36: Ah, good to know

client/landing/stepper/declarative-flow/site-assembler-flow.ts
bangank36: @arthur791004 I think we can leave this as it is and address it on metrics issue, how do you think?

arthur791004: Sure!

arthur791004 (approved):
Tested and look good to me 👍

client/landing/stepper/declarative-flow/site-assembler-flow.ts
bangank36: Great!

client/landing/stepper/declarative-flow/site-assembler-flow.ts
```diff
@@ -0,0 +1,68 @@
...
+import type { Design } from '@automattic/design-picker/src/types';
+
+const siteAssemblerFlow: Flow = {
+	name: SITE_ASSEMBLER_FLOW,
+	useSteps() {
+		const { setSelectedDesign } = useDispatch( ONBOARD_STORE );
+		const selectedTheme = useQuery().get( 'theme' );
+
+		useEffect( () => {
+			if ( selectedTheme === 'blank-canvas-3' && this.name === SITE_ASSEMBLER_FLOW ) {
```
miksansegundo: We can use `selectedTheme === BLANK_CANVAS_DESIGN.slug` rather than `selectedTheme === 'blank-canvas-3'`. That would make easier future updates.

client/signup/config/flows.js
```diff
@@ -108,11 +109,23 @@ function getThankYouNoSiteDestination() {
 }
 
 function getChecklistThemeDestination( { siteSlug, themeParameter } ) {
+	const canGoToAssemblerFlow = isDesktop();
+
 	if (
 		themeParameter === 'blank-canvas-3' &&
```
miksansegundo: Here we can use `BLANK_CANVAS_DESIGN.slug` from the constants file too.

packages/onboarding/src/utils/flows.ts
```diff
@@ -13,6 +13,8 @@ export const MIGRATION_FLOW = 'import-focused';
 export const COPY_SITE_FLOW = 'copy-site';
 export const BUILD_FLOW = 'build';
 export const WRITE_FLOW = 'write';
+export const SITE_ASSEMBLER_FLOW = 'site-assembler';
```
miksansegundo, line 16: I find the new flow name confusing:

- `/setup/site-assembler/patternAssembler` - Used for a theme switch or an empty theme search.
- `/setup/site-setup/patternAssembler` - Used for the onboarding flow.

Can we find a clearer name? 
- Maybe `/setup/site-theme/patternAssembler`
- Maybe `/setup/site-design/patternAssembler`

miksansegundo (approved):
The changes work well 👍. Great job!

We could find a clearer name for the flow. See my suggestions. @Automattic/lego any better ideas?

This new flow will be triggered from a theme switch and also from an empty theme search (see [Themes: Improve the Empty Search Results Screen](https://github.com/Automattic/dotcom-forge/issues/1453)) and possibly from other places where we want to help users create their own theme from scratch.

image

miksansegundo:
BTW. When testing this PR, I found a critical issue, not on production, and it seems not related to these changes. Can you confirm? @bangank36 

When switching to any theme preserving the homepage within the logged-in showcase, the API `/themes/mine` returns an error 500 when the passing `dont_change_homepage: true`.

Screenshot 2566-01-30 at 16 56 53
Screenshot 2566-01-30 at 16 55 48
Screenshot 2566-01-30 at 16 55 29

client/landing/stepper/declarative-flow/site-assembler-flow.ts
bangank36: Nice catch

packages/onboarding/src/utils/flows.ts
bangank36: `site-theme` does not make sense to me, and `site-design` sounds like it more focus on the visualization of the site, and not include the content
I think this is not blocker though, will bring this up on tomorrow's sync
cc: @Automattic/lego

bangank36:
> Can you confirm? @bangank36

@miksansegundo I can reproduce the issue, I also think it is not related to this PR, since I was testing this issue on trunk
It should be something wrong with persistent template on theme switch, which is enabled via this flag `themes/theme-switch-persist-template`
Disable the flag via `?flags=-themes/theme-switch-persist-template` makes the issue gone, I think we should create an issue to investigate this. In the mean time, we should set the flag `themes/theme-switch-persist-template` to false on all enviroment

client/landing/stepper/declarative-flow/site-assembler-flow.ts
autumnfjeld: We have discussed the topic of intent in a few different places:
* p1671807817094079/1671770959.129519-slack-CRWCHQGUB
* pbxlJb-3bS-p2#comment-2201
* p1673498055131429-slack-C048CUFRGFQ


The problem with using `intent=assembler` is that we can't keep track of the build and write intent if we overwrite it with `assembler`. And the write intent is used in My Home to conditionally show content.   Thus, we have some constraints we must respect.  

Can you point to all the different `flowName` options we have?  To give me a sense of where else flow is used.

Is it this same flow. And [these](https://github.com/Automattic/wp-calypso/blob/a3402293c7319398fe5d1a60d0f183e17e6e55f6/client/landing/stepper/declarative-flow/registered-flows.ts) are all the flow possibilities?

image


I'm worried we are using flow in two different ways:
* stepper flow definition
* singup flow definition (e.g. when `flow=onboarding`)
* maybe these are all compatible? 
* see [FigJam](58MP46f44oGKl5Bin5bSp1-fi-0%3A1) (which I didn't yet update! oops!)
* 
image

autumnfjeld: I started a[ new metrics map](58MP46f44oGKl5Bin5bSp1-fi-0%3A1) for the logged out PA flow.  Let's fill in all the items correctly:

image

packages/onboarding/src/utils/flows.ts
autumnfjeld: Why are we mixing snake-case and camelCase?  This seems wrong.

bangank36: This is the current setup for the names I think
snake-case is for the [flow name](https://github.com/Automattic/wp-calypso/blob/2406cb8170dd0093027b05ff1400cc7ae8d91de5/packages/onboarding/src/utils/flows.ts), while camelCase is used for [step name](https://github.com/Automattic/wp-calypso/blob/9274711f50cb2334569b59cedd95c135f105e934/client/landing/stepper/declarative-flow/anchor-fm-flow.ts#L33). 
eg: /setup/site-setup/designSetup

autumnfjeld:
>  I think we should create an issue to investigate this

Thank you.
An issue to add to the theme switch view https://github.com/orgs/Automattic/projects/338/views/35

packages/onboarding/src/utils/flows.ts
autumnfjeld: Thank you.  I was just realizing that, going to other steps, like `https://wordpress.com/setup/site-setup/designSetup`

autumnfjeld: For flows coming from the Theme Showcase, maybe to be cohesive with `with-theme`  
 `https://wordpress.com/start/with-theme/user?ref=calypshowcase&theme=upsidedown`:

`/setup/with-assembler/patternAssembler`

autumnfjeld: To be clear, we need two path names, where the differentiation in path name is 
* logged-in -> theme switch PA flow
* logged-out -> onboarding PA flow

bangank36: CMIIW, but adding 2 path names means we have to create 2 flows in the Stepper Framework

bangank36:
I did the final test before merging

https://user-images.githubusercontent.com/10071857/215674417-9cc8e64b-7898-40e4-bf3d-e70d3e945d4f.mov

PR #23815, opened 2018-03-30
Analytics: do not omit site id when a site fragment is present in the path

gwwar:
This PR builds on top of #23728 to assume that when analytics paths are reported **correctly**, we only include the blog id as part of the payload when a site fragment `:site` is available. Eg `/comments/all/:site` but not for `/comments`.

On further investigation when we call `shouldReportOmitBlogId` the input given is the normalized path rather than the original one. Since this is the case we can also remove the section blacklist.

### Testing Instructions
- Enable debugging `localStorage.setItem('debug', 'calypso:analytics:tracks')`
- When visiting urls with a site fragment we see a `blog_id` that matches the slug
- When visiting urls without a site fragment we `blog_id` is not set.
- Click on My Sites and select a site. **Without refreshing/or pasting in the url** click on your profile, or visit the reader. We should not see blog_id set.

screen shot 2018-03-30 at 2 17 39 pm

Needed as part of #23647 

cc @yanirs if you prefer this general behavior or if you would prefer a blacklist for particular sections. For the latter there are corner cases with paths like `/me/purchases/:site/:purchaseId/cancel-privacy-protection` because it is in the profile section.

PR #23728, opened 2018-03-28
Analytics: Refactor `tracks.recordEvent` to omit `blog_id` when unneeded

Copons:
Fix #16900 

Events recorded from some specific paths should not report the `blog_id` value because it's used to tell general admin and site-specific activities apart.

This PR checks if the path from where an event record is originating begins with one of the paths from the "blacklist", and in that case, `null`s the event's `blog_id` property.

## Testing instructions

- Enable tracks debugging with `localStorage.setItem('debug', 'calypso:analytics:tracks')` and filter by `page_view`.
- Try loading the following list of paths, and make sure the `blog_id` is reported as `null`.
  - `/me`
  - `/help`
  - `/` (Reader root)
  - `/read`
  - `/following`
  - `/discover`
  - `/activities`
  - `/tag`
- Try loading any other path, and make sure that the `blog_id` is reported with an actual value (unless it's an "All Sites" page: there it should be `undefined`).

PR #72782, opened 2023-01-30
Themes: Add force_site_id in the recordTracksEvent for the theme search

BogdanUngureanu:
In order to collect the blog_id and the site_plan_id, we can pass the `force_site_id` flag so that they are automatically passed to the events.

#### Proposed Changes

* Pass `force_site_id` for the `calypso_themeshowcase_theme_click` and `calypso_themes_search_results`  events

#### Testing Instructions

* Use the calypso.live link
* Open DevTools
* Go to /themes/:site-id and search for a string
* Check if the t.gif for the events are now containing `blog_id`, `site_plan_id` and `site_id_label`

#### Pre-merge Checklist

### DIFF
--- a/client/my-sites/themes/search-themes-tracks.js
+++ b/client/my-sites/themes/search-themes-tracks.js
@@ -74,6 +74,7 @@ export default function SearchThemesTracks( { query = {}, themes = [], wporgThem
 		actual_count: actualCount,
 		wpcom_result_count: themesCount,
 		wporg_result_count: wporgThemesCount,
+		force_site_id: true,
 	} );
 
 	useEffect( () => {
--- a/client/my-sites/themes/themes-selection.jsx
+++ b/client/my-sites/themes/themes-selection.jsx
@@ -85,6 +85,7 @@ class ThemesSelection extends Component {
 			page_number: query.page,
 			theme_on_page: parseInt( ( resultsRank + 1 ) / query.number ),
 			action: snakeCase( action ),
+			force_site_id: true,
 		} );
 	};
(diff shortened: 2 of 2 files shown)


dsas:
This does add `blog_id` et al to the tracking gif but I'm not sure why, something seems off 🤔🤔🤔 

`force_site_id` is [only read in one place](https://github.com/Automattic/wp-calypso/blob/35a31624ce8a9a93b665276a2fad96930fa121d4/client/lib/analytics/super-props.js#L17-L21) and it's used to force an exception to the 'paths that shouldn't include the site ID' rule.

In both of the events above we don't pass a `path` property, so it counts as a path that shouldn't include the site ID. Even if we did pass a path I suspect the path would be something like `/themes/:theme_name/:site` which would count as a 'paths that shouldn't include the site ID' rule.

It looks like when `shouldReportOmitBlogId` was [implemented](https://github.com/Automattic/wp-calypso/pull/23728) the intention was that it should null blog_id on non site specific pages like /me and at some point since, that behaviour has been inverted?

Can @Copons shed any light?

dsas:
Oh in fact the problem was that `shouldReportOmitBlogId` originally returned false when there was no path (see #23728) and this was later changed to return true in #23815.  See https://jsfiddle.net/n9epcw1y/

I think we should abandon this PR and instead we should fix the bug introduced four years ago in #23815

BogdanUngureanu:
Good idea! I've opened https://github.com/Automattic/wp-calypso/pull/73248 instead. :)

PR #72786, opened 2023-01-30
Plans 2023 : Adds the tooltip component,and some examples of the usage of the tooltip

ddc22:
#### Proposed Changes

* Adds the initial tooltip component
* Showcases a few examples of a few tooltips in operation

#### Please Note The Following Caveats
* All of the tooltips have not been added in this PR and will be added by a separate change (Automattic/martech#1460) . This PR is shipped prematurely to unblock any other developers that want to use the tooltip.

image

#### Testing Instructions
- Go to /start/plans?flags=onboarding/2023-pricing-grid
- Make sure the tooltip popover works by hovering over `Storage` or any other features that have been provided with a feature explanation

#### Pre-merge Checklist

Related to #
Fixes Automattic/martech#1390

### DIFF
--- a/client/lib/plans/features-list.tsx
+++ b/client/lib/plans/features-list.tsx
@@ -229,13 +229,14 @@ import ExternalLinkWithTracking from 'calypso/components/external-link/with-trac
 import MaterialIcon from 'calypso/components/material-icon';
 import { DOMAIN_PRICING_AND_AVAILABLE_TLDS } from 'calypso/lib/url/support';
 
+const is2023OnboardingPricingGrid = isEnabled( 'onboarding/2023-pricing-grid' );
+
 export type FeatureObject = {
 	getSlug: () => string;
 	getTitle: ( domainName?: string ) => TranslateResult;
 	getAlternativeTitle?: () => TranslateResult;
 	getHeader?: () => TranslateResult;
 	getDescription?: ( domainName?: string ) => TranslateResult;
-	getExplanation?: () => TranslateResult;
 	getStoreSlug?: () => string;
 	getCompareTitle?: () => TranslateResult;
 	getIcon?: () => string | { icon: string; component: MemoExoticComponent< any > } | JSX.Element;
@@ -572,7 +573,11 @@ export const FEATURES_LIST: FeatureList = {
 					args: domainName,
 				} );
 			}
-
+			if ( is2023OnboardingPricingGrid ) {
+				return i18n.translate(
+					'Get a custom domain – like yoursite.com – free for the first year.'
+				);
+			}
 			return i18n.translate(
 				'All paid WordPress.com plans purchased for an annual term include one year of free domain registration. ' +
 					'Domains registered through this promotion will renew at our {{a}}standard rate{{/a}}, plus applicable taxes, after the first year.{{br /}}{{br /}}' +
@@ -591,8 +596,6 @@ export const FEATURES_LIST: FeatureList = {
 				}
 			);
 		},
-		getExplanation: () =>
-			i18n.translate( 'Get a custom domain – like yoursite.com – free for the first year.' ),
 	},
 
 	[ FEATURE_JETPACK_ESSENTIAL ]: {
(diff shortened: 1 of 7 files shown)


client/lib/plans/features-list.tsx
```diff
@@ -235,6 +235,7 @@ export type FeatureObject = {
 	getAlternativeTitle?: () => TranslateResult;
 	getHeader?: () => TranslateResult;
 	getDescription?: ( domainName?: string ) => TranslateResult;
+	getExplanation?: () => TranslateResult;
```
chriskmnds: NIT - These two ("explanation", and "description") are essentially equivalent in meaning. What makes their values and functionality different (is "explanation" something like "abbreviated description")? Maybe we can be more explicit in the name. :-)

client/my-sites/plan-features-2023-grid/plans-2023-tooltip.tsx
```diff
@@ -0,0 +1,42 @@
+import styled from '@emotion/styled';
+import { TranslateResult } from 'i18n-calypso';
+import { PropsWithChildren } from 'react';
+
+const TooltipContainer = styled.div`
+	position: relative;
+	cursor: default;
+`;
+const Tooltip = styled.div`
+	background: #101517;
```
chriskmnds: I think it's best, in general, if we avoid static values this way and find an equivalent (or near equivalent) variable from color palette. Same for the other color used below.

chriskmnds (changes requested):
Hey @jdc91 . Thanks for putting these backbone parts together and for the example usage. 😎 

I left some code comments/suggestions, but I am also not sure we need to be creating a separate tooltip component. We have a couple of variations in Calypso (e.g. [Tooltip component](https://github.com/Automattic/wp-calypso/tree/trunk/client/components/tooltip), or [Popover package](https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/popover)) and Gutenberg. I need to test these myself, but have we considered using one of those instead? Is there something we are trying to do differently (and is it worth the direction)?

dzver:
Is it supposed to also work here?

Screenshot 2023-01-31 at 12 45 20

Also here?
Screenshot 2023-01-31 at 12 46 30

And last but not least, when I resize the window, the tooltip stops showing. When I resize back to a similar resolution that I  had before, it appears.

client/lib/plans/features-list.tsx
ddc22: Good call, will reuse description.

ddc22:
@dzver I added the same examples on the main plans grid as well. However in mobile I am not sure how a hover feature should work because technically we cannot hover when in mobile 🤔 

@chriskmnds I have reused the Tooltip component and overridden it's styles in a wrapper tooltip component. Let me know your thoughts on this approach.

ddc22:
Fixing tests 👀

dzver (approved):
The tooltips don't seem to disappear now. The feedback by Christos seems to be addressed.

Do we have plans to add tooltips for all of the features? It looks like some don't have them.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7785034

client/my-sites/plan-features-2023-grid/plans-2023-tooltip.tsx
```diff
@@ -0,0 +1,52 @@
+import styled from '@emotion/styled';
+import { TranslateResult } from 'i18n-calypso';
+import { PropsWithChildren, useRef, useState } from 'react';
+import Tooltip from 'calypso/components/tooltip';
+
+const HoverAreaContainer = styled.span``;
+
+const StyledTooltip = styled( Tooltip )`
+	&.tooltip.popover .popover__inner {
```
chriskmnds, line 9: I guess the Tooltip component wasn't built flexibly enough to be restyled heavily. Ideally, we shouldn't be referencing external classes and structure. It creates coupling / these may change, esp. if component moves to a package, etc. It doesn't look like something we can address. I recall battling to avoid this (the need to reference on the consumer end external classes) when working on another package (TourKit). 🤷 

On another end, it's also not bad, and perhaps we shouldn't really be modifying some things too heavily (like background colors, borders, font-weight, etc.) on every use case. Having consistent styling for things like tooltips, popovers (etc.) across the app may be a good thing. cc @lucasmendes-design

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7785034

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7785034).

PR #72797, opened 2023-01-30
Add trial Plans page

### LINKED ISSUE #71807: WooExpress Trial: Plans page
We're waiting on designs for the proposed Plans page at `/plans/[siteSlug]` for the eCommerce trial. As of now, the Plans page does not allow a user to upgrade when on the trial:

screen-shot-2023-01-05-at-11 00 53-am

nelsonec87:
#### Proposed Changes

* This PR adds the custom Plans page for the eCommerce trial Plan.
* **It's still missing the final copy and actions for the buttons/links.**
* **Monthly/Annually toggle is gonna be added in a future PR**

#### Testing Instructions
* Open the "Plan" page (`/plans/<site-slug>`) for a site with the eCommerce trial plan
* Check that the new design is shown and that it matches the Figma design 

Figma:

HTML:

* Test accordion functionality.

* Open in mobile mode and compare with the designs:

Figma:

HTML:

Closes https://github.com/Automattic/wp-calypso/issues/71807


client/my-sites/plans/doughnut-chart/index.tsx
```diff
@@ -10,7 +10,7 @@ interface CustomPercentageVariable extends CSSProperties {
 	'--percentage': number;
 }
 
-const DoughnutChart: React.FunctionComponent< Props > = ( props ) => {
+const DoughnutChart = ( props: Props ) => {
```
daledupreez: Yay for simplicity!
We can also consider renaming `Props` to `DoughnutChartProps`.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+	getECommerceTrialExpiration,
+	isECommerceTrialExpired,
+} from 'calypso/state/sites/plans/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import DoughnutChart from '../doughnut-chart';
+import TrialFeatureCard from './trial-feature-card';
+
+import './style.scss';
+
+const ECommerceTrialPlans = () => {
```
daledupreez: Nit: should this not be `ECommerceTrialPlan`? We can only have one plan on a site.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+	 */
+	const trialProgress = 1 - eCommerceTrialDaysLeft / trialDuration;
+	const eCommerceTrialDaysLeftToDisplay = isTrialExpired ? 0 : eCommerceTrialDaysLeft;
+
+	// moment.js doesn't have a format option to display the long form in a localized way without the year
+	// https://github.com/moment/moment/issues/3341
+	const readableExpirationDate = eCommerceTrialExpiration?.toDate().toLocaleDateString( locale, {
+		month: 'long',
+		day: 'numeric',
+	} );
```
daledupreez: This looks super-similar to the My Plan content and logic. It may be worth creating a `ECommerceTrialDaysRemainingDoughnutChart` wrapper component, though that can be handled in a follow-up.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+				{
+					title: translate( 'Sell in over 60 countries' ),
+					subtitle: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. ',
+				},
+				{
+					title: translate( 'Advanced subscriptions' ),
+					subtitle: translate(
+						'Add subscription plans to any product – plus discounts for subscribing and a way for customers to purchase subscriptions for others.'
+					),
+				},
```
daledupreez: Maybe keep these untranslated for now? I am not sure if we're going to stick with these features or details.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+					title: translate( 'Sales channels' ),
+					subtitle: translate(
+						'Promote and sell your products on multiple sales channels, including social media and online marketplaces.'
+					),
+				},
+				{
+					title: translate( 'Marketing automation' ),
+					subtitle: translate(
+						'Build custom automations so you can engage customers at every stage of their journey.'
+					),
```
daledupreez: Maybe keep this untranslated? We've been having a lengthy internal discussion about whether we're going to include AutomateWoo or MailPoet first, so the features here may depend on which one we pick.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+						'Manage the RMA process, add warranties to products and let customers request/manage returns from their account.'
+					),
+				},
+			],
+		},
+	];
+
+	return (
+		<>
+			<BodySectionCssClass bodyClass={ [ 'is-trial-plan' ] } />
```
daledupreez: To make this more specific, can we use `is-ecommerce-trial-plan`?

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+	const eCommerceTrialDaysLeftToDisplay = isTrialExpired ? 0 : eCommerceTrialDaysLeft;
+
+	// moment.js doesn't have a format option to display the long form in a localized way without the year
+	// https://github.com/moment/moment/issues/3341
+	const readableExpirationDate = eCommerceTrialExpiration?.toDate().toLocaleDateString( locale, {
+		month: 'long',
+		day: 'numeric',
+	} );
+
+	const [ interval, setInterval ] = useState( 'yearly' );
```
daledupreez: Two thoughts here:
1. The default interval should be based on the interval for the user's current plan
2. It may be worth using `useState<PlanInterval>( 'yearly' );` here so we can apply type checks to the downstream `setInterval()` calls.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+					subtitle: translate(
+						'Streamline your checkout and boost conversions with the Cart and Checkout blocks.'
+					),
+				},
+			],
+		},
+		{
+			illustration: '/calypso/images/plans/wpcom/ecommerce-trial/growth.png',
+			title: translate( 'Marketing and growth' ),
+			subtitle: 'Optimize your store for sales by adding in email and social integrations.',
```
daledupreez, line 174: Note: We should switch this back to lorem ipsum, wrap the content in `translate()`, or make a note of the untranslated string to make sure we don't miss it later.

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+							// TODO: translate when final copy is available
+							'per month, $540 billed annually'
+						}
+					</span>
+					<span className="e-commerce-trial-plans__price-card-savings">
+						{
+							// TODO: translate when final copy is available
+							'SAVE 31% BY PAYING ANNUALLY'
+						}
+					</span>
```
daledupreez: I think we should implement this now, but we should use at least some interpolation. Something like the following:
```suggestion
					const priceContent = isAnnualSubscription
						translate(
							'{{priceWrapper}}%(monthlyPrice)s{{/priceWrapper}} per month, %(annualPrice)s billed annually',
							{
								args: {
									monthlyPrice: '$45', // we should be able to compute both prices, including the symbol from the currentPlan object and/or a related selector
									annualPrice: '$540',
								},
								components: {
									priceWrapper: <span className="e-commerce-trial-plans__price-card-value" />,
								},
							}
						)
						: translate(
							'{{priceWrapper}}%(monthlyPrice)s{{/priceWrapper}} per month',
							{
								components: {
									priceWrapper: <span className="e-commerce-trial-plans__price-card-value" />,
								}
							}
						);
					const annualUpsellContent = ! isAnnualSubscription
						? null
						: (
							<span className="e-commerce-trial-plans__price-card-savings">
								{ translate(
									// apply upper-casing via CSS
									'Save %(percentageSavings)s%% by paying annually',
									{
										percentageSavings: 31, // compute this here
									}
								) }
							</span>
						);
					// ...
					{ priceContent }
					{ annualUpsellContent }
```

client/my-sites/plans/ecommerce-trial/style.scss
```diff
@@ -0,0 +1,191 @@
...
+
+				@media (max-width: $break-mobile) {
+					font-size: $font-headline-small;
+					margin-top: 30px;
+					font-weight: 400;
+				}
+
+				.e-commerce-trial-plans__price-card-value-symbol {
+					@media (max-width: $break-mobile) {
+						// Make a <sup>-like style
```
daledupreez: Can we not just use `<sup>` as a wrapper element?

client/my-sites/plans/ecommerce-trial/style.scss
```diff
@@ -0,0 +1,191 @@
...
+						/* stylelint-disable-next-line declaration-property-unit-allowed-list */
+						font-size: 70%;
```
daledupreez: Can we use `0.7em` instead? I am not sure if that's possible, but it would be better to be explicit so we don't need to ignore the style rule.

client/my-sites/plans/ecommerce-trial/trial-feature-card/index.tsx
```diff
@@ -0,0 +1,61 @@
+import { Card, Gridicon } from '@automattic/components';
+import { useState } from 'react';
+
+import './style.scss';
+
+interface Props {
```
daledupreez: Nit: I would prefer this to be called `TrialFeatureCardProps` so the intent is super-explicit.

client/my-sites/plans/ecommerce-trial/trial-feature-card/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+						) ) }
+					</div>
+				) }
+			</div>
+			<div className="trial-feature-card__accordion-toggle">
+				{ expanded ? (
+					<Gridicon icon="chevron-down" size={ 16 } onClick={ toggleExpanded } />
+				) : (
+					<Gridicon icon="chevron-up" size={ 16 } onClick={ toggleExpanded } />
+				) }
```
daledupreez: Tiny nit/suggestion: this could be simplified into an internal condition.
```suggestion
				<Gridicon icon={ expanded? 'chevron-down' : 'chevron-up' } size={ 16 } onClick={ toggleExpanded } />
```

client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
```diff
@@ -0,0 +1,98 @@
...
+
+	@media (max-width: $break-mobile) {
+		padding: 24px 16px;
+	}
+
+	&:last-child {
+		border-radius: 0 0 2px 2px;
+	}
+
+	&__illustration {
```
daledupreez: I am not sure if using `&` in this fashion is helpful overall. That makes it impossible to use simple _Find_ operations to locate the relevant CSS classes in the code. I think it is better to list out the full class names explicitly.

I've kicked off this thread with the team to discuss: p1675769680790209-slack-C0Q664T29

client/my-sites/plans/main.jsx
```diff
@@ -194,8 +195,8 @@ class Plans extends Component {
 				<QueryPlans />
 				<TrackComponentView eventName="calypso_plans_view" />
 				<Main
-					fullWidthLayout={ is2023OnboardingPricingGrid }
-					wideLayout={ ! is2023OnboardingPricingGrid }
+					fullWidthLayout={ is2023OnboardingPricingGrid && ! isEcommerceTrial }
+					wideLayout={ ! ( is2023OnboardingPricingGrid && ! isEcommerceTrial ) }
```
daledupreez: Maybe use a positive check here? As written, this is not obvious to me.
```suggestion
					wideLayout={ isEcommerceTrial || ! is2023OnboardingPricingGrid }
```

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+					</SegmentedControl.Item>
+
+					<SegmentedControl.Item
+						selected={ interval === 'yearly' }
+						onClick={ () => setInterval( 'yearly' ) }
+					>
+						<span>{ translate( 'Pay Annually' ) }</span>
+					</SegmentedControl.Item>
+				</SegmentedControl>
+			</div>
```
daledupreez: The main Plans UI defaults to the current term if no term is specified in the URL, but the toggles switch to a URL that includes the term in the path, i.e. `/plans/[yearly|monthly]/[siteSlug]`, and this URL is kept when switching sites.
So I think it would be best to skip the toggle for this PR, and implement it in a follow-up, as the behaviour is just weird as it is. Alternatively, we should let the interval be injected as a prop from the plans wrapper - it looks like it is available in the `intervalType` prop in the main `Plans` component.

Also, we probably want the CTAs and flows to default to monthly for eCommerce trials, so we should cater for that too.

daledupreez (commented):
Things are looking pretty good overall, but I do have some feedback I think we should tackle before we get this shipped.

client/my-sites/plans/ecommerce-trial/index.tsx
nelsonec87: The idea behind it was `ECommerceTrial[version of]Plans[page]`. But it wasn't clear, obviously :)

client/my-sites/plans/ecommerce-trial/index.tsx
nelsonec87: Agreed. As I mentioned in the first comment, maybe we can reuse the whole first card.

client/my-sites/plans/doughnut-chart/index.tsx
nelsonec87: Np! I used `Props` because the pattern is used in 100+ other components, but I agree, the more explicit, the better :)

client/my-sites/plans/ecommerce-trial/style.scss
nelsonec87: `em` is also not allowed, but I just put the specific size for this instance (`$font-title-medium`)

client/my-sites/plans/ecommerce-trial/style.scss
nelsonec87: I did it in the css because it's only "sup" in the mobile view. 
I think it's a bit cleaner styling the element than having two different elements (one of them `<sup>`) and toggling between them. What do you think?

client/my-sites/plans/ecommerce-trial/index.tsx
nelsonec87: I think it has complexity enough to have its own PR.

client/my-sites/plans/ecommerce-trial/index.tsx
nelsonec87: I've changed it to `ECommerceTrialPlansPage` to indicate it's the "Plans Page" customized to the "eCommerce Trial".
Please, let me know if it's still confusing.

client/my-sites/plans/ecommerce-trial/index.tsx
nelsonec87: Revolving conversation since we are removing the interval selection from this PR.

nelsonec87:
Rebased several changes from https://github.com/Automattic/wp-calypso/pull/72960

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,314 @@
...
+	const isAnnualSubscription = true;
+
+	const percentageSavings = Math.floor(
+		( 1 - eCommercePlanPrices.annualPlanMonthlyPrice / eCommercePlanPrices.monthlyPlanPrice ) * 100
+	);
+
+	const features = [
+		{
+			expanded: true,
+			illustration: '/calypso/images/plans/wpcom/ecommerce-trial/general-features.png',
```
sixhours: We're phasing out the `/static/images` folder according to [the directory's README](https://github.com/Automattic/wp-calypso/tree/trunk/static/images). We should move these into `/client/assets/images`, that way we can take advantage of webpack. Example:

```
// Import images at the top
import generalFeatures from 'calypso/assets/images/ecommerce-trial/general-features.png';
```

Then:

`illustration: generalFeatures,`

sixhours:
Looking good so far! I have some pretty minor non-blocking feedback.

Clicking between the My Plan and Plans tabs causes the text/padding in the trial banner to shift; it's a little jarring:

https://user-images.githubusercontent.com/2124984/217658381-eb10175c-d0f9-4c62-8888-c7391bfb88ff.mov

Should the annual total have a currency symbol?

Screen Shot 2023-02-08 at 4 37 46 PM

I'm guessing the doughnut chart isn't quite working yet? This visual seems inaccurate as a percentage of the amount of time elapsed. 12 days out of 31 is ~ 38%, so I'd expect to see more than a third of the wheel bright pink.

Screen Shot 2023-02-08 at 4 38 00 PM

I also noticed the hit area on the accordions is quite small, limited to just the arrow icons; I expected to be able to click the accordion section title and have it work the same way.

nelsonec87:
@sixhours, Thanks for the feedback!

> Clicking between the My Plan and Plans tabs causes the text/padding in the trial banner to shift; it's a little jarring:
> 

Yes! After this PR is merged we'll use the page component for both pages, so it'll be solved :) 

> Should the annual total have a currency symbol?

Bug fixed

> I'm guessing the doughnut chart isn't quite working yet? This visual seems inaccurate as a percentage of the amount of time elapsed. 12 days out of 31 is ~ 38%, so I'd expect to see more than a third of the wheel bright pink.

It should be working! Do you mind checking the plans subscription date and expiry date?

My test site:

image
image


> I also noticed the hit area on the accordions is quite small, limited to just the arrow icons; I expected to be able to click the accordion section title and have it work the same way.

Fixed!

client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+			<div className="e-commerce-trial-banner__content">
+				<p className="e-commerce-trial-banner__title">{ translate( 'You’re in a free trial' ) }</p>
+				<p className="e-commerce-trial-banner__subtitle">
+					{ isTrialExpired
+						? translate(
+								'Your free trial has expired. Sign up to a plan to unlock new features and keep your store running.'
+						  )
+						: translate(
+								'Your free trial will end in %(daysLeft)d day. Upgrade to a plan by %(expirationdate)s to unlock new features and start selling.',
+								'Your free trial will end in %(daysLeft)d days. Upgrade to a plan by %(expirationdate)s to unlock new features and start selling.',
```
daledupreez: You're updating this language/wording in #73068. Can you get the correct language in this PR as well?

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+				{
+					args: {
+						currency: currencySymbol,
+						monthlyPrice: eCommercePlanPrices.annualPlanMonthlyPrice,
+						annualPrice: eCommercePlanPrices.annualPlanPrice,
+					},
+					components: {
+						monthlyPriceWrapper: <span className="e-commerce-trial-plans__price-card-value" />,
+						priceDescription: <span className="e-commerce-trial-plans__price-card-interval" />,
+						currencySymbol: <span className="e-commerce-trial-plans__price-card-value-symbol" />,
```
daledupreez: Minor note: we could use dedicated `const`s for these components, as they are the same across the two logic branches, and it may be simpler to define them once and use them twice.
```suggestion
						monthlyPriceWrapper,
						priceDescription,
						currencySymbol,
```

client/my-sites/plans/ecommerce-trial/style.scss
daledupreez: My only concern is that we could easily fail to mimic the `<sup>` in some way. If we're _sure_ we aren't getting it wrong, we can keep this as-is, but if we aren't, I think that using two elements and a show/hide at different breakpoints would be more foolproof and futureproof. The `top: -0.3em` in particular feels like it could cause weird behaviour in some edge cases.

client/my-sites/plans/style.scss
```diff
@@ -110,3 +110,103 @@ body.is-section-plans.is-domain-sidebar-experiment-user {
 		}
 	}
 }
+
+body.is-section-plans.is-trial-plan {
```
daledupreez: Should we not remove all this CSS? It looks like we've moved the content to `client/my-sites/plans/ecommerce-trial/style.scss`. The red flag for me is that we're using `.is-trial-plan` instead of `.is-ecommerce-trial-plan`, which tells me these styles aren't going to work!

client/my-sites/plans/ecommerce-trial/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+	const eCommercePlanMonthly = getPlans()[ PLAN_ECOMMERCE_MONTHLY ];
+
+	const eCommercePlanPrices = useSelector( ( state ) => ( {
+		annualPlanPrice: getPlanRawPrice( state, eCommercePlanAnnual.getProductId(), false ) || 0,
+		annualPlanMonthlyPrice: getPlanRawPrice( state, eCommercePlanAnnual.getProductId(), true ) || 0,
+		monthlyPlanPrice: getPlanRawPrice( state, eCommercePlanMonthly.getProductId() ) || 0,
+		currencyCode: getPlan( state, eCommercePlanAnnual.getProductId() )?.currency_code || '',
+	} ) );
+
+	const { symbol: currencySymbol } = getCurrencyObject( 0, eCommercePlanPrices.currencyCode );
```
daledupreez, line 32: This approach is technically correct, but I think we're going to need to revisit this in a follow-up PR as there is a subtle issue we're not handling at the moment. The issue is that when we're showing USD prices in countries outside of the US, we need to show `US$` as the currency symbol, not just `$`, as some countries use `$` to represent their local currency. This is likely available from other selectors, but I think we should get this first iteration out and fix that in a follow-up.

client/my-sites/plans/ecommerce-trial/style.scss
```diff
@@ -0,0 +1,126 @@
...
+	&.color-scheme {
+		--color-surface-backdrop: var(--color-surface);
+	}
+
+	.section-nav {
+		box-shadow: inset 0 -1px 0 rgba(0, 0, 0, 0.05);
+	}
+
+	.e-commerce-trial-plans__banner-wrapper {
+		margin: 20px 0;
```
daledupreez, line 13: I noticed that this is the cause of the padding "jump" @sixhours referred to -- the `.section-nav` element has a padding of 17px, which is what is currently in effect on the My Plan page. Just calling this out to make sure we don't miss it now.

client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
```diff
@@ -0,0 +1,99 @@
...
+						font-size: $font-body-extra-small;
+						color: var(--color-neutral-40);
+					}
+
+				}
+			}
+		}
+	}
+
+	.trial-feature-card__accordion-toggle {
```
daledupreez, line 89: The approach you've taken for the accordion "expansion" introduces some weird behaviour. I think we should keep the small landing area (possibly with increased padding), but drop the `inline-flex` addition.

With things as they are, the button becomes as tall as the entire visual area, but stays super-narrow. I think the right way to solve for this may be to restructure the HTML so the title, subtitle, and accordion toggle are in a single `header` wrapper, as that should allow for simpler styling for them as a group. We could then make all three elements trigger the toggle.

But I think that's something we should tackle in a follow-on PR, as this PR is already huge.

client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+			eCommerceTrialDaysLeft: Math.round( getECommerceTrialDaysLeft( state, selectedSiteId ) || 0 ),
+			eCommerceTrialExpiration: getECommerceTrialExpiration( state, selectedSiteId ),
+		} ) );
+
+	const locale = useLocale();
+	const moment = useLocalizedMoment();
+	const translate = useTranslate();
+
+	const trialStart = moment( currentPlan?.subscribedDate );
+	const trialEnd = moment( currentPlan?.expiryDate );
```
daledupreez: I think we should be using `moment.utc()` here, as these times are in UTC, not local time.

daledupreez (commented):
I think we are super-close, but there is still some cleanup to handle here, as well as some follow-ups.

Given how close we are, it may be worth deferring any rebases until after we wrap this, as that makes it much easier to track what changes you're making.

client/my-sites/plans/ecommerce-trial/style.scss
nelsonec87: I've replaced the `span` with a `sup`, but the styles are overwritten and standardized so it looks like any other text =/
I then added some style to the mobile version to make it look like the standard `sup`:
image

Result:

client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
nelsonec87: This is being handled here: https://github.com/Automattic/wp-calypso/pull/73174

daledupreez (commented):
Everything looks good _except_ for the `<sup>` implementation/emulation. I think should drop the `<sup>` implementation for this PR. When I tested using Chrome and Firefox mobile emulators, things didn't look great. I feel that's starting to be an unnecessary rabbit hole for a small visual effect.

#### Screenshots

**Chrome mobile emulation**
Screenshot 2023-02-09 at 19 58 21
**Firefox mobile emulation**
Screenshot 2023-02-09 at 19 59 22

daledupreez (approved):
With the `<sup>` emulation gone, this is LGTM!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7801996

client/my-sites/plans/ecommerce-trial/style.scss
daledupreez: This is getting addressed in #73192.

client/my-sites/plans/ecommerce-trial/index.tsx
daledupreez: I've created #73223 to track this specific issue.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7801996).

PR #72387, opened 2023-01-20
Hide Writing Prompte View all responses link

eoigal:
We want to hide the `View All Responses` link until the `dailyprompt-{id}` tag is being saved. 

This tag will be saved when Jetpack version updates on WPCOM. After that we will need to double check that the tag is live. If so, we can show this link again.

### Before
Screenshot 2023-01-20 at 19 11 18
### After
Screenshot 2023-01-20 at 19 10 30

Ref: pe7F0s-vl#comment-555

PR #72820, opened 2023-01-31
Writing Prompts: Show View Responses link in home card

eoigal:
This PR simply toggles the `View Responses` link in the writing prompt home card.

Screenshot 2023-01-31 at 12 33 10

This link will direct users to reader where they can see answers/responses to this specific writing prompt. Each writing prompt has a tag `dailyprompt-{prompt-id}` that is used to identify these responses.

This link was hidden in https://github.com/Automattic/wp-calypso/pull/72387 because the tag was not set on posts yet. 

This was introduced in https://github.com/Automattic/jetpack/pull/28387.
We also ran a script to backfill posts that were missing this tag in https://github.com/Automattic/loop/issues/59.

It should be ok to show this link again now that posts have this tag and the reader experience is complete.

### DIFF
--- a/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
+++ b/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
@@ -63,7 +63,7 @@ const BloggingPromptCard = () => {
 						</Button>
 					</EllipsisMenu>
 				</CardHeading>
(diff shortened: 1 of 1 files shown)


eoigal:
This will not be needed with https://github.com/Automattic/wp-calypso/pull/72834

PR #72823, opened 2023-01-31
Change behaviour of dotorg theme nudge

dsas:
#### Proposed Changes

Link the dotorg theme nudge displayed to sites that are not atomic
capable straight to the checkout for the business plan.

Include a redirect so that post-purchase, the user is returned back to
the themes page so they can then activate the theme.
    
Make the nudge wording clearer so that what you are buying is more
explicit.

image

#### Testing Instructions

1. Use live link go to /themes on a < business site
2. Search for a dot org only theme (not a TTn theme)
3. Click on the theme
4. You should see the theme detail page, including a nudge saying 'Upgrade your plan to access third-party themes' and directly mentioning the business plan.
5. Click the nudge and you should get sent to the checkout screen for a business plan 
6. After checkout you should get redirected back to the theme with a notification in the top right saying words to the effect of 'purchase complete'. The nudge shouldn't appear and instead an 'Activate' button should.

Repeat from the logged out theme showcase, this should go via the plans page instead so you can choose / create a site after you register or login.

#### Pre-merge Checklist

Related to https://github.com/Automattic/dotcom-forge/issues/1299

### DIFF
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -885,6 +885,7 @@ class ThemeSheet extends Component {
 		}
 
 		if ( hasWpOrgThemeUpsellBanner || hasThemeUpsellBannerAtomic ) {
+			const thisPageUrl = `/theme/${ themeId }/${ siteSlug }`;
 			pageUpsellBanner = (
 				<UpsellNudge
 					plan={ PLAN_BUSINESS }
@@ -896,13 +897,19 @@ class ThemeSheet extends Component {
 					}
 					description={ preventWidows(
 						translate(
-							'Instantly unlock thousands of different themes and install your own when you upgrade.'
+							'Instantly unlock thousands of different themes and install your own when you upgrade to the Business plan.'
 						)
 					) }
 					forceHref
 					feature={ FEATURE_UPLOAD_THEMES }
 					forceDisplay
-					href={ ! siteId ? '/plans' : null }
+					href={
+						siteId
+							? `/checkout/${ siteSlug }/business?redirect_to=${ encodeURIComponent(
+									thisPageUrl
+							  ) }`
+							: plansUrl
+					}
 					showIcon
 					event="theme_upsell_plan_click"
 					tracksClickName="calypso_theme_upsell_plan_click"
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -908,7 +908,7 @@ class ThemeSheet extends Component {
 							? `/checkout/${ siteSlug }/business?redirect_to=${ encodeURIComponent(
 									thisPageUrl
 							  ) }`
-							: plansUrl
+							: localizeUrl( 'https://wordpress.com/pricing' )
 					}
 					showIcon
 					event="theme_upsell_plan_click"
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -890,11 +890,7 @@ class ThemeSheet extends Component {
 				<UpsellNudge
 					plan={ PLAN_BUSINESS }
 					className="theme__page-upsell-banner"
-					title={
-						siteId
-							? translate( 'Upgrade your plan to access third-party themes' )
-							: translate( 'Access third-party themes with a Business plan' )
-					}
+					title={ translate( 'Access this third-party theme with the Business plan!' ) }
 					description={ preventWidows(
 						translate(
 							'Instantly unlock thousands of different themes and install your own when you upgrade to the Business plan.'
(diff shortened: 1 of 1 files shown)


client/my-sites/theme/main.jsx
```diff
@@ -885,20 +885,31 @@ class ThemeSheet extends Component {
...
 		if ( hasWpOrgThemeUpsellBanner || hasThemeUpsellBannerAtomic ) {
+			const thisPageUrl = `/theme/${ themeId }/${ siteSlug }`;
 			pageUpsellBanner = (
 				<UpsellNudge
 					plan={ PLAN_BUSINESS }
 					className="theme__page-upsell-banner"
-					title={ translate( 'Access this theme for FREE with a Business plan!' ) }
+					title={
+						siteId
+							? translate( 'Upgrade your plan to access third-party themes' )
```
mmtr: Nitpick: What if we also mention the Business plan in the nudge title?
```suggestion
							? translate( 'Upgrade to a Business plan to access third-party themes' )
```

client/my-sites/theme/main.jsx
```diff
@@ -885,20 +885,31 @@ class ThemeSheet extends Component {
...
 					) }
 					forceHref
 					feature={ FEATURE_UPLOAD_THEMES }
 					forceDisplay
-					href={ ! siteId ? '/plans' : null }
+					href={
+						siteId
+							? `/checkout/${ siteSlug }/business?redirect_to=${ encodeURIComponent(
+									thisPageUrl
+							  ) }`
```
mmtr: Nitpick: `encodeURIComponent` replaces `/` with `%2F` (e.g. `?redirect_to=%2Ftheme%2Fastra%2Fexample.wordpress.com#step2`), but slashes don't need to be escaped (`?redirect_to=/theme/astra/example.wordpress.com#step2` works too). Since `thisPageUrl` is defined a few lines above (i.e. it is not given as param), we can guarantee it doesn't include any character that needs to be escaped and simplify this a little bit.
```suggestion
							? `/checkout/${ siteSlug }/business?redirect_to=${ thisPageUrl }`
```

mmtr (approved):
Works great! Just left a couple of minor pieces of feedback.

client/my-sites/theme/main.jsx
```diff
@@ -885,20 +885,31 @@ class ThemeSheet extends Component {
...
+							: localizeUrl( 'https://wordpress.com/pricing' )
+					}
```
mmtr: Should this link to `/start/business` instead, similarly to the CTA displayed when there are no results in the theme showcase?

https://github.com/Automattic/wp-calypso/blob/7c125bd5721a55574a0bc3768fb7451b796225f7/client/components/themes-list/index.jsx#L255

dsas: That's a better idea yes 👍 

I think ideally when logged out then you'd go to /start/business if you sign-up and possibly to choose a site and checkout if you log-in. That's a different kettle of fish I think though, so lets go with this.

client/my-sites/theme/main.jsx
dsas: Cheers!

client/my-sites/theme/main.jsx
dsas: The repeated use of 'to' is valid English but it doesn't feel as slick for an upgrade nudge. I used some different wording instead that explicitly mentions the business plan and is still clear about this being a third party theme

client/my-sites/theme/main.jsx
```diff
@@ -890,11 +890,7 @@ class ThemeSheet extends Component {
 				<UpsellNudge
 					plan={ PLAN_BUSINESS }
 					className="theme__page-upsell-banner"
-					title={
-						siteId
-							? translate( 'Upgrade your plan to access third-party themes' )
-							: translate( 'Access third-party themes with a Business plan' )
-					}
+					title={ translate( 'Access this third-party theme with the Business plan!' ) }
```
mmtr, line 893: I was surprised to see that final exclamation mark (`!`). Is that intentional? Just asking because I think we don't use that tone very often (at least on nudges).

mmtr: Never mind, I didn't notice the previous nudge also had the exclamation mark 🤦. I was comparing just the changes from the latest commits, which is why I didn't see the original text:
Screenshot 2023-02-01 at 10 15 16

mmtr (approved):

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7787529

Thank you @dsas for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7787529).

PR #72827, opened 2023-01-31
Checkout: Only allow VAT fields in checkout version of TaxFields

sirbrillig:
#### Proposed Changes

Recently we added a special `VatForm` component to the `TaxFields` component so that we can request VAT details during checkout's contact step if no domain is in the cart. However, `TaxFields` is used in many places besides checkout, like the "Add credit card" form. Because `VatForm` relies on the checkout data store being available, it cannot be used in those other places.

In this diff we make showing the `VatForm` opt-in via a new Prop on `TaxFields` and only enable it in checkout.

Before:

Screenshot 2023-01-31 at 10 57 09 AM

After:

Screenshot 2023-01-31 at 10 57 32 AM

#### Testing Instructions

Visit `/me/purchases/add-payment-method` and verify that no errors are shown.


martian77 (approved):
LGTM. 👍

PR #72833, opened 2023-01-31
Marketplace Themes: Fix the CTA on the Theme page when the user isn't logged in

valterlorran:
#### Proposed Changes

* Clicking on the CTA button or on the upsell banner, wouldn't redirect the user to the appropriate page.

#### Testing Instructions

* Navigate to `/theme/makoney` in an incognito tab.
* Click on the CTA button. It should take you to the signup flow.
* Navigate again to `/theme/makoney` in an incognito tab.
* Click on the upsell banner, and you should be redirected to the pricing page.

Slack discussion #p1675112160128929-slack-C048CUFRGFQ

### DIFF
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -675,7 +675,7 @@ class ThemeSheet extends Component {
 		const label = this.getDefaultOptionLabel();
 		const price = this.renderPrice();
 		const placeholder = <span className="theme__sheet-button-placeholder">loading......</span>;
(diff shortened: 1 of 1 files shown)


sixhours (approved):
This works as described, but I'm encountering the same issue with logging in. I don't think that's a result of these changes but something related to login via localhost or calypso.live.

sixhours:
Aha, the lack of ability to log in is due to incognito mode :)

PR #72835, opened 2023-01-31
Checkout: Add types to checkout data store actions and selectors

sirbrillig:
#### Proposed Changes

Checkout uses a global data store built using the `@wordpress/data` package. This data store is somewhat like Redux and has Actions and Selectors, but there are some differences. Adding TypeScript types to a store is fairly complicated, so up until this point we've relied on the consumer to clarify the types of actions and selectors but this has several downsides. The biggest downside is that sometimes the data store has not been initialized, and in that case TypeScript will say that an action or selector is available when in fact it is undefined (one bug caused by this is the one fixed by https://github.com/Automattic/wp-calypso/pull/72827).

In this PR we jump through the hoops required to add types to the `'wpcom-checkout'` data store and then fix the type errors that are exposed as a result.

#### Testing Instructions

Automated tests should cover this.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -62,11 +62,7 @@ import WPCheckout from './wp-checkout';
 import type { PaymentProcessorOptions } from '../types/payment-processors';
 import type { CheckoutPageErrorCallback } from '@automattic/composite-checkout';
 import type { MinimalRequestCartProduct } from '@automattic/shopping-cart';
-import type {
-	ManagedContactDetails,
-	CountryListItem,
-	CheckoutPaymentMethodSlug,
-} from '@automattic/wpcom-checkout';
+import type { CountryListItem, CheckoutPaymentMethodSlug } from '@automattic/wpcom-checkout';
 
 const { colors } = colorStudio;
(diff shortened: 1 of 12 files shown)


sirbrillig:
Thanks @pottedmeat! TIL about the helper types that `@wordpress/data` [already has defined](https://github.com/Automattic/wp-calypso/blob/trunk/packages/data-stores/src/mapped-types.ts)!

However, it seems that there's still a bug in their types [where it claims that a dispatched action returns](https://github.com/Automattic/wp-calypso/blob/3df5299cdf7b9e4bd3dd239cd9af8972d6a0ae67/packages/data-stores/src/mapped-types.ts#L43) `void` when in fact it returns `Promise<void>`. I had [fixed this once in the DT types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60693), but it looks like I'll have to fix it again. 😩 

~In the meantime I've rebased out [your last change](https://github.com/Automattic/wp-calypso/pull/72835/commits/dfb93f1063ac8d69905ec5d10c19a875d20de309) and I'll rely on my custom types until that can be fixed.~ Nevermind. I'm going to try to fix it ASAP instead.

sirbrillig:
Trying to fix the `wordpress/data` types in https://github.com/Automattic/wp-calypso/pull/72841

PR #72837, opened 2023-01-31
Marketplace Themes: Adds a loading state while the site is transferred to atomic

valterlorran:
#### Proposed Changes

* Adds a loading state while the site is transferred to atomic on the theme page.

https://user-images.githubusercontent.com/1234758/216349115-bef9673b-881a-4ba9-8a17-dae21b440fa2.mov

* I'm using the `/rest/v1.1/sites/{SITE_ID}/themes/mine` endpoint to check if the active theme is the recently subscribed theme.

#### Testing Instructions

* Create a free site
* Navigate to `/theme/makoney/{YOUR_FREE_SITE_SLUG}`
* Click on the CTA `Upgrade to subscribe` and go through the checkout flow.
* Once you land on the theme page again, you should see a loading state while the site is being transferred.

Slack discussion: p1675165572273799-slack-C9EJ7KSGH


nelsonec87:
Hi! I tested it using calypso-live, but after the checkout, I was redirected to My Home. I went to the theme's page again, but it still had the regular "Activate this design" button. After a while, I refreshed and got the "Customize site" button instead.

Do I need to follow some specific flow?
Should I have been redirected to the theme page after checkout?

markbiek:
This is super weird. When I'm on this branch, I don't see the Free plan on the plans page.
image

If I switch back to `trunk`, the Free plan is there.
image

Not sandboxed or anything.

nelsonec87:
> This is super weird. When I'm on this branch, I don't see the Free plan on the plans page.


The free plan shows up for me on calypso live!

markbiek:
> The free plan shows up for me on calypso live!

I also see the free plan on calypso live. It's only on calypso.localhost that I'm not seeing it.

client/components/data/sync-active-theme/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	const [ attempts, setAttempts ] = useState( 0 );
+
+	useEffect( () => {
+		if ( activeTheme === themeId ) {
+			return onThemeActive();
+		}
+		if ( ! requestStarted && attempts < maxAttempts ) {
+			setAttempts( attempts + 1 );
+			waitFor( 5 ).then( () => dispatch( requestActiveTheme( siteId ) ) );
```
daledupreez: Should the delay here be configurable? I feel like it may be better to check every 2-3 seconds to keep the delay as short as possible. We might need to bump `maxAttempts` up a bit to compensate, but I would rather us check more frequently so users don't have to wait some multiple of 5 seconds.

client/my-sites/theme/main.jsx
```diff
@@ -130,12 +133,17 @@ class ThemeSheet extends Component {
...
+
 	scrollToTop = () => {
 		window.scroll( 0, 0 );
 	};
 
 	componentDidMount() {
 		this.scrollToTop();
+		if ( getQueryArgs()?.[ 'sync-active-theme' ] === 'true' ) {
+			this.props.themeStartActivationSync( this.props.siteId, this.props.themeId );
+		}
```
daledupreez: Might it be cleaner if we handled this via a prop that we inject from the controller in `client/my-sites/theme/controller.jsx` based on `context.query?.['sync-active-theme']`.

client/my-sites/theme/main.jsx
```diff
@@ -760,6 +782,20 @@ class ThemeSheet extends Component {
 		);
 	};
 
+	onThemeActive = () => {
```
daledupreez: Nit: Maybe call this `onAtomicThemeActive()` to make it clear that the function is intended for Atomic sites?

client/components/data/sync-active-theme/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	const activeTheme = useSelector( ( state ) => getActiveTheme( state, siteId ) );
+
+	const [ attempts, setAttempts ] = useState( 0 );
+
+	useEffect( () => {
+		if ( activeTheme === themeId ) {
+			return onThemeActive();
+		}
+		if ( ! requestStarted && attempts < maxAttempts ) {
```
daledupreez: Can we add a debug message or a Tracks event for the case where we hit/exceed maxAttempts? It feels like it would be _really_ useful to know how often that occurs, and what the total time was for that case.

client/my-sites/theme/main.jsx
```diff
@@ -760,6 +782,20 @@ class ThemeSheet extends Component {
...
+		if ( ! this.state.isAtomicTransferCompleted ) {
+			this.setState( {
+				isAtomicTransferCompleted: true,
+			} );
+
+			const { isAtomic, siteSlug } = this.props;
+			if ( ! isAtomic ) {
+				const newSiteSlug = siteSlug.replace( /\b.wordpress.com/, '.wpcomstaging.com' );
+				return page( `/theme/makoney/${ newSiteSlug }` );
+			}
```
daledupreez: Bug: You've hard-coded `makoney` in the URL.
```suggestion
			const { isAtomic, siteSlug, themeId } = this.props;
			if ( ! isAtomic ) {
				const newSiteSlug = siteSlug.replace( /\b.wordpress.com/, '.wpcomstaging.com' );
				return page( `/theme/${ themeId }/${ newSiteSlug }` );
			}
```

daledupreez (commented):
Overall, this is pretty simple and smooth, but I did find one outright bug that we need to address before deploying. Some of the other items can be tackled in follow-up PRs (e.g. the Tracks even when we exceed the maximum attempts).

client/components/data/sync-active-theme/index.tsx
daledupreez: Also, can we add a check for whether a transfer is in place, just in case somebody does land on the URL with the flag set to `true`, but the site isn't actually transferring?

sixhours (commented):
I didn't have a chance to review the code entirely but I did test this and it's working! 🥳

client/components/data/sync-active-theme/index.tsx
```diff
@@ -1,36 +1,54 @@
...
 	const requestStarted = useSelector( ( state ) => isRequestingActiveTheme( state, siteId ) );
 
 	const activeTheme = useSelector( ( state ) => getActiveTheme( state, siteId ) );
 
 	const [ attempts, setAttempts ] = useState( 0 );
+	const [ exceededMaxAttempts, setExceededMaxAttempts ] = useState( false );
+
+	useEffect( () => {
+		if ( exceededMaxAttempts ) {
+			dispatch( recordTracksEvent( 'calypso_sync_active_theme_exceeded_max_attempts' ) );
```
daledupreez: Can we add some props to the event so it's more useful?
```suggestion
			dispatch( recordTracksEvent(
				'calypso_sync_active_theme_exceeded_max_attempts',
				{
					delay,
					max_attempts,
					theme: themeId,
				}
			) );
```
We'll also need to add these to the deps for the effect. I did consider adding `attempts`, but that would trigger the effect every time `attempts` is incremented, which we don't want.

client/components/data/sync-active-theme/index.tsx
```diff
@@ -1,36 +1,54 @@
...
+			return onAtomicThemeActive();
 		}
 		if ( ! requestStarted && attempts < maxAttempts ) {
 			setAttempts( attempts + 1 );
-			waitFor( 5 ).then( () => dispatch( requestActiveTheme( siteId ) ) );
+			waitFor( delay ).then( () => dispatch( requestActiveTheme( siteId ) ) );
+		} else if ( attempts >= maxAttempts ) {
+			setExceededMaxAttempts( true );
 		}
 	}, [ siteId, dispatch, requestStarted, activeTheme ] );
```
daledupreez: Question: should attempts, maxAttempts, delay, and onAtomicThemeActive not be dependencies for this effect? They are all reference in the effect.

client/my-sites/theme/main.jsx
```diff
@@ -782,20 +784,24 @@ class ThemeSheet extends Component {
...
 			if ( ! isAtomic ) {
 				const newSiteSlug = siteSlug.replace( /\b.wordpress.com/, '.wpcomstaging.com' );
-				return page( `/theme/makoney/${ newSiteSlug }` );
+				return page( `/theme/${ themeId }/${ newSiteSlug }` );
 			}
 		}
 	};
 
+	onAtomicThemeActiveFailure = ( message ) => {
+		return this.props.errorNotice( message );
```
daledupreez: Do we need to `return` here? Isn't the call to `errorNotice()` effectively a `dispatch()` call under the covers?

daledupreez (commented):
I have some feedback and questions, but the only blocking thing for me is making sure we get more data in the Tracks event.

Also, I am concerned about the dependencies in the effects. I think we should try to avoid missing any of them.

client/components/data/sync-active-theme/index.tsx
valterlorran: Actually, I think we should only use `requestStarted` and `activeTheme`. Because these are the only 2 things that we need to get the update to execute this effect. What do you think?

client/components/data/sync-active-theme/index.tsx
```diff
@@ -0,0 +1,64 @@
+import { translate } from 'i18n-calypso';
```
daledupreez: We shouldn't import `translate` directly in functional components. We should be importing the `useTranslate()` hook so we get the appropriate context applied.
```suggestion
import { useTranslate } from 'i18n-calypso';
...
const translate = useTranslate();
```

client/components/data/sync-active-theme/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+					delay,
+					max_attempts: maxAttempts,
+					theme: themeId,
+				} )
+			);
+			return onFailure?.(
+				translate( 'Failed to check the theme activation status, please try again.' )
+			);
+		}
+	}, [ exceededMaxAttempts ] );
```
daledupreez: I still think we should include `delay`, `maxAttempts`, `themeId`, and `translate` as deps. (That assumes we build `translate` via the `useTranslate()` hook.) We want to make sure the effect is correctly updated when any of those change, even if we don't think they will change much.

client/components/data/sync-active-theme/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+		if ( activeTheme === themeId ) {
+			return onAtomicThemeActive();
+		}
+		if ( ! requestStarted && attempts < maxAttempts ) {
+			setAttempts( attempts + 1 );
+			waitFor( delay ).then( () => dispatch( requestActiveTheme( siteId ) ) );
+		} else if ( attempts >= maxAttempts ) {
+			setExceededMaxAttempts( true );
+		}
+	}, [ requestStarted, activeTheme ] );
```
daledupreez: I think we also need deps on the following values: `themeId`, `attempts`, `maxAttempts`, `delay`, `siteId`, and `setExceededMaxAttempts`. See the React docs for more on why: https://beta.reactjs.org/reference/react/useEffect#specifying-reactive-dependencies

client/my-sites/theme/main.jsx
```diff
@@ -130,12 +133,19 @@ class ThemeSheet extends Component {
...
 		window.scroll( 0, 0 );
 	};
 
 	componentDidMount() {
 		this.scrollToTop();
+
+		const { syncActiveTheme } = this.props;
+		if ( syncActiveTheme ) {
+			this.props.themeStartActivationSync( this.props.siteId, this.props.themeId );
+		}
```
daledupreez: It may be worth being a bit defensive against either `siteId` or `themeId` not being defined.
```suggestion
		const { siteId, syncActiveTheme, themeId, themeStartActivationSync } = this.props;
		if ( syncActiveTheme && siteId && themeId ) {
			themeStartActivationSync( siteId, themeId );
		}
```

client/state/themes/actions/add-external-managed-theme-to-cart.tsx
```diff
@@ -127,8 +127,12 @@ export function addExternalManagedThemeToCart( themeId: string, siteId: number )
 		const { origin = 'https://wordpress.com' } =
 			typeof window !== 'undefined' ? window.location : {};
 
+		const redirectTo = addQueryArgs( `${ origin }/theme/${ themeId }/${ siteSlug }`, {
+			...( ! isSiteEligibleForManagedExternalThemes ? { 'sync-active-theme': true } : {} ),
```
daledupreez: Might it be worth using the string `'true'` here given that we're expecting that as the incoming value? The transform from `true` to `'true'` feels closely tied to the underlying implementation, when that could easily be understood to represent `&sync-active-theme=1` or just `&sync-active-theme` with no value.
```suggestion
			...( ! isSiteEligibleForManagedExternalThemes ? { 'sync-active-theme': 'true' } : {} ),
```

daledupreez (commented):
I have some additional cleanup (mostly the deps), and some non-blocking suggestions. I think it's worth getting those fixed before another round of review.

Apologies for the super-slow turnaround here -- I lost track of this PR.

markbiek:
I'm still having trouble getting this one to work. I'm not sandboxed if that makes any difference.

When I tested the calypso.live link, I ended up at the Thank You page instead of being redirected back to the theme page.

https://user-images.githubusercontent.com/917632/220416423-72a42c48-543e-4bd3-8d06-77bc49d5b978.mp4

Then I tested on calypso.localhost. In that case, I _did_ get redirected back to the theme page and I see the loading state but I also got an activation error so it stayed stuck on that page forever.

https://user-images.githubusercontent.com/917632/220416645-cf3e75c4-b01a-43cc-84d4-5c496428a626.mp4

markbiek (approved):
Working as expected on calypso.localhost! I'm still seeing the same issue with calypso.live but my suspicion is that there's a flag getting lost somewhere (unrelated to this PR).

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7829296

Hi @valterlorran, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include this string: `Failed to check the theme activation status, please try again.`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7829296).

PR #72840, opened 2023-01-31
Remove links to /courses and replace it with /learn in Universal Header & Footer Navbars for Calypso

dkd903:
#### Proposed Changes

* Contexte: p5uIfZ-cyH-p2 & 1244-gh-Automattic/martech

#### Testing Instructions
* Navigate to /themes or /plugins in an incognito/logged-out session.
* Ensure that you see the "Learn WordPress" link in the "Resources" menu item in the header (for /themes and /plugins)
* Ensure that you see the "Learn WordPress" link in the "Resources" section in the footer (for /plugins)

* Switch locale via the footer select field, or by going to /es/plugins or /es/themes
    * Verify that the link to WPcourses is not visible, and the new /learn link is hidden

Note: If you switch locale, and still see the page in English, you probably need to build the language files. You can do that by running `yarn build-languages` or `BUILD_TRANSLATION_CHUNKS=true yarn start`.

Alternatively, test it via the Calypso live below, as that build will have the language files. 

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/layout/universal-navbar-footer/index.tsx
+++ b/client/layout/universal-navbar-footer/index.tsx
@@ -1,5 +1,9 @@
 import './style.scss';
-import { useLocalizeUrl, removeLocaleFromPathLocaleInFront } from '@automattic/i18n-utils';
+import {
+	useLocalizeUrl,
+	removeLocaleFromPathLocaleInFront,
+	useIsEnglishLocale,
+} from '@automattic/i18n-utils';
 import { useTranslate, getLocaleSlug } from 'i18n-calypso';
 import { useCallback, useState } from 'react';
 import { useSelector } from 'react-redux';
@@ -18,7 +22,7 @@ const UniversalNavbarFooter = () => {
 	const { shouldSeeDoNotSell, isDoNotSell, onSetDoNotSell } = useDoNotSell();
 	const [ isDialogOpen, setIsDialogOpen ] = useState( false );
 	const isLoggedIn = useSelector( isUserLoggedIn );
-
+	const isEnglishLocale = useIsEnglishLocale();
 	const currentRoute = useSelector( getCurrentRoute );
 	const pathNameWithoutLocale = removeLocaleFromPathLocaleInFront( currentRoute ).slice( 1 );
 
@@ -176,9 +180,11 @@ const UniversalNavbarFooter = () => {
 									</a>
 								</li>
 								<li>
-									<a href={ localizeUrl( 'https://wordpress.com/courses/' ) } target="_self">
-										{ translate( 'WordPress Courses' ) }
-									</a>
+									{ isEnglishLocale && (
+										<a href={ localizeUrl( 'https://wordpress.com/learn/' ) } target="_self">
+											{ translate( 'Learn WordPress' ) }
+										</a>
+									) }
 								</li>
 								<li>
 									<a
--- a/client/layout/universal-navbar-header/index.jsx
+++ b/client/layout/universal-navbar-header/index.jsx
@@ -1,5 +1,5 @@
 import './nav-style.scss';
-import { useLocalizeUrl } from '@automattic/i18n-utils';
+import { useLocalizeUrl, useIsEnglishLocale } from '@automattic/i18n-utils';
 import { useTranslate, getLocaleSlug } from 'i18n-calypso';
 import { useState } from 'react';
 import { useSelector } from 'react-redux';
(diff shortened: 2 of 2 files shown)


chickee0311 (commented):
@dkd903 I tested and things look good to me
- Ensure that you see the Learn WordPress Link in the Resources Menu Item in Header (For both themes and plugins sections) ✅ 
- Ensure that you see the Learn WordPress Link in the Resources Section in Footer (Only for Plugins section) ✅ 
- Ensure that the /learn link is not visible in the Menus for Non English Locale URLs (eg. /es/themes or /es/plugins) ✅

client/layout/universal-navbar-footer/index.tsx
```diff
@@ -171,11 +174,14 @@ const UniversalNavbarFooter = () => {
 										{ translate( 'Daily Webinars' ) }
 									</a>
 								</li>
-								<li>
-									<a href={ localizeUrl( 'https://wordpress.com/courses/' ) } target="_self">
-										{ translate( 'WordPress Courses' ) }
-									</a>
-								</li>
+								{ ( localeFromPath === 'en' ||
```
yuliyan: I assume you switch to using `localeFromPath`, because `getLocaleSlug()` always returns the default `en` locale slug on the SSR.

That's happens because `getLocaleSlug()` reads the locale slug from the [global i18n-calypso instance](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/src/index.js#L18), whereas for the plugins and themes sections, we create a [private i18n-calypso instance](https://github.com/Automattic/wp-calypso/blob/trunk/client/controller/ssr-setup-locale.js#L14).

If you want to get the actual locale slug from the private i18n-calypso instance, you can use either:

```js
const { localeSlug } = useTranslate();
```

or:

```js
```

Both approaches are using the same source for the locale slug (i.e. `i18n.getLocaleSlug()` [#1](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/src/use-translate.js#L6) [#2](https://github.com/Automattic/wp-calypso/blob/trunk/client/components/calypso-i18n-provider/index.tsx#L13)), so it shouldn't really matter which one you use.

As for the conditional rendering check - it could probably be replaced by [isDefaultLocale](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-utils/src/utils.ts#L28) or [useIsEnglishLocale](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-utils/src/locale-context.tsx#L118).

That being said, the use of `localeFromPath` seems to be redundant, unless there's another reason for using it that I'm missing.

dkd903: @yuliyan

The reason for using `localeFromePath` is we need to determine what language is the given page translated in. In english Locale, both `/themes` and `/es/themes` returns `true` for the method `useIsEnglishLocale` and returns `en` for the method `useTranslate` - we do not want this.

Basically, for GoogleBot and users who land on the page /es/themes (and others in non English language) the /learn/ link should not be displayed in the menu as /learn/ is not available in any non-english language.

yuliyan: @dkd903, if `useIsEnglishLocale` returns `true` in all cases on your local testing, it could be due to missing translation files. Could you try booting up Calypso with `BUILD_TRANSLATION_CHUNKS=true yarn start` to ensure translation files are downloaded?

I've just tested on my environment and I'm seeing the expected results.

dkd903: @yuliyan - that actually did the trick. I have updated the PR and thank you so much for your help and guidance here!

client/layout/universal-navbar-footer/index.tsx
```diff
@@ -11,6 +11,7 @@ import { navigate } from 'calypso/lib/navigate';
 import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 
 const UniversalNavbarFooter = () => {
+	const isEnglishLocale = useIsEnglishLocale();
 	const translate = useTranslate();
 	const localizeUrl = useLocalizeUrl();
 	const locale = getLocaleSlug() ?? undefined;
```
yuliyan: This incorrect locale value seems to be used in a few `localizeUrl` calls, which may cause unexpected results on the SSR. However, that seems to be out of the scope of this PR so we should probably address it in a separate PR.

yuliyan (approved):
Tested the changes and everything seems to be working as expected.

I noticed that some e2e tests are failing, but it seems to be unrelated to the changes here and will probably be resolved after a rebase.

LGTM 🚀

yuliyan (approved):
Tested again - works as expected and all checks are green ✅, LGTM!

PR #43092, opened 2020-06-09
Add error handling for new site creation failure modes

lsl:
#### Changes proposed in this Pull Request

* Displays an error message if create site returns with an error.

#### Testing instructions

Full flow:
* Apply D44621-code
* Attempt to create a site using http://calypso.localhost:3000/new
* Clicking `Continue` on the `/new/plans` step should begin showing `<CreateSite />` before transitioning to the error message showing a "Go Back" button to return to the last known good step.

Error only:
* `ENTRY_LIMIT=entry-gutenboarding yarn start`
* Visit http://calypso.localhost:3000/new/error (I'll remove this route before merge)

Mobile:
Peek 2020-06-15 13-57-mobile

Desktop:
Peek 2020-06-15 13-54

Fixes #41067

PR #72841, opened 2023-01-31
@wordpress/data: Change DispatchFromMap to return Promise<void> instead of void

sirbrillig:
#### Proposed Changes

This changes the types of `@wordpress/data` to show that a dispatched action will return `Promise<void>` instead of `void`.

See https://github.com/WordPress/gutenberg/blob/6aef6e6b9b5cf2f5aeff6145cd24181edf0d43f0/packages/data/README.md#user-content-dispatch

> Note: Action creators returned by the dispatch will return a promise when they are called.

This was previously [fixed in the DT types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60693) but it's come back in the package types.

This was discovered when I tried to use the mapped dispatch type in https://github.com/Automattic/wp-calypso/pull/72835 (props @pottedmeat).

#### Testing Instructions

None should be required. You can see that this solves the issue in https://github.com/Automattic/wp-calypso/pull/72835. I guess just verify that I'm not wrong about this return type.

### DIFF
--- a/packages/data-stores/src/mapped-types.ts
+++ b/packages/data-stores/src/mapped-types.ts
@@ -40,7 +40,7 @@ export type DispatchFromMap< A extends Record< string, ( ...args: any[] ) => any
 		...args: Parameters< A[ actionCreator ] >
 	) => A[ actionCreator ] extends ( ...args: any[] ) => Generator
 		? Promise< GeneratorReturnType< A[ actionCreator ] > >
-		: void;
+		: Promise< void >;
 };
 
 /**
--- a/client/landing/gutenboarding/onboarding-block/create-site-error/index.tsx
+++ b/client/landing/gutenboarding/onboarding-block/create-site-error/index.tsx
@@ -20,7 +20,7 @@ const CreateSiteError: FunctionComponent< Props > = ( { linkTo } ) => {
 
 	const { resetNewSiteFailed } = useDispatch( SITE_STORE );
 
-	useEffect( () => resetNewSiteFailed );
+	useEffect( () => void resetNewSiteFailed() );
 
 	return (
 		<div className="create-site-error">


sirbrillig:
Note: This changes some code in `CreateSiteError` added in https://github.com/Automattic/wp-calypso/pull/43092 where the line 

```js
useEffect( () => resetNewSiteFailed );
```

was added... which I think will do nothing??, but I'm _assuming_ that was a bug and it was meant to be 

```js
useEffect( () => void resetNewSiteFailed() );
``` 

which would run the action on every render. Or maybe even 

```js
useEffect( () => void resetNewSiteFailed(), [] );
```

which would run only once on mount? I opted for the first one but I'm not 100% sure what the intent was. cc @lsl

p-jackson:
@sirbrillig the `/landing/gutenboarding` code is no longer used so we can't really test the change. The server redirects all traffic to other signup flows: https://github.com/Automattic/wp-calypso/blob/4d22352b52290c917233b713eea0bed02677ac0a/client/server/pages/index.js#L882

`useEffect( () => resetNewSiteFailed );` does look like a bug to me. My instinct is that we only wanted to dispatch the action on mount (`resetNewSiteFailed` clears the error state flag). But it looks like that'd cause a bug too. The thing that causes `<CreateSiteError>` to be mounted is the value of the error state flag https://github.com/Automattic/wp-calypso/blob/4d22352b52290c917233b713eea0bed02677ac0a/client/landing/gutenboarding/onboarding-block/edit.tsx#L151. So as soon as the component mounts it will immediately unmount.

I can't verify whether this is true of course 😄 

I think the best thing to do for this PR is just to do whatever it takes to silence the TS error and move on. I don't know what the plan with the gutenboarding code is. @alshakero now that the stepper is in place would Vertex remove this?

packages/data-stores/src/mapped-types.ts
```diff
@@ -40,7 +40,7 @@ export type DispatchFromMap< A extends Record< string, ( ...args: any[] ) => any
 		...args: Parameters< A[ actionCreator ] >
 	) => A[ actionCreator ] extends ( ...args: any[] ) => Generator
 		? Promise< GeneratorReturnType< A[ actionCreator ] > >
-		: void;
+		: Promise< void >;
```
p-jackson, line 43: This looks good to me

p-jackson (approved):

tyxla (approved):
Thanks @sirbrillig 👍

PR #72844, opened 2023-02-01
Stats: Update Videos summary page

grzegorz-cp:
#### Proposed Changes

* Add a feature flag and modernise the `Video` summary details page
* Extend the Horizontal bar component and its card component to:
   * render additional value columns by accepting a new parameter
   * add an option for an alternate header displaying column names below the card's title
   * update the width of the value column to accommodate larger numbers (a fixed width is needed for the label ellipsis)
   * add an option for a flexible name of the main item column (value column name is already flexible)
   * hide displaying data is loading property is passed (unlike other components, Videos summary page already have partial data when navigating to a page but still loads more)

I plan to update click events for each number in a separate PR.

#### Testing Instructions

* Navigate to the main Stats page on the live branch
* append `?flags=stats/new-video-summary` to the URL
* visit `Videos` summary page by clicking `View all` on the component on the `Traffic` page
* make sure that the component loads properly
* Bonus: smoke test other horizontal bar component on different screen sizes

| Before | After |
| --- | --- |
| SCR-20230201-hkp | SCR-20230201-hgq |

#### Pre-merge Checklist

Related to #71018


client/my-sites/stats/videopress-stats-module/index.jsx
```diff
@@ -151,93 +155,171 @@ class VideoPressStatsModule extends Component {
...
+							additionalColumns={ {
+								header: (
+									<>
+										<span>{ translate( 'Impressions' ) }</span>
+										<span>{ translate( 'Hours watched' ) }</span>
+									</>
+								),
+								body: ( item ) => (
+									<>
+										{ /* eslint-disable-next-line jsx-a11y/no-static-element-interactions */ }
```
grzegorz-cp, line 214: This will be addressed in a separate PR.

packages/components/src/horizontal-bar-list/_stats-variables.scss
```diff
@@ -17,3 +17,16 @@ $header-font-size: 20px;
...
+@mixin additional-columns-wrapper() {
+	display: flex;
+	text-align: right;
+	gap: 0 4px;
+	padding: 0 4px;
+	font-weight: 400;
+
+	> div,
+	> span {
+		min-width: 120px;
```
grzegorz-cp, line 30: I'm not sure how to avoid adding this value due to the need of aligning all columns and maintaining ellipsis for long titles. If showing more columns (with this component) will be used in any other place this will have to be moved to inline styles.

client/my-sites/stats/videopress-stats-module/index.jsx
```diff
@@ -151,93 +155,171 @@ class VideoPressStatsModule extends Component {
 			page( url );
 		};
 
+		const isNewVideoPage = config.isEnabled( 'stats/new-video-summary' );
+
+		const downloadCSV = (
+			<div className="stats-module__heaver-nav-button">
```
grzegorz-cp, line 161: I need to fix the breakpoint for this to only go under the title for small screens.

packages/components/src/horizontal-bar-list/stats-card.tsx
```diff
@@ -14,27 +14,58 @@ const StatsCard = ( {
...
 	heroElement,
+	splitHeader,
+	metricLabel,
+	mainItemLabel,
+	additionalHeaderColumns,
 }: StatsCardProps ) => {
 	const translate = useTranslate();
 
+	const titleNode = titleURL ? (
+		<a href={ `${ titleURL }` } className={ `${ BASE_CLASS_NAME }--header--title` }>
```
kangzj: Should we follow the Calypso CSS naming convention to name this `${ BASE_CLASS_NAME }__header-title`? And also for a few other places?

REF: 
- https://github.com/Automattic/wp-calypso/blob/trunk/docs/coding-guidelines/css.md
- https://getbem.com/naming/

kangzj (changes requested):
1. I did notice the views would overlap with other column heads now for smaller screens. Also should we spare more space to title column?

image

image

2. I would like the CSS class names to following the Calypso/BEM convention. 🙂

3. I would prefer not showing the columns when there's no data. What do you think?
image

packages/components/src/horizontal-bar-list/stats-card.tsx
kangzj: Oh, actually the titleNode could be a block too. In that case, the name would be something like:

`${ BASE_CLASS_NAME }-header__title`

client/my-sites/stats/videopress-stats-module/index.jsx
grzegorz-cp: Actually, I won't be fixing this because the period selection will replace the button like on the other summary pages.

packages/components/src/horizontal-bar-list/stats-card.tsx
grzegorz-cp: Thanks for noticing this.

jsnmoon (approved):
The overall approach appears solid. I agree with the changes Jasper suggested, but it would get good to ship this first and continue iterating behind a feature flag.

This change would benefit from some design iterations, specifically:

- What would be the best way to present the information in a mobile viewport? Perhaps it'd be better to hide the `Impressions` and `Hours Watched` columns (and figure out ways to present them somehow in a more mobile-friendly manner without clipping the title):

image

- How would it feel to extend the 100% horizontal bar all the way to the final value?

image

cc @jeffgolenski and @ederrengifo for feedback.

grzegorz-cp:
Thanks for the comment @jsnmoon. I was considering adding a scroll to the table below 480px but this would involve adding an arbitrary size which is not great and would differ from table to table if we have more examples of multiple columns.

Hiding the additional columns for a small screen is quite clever. Another solution would involve a pattern of switching from a table to some sort of list view, but this would require a new component and can get complex.

kangzj (approved):
Looks good to me. The styles could be iterated in following up PRs since it's behind a feature flag 👍

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7789899

Thank you @grzegorz-cp for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7789899).

jeffgolenski:
@grzegorz-cp @jsnmoon 

> I was considering adding a scroll to the table below 480px but this would involve adding an arbitrary size which is not great and would differ from table to table if we have more examples of multiple columns.

I think for now we could ship that idea because it's a pattern we're using elsewhere. Then we can iterate on different cards in future projects. Up to you!

PR #72856, opened 2023-02-01
Stats: Better localized formatting in highlights

### LINKED ISSUE #72492: Stats: Investigate WordAds currency presentation
Per [this comment](https://github.com/Automattic/wp-calypso/pull/72320#discussion_r1083575318), we should confirm amounts on the Stats → Ads page are being handled correctly.

The current behaviour is to always display USD amounts with US formatting and a "$" symbol. I think this makes sense for WordAds if it's a USD-only program but it's worth confirming.

1. Is WordAds a USD-only program?
2. If so, does it make sense to localize the numbers?
3. If not, we need to update it to show localized amounts.

If we need to update the behaviour, the highlights (top row) and the tables should remain consistent.

**Update** — Currently we do localize the amounts. For example, on my CAD account, if switched to French, I see USD amounts as $0,00 instead of $0.00. When switching back to English, there's no indication that the amounts are USD as opposed to CAD either. Potentially confusing?

a8ck3n:
Watch for and handle special case before formatting as a string.

#### Proposed Changes

Tweak to special casing how zero amounts are handled so it works as expected in languages that don't use a period as the decimal separator.

**Before:**

SCR-20230201-phh

**After:**

SCR-20230201-pik

#### Testing Instructions

* Open Calypso Live link.
* Visit a site with Ads enabled.
* Visit Stats → Ads.
* Confirm highlight cards with zero amount show as "$0".
* Change language (try German) and reload the page.
* Confirm amounts are still shown as "$0".

#### Pre-merge Checklist

Related to #72492.

### DIFF
--- a/client/my-sites/stats/wordads/highlights-section.jsx
+++ b/client/my-sites/stats/wordads/highlights-section.jsx
@@ -11,14 +11,15 @@ import HighlightCardSimple from './highlight-card-simple';
 // Then refactor this Comp to use HighlightCard again.
 
 function getAmountAsFormattedString( amount ) {
+	if ( amount === 0 ) {
+		// Per design spec we don't want "$0.00" as a result.
+		// https://github.com/Automattic/wp-calypso/issues/72045
+		// We'll return "$0" in this scenario.
+		return '$0';
+	}
 	// Takes a Number, formats it to 2 decimal places, and prepends a "$".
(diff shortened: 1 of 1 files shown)


kangzj (approved):
LGTM

jsnmoon (approved):
Agreed with Jasper. Looks good!

PR #72867, opened 2023-02-01
Fix premium domain not rendering the correct price when not returned from the suggestion endpoint

hambai:
When we show domain suggestions for FQDNs we actually merge and render the result of two separate API calls - the FQDN availability check and the suggestion results. In some cases we might not get the FQDN in the suggestion results even though it is available domain. When this domain is premium domain we're in trouble because the availability properties are different than the suggestion properties. This is a quick fix for that - let's just set `is_premium` to true if there is `is_supported_premium_domain` property in the results. I'll also check why we don't get the domain entered from the suggestion endpoint.

#### Proposed Changes

* Set `is_premium` to true in the suggestion results if the `is_supported_premium_domain` is true

#### Testing Instructions

* I've noticed this in sandboxed mode where searching for a premium .link domain doesn't return the domain from the suggestion endpoint.

#### Pre-merge Checklist

### DIFF
--- a/client/components/domains/register-domain-step/index.jsx
+++ b/client/components/domains/register-domain-step/index.jsx
@@ -921,6 +921,14 @@ class RegisterDomainStep extends Component {
 						AVAILABLE_PREMIUM === status &&
 						result?.is_supported_premium_domain;
 
+					/**
+					 * In rare cases we don't get the FQDN as suggestion from the suggestion engine but only
+					 * from the availability endpoint. Let's make sure the `is_premium` flag is set.
+					 */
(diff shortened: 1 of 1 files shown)


leonardost (approved):
I tested this together with D99961-code and LGTM! I tested with the `nutrient.link` domain but it seems to be returned by the suggestions endpoint. These are the results I got without this PR and with:

- WIthout PR:

Screen Shot 2023-02-01 at 13 08 06

- With PR:

Screen Shot 2023-02-01 at 13 08 00

With the PR the domain shows correctly as a premium domain, so I think that's correct, right?

PR #72712, opened 2023-01-27
fix: Disable copy site for sites that are not atomic

kozer:
#### Proposed Changes

Don't allow copy site when a site has copy site feature but it's not atomic.

#### Testing Instructions

1. Create a site with a `Business` plan
2. Make sure that it's not atomic.
3. Create another site with a `Business` plan
4. Make sure that the site is atomic.

##### Test copy site from sites dashboard
1. Navigate to `/sites`
2.  Make sure that the newly created site (**not atomic**) _can't_ be copied (no menu item appears)
3. Make sure that the newly created site (**atomic**) _can_ be copied (menu item appears)

##### Test copy site from site tools
1. Navigate to Settings > General > Site Tools (**on your not atomic site**)
2. Make sure that "Copy Site" option _doesn't_ appear.
3. Navigate to Settings > General > Site Tools (**on your atomic site**)
4. Make sure that "Copy Site" option _appears_.

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

fixes: https://github.com/Automattic/wp-calypso/issues/72707

PR #72868, opened 2023-02-01
Copy Site: Prefetch data for site actions

### LINKED ISSUE #72275: Sites Management Page: Pre-fetch data for Site Actions
It's a little annoying to have 'Copy Site' and 'Manage Plugins' pop in after a delay

https://user-images.githubusercontent.com/36432/213200432-30da926f-3b76-4cfa-800f-c9536fe2759e.mp4

We could avoid it if we pre-fetched the necessary data. We could do this if the user hovers over the site row, or if the site is in the viewport.

From p1674052767199949-slack-C0347E545HR

**Done is:**
* 'Copy Site' and 'Manage Plugins' appear instantly when applicable, instead of after a delay.

kozer:
#### Proposed Changes

'Copy Site' and 'Manage Plugins' appear instantly when applicable, instead of after a delay. (Prefetch data)

NOTE:

Has changes that already esists in https://github.com/Automattic/wp-calypso/pull/72291.

I'll make sure that I'll update this pr when this is merged.

#### Testing Instructions

1. Open the dev console.
2. Make sure you have a  site with `Business` plan and atomic.
3. Navigate to `/sites`
4. Ensure that the requests you see, exist:
        - Only one request to `me/purchases`
        - `/sites/<site id>/features` requests only for the sites you see on your viewport.
5. Make sure that you find the site in your list. Click on the three dots. `Copy Site` menu and `Manage Plugin` menu should be there.
prefetch-sites-menu

#### Pre-merge Checklist

Fixes https://github.com/Automattic/wp-calypso/issues/72275


danielbachhuber:
@kozer Can you make sure the unit tests are passing here?

kozer:
> @kozer Can you make sure the unit tests are passing here?

Surely!

Update:
@danielbachhuber , done!

client/sites-dashboard/components/sites-ellipsis-menu.tsx
```diff
@@ -284,7 +268,9 @@ export const SitesEllipsisMenu = ( {
 					<ManagePluginsItem { ...props } />
 					{ showHosting && <HostingConfigItem { ...props } /> }
 					{ site.is_coming_soon && <PreviewSiteModalItem { ...props } /> }
-					{ isEnabled( 'sites/copy-site' ) && <CopySiteItem { ...props } /> }
+					{ isEnabled( 'sites/copy-site' ) && shouldShowSiteCopyItem && (
+						<CopySiteItem recordTracks={ startSiteCopy } site={ site } />
```
p-jackson: I wasn't expecting the trick where `startSiteCopy` can be used in place of a tracks recording function 😅 

I think it might be clearer to add an `onClick` callback to `<CopySiteItem>`. The component would record the track event and then call any `onClick` prop it's been passed.

client/components/data/query-user-purchases/index.jsx
```diff
@@ -29,4 +29,16 @@ function QueryUserPurchases() {
...
+export const useQueryUserPurchases = () => {
+	const userId = useSelector( ( state ) => getCurrentUserId( state ) );
+	const isRequesting = useSelector( ( state ) => state.purchases.isFetchingUserPurchases );
+	const hasLoaded = useSelector( ( state ) => state.purchases.hasLoadedUserPurchasesFromServer );
+	const reduxDispatch = useDispatch();
+
+	if ( ! userId || isRequesting || hasLoaded ) {
+		return;
+	}
+	reduxDispatch( fetchUserPurchases( userId ) );
```
p-jackson: This doesn't seem right, I think you'd want to have a `useEffect` hook and then within that `useEffect` callback you'd dispatch the redux action. The idea is that you want to dispatch the action as a "side effect" of the component being mounted, not as part of the main render path itself.

Take a look at how `useQuerySitePurchases` is implemented for an example of how it can be done.

client/sites-dashboard/components/sites-ellipsis-menu.tsx
```diff
@@ -204,8 +189,6 @@ const CopySiteItem = ( { recordTracks, site }: SitesMenuItemProps ) => {
 		<MenuItemLink
 			href={ copySiteHref }
 			onClick={ () => {
-				clearSignupDestinationCookie();
-				setPlanCartItem( { product_slug: plan.product_slug } );
 				recordTracks( 'calypso_sites_dashboard_site_action_copy_site_click' );
```
p-jackson, line 194: I actually like how this was before. It feels a little strange pulling code from the `/client/landing/stepper` area of the code into the site's dashboard. And as far as I can tell there's nowhere else using the `startSiteCopy` that means it needs to be shared.

p-jackson (commented):
Much faster! All the requests are being made as expected and by the time I click the ellipsis menu the Copy Site link is already there 👍 

I wouldn't have hidden the ellipses menu until it was in view. I'm not sure, but I think removing controls from the DOM until they're in view can't be as good from an accessibility point of view. Assistive technology won't be able to analyse what interactive things are in the DOM.
You could achieve the same thing by moving the `useInView` code into the `<SitesEllipsisMenu>` and then passing a `{ disable: true }` options object to the `useSiteCopy` hook to prevent it from making any requests. This is the pattern used by the ReactQuery hooks (see `useSiteExcerptsQuery`) to declaratively turn network requests on and off and I think it's a really good pattern for disabling queries for whatever reason (like being out of view).
This method does work too though, and maybe it's just a hunch I have about the a11y issue.

client/sites-dashboard/components/sites-ellipsis-menu.tsx
kozer: We did it like this to avoid writing the same code twice. This code is used exactly as is to the `Settings > General > Site Tools`. And this is the reason we refactor it to the hook in the first place.

client/components/data/query-user-purchases/index.jsx
kozer: I had a difficult time with this one. Initially, I wanted to do it just like `useQuerySitePurchases`, and I tried to do like so, but:

This function is triggered many times, as it's inside the `use-site-copy` hook. (it does for each row item).

I was expecting that having a `useEffect` will actually prevent the dispatch of `fetchUserPurchaces` again, after `PURCHASES_USER_FETCH` action.

However this is not the case, and if I put this code inside a `useEffect`, the request is triggered for each row item, no matter what.

Moving this code:

```
if ( isRequesting || hasLoaded ) {
		return;
	}
```
inside `fetchUserPurchases`, solves the issue, but, I wasn't sure if we want to put that logic there.

Having said that, and because of the things I mentioned, I'm not sure that `useQuerySitePurchases` works as expected tbh.

Also, I think this works without any `useEffect` because this is a hook, and it includes state properties (using `useSelector`). This means that whether there are updates to those parts of the state, this function will run. Given that, having that if statement in place, is enough to avoid refetching the data without the need of a `useEffect`

client/sites-dashboard/components/sites-ellipsis-menu.tsx
kozer: I'll revisit it on Monday. Thanks for your feedback!

client/components/data/query-user-purchases/index.jsx
p-jackson: If the checks for whether the purchases data is inside the effect too, then wouldn't this prevent the repeated queries?

```ts
useEffect( () => {
	if ( ! userId || isRequesting || hasLoaded ) {
		return;
	}
	reduxDispatch( fetchUserPurchases( userId ) );
}, [ userId, isRequesting, hasLoaded, reduxDispatch ] );
```

The reason we want to put the fetch in an effect hook is that it moves it out of the high-priority part of the React lifecycle and into the low-priority part.

CleanShot 2023-02-07 at 15 28 26@2x

[[source]](https://github.com/donavon/hook-flow)

Currently `fetchUserPurchases` is being called in the green part, but we want it at the end in the yellow part.

kozer: First of all, thanks for the info!!! REALLY handy!
I was about to respond that I had tried it but it did not work. However, I thought, let's try that again to confirm that, as I had already tried it!
But miraculously it worked!! So I pushed that change! Thanks!

kozer:
> You could achieve the same thing by moving the useInView code into the <SitesEllipsisMenu> and then passing a { disable: true } options object to the useSiteCopy hook to prevent it from making any requests. This is the pattern used by the ReactQuery hooks (see useSiteExcerptsQuery) to declaratively turn network requests on and off and I think it's a really good pattern for disabling queries for whatever reason (like being out of view).
This method does work too though, and maybe it's just a hunch I have about the a11y issue.

@p-jackson 
I updated the hook as you proposed, however, I had to pass the disabled property, down to the other hooks as well. Do you have any pattern in mind to prevent that? It's not a big deal, however I kinda don't like it.

client/sites-dashboard/components/sites-ellipsis-menu.tsx
```diff
@@ -196,7 +193,9 @@ const CopySiteItem = ( { recordTracks, site }: SitesMenuItemProps ) => {
 			href={ copySiteHref }
 			onClick={ () => {
 				recordTracks( 'calypso_sites_dashboard_site_action_copy_site_click' );
-				startSiteCopy();
+				if ( onClick ) {
+					onClick();
+				}
```
sejas: We can simplify the code like this:
```suggestion
				onClick?.();
```

sejas (commented):
@kozer, I've tested it and it doesn't work for me on sites that were out of the view.
Steps to reproduce:
- Go to `/site`
- Scroll the page to display hidden rows.
- Click in the three dots on a business site.
- Observe there are no `Copy Site` item in the menu.

I wonder if we should optimize the `useSiteCopy` moving the logic to a new API endpoint. So all the logic in the frontend would be reduced to only one request to the network. (Initially suggested here https://github.com/Automattic/wp-calypso/pull/72712#issuecomment-1407170444)


https://user-images.githubusercontent.com/779993/217232909-9b11c7d3-6c8b-4471-9aef-f01bd16daf5a.mp4

kozer:
> @kozer, I've tested it and it doesn't work for me on sites that were out of the view.

It should be fixed now

client/components/data/query-user-purchases/index.jsx
```diff
@@ -29,4 +29,18 @@ function QueryUserPurchases() {
 	return null;
 }
 
+export const useQueryUserPurchases = ( disabled = false ) => {
```
danielbachhuber: This `disabled` flag feels weird. Shouldn't "whether or not to query the data" be the responsibility of the calling component?

client/landing/stepper/declarative-flow/copy-site.tsx
```diff
@@ -145,12 +142,6 @@ const copySite: Flow = {
 				}
 
 				case 'processing-copy': {
-					const processingResult = params[ 0 ] as ProcessingResult;
-
-					if ( processingResult === ProcessingResult.FAILURE ) {
-						// @TODO:Create a retry step
-						return navigate( 'retry' );
-					}
```
danielbachhuber: Why is this being removed?

kozer: This was added initially as a first step to implement a retry step. It's not actually functional. However, we took another approach on resuming implementation, so this code is no longer necessary

client/components/data/query-user-purchases/index.jsx
kozer: Unfortunately, we can't. Hooks cannot be conditionally rendered. I don't know if there is a pattern that I'm not aware of.
I'm open to suggestions!

danielbachhuber: @p-jackson Any great suggestions here?

sejas: I think passing the disabled parameter is a common practice among hooks. The component can avoid making the requests using that parameter.

sejas (approved):
Tested and working as described.
We increase the number of requests on load and on scroll. The `Copy site` item is already there when we open the dropdown.

client/components/data/query-user-purchases/index.jsx
danielbachhuber: Such hack. So wow.

p-jackson: Yeah this is one of the things about hooks; things that you're used to treating as imperative (`if condition is met then query purchases`) have become declarative (`declare a purchase query with property of condition being met`).
This is how ReactQuery does it. [It's often used for waiting for all endpoint params to be available](https://github.com/Automattic/wp-calypso/blob/dd409ed59f831e4a0a1bb59cc8f3a8d3c7703e2a/client/data/site-roles/use-site-roles-query.js#L12).

If you were inside a component you'd be able to do something like `isCondtion && <QueryUserPurchases />`, but since you want to be able to do this inside the `useCopySite()` mega-hook you've got to do it with a hook.

I know I originally suggested the `disabled` flag, but if you wanted to make it look more familiar you could copy the ReactQuery API by flipping it to an `enabled` flag and passing it as part of an options object. That way devs will see it as just another hook with "Query" in the name, and they'll be like: oh yeah I know how this works.

kozer: Thanks @p-jackson for your suggestion. I just changed `disable` to `enable` to match ReactQuery API.

wojtekn:
I've tested it, and it works much better now.

wojtekn (approved):

client/landing/stepper/hooks/use-site-copy.tsx
```diff
@@ -67,22 +79,28 @@ export const useSiteCopy = (
...
-				[ 'marketplace_plugin', 'marketplace_theme' ].includes( purchase.productType )
+		const marketplacePluginProducts = ( purchases || [] )
+			.filter(
+				( purchase ) =>
+					[ 'marketplace_plugin', 'marketplace_theme' ].includes( purchase.productType ) &&
+					purchase.siteId === site?.ID
+			)
+			.filter(
+				( purchase ) =>
+					purchase.productType === 'marketplace_plugin' && purchase.siteId === site?.ID
```
sejas, line 90: @kozer, I caught this now. Do we need to filter twice? I think the second filter is "overwriting" the first one.
```diff
			.filter(
				( purchase ) =>
					[ 'marketplace_plugin', 'marketplace_theme' ].includes( purchase.productType ) &&
					purchase.siteId === site?.ID
			)
-			.filter(
-				( purchase ) =>
-					purchase.productType === 'marketplace_plugin' && purchase.siteId === site?.ID
```

kozer: Actually, it's not, it is applied after the first one, but it's unnecessary of course.

kozer: Since it's merged, what do you think it's better to tackle this?
1. Creating a quick fix
2. Tacke it to your branch, as you heavily changed the hook?

sejas: Do you know which one is the correct one?
I think `[ 'marketplace_plugin', 'marketplace_theme' ]`  is what we need to use, can you confirm it?

kozer: Oh no! You are right @sejas ! it's overwritten!
Let's apply a quick fix

kozer: Here is the new pr that fixes that https://github.com/Automattic/wp-calypso/pull/73154

sejas: Thanks for the quick fix!

PR #72875, opened 2023-02-01
Themes: Add new an action to fetch a specific theme

valterlorran:
#### Proposed Changes

* Adds a new action to fetch a specific theme.  This will be used on #72837 to determine if the theme is ready to be activated after the site is transferred to atomic.

#### Testing Instructions

* This action is not being used yet. I added some unit tests that can be run with the following command:
```bash
node 'node_modules/.bin/jest' './client/state/themes/test/actions.js' -c './test/client/jest.config.js'
```

#### Pre-merge Checklist

Related to #72837


markbiek (approved):
LGTM!

PR #71912, opened 2023-01-11
Plans 2023 : Refactor and typescriptify code

ddc22:
#### Proposed Changes

* Feature strings used by the plans details grid have to be used in the plans comparison grid.
* This PR tries to refactor the code as well as introduce typescript so that the code is more manageable
* Also general cleanup of typescript errors are also performed. 

image

#### Testing Instructions

* Go to `/start/onboarding-2023-pricing-grid/plans`
* Make sure the existing grid works as the shipped code 
* Make sure the title for the comparison grid is showing

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #

PR #72878, opened 2023-02-01
Plans: Update getPlans selector to clarify it can return undefined

sirbrillig:
#### Proposed Changes

In https://github.com/Automattic/wp-calypso/pull/72032 (which was merged to trunk as part of https://github.com/Automattic/wp-calypso/pull/71912), I changed the `getPlan()` selector to switch from using lodash's `find` to native JS `find`. This seemed safe because it appeared that the `getPlans()` selector (on which `getPlan` is based) always returned an array. However, recent fatal errors have shown that this is not always the case. Presumably lodash `find()` was hiding this fact.

In this PR we clarify that `getPlans` can return undefined and use optional chaining in `getPlan` to prevent fatals.

#### Testing Instructions

TypeScript should cover these changes.

### DIFF
--- a/client/state/plans/selectors/plan.ts
+++ b/client/state/plans/selectors/plan.ts
@@ -8,7 +8,7 @@ import 'calypso/state/plans/init';
 /**
  * Return WordPress plans getting from state object
  */
-export const getPlans = ( state: AppState ): PricedAPIPlan[] => {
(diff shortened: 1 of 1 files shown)


pottedmeat (approved):
Fix is straightforward and looks good.

Issue #72021, opened 2023-01-12
LaunchPad - Celebration animation


Addison-Stavlo:
We should create some visuals to help celebrate important moments in our flows on dotcom, such as Launching a site!

One idea is a celebration / confetti / "popping the cork"  type animation that can be played on top of the screen, maybe with floating text that says "Congratulations!" or something similar.  The idea/design is vague, but this could really help the feeling users get by going through our flows.

Addison-Stavlo:
note - we may be able to reuse similar recent work by other teams ✨   - https://github.com/Automattic/wp-calypso/pull/71847#issuecomment-1378881522

Addison-Stavlo:
more design context here - p9Jlb4-6dJ-p2

Addison-Stavlo:
closing this with - https://github.com/Automattic/wp-calypso/pull/72880

PR #72880, opened 2023-02-01
Launchpad: Launch Celebration Modal

sharpirate:
### Time Estimate
Review: Mid
Test: Short

### Proposed Changes
This PR adds a new modal celebrating launching a site

#### Dev Notes
I am currently using the `CelebrateSiteLaunch` component to render the modal which I suspect it's not ideal. I would love some suggestions on whether we should make the modal its own component and ideas on how to hook into the site launch and know when to show the modal.
image

### Testing
1. Checkout this branch
2. Go to http://calypso.localhost:3000/setup/free/intro or any other Launchpad flow
3. Create & Launch a site
4. Once you are redirected to My Home you will see the modal pop-up. Verify the `?celebrateLaunch=true` is appended to the URL.
5. Make sure it looks good according to the mockup (you can find it in the thread linked below)
6. As soon as the modal opens verify the `?celebrateLaunch=true` is removed from the URL without the page being reloaded
7. Visit Site button should redirect you to the site
8. Customize domain should redirect you to the domain screen. When you go back you should go to my home without seeing the modal again (no `?celebrateLaunch=true` in the URL)
9. Clicking the close button, pressing Escape or clicking outside the modal should close the modal

https://user-images.githubusercontent.com/20927667/216691284-c041de5f-af18-4ccc-ac59-0d7b55e87c5b.mov

### Project Thread
Related to #72021 
p9Jlb4-6dJ-p2


sharpirate:
### Notes
I wanted to start a discussion about what we'll consider a good first iteration of the modal since there are a lot of things going on in the P2 - there were concerns about having the modal take up the fullscreen, suggestions to modify the current confetti animation to use WP colors or the user's palette, having a "don't show again" checkbox on the modal etc.

I think we can definitely iterate over these things but wanted to get a clear picture of what we can ship as v1. Any input here is more than welcome! @nuriapenya and others

nuriapenya:
>I think we can definitely iterate over these things but wanted to get a clear picture of what we can ship as v1.

💯 
This is already a much better experience! I'd say to ship it as it is, monitor (even test with some users), and iterate. 

The only thing is that if you click "Choose domain" link and then hit the back button, the modal & confetti are triggered again. Is this an expected behaviour? Or should we just simply bring users back to My Home?

https://user-images.githubusercontent.com/7931304/216922966-a81b4635-aab8-4d54-96fe-f641792df861.mov

Addison-Stavlo (approved):
This is awesome!

The code looks good and this tests well on my end.

edanzer:
Tests nicely here! There seems to be a small bit of jumpiness in the animation/transition, but it's subtle, and if its a concern, it's definitely something we could address in a follow up.

client/my-sites/customer-home/cards/notices/celebrate-site-launch/index.jsx
```diff
@@ -15,6 +15,7 @@ import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 const CelebrateSiteLaunch = ( { isSiteSetupComplete, pendingSiteSetupTasks, siteId } ) => {
 	const translate = useTranslate();
 	const dispatch = useDispatch();
+
```
edanzer: Seems like we added an empty line here? No harm, but not sure if that's intentional.

client/my-sites/customer-home/celebrate-launch-modal.jsx
```diff
@@ -0,0 +1,53 @@
+import { ConfettiAnimation, Gridicon } from '@automattic/components';
```
edanzer: I see we're just adding the component file and associate css directly to the customer-home folder. Probably not a requirement, but I was wondering if it be cleaner to put these inside a components folder, or otherwise separate them out of the root of customer home?

sharpirate: `components` folder is definitely a good idea, I'm on it

client/my-sites/customer-home/main.jsx
```diff
@@ -48,6 +49,10 @@ const Home = ( {
 	sitePlan,
 	isNew7DUser,
 } ) => {
+	const [ celebrateLaunchModalIsOpen, setCelebrateLaunchModalIsOpen ] = useState(
+		window.location.href.includes( 'celebrateLaunch=true' )
```
edanzer, line 53: Not a big deal, but seems a bit unusual that we're just checking if the entire celebrateLaunch=true string exists within the url. Should work and can't see an obvious problem. But I wonder if it would be better to use something that actually parses out the query params. Options might include `useQuery` hook within stepper, `useLocation` hook (from react-router-dom, which useQuery uses), something like `new URLSearchParams(location.search)` or something else. 

Not sure if it's worth making this change though.

edanzer (approved):

client/my-sites/customer-home/main.jsx
Addison-Stavlo: Good call out. I passed over this due to seeing it a couple times recently, but added similar comments on this PR from last sprint - https://github.com/Automattic/wp-calypso/pull/72930#discussion_r1095052252 (and explanation in comments below that) - so i have similar thoughts. Non-blocking but good call out.

edanzer:
Just re-tested after latest changes. Still testing well here.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7816061

Thank you @Sharpirate for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7816061).

ebinnion:
Could we add tracks events to this modal for the actions the user could take? I was attempting to do a bit of analysis to track down an increase in domain purchases and it would be helpful to know how users are interacting with this.

Issue #72870, opened 2023-02-01
LaunchPad -  General onboarding - give "skip to dashboard" from goals page a launchpad


Addison-Stavlo:
we recently added the launchpad to general onboarding for `build` and `write` flows.

This however does not show up if we "skip to dashboard" from the goals page:
![Screen Shot 2023-02-01 at 8 38 12 AM](https://user-images.githubusercontent.com/28742426/216058465-8e0089f5-c8fc-4368-ba16-c6ce52451fbe.png)

Upon inspection this does not set a site intent, however it would make sense to set this to `build` as we do in the "Other" option, and enable the launchpad for this flow.

edanzer:
Closing this as complete based on https://github.com/Automattic/wp-calypso/pull/74082

Issue #72872, opened 2023-02-01
LaunchPad - general onboarding - can land in my home with no launchpad.


Addison-Stavlo:
There seem to be some cases where we land in my home but are not redirected to the launchpad. This may be related to leaving the flow early with a "skip to dashboard" type link. One example to reproduce:

* create a new site
* get to the "goals"
* select "other"
* at the first page inside the "other" flow, select "skip to dashboard"
* notice we land in "my home"
* reload the browser, notice a redirect to the launchpad.

![no-launchpad-at-first](https://user-images.githubusercontent.com/28742426/216059861-e4d1d21e-fa2c-4ec8-be1a-150dd4d36525.gif)

Addison-Stavlo:
should be resolved now with https://github.com/Automattic/wp-calypso/pull/72930

i can no longer repro in production ✨

PR #72883, opened 2023-02-01
Update signup goals page skip to dashboard set site_intent to build

agrullon95:
#### Testing/Review instructions

Test: short
Review: short

#### Proposed Changes

Currently, users that click the `Skip to dashboard` link on the top right-hand corner of the Goals onboarding step will be navigated to the home screen with the following site options:

`site_intent` = ""
`launchpad_screen` = false

The changes in this PR will set the `site_intent` site option to `build` if the user clicks `Skip to dashboard` on the Goals page. This will also enable the launchpad and set the `launchpad_screen` site option to `full` before the user is navigated to the home screen.

P2 post to give heads up: paYE8P-2yP-p2

Goals step:
image

#### Testing Instructions

* Checkout this branch
* Navigate to the onboarding flow (Click `< Switch Site` on the Calypso sidebar and then click `Add a new site`)
* Create your free site and you should land on the Goals step.
* Do not select a goal and click `Skip to dashboard`. The `site_intent` and `launchpad_screen` values should be updated. The home screen should redirect you to the launchpad screen (There may be a bug that requires multiple refreshes before you redirected - )
* Create a new site and test another goal such as `Write & Publish` to verify the correct `site_intent` value is set

#### Pre-merge Checklist

Related to #72870


Sharpirate:
### Testing
Go to `http://calypso.localhost:3000/start`

sharpirate:
### Testing
1. Go to `http://calypso.localhost:3000/start`
2. On the Goals page click on "Skip to dashboard"
3. On My Home refresh we are redirected to Launchpad which indicates the site options were updated successfully.
5. Tested with the `Write & Publish` flow continues to work when selecting Goals.

sharpirate (approved):

jeyip:
## Testing
### Requirements
- [x] Do not select a goal and click Skip to dashboard. The site_intent and launchpad_screen values should be updated. The home screen should redirect you to the launchpad screen
- [x] Create a new site and test another goal such as Write & Publish to verify the correct site_intent value is set
- [x] Could complete Launchpad tasks and launch site as expected
- [x] Launchpad iframe overlay, url copy to clipboard button, customize pill, and go to admin button continued to work as expected 

### Browsers
- [x] Chrome

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -164,12 +164,17 @@ const siteSetupFlow: Flow = {
 					if ( ! siteSlug ) {
 						return;
 					}
+
+					// If user clicks "Skip to Dashboard" on goals step,
+					// Default site_intent to "build"
+					const updatedSiteIntent = intent ? intent : SiteIntent.Build;
```
Addison-Stavlo: Note this isnt specifically tied to this dashboard button on the goals step, so im curious if there are any other areas this change could potentially bleed into?

That said, testing around and escaping out of other flows early I see the intent the same as expected. So it seems to be looking good in my testing.

Addison-Stavlo (approved):
This tests well on my end, minus the other issue tracked elsewhere that shouldnt block this PR. https://github.com/Automattic/wp-calypso/issues/72872

client/landing/stepper/declarative-flow/site-setup-flow.ts
jeyip: Since we're setting a default value for _any_ flow with some falsey value for `intent`, would we inadvertently enable launchpad for situations where it's undesired?

I can't think of anything off the top of my head, but I suppose, if we wanted to be safer, we could only set default values for general onboarding

jeyip (commented):
I left one comment. +1 from me when we decide how we want to approach the concern

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -520,6 +523,7 @@ const siteSetupFlow: Flow = {
 
 				case 'goals':
 					// Skip to dashboard must have been pressed
+					setIntent( SiteIntent.Build );
```
agrullon95, line 550: Sets site_intent to `build` in redux store if user clicks "Skip to dashboard"

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -191,7 +194,7 @@ const siteSetupFlow: Flow = {
 			navigate( 'processing' );
 
 			// Clean-up the store so that if onboard for new site will be launched it will be launched with no preselected values
-			resetOnboardStoreWithSkipFlags( [ 'skipPendingAction' ] );
+			resetOnboardStoreWithSkipFlags( [ 'skipPendingAction', 'skipIntent' ] );
```
agrullon95, line 221: Added `skipIntent` flag to prevent the site intent from being deleted before its used in the setPendingAction  function above.

packages/data-stores/src/onboard/reducer.ts
```diff
@@ -294,6 +294,9 @@ const intent: Reducer< string, OnboardAction > = ( state = '', action ) => {
 	if ( action.type === 'SET_INTENT' ) {
 		return action.intent;
 	}
+	if ( action.type === 'RESET_ONBOARD_STORE' && action?.skipFlags?.includes( 'skipIntent' ) ) {
```
agrullon95, line 310: If we reset the onboard store and include the `skipIntent` flag, then we don't want to delete the site_intent state value.

sharpirate:
✅ Still tests well after the changes

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -164,12 +165,14 @@ const siteSetupFlow: Flow = {
 					if ( ! siteSlug ) {
 						return;
 					}
+
+					const siteIntent = getIntent();
 					const siteId = site?.ID;
 					const pendingActions = [
-						setIntentOnSite( siteSlug, intent ),
+						setIntentOnSite( siteSlug, siteIntent ),
```
Addison-Stavlo, line 185: why cant we use the regular `intent` here (and below)? Is there a reason we need to have both:

```
		const intent = useSelect( ( select ) => select( ONBOARD_STORE ).getIntent() );
		const { getIntent } = useSelect( ( select ) => select( ONBOARD_STORE ) );
```

client/landing/stepper/declarative-flow/site-setup-flow.ts
Addison-Stavlo: I dont see this reset being called in many other places. So it seems like we are adding `skipIntent` to the main case here?

client/landing/stepper/declarative-flow/site-setup-flow.ts
Addison-Stavlo: Im not understanding the need for the other changes, but this seems like the main thing we need here?

agrullon95: Yes this is the main change, but I needed to add the other changes in order to get the latest site_intent value.

client/landing/stepper/declarative-flow/site-setup-flow.ts
agrullon95: Im not able to get the correct value from using the existing `intent` variable. It appears as an empty string.

Addison-Stavlo (commented):
Behavior tests well on my end 👍 

To be safe and communicative here, would you mind writing a P2 about this and crossposting for dotcom and related teams? essentially that we are making this "skip" from this goals page default to `build`, similar to how "other" flow defaults to `build`? Mostly as a heads up, but also to double check for any concerns with this we may have missed?

client/landing/stepper/declarative-flow/site-setup-flow.ts
agrullon95: Yes im adding it here to ensure Im still able to retrieve the site intent value from redux inside the `exitFlow` pendingActions code.

Without adding this and modifying redux, I noticed that site intent value was always an empty string when I used the `getIntent` selector.

client/landing/stepper/declarative-flow/site-setup-flow.ts
edanzer: I was going to ask the same question. Any thoughts on why? I assume it may a JS timing issue related to how and when pendingAction is invoked? Like in one case, you're getting the value of intent NOW as you set up the function (essentially acts as a closure and stores the current value), and in the other, your fetching site intent later, when the method itself is fired. 

I will say it seems odd in this file to be fetching intent here twice via two different methods, in two different places.

edanzer: I think one other thing I'm puzzle by here is that before this PR, we WERE using the first/original intent value. Does that mean that code was not working as expected?

edanzer: I'm a bit concerned about whether this change (since it clearly does change the behavior and possible value of intent) will have unexpected impact on other existing flows beyond what we're wanting to do just for ours.

edanzer: Reviewed this @agrullon95 and just did some extensive testing of other scenarios and flows to confirm I'm not seeing any issues.

edanzer (approved):
This tests as expected. I have some concerns that these code changes could plausibly impact how intent is set in a range of circumstances. I reviewed with @agrullon95 and I tested both the intended scenario (Skip to Dashboard) but also a wide mix of other flows and scenarios, and everything tests as expected. 

In another comment @Addison-Stavlo mentioned maybe posting about this. Not sure if it's worth flagging one of the people who worked on the site-setup-flow file to confirm they're good with it too? 

All that said, given all the checks and testing, I'm also approving now.

agrullon95:
@edanzer - I added a link to the P2 in the description where I gave other teams a heads-up about these changes

PR #72888, opened 2023-02-01
Launchpad: Disable Launchpad on Goals page revisit

### LINKED ISSUE #72885: LaunchPad in General Onboarding - unset launchpad screen if not a launchpad goal.
currently there can be some edge cases where users activate a non-launchpad supported `site_intent` after having activated a supported one and having the launchpad enabled. - resulting in a launchpad enabled for a non supported flow/intent.

currently, we check to enable the launchpad here: https://github.com/Automattic/wp-calypso/blob/ec8706020e2a31bc0616a47a8e339394a429e0ff/client/landing/stepper/declarative-flow/site-setup-flow.ts#L183-L184 

Perhaps we should add a second condition: `if NOT isLaunchPadIntent -> set launchpad_screen = off` ?

### How to Repro/ Where the user could encounter this:

Imagine a user goes through a general flow that enables launchpad (such as "other") but saves a link to the goals page for later use just in case they want to change their selection. Later they visit the goals page, select another flow like "built-by-express", and checkout that way for their site. The result will be a launchpad enabled for a non-launchpad flow.

To Repro:
* create a new site in general onboarding.
* at the goals page, open a new tab in the broswer and copy/paste the URL from the goals page.
* on one goals page, select "other" and go through the flow to my home.
* after this is done, go to the tab with the other goals page open.
* select the "seller" option and follow this through to my home as well.
* check the sites `site_intent` and `launchpad_screen` options, you will notice we have the launchpad enabled for the seller flow which we do not currently support.

sharpirate:
Fixes #72885 

### Time Estimate
Review: Short
Test: Short

### Summary
Currently, if the user selects a Launchpad selected flow on the Goals page and later revisits the Goals page and select a non-Launchpad flow we are not disabling Launchpad.

##### Test current behavior
1. Go to https://wordpress.com/start
2. Once you arrive at the Goals page save the URL
3. On the Goal page select "Other"
4. After finishing the flow confirm that you are redirected to Launchpad
5. Go back to the Goals page from the URL you saved
6. On the Goal page select "Sell"
7. After finishing the flow confirm that you are redirected to Launchpad

##### Test Fix
1. Go to http://calypso.localhost:3000/start
2. Once you arrive at the Goals page save the URL
3. On the Goal page select "Other"
4. After finishing the flow confirm that you are redirected to Launchpad
5. Go back to the Goals page from the URL you saved
6. On the Goal page select "Sell"
7. After finishing the flow confirm that you are NOT redirected to Launchpad

### DIFF
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -184,7 +184,7 @@ const siteSetupFlow: Flow = {
 					if ( typeof siteId === 'number' ) {
 						pendingActions.push(
 							saveSiteSettings( siteId, {
-								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : null,
+								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : 'off',
 							} )
 						);
 					}
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -180,9 +180,13 @@ const siteSetupFlow: Flow = {
 						);
 					}
 
-					// Add Launchpad to selected intents in General Onboarding
-					if ( isLaunchpadIntent( intent ) && typeof siteId === 'number' ) {
-						pendingActions.push( saveSiteSettings( siteId, { launchpad_screen: 'full' } ) );
+					// Update Launchpad option based on site intent
+					if ( siteId ) {
+						pendingActions.push(
+							saveSiteSettings( siteId, {
+								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : 'off',
+							} )
+						);
 					}
 
 					let redirectionUrl = to;
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -181,7 +181,7 @@ const siteSetupFlow: Flow = {
 					}
 
 					// Update Launchpad option based on site intent
-					if ( siteId ) {
+					if ( typeof siteId === 'number' ) {
 						pendingActions.push(
 							saveSiteSettings( siteId, {
 								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : null,
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -184,7 +184,7 @@ const siteSetupFlow: Flow = {
 					if ( siteId ) {
 						pendingActions.push(
 							saveSiteSettings( siteId, {
-								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : 'off',
+								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : null,
 							} )
 						);
 					}


client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -179,9 +179,13 @@ const siteSetupFlow: Flow = {
...
-					// Add Launchpad to selected intents in General Onboarding
-					if ( isLaunchpadIntent( intent ) && typeof siteId === 'number' ) {
-						pendingActions.push( saveSiteSettings( siteId, { launchpad_screen: 'full' } ) );
+					// Update Launchpad option based on site intent
+					if ( siteId ) {
+						pendingActions.push(
+							saveSiteSettings( siteId, {
+								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : 'off',
+							} )
+						);
```
Addison-Stavlo: This seems reasonable. Although Im wondering if there could be any issues setting it to `'off'` this early in the process (i know i recommended that in the issue but now having second thoughts).

IIRC we might use null vs 'off' as a distinguishing factor in some places like the my home stale data redirect logic? cc @jeyip.  If that is the case maybe setting it as null or empty string might make more sense. 🤔

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -179,9 +179,13 @@ const siteSetupFlow: Flow = {
...
+					// Update Launchpad option based on site intent
+					if ( siteId ) {
+						pendingActions.push(
+							saveSiteSettings( siteId, {
+								launchpad_screen: isLaunchpadIntent( intent ) ? 'full' : null,
+							} )
+						);
 					}
 
 					Promise.all( pendingActions ).then( () => window.location.assign( to ) );
```
jeyip: `NB:` Do we need to be catching exceptions here? Wondering what we do in the event that site settings fail to be toggled properly cc @Addison-Stavlo

client/landing/stepper/declarative-flow/site-setup-flow.ts
jeyip: From what I can tell, using either value of null or ‘off’ would not affect cache invalidation and redirection in my home for the launchpad, but I’d vote for null or false as well

We only care about whether or not the value is full , and if it is, refetch data and make sure that the Launchpad wasn’t set to off recently

https://github.com/Automattic/wp-calypso/blob/e05b2fa84ddfd9671c51cb51f4aea3492228a71e/client/my-sites/customer-home/controller.jsx#L48-L50

client/landing/stepper/declarative-flow/site-setup-flow.ts
Addison-Stavlo: it looks like we were previously ensuring `siteId` was a number `&& typeof siteId === 'number'`. Should we still be doing that as well?

sharpirate: @jeyip I've updated the PR to add back the `typeof` check but I'm not sure that was what you were referring to..

agrullon95:
@Sharpirate - I was able to create a site using the `Write & publish` goal and was navigated to the launchpad screen correctly. I went back to the onboarding goals page using the same site slug and walked through the flow using the `Sell online` goal. The `site_intent` was updated to `sell` , but `launchpad_screen` didn't get updated to `null`.

Addison-Stavlo:
> I was able to create a site using the Write & publish goal and was navigated to the launchpad screen correctly. I went back to the onboarding goals page using the same site slug and walked through the flow using the Sell online goal. The site_intent was updated to sell , but launchpad_screen didn't get updated to null.

Same (I went through the same flow as outlined above). Initially I wondered if it was stale calypso data but I checked the options on the backend as well and it was still `full`. 

After some digging I added some notes/links in this slack thread as to why - p1675431804346469-slack-CHN6J22MP

sharpirate:
I've updated the PR to use `off` again and I create a [follow up for deleting the option in the future ](https://github.com/Automattic/wp-calypso/issues/72976)

agrullon95 (approved):
#### Test results

Tested using existing site with site_intent set to "Write" and launchpad screen enabled.
I revisited the goals page, selected the `Sell online` goal and skipped to the dashboard. I was able to verify that the site_intent was updated to `Sell` and launchpad_screen option was set to `off`.

Would we also want to set `launchpad_checklist_tasks_statuses` to an empty array to clear out the launchpad checklist-related data?

sharpirate:
Thanks @agrullon95, I'll be addressing the `launchpad_checklist_tasks_statuses` in a follow up PR shortly

PR #72907, opened 2023-02-02
Logged-out TS: Hide eCommerce plan if destination flow is site assembler

arthur791004:
#### Proposed Changes

Referring to p1675233608414999/1675233514.738739-slack-CRWCHQGUB and p1675305886405959/1675301136.567349-slack-CRWCHQGUB, there are 2 issues for leading people from the Logged-out Theme Showcase to the Site Assembler flow
  * The Site Assembler doesn't support atomic sites, so if the user selects an eCommerce plan, the result is not expected. Moreover, we will lead to eCommerce flow if the user selects an eCommerce plan and it's weird for the user who clicks on the Pattern Assembler CTA in the Logged-out Theme Showcase.
  * For now we only use the `theme` from the query string to determine whether to go to the Site Assembler flow. However, if we add the Blank Canvas theme back to the Theme Showcase, the user will land on the Site Assembler flow too even if they just want to activate the Blank Canvas theme.

Hence, this PR is proposing adding the `destination_flow=site_assembler` (any better naming?) to the query string when the user clicks on the Pattern Assembler CTA.

https://user-images.githubusercontent.com/13596067/216254448-b2b34877-1b31-4e6a-99a6-9e90e22e7dfd.mov

#### Testing Instructions

* Open the browser with Incognito mode
* Open the developer console and run below code to make sure the feature flag is enabled
  * `document.cookie = 'flags=pattern-assembler/logged-out-showcase;max-age=1209600;path=/'`
* Go to /themes
* Select the Pattern Assembler CTA
* When you land on the plan step, check whether the eCommerce plan is hidden
* Finish the flow and check you land on the Site Assembler

#### Pre-merge Checklist

Related to p1675233608414999/1675233514.738739-slack-CRWCHQGUB and p1675305886405959/1675301136.567349-slack-CRWCHQGUB

### DIFF
--- a/client/components/themes-list/index.jsx
+++ b/client/components/themes-list/index.jsx
@@ -2,6 +2,7 @@ import { isEnabled } from '@automattic/calypso-config';
 import { FEATURE_INSTALL_THEMES, PLAN_BUSINESS } from '@automattic/calypso-products';
 import { Button } from '@automattic/components';
 import { PatternAssemblerCta, BLANK_CANVAS_DESIGN } from '@automattic/design-picker';
+import { SITE_ASSEMBLER_FLOW } from '@automattic/onboarding';
 import { localize } from 'i18n-calypso';
 import { isEmpty, times } from 'lodash';
 import page from 'page';
@@ -33,6 +34,15 @@ export const ThemesList = ( props ) => {
 		[ props.fetchNextPage ]
 	);
 
+	const goToSiteAssemblerFlow = () => {
+		const params = new URLSearchParams( {
+			ref: 'calypshowcase',
+			theme: BLANK_CANVAS_DESIGN.slug,
+			destination_flow: SITE_ASSEMBLER_FLOW,
+		} );
+		window.location.assign( `/start/with-theme?${ params }` );
+	};
+
 	if ( ! props.loading && props.themes.length === 0 ) {
 		return (
 			<Empty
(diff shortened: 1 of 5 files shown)


taipeicoder (approved):
Works as intended 👍
One thing I noticed was that if I selected a paid plan (Business, in my testing) and then canceled my purchase, I would be redirected to the plans page of the homepage. In this scenario, users bypass the PA and ends up with the Index template. I wonder if purchasing the plan will also bypass the PA flow.

In any case, this is not part of this PR's scope so it's good to go IMO 🙂

taipeicoder:
I also think that having something like `isNewsletterOrLinkInBioFlow( flowName )` is cleaner for determining whether users are on the PA flow, but I think it's currently not possible since the `flowName` is shared with normal signup flows? CMIIW.

arthur791004:
> One thing I noticed was that if I selected a paid plan (Business, in my testing) and then canceled my purchase, I would be redirected to the plans page of the homepage. In this scenario, users bypass the PA and ends up with the Index template. I wonder if purchasing the plan will also bypass the PA flow.

Yes, this is the existing behavior even for the normal signup flow. For example
* If the user selected a paid plan but then canceled, they won't land on the site setup flow.
* If the user wants to unlock the theme in the Design Picker but then canceled, they won't go back to the Design Picker (CMIIW)

For the latter case, we tried to open the full-screen checkout modal before but it was removed after we migrate to the stepper framework 😔

arthur791004:
> I also think that having something like isNewsletterOrLinkInBioFlow( flowName ) is cleaner for determining whether users are on the PA flow, but I think it's currently not possible since the flowName is shared with normal signup flows? CMIIW.

Yes. Newsletter or LiB is able to use `flowName` because they use the Plan step in the stepper framework. However, the Site Assembler in the logged-out TS is using the signup flow to let the user create a new site, so we cannot use the `flowName`. I tried to make the site assembler flow support the domain step, plan step, and site creation step, but it's more complicated than this way.

RyanaHunter:
I understand that you're facing issues related to the eCommerce plan in the Logged-out Theme Showcase. You're trying to hide it when the destination flow is Site Assembler. You have proposed a solution in your PR, which is adding the destination_flow=site_assembler to the query string. I hope that works for you!
Regarding your questions, there are [Ecommerce Integrations](https://transformagency.com/) tools and services. Besides, there are various tools available that can help you integrate your online store with your website. You can check them out and see which suits your needs and budget best.

PR #72912, opened 2023-02-02
HappyBlocks: Remove automatic domain detection in Pricing Plans block

msurdi-a8c:
#### Proposed Changes

fixes 237-gh-Automattic/lighthouse-forums

This PR changes the behaviour of the Pricing Plans block so that the domain has to be explicitly set in the block settings.

There are several behavioural corner cases we'll need to carefully rethink for this to work properly, such as:

* Priority of values: what's more important, the custom entered value or the selected one from the topic's form?
* In case one of those sources of information changes, should we update or should we preserve the previous value?
* If there's no custom value, does that mean we want the block to not have a domain? Or does that mean we should default to the one from the topic? Etc.

Until we clearly define the behaviour it's better to remove this functionality to avoid bugs as the one linked above.

#### Testing Instructions

* Sync these changes with `yarn dev --sync`
* Create a new topic, add a new Pricing Plan block by typing `/personal` in the editor
* Ensure the block is shown without the byline "for [domain]..." under the plan name
* In the block properties, set a domain
* Ensure the block now shows the byline with "for [domain_you_entered]".

#### Pre-merge Checklist

Related to #

### DIFF
--- a/apps/happy-blocks/src/pricing-plans/edit.tsx
+++ b/apps/happy-blocks/src/pricing-plans/edit.tsx
@@ -40,38 +40,6 @@ export const Edit: FunctionComponent< BlockEditProps< BlockAttributes > > = ( {
 		} );
 	}, [ attributes.planTypeOptions.length, attributes.productSlug, plans, setAttributes ] );
 
-	useEffect( () => {
-		if ( attributes.domain ) {
-			return;
-		}
-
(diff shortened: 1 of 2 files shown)


johngodley (approved):
Confirmed that the domain doesn't show until it has been added, and the code is all deletion so an easy review! 👍

Issue #72015, opened 2023-01-12
Writing Prompts - Add Reader Writing Prompt Promo


davemart-in:
## Description

Let’s repurpose the My Home card and start occasionally showing it in the reader in your “Following” feed. We’ll need to figure out a frequency which won’t be too repetitive.

![Image](https://user-images.githubusercontent.com/5634774/212122435-eca3f336-a1b9-4623-a02b-37763673768b.png)

PR #72923, opened 2023-02-02
Inject writing prompt card into reader stream

eoigal:
This PR intends to inject the writing prompt home card into the Reader following stream.

Image

Ref - https://github.com/Automattic/wp-calypso/issues/72015

Things to work out;

- [x] How do we inject an component into the post stream on Reader?
- [x] How often should the prompt card be rendered in that stream?
- [x] What site will link to when rendering the prompt card? (where will the answer be posted?)

### DIFF
--- a/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
+++ b/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
@@ -1,6 +1,7 @@
 import { Card, Button, Gridicon } from '@automattic/components';
 import classnames from 'classnames';
 import { useTranslate } from 'i18n-calypso';
+import { sortBy } from 'lodash';
 import { useSelector, useDispatch } from 'react-redux';
 import CardHeading from 'calypso/components/card-heading';
 import EllipsisMenu from 'calypso/components/ellipsis-menu';
@@ -8,6 +9,8 @@ import { useBloggingPrompts } from 'calypso/data/blogging-prompt/use-blogging-pr
 import useSkipCurrentViewMutation from 'calypso/data/home/use-skip-current-view-mutation';
 import { SECTION_BLOGGING_PROMPT } from 'calypso/my-sites/customer-home/cards/constants';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import getSitesItems from 'calypso/state/selectors/get-sites-items';
+import { setSelectedSiteId } from 'calypso/state/ui/actions';
 import { getSelectedSiteId, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 import BellOffIcon from './bell-off-icon';
 import LightbulbIcon from './lightbulb-icon';
@@ -18,7 +21,22 @@ import './style.scss';
 const BloggingPromptCard = () => {
 	const dispatch = useDispatch();
 	const translate = useTranslate();
-	const siteId = useSelector( ( state ) => getSelectedSiteId( state ) );
+	const selectedSiteId = useSelector( ( state ) => getSelectedSiteId( state ) );
+	let sites = useSelector( ( state ) => getSitesItems( state ) );
+	let siteId = selectedSiteId;
+
+	// Set selected site ID to first site with write site_intent
+	if ( siteId === null ) {
+		sites = sortBy( sites, [ 'ID' ] );
+		const blogs = Object.values( sites ).filter( ( site ) => {
+			return site.options?.site_intent === 'write';
+		} );
+		if ( blogs.length > 0 ) {
+			siteId = blogs[ 0 ].ID;
+			dispatch( setSelectedSiteId( siteId ) );
+		}
+	}
+
 	const siteSlug = useSelector( ( state ) => getSelectedSiteSlug( state ) );
 	const maxNumberOfPrompts = 10;
 	const { data: prompts } = useBloggingPrompts( siteId, maxNumberOfPrompts );
--- a/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
+++ b/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
@@ -9,6 +9,7 @@ import { useBloggingPrompts } from 'calypso/data/blogging-prompt/use-blogging-pr
 import useSkipCurrentViewMutation from 'calypso/data/home/use-skip-current-view-mutation';
 import { SECTION_BLOGGING_PROMPT } from 'calypso/my-sites/customer-home/cards/constants';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import getPrimarySiteId from 'calypso/state/selectors/get-primary-site-id';
 import getSitesItems from 'calypso/state/selectors/get-sites-items';
 import { getSelectedSiteId, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 import BellOffIcon from './bell-off-icon';
@@ -21,6 +22,7 @@ const BloggingPromptCard = () => {
 	const dispatch = useDispatch();
 	const translate = useTranslate();
 	const selectedSiteId = useSelector( ( state ) => getSelectedSiteId( state ) );
+	const primarySiteId = useSelector( ( state ) => getPrimarySiteId( state ) );
 	let sites = useSelector( ( state ) => getSitesItems( state ) );
 	let siteId = selectedSiteId;
 
@@ -32,6 +34,9 @@ const BloggingPromptCard = () => {
 		} );
 		if ( blogs.length > 0 ) {
 			siteId = blogs[ 0 ].ID;
+		} else {
+			// Fallback to using Primary blog id so we can request prompts
+			siteId = primarySiteId;
 		}
 	}
 
--- a/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
+++ b/client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
@@ -1,7 +1,6 @@
 import { Card, Button, Gridicon } from '@automattic/components';
 import classnames from 'classnames';
 import { useTranslate } from 'i18n-calypso';
-import { sortBy } from 'lodash';
 import { useSelector, useDispatch } from 'react-redux';
 import CardHeading from 'calypso/components/card-heading';
 import EllipsisMenu from 'calypso/components/ellipsis-menu';
@@ -10,7 +9,6 @@ import useSkipCurrentViewMutation from 'calypso/data/home/use-skip-current-view-
 import { SECTION_BLOGGING_PROMPT } from 'calypso/my-sites/customer-home/cards/constants';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
 import getPrimarySiteId from 'calypso/state/selectors/get-primary-site-id';
-import getSitesItems from 'calypso/state/selectors/get-sites-items';
 import { getSelectedSiteId, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 import BellOffIcon from './bell-off-icon';
 import LightbulbIcon from './lightbulb-icon';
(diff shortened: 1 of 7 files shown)


roo2:
> What site will link to when rendering the prompt card? where will the answer be posted?
I think there are a few options:
* The default site
* The site selector page?
* The "first blogging site" using similar logic to what we used for the notifications

I'm in favor of linking to the site selector page. You can link to `/post` (with no site) and it will take them to this selector. When they select a site it will take them through to the editor. 

What do you think @davemart-in ?
Screen Shot 2023-02-06 at 7 15 48 am

eoigal:
I've updated the patch to link to just `/post` when answering a writing prompt and the site ID is not selected - this will mean the site selector will be shown.

I have a concern with how we track stats with no site ID set though. It will appear as null in cases where the site ID is not selected.

I also added some logic in to space out when we show the prompts in the reader stream. I've space it out more than we do the recommended posts as I think we should but let me know if you think differently.

davemart-in:
@eoigal I did a walkthrough. Everything looks really great.

I noticed a reproducible JS error in the editor every time I clicked to write a blog post, but I'm not sure it's related to your PR. Here's what I'm seeing:

https://user-images.githubusercontent.com/5634774/217586382-184104b2-10a6-49f6-8386-c428f6310c37.mp4

```
Block validation: Block validation failed for `core/button` ({name: 'core/button', icon: {…}, keywords: Array(1), attributes: {…}, providesContext: {…}, …}).

Content generated by `save` function:

<div class="wp-block-button has-custom-font-size has-medium-font-size" style="text-transform:uppercase"><a class="wp-block-button__link wp-element-button">Get Directions →</a></div>

Content retrieved from post body:

<div class="wp-block-button" style="text-transform:uppercase"><a class="wp-block-button__link wp-element-button has-custom-font-size has-medium-font-size">Get Directions →</a></div>
```

It seems to be talking about a **Get Directions** button? I have no idea where that is coming from.

Note: I see this error when I click the "Post Answer" button on My Home as well, so this doesn't appear to be stemming from your PR.

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
```diff
@@ -24,7 +24,9 @@ const PromptsNavigation = ( { prompts } ) => {
 		return prompts !== undefined ? prompts[ promptIndex ] : null;
 	};
 
-	const getNewPostLink = () => addQueryArgs( editorUrl, { answer_prompt: getPrompt()?.id } );
+	// If no site ID set, go through site selector before rendering post editor
+	const getNewPostLink = () =>
+		addQueryArgs( siteId ? editorUrl : '/post', { answer_prompt: getPrompt()?.id } );
 
 	const goToPreviousStep = () => {
 		let nextIndex = promptIndex - 1;
```
creativecoder, line 32: We should pull this out of `client/my-sites/customer-home/cards/features/` if it's going to be used in both My Sites and Reader. Maybe somewhere in `client/components`? What do you think?

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
```diff
@@ -18,8 +20,24 @@ import './style.scss';
...
+	if ( siteId === null ) {
+		sites = sortBy( sites, [ 'ID' ] );
+		const blogs = Object.values( sites ).filter( ( site ) => {
+			return site.options?.site_intent === 'write';
+		} );
+		if ( blogs.length > 0 ) {
+			siteId = blogs[ 0 ].ID;
+		}
+	}
+
```
creativecoder: What is the selected site when on reader? Is this just the most recent site viewed in the My Sites section?

I thinking falling back to the `getPrimarySiteId` selector if there are no blogs with write intent (which is a fairly recent thing added to onboarding and likely not present for many users) would be a good addition to this block.

eoigal: Good idea, I'll add a fallback to primary blog id if no blog with write intent is found.

In this case we need a site ID to allow us to get the list of prompts.

When rendering the prompts nav though the selected site ID is used in track stats - I'm not sure how to handle this.

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
eoigal: This is a good idea too, I'll create a task to handle this in a follow up PR. Thanks.

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
```diff
@@ -1,13 +1,15 @@
 import { Card, Button, Gridicon } from '@automattic/components';
 import classnames from 'classnames';
 import { useTranslate } from 'i18n-calypso';
+import { sortBy } from 'lodash';
```
roo2: We're trying to avoid bringing in lodash dependencies p4TIVU-9GZ-p2 (@creativecoder just reminded me on my pr 😄) it looks like sortBy can be replaced by the ES6 `sort` which takes a comparison function

roo2:
I can see a few further design tweaks to match the design shown in https://github.com/Automattic/wp-calypso/issues/72015:

* Vertical space above and below the prompt should be increased to match the internal padding in the other cards
* Prompt should be left aligned with the text, not with the edge of the cards
* Card should have rounded corners to match the images.
* Post Answer button should also have rounded corners.

Screen Shot 2023-02-09 at 9 39 19 am

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
```diff
@@ -18,8 +20,24 @@ import './style.scss';
 const BloggingPromptCard = () => {
 	const dispatch = useDispatch();
 	const translate = useTranslate();
-	const siteId = useSelector( ( state ) => getSelectedSiteId( state ) );
+	const selectedSiteId = useSelector( ( state ) => getSelectedSiteId( state ) );
+	let sites = useSelector( ( state ) => getSitesItems( state ) );
+	let siteId = selectedSiteId;
+
+	// If no selected site ID, set site ID to first site with write site_intent
+	if ( siteId === null ) {
```
roo2, line 28: Since siteId is only used to retrieve the blogging prompts, I think that we don't need this method to find the first blogging site, the default site would work just as well. 

For the "Post Answer" link, we're just linking to the site selector regardless.

client/reader/stream/utils.js
```diff
@@ -89,3 +89,45 @@ export function getDistanceBetweenRecs( totalSubs ) {
...
+	if ( totalSubs <= 0 ) {
+		// 0 means either we don't know yet, or the user actually has zero subs.
+		// if a user has zero subs, we don't show posts at all, so just treat 0 as 'unknown' and
+		// push recs to the max.
+		return MAX_DISTANCE_BETWEEN_PROMPTS;
+	}
+
+	return Math.min(
+		Math.max(
+			Math.floor( Math.log( totalSubs ) * Math.LOG2E * 5 - 3 ),
```
roo2: It seems too complicated, but also quite genius 😄

roo2: I think it should generally be less frequent and converge to 50 sooner, here's some outputs for `totalSubs` I just tested in the dev tools panel:

2 => 2  (rounded to 10)
5 => 8  (rounded to 10)
10 => 13
20 => 18
50 => 25
100 => 30
200 => 35
500 => 41

client/state/reader/streams/selectors/get-reader-stream-transformed-items.js
```diff
@@ -26,6 +31,8 @@ export const getTransformedStreamItems = treeSelect(
 			items = injectRecommendations( items, recs, getDistanceBetweenRecs( follows.length ) );
 		}
 
+		items = injectPrompts( items, getDistanceBetweenPrompts( follows.length ) );
```
roo2, line 34: Nice that this hook already exists for this 👍

client/reader/stream/utils.js
eoigal: Updated this to converge to 50 sooner

client/state/reader/streams/selectors/get-reader-stream-transformed-items.js
eoigal: Agree! Easier to reuse  existing pattern

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
eoigal: Getting rid of lookup on sites with write intent - doesn't serve much purpose now since not setting selected site id.

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
eoigal: Removed.

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
eoigal: https://github.com/Automattic/wp-calypso/issues/73158

davemart-in (approved):
Visually this looks good to me. Nice work!

creativecoder (dismissed):
I think there's some kind of bug in the injection.

As I scroll down the feed page, the injected card duplicates and starts appearing in other places, like the top of the page.

image

creativecoder:
Less critical and more opinion, the inset card looks odd to me. Other items injected use the full width, I think we should do the same with the blogging prompts card.

image

image

ollierozdarz:
> the inset card looks odd to me. Other items injected use the full width, I think we should do the same with the blogging prompts card.

@creativecoder @eoigal yeah, now that I see this in context to the entire page, I agree with the above. Let's extend it back out full-width. I beleive it's just removing the `margin-left: 52px;` below.

.reader-stream__blogging-prompt {
    ~margin-left: 52px;~
}

client/reader/stream/utils.js
```diff
@@ -89,3 +89,45 @@ export function getDistanceBetweenRecs( totalSubs ) {
...
+export function injectPrompts( posts, itemsBetweenPrompts ) {
+	if ( posts.length < itemsBetweenPrompts ) {
+		return posts;
+	}
+
+	return flatMap( posts, ( post, index ) => {
+		if ( index && index % itemsBetweenPrompts === 0 ) {
+			const promptBlock = {
+				isPromptBlock: true,
+				index: index,
```
creativecoder, line 127: I'm not sure if there's any practical impact, but the index passed in the `injectRecommendations` function is a separate `recIndex` that's incremented for each recommendation that's injected. Maybe we should do the same here for prompts?

creativecoder (commented):
It can be in a follow-up, but a nice improvement to this would be showing a different prompt each time the prompt card is injected, rather that the same prompt over and over again.

client/reader/stream/utils.js
eoigal: This is a good idea. Will leave it for a followup however.

davemart-in:
I tried replicating Gran't bug report. I tried:

- Scrolling slowly
- Scrolling quickly
- Scrolling multiple times down and up the page in the same page load
- Scrolling REALLY far (10+ instances of the writing prompt component)
- Scrolling via the keyboard down arrow

None of which seemed to replicate the bug or produce any sort of JS error.

creativecoder:
In case it helps, the account showing the bug is a non-a11n test account, and the default feed has a lot of "Feed: All Latest" posts included in it.

creativecoder:
FYI, "Feed: All Latest" is actually the main feed from wired.com.

client/my-sites/customer-home/cards/features/blogging-prompt/index.jsx
```diff
@@ -18,8 +19,19 @@ import './style.scss';
...
+	let siteId = selectedSiteId;
+
+	// We need a site ID to request prompts
+	// If no selected site ID, fallback to using Primary site ID
+	if ( siteId === null ) {
+		siteId = primarySiteId;
+	}
+
 	const siteSlug = useSelector( ( state ) => getSelectedSiteSlug( state ) );
+	const notificationSettingsLink = '/me/notifications' + siteSlug ? '#' + siteSlug : '';
```
creativecoder: When I go back to use this card on My Home, the "Manage Notifications" link doesn't work any more.

creativecoder (commented):
Nice job tracking down the `key` prop issue--I no longer see a console error or duplicated prompt cards with the fix you added.

Just one small problem I'm having with the Notifications link, otherwise this is all now working as expected for me.

client/reader/stream/utils.js
creativecoder: Since it was a 2 line change, I went ahead and did it in https://github.com/Automattic/wp-calypso/pull/72923/commits/4201cf5cb0fb3fb2fa1adcbfdecc5fed38d57053 -- but feel free to revert that if you think it's problematic.

creativecoder:
For anyone following along, the key prop bug only seemed to be reproducible using a non-a11n account, for unknown reasons. But it looks to be fixed by https://github.com/Automattic/wp-calypso/pull/72923/commits/022f97750a5abcde7625ca31dc2b72953b706b70. I'm glad we caught that and were able to figure out a fix!

creativecoder:
@eoigal I created these follow-up issues that came out of reviews/discussions on this PR

- https://github.com/Automattic/wp-calypso/issues/73257
- https://github.com/Automattic/wp-calypso/issues/73258

PR #72930, opened 2023-02-02
Add queryParam to invalid cache when navigating away from site setup flow

agrullon95:
#### Testing/Review: 

Test: short
Review: short

#### Proposed Changes

* After a user walks through the general onboarding and selects a launchpad enabled goal (`Write` or `Build`), the user may be navigated to the home screen instead of the launchpad screen. These changes add the `postOnboarding=true` query parameter to site setup exit flow redirection URL, invalids the site data and fetches it again.

#### Testing Instructions

- Check out this branch
- `yarn start`
- Start general onboarding at calypso.localhost:3000/start
- At the goals step, select the `write` site intent
- Click on continue
- Click on the skip to dashboard button
- The user should see the `postOnboarding=true` query param in the /home url before being redirected to Launchpad
- If the redirect is too fast to see, then comment out the following code https://github.com/Automattic/wp-calypso/blob/70ac415c7a013ea1a182511ec3d2cd77ede3a328/client/my-sites/customer-home/controller.jsx#L55-L62

#### Pre-merge Checklist

Related to #72872

### DIFF
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -5,6 +5,7 @@ import { useSelect, useDispatch } from '@wordpress/data';
 import { useDispatch as reduxDispatch, useSelector } from 'react-redux';
 import { ImporterMainPlatform } from 'calypso/blocks/import/types';
 import { useQuery } from 'calypso/landing/stepper/hooks/use-query';
+import { addQueryArgs } from 'calypso/lib/route';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
 import { getCurrentUser } from 'calypso/state/current-user/selectors';
 import { getActiveTheme, getCanonicalTheme } from 'calypso/state/themes/selectors';
@@ -184,7 +185,14 @@ const siteSetupFlow: Flow = {
 						pendingActions.push( saveSiteSettings( siteId, { launchpad_screen: 'full' } ) );
 					}
 
-					Promise.all( pendingActions ).then( () => window.location.assign( to ) );
+					let redirectionUrl = to;
+
+					// Forcing cache invalidation to retrieve latest launchpad_screen option value
+					if ( isLaunchpadIntent( intent ) ) {
+						redirectionUrl = addQueryArgs( { postOnboarding: true }, to );
+					}
+
+					Promise.all( pendingActions ).then( () => window.location.assign( redirectionUrl ) );
 				} );
 			} );
 
--- a/client/landing/stepper/declarative-flow/site-setup-flow.ts
+++ b/client/landing/stepper/declarative-flow/site-setup-flow.ts
@@ -189,7 +189,7 @@ const siteSetupFlow: Flow = {
 
 					// Forcing cache invalidation to retrieve latest launchpad_screen option value
 					if ( isLaunchpadIntent( intent ) ) {
-						redirectionUrl = addQueryArgs( { postOnboarding: true }, to );
+						redirectionUrl = addQueryArgs( { showLaunchpad: true }, to );
 					}
 
 					Promise.all( pendingActions ).then( () => window.location.assign( redirectionUrl ) );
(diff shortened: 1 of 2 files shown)


jeyip (approved):

jeyip:
Paired with @agrullon95 on this so +1 from me

client/landing/stepper/declarative-flow/site-setup-flow.ts
```diff
@@ -184,7 +185,14 @@ const siteSetupFlow: Flow = {
 						pendingActions.push( saveSiteSettings( siteId, { launchpad_screen: 'full' } ) );
 					}
 
-					Promise.all( pendingActions ).then( () => window.location.assign( to ) );
+					let redirectionUrl = to;
+
+					// Forcing cache invalidation to retrieve latest launchpad_screen option value
+					if ( isLaunchpadIntent( intent ) ) {
+						redirectionUrl = addQueryArgs( { postOnboarding: true }, to );
```
Addison-Stavlo: if we are only adding the arg for the launchpad flows, maybe a `showLaunchpad` or similar verbage makes more sense than `postOnboarding` ?

client/my-sites/customer-home/controller.jsx
```diff
@@ -38,14 +38,18 @@ export async function maybeRedirect( context, next ) {
...
 	// We need the latest site info to determine if a user should be redirected to Launchpad.
 	// As a result, we can't use locally cached data, which might think that Launchpad is
 	// still enabled when, in reality, the final tasks have been completed and everything is
 	// disabled. Because of this, we refetch site information and limit traffic by scoping down
 	// requests to launchpad enabled sites.
 	// See https://cylonp2.wordpress.com/2022/09/19/question-about-infinite-redirect/#comment-1731
-	if ( maybeStalelaunchpadScreenOption && maybeStalelaunchpadScreenOption === 'full' ) {
+	if (
+		( maybeStalelaunchpadScreenOption && maybeStalelaunchpadScreenOption === 'full' ) ||
+		currentUrl?.includes( 'postOnboarding=true' )
```
Addison-Stavlo: should we get and read the query args specifically here?

Addison-Stavlo: side note, if we are only adding the arg from launchpad flows, then we could maybe just use that itself as the indication to redirect back to the launchpad here as opposed to refetching? Although, maybe the refetching is good to do in the first place anyways to make sure calypso store is truly up to date.

sharpirate:
#### Testing
1. Start general onboarding at calypso.localhost:3000/start
2. At the goals step, select the write site intent
3. Click on continue
4. Click on the skip to dashboard button
✅ The user should see the postOnboarding=true query param in the /home url before being redirected to Launchpad

sharpirate (approved):

client/my-sites/customer-home/controller.jsx
jeyip: > should we get and read the query args specifically here?

Could you elaborate on this?

> if we are only adding the arg from launchpad flows, then we could maybe just use that itself as the indication to redirect back to the launchpad here as opposed to refetching?

If we don't force a refetch, even if we redirect to Launchpad, the site option data in redux still might be stale if they happen to navigate to my home. I believe, however, if the user visits `/view` or the post or site editor ( the only pages the user can navigate to from the Launchpad ), the cache should be invalidated and things should behave as expected, but I'm not sure

Addison-Stavlo: > Could you elaborate on this?

Sure, its probably not a big deal at all but right now we are just parsing the entire string of the URL to see if "postOnboarding=true" is part of the string. As opposed to getting the specific params out of the URL and checking their values. This might generally be something like [$_GET ](https://www.php.net/manual/en/reserved.variables.get.php) in php or maybe [searchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) in JS (maybe theres another method im not remembering). That said, Its probably not a big deal. We aren't going to see `postOnboarding=true` elsewhere in the URL string so it should work the same way.

> If we don't force a refetch, even if we redirect to Launchpad, the site option data in redux still might be stale if they happen to navigate to my home. I believe if we visit /view or the editors the cache should be invalidated, but I'm not sure

That makes sense 👍

Issue #72894, opened 2023-02-01
Sidebar nudge upsell: Update the domain filter


paulopmt1:
[Based on our POC](https://github.com/Automattic/wp-calypso/pull/72831), we expect:

- [ ] Update the domain filter to be at the bottom of the text input
- [ ] It should look like the design: pebzTe-Fx-p2#comment-1131
- [ ] Make sure the changes are connected to the experiment and only show when the user is in the `treatment` group. [Example of code here](https://github.com/Automattic/wp-calypso/pull/72831/commits/8148a79a089580437ddc43cbaa27d6ad03f381e0).

### Visual reference

![Image](https://user-images.githubusercontent.com/1044309/216156268-6b09c94d-ba09-42f5-83e9-490a02fdcfb5.png)

task-related: https://github.com/Automattic/wp-calypso/issues/72516

Issue #72892, opened 2023-02-01
Sidebar nudge upsell: Implement the navigation header


paulopmt1:
[Based on our POC](https://github.com/Automattic/wp-calypso/pull/72831), we expect:

- [x] Implement the new header for the flow
- [x] At the domain page, `Select a Domain` should be selected
- [x] At the plans page, `Select a plan` should be selected
- [x] Back button should go back to the previous page user was
- [ ] Gate the component with `?domainAndPlanPackage=true` and experiment = `treatment`
- [x] It should look like the design: pebzTe-Fx-p2#comment-1131
- [ ] Make sure the changes are connected to the experiment and only show when the user is in the `treatment` group. [Example of code here](https://github.com/Automattic/wp-calypso/pull/72831/commits/8148a79a089580437ddc43cbaa27d6ad03f381e0).

### Design reference

![header](https://user-images.githubusercontent.com/1044309/216154781-ef82de51-4ad9-4f0f-8a7f-74a44739a95e.png)

task-related: https://github.com/Automattic/wp-calypso/issues/72516

DustyReagan:
The [navigation component](https://github.com/Automattic/wp-calypso/pull/72938) exists in trunk now. But we still need to implement it once this [PR](https://github.com/Automattic/wp-calypso/pull/73017) is merged.

paulopmt1:
Added the least [piece of it here](https://github.com/Automattic/wp-calypso/pull/73062/commits/afbc90de0bb8c221ec54a659b986ccf17fccbbad) @DustyReagan, so no extra connection task needed.

DustyReagan:
Awesome! Thanks, @paulopmt1!

PR #72932, opened 2023-02-02
Prepare the domain page to work with the experiment

### LINKED ISSUE #72889: Sidebar nudge upsell: Prepare the domain page to work with the experiment
[Based on our POC](https://github.com/Automattic/wp-calypso/pull/72831), we expect:

- [ ] We're not handling the page header here
- [ ] We're not handling the domain filter changes here
- [ ] Prepare the domain page to change its behavior if the user participates in the experiment
- [ ] Make sure we're implementing a good strategy to avoid changing the URL
- [ ] It should look like the design: pebzTe-Fx-p2#comment-1131
- [ ] Make sure the changes are connected to the experiment and only show when the user is in the `treatment` group. [Example of code here](https://github.com/Automattic/wp-calypso/pull/72831/commits/8148a79a089580437ddc43cbaa27d6ad03f381e0).

### Visual reference

Image

task-related: https://github.com/Automattic/wp-calypso/issues/72516

paulopmt1:
#### Proposed Changes

* Prepare the domain page to work with the experiment. We're focusing on the New header, background color, and centralizing the content. Domain filtering section [will be handled in another task](https://github.com/Automattic/wp-calypso/issues/72894), as well as [the Domain Header](https://github.com/Automattic/wp-calypso/issues/72892).

image

#### Testing Instructions

* Access the current Domain upsell flow starting at `Upgrade` button on your side menu
* The domain page should look like the production one (assuming you're at the control group)
* Using tab `Application` of your console development tool, delete any previous local storage data related to the experiment (here how to do that: pb5gDS-2nD-p2#pro-tip-for-calypso-users) and also delete the Session Storage key `calypso_sidebar_upsell_experiment`
* Change your variation to `treatment` at the experiment `21025-calypso-sidebar-upsell-dedicated-upgrade-flow-202301` following these tips: pb5gDS-2nD-p2#connecting-an-experiment-to-a-jitm
* **IMPORTANT**: First access [the Calypso home](http://calypso.localhost:3000/), and after click at the CTA on the left menu (to ensure our [CSS rules](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4R105) will be applied as expected)
* Make sure your change will work as expected
* You should now see the new interface, make sure it aligns with our design expectations: pebzTe-Fx-p2#comment-1131

#### Pre-merge Checklist

Related to #

### DIFF
--- a/client/controller/index.web.js
+++ b/client/controller/index.web.js
@@ -2,6 +2,7 @@ import config from '@automattic/calypso-config';
 import { getLanguageSlugs } from '@automattic/i18n-utils';
 import { translate } from 'i18n-calypso';
 import page from 'page';
+import { useEffect } from 'react';
 import { QueryClientProvider } from 'react-query';
 import { Provider as ReduxProvider } from 'react-redux';
 import CalypsoI18nProvider from 'calypso/components/calypso-i18n-provider';
@@ -10,6 +11,7 @@ import MomentProvider from 'calypso/components/localized-moment/provider';
 import { RouteProvider } from 'calypso/components/route';
 import Layout from 'calypso/layout';
 import LayoutLoggedOut from 'calypso/layout/logged-out';
+import { loadExperimentAssignment } from 'calypso/lib/explat';
 import { login } from 'calypso/lib/paths';
 import { CalypsoReactQueryDevtools } from 'calypso/lib/react-query-devtools-helper';
 import { getSiteFragment } from 'calypso/lib/route';
@@ -39,6 +41,18 @@ export const ProviderWrappedLayout = ( {
 } ) => {
 	const state = store.getState();
 	const userLoggedIn = isUserLoggedIn( state );
+
+	useEffect( () => {
+		async function loadExperiment() {
+			const experimentAssignment = await loadExperimentAssignment(
+				'calypso_sidebar_upsell_dedicated_upgrade_flow_202301'
+			);
+			const variationName2 = experimentAssignment?.variationName;
+			sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName2 );
+		}
+
+		loadExperiment();
+	}, [] );
 
 	const layout = userLoggedIn ? (
 		<Layout primary={ primary } secondary={ secondary } />
--- a/client/controller/index.web.js
+++ b/client/controller/index.web.js
@@ -2,7 +2,7 @@ import config from '@automattic/calypso-config';
 import { getLanguageSlugs } from '@automattic/i18n-utils';
 import { translate } from 'i18n-calypso';
 import page from 'page';
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { QueryClientProvider } from 'react-query';
 import { Provider as ReduxProvider } from 'react-redux';
 import CalypsoI18nProvider from 'calypso/components/calypso-i18n-provider';
@@ -42,17 +42,33 @@ export const ProviderWrappedLayout = ( {
 	const state = store.getState();
 	const userLoggedIn = isUserLoggedIn( state );
 
-	useEffect( () => {
-		async function loadExperiment() {
-			const experimentAssignment = await loadExperimentAssignment(
-				'calypso_sidebar_upsell_dedicated_upgrade_flow_202301'
-			);
-			const variationName = experimentAssignment?.variationName;
-			sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName );
-		}
+	const [ isExperimentLoaded, setIsExperimentLoaded ] = useState( null );
 
-		loadExperiment();
-	}, [] );
+	async function loadDomainExperiment() {
+		const experimentAssignment = await loadExperimentAssignment(
+			'calypso_sidebar_upsell_dedicated_upgrade_flow_202301'
+		);
+		const variationName = experimentAssignment?.variationName;
+		sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName );
+		setIsExperimentLoaded( true );
+	}
+
+	function shouldLoadDomainExperiment() {
+		return (
+			window.location.pathname.startsWith( '/domains/add' ) ||
+			window.location.pathname.startsWith( '/plans/yearly' )
+		);
+	}
+
+	useEffect( () => {
+		if ( shouldLoadDomainExperiment() ) {
+			loadDomainExperiment();
+		}
+	}, [ currentRoute ] );
+
+	if ( shouldLoadDomainExperiment() && ! isExperimentLoaded ) {
+		return null;
+	}
 
 	const layout = userLoggedIn ? (
 		<Layout primary={ primary } secondary={ secondary } />
--- a/client/controller/index.web.js
+++ b/client/controller/index.web.js
@@ -47,8 +47,8 @@ export const ProviderWrappedLayout = ( {
 			const experimentAssignment = await loadExperimentAssignment(
 				'calypso_sidebar_upsell_dedicated_upgrade_flow_202301'
 			);
-			const variationName2 = experimentAssignment?.variationName;
-			sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName2 );
+			const variationName = experimentAssignment?.variationName;
+			sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName );
 		}
 
 		loadExperiment();
--- a/client/controller/index.web.js
+++ b/client/controller/index.web.js
@@ -2,7 +2,7 @@ import config from '@automattic/calypso-config';
 import { getLanguageSlugs } from '@automattic/i18n-utils';
 import { translate } from 'i18n-calypso';
 import page from 'page';
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { QueryClientProvider } from 'react-query';
 import { Provider as ReduxProvider } from 'react-redux';
 import CalypsoI18nProvider from 'calypso/components/calypso-i18n-provider';
@@ -53,18 +53,20 @@ export const ProviderWrappedLayout = ( {
 		setIsExperimentLoaded( true );
 	}
 
-	function shouldLoadDomainExperiment() {
-		return (
+	const shouldLoadDomainExperiment = useCallback( () => {
+		const domainAndPackage = currentQuery && 'true' === currentQuery.domainAndPlanPackage;
+		const experimentPages =
 			window.location.pathname.startsWith( '/domains/add' ) ||
-			window.location.pathname.startsWith( '/plans/yearly' )
-		);
-	}
+			window.location.pathname.startsWith( '/plans/yearly' );
+
+		return domainAndPackage && experimentPages;
+	}, [ currentQuery ] );
 
 	useEffect( () => {
 		if ( shouldLoadDomainExperiment() ) {
 			loadDomainExperiment();
 		}
-	}, [ currentRoute ] );
+	}, [ currentRoute, currentQuery, shouldLoadDomainExperiment ] );
 
 	if ( shouldLoadDomainExperiment() && ! isExperimentLoaded ) {
 		return null;
--- a/client/layout/index.jsx
+++ b/client/layout/index.jsx
@@ -384,8 +384,15 @@ export default withCurrentRoute(
 		const isJetpack =
 			( isJetpackSite( state, siteId ) && ! isAtomicSite( state, siteId ) ) ||
 			currentRoute.startsWith( '/checkout/jetpack' );
-		const noMasterbarForRoute = isJetpackLogin || currentRoute === '/me/account/closed';
-		const noMasterbarForSection = [ 'signup', 'jetpack-connect' ].includes( sectionName );
+		const userBelongsToExperiment =
+			'treatment' === sessionStorage.getItem( 'calypso_sidebar_upsell_experiment' );
+		const isStartDomainExperiment =
+			( currentRoute.startsWith( '/domains/add' ) || currentRoute.startsWith( '/plans/yearly' ) ) &&
+			userBelongsToExperiment;
+		const noMasterbarForRoute =
+			isStartDomainExperiment || isJetpackLogin || currentRoute === '/me/account/closed';
+		const noMasterbarForSection =
+			isStartDomainExperiment || [ 'signup', 'jetpack-connect' ].includes( sectionName );
 		const masterbarIsHidden =
 			! masterbarIsVisible( state ) ||
 			noMasterbarForSection ||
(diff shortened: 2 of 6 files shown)


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7791902

Thank you @paulopmt1 for including a screenshot in the description! This is really helpful for our translators.

DustyReagan:
Nice work on this, Paulo! It's working for me, but I'm seeing this error below when I run `yarn start`. Looks like it's something to do with our experiment, but I'm not sure what to make of it yet.

error

candy02058912:
+1 with things are working for the code and also seeing the error mentioned [above](https://github.com/Automattic/wp-calypso/pull/72932#issuecomment-1414511295).

Guessing from [this test](https://github.com/Automattic/wp-calypso/blob/update/implement-domains-page-for-experiment/client/lib/explat/internals/test/anon-id.ts#L41), this error is due to when we're trying to do some SSR stuff.

> should return null and log error when run under SSR

The error currently isn't fatal because for logged-in users, [we're not using SSR](https://github.com/Automattic/wp-calypso/blob/update/implement-domains-page-for-experiment/client/controller/index.node.js#L15).

> Server side rendering (SSR) is used exclusively for logged out users.

I've shared another approach here: https://github.com/Automattic/wp-calypso/pull/72945

lupus2k (approved):
Good job, @paulopmt1 🎉

client/controller/index.web.js
```diff
@@ -40,6 +42,18 @@ export const ProviderWrappedLayout = ( {
 	const state = store.getState();
 	const userLoggedIn = isUserLoggedIn( state );
 
+	useEffect( () => {
```
paulopmt1, line 65: Changed the approach to load the experiment based on @candy02058912  [PR](https://github.com/Automattic/wp-calypso/pull/72945)

DustyReagan: I saw @tyxla [commented about this code](https://github.com/Automattic/wp-calypso/pull/72945/files#r1095820066) in @candy02058912's PR. Linking here to pull it into the loop.

paulopmt1: Thanks, @DustyReagan, added [the validation here](https://github.com/Automattic/wp-calypso/pull/72932/commits/8467502c2827e3d4a308f65763af209fe915a728). Now we should be good to go 🤞

client/controller/index.web.js
```diff
@@ -42,17 +42,33 @@ export const ProviderWrappedLayout = ( {
...
-			const variationName = experimentAssignment?.variationName;
-			sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName );
+		if ( shouldLoadDomainExperiment() ) {
+			loadDomainExperiment();
 		}
+	}, [ currentRoute ] );
 
-		loadExperiment();
-	}, [] );
+	if ( shouldLoadDomainExperiment() && ! isExperimentLoaded ) {
```
paulopmt1, line 71: We need this to ensure we'll only Load the subcomponents of the `domain` and `plans` pages once the experiment was loaded. To ensure [codes like this](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4R105) works as expected

client/controller/index.web.js
paulopmt1: Hey @tyxla, can you have a glance at this code to ensure it's good to go?

client/controller/index.web.js
DustyReagan: Question. Do we also want to limit this to routes with `domainAndPlanPackage` set in the query string? For example, ATM if I'm in the treatment group and I go the domain search via the standard path (/domains/add/example.wordpress.com) I see our new design.

paulopmt1: That's a good question @DustyReagan. It will depend if we want to restrict to free users only. Asked here: pebzTe-Fx-p2#comment-1215

paulopmt1: Good catch @DustyReagan, adding our Slack thread here for future reference: p1675449243063399/1675446834.367269-slack-CKZHG0QCR

paulopmt1: [Added it here](https://github.com/Automattic/wp-calypso/pull/72932/commits/98fbcb52c9da99e9ed9e4299d62efad93917e4c9) @DustyReagan

lupus2k (approved):
After a lot of deep tests, it works like a charm, good job @paulopmt1

DustyReagan (approved):
This seems good to me! It's working, constrained to our two routes and `domainAndPlanPackage`, and no build errors.

candy02058912 (approved):
LGTM!
Tested:
* Loads experiment for `/domains/add` and `/plans/yearly` when `domainAndPlanPackage` is set to true.
* Doesn't run `loadDomainExperiment` in other circumstances
* Shows correct UI when I'm in treatment/control group

client/controller/index.web.js
```diff
@@ -40,6 +42,36 @@ export const ProviderWrappedLayout = ( {
 	const state = store.getState();
 	const userLoggedIn = isUserLoggedIn( state );
 
+	const [ isExperimentLoaded, setIsExperimentLoaded ] = useState( null );
```
tyxla, line 45: This runs literally everywhere. At the same time, it's pretty specific and applicable only to part of the application.

IMHO it would be ideal if we could keep experiments for specific workflows away from code that runs everywhere, on every render. 

Just imagine if we add code for dozens more experiments here. I don't think it would be a sustainable way forward.

Is there a really good reason for it to live here? Could we consider moving it somewhere else? A good place could be creating a custom controller method that runs before `makeLayout`. We could then add that only for the routes that actually care about it. 

WDYT?

paulopmt1: > IMHO it would be ideal if we could keep experiments for specific workflows away from code that runs everywhere, on every render.

Agree, and I would love to do that. 

> We could then add that only for the routes that actually care about it.

I tried to achieve this goal [with this strategy](https://github.com/Automattic/wp-calypso/pull/72932/commits/50a82020b814426d3214e248b67354c27a6687e5#diff-18e8c98551ffdb30f512920db2f0637cbe65e6ac35b2a7d7cbb6164ff81b621eR431) - loading the code only on specific routes. Unfortunately, [I had this issue](https://github.com/Automattic/wp-calypso/pull/72932#issuecomment-1414511295) while trying that approach. Do you know how ew can overcome that @tyxla?

tyxla: Ah, I see. I guess we can't run this on the server for an intentional reason. 

In that case, can we wrap this functionality up as something like a query component then, and drop it only where necessary, instead of loading it for everyone in `<Layout />`?

paulopmt1: Yeah, I would love to load it only at the [Domain Search component](client/my-sites/domains/domain-search/index.jsx) but the problem we're having is we also need that data available at the Layout, as [we need to decide showing or not](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-7d1eb73ef67c978553b9d47cf17a8f4b9d93bd396ee6cb5ce89e75aedb8b0f19R393) the masterBar and sideBar while mounting the Layout.

Can you point us a better way to load this, making it available for [the Layout menus decision](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-7d1eb73ef67c978553b9d47cf17a8f4b9d93bd396ee6cb5ce89e75aedb8b0f19R388) and in the same time loading only at the Domain and Plans page @tyxla?

client/controller/index.web.js
```diff
@@ -40,6 +42,36 @@ export const ProviderWrappedLayout = ( {
...
+
+	async function loadDomainExperiment() {
+		const experimentAssignment = await loadExperimentAssignment(
+			'calypso_sidebar_upsell_dedicated_upgrade_flow_202301'
+		);
+		const variationName = experimentAssignment?.variationName;
+		sessionStorage.setItem( 'calypso_sidebar_upsell_experiment', variationName );
+		setIsExperimentLoaded( true );
+	}
```
jessie-ross, line 54: [`loadExperimentAssignment`](https://github.com/Automattic/wp-calypso/blob/trunk/packages/explat-client/README.md#api-explatclientloadexperimentassignment) already caches state both in memory and in localStorage, this function and the use-state above are redundant.

As we are in a React context we should use [`useExperimentAssignment`](https://github.com/Automattic/wp-calypso/blob/trunk/packages/explat-client-react-helpers/README.md#api-useexperimentexperiment_name) with the [`isEligible`](https://github.com/Automattic/wp-calypso/blob/trunk/packages/explat-client-react-helpers/README.md#experimentoptions) option set to `shouldLoadDomainExperiment()`. This will likely result in better safety - it is well tested across Calypso and we won't be doubling anything up - and it will remove a lot of code so there is less to go wrong (`loadDomainExperiment`, `useEffect` can be removed.) It returns `[isLoaded, ExperimentAssignment]` and `isLoaded` will respect the eligibility - so you only have to add in eligibility where-ever you load an experiment assignment, so `shouldLoadDomainExperiment() && ! isExperimentLoaded` can be replaced by `! isLoaded`.

We should also use the HOC [`<ProvideExperimentData>`](https://github.com/Automattic/wp-calypso/blob/trunk/packages/explat-client-react-helpers/README.md#api-provideexperimentdata) rather than adding `isDomainSidebarExperimentUser()` as it is well instrumented and contains several safeguards. We also cache experiment assignments in a particular way so using another additional caching method might lead to subtle bugs.

client/controller/index.web.js
```diff
@@ -40,6 +42,36 @@ export const ProviderWrappedLayout = ( {
...
+	useEffect( () => {
+		if ( shouldLoadDomainExperiment() ) {
+			loadDomainExperiment();
+		}
+	}, [ currentRoute, currentQuery, shouldLoadDomainExperiment ] );
+
+	if ( shouldLoadDomainExperiment() && ! isExperimentLoaded ) {
+		return null;
+	}
+
```
jessie-ross, line 74:
```suggestion
    const isDomainAndPlanPackage = currentQuery && 'true' === currentQuery.domainAndPlanPackage;
    const isExperimentPages =
			window.location.pathname.startsWith( '/domains/add' ) ||
			window.location.pathname.startsWith( '/plans/yearly' );
    const isEligibleForExperiment = isDomainAndPlanPackage && isExperimentPages;
    
    const [ isExperimentLoaded, experimentAssignment ] = useExperimentAssignment(
        'calypso_sidebar_upsell_dedicated_upgrade_flow_202301',
        { isEligible: isEligibleForExperiment }
    );
	
    if ( ! isExperimentLoaded ) {
        return null;
    }
```

jessie-ross (changes requested):
See comments, I recommend using the `useExperimentAssignment` hook as well as the `ProvideExperimentAssignment` HOC.

client/controller/index.web.js
jessie-ross: See https://github.com/Automattic/wp-calypso/pull/72932#discussion_r1098290645 for the change suggested.

client/controller/index.web.js
jessie-ross: > The error currently isn't fatal because for logged-in users, [we're not using SSR](https://github.com/Automattic/wp-calypso/blob/update/implement-domains-page-for-experiment/client/controller/index.node.js#L15).

If this is still causing SSR issues (if the path can be accessed while logged out,) you should add `userLoggedIn` to the isEligible check.

client/controller/index.web.js
tyxla: One way we could do it is by checking if the user is logged in. That's usually a good way to check if we're SSR-ing or not on those isomorphic routes. I see @jessie-ross is also suggesting that [here](https://github.com/Automattic/wp-calypso/pull/72932#discussion_r1098305933).

Note that what you're suggesting is architecturally wrong: blocking the rest of the layout for all screens on Calypso just to check for the value of an experiment that runs only on two sections. This sounds undesirable.

Have we considered trying to provide the experiment value from the server side, as provided by the [current user bootstrap](https://github.com/Automattic/wp-calypso/blob/a14b6727ffc1ed68b951c95f8a43ca6e1cb15ccc/client/server/user-bootstrap/index.js#L28)? That would cause fewer hops and we'll have that data immediately in `<Layout />`. WDYT?

razvanpapadopol: Hi, @tyxla 👋🏼 
Thanks for thoughtfully reviewing these bits! Your help is invaluable, as always ❤️ 
I find this scenario a very common one and @Automattic/experimentation-platform is also committed to finding a sustainable approach for future experiments.

> blocking the rest of the layout for all screens on Calypso just to check for the value of an experiment that runs only on two sections

I thought this issue (returning `null` on all screens) is addressed by evaluating `shouldLoadDomainExperiment` (checking for the two sections) in this [condition](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-00073ede4cd5690e0d36147c6bfd9f6e541539a11a8bec5dd8c4d3e36a48ffa2R71).

> Have we considered trying to provide the experiment value from the server side

Yes, the backend approach is detailed by @paulopmt1 in https://github.com/Automattic/wp-calypso/pull/73017, the only caveat being that all Calypso users will be assigned to the experiment instead of doing this based on intent (accessing page `/domains/add` or `/plans/yearly`). If you find it cleaner and more performant, it can work with setting an exposure event (Tracks event) in ExPlat to filter out ineligible users. Then we should definitely document this as best practice 🤓

tyxla: > Yes, the backend approach is detailed by @paulopmt1 in https://github.com/Automattic/wp-calypso/pull/73017, the only caveat being that all Calypso users will be assigned to the experiment instead of doing this based on intent (accessing page /domains/add or /plans/yearly). If you find it cleaner and more performant, it can work with setting an exposure event (Tracks event) in ExPlat to filter out ineligible users. 

That sounds like a very reasonable approach 👍 Any downsides if we decide to go with it from the explat PoV?

> Then we should definitely document this as best practice

💯 Yes, please! This sounds worthy for both Calypso docs and p2.

jessie-ross: (Sorry there was also a convo happening in slack around this:
p1675758293036799/1675715733.464879-slack-CKZHG0QCR)

paulopmt1:
We decided to move forward (Slack thread: p1675771334208239/1675715733.464879-slack-CKZHG0QCR) using [the backend strategy](https://github.com/Automattic/wp-calypso/pull/73017), so we don't need to load any unnecessary code in pages which won't use the experiment, and we'll use the calypso_upgrade_nudge_cta_click as our exposure event.

Thanks everyone for your reviews, and @jessie-ross, @tyxla, and @razvanpapadopol  for detailed instructions and code suggestions.

tyxla:
Thanks @paulopmt1, I appreciate you taking the path to a more sustainable approach 🙌

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7791902).

PR #72936, opened 2023-02-02
Merge trunk to branch docker image builds

noahtallen:
### Proposed Changes
In #72687, we update the webpack cache whenever it invalidates on trunk. To fully utilize this, a PR's code needs to be up-to-date with trunk. Otherwise, the PR can contain an older file which will also cause the webpack cache to invalidate. (Which adds 2 minutes to the build). 

This does not run on trunk, and merges are skipped if there are conflicts to avoid blocking developers.

### Testing
Check the TeamCity output.

### DIFF
--- a/.teamcity/_self/projects/WebApp.kt
+++ b/.teamcity/_self/projects/WebApp.kt
@@ -110,7 +110,7 @@ object BuildDockerImage : BuildType({
 			"""
 		}
 
-		// We want calypso.live and e2e tests to run even if there's a merg conflict,
+		// We want calypso.live and Calypso e2e tests to run even if there's a merge conflict,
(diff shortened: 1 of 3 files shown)


.teamcity/_self/projects/WebApp.kt
```diff
@@ -110,6 +110,12 @@ object BuildDockerImage : BuildType({
 			"""
 		}
 
+		// We want calypso.live and e2e tests to run even if there's a merg conflict,
```
worldomonation:
```suggestion
		// We want calypso.live and Calypso e2e tests to run even if there's a merge conflict,
```

worldomonation (approved):

Issue #61318, opened 2022-02-21
Reader: Improve keyboard navigation accessibility


tyxla:
### Quick summary

While reviewing #61112, @pento noticed that there are several accessibility issues with keyboard navigation around the reader. 

This issue quotes @pento's thoughts from https://github.com/Automattic/wp-calypso/pull/61112#pullrequestreview-883931745.

### Steps to reproduce

When tabbing through tab-able elements in the reader list view, those elements can only be triggered by pressing `space`, pressing `enter` opens the currently selected post. However, the equivalent elements can be triggered with `space` or `enter` in the single post view. I noticed this with the Comment and Like buttons, but there may be others.

The currently tabbed element has no relation to the current post in the reader list view. This has two noticeable effects:
- After using `j` to move down a few posts, the tabbed element is now several posts behind, you have to press `tab` a bunch to catch up.
- When you open a single post from the reader list view, then close that post again, the currently tabbed element is reset to the start of the page, but the current post stays as it was prior to it being opened.

### What you expected to happen

Keyboard navigation should be improved.

### What actually happened

This is described in the steps to reproduce

### Context

Discovered by @pento while reviewing #61112.

PR #72940, opened 2023-02-03
Improve reader focus management when using "J" and "K" keyboard navigation

roo2:
#### Proposed Changes
Part of https://github.com/Automattic/wp-calypso/issues/61318

When navigating with "J" and "K" on the keyboard, update the focus to the first anchor in the selected post.

https://user-images.githubusercontent.com/22446385/216486493-e5d6ff8a-ab18-4bf0-9453-6a1a07e3aea4.mov

I haven't included the second part of #61318, as the focus will be moved when they next use the "J" key to change selection.
> When you open a single post from the reader list view, then close that post again, the currently tabbed element is reset to the start of the page, but the current post stays as it was prior to it being opened.

#### Testing Instructions
go too /read and use the "J" key to navigate through the posts. 
Press the TAB key, and see that the focus has been moved to the start of the selected post

### DIFF
--- a/client/reader/stream/index.jsx
+++ b/client/reader/stream/index.jsx
@@ -125,6 +125,7 @@ class ReaderStream extends Component {
 
 		if ( ! keysAreEqual( selectedPostKey, this.props.selectedPostKey ) ) {
 			this.scrollToSelectedPost( true );
+			this.focusSelectedPost( this.props.selectedPostKey );
 		}
 
 		if ( this.props.shouldRequestRecs ) {
@@ -134,6 +135,23 @@ class ReaderStream extends Component {
 			} );
 		}
 	}
+
+	focusSelectedPost = ( selectedPostKey ) => {
+		const postRefKey = this.getPostRef( selectedPostKey );
+		const ref = this.listRef.current.refs[ postRefKey ];
+		const node = ReactDom.findDOMNode( ref );
+
+		// if the post is found, focus the first anchor tag within it.
+		if ( node ) {
+			const firstLink = node.querySelector( 'a' );
+
+			if ( firstLink ) {
+				setTimeout( () => {
+					firstLink.focus();
+				} );
+			}
+		}
+	};
 
 	_popstate = () => {
 		if ( this.props.selectedPostKey && window.history.scrollRestoration !== 'manual' ) {
--- a/client/reader/stream/index.jsx
+++ b/client/reader/stream/index.jsx
@@ -138,7 +138,7 @@ class ReaderStream extends Component {
 
 	focusSelectedPost = ( selectedPostKey ) => {
 		const postRefKey = this.getPostRef( selectedPostKey );
-		const ref = this.listRef.current.refs[ postRefKey ];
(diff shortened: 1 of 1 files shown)


client/reader/stream/index.jsx
```diff
@@ -135,6 +136,23 @@ class ReaderStream extends Component {
...
+		const node = ReactDom.findDOMNode( ref );
+
+		// if the post is found, focus the first anchor tag within it.
+		if ( node ) {
+			const firstLink = node.querySelector( 'a' );
+
+			if ( firstLink ) {
+				setTimeout( () => {
+					firstLink.focus();
+				} );
```
tyxla: Is the timeout necessary?

client/reader/stream/index.jsx
```diff
@@ -135,6 +136,23 @@ class ReaderStream extends Component {
 		}
 	}
 
+	focusSelectedPost = ( selectedPostKey ) => {
+		const postRefKey = this.getPostRef( selectedPostKey );
+		const ref = this.listRef.current.refs[ postRefKey ];
```
tyxla: `postRefKey` could theoretically be `null` because `keyToString()` could return `null`. Let's make sure to check for that before we try to reference `postRefKey`.

tyxla: `this.listRef.current` can also be `undefined`, let's make sure to check it before using it. Could be with optional chaining

client/reader/stream/index.jsx
```diff
@@ -135,6 +136,23 @@ class ReaderStream extends Component {
...
+		const node = ReactDom.findDOMNode( ref );
```
tyxla, line 142: Given that `ref` can be `undefined` based on the above facts, we should make sure it's not before we pass it to `React.findDOMNode`.

tyxla (commented):
This works great in my testing 👍 

I've left some feedback and questions, let me know what you think.

Thanks, it's a great improvement 🚀

client/reader/stream/index.jsx
roo2: I thought it was necessary but retested and it isn't 👍

client/reader/stream/index.jsx
roo2: I don't think that check is necessary though because if postRefKey is `undefined` it will just not find a ref in the list and won't throw an error.

client/reader/stream/index.jsx
roo2: but if it's undefined in that case, it won't throw an error, it will just not find a node, so I think the check for node on the next line is enough.

client/reader/stream/index.jsx
roo2: good catch!

roo2:
Thanks for the review @tyxla! I went and hacked the code to test what would happen in case of `undefined` for those cases, I think some are ok. I fixed the other feedback.

tyxla (approved):
Works well and the code looks good! Thanks again for the great improvement 🚀

PR #72959, opened 2023-02-03
Theme: Fix DIFM link in empty search (take 2)

mmtr:
#### Proposed Changes

Fixes the DIFM link on the empty search card for the logged-out theme showcase

#### Testing Instructions

- Use the Calypso live link below on an incognito window
- Go to `/themes`
- Enter a term that doesn't produce any result
- Click on the DIFM option
- Make sure it redirects to the DIFM page


dsas (approved):
Works on calypso live 😬

PR #72960, opened 2023-02-03
Plans section: Modernized header and navbar / part 1

chriskmnds:
### Proposed Changes

Partially addresses https://github.com/Automattic/martech/issues/1463

This brings the "modernized" header & navbar by default to `/plans/my-plan` & `/plan` (not behind a feature flag). This is another temp solution with follow-ups in TODO list below.

_Notes:_ Plans header & navigation are used for [jetpack-plans](https://github.com/Automattic/wp-calypso/tree/trunk/client/my-sites/plans/jetpack-plans) code, but the changes should not affect these (something for a follow-up)

#### Media

with some of the states I notice for the plans pages (info & upsell notices):

_Old `/plans`_

Screenshot 2023-02-06 at 4 25 02 PM

_New `/plans`_

Screenshot 2023-02-06 at 4 23 42 PM

_New `/plans` free site_

Screenshot 2023-02-06 at 5 12 03 PM

_`/plans/my-plan`_

Screenshot 2023-02-06 at 4 24 30 PM

### Testing Instructions

- Navigate to `/plans/[site_slug]?flags=onboarding/2023-pricing-grid`
- Confirm the changes
- Do the same without the flag `/plans/[site_slug]?flags=-onboarding/2023-pricing-grid`
- Confirm no changes in `jetpack-plans` (and please post instructions in a comment on how to test that while at it 😄)

### TODO

- [x] Extract into a layout component to allow reuse | [issue](https://github.com/Automattic/wp-calypso/issues/73102)
- [x] Extract navigation into a component to allow reuse | [issue](https://github.com/Automattic/wp-calypso/issues/73102)
- [x] Consider navbar inside header to reduce some of the complexity | [issue](https://github.com/Automattic/wp-calypso/issues/73102)

### Pre-merge Checklist

Related to Automattic/martech#1463


chriskmnds:
@lucasmendes-design for design review :-)

ddc22:
Ah this looks elegant 💯 @chriskmnds

lucasmendes-design:
Hi there, 

**Monthly:**

When I selected monthly, the copy is "Save 56% by paying annually", but I believe It should be "per month".

<br class="Apple-interchange-newline">
image

**Visitor stats:**

Sometimes, this feature is misaligned. I checked again and it's fine 🤔 

image
image



**Sticky header:** 

Are we going to have a sticky header in Calypso? 

**Check** 

I checked on the design and the y'all did the right thing but looking at the plans comparison grid right now. I think the check ✔ mark is too big. Is it simple to change from 24px to 20px? We can also upload that on different pages. 

image


**Chevron icon**

Is it possible to make the chevron more prominent? I think the current one is to subtle. 

image


**Tooltip**

Matias asked for the tooltips for the logos as well. It's missing the tooltip for Jetpack, WP Cloud, VIP and Woo. 

image

ddc22:
@lucasmendes-design The generic feedback you have raised is important but needs to be [raised elsewhere](https://github.com/orgs/Automattic/projects/552/views/1). This PR is focussed on the header area as shown below. Since there is not much feedback on that area, I hope we can assume that the design is pixel perfect (for the exact area being worked on below)?
image

I will carry this feedback to a [ticket](https://github.com/Automattic/martech/issues/1479), but, going forward, please do raise generic feedback on the [task board](https://github.com/orgs/Automattic/projects/552/views/1)  :)

client/my-sites/plans/current-plan/index.jsx
```diff
@@ -232,71 +233,75 @@ class CurrentPlan extends Component {
...
-						</Notice>
-					) }
-
-					{ legacyPlanNotice( eligibleForProPlan, selectedSite ) }
-
-					{ isEcommerceTrial ? this.renderEcommerceTrialPage() : this.renderMain() }
-
-					<TrackComponentView eventName="calypso_plans_my_plan_view" />
+				{ selectedSiteId && (
+					// key={ selectedSiteId } ensures data is refetched for changing selectedSiteId
```
southp: I know it's been this way for ages, but let's remove this since it doesn't make sense anymore.

client/my-sites/plans/main.jsx
```diff
@@ -184,38 +187,39 @@ class Plans extends Component {
...
+					<div>
+						<PlansHeader />
+						<div id="plans" className="plans plans__has-sidebar">
+							<PlansNavigation path={ this.props.context.path } modernized={ true } />
+							<Main
+								fullWidthLayout={ is2023OnboardingPricingGrid }
+								wideLayout={ ! is2023OnboardingPricingGrid }
+							>
+								{ domainAndPlanPackage && <DomainAndPlanUpsellNotice /> }
+								{ ! canAccessPlans && (
```
southp: This section is nested under `canAccessPlans && (...`, which means it will never be reached. From the original code, we should move this section outside of this `canAccessPlans` expression.

client/my-sites/plans/navigation.scss
```diff
@@ -0,0 +1,89 @@
...
+ * */
+
+@import "@wordpress/base-styles/breakpoints";
+
+.plans-navigation {
+	&.is-modernized {
+		box-shadow: inset 0 -1px 0 #0000000d;
+		background-color: var(--studio-white);
+
+		// .section-nav__panel {
```
southp: If there is no good reason for keeping it, let's remove this commented code block :)

southp (commented):
Outside of the inline comments, here are a few more things I'd like to raise:

Comparing My Plans and Plans tab, the former has two horizontal lines while the latter has only one. Which one is the intended result?
截圖 2023-02-07 上午11 05 29
截圖 2023-02-07 上午11 05 53

Testing the Jetpack plans page, the Plans tab is unaffected while My Plan tab is affected, presenting inconsitent look upon switching tabs:

https://user-images.githubusercontent.com/1842898/217161882-0f36268a-6454-470e-9219-1cf64fc35adb.mov

Finally, the idea of `ModernizedLayout` component is quite intersting and works in general. However, I'm a bit confused that we have two ways of adding the "modernized" classname: either adding it directly to the component like `PlansHeaders` does or conditionally adding it through a prop like `PlansNavigation` does. I'm wondering if we can make it more consistent? This is more of a question than a blocking comment, so please feel free to keep it the way it is :)

southp:
Ohhhh right, I forgot about this:

> Confirm no changes in jetpack-plans (and please post instructions in a comment on how to test that while at it 😄)

Here goes:

1. Go to https://jurassic.ninja to create a Jetpack site.
2. Once it's done, go to its wp-admin and access the Jetpack dashboard there. 
3. There, you should be able to find a button to connect the site to your WordPress.com account. Click it and follow the instructions.
4. Go back to calypso, and click "switch site" to change the selected site to this Jetpack site.
5. Now the plans tab will present you with `jetpack-plans` component :)

client/my-sites/plans/modernized-layout.tsx
```diff
@@ -0,0 +1,150 @@
...
+	const backgroundColor = '#fdfdfd';
+	const sectionMaxWidth = '1224px';
+	const layoutContentPaddingTop = '79px';
+	const sidebarAppearanceBreakPoint = '783px';
+	// from @automattic/components/src/highlight-cards/variables.scss:
+	const customMobileBreakpoint = '660px';
+	const verticalMargin = '32px';
+
+	const globalOverrides = css`
+		.${ subSection ?? section }__section-header {
```
chriskmnds: These variable class names are (ought to be) really temp stuff. Ideally, what's mapped on the front e.g. `.plans__section-header` should be searchable in the code base.

client/my-sites/plans/main.jsx
chriskmnds: Good catch :-)

client/my-sites/plans/modernized-layout.tsx
chriskmnds: refactored back to static classes for now

chriskmnds:
@southp @jdc91 @lucasmendes-design Thanks for the thorough reviews, and the detailed JP instructions. I believe I've addressed your concerns, but pls do take another look. I'll do few more tests on my end.

> Comparing My Plans and Plans tab, the former has two horizontal lines while the latter has only one. Which one is the intended result?

There should only be one. Just clumsy me :p

> Finally, the idea of ModernizedLayout component is quite intersting and works in general. However, I'm a bit confused that we have two ways of adding the "modernized" classname: either adding it directly to the component like PlansHeaders does or conditionally adding it through a prop like PlansNavigation does. I'm wondering if we can make it more consistent? This is more of a question than a blocking comment, so please feel free to keep it the way it is :)

Yeah, it's a bit of a soup. I think I wanted to keep them closer to how they were defined elsewhere (one being part of the nav component styles, the other being actual section overrides), but that idea was short-lived. I've consolidated the styles into `ModernizedLayout` now and removed the smart classes along. It's all still temporary. Ideally, we'd want to follow up and extract these overrides/styles to a place and form that are convenient for reuse in other sections (and update all the sections to sync! 😬)

client/my-sites/plans/main.jsx
```diff
@@ -186,36 +189,36 @@ class Plans extends Component {
 		const isEcommerceTrial = currentPlanSlug === PLAN_ECOMMERCE_TRIAL_MONTHLY;
 
 		return (
-			<div className={ is2023OnboardingPricingGrid ? 'is-2023-pricing-grid' : '' }>
+			<div>
+				<ModernizedLayout dropShadowOnHeader={ isFreePlan( currentPlanSlug ) } />
```
chriskmnds: should probably check for Jetpack here also, for sanity

client/my-sites/plans/current-plan/index.jsx
```diff
@@ -329,5 +334,6 @@ export default connect( ( state, { requestThankYou } ) => {
 		showThankYou: requestThankYou && isJetpackNotAtomic,
 		scheduleId: getConciergeScheduleId( state ),
 		eligibleForProPlan,
+		isJetpack,
```
ddc22: This will also flag atomic sites as Jetpack sites and the Modernized header will not be visible. 

image


I think the flag on line 318 accounts for this and can be used for our use case as well.

https://github.com/Automattic/wp-calypso/blob/fd4d73e86ebc9dc8af48c2fd0134406ae7b4094c/client/my-sites/plans/current-plan/index.jsx#L318

ddc22 (changes requested):
Overall looks good but a minor fix is required for Atomic sites.

lucasmendes-design:
Hi there, there is one detail that I believe we can fix it. 

On this page, the common alignment is center-align. To give an example here is how stats and themes page look like with the new header: 

**Stats page:**
image

Themes page:
image

I would change the size of the notification to follow the size of the plans grid and the header itself. [Figma file](https://www.figma.com/file/xkhhUlDowF13dRoXJqlRDM?node-id=1:95#363744244)

**1024px**
image

**1440px**
image

**Notification:**
image

chriskmnds:
> I would change the size of the notification to follow the size of the plans grid and the header itself.

@lucasmendes-design I'm not exactly sure how that's _not_ the case right now. Notification seems to be aligned with the same padding, so it follows the size of the header? 🤔 

Screenshot 2023-02-08 at 11 12 01 AM

client/my-sites/plans/current-plan/index.jsx
chriskmnds: Ah! Good catch. totally ignored this.

chriskmnds: (and funnily enough, I was testing with an atomic site to confirm the jetpack condition at some point. so I'm the worst! 😆 )

client/my-sites/plans/main.jsx
```diff
@@ -241,5 +248,6 @@ export default connect( ( state ) => {
 		showTreatmentPlansReorderTest: isTreatmentPlansReorderTest( state ),
 		plansLoaded: Boolean( getPlanSlug( state, getPlan( PLAN_FREE )?.getProductId() || 0 ) ),
 		is2023OnboardingPricingGrid,
+		isJetpackNotAtomic: isJetpackSite( state, selectedSiteId, { treatAtomicAsJetpackSite: false } ),
```
chriskmnds, line 251: Apparently, there's a `options.treatAtomicAsJetpackSite` parameter that can be set on this selector (and avoid having to [combine two selectors](https://github.com/Automattic/wp-calypso/pull/72960/files#diff-9a7bb16689c221cf314662dc7929cea29ae46431f9bc11fd4123e233b0b208d2R317-R318) for it as in the other file). Not sure I'm missing something though. @jdc91

chriskmnds:
@jdc91 Thanks for reviewing. Sorry that I'm dragging you into spotting errors for me 😬 

Let's give this another review pls, I also noticed console errors with the CSS parts that needed to resolve. will do a few more tests on my end

client/my-sites/plans/main.jsx
ddc22: Nice this is a better approach!

ddc22 (approved):
LGTM :shipit:

lucasmendes-design:
Hi there, Dustin Reagon mention [that here (Slack). ](https://a8c.slack.com/archives/C02FMH4G8/p1675875224401949)

The problem with this layout is that the header with the copy "Plans" and "See and compare features available on each WordPress.com plan" is misaligned. 
image

Everything should be center-aligned, not only the plans grid. Domains example below:
image

cc: @jdc91

PR #72968, opened 2023-02-03
Removing monthly toggle from paid media signup flow

donlair:
#### Proposed Changes

* Run an A/B test that removes the monthly subscription option for paid media traffic. Context is at pcbrnV-4Pi-p2.
* Creates separate components for the paid media version of the plans step (`plans-pm`). I chose to implement separate components to avoid adding yet another use case to the display logic, especially since the timing overlaps with the broader plans page overhaul.
* Creates a new `util` file that contains functions shared by the `PlansStep` and `PlansStepPM` components.
* The test is implemented using Explat, experiment name `paid_media_signup_2023_02_hide_monthly`
* Experiment post is available at pbxNRc-2jS-p2.

#### Testing Instructions
* Checkout this PR and start Calypso locally.
* Open the Abacus page for this experiment (`paid_media_signup_2023_02_hide_monthly`) and assign yourself to either the control or treatment variation.
* Navigate to `/start/onboarding-pm`
* Select a domain and proceed to the Plans step.
* Confirm that the monthly toggle is showing for the control variation and not showing for the treatment variation.
* Confirm that the Plans step works in both mobile and desktop viewports
* Select a plan and proceed to checkout.
* Confirm that checkout works as expected.
* Start a new signup session and stress test the domains step, choosing a variety of options from that step, including the 'choose later' and the 'use a domain I already own' paths.
* Test other flows that use the `Plans` step: `onboarding`, the tailored flows, wpcc, and Calypso admin.

Related to #

### DIFF
--- a/client/lib/signup/step-actions/index.js
+++ b/client/lib/signup/step-actions/index.js
@@ -1,5 +1,10 @@
 import config from '@automattic/calypso-config';
-import { WPCOM_DIFM_LITE } from '@automattic/calypso-products';
+import {
+	WPCOM_DIFM_LITE,
+	planHasFeature,
+	FEATURE_UPLOAD_THEMES_PLUGINS,
+	isEcommerce,
+} from '@automattic/calypso-products';
 import { getUrlParts } from '@automattic/calypso-url';
 import { Site } from '@automattic/data-stores';
 import { isBlankCanvasDesign } from '@automattic/design-picker';
@@ -39,7 +44,6 @@ import { getSiteType } from 'calypso/state/signup/steps/site-type/selectors';
 import { getWebsiteContent } from 'calypso/state/signup/steps/website-content/selectors';
 import { requestSite } from 'calypso/state/sites/actions';
 import { getSiteId } from 'calypso/state/sites/selectors';
-
 const Visibility = Site.Visibility;
 const debug = debugFactory( 'calypso:signup:step-actions' );
 
@@ -1164,3 +1168,79 @@ export function isNewOrExistingSiteFulfilled( stepName, defaultDependencies, nex
 		flows.excludeStep( stepName );
 	}
 }
+
+export const buildUpgradeFunction = ( planProps, cartItem ) => {
+	const {
+		additionalStepData,
+		flowName,
+		launchSite,
+		selectedSite,
+		stepName,
+		stepSectionName,
+		themeSlugWithRepo,
+		goToNextStep,
+		submitSignupStep,
+	} = planProps;
+
+	if ( cartItem ) {
+		recordTracksEvent( 'calypso_signup_plan_select', {
+			product_slug: cartItem.product_slug,
+			free_trial: cartItem.free_trial,
+			from_section: stepSectionName ? stepSectionName : 'default',
+		} );
+
+		// If we're inside the store signup flow and the cart item is a Business or eCommerce Plan,
+		// set a flag on it. It will trigger Automated Transfer when the product is being
+		// activated at the end of the checkout process.
+		if (
+			flowName === 'ecommerce' &&
+			planHasFeature( cartItem.product_slug, FEATURE_UPLOAD_THEMES_PLUGINS )
+		) {
+			cartItem.extra = Object.assign( cartItem.extra || {}, {
+				is_store_signup: true,
+			} );
+		}
+	} else {
+		recordTracksEvent( 'calypso_signup_free_plan_select', {
+			from_section: stepSectionName ? stepSectionName : 'default',
+		} );
+	}
+
+	const step = {
+		stepName,
+		stepSectionName,
+		cartItem,
+		...additionalStepData,
+	};
+
+	if ( selectedSite && flowName === 'site-selected' && ! cartItem ) {
+		wpcom.req.post(
+			`/domains/${ selectedSite.ID }/${ selectedSite.name }/convert-domain-only-to-site`,
+			{},
+			( error ) => {
+				if ( error ) {
+					errorNotice( error.message );
+					return;
+				}
+				submitSignupStep( step, {
+					cartItem,
+				} );
+				goToNextStep();
+			}
+		);
+		return;
+	}
+
+	const signupVals = {
+		cartItem,
+		...( themeSlugWithRepo && { themeSlugWithRepo } ),
+		...( launchSite && { comingSoon: 0 } ),
+	};
+
+	if ( cartItem && isEcommerce( cartItem ) ) {
+		signupVals.themeSlugWithRepo = 'pub/twentytwentytwo';
+	}
+
+	submitSignupStep( step, signupVals );
+	goToNextStep();
+};
--- a/client/lib/signup/step-actions/index.js
+++ b/client/lib/signup/step-actions/index.js
@@ -1180,10 +1180,11 @@ export const buildUpgradeFunction = ( planProps, cartItem ) => {
 		themeSlugWithRepo,
 		goToNextStep,
 		submitSignupStep,
+		recordTracksEvent: rte,
 	} = planProps;
 
 	if ( cartItem ) {
-		recordTracksEvent( 'calypso_signup_plan_select', {
+		rte( 'calypso_signup_plan_select', {
 			product_slug: cartItem.product_slug,
 			free_trial: cartItem.free_trial,
 			from_section: stepSectionName ? stepSectionName : 'default',
@@ -1201,7 +1202,7 @@ export const buildUpgradeFunction = ( planProps, cartItem ) => {
 			} );
 		}
 	} else {
-		recordTracksEvent( 'calypso_signup_free_plan_select', {
+		rte( 'calypso_signup_free_plan_select', {
 			from_section: stepSectionName ? stepSectionName : 'default',
 		} );
 	}
--- a/client/lib/signup/step-actions/index.js
+++ b/client/lib/signup/step-actions/index.js
@@ -1180,11 +1180,10 @@ export const buildUpgradeFunction = ( planProps, cartItem ) => {
 		themeSlugWithRepo,
 		goToNextStep,
 		submitSignupStep,
-		recordTracksEvent: rte,
 	} = planProps;
 
 	if ( cartItem ) {
-		rte( 'calypso_signup_plan_select', {
+		planProps.recordTracksEvent( 'calypso_signup_plan_select', {
 			product_slug: cartItem.product_slug,
 			free_trial: cartItem.free_trial,
 			from_section: stepSectionName ? stepSectionName : 'default',
@@ -1202,7 +1201,7 @@ export const buildUpgradeFunction = ( planProps, cartItem ) => {
 			} );
 		}
 	} else {
-		rte( 'calypso_signup_free_plan_select', {
+		planProps.recordTracksEvent( 'calypso_signup_free_plan_select', {
 			from_section: stepSectionName ? stepSectionName : 'default',
 		} );
 	}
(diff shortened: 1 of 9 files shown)


client/signup/steps/plans/index.jsx
```diff
@@ -439,8 +440,10 @@ export class PlansStep extends Component {
 	}
 
 	render() {
+		const { isPaidMediaFlow } = this.props;
 		const is2023OnboardingPricingGrid = is2023PricingGridEnabled();
 
+
```
southp: This redundant newline is causing [the code style check](https://teamcity.a8c.com/viewLog.html?buildId=9475018&tab=Inspection&buildTypeId=calypso_calypso_WebApp_Check_Code_Style_Branch) to complain :) 

image

southp (changes requested):
I've checked both the control variation and the treatment one and they both work as expected. However, I only tested with a logged in user; somehow I didn't manage to force a variation when I was logged-out. I'm wondering if it's the same `tk_ai` cookie issue that we saw before. In that case, it won't be an issue in production. @aaronyan could you help me confirm?

As explained in the PR description, overall the code contains two parts:

1. The A/B test of removing the monthly/yearly toggle.
2. Create a separate plans step from the primary one as a foundation for future tweaking.

The 1st is quite intuitive so I have no concern for that. However, we can architect the 2nd better :) Currently the new plan step component is being returned within the current primary step component. That means no matter which flow we go, `/start/` or `/start/onboarding-pm`, both components will be loaded. That's why [the section size](https://github.com/Automattic/wp-calypso/pull/72968#issuecomment-1419764835) has been increased quite a bit:
image

Since we've amended the original plans step component into a leaner one anyway, it would be much cleaner if we completely decouple `plans` and `plans-pm` components. Those shared callback functions can be then extracted accordingly, e.g. as utility functions shared in a module, etc.

donlair:
messed up that rebase

donlair:
Thanks for the feedback @southp, great suggestions! This is ready for another look when you get a chance :).

aaronyan:
> I've checked both the control variation and the treatment one and they both work as expected. However, I only tested with a logged in user; somehow I didn't manage to force a variation when I was logged-out. I'm wondering if it's the same tk_ai cookie issue that we saw before. In that case, it won't be an issue in production. @aaronyan could you help me confirm?

I would recommend only running the experiment for logged-in users, is that possible? We recently disabled (https://github.com/Automattic/experimentation-platform/issues/814) support for the Logged-Out Home Page due to caching issues on the Edge browser. Keeping the experience only for logged-in users also protects against crossover effects (logged-out users that clear cache, view from a different device, etc is easier to see both `control` and `treatment`).

donlair:
> I would recommend only running the experiment for logged-in users, is that possible? We recently disabled ([Automattic/experimentation-platform#814](https://github.com/Automattic/experimentation-platform/issues/814)) support for the Logged-Out Home Page due to caching issues on the Edge browser. Keeping the experience only for logged-in users also protects against crossover effects (logged-out users that clear cache, view from a different device, etc is easier to see both `control` and `treatment`).

@aaronyan That's not practical for this experiment because the vast majority of users who see this will be logged out (coming from paid ads). I'll take a look at the issue you linked to, but seems like it's related to the home page so I'll highlight that this traffic won't come from LOHP (this signup flow is only linked to from paid media landing pages).

aaronyan:
I caught up with @donlair about the specifics of this experiment! Users that arrive anonymously will provide an email to "log in", and then they are later shown the experiment variations that occur on the plan page. As long as the variation assignment occurs after the user provides an email address, then the experiment is considered logged-in and you don't need to test for anonymous user assignment!

client/signup/steps/plans/index.jsx
```diff
@@ -76,86 +74,6 @@ export class PlansStep extends Component {
...
-			flowName,
-			themeSlugWithRepo,
-			launchSite,
-			goToNextStep,
-		} = this.props;
-
-		if ( cartItem ) {
-			this.props.recordTracksEvent( 'calypso_signup_plan_select', {
-				product_slug: cartItem.product_slug,
-				free_trial: cartItem.free_trial,
```
southp, line 94: This line is missing in the newly extracted `buildUpgradeFunction()`. Are we sure about removing it?

southp (commented):
Thanks for the fix, @donlair . It looks very close now. 
Except for the missing line of `free_trial`, [the unit tests is now broken](https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Run_All_Unit_Tests/9487213?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildTestsSection=true) and have to be fixed.

I also have a slight concern regarding using the expression that returning an anonymoous function right in the prop. [It used to be an issue](https://stackoverflow.com/questions/42280526/how-unperformant-are-anonymous-functions-in-react-component-attributes), but I have a vague memory that React has done some clever things making it no longer the case.  @jdc91 Could you help me confirm?

Also, one thing I forgot to confirm with you: would this flow need the new pricing grid that we are launching next week? 

Lastly, I'd like to share one more alternative strategy here in case you'd be interested. Except for extracting the utility functions, one more way is to inheret the original plans component and only override the part that is needed. I didn't mention this previously because I had the feeling that there are intended clean-up happened to `plans-pm`; just wanted to share in case it inspires some more thoughts :)

client/signup/steps/plans/index.jsx
```diff
@@ -246,9 +152,9 @@ export class PlansStep extends Component {
 					hideEcommercePlan={ this.shouldHideEcommercePlan() }
 					isInSignup={ true }
 					isLaunchPage={ isLaunchPage }
-					intervalType={ this.getIntervalType() }
-					onUpgradeClick={ this.onSelectPlan }
-					domainName={ this.getDomainName() }
+					intervalType={ getIntervalType() }
+					onUpgradeClick={ this.onSelectPlan() }
```
ddc22: I think a lot of the callbacks that are returned can be removed if we switch to the approach below

```javascript
	onUpgradeClick={ (cartItem)=>  buildUpgradeFunction(planProperties, cartItem) }
```

client/signup/steps/plans/util.ts
```diff
@@ -0,0 +1,130 @@
...
+	// Default value
+	return 'yearly';
+};
+
+export const getDomainName = ( domainItem: DomainItem ): string | undefined => {
+	return domainItem && domainItem.meta;
+};
+
+export const buildUpgradeFunction = (
+	planProps: PlanProperties
```
ddc22: Here we can just accept two parameters.

client/signup/steps/plans-pm/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+			return translate( 'Choose a plan' );
+		}
+
+		return translate( "Pick a plan that's right for you." );
+	}
+
+	getSubHeaderText() {
+		const { translate } = this.props;
+		const freePlanButton = (
+			<Button onClick={ handleFreePlanButtonClick( this.props ) } borderless />
```
ddc22: This can be changed to 
```javascript
     <Button onClick={ ()=> buildUpgradeFunction( this.props, null ) } borderless />
```

ddc22 (commented):
@southp I am not sure about the caveat related to anonymous functions.

However @donlair I think we can simplify the code by avoid returning callbacks within functions specially if there is no async tasks that are involved that require a future value of a function. (Even then we can often use async await :).

I have left some related suggestions below.

client/signup/steps/plans/util.ts
```diff
@@ -0,0 +1,130 @@
...
+
+	// Default value
+	return 'yearly';
+};
+
+export const getDomainName = ( domainItem: DomainItem ): string | undefined => {
+	return domainItem && domainItem.meta;
+};
+
+export const buildUpgradeFunction = (
```
ddc22: If we are moving this function out of the component it should ideally live in step-actions. This is sort of where the Async Api Calls live in relation to Steps.

https://github.com/Automattic/wp-calypso/blob/cf7823160cc6baf87f4631051b0127dcd44971d2/client/lib/signup/step-actions.js

client/signup/steps/plans/index.jsx
donlair: That's a nice solution, I'll give it a shot

client/signup/steps/plans/index.jsx
donlair: Afaik there aren't any free trials floating around, but I added it back in just in case!

donlair:
> Also, one thing I forgot to confirm with you: would this flow need the new pricing grid that we are launching next week?

@southp I'd like to hold off on launching the new pricing grid in this paid media signup flow until we get a chance to look at the performance. That'll let us be more cautious for the traffic that costs us money.

donlair:
Thanks @southp @jdc91, this version of `buildUpgradeFunction` feels much better to me, thanks for the suggestion! I believe I've addressed all feedback so this is ready for another look when you get a chance!

donlair:
@southp @jdc91 the tests are failing now because I refactored out the onSelectPlan function - would it be better to update the tests to account for that change, or rearrange it to retain the onSelectPlan function that the tests expect to be there?

client/signup/steps/plans/index.jsx
```diff
@@ -76,86 +70,6 @@ export class PlansStep extends Component {
 		this.unsubscribe();
 	}
 
-	onSelectPlan = ( cartItem ) => {
```
ddc22, line 80: It's unfortunate that the tests are failing @donlair, however, I think a good compromise would be to retain this function.
And call the function `buildUpgradeFunction`,  within this function. 

It's a bit dirty and I would rather move all those related tests to `client/lib/signup/step-actions.js` but the above fix should also work without increasing the workload even more.

southp (commented):
@donlair It looks pretty good now. Thanks for the great effort. Please let me know once you've resolved the code conflict, and I'll come for a final checking 🙇🏼

donlair:
Thanks @southp and @jdc91, this is ready for another look - I rebased this and fixed one remaining unit test that was failing.

client/lib/signup/step-actions/index.js
```diff
@@ -1180,10 +1180,11 @@ export const buildUpgradeFunction = ( planProps, cartItem ) => {
 		themeSlugWithRepo,
 		goToNextStep,
 		submitSignupStep,
+		recordTracksEvent: rte,
 	} = planProps;
 
 	if ( cartItem ) {
-		recordTracksEvent( 'calypso_signup_plan_select', {
+		rte( 'calypso_signup_plan_select', {
```
southp: It's in general uncommon to abbreviate a function this way in calypso. When it happens, it's common to just use `this.props.recordTracksEvent()` to work around the linting rule, e.g. https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plans/current-plan/jetpack-checklist/index.tsx#L131. 

Let's use a more verbose form here :)

southp (approved):
LGTM 👍🏼 I've left one final inline comment regarding the naming convention. I do think it needs to be addressed, but it's not so huge that would require another round of review. Thus, please feel free to ship this once it's done.

One more deeper code structure wise thing I'd like to note is that `step-actions/index.js` now have two versions of `recordTracksEvent()`: one imported from `calypso/lib/analytics/tracks` and one from a prop. This won't be a big deal for now or even in a forseeable future, so it shouldn't block this PR. However, at some points in the future we should make sure we only use one version of the same functionality in this file.

PR #72975, opened 2023-02-03
Docker webpack cache changes

noahtallen:
This restores the changes we had to revert yesterday, since we ran out of time. Will do more thorough testing on this branch before merging it next week. See more in https://github.com/Automattic/wp-calypso/pull/72939. In particular, see a long description about this in https://github.com/Automattic/wp-calypso/pull/72687!

General idea is that when the webpack cache totally invalidates, it costs at least 2 minutes in the build. What was a 3-4 minute build is now 6 minutes. The cache invalidates when several files (dependencies of the webpack config) change. These files change several times a day, but we can only build the base image a handful of times throughout the day since it's a big build.

So, why not update the cache whenever it invalidates? We accomplish that by detecting an invalidation. When we do, we add a layer to the base docker image with the updated webpack cache. We push the base image to latest, meaning it'll be used for subsequent builds.

Note that this behavior only happens on trunk, and happens after sending a webhook to the deploy system, so it shouldn't block deploys.

To fully take advantage of this, we'll need to merge in the trunk branch in PR builds so that they don't contain outdated files which would invalidate the cache. See #72936 for that.

### Todo:
- [x] Correctly disable readonly cache plugin. (This is why we had to revert, since it wasn't working at all without this change)
- [x] Figure out how to better use the layer cache. If you look at https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/9428035, it ran webpack twice. It doesn't need to since the layer which builds webpack should be cached. This adds 5 minutes or more to the build, which is really expensive. It makes the total build almost 14min.

Noting that the layer cache worked well today. I think we can merge and keep an eye on it.


.teamcity/_self/projects/WebApp.kt
```diff
@@ -4,25 +4,13 @@ import Settings
...
-import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.schedule
-import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
-import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.finishBuildTrigger
+
+import jetbrains.buildServer.configs.kotlin.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.failureConditions.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.*
```
noahtallen, line 13: Just learned we can do this

Dockerfile.base
```diff
@@ -8,7 +8,7 @@ ARG node_memory=8192
 WORKDIR /calypso
 ENV NPM_CONFIG_CACHE=/calypso/.cache
 ENV PERSISTENT_CACHE=true
-ENV READONLY_CACHE=false
+ENV GENERATE_CACHE_IMAGE=true
```
noahtallen, line 11: Changing this for consistency. It's no longer good enough to just set `READONLY_CACHE`, and it wasn't easy to get the logic working in Kotlin/TeamCity.

worldomonation (approved):

worldomonation:
Thanks for making this change in one PR for ease of monitoring and revert.

PR #59876, opened 2022-01-07
woop: uses combobox to navigate countries

Rolilink:
#### Changes proposed in this Pull Request
* Updating Country/Region dropdown to use ComboboxControl

Screen Shot 2022-01-10 at 2 25 40 PM

#### Testing instructions
* Apply D72720-code and sandbox the API.
* Navigate to `http://calypso.localhost:3000/start/woocommerce-install/store-address?site=<siteId>`
* Select a country (and region, if available) in the country dropdown
* Fill the rest and click continue
* With D72639-code it should send an http request to settings with `woocommerce_default_country` being the code of the selected country.

Related to #
* https://github.com/Automattic/wp-calypso/issues/59833

PR #72978, opened 2023-02-04
Redux: Convert countries state slice to TypeScript

sirbrillig:
#### Proposed Changes

This PR attempts to add TypeScript types to the `countries` Redux reducer slice.

In so doing this corrects an existing type bug: previously it was assumed that every state slice under `state.countries` was an array, but the `woocommerce` data slice is an object. As a result of fixing this, the reducer default has been changed; previously when `state.countries.woocommerce` was first initialized, it was `[]` like the other `countries` slices; its value was only set to an object once data had been retrieved. Now it will default to `{}`. This bug was added in https://github.com/Automattic/wp-calypso/pull/59876

#### Testing Instructions

This should mainly only change types so if the tests pass then there shouldn't be anything to do.

However, one thing that is altered by this PR is replacing lodash `get()` in the `getCountries` selector with native JS optional chaining. To verify that this doesn't cause any problems, it's probably a good idea to test that a countries list still loads. One way to test this is to visit checkout and verify that the list of countries in the billing details/contact details step shows up as expected.


client/signup/steps/woocommerce-install/step-store-address/index.tsx
```diff
@@ -48,7 +49,8 @@ export default function StepStoreAddress( props: WooCommerceStoreAddressProps )
 
 	const siteId = useSelector( getSelectedSiteId ) as number;
 
-	const countriesList = useSelector( ( state ) => getCountries( state, 'woocommerce' ) ) || [];
+	const countriesList =
+		useSelector( ( state: IAppState ) => getCountries( state, 'woocommerce' ) ) || {};
```
tyxla, line 53: Can we be certain that this default value change is going to be necessary only for the `woocommerce` subtree?

client/state/countries/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface SmsCountry {
+	code: string;
+	name: string;
+	numeric_code: string;
+	country_name: string;
+}
+
+export interface CountriesState {
+	domains: CountryListItem[] | undefined;
```
tyxla: Is there a reason to prefer ` | undefined` rather than making the property optional by appending `?` to its name?

client/state/selectors/get-countries.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Retrieves the list of countries from the specified type.
+ *
+ * The country data types are 'domains, 'payments', or 'sms'.
+ */
+export default function getCountries< TType extends keyof CountriesState >(
+	state: IAppState,
+	type: TType
+) {
+	return state.countries?.[ type ] ?? null;
```
tyxla, line 15: Since we're including the reducer (L4), `state.countries` is expected to always be there, therefore `state.countries` shouldn't need optional chaining, right?

client/state/countries/reducer.ts
```diff
@@ -0,0 +1,59 @@
...
+				return state;
+		}
+	};
+}
+
+const combinedReducer = combineReducers( {
+	// Note: it would be nice to not need to specify the second generic type
+	// here, since it can be inferred from the argument, but if you pass any
+	// generics, you must pass all generics. May possibly be solved by
+	// https://github.com/microsoft/TypeScript/pull/26349
```
tyxla: Yep, I was going to comment on this. Makes the types a bit more complex than we'd like. I wonder what other alternatives we have for simplifaction through inference here.

tyxla (commented):
This is looking great @sirbrillig, thanks! I only have a few questions.

client/state/countries/reducer.ts
pottedmeat: Why is `TActionType` needed in these functions? `updatedActionType` could just be typed string and it should all work.

pottedmeat: What do you mean it can be inferred from the argument? If you didn't pass a second value, the argument would accept all strings.

pottedmeat: I think perhaps `TActionType` is meant to be used beyond the argument, but that's the only place it's being used right now.

sirbrillig: In a previous version of the reducer factory, I was using `TActionType` to type the actions, but that code has all since been removed (the actions are now untyped strings). So actually we can just remove the second generic here entirely.

client/state/countries/types.ts
sirbrillig: I think yes, although maybe it's not super relevant here: Using `?` makes the property optional, whereas `| undefined` makes the property required (but it can be explicitly undefined). If I were to make the property optional, a type check on the combined state object would not notice if a the property were somehow deleted.

However, the `combineReducers()` function in calypso actually loses all state information because it does not accept any generics. As a result, there's no type checking of the combined state anyway and this point is kind of moot. I started trying to change this behavior and add proper state/action types to `combineReducers` in https://github.com/Automattic/wp-calypso/pull/72977 but it quickly became very tangled and I think I'm going to have to give up for now.

There's another factor as well: lots of the non-TS code that uses the `getCountries` selector does not guard for `undefined` values. Because these types are just informative and are not actually enforced by `combineReducers`, I opted to give these an `undefined` initial state to force consumers to be extra careful (for example, if the initialization was ever changed to allow undefined there would be no way to catch it in the compiler), but in practice they should never be undefined since their initial reducers all have a default value.

In summary: this could equally be `domains: CountryListItem[]` or `domains?: CountryListItem[]`. I don't have a strong preference.

client/state/selectors/get-countries.ts
sirbrillig: I did this because in the non-TS code the line was `get( state, [ 'countries', type ], null )` which corresponds to `state.countries?.[ type ] ?? null` and I didn't want to change the behavior. 

On the other hand, if the types are right and the `init` file has correctly registered the reducer, you're right that `countries` will always exist.

On the other other hand, unit tests in calypso often create a Redux store without the code that allows reducers to register themselves and leave out parts of state that they do not need. I wouldn't be surprised if removing this optional chaining could break some tests. I'll give it a try and find out.

client/signup/steps/woocommerce-install/step-store-address/index.tsx
sirbrillig: I think so. The other subtrees are currently all arrays. This default value should always have been an object but from the history it looks like when someone changed this subtree from `'payments'` to `'woocommerce'` (and thus from an array to an object) they forgot to change its default value.

To echo some of my other comments: in practice `getCountries` should never be able to return a falsy value for a known subtree. Its initializers always specify a default value. So this default (and the `null` default in the selector) is mostly academic and would probably apply only if the data returned by the API was falsy (there are no types or guards on the functions which fetch data from the API to this tree which is not great but beyond the scope of this PR).

client/state/selectors/get-countries.ts
sirbrillig: Yeah, as I suspected, [there are tests that fail](https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Run_All_Unit_Tests/9433446?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildTestsSection=true) if this is changed.

🤔 I'm not sure what the right solution is, though.

1. I could fix the failing tests to also mock the `countries` state and remove the optional chaining here.
2. I could type the `state` parameter to `state: Partial< IAppState >` as defensive coding.
3. I could make `countries` an optional property of `IAppState` since it may not always be defined if the reducer is not registered.

Honestly I think that the "correct" thing to do is to make it easier for tests to initialize the actual Redux state tree used by the application _and_ provide a default state at the same time. I was able to [jump through some hoops to do that](https://github.com/Automattic/wp-calypso/pull/72839/files#diff-3b1fead302c83b8fc4e5faab9e6193e065e2b84a656909003498c7ebe3e68b14R142-R181) recently but it's not ideal.

If we could do that, then all tests that need to mock the Redux store could do so directly without needing to [create elaborate mocks](https://github.com/Automattic/wp-calypso/blob/78d77ba0cc2009f86ba9d3a822480510cd51fbb5/client/my-sites/checkout/composite-checkout/test/util/index.ts#L760-L858) that [mock data unrelated to the test](https://github.com/Automattic/wp-calypso/blob/78d77ba0cc2009f86ba9d3a822480510cd51fbb5/client/me/purchases/manage-purchase/test/purchase-management-buttons.js#L91-L118). The tests would be more accurate to actual usage and then code like this optional chaining would be unnecessary. I've been [exploring how this might be possible](https://github.com/Automattic/wp-calypso/pull/73025) but learning how Redux is set up in calypso is taking me a while.

sirbrillig: I pushed an update to do option 3, so this should be resolved now.

sirbrillig:
Based on comments, I realized that there's a fundamental unanswered (at least to me) question about how we will type the Redux state going forward: since the state is dynamically registered and there's no current way to know if dynamic state is loaded or to enforce the types of a state slice that uses `combineReducers`, should we:

1. Type each state slice as optional (the defensive case) or 
2. Type each slice assuming dynamic reducers are already active and their initializations have complete (the most likely case)?

I tend to lean toward defensive coding (option 1) as I'd rather have unnecessary guards than fatal errors, so I've updated this PR to make each state slice optional (except `marketplace` because that would require a lot of refactoring and I don't know what it might affect). However, feel free to disagree with me and we could go in the other direction.

client/state/countries/types.ts
sirbrillig: I pushed an update to make these all optional.

client/state/selectors/get-countries.ts
tyxla: From 3), how is it possible that the `countries` reducer will not be registered? With the modular state, we've been making sure that every selector and action that make use of that state tree will include that reducer, so there should never be the scenario where that tree doesn't exist.

Furthermore, I do agree that in tests it can be difficult to get it all working with requiring all required reducer trees and their corresponding initial state. To fix that, we do need a helper and I appreciate you working on that. One idea for it could be including the `addReducerEnhancer` enhancer when creating the mock store. That will allow modular reducers to dynamically get included in the state tree as relevant selectors and actions initialize those reducers. Then the only thing we might need to provide would be initial state when necessary. 

This does go a bit beyond what the purpose of this PR is, but let me know what you think.

tyxla:
Type-wise, I think that since there's no way for TS to ensure that a certain state tree has been loaded, it should be a `Partial` of the state. The main rationale is that parts of the state tree will most certainly not be loaded yet (most of the state tree is modular) and won't even be necessary. It might require some defensive code to explicitly check if a state exists in order to satisfy TS, but ideally, this could be in its own function where it only happens once. 

I've also left [some thoughts](https://github.com/Automattic/wp-calypso/pull/72978#discussion_r1098363804) on making testing easier with regards to dynamically loaded reducer trees.

spraveenitpro:
1. Functional testing works fine as expected.
2. I did do static analysis and nothing specific stands out apart form the E2E tests that are currently failing. @sirbrillig

sirbrillig:
>  I think that since there's no way for TS to ensure that a certain state tree has been loaded, it should be a Partial of the state. 

👍 I think we're agreeing here, although I didn't explicitly use the `Partial` helper and instead just made most of the properties optional. Part of the reason is that we can't make the `marketplace` slice optional without a fair number of code changes (which are probably worth doing but would make this PR much more risky; I'll follow up on that later). If I'm misunderstanding what you mean, though, let me know. It's possible you meant that we should leave the `IAppState` properties required and then use `Partial<IAppState>` wherever it's used?

What do you mean by "it" in "it should be a Partial of the state"?

> but ideally, this could be in its own function where it only happens once

🤔 I think I'm starting to understand what you mean. Leaving the slices of `IAppState` as non-optional but creating some top-level selector for each slice (or the whole tree) that all other selectors use to verify that the state is loaded?

sirbrillig:
> but ideally, this could be in its own function where it only happens once

🤔 I think I'm starting to understand what you mean. Leaving the slices of `IAppState` as non-optional but creating some top-level selector for each slice (or the whole tree) that all other selectors use to verify that the state is loaded and convince TS of that fact. Let me experiment with that.

tyxla:
I think both optional properties and `Partial` will do the job, really, as long as TS knows that those state pieces are optional since it doesn't know about our modular state mechanics. Feel free to try any of these approaches. Really, what difference does it make if they're required but we're using a partial type, or if they're optional?

sirbrillig:
Cool. Thank you, @tyxla! I just wanted to make sure we were talking about the same thing.

I bet we can improve how things are typed here even further in the future but I believe this will be an improvement at least.

Issue #71988, opened 2023-01-12
Pattern Assembler: Add a site option to ID assembler sites in a8c tools


miksansegundo:
# Description

As a first iteration, let' only add a site option to identify the sites that click on the Blank canvas CTA and complete the Assembler step to create their site.

* site option: `wpcom_site_setup = assembler`

This will be used in filtering in Site Snapshots, site quality analysis, data analysis, and possibly downstream in calypso UI.

The query for the options of all sites in `wpcom` must be done in the table `wpcom/blog_properties` in Hue. Something like:
```
select distinct blog_id from wpcom.blog_properties
where wpcom_site_setup = assembler;
```

# Acceptance Criteria
- [X] Be able to filter sites by the option  to know the count, URL, etc...
- [X] Be able to know when a site is created with the Site Assembler flow in Calypso, in MC tools like the Blog Report Card.

# Dev Tasks
- [X] Set the site option `wpcom_site_setup=assembler` for newly created sites, expect for virtual themes
- [x] Document the option `wpcom_site_setup` somewhere

# Testing 
* Create a new site and access the PA flow
* Click "Continue" to complete the PA flow
* Go to `/wp-admin/options.php` in your site to confirm that option `design_type: assembler` is there

## Related
p1673503178795799/1673498055.131429-slack-C048CUFRGFQ
https://github.com/Automattic/dotcom-forge/issues/1430

autumnfjeld:
Maybe we should discuss this more widely.   I'm not sure `design_type` is good.   For example we might want to use language that encompasses a broader idea.  That could be used for LIB, Newsletter, Build, Write, Assembler.     I'm not yet sure.

miksansegundo:
I like `wpcom_signup_flow`.

autumnfjeld:
See p1675390590190449-slack-C014TPFLP4Z

I don't think it should include the word `signup`.   Should be broader, like it could include a site going through a theme switch.

Seems like your idea `design_type is popular in that slack thread.  Maybe `wpcom_design_type` ? 

cc: @Automattic/lego

autumnfjeld:
Related comment left here https://github.com/Automattic/wp-calypso/pull/72989#issuecomment-1423742277

autumnfjeld:
Let's check in with Nuria if this would be valuable for the launchpad. cc: @miksansegundo

miksansegundo:
[...]

PR #70641, opened 2022-12-01
New Welcome Tour Slide for Site Editor

arthur791004:
#### Proposed Changes

* Introduce `completedFlow` to the site editor and it will be used only once when you finish the flow and land on the site editor
* Replace the first slide of the welcome tour with the new one if the completed flow is pattern assembler
* When you click on the image or link on the new slide, you will see the site editor course on a new tab
* Add a new event to track the link is clicked
* Update the copy of the first slide from "Try it out" to "Take tour"

**TODOs**

- [x] register event: calypso_editor_wpcom_tour_site_editor_course_link_click

**Screenshot**

image

**Event**

image

#### Testing Instructions

* Open another terminal and run `cd apps/editing-toolkit && yarn dev --sync` to sync the ETK plugin to your sandbox. Or you can run `install-plugin.sh editing-toolkit feat/pa-welcome-tour` on your sandbox.
* Sandbox **_your site_**
* Go to `/setup?siteSlug=<your_site>&flags=signup/design-picker-pattern-assembler` 
* Continue until you land on the Design Picker step
* Scroll to bottom and click the Blank Canvas CTA
* Select any patterns and continue
* When you land on the site editor, verify the first slide is the new one
* Click on the image or link, and the browser will open a new tab and the site editor course will show
* Check your network status and verify `calypso_editor_wpcom_tour_site_editor_course_link_click` event is sent
* Refresh the site editor, and verify the first slide is the same as before

#### Pre-merge Checklist

- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
[...]

PR #72989, opened 2023-02-06
Site Assembler: Replace with-theme flow with with-theme-assembler

arthur791004:
#### Proposed Changes

* Referring to p1675649656104429-slack-CRWCHQGUB, we want to have a different flow name for the `with-theme` flow that ends with `site-assembler`. So, this PR proposes to create a new flow called `with-theme-assembler`, and all behaviors are the same as `with-theme` flow except the final destination.

| Land on the `with-theme-assembler` flow | Land on the `pattern-assembler` step |
| - | - |
| image | image |

#### Testing Instructions

* Open the browser in Incognito mode
* Go to /themes
* Scroll down and select the Pattern Assembler CTA
* Verify you're going to `/start/with-theme-assembler` and the flow name of the track events is `with-theme-assembler`
* Verify you're going to `/setup/with-theme-assembler` when you finish the signup and create a new site. Moreover, the flow name of the track event should be `with-theme-assembler` as well.

#### Pre-merge Checklist

Related to p1675649656104429-slack-CRWCHQGUB

### DIFF
--- a/client/components/themes-list/index.jsx
+++ b/client/components/themes-list/index.jsx
@@ -2,7 +2,7 @@ import { isEnabled } from '@automattic/calypso-config';
 import { FEATURE_INSTALL_THEMES } from '@automattic/calypso-products';
 import { Button } from '@automattic/components';
 import { PatternAssemblerCta, BLANK_CANVAS_DESIGN } from '@automattic/design-picker';
-import { SITE_ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { SITE_ASSEMBLER_FLOW, WITH_THEME_ASSEMBLER_FLOW } from '@automattic/onboarding';
 import { Icon, addTemplate, brush, cloudUpload } from '@wordpress/icons';
 import { localize } from 'i18n-calypso';
 import { isEmpty, times } from 'lodash';
@@ -40,7 +40,7 @@ export const ThemesList = ( props ) => {
 			theme: BLANK_CANVAS_DESIGN.slug,
 			destination_flow: SITE_ASSEMBLER_FLOW,
 		} );
-		window.location.assign( `/start/with-theme?${ params }` );
+		window.location.assign( `/start/${ WITH_THEME_ASSEMBLER_FLOW }?${ params }` );
 	};
 
 	const matchingWpOrgThemes = useMemo(
--- a/client/components/themes-list/index.jsx
+++ b/client/components/themes-list/index.jsx
@@ -2,7 +2,7 @@ import { isEnabled } from '@automattic/calypso-config';
 import { FEATURE_INSTALL_THEMES } from '@automattic/calypso-products';
 import { Button } from '@automattic/components';
 import { PatternAssemblerCta, BLANK_CANVAS_DESIGN } from '@automattic/design-picker';
-import { SITE_ASSEMBLER_FLOW, WITH_THEME_ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { WITH_THEME_ASSEMBLER_FLOW } from '@automattic/onboarding';
 import { Icon, addTemplate, brush, cloudUpload } from '@wordpress/icons';
 import { localize } from 'i18n-calypso';
 import { isEmpty, times } from 'lodash';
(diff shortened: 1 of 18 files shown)


autumnfjeld:
@arthur791004  In this comment I have done an audit of the key events that fire at each step.  Not all of these events need `flow=with-theme-assembler`, but I think there are a few events that are firing that DON'T have the flow and we should add it.   In the next comment I'll list these events.   Probably tomorrow, because I'm getting tired and this stuff hurts my brain. 😆 


image


----
Click here
image
Land here:
image


Events:
image

@arthur791004  

----
Fill in input fields & click "Create your account"
image
image

image

---
On the domain screen, input domain and select free domain, then land on Plans (Choose your flavor of WP 🤣) 

image

image


---
On the plans screen click "Start with free"  & see loading screen and then land on Assembler screen
image

image


-----
Clicking "Continue" = end of Assembler step, goes to site editor
image


image

image




-----
Landing in Site Editor
image


image
image

autumnfjeld:
Let's add prop `flow=with-theme-assembler` on these events.  



image


image

image


For the next four events, let's add both `flow` and `intent` prop. Because soon it will be valuable to know if user came through the build or write flow (or null if coming via the theme showcase). Do you agree?

image

image


image
image



Do you see any reason not to add `flow` in these events?

autumnfjeld:
Perhaps we should open another issue to update all the tour events to include `flow`, `intent`, and `editor_type = site | post | page`.  I think those are the three editor types?   A good first issue for Tadahiro  @ okmttdhr

arthur791004:
> Do you see any reason not to add flow in these events?

I think adding flow to those events is good for the signup event. But for the editor or tour events, they're not highly related to the flow. The flow exists only when users just finished the post-signup flow and are landing on the editor for the first time. Maybe adding the intent is enough. What do you think?

> Perhaps we should open another issue to update all the tour events to include flow, intent, and editor_type = site | post | page. I think those are the three editor types? A good first issue for Tadahiro @ okmttdhr

Sounds good! I'll create a new issue for it 🙂

autumnfjeld:
>  But for the editor or tour events, they're not highly related to the flow. The flow exists only when users just finished the post-signup flow and are landing on the editor for the first time. Maybe adding the intent is enough. What do you think?

We have special condition for the Tour that it will conditionally content for the Assembler 
* https://github.com/Automattic/wp-calypso/pull/70641#issue-1471143253, 
*  (https://github.com/Automattic/wp-calypso/blob/4ceb8e33c74dabea7ed2d70fa4df96dab6ac6e9b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/welcome-tour/tour-steps.tsx)
   *  `const isPatternAssemblerFlow = 'pattern_assembler' === completedFlow;` then 

Even though it is not same exact usage as "flow" with `flow=with-theme-assembler`  the sentiment is the same.  

So I'm just thinking, are there future analysis we might want to do where we know a user interacts more with Tour in certain cases, in the case of write flow or using the Assembler?   It's not really a high priority, but could be important if we make better use of the Tour in the future.  


This is not directly related to  your question, but I'm beginning to wonder if we should use `intent=with-theme-assembler`. 🤔.   As @Addison-Stavlo  says here the site option `site_intent` is set to `link-in-bio` or `newsletter` etc.    So there is an analogy there to make the intent be `with-theme-assembler`.  There is no conflict with `intent=write or build` because it is a very specific flow.   

We use intent in two ways:
* tracks prop
* site option 

For user's coming from the Theme Showcase, I think it would be ok to use the value `with-theme-assembler` in both these cases.   And then also we don't need this: [Pattern Assembler: Add a site option to save design_type: assembler#71988](https://github.com/Automattic/wp-calypso/issues/71988) because we can keep using `site_intent` for a wide variety of uses.    What do you think?   @miksansegundo & @arthur791004

arthur791004:
I've added flow, step, and intent to signup-related events! For the editor or tour event, I'll create an issue for it as mentioned above.

image

image

image

image

image

image
(I'm not sure why my extension cannot capture this event)

arthur791004:
> This is not directly related to your question, but I'm beginning to wonder if we should use intent=with-theme-assembler. 🤔. As @Addison-Stavlo says here the site option site_intent is set to link-in-bio or newsletter etc. So there is an analogy there to make the intent be with-theme-assembler. There is no conflict with intent=write or build because it is a very specific flow.

Let's do it!

arthur791004:
@autumnfjeld I've [added flow, step, and intent to signup-related events](https://github.com/Automattic/wp-calypso/pull/72989#issuecomment-1423832078) and set the site intent to `with-theme-assembler`. For the tour events, here is a new [issue](https://github.com/Automattic/wp-calypso/issues/73142)

autumnfjeld:
Thank you @arthur791004 good to go from my end. :)

PR #73010, opened 2023-02-06
Copy Site: Customize text used on initial Processing screen

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/1589

## Proposed Changes

Customizes the text used on the initial `Copy Site i1` Processing screen (pre-checkout):

CleanShot 2023-02-06 at 12 15 48@2x

## Testing Instructions

1. Visit `/sites`.
2. Click on 'Copy site' from the triple dot menu.
3. Verify the strings are used in the expected order.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/processing-step/hooks/use-processing-loading-messages.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/processing-step/hooks/use-processing-loading-messages.ts
@@ -13,8 +13,8 @@ export function useProcessingLoadingMessages( flow?: string ) {
 
 	if ( flow === 'copy-site' ) {
 		return [
-			{ title: __( 'Laying the foundations' ), duration: 5000 },
-			{ title: __( 'Securing your data' ), duration: 4000 },
+			{ title: __( 'Laying the foundations' ), duration: 3500 },
+			{ title: __( 'Securing your data' ), duration: 4500 },
 			{ title: __( 'Enabling encryption' ), duration: 5000 },
 			{ title: __( 'Applying a shiny top coat' ), duration: 4000 },
 		];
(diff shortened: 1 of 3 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/processing-step/hooks/use-processing-loading-messages.ts
```diff
@@ -11,6 +11,14 @@ export function useProcessingLoadingMessages() {
 
 	const stepData = useSelect( ( select ) => select( STEPPER_INTERNAL_STORE ).getStepData() );
 	switch ( stepData.intent ) {
+		case SiteIntent.Copy:
```
danielbachhuber: @wojtekn Are we setting `site_intent` for the Simple site? I thought we were, but I don't see it anywhere...

wojtekn: No, you may look for `setIntentOnSite`.

danielbachhuber: I ended up directly referencing `flow` 851c6b91b2

`site_intent` is conceptually slightly different, so I didn't want to add bad data into that system.

wojtekn (approved):
@danielbachhuber looks good, and it works for me. I just noticed that with current values of `duration,` I rarely see the last message. I experience better behavior with values closer to 3000ms.

danielbachhuber:
> looks good, and it works for me. I just noticed that with current values of `duration,` I rarely see the last message. I experience better behavior with values closer to 3000ms.

@wojtekn Thanks! Yeah, I think there isn't a perfect number. I generally prefer "slow enough to read" so we can reinforce the messaging.

Issue #44992, opened 2020-08-18
Reader: centered paragraphs are displayed with left text alignment


designsimply:
#### Steps to reproduce

1. Go to WordPress.com > My Site > Posts > Add new post.
1. Add a paragraph block and change the color for it in the block settings.
1. Add a paragraph with "Align text left".
1. Add a paragraph with "Align text center".
1. Add a paragraph with "Align text right".
1. Publish the post.
1. View the post in the Reader and check to see if the paragraph alignment looks correct.

#### What I expected

I expected that centered text would appear center-aligned in the Reader.

#### What happened instead

No alignment options are respected in the Reader.

#### Browser / OS version

* Safari 13.1.1 on macOS 10.15.4 
* Safari iOS 13.6 on iPhone 11
* WPiOS 15.5.01 on iPhone 11

#### Screenshot / Video

Web|Reader web|Reader mobile web|Reader iOS app Dark Mode
---|---|---|---
<img width="1280" alt="Screen Shot 2020-08-18 at Aug 18 9 03 24 AM" src="https://user-images.githubusercontent.com/1119271/90530364-0a8f4f80-e132-11ea-9e4a-46b1a720c793.png">|<img width="1280" alt="Screen Shot 2020-08-18 at Aug 18 8 59 31 AM" src="https://user-images.githubusercontent.com/1119271/90530395-13802100-e132-11ea-8885-52bb17acbe29.png">|![IMG_4513](https://user-images.githubusercontent.com/1119271/90530469-2a267800-e132-11ea-9f2d-9a24d967ff98.png)|![IMG_4515](https://user-images.githubusercontent.com/1119271/90530480-2c88d200-e132-11ea-93ab-e3b75b8da866.png)

* Web: https://made4testing0318.blog/2020/08/18/paragraph-color-alignment/
* Reader: https://wordpress.com/read/feeds/96421692/posts/2866304892

#### Context / Source

In mobile support, we have observed that users have asked for text alignment and colors to be respected in the Reader. We discussed which of those things should be respected in an intentionally style-neutral environment such as the Reader at https://github.com/wordpress-mobile/WordPress-iOS/issues/13930#issuecomment-615465956 and the consensus was that we should not support color changes in the Reader but we are open to supporting alignment. 

[...]

PR #73011, opened 2023-02-06
Fix bug where centered paragraphs show up left aligned on reader full post view

davemart-in:
## Description

When a post is published with center aligned text, it currently shows up as left aligned in the reader full page view. It should show us as left aligned in the feed, but in the full-page view we should try and keep the post as 1:1 to the actual post as possible.

### Before

CleanShot 2023-02-06 at 10 37 28

### After

CleanShot 2023-02-06 at 10 41 30

### Note

Text in feed should still show up left aligned:

CleanShot 2023-02-06 at 10 45 54

## Related

#44992, #73014

### DIFF
--- a/client/blocks/reader-full-post/style.scss
+++ b/client/blocks/reader-full-post/style.scss
@@ -158,8 +158,12 @@
 	color: var(--color-neutral-70);
 	font-size: rem(17px);
 	line-height: 28px;
+
 	figure.wp-block-table {
 		overflow: auto;
+	}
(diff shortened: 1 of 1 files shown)


creativecoder (approved):
Works as advertised!

image

Any reason not to do right alignment, as well?

roo2 (approved):
LGTM

davemart-in:
> Any reason not to do right alignment, as well?

@creativecoder see #73014

PR #73014, opened 2023-02-06
Fix bug where right aligned paragraphs show up left aligned on reader full post view

davemart-in:
## Description

When a post is published with right aligned text, it currently shows up as left aligned in the reader full page view. It should show us as left aligned in the feed, but in the full-page view we should try and keep the post as 1:1 to the actual post as possible.

### Before
CleanShot 2023-02-06 at 13 37 42

### After
CleanShot 2023-02-06 at 13 38 56

### Note
Text in feed should still show up left aligned:
CleanShot 2023-02-06 at 13 39 16

## Related
#44992, #73011


roo2 (approved):
LGTM

creativecoder (approved):
Looks good on my test site.

image

PR #73017, opened 2023-02-06
Prepare the domain page to work with the experiment - backend strategy

### LINKED ISSUE #72889: Sidebar nudge upsell: Prepare the domain page to work with the experiment

paulopmt1:
#### Proposed Changes

* Prepare the domain page to work with the experiment. We're focusing on the New header, background color, and centralizing the content. Domain filtering section [will be handled in another task](https://github.com/Automattic/wp-calypso/issues/72894), as well as [the Domain Header](https://github.com/Automattic/wp-calypso/issues/72892).

image

#### Testing Instructions

* Apply this diff to your sandbox API: D100077-code
* Access the current Domain upsell flow starting at `Upgrade` button on your side menu
* The domain page should look like the production one (assuming you're at the control group)
* Change your variation to `treatment` at the experiment `21025-calypso-sidebar-upsell-dedicated-upgrade-flow-202301` following these tips: pb5gDS-2nD-p2#connecting-an-experiment-to-a-jitm
* **IMPORTANT**: First access [the Calypso home](http://calypso.localhost:3000/), and after click at the CTA on the left menu (to ensure our [CSS rules](https://github.com/Automattic/wp-calypso/pull/72932/files#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4R105) will be applied as expected)
* Make sure your change will work as expected
* You should now see the new interface, make sure it aligns with our design expectations: pebzTe-Fx-p2#comment-1131
* Make sure the `calypso_upgrade_nudge_cta_click` (with property cta_name: free-to-paid-sidebar) is the correct exposure event for triggering this experiment.

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/72889


client/my-sites/controller.js
```diff
@@ -742,3 +743,15 @@ export function selectSiteIfLoggedIn( context, next ) {
...
+ * Whether the user is in the Domain sidebar upsell experiment.
+ *
+ * @returns {boolean} Whether the user is in the Domain sidebar upsell experiment.
+ */
+export const isDomainSidebarExperimentUser = ( state ) => {
+	const currentUser = getCurrentUser( state );
+	const domainAndPlanPackage = getCurrentQueryArguments( state ).domainAndPlanPackage;
+
+	return domainAndPlanPackage && 'treatment' === currentUser?.calypso_sidebar_upsell_experiment;
+};
```
tyxla: This looks like a selector and should live in its own file in `client/state/selectors/`.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -308,6 +341,7 @@ export default connect(
 			isSiteOnMonthlyPlan: isSiteOnMonthlyPlan( state, siteId ),
 			productsList: getProductsList( state ),
 			userCanPurchaseGSuite: canUserPurchaseGSuite( state ),
+			stateObject: state,
```
tyxla: Passing the entire state object to a component is an antipattern, because it will re-render on any state change. This is a good place to use the `isDomainSidebarExperimentUser` selector, which will limit the area of state object interest for this component.

client/my-sites/domains/style.scss
```diff
@@ -10,3 +11,45 @@
 		align-items: center;
 	}
 }
+
+body.is-section-domains.is-experiment-user {
```
tyxla: Do we need the `is-section-domains` class here? Doesn't this stylesheet (`/my-sites/domains/style.scss`) imply that we're already in the domains section?

tyxla: The `is-experiment-user` class also feels a bit broad. What does it actually mean? What if there are multiple experiments? Which one does this apply to?

client/my-sites/domains/style.scss
```diff
@@ -1,5 +1,6 @@
 @import "@wordpress/base-styles/breakpoints";
 @import "@wordpress/base-styles/mixins";
+@import "@automattic/typography/styles/variables";
```
tyxla: Is this necessary? Isn't it already imported here:

https://github.com/Automattic/wp-calypso/blob/b7a4a072ae2a13785fb6ccbf9c34ed8e65739705/client/assets/stylesheets/shared/_typography.scss#L2

client/my-sites/domains/style.scss
```diff
@@ -10,3 +11,45 @@
...
+			font-size: $font-body;
+			color: var(--color-text-subtle);
+			text-align: center;
+			padding: 0 10px;
+		}
+
+		a {
+			color: var(--studio-gray-90);
+			font-weight: 500;
+			padding-left: 5px;
```
tyxla, line 35: Do we really want to add a left space to every link? This seems like a code smell.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -196,12 +202,17 @@ class DomainSearch extends Component {
...
 	render() {
-		const { selectedSite, selectedSiteSlug, translate, isManagingAllDomains, cart } = this.props;
+		const { selectedSite, selectedSiteSlug, translate, isManagingAllDomains, cart, stateObject } =
+			this.props;
 
 		if ( ! selectedSite ) {
 			return null;
 		}
 
+		if ( isDomainSidebarExperimentUser( stateObject ) ) {
```
tyxla: Calling Redux selectors directly on rendering is an antipattern. They should be called either in `connect()` if using a class component (like here) or through `useSelector()` in a functional component.

tyxla (changes requested):
Thanks for considering the server-side approach 👍 Nice work too.

I think this is going in a good direction overall. I've left a few suggestions that mostly are about antipatterns that need to be refactored.

I'm happy to give another review once they're addressed.

client/my-sites/domains/style.scss
paulopmt1: I see the same pattern being used in other places [like the /plans page](https://github.com/Automattic/wp-calypso/blob/update/implement-domains-page-for-experimentV2/client/my-sites/plans/header.scss#L3). In this particular case, we if remove it, the `is-experiment-user` will be enough to filter it, should we be less restrictive in this CSS?

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Oh, that's great to know, thanks!
[Fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/95e64239f4ea24b59fc9ca5d33e2a2d428abc245).

client/my-sites/domains/style.scss
tyxla: Committed code doesn't always stand the standards and best practices. Plus, there are always bits to improve. [It's a marathon, not a sprint](https://automattic.com/creed/marathon-sprint/), after all.

I'd say, if `is-experiment-use` is enough, then no need to add another class name.

client/my-sites/controller.js
paulopmt1: Nice, [fixed it here](https://github.com/Automattic/wp-calypso/pull/73017/commits/95e64239f4ea24b59fc9ca5d33e2a2d428abc245).

client/my-sites/domains/style.scss
paulopmt1: Yeah, agree, it's too generic. [Improved it here](https://github.com/Automattic/wp-calypso/pull/73017/commits/ccaaa9cb70837c2e4fd33ab7f6b5656062617386).

client/my-sites/domains/style.scss
paulopmt1: Yeah, it's not necessary, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/ccaaa9cb70837c2e4fd33ab7f6b5656062617386).

client/my-sites/domains/style.scss
paulopmt1: It would be applied only at the experiment header, but assuming this experiment will be successful we'll make this code permanent, so good point, [I improved it here](https://github.com/Automattic/wp-calypso/pull/73017/commits/7f91d2e9e3df70265ae00ba0722efecaaf4e67b9).

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Nice, [improved it here](https://github.com/Automattic/wp-calypso/pull/73017/commits/95e64239f4ea24b59fc9ca5d33e2a2d428abc245).

DustyReagan (approved):
This is working functionally! 👍

client/layout/index.jsx
```diff
@@ -384,7 +385,10 @@ export default withCurrentRoute(
 		const isJetpack =
 			( isJetpackSite( state, siteId ) && ! isAtomicSite( state, siteId ) ) ||
 			currentRoute.startsWith( '/checkout/jetpack' );
-		const noMasterbarForRoute = isJetpackLogin || currentRoute === '/me/account/closed';
+		const noMasterbarForRoute =
+			isJetpackLogin ||
+			currentRoute === '/me/account/closed' ||
+			isDomainSidebarExperimentUser( state );
```
tyxla: The `isDomainSidebarExperimentUser` selector is called twice in this `connect()` call. We could call it once by assigning it to a constant and using that constant here.

client/my-sites/domains/style.scss
```diff
@@ -10,3 +10,45 @@
 		align-items: center;
 	}
 }
+
+body.is-section-domains.is-domain-sidebar-experiment-user {
+	background: #fdfdfd;
```
tyxla: Ideally, we should be using color variables instead of hardcoded color values.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -196,18 +202,30 @@ class DomainSearch extends Component {
...
+			domainSidebarExperimentUser,
+		} = this.props;
 
 		if ( ! selectedSite ) {
 			return null;
 		}
 
+		if ( domainSidebarExperimentUser ) {
+			document.body.classList.add( 'is-domain-sidebar-experiment-user' );
+		}
```
tyxla: I see we're removing this class on unmount above. That makes a lot of sense. However, does it make sense to add the class on `componentDidUpdate()` instead of doing it on `render()`? Seems so, because it's a side effect, and performing side effects during rendering is not preferred. TL;DR: let's move these into `componentDidUpdate()`.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -196,18 +202,30 @@ class DomainSearch extends Component {
...
+
 		const classes = classnames( 'main-column', {
 			'domain-search-page-wrapper': this.state.domainRegistrationAvailable,
 		} );
 		const { domainRegistrationMaintenanceEndTime } = this.state;
 
 		const hasPlanInCart = hasPlan( cart );
+		const hrefForDecideLater = `/plans/yearly/${ selectedSite.domain }?domainAndPlanPackage=true`;
```
tyxla: Do we always want to link to `yearly` plans? Is there any possibility that we were on monthly plans before and would like to continue on monthly plans again?

tyxla: Furthermore, when adding a query argument to a route or URL, we tend to use `addQueryArgs()` from `calypso/lib/url`, or [`addQueryArgs` from `@wordpress/url`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-url/#addqueryargs).

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -236,24 +254,48 @@ class DomainSearch extends Component {
...
 				<span>
 					<div className="domain-search__content">
-						<BackButton
-							className="domain-search__go-back"
-							href={ domainManagementList( selectedSiteSlug ) }
-						>
-							<Gridicon icon="arrow-left" size={ 18 } />
-							{ translate( 'Back' ) }
-						</BackButton>
+						{ false === domainSidebarExperimentUser && (
```
tyxla: `domainSidebarExperimentUser` should be a boolean already since the selector returns a boolean, so we could simplify it.

```suggestion
						{ ! domainSidebarExperimentUser && (
```

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -236,24 +254,48 @@ class DomainSearch extends Component {
...
+											'Stake your claim on your corner of the web with a custom domain name that’s easy to find, share, and follow. Not sure yet?'
+										) }
+										<a className="domains__header-decide-later" href={ hrefForDecideLater }>
+											{ translate( 'Decide later.' ) }
+										</a>
+									</p>
+								</>
+							) }
+
+							{ false === domainSidebarExperimentUser && (
```
tyxla:
```suggestion
							{ ! domainSidebarExperimentUser && (
```

client/state/selectors/is-domain-sidebar-experiment-user.js
```diff
@@ -0,0 +1,14 @@
...
+import getCurrentQueryArguments from 'calypso/state/selectors/get-current-query-arguments';
+
+/**
+ * Whether the user is in the Domain sidebar upsell experiment.
+ *
+ * @returns {boolean} Whether the user is in the Domain sidebar upsell experiment.
+ */
+export const isDomainSidebarExperimentUser = ( state ) => {
+	const currentUser = getCurrentUser( state );
+	const domainAndPlanPackage = getCurrentQueryArguments( state ).domainAndPlanPackage;
```
tyxla: `getCurrentQueryArguments` could be `null`, and trying to access a property on a `null` will trigger a `TypeError`. We could use optional chaining to address that.

Furthermore, we're using this query argument value as a boolean, so let's make sure to cast it. It could theoretically contain anything - including falsy values, so casting here would be required. We can prefix with `!!` to cast it to a boolean.

```suggestion
	const domainAndPlanPackage = !! getCurrentQueryArguments( state )?.domainAndPlanPackage;
```

client/state/selectors/is-domain-sidebar-experiment-user.js
```diff
@@ -0,0 +1,14 @@
+import { getCurrentUser } from 'calypso/state/current-user/selectors';
+import getCurrentQueryArguments from 'calypso/state/selectors/get-current-query-arguments';
+
+/**
+ * Whether the user is in the Domain sidebar upsell experiment.
+ *
+ * @returns {boolean} Whether the user is in the Domain sidebar upsell experiment.
+ */
+export const isDomainSidebarExperimentUser = ( state ) => {
```
tyxla, line 10: This selector could use a unit test or two.

client/state/selectors/is-domain-sidebar-experiment-user.js
```diff
@@ -0,0 +1,14 @@
+import { getCurrentUser } from 'calypso/state/current-user/selectors';
+import getCurrentQueryArguments from 'calypso/state/selectors/get-current-query-arguments';
+
+/**
+ * Whether the user is in the Domain sidebar upsell experiment.
+ *
```
tyxla, line 6: Nit: `state` could be added as a `@param` in this JSDoc block.

tyxla (commented):
This is looking way better, thank you @paulopmt1! 🚀 

I think there are a few more details that needed to be ironed out, I've left feedback and suggestions.

I think once they're addressed this should be good to go, but I'm happy to give it one last look once the feedback has been addressed.

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Good catch! We don't plan to offer a monthly option in this flow.

client/layout/index.jsx
paulopmt1: Nice, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e).

client/my-sites/domains/style.scss
paulopmt1: Sounds good, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e).

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Great, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e).

client/my-sites/domains/domain-search/index.jsx
paulopmt1: [Changed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4R229). What's the advantage of using `addQueryArgs` in this case as the only parameter has a fixed value?

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Good, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4L257).

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Good, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/db39021a84115d37704b723a4f414228a102c42e#diff-e9d5bbfeb0cc388a1ec2501f08f49524a97724605eca06bc1f3f626bf12cfab4L288).

client/state/selectors/is-domain-sidebar-experiment-user.js
paulopmt1: Interesting to know, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/3aeff691fe16323df3aa05089b1791e5309eff8b).

client/state/selectors/is-domain-sidebar-experiment-user.js
paulopmt1: Nice, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/3aeff691fe16323df3aa05089b1791e5309eff8b).

client/state/selectors/is-domain-sidebar-experiment-user.js
paulopmt1: Tests are great! As this code will be deleted after the experiment ends, in less than a month, I'm wondering if it is worth writing a test for it 🤔

paulopmt1:
Thanks again for your reviews @tyxla, waiting for the final review 🤞

client/state/selectors/is-domain-sidebar-experiment-user.js
candy02058912: Added unit tests since this was a quick one.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -107,12 +110,20 @@ class DomainSearch extends Component {
 	}
 
 	componentDidUpdate( prevProps ) {
+		if ( this.props.domainSidebarExperimentUser ) {
+			document.body.classList.add( 'is-domain-sidebar-experiment-user' );
+		}
```
tyxla: Something else just came up as I saw this: we'll also need to cover the opposite scenario: on update, if `domainSidebarExperimentUser` is false, we should likely remove the class, just like we do on unmount.

client/my-sites/domains/domain-search/index.jsx
```diff
@@ -208,6 +226,12 @@ class DomainSearch extends Component {
 		const { domainRegistrationMaintenanceEndTime } = this.state;
 
 		const hasPlanInCart = hasPlan( cart );
+		const hrefForDecideLater = addQueryArgs(
+			{
+				domainAndPlanPackage: true,
+			},
+			`/plans/yearly/${ selectedSite.domain }`
```
tyxla: Note: theoretically, using the site slug is a better idea than using the domain. The rationale: IIRC `siteSlug` is better suited for working with sites installed in subdirectories.

client/state/selectors/is-domain-sidebar-experiment-user.js
tyxla: @paulopmt1 in that case tests would be understandable to be skipped, but I appreciate you adding some @candy02058912 🙇‍♂️

tyxla (approved):
This looks good to me 👍 Nice work @paulopmt1 🙌 

I've left a couple of preferred suggestions, but feel free to ship once they're addressed.

🚀

client/my-sites/domains/domain-search/index.jsx
tyxla: It's about sustainability, mostly. Often we end up adding more args in the future and this ends up being a more practical way to maintain query args.

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Nice, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/504282123d8ac50cb8978af6f669603f23727aba).

client/my-sites/domains/domain-search/index.jsx
paulopmt1: Nice, [fixed here](https://github.com/Automattic/wp-calypso/pull/73017/commits/504282123d8ac50cb8978af6f669603f23727aba).

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7835555

Thank you @paulopmt1 for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7835555).

PR #73018, opened 2023-02-06
Writing prompts: fix Home card analytics issues

creativecoder:
#### Proposed Changes

image

Fixes these issues with the My Home writing prompts card

- Fixes a bug where `calypso_customer_home_answer_prompt` and `calypso_customer_home_skip_prompt` events are triggered on page load, rather than when clicking to answer a prompt.
- Prevents JS errors if a malformed API response returns an empty prompt.
- Prevents a race condition where the page may navigate away before logging events when answering a prompt
- React development errors for `stroke-width` attribute and missing key in `answered_users_sample` list.

#### Testing Instructions

* Using a blog that is set to show posts on the home page, visit `/home/{site}` and find the Daily writing prompt card.
* In the JS console of the browser, use `localStorage.setItem( 'debug', 'calypso:analytics*' );` to enable tracks debugging, and reload the page.
* Make sure the `calypso_customer_home_answer_prompt` and `calypso_customer_home_skip_prompt` events **do not** trigger when the page is loaded.
* Click on "View all responses". You should see the `calypso_customer_home_view_all_responses` event logged in the console and navigate to the reader tag page for the prompt.
* Go back and click the "Post Answer" button. You should see the `calypso_customer_home_answer_prompt` event logged and navigate to the editor.
* Go back, change the prompt, and click on the "Post Answer" button again. You should additionally see the `calypso_customer_home_skip_prompt` even recorded.
* Switch to an Atomic site that loads the editor outside calypso, and test that the Post Answer button works in the same way.


client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
```diff
@@ -67,13 +74,25 @@ const PromptsNavigation = ( { prompts } ) => {
...
+		const newPostLinkType = determineUrlType( getNewPostLink() );
+
+		if ( newPostLinkType === URL_TYPE.PATH_ABSOLUTE ) {
+			// Internal link, use router navigation to avoid reloading Calypso.
+			page( newPostLink );
+		} else {
+			// External link, use a full page reload to the new url.
+			window.location.href = newPostLink;
+		}
 	};
```
roo2: I think that the `navigate( path )` function performs this check https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/navigate/index.ts

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
```diff
@@ -48,17 +52,20 @@ const PromptsNavigation = ( { prompts } ) => {
...
-				prompt_id: getPrompt().id,
+				prompt_id: getPrompt()?.id,
 			} )
 		);
 
 		// Track if a user skipped todays prompt and choose to answer another prompt
 		const todayPromptId = prompts[ 0 ].id;
-		const skippedPromptId = getPrompt().id;
+		const skippedPromptId = getPrompt()?.id;
 		if ( todayPromptId !== skippedPromptId ) {
```
roo2: what's recorded as `prompt_id` is currently the id of the prompt that they chose. I think we want to be logging the id of the prompt that they skipped, which would be `todayPromptId`.

The variable `skippedPromptId` is confusingly named, it's actually the _selected_ prompted id.

creativecoder: Great catch, thank you! Updated.

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
creativecoder: Thank you, this is the function I was looking for! ❤️

eoigal (approved):
lgtm

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
eoigal: Good catch alright. I was trying to log what prompt the user skipped to but I see now that we already log that in the `calypso_customer_home_answer_prompt` event.

client/my-sites/customer-home/cards/features/blogging-prompt/prompts-navigation.jsx
```diff
@@ -140,13 +151,10 @@ const PromptsNavigation = ( { prompts } ) => {
 	};
 
 	const renderPromptAnswer = () => {
-		const newPostLink = addQueryArgs( editorUrl, {
-			answer_prompt: getPrompt().id,
-		} );
 		return (
 			<div className="blogging-prompt__prompt-answers">
 				{ renderResponses() }
-				<Button href={ newPostLink } onClick={ trackBloggingPromptClick() } target="_blank">
```
eoigal, line 149: Ah ok.. so the bug here was by using `trackBloggingPromptClick()` instead of `trackBloggingPromptClick` the function is called on load and not just when clicked.

creativecoder: Yes, that was the main issue, thanks for the review!

PR #73019, opened 2023-02-06
Run cache check in separate Dockerfile step

noahtallen:
#### Proposed Changes
When we merged the initial PR (#72975), we noticed that the output of this script is never shown, since the log reaches Docker's internal logfile size limit. Easy fix is to run the script as a separate step, and since the bit about cache invalidation is early on in the file, it will still work.

#### Testing Instructions
Check that the build param `env.WEBPACK_CACHE_INVALIDATED` is ultimately set by the build. See https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/9434853?buildTab=log&focusLine=1526&logView=flowAware&linesState=186

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,8 +68,10 @@ RUN yarn install --immutable --check-cache
 # This contains built environments of Calypso. It will
 # change any time any of the Calypso source-code changes.
 ENV NODE_ENV production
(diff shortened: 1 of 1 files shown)


worldomonation (approved):
Let's try it.

noahtallen:
Tests passed on a re-run

PR #73027, opened 2023-02-07
Checkout: Support setting quantity of cart items by URL

sirbrillig:
#### Proposed Changes

When visiting the checkout URL in calypso, a series of product aliases can be included at the end of the URL that will be automatically added to the cart as it loads. For example, `/checkout/personal` will add a Personal plan. Some products require additional metadata, like `/checkout/domain_reg:example.com` which will add a Domain registration for the `example.com` domain.

This PR extends this behavior to allow setting the `quantity` of a product via the URL. If a URL's product alias includes a colon followed by the code `-q-` and a number, that number will be interpreted as a quantity (we know this is not a domain because domains cannot start with a hyphen). So a URL like `/checkout/some_product:-q-20` will add a product with a quantity of 20.

**Note:** few products can have a quantity set and even fewer products allow a user-supplied quantity, so although this PR will allow _requesting_ a quantity, the choice to actually _use_ that quantity lies within the shopping-cart API endpoint.

This implementation also creates a pattern for future reuse; if we need to encode additional metadata for a product into the URL, we can use a colon followed by a hyphen and a string to label that data. For example if we ever wanted to encode the `extra.signup` boolean flag, we could accept products like `some_product:-signup-1`.

This feature was requested in a p2 discussion here: p1HpG7-jya-p2#comment-60450

#### Testing Instructions

Automated tests are included. To test this manually:

- Open calypso with your browser's devtools open and recording network traffic to the `/me/shopping-cart` endpoint.
- Visit a checkout url with a quantity (it doesn't have to be real) like `/checkout/business:-q-23`.
- Verify that a network request is made to the shopping-cart endpoint with the quantity set to the number in the URL.

Screenshot 2023-02-06 at 7 35 24 PM

### DIFF
--- a/client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
@@ -432,12 +432,12 @@ function getProductPartsFromAlias( productAlias: string ): {
 } {
 	const [ slug, ...productParts ] = productAlias.split( ':' );
 
-	let meta = null;
-	let quantity = null;
+	let meta: string | null = null;
+	let quantity: number | null = null;
 
 	productParts.forEach( ( part ) => {
 		if ( /^-q-\d+$/.test( part ) ) {
-			quantity = part.replace( /^-q-/, '' );
+			quantity = parseInt( part.replace( /^-q-/, '' ), 10 );
 			return;
 		}
 		meta = part;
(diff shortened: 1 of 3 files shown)


michaeldcain (approved):
This works well. I tested the following:
- Existing meta functionality with a premium theme (`/checkout/theme:varese`). Works as expected - the theme was added to the cart.
- A quantity-based product with the Akismet Personal plan (sandboxed - `/checkout/ak_personal_yearly:-q-20`). Works as expected - the product was added to the cart with a $20 sub-total.
- Adding a quantity to a product that doesn't support quantity (`/checkout/personal:-q-20`). Works as expected - added the Personal plan to the cart with a "Sorry, WordPress.com Personal does not support specifying a quantity." dismissible error.

PR #73030, opened 2023-02-07
GH Integration: Add card to display list of repositories and branches to select from.

### LINKED ISSUE #70378: Created Connect UI (post-authorize)
### Details

After connecting to GitHub, a connection should be available in `/me/connections` that looks like this:

```json
"connections": [
      {
        "ID": 28746642,
        "user_ID": 219393573,
        "type": "other",
        "service": "github",
        "label": "GitHub",
         ...
      }
    ]
```

This endpoint should be queried and if a connection exists with `service === 'github'`, then the Connect Branch UI below can be shown. 

connect_branch

### Checklist

- [ ] UI looks like the mock-up
- [ ] Works for RTL
- [ ] This UI is shown conditionally. If no connection exists, the [Authorize UI](https://github.com/Automattic/wp-calypso/issues/70376) should be shown instead.
- [ ] The server can return a list of repos for the connection
- [ ] The server can return a list of branches for the selected repo
- [ ] The user can click "Connect' if a repo and branch are selected, which then calls the API to register a webhook event handler for the repo. 
- [ ] The selected repo and branch and persisted on the server
- [ ] User can specify a base path (defaults to root)

### Related

_No response_

### LINKED ISSUE #73404: Add repository search on GitHub card
When browsing repositories in the GitHub card so you can connect it to WPCOM, the existing list doesn't bring all repositories the user has access to. Being an organization member already brings all of their repos, and organizations often have a lot of repos, so fetching the list every time (GH API doesn't allow searching on [/user/repos](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-the-authenticated-user)) is very resource intensive.

Just being a member of Automattic, for example, already brings me 500+ repos. There's a filter to hide org repos (`affiliation`) but that's probably not a good idea, especially for developers, since they're often not working on personal repos.

`collaborator` doesn't really work because, for example, I've never been added to `wp-calypso` as a collaborator, still I can commit to the repo. I do not see it when using this filter... So we'll need to use a text input with a dropdown, just as we're using right now on @Vykes’s [PR](https://github.com/Automattic/wp-calypso/pull/73030).

The difference is that it'll be lazy, meaning that we'll start fetching repos as soon as the user starts typing.

Done is: I can pick any repo I want.

vykes-mac:
Closes https://github.com/Automattic/wp-calypso/issues/73404 and part of https://github.com/Automattic/wp-calypso/issues/70378.

#### Proposed Changes

* Add card to display list of user's github repositories and branches that may select from when connecting their site to github.

#### Testing Instructions

See pet6gk-5n-p2. You need to patch your sandbox with D101662-code for the search to work.

**Pre-requisites** 

```
- Sandbox the endpoints from this [diff](D100978-code)
```

- On any AT sites go to the hosting page
- Scroll down the github connect branch section
- A card should be displayed with your list of repositories
- Verify changing repo updates branches
- Verify display on smaller screens
- Verify it renders correctly on RTL languages

image

*

#### Pre-merge Checklist

Related to https://github.com/Automattic/wp-calypso/issues/70378


client/my-sites/hosting/github/github-connect-card/use-github-branches.ts
```diff
@@ -0,0 +1,30 @@
+import { useQuery } from 'react-query';
+import wp from 'calypso/lib/wp';
+
+const USE_GITHUB_BRANCHES_QUERY_KEY = 'github-branches-query-key';
+const CACHE_TIME = 1000 * 60 * 5; // 5 mins
```
p-jackson, line 5: Is there a reason 5 minutes is important for this case? 5 minutes is already the default value for `cacheTime`

client/my-sites/hosting/github/github-authorize-card/index.tsx
```diff
@@ -26,7 +26,7 @@ export const GithubAuthorizeCard = () => {
 
 	return (
 		<Card className="github-hosting-card">
-			<img className="github-hosting-icon" src={ iconGitHub } alt="" />
+			<SocialLogo className="material-icon" icon="github" size={ 32 } />
```
p-jackson, line 32: Nice one switching to the standard social icon 👍 Does that mean we can delete `github.svg`?

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,16 +1,106 @@
...
+	);
+
+	useEffect( () => {
+		if ( repos ) {
+			setSelectedRepo( selectedRepo || repos[ 0 ] );
+		}
+		if ( branches ) {
+			setSelectedBranch( branches[ 0 ] );
+		}
+	}, [ repos, branches, selectedRepo ] );
```
p-jackson: I think you need to do the `selectedBranch || branches[ 0 ]` trick here too?

We shouldn't need to have a `useEffect` here. Updating state based on some other state change causes a cascade of updates. The list of repos will load, causing a rerender, then this effect callback runs, which updates the state, which causes a rerender.

The idiomatic ways to remove the `useEffect` would be either to [calculate what the selected value should be](https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state) i.e. when setting the `value` prop something like this:

```tsx
value={ selectedRepo?.full_name || repos?.[0].full_name }
```

Or we could do this trick where [the state is updated directly in the render path](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) (I don't see this pattern very often).

Or we could pass an `onSuccess` handler to the query hooks, and we set the selected repo on success. I actually like the pattern of exposing all the ReactQuery options as one of the arguments to `useGithubRepos`, and I also think it's much clearer to just say "on success set the default dropdown selection" than to have to deal with `useEffect`

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,16 +1,106 @@
...
+
 	return (
-		<Card className="github-hosting-card">
-			<img className="github-hosting-icon" src={ iconGitHub } alt="" />
+		<Card className="connect-branch-card">
+			<SocialLogo className="material-icon" icon="github" size={ 32 } />
 			<CardHeading>{ __( 'Connect Branch' ) }</CardHeading>
+			<div>
+				<p>{ __( 'Changes pushed to connected branch will be deployed automatically' ) }</p>
+				{ ! showSpinner ? (
```
p-jackson: Personal preference, but I think it's more readable to put the shorter ternary branch first.

```suggestion
				{ showSpinner ? <Spinner /> : (
```

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,16 +1,106 @@
...
+					<FormFieldset className="connect-branch__fields">
+						<FormSelect
+							id="repositories"
```
p-jackson: I don't know where we reference these element IDs, but if we need them I think we should use [the `useId()` hook](https://beta.reactjs.org/reference/react/useId) to generate an ID. That way we know it'll be unique.

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,16 +1,106 @@
...
+				<div>
+					<span>
+						<strong>{ __( 'Root Directory' ) }</strong>
+						<small>
+							{ __(
+								'(you can specify a root directory and only changes in that directory will be deployed)'
+							) }
+						</small>
+					</span>
+					<FormTextInput id="root_path" placeholder="/wp-content" />
```
p-jackson: I'm not sure whether this placeholder should be using `__()` 🙃 Probably not right?
I do think the input should have a `dir="ltr"` attribute though so that even in RTL languages the browser knows to render the contents as a LTR string. This is how we present domains in the Calypso UI, so I think it makes sense for file paths too.

p-jackson: Yes I think `dir="ltr"` is what we want.

Without the attribute it's unable to render an absolute path correctly.
CleanShot 2023-02-07 at 21 27 09@2x

This is what it's like with the attribute.
CleanShot 2023-02-07 at 21 28 35@2x

client/my-sites/hosting/github/github-connect-card/use-github-repos.ts
```diff
@@ -0,0 +1,31 @@
...
+		[ USE_GITHUB_REPOS_QUERY_KEY, siteId ],
+		(): Repo[] =>
+			wp.req.get( {
+				path: `/sites/${ siteId }/hosting/github/repos`,
+				apiNamespace: 'wpcom/v2',
+			} ),
+		{
+			enabled: !! siteId,
+			select: ( data ) =>
+				data.map( ( repo: Repo ) => ( {
```
mpkelly: I will simplify the server data to be a list of strings so you can remove this mapping code later on.

client/my-sites/hosting/github/github-connect-card/use-github-branches.ts
```diff
@@ -0,0 +1,30 @@
...
+		[ USE_GITHUB_BRANCHES_QUERY_KEY, repoName, siteId ],
+		(): RepoBranch[] =>
+			wp.req.get( {
+				path: `/sites/${ siteId }/hosting/github/branches?repo=${ repoName }`,
+				apiNamespace: 'wpcom/v2',
+			} ),
+		{
+			enabled: !! siteId && !! repoName,
+			select: ( data ) =>
+				data.map( ( branch: RepoBranch ) => ( {
```
mpkelly: Same as for repos, the server will send a list of strings.

client/my-sites/hosting/github/github-connect-card/index.tsx
mpkelly: The initial value for these two properties, and the base path, should come from site options. I am using these constants:

```php
define( 'GITHUB_HOSTING_REPO_NAME', 'GITHUB_HOSTING_REPO_NAME' );
define( 'GITHUB_HOSTING_BRANCH_NAME', 'GITHUB_HOSTING_BRANCH_NAME' );
define( 'GITHUB_HOSTING_BASE_PATH', 'GITHUB_HOSTING_BASE_PATH' );
```
>We shouldn't need to have a useEffect here.

And then should be able to remove this `useEffect` as mentioned

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,16 +1,106 @@
...
+							onChange={ ( event ) => handleBranchSelect( event.currentTarget.value ) }
+							value={ selectedBranch?.name }
+						>
+							{ branches?.map( ( branch ) => (
+								<option value={ branch.name } key={ branch.name }>
+									{ branch.name }
+								</option>
+							) ) }
+						</FormSelect>
+						<Button
```
mpkelly: Can we connect this up so that it calls `/sites/{site_id}/hosting/github/connect?repo=...&branch=...&base_path=....`

The `base_path` will be optional but the others need to be set before this button can be clicked.

client/my-sites/hosting/github/github-connect-card/index.tsx
vykes-mac: Thanks didn't know about the dir. 

> I'm not sure whether this placeholder should be using __() 🙃 Probably not right?

I think it should remain as is since it's a folder path that's language agnostic. Might be difficult to say how translators should represent a folder path.

client/my-sites/hosting/github/github-connect-card/index.tsx
vykes-mac: They are not used, I'll remove them.

client/my-sites/hosting/github/github-connect-card/index.tsx
vykes-mac: @mpkelly I don't think we need to use those for the connect branch card which is only shown after user is authorized, those will be shown on the disconnect branch card at which point I assume thos site options will also be cleared and the connect branch card will also fetch list of repo and branches again.

@p-jackson I knew something smelled with that `useEffect` since I was using ReactQuery which internally has `useEffect`.  I even had a discussion with chatGPT on it 😆 asking it if it's best practice to use useEffect while using react query. 

> we could pass an onSuccess handler to the query hooks
I like this approach better.

client/my-sites/hosting/github/github-connect-card/use-github-branches.ts
vykes-mac: I don't think branches are changed often and for deploying most times we gonna need master/main branch. I think it's a good candidate for caching, Can't scientifically justify 5 mins, I just thought it would be a good buffer in case user modify their branches. Would you recommend something higher like say 24hr?

p-jackson: Nah I think I'd just leave it as the default.

client/my-sites/hosting/github/github-connect-card/use-github-repos.ts
vykes-mac: maybe we should allow it as is in case we decided to user other fields from the returned object.

p-jackson:
@vykes-mac apologies for the merge conflict. My intention is that you'll render the `<DisconnectGitHubExpander>` at the very bottom of the card which shows the repo/branch selector.

mpkelly:
@vykes-mac, just a heads up that I changed repos and branches to be a list of strings. I made the corresponding change in D100978-code

mpkelly:
I've noticed a couple of small issues after using this UI a bit:
1. Sometimes when you hit "Connect", the dropdown menus reset
2. ~~It's possible to change the repo and then hit "Connect" and connect to a branch from the old repo. I think we need to reset the branches array when the repo changes.~~ I fixed this.

zaguiini:
First of all thanks for the great effort here, it's looking good. We could use a few improvements for it to be great:

- Let’s not make the root directory description small but instead put it after the text box, which is a common pattern across WPCOM forms

- Let’s put `/wp-content` directly in the text box. Let's not use a placeholder and show that `/wp-content` is the default 👍 

- Let’s make the disconnect button visible by default instead of hidden in the accordion

- Let's add a label to both repo and branch text inputs/select

- It would be very good to link to the docs about the GitHub integration, so the user knows what files are forbidden for example

- The list is not alphabetically ordered, making it hard to find the repo

- I think we’ll need to fall back to [SelectDropdown](http://calypso.localhost:3000/devdocs/client/components/select-dropdown/README.md) or something like it that allows searching for a specific repo. In my case, I have 103. By default, GitHub returns 30 repos and it's up to 100, so chances are I'll miss it when trying to connect.

- It’d also be cool to have the repo name and branch in a monospaced font so it's easier for the user to read.

---

Let me know if you'd like to pair or split the work between the two of us, so we can implement them all or split it into two or more PRs.

javierarce:
I echo everything @zaguiini mentioned, and I'd also like to suggest some other minor tweaks:


image

- We should use the same title for the authorized/non-authorized state of the card: I suggest `Deploy from GitHub`
- We could disable the connect button after the link is established. If the user changes the repo, the branch, or the directory, we enable the button back. 

I've also modified the descriptions to make them a bit more concise. 

Here's a Figma file with the designs above: mscADrGBDjxazOtDcy4qAw-fi

client/my-sites/hosting/github/disconnect-github-expander/index.tsx
```diff
@@ -15,40 +13,20 @@ interface DisconnectGitHubExpanderProps {
...
-					busy={ isDisconnecting }
-					disabled={ isDisconnecting } // `busy` doesn't actually disable the button
-				>
-					{ __( 'Disconnect' ) }
-				</Button>
-			</div>
-		</>
+		<Button
+			onClick={ () => disconnect( connection ) }
+			busy={ isDisconnecting }
```
zaguiini: I think we should always display this button as `busy` so it represents a destructive action.

client/my-sites/hosting/github/github-connect-card/index.tsx
```diff
@@ -1,23 +1,132 @@
...
+									<option value={ branch } key={ branch }>
+										{ branch }
+									</option>
+								) ) }
+							</FormSelect>
+						</FormFieldset>
+					</FormFieldset>
+				) }
+			</div>
+			<div style={ { marginTop: '10px', display: 'inline-flex', gap: '16px' } }>
```
zaguiini: Would be nice to have the disconnect button on the other side of the card so the user won't misclick it.

zaguiini (commented):
Can we make the options popover float instead of pushing the content below it, specifically for this instance of `SuggestionSearch`?

image

Something like that does the trick already:

```diff
diff --git a/client/my-sites/hosting/github/github-connect-card/index.tsx b/client/my-sites/hosting/github/github-connect-card/index.tsx
index f1a36f24b1..442243cf79 100644
--- a/client/my-sites/hosting/github/github-connect-card/index.tsx
+++ b/client/my-sites/hosting/github/github-connect-card/index.tsx
@@ -89,6 +89,7 @@ export const GithubConnectCard = ( { connection }: GithubConnectCardProps ) => {
 							<FormLabel htmlFor="repository">{ __( 'Repository' ) }</FormLabel>
 							<Search
 								id="repository"
+								className="connect-branch__repository-field"
 								placeholder={ __( 'Start typing a repo..' ) }
 								options={ repos ?? [] }
 								onSelect={ handleRepoSelect }
diff --git a/client/my-sites/hosting/github/github-connect-card/search.tsx b/client/my-sites/hosting/github/github-connect-card/search.tsx
index 071ba5ef57..060764f083 100644
--- a/client/my-sites/hosting/github/github-connect-card/search.tsx
+++ b/client/my-sites/hosting/github/github-connect-card/search.tsx
@@ -7,14 +7,23 @@ interface SearchProps {
 	options: string[];
 	onSelect( option: string ): void;
 	disabled?: boolean;
+	className?: string;
 }
 
-export const Search = ( { id, placeholder, options, onSelect, disabled }: SearchProps ) => {
+export const Search = ( {
+	id,
+	className,
+	placeholder,
+	options,
+	onSelect,
+	disabled,
+}: SearchProps ) => {
 	const [ query, setQuery ] = useState( '' );
 
 	return (
 		<SuggestionSearch
 			id={ id }
+			className={ className }
 			placeholder={ placeholder }
 			disabled={ disabled }
 			showIcon={ false }
diff --git a/client/my-sites/hosting/github/github-connect-card/style.scss b/client/my-sites/hosting/github/github-connect-card/style.scss
index 839d3d6f40..2dd39131ab 100644
--- a/client/my-sites/hosting/github/github-connect-card/style.scss
+++ b/client/my-sites/hosting/github/github-connect-card/style.scss
@@ -14,6 +14,10 @@
 	width: 100%;
 }
 
+.connect-branch__repository-field .suggestions {
+	position: absolute;
+}
+
 @media ( max-width: 860px ) {
 	.connect-branch__fields {
 		flex-direction: column;

```

zaguiini:
One thing that's striking me as odd is that:

- the `Connect to repository` button is enabled if I remove the text from the repo input;
- the `Connect to repository` button is shown in a loading state even though we're loading branches.

client/my-sites/hosting/github/github-connect-card/index.tsx
zaguiini: I think we can get rid of this top margin 👍

client/my-sites/hosting/github/disconnect-github-expander/index.tsx
vykes-mac: is that a common pattern? this seems like it would draw more attention to this button

client/my-sites/hosting/github/github-connect-card/index.tsx
vykes-mac: I tried but the UI looks unbalance.
image

vykes-mac:
@zaguiini 

> the Connect to repository button is shown in a loading state even though we're loading branches.

This is intentional to indicate some fetching is happening. You think we should remove it?

zaguiini:
>You think we should remove it?

I think we should have a separate loading indicator for the branch fetching. We should disable the Connect button when we have either repo or branch empty 👍

client/my-sites/hosting/github/github-connect-card/index.tsx
zaguiini: It does 😢

Can we try moving the CTA to the top-right corner of the screen and moving the disconnect button to the bottom left?

Another alternative would be to flip the buttons (move disconnect to the left). Or, keep it on the right, but remove the border and leave it as red, just text -- the `transparent` variant

Or, if we're feeling like it, we can just get rid of this button altogether 🤷‍♂️  I can't imagine a scenario where a user would need to totally disconnect GitHub. They might disconnect the repo, but unauthorize GitHub? Can't see that happening. And if it does, I don't think it's gonna be frequent.

client/my-sites/hosting/github/disconnect-github-expander/index.tsx
zaguiini: It is... most destructive CTAs are red. We can either make it busy all the time or remove the border, making it less prominent.

zaguiini: Oh, sorry about that. I mistakenly thought of `busy` as the destructive, red button. But `busy` actually means  `loading` 🤦 

It's supposed to be `scary` and preferentially `borderless` too, so it doesn't draw too much attention, but it's still there.

client/my-sites/hosting/github/github-connect-card/index.tsx
zaguiini: cc @javierarce @mpkelly would like some input on the above ☝️

p-jackson:
> Let’s make the disconnect button visible by default instead of hidden in the accordion

Since the disconnect button no longer has the "scary" style, IMO we should add a confirmation dialog.

client/my-sites/hosting/github/github-connect-card/index.tsx
mpkelly: image

I like this new UI 👍 . Visually, I think the buttons are ok as they are. But pinning them in the corners also works for me. 

We can improve it by showing that a query is in progress and by explicitly showing that no results were returned If you type "not-a-repo-name", it just sits there and is no visual feedback. 

>Another alternative would be to flip the buttons

On this page, the other cards have the CTA on the bottom left side. 

>Or, if we're feeling like it, we can just get rid of this button altogether 🤷‍♂️ 

I like having a complete reset option. The user can modify stuff at GitHub that could break this feature, so we should have a way to reset it. I agree with @javierarce on Slack that it should be more difficult to disconnect once you've connected a repo.

mpkelly:
Found an issue where the search input text gets reset:

https://user-images.githubusercontent.com/6851384/219307446-73bd8d68-936a-44d6-8782-fc2bbd1e9cb7.mp4

mpkelly:
I think it would be good to get this merged and create a new task for any follow-up work.

zaguiini:
This is how it looks now. Let's merge it!

image

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7820502

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7820502).

PR #73032, opened 2023-02-07
Update backup features tests to use strings from function

CodeyGuyDylan:
#### Proposed Changes

Updating tests to pull strings from function instead of redefining them in tests

This was at the request of @creativecoder [here](https://github.com/Automattic/wp-calypso/pull/72525#discussion_r1094902203)

#### Testing Instructions

1. From the root directory, run the following command
`yarn test-client -- client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx`
2. Make sure tests still pass

#### Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx
+++ b/client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx
@@ -12,9 +12,11 @@ import { checkoutTheme } from '@automattic/composite-checkout';
 import { ShoppingCartProvider, createShoppingCartManagerClient } from '@automattic/shopping-cart';
 import { ThemeProvider } from '@emotion/react';
 import { render, screen, waitFor } from '@testing-library/react';
+import { useTranslate } from 'i18n-calypso';
 import React from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
 import WPCheckoutOrderSummary from '../components/wp-checkout-order-summary';
+import getJetpackProductFeatures from '../lib/get-jetpack-product-features';
 import {
 	mockSetCartEndpointWith,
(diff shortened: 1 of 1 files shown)


client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx
```diff
@@ -32,13 +34,9 @@ jest.mock( '@automattic/calypso-config', () => {
...
-const jetpackBackupFeatureList = [
-	'Real-time cloud backups',
-	'10GB of backup storage',
-	'30-day archive & activity log',
-	'One-click restores',
-	'Priority support',
-];
+const translate: ( original: string ) => string = ( original ) => {
+	return original;
+};
```
creativecoder: In case you didn't know, a function like this is often called an "identity" function. While I see you're trying to base it off the existing function type, I think you could simplify the stub even more to make it more readable.

See this other test that does the same stub of the translate function as an example: https://github.com/Automattic/wp-calypso/blob/b55a5036225d4faf6b94e0eff474816fa8e3b5e1/client/my-sites/plan-features/test/index.jsx#L43

creativecoder (approved):
This is a nice improvement to prevent the need for updating tests if the feature strings ever change.

Left one non-blocking suggestion to simplify the translate function stub, otherwise looks good to me, as I see that tests are all passing in CI.

client/my-sites/checkout/composite-checkout/test/wp-checkout-order-summary.tsx
CodeyGuyDylan: Thanks for the review and the tip! Got that updated 😄

PR #73033, opened 2023-02-07
Filter out duplicate pending comments

### LINKED ISSUE #61812: [Bug]: Reader - Duplicate comment temporarily appears when making a new comment
### Quick summary

When adding a new comment to a post in the reader, a duplicate comment temporarily appears: 

Screen Capture on 2022-03-10 at 10-13-40

### Steps to reproduce

1. Go to [wordpress.com/read/](https://wordpress.com/read/) to access Reader
2. Navigate to any posts comment section
3. Create a comment
4. See duplicate comment briefly appear 

Expected to only see one comment appear. 

### What actually happened

A second comment briefly appears. 

### Browser

Google Chrome/Chromium, Mozilla Firefox

### Simple/Atomic

Simple

### Other notes

Safari behaves a bit differently but it still seems like the comment is being added twice. In GIF below, wait to see second highlight to see Safari result: 

Screen Capture on 2022-03-10 at 10-28-55

### Reproducibility

Consistent

### Severity

All

roo2:
#### Proposed Changes
Fixes https://github.com/Automattic/wp-calypso/issues/61812

I found on inspecting the commentTree object, that in some instances it can contain duplicates where there is a pending comment, and an approved comment that briefly exist at the same time. This causes the issue described in https://github.com/Automattic/wp-calypso/issues/61812.

This is a somewhat "quick and dirty" patch as I haven't dug right down to the cause of why there can sometimes be duplicates, but it fixes the issue and I don't think it adds too much complexity.

#### Testing Instructions
see https://github.com/Automattic/wp-calypso/issues/61812#issuecomment-1419023924
I personally found that I could only reproduce the issue when
* I wasn't sandboxed
* I was replying to a post that already had many (10+) comments.

### DIFF
--- a/client/state/comments/selectors/get-post-comments-tree.js
+++ b/client/state/comments/selectors/get-post-comments-tree.js
@@ -28,17 +28,6 @@ export const getPostCommentsTree = treeSelect(
 			) {
 				return false;
 			}
-
-			// Filter out pending comments that are duplicates of approved comments
-			if ( item.status === 'pending' ) {
-				const duplicate = allItems.find(
-					( comment ) => comment.content === item.content && comment.ID !== item.ID
-				);
-				if ( duplicate ) {
-					return false;
-				}
-			}
-
 			if ( status !== 'all' ) {
 				return item.isPlaceholder || item.status === status;
 			}
--- a/client/state/comments/selectors/get-post-comments-tree.js
+++ b/client/state/comments/selectors/get-post-comments-tree.js
@@ -28,6 +28,17 @@ export const getPostCommentsTree = treeSelect(
 			) {
 				return false;
 			}
+
+			// Filter out pending comments that are duplicates of approved comments
+			if ( item.status === 'pending' ) {
+				const duplicate = allItems.find(
+					( comment ) => comment.content === item.content && comment.ID !== item.ID
+				);
+				if ( duplicate ) {
+					return false;
+				}
+			}
+
 			if ( status !== 'all' ) {
 				return item.isPlaceholder || item.status === status;
 			}
--- a/client/state/lasagna/post-channel/middleware.js
+++ b/client/state/lasagna/post-channel/middleware.js
@@ -57,14 +57,19 @@ const joinChannel = async ( store, joinParams ) => {
 			return;
 		}
 
-		store.dispatch(
-			receiveComments( {
-				siteId: comment.post.site_ID,
-				postId: comment.post.ID,
-				comments: [ comment ],
-				commentById: true,
-			} )
-		);
+		// If the comment's author is 0, that means it was authored by the current user.
+		// In that case do not dispatch the event as it may arrive before the pending comment has been deleted,
+		// causing a duplicate comment. See https://github.com/Automattic/wp-calypso/issues/61812
+		if ( comment.author && comment.author.ID !== 0 ) {
+			store.dispatch(
+				receiveComments( {
+					siteId: comment.post.site_ID,
+					postId: comment.post.ID,
+					comments: [ comment ],
+					commentById: true,
+				} )
+			);
+		}
 	} );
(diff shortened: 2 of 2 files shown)


client/state/comments/selectors/get-post-comments-tree.js
```diff
@@ -28,6 +28,17 @@ export const getPostCommentsTree = treeSelect(
...
+
+			// Filter out pending comments that are duplicates of approved comments
+			if ( item.status === 'pending' ) {
+				const duplicate = allItems.find(
+					( comment ) => comment.content === item.content && comment.ID !== item.ID
+				);
+				if ( duplicate ) {
+					return false;
+				}
+			}
```
tyxla: I think this solution has some downsides - what happens if someone submits a pending comment that has the same content as an existing approved comment? 

I think we have to look for the original reason for the comment to get duplicated and fix it there. It could be something we're doing twice in the reducer, or it could be two conflicting actions - one that adds the comment optimistically and then one that adds it after it has actually been submitted. Maybe those two are supposed to be compared and the second one would be inserted only if the comparison fails, and the comparison has been failing for some time? 

I'd suggest diving deeper and finding the root cause, instead of patching it up and concealing it under yet another level of complexity. 

I'd also encourage writing some tests. This is a great use case for some component tests that ensure this is being fixed, and won't happen again. I'm happy to provide on how to write component tests with `@testing-library/react` if that is going to help.

tyxla (changes requested):
Thanks for working on it @roo2! 

Unfortunately, that doesn't seem to solve the issue, but rather hides it under yet another level of complexity. See my inline comment for more details.

client/state/comments/selectors/get-post-comments-tree.js
roo2: ok, I will keep digging 👍

roo2:
I've found that the reason the extra comment is being added is because apart from the chain of redux events that is expected, the `Lasagna` middleware subscribes to `new_comment` events and dispatches the `receiveComment` action directly https://github.com/Automattic/wp-calypso/blob/trunk/client/state/lasagna/post-channel/middleware.js#L61.

I'm looking into what lasagna is, and if it can be turned off in this case or if we can work around having `receiveComment` called immediately when the new comment is added.

roo2:
OK, @tyxla I think I have a solution.

Lasagna is used to asynchronously load other people's comments via a websocket so that you don't have to reload the page to see new comments. It is redundant and comments will otherwise be loaded manually on page load.

Lasagna isn't needed to load the users own comments; they are loaded manually.

The users own comments come through with `comment.author.ID === 0` so it is easy to filter them out and stop lasagna from raising `receiveComments` actions for them.

🎉

client/state/lasagna/post-channel/middleware.js
```diff
@@ -57,14 +57,19 @@ const joinChannel = async ( store, joinParams ) => {
...
-			receiveComments( {
-				siteId: comment.post.site_ID,
-				postId: comment.post.ID,
-				comments: [ comment ],
-				commentById: true,
-			} )
-		);
+		// If the comment's author is 0, that means it was authored by the current user.
+		// In that case do not dispatch the event as it may arrive before the pending comment has been deleted,
+		// causing a duplicate comment. See https://github.com/Automattic/wp-calypso/issues/61812
```
tyxla, line 62: Thanks for adding an explanation 👍

tyxla (approved):
Works well and the code looks good to me 👍 

Thanks @roo2 🙌 🚀

client/state/lasagna/post-channel/middleware.js
roo2: yes this was a confusing one but I think it makes sense with an explanation. Thank you 😄

PR #73031, opened 2023-02-07
Jetpack Backup: Fetch last backup size

Initsogar:
#### Proposed Changes

* Add the necessary logic to fetch the last backup size as part of the `rewind/size` API call. It includes reducer, selector and tests.

#### Testing Instructions

* Ensure running testing passes by executing `yarn run test-client client/state/rewind`
* You can validate that the new `state.rewind[site_id].size.lastBackupSize` has been added to the state using Redux DevTools or developer console by running: `state.rewind[YOUR_BLOG_ID].size.lastBackupSize`

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] ~Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)~
- [ ] ~Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?~
- [ ] ~For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~

PR #73007, opened 2023-02-06
Jetpack Backup: Update component to use custom backup retention days if set

rustamshaikh:
#### Proposed Changes

* Update the component to make use of a custom backup retention period (in days) if set by the customer otherwise use the retention from the plan.

Requires D100421-code on VP.

#### Testing Instructions

1) Take a site whose backups are being discarded due to the storage limit has been reached.
2) Check the current backup dashboard status, it should be similar to the below screenshot. Ensure that no custom retention has been set yet or delete if it is, from the vpshell.
     Screenshot 2023-02-07 at 5 49 05 PM
3) Now set a custom retention period lower than that of the current `days_of_backups_allowed` from the size API either by following D100335-code or via from vpshell (e.g. in D100421-code)
4) Refresh the Backup Dashboard on Calypso, it should show normal usage with no warning message as below:
     Screenshot 2023-02-07 at 5 55 28 PM

Also, ensure unit tests passes by executing `yarn run test-client client/state/rewind`

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to D100421-code

PR #73035, opened 2023-02-07
Jetpack Backup: Add Backup retention management setting

Initsogar:
#### Proposed Changes

* Add a new `Days of backups saved` (backup retention period) setting in Jetpack Cloud, under Settings page. For this PR, it will be available only for A12s (dev & stage environment). The idea is to start the review process of this new functionality.
* This iteration includes:
  * Allow to update the setting only to a number of days that doesn't require a storage upgrade.
  * Fetch and update the customer-defined retention period to our backend.
  * Confirmation dialog when attempting to update to a lower retention period.

#### Demo

https://user-images.githubusercontent.com/1488641/218225112-ce96783e-d079-447b-8df2-24d9f54666fd.mov

↪ [Demo v1 (deprecated)](https://user-images.githubusercontent.com/1488641/217998763-cc89187c-9ce8-4536-a8fb-785c153ab76c.mov)

#### Screenshots

| Description | Screenshot |
|---|---|
| Setting page | image |
| Confirmation dialog | image |
| Tooltip | image |
| Success notice | image |
| Error notice | image |

#### Testing Instructions

* The `Update settings` button will be disabled when the current plan or an option that requires an upgrade are selected.

Notes:
* If you haven't configured this setting before, the current retention period associated to your plan will be preselected. Otherwise, your current configuration will be preselected.

#### Pre-merge Checklist

### DIFF
--- a/client/components/backup-retention-management/constants.ts
+++ b/client/components/backup-retention-management/constants.ts
@@ -1,0 +1,11 @@
+export const RETENTION_OPTIONS = {
+	RETENTION_DAYS_7: 7,
+	RETENTION_DAYS_30: 30,
+	RETENTION_DAYS_120: 120,
+	RETENTION_DAYS_365: 365,
+} as const;
+
+export const STORAGE_ESTIMATION_ADDITIONAL_BUFFER = 0.25;
+
+export const STORAGE_RETENTION_LEARN_MORE_LINK =
+	'https://jetpack.com/support/backup/#how-is-storage-usage-calculated';
--- a/client/components/backup-retention-management/consts.tsx
+++ b/client/components/backup-retention-management/consts.tsx
@@ -1,0 +1,15 @@
+export const RetentionOptions = {
+	RETENTION_DAYS_7: 7,
+	RETENTION_DAYS_30: 30,
+	RETENTION_DAYS_120: 120,
+	RETENTION_DAYS_365: 365,
+} as const;
+
+export interface RetentionRadioOptionType {
+	label: string; // 7 days
+	spaceNeeded: any; // 700MB
+	upgradeRequired: boolean;
+	isCurrentPlan: boolean;
+	value: number; // 7
+	checked: boolean;
+}
--- a/client/components/backup-retention-management/consts.tsx
+++ b/client/components/backup-retention-management/consts.tsx
@@ -1,6 +1,0 @@
-export const RetentionOptions = {
-	RETENTION_DAYS_7: 7,
-	RETENTION_DAYS_30: 30,
-	RETENTION_DAYS_120: 120,
-	RETENTION_DAYS_365: 365,
-} as const;
--- a/client/components/backup-retention-management/consts.tsx
+++ b/client/components/backup-retention-management/consts.tsx
@@ -1,3 +1,5 @@
+import type { TranslateResult } from 'i18n-calypso';
+
 export const RetentionOptions = {
 	RETENTION_DAYS_7: 7,
 	RETENTION_DAYS_30: 30,
@@ -7,7 +9,7 @@ export const RetentionOptions = {
 
 export interface RetentionRadioOptionType {
 	label: string; // 7 days
-	spaceNeeded: any; // 700MB
+	spaceNeeded: TranslateResult | string; // 700MB
 	upgradeRequired: boolean;
 	isCurrentPlan: boolean;
 	value: number; // 7
--- a/client/components/backup-retention-management/consts.tsx
+++ b/client/components/backup-retention-management/consts.tsx
@@ -1,17 +1,6 @@
-import type { TranslateResult } from 'i18n-calypso';
-
 export const RetentionOptions = {
 	RETENTION_DAYS_7: 7,
 	RETENTION_DAYS_30: 30,
 	RETENTION_DAYS_120: 120,
 	RETENTION_DAYS_365: 365,
 } as const;
-
-export interface RetentionRadioOptionType {
-	label: string; // 7 days
-	spaceNeeded: TranslateResult | string; // 700MB
-	upgradeRequired: boolean;
-	isCurrentPlan: boolean;
-	value: number; // 7
-	checked: boolean;
-}
--- a/client/components/backup-retention-management/hooks.tsx
+++ b/client/components/backup-retention-management/hooks.tsx
@@ -2,6 +2,7 @@ import { useMemo } from '@wordpress/element';
 import { useSelector } from 'react-redux';
 import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
 import getActivityLogVisibleDays from 'calypso/state/rewind/selectors/get-activity-log-visible-days';
+import getBackupCurrentSiteSize from 'calypso/state/rewind/selectors/get-backup-current-site-size';
 import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import { RetentionRadioOptionType } from './consts';
@@ -10,7 +11,10 @@ import { RetentionRadioOptionType } from './consts';
  * Estimate the current site size based on the last backup size, adding a 25% buffer
  */
 export function useEstimatedCurrentSiteSize(): number {
-	const lastBackupSize = 2 ** 30; // @TODO: Fetch this from #73031 when it's ready. We will need to call getBackupCurrentSiteSize() from the state.
+	const siteId = useSelector( getSelectedSiteId ) as number;
+	const lastBackupSize = useSelector(
+		( state ) => getBackupCurrentSiteSize( state, siteId ) as number
+	);
 	return useMemo( () => lastBackupSize + lastBackupSize * 0.25, [ lastBackupSize ] );
 }
 
--- a/client/components/backup-retention-management/hooks.tsx
+++ b/client/components/backup-retention-management/hooks.tsx
@@ -1,0 +1,73 @@
+import { useMemo } from '@wordpress/element';
+import { useSelector } from 'react-redux';
+import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
+import getActivityLogVisibleDays from 'calypso/state/rewind/selectors/get-activity-log-visible-days';
+import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import { RetentionRadioOptionType } from './consts';
+
+/**
+ * Estimate the current site size based on the last backup size, adding a 25% buffer
+ */
+export function useEstimatedCurrentSiteSize(): number {
+	const lastBackupSize = 2 ** 30; // @TODO: Fetch this from #73031 when it's ready. We will need to call getBackupCurrentSiteSize() from the state.
+	return useMemo( () => lastBackupSize + lastBackupSize * 0.25, [ lastBackupSize ] );
+}
+
+/**
+ * Estimate the space needed based on the current site size and the retention days
+ *
+ * @param currentSiteSizeInBytes Current site size in bytes
+ * @param desiredRetentionDays   Desired backup retention days
+ * @returns Space needed in bytes
+ */
+export function useEstimateSpaceNeeded(
+	currentSiteSizeInBytes: number,
+	desiredRetentionDays: number
+): number {
+	return useMemo( () => {
+		return currentSiteSizeInBytes * desiredRetentionDays;
+	}, [ currentSiteSizeInBytes, desiredRetentionDays ] );
+}
+
+/**
+ * Prepare the retention options for the RetentionOptionsControl component
+ */
+export function usePrepareRetentionOptions(
+	label: string,
+	desiredRetentionDays: number,
+	checked: boolean
+): RetentionRadioOptionType {
+	const siteId = useSelector( getSelectedSiteId ) as number;
+	const currentSiteSizeInBytes = useEstimatedCurrentSiteSize();
+
+	const storageLimitBytes = useSelector( ( state ) =>
+		getRewindBytesAvailable( state, siteId )
+	) as number;
+
+	const planRetentionPeriod = useSelector( ( state ) =>
+		getActivityLogVisibleDays( state, siteId )
+	);
+
+	const spaceNeeded = useEstimateSpaceNeeded( currentSiteSizeInBytes, desiredRetentionDays );
+	const spaceNeededText = useStorageText( spaceNeeded );
+
+	return useMemo( () => {
+		return {
+			label,
+			spaceNeeded: spaceNeededText,
+			upgradeRequired: spaceNeeded > storageLimitBytes,
+			isCurrentPlan: desiredRetentionDays === planRetentionPeriod,
+			value: desiredRetentionDays,
+			checked,
+		};
+	}, [
+		label,
+		spaceNeededText,
+		spaceNeeded,
+		storageLimitBytes,
+		desiredRetentionDays,
+		planRetentionPeriod,
+		checked,
+	] );
+}
--- a/client/components/backup-retention-management/hooks.tsx
+++ b/client/components/backup-retention-management/hooks.tsx
@@ -4,7 +4,7 @@ import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
 import getBackupCurrentSiteSize from 'calypso/state/rewind/selectors/get-backup-current-site-size';
 import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
-import { RetentionRadioOptionType } from './consts';
+import type { RetentionRadioOptionType } from './types';
 
 /**
  * Estimate the current site size based on the last backup size, adding a 25% buffer
--- a/client/components/backup-retention-management/hooks.tsx
+++ b/client/components/backup-retention-management/hooks.tsx
@@ -4,6 +4,7 @@ import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
 import getBackupCurrentSiteSize from 'calypso/state/rewind/selectors/get-backup-current-site-size';
 import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import { STORAGE_ESTIMATION_ADDITIONAL_BUFFER } from './constants';
 import type { RetentionRadioOptionType } from './types';
 
 /**
@@ -14,7 +15,10 @@ export function useEstimatedCurrentSiteSize(): number {
 	const lastBackupSize = useSelector(
 		( state ) => getBackupCurrentSiteSize( state, siteId ) as number
 	);
-	return useMemo( () => lastBackupSize + lastBackupSize * 0.25, [ lastBackupSize ] );
+	return useMemo(
+		() => lastBackupSize + lastBackupSize * STORAGE_ESTIMATION_ADDITIONAL_BUFFER,
+		[ lastBackupSize ]
+	);
 }
 
 /**
--- a/client/components/backup-retention-management/hooks.tsx
+++ b/client/components/backup-retention-management/hooks.tsx
@@ -1,7 +1,6 @@
 import { useMemo } from '@wordpress/element';
 import { useSelector } from 'react-redux';
 import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
-import getActivityLogVisibleDays from 'calypso/state/rewind/selectors/get-activity-log-visible-days';
 import getBackupCurrentSiteSize from 'calypso/state/rewind/selectors/get-backup-current-site-size';
 import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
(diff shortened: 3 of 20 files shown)


grbicsanja:
This is looking great! 🙌
Can we please have the confirmation notice say:
`You have successfully changed the time period of saved backups to X days`

client/components/backup-retention-management/hooks.tsx
```diff
@@ -0,0 +1,73 @@
+import { useMemo } from '@wordpress/element';
+import { useSelector } from 'react-redux';
+import { useStorageText } from 'calypso/components/backup-storage-space/hooks';
+import getBackupCurrentSiteSize from 'calypso/state/rewind/selectors/get-backup-current-site-size';
+import getRewindBytesAvailable from 'calypso/state/rewind/selectors/get-rewind-bytes-available';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import { RetentionRadioOptionType } from './consts';
```
rustamshaikh:
```suggestion
import type { RetentionRadioOptionType } from './consts';
```
Since we are importing type.

client/components/backup-retention-management/hooks.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Estimate the current site size based on the last backup size, adding a 25% buffer
+ */
+export function useEstimatedCurrentSiteSize(): number {
+	const siteId = useSelector( getSelectedSiteId ) as number;
+	const lastBackupSize = useSelector(
+		( state ) => getBackupCurrentSiteSize( state, siteId ) as number
+	);
+	return useMemo( () => lastBackupSize + lastBackupSize * 0.25, [ lastBackupSize ] );
```
rustamshaikh: `0.25` - Maybe we can define it as a constant?

client/components/backup-retention-management/consts.tsx
```diff
@@ -0,0 +1,17 @@
+import type { TranslateResult } from 'i18n-calypso';
+
+export const RetentionOptions = {
+	RETENTION_DAYS_7: 7,
+	RETENTION_DAYS_30: 30,
+	RETENTION_DAYS_120: 120,
+	RETENTION_DAYS_365: 365,
+} as const;
+
+export interface RetentionRadioOptionType {
```
rustamshaikh: Maybe we can move this type `RetentionRadioOptionType` to a 'type only' file say `types.ts`

grbicsanja:
The info icon and tooltip is missing. [See design here](https://www.figma.com/file/zEkRPQwNT9aykpCyJvrSt5/Allow-customers-to-define-their-own-backup-retention?node-id=3147%3A38368&t=u5jSmMRkhRo6u8Sm-4).
Screenshot 2023-02-10 at 13 26 14

grbicsanja:
I see the warning prompt even when I'm choosing a longer time period. It should only be shown when I am reducing the time and loosing saved backups.

Here in this screenshot, I was moving from 7 days to a year.
Screenshot 2023-02-10 at 13 27 26

client/components/backup-retention-management/retention-options/retention-options-control.tsx
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState, useCallback } from '@wordpress/element';
+import React from 'react';
+import { RetentionRadioOptionType } from '../consts';
```
rustamshaikh: `import type` - Same w.r.t import type and separating constants from types.

client/components/backup-storage-space/hooks.tsx
```diff
@@ -111,3 +111,49 @@ export const useDaysOfBackupsSavedText = (
...
+					if ( unitAmount % 1 === 0 ) {
+						return translate( '%(storageInBytes)dTB', {
+							args: { storageInBytes: unitAmount },
+							comment: 'Must use unit abbreviation; describes an storage amounts (e.g., 1TB)',
+						} );
+					}
+
+					return translate( '%(storageInBytes).1fTB', {
+						args: { storageInBytes: unitAmount },
+						comment: 'Must use unit abbreviation; describes an storage amounts (e.g., 1.5TB)',
```
rustamshaikh:
```suggestion
						comment: 'Must use unit abbreviation; describes an storage amounts with 1 decimal point (e.g., 1.5TB)',
```
Nitpicking, but it may be helpful to translators.

client/jetpack-cloud/sections/settings/controller.tsx
```diff
@@ -22,19 +24,24 @@ export const advancedCredentials: PageJS.Callback = ( context, next ) => {
...
-					siteId={ siteId }
-					trueComponent={ sectionElt }
-					falseComponent={ <NoSitesPurchasesMessage /> }
-					loadingComponent={ <AdvancedCredentialsLoadingPlaceholder /> }
-				/>
-			}
-			loadingComponent={ <AdvancedCredentialsLoadingPlaceholder /> }
-		/>
+		<>
+			{ config.isEnabled( 'jetpack/backup-retention-settings' ) ? (
```
rustamshaikh, line 28: Great!

rustamshaikh (commented):
Great work @Initsogar 👍 Tested it from Jetpack Cloud live and works as expected. Left some feedback. Let me know what you think.

rustamshaikh:
In addition to what @grbicsanja said, these are some UI-related differences that might have been missed (in any case forgive me I am not a good designer if any 🙂)

The height of the loading button:
Screenshot 2023-02-10 at 6 51 34 PM

The spacings which don't match with Design P2 are highlighted (with <--) (Uploaded Design P2 screenshot for simplicity).

Screenshot 2023-02-10 at 6 43 08 PM

client/components/backup-storage-space/hooks.tsx
Initsogar: Done! Thanks for the suggestion.

client/components/backup-retention-management/hooks.tsx
Initsogar: Added it to the constants.ts file, thanks!

client/components/backup-retention-management/consts.tsx
Initsogar: I agree. Just moved it to a new types.ts file.

Initsogar:
Thanks @grbicsanja and @rustamshaikh, just addressed all your comments and suggestions:
* Added tooltip. Also included it on the PR description, so we can double check the wording there.
* Updated success notice text to `You have successfully changed the time period of saved backups to X days`
* Update confirmation dialog to only be required when updating to a lower retention period
* Adjusted translations comments
* Refactored types and constants
* Adjusted spacing between elements

rustamshaikh (approved):
Verified the fixes and it looks great to me 🎉 - Thanks @Initsogar for adding more tests 👍

client/components/backup-storage-space/hooks.tsx
```diff
@@ -111,3 +111,50 @@ export const useDaysOfBackupsSavedText = (
...
+/**
+ * The idea is to convert any storage amount in bytes to a human readable format.
+ *
+ * @param storageInBytes The storage amount in bytes
+ * @returns				 The storage amount in a human readable format
+ */
+export const useStorageText = ( storageInBytes: number ): TranslateResult | string => {
+	const translate = useTranslate();
+
+	return useMemo( () => {
```
tyxla, line 124: Is the memoization here worth the effort? What's the performance impact with and without it?

client/components/backup-retention-management/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { TranslateResult } from 'i18n-calypso';
+
+export interface RetentionRadioOptionType {
+	label: string; // 7 days
+	spaceNeeded: TranslateResult | string; // 700MB
+	upgradeRequired: boolean;
+	isCurrentPlan: boolean;
+	value: number; // 7
```
tyxla, line 8: Is this comment required or incomplete perhaps?

```suggestion
	value: number;
```

client/components/backup-retention-management/loading/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
```
tyxla: Should we import only the `React.FC` type if we really need that only?

client/components/backup-retention-management/loading/style.scss
```diff
@@ -0,0 +1,16 @@
+.backup-retention-management__loading {
+	&.loading {
```
tyxla, line 2: If this adds an extra `loading` class, why do we need the `__loading` suffix for `backup-retention-management__loading` ?

client/components/backup-retention-management/info-tooltip/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+						onClick={ closeTooltip }
+					>
+						<Gridicon icon="cross" size={ 18 } />
+					</Button>
+				</p>
+				<hr />
+				{ translate( '{{a}}Learn more…{{/a}}', {
+					components: {
+						a: (
+							<a
```
tyxla, line 57: Could use `<ExternalLink />` here.

client/components/backup-retention-management/info-tooltip/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+					<Button
+						borderless
+						compact
+						className="retention-setting-info-tooltip__close-tooltip"
+						onClick={ closeTooltip }
+					>
+						<Gridicon icon="cross" size={ 18 } />
+					</Button>
+				</p>
+				<hr />
```
tyxla, line 53: Using `hr`s is a suboptimal way to lay out the presentation. Could be a better bet to add an extra wrapper and use CSS.

client/components/backup-retention-management/info-tooltip/index.tsx
```diff
@@ -0,0 +1,69 @@
+import { Button, Gridicon } from '@automattic/components';
+import { useTranslate } from 'i18n-calypso';
+import * as React from 'react';
```
tyxla: Do we need all of `react` to be imported?

client/components/backup-retention-management/retention-options/retention-option-card.tsx
```diff
@@ -0,0 +1,75 @@
+import classnames from 'classnames';
+import { useTranslate } from 'i18n-calypso';
+import React from 'react';
```
tyxla: Same on importing all of `react`

client/components/backup-retention-management/retention-options/retention-option-card.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type { TranslateResult } from 'i18n-calypso';
+
+interface RetentionOptionCardProps {
+	label: string;
+	spaceNeeded: TranslateResult | string;
```
tyxla, line 8: Won't the `TranslateResult` include `string` already?

Also, it's an odd prop name - it sounds like a boolean but it's not. Thoughts?

client/components/backup-retention-management/retention-options/retention-option-card.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	return (
+		<div
+			className="retention-option"
+			role="radio"
+			aria-checked={ checked }
+			onClick={ handleOptionClick }
+			onKeyDown={ handleOptionClick }
+			tabIndex={ 0 }
+		>
```
tyxla, line 39: Any more accessible way to compose a radio button? We have a bunch of radio control components available for use.

client/components/backup-retention-management/retention-options/retention-options-control.tsx
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}, [ selectedOption, onChange ] );
+
+	const handleRetentionOptionChange = useCallback( ( value: number ) => {
+		setSelectedOption( value );
+	}, [] );
+
+	return (
+		<div className="retention-form__options">
+			{ Object.values( retentionOptions ).map( ( option ) => (
```
tyxla, line 31: If we're going to use `retentionOptions` as an array of values, should we reform the component to expect an array and use that instead of calling `Object.values()`  on an object every time?

client/components/backup-retention-management/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+	useEffect( () => {
+		if ( updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.SUCCESS ) {
+			// Update the current retention plan to the one updated by the user.
+			setCurrentRetentionPlan( retentionSelected );
+			setConfirmationDialogVisible( false );
+		} else if ( updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.FAILED ) {
+			setConfirmationDialogVisible( false );
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
tyxla, line 159: Why is this necessary?

client/components/backup-retention-management/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+			currentRetentionPlan,
+			retentionSelected === RETENTION_OPTIONS.RETENTION_DAYS_120
+		),
+		[ RETENTION_OPTIONS.RETENTION_DAYS_365 ]: usePrepareRetentionOptions(
+			translate( '1 year' ),
+			RETENTION_OPTIONS.RETENTION_DAYS_365,
+			currentRetentionPlan,
+			retentionSelected === RETENTION_OPTIONS.RETENTION_DAYS_365
+		),
+	};
```
tyxla, line 103: Do we really need to redeclare those on every render?

client/components/backup-retention-management/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const RETENTION_OPTIONS = {
+	RETENTION_DAYS_7: 7,
```
tyxla, line 2: Is there a good reason to have those declared in an object? They seem pretty simple and used only in one place.

client/components/backup-retention-management/hooks.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * @param desiredRetentionDays   Desired backup retention days
+ * @returns Space needed in bytes
+ */
+export function useEstimateSpaceNeeded(
+	currentSiteSizeInBytes: number,
+	desiredRetentionDays: number
+): number {
+	return useMemo( () => {
+		return currentSiteSizeInBytes * desiredRetentionDays;
+	}, [ currentSiteSizeInBytes, desiredRetentionDays ] );
```
tyxla, line 37: Have we evaluated the need for each of those memoizations? What's the performance impact and why do we need them? Can we do without them? Would be good to measure and use that to decide whether we need them.

tyxla (commented):
Hey folks! This is a pretty big PR, have you considered splitting it into smaller pieces? Usually, we recommend such bigger PRs to land in smaller places because that allows us to thoroughly test and review, and ensure that we don't introduce regressions. The more the code, the more difficult it is to catch a bug or regression.

Furthermore, I've left some drive-by feedback and questions.

Please let me know if you're interested in me reviewing a particular aspect of the PR.

Thanks 🙌

client/components/backup-retention-management/index.tsx
Initsogar: @tyxla in this case, I would like this check to occur only when `updateRetentionRequestStatus` changes. If I don't include this line, the linter will keep expecting to pass all as dependencies.

tyxla: Technically, the rest of the dependencies can also change and that theoretically should re-run the effect. Is there a good reason to seek workarounds for this default behavior?

client/components/backup-retention-management/info-tooltip/index.tsx
Initsogar: You are right! Just fixed it.

client/components/backup-retention-management/retention-options/retention-option-card.tsx
Initsogar: Fixed!

client/components/backup-retention-management/loading/index.tsx
Initsogar: Fixed!

Initsogar:
> Hey folks! This is a pretty big PR, have you considered splitting it into smaller pieces? Usually, we recommend such bigger PRs to land in smaller places because that allows us to thoroughly test and review, and ensure that we don't introduce regressions. The more the code, the more difficult it is to catch a bug or regression.
> 
> Furthermore, I've left some drive-by feedback and questions.

@tyxla thanks for all this valuable feedback! Indeed, it has been quite complicated to make this PR smaller. We tried to add some dependencies in separate PRs in #73031, #73108 and #73007, but I agree that we could do better.

As we need to continue iterating on this new component, I would appreciate it if we could merge this PR and address the comments and suggestion in a new PR. What do you think?

tyxla:
> @tyxla thanks for all this valuable feedback! Indeed, it has been quite complicated to make this PR smaller. We tried to add some dependencies in separate PRs in #73031, #73108 and #73007, but I agree that we could do better.
> 
> As we need to continue iterating on this new component, I would appreciate it if we could merge this PR and address the comments and suggestion in a new PR. What do you think?

I really appreciate the effort. Ideally, we should approach changes with the atomic change in mind, and keep large PRs like this only for testing or prototyping. I realize it's hard and time-taking sometimes, but it's worth the effort and our future selves will thank us for that. Anyway, I won't be blocking this PR if you can commit to addressing the feedback in subsequent PRs in the shorter term.

Initsogar:
> I really appreciate the effort. Ideally, we should approach changes with the atomic change in mind, and keep large PRs like this only for testing or prototyping. I realize it's hard and time-taking sometimes, but it's worth the effort and our future selves will thank us for that. Anyway, I won't be blocking this PR if you can commit to addressing the feedback in subsequent PRs in the shorter term.

Thank you, @tyxla, I commit to address the feedback in a new PR in short term.

client/components/backup-retention-management/index.tsx
Initsogar: I think I see now what's the issue and how it should be.

After updating the retention period, I'm setting the `retentionSelected` at that moment as the `currentRetentionPlan`, which is not the best approach. The right way should be to update the retentionPlan on the redux state using an action dispatcher in the [onSuccess callback](https://github.com/Automattic/wp-calypso/pull/73035/files#diff-d9ef76a91dcc2d8e7451ccddfbc3fc6b732bba3c09dc25de5f55da431b361c2fR24).

I'll take a double look and handle it in a new PR.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7814445

Thank you @Initsogar for including a screenshot in the description! This is really helpful for our translators.

client/components/backup-retention-management/index.tsx
Initsogar: Just [submitted a PR](https://github.com/Automattic/wp-calypso/pull/73300) for review related to this, it would be great if you can take a look :) https://github.com/Automattic/wp-calypso/pull/73300

client/components/backup-retention-management/constants.ts
Initsogar: Double-checked and no. It was an attempt to declare what are the options that we are going to use. I removed this and instead, added the `RetentionPeriod` from `calypso/state/rewind/retention/types` to validate that we use valid values.

Addressed in #73303

client/components/backup-retention-management/retention-options/retention-options-control.tsx
Initsogar: Agree, it makes sense. We just prepared a refactor for this.

Addressed in #73303

client/components/backup-retention-management/index.tsx
Initsogar: I think it makes more sense to have it as a component state. We just refactored it.

Addressed in #73303

client/components/backup-retention-management/retention-options/retention-option-card.tsx
Initsogar: > Won't the TranslateResult include string already?

It seems [it is actually defined as a React.ReactChild](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/types/index.d.ts#L59) and Typescript doesn't like the assumption that it is a string.

Either way, we revisited it and noticed that it is preferable to pass the space needed in bytes, so we made a little refactor there.

Addressed in #73303

client/components/backup-retention-management/types.ts
Initsogar: Incomplete 🥲. The 7 was actually an example.

Either way, [we refactored it a bit](https://github.com/Automattic/wp-calypso/pull/73303/files#diff-71e240ea9b6bd13fea335936ce41b9d60fc06babcbc766ce1ed40cde9f4905e5), including a `RetentionPeriod` type instead of `number`, so we can enforce to only use the valid values (7, 30, 120, 365 for now).

Addressed in #73303

client/components/backup-retention-management/hooks.tsx
Initsogar: We did some refactors on the way we handle the retention options, and we needed to stop using hooks, as they must only be called at top level.

Either way, at first sight and given that they are simple math operations and the fact that we are not calling too many times those functions, the memoization is not required for `useEstimatedCurrentSiteSize()` and `useEstimateSpaceNeeded()`.

Addressed in #73303

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7814445

Thank you @Initsogar for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7814445

Thank you @Initsogar for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7814445).

PR #73050, opened 2023-02-07
Reading Settings: Fix unexpected dialog displayed on Reading settings page save

### LINKED ISSUE #71071: Reading Settings: Fix unexpectedly displayed "Changes you made may not be saved." dialog
When a setting with a toggle is changed and saved, the "Changes you made may not be saved." browser dialog appears on a page reload (or when trying to navigate away) - even though there's no need for it:

Image

We could reproduce this issue with the following two settings:

* Enable featured image on your new post emails
* Limit feed to excerpt only

ivan-ottinger:
Fixes https://github.com/Automattic/wp-calypso/issues/71071.

There's a bug (https://github.com/Automattic/wp-calypso/issues/71071) that is causing "Changes you made may not be saved." browser dialog to display even when not necessary. Please see the issue description for more details.

### tl;dr;

`nextDirtyFields` array inside `updateDirtyFields()` of `WrappedSettingsForm` component (in `client/my-sites/site-settings/wrap-settings-form.jsx`) keeps the updated field in it - even after it has been saved correctly. As a result, the form "thinks" it hasn't been saved and the dialog appears.

## Longer explanation:

`nextDirtyFields` array is calculated with the help of `persistedFields` that holds already-saved settings:

```js
const persistedFields = getFormSettings( this.props.settings );
```

The issue can be observed when we look at the `return` statement inside `getFormSettings` in `client/my-sites/site-settings/settings-reading/main.tsx`. When one of the settings, e.g. `wpcom_featured_image_in_email` is `false`, it doesn't make it to `persistedFields`.

Consequently, `persistedFields` will be missing that setting completely (i.e. in our example, `wpcom_featured_image_in_email: false` won't be there) - and as a result, `persistedFields` won't match with `currentFields` (that holds current setting form selection, e.g. `wpcom_featured_image_in_email: false`). Hence, the form will think it still needs to be saved.

ℹ️ This PR addresses the issue manifesting on settings of the new Reading settings page only. → I have created a separate task that will address the issue on other settings pages. We can take it on later - as a part of maintenance: https://github.com/Automattic/wp-calypso/issues/73147

ℹ️ Similar issue is affecting `rss_use_excerpt`, but the cause is different there (not fully understood yet). → I have created a separate task for it as well: https://github.com/Automattic/wp-calypso/issues/73149

## Proposed Changes

* Make sure the `getFormSettings()` returns object that includes the settings even if they are `false` or equal to empty strings.

## Testing Instructions

1. Check out and build this PR's branch locally.
2. Navigate to `https://wordpress.com/settings/reading/[site-address]`.
3. Try to reproduce the https://github.com/Automattic/wp-calypso/issues/71071 issue with the following fields. It should not be possible.
- Your homepage displays
- Default posts page
- all toggles of Related Posts setting
- Enable featured image on your new post emails
- For each new post email, include
4. There should be no regressions. All settings should save correctly.

ℹ️ Please note that if you reload the page very quickly after the setting has been saved, it may not yet reflect in the form.

## Pre-merge Checklist


client/my-sites/site-settings/settings-reading/main.tsx
```diff
@@ -59,20 +59,20 @@ const getFormSettings = ( settings: unknown & Fields ) => {
...
-		...( jetpack_relatedposts_show_thumbnails && { jetpack_relatedposts_show_thumbnails } ),
-		...( page_for_posts && { page_for_posts } ),
-		...( page_on_front && { page_on_front } ),
+		jetpack_relatedposts_enabled: !! jetpack_relatedposts_enabled,
+		jetpack_relatedposts_show_context: !! jetpack_relatedposts_show_context,
+		jetpack_relatedposts_show_date: !! jetpack_relatedposts_show_date,
+		jetpack_relatedposts_show_headline: !! jetpack_relatedposts_show_headline,
+		jetpack_relatedposts_show_thumbnails: !! jetpack_relatedposts_show_thumbnails,
+		...( page_for_posts ? { page_for_posts } : { page_for_posts: '' } ),
+		...( page_on_front ? { page_on_front } : { page_on_front: '' } ),
```
mashikag: Spreading seems to be unnecessary here. Based on the new statement we will always end up with `page_on_front` property. 
Previously the spread operator was used because if `page_on_front` was `undefined` (or of falsy value to be more precise) it would not be included in the object returned. See example of what I mean below:
Screen Shot 2023-02-09 at 11 35 13

With the new statement we will include the `page_on_front` prop in the result object even when it is `undefined`:
Screen Shot 2023-02-09 at 11 39 24

Therefore to reduce the complexity of the statement I would suggest dropping the spread operator. It could be re-written as:
```
page_on_front: page_on_front ?? '',
```

The same goes for `page_for_posts` on the line above.

mashikag (changes requested):
Great job on finding the root cause and coming up with the solution! I suggested just a slight change to how two statements are written, it should make the code easier to understand. 🙂
Feel free to argue on this one if you do not agree (this should be always implied). 😅

BTW I am now very curious as to what is the other issue that causes the same symptom for `rss_use_excerpt`. 😁

client/my-sites/site-settings/settings-reading/main.tsx
ivan-ottinger: Thanks for noticing and sharing the details, Maciej! 🙂 Yes, that makes total sense. I will apply the change in a new commit.

ivan-ottinger: Resolved in https://github.com/Automattic/wp-calypso/pull/73050/commits/5c11c1b63ccbba1856381af63850da6b40024506.

mashikag (approved):
Thanks for addressing my suggestion! 🙇 
I tested locally with the PR applied and the issue is addressed as expected. 👌

PR #73052, opened 2023-02-07
E2E: Enable CoBlocks tests against Atomic env

WunderBart:
Related to p4TIVU-apL-p2

## Proposed Changes

Update and enable CoBlocks tests against the Atomic environment.

## How?

For now, let's go with the scenario where run (Atomic) CoBlocks tests within our regular Gutenberg upgrade test suite. The whole suite should run against the no-CoBlocks account (after the cutoff date), and only the CoBlocks tests would run against the account with the plugin enabled (before the cutoff date).

## Testing Instructions

As a part of the Gutenberg group, all CoBlocks tests should run against both Simple and Atomic sites (CI).

To verify locally, run the following and make sure correct accounts are being used (tip: exporting `DEBUG=1` will log out currently used accounts):
   
- `coBlocksAtomicSiteEdgeUser` (new, used only for CoBlocks tests, switched on via the atomic flag)
   
   ```
   TEST_ON_ATOMIC=true yarn test specs/blocks/blocks__coblocks__blocks.ts
   ```

- `gutenbergAtomicSiteUser` (as it was, used for tests other than CoBlocks)

   ```
   TEST_ON_ATOMIC=true yarn test specs/blocks/blocks__core.ts
   ```

- `coBlocksSimpleSiteEdgeUser` (as it was)

   ```
   COBLOCKS_EDGE=true yarn test specs/blocks/blocks__coblocks__blocks.ts
   ```

- `gutenbergSimpleSiteEdgeUser` (as it was)

   ```
   GUTENBERG_EDGE=true yarn test specs/blocks/blocks__coblocks__blocks.ts
   ```

- `simpleSitePersonalPlanUser` (as it was)

   ```
   yarn test specs/blocks/blocks__coblocks__blocks.ts
   ```

### DIFF
--- a/packages/calypso-e2e/src/lib/blocks/pricing-table-block.ts
+++ b/packages/calypso-e2e/src/lib/blocks/pricing-table-block.ts
@@ -1,10 +1,5 @@
 import { Frame, Page, ElementHandle } from 'playwright';
 
-// Workaround to get a type from predefined array.
-// See https://stackoverflow.com/a/59857409.
-const gutterValuesArray = [ 'None', 'Small', 'Medium', 'Large', 'Huge' ] as const;
-export type gutterValues = typeof gutterValuesArray[ number ];
-
 const selectors = {
 	block: '.wp-block-coblocks-pricing-table',
 	pricing: '.wp-block-coblocks-pricing-table-item__amount',
@@ -18,7 +13,7 @@ export class PricingTableBlock {
 	// Static properties.
 	static blockName = 'Pricing Table';
 	static blockEditorSelector = '[aria-label="Block: Pricing Table"]';
-	static gutterValues = gutterValuesArray;
+
 	block: ElementHandle;
 
 	/**
@@ -47,21 +42,21 @@ export class PricingTableBlock {
 	/**
 	 * Given an appropriate value, select the gutter value on the settings sidebar for Pricing Table block.
 	 *
-	 * @param {string} value Value to set the gutter to.
+	 * @param {string} name Name of the gutter preset.
+	 * @param {number} value Value of the gutter for the "Custom" preset (Atomic).
 	 * @returns {Promise<void>} No return value.
 	 */
-	async setGutter( value: gutterValues ): Promise< void > {
+	async setGutter( name: string, value?: number ): Promise< void > {
 		const frame = ( await this.block.ownerFrame() ) as Frame;
+		const buttonSelector = `${ selectors.gutterControl } button[aria-label="${ name }"]`;
 
-		const selector = `${ selectors.gutterControl } button[aria-label="${ value }"]`;
-		await frame.click( selector );
-		const elementHandle = await frame.waitForSelector( selector );
-		// waitForFunction will do its own validation, so no need to return the value
-		// to caller for a check.
-		await frame.waitForFunction(
-			( element: SVGElement | HTMLElement ) => element.ariaPressed === 'true',
-			elementHandle
-		);
+		await frame.locator( buttonSelector ).click();
+		await frame.locator( `${ buttonSelector }[aria-pressed="true"]` ).waitFor();
+
+		if ( name === 'Custom' && value !== undefined ) {
+			const valueInput = frame.locator( 'input[type="number"]:below(button[aria-label="Custom"])' );
+			await valueInput.fill( String( value ) );
+		}
 	}
 
 	/**
--- a/test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
+++ b/test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
@@ -15,9 +15,12 @@ import { skipItIf } from '../../jest-helpers';
 
 declare const browser: Browser;
 
-const features = envToFeatureKey( envVariables );
 const isAtomic = envVariables.TEST_ON_ATOMIC;
 const isSimple = ! envVariables.TEST_ON_ATOMIC;
+const features = envToFeatureKey( {
+	...envVariables,
+	COBLOCKS_EDGE: isAtomic || envVariables.COBLOCKS_EDGE,
+} );
 
 describe( DataHelper.createSuiteTitle( 'CoBlocks: Extensions: Gutter Control' ), () => {
 	const accountName = getTestAccountByFeature( features );
--- a/test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
+++ b/test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
@@ -11,73 +11,90 @@ import {
 	envToFeatureKey,
 } from '@automattic/calypso-e2e';
 import { Page, Browser } from 'playwright';
-import { skipDescribeIf } from '../../jest-helpers';
+import { skipItIf } from '../../jest-helpers';
 
 declare const browser: Browser;
 
 const features = envToFeatureKey( envVariables );
+const isAtomic = envVariables.TEST_ON_ATOMIC;
+const isSimple = ! envVariables.TEST_ON_ATOMIC;
 
-skipDescribeIf( features.siteType === 'atomic' )(
-	DataHelper.createSuiteTitle( 'CoBlocks: Extensions: Gutter Control' ),
-	() => {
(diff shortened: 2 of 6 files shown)


WunderBart:
@fullofcaffeine, I think we need a dedicated CoBlocks job here, unfortunately. This is because there's a separate environment that is used for those tests (Atomic site with enabled CoBlocks), which doesn't make sense to use as a default one for testing Gutenberg upgrades. What do you think? cc @worldomonation

packages/calypso-e2e/src/lib/blocks/pricing-table-block.ts
```diff
@@ -47,21 +42,21 @@ export class PricingTableBlock {
 	/**
 	 * Given an appropriate value, select the gutter value on the settings sidebar for Pricing Table block.
 	 *
-	 * @param {string} value Value to set the gutter to.
+	 * @param {string} name Name of the gutter preset.
+	 * @param {number} value Value of the gutter for the "Custom" preset (Atomic).
 	 * @returns {Promise<void>} No return value.
 	 */
-	async setGutter( value: gutterValues ): Promise< void > {
+	async setGutter( name: string, value?: number ): Promise< void > {
```
dpasque, line 49: **Suggestion (non-blocking):** Any specific reason for getting rid of the stronger typing here? I personally tend to favor having strings with specific options as a union type of specific strings. That's also just my personal preference though!

test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
```diff
@@ -11,73 +11,90 @@ import {
...
 
-		it( 'Verify the class for "Huge" gutter is present', async () => {
-			await page.waitForSelector( '.wp-block-coblocks-pricing-table .has-huge-gutter' );
-		} );
-	}
-);
+	skipItIf( isSimple )( 'Verify the proper value for "Custom" gutter is set', async () => {
+		await page
+			.locator( '.wp-block-coblocks-pricing-table [style="--coblocks-custom-gutter:2.7em"]' )
+			.waitFor();
```
dpasque, line 102: **Suggestion (blocking):** When I was running this locally, I was getting a strict mode violation here. It looks like there's a debug element that conflicts with this:

```
    locator.waitFor: Error: strict mode violation: locator('.wp-block-coblocks-pricing-table [style="--coblocks-custom-gutter:2.7em"]') resolved to 2 elements:
        1) <div class="wp-block-coblocks-pricing-table__inner ha…>…</div> aka locator('#post-101').getByText('4.99 9.99')
        2) <div class="wp-block-coblocks-pricing-table__inner ha…>…</div> aka locator('#debug-bar-wp-query').getByText('4.99 9.99')
```

Maybe we need to just add a `first()` here?

dpasque (approved):
Left a couple comments -- one of the scripts was failing locally, so that should be fixed before merge!

But otherwise, the code LGTM!

Also, let us know what you end up preferring in terms of controlling the atomic environment for these tests, and how we can help with that!

I think you probably have two main ways of doing that...
1. Just make all coblocks tests on atomic use the coblocks user instead of the default gutenberg one.
2. You could split these tests into their own `jest-group` and we could tweak which TeamCity builds run which tests

test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
WunderBart: This happens only when you run tests AutoProxxied. For some reason, there's this `#debug-bar-wp` being added to DOM in the mobile viewport, which is breaking the strict rule of the locator because it contains cloned target block element. I think that AutoProxxy should be disabled when running tests locally because it can change the environment to the point where some tests break, like this one. Do you think it would be worth adding this to the general E2E readme?

packages/calypso-e2e/src/lib/blocks/pricing-table-block.ts
WunderBart: Good question. I removed it because it started getting overengineered, TBH. The `PricingTable` utility will not be used anywhere except for the few smoke tests of the CoBlocks plugin (context: p4TIVU-apL-p2, p9oQ9f-1yg-p2). The important part is verifying that those gutter settings work well on our side because we customize them to the point where they're different on Simple vs. Atomic sites. That bit is covered [within the Gutter Control extension suite](https://github.com/Automattic/wp-calypso/blob/733264178943a777815eca6d9fd64a9a02044966/test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts#L54-L76), so the additional layer of conditional Simple vs. Atomic gutter typing felt overkill to me as it doesn't really bring any value at this point. Does that make sense?

WunderBart:
Thanks for the review, @dpasque! 🙌 
> (…) let us know what you end up preferring in terms of controlling the atomic environment for these tests, and how we can help with that!
>
> I think you probably have two main ways of doing that...
> 
> 1. Just make all coblocks tests on atomic use the coblocks user instead of the default gutenberg one.
> 2. You could split these tests into their own `jest-group` and we could tweak which TeamCity builds run which tests

I think those tests need to be addressed individually because of the difference in environments of CoBlocks vs. no-CoBlocks (Atomic) users. CoBlocks add a lot of deeply integrated functionality to Gutenberg, which can cause problems unrelated to the users that don't have the plugin enabled (accounts created after the cutoff date). That means we can't really use the CoBlocks account instead of the no-CoBlocks one to run the Gutenberg upgrade tests against, so option `1` is off the table. Regarding option `2`, the question is whether we want to run the full Gutenberg suite against an Atomic site with CoBlocks enabled or just a selected part. If we decide to run the full suite, it would be best to have a separate job (something like the `COBLOCKS_EDGE` for Simple sites). Otherwise, I don't think we need another group because we can run our CoBlocks tests within the regular Gutenberg upgrade suite - we'll just need to use the dedicated account, just for them.

For now, I'd go with the scenario where run (Atomic) CoBlocks tests within our regular Gutenberg upgrade test suite. The whole suite would run against the no-CoBlocks account, and only the CoBlocks tests would run against the account with the plugin enabled. How does that sound?

WunderBart:
@dpasque, I've added more testing instructions in the PR description for a better explanation of my suggestion [here](https://github.com/Automattic/wp-calypso/pull/73052#issuecomment-1422500844). Let me know if it makes sense!

packages/calypso-e2e/src/lib/blocks/pricing-table-block.ts
dpasque: For sure, that's fine with me! 😄

test/e2e/specs/blocks/blocks__coblocks__extensions__gutter-control.ts
dpasque: Ahhh, _that's_ what that is! Sorry for the confusion! 

> Do you think it would be worth adding this to the general E2E readme?

Yeah, that's not a bad idea!

dpasque:
> For now, I'd go with the scenario where run (Atomic) CoBlocks tests within our regular Gutenberg upgrade test suite. The whole suite would run against the no-CoBlocks account, and only the CoBlocks tests would run against the account with the plugin enabled. How does that sound?

I think this is a great approach! This is actually what I was proposing with option `1`, but just worded really poorly 🙃

dpasque:
@WunderBart Btw - new changes look good too! 👍

fullofcaffeine:
> For now, I'd go with the scenario where run (Atomic) CoBlocks tests within our regular Gutenberg upgrade test suite. The whole suite would run against the no-CoBlocks account, and only the CoBlocks tests would run against the account with the plugin enabled. How does that sound?

I think this sounds good! The criteria table looks good, though it will also select these sites for the stable GB AT build, which is confusing at first glance, but I guess makes sense, given CoBlocks in AT is "schrodinger"'s CoBlocks, it might be stable or edge, depending on how you see it :D

WunderBart:
> (…) I guess makes sense, given CoBlocks in AT is "schrodinger"'s CoBlocks, it might be stable or edge, depending on how you see it :D

Ha! I guess we can say that Atomic CoBlocks are always on the `edge` since we're testing against the Gutenberg plugin, right? 😅 (p9oQ9f-1yg-p2#comment-2299)

PR #73070, opened 2023-02-08
reader: Restore focus when using back button from full post view to feed

roo2:
Builds on https://github.com/Automattic/wp-calypso/pull/72940 to finish off https://github.com/Automattic/wp-calypso/issues/61318

> When you open a single post from the reader list view, then close that post again, the currently tabbed element is reset to the start of the page, but the current post stays as it was prior to it being opened.

This PR fixes this behavior by checking for a selected post and moving the focus to that post.

The timeout is necessary here because the DOM is not available to `focus()` similar to [this question on SO](https://stackoverflow.com/questions/35522220/react-ref-with-focus-doesnt-work-without-settimeout-my-example). I played with different values and found `100ms` works but `10ms` does not.

https://user-images.githubusercontent.com/22446385/217420161-8960ba59-3e4c-42b7-aa48-d396d81b02a9.mov

### DIFF
--- a/client/reader/stream/index.jsx
+++ b/client/reader/stream/index.jsx
@@ -190,6 +190,12 @@ class ReaderStream extends Component {
 			window.history.scrollRestoration = 'manual';
 		}
 
+		if ( this.props.selectedPostKey ) {
+			setTimeout( () => {
+				this.focusSelectedPost( this.props.selectedPostKey );
+			}, 100 );
+		}
+
 		document.addEventListener( 'keydown', this.handleKeydown, true );
 	}
 


tyxla (approved):
Nice, works great @roo2, thanks 🚀

client/reader/stream/index.jsx
```diff
@@ -190,6 +190,12 @@ class ReaderStream extends Component {
 			window.history.scrollRestoration = 'manual';
 		}
 
+		if ( this.props.selectedPostKey ) {
+			setTimeout( () => {
+				this.focusSelectedPost( this.props.selectedPostKey );
+			}, 100 );
+		}
```
tyxla, line 197: Actually one thing that just occurred to me: does it make sense to reuse the existing `componentDidUpdate()` behavior:

https://github.com/Automattic/wp-calypso/blob/323f51b38fad7e34634b2a2eca7d9eeff8d1baea/client/reader/stream/index.jsx#L126-L129

and ideally, have a single approach for focusing the post on load and on update?

roo2: I don't understand. I don't think we can reuse it

* We can't use the condition `keysAreEqual( selectedPostKey, this.props.selectedPostKey )` in `componentDidMount` because `{ selectedPostKey }` is an argument to `componentDidUpdate` and unavailable in `componentDidMount`.
* The scroll restoration seems to be handled differently already and depends on `if ( 'scrollRestoration' in window.history )`
* and we need the `setTimeout` here because DOM isn't fully ready yet, whereas it is not neccessary in the `componentDidUpdate` function

tyxla: Makes sense, thanks for confirming! 🚀

Issue #72548, opened 2023-01-25
Site Assembler - Pattern fast preview component optimizations


miksansegundo:
## Description
The rendering is pretty snappy, but we wonder if this could be snappier.

## Are there any possible optimizations?

We could try some of the following:

 - use a virtual list (virtual scroll) to keep a low number of DOM elements
 - use infinite scroll to avoid rendering all the patterns in a category at the same time 
 - render the first patterns of each category in the background
 - render all patterns in the background
 - move some JS tasks to a web worker
 - load the pattern CSS in the iframe first and then the HTML
 - load the patterns in a category in small batches (3 at a time)
 - add an animation to show patterns with some delay, so they have time to load their CSS and images
 - cache the server requests 

##  How it works?

The fast preview component fetches the patterns HTML already rendered from the server and applies their CSS in an iframe in the browser.

https://user-images.githubusercontent.com/1881481/216924374-5dfec33d-a235-4a84-b10d-1d1adaa9887a.mov

https://user-images.githubusercontent.com/1881481/214501461-26cbc35e-43ab-4bbd-88d6-1c94a2fe83a1.mov

## Considerations

We should be aware of client-side performance issues caused by the following:

- blocking the UI because of too many synchronous JS tasks - tasks should be queued 
- lagging the UI because of having a large amount of DOM elements in the document  
- unnecessary renders
- memory leaks

## Questions

> Can we have these loading in the background somewhere continually while someone is browsing so that rendering just gets snappier and snappier in a session?

The loading time is now not caused by the server requests but by how the patterns are rendered in the browser. The pattern HTML and CSS are downloaded in the background, but they are rendered together at the same time that a pattern is displayed. 

> Can we cache all these so they render instantly?

We had a client-side cache for the server requests, it was removed because it caused an annoying issue only for developers. 

We could try to add a new cache system. Maybe using a browser service worker or a server-side cache using the site as a cache key. This won't make the patterns render instantly but will reduce the download time of the HTML & CSS. 

## Related

https://github.com/Automattic/wp-calypso/pull/71311

arthur791004:
> use a virtual list (virtual scroll) to keep a low number of DOM elements
> use infinite scroll to avoid rendering all the patterns in a category at the same time

It would be great but I think we can hold it on. I assume the performance would be better after adding categories since there are not so many patterns in each category. But if we still feel a bit slow at that time, we can start to work on this one.

> render the first patterns of each category in the background

[...]

PR #73077, opened 2023-02-08
Site Assembler: Render patterns on demand

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/72548

## Proposed Changes

* Render the specified patterns when the pattern selector is open
* Render the list of patterns by `useAsyncList` to render the patterns one by one. The `useAsyncList` hook has an option called `step` to control how many items you want to render every time. The default value is 1 and I think it's enough for now.
* Preload the styles assets as much as possible so that the styles will be loaded when the user enters the site assembler for the first time

https://user-images.githubusercontent.com/13596067/217470562-9f1ce6ca-c67a-48fe-973d-914d1798abf8.mov

## Testing Instructions

* On the Pattern Assembler screen, select any patterns to see whether it's faster than before

## Pre-merge Checklist


autumnfjeld:
@arthur791004 I did the "Lego fast test", meaning as soon as I landed on the Assembler screen I clicked "Add sections" (which I don't think a user would do, a user would be much slower and probably choose Headers first and so they would not experience the same delay as in the "Lego fast test").  There is still a few second delay as patterns load, but it is better with `useAsyncList` than what is in production.

For these tests I "emptied cash"
image

In production:

https://user-images.githubusercontent.com/5665959/217732077-9c3357e9-6dd7-4e30-93a2-92ff788bc368.mov



Your changes:
I did it twice because there was an odd delay in show the Assembler screen in one test (maybe slowness in my internet?)

https://user-images.githubusercontent.com/5665959/217732660-591ae776-0a2a-4ad6-b1bd-15864ff452ca.mp4


https://user-images.githubusercontent.com/5665959/217731685-cdf6c4b4-9ffe-4e93-9512-dc8b6eb5d703.mp4

arthur791004:
> There is still a few second delay as patterns load, but it is better with useAsyncList than what is in production.

Yep, there are still too many patterns there and it will make the network requests queued for a short time. I think it will be better after we show them in categories

miksansegundo:
I still haven't had much time to review this PR and help improve it. I'd love to work on it after we launch the Write flow.

I think we can still improve how the patterns load and the animation. Ideally, for a pretty load, patterns should be hidden until the load is done, so they look good.

**Can we hide the patterns until the styles are applied?**
I think the pattern HTML without styles introduces a glitch that is not desired 

Screenshot 2566-02-13 at 18 18 53

**Can we remove the gray background?** 
I think the background color change during the load introduces a glitch that is not desired 

Screenshot 2566-02-13 at 18 18 25

This video is in slow motion to help visualize the glitches:

https://user-images.githubusercontent.com/1881481/218445726-67f90065-4c43-4d22-b621-c8b38acde3a3.mov

arthur791004:
> Can we hide the patterns until the styles are applied?
> I think the pattern HTML without styles introduces a glitch that is not desired

I think this is not a blocker for this one. We can continue to improve the experience in the follow-up PR as it's the existing behavior in production. Does it make sense?

> Can we remove the gray background?
> I think the background color change during the load introduces a glitch that is not desired

Sorry, I'm not quite sure what you're meaning. If you mean the first pattern in your screenshot, I think it's due to the animation of the opacity. The loading placeholder and the rendered pattern won't be overlapped as either one of them would be rendered.

miksansegundo:
> I think this is not a blocker for this one. We can continue to improve the experience in the follow-up PR as it's the existing behavior in production. Does it make sense?

Comparing the changes in this PR with production, these changes accelerate the pattern load the first time, but then in the following loads render all patterns again, which I find a worse experience than in production.

I find that a blocker for this PR because when going back and forth, you have to wait for the patterns to load every time, not just the first time as in production.

miksansegundo:
> Sorry, I'm not quite sure what you're meaning. 

I mean the onboarding placeholder, it has a gray background.

miksansegundo:
Have a look at these [changes](https://github.com/Automattic/wp-calypso/pull/73323). The initial loading is cleaner IMO. Can you take the idea from my changes to improve this PR?

We still should avoid rendering the patterns on demand after the first load.

miksansegundo:
Here is a video of the pattern load improved by removing the placeholder and waiting for the style links to be loaded.

https://user-images.githubusercontent.com/1881481/218730147-493985a9-69d8-4a79-8d48-66fbe8a3cc3a.mov

miksansegundo:
Here is a video explaining why I think the UX in this PR is not better than in production.

https://user-images.githubusercontent.com/1881481/218732223-968090f9-2f80-4cf8-9413-438066e70db2.mov

arthur791004:
I made a change to avoid showing any unstyled content and remove the placeholder. For the placeholder, IMO we still need it since the pattern shows the white content for a few seconds might make users feel weird or something is wrong.

> Comparing the changes in this PR with production, these changes accelerate the pattern load the first time, but then in the > following loads render all patterns again, which I find a worse experience than in production.
>
> I find that a blocker for this PR because when going back and forth, you have to wait for the patterns to load every time, not just the first time as in production.
>
> We still should avoid rendering the patterns on demand after the first load.

We should not keep all of the patterns after the first load when we have more and more patterns. If we did, whenever something changed (e.g. upcoming colors & fonts), calypso would re-render all of them at the same time, and lead the browser stuck for a while. Additionally, it will take up too much memory on your computer (now it took 2.5 GB after loading all patterns on production), and I don't think it's good for our customers.

Instead, we have to keep the rendering process as fast as possible and that's also what the site editor tries to do. Now, the issue is there are too many patterns there and it leads to loading styles being a bit slow. Some of the requests might be queued for a while because there are too many requests at the same time.

So, I think our next steps are
1. Show patterns by categories as we planed
2. Remove unnecessary external styles (like what we did for the block-preview endpoint before)

What do you think?

miksansegundo:
> For the placeholder, IMO we still need it since the pattern shows the white content for a few seconds might make users feel weird or something is wrong.

Agree. My concern is about the glitches that the placeholder cause and the gray background opacity animation. 

> We should not keep all of the patterns after the first load. If we did, whenever something changed (e.g. upcoming colors & fonts), calypso would re-render all of them at the same time, and lead the browser stuck for a while.
 
Agree. We can keep a controlled pool (array) of DOM nodes (previews) displayed. Let's see how many. 

Have you had the chance to read the optimizations I suggested in [Site Assembler - Pattern fast preview component optimizations#72548](https://github.com/Automattic/wp-calypso/issues/72548)? 

> we have to keep the rendering process as fast as possible

Agree. 

I think preloading some pattern previews would be a great optimization too. We can preload the first 5 pattern previews for each category or the most popular categories, every time there are changes in colors and keep them in the pool.

arthur791004:
> My concern is about the glitches that the placeholder cause and the gray background opacity animation.

Agree. So, I remove it temporarily and we can add it back when we have a better implementation

> Have you had the chance to read the optimizations I suggested in https://github.com/Automattic/wp-calypso/issues/72548?

Sure! I replied [there](https://github.com/Automattic/wp-calypso/issues/72548#issuecomment-1432591979)

> I think preloading some pattern previews would be a great optimization too. We can preload the first 5 pattern previews for each category or the most popular categories, every time there are changes in colors and keep them in the pool.

I also described the issue of the preload [here](https://github.com/Automattic/wp-calypso/issues/72548#issuecomment-1432591979). It's great but we still need to take caution. Maybe we can preload the first 5 patterns for the category that still has lots of patterns, e.g. Call to Action.

miksansegundo (approved):
Great job 🙌  
I just tested the latest changes, and the pattern loading looks much better now, with fewer glitches and much faster. 

I think this PR is ready. I'd ask our designers for feedback on this change before we ship it. Would you like to do that?

Later, we can improve the loading animation with a fade-in and test others, maybe a partial slide-in from the top.

arthur791004:
> I think this PR is ready. I'd ask our designers for feedback on this change before we ship it. Would you like to do that?

Sure! We can ask people on our channel 🙂

BTW, I added the `useInView` hook to load the patterns when they're in the viewport. So, we won't load all section patterns simultaneously!

PR #73088, opened 2023-02-08
Jetpack Backup: Add the logic to update backup retention settings

rustamshaikh:
Related to #
## Proposed Changes

* TBD

## Testing Instructions

* TBD

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #73108, opened 2023-02-08
Jetpack Backup: Add the logic to update backup retention settings #73088

rustamshaikh:
## Proposed Changes

* This implements the logic needed to call WPCOM API to update the backup retention period from a Calypso component.
* Also includes a selector `getBackupRetentionUpdateRequestStatus` to query the update request's status.

## Testing Instructions

* Unit tests:  `yarn run test-client client/state/rewind`
* Using Redux Dev Tool:
  1) Test `retention.updateRequestStatus` state, which is updated while calling the update API.  
  
      Initially, the `state.rewind[BLOG_ID].updateBackupRetentionRequestStatus` should be in the `unsubmitted` state.

      Now, for the following action when dispatched:
      ```javascript
      {
        type: 'JETPACK_BACKUP_RETENTION_UPDATE',
        siteId: '<blog-id>',
        rententionDays: 7,
      }
      ```
      the `state.rewind[BLOG_ID].updateBackupRetentionRequestStatus` should be in the `pending` state.
      
      For the following action:
      ```javascript
      {
        type: 'JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS',
        siteId: '<blog-id>',
      }
      ```
      the `state.rewind[BLOG_ID].updateBackupRetentionRequestStatus` should be in the `success` state.
    
      For the following action:
      ```javascript
      {
        type: 'JETPACK_BACKUP_RETENTION_UPDATE_ERROR',
        siteId: '<blog-id>',
      }
      ```
      the `state.rewind[BLOG_ID].updateBackupRetentionRequestStatus` should be in the `failed` state.

## Pre-merge Checklist


client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
+import {
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import retentionReducer from '../reducer';
+
+describe( 'rewind.retention reducers', () => {
+	describe( 'lastBackupSize', () => {
```
Initsogar:
```suggestion
	describe( 'updateBackupRetentionRequestStatus', () => {
```

Shouldn't we use `updateBackupRetentionRequestStatus` here?

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,26 @@
...
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { combineReducers } from 'calypso/state/utils';
+import { AppState } from 'calypso/types';
+
+const updateBackupRetentionRequestStatus = ( state: AppState = null, { type }: AnyAction ) => {
+	switch ( type ) {
+		case JETPACK_BACKUP_RETENTION_UPDATE:
+			return 'pending';
```
Initsogar: Maybe it would be better to add the different request statuses as constants and use it anywhere, so we don't need to hard-code the strings anytime we need to validate the status. It can be something like this:

```
const BACKUP_RETENTION_UPDATE_PENDING = 'pending';
const BACKUP_RETENTION_UPDATE_SUCCESS = 'success';
const BACKUP_RETENTION_UPDATE_FAILED = 'failed';
```

What do you think?

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
...
+import 'calypso/state/data-layer/wpcom/sites/rewind/retention';
+
+type RequestActionType = Action< typeof JETPACK_BACKUP_RETENTION_UPDATE > & {
+	siteId: number | null;
+	rententionDays: RetentionPeriod;
+};
+
+export const updateBackupRetention = (
+	siteId: number | null,
+	rententionDays: RetentionPeriod
```
Initsogar:
```suggestion
	retentionDays: RetentionPeriod
```

A small typo 😃

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
...
+): RequestActionType => ( {
+	type: JETPACK_BACKUP_RETENTION_UPDATE,
+	siteId,
+	rententionDays,
```
Initsogar:
```suggestion
	retentionDays,
```

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
+import { Action } from 'redux';
+import { JETPACK_BACKUP_RETENTION_UPDATE } from 'calypso/state/action-types';
+import type { RetentionPeriod } from './types';
+import 'calypso/state/data-layer/wpcom/sites/rewind/retention';
+
+type RequestActionType = Action< typeof JETPACK_BACKUP_RETENTION_UPDATE > & {
+	siteId: number | null;
+	rententionDays: RetentionPeriod;
```
Initsogar:
```suggestion
	retentionDays: RetentionPeriod;
```

client/state/data-layer/wpcom/sites/rewind/retention/index.ts
```diff
@@ -0,0 +1,47 @@
...
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { registerHandlers } from 'calypso/state/data-layer/handler-registry';
+import { http } from 'calypso/state/data-layer/wpcom-http/actions';
+import { dispatchRequest } from 'calypso/state/data-layer/wpcom-http/utils';
+import { RetentionPeriod } from 'calypso/state/rewind/retention/types';
+
+type RequestActionType = Action< typeof JETPACK_BACKUP_RETENTION_UPDATE > & {
+	siteId: number | null;
+	rententionDays: RetentionPeriod;
```
Initsogar:
```suggestion
	retentionDays: RetentionPeriod;
```

client/state/data-layer/wpcom/sites/rewind/retention/index.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const updateRetention = ( action: RequestActionType ) =>
+	http(
+		{
+			apiNamespace: 'wpcom/v2',
+			method: 'POST',
+			path: `/sites/${ action.siteId }/backup/retention/update`,
+			body: { retention_days: action.rententionDays },
```
Initsogar:
```suggestion
			body: { retention_days: action.retentionDays },
```

Initsogar (commented):
Overall, it looks great and tests are passing as expected.
Just added some suggestion and a typo detected.

Also, not a stopper, but what do you think about using `client/state/backup` instead of `client/state/rewind`, given that we would like to get rid of `rewind` in our codebases.

@tyxla if you have some time, it would be great if you can also take a look 😊

client/state/rewind/retention/reducer.ts
rustamshaikh: Good suggestion. Maybe we can declare such constants without being specific to backup retention API, given that those are generic states an API call can have and can be reused with other API calls in the future. What do you think about moving these to the root of the rewind/backup folder? like this:

```
const BACKUP_API_REQUEST_PENDING = 'pending';
const BACKUP_API_REQUEST_SUCCESS = 'success';
const BACKUP_API_REQUEST_FAILED = 'failed';
```

client/state/rewind/retention/test/reducer.js
rustamshaikh: ✅

client/state/rewind/retention/actions.ts
rustamshaikh: ✅

client/state/data-layer/wpcom/sites/rewind/retention/index.ts
```diff
@@ -0,0 +1,47 @@
+import { Action } from 'redux';
```
tyxla: Did you mean to do `import type` instead of `import`?

client/state/data-layer/wpcom/sites/rewind/retention/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { registerHandlers } from 'calypso/state/data-layer/handler-registry';
+import { http } from 'calypso/state/data-layer/wpcom-http/actions';
+import { dispatchRequest } from 'calypso/state/data-layer/wpcom-http/utils';
+import { RetentionPeriod } from 'calypso/state/rewind/retention/types';
```
tyxla: In general, when importing types, there are certain benefits we get when using `import type` instead of `import`. 

See this for more info: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
+import { Action } from 'redux';
```
tyxla: Same as above wrt `import type`.

client/state/rewind/reducer.js
```diff
@@ -14,6 +15,7 @@ const rewind = combineReducers( {
 	size,
 	state,
 	storage,
+	retention,
```
tyxla: Nit: so far, the sub reducers were alphabetically sorted. We might want to preserve that order when we introduce a new sub reducer.

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
...
+import { JETPACK_BACKUP_RETENTION_UPDATE } from 'calypso/state/action-types';
+import type { RetentionPeriod } from './types';
+import 'calypso/state/data-layer/wpcom/sites/rewind/retention';
+
+type RequestActionType = Action< typeof JETPACK_BACKUP_RETENTION_UPDATE > & {
+	siteId: number | null;
+	retentionDays: RetentionPeriod;
+};
```
tyxla: This type seems repeated between here and the data layer. Should we declare at a single location and reuse?

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const updateBackupRetention = (
+	siteId: number | null,
```
tyxla: See above for using `null` for a `siteId`.

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
+import { Action } from 'redux';
+import { JETPACK_BACKUP_RETENTION_UPDATE } from 'calypso/state/action-types';
+import type { RetentionPeriod } from './types';
+import 'calypso/state/data-layer/wpcom/sites/rewind/retention';
+
+type RequestActionType = Action< typeof JETPACK_BACKUP_RETENTION_UPDATE > & {
+	siteId: number | null;
```
tyxla: I don't think a `null` `siteId` makes sense here, because that will trigger an HTTP request to `/sites/null/backup/retention/update`, and I don't think we want that.

If TS complains about type incompatibility, this isn't a good way to fix it; rather, we should make sure that this is only used with a non-null `siteId` by additional checks and guards.

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,26 @@
+import { AnyAction } from 'redux';
```
tyxla: Same as above wrt `import type`.

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,26 @@
...
+import {
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { combineReducers } from 'calypso/state/utils';
+import { AppState } from 'calypso/types';
```
tyxla:
```suggestion
import type { AppState } from 'calypso/types';
```

client/state/rewind/retention/reducer.ts
tyxla: Agreed with @Initsogar. We usually declare those as `constants.ts` in the state subtree directory and reuse them where needed.

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+		case JETPACK_BACKUP_RETENTION_UPDATE_ERROR:
+			return 'failed';
+	}
+	return state;
+};
+
+export default combineReducers( {
+	updateBackupRetentionRequestStatus,
+} );
```
tyxla: Do we really need another level here? If we're going to have only the `updateBackupRetentionRequestStatus` sub-reducer, we might as well remove that level and keep the sub-reducer narrow.

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,26 @@
+import { AnyAction } from 'redux';
+import {
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { combineReducers } from 'calypso/state/utils';
+import { AppState } from 'calypso/types';
+
+const updateBackupRetentionRequestStatus = ( state: AppState = null, { type }: AnyAction ) => {
```
tyxla: Since `getBackupRetentionUpdateRequestStatus()` returns `unsubmitted` when there's no data about this site ID, should this reducer default to `unsubmitted` too?

client/state/rewind/selectors/get-backup-retention-update-status.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Returns the update backup retention request's status.
+ *
+ * @param state The application state.
+ * @param siteId The site ID for which to retrieve days of backups saved.
+ * @returns The the status of the request.
+ */
+const getBackupRetentionUpdateRequestStatus = ( state: AppState, siteId: number ): string =>
+	state.rewind[ siteId ]?.retention?.updateBackupRetentionRequestStatus ?? 'unsubmitted';
```
tyxla: Just like the rest of the states of this reducer are suggested to be constants, `unsubmitted` should likely be one of those constants.

client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
+import {
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import retentionReducer from '../reducer';
+
+describe( 'rewind.retention reducers', () => {
```
tyxla: I believe this `describe()` level could be omitted. The fact that these tests live in `client/state/rewind/retention/test/reducer.js` is enough to explain that fact.

client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
...
+		it.each( [
+			{
+				state: undefined,
+				action: {},
+				expected: null,
+			},
+			{
+				state: undefined,
+				action: { type: JETPACK_BACKUP_RETENTION_UPDATE },
+				expected: 'pending',
```
tyxla: We should make sure to use the new constants in tests, too.

client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
...
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import retentionReducer from '../reducer';
+
+describe( 'rewind.retention reducers', () => {
+	describe( 'updateBackupRetentionRequestStatus', () => {
+		it.each( [
+			{
+				state: undefined,
+				action: {},
```
tyxla: Is that a realistic test? Will we ever have an action without a type? I'd suggest adding a random type here to make it a bit more realistic.

client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
...
+				action: { type: JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS },
+				expected: 'success',
+			},
+			{
+				state: { updateBackupRetentionRequestStatus: 'pending' },
+				action: { type: JETPACK_BACKUP_RETENTION_UPDATE_ERROR },
+				expected: 'failed',
+			},
+		] )(
+			'should return a value different than null only when the state has a value or the action is (JETPACK_BACKUP_RETENTION_UPDATE, JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS, REWIND_BACKUP_RETENTION_ERROR)',
```
tyxla: I think this test name isn't very comprehensible. We could use the [positional arguments](https://jestjs.io/docs/api#1-testeachtablename-fn-timeout) of `it.each()`.

Furthermore, I'd suggest creating separate tests, since the `it.each()` syntax doesn't make the tests more readable, and "less is more" in tests.

client/state/rewind/retention/test/reducer.js
```diff
@@ -0,0 +1,40 @@
...
+	describe( 'updateBackupRetentionRequestStatus', () => {
+		it.each( [
+			{
+				state: undefined,
+				action: {},
+				expected: null,
+			},
+			{
+				state: undefined,
+				action: { type: JETPACK_BACKUP_RETENTION_UPDATE },
```
tyxla: I believe we should use the original action creators from `actions.ts`.

client/state/rewind/retention/types.ts
```diff
@@ -0,0 +1 @@
+export type RetentionPeriod = 7 | 30 | 120 | 365;
```
tyxla, line 4: This is a good place for the rest of the types we introduce inline in this state subtree, especially when we repeat them in more than 1 place.

client/state/rewind/retention/actions.ts
```diff
@@ -0,0 +1,18 @@
+import { Action } from 'redux';
+import { JETPACK_BACKUP_RETENTION_UPDATE } from 'calypso/state/action-types';
+import type { RetentionPeriod } from './types';
+import 'calypso/state/data-layer/wpcom/sites/rewind/retention';
+
```
tyxla: This action creators file is missing:

```
import 'calypso/state/rewind/init';
```

which is necessary to ensure that when running an action creator, the corresponding dynamic reducer has already been loaded.

client/state/rewind/selectors/get-backup-retention-update-status.ts
```diff
@@ -0,0 +1,13 @@
+import type { AppState } from 'calypso/types';
+
```
tyxla, line 4: This selector is missing:

```
import 'calypso/state/rewind/init';
```

which is necessary to ensure that the rewind state tree will be loaded in the modular reducer.

That applies to any selector and any action creator that use a certain dynamically loaded state tree.

client/state/rewind/selectors/get-backup-retention-update-status.ts
tyxla: Let's make sure that we use optional chaining only when really necessary. For example, in this case, if `rewind[ siteId ]` exists, then its `retention` property will exist, as well as its `updateBackupRetentionRequestStatus` sub-property.

client/state/rewind/test/selectors.js
```diff
@@ -107,3 +108,37 @@ describe( 'getBackupRetentionDays()', () => {
 		}
 	);
 } );
+describe( 'getBackupRetentionUpdateRequestStatus()', () => {
+	it.each( [
```
tyxla: Same as above for `it.each()`. I believe in this case it doesn't make the tests more readable and would recommend just having separate tests that explain what each test is doing.

client/state/rewind/test/selectors.js
```diff
@@ -107,3 +108,37 @@ describe( 'getBackupRetentionDays()', () => {
...
+		{
+			state: {
+				rewind: {
+					123: {
+						retention: {},
+					},
+				},
+			},
+			siteId: 123,
+			expected: 'unsubmitted',
```
tyxla: We might want to use the constants here as well.

tyxla (commented):
Nice work @rustamshaikh! 👍 

I've left some feedback, let me know what you think.

rustamshaikh:
> Nice work @rustamshaikh! 👍
> 
> I've left some feedback, let me know what you think.

@tyxla thanks a lot for the review and feedback 👍. Lot of learning for me 🙂. I will resolve them one by one and ask for a re-review. Thanks again!

tyxla:
> @tyxla thanks a lot for the review and feedback 👍. Lot of learning for me 🙂. I will resolve them one by one and ask for a re-review. Thanks again!

No worries! Let me know if you need anything or have questions in the meantime.

client/state/rewind/retention/reducer.ts
rustamshaikh: @tyxla Could you give some detail on what you mean by this? I think you are referring to getting rid of `combineReducers` since `updateBackupRetentionRequestStatus` is the only reducer currently present. If we do, how can I declare a single reducer?

tyxla: If a reducer has a single subtree, you don't need that it to be defined as a subtree and can just use the parent reducer. That means, exporting the reducer is just fine without a `combineReducers` call. Just like it's done here:

https://github.com/Automattic/wp-calypso/blob/1d652f90440b0e5b0877e9d96cf876518fae7a78/client/state/concierge/is-user-blocked/reducer.js

client/state/rewind/retention/actions.ts
rustamshaikh: Moved this type to `retention/types.ts`

rustamshaikh:
> Also, not a stopper, but what do you think about using `client/state/backup` instead of `client/state/rewind`, given that we would like to get rid of `rewind` in our codebases.

Probably yes, but along with using the directory with `backup` name in it, I think we should also refactor/rename the internal state tree or any other related things (for rewind -> backup) but then it would probably involve a lot of changes and will make this PR big 🤔

rustamshaikh:
@tyxla could you take a look again 🙂

Initsogar:
> Probably yes, but along with using the directory with `backup` name in it, I think we should also refactor/rename the internal state tree or any other related things (for rewind -> backup) but then it would probably involve a lot of changes and will make this PR big 🤔

@rustamshaikh Thank you! I think we could consider addressing that later. It will also involve to add more unit tests there.

Initsogar (approved):
Looks good to me! Thanks for addressing the comments! :)

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,28 @@
...
+	JETPACK_BACKUP_RETENTION_UPDATE,
+	JETPACK_BACKUP_RETENTION_UPDATE_ERROR,
+	JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS,
+} from 'calypso/state/action-types';
+import { BACKUP_RETENTION_UPDATE_REQUEST } from './constants';
+import type { AppState } from 'calypso/types';
+import type { AnyAction } from 'redux';
+
+const initialState = {
+	updateRequestStatus: BACKUP_RETENTION_UPDATE_REQUEST.UNSUBMITTED,
```
tyxla: Perhaps I didn't cover this well enough the last time, but don't see a good reason to need the form of the state to be an object with a `updateRequestStatus` field. `initialState` could directly hold the string value. We should just set it to default to  `BACKUP_RETENTION_UPDATE_REQUEST.UNSUBMITTED` and then in the reducer, we could set the value directly, and not to a property of an object. This would also simplify the reducer tests, and the selector (and its tests) when retrieving from this state tree.

rustamshaikh: I agree with you and I was doing the same while addressing the feedback, but then I saw other rewind reducers', and all were having the  `requestStatus` as a sub-field, so decided to keep the structure similar in fashion. Also, we might add an action to retrieve the retention days separately (which is currently part of the other endpoint's response and needs refactoring) so as to add related reducers here later on.

tyxla: Other reducers keep that state because it's needed somewhere (whether it's really needed is a different topic). If it's not needed here, then we shouldn't add it. 

With regards to adding a different subtree in the future - that is a good excuse to keep this structure, but I've historically seen often that this doesn't happen, and the extra level remains there. My usual suggestion is to keep things simple and follow the YAGNI principle, and if we end up needing it, that's when we can refactor it.

rustamshaikh: > Other reducers keep that state because it's needed somewhere (whether it's really needed is a different topic). If it's not needed here, then we shouldn't add it.

It's needed and will be used by @Initsogar upcoming PR to confirm updation.

>With regards to adding a different subtree in the future - that is a good excuse to keep this structure, but I've historically seen often that this doesn't happen, and the extra level remains there. My usual suggestion is to keep things simple and follow the YAGNI principle, and if we end up needing it, that's when we can refactor it.

Ok, will revert it and will add it again when needed :)

tyxla: Well, feel free to keep it if it's really needed and going to be used soon.

rustamshaikh: Done, refactored the reducer to hold the status directly.

client/state/rewind/retention/reducer.ts
```diff
@@ -7,22 +7,21 @@ import { BACKUP_RETENTION_UPDATE_REQUEST } from './constants';
...
 
 		case JETPACK_BACKUP_RETENTION_UPDATE_ERROR:
-			return { updateRequestStatus: BACKUP_RETENTION_UPDATE_REQUEST.FAILED };
+			return BACKUP_RETENTION_UPDATE_REQUEST.FAILED;
 	}
 	return state;
 };
 
-export default retention;
+export default updateBackupRetentionRequestStatus;
```
Initsogar, line 27: Thanks @rustamshaikh and @tyxla. I think it makes sense to handle it this way, and if we include the new reducer for the `retentionDays` later in another PR, we can then export the `combinedReducers` and refactor where needed. But that's not high priority for now.

client/state/rewind/retention/reducer.ts
rustamshaikh: > Well, feel free to keep it if it's really needed and going to be used soon.

Will segregate the update request status into a property when we refactor the other endpoint, for now, we can use the status directly as you suggested.

client/state/rewind/retention/reducer.ts
```diff
@@ -0,0 +1,27 @@
...
+		case JETPACK_BACKUP_RETENTION_UPDATE:
+			return BACKUP_RETENTION_UPDATE_REQUEST.PENDING;
+
+		case JETPACK_BACKUP_RETENTION_UPDATE_SUCCESS:
+			return BACKUP_RETENTION_UPDATE_REQUEST.SUCCESS;
+
+		case JETPACK_BACKUP_RETENTION_UPDATE_ERROR:
+			return BACKUP_RETENTION_UPDATE_REQUEST.FAILED;
+	}
+	return state;
```
tyxla, line 24: Looks way simpler 👍

client/state/rewind/selectors/get-backup-retention-update-status.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Returns the update backup retention request's status.
+ *
+ * @param state The application state.
+ * @param siteId The site ID for which to retrieve days of backups saved.
+ * @returns The the status of the request.
+ */
+const getBackupRetentionUpdateRequestStatus = ( state: AppState, siteId: number ): string =>
+	state.rewind[ siteId ]?.updateBackupRetentionRequestStatus ??
```
tyxla, line 13: Considering that the reducer has a default value, `updateBackupRetentionRequestStatus` should be set to `BACKUP_RETENTION_UPDATE_REQUEST.UNSUBMITTED` by default right? So the nullish coalescing is likely unnecessary.

client/state/rewind/test/selectors.js
```diff
@@ -107,3 +109,30 @@ describe( 'getBackupRetentionDays()', () => {
...
+	test( 'should default to UNSUBMITTED when no backup retention update request is being made yet.', () => {
+		const state = {
+			rewind: {
+				[ TEST_SITE_ID ]: {},
+			},
+		};
+		expect( getBackupRetentionUpdateRequestStatus( state, TEST_SITE_ID ) ).toEqual(
+			BACKUP_RETENTION_UPDATE_REQUEST.UNSUBMITTED
+		);
+	} );
```
tyxla, line 133: Worth having a test with a site that doesn't exist in `state.rewind` at all.

tyxla (approved):
This looks great 👍 Nice work!

Left a few final observations, feel free to ship once they're addressed.

🚀

client/state/rewind/selectors/get-backup-retention-update-status.ts
rustamshaikh: Correct! But the unit tests were failing since they run isolated - when testing `getBackupRetentionUpdateRequestStatus` the first case fails and returns `undefined` instead of `UNSUBMITTED`, browser console does show `UNSUBMITTED` as default.

client/state/rewind/test/selectors.js
rustamshaikh: Added :)

rustamshaikh:
Thanks @tyxla and @Initsogar for the comprehensive review 👍

client/state/rewind/selectors/get-backup-retention-update-status.ts
tyxla: Aren't the unit tests supposed to reflect the actual real use case and behavior?

But I do see some rationale in making it tighter 😉 I don't mind keeping the nullish coalescing, just wanted to point it out.

PR #73109, opened 2023-02-08
Show virtual themes in onboarding design picker

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1598
Fixes https://github.com/Automattic/dotcom-forge/issues/1599

## Proposed Changes

- Introduces a new `virtual-themes/onboarding` feature flag so we can ship small iterations of the Virtual Themes work without exposing the feature in production.
- Displays the style variations as separate themes in the design picker step of the onboarding (*).

(*) Thumbnails don't show an actual preview of the virtual theme yet – that will be handled separately in https://github.com/Automattic/dotcom-forge/issues/1603.

Screenshot 2023-02-08 at 17 28 07

## Testing Instructions

- Make sure the virtual themes don't show up.
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>&flags=virtual-themes/onboarding`.
- Make sure the virtual themes show up.
- Make sure their title follow the "Theme - Style variation" format.
- Select a virtual theme.
- Make sure the "actual" theme is selected, and the relevant style variation has been chosen automatically.

### DIFF
--- a/packages/design-picker/src/components/unified-design-picker.tsx
+++ b/packages/design-picker/src/components/unified-design-picker.tsx
@@ -214,7 +214,7 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 
 	const getTitle = () => {
 		if ( design.is_virtual && design.style_variation ) {
-			return `${ design.title } – ${ design.style_variation?.title }`;
+			return `${ design.title } – ${ design.style_variation.title }`;
 		}
 
 		return design.title;
--- a/packages/design-picker/src/components/unified-design-picker.tsx
+++ b/packages/design-picker/src/components/unified-design-picker.tsx
@@ -205,12 +205,24 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 		);
 	}
 
+	const previewDesign = () => {
+		if ( design.is_virtual && design.style_variation ) {
+			return onPreview( design, design.style_variation );
+		}
+		return onPreview( design );
+	};
+
+	const getTitle = () => {
+		if ( design.is_virtual && design.style_variation ) {
+			return `${ design.title } – ${ design.style_variation?.title }`;
+		}
+
+		return design.title;
+	};
+
 	return (
 		<div className="design-picker__design-option">
-			<button
-				data-e2e-button={ isPremium ? 'paidOption' : 'freeOption' }
-				onClick={ () => onPreview( design ) }
-			>
+			<button data-e2e-button={ isPremium ? 'paidOption' : 'freeOption' } onClick={ previewDesign }>
 				<span
 					className={ classnames(
 						'design-picker__image-frame',
@@ -225,7 +237,7 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 				</span>
 				<span className="design-picker__option-overlay">
 					<span id={ makeOptionId( design ) } className="design-picker__option-meta">
-						<span className="design-picker__option-name">{ design.title }</span>
+						<span className="design-picker__option-name">{ getTitle() }</span>
 						{ style_variations.length > 0 && (
 							<div className="design-picker__options-style-variations">
 								<StyleVariationBadges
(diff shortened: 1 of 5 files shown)


Copons:
I've added a screenshot to give a preliminary idea of what this is going to look like.

With the repeated thumbnails (which is known: we are working on that in parallel), I feel huge fatigue in scrolling the list. But, I'm excited to think of how colorful this will become once the thumbnails will be updated!

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -205,12 +205,24 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
...
+		return onPreview( design );
+	};
+
+	const getTitle = () => {
+		if ( design.is_virtual && design.style_variation ) {
+			return `${ design.title } – ${ design.style_variation.title }`;
+		}
+
+		return design.title;
+	};
```
Copons, line 221: As mentioned in D100722#2062749-code, we could do this in the endpoint.

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -205,12 +205,24 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 		);
 	}
 
+	const previewDesign = () => {
+		if ( design.is_virtual && design.style_variation ) {
+			return onPreview( design, design.style_variation );
```
Copons, line 210: There is an instance of `onPreview` that only accepts one parameter:

https://github.com/Automattic/wp-calypso/blob/babd265c6114418f9b59413da06ea57905574ec0/client/signup/steps/design-picker/index.jsx#L164-L176

I'm not quite sure how to get there (my head spins looking at all these onboarding flows!) but we should ensure we don't risk regressions.

Copons (commented):
This works as advertised!
I've left a couple minor comments, but otherwise LGTM!

packages/design-picker/src/types.ts
```diff
@@ -96,6 +96,8 @@ export interface Design {
 	price?: string;
 	software_sets?: SoftwareSet[];
 	is_bundled_with_woo_commerce?: boolean;
+	is_virtual?: boolean;
```
BogdanUngureanu, line 99: I think `is_virtual` and `style_variation` are always defined in the API, so I think these properties should be non-optional.

packages/design-picker/src/components/unified-design-picker.tsx
mmtr: Already replied in the diff, but cross-posting here for visibility.

I actually did that first, but the `onPreview( design, styleVariation )` in the frontend opened the preview with a big "Theme - Title" title:

Screenshot 2023-02-09 at 11 28 45

That's because `design.title` comes with the title already built by the backend, so we'd need to add more logic in the frontend to "unbuild" the title. That seemed more complex / confusing, so I ultimately decided to build the title in the frontend.

packages/design-picker/src/components/unified-design-picker.tsx
mmtr: I think that's a different function. The one here also accepts a 2nd parameter:

https://github.com/Automattic/wp-calypso/blob/4d70821f6e85a329cbb058606b6e91c51d3a9b08/packages/design-picker/src/components/unified-design-picker.tsx#L147

https://github.com/Automattic/wp-calypso/blob/4d70821f6e85a329cbb058606b6e91c51d3a9b08/packages/design-picker/src/components/unified-design-picker.tsx#L233

packages/design-picker/src/types.ts
mmtr: They are always defined for [`StaticDesign`](https://github.com/Automattic/wp-calypso/blob/babd265c6114418f9b59413da06ea57905574ec0/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts#L31-L43) but not for [`GeneratedDesign`](https://github.com/Automattic/wp-calypso/blob/babd265c6114418f9b59413da06ea57905574ec0/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts#L45-L50).

BogdanUngureanu (approved):
LGTM!

PR #73117, opened 2023-02-08
Revert "Plans section: Modernized header and navbar / part 1"

ddc22:
Reverts Automattic/wp-calypso#72960

The above change breaks the header of a domain only view. So the change is to be reverted.

image

Context :  p1675875224401949-slack-C02FMH4G8

@chriskmnds please do have a look at resolving this issue and merging this change back :) .

### Steps to test
1 - Create a free site without a domain
2 - Click on the following card

image

3 - Make sure the following unbroken header view is visible.

image


DustyReagan (approved):
This appears to fix the issue. 👍 

fixed

fixed2

chriskmnds:
It's actually not an edge case per se. I completely missed to update a few cases of the headers in [plans/header.scss](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plans/header.jsx), despite including the layout overrides for simple sites by default.

ddc22:
Hey @chriskmnds but does the existing design handle the extra text and buttons on this view?

image

PR #73121, opened 2023-02-08
Checkout: Add fields for per-country tax requirements

pottedmeat:
Some countries require city, state, and organization to be provided in order for us to store the transaction with our tax provider. This PR provides API and UI updates.

References 1348-gh-Automattic/payments-shilling

## Proposed Changes

* Type information for the updates to the countries API
* Dynamically adds fields during checkout to support these requirements
* Stores the new data in tax location.

## Testing Instructions

* Apply D100762-code
* Run this test in store sandbox (PCYsg-IA-p2)
* Add a product (like a plan upgrade) to your cart and make sure you're on the checkout page
* Verify the previous behavior continues for countries unaffected by the parent ticket (country + postal code if relevant):
* image
* Switch between the countries in the parent ticket
* Verify that each one shows the expected fields:
* Screen Shot 2023-02-08 at 9 58 57 AM
* Screen Shot 2023-02-08 at 10 00 02 AM
* Screen Shot 2023-02-08 at 10 01 00 AM
* Screen Shot 2023-02-08 at 10 01 40 AM
* Verify that if city appears, it can't be left empty:
* Screen Shot 2023-02-08 at 2 16 49 PM
* Add a `.ca` domain name to the cart (skip email)
* Verify the billing information is replaced with contact information
* Switch between the countries in the parent ticket
* Verify that countries that want a business name provided have the organization field automatically expanded:
* image
* Switch the country to "United Kingdom"
* Check "Add VAT details"
* Verify the organization field doesn't appear (it's already provided above)


sirbrillig:
Thanks for doing this! I updated the PR title to clarify that this is designed to affect checkout and not other parts of calypso (although we should be sure to test the other components that use `TaxFields`; I made [that mistake](https://github.com/Automattic/wp-calypso/pull/72827) last time).

Could you add before/after screenshots to the PR description? A picture is worth a thousand words.

client/my-sites/checkout/composite-checkout/components/contact-details-container.tsx
```diff
@@ -59,13 +59,21 @@ export default function ContactDetailsContainer( {
...
+		updateState,
+		updateEmail,
+	} = checkoutActions;
 	const contactDetails = prepareDomainContactDetails( contactInfo );
 	const contactDetailsErrors = prepareDomainContactDetailsErrors( contactInfo );
 	const onChangeContactInfo = ( newInfo: ManagedContactDetails ) => {
 		updateCountryCode( newInfo.countryCode?.value ?? '' );
 		updatePostalCode( newInfo.postalCode?.value ?? '' );
+		updateCity( newInfo.city?.value ?? '' );
+		updateState( newInfo.state?.value ?? '' );
```
sirbrillig: I think you need to add one of these (and all the other associated machinery) for the organization field also. When I type into it, nothing happens.

We might consider doing https://github.com/Automattic/wp-calypso/pull/72507 before this PR so that less code is needed and we would need only one call no matter how many pieces of data there are. I'll clean that PR up and put it back in review.

client/my-sites/checkout/composite-checkout/lib/update-cart-contact-details-for-checkout.ts
```diff
@@ -43,5 +43,7 @@ export async function updateCartContactDetailsForCheckout(
 		vatId: vatDetails.id ?? '',
 		organization: vatDetails.name ?? '',
 		address: vatDetails.address ?? '',
+		city: contactInfo.city?.value ?? '',
+		state: contactInfo.state?.value ?? '',
```
sirbrillig: Something we need to solve, although I'm not sure of the exact place to solve it: even if the fields are hidden, the data for these fields remains in the store. If it was previously filled (eg: via auto-fill from a previous entry or manually before the country was changed) then the (now incorrect) fields will still be sent to the shopping cart.

For example, here's what happens if you fill in the form for Canada, but then change it to Jamaica. The shopping-cart endpoint receives this payload (note that it still sends "Toronto"):

Screenshot 2023-02-09 at 3 53 11 PM

Solving a similar problem in the VAT form was done by explicitly deleting values when the form changes:

https://github.com/Automattic/wp-calypso/blob/d798bd9b87e699f4672077dc7e0bab10d89e0592/client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx#L150-L152

I'm not sure that's the right solution here (we might just want to check which fields are required instead) but I thought I would mention it.

pottedmeat: Good point. I should be able to fix this and make `onChange` more typesafe at the same time.

sirbrillig:
I think this PR will also suffer from this bug I just discovered: https://github.com/Automattic/wp-calypso/issues/73190

If you change the country, the state/province dropdown will be wrong unless it's changed. Since it's an existing problem, I don't think we have to solve it in this PR but if there's an easy solution then we might consider it.

client/my-sites/checkout/composite-checkout/lib/update-cart-contact-details-for-checkout.ts
```diff
@@ -43,5 +43,7 @@ export async function updateCartContactDetailsForCheckout(
 		vatId: vatDetails.id ?? '',
 		organization: vatDetails.name ?? '',
```
sirbrillig: I think this will need to change slightly to properly support `organization`. Currently it's assuming that the organization comes from the VAT details endpoint, but with this change we might instead want to use the contact details version of the value. 🤔  But how do we know when to choose? We don't want to use the contact details organization value if the VAT details are active or if the organization field is not displayed (not required by the country).

sirbrillig:
Thanks for the updates! I think this is nearly there, but I noticed that `organization` is not being sent to the cart. I think I know why and I left a comment above to suggest a solution.

Screenshot 2023-02-10 at 12 27 35 PM

client/my-sites/checkout/composite-checkout/lib/update-cart-contact-details-for-checkout.ts
pottedmeat: Right now, it would only show the organization in contact details if both are present.

pottedmeat: * If domain contact details is being used, organization is excluded from VAT details
* If an organization is required for tax, organization is excluded from VAT details
* If an organization is not required for tax, it's set to undefined in `contactInfo`
* I think this means we can default to `contactInfo.organization` and fall back to `vatDetails.name` as long as `vatDetails.name` is unset if that form is hidden

sirbrillig: > I think this means we can default to contactInfo.organization and fall back to vatDetails.name as long as vatDetails.name is unset if that form is hidden

I think you're right since in this case `vatDetails` is the internal `wpcom-checkout` data store version of the value. However, you can prove it. You can write tests for the different scenarios by adding to these:

https://github.com/Automattic/wp-calypso/blob/feab57902d951c3e9b84e997f4a76def4bdaca94/client/my-sites/checkout/composite-checkout/test/checkout-contact-step.js#L706

sirbrillig:
In the interest of getting this PR merged soonish, here's an idea: if the country requires an organization field, always display it in the tax fields, and if they also are eligible for VAT, show both fields. One will read "Organization" and the other will read "Organization for VAT" (as it is now, but you need to remove the `excludeOrganizationField` prop).  If the VAT organization field is filled in, we send that to the cart endpoint and ignore the other field entirely (for TaxFields; domains can still use it for their thing). It's not the most wonderful UX but I think the simplicity of the logic makes it worth it for now.

sirbrillig:
Nice! Organization works as expected now. 

If only Tax organization is filled, it uses that value:

Screenshot 2023-02-15 at 6 04 40 PM

Screenshot 2023-02-15 at 6 06 52 PM


If both VAT organization and Tax organization are filled, it uses VAT organization:

vat-organization-overriding-tax-organization

Screenshot 2023-02-15 at 6 07 50 PM

sirbrillig (approved):
This looks good to me and works as described! Please correct me if I'm wrong but I think this should be totally safe to merge without the changes in D100762-code since the new "needs..." fields are optional and will just all be empty for now.

sirbrillig:
Ah, I see. We can use subdivision instead of state. That makes sense.

I modified D100762-code locally to provide `tax_needs_subdivision`, and everything seems to work as expected:

Screenshot 2023-02-16 at 12 43 04 PM

The unit tests are failing with a minor (I think) TS error:
```
client/my-sites/checkout/composite-checkout/lib/translate-cart.ts(110,3): error TS2339: Property 'state' does not exist on type '{ country_code?: string | undefined; postal_code?: string | undefined; subdivision_code?: string | undefined; vat_id?: string | undefined; organization?: string | undefined; address?: string | undefined; city?: string | undefined; }'.
```

sirbrillig:
Looks like we forgot Google Workspace: https://github.com/Automattic/wp-calypso/issues/77759

PR #73126, opened 2023-02-08
Updating sharing button mockup styles in marketing section

davemart-in:
## Description

As outlined in [this JP issue](https://github.com/Automattic/jetpack/issues/27817), we plan to update the sharing button styles that are seen at the bottom of each post. This is one of a couple of PRs that need to be updated. 

## Testing

- Visit https://wordpress.com/marketing/sharing-buttons/{site-url}

## Before

before

## After

after

## Related
- https://github.com/Automattic/jetpack/issues/27817
- https://github.com/Automattic/jetpack/pull/28874
- https://github.com/Automattic/jetpack/pull/28876

### DIFF
--- a/client/my-sites/marketing/style.scss
+++ b/client/my-sites/marketing/style.scss
@@ -832,7 +832,7 @@
 	background: #fff;
 	box-shadow: 0 1px 2px rgba(0, 0, 0, 0.12), 0 0 0 1px rgba(0, 0, 0, 0.12);
 	line-height: 23px;
-	padding: 5px 11px 4px 9px;
+	padding: 4px 11px 3px 9px;
 
 	&:hover {
 		box-shadow: 0 1px 2px rgba(0, 0, 0, 0.22), 0 0 0 1px rgba(0, 0, 0, 0.22);
--- a/client/my-sites/marketing/style.scss
+++ b/client/my-sites/marketing/style.scss
@@ -612,7 +612,8 @@
 }
 
 .sharing-buttons-preview__reblog-like .sharing-buttons-preview-button {
-	margin: 0 0 3px;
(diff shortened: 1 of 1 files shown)


creativecoder (commented):
Here's the default view on one of my test sites. Looks like there's a margin missing?

| **Before** | **After** |
| - | - |
| image | image |

---

I see you're removing some of the variable values for color and font size--is that intentional?

creativecoder (commented):
Also, the buttons when adding more no longer look like buttons (and they're too faint in both cases 😦 )

| **Before** | **After** |
| - | - |
| image | image |

davemart-in:
@creativecoder mind retesting when you have a sec?

> Looks like there's a margin missing?

This should be fixed now.

> I see you're removing some of the variable values for color and font size--is that intentional?

Yep. The variable color went with the border that we intentionally removed. The font-size was updated to 13px intentionally by @ollierozdarz's request. There is no preset variable for a size 13 font, so I made an override in this case. I noticed that we do this already in a handful of spots in the codebase.

> the buttons when adding more no longer look like buttons

Oh! Nice catch. This should be fixed now.

> they're too faint in both cases

Yeah, this is because once you click a button you activate it. The faint grey is the deactivated state. I'm not going to touch this aspect of the design within this iteration. Happy to circle back to it though.

creativecoder (approved):
Looks good.

| **Before** | **After** |
| - | - |
| image | image |

davemart-in:
Ignoring failed E2E Tests (mobile) (Web app) test because I've re-run it like 8 times and it seems unrelated to my CSS changes.

davemart-in:
Pre-release tests passed, so I deployed this change.

PR #46224, opened 2020-10-06
Site-level billing: Add receipt view

sirbrillig:
#### Changes proposed in this Pull Request

This adds the site-level version of the receipt page at the url `/purchases/billing-history/:siteSlug/:receiptId` building on the site-level billing-history section added in https://github.com/Automattic/wp-calypso/pull/46175.

Fixes https://github.com/Automattic/wp-calypso/issues/46203

Main project thread: pbOQVh-sc-p2

#### Screenshots

Screen Shot 2020-10-07 at 4 33 42 PM

#### To do

- [x] Add route
- [x] Add page
- [x] Add link to receipt page from history page
- [x] Add Back button and header
- [x] Add redirect for invalid receipts
- [x] Add redirect for invalid site
- [x] Add analytics calls (eg: Print button)

#### Testing instructions

First, because this modifies the account-level receipt page (for easier re-use), we should test that to be sure it is not broken.

- Visit `http://calypso.localhost:3000/me/purchases/billing`.
- Click on "View receipt" for any purchase.
- Verify that you see the correct receipt.
- Verify that the sidebar remains at the account-level.
- Click the "Back" link in the header and verify that you're taken back to the account-level billing history page.

Then we can test the site-level receipt page.

- Visit `http://calypso.localhost:3000/purchases/billing-history/:siteSlug`.
- Click on "View receipt" for any purchase.
- Verify that you see the correct receipt.
- Verify that the sidebar remains at the site-level.
- Click the "Back" link in the header and verify that you're taken back to the site-level billing history page.
- Visit the receipt page again.
- Verify that the "Print" button causes the browser to open a print dialog.
- Manually visit the URL of a receipt that doesn't exist and verify that you are redirected to the site-level billing history page for that site.
- Manually visit the URL of a receipt that _does exist_ but one that exists for a different site than the slug in the URL and verify that you are redirected to the site-level billing history page for the site in the URL.

PR #73129, opened 2023-02-08
Redux: Convert billingTransactions state slice to TypeScript

sirbrillig:
## Proposed Changes

This PR adds TypeScript types to the `state.billingTransactions` Redux state slice. It adds the state types to `IAppState` which can be used in selectors and converts many of the selectors, actions, and reducers to TypeScript.

This makes a subtle change to `useRedirectToHistoryPageOnWrongSiteForTransaction` because it could previously have caused a fatal error when used by the `ReceiptView` component. It's passed the result of the `getPastBillingTransaction()` Redux selector, but that selector can return either a past billing transaction or an "individual" billing transaction (a quite different object stored in a different subtree of the `billingTransactions` state). That issue was added in https://github.com/Automattic/wp-calypso/pull/46224.

## Testing Instructions

Automated tests should be sufficient, although this does replace some uses of lodash, so it's worth manually looking to see if those were done safely.

To test to make sure these changes don't break anything, visit the site-level purchases page for a site with at least one prior purchase. Click on "Billing history" (so the URL should be something like `/purchases/billing-history/example.com`). Then click on a receipt and verify that it displays as expected.


sirbrillig:
Note to self: this may be ready but could use a final walk-through and I should check to make sure my solution for `useRedirectToHistoryPageOnWrongSiteForTransaction` was the right one. Also that removing lodash `find` didn't break the selector.

client/state/selectors/get-past-billing-transaction.ts
```diff
@@ -1,33 +1,30 @@
...
- * @returns {?Object}         The transaction object
  */
 export default createSelector(
-	( state, id ) =>
-		find( getPastBillingTransactions( state ), { id } ) ||
-		getIndividualBillingTransaction( state, id ),
-	( state, id ) => [
+	( state: IAppState, id: number ) =>
+		getPastBillingTransactions( state )?.find(
+			( transaction ) => String( transaction.id ) === String( id )
```
sirbrillig, line 25: This is the biggest change in this PR. Changing from lodash `find()` to `Array.prototype.find()`.

client/my-sites/purchases/billing-history/use-redirect-to-history-page-on-wrong-site-for-transaction.tsx
```diff
@@ -16,6 +20,7 @@ export default function useRedirectToHistoryPageOnWrongSiteForTransaction(
 	const doesTransactionMatchSite =
 		doesTransactionExist &&
 		selectedSiteId &&
+		'items' in transaction &&
```
sirbrillig, line 23: This is the other big change in this PR, although if this were not added and we did pass the wrong object type to this component we would have gotten a fatal error, so hopefully it's safe.

tyxla: Looks safe enough to me 👍 Judging by the fact we're already using `transaction.items` as an array, it's fine to have this check to satisfy TS.

client/state/billing-transactions/individual-transactions/actions.ts
```diff
@@ -0,0 +1,74 @@
...
+	BILLING_TRANSACTION_REQUEST_SUCCESS,
+} from 'calypso/state/action-types';
+import { errorNotice } from 'calypso/state/notices/actions';
+import type { IndividualReceipt } from '../types';
+import type { CalypsoDispatch } from 'calypso/state/types';
+
+import 'calypso/state/billing-transactions/init';
+
+export const requestBillingTransaction =
+	( transactionId: number | string ) => ( dispatch: CalypsoDispatch ) => {
```
tyxla: I wonder if it makes sense to cast this to a number or a string somewhere, so we're certain that we're working with a single data type. I've seen subtle issues coming from the fact that we expect data to be either of these types but we use it as the other one.

client/state/billing-transactions/types.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface BillingTransaction {
+	id: string;
+	service: string;
+	amount: string;
+	tax_amount: string;
+	icon: string;
+	date: string;
+	desc: string;
+	org: string;
```
tyxla, line 21: Is each of these required fields guaranteed to be there?

client/state/billing-transactions/types.ts
```diff
@@ -0,0 +1,103 @@
...
+	credit: string;
+	items: BillingTransactionItem[];
+}
+
+export interface BillingTransactionItem {
+	id: string;
+	type: string;
+	type_localized: string;
+	domain: string;
+	site_id: string;
```
tyxla, line 39: It might be odd that the site ID is a string.

client/state/selectors/get-billing-transactions-by-type.ts
```diff
@@ -0,0 +1,34 @@
...
+	state: IAppState,
+	transactionType: BillingTransactionsTypeUpcoming
+): UpcomingCharge[] | null;
+export function getBillingTransactionsByType(
+	state: IAppState,
+	transactionType: BillingTransactionsTypePast
+): BillingTransaction[] | null;
+export function getBillingTransactionsByType(
+	state: IAppState,
+	transactionType: BillingTransactionsTypeUpcoming | BillingTransactionsTypePast
```
tyxla: Should this be:

```suggestion
	transactionType: BillingTransactionsType
```

client/state/selectors/get-past-billing-transaction.ts
tyxla: Thank you for this 👍 Looks good and quite safe, especially with the explicit type casting.

tyxla (approved):
Great work @sirbrillig! Works well in my testing and the code looks good!

I've left a few observations and suggestions, but none seem to be blocking.  

🚀

client/state/billing-transactions/individual-transactions/actions.ts
sirbrillig: I think you're right, but I typed this to match [the PropType on this data's query component](https://github.com/Automattic/wp-calypso/blob/8349c684ccb816603498bf520220674ee50b2a9c/client/components/data/query-billing-transaction/index.jsx#L28). That said, it's only in use in two places, `ReceiptView` and `BillingReceipt`. `ReceiptView` gets it directly from the URL here (which I assume makes it a string):

https://github.com/Automattic/wp-calypso/blob/8349c684ccb816603498bf520220674ee50b2a9c/client/my-sites/purchases/controller.js#L87

And `BillingReceipt` gets it from the URL also:

https://github.com/Automattic/wp-calypso/blob/8349c684ccb816603498bf520220674ee50b2a9c/client/me/purchases/billing-history/controller.js#L11

So I think it's safe to consider it a string. The same is true for `clearBillingTransactionError()`.

The tests treat the ID as a number, though...

https://github.com/Automattic/wp-calypso/blob/8349c684ccb816603498bf520220674ee50b2a9c/client/state/billing-transactions/individual-transactions/test/actions.js#L117

I have a feeling that the proper thing to do is to actually use a number here and do some light casting. I've pushed commits to do that.

client/state/billing-transactions/types.ts
sirbrillig: I looked at the endpoint to come up with this type and I'm pretty sure all of these are always set (some might be empty strings, but I think they're always strings).

client/state/billing-transactions/types.ts
sirbrillig: I agree, but that's what the endpoint returns:

Screenshot 2023-02-17 at 1 48 20 PM

PR #70452, opened 2022-11-26
Checkout: Use shopping cart items to provide product variants

sirbrillig:
#### Proposed Changes

When viewing checkout, some products have billing term variants available (eg: for a Personal plan you can choose between 1 month, 1 year, or 2 years). These variants are fetched with a complicated combination of data by the `useGetProductVariants()` hook. The hook uses the product slug and quantity from the shopping cart, fetches plan and product data from two other endpoints, then combines that with hard-coded product data in calypso. The process of doing this has several downsides, notably that its prices can be wrong (https://github.com/Automattic/wp-calypso/issues/69964 and https://github.com/Automattic/wp-calypso/issues/55019) and that it requires quite a lot of frontend machinery to recreate data the backend already has.

In D93550-code, we amend the shopping-cart endpoint to include product variant data for each cart item. This PR uses that data to greatly simplify `useGetProductVariants()` so that it no longer needs the product, plan, or hard-coded information. In addition, many of the product variant guards (like not being allowed to purchase plan upgrades that are term downgrades) are no longer required as they will now be handled by the endpoint.

If it works correctly, this PR should not affect the behavior of calypso at all.

#### Screenshots

Screenshot 2022-11-26 at 1 00 44 AM

Screenshot 2022-11-26 at 1 00 56 AM

Screenshot 2022-11-26 at 1 06 58 AM

#### Testing Instructions

To test this properly we'll need to load the shopping cart with various types of products with and without this PR to verify that the product variants are unchanged. 

Some situations to check, comparing each one against the variant picker in production:

- WordPress.com plans with and without discounts.
- Jetpack bundles with and without discounts.
- Jetpack products that have a quantity set.

PR #59969, opened 2022-01-11
Support new Intro Offers Endpoint

cbauerman:
#### Changes proposed in this Pull Request

- Support new `/introductory-offers` endpoint ( D72857-code )
- Use introductory offers to display discount prices on the variant picker in the cart

| Before ( introductory offer not displayed on variant ) | After ( introductory offer displayed on variant ) |
| ------------------------------------------------------ | --------------------------------------------------------|
|  |  |

#### Testing instructions

1. Add an assortment of Jetpack annual plans and products to the cart
2. Click "Edit" in cart to open the product variant selector
2. Verify that in the product variant picker the correct introductory offer ( 50% off ).
3. Verify the cart total is correct.
4. Add an annual product to the cart via `cloud.jetpack.com/pricing` in a logged-out/incognito window.
5. Verify the correct discount and price is shown on the variant selector.

Related to 1201295898168937-as-1201646705905156

PR #73131, opened 2023-02-09
Checkout: Do not load introductory offers when loading checkout

sirbrillig:
## Proposed Changes

https://github.com/Automattic/wp-calypso/pull/59969 added code to load data from the introductory offers endpoint into the Redux store when loading checkout and to show checkout's loading screen while waiting for that data. The purpose of this was to improve the discounts shown in the term variant picker for Jetpack products. However, in https://github.com/Automattic/wp-calypso/pull/70452 we refactored the term variant picker to get all its data from the shopping-cart endpoint instead and the introductory offers data is no longer used.

This PR removes the introductory offers query component from checkout and removes the code that caused checkout to wait for that data to load.

I'm not sure if any other code in calypso uses the introductory offers endpoint, but if not, we could probably remove it from Redux entirely in a follow-up PR. In any case, this should increase the speed at which checkout loads.

## Testing Instructions

You can see that the Introductory Offers data is unused by looking for the selector added by https://github.com/Automattic/wp-calypso/pull/59969. Grep for `getIntroOfferPrice` and you'll see that it's not used by checkout or any of its dependencies. Neither are `getIsIntroOfferRequesting` or `getIntroOfferRequestStatus`.

Follow the testing instructions for https://github.com/Automattic/wp-calypso/pull/59969 to make sure that the bug it fixed does not reappear:

- Add an assortment of Jetpack annual plans and products to the cart
- Verify that in the product variant picker the correct introductory offer ( 50% off ) is shown.
- Verify the cart total is correct.
- Add an annual product to the cart via cloud.jetpack.com/pricing in a logged-out/incognito window.
- Verify the correct discount and price is shown on the variant selector.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -10,7 +10,6 @@ import { useTranslate } from 'i18n-calypso';
 import { Fragment, useCallback, useMemo } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import QueryContactDetailsCache from 'calypso/components/data/query-contact-details-cache';
-import QueryIntroOffers from 'calypso/components/data/query-intro-offers';
 import QueryJetpackSaleCoupon from 'calypso/components/data/query-jetpack-sale-coupon';
 import QueryPlans from 'calypso/components/data/query-plans';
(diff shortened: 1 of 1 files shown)


michaeldcain (approved):
This works well for me. From our timing data, the `/introductory-offers/` endpoint takes ~2s, so this should result in a noticeable gain.

PR #73136, opened 2023-02-09
Stats: Converting Shares component

grzegorz-cp:
Related to #72134 
## Proposed Changes

* replace the current `Shares` component to use horizontal bars
* extend the horizontal bar component to display shortened numbers for values

## Testing Instructions

* open the live branch
* apply `?flags=stats/insights-page-grid` to the URL
* verify that `Shares` module on the `Insights` page has horizontal bars
* verify that without the feature flag the current version of the component loads without problems

| Before | After |
|---|---|
| SCR-20230209-nrn | SCR-20230209-nmq |

## Pre-merge Checklist


grzegorz-cp:
This component is missing social media icons on the left side of the labels, and the card header should be altered (to allow column labels to take up more space than the columns below).
I was planning to address both after converting other modules on the page.

client/my-sites/stats/stats-insights/index.jsx
```diff
@@ -89,8 +89,9 @@ const StatsInsights = ( props ) => {
 							hideNewModule // remove when cleaning 'stats/horizontal-bars-everywhere' FF
 						/>
 
+						{ /* Replaced by new modules on top of the page */ }
 						{ ! isLatestPostReplaced && <LatestPostSummary /> }
-						<Reach />
+						{ /* <Reach /> */ }
```
jsnmoon, line 94:
```suggestion
```

client/my-sites/stats/stats-insights/index.jsx
```diff
@@ -89,8 +89,9 @@ const StatsInsights = ( props ) => {
 							hideNewModule // remove when cleaning 'stats/horizontal-bars-everywhere' FF
 						/>
 
+						{ /* Replaced by new modules on top of the page */ }
 						{ ! isLatestPostReplaced && <LatestPostSummary /> }
-						<Reach />
```
jsnmoon, line 93: I know it's part of our design specifications, but I don't feel comfortable getting rid of the `Subscriber totals` card without preserving its functionality somewhere else in the interface.

I know WPCOM and Email subscriber totals remain accessible via the `Subscribers` card, but social subscriber total isn't easily obtainable in the interface without manual summation.

What are your thoughts?

jsnmoon (approved):
Everything looks good, both with and without the feature flag. Let's ship!

client/my-sites/stats/stats-insights/index.jsx
grzegorz-cp: That's a good point. But on the other hand, I like how clean the cards look without this additional information, and I was wondering if we couldn't put the data into the highlight sections. That's when I realised that we already have this information in the year in review. So the number is detached from the card, but we gain a unified look.
Do you think we should put it in a different place?

grzegorz-cp: The email subscriber total is not included in the year in review. 
Another card at the top could be helpful, aggregating totals for many items.
We could also extend the component to show some footer with additional information and put the totals back on the cards.
The second option is better because it doesn't push down the content that was previously at the top of the page.

grzegorz-cp: I looked into the components further, and since we preserve the information about total wp.com subscribers as part of `Social subscribers(followers)`, we could maintain the information about social networks by adding a row that can be expanded into a list of networks (just like we do on the Traffic page with Authors).

jsnmoon: I also like how much cleaner the interface is. Your observation is correct; the subscribers card in the annual highlights section only counts WPCOM subscribers, which is already represented elsewhere on the page:

image

jsnmoon: At any rate, we can iterate on this refinement next week during our maintenance sprint 🙂

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7803087

Thank you @grzegorz-cp for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7803087).

PR #73078, opened 2023-02-08
WIP Upgrade @wordpress/edit-site to 4.9.0

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/71038, https://github.com/Automattic/wp-calypso/pull/71917

## Proposed Changes

Referring to https://github.com/Automattic/wp-calypso/pull/71917, when we're working on adding the font pairings variations to the site assembler, we need to `"@wordpress/edit-site: ^4.9.0"` and `"@wordpress/blocks: ^11.11.0"` to support changing the font family of the heading via the `elements.heading`. So, this PR proposes to upgrade those 2 packages first

TODOs

- [ ] Fix unit tests

## Testing Instructions

* TBD

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #73155, opened 2023-02-09
Try: Upgrade @wordpress packages to pre-React 18

tyxla:
This PR aims to upgrade the `@wordpress` packages to versions before upgrading to React 18.

Still in progress and very far from ready. 

DO NOT MERGE.


noahtallen:
Ah, I remember starting to look at some of these. I think the big change is that `@wordpress/data` now bundles its own types. As a result, we can remove `@types/wordpress__data`, but this also means we have to rework how we define types for some of our data stores.

Essentially, the issue `Cannot redeclare block-scoped variable 'select'.` happens because we use `declare module '@wordpress/data'` to provide our own types for data stores currently (along with `registerStore`). But `registerStore` has been deprecated now in favor of `register`, which I think allows types to work successfully without declaring the module ourselves.

Once I tried switching from `registerStore` to `register`, I recall an issue with the persistence plugin, which was never updated to handle the newer API

At that point, I had some other priorities so I didn't get a chance to follow-up.

tyxla:
Yup 👍 What you said, essentially. I'm happy to give it a try and see where it goes. Will ping you when I get it closer 😉

tyxla:
This got way messier than I was hoping it to be. I'm going to close it and retry another time. Maybe we should do it in smaller chunks of versions and not with such a huge leap forward.

arthur791004:
Okay! I'll continue on https://github.com/Automattic/wp-calypso/pull/73078 to see how it works 🙂

tyxla:
Thanks, @arthur791004. I've also left some thoughts in https://github.com/Automattic/wp-calypso/pull/73629#issuecomment-1443124320 in case that helps your scenario.

tyxla:
I've started work on a second take in #73890 and will be working on it this week.

PR #73169, opened 2023-02-09
VAT Info Page: Add logging for errors while fetching VAT details

sirbrillig:
When we load the `/me/purchases/vat-details` page, an API request is made to the `/me/vat-info` API endpoint. If that request fails, we do not log it anywhere. If the form on that page is used to change the VAT details, and the update fails, we do log that error.

Recently we've gotten several reports of the GET request failing without an obvious cause. See pNPgK-6n6-p2

Here's what the user sees:

Screenshot 2023-02-09 at 1 06 13 PM

## Proposed Changes

In this PR we add logging for the fetch error as well. (While not required, D100880-code will improve the error message that we get.)

## Testing Instructions

- Sandbox the API and add a fatal error the the `/me/vat-info` GET endpoint.
- In the JS console, run `localStorage.setItem('debug', 'calypso:analytics')` and reload the page to activate the additional logging.
- Verify that you see the `calypso_vat_details_fetch_failure` error logged. (It will include the error message you added if D100880-code is also applied but it should log regardless.)

Screenshot 2023-02-09 at 1 24 43 PM

### DIFF
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -285,7 +285,10 @@ function useRecordVatEvents( {
 	useEffect( () => {
 		if ( fetchError && lastFetchError.current !== fetchError ) {
 			reduxDispatch(
-				recordTracksEvent( 'calypso_vat_details_fetch_failure', { error: fetchError.error } )
+				recordTracksEvent( 'calypso_vat_details_fetch_failure', {
+					error: fetchError.error,
+					message: fetchError.message,
+				} )
 			);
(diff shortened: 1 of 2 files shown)


pottedmeat (approved):
* I applied the backend patch
* I returned a `WP_Error` from `WPCOM_JSON_API_VAT_Info_Endpoint` and saw it displayed in the console in an error matching the name I gave it
* I threw an error in the `get_vat` `try` statement and saw the message in the console as a `VatFetchErrror`
* Code looks good

PR #73174, opened 2023-02-09
WooExpress: Update trial "Plans"page accordion

nelsonec87:
## Proposed Changes

* This PR changes the way the accordion works in the Plans page (for the eCommerce trial)

Now the whole header triggers the toggle, not just the arrow on the right:

Before:
image

After:
image

## Testing Instructions

* Open the Plans page of a site on the eCommerce trial plan (`/plans/<site-slug>`)
* Check that you can toggle the accordion by clicking on any part of the header in both Desktop and Mobile versions.


client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
```diff
@@ -13,87 +14,108 @@
...
+		}
+	}
 
-				.trial-feature-card__content-items-item-icon {
-					padding: 3px 0;
+	.trial-feature-card__items {
+		flex: 1;
+		display: flex;
+		flex-direction: column;
+		max-width: 590px;
```
daledupreez, line 80: This rule feels wrong - I think we should be curbing the size of the container, not some specific inner element. If we keep this, the text in the items ends well before the right edge of the card on wide screens, and that looks odd. Things look fine with it removed, both in mobile and in desktop.
```suggestion
```

For reference, why do the lines end/wrap in the middle of the screen?

Screenshot 2023-02-09 at 22 33 03

client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
```diff
@@ -13,87 +14,108 @@
...
 
-	.trial-feature-card__accordion-toggle {
-		display: inline-flex;
-		cursor: pointer;
-		padding: 10px;
+			.trial-feature-card__item-content {
+				flex: 1;
+				display: flex;
+				flex-direction: column;
+				margin-left: 20px;
```
daledupreez: For mobile screens, I think we should add a rule to either reduce this to 16px to match the right padding, or increase the padding on the right of these items by 4px.

As things stand, the whitespace between the check mark and the text is 20px, but the space between the text and the right edge is 16px, and it looks a bit unbalanced to me. (See screenshot below.)

Screenshot 2023-02-09 at 22 35 08

daledupreez (approved):
I think this is ready to ship as soon as we can address the two small items I flagged. I think they should both block this, but they're pretty minor changes to correct.

client/my-sites/plans/ecommerce-trial/trial-feature-card/style.scss
nelsonec87: Figma:
image

nelsonec87: Not sure why, though

daledupreez: 🤷 Then let's go with it for now. We should probably do a walkthrough with the designers early next week.

PR #73177, opened 2023-02-09
[WordAds] Update ad dashboard for free sites removed from wordads

Alameen688:
## Proposed Changes

Going forward sites enrolled in WordAds would be temporarily removed once they no longer have a valid WordPress.com plan. This PR updates the UX for those sites to:
- Inform them they would be removed from WordAds in the process of canceling their plan.
- Allow them to view their earning and payment history on the Earnings dashboard but prevent updates to Settings
- Provide them an option to rejoin WordAds by upgrading.

## Testing Instructions

* Create a new WordPress.com site and apply to join WordAds under, `My site -> Tools -> Earn -> Earn ad revenue`
* You would subscribe to a premium plan or higher (as part of Step 1)
* Navigate to `My site -> Upgrades -> Manage plan -> Cancel plan`, by clicking on the cancel plan button you should see a modal informing you of the consequences of canceling a plan.

 _In a scenario where the site has a primary domain the message will be combined_
notice

_In case the site has no primary domain at the time the message will be independent_
no primary domain

* Now we need to simulate a subscription plan expiry by manually removing the subscription via the StoreAdmin. Before removing the subscription, make sure you are running the D100709-code revision in your sandbox
* Once you remove the subscription your WordAds site should now have an ineligible status
* Navigate to `My site  -> Tools -> Earn -> View ad dashboard`

* You should be able to view your Earnings and payment history, but the Settings tab should only have a preview with options disabled
*  You should also see an upgrade banner on top of the page
settings

* Upgrade your plan by clicking the "Upgrade" button
* After upgrading, navigate back to `My site -> Tools -> Earn -> View ad dashboard`, you should no longer see the upgrade banner and you should be able to make changes to the site Settings

## Pre-merge Checklist

### DIFF
--- a/client/me/purchases/non-primary-domain-dialog/index.jsx
+++ b/client/me/purchases/non-primary-domain-dialog/index.jsx
@@ -76,24 +76,30 @@ class NonPrimaryDomainDialog extends Component {
 									strong: <strong />,
 								},
 							}
+						) }{ ' ' }
+						{ hasSetupAds && (
+							<>
+								<br />
+								<br />
+								{ translate(
+									'You will also be ineligible for the WordAds program. Visit {{a}}our FAQ{{/a}} to learn more.',
+									{
+										components: {
+											a: (
+												<a
+													href="https://wordads.co/faq/#eligibility-for-wordads"
+													target="_blank"
+													rel="noopener noreferrer"
+												/>
+											),
+										},
+									}
+								) }
+								<br />
+								<br />
+							</>
 						) }
-						{ hasSetupAds &&
-							translate(
-								'{{br/}}{{br/}}You will also be ineligible for the WordAds program. Visit {{a}}our FAQ{{/a}} to learn more.{{br/}}{{br/}}',
-								{
-									components: {
-										a: (
-											<a
-												href="https://wordads.co/faq/#eligibility-for-wordads"
-												target="_blank"
-												rel="noopener noreferrer"
-											/>
-										),
-										br: <br />,
-									},
-								}
-							) }
-						{ translate( ' Would you still like to downgrade your plan?' ) }
+						{ translate( 'Would you still like to downgrade your plan?' ) }
 					</p>
 				</Fragment>
 			</Dialog>
--- a/client/me/purchases/non-primary-domain-dialog/index.jsx
+++ b/client/me/purchases/non-primary-domain-dialog/index.jsx
@@ -11,6 +11,7 @@ class NonPrimaryDomainDialog extends Component {
 		planName: PropTypes.string.isRequired,
 		oldDomainName: PropTypes.string.isRequired,
 		newDomainName: PropTypes.string.isRequired,
+		hasSetupAds: PropTypes.bool,
 	};
 
 	close = () => {
@@ -22,7 +23,7 @@ class NonPrimaryDomainDialog extends Component {
 	};
 
 	render() {
-		const { planName, oldDomainName, newDomainName, translate } = this.props;
+		const { planName, oldDomainName, newDomainName, translate, hasSetupAds } = this.props;
 		const buttons = [
 			{
 				action: 'cancel',
@@ -66,8 +67,7 @@ class NonPrimaryDomainDialog extends Component {
 						) }
 						<br />
 						{ translate(
-							'{{strong}}%(newDomain)s{{/strong}} will be the address that people see when they visit ' +
-								'your site. Would you still like to downgrade your plan?',
+							'{{strong}}%(newDomain)s{{/strong}} will be the address that people see when they visit your site.',
 							{
 								args: {
 									newDomain: newDomainName,
@@ -77,6 +77,23 @@ class NonPrimaryDomainDialog extends Component {
 								},
 							}
 						) }
+						{ hasSetupAds &&
+							translate(
+								'{{br/}}{{br/}}You will also be ineligible for the WordAds program. Visit {{a}}our FAQ{{/a}} to learn more.{{br/}}{{br/}}',
+								{
+									components: {
+										a: (
+											<a
+												href="https://wordads.co/faq/#eligibility-for-wordads"
+												target="_blank"
+												rel="noopener noreferrer"
+											/>
+										),
+										br: <br />,
+									},
+								}
+							) }
+						{ translate( ' Would you still like to downgrade your plan?' ) }
 					</p>
 				</Fragment>
 			</Dialog>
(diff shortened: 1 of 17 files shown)


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7817399

Thank you @Alameen688 for including a screenshot in the description! This is really helpful for our translators.

client/me/purchases/non-primary-domain-dialog/index.jsx
```diff
@@ -77,6 +77,23 @@ class NonPrimaryDomainDialog extends Component {
...
+												href="https://wordads.co/faq/#eligibility-for-wordads"
+												target="_blank"
+												rel="noopener noreferrer"
+											/>
+										),
+										br: <br />,
+									},
+								}
+							) }
+						{ translate( ' Would you still like to downgrade your plan?' ) }
```
sirbrillig: Translating strings that start with a space for layout purposes is not a great idea for localization. What if you just put a space between this string and the previous ones instead?

client/me/purchases/non-primary-domain-dialog/index.jsx
```diff
@@ -77,6 +77,23 @@ class NonPrimaryDomainDialog extends Component {
 								},
 							}
 						) }
+						{ hasSetupAds &&
+							translate(
+								'{{br/}}{{br/}}You will also be ineligible for the WordAds program. Visit {{a}}our FAQ{{/a}} to learn more.{{br/}}{{br/}}',
```
sirbrillig: The BR line break tags probably don't need to be translated and it will make any future modifications to this string harder (if we ever want to increase or decrease the spacing, the whole string will have to go through a two week translation process and many hours of work for translators). Can you put them outside the string?

client/me/purchases/wordads-eligibility-warning-dialog/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+		>
+			<>
+				<FormSectionHeading>
+					{ translate( 'Remove %(plan)s', {
+						args: { plan: planName },
+					} ) }
+				</FormSectionHeading>
+				<p>
+					{ translate(
+						'When you downgrade your plan, you will become ineligible for the WordAds program. Visit {{a}}our FAQ{{/a}} to learn more.{{br/}}{{br/}}',
```
sirbrillig: Same here: the BR tags should not be part of the string sent for translation. Just put them outside the string.

client/me/purchases/wordads-eligibility-warning-dialog/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+										href="https://wordads.co/faq/#eligibility-for-wordads"
+										target="_blank"
+										rel="noopener noreferrer"
+									/>
+								),
+								br: <br />,
+							},
+						}
+					) }
+					{ translate( ' Would you still like to downgrade your plan?' ) }
```
sirbrillig: Same here: please don't translate strings that start with spaces.

client/state/selectors/get-site-wordads-status.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Returns the wordads status for a site
+ */
+export default function getSiteWordadsStatus(
+	state: AppState,
+	siteId: number | undefined | null
+): string | null {
+	if ( ! siteId ) {
+		return null;
+	}
+	return get( state, [ 'wordads', 'status', siteId, 'status' ], null );
```
sirbrillig: Calypso folks have been hard at work for several years now to remove lodash from the codebase. Could you use regular JS here?

```suggestion
	return state.wordads?.status?.[ siteId ]?.status ?? null;
```

Alameen688:
@sirbrillig Thanks for your feedback on the recommended practice for translation and the attempt to discontinue using lodash. I've now implemented the suggested changes.

sirbrillig (commented):
Looks good to me. I'll leave final approval to your team but you took care of all the things I was concerned about.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7817399

Thank you @Alameen688 for including a screenshot in the description! This is really helpful for our translators.

JessBoctor:
👋 

I did a functional review and I ran into a couple of strange behaviors.

**1. Plans page formatting**

On my free site, when I go to `My Site > Tools > Earn > Earn Ad Revenue` and then click on `Unlock this feature`, the plans page layout doesn't match the layout if I go straight to `My Site > Upgrades > Plans`. I am not sure if this is intentional or not.

GHPR73177-PlansPage1Column

GHPR73177-PlansPageLayout

I didn't have this issue when I tested the flow with a German user account. It looks like the plans page is using an older template.

GHPR73177-DEPlansPageFromUnlockAds

**2. Ad Settings Access**

When I canceled the Premium plan from both accounts, I saw the warning modal that I would no longer have access to WordAds. However, when I navigate to the Ads Dashboard, I still see the Ad Settings and do not see the upgrade banner. Removing the plan via StoreAdmin rather than the cancellation flow didn't have an effect on this. I was still able to access the settings in both scenarios.

GHPR73177-AdsSettings-PostCancel

**3. Post Checkout Landing**

This is likely a different issue, but I really would have liked to have been returned to the ads dashboard after completing checkout. It was a bit jarring to have to get back to where I started after checkout.

Tested with: 
- MacOS Ventura 13.1
- Chrome 110.0.5481.100 
- 15" MacBook Pro M1 Chip

Alameen688:
@JessBoctor Thanks a lot for testing this 🙏🏽

1) This is intentional, there is a query parameter that makes the page only show annual Premium plans and higher when you go through the `Unlock this feature` button, and lesser plans are hidden which explains the layout difference.  As for the German user account, I see it also starts from the Premium plan when you go via the `Unlock this feature` button just that the UI is the older design (_I don't know what toggles the old or new design though, maybe a flag_).

2) If you cancel your plan via `My site -> Upgrades -> Manage plan -> Cancel plan`, you would still be on WordAds till your plan actually expires where you would no longer have an active Premium (or higher plan).
However, removing it from StoreAdmin should work. I believe it doesn't work for you because you need to have the corresponding WPCOM revision D100709-code running on your sandbox. _(I just updated the PR description to make that clear)_

3) This is definitely good feedback, the WordAds team is aware of this but it's not part of the improvement we want to make in this change. Ideally, we want to redirect the user back to the ad dashboard.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7817399).

PR #73183, opened 2023-02-09
Hosting redesign: Add hosting onboarding flow

zaguiini:
Part of https://github.com/Automattic/dotcom-forge/issues/1620.

## Proposed Changes

We want to show the plans grid as the first step for users coming from [wordpress.com/hosting](https://wordpress.com/hosting) because as the CTA is on top, they won't see anything they're getting.

This is a temporary fix until we add more CTAs to the page, then we can revert to using the normal onboarding flow steps (plans after domain).

Reasons we're adding this new flow:

- show the plans grid first so the user knows what they're buying (features are shown below the price tag)
- highlight the Business plan instead of the Personal one since that's the one we're selling on wordpress.com/hosting

## Testing Instructions

- Check out this branch and browse to `/start/hosting`
- You should see the plans grid first. Check that `Business` is highlighted and choose that plan
- Sign up and choose a domain name

You should be presented with the checkout page. Enter the payment information and the flow should be completed with a `/home` redirection.

## Additional information

We tried skipping the domain part but it looks like it's actually necessary to have a domain (I think the plan is tied to the domain). Apparently, it's possible to add a stub one (email or username) but you cannot buy a plan without a domain, and you also need to be signed in. That makes it impossible to go to the cart without signing up, so we need the `user` step too.

At the end of the day, we're just shuffling the steps to make `plans` appear first and with the Business plan highlighted.

### DIFF
--- a/client/signup/config/flows-pure.js
+++ b/client/signup/config/flows-pure.js
@@ -16,12 +16,13 @@ export function generateFlows( {
 	getDestinationFromIntent = noop,
 	getDIFMSignupDestination = noop,
 	getDIFMSiteContentCollectionDestination = noop,
+	getHomeDestination = noop,
 } = {} ) {
 	const flows = [
 		{
 			name: HOSTING_LP_FLOW,
 			steps: [ 'plans', 'user', 'domains' ],
-			destination: getSignupDestination,
+			destination: getHomeDestination,
 			description:
 				'Create an account and a blog and give the user the option of adding a domain and plan to the cart.',
 			lastModified: '2023-02-09',
--- a/client/signup/config/flows-pure.js
+++ b/client/signup/config/flows-pure.js
@@ -22,7 +22,8 @@ export function generateFlows( {
 			name: HOSTING_LP_FLOW,
 			steps: [ 'plans', 'user', 'domains' ],
 			destination: getSignupDestination,
-			description: 'Create an account and a blog and then add the starter plan to the users cart.',
+			description:
+				'Create an account and a blog and give the user the option of adding a domain and plan to the cart.',
 			lastModified: '2023-02-09',
 			showRecaptcha: true,
 		},
--- a/client/signup/config/flows-pure.js
+++ b/client/signup/config/flows-pure.js
@@ -1,5 +1,5 @@
 import { isEnabled } from '@automattic/calypso-config';
-import { setupSiteAfterCreation } from '@automattic/onboarding';
+import { HOSTING_LP_FLOW, setupSiteAfterCreation } from '@automattic/onboarding';
 import { translate } from 'i18n-calypso';
 
 const noop = () => {};
@@ -18,6 +18,14 @@ export function generateFlows( {
 	getDIFMSiteContentCollectionDestination = noop,
 } = {} ) {
 	const flows = [
+		{
+			name: HOSTING_LP_FLOW,
+			steps: [ 'plans', 'user', 'domains' ],
+			destination: getSignupDestination,
+			description: 'Create an account and a blog and then add the starter plan to the users cart.',
+			lastModified: '2023-02-09',
+			showRecaptcha: true,
+		},
 		{
 			name: 'account',
 			steps: [ 'user' ],
--- a/client/signup/config/flows.js
+++ b/client/signup/config/flows.js
@@ -169,6 +169,10 @@ function getDIFMSiteContentCollectionDestination( { siteSlug } ) {
 	return `/home/${ siteSlug }`;
 }
 
+function getHomeDestination( { siteSlug } ) {
+	return `/home/${ siteSlug }`;
+}
+
 const flows = generateFlows( {
 	getSiteDestination,
 	getRedirectDestination,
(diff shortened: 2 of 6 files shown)


client/signup/config/flows-pure.js
```diff
@@ -18,6 +18,14 @@ export function generateFlows( {
 	getDIFMSiteContentCollectionDestination = noop,
 } = {} ) {
 	const flows = [
+		{
+			name: HOSTING_LP_FLOW,
+			steps: [ 'plans', 'user', 'domains' ],
```
p-jackson, line 24: Is there a rationale for having a different order from the rest of the flows? It might seem good to delay the `user` step, but last time we tested this (we tried putting the user step as the last step) it didn't actually improve performance.

The other thing is that `domains` is supposed to come before `plans` because otherwise we end up showing you plans that are ineligible for a custom domain. For example what happens if the user chooses the free plan and then they go ahead and pick a domain? Better to have the user pick the domain first and then we can only show plans which allow custom domains.

client/signup/config/flows-pure.js
```diff
@@ -18,6 +18,14 @@ export function generateFlows( {
...
+			destination: getSignupDestination,
+			description: 'Create an account and a blog and then add the starter plan to the users cart.',
```
p-jackson: These don't really get used anywhere but:

```suggestion
			description: 'Create an account and a blog and give the user the option of adding a domain and plan to the cart.',
```

client/signup/config/flows-pure.js
```diff
@@ -18,6 +18,14 @@ export function generateFlows( {
 	getDIFMSiteContentCollectionDestination = noop,
 } = {} ) {
 	const flows = [
+		{
+			name: HOSTING_LP_FLOW,
+			steps: [ 'plans', 'user', 'domains' ],
+			destination: getSignupDestination,
```
p-jackson: We don't have to do it in this PR, but the other customisation we might want to do with the `hosting` flow is to skip the site setup that happens after the site is created. That is, if you continue through the site creation process you end up here.
CleanShot 2023-02-10 at 21 19 15@2x

You can change that by changing the destination to something else.

```suggestion
			destination: getDIFMSiteContentCollectionDestination,
```

This will work, although you'll probably want to copy-paste this function and give it a more semantically correct name.

p-jackson (commented):
You've got precisely the right idea, it's mainly just adding a new flow to `flows-pure.js`.

Given that the original suggestion (pet6gk-4w-p2#comment-86) was to use the `/start/business` flow, I wonder why we want to show the `plans` step at all. It'll just present a bunch of plans that are unrelated to `/hosting`. The `/start/business` flow silently adds the Business plan to the cart, so we could go with that suggestion.
If we do that then we'd want our steps to be something like `[ 'user', 'domains', 'plans-business' ]` (the `plans-business` step magically adds the plan to the cart).

We might even want to remove the `domains` step from the `hosting` flow. When you create a new site on Pressable or AWS or Fly I'm pretty used to the idea of getting a site with a placeholder name that I can change later.

zaguiini:
> Given that the original suggestion (pet6gk-4w-p2#comment-86) was to use the `/start/business` flow, I wonder why we want to show the `plans` step at all.

For people clicking the top CTA, they won't see any details at all of what they're getting -- this is similar to the hosting configuration upsell project that we worked on a few weeks ago.

> It'll just present a bunch of plans that are unrelated to /hosting. The /start/business flow silently adds the Business plan to the cart, so we could go with that suggestion.

The first part is true, but they'll actually see what they're getting at least... I'd go with the latter if the user were presented with the features before adding it to the cart.

> We might even want to remove the domains step from the hosting flow. 

Could be, we could experiment with that.

---

@p-jackson, just to add some context: this is just a temporary solution until we add more CTAs to the page (task on pet6gk-4d-p2), then we can start adding the Business plan to the cart directly.

client/signup/config/flows-pure.js
zaguiini: >Is there a rationale for having a different order from the rest of the flows?

I feel it's a good idea to show the plans grid with the features before asking the user for credentials.

>The other thing is that domains is supposed to come before plans because otherwise we end up showing you plans that are ineligible for a custom domain. For example what happens if the user chooses the free plan and then they go ahead and pick a domain? Better to have the user pick the domain first and then we can only show plans which allow custom domains.

I see. Then maybe we can skip the domain selection altogether? I don't think it's a highlight for hosting services and we're creating a new flow anyway -- shouldn't be a hard sell. Plus, they can buy that later when creating a site for their business plan.

p-jackson:
> just to add some context: this is just a temporary solution until we add more CTAs to the page

Ah! That makes sense.

client/signup/config/flows-pure.js
p-jackson: > I feel a good idea to show the plans grid with the features before asking the user for credentials.

Fair enough. When this PR goes ready for review and Martech take a look they might have a better idea about whether there are gotchas with not having `user` at the front. But I don't see anything wrong with the idea in principle.

> Then maybe we can skip the domain selection altogether?

I think so. As a builder when I'm spinning up some resources on a new hosting provider I'm not expecting to have to think about domain name before I can use it.

p-jackson (approved):
lgtm

I tried to test the flow signing up with an Apple ID too, but that can't be tested until it's in production (can't test using `calypso.live`). Otherwise everything works as expected.

PR #73191, opened 2023-02-09
Added actions to My Plan page

nelsonec87:
## Proposed Changes

* This PR adds action to both "Upgrade now" buttons from the My Plan page, so it redirects the user to the Plans page.

## Testing Instructions

* Open the My Plan page (using a trial site) (`/plans/my-plan/<site-slug>`)
* Check that you are redirected to the Plans page when clicking the buttons
image

### DIFF
--- a/client/my-sites/plans/current-plan/ecommerce-trial/index.tsx
+++ b/client/my-sites/plans/current-plan/ecommerce-trial/index.tsx
@@ -2,6 +2,7 @@ import { Button, Card } from '@automattic/components';
 import { useLocale } from '@automattic/i18n-utils';
 import { useMediaQuery } from '@wordpress/compose';
 import { useTranslate } from 'i18n-calypso';
+import page from 'page';
 import { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { useLocalizedMoment } from 'calypso/components/localized-moment';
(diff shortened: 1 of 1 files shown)


client/my-sites/plans/current-plan/ecommerce-trial/index.tsx
```diff
@@ -44,6 +46,10 @@ const ECommerceTrialCurrentPlan = () => {
 	const isMobile = useMediaQuery( '(max-width: 480px)' );
 	const displayAllIncluded = ! isMobile || showAllTrialFeaturesInMobileView;
 
+	const gotoPlansPage = () => {
+		page.redirect( '/plans/' + selectedSite?.slug );
```
daledupreez, line 50: Nit: it may be worth using the `plansLink()` function from `@automattic/calypso-products` for this URL, just to be more future-proof.
```suggestion
		page.redirect( plansLink( '/plans', selectedSite?.slug ) );
```

daledupreez (approved):
LGTM! We can add tracking in a follow-up PR, as I think it's more important to get this functional for now.

client/my-sites/plans/current-plan/ecommerce-trial/index.tsx
daledupreez: On second thought, I think that may be overkill for now! I'll get this merged shortly.

PR #73192, opened 2023-02-09
Added Banner component to My Plan page

nelsonec87:
## Proposed Changes

* Changed the My Plan page to use the Banner component (Same as Plans page)
image

## Testing Instructions

* Open the My Plan page (using a trial site) (`/plans/my-plan/<site-slug>`)
* Check that you see the top banner and the button is working correctly.
* Open the Plans page and check that the Banner is shown without the button.
 
image

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
@@ -1,4 +1,4 @@
-import { Button, Card } from '@automattic/components';
+import { Card } from '@automattic/components';
 import { useLocale } from '@automattic/i18n-utils';
 import { useTranslate } from 'i18n-calypso';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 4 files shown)


client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
```diff
@@ -67,6 +72,15 @@ const ECommerceTrialBanner = () => {
...
 				</p>
+				{ props.showButton && (
+					<Button
+						className="e-commerce-trial-current-plan__trial-card-cta"
+						primary
+						onClick={ props.onClick }
+					>
+						{ translate( 'Upgrade now' ) }
+					</Button>
+				) }
```
daledupreez: To align with my other comment about this being passed in as a prop.
```suggestion
				{ props.callToAction } // could also just be { callToAction } if we destructure props
```

client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
```diff
@@ -14,7 +14,12 @@ import DoughnutChart from '../../doughnut-chart';
 
 import './style.scss';
 
-const ECommerceTrialBanner = () => {
+interface ECommerceTrialBannerProps {
+	showButton?: boolean;
+	onClick?: () => void;
+}
```
daledupreez: I think the button as a whole should be passed in as a prop -- I think the interface is much simpler if we do that, as the calling context gets to define the styling and behaviour for the button. We already have a problem where the className (and thus the CSS) for the button is specific to the calling context, which is a poor pattern that's likely to break.


```suggestion
interface ECommerceTrialBannerProps {
	callToAction?: JSX.Element | null;
}
```

daledupreez (changes requested):
This is working, but I think the interface needs to be changed before we ship this.

daledupreez (approved):
Thanks for restructuring this! I think that's much cleaner and more flexible.

LGTM!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7807201

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7807201).

PR #73196, opened 2023-02-09
Sensei Tailored Flow: Fix styling for the domain suggestion badges

mikeyarce:
## Proposed Changes

* Move the domain suggestion badge above the domain to prevent overflow layout issues

## Testing Instructions

* Go to http://calypso.localhost:3000/setup/sensei/senseiDomain
* Enter a really really long domain
* See that the badge is above the domain and that there is no overlap

## Before
Screenshot 2023-02-09 at 12 58 05 PM

## After
Screenshot 2023-02-09 at 12 57 06 PM

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/sensei-domain/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/sensei-domain/style.scss
@@ -220,6 +220,15 @@
 	}
 
 	.featured-domain-suggestions {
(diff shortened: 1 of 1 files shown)


yscik:
Seeing a small alignment issue in some cases for the domain (right-aligned because of the reverse as it was set to flex-start):
image

Pushed a small fix for it.

yscik (approved):

Issue #43595, opened 2020-06-23
Reader: some Gutenberg blocks don't display correctly


leandroalonso:
Some blocks don't appear correctly when shown in the Reader.

While I do think that the Reader should keep a consistent visual (and not apply some random colors that the post authors might use) there are a few improvements that we could make when showing them (an example is a table, that doesn't display any border).

(*JFYI: I'm creating that as a single issue so we can keep the discussion centered here. If we decide to fix one type of block or just a small amount of them I can open another issue.*)

- [ ] Cover block
- [x] Button block - #43779 
- [x] Headings (Settings) - #46720, D51569-code
- [x] Group block
- [x] Latest posts block
- [x] Tables  #44276
- [x] File block
- [x] Pay with Paypal block - https://github.com/Automattic/jetpack/pull/17256

### Cover block

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/964
How it looks on the site: https://made4testing0318.blog/2020/04/15/cover-blocks-2/

### Button block

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/1363
How it looks on the site: https://made4testing0318.blog/2020/06/02/many-buttons/

### Headings (Settings)

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/984
How it looks on the site: https://made4testing0318.blog/2020/04/16/headings-paragraphs/

### Group Block

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/1037
How it looks on the site: https://made4testing0318.blog/2020/04/17/group-block-with-nested-groups/

### Latest Posts Block

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/1003
How it looks on the site: https://made4testing0318.blog/2020/04/16/latest-posts-block/
On comments `latest-posts` don't appear at all in the Reader

### Tables

How it looks on Reader: https://wordpress.com/read/blogs/165243437/posts/1717
How it looks on the site: https://charlietest12345.blog/2020/06/23/table-block/

### File block

How it looks on Reader: https://wordpress.com/read/blogs/159889361/posts/1219
How it looks on the site: https://made4testing0318.blog/2020/04/29/core-file-block/

tophertoy:
@leandroalonso I'd be keen to start this one off by looking at the "button block" if that works for you.

I've started something, it's work in progress at the moment.

tophertoy:
[...]

PR #73201, opened 2023-02-10
Fix padding on paragraph in group block

roo2:
Part of https://github.com/Automattic/wp-calypso/issues/43595

I don't think we want to support the `group` block with any special display CSS, but currently it can cause the whitespace to be a little broken in the full post page.

This ensures that paragraphs within group blocks do not have their padding stripped

### Before:
Screen Shot 2023-02-10 at 2 27 55 pm

### After
Screen Shot 2023-02-10 at 2 34 58 pm

### Testing instructions 
see https://github.com/Automattic/wp-calypso/issues/43595

example post: https://wordpress.com/read/blogs/159889361/posts/1037

### DIFF
--- a/client/blocks/reader-full-post/style.scss
+++ b/client/blocks/reader-full-post/style.scss
@@ -168,6 +168,9 @@
 	.has-text-align-right {
 		text-align: right;
 	}
+	.wp-block-group p:last-child {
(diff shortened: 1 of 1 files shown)


davemart-in (approved):
Tested. Works as expected. Nice work!

PR #73202, opened 2023-02-10
Reader: Tidy up "latest posts" block from reader full post page

roo2:
Part of https://github.com/Automattic/wp-calypso/issues/43595

Given the purpose of reader and the full post page is to read one blog post at a time, I don't think that displaying a nested list of posts is in alignment with that goal.

Currently the "latest posts" block looks somewhat broken in reader.

This PR simply hides the block.

### Before
Screen Shot 2023-02-10 at 2 49 32 pm

### After 
Screen Shot 2023-02-10 at 2 49 39 pm

### Testing instructions 

Example post: https://wordpress.com/read/blogs/159889361/posts/1003

### DIFF
--- a/client/blocks/reader-full-post/style.scss
+++ b/client/blocks/reader-full-post/style.scss
@@ -647,6 +647,6 @@
 	}
 }
 
-.wp-block-latest-posts {
-	display: none;
+.wp-block-latest-posts__post-date::before {
+	content: " - ";
 }
--- a/client/blocks/reader-full-post/style.scss
+++ b/client/blocks/reader-full-post/style.scss
@@ -646,3 +646,7 @@
 		}
 	}
 }
+
+.wp-block-latest-posts {
+	display: none;
+}


davemart-in:
Hmm.. I'm not sure we should be hiding this block. As an end-user I suspect I'd see this as a bug if a portion of my content was just missing with no explanation. I wonder if it would be possible to just inject a hyphen after each of the three links.

- [Final Publish Date Test](paOSwF-xG-p2) - August 25, 2021
I created this post using WPiOS 18.1.0.2 (TestFlight beta) at […]
- [testing](saOSwF-testing-p2) - July 13, 2021
blogging reminders
- [Scheduled post](paOSwF-xf-p2) - March 17, 2021
This is a scheduled post made with WPAndroid 16.9-rc-3 Pixel […]

roo2:
but in what context would you add "latest posts" to a _post_? I think it would almost always be used on _pages_ such as the home page.

So if someone does put "Latest Posts" within a post, I think there's no befit really for readers to even see those, since they're available at the top of reader's feed anyway 😄.

davemart-in:
I guess I would say who are we to dictate what blocks someone finds useful to use in a post? For example, I could see someone who published blog posts daily wanting to do a weekly summary post on the weekend that links to all of their posts for the week. The latest posts block could be handy for that.

roo2:
Ok, I've updated this PR to just add the "-" as you suggested 👍

Screen Shot 2023-02-14 at 11 32 56 am

davemart-in (approved):
Nice and simple! 

I know this isn't the solution you intended. But I do think this is the right path forward.

Thanks Eru!

roo2:
yup, no dramas! I don't think it will be the worst thing if we render the latest posts block nicely 😄

PR #73206, opened 2023-02-10
Plugins: Add getFilteredAndSortedPlugins to selectors-ts

robertf4:
## Proposed Changes

This PR adds the `getFilteredAndSortedPlugins` to `selectors-ts`. It will be a replacement for the existing `getPlugins` function in `selectors.js`, with the name being updated to better reflect what it does.

It is the same as the version found in https://github.com/Automattic/wp-calypso/pull/72363, except that I have removed all usages of `lodash` from it as I decided to remove `lodash` from `selectors-ts` in accordance with the plans to remove `lodash` from Calypso.

## Testing Instructions

Review the code and run the tests.

## Pre-merge Checklist

### DIFF
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -135,7 +135,7 @@ export const getFilteredAndSortedPlugins = createSelector(
 		const allPluginsIndexedBySiteId = getAllPluginsIndexedBySiteId( state );
 
 		// Properties on the objects in allPluginsIndexedBySiteId will be modified and the
-		// selector memoization always returns the same object, so use cloneDeep to avoid
+		// selector memoization always returns the same object, so use `structuredClone` to avoid
 		// altering it for everyone.
 		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = structuredClone(
 			siteIds
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -160,28 +160,27 @@ export const getFilteredAndSortedPlugins = createSelector(
 		// Filter the plugins using the pluginFilter if it is set
 		const pluginList =
 			pluginFilter && _filters[ pluginFilter ]
-				? Object.entries( allPluginsForSites )
-						.filter( ( [ , plugin ] ) => _filters[ pluginFilter ]( plugin ) )
-						.reduce( ( obj, [ pluginSlug, plugin ] ) => {
-							obj[ pluginSlug ] = plugin;
+				? Object.values( allPluginsForSites )
+						.filter( ( plugin ) => _filters[ pluginFilter ]( plugin ) )
+						.reduce( ( obj, plugin ) => {
+							obj[ plugin.slug ] = plugin;
 							return obj;
 						}, {} as { [ pluginSlug: string ]: Plugin } )
 				: allPluginsForSites;
 
 		// Sort the plugins alphabetically by slug
-		const sortedPluginListEntries = Object.entries( pluginList ).sort(
-			( [ pluginSlugA ], [ pluginSlugB ] ) => {
-				const pluginSlugALower = pluginSlugA.toLowerCase();
-				const pluginSlugBLower = pluginSlugB.toLowerCase();
-				if ( pluginSlugALower < pluginSlugBLower ) {
-					return -1;
-				} else if ( pluginSlugALower > pluginSlugBLower ) {
-					return 1;
-				}
-				return 0;
+		const sortedPluginListEntries = Object.values( pluginList ).sort( ( pluginA, pluginB ) => {
+			const pluginSlugALower = pluginA.slug.toLowerCase();
+			const pluginSlugBLower = pluginB.slug.toLowerCase();
+			if ( pluginSlugALower < pluginSlugBLower ) {
+				return -1;
+			} else if ( pluginSlugALower > pluginSlugBLower ) {
+				return 1;
 			}
-		);
-		return sortedPluginListEntries.map( ( [ , plugin ] ) => plugin );
+			return 0;
+		} );
+
+		return sortedPluginListEntries;
 	},
 	( state: AppState ) => [ getAllPluginsIndexedBySiteId( state ) ],
 	( state: AppState, siteIds: number[], pluginFilter?: PluginFilter ) => {
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -157,6 +157,7 @@ export const getFilteredAndSortedPlugins = createSelector(
 				}, {} as PluginSites );
 		}
 
+		// Filter the plugins using the pluginFilter if it is set
 		const pluginList =
 			pluginFilter && _filters[ pluginFilter ]
 				? Object.entries( allPluginsForSites )
@@ -167,6 +168,7 @@ export const getFilteredAndSortedPlugins = createSelector(
 						}, {} as { [ pluginSlug: string ]: Plugin } )
 				: allPluginsForSites;
 
+		// Sort the plugins alphabetically by slug
 		const sortedPluginListEntries = Object.entries( pluginList ).sort(
 			( [ pluginSlugA ], [ pluginSlugB ] ) => {
 				const pluginSlugALower = pluginSlugA.toLowerCase();
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -1,5 +1,4 @@
 import { createSelector } from '@automattic/state-utils';
-import { cloneDeep } from 'lodash';
 import { isRequesting, isRequestingForAllSites } from './selectors';
 import type {
 	InstalledPlugins,
@@ -138,7 +137,7 @@ export const getFilteredAndSortedPlugins = createSelector(
 		// Properties on the objects in allPluginsIndexedBySiteId will be modified and the
 		// selector memoization always returns the same object, so use cloneDeep to avoid
 		// altering it for everyone.
-		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = cloneDeep(
+		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = structuredClone(
 			siteIds
 				.map( ( siteId: number ) => allPluginsIndexedBySiteId[ siteId ] )
 				.filter( Boolean )
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -1,6 +1,13 @@
 import { createSelector } from '@automattic/state-utils';
+import { cloneDeep } from 'lodash';
 import { isRequesting, isRequestingForAllSites } from './selectors';
-import type { InstalledPlugins, InstalledPluginData, Plugin } from './types';
+import type {
+	InstalledPlugins,
+	InstalledPluginData,
+	Plugin,
+	PluginFilter,
+	PluginSites,
+} from './types';
 import type { AppState } from 'calypso/types';
 
 import 'calypso/state/plugins/init';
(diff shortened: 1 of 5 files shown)


client/state/plugins/installed/selectors-ts.ts
```diff
@@ -65,6 +72,34 @@ export const getAllPluginsIndexedByPluginSlug = createSelector(
 	]
 ) as { ( state: AppState ): { [ pluginSlug: string ]: Plugin } };
 
+const _filters: { [ key in PluginFilter ]: ( plugin: Plugin ) => boolean } = {
```
robertf4, line 74: This is duplicated with `selectors.js`. My current plan is to eventually duplicate everything here until `selectors-ts.ts` is equivalent to `selectors.js`, at which point I will begin migrating component code to use `selectors-ts.ts`. Once that is complete I will delete `selectors.js` and rename `selectors-ts.ts` to `selectors.ts`, so this state of duplication will be a temporary one.

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,51 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
 		expect( pluginsOne ).toBe( pluginsTwo );
 	} );
 } );
+
+describe( 'getFilteredAndSortedPlugins', () => {
```
robertf4, line 221: These tests could definitely be better, but for now I've just preserved them as they were. All of the tests on these selectors could be improved but that's probably better done in a later PR.

Just for fun, I asked ChatGPT to give me some equivalence classes for this function and evaluate how well the current tests cover them. This was the response:

image

robertf4:
@tyxla, if there's a group that I should be adding for review on these instead of you personally then feel free to let me know. Thanks for all your work reviewing these PRs!

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -1,6 +1,13 @@
 import { createSelector } from '@automattic/state-utils';
+import { cloneDeep } from 'lodash';
```
tyxla: Considering that we now use Node 18+, let's consider using `structuredClone` instead of `cloneDeep`. Any good reason to prefer `cloneDeep`?

client/state/plugins/installed/selectors-ts.ts
tyxla: We could as well have moved this to a separate file if that will help.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +130,61 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+				const pluginSlugBLower = pluginSlugB.toLowerCase();
+				if ( pluginSlugALower < pluginSlugBLower ) {
+					return -1;
+				} else if ( pluginSlugALower > pluginSlugBLower ) {
+					return 1;
+				}
+				return 0;
+			}
+		);
+		return sortedPluginListEntries.map( ( [ , plugin ] ) => plugin );
```
tyxla: We could avoid this `map()` if we use `Object.values()` instead of `Object.entries()`, can't we?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,51 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+		};
+		const { active, autoupdate, update, version, ...pluginProperties } = jetpack;
+		expect( plugins ).toEqual(
+			expect.arrayContaining( [ { ...pluginProperties, sites: siteWithPlugin } ] )
+		);
+	} );
+
+	test( 'Should get a plugin list of length 2 if only site 1 is requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId ], undefined );
+		expect( plugins ).toHaveLength( 2 );
```
tyxla, line 244: Noting that for filtering, these kinds of tests that only test the length are not necessarily precise. A filter could return the same number of items, but the wrong items, and this test would still pass.

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,51 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+	test( 'Should get a plugin list of length 2 if active plugins on both sites are requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId, siteTwoId ], 'active' );
+		expect( plugins ).toHaveLength( 2 );
+	} );
+
+	test( 'Should get a plugin list of length 1 if inactive plugins on site 1 is requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId ], 'inactive' );
+		expect( plugins ).toHaveLength( 1 );
+	} );
+} );
```
tyxla, line 256: Do we also need some test cases for the sorting?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,51 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+	test( 'Should get a plugin list containing jetpack if both sites are requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId, siteTwoId ], undefined );
+		const siteWithPlugin = {
+			[ siteTwoId ]: {
+				active: jetpack.active,
+				autoupdate: jetpack.autoupdate,
+				update: jetpack.update,
+				version: jetpack.version,
+			},
+		};
```
tyxla: Nit: this could be simplified by using destructuring:

```suggestion
		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId, siteTwoId ], undefined );
		const { active, autoupdate, update, version } = jetpack;
		const siteWithPlugin = {
			[ siteTwoId ]: { active, autoupdate, update, version },
		};
```

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,51 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+		const { active, autoupdate, update, version, ...pluginProperties } = jetpack;
+		expect( plugins ).toEqual(
+			expect.arrayContaining( [ { ...pluginProperties, sites: siteWithPlugin } ] )
```
tyxla: Does the order matter?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +130,61 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+		// selector memoization always returns the same object, so use cloneDeep to avoid
+		// altering it for everyone.
+		const allPluginsForSites: { [ pluginSlug: string ]: Plugin } = cloneDeep(
+			siteIds
+				.map( ( siteId: number ) => allPluginsIndexedBySiteId[ siteId ] )
+				.filter( Boolean )
+				.reduce( ( accumulator, current ) => ( { ...accumulator, ...current } ), {} )
+		);
+
+		// Filter the sites object on the plugins so that only data for the requested siteIds is present
```
tyxla, line 147: Can you elaborate on this one? I'm not sure I understand how this is related to data only for requested site IDs.

tyxla (commented):
Thanks for the good work @robertf4 👍 

I think this looks good in general, I've left some feedback and questions, let me know what you think. 🙌

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I agree, I think these tests could use a PR to clean up and expand them. But for this PR I want to keep them mostly as they were before.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I decided to change the test that was looking only for the Jetpack plugin to look for all of them in sorted order to account for this.

client/state/plugins/installed/selectors-ts.ts
robertf4: Yes, I made that change.

client/state/plugins/installed/selectors-ts.ts
robertf4: In my opinion it shouldn't be, but that is how the code was written before. When I first refactored I left the site ID filtering out and it broke a lot of component code which was assuming that this selector filtered out the other sites for it. This was one of the major difficulties I had, so eventually I decided to just preserve the existing behavior and comment it to make it more obvious.

client/state/plugins/installed/test/selectors-ts.ts
tyxla: Sounds good, I'm fine with following up on this in another PR 👍

client/state/plugins/installed/test/selectors-ts.ts
tyxla: I'd say that a single test should not test multiple traits/behaviors. That's why they're called unit tests - because each test tests one unit. This is a minor issue though and not a blocker.

client/state/plugins/installed/selectors-ts.ts
tyxla: FWIW I did share a workaround if the old jest version is the only blocker to using `structuredClone`:

https://github.com/Automattic/wp-calypso/pull/72363#discussion_r1105737449

client/state/plugins/installed/selectors-ts.ts
tyxla: It's worth spending some more time, perhaps in another PR, to make this clearer, even if it's just with some additional inline comments and/or additional unit tests.

client/state/plugins/installed/selectors-ts.ts
robertf4: Good idea, I implemented the workaround.

test/client/setup-test-framework.js
```diff
@@ -29,3 +29,10 @@ global.CSS = {
 jest.mock( 'wpcom-proxy-request', () => ( {
 	__esModule: true,
 } ) );
+
+// structuredClone wasn't available in Jest before verson 28 so this creates
```
tyxla: Nit: I'd suggest prefixing this comment with TODO to make it more actionable.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +129,60 @@ export const getAllPluginsIndexedBySiteId = createSelector(
 	},
 	( state ) => [ getAllPluginsIndexedByPluginSlug( state ), getSiteIdsThatHavePlugins( state ) ]
 ) as { ( state: AppState ): { [ siteId: number ]: { [ pluginSlug: string ]: Plugin } } };
+
+export const getFilteredAndSortedPlugins = createSelector(
+	( state: AppState, siteIds: number[], pluginFilter?: PluginFilter ) => {
+		const allPluginsIndexedBySiteId = getAllPluginsIndexedBySiteId( state );
+
+		// Properties on the objects in allPluginsIndexedBySiteId will be modified and the
+		// selector memoization always returns the same object, so use cloneDeep to avoid
```
tyxla:
```suggestion
		// selector memoization always returns the same object, so use `structuredClone` to avoid
```

tyxla (commented):
Hey @robertf4 let me know if you're planning to further iterate or if this is ready for a final look. Thanks for the great work 🚀

robertf4:
@tyxla it's ready for a final look.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -65,6 +71,34 @@ export const getAllPluginsIndexedByPluginSlug = createSelector(
 	]
 ) as { ( state: AppState ): { [ pluginSlug: string ]: Plugin } };
 
+const _filters: { [ key in PluginFilter ]: ( plugin: Plugin ) => boolean } = {
```
tyxla, line 74: Before we ship this, let's add a TODO comment, indicating that this almost duplicates `_filters` from `selectors.js` as part of the migration to TS and we're about to remove the duplication soon. 

We might want to add that `_filters.isEqual()` is missing right now but it should be added afterward.

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +217,40 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+	} );
+
+	test( 'Should get a plugin list of length 2 if active plugins on both sites are requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId, siteTwoId ], 'active' );
+		expect( plugins ).toHaveLength( 2 );
+	} );
+
+	test( 'Should get a plugin list of length 1 if inactive plugins on site 1 is requested', () => {
+		const plugins = getFilteredAndSortedPlugins( state, [ siteOneId ], 'inactive' );
+		expect( plugins ).toHaveLength( 1 );
```
tyxla, line 254: Ideally, we should alter all these tests to have better precision. Checking for the length is almost never enough, because the item(s) might be incorrect while the length would match.

test/client/setup-test-framework.js
```diff
@@ -29,3 +29,10 @@ global.CSS = {
 jest.mock( 'wpcom-proxy-request', () => ( {
 	__esModule: true,
 } ) );
+
+// TODO: structuredClone wasn't available in Jest before verson 28 so this creates
```
tyxla, line 33: 👍

tyxla (approved):
This is looking great 👍  Thanks @robertf4!

I've left a few last-minute suggestions, I'd ideally prefer them addressed before shipping if you get a chance.

Thanks again!

client/state/plugins/installed/selectors-ts.ts
robertf4: I took out `_filters.isEqual()` because as far as I can tell it's never used. It's not a predicate function so it can't be used like the other filters are (i.e. you can't pass it to `Array.filter`).

I added a comment regarding the entire file instead of just `_filters` because there will be other functions that are duplicated during the transition.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I agree, but these are all pre-existing tests and I haven't added new ones because I am trying to keep these PRs small and focused.

PR #73242, opened 2023-02-10
Virtual themes: Show Premium badge

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1611.
Requires https://github.com/Automattic/wp-calypso/pull/73109.

## Proposed Changes

Virtual themes are now labeled as Premium since they require the Premium plan to unlock the style variation.

Screenshot 2023-02-10 at 14 30 12

## Testing Instructions

- Grab the domain of a site with Limited Global Styles
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>&flags=virtual-themes/onboarding`.
- Make sure the virtual themes show the Premium badge.
- Select a virtual theme
- Make sure the CTA in the preview says "Unlock this style"
- In the same preview screen, pick the default variation
- Make sure the CTA changes to "Continue"
- Switch to a site without Limited Global Styles
- Make sure the virtual themes are labeled as Free

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
@@ -92,6 +92,7 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow } ) => {
 				site && select( SITE_STORE ).siteHasFeature( site.ID, WPCOM_FEATURES_PREMIUM_THEMES )
 		)
 	);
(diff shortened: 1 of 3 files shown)


Copons (approved):
This LGTM.
I don't love how we are forced to send the `shouldLimit` prop across the whole universe, but IIRC we can't help it.

There's just one thing worth mentioning: the tooltip says "Premium theme" which is not accurate.

Screenshot 2023-02-10 at 15 39 50

Since this is behind feature flag, I don't consider it a blocker. We can fix it easily enough after a design pass on the whole screen. 🙂

PR #73243, opened 2023-02-10
WooExpress: Plans page bottom margin

nelsonec87:
## Proposed Changes

* In desktop mode, the trial's Plans page has no space between the last button and the end of the page.
image

* This PR adds the margin
image

## Testing Instructions

* Open the Plans page of a trial site and check margin at the bottom of the page


daledupreez (approved):
LGTM!

I did notice that the complete lack of padding only kicks in for widths > 1400px due to [this rule](https://github.com/Automattic/wp-calypso/blob/34f0eef46774cac04c179372fe4cfd33c6a9c3e1/client/layout/style.scss#L106), so there is lots of trailing whitespace for narrower screens, but I think that's a general issue. That said, I don't know if that's necessary any more, as I don't think we're using help in that location. I'll follow up on that separately.

PR #73246, opened 2023-02-10
Settings: Adjust setting section title's padding

### LINKED ISSUE #72721: Reading Settings: Adjust bottom margin for all setting titles
Initially @saygunnyc has reported that the bottom margin on the **Related posts** setting title seems to be less than of the other settings on the page.
Once inspected it looks like the bottom margin varies for many of the settings. It would be nice to have it justify for all. 

Image

mashikag:
Resolves https://github.com/Automattic/wp-calypso/issues/72721

## Proposed Changes

* change the title's bottom margin to 12px as suggested by @saygunnyc in https://github.com/Automattic/wp-calypso/issues/72721
* note that this change affects the bottom title margin on all caylpso's settings pages

## Testing Instructions

* Go to http://calypso.localhost:3000/settings/reading/your-test-site-here

## Pre-merge Checklist

#### With the margin adjusted
calypso localhost_3000_settings_reading_simple-site blog (1)

### DIFF
--- a/client/my-sites/site-settings/reading-rss-feed-settings/ExcerptSetting.tsx
+++ b/client/my-sites/site-settings/reading-rss-feed-settings/ExcerptSetting.tsx
@@ -19,7 +19,9 @@ export const ExcerptSetting = ( {
 	const translate = useTranslate();
 	return (
 		<FormFieldset>
-			<FormLabel>{ translate( 'For each post in a feed, include' ) }</FormLabel>
+			<FormLabel className="increase-margin-bottom-fix">
+				{ translate( 'For each post in a feed, include' ) }
+			</FormLabel>
 			<FormLabel>
 				{ /* @ts-expect-error FormRadio is not typed and is causing errors */ }
 				<FormRadio
--- a/client/my-sites/site-settings/reading-rss-feed-settings/SyndicationFeedsSetting.tsx
+++ b/client/my-sites/site-settings/reading-rss-feed-settings/SyndicationFeedsSetting.tsx
@@ -23,7 +23,11 @@ export const SyndicationFeedsSetting = ( {
 	const siteFeedUrl = siteUrl ? `${ siteUrl }/feed/` : '';
 	return (
 		<FormFieldset>
-			<FormLabel id={ `${ SYNDICATION_FEEDS_OPTION }-label` } htmlFor={ SYNDICATION_FEEDS_OPTION }>
+			<FormLabel
+				id={ `${ SYNDICATION_FEEDS_OPTION }-label` }
+				className="reduce-margin-bottom-fix"
+				htmlFor={ SYNDICATION_FEEDS_OPTION }
+			>
 				{ translate( 'Syndication feeds' ) }
 			</FormLabel>
 			<div>
--- a/client/my-sites/site-settings/reading-site-settings/BlogPagesSetting.tsx
+++ b/client/my-sites/site-settings/reading-site-settings/BlogPagesSetting.tsx
@@ -16,7 +16,11 @@ export const BlogPagesSetting = ( { value = 10, onChange, disabled }: BlogPagesS
 	const translate = useTranslate();
 	return (
 		<FormFieldset>
-			<FormLabel id={ `${ BLOG_PAGES_OPTION }-label` } htmlFor={ BLOG_PAGES_OPTION }>
+			<FormLabel
+				id={ `${ BLOG_PAGES_OPTION }-label` }
+				htmlFor={ BLOG_PAGES_OPTION }
+				className="reduce-margin-bottom-fix"
+			>
 				{ translate( 'Blog pages' ) }
 			</FormLabel>
 			<div>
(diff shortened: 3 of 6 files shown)


mashikag:
@saygunnyc 👋 
Above in the PR description I included the screenshot of the Reading Settings page with the change applied. Does the change meet yours expectations?

Note that this will apply to all Calypso setting pages, e.g.: General, Writing, Performance, etc.

saygunnyc:
Looking good from what I can see!
It might be better to style this one as a title as well and apply the same spacing. 

Screenshot 2023-02-10 at 3 25 28 PM

yansern:
I'm not sure if changing the whitespace for all the settings page is all that desirable, take this as an example.

I think there's too much whitespace between the label and the explanation text.

image

image

mashikag:
@yansern got it! 👍 I will apply the change to Reading Settings page only then.

yansern:
I think the fix needs to be more targeted than a fix-all. For example, in this screenshot below, the issue mostly because this label comes with a toggle, hence a whitespace is introduced, so perhaps the whitespace need to account for both situations, with or without toggle.
image

mashikag:
> I think the fix needs to be more targeted than a fix-all.

In order to have standardised UI across pages it is crucial to have such spacings well defined for all (a.k.a. design system). It should be as simple as: all setting title's should have bottom-margin of 12px. 

Unfortunately we are lacking well designed components. There is enough inconsistency in the already implemnted UI which tempts us to implement very specific UI rules such as the case @yansern mentions above:
> For example, in this screenshot below, the issue mostly because this label comes with a toggle, hence a whitespace is introduced, so perhaps the whitespace need to account for both situations, with or without toggle.
image

## The Reading Settings page
In the **Reading Settings** page we have no such settings which would have the toggle beside the setting's "title"/label (this is an example of inconsistent UI). Yet the issue was raised after the Reading Settings page was reviewed.
Screen Shot 2023-02-14 at 10 27 48
Above is an example of such spacing inconsistency in the Reading Settings page. The space difference occurs depending on what inline element is inserted in the line below the setting's title. If an inline-element, such as radio button or text input box, is taller than the line height, this will increase the spacing between the title and the text line directly below. 

### Reading Settings Example: More space
Screen Shot 2023-02-14 at 10 39 14
In the example above you can see how we end up with more space in-between the lines when the text input box is inserted. The text input box is taller than the line height (`normal`). The line height set for input box is `1.5` (1.5x the font size) + border size is added to the final height of the line.  

### Reading Settings Example: Less space
Screen Shot 2023-02-14 at 10 40 35
In this example we get an opposite effect as in the example above. The inline toggle button which is inserted into the line directly below the setting's title, has height of `18px` which is only `1px` more than the line height (`normal`). The toggles placement in the front of the line adds a little bit more to the weight of the effect that there is even less space between the title and the next line.

## Possible Solutions
1. We can do a very specific fix and address the specific lines that look "off" in the Reading Settings page only.
   + Quick to implement
    - Adds to styling complexity. 
    - Makes the code more difficult to maintain 
    - Makes it more difficult to introduce new settings
    - Adds on to the problem of non-standardized UI
2. ~~Create a mini-project out of this to address all styling inconsistencies for setting pages in calypso.
    (I do not think this is feasible at the moment)~~
3. This is not a real solution but... it may be wiser that we drop all the non-critical setting page issues related to styles. Looks like there may be a move to transition back to old pages... 🤦‍♂️ pdKhl6-1JP-p2

yansern:
We could expand this work into something more proper, like making a plan by "identifying what combination of common elements under what scenario needs spacing adjusted", like what you said in no. 2. By going this route, I'd recommend clearer test instructions on what specific areas are affected in the settings page so we can check specifically those areas.

Point no.1 where we would start with just fixing what looks "off" in Reading Settings is not a bad idea for a quick turnaround for end-users, targeted CSS isn't a bad thing as long as it is accompanied by one-line of CSS comment that explains what it is actually fixing (and thus reduces the issue with complexity/maintainability by a little bit).

saygunnyc:
@yansern has a good point here; anything across other Settings pages would require testing and adjustments within each of those Settings pages, which probably would not be the most impactful project. 

I am leaning towards focusing on the Reading Settings page itself, although, of course, that would be ideal to have a system. Not sure if it's worth spending time on this now as we might want to challenge that UI too.

mashikag:
> Point no.1 where we would start with just fixing what looks "off" in Reading Settings is not a bad idea for a quick turnaround for end-users, targeted CSS isn't a bad thing as long as it is accompanied by one-line of CSS comment that explains what it is actually fixing (and thus reduces the issue with complexity/maintainability by a little bit).

@yansern I have done so as suggested. Please review the changes if you find time to do so, otherwise I can ask someone else from the team to help review the PR. 🙂

yansern (approved):
The use of `.increase-margin-bottom-fix` and `.reduce-margin-bottom-fix` feels a little janky.

Perhaps an inline CSS like `style={ marginBottom: 2px }"` would've been simpler, with a JSX comment above the `<FormLabel>` line.

Either that, or a more idiomatic CSS selector expression, like `.excerpt-setting label { margin-bottom: 2px; }`.

I'm quite two minds about approving this PR, but since it's just a cosmetic fix, I'm just going ahead and approve it.

TimBroddin:
@apex We seem to have forgotten to merge this 🫠

andregardi:
Just rebased and tested.
Still looking good today.

PR #73248, opened 2023-02-10
Fixed super props not being added in tracks when path is undefined

BogdanUngureanu:
When the event properties don't have a property called 'path',  then we should fallback to the currentPath from Redux.

SInce the Redux path contains the actual path (not the route), we'll use getSiteFragment to guess if this is a site-specific page.
In this manner, every track event will be automatically populated with the blog_id and plan id automatically.

Related to #

## Proposed Changes

* Automatically pass the global information for every event (blog_id, blog_lang, site_plan). 

## Testing Instructions

* Use the calypso.live link and test a random event
* Make sure that the `blog_id`, `blog_lang` and `site_plan_id` parameters are automatically passed for paths that have a /:site in the URL.

## Pre-merge Checklist


dsas (commented):
With this change i'm getting blog_id, blog_lang and site_plan_id for the `calypso_community_translator_disabled` even on /me, whereas without this change I'm not.

Ditto for `calypso_traintracks_render` on /read. 

Should those pages have the events recorded, or is this working too well? 😆

BogdanUngureanu:
No, it shouldn't work like that. The global properties shouldn't be available on those pages. 😿

BogdanUngureanu:
Looking at the original PR from 2018, that was the purpose of that function: to filter these properties for non-site pages.

mmtr (approved):
Cool, works great!

client/lib/analytics/super-props.js
```diff
@@ -14,8 +15,9 @@ const getSuperProps = ( reduxStore ) => ( eventProperties ) => {
 		client: config( 'client_slug' ),
 	};
 
-	const omitSelectedSite =
-		! eventProperties.force_site_id && shouldReportOmitBlogId( eventProperties.path );
+	const path = eventProperties.path ?? getCurrentRoute( state );
```
tyxla, line 18: Note that this will fall back to the current route not only if `eventProperties.path` is `undefined`, but also if it is `null`.

tyxla (approved):
Looks great 👍 

Seems like a scenario that could be worth some unit tests, but that's definitely not a blocker.

client/lib/analytics/super-props.js
miksansegundo: Any idea where could be a bug causing `path` to be a "random" number between 1 and 103?

It should always be a string. See https://github.com/Automattic/wp-calypso/pull/85821

I investigated [getCurrentRoute](https://github.com/Automattic/wp-calypso/blob/b54b89472ed0df5342e6c153769d08fe3b7f8ed0/client/state/selectors/get-current-route.js#L10) and [ROUTE_SET](https://github.com/search?q=repo%3AAutomattic%2Fwp-calypso%20ROUTE_SET&type=code) without finding much, only a suspicious [timestamp](https://github.com/Automattic/wp-calypso/blob/b54b89472ed0df5342e6c153769d08fe3b7f8ed0/client/state/route/reducer.js#L12).

PR #73255, opened 2023-02-10
Shopping Cart: Show persistent error messages in checkout only

sirbrillig:
The shopping-cart endpoint returns an array of `errors` for the client to display. These messages are displayed on every page in calypso that has access to the shopping cart by the `CartMessages` component inside the `CalypsoShoppingCartProvider` (since the masterbar includes a shopping cart, this is pretty much every page). Typically these messages are related to the most recent cart action (for example: adding a product or a coupon). When we are just fetching an existing cart and not making any changes, errors might still occur (for example: if a product is no longer available) but they will not be returned by the endpoint.

However, some error messages are not related to the transaction. I'm calling these "persistent errors" as opposed to the more common "transient errors". Currently these include the "Your purchases are blocked" message and the "You have to pay a chargeback fee" message. Persistent errors have special rules about when they should be shown to a user:

- They should never be shown to the user when we are not in a checkout context (like transient errors).
- They should always be shown to the user in a checkout context (even if no changes are being made to the cart).

This is difficult because it's hard to differentiate between the two types of messages in calypso.

See https://github.com/Automattic/payments-shilling/issues/1384

## Proposed Changes

In D101050-code we separated out persistent messages from transient messages and returned them in a new property, `persistent_errors`. Because they are in a separate property, `CartMessages` won't display them by default on every page (meeting our first requirement above). In this PR we add a flag to enable displaying these messages and use that flag in checkout (meeting our second requirement).

Screenshot 2023-02-10 at 6 34 22 PM

## Testing Instructions

- First, apply D101050-code to your sandbox and sandbox the API.
- On your sandbox, cause a persistent error to be set on the shopping cart. One way to do this is to alter the code that adds the `'blocked'` error in the shopping cart validator so that it always runs. Another way to do this would be to set the `'block_purchases'` user attribute on your test user.
- Reload checkout and verify that you see the "Purchases are currently disabled" error displayed. Dismiss the error.
- Visit another page in calypso (like the Plans page) and verify that you do not see the error displayed. Reload and check again for good measure.

### DIFF
--- a/client/my-sites/checkout/calypso-shopping-cart-provider.tsx
+++ b/client/my-sites/checkout/calypso-shopping-cart-provider.tsx
@@ -7,7 +7,19 @@ import { cartManagerClient } from './cart-manager-client';
 // for calypso and to display error and success messages returned from calls to
 // the cart endpoint.
 // eslint-disable-next-line @typescript-eslint/ban-types
-export default function CalypsoShoppingCartProvider( { children }: PropsWithChildren< {} > ) {
+export default function CalypsoShoppingCartProvider( {
+	children,
+	shouldShowPersistentErrors,
+}: PropsWithChildren< {
+	/**
+	 * Persistent errors like "Purchases are disabled for this site" are returned
+	 * during cart fetch (regular cart errors are transient and only are returned
+	 * when changing the cart). We want to display these errors only in certain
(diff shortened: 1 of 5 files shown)


martian77:
Works as described and makes sense. 👍

PR #73261, opened 2023-02-12
Checkout: add page load performance tracking

michaeldcain:
This adds page load performance tracking for the Checkout section.

#### Testing instructions

* Navigate to `Checkout`. You should see a request to `/logstash` in the network inspector. The call to `/logstash` should happen as close to the render of useful UI as possible (not just Calypso loading or the cart loading). Verify the payload contains data.
* Go to logstash and query for `message:"perf.nav"` and `properties.id:checkout` and check the data is there.


sirbrillig (approved):
I don't have a full understanding of the way that page load tracking works, but this seems to work as expected for me.

PR #73263, opened 2023-02-13
Reader: fix media inside comments

### LINKED ISSUE #52021: Reader: comment wrapper is not clear floating rules
Encountered while using Reader.

It looks like the `wp-block-embed` is set to `float:left`. And because the parent wrapper doesn't apply a clearfix solution, the next comment is displayed incorrectly

image

roo2:
fixes: https://github.com/Automattic/wp-calypso/issues/52021

Exactly as described in the issue, when media is posted in a comment it can be `float: left`, but the parent element doesn't have any kind of clear fix, causing the other comment to come up alongside it and appear broken.

### Before:
115054959-288b9a80-9ee1-11eb-9323-74fae45e9ee9

### After:
Screen Shot 2023-02-13 at 11 47 23 am

### Testing instructions:

see https://wordpress.com/read/blogs/11326809/posts/33863

### DIFF
--- a/client/blocks/comments/post-comment-content.scss
+++ b/client/blocks/comments/post-comment-content.scss
@@ -4,6 +4,7 @@
 // Post comments line-clamping
 .comments__comment-content-wrapper {
 	position: relative;
+	@include clear-fix;
 
 	&.is-single-line,
 	&.is-single-line .comments__comment-content {


davemart-in (approved):
Looks great. Nice fix!

PR #73284, opened 2023-02-13
Refactor blogging prompt card

### LINKED ISSUE #73158: Writing prompts: Move BloggingPromptCard component
### Details

The home card blogging prompt components are used in my sites and in reader. We should move them to `client/components` to reflect the fact they are used in multiple places.

### Checklist

- [ ] Move `BloggingPromptCard` to `client/components` directory

### Related

_No response_

### LINKED ISSUE #73257: Blogging Prompts card: add tracks property to indicate context
### What

Now that we are showing the blogging prompts card in multiple places (Reader and My Home), let's add an additional property that indicates where the Card is being shown.

### Why

This will let us see if one location or another has more engagement, and help inform where else to best place writing prompts in the UI.

### How

_No response_

eoigal:
This PR refactors the blogging-prompt card. The component is now used in 2 places;

* My Sites Home page
* Reader following stream

This PR will move the component into the `client/components` directory.

This PR adds property to set the `viewContext` to where the component is being used. We can include this context when logging activity on tracks so we know where the interaction with the blogging prompt card took place.

This PR also refactors how we handle siteID where the parent component passes in the site ID it wishes to use with the blogging prompt component. In the case of the reader, the primary blog ID is passed. For home page, the selected site ID will be passed.


davemart-in (approved):
I'm not sure I can comment on the code itself, but I did test the code in both spots and didn't see any JS errors or issues. 👍

creativecoder (commented):
A couple suggestions:

The word "context" has a [very specific meaning in React](https://reactjs.org/docs/context.html). To avoid confusion, I'd suggest a different name for the prop. Brainstorming some options: `location`, `renderLocation`, `cardLocation`? (Naming is hard!)

It seems better to use a more specific folder, like `client/components/blogging-prompts/card` or `client/components/blogging-prompts-card`, so we have room to move other blogging prompt related components here as well, if we need to in the future.

creativecoder:
> suggest a different name for the prop.

Another option that comes to mind: `tracksContext`

eoigal:
> Another option that comes to mind: tracksContext

I'll change it to this, makes sense. Thanks.

client/components/blogging-prompt/index.jsx
```diff
@@ -8,47 +8,61 @@ import { useBloggingPrompts } from 'calypso/data/blogging-prompt/use-blogging-pr
...
 		dispatch(
 			recordTracksEvent( 'calypso_customer_home_task_skip', {
 				task: SECTION_BLOGGING_PROMPT,
+				context: viewContext,
 			} )
 		);
 	};
 
+	const renderMenu = () => {
+		// Only render the menu in home view context
```
creativecoder: Why not use a prop here, like `showMenu`, to alter the how the component is rendered? Then the component doesn't need any special knowledge about the surrounding page--it just renders in whatever way the props indicate.

Additionally adding prompts in other locations won't require updating the component, itself.

client/components/blogging-prompt/index.jsx
```diff
@@ -8,47 +8,61 @@ import { useBloggingPrompts } from 'calypso/data/blogging-prompt/use-blogging-pr
...
-
 	const { skipCard } = useSkipCurrentViewMutation( siteId );
 
 	if ( prompts === undefined ) {
 		return null;
 	}
+
+	if ( viewContext !== 'reader' ) {
+		viewContext = 'home';
+	}
```
creativecoder: This doesn't seem like it should be needed? Can we pass the viewContext prop in from /home, as well?

client/components/blogging-prompt/index.jsx
```diff
@@ -8,47 +8,61 @@ import { useBloggingPrompts } from 'calypso/data/blogging-prompt/use-blogging-pr
...
 	const hidePrompts = () => {
 		skipCard( SECTION_BLOGGING_PROMPT );
 		dispatch(
 			recordTracksEvent( 'calypso_customer_home_task_skip', {
 				task: SECTION_BLOGGING_PROMPT,
+				context: viewContext,
```
creativecoder: I tried to see if there's an existing convention for attaching the UI location to a track event. I found both `section` and `location` props used on other events.

client/components/blogging-prompt/prompts-navigation.jsx
```diff
@@ -61,6 +59,7 @@ const PromptsNavigation = ( { prompts } ) => {
 			recordTracksEvent( `calypso_customer_home_answer_prompt`, {
```
creativecoder: I just noticed that the event names here use "customer_home" 🙁. That's very deceptive to someone browsing events since these clicks may come from elsewhere... I think we need to address it sooner rather than later by introducing new event names that don't assume the context of the prompts.

client/components/blogging-prompt/no-responses-icon.jsx
```diff
@@ -4,7 +4,7 @@ export default function NoResponsesIcon() {
 			<path
 				d="m7.82094 2.62187c.06408-.16249.29404-.16249.35812 0l1.36914 3.47149c.01957.04961.05883.08887.10844.10844l3.47146 1.36914c.1625.06408.1625.29404 0 .35812l-3.47146 1.36914c-.04961.01957-.08887.05883-.10844.10844l-1.36914 3.47146c-.06408.1625-.29404.1625-.35812 0l-1.36914-3.47146c-.01957-.04961-.05884-.08887-.10844-.10844l-3.47149-1.36914c-.16249-.06408-.16249-.29404 0-.35812l3.47149-1.36914c.04961-.01957.08887-.05884.10844-.10844z"
 				stroke="#787c82"
-				stroke-width="1.25"
+				strokeWidth="1.25"
```
creativecoder: Nice catch.

client/my-sites/customer-home/locations/secondary/index.jsx
```diff
@@ -32,6 +32,7 @@ const Secondary = ( { cards } ) => {
 					cardComponents[ card ] &&
 					createElement( cardComponents[ card ], {
 						key: card,
+						siteId: siteId,
```
creativecoder: In the future, we could redefine `cardComponents` above to include card specific props. But for now something like this should do it:

```suggestion
						...(card === SECTION_BLOGGING_PROMPT ? { siteId, showMenu: true, viewContext: 'home' } : {}),
```

client/components/blogging-prompt/index.jsx
eoigal: Good suggestion, I'll add a prop for this.

client/components/blogging-prompt/index.jsx
eoigal: We can. I was reluctant as this was the only card on the home page that used it however. I'll make this change and see how it looks.

client/components/blogging-prompt/index.jsx
eoigal: Thanks, I'll go with `location` I think

client/components/blogging-prompt/prompts-navigation.jsx
eoigal: Yeah good catch. I'm going to leave that to a follow up PR though.

client/my-sites/customer-home/locations/secondary/index.jsx
eoigal: Thanks, added this.

client/components/blogging-prompt/prompts-navigation.jsx
eoigal: I added a function to prefix the tracks events based on the `viewContext`

eoigal: I will need to register the new events when they show up.

client/components/blogging-prompt/index.jsx
eoigal: Taken out

creativecoder (approved):
Thanks for all your work on this @eoigal !

This is working well for me. I can see all events expected within home and reader sent to tracks, including the new `calypso_reader_` prefixed events.

PR #73285, opened 2023-02-13
Writing Prompts: Pass index of prompts to card

### LINKED ISSUE #73258: Reader: use a different prompt every time the Blogging Prompts card is shown
### What

The daily writing prompt is injected into the reader feed, but it shows the same prompt every time.

Let's make sure the card uses a different prompt every time it's shown, to help keep it more engaging.

### Why

I always find seeing something repeat in a feed annoying. Rotating the content keep it more interesting.

### How

_No response_

eoigal:
This PR passes an index when rendering the blogging prompt card to allow reader to render a different prompt each time as the user scrolls down the page.

This will make the prompts more interesting and should help with engagement.

### Testing

* Go to Following feed on Reader - http://calypso.localhost:3000/read
* The first prompt should be the first on the list (where the left arrow is disabled)
Screenshot 2023-02-14 at 21 46 55
* Scroll to the next prompt and confirm the prompt increments by 1 (you can check the `Post Answer` URL to see the prompt ID - i.e. http://calypso.localhost:3000/post/testreader9.wordpress.com?answer_prompt=1846
* Scroll through all the prompts until you get to the last prompt ( where the right arrow is disabled)
Screenshot 2023-02-14 at 21 49 25
* Scroll to the next prompt and it should start from the first prompt again (where the left arrow is disabled).

### DIFF
--- a/client/reader/stream/utils.js
+++ b/client/reader/stream/utils.js
@@ -90,8 +90,8 @@ export function getDistanceBetweenRecs( totalSubs ) {
 	);
 }
 
-const MIN_DISTANCE_BETWEEN_PROMPTS = 10;
-const MAX_DISTANCE_BETWEEN_PROMPTS = 50;
+const MIN_DISTANCE_BETWEEN_PROMPTS = 1;
+const MAX_DISTANCE_BETWEEN_PROMPTS = 5;
 
 export function getDistanceBetweenPrompts( totalSubs ) {
 	// the distance between recs changes based on how many subscriptions the user has.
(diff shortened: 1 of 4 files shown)


client/reader/stream/utils.js
```diff
@@ -90,8 +90,8 @@ export function getDistanceBetweenRecs( totalSubs ) {
 	);
 }
 
-const MIN_DISTANCE_BETWEEN_PROMPTS = 10;
-const MAX_DISTANCE_BETWEEN_PROMPTS = 50;
+const MIN_DISTANCE_BETWEEN_PROMPTS = 1;
```
eoigal: This is just to aid testing - will revert these 2 lines before merging

roo2 (approved):
LGTM. Remember to revert the constants!

`MIN_DISTANCE_BETWEEN_PROMPTS`
`MAX_DISTANCE_BETWEEN_PROMPTS`

PR #73063, opened 2023-02-07
Add better pricing messaging on trial plans

jboland88:
## Proposed Changes

* This diff updates the language on the the Jetpack pricing page to better indicate how the $1 for the first month offers work.

## Testing Instructions

* Apply this patch to your testing environment or use the Calypso live link
* Check the `/pricing` page for the Jetpack Cloud environment
* Check the language next to the pricing for the VaultPress Backup and Social Basic, it should read something like "for the first month, then $10 /month billed yearly"
* Click the "More about" link for both these products to view the lightbox - the same language should be used within the lightbox

Screen Shot 2023-02-07 at 4 27 12 PM
Screen Shot 2023-02-07 at 4 26 58 PM

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203923622825716

PR #73291, opened 2023-02-13
Plans: Replace priceAsString with formatCurrency

sirbrillig:
In https://github.com/Automattic/wp-calypso/pull/73063, a helper function called `priceAsString()` was added to the `PlanPrice` component file to be used by Jetpack product cards on the `/plans` page (it was not actually used by `PlanPrice` but presumably it was added there because that component can be used to display prices in HTML?). The goal of this function was to format a raw price into a localized format without HTML. It operated using the `getCurrencyObject()` export of the `@automattic/format-currency` package, building up the formatted string from its parts.

However, the main export of the `@automattic/format-currency` package is the function `formatCurrency()`, which does exactly this behavior, and does it more accurately for all locales (see pbOQVh-2WH-p2). `getCurrencyObject()` is only made available to build HTML with a price.

## Proposed Changes

This PR removes the unnecessary `priceAsString()` function and replaces its call-sites with calls to `formatCurrency()` instead. I've specified `stripZeros: true` for each call because that behavior was included implicitly in `priceAsString()`. (The function also included an optional parameter for `isSmallestUnit` but it was never used so I've excluded it from the new calls.)

## Testing Instructions

This code is all TypeScript so little testing should be required. However, we can run through the testing instructions of https://github.com/Automattic/wp-calypso/pull/73063 to be sure:

* Apply this patch to your testing environment or use the Calypso live link.
* Check the `/pricing` page for the Jetpack Cloud environment.
* Check the language next to the pricing for the VaultPress Backup and Social Basic, it should read something like "for the first month, then $10 /month billed yearly".
* Click the "More about" link for both these products to view the lightbox - the same language should be used within the lightbox.

Screen Shot 2023-02-07 at 4 27 12 PM
Screen Shot 2023-02-07 at 4 26 58 PM

### DIFF
--- a/client/components/jetpack/card/jetpack-product-card/display-price/paid.tsx
+++ b/client/components/jetpack/card/jetpack-product-card/display-price/paid.tsx
@@ -1,6 +1,7 @@
+import formatCurrency from '@automattic/format-currency';
 import { TranslateResult } from 'i18n-calypso';
 import InfoPopover from 'calypso/components/info-popover';
-import PlanPrice, { priceAsString } from 'calypso/my-sites/plan-price';
+import PlanPrice from 'calypso/my-sites/plan-price';
 import PriceAriaLabel from './price-aria-label';
 import TimeFrame from './time-frame';
 import type { Duration } from 'calypso/my-sites/plans/jetpack-plans/types';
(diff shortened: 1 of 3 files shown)


mattgawarecki (approved):
I tested according to the instructions, with several currencies. Everything looks more-or-less identical to production, so I feel comfortable shipping these changes as-is. 🚢

PR #73296, opened 2023-02-13
Plugins: Add getPluginOnSites and related selectors to selectors-ts

robertf4:
## Proposed Changes

This PR adds `getPluginsOnSites`, `getPluginOnSites`, `getPluginOnSite`, and `getPluginsOnSite` to selectors-ts.

## Testing Instructions

Review the code and tests, and ensure that the tests pass.

## Pre-merge Checklist

### DIFF
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -240,5 +240,6 @@ export const getPluginsOnSite = createSelector(
 	},
 	( state, siteId, pluginSlugs ) => [
 		...pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) ),
-	]
+	],
+	( state, siteId, pluginSlugs ) => [ siteId, ...pluginSlugs ].join()
 );
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -186,3 +186,55 @@ export const getFilteredAndSortedPlugins = createSelector(
 		return [ siteIds, pluginFilter ].flat().join( '-' );
 	}
 );
+
+export function getPluginsOnSites( state: AppState, plugins: Plugin[] ) {
+	return Object.values( plugins ).reduce(
+		( acc: { [ pluginSlug: string ]: Plugin }, plugin: Plugin ) => {
+			const siteIds = Object.keys( plugin.sites ).map( Number );
+			const pluginOnSites = getPluginOnSites( state, siteIds, plugin.slug );
+			if ( pluginOnSites ) {
+				acc[ plugin.slug ] = pluginOnSites;
+			}
+			return acc;
+		},
+		{}
+	);
+}
+
+export function getPluginOnSites( state: AppState, siteIds: number[], pluginSlug: string ) {
+	const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
+
+	if ( ! plugin ) {
+		return undefined;
+	}
+
+	for ( const siteId of siteIds ) {
+		if ( plugin.sites[ siteId ] ) {
+			return plugin;
+		}
+	}
+
+	return undefined;
+}
+
+export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: string ) {
+	const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
+
+	const { sites, ...pluginWithoutSites } = plugin;
+
+	return plugin && plugin.sites[ siteId ]
+		? // Because this is a plugin for a single site context only the data for the selected
+		  // site is included. When I changed this file to TypeScript I found that some code
+		  // assumes it will be on the plugin object, while other code assumes it will be on the
+		  // sites object, so I included both.
+		  {
+				...pluginWithoutSites,
+				...plugin.sites[ siteId ],
+				...{ sites: { [ siteId ]: plugin.sites[ siteId ] } },
+		  }
+		: undefined;
+}
+
+export function getPluginsOnSite( state: AppState, siteId: number, pluginSlugs: string[] ) {
+	return pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) );
+}
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -205,10 +205,8 @@ export function getPluginOnSites( state: AppState, siteIds: number[], pluginSlug
 		return undefined;
 	}
 
-	for ( const siteId of siteIds ) {
-		if ( plugin.sites[ siteId ] ) {
-			return plugin;
-		}
+	if ( siteIds.some( ( siteId ) => !! plugin.sites[ siteId ] ) ) {
+		return plugin;
 	}
 
 	return undefined;
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -212,25 +212,33 @@ export function getPluginOnSites( state: AppState, siteIds: number[], pluginSlug
 	return undefined;
 }
 
-export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: string ) {
-	const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
+export const getPluginOnSite = createSelector(
+	( state: AppState, siteId: number, pluginSlug: string ) => {
+		const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
 
-	const { sites, ...pluginWithoutSites } = plugin;
+		const { sites, ...pluginWithoutSites } = plugin;
 
-	if ( ! plugin || ! plugin.sites[ siteId ] ) {
-		return undefined;
-	}
+		if ( ! plugin || ! plugin.sites[ siteId ] ) {
+			return undefined;
+		}
 
-	// To keep compatibility with some behavior that existed before the refactor
-	// in #73296 the returned object has the site specific data lifted onto it, and
-	// the sites property has only the site data for the requested site.
-	return {
-		...pluginWithoutSites,
-		...plugin.sites[ siteId ],
-		...{ sites: { [ siteId ]: plugin.sites[ siteId ] } },
-	};
-}
+		// To keep compatibility with some behavior that existed before the refactor
+		// in #73296 the returned object has the site specific data lifted onto it, and
+		// the sites property has only the site data for the requested site.
+		return {
+			...pluginWithoutSites,
+			...plugin.sites[ siteId ],
+			...{ sites: { [ siteId ]: plugin.sites[ siteId ] } },
+		};
+	},
+	( state ) => [ getAllPluginsIndexedByPluginSlug( state ) ]
+);
 
-export function getPluginsOnSite( state: AppState, siteId: number, pluginSlugs: string[] ) {
-	return pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) );
-}
+export const getPluginsOnSite = createSelector(
+	( state: AppState, siteId: number, pluginSlugs: string[] ) => {
+		return pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) );
+	},
+	( state, siteId, pluginSlugs ) => [
+		...pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) ),
+	]
+);
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -221,6 +221,9 @@ export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: st
 		return undefined;
 	}
 
+	// To keep compatibility with some behavior that existed before the refactor
+	// in #73296 the returned object has the site specific data lifted onto it, and
+	// the sites property has only the site data for the requested site.
 	return {
 		...pluginWithoutSites,
 		...plugin.sites[ siteId ],
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -217,17 +217,15 @@ export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: st
 
 	const { sites, ...pluginWithoutSites } = plugin;
 
-	return plugin && plugin.sites[ siteId ]
-		? // Because this is a plugin for a single site context only the data for the selected
-		  // site is included. When I changed this file to TypeScript I found that some code
-		  // assumes it will be on the plugin object, while other code assumes it will be on the
-		  // sites object, so I included both.
-		  {
-				...pluginWithoutSites,
-				...plugin.sites[ siteId ],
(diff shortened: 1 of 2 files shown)


client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
 	},
 	( state ) => [ getAllPluginsIndexedByPluginSlug( state ), getSiteIdsThatHavePlugins( state ) ]
 ) as { ( state: AppState ): { [ siteId: number ]: { [ pluginSlug: string ]: Plugin } } };
+
+export function getPluginsOnSites( state: AppState, plugins: Plugin[] ) {
+	return Object.values( plugins ).reduce(
```
tyxla: Why is `Object.values()` necessary if `plugins` is already an array?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
 	},
 	( state ) => [ getAllPluginsIndexedByPluginSlug( state ), getSiteIdsThatHavePlugins( state ) ]
 ) as { ( state: AppState ): { [ siteId: number ]: { [ pluginSlug: string ]: Plugin } } };
+
+export function getPluginsOnSites( state: AppState, plugins: Plugin[] ) {
```
tyxla: Is this worth some memoization?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+		return undefined;
+	}
+
+	for ( const siteId of siteIds ) {
+		if ( plugin.sites[ siteId ] ) {
+			return plugin;
+		}
+	}
+
+	return undefined;
```
tyxla: Maybe could be simplified:

```suggestion
	if ( ! plugin ) {
		return undefined;
	}
	
	if ( siteIds.some( ( siteId ) => !! plugin.sites[ siteId ] ) ) {
		return plugin;
	}

	return undefined;
```

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+	const { sites, ...pluginWithoutSites } = plugin;
+
+	return plugin && plugin.sites[ siteId ]
+		? // Because this is a plugin for a single site context only the data for the selected
+		  // site is included. When I changed this file to TypeScript I found that some code
+		  // assumes it will be on the plugin object, while other code assumes it will be on the
+		  // sites object, so I included both.
+		  {
+				...pluginWithoutSites,
+				...plugin.sites[ siteId ],
```
tyxla: Dropping the plugin data with the plugin site data feels odd. Could be worth some comments here that explain what we're adding to this object.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+
+	return undefined;
+}
+
+export function getPluginOnSite( state: AppState, siteId: number, pluginSlug: string ) {
+	const plugin = getAllPluginsIndexedByPluginSlug( state )[ pluginSlug ];
+
+	const { sites, ...pluginWithoutSites } = plugin;
+
+	return plugin && plugin.sites[ siteId ]
```
tyxla: We could have an early return and avoid the ternary.

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +229,108 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+
+		const result = getPluginsOnSites( state, plugins );
+
+		expect( result ).toEqual( {
+			jetpack: jetpackWithSites,
+			akismet: akismetWithSites,
+		} );
+	} );
+
+	it( 'returns an empty object if the plugins have no sites', () => {
```
tyxla: "the plugins have no sites" feels a bit odd. Should we say it the other way around?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +229,108 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+		};
+		const expectedPlugin = {
+			...akismet,
+			...{ sites: { [ siteOneId ]: siteProperties } },
+		};
+		expect( plugin ).toEqual( expectedPlugin );
+	} );
+} );
+
+describe( 'getPuginsOnSite', () => {
```
tyxla:
```suggestion
describe( 'getPluginsOnSite', () => {
```

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,55 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
 	( state ) => [ getAllPluginsIndexedByPluginSlug( state ), getSiteIdsThatHavePlugins( state ) ]
 ) as { ( state: AppState ): { [ siteId: number ]: { [ pluginSlug: string ]: Plugin } } };
+
+export function getPluginsOnSites( state: AppState, plugins: Plugin[] ) {
+	return Object.values( plugins ).reduce(
+		( acc: { [ pluginSlug: string ]: Plugin }, plugin: Plugin ) => {
+			const siteIds = Object.keys( plugin.sites ).map( Number );
+			const pluginOnSites = getPluginOnSites( state, siteIds, plugin.slug );
+			if ( pluginOnSites ) {
+				acc[ plugin.slug ] = pluginOnSites;
```
tyxla: Do we really want to have arrays with `undefined` in them? How is this being handled at the component level afterwards?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +229,108 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+	test( 'Should get an array of undefined if the requested site is not in the current state', () => {
+		expect( getPluginsOnSite( state, siteThreeId, [ 'jetpack' ] ) ).toEqual( [ undefined ] );
```
tyxla, line 368: That test is an example of what I asked above - doesn't it feel unexpected to have an array of `undefined`, and how are we going to be handling this at the component level? Should we just filter out the undefined values?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +229,108 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+
+	test( 'Should get an array with the plugin objects for the requested plugins with site properties lifted onto the plugin object', () => {
+		const plugins = getPluginsOnSite( state, siteOneId, [ 'akismet' ] );
+		expect( plugins ).toEqual( [
+			{
+				...akismetWithSites,
+				...{
+					active: akismetWithSites.sites[ siteOneId ].active,
+					autoupdate: akismetWithSites.sites[ siteOneId ].autoupdate,
+					version: akismetWithSites.sites[ siteOneId ].version,
```
tyxla, line 383: Intentionally missing the `update` key?

client/state/plugins/installed/test/selectors-ts.ts
```diff
@@ -213,3 +229,108 @@ describe( 'getAllPluginsIndexedBySiteId', () => {
...
+	it( 'returns an empty object if the plugins have no sites', () => {
+		const plugins = [
+			{ ...jetpackWithSites, ...{ sites: {} } },
+			{ ...akismetWithSites, ...{ sites: {} } },
+		];
+
+		const result = getPluginsOnSites( state, plugins );
+
+		expect( result ).toEqual( {} );
+	} );
```
tyxla, line 292: Are we missing a test with an empty array of plugins, and/or with incorrect plugin slugs?

tyxla (commented):
Thanks for this, @robertf4! Nice work!

I've left some feedback, let me know what you think.

client/state/plugins/installed/selectors-ts.ts
robertf4: For now I think it's not. The heavy calculation is done in `getAllPluginsIndexedByPluginSlug` which is memoized and this selector depends on, so I believe that if this one is memoized it will have to declare the same dependents as `getAllPluginsIndexedByPluginSlug` which someone could miss updating if future changes are made to those dependents. 

In my tests on large customers I never saw this selector causing performance issues other than the ones it had from calling `getPlugins`, and memoizing introduces future maintenance concerns, so I'm inclined to not memoize it.

client/state/plugins/installed/selectors-ts.ts
robertf4: I think that was just a holdover from how it worked before so I removed it.

client/state/plugins/installed/selectors-ts.ts
robertf4: This is how it worked before, and I tried to preserve the existing behavior of each selector apart from putting the site dependent properties on site objects instead of the plugin object. The components would need to be investigated to see how they would handle it.

client/state/plugins/installed/selectors-ts.ts

client/state/plugins/installed/selectors-ts.ts
robertf4: Good call.

client/state/plugins/installed/selectors-ts.ts
robertf4: I changed the comment to try to clarify this.

It is odd behavior. This is basically a compromise between how the code worked before (putting site properties on the plugin object), and how it works now (site properties remain on site objects).

For this specific case only a single `siteId` is requested, and `getPlugins` was only adding data for the the sites that were requested. That means the object returned by this function always had only the data for a single site so there wasn't undefined behavior when the properties from different site objects were being put onto the same plugin object.

There was some component code calling this selector seeking data for only a single site, and assuming that data was on the plugin object itself. I don't remember exactly why, but when I tried to refactor the component code it was difficult and so I made this compromise here.

I do think this could be worth addressing in a future PR, but I'm not sure if the dev time would be worth it in the end. 

If I had the time I would consider getting rid of all of the selectors in this PR but it would require a lot of changes in the component code. 

First, it is quite confusing to have four selectors all named in the form `/getPlugins?OnSites?/`. My head spins every time I try to distinguish them from one another and the differences in their behavior.

`getPluginOnSites` should be easily replaceable by having components call `getAllPluginsIndexedByPluginSlug` or `getAllPluginsIndexedBySiteId` directly.

`getPluginOnSite` should also be replaceable by having component code call `getAllPluginsIndexedByPluginSlug` or `getAllPluginsIndexedBySiteId` directly, but would require some refactoring so that component code doesn't assume that site data is on the plugin object, and also not assume that the plugin object has only data for the site requested instead of all the sites the plugin is on.

`getPluginsOnSite` is simple logic that doesn't really need to be in a selector, it is essentially an overload of `getPluginOnSite` where the final parameter is an array instead of a string.

`getPluginsOnSites` is sort of self-reflexive where it looks at each plugin object to see what sites it is on, and then gets the plugin for those sites. That should always be that plugin as far as I can tell: a plugin retrieved by its own sites should be itself every time, unless the component code was somehow passing in an incomplete `Plugin` object and getting back a completed one (which shouldn't be possible, and especially so now that the selectors are all in TypeScript). I'm not fully sure why this logic is needed, and I think instead a component could just call `getAllPluginsIndexedByPluginSlug` and then filter that down to only the plugins it needs, but such a change would have been more changes to component code so I didn't do it.

client/state/plugins/installed/selectors-ts.ts
tyxla: Sounds good 👍

client/state/plugins/installed/selectors-ts.ts
tyxla: I'd very much like to see why the components consider an array of `undefined` as expected in this case. Could you provide more details on where and how it worked that way, and why? Thanks in advance!

client/state/plugins/installed/test/selectors-ts.ts
robertf4: In trying to minimize the scope of both this PR and the project overall I think it is best to keep the existing contracts where possible even if they are not optimal.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: Yes, there is no `update` object on the akismet plugin fixture which simulates the case where a plugin has no updates.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I added tests for these cases.

client/state/plugins/installed/test/selectors-ts.ts
tyxla: The thing is that this is a behavior I'm not convinced works well in the first place. If you don't have a chance to work on it, perhaps it could be a good thing to raise it as a separate issue. The thing is that you have more context than most folks who work in the codebase. I'm fine with a follow-up, but I think we should make sure that this doesn't cause any previously uncaught unexpected behavior anywhere in the component tree.

tyxla:
Hey, @robertf4 let me know when you've finished iterating and this is ready for another look.

client/state/plugins/installed/test/selectors-ts.ts
robertf4: I looked a bit through component code uses of this selector and did find a place that depends on the `undefined` values:

https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx#L93

If `undefined` is filtered out then `pluginsOnSite.every( ( pluginOnSite ) => !! pluginOnSite )` would return true every time, so `areAllPluginsOnSite` would always be `true`, and I don't think that is intended to be the case there.

robertf4:
@tyxla this is ready for final review.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,51 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+	const { sites, ...pluginWithoutSites } = plugin;
+
+	if ( ! plugin || ! plugin.sites[ siteId ] ) {
+		return undefined;
+	}
+
+	// To keep compatibility with some behavior that existed before the refactor
+	// in #73296 the returned object has the site specific data lifted onto it, and
+	// the sites property has only the site data for the requested site.
+	return {
```
tyxla: This will always return a new object, regardless of whether there are differences in the data. Memoize?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -95,3 +95,51 @@ export const getAllPluginsIndexedBySiteId = createSelector(
...
+		...pluginWithoutSites,
+		...plugin.sites[ siteId ],
+		...{ sites: { [ siteId ]: plugin.sites[ siteId ] } },
+	};
+}
+
+export function getPluginsOnSite( state: AppState, siteId: number, pluginSlugs: string[] ) {
+	return pluginSlugs.map( ( pluginSlug ) => getPluginOnSite( state, siteId, pluginSlug ) );
```
tyxla: This will always return a new array, regardless of whether there are differences in the data. Memoize?

tyxla (approved):
LGTM 👍 

I've just left a few suggestions for further optimization, could be addressed in subsequent PRs if you prefer.

Thanks @robertf4 🚀

PR #73300, opened 2023-02-14
BackupRetention: Update request status state after successful API call

Initsogar:
Related to #73035

The idea of this PR is to set the retentionDays in the redux state once the `/backup/retention/update` WPCOM API call succeed.

This started being discussed in [this comment](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104531849) review in #73035.

## Proposed Changes
* Create new `JETPACK_BACKUP_RETENTION_SET` action type to handle the backup retention set in the state.
* Set the `retentionDays` after the `updateBackupRetention` action API succeed.

### Unit testing
You can run the unit testing using: `yarn run test-client client/state/rewind`

### Acceptance testing
* The Update settings button will be disabled when the current plan or an option that requires an upgrade are selected.

## Pre-merge Checklist


client/components/backup-retention-management/index.tsx
```diff
@@ -149,14 +149,12 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
 	useEffect( () => {
-		if ( updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.SUCCESS ) {
-			// Update the current retention plan to the one updated by the user.
-			setCurrentRetentionPlan( retentionSelected );
-			setConfirmationDialogVisible( false );
-		} else if ( updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.FAILED ) {
+		if (
+			updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.SUCCESS ||
+			updateRetentionRequestStatus === BACKUP_RETENTION_UPDATE_REQUEST.FAILED
+		) {
```
tyxla, line 155: Could be simplified with `includes()`, can't it (untested)? 
```suggestion
		if ( [ BACKUP_RETENTION_UPDATE_REQUEST.SUCCESS, BACKUP_RETENTION_UPDATE_REQUEST.FAILED ].includes( updateRetentionRequestStatus ) ) {
```

client/state/rewind/size/reducer.ts
```diff
@@ -67,13 +68,16 @@ const daysOfBackupsSaved = (
 
 const retentionDays = (
 	state: AppState = null,
-	{ type, size }: AnyAction
+	{ type, size, retentionDays }: AnyAction
```
tyxla: It feels off to destructure this for all actions, since it won't exist for the `REWIND_SIZE_SET` action. We should likely do the destructuring inline for each action, or just not destructure at all.

This is the only blocking change IMHO.

client/state/rewind/size/test/reducer.js
```diff
@@ -61,5 +65,23 @@ describe( 'rewind.size reducers', () => {
...
+				action: { type: JETPACK_BACKUP_RETENTION_SET, retentionDays: 30 },
+				expected: 30,
+			},
+			{
+				state: { retentionDays: 30 },
+				action: { type: JETPACK_BACKUP_RETENTION_SET, retentionDays: 7 },
+				expected: 7,
+			},
+		] )(
+			'should return the new retentionDays value if we set it using the action JETPACK_BACKUP_RETENTION_SET',
```
tyxla: Note that it's useful to have the different values inside the test's name if we're doing this, otherwise when this test fails it will display a cryptic message. 

That's why I generally recommend avoiding `it.each()` unless it really saves a lot of effort. Here it only does a simple `expect()` assertion so we're not saving much code. That's the other thing about testing: saving code often means we sacrifice readability and comprehensibility.

tyxla (commented):
The code looks good, thanks @Initsogar 👍 
I've left a few suggestions, just one is blocking, the rest are optional. Thanks!

client/components/backup-retention-management/index.tsx
Initsogar: I did try this, but it was throwing this error:
`Argument of type 'string' is not assignable to parameter of type '"success" | "failed"'.ts(2345)`

It seems related to the fact that we are returning a string from `getBackupRetentionUpdateRequestStatus()`, so we should probably add another type there to ensure the function is returning either one of the `BACKUP_RETENTION_UPDATE_REQUEST` values.

client/state/rewind/size/reducer.ts
Initsogar: Thanks for clarifying this. It does make sense!
Just fixed it :)

client/state/rewind/size/test/reducer.js
Initsogar: Yeah, I think we can get rid of it.each in this scenario. Just made the change :)
Thank you for the suggestion!

Initsogar:
@tyxla I think this is ready for a re-review. Could you take a look? :)

client/state/rewind/size/reducer.ts
```diff
@@ -65,15 +66,15 @@ const daysOfBackupsSaved = (
 	return size.daysOfBackupsSaved ?? null;
 };
 
-const retentionDays = (
-	state: AppState = null,
-	{ type, size }: AnyAction
-): AppState | number | null => {
-	if ( type !== REWIND_SIZE_SET ) {
-		return state;
+const retentionDays = ( state: AppState = null, payload: AnyAction ): AppState | number | null => {
```
tyxla: I'd rename `payload` to `action` since it actually is an action and the `payload` is usually only part of the action.

client/state/rewind/size/test/reducer.js
```diff
@@ -61,5 +65,21 @@ describe( 'rewind.size reducers', () => {
 				expect( sizeReducer( state, action ).retentionDays ).toEqual( expected );
 			}
 		);
+
+		it( 'should return the new retentionDays value if we set it using the action JETPACK_BACKUP_RETENTION_SET and state is undefined', () => {
+			const state = undefined;
+			const action = { type: JETPACK_BACKUP_RETENTION_SET, retentionDays: 30 };
+			const expected = 30;
+
+			expect( sizeReducer( state, action ).retentionDays ).toEqual( expected );
```
tyxla: Now that we're using `state` just once, we can just call it inline. Same for all other variables - `action` and `expected`. This makes the test quite short and straightforward, as you can see 😉

tyxla (approved):
Left 2 more minor suggestions, but this looks good to me otherwise 🚀 Feel free to ship regardless.

client/state/rewind/size/reducer.ts
Initsogar: Makes sense! Just changed it :D

client/state/rewind/size/test/reducer.js
Initsogar: Great, just made it shorter!

PR #73303, opened 2023-02-14
BackupRetention: Refactor radio retention options to be in a component state

Initsogar:
Related to #73035

The idea of this PR is to address multiple comments made as part of the code review in #73035. This includes a refactor in the way we are handling the `retentionOptionsCards` which will be redeclaring those on every render.

## Proposed Changes

* Move `retentionOptionsCards` to a component state as an array instead of an object and modify it dynamically when any value changes. **Addresses:** [[redeclare on every render](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104532289)], [[array vs object](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104531199)].
* Rename spaceNeeded prop to spaceNeededInBytes for more clarity. **Addresses:** [[odd prop name](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104529533)].
* Remove unnecessary `RETENTION_OPTIONS` constant. Use RetentionPeriod type instead on `RetentionRadioOptionType` type. **Addresses:** [[simple and probably unnecessary object](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104532911)], [[comment incomplete?](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104519512)]
* Stop using `useMemo` on helper functions. Given that they are not really complex, and we moved the logic to a useEffect() it is preferable to not use a hook for those calculations. **Addresses:** [[are those memoizations needed?](https://github.com/Automattic/wp-calypso/pull/73035/files#r1104533766)]
* Remove unnecessary `usePrepareRetentionOptions` hook. No longer needed.
* Refactored unit tests

### Acceptance testing
* The Update settings button will be disabled when the current plan or an option that requires an upgrade are selected.

## Pre-merge Checklist


client/components/backup-retention-management/hooks.tsx
```diff
@@ -1,24 +1,10 @@
...
-	const siteId = useSelector( getSelectedSiteId ) as number;
-	const lastBackupSize = useSelector(
-		( state ) => getBackupCurrentSiteSize( state, siteId ) as number
-	);
-	return useMemo(
-		() => lastBackupSize + lastBackupSize * STORAGE_ESTIMATION_ADDITIONAL_BUFFER,
-		[ lastBackupSize ]
-	);
+export function getEstimatedCurrentSiteSize( lastBackupSize: number ): number {
+	return lastBackupSize + lastBackupSize * STORAGE_ESTIMATION_ADDITIONAL_BUFFER;
```
tyxla: Might as well be:

```suggestion
	return lastBackupSize * ( STORAGE_ESTIMATION_ADDITIONAL_BUFFER + 1 );
```

client/components/backup-retention-management/hooks.tsx
```diff
@@ -28,50 +14,9 @@ export function useEstimatedCurrentSiteSize(): number {
  * @param desiredRetentionDays   Desired backup retention days
  * @returns Space needed in bytes
  */
-export function useEstimateSpaceNeeded(
+export function getSpaceNeededInBytes(
```
tyxla: Do we really need a function that multiplies 2 numbers? I think we can just replace it with a simple multiplication where necessary, can't we?

client/components/backup-retention-management/test/hooks.tsx
```diff
@@ -1,80 +1,25 @@
...
-				const { result } = renderHook( () => useEstimatedCurrentSiteSize() );
-				expect( result.current ).toBe( expectedEstimation );
+				const result = getEstimatedCurrentSiteSize( lastBackupSize );
+				expect( result ).toBe( expectedEstimation );
 			}
 		);
 	} );
 
-	describe( 'useEstimateSpaceNeeded', () => {
+	describe( 'getSpaceNeededInBytes', () => {
```
tyxla: These tests are an indication that we don't really need this function and we can just inline the multiplication.

client/components/backup-retention-management/hooks.tsx
```diff
@@ -1,24 +1,10 @@
...
-export function useEstimatedCurrentSiteSize(): number {
-	const siteId = useSelector( getSelectedSiteId ) as number;
-	const lastBackupSize = useSelector(
-		( state ) => getBackupCurrentSiteSize( state, siteId ) as number
-	);
-	return useMemo(
-		() => lastBackupSize + lastBackupSize * STORAGE_ESTIMATION_ADDITIONAL_BUFFER,
-		[ lastBackupSize ]
-	);
+export function getEstimatedCurrentSiteSize( lastBackupSize: number ): number {
```
tyxla: I think we can remove this function, too. Seems like a straightforward calculation we can do inline.

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,104 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
+		},
+		{
+			label: translate( '1 year' ),
+			spaceNeededInBytes: 0,
+			upgradeRequired: false,
+			isCurrentPlan: false,
+			value: 365,
+			checked: false,
+		},
+	] );
```
tyxla: This object will be reinitialized on every re-render. We could benefit from use lazy initialization. See [this](https://kentcdodds.com/blog/use-state-lazy-initialization-and-function-updates) for more info.

It also feels like a weird data type to live in state. Let's try to refactor and use as little state as possible. The rest that is always the same can live outside of the state.

tyxla (commented):
I love seeing all the red lines 🔴 ! Thanks!

I've left a few more suggestions for simplification. Let me know what you think!

client/components/backup-retention-management/hooks.tsx
Initsogar: Yeah, it looks cleaner. Just changed it :)

client/components/backup-retention-management/hooks.tsx
Initsogar: We were anticipating to future changes, but as you said previously, it is better to follow YAGNI principle. If we needed in the future, then we will create those functions again.

Initsogar: We just removed both functions and its tests :)

client/components/backup-retention-management/index.tsx
Initsogar: > This object will be reinitialized on every re-render. We could benefit from use lazy initialization. See [this](https://kentcdodds.com/blog/use-state-lazy-initialization-and-function-updates) for more info.

Great resource. Thanks for the tip. I just updated this to add lazy initialization.

> It also feels like a weird data type to live in state. Let's try to refactor and use as little state as possible. The rest that is always the same can live outside of the state.

We double-checked the code and we were able to decouple some props that are not necessarily needed on the state.
* Moved `label` to a constant and use it directly when rendering the radio button.
* Handled `checked` and `isCurrentPlan` on the radio control.

client/components/backup-retention-management/hooks.tsx
Initsogar: Done!

Initsogar:
@tyxla  I think this is ready for a re-review. It would be great if you could take another look :)

client/components/backup-retention-management/constants.ts
```diff
@@ -1,11 +1,15 @@
...
 
 export const STORAGE_RETENTION_LEARN_MORE_LINK =
 	'https://jetpack.com/support/backup/#how-is-storage-usage-calculated';
+
+export const RETENTION_OPTIONS_LABELS: Record< RetentionPeriod, TranslateResult > = {
+	[ 7 ]: translate( '7 days' ),
+	[ 30 ]: translate( '30 days' ),
+	[ 120 ]: translate( '120 days' ),
+	[ 365 ]: translate( '1 year' ),
+};
```
tyxla: This way of declaring the labels once at runtime during the first import may cause some i18n issues. If the locale gets changed, or new translations get loaded in the meantime AFTER this constant has been declared, the strings will remain unchanged, and `translate()` will not be called again. 

One way forward would be to just avoid declaring these strings separately in an object, and use them inline as needed. That would be my preferred way - I don't see a good reason to reuse those strings. What do we earn by reusing them?

If we really insist these strings to be reused, we can have them in a hook like `useStorageStatusText()` [does](https://github.com/Automattic/wp-calypso/blob/9f9d18171543a4bc366b0edd2ade4631bedf57d9/client/components/backup-storage-space/usage-warning/use-storage-status-text.ts#L10), or (not recommended) use [getters](https://github.com/Automattic/wp-calypso/blob/try/upgrade-wordpress-packages-pre-react18/client/components/domains/use-my-domain/utilities/option-info.js#L39) or [callbacks](https://github.com/Automattic/wp-calypso/blob/1740a4c11f8e8fcc313e04d84e636ceacc8c4aae/client/lib/plans/features-list.tsx#L851).

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,82 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
+
+	// Update the retention options cards with the correct values dynamically
+	useEffect( () => {
+		if ( isFetching ) {
+			return;
+		}
+
+		setRetentionOptionsCards( ( previousOptions ) => {
+			const newOptions = previousOptions.map( ( option ) => {
+				const spaceNeededInBytes = estimatedCurrentSiteSize * option.id;
```
tyxla: Looks like the formula for calculating `spaceNeededInBytes` is predictable: it's always equal to the ID, times `estimatedCurrentSiteSize`. That means that `spaceNeededInBytes` is likely not necessary to be in the state and can be derived from the `id`.

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,82 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
+				const upgradeRequired = spaceNeededInBytes > storageLimitBytes;
```
tyxla: Same as above for `upgradeRequired` - it can be derived from `spaceNeededInBytes` and `storageLimitBytes` therefore it does not need to be stored in `state`. 

There's an interesting parallel we can draw between that kind of functional component derived state and the original React derived state that is supported for the class component syntax. You can read more about `getDerivedStateFromProps()` [here](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) and why it's usually not necessary [in this article](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html).

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,82 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
-		if ( value ) {
-			setRetentionSelected( value );
-		}
-	}, [] );
+	const onRetentionSelectionChange = useCallback(
+		( value: number ) => {
+			if ( value ) {
+				setRetentionSelected( value );
+				const selectedOption = retentionOptionsCards.find( ( option ) => option.id === value );
+				setStorageUpgradeRequired( selectedOption?.upgradeRequired ?? false );
```
tyxla: Should we call `setStorageUpgradeRequired` only if the new `storageUpgradeRequired` value is different from the current one?

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,82 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
+			id: 365,
+			spaceNeededInBytes: 0,
+			upgradeRequired: false,
+		},
+	];
+
+	const [ retentionOptionsCards, setRetentionOptionsCards ] = useState( getOptionsInitialState );
+
+	// Update the retention options cards with the correct values dynamically
+	useEffect( () => {
```
tyxla: This effect seems to be unnecessary. In the subsequent comments, I'm suggesting the removal of some state properties, and after removing them, the options become independent and unchanged by this effect. 

If we really need to render something different while loading, we can instead render a placeholder while `isFetching()` is `true` and then render the options as they are. See comments below.

client/components/backup-retention-management/index.tsx
tyxla: Is there really a possibility that `selectedOption` will be nullish? Is the optional chaining necessary to satisfy TS? If yes, and `selectedOption` will always be there, perhaps we need to tell the TS compiler about that fact?

client/components/backup-retention-management/index.tsx
```diff
@@ -70,53 +73,82 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
 
 	// Set the retention period selected when the user selects a new option
-	const onRetentionSelectionChange = useCallback( ( value: number ) => {
-		if ( value ) {
-			setRetentionSelected( value );
-		}
-	}, [] );
+	const onRetentionSelectionChange = useCallback(
+		( value: number ) => {
+			if ( value ) {
```
tyxla, line 95: In which case is `value` fasly? Are we intentionally not doing anything in that case?

client/components/backup-retention-management/retention-options/retention-option-card.tsx
```diff
@@ -52,7 +52,7 @@ const RetentionOptionCard: FunctionComponent< RetentionOptionCardProps > = ( {
 			</div>
 			<div className="retention-option__space-needed">
 				<div className="space-needed__label">{ translate( 'Space needed:' ) }</div>
-				<div className="space-needed__value">{ spaceNeeded }</div>
+				<div className="space-needed__value">{ useStorageText( spaceNeededInBytes ) }</div>
```
tyxla: You might want to move the hook call to the top of the function. Not a requirement, but it gives better visibility on the fact that we're calling a hook and reduces the chance of someone breaking the [rules of hooks](https://reactjs.org/docs/hooks-rules.html) in the future.

client/components/backup-retention-management/retention-options/retention-options-control.tsx
```diff
@@ -28,10 +30,14 @@ const RetentionOptionsControl: React.FC< RetentionOptionsControlProps > = ( {
...
+			{ retentionOptions.map( ( option ) => (
 				<RetentionOptionCard
-					{ ...option }
-					key={ option.value }
+					value={ option.id }
+					spaceNeededInBytes={ option.spaceNeededInBytes }
+					upgradeRequired={ option.upgradeRequired }
+					checked={ retentionSelected === option.id }
+					isCurrentPlan={ currentRetentionPlan === option.id }
+					key={ option.id }
```
tyxla: I'd like to point out that using the value as a key might have unexpected consequences and cause unexpected behavior. 

That can mean that if the value changes, the key will also change, and React will be unable to reuse the component, creating a new one. It can also mean that React might unintentionally reuse an old component that had the same value (respectively, key) as the one we're now about to render.

Initsogar: Given that for this specific scenario, the `id` is the unique identifier for that option (which only can be 7, 30, 120 or 365 for now), I don't see how this could be an issue. What do you think?

client/components/backup-retention-management/constants.ts
Initsogar: Makes totally sense! We are using this constant only in `backup-retention-management/retention-options/retention-option-card.tsx` so I think we can just put it there where we need to render them.

client/components/backup-retention-management/retention-options/retention-options-control.tsx
tyxla: I don't see it problematic in this particular scenario, but it is a red flag because someone might borrow this code later and use it with a larger amount of (or infinite) options. It partially defies the idea of how React keys are supposed to serve.

That being said, feel free to add a prefix to the `key` so it's at least different from the value. Adding a comment that this is fine because we're working with a small set of option IDs could also help our future selves.

client/components/backup-retention-management/index.tsx
Initsogar: Double-checking noticed that we added this check because we were setting the default selection to 0. Fixing it to set a valid option as default.

client/components/backup-retention-management/index.tsx
Initsogar: Hmm, based on this, then we actually don't need a state at all.

We can replace the retentionOptionsCards state with:

On `constants.tsx`:
```javascript
export const RETENTION_OPTIONS = [ 7, 30, 120, 365 ] as RetentionPeriod[];
```

On `BackupRetentionManagement`:
```javascript
const retentionOptionsCards = RETENTION_OPTIONS.map( ( retentionDays ) => {
	return {
		id: retentionDays,
		spaceNeededInBytes: estimatedCurrentSiteSize * retentionDays,
		upgradeRequired: estimatedCurrentSiteSize * retentionDays > storageLimitBytes,
	};
} );
```

Way to simpler and we can get rid of unnecessary useEffect(s).

What do you think?

client/components/backup-retention-management/retention-options/retention-option-card.tsx
Initsogar: Done

client/components/backup-retention-management/retention-options/retention-options-control.tsx
Initsogar: Done!

client/components/backup-retention-management/constants.ts
Initsogar: Done!

client/components/backup-retention-management/index.tsx
tyxla: I'm really glad you got to this conclusion! Less state (and effects that maintain it) is always a good idea 👍

client/components/backup-retention-management/index.tsx
Initsogar: We just get rid of that useEffect with the mentioned approach [here](https://github.com/Automattic/wp-calypso/pull/73303#discussion_r1107153191), so marking this as solved.

client/components/backup-retention-management/index.tsx
Initsogar: Yeah, I got carried away by TS in this one. Just fixed it :)

Initsogar: You are right! Done!

Initsogar:
@tyxla Thanks for the detailed review. Great learning opportunity and ensuring code quality.
I think this is ready to re-review. Could you take another look?

client/components/backup-retention-management/index.tsx
Initsogar: It appears that just adding a valid option as default is not too straightforward and, for timely reasons, I think it would be better to address this improvement in another PR. I commit to it. Would it be okay?

tyxla: Sounds good, thanks 👍

client/components/backup-retention-management/index.tsx
```diff
@@ -65,53 +68,48 @@ const BackupRetentionManagement: FunctionComponent = () => {
...
-			currentRetentionPlan,
-			retentionSelected === RETENTION_OPTIONS.RETENTION_DAYS_120
-		),
-		[ RETENTION_OPTIONS.RETENTION_DAYS_365 ]: usePrepareRetentionOptions(
-			translate( '1 year' ),
-			RETENTION_OPTIONS.RETENTION_DAYS_365,
-			currentRetentionPlan,
-			retentionSelected === RETENTION_OPTIONS.RETENTION_DAYS_365
-		),
-	};
```
tyxla, line 98: 😍

tyxla (approved):
Nice work on the code, looks great 👍 🚀 

I haven't tested, so I'll leave the final approval to your team.

mavegaf (approved):

PR #73334, opened 2023-02-14
Virtual themes: Fix site preview for default variation

mmtr:
## Proposed Changes

Fixes an issue that was preventing the default style variation from being previewed after selecting a virtual theme

## Testing Instructions

- In `wordpress.com` (production) go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>&flags=virtual-themes/onboarding`
- Select a virtual theme to open the "site preview"
- While in the "site preview", select the default style variation
- Note how it is not previewed
- Now use the Calypso live link below and repeat the same steps
- Make sure the default style variation is previewed


dsas (approved):

mmtr:
I noted that, although the default style variation is eventually previewed, it takes a bit of time to load:

https://user-images.githubusercontent.com/1233880/219010531-12f0c500-ff0c-4ac9-a5a5-d947f27dc613.mov

The problem is not present somehow when using the discs on a non-virtual theme to preview a premium variation and then switching to the default one on the "site preview".

It only happens when selecting a virtual theme. It looks like the `onPreview` callback expects the first param to be a non-virtual theme, so I changed the approach in https://github.com/Automattic/wp-calypso/pull/73334/commits/0c831d285589f3ee502906fee2af1fb64eb71825 to fix that.

@dsas would you mind giving another look?

dsas:
This seems to have broken it for me, the initial preview is the default theme not the selected variation.


https://user-images.githubusercontent.com/93301/219018233-e683b7d9-e687-41c8-90ac-21ef408c2013.mp4

mmtr:
> This seems to have broken it for me, the initial preview is the default theme not the selected variation.

That's existing bug that already happens in `trunk` and also when selecting a variation from the discs on non-virtual themes. Apparently it's a race condition issue, since sometimes the selected variation is previewed, and sometimes it is not.

dsas:
Oh, I hadn't noticed that before but yes I eventually reproduced it on trunk.

It does seem more prevalent with this change though - I can reproduce it every time.

mmtr:
> I can reproduce it every time.

Definitely not my case 🤔. Are you sandboxed by any chance? I think that plays some role in the likelihood of the race condition issue to happen.

dsas (approved):
Not sandboxed now, but possibly was earlier. When I checkout and build your branch locally I can't reproduce it.

Happy to hit approve, but lets keep a very close eye post-deploy.

PR #65760, opened 2022-07-19
Sites Management Dashboard: Use fuzzy search instead of simple search

zaguiini:
#### Proposed Changes

After the introduction of the `useFuzzySearch` hook, it's time for us to use it.

#### Testing Instructions

Open SMD and try mistakenly searching for a blog. It should return the result, even with typos.

#### Pre-merge Checklist

Complete applicable items on this checklist **before** merging into trunk. Inapplicable items can be left unchecked.

Both the PR author and reviewer are responsible for ensuring the checklist is completed.

- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in [Simple](P9HQHe-k8-p2-p2), [Atomic](P9HQHe-jW-p2-p2), and [self-hosted Jetpack sites](PCYsg-g6b-p2-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Are we memoizing when appropriate (for expensive computations)? More info in [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md) and [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors)
- [x] Have we sent any new strings [for translation](PCYsg-1vr-p2-p2) ASAP?

Related to https://github.com/Automattic/wp-calypso/issues/65169.

PR #73335, opened 2023-02-14
Sites Management Page: Improve search input performance with lazyloading

danielbachhuber:
Related to p1675986927280589-slack-C0347E545HR

## Proposed Changes

Improves search input performance on Sites Management Page by lazyloading `<SiteItemThumbnail />` and `<SitesEllipsisMenu />`.

Before this change, the first keystroke takes ~150-190ms to complete:

image

With this change, the first keystroke takes ~100-130ms to complete:

image

We save on performance by deferring some rendering until the given site is in view.

Another approach I considered was to reintroduce `delaySearch` (https://github.com/Automattic/wp-calypso/pull/65760/files#r924953509). However, doing so introduces some delay between keystroke and render, which is an inferior user experience. Plus, it somewhat ignores the underlying problem.

Lastly, another alternative would be to only show 36 sites instead of 96. The first keystroke looks to be in the 100ms range:

image

```diff
diff --git a/client/sites-dashboard/components/sites-dashboard.tsx b/client/sites-dashboard/components/sites-dashboard.tsx
index ea0b4e4f95..71153aafdf 100644
--- a/client/sites-dashboard/components/sites-dashboard.tsx
+++ b/client/sites-dashboard/components/sites-dashboard.tsx
@@ -138,7 +138,7 @@ const ScrollButton = styled( Button, { shouldForwardProp: ( prop ) => prop !== '
 const SitesDashboardSitesList = createSitesListComponent();
 
 export function SitesDashboard( {
-	queryParams: { page = 1, perPage = 96, search, status = 'all' },
+	queryParams: { page = 1, perPage = 36, search, status = 'all' },
 }: SitesDashboardProps ) {
 	const { __, _n } = useI18n();
 	const { data: allSites = [], isLoading } = useSiteExcerptsQuery();

```

## Testing Instructions

1. Navigate to `/sites`.
2. Test a search in List view and verify that it feels snappier.
3. Test a search in Grid view and verify that it feels snappier.


client/sites-dashboard/components/sites-site-item-thumbnail.tsx
```diff
@@ -25,12 +25,33 @@ interface SiteItemThumbnailProps extends Omit< ComponentProps< typeof SiteThumbn
 	displayMode: SitesDisplayMode;
 	site: SiteExcerptData;
 	alt?: string;
+	inViewOnce?: boolean;
 }
 
-export const SiteItemThumbnail = ( { displayMode, site, ...props }: SiteItemThumbnailProps ) => {
+export const SiteItemThumbnail = ( {
+	displayMode,
+	inViewOnce,
```
p-jackson: `placeholder` might be a more descriptive prop name

client/sites-dashboard/components/sites-grid-tile.tsx
```diff
@@ -23,14 +23,16 @@ interface SitesGridTileProps {
...
 
-export const SitesGridTile = ( { leading, primary, secondary }: SitesGridTileProps ) => {
-	return (
-		<div className={ container }>
-			{ leading }
-			<div>
-				<div className={ primaryContainer }>{ primary }</div>
-				{ secondary }
+export const SitesGridTile = forwardRef(
+	( { leading, primary, secondary }: SitesGridTileProps, ref: React.Ref< HTMLDivElement > ) => {
```
p-jackson: Slightly less typing 🤷 Just a TypeScript personal preference thing. They mean the same things.

```suggestion
export const SitesGridTile = forwardRef< HTMLDivElement, SitesGridTileProps >(
	( { leading, primary, secondary }, ref ) => {
```

p-jackson (approved):
> However, doing so introduces some delay between keystroke and render, which is an inferior user experience. Plus, it somewhat ignores the underlying problem.

I agree. We'd be introducing a delay for users who have very few sites too, and that's no good.

The PR LGTM. But if we're going to only render some components when they're in view, why not go the full hog and virtualise this list?

### Check for regressions

Putting aside this PR initially, first I double checked that there hadn't been some obvious regression in the way we were using `React.memo`, which ensures we don't re-render rows during search, we just rearrange the existing DOM elements.
CleanShot 2023-02-15 at 10 58 54@2x

Nope no regression.

### Test environment

The dev mode of React is much slower so I only ran comparisons in production builds. So I compared `wordpress.com` to this PR running on `calypso.live`. No `localhost` here!

I also used a `?page-query=5000` query param to effectively disable pagination of the React components. Otherwise it confuses the results because React will have to render rows that come into view for the first time, but how many rows come into view will depend on the search term. I wanted to keep the comparison simple.

### Results

- Before PR keypress took **150ms**
- After PR keypress took **80ms**

It's still slow (equivalent to 12fps) but a definite improvement.

Given we're not unnecessarily re-rendering rows, we've made the React about as fast as we can make it. And half the time is just spent processing CSS.
CleanShot 2023-02-15 at 11 09 04@2x
There's just a lot of DOM elements, so reducing the amount we have to render at once is a good approach.

But rather than going piecemeal why not go the whole hog and only render rows/tiles when they're in view? This is a relatively common practice for long lists in React. Using [the `react-window` package](https://react-window.vercel.app/#/examples/list/fixed-size) is one way to only render rows that are in the viewport but also have the scrolling work correctly.

client/sites-dashboard/components/sites-grid-item.tsx
```diff
@@ -73,6 +74,9 @@ export const SitesGridItem = memo( ( { site }: SitesGridItemProps ) => {
 	const isP2Site = site.options?.is_wpforteams_site;
 	const translatedStatus = useSiteLaunchStatusLabel( site );
 
+	const [ inViewOnce, setInViewOnce ] = useState( false );
+	const ref = useInView< HTMLDivElement >( () => setInViewOnce( true ) );
```
p-jackson, line 82: Using the `inViewOnce` trick does speed things up in practice, but I'm not really sure why we'd do this over just only rendering it when it was `inView` i.e. removing the DOM elements when they're no longer in view. Since reducing the number of DOM elements is what's giving us the perf boost.

danielbachhuber:
Excellent code review, @p-jackson 😍 Thank you!

I'll dive into this some more tomorrow, and wanted to quickly respond to this:

> But rather than going piecemeal why not go the whole hog and only render rows/tiles when they're in view? This is a relatively common practice for long lists in React. Using [the react-window package](https://react-window.vercel.app/#/examples/list/fixed-size) is one way to only render rows that are in the viewport but also have the scrolling work correctly.

I considered this. However, I find virtual lists annoying because they break in-browser search (cmd + F). The element has to be at least partially rendered in order to search the page for it.

> And half the time is just spent processing CSS.

Ah, good find!

p-jackson:
> they break in-browser search (cmd + F)

Yeah that's a really good point.

I'd be open to reducing the default page size too. Searching works well.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
danielbachhuber: Good suggestion, b7e437788f

PR #73336, opened 2023-02-14
Site Selector: Remove search debounce for better UX

danielbachhuber:
Related https://github.com/Automattic/wp-calypso/pull/73335

## Proposed Changes

Removes the debounce from Site Selector search so new results appear immediately, instead of after a delay.

## Testing Instructions

1. Navigate to My Home and open up the Site Selector.
2. Enter a search in the Site Selector and observe the performance boost.


JessBoctor:
👋 I tested and really like this change! Feels much more responsive.

LGTM 👍

wojtekn (approved):
Change looks good to me, however, I couldn't see any performance difference - it was quick in both cases. I looked at the code and it looks that `delayTimeout` is set to 300ms as a default value.

PR #73367, opened 2023-02-15
Tooltip alignment polishing for the 2023 plan grid.

southp:
Related to Automattic/martech#1485

## Proposed Changes

This PR addresses the calypso part of tooltip alignment issue reported in Automattic/martech#1485 by @lucasmendes-design 

First, the space between the hovered item and the tooltip should be updated to 8px. Currently it is simply using the default `top: -10px` defined here: https://github.com/Automattic/wp-calypso/blob/trunk/packages/components/src/popover/style.scss#L113. Thie PR overrides it into 8px. I don't know how to capture a screenshot and mark the pixel distance for this, so tips are welcome :) 

Second, the tooltip is not centerly-aligned to the logo. It is due to that the hovering element is the encapsulated `div` instead of the logo image itself. This PR rearranges the nesting and extracts that as `<ServiceLogo>` component to fix the issue:

Before:
image

After:
image

## Testing Instructions

1. Go to /start/plans and confirm the above changes hold true and not breaking anything else.
2. Do the same to /plans while logged-in.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/index.tsx
+++ b/client/my-sites/plan-features-2023-grid/index.tsx
@@ -26,7 +26,7 @@ import formatCurrency from '@automattic/format-currency';
 import { MinimalRequestCartProduct } from '@automattic/shopping-cart';
 import { Button } from '@wordpress/components';
 import classNames from 'classnames';
-import { localize, LocalizeProps } from 'i18n-calypso';
+import { localize, LocalizeProps, TranslateResult } from 'i18n-calypso';
 import { last } from 'lodash';
 import page from 'page';
 import { Component, createRef } from 'react';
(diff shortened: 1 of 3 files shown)


claudiucelfilip:
The tooltips look well-aligned in all cases except on `/start/plans` in the Comparison grid:
Screenshot 2023-02-15 at 14 03 16

southp:
@claudiucelfilip Good catch. However, it's the `autoPosition` behavior from the underlying `<Popover />` component. The layout logic of the tooltip component when `position` prop is `top` is that it will attempt to center itself on top of the encapsulated element. When the space is not enough for it to do so and `autoPosition` is `true`, it will automatically adjust the position(see https://github.com/Automattic/wp-calypso/blob/trunk/packages/components/src/popover/util.js#L63), and that's why some of the tooltips are not centerly aligned here: some have the space, some are "auto-positioned".

We can turn off `autoPosistion`, but I don't think it looks any better:

https://user-images.githubusercontent.com/1842898/219289283-332338f7-cc3c-45f7-8ced-f60440ea844d.mov

It also causes an unpleasant jump when the tooltip is near the top:

https://user-images.githubusercontent.com/1842898/219289436-1f9bca35-011a-4901-b168-05f9f5c0c018.mov

Since we'd likely have to fix from the underlying component level, I'd suggest us to do it as a follow-up. Does that make sense to you?

ddc22 (approved):
Thanks for the change James, LGTM.
We can explore the cosmetic change proposed by @claudiucelfilip later.

southp:
Upon deploying, I've documented the follow-up as Automattic/martech#1534.

Issue #73223, opened 2023-02-10
[WooExpress] Task: Ensure we display USD prices with US prefix outside of the US on eCommerce trial Plans page


daledupreez:
### Details

This issue was originally identified in https://github.com/Automattic/wp-calypso/pull/72797#discussion_r1101420149, where the eCommerce trial Plans page always shows USD prices with the currency symbol `$`. However, there are a number of countries where `$` is used as the local currency symbol, so we show USD prices with the symbol `US$` whenever we detect that users are outside of the US.

We need to update the new page to show the price appropriately.

### Checklist

- [ ] Show `US$` for prices on the eCommerce trials Plans page for USD prices outside the US

### Related

#72797

markbiek:
@daledupreez I did a quick test where I manually passed a different locale (`es`) to the call to [getCurrencyObject](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plans/ecommerce-trial/index.tsx#L43) on the plans page and it seems to be displaying the prices correctly.

![image](https://user-images.githubusercontent.com/917632/218869423-d48fe3f2-4d81-406c-b70c-c2937c25c9bf.png)

Is there a recommended/reliable way of checking if a user is outside the US other than using the browser locale settings?

daledupreez:
I should have provided way more context here, as there was some important discussions on how this should be done in the last month:
* pbOQVh-2WH-p2 discusses how we've updated Calypso to handle price formatting in a much more robust and locale-aware way
* #71387 is the PR that implemented the core API change for this

daledupreez:
I've put up a PR for this in #73372, but that doesn't always add the `US` prefix -- I've asked on the original PR about the expected behaviour for that.

markbiek:
> I've put up a PR for this in #73372, but that doesn't always add the `US` prefix -- I've asked on the original PR about the expected behaviour for that.

@daledupreez Can you drop a link to your question? If it's on either #73372 or #72797, I'm not seeing it.

daledupreez:
Sorry for not including the context, @markbiek!
My questions on the original issue are in [this comment](https://github.com/Automattic/wp-calypso/pull/71387#issuecomment-1430869326) and [this comment](https://github.com/Automattic/wp-calypso/pull/71387#issuecomment-1431105505).

markbiek:
@daledupreez I'm not sure what the next steps are.

[...]

PR #73372, opened 2023-02-15
Use centralised formatCurrency() function for ecommerce trial plans

daledupreez:
Related to #71387, #73223

## Proposed Changes

* This change updates the recently added eCommerce trial plans page to use the general-purpose `formatCurrency()` function to format prices in a locale-aware fashion - we were previously using the `getCurrencyObject()` function to find the currency symbol, but were manually setting up the currency location and symbol
* I also removed a `width` restriction on a button as that was going to cause trouble for some languages

At an implementation level, this _should_ ensure that we use `US$` as a prefix for USD in various cases, but I think there are cases where that is not the case, which I flagged on the original PR [here](https://github.com/Automattic/wp-calypso/pull/71387#issuecomment-1431105505).

Given that I am still seeing `$` for `USD` with my locale set to `en`, I am not sure this is a full fix for the original issue, but it does mean we will handle other currencies and locales correctly even if we're not handling USD as I think we should.

## Testing Instructions

There should be no functional change except when you have a currency and language/locale pairing that would switch the display order.
You can test this as follows:
* Run this branch locally or via the calypso.live branch
* Update your WordPress.com language to Italian (or another European language that defaults to showing the Euro symbol _after_ the value(
* Update your preferred currency to `EUR` via SA
* Navigate to _Upgrades_ -> _Plans_ for a site on the free eCommerce trial
* Verify that the price has the Euro symbol after the currency
* Verify that the CTA button doesn't wrap

### Before

Screenshot 2023-02-15 at 13 04 46

### After

Screenshot 2023-02-15 at 13 03 22

## Pre-merge Checklist

- [N/A] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [N/A] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [N/A] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/index.tsx
@@ -5,7 +5,7 @@ import {
 	PLAN_ECOMMERCE_MONTHLY,
 } from '@automattic/calypso-products';
 import { Button, Card } from '@automattic/components';
-import { getCurrencyObject } from '@automattic/format-currency';
+import { formatCurrency } from '@automattic/format-currency';
 import { useTranslate } from 'i18n-calypso';
 import page from 'page';
 import { useCallback, useMemo } from 'react';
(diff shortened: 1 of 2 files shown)


nelsonec87 (approved):
LGTM, assuming the before/after screenshots are swapped!
The failing test seems to be unrelated.

PR #73381, opened 2023-02-15
WooExpress: Trial flow checks that WooCommerce is installed before completing

### LINKED ISSUE #73307: [WooExpress] Calypso home is loaded after completing signup
### Quick summary

After completing the `/setup/wooexpress` signup flow, we end up loading the Calypso home page at `/home/:siteSlug`. We should be redirected to the WooCommerce home page, which suggests that WooCommerce may not be fully installed by the time we wrap up the signup flow.

### Steps to reproduce

1. Visit `/setup/wooexpress`
2. Wait it out - you may or may not be redirected to WooCommerce home after the flow completes

I should have been redirected to the WooCommerce task list in the WooCommerce home page.

### What actually happened

The page did not redirect and I stayed on the Calypso home page.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

Atomic

### Logs or notes

_No response_

markbiek:
Fixes #73307

## Proposed Changes

Previously, we redirected to `/home` as soon as the Atomic transfer was complete. This lead to issues because, sometimes, WooCommerce wasn't finished installed before the redirect so we'd end up on the main dashboard instead of the WooCommerce launch checklist.

This adds a new step where we wait for WooCommerce to finish installing before we redirect.

## Testing Instructions

Prior to applying this patch, visit http://calypso.localhost:3000/setup/wooexpress/. In some cases, you may end up on `/home` instead of `/wp-admin/admin.php?page=wc-admin`. It seems slightly more likely to happen if you're sandboxed but you may have try multiple times to reproduce the behavior.

After applying the patch, you should always end up at `/wp-admin/admin.php?page=wc-admin`.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
@@ -55,11 +55,7 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 
 	const username = useSelector( ( state ) => getCurrentUserName( state ) );
 
-	const { setPendingAction, setPluginsToVerify } = useDispatch( ONBOARD_STORE );
-
-	if ( isWooExpressFlow( flow ) ) {
-		setPluginsToVerify( [ 'woocommerce' ] );
-	}
+	const { setPendingAction } = useDispatch( ONBOARD_STORE );
 
 	let theme: string;
 	if ( isMigrationFlow( flow ) || isCopySiteFlow( flow ) ) {
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
@@ -55,7 +55,11 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 
 	const username = useSelector( ( state ) => getCurrentUserName( state ) );
 
-	const { setPendingAction } = useDispatch( ONBOARD_STORE );
+	const { setPendingAction, setPluginsToVerify } = useDispatch( ONBOARD_STORE );
+
+	if ( isWooExpressFlow( flow ) ) {
+		setPluginsToVerify( [ 'woocommerce' ] );
+	}
 
 	let theme: string;
 	if ( isMigrationFlow( flow ) || isCopySiteFlow( flow ) ) {
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
@@ -1,0 +1,120 @@
+import config from '@automattic/calypso-config';
+import { isWooExpressFlow } from '@automattic/onboarding';
+import { useDispatch } from '@wordpress/data';
+import { useEffect } from 'react';
+import wpcomRequest from 'wpcom-proxy-request';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
+import { logToLogstash } from 'calypso/lib/logstash';
+import { ONBOARD_STORE } from '../../../../stores';
+import type { Step } from '../../types';
+
+export interface Plugin {
+	slug: string;
+}
+
+export interface PluginsResponse {
+	plugins: Plugin[];
+}
+
+export interface FailureInfo {
+	type: string;
+	code: number | string;
+	error: string;
+}
+
+export const installedStates = {
+	PENDING: 'pending',
+	INSTALLED: 'installed',
+	ERROR: 'error',
+} as const;
+
+const wait = ( ms: number ) => new Promise( ( res ) => setTimeout( res, ms ) );
+
+const WaitForPluginInstall: Step = function WaitForAtomic( { navigation, data, flow } ) {
+	const { submit } = navigation;
+	const { setPendingAction } = useDispatch( ONBOARD_STORE );
+
+	const siteId = data?.siteId;
+	const siteSlug = data?.siteSlug;
+
+	const handlePluginCheckFailure = ( failureInfo: FailureInfo ) => {
+		recordTracksEvent( 'calypso_stepper_plugin_check_error', {
+			action: failureInfo.type,
+			site: siteId,
+			code: failureInfo.code,
+			error: failureInfo.error,
+		} );
+
+		logToLogstash( {
+			feature: 'calypso_client',
+			message: failureInfo.error,
+			severity: config( 'env_id' ) === 'production' ? 'error' : 'debug',
+			blog_id: siteId,
+			properties: {
+				env: config( 'env_id' ),
+				type: 'calypso_stepper_plugin_check_error',
+				action: failureInfo.type,
+				site: siteId,
+				code: failureInfo.code,
+			},
+		} );
+	};
+
+	useEffect( () => {
+		if ( ! siteId || ! siteSlug ) {
+			return;
+		}
+
+		setPendingAction( async () => {
+			const startTime = new Date().getTime();
+			const totalTimeout = 1000 * 300;
+			const maxFinishTime = startTime + totalTimeout;
+
+			// Poll for transfer status
+			let stopPollingPlugins = false;
+
+			while ( ! stopPollingPlugins ) {
+				await wait( 3000 );
+
+				try {
+					const response: PluginsResponse = await wpcomRequest( {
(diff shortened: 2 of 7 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+		setPendingAction( async () => {
+			const startTime = new Date().getTime();
+			const totalTimeout = 1000 * 300;
+			const maxFinishTime = startTime + totalTimeout;
+
+			// Poll for transfer status
+			let stopPollingPlugins = false;
+
+			while ( ! stopPollingPlugins ) {
+				await wait( 3000 );
```
daledupreez: I think we should make this interval shorter so we don't introduce unnecessary delays of up to 3 seconds. Maybe 0.5 to 1 second intervals could work?

client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+					// For WooExpress, we need to check for the WooCommerce plugin.
+					if ( response?.plugins && isWooExpressFlow( flow ) ) {
+						const woocommercePlugin = response?.plugins.find(
+							( plugin: { slug: string } ) => plugin.slug === 'woocommerce'
+						);
+
+						if ( woocommercePlugin ) {
+							stopPollingPlugins = true;
+						}
+					}
```
daledupreez: This feels icky to me. Can we push the set of desired plugin slugs into the `ONBOARD_STORE` via a dedicated storage location/key? We could persist the set of plugins at the start of a flow (or during any step before this one), and simply have this step verify that a list of plugin slugs have been installed on the site.

client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+					throw new Error( 'plugin check timeout' );
+				}
+			}
+
+			return { pluginsInstalled: true, siteSlug, siteId };
+		} );
+
+		submit?.();
+
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
daledupreez, line 111: This is a red flag to me: why are we disabling the check? At the very least, please add a comment to note why.

daledupreez (commented):
This isn't working for me. All the API calls to `/sites/:siteSlug/lpugins` are returning a 403 status with the following error content:
```
{"error":"unauthorized","message":"This user is not authorized to activate_plugins on this blog."}
```
This is a weird catch-22, as we aren't allowing _anyone_ plugin activate or install permissions on the site. :-/

client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
markbiek: Most steps are using this to one degree or another, but I'll add a comment for clarity.

image

client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
```diff
@@ -55,7 +55,11 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 
 	const username = useSelector( ( state ) => getCurrentUserName( state ) );
 
-	const { setPendingAction } = useDispatch( ONBOARD_STORE );
+	const { setPendingAction, setPluginsToVerify } = useDispatch( ONBOARD_STORE );
+
+	if ( isWooExpressFlow( flow ) ) {
+		setPluginsToVerify( [ 'woocommerce' ] );
```
daledupreez: Should we not run this logic from the flow itself? It might make sense to do it just before we redirect to the `waitForPluginInstall` step.

Having this logic embedded in the step doesn't feel right to me at all!

markbiek: I was going from your earlier comment:

> We could persist the set of plugins at the start of a flow _(or during any step before this one)_, and simply have this step verify that a list of plugin slugs have been installed on the site.

and so I put it in the first step of the flow. But I can definitely move it to the flow itself if you think that makes more sense.

daledupreez: From my side it's more about the separation of concerns than about sequencing, if that makes sense. Having the _step_ responsible for logic that is flow-specific feels like it could easily be fragile. I know we have lots of other logic doing that already, but I think we should aim to minimise that as much as possible.

markbiek: That all makes total sense. I'll get that updated.

client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+	}, [ siteId ] );
```
daledupreez: Given that the upfront check in the code also depends on `siteSlug`, it may be worth adding it to the deps. Admittedly, one won't change without the other, so this is not blocking.

client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-plugin-install/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+				} catch ( err ) {
+					// Ignore errors. It's normal to get errors the first couple of times we poll. The timeout will eventually catch it if the failures continue.
+				}
+
+				if ( maxFinishTime < new Date().getTime() ) {
+					handlePluginCheckFailure( {
+						type: 'plugin_check_timeout',
+						error: 'plugin check took too long',
+						code: 'plugin_check_timeout',
+					} );
```
daledupreez: It might be useful to log the amount of actual elapsed time, or at least the value of `totalTimeout`.

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -74,7 +76,8 @@ const wooexpress: Flow = {
 		const intent = useSelect( ( select ) => select( ONBOARD_STORE ).getIntent() );
 		const siteSlugParam = useSiteSlugParam();
 
-		const { setStepProgress } = useDispatch( ONBOARD_STORE );
+		const { setStepProgress, setPluginsToVerify } = useDispatch( ONBOARD_STORE );
+		setPluginsToVerify( [ 'woocommerce' ] );
```
daledupreez, line 80: Not something for this PR (and maybe something we never need), but we may eventually want to have this data keyed by the flow name. I am not sure if we clear the data across flows, but that might be a situation where we could have subtle data pollution or bugs.
```suggestion
		setPluginsToVerify( flowName, [ 'woocommerce' ] );
```

packages/data-stores/src/onboard/reducer.ts
```diff
@@ -483,6 +483,16 @@ const isMigrateFromWp: Reducer< boolean, OnboardAction > = ( state = false, acti
 	return state;
 };
 
+const pluginsToVerify: Reducer< string[] | undefined, OnboardAction > = ( state, action ) => {
+	if ( action.type === 'SET_PLUGIN_SLUGS_TO_VERIFY' ) {
+		return action.pluginSlugs;
```
daledupreez, line 488: Is it worth enforcing that this has type `string[]`? That way we can have `getPluginsToVerify()` be strictly typed to return `undefined | string[]`. (Definitely not blocking, especially seeing as that doesn't seem to be an existing pattern.)

daledupreez (approved):
LGTM! The flow is pretty slow for me, especially the API call to add the eCommerce trial (which took ~43 seconds!), but the transfer and wait for the plugin to be installed also add up.

We have #73419 open to debug this, but I wanted to note the delays given that they are now longer, even if it is for good reason.

markbiek:
> The flow is pretty slow for me, especially the API call to add the eCommerce trial (which took ~43 seconds!), but the transfer and wait for the plugin to be installed also add up.

It's funny, for me the `assignTrial` step is usually close to instant and the atomic transfer takes forever. 

But agreed that we definitely need to do some work to speed things up!

nelsonec87 (approved):
LGTM!
I was able to reproduce the problem on `trunk` and, after this patch, created 5 new sites successfully.

PR #73387, opened 2023-02-15
Offer Jetpack Complete, CTA button component

TheSteveK:
Per 1203759331596262-as-1203791007934538/f this PR adds a new button component to the Offer Complete card

It adds a call to action button that will redirect to the cart with Complete added.  The other sends the user to their site's wp-admin page

Screenshot 2023-02-15 at 12 04 24 PM

I also added logic for generic redirects if there is no available site slug.  I'm not sure if that will ever be possible, but it seems a good idea to include it. 

Tracks events are not yet integrated.  I figured I would do that in another PR to sort of keep the work compartmentalized in my mind

## Testing Instructions

Have a hand-dandy JN site available that's connected to Jetpack, but has no upgrades.

Head over to the calypso.live link below (or roll your own with yarn-start)

Append /jetpack/connect/plans/complete/[site] to the base URL

You should see the two buttons on the card.  Click each one and verify the get complete drops you in the cart with the complete plan already selected.  The other one should just send you over to wp-admin.

## Pre-merge Checklist

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203791007934538


client/my-sites/plans/jetpack-plans/jetpack-complete-page/style.scss
```diff
@@ -66,3 +66,44 @@
 		}
 	}
 }
+
+.jetpack-complete-page__cta-buttons {
```
andrii-lysenko: Maybe we could place this into separate `styles.scss` in the same folder as `buttons` component.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons.tsx
```diff
@@ -0,0 +1,44 @@
+//imports
```
andrii-lysenko: Maybe let's remove this unused import

client/my-sites/plans/jetpack-plans/jetpack-complete-page/style.scss
TheSteveK: I was debating whether or not to do that.  I moved them to a subfolder in components to keep everything tidy

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
```diff
@@ -0,0 +1,45 @@
...
+		return `https://wordpress.com/checkout/${ planToOffer }`;
+	};
+
+	//Start for free with no site, just redirect to jetpack.com
+	const stayFreeURL = () => {
+		if ( siteSlug ) {
+			return `https://${ siteSlug }/wp-admin`;
+		}
+		return `https://jetpack.com/`;
+	};
```
mattgawarecki: It looks like we're trying to redirect folks to their site's main wp-admin page, but `https://${ siteSlug }/wp-admin` doesn't always get us there, for the following reasons:
* `siteSlug` doesn't always format as a valid URL. For example, sites that are part of a multi-site network often have slugs that look like `mymultisite.example::site1` when the actual site URL is `https://mymultisite.example/site1`.
* Not all Jetpack sites use HTTPS. To know which is the proper protocol to use, we have to rely on what we know about a site from its connection to Jetpack.
 
To fix this, we can pull from prior art: [components/jetpack/sidebar/index.jsx:72](https://github.com/Automattic/wp-calypso/blob/f2e6f88ae65c8571bbac21b284c29b48ec987598/client/components/jetpack/sidebar/index.jsx#L72) reliably sends people directly to their site's Jetpack dashboard by extending the `getSiteAdminUrl` selector. _(Aside: subjectively, I also believe linking to the Jetpack dashboard might be more appropriate than the wp-admin root, since in our case we're working with Jetpack customers.)_

To avoid duplicating code, perhaps we can extract the `getJetpackAdminUrl` function into a new file, and reference it here? It could live in `state/sites/selectors/get-jetpack-admin-url.ts`, since two very separate areas of Jetpack Cloud would now depend on it.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
```diff
@@ -0,0 +1,45 @@
...
+	const planToOffer = PLAN_JETPACK_COMPLETE;
+	const siteSlug = useSelector( getSelectedSiteSlug );
+
+	// handle no site selected
+	const getCompleteURL = () => {
+		if ( siteSlug ) {
+			return `/checkout/${ siteSlug }/${ planToOffer }`;
+		}
+		return `https://wordpress.com/checkout/${ planToOffer }`;
+	};
```
mattgawarecki: Depending on whether or not we have a site slug, someone testing in an environment outside of production or staging (e.g., `http://jetpack.cloud.localhost:3000`) could be sent to either of two different domains, which could be a big source of frustration if they don't notice the difference.

If we're satisfied with excluding the domain for one case, let's do it for both:

```suggestion
	const getCompleteURL = () => {
		if ( siteSlug ) {
			return `/checkout/${ siteSlug }/${ planToOffer }`;
		}
		return `/checkout/${ planToOffer }`;
	};
```

---

Second, lower priority: using ternaries, we could make this a variable instead of a function 🙂 Using exactly what you've written, here's an example:

```js
const completeURL = siteSlug
	? `/checkout/${ siteSlug }/${ planToOffer }`
	: `https://wordpress.com/checkout/${ planToOffer }`;
```

---

Third, since you've mentioned that you're studying hooks like `useMemo`: it wouldn't help much in this case since the logic is very inexpensive and the component itself doesn't have a complex enough state to cause it to re-render often, but a memoized version of exactly the code you have here might look like so:

```js
const completeURL = useMemo( () => {
	if ( siteSlug ) {
			return `/checkout/${ siteSlug }/${ planToOffer }`;
		}
		return `https://wordpress.com/checkout/${ planToOffer }`;
}, [ siteSlug, planToOffer ] );
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useTranslate } from 'i18n-calypso';
+import { useSelector } from 'react-redux';
+import Button from 'calypso/components/forms/form-button';
+import getSelectedSiteSlug from 'calypso/state/ui/selectors/get-selected-site-slug';
+
+import './style.scss';
+
+const CtaButtons = () => {
+	const translate = useTranslate();
+	const planToOffer = PLAN_JETPACK_COMPLETE;
```
mattgawarecki: Normally I'm a big fan of variables that clarify other variables, like it seems you've done here. However, since this component is only ever referenced in relation to Jetpack Complete, I think it's safe to remove this variable and replace all its references with `PLAN_JETPACK_COMPLETE` directly.

For example, instead of `/checkout/${ siteSlug }/${ planToOffer }`, I think it's perfectly readable to say `/checkout/${ siteSlug }/${ PLAN_JETPACK_COMPLETE }`.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
```diff
@@ -0,0 +1,45 @@
+import { PLAN_JETPACK_COMPLETE } from '@automattic/calypso-products';
```
mattgawarecki: Conventionally, the "main" component file inside a directory is named `index.tsx` so that it can be referenced by the folder name.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/index.tsx
```diff
@@ -12,6 +12,7 @@ import { JPC_PATH_PLANS } from 'calypso/jetpack-connect/constants';
 import { successNotice } from 'calypso/state/notices/actions';
 import getSelectedSiteId from 'calypso/state/ui/selectors/get-selected-site-id';
 import { QueryArgs, Duration } from '../types';
+import CtaButtons from './components/cta-buttons/cta-buttons';
```
mattgawarecki: Paired with my other comment about conventional file names, the import statement would change to this:

```suggestion
import CtaButtons from './components/cta-buttons';
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
+.jetpack-complete-page__cta-buttons {
```
mattgawarecki, line 1: Note: I'm holding off on a review of the styles here, until the functionality is fully tested.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/index.tsx
```diff
@@ -67,6 +68,7 @@ const JetpackCompletePage: React.FC< Props > = ( {
 							all CRM extensions, and extra storage for backups and video.
 						</p>
 					</>
```
mattgawarecki, line 88: Not super related to your PR itself, but we have an unnecessary React Fragment here. We can simplify the markup by removing it, as follows:

```tsx
<JetpackRnaDialogCard ...stuff>
  <h1 ...stuff>
    ...
  </h1>
  <p ...stuff>
    ...
  </p>
  <CtaButtons />
</JetpackRnaDialogCard>
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
andrii-lysenko: Alternatively, we could probably also use [buildCheckoutUrl](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plans/jetpack-plans/get-purchase-url-callback.ts#L27) function. Though, I'm also ok with manual links.

mattgawarecki: @andrii-lysenko That's a great point -- I wasn't aware of that function, but re-using well-worn code gives us an extra level of confidence that what we're writing will be correct and require less maintenance.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
TheSteveK: Hey @mattgawarecki I just pushed a commit with my first whack at creating the component, but I'm confused about the implementation.  Unlike the other components, which just use state, this one won't work with just state like the others such as getSiteSlug do, I have to explicitly use a SiteId and pass that along from the buttons component via a callback.

mattgawarecki: What you did to adapt `getJetpackAdminUrl` -- passing an anonymous function into `useSelector` -- was the correct move. `useSelector` expects a function that accepts only `state` and passes back a result; so, to satisfy this condition, we have to reference data from outside the function (i.e., what we call a [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures)).

TheSteveK: Oh, because the components that the getJeptackAdminUrl relies on, like getSiteAdminUrl also expect a siteId to be passed along with state 💡

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
```diff
@@ -2,30 +2,22 @@ import { PLAN_JETPACK_COMPLETE } from '@automattic/calypso-products';
...
-	};
-
-	//Start for free with no site, just redirect to jetpack.com
-	const stayFreeURL = () => {
-		if ( siteSlug ) {
-			return `https://${ siteSlug }/wp-admin`;
-		}
-		return `https://jetpack.com/`;
-	};
+	const stayFreeURL = siteSlug ? adminURL : `https://jetpack.com/`;
```
mattgawarecki: Since `adminURL` only depends on `siteSlug` being present, and is null/undefined otherwise, we can tighten up our selector using the null-coalesce (`??`) operator:

```ts
	// adminURL variable is no longer needed

	const stayFreeURL = useSelector( ( state ) =>
		getJetpackAdminUrl( state, siteId )
		?? 'https://jetpack.com/'
	);
	const getCompleteURL = siteSlug
		? `/checkout/${ siteSlug }/${ planToOffer }`
		: `/checkout/${ planToOffer }`;
```

client/state/sites/selectors/get-jetpack-admin-url.js
```diff
@@ -0,0 +1,15 @@
+import { format as formatUrl, getUrlParts, getUrlFromParts } from '@automattic/calypso-url';
+import getSiteAdminPage from './get-site-admin-page';
+import getSiteAdminUrl from './get-site-admin-url';
+
+export default function getJetpackAdminUrl( state, siteId ) {
+	const siteAdminUrl = getSiteAdminUrl( state, siteId );
+	if ( null === siteAdminUrl ) {
```
mattgawarecki: I know this wasn't your doing, but in the name of leaving it better than we found it, let's re-write this conditional so it's not Yoda-style:

```suggestion
	if ( siteAdminUrl === null ) {
```

TheSteveK: I see you and I raise you` if ( ! siteAdminUrl )`

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
TheSteveK: Question on the finer points of buildCheckoutUrl.  Even though it appears to be able to handle if you don't pass a site to it, siteId is typed as string.  I'm not sure what to pass if there's no ID to make the linter not whine at me for passing in something that could have a null value.

andrii-lysenko: @TheSteveK good question. So it could be either `url = buildCheckoutURL( siteSlug || '', ...)` or `url = siteSlug ? buildCheckoutURL( siteSlug,) : ''`. It should work as we should have siteId always here. However, I see also an option to modify the `buildCheckoutURL` so it could work without necessarily passing `siteSlug`.

client/state/sites/selectors/get-jetpack-admin-url.js
andrii-lysenko: @TheSteveK it's better to be precise, so I'd stick with `if ( siteAdminUrl === null )` because function returns `string | null`. Not to this example, but there are situations where `if ( ! siteAdminUrl )` would give you the wrong result (in case of numbers and 0 for example).

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/cta-buttons.tsx
TheSteveK: Yeah I tried to test the page without a site, but it doesn't work at all, so maybe this page never gets accessed if there's not a site in the URL?

For now, maybe just use a manual link if there somehow is no available site?

```
href={
					siteSlug
						? buildCheckoutURL( siteSlug, PLAN_JETPACK_COMPLETE )
						: `/checkout/${ PLAN_JETPACK_COMPLETE }`
				}
```

client/state/sites/selectors/get-jetpack-admin-url.js
TheSteveK: Ok, I changed it back, but it's worth mentioning most of the other selectors use ! (typically !siteId)

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+		( state ) => getJetpackAdminUrl( state, siteId ) ?? 'https://jetpack.com/'
+	);
+
+	return (
+		<div className="jetpack-complete-page__cta-buttons">
+			<Button
+				className="jetpack-complete-page__get-complete-button"
+				primary
+				href={
+					siteSlug
```
andrii-lysenko: Maybe we could still use a `const` here, to have it a bit cleaner. You can define it above like:
`const checkoutUrl = siteSlug ? buildCheckoutURL...`

mattgawarecki: +1 to this 🙂

client/state/sites/selectors/get-jetpack-admin-url.js
```diff
@@ -0,0 +1,15 @@
+import { format as formatUrl, getUrlParts, getUrlFromParts } from '@automattic/calypso-url';
+import getSiteAdminPage from './get-site-admin-page';
+import getSiteAdminUrl from './get-site-admin-url';
+
+export default function getJetpackAdminUrl( state, siteId ) {
```
mattgawarecki, line 5: Note for a separate PR, or an extra-small maintenance task: return to `components/jetpack/sidebar` to remove its implementation and replace it with this shared one.

TheSteveK: I added a task to our board

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+	const translate = useTranslate();
+	const siteSlug = useSelector( getSelectedSiteSlug );
+	const siteId = useSelector( getSelectedSiteId );
+
+	const checkoutURL = siteSlug
+		? buildCheckoutURL( siteSlug, PLAN_JETPACK_COMPLETE )
+		: `/checkout/${ PLAN_JETPACK_COMPLETE }`;
+
+	const stayFreeURL = useSelector(
+		( state ) => getJetpackAdminUrl( state, siteId ) ?? 'https://jetpack.com/'
```
mattgawarecki: Instead of Jetpack.com, I just realized there's an appropriate page for introducing folks to Free: `/pricing/jetpack-free/welcome`.

Benefits of using this instead:
- Is more relevant to people seeking information about Free
- Has specific instructions for situations where we can't detect a site ID (i.e., a connection)
- Keeps people in the same environment, even if they're testing

Drawbacks:
- This page is less polished than the Jetpack.com root, IMO

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	width: 100%;
+	& > * {
+		flex: 1;
+	}
+	@include breakpoint-deprecated( "<960px" ) {
+		flex-direction: column;
+	}
```
mattgawarecki: Conventionally we try to do "mobile first," so larger screens are the ones that end up with breakpoints. The equivalent would be something like this:

```scss
flex-direction: column;

@include breakpoint-deprecated( '>960px' ) {
	flex-direction: row;
}
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+
+	.jetpack-complete-page__get-complete-button {
+		@include breakpoint-deprecated( ">960px" ) {
+			margin-right: 2em;
+		}
```
mattgawarecki: Since we're using `inline-flex`, there's a more suitable property we can use to keep the buttons separated: `gap`. I see in the original designs there's a spacing of 24px between the two (we generally avoid `em`, for... reasons), so something like this:

```scss
// starting in the scope of __cta-buttons
@include breakpoint-deprecated( '>960px' ) {
	gap: 24px;
}

.jetpack-complete-page__get-complete-button {
	// remove breakpoint rule for margin-right
	// ...
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
mattgawarecki, line 1: According to the designs, our buttons have the following specs:

- `border-radius` of 4px
- `font-weight` of 600

We need new rules for these inside the `__cta-buttons` scope, probably targeted at `.button` or `.form-button`.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+			&:hover,
+			&:active {
+				background-color: var(--studio-gray-90);
+			}
+		}
+	}
+	.jetpack-complete-page__start-free-button {
+		@include breakpoint-deprecated( "<960px" ) {
+			margin-top: 1em;
+		}
```
mattgawarecki: 1. For i18n purposes, it's better to use `margin-block-start` instead of `margin-top`.
2. The original designs show a top margin of 24px.
3. This is also another instance where the margin would be present by default, with larger screens getting a breakpoint. We can reset the margin to whatever the default value is with `unset`.

```scss
margin-block-start: 24px;

@include breakpoint-deprecated( '>960px' ) {
	margin-block-start: unset;
}
```

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+		&.form-button {
+			color: var(--studio-black);
+			background-color: var(--studio-white);
+			border: 2px solid var(--studio-black);
```
mattgawarecki: Original designs have the button border at 1px thick.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
+.jetpack-complete-page__cta-buttons {
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	width: 100%;
+	& > * {
+		flex: 1;
+	}
```
mattgawarecki: Design mock-ups don't seem to indicate that buttons should grow, which is what this rule does. (Also, it's quite vague. If we kept it, I'd recommend simply `.button` instead of `> *`; but that's neither here or there.)

Instead, let's remove the `flex` rule here entirely.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
+.jetpack-complete-page__cta-buttons {
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
```
mattgawarecki: Design mock-ups have these buttons as left-aligned, so we can safely remove `justify-content: center;`.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+		@include breakpoint-deprecated( ">960px" ) {
+			margin-right: 2em;
+		}
+		&.form-button {
+			color: var(--studio-white);
+			background-color: var(--studio-black);
+			&:hover,
+			&:active {
+				background-color: var(--studio-gray-90);
+			}
```
mattgawarecki: I'm not sure about this background color mixed with the white text, but I don't see design specs for what it should be. @Automattic/jetpack-design, input here? (Ignore all the other aspects, those are addressed in other comments.)

image

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
```diff
@@ -0,0 +1,40 @@
...
+	}
+	@include breakpoint-deprecated( "<960px" ) {
+		flex-direction: column;
+	}
+
+	.jetpack-complete-page__get-complete-button {
+		@include breakpoint-deprecated( ">960px" ) {
+			margin-right: 2em;
+		}
+		&.form-button {
```
mattgawarecki, line 20: We need a rule inside this `.form-button` scope to specify that the Complete button has a `var(--studio-black)` border color, not the default green.

mattgawarecki (commented):
This was mostly a design review. (I used `R1U6Ww74mdgMrKUJ21BnTL-fi-3061%3A20967` as my source material, looking at the diagram inside the box marked "Change.")

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/index.tsx
TheSteveK: Since the page exists and is specific, I've changed it to that one.  We can always address that page later either ourselves with some help with marketing or maybe MarTech can take it on at some point.

client/my-sites/plans/jetpack-plans/jetpack-complete-page/components/cta-buttons/style.scss
TheSteveK: FWIW it looks like the white text is probably also the original button component's default styling.  I went ahead and switched the hover color to black and it looks more presentable:

Screenshot 2023-02-23 at 7 48 14 AM

mattgawarecki:
I think I may have led you astray with one of my earlier review comments, @TheSteveK. I had the false impression that this Complete offer page was part of the Jetpack Cloud environment, but in fact it's part of Calypso Blue. Redirecting visitors to `/pricing/jetpack-free/welcome` will give them a 404, because that path only exists in Jetpack Cloud.

The lack of a specific Free page is a gap to be sure, but since the presence of a site is virtually guaranteed, I think it's safer to revert back to `jetpack.com` as our redirect when we can't find a site.

TheSteveK:
> The lack of a specific Free page is a gap to be sure, but since the presence of a site is virtually guaranteed, I think it's safer to revert back to `jetpack.com` as our redirect when we can't find a site.

I changed it back to jetpack.com but we could just hard link the start free page `https://cloud.jetpack.com/pricing/...`

mattgawarecki:
As for the hover behavior/coloring: I took some time to compare similar buttons (on our pricing page, and in Jetpack Connect). In those places, it looks like there's no change in appearance when buttons are hovered, so let's set the `background-color` et al to the same values on :hover and :active as they are without being interacted with.

TheSteveK:
> As for the hover behavior/coloring: I took some time to compare similar buttons (on our pricing page, and in Jetpack Connect). In those places, it looks like there's no change in appearance when buttons are hovered, so let's set the `background-color` et al to the same values on :hover and :active as they are without being interacted with.

As part of the WCAG specs, buttons should have a visual indicator when they are hovered/selected maybe we can ask @Automattic/jetpack-design what it should look like.

I might also suggest maybe the Get Complete button needs a bit more of a color variation, the hover effect is very subtle.

TheSteveK:
Design was kind enough to whip up some quick layouts for the buttons in p1677160334601789-slack-C0D96691V and they have been implemented

andrii-lysenko (approved):
This is looking good to me! Good job in addressing all the suggestions.

I think we can release it as is, and, of course, if something comes up in a final review from a design team we could address that.

Tested and also works as expected.

PR #73390, opened 2023-02-15
WooExpress: Prevent negative value on the trial progress

### LINKED ISSUE #73312: Task: Improve time remaining calculation in eCommerce trial banner
### Details

It's possible that our calculations in the `ECommerceTrialBanner` component for computing `trialProgress` results in a value < 0. We noticed this in a walkthrough due to the progress "ring" being dots, one at 12 o'clock and a second at ~11 o'clock.

Walkthrough action items: peapX7-1DZ-p2

### Checklist

- [ ] Ensure trialProgress is between 0 and 1.

### Related

_No response_

valterlorran:
## Proposed Changes

* Prevents negative values for the trial progress, which would cause two dots to show up on the ring.

**Before**

Screen Shot 2023-02-15 at 15 29 41

## Testing Instructions

* Navigate to `/plans/my-plan/:siteSlug` on a WooExpress site.
* TBD

Closes #73312

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
@@ -39,10 +39,10 @@ const ECommerceTrialBanner = ( props: ECommerceTrialBannerProps ) => {
 	const trialDuration = trialEnd.diff( trialStart, 'days' );
 
 	/**
-	 * Trial progress from 0 to 1
+	 * Trial progress from 0 to 1.
+	 * Protect against values < 0 by wrapping the code in Math.max()
 	 */
-	const trialProgress =
-		eCommerceTrialDaysLeft <= trialDuration ? 1 - eCommerceTrialDaysLeft / trialDuration : 0;
+	const trialProgress = Math.max( 1 - eCommerceTrialDaysLeft / trialDuration, 0 );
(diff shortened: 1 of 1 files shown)


client/my-sites/plans/ecommerce-trial/ecommerce-trial-banner/index.tsx
```diff
@@ -41,7 +41,8 @@ const ECommerceTrialBanner = ( props: ECommerceTrialBannerProps ) => {
 	/**
 	 * Trial progress from 0 to 1
 	 */
-	const trialProgress = 1 - eCommerceTrialDaysLeft / trialDuration;
+	const trialProgress =
+		eCommerceTrialDaysLeft <= trialDuration ? 1 - eCommerceTrialDaysLeft / trialDuration : 0;
```
daledupreez: Two thoughts:
1. This is really complicated to read! Could we use `Math.max()` to simplify the expression?
2. I was testing with a RTL language earlier today (I think it was Hebrew) and things looked a bit weird -- it may be worth looking at that as well while you're looking at this component.
```suggestion
	// Protect against values < 0 by wrapping the code in Math.max()
	const trialProgress = Math.max( 1 - eCommerceTrialDaysLeft / trialDuration, 0 );
```

valterlorran: I created this issue with the problems I found: https://github.com/Automattic/wp-calypso/issues/73392

valterlorran:
Closing this, as #73435 was deployed and the value can't be negative anymore.

PR #73326, opened 2023-02-14
Domain Upsell: Send Search domain button to the correct flow

lupus2k:
Fixes https://github.com/Automattic/dotcom-forge/issues/1610

## Proposed Changes

Send the `Search for a domain` to the correct flow

## Testing Instructions

- With a Free site go to `/home`
- Click on the `Search for a domain` button on the Domain Upsell Banner 
- It should send you to select a domain
- After the domain you should see the Plans page.

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] ~~[Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?~~
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] ~~Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)~~
- [ ] ~~Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?~~
- [ ] ~~For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~~

PR #73391, opened 2023-02-15
Created test for My home domain-upsell custom domain flow

paulopmt1:
Related to 1685-gh-Automattic/dotcom-forge

## Proposed Changes

* Created component test for domain-upsell flow

## Testing Instructions

* Run `yarn run test-client client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx` and make sure the test passes
* Ensure this test covers the previous change [we did at this task](https://github.com/Automattic/wp-calypso/pull/73326/files)

### E2E testing instructions
- First of all, build the calypso-e2e package with `yarn build` on `/packages/calypso-e2e`.
- Follow [this guide](https://github.com/Automattic/wp-calypso/blob/trunk/test/e2e/README.md#quick-start) to make sure to have the environment working.
- Run `yarn workspace wp-e2e-tests test -- domain-upsell`
- Make sure all test passes

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
+++ b/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
@@ -9,31 +9,32 @@ import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 import DomainUpsell from '../';
 
-jest.mock( 'calypso/state/ui/selectors', () => {
-	return {
-		getSelectedSite: () => {
-			return {
-				ID: 1,
-				plan: 'free',
-			};
-		},
-		getSelectedSiteSlug: () => 'example.wordpress.com',
-	};
-} );
+const sites = [];
+sites[ 1 ] = {
+	ID: 1,
+	URL: 'example.wordpress.com',
+	plan: {
+		product_slug: 'free_plan',
+	},
+};
 
-jest.mock( 'calypso/state/sites/domains/selectors', () => {
-	return {
-		getDomainsBySiteId: () => {
-			return [ 'example.wordpress.com' ];
+const initialState = {
+	sites: {
+		items: sites,
+		domains: {
+			items: [ 'example.wordpress.com' ],
 		},
-	};
-} );
-
-jest.mock( '@automattic/calypso-products', () => {
-	return {
-		isFreePlanProduct: () => true,
-	};
-} );
+	},
+	ui: {
+		selectedSiteId: 1,
+	},
+	currentUser: {
+		id: 12,
+		user: {
+			email_verified: true,
+		},
+	},
+};
 
 jest.mock( '@automattic/domain-picker/src', () => {
 	return {
@@ -50,18 +51,11 @@ jest.mock( '@automattic/domain-picker/src', () => {
 	};
 } );
 
-jest.mock( 'calypso/state/current-user/selectors', () => ( {
-	getCurrentUserId: jest.fn( () => 12 ),
-	isUserLoggedIn: jest.fn( () => true ),
-	isCurrentUserEmailVerified: jest.fn( () => true ),
-} ) );
-
 let pageLink = '';
 jest.mock( 'page', () => ( link ) => ( pageLink = link ) );
 
 describe( 'index', () => {
 	test( 'Should show H3 content for the Home domain upsell and test search domain button link', async () => {
-		const initialState = {};
 		const mockStore = configureStore();
 		const store = mockStore( initialState );
 
@@ -87,10 +81,9 @@ describe( 'index', () => {
 		nock.cleanAll();
 		nock( 'https://public-api.wordpress.com' )
 			.persist()
-			.post( '/rest/v1.1/me/shopping-cart/no-site' )
+			.post( '/rest/v1.1/me/shopping-cart/1' )
 			.reply( 200 );
 
-		const initialState = {};
 		const mockStore = configureStore();
 		const store = mockStore( initialState );
 
--- a/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
+++ b/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
@@ -1,0 +1,67 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { render } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+import DomainUpsell from '../';
+
+jest.mock( 'calypso/state/ui/selectors', () => {
+	return {
+		getSelectedSite: () => {
+			return {
+				ID: 1,
+				plan: 'free',
+			};
+		},
+		getSelectedSiteSlug: () => 'example.wordpress.com',
+	};
+} );
+
+jest.mock( 'calypso/state/sites/domains/selectors', () => {
+	return {
+		getDomainsBySiteId: () => {
+			return [ 'example.wordpress.com' ];
+		},
+	};
+} );
+
+jest.mock( '@automattic/calypso-products', () => {
+	return {
+		isFreePlanProduct: () => true,
+	};
+} );
+
+jest.mock( 'calypso/state/current-user/selectors', () => ( {
+	getCurrentUserId: jest.fn( () => 12 ),
+	isUserLoggedIn: jest.fn( () => true ),
+	isCurrentUserEmailVerified: jest.fn( () => true ),
+} ) );
+
+describe( 'index', () => {
+	describe( 'verification nudge', () => {
+		test( 'Should show H3 content for the Home domain upsell and the correct /domains/add link', () => {
+			const initialState = {};
+			const mockStore = configureStore();
+			const store = mockStore( initialState );
+
+			const { container } = render(
+				<Provider store={ store }>
+					<DomainUpsell />
+				</Provider>
+			);
+
+			const notice = container.querySelector( 'h3' );
+			expect( notice ).toBeVisible();
+			expect( notice ).toHaveTextContent( 'Own your online identity with a custom domain' );
+
+			const links = [].slice.call( container.querySelectorAll( 'a' ) );
+			expect(
+				links.some( ( link ) =>
+					link.href.endsWith( '/domains/add/example.wordpress.com?domainAndPlanPackage=true' )
+				)
+			).toBeTruthy();
+		} );
+	} );
+} );
--- a/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
+++ b/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
@@ -2,7 +2,7 @@
  * @jest-environment jsdom
  */
 
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import nock from 'nock';
 import { Provider } from 'react-redux';
@@ -59,21 +59,20 @@ describe( 'index', () => {
 		const mockStore = configureStore();
 		const store = mockStore( initialState );
 
-		const { container } = render(
+		render(
 			<Provider store={ store }>
 				<DomainUpsell />
 			</Provider>
 		);
+		expect( 1 ).toBe( 1 );
 
-		const notice = container.querySelector( 'h3' );
-		expect( notice ).toBeVisible();
-		expect( notice ).toHaveTextContent( 'Own your online identity with a custom domain' );
+		const h3 = screen.getByText( 'Own your online identity with a custom domain' );
+		expect( h3 ).toBeInTheDocument();
 
-		const links = [].slice.call( container.querySelectorAll( 'a' ) );
+		const searchLink = screen.getByText( 'Search for a domain' );
+		expect( searchLink ).toBeInTheDocument();
 		expect(
-			links.some( ( link ) =>
-				link.href.endsWith( '/domains/add/example.wordpress.com?domainAndPlanPackage=true' )
-			)
+			searchLink.href.endsWith( '/domains/add/example.wordpress.com?domainAndPlanPackage=true' )
 		).toBeTruthy();
 	} );
 
@@ -87,14 +86,14 @@ describe( 'index', () => {
 		const mockStore = configureStore();
 		const store = mockStore( initialState );
 
-		const { getByText } = render(
+		render(
 			<Provider store={ store }>
 				<DomainUpsell />
 			</Provider>
 		);
 
 		const user = userEvent.setup();
-		await user.click( getByText( 'Buy this domain' ) );
+		await user.click( screen.getByText( 'Buy this domain' ) );
 		expect( pageLink ).toBe( '/plans/yearly/example.wordpress.com?get_domain=example.blog' );
 	} );
 } );
--- a/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
+++ b/client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
@@ -60,50 +60,48 @@ let pageLink = '';
 jest.mock( 'page', () => ( link ) => ( pageLink = link ) );
 
 describe( 'index', () => {
-	describe( 'verification nudge', () => {
-		test( 'Should show H3 content for the Home domain upsell and test search domain button link', async () => {
-			const initialState = {};
-			const mockStore = configureStore();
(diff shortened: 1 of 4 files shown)


test/e2e/README.md
```diff
@@ -101,7 +101,7 @@ yarn workspace wp-e2e-tests build --watch
 12. run test.
 
 ```bash
-yarn workspace wp-e2e-tests start -- <test_path>
+yarn workspace wp-e2e-tests test -- <test_path>
```
paulopmt1, line 104: Thanks for updating the docs @lupus2k!

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
```diff
@@ -0,0 +1,130 @@
+/**
```
paulopmt1: Great test @lupus2k, it's working well and validates the whole flow!

lupus2k: Mi next plan is to test this flow with the experiment

packages/calypso-e2e/src/lib/components/sidebar-component.ts
```diff
@@ -7,6 +7,7 @@ type FocusType = 'Sites' | 'Sidebar';
 
 const selectors = {
 	sidebar: '.sidebar',
+	sidebarNotice: '.sidebar .current-site__notices .banner__action button:visible',
```
dpasque: **Suggestion (non-blocking)**: Since you pass in the button name ("item"), in `openNotice` below, what if we made this a dynamic selector that is used?

```suggestion
	sidebarNoticeButton: ( name: string ) =>
		`.sidebar .current-site__notices button:text("${ name }")`,
```

And then later...
```typescript
		const itemSelector = selectors.sidebarNoticeButton( item );
		await this.page.dispatchEvent( itemSelector, 'click' );
```

packages/calypso-e2e/src/lib/components/sidebar-component.ts
```diff
@@ -106,6 +107,37 @@ export class SidebarComponent {
...
 	}
 
+	/**
+	 * Open a notice of the sidebar menu.
+	 *
+	 * @param {string} item Plaintext representation of the top level heading.
+	 * @param {string} expectedUrl expected URL after clicking on the notice.
+	 * @returns {Promise<void>} No return value.
+	 */
+	async openNotice( item: string, expectedUrl: string ): Promise< void > {
```
dpasque: **Nit (non-blocking)**: On my first read, I was a bit confused by the `item` parameter name and JSDoc. It looks like it's being treated like the name of the button in the notice -- what do you think of making that a bit more explicit? 

(Like `noticeButtonName`, e.g.)

packages/calypso-e2e/src/lib/components/sidebar-component.ts
```diff
@@ -106,6 +107,37 @@ export class SidebarComponent {
...
+		// Do not verify selected menu items or retry if navigation takes user out of Calypso (eg. WP-Admin, Widgets editor)...
+		if ( ! currentURL.startsWith( getCalypsoURL() ) ) {
+			return;
+		}
+		// ... or to a page in Calypso that closes the sidebar.
+		if ( currentURL.match( /\/(post|page|site-editor)\// ) ) {
+			return;
+		}
+
+		await this.page.waitForURL( expectedUrl );
```
dpasque: **Suggestion (non-blocking)**: I wonder if you might be able to simplify this section just a little bit?

The way I see it, is people may not always need to validate the landing URL on clicking an upsell. And even if they do provide one that takes them outside Calypso, they still may want to wait for that URL!

So what if `expectedUrl` was an optional parameter, and this section was just...
```typescript
if ( expectedUrl ) {
await this.page.waitForUrl( epectedUrl );
```

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
```diff
@@ -0,0 +1,130 @@
...
+			const credentials = SecretsManager.secrets.testAccounts.simpleSiteFreePlanUser;
+
+			restAPIClient = new RestAPIClient( credentials );
+			console.info( 'Creating a new test site.' );
+			newSiteDetails = await restAPIClient.createSite( {
+				name: blogName,
+				title: blogName,
+			} );
+			console.info( `New site created: ${ newSiteDetails.blog_details.url }` );
+			siteCreatedFlag = true;
```
dpasque: **Question (not blocking per say, but we should discuss pre-merge!)**: This test design certainly works fine, but I'm always curious if there's a way to design the test so that you don't have to make a new site!

It looks like you don't actually make the upgrade/purchase, so the site doesn't actually change state. So that just means we really have to handle concurrency issues -- but from what I can tell, we may just have to add some resiliency around the domain selection step!

The other main perk of this would be that you could move this from only running post merge ( the `calypso-release` group) to running on every PR! That might get you feedback even sooner!

Thoughts or ideas? I'd be glad to help with the test design however I can! 😄

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+let pageLink = '';
+jest.mock( 'page', () => ( link ) => ( pageLink = link ) );
+
+describe( 'index', () => {
+	test( 'Should show H3 content for the Home domain upsell and test search domain button link', async () => {
+		const initialState = {};
+		const mockStore = configureStore();
+		const store = mockStore( initialState );
+
+		const { container } = render(
```
dpasque: **Suggestion (non-blocking)**: There's actually a `screen` object you can import directly from `testing-library/react` that they recommend using for your queries! [Context](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)

Not required at all, but a nice convention :)

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+		const notice = container.querySelector( 'h3' );
+		expect( notice ).toBeVisible();
+		expect( notice ).toHaveTextContent( 'Own your online identity with a custom domain' );
+
+		const links = [].slice.call( container.querySelectorAll( 'a' ) );
+		expect(
+			links.some( ( link ) =>
+				link.href.endsWith( '/domains/add/example.wordpress.com?domainAndPlanPackage=true' )
+			)
+		).toBeTruthy();
```
dpasque: **Suggestion (non-blocking)**: If possible, I always try to use the more "what the user sees" centric queries provided by React Testing Library! Like `getByRole`.  [Context](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time) and [API](https://testing-library.com/docs/queries/about/#priority).

So, for example...
```javascript
expect( screen.getByRole( 'heading', { name: 'Own your online identity with a custom domain' } ) ).toBeInTheDocument();
```

I like those because they...
1. Test your component more like a real user
2. End up being more resilient over time! (Like, if you ended up swapping out the heading level.)

Not required at all, but just a suggestion! 😄

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+	};
+} );
+
+jest.mock( 'calypso/state/sites/domains/selectors', () => {
+	return {
+		getDomainsBySiteId: () => {
+			return [ 'example.wordpress.com' ];
+		},
+	};
+} );
```
dpasque: **Question (non-blocking)**: How hard would it be to set these values in the `initialState` that you pass into the render function below?

Sometimes it's too onerous, and it's just better to mock selectors. But, whenever I can just set redux state directly, I always try to!

This usually gets you a couple of benefits:
1. It expands your test coverage a pinch - it makes sure that the selectors you are using actually select as you expect!
2. In that way, it can actually provide a pinch more flexibility over time. If you refactor the selectors without changing the redux state tree shape, these tests wouldn't need to be updated!

dpasque (commented):
Looking great! 🥳 The E2E script especially runs very nicely! We really appreciate you taking the time to add tests for your work! 👏 

I've added some non-blocking suggestions with some conventions and tips. Feel free to grab and use any of them you find helpful!

The one thing that might be good to discuss pre-merge is the question I have about potentially redesigning the test to be able to run without making a new site. It might not be the right call, but is worth considering for sure. It allows more frequent runs and minimizes data artifacts!

Feel free to ping me with any questions! 😄 I'll be AFK tomorrow, but you can also ping any KitKatters!

packages/calypso-e2e/src/lib/components/sidebar-component.ts
lupus2k: You're right, I started with an approach and then change it but didn't change the parameter name :)

packages/calypso-e2e/src/lib/components/sidebar-component.ts
lupus2k: You're right; sometimes, the notice CTA is not to open a page

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
lupus2k: Thanks for your explanation, now I understand that as we are using the `simpleSiteFreePlanUser` account, which has simple sites, we can use the primary site to test it without creating a new one and remove the `calypso-release` group

packages/calypso-e2e/src/lib/components/sidebar-component.ts
lupus2k: Sounds good to re-use it in the future!

lupus2k:
Hey, @dpasque, I [followed all your suggestions](https://github.com/Automattic/wp-calypso/tree/449d34e5266a0815c7422635e1ae35e07fefd337) for the e2e tests, and all is working fine except when trying to empty the cart, apparently is a problem with the `NavbarComponent` because it does not even open the cart overlay. 
I see `navbarCartComponent.emptyCart()` is only used [here](https://github.com/Automattic/wp-calypso/blob/fcd722ed7328dcfa760e274883b19c76c9c15c07/test/e2e/specs/domains/domains__add.ts#L54), but this test is skipped, should I take care of this, or can I leave the test as it is?

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
paulopmt1: Loved your suggestion @dpasque, [did this change here](https://github.com/Automattic/wp-calypso/pull/73391/commits/2e2cc3b03b46a1394f152c0c15ac3e358fe4975e). 

It feels like mocking helps us to write the first iteration of the test, so we don't need to be concerned about the state, but once it's working, changing to state improves a lot the code coverage.

Thanks!

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
paulopmt1: Nice! [Improved it here](https://github.com/Automattic/wp-calypso/pull/73391/commits/7f5f1893912aef00293eb27d77838354ac74acc3).

client/my-sites/customer-home/cards/features/domain-upsell/test/index.jsx
paulopmt1: Loved this! Thanks for the code example, helped a lot!
[Changed it here](https://github.com/Automattic/wp-calypso/pull/73391/commits/87f29cda08571066d012ce170859c50de265f366).

paulopmt1:
Thanks for all your suggestions @dpasque, we're ready for the next round of review :)

worldomonation:
> Hey, @dpasque, I [followed all your suggestions](https://github.com/Automattic/wp-calypso/tree/449d34e5266a0815c7422635e1ae35e07fefd337) for the e2e tests, and all is working fine except when trying to empty the cart, apparently is a problem with the `NavbarComponent` because it does not even open the cart overlay. I see `navbarCartComponent.emptyCart()` is only used [here](https://github.com/Automattic/wp-calypso/blob/fcd722ed7328dcfa760e274883b19c76c9c15c07/test/e2e/specs/domains/domains__add.ts#L54), but this test is skipped, should I take care of this, or can I leave the test as it is?

Right, the cart method in `NavbarComponent` is likely out of date since as you said, the only spec that uses the component has been in quarantine for over a year.

I'll take a look at the revised code and edit this comment later with information on what you can do.

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
```diff
@@ -0,0 +1,96 @@
...
+	SecretsManager,
+	NavbarCartComponent,
+} from '@automattic/calypso-e2e';
+import { Page, Browser } from 'playwright';
+
+declare const browser: Browser;
+
+describe(
+	DataHelper.createSuiteTitle(
+		'Domain Upsell: Click on the sidebar Domain Upsell, search for a domain, select the Premium plan and check both items on the checkout page'
```
worldomonation: **nit:non-blocking** - by no means a blocker, just a nit - ideally, the suite title should be fairly short (one screen width at most) and just describe the overall feature or flow being tested.

I would title it something like:

```suggestion
		'Sidebar: Domain upsell'
```

and that should convey the overall intention of the suite.

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
```diff
@@ -0,0 +1,96 @@
...
+		beforeAll( async function () {
+			// Launch browser.
+			page = await browser.newPage();
+
+			// Authenticate as simpleSiteFreePlanUser.
+			const testAccount = new TestAccount( 'simpleSiteFreePlanUser' );
+			await testAccount.authenticate( page );
+		} );
+
+		describe( `Upgrade to WordPress.com ${ planName }`, function () {
```
worldomonation: **nit:non-blocking** - maybe this is intentional, but it seems the sub-suite title here (upgrading to Premium) doesn't appear to correlate to the actual test flow (Domain upsell). Maybe this is just my misunderstanding, since the rest of the test steps also mention the WordPress.com Premium plan throughout.

----

Additionally, since the spec isn't subdivided into sub-suites (which is the use case for nested `describe` blocks), one approach to simplify is to remove this describe entirely and have just `it` steps hanging below the top level `describe` block.

If you do that though, make sure to combine the `beforeAll` steps into one :)

worldomonation (approved):
I'm filling in for @dpasque who is off on Fridays - generally looking good, just a couple of minor (style-related) nits that I had. I'll approve to unblock progress and since the nits are non-blocking it isn't necessary to address before merge.

worldomonation:
Re-triggered the E2E builds for desktop/mobile. It's always the `Enter post content` that fails :\

image

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
lupus2k: Thanks for your suggestion, I updated it.

test/e2e/specs/domain-upsell/domain-upsell__sidebar.ts
lupus2k: You are right, and this sub-suite was different from my intention, I made some tests and forgot to remove it 🤦‍♂️  
I updated the code now.

dpasque:
@lupus2k @paulopmt1 -- this is so cool to see that you were able to redesign the test to avoid site creation! 😄 🥳 That should be really helpful in the long run, thanks for making that update!

I noticed though that the new spec doesn't have a jest group added to the top, so it won't automatically be picked up by our CI jobs. Was this ready to be run as part of the `calypso-pr` group? If so, I would add this to the top of the file...

```
/**
 * @group calypso-pr
 */
```

lupus2k:
Thanks, @dpasque. As this PR is already merged, do you think we have to create a new PR just for this, or we can add it to another working PR?

worldomonation:
> Thanks, @dpasque. As this PR is already merged, do you think we have to create a new PR just for this, or we can add it to another working PR?

I'd say it should be in its own PR to separate out unrelated work. You should be able to edit it from GitHub web UI as well, and cut a PR quickly from there.

lupus2k:
Thank you all. 
I just created this: https://github.com/Automattic/wp-calypso/pull/73608

PR #67835, opened 2022-09-14
Launchpad: Add Link in bio post-setup flow

edanzer:
### Proposed Changes

**Related to:** https://github.com/Automattic/wp-calypso/issues/67429

Before arriving at Launchpad, users go through the linkInBioSetup stepper screen. We want to allow them to edit the same values from that screen after arriving at Launchpad. Following [discussion from a prior PR,](https://github.com/Automattic/wp-calypso/pull/67523#pullrequestreview-1107102774) rather than send users back to the original Setup screen, we are creating a new stepper flow and screen. Users can navigate between Launchpad and the new Setup screen, and changes made on new screen should be reflected on their site, including the Launchpad website preview. 

Specific changes include: 
* Separate setup form in original linkInBioSetup screen to its own component for re-use.
* Add link-in-bio-post-setup flow, with linkInBioPostSetup screen (uses same form as the original).
* Ensure linkInBioPostSetup screen pre-populates site title/tagline/icon from the redux SITE_STORE.
* Ensure changes to site title/tagline/icon are updated on the site.
* Add link to new flow/screen from Launchpad. 

### Testing Instructions

1. Checkout this branch, and run yarn and yarn start if needed.
2. Start a new link-in-bio site from https://wordpress.com/hp-2022-tailored-flows/
3. As soon as you've started the process (and before you arrive at the Link in Bio setup screen), switch to your local Calypso instance. Replace `https://worpdress.com` with `http://calypso.localhost:3000/` in your browser. 
4. TEST: When you arrive on the first link-in-bio set up screen where you set title/tagline/icon, confirm that everything works as expected. I moved a large part of this screen to a new component, so we want to be sure no regressive breakages.
5. TEST: Continue as normal until Launchpad, and confirm that the site title/tagline/icon in the Launchpad web preview is correct. Again, we want to ensure no regressive breakages. 
6. TEST: From Launchpad, confirm the 'Set up Link in Bio' task is clickable. 
7. TEST: Click the 'Set up Link in Bio' task. You should arrive at a screen that looks exactly like the original setup screen, but the url of this screen should be different: http://calypso.localhost:3000/setup/linkInBioPostSetup?flow=link-in-bio-post-setup&siteSlug=[YOUR_SITE_SLUG]
8. TEST: Confirm you can edit the site title, tagline, and icon. 
9. TEST: Click the 'Continue' button and confirm you are returned to Launchpad. 
10. TEST: Confirm in the Launchpad web preview that your new site title, tagline, and icon are there.

[...]

PR #73401, opened 2023-02-15
Launchpad: Add flow and link to update design

edanzer:
### Proposed Changes

For our general flows, we want to provide a link from launchpad back to the design/theme selection screen. To do that, this PR makes the following changes: 
  - Adds a new update-design new flow that just uses the same designSetup screen as usual. 
  - Adds a link from launchpad to the new update-design flow/screen. 

### Testing Instructions

Review Time: Medium
Testing Time: Short

1) Check out this branch and run yarn and yarn start if needed. 
2) Go to http://calypso.localhost:3000/setup/free/intro and start a free site. 
3) Go all the way through to Launchpad.
4) TEST: Confirm the 'Choose Design' link is now active. 
5) TEST: Click the 'Choose Design link and confirm you go to the design picker screen.
6) TEST: Choose a new theme. Confirm you are taken back to Launchpad and that your new theme shows in the preview.
7) TEST: Also just make sure that everything else about the flow back-and-forth between launchpad and design selection looks and feels as it should.

### DIFF
--- a/client/landing/stepper/declarative-flow/update-design.ts
+++ b/client/landing/stepper/declarative-flow/update-design.ts
@@ -12,8 +12,8 @@ import Processing from './internals/steps-repository/processing-step';
 import { ProvidedDependencies } from './internals/types';
 import type { Flow } from './internals/types';
 
-const designSetup: Flow = {
-	name: 'design-post-setup',
+const updateDesign: Flow = {
+	name: 'update-design',
 	get title() {
 		return translate( 'Choose Design' );
 	},
@@ -30,7 +30,7 @@ const designSetup: Flow = {
 		const flowToReturnTo = useQuery().get( 'flowToReturnTo' ) || 'free';
 
 		function submit( providedDependencies: ProvidedDependencies = {} ) {
-			recordSubmitStep( providedDependencies, 'design-post-setup', flowName, currentStep );
+			recordSubmitStep( providedDependencies, 'update-design', flowName, currentStep );
 			switch ( currentStep ) {
 				case 'processing':
 					return window.location.assign(
@@ -61,4 +61,4 @@ const designSetup: Flow = {
 	},
 };
 
-export default designSetup;
+export default updateDesign;
--- a/client/landing/stepper/declarative-flow/design-post-setup.ts
+++ b/client/landing/stepper/declarative-flow/design-post-setup.ts
@@ -1,4 +1,3 @@
-// import { useSelect } from '@wordpress/data';
 import { translate } from 'i18n-calypso';
 import {
 	setSignupCompleteSlug,
@@ -7,7 +6,6 @@ import {
 } from 'calypso/signup/storageUtils';
 import { useQuery } from '../hooks/use-query';
 import { useSiteSlug } from '../hooks/use-site-slug';
-// import { ONBOARD_STORE } from '../stores';
 import { recordSubmitStep } from './internals/analytics/record-submit-step';
 import DesignSetup from './internals/steps-repository/design-setup';
 import Processing from './internals/steps-repository/processing-step';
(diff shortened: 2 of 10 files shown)


jeyip:
> Confirm you are taken back to Launchpad and that your new theme shows in the preview.

I'm unable to get the theme to render in the preview. When I navigate to Appearance > Themes, it shows that my new design hasn't been selected. Did you happen to see anything like this during development @edanzer ?

I checked out this branch, ran `yarn start`, and started going through http://calypso.localhost:3000/setup/free/intro

edanzer:
> I'm unable to get the theme to render in the preview. When I navigate to Appearance > Themes, it shows that my new design hasn't been selected. Did you happen to see anything like this during development @edanzer ?

To confirm, is the issue that the preview is not rendering at all, or that it is rendering, but with the old theme (ie, theme just isn't updating)? Early on, I had issues where the theme would not update, and you'd see same/original theme in the launchpad preview, but that's been resolved.

jeyip:
## Testing
### Requirements
- [x] Confirm the 'Choose Design' link is now active.
- [x] Confirm you go to the design picker screen.
- [x] Confirm you are taken back to Launchpad and that your new theme shows in the preview.

### Browsers
- [x] Chrome

client/landing/stepper/declarative-flow/update-design.ts
```diff
@@ -0,0 +1,64 @@
...
+				case 'processing':
+					return window.location.assign(
+						`/setup/${ flowToReturnTo }/launchpad?siteSlug=${ siteSlug }`
+					);
+
+				case 'designSetup':
+					if ( providedDependencies?.goToCheckout ) {
+						const destination = `/setup/${ flowToReturnTo }/launchpad?siteSlug=${ providedDependencies.siteSlug }`;
+						persistSignupDestination( destination );
+						setSignupCompleteSlug( providedDependencies?.siteSlug );
```
jeyip, line 44: `NB:`

This doesn't appear to be affecting functionality, but do we need to set `setSignupCompleteSlug` and `setSignupCompleteFlowName`?

jeyip (approved):

agrullon95 (approved):
Changes are working well for me.

Is there a reason why we create a new "update-design" flow instead of reusing/navigating to the existing designSetup component within each flow? I'm curious about the approach since I may have missed some context.

### Testing

- [x]  Confirm the 'Choose Design' link is now active.
- [x]   Click the 'Choose Design link and confirm you go to the design picker screen.
- [x]  Choose a new theme. Confirm you are taken back to Launchpad and that your new theme shows in the preview.

edanzer:
> Is there a reason why we create a new "update-design" flow instead of reusing/navigating to the existing designSetup component within each flow? I'm curious about the approach since I may have missed some context.

@agrullon95 - Yes, there's a preference for stepper to avoid navigation back and forth between steps within a flow. This is same reason we created the Post Setup flows for LIB / Newsletter / Free. For origin of the conversation, see [this comment on a past PR](https://github.com/Automattic/wp-calypso/pull/67523#pullrequestreview-1107102774). I'd also commented/referred to this when doing the first "Post Setup" flow ([see PR description](https://github.com/Automattic/wp-calypso/pull/67835)), which creates a small separate flow with an extra setup screen, rather than just navigating user back to the original setup screen.

client/landing/stepper/declarative-flow/update-design.ts
edanzer: @jeyip - I copied this logic from the free flow [here](https://github.com/Automattic/wp-calypso/blob/00f4e78eb45b97ea108c495d9b206b13e03443c7/client/landing/stepper/declarative-flow/free.ts#L98). 

I considered that it may not be needed here, as this flow is accessed later in the onboarding flow than the original designSetup step. 

But this logic is here to handle if a user upgrades to a premium theme. It seemed plausible that a user would pick a free theme the first time, and then when changing themes here, pick a premium them. So they'd still need to be routed through the same upgrade/cart process. Thus leaving all that logic here and intact. 

Let me know if you have other concerns.

PR #73412, opened 2023-02-16
Add/config flags storage update

gmovr:
Related to Automattic/martech#1509

## Proposed Changes

* Adding a feature flag for storage add-on. E.g. for conditional visibility, testing, and to guard this feature until ready.
* Adding a selector to check the flag. For now it's just checking the config flag, but we will extend this with other conditions, e.g. certain plans if needed, and it makes it easier in case the feature flag gets renamed, or the logic changes otherwise. I didn't add any tests for it, because it's just taking a boolean from the config object and returns it, and casting it to boolean via `!!` (since it's reading from an untyped JSON file). 

## Testing Instructions

* Check that all the environments have this flag, and that it's set to `false`.
* For manual testing, you can also import the selector anywhere you'd like, e.g. the plans page, and `console.log` it, or add a `debugger;` statement inside the selector and verify that the value is read correctly (in this case it should return `false`).

## Pre-merge Checklist

Resolves Automattic/martech#1509


client/state/selectors/is-storage-addon-available.ts
```diff
@@ -0,0 +1,10 @@
+import config from '@automattic/calypso-config';
+
+/**
+ * Is the storage addon (space upgrade) available, based on config flag vs. environment.
+ *
+ * @returns {boolean} - Whether or not the storage addon is available
+ */
+const isStorageAddonAvailable = (): boolean => !! config.isEnabled( 'storage-addon' );
```
chriskmnds: This probably shouldn't live under `state/selectors` if it's not interacting with state directly :-)

gmovr: Cool, that makes sense. I was kind of envisioning the config as global state, but technically it isn't. Will move it into the addons folder👍

chriskmnds: Yeah. The selectors here don't have any connection to state, not outside of what is passed into them. So unless you are passing state into the exported function, then it probably doesn't belong here. I don't _think_ there are exceptions.

ddc22:
Probably stale and can be closed off @gmovr @jeyip ?

jeyip:
> Probably stale and can be closed off @gmovr @jeyip ?

After reading the PR description, I believe so 👍. Closing this -- we can always reopen if I'm incorrect @gmovr

PR #73428, opened 2023-02-16
Premium Global Styles: Extend Gating to Older Sites

Copons:
Fixes https://github.com/Automattic/dotcom-forge/issues/1657

## Proposed Changes

* Extend the Global Styles gating to older sites that never used Global Styles before.

Premium Global Styles has been soft-launched to sites created after 15 Dec 2022; we are now extending it to older sites, as long as they never customized Global Styles before.

To avoid performing this expensive check every time, we will label older sites with the `wpcom-premium-global-styles-exemption-checked` sticker and with the `wpcom-premium-global-styles-exempt` to identify those that are exempt from Premium Global Styles (meaning: they had used Global Styles before).

Since stickers are not available on Atomic, we will ignore the edge case of Atomic Free sites, as they are not enough to justify the effort required.
Having said that, I'm open to alternative solutions!

## Testing Instructions

* Apply these changes to the sandbox with `install-plugin.sh etk try/extend-global-styles-gating`.
* Pick two Free sites created before 15 Dec 2022.
  * One should have used Global Styles in the past.
  * The other shouldn't. If no such site exists, it should be enough to delete all its `wp_global_styles` posts.
* Test on these kind of sites:
  * A: Free site created before 15 Dec 2022, that has used GS in the past.
  * B: Free site created before 15 Dec 2022, that has never used GS. If no such site exists, it should be enough to delete all its `wp_global_styles` posts.
  * C: Free site created after 15 Dec 2022.
  * D: Premium site created before 15 Dec 2022.
  * E: Premium site created after 15 Dec 2022.
* Sandbox all sites.
* Site A:
  * Ensure it has "unlimited" GS (which is the current behaviour).
* Site B:
  * Ensure it gets a `wpcom-premium-global-styles-exempt` sticker, and has limited GS (new behaviour).
* Site C:
  * Ensure it has limited GS (current behaviour).
* Site D and E:
  * Ensure they have unlimited GS (current behaviour). 

## Pre-merge Checklist

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
@@ -295,17 +295,24 @@ function wpcom_premium_global_styles_is_site_exempt( $blog_id = 0 ) {
 		return false;
 	}
 
-	// Non-Simple sites are edge cases; other conditions will determine whether they can use GS.
-	if ( ! defined( 'IS_WPCOM' ) || ! IS_WPCOM ) {
-		return false;
-	}
-
 	// If the exemption check has already been performed, just return if the site is exempt.
 	if ( wpcom_global_styles_has_blog_sticker( 'wpcom-premium-global-styles-exemption-checked', $blog_id ) ) {
 		return wpcom_global_styles_has_blog_sticker( 'wpcom-premium-global-styles-exempt', $blog_id );
 	}
 
+	// Non-Simple sites are edge cases; other conditions will determine whether they can use GS.
+	if ( ! defined( 'IS_WPCOM' ) || ! IS_WPCOM ) {
+		return false;
+	}
+
 	switch_to_blog( $blog_id );
+
+	// If the current user cannot modify the `wp_global_styles` CPT, the exemption check is not needed;
+	// other conditions will determine whether they can use GS.
+	$wp_global_styles_cpt = get_post_type_object( 'wp_global_styles' );
+	if ( ! current_user_can( $wp_global_styles_cpt->cap->publish_posts ) ) {
+		return false;
+	}
 
 	add_blog_sticker( 'wpcom-premium-global-styles-exemption-checked', null, null, $blog_id );
 
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
@@ -286,6 +286,9 @@ function wpcom_global_styles_in_use() {
  * it was created before the Premium Global Styles launch date (2022-12-15)
  * and had already customized its Global Styles.
  *
+ * We use blog stickers and other strategies to only perform the intensive check
+ * when strictly needed.
+ *
  * @param  int $blog_id Blog ID.
  * @return bool Whether the site is exempt from Premium Global Styles.
  */
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
@@ -45,10 +45,10 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 	}
 
 	// Do not limit Global Styles on sites created before we made it a paid feature (2022-12-15),
-	// that have never used Global Styles.
+	// that had already used Global Styles.
 	if (
 		$blog_id < 213403000 &&
-		defined( 'IS_WPCOM' ) && IS_WPCOM && ! wpcom_global_styles_has_been_used( $blog_id )
+		defined( 'IS_WPCOM' ) && IS_WPCOM && wpcom_global_styles_has_been_used( $blog_id )
 	) {
 		return false;
 	}
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
@@ -34,11 +34,6 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 		}
 	}
 
-	// Limit Global Styles on sites with the 'wpcom-limit-global-styles' sticker.
-	if ( wpcom_global_styles_has_blog_sticker( 'wpcom-limit-global-styles', $blog_id ) ) {
-		return true;
-	}
-
 	// Do not limit Global Styles on demo sites.
 	if ( wpcom_global_styles_has_blog_sticker( 'theme-demo-site', $blog_id ) ) {
 		return false;
@@ -46,10 +41,7 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 
 	// Do not limit Global Styles on sites created before we made it a paid feature (2022-12-15),
 	// that had already used Global Styles.
-	if (
-		$blog_id < 213403000 &&
-		defined( 'IS_WPCOM' ) && IS_WPCOM && wpcom_global_styles_has_been_used( $blog_id )
-	) {
+	if ( wpcom_premium_global_styles_is_site_exempt( $blog_id ) ) {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -34,26 +34,50 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
...
+	if ( wpcom_global_styles_has_blog_sticker( 'theme-demo-site', $blog_id ) ) {
 		return false;
 	}
 
-	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'theme-demo-site', $blog_id ) ) {
+	// Do not limit Global Styles on sites created before we made it a paid feature (2022-12-15),
+	// that have never used Global Styles.
+	if (
+		$blog_id < 213403000 &&
+		defined( 'IS_WPCOM' ) && IS_WPCOM && ! wpcom_global_styles_has_been_used( $blog_id )
```
BogdanUngureanu: Shouldn't this condition be inverted? From what I understand, this it should **not** limit the GS if there is no usage of GS.

Copons: Yes, yes it should!

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -256,6 +289,39 @@ function wpcom_global_styles_in_use() {
 	return $global_styles_in_use;
 }
 
+/**
+ * Checks if Global Styles has been customized by the given site in the past.
+ *
+ * @param  int $blog_id Blog ID.
+ * @return bool Whether Global Styles have been customized.
+ */
+function wpcom_global_styles_has_been_used( $blog_id = 0 ) {
```
BogdanUngureanu: I have mixed feelings about using this here. This might cause some performance issues since this code should run on the front-end too, which means regular site visitors will hit this code path. 

Now, even though this should run for each site only once, it might cause a spike of queries at once, after the code lands. 

Also, since the sticker is added for just one type of site (sites that should be limited), it means that this query will run on every request for the other type of sites, since we are not caching the result for those sites. So, to sum it up:
 - For sites limited this will run only once
 - For sites that are not limited, this code will run on every request since we are not caching the result with a blog sticker.

BogdanUngureanu: Since `get_posts` is a read operation, it's probably cached - although `wpcom_global_styles_in_use_by_wp_global_styles_post` looks a bit heavy (since we are doing a json decode and some array manipulations). My concern is related to `add_blog_sticker` since that should be a write operation (most likely uncached).

Copons: I'm not super concerned about the spike, as TBH I would expect most sites to only have a handful of GS posts.

But you make it a good point: the sticker doesn't seem a good way to store this info, as we need to know both the positive and negative states, not just its existence. 🤔 

I was thinking about site options, but would rather find something more private if possible 🤔

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -34,26 +34,42 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
...
-	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'theme-demo-site', $blog_id ) ) {
+	// Do not limit Global Styles on sites created before we made it a paid feature (2022-12-15),
+	// that had already used Global Styles.
+	if ( wpcom_premium_global_styles_is_site_exempt( $blog_id ) ) {
 		return false;
 	}
 
-	if ( function_exists( 'wpcomsh_is_site_sticker_active' ) && wpcomsh_is_site_sticker_active( 'theme-demo-site' ) ) {
+	// Do not limit Global Styles if the site paid for it.
+	if ( wpcom_site_has_feature( WPCOM_Features::GLOBAL_STYLES, $blog_id ) ) {
```
BogdanUngureanu: Can we move this condition above `wpcom_premium_global_styles_is_site_exempt`? This way, the exempt logic is not executed on sites that have a valid subscription. There's no reason for doing that on them.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -256,6 +281,65 @@ function wpcom_global_styles_in_use() {
...
+	}
+
+	// If the exemption check has already been performed, just return if the site is exempt.
+	if ( wpcom_global_styles_has_blog_sticker( 'wpcom-premium-global-styles-exemption-checked', $blog_id ) ) {
+		return wpcom_global_styles_has_blog_sticker( 'wpcom-premium-global-styles-exempt', $blog_id );
+	}
+
+	// Non-Simple sites are edge cases; other conditions will determine whether they can use GS.
+	if ( ! defined( 'IS_WPCOM' ) || ! IS_WPCOM ) {
+		return false;
```
BogdanUngureanu: I believe this should be `true`. With `false`, it would mean that a Business site that has used GS on their site, it would lose that feature when their subscription expires.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -34,26 +34,42 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
...
+ *
+ * @param string $blog_sticker The blog sticker.
+ * @param int    $blog_id The WPCOM blog ID.
+ * @return bool Whether the site has the blog sticker.
+ */
+function wpcom_global_styles_has_blog_sticker( $blog_sticker, $blog_id ) {
+	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( $blog_sticker, $blog_id ) ) {
+		return true;
+	}
+	if ( function_exists( 'wpcomsh_is_site_sticker_active' ) && wpcomsh_is_site_sticker_active( $blog_sticker ) ) {
```
BogdanUngureanu, line 67: Since we are hardcoding the result for Atomic sites in `wpcom_premium_global_styles_is_site_exempt`, is this code necessary? 🤔

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -34,26 +34,42 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 		}
 	}
 
-	// Do not limit Global Styles on sites created before we made it a paid feature (2022-12-15).
-	if ( $blog_id < 213403000 ) {
+	// Do not limit Global Styles on demo sites.
+	if ( wpcom_global_styles_has_blog_sticker( 'theme-demo-site', $blog_id ) ) {
```
BogdanUngureanu, line 38: Non-blocking: Maybe we can move this condition inside of `wpcom_premium_global_styles_is_site_exempt`? 🤔 In the end, it's a site exception condition.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
dsas: It's necessary so that the check for `theme-demo-site` works.

It's unnecessary for anything else, only four stickers get synced to atomic sites - `woop`, `wpcom-marketplace`, `theme-demo-site` and `difm-lite-in-progress` See:
 * fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Sjcpbz%2Qoybt%2Qfgvpxref.cuc%3Se%3Qp0n1pn5r%23138-og for the list of stickers
 * fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Sjcpbz%2Qoybt%2Qfgvpxref.cuc%3Se%3Qp0n1pn5r%2331-og for `add_block_sticker`
 * https://github.com/Automattic/wpcomsh/blob/7e9d0d35d8359b5866fd68cde364171cac016a3c/feature-plugins/masterbar.php#L213 for `wpsomsh_is_site_sticker_active`

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
Copons: I'd rather keep it outside as an exception to the rule, rather than being part of the exemption logic.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -238,14 +270,7 @@ function wpcom_global_styles_in_use() {
 
 	$user_cpt = WP_Theme_JSON_Resolver::get_user_data_from_wp_global_styles( wp_get_theme() );
 
-	if ( ! isset( $user_cpt['post_content'] ) ) {
-		do_action( 'global_styles_log', 'global_styles_not_in_use' );
```
mmtr, line 242: We are no longer triggering this action. Is that intentional? I think it was used to add some debug traces.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -256,6 +281,69 @@ function wpcom_global_styles_in_use() {
...
+		return true;
+	}
+
+	switch_to_blog( $blog_id );
+
+	// If the current user cannot modify the `wp_global_styles` CPT, the exemption check is not needed;
+	// other conditions will determine whether they can use GS.
+	$wp_global_styles_cpt = get_post_type_object( 'wp_global_styles' );
+	if ( ! current_user_can( $wp_global_styles_cpt->cap->publish_posts ) ) {
+		return false;
```
mmtr, line 319: I think we should run `restore_current_blog` here as well.

mmtr (commented):
Nice, it works quite well! Just left a couple of comments.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
Copons: Uh. Good catch, it's 1) not intentional, and 2) no idea how it got lost in the copy-paste. 😮

Copons: Actually! Maybe it was intentional. 😄 

The same `not_in_use` is logged in the caller, just a bit later, by the `wpcom_global_styles_in_use()` function.

I think that this is the most appropriate behaviour, as it records whether GS is **currently** in use.

If I put this additional log back here, it would also record whether GS is (not) used by all GS posts on the site.
We can discuss if it's interesting to know, but it's different than what we were recording before this change. 🙂

mmtr (approved):

PR #73457, opened 2023-02-16
Add Jetpack CI infrastructure to target Jetpack tests against staging

dpasque:
Related to #

## Proposed Changes

See discussion here: pd5faL-rf-p2

In short, we want to expand the Jetpack E2E testing we're doing on WPCOM!

Currently, we just use the `TEST_ON_JETPACK` env var flag to switch to an account that has a site hosted on Pressable, but then Jetpack connected to be used in Calypso.

As discussed in the P2, we want to also start running tests against the Jetpack installation on WPCOM, especially in its staging mode during releases.

So, we've made a new env var: `JETPACK_TARGET`. It can be one of three values:
1. `wpcom-production`: this is the jetpack install that any ye-old WPCOM site is using! The baseline, the deafult.
2. `wpcom-staging`: during jetpack releases on WPCOM, a couple of sites have access to the new release in its staging form. We target that staging version on WPCOM here.
3. `not-wpcom`: this represents what we used to target with jetpack testing: a site hosted elsewhere (not on WPCOM) that we've since Jetpack connected and are testing through Calypso. 

There are new CI builds to represent these different options!

We'll be enabling more tests and users for these different cases in a future PR!

## Testing Instructions

- [ ] No TeamCity errors
- [ ] The previous Jetpack tests should continue to target a site hosted on Pressable and pass.

## Pre-merge Checklist

### DIFF
--- a/.teamcity/_self/projects/WPComTests.kt
+++ b/.teamcity/_self/projects/WPComTests.kt
@@ -49,8 +49,17 @@ object WPComTests : Project({
 	// Editor Tracking Edge
 	buildType(editorTrackingBuildType("desktop", "c752ca9a-e77d-11ec-8fea-0242ac120002", atomic=false, edge=true));
 
-	buildType(jetpackPlaywrightBuildType("desktop", "68fe6336-5869-4244-b236-cca23ba03487"));
-	buildType(jetpackPlaywrightBuildType("mobile", "a80b5c10-1fef-4c7f-9e2c-5c5c30d637c8"));
+	// Jetpack for Connected Non-WPCOM Sites
+	buildType(jetpackPlaywrightBuildType("desktop", "68fe6336-5869-4244-b236-cca23ba03487", "not-wpcom"));
+	buildType(jetpackPlaywrightBuildType("mobile", "a80b5c10-1fef-4c7f-9e2c-5c5c30d637c8", "not-wpcom"));
+
+	// Jetpack for WPCOM Sites Running Staging
+	buildType(jetpackPlaywrightBuildType("desktop", "3007d7a1-5642-4dbf-9935-d93f3cdb4dcc", "wpcom-staging"));
+	buildType(jetpackPlaywrightBuildType("mobile", "ccfe7d2c-8f04-406b-8b83-3db6c8475661", "wpcom-staging"));
+
+	// Jetpack for WPCOM Sites Running Prod
+	buildType(jetpackPlaywrightBuildType("desktop", "de6df2e1-3bad-46a4-9764-9c7e0974d4ff", "wpcom-production"));
+	buildType(jetpackPlaywrightBuildType("mobile", "79479054-e30e-4177-937f-4197c4846a0f", "wpcom-production"));
 
 	buildType(I18NTests);
 	buildType(P2E2ETests)
@@ -172,12 +181,12 @@ fun editorTrackingBuildType( targetDevice: String, buildUuid: String, atomic: Bo
 	)
 }
 
-fun jetpackPlaywrightBuildType( targetDevice: String, buildUuid: String): E2EBuildType {
+fun jetpackPlaywrightBuildType( targetDevice: String, buildUuid: String, jetpackTarget: string = "wpcom-production" ): E2EBuildType {
 	return E2EBuildType (
 		buildId = "WPComTests_jetpack_Playwright_$targetDevice",
 		buildUuid = buildUuid,
-		buildName = "Jetpack E2E Tests ($targetDevice)",
-		buildDescription = "Runs Jetpack E2E tests as $targetDevice",
+		buildName = "Jetpack E2E Tests [$jetpackTarget] ($targetDevice)",
+		buildDescription = "Runs Jetpack E2E tests as $targetDevice against Jetpack install $jetpackTarget",
 		testGroup = "jetpack",
 		buildParams = {
 			text(
@@ -188,14 +197,18 @@ fun jetpackPlaywrightBuildType( targetDevice: String, buildUuid: String): E2EBui
 				allowEmpty = false
 			)
 			param("env.VIEWPORT_NAME", "$targetDevice")
-			checkbox(
-				name = "env.TEST_ON_JETPACK",
-				value = "true",
-				label = "Test on Jetpack",
-				description = "Use a Jetpack blog to test against",
-				checked = "true",
-				unchecked = "false"
+			select(
+				name = "env.JETPACK_TARGET",
+				value = jetpackTarget,
+				label = 'Jetpack Target',
+				description = 'Which Jetpack install are we targeting through Calypso?',
+				options = [
+					"wpcom-production",
+					"wpcom-staging",
+					"not-wpcom"
+				]
 			)
+
 		},
 		buildFeatures = {},
 		buildTriggers = {
--- a/.teamcity/_self/projects/WPComTests.kt
+++ b/.teamcity/_self/projects/WPComTests.kt
@@ -200,13 +200,13 @@ fun jetpackPlaywrightBuildType( targetDevice: String, buildUuid: String, jetpack
 			select(
 				name = "env.JETPACK_TARGET",
 				value = jetpackTarget,
-				label = 'Jetpack Target',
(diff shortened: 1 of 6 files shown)


dpasque:
Lol, clearly I need to stop using the remote CI servers as my Kotlin compiler 🤦

worldomonation:
> Lol, clearly I need to stop using the remote CI servers as my Kotlin compiler 🤦

Why not? That's what @noahtallen and I both do 🤣

dpasque:
@worldomonation -- it's been a while since I've done one of these PRs, is there any way to see the Kotlin changes for the new build pre-merge? I was in TeamCity looking for this branch, but didn't see them

noahtallen:
> Why not? That's what @noahtallen and I both do 🤣

lmao. On a previous computer, I got it working locally, but I never figured out how to do it again. I should probably take a look again, because it'd be super useful to troubleshoot locally. The main problem is that Gradle or maven or whatever the dependency manager is has to connect to the teamcity server to download the teamcity kotlin library, but that's behind the proxy

noahtallen:
> I was in TeamCity looking for this branch, but didn't see them

Not for new builds, unfortunately :( TeamCity has to know the build exists to be able to trigger it, and it only does that once it's merged to trunk. After that, you can test most changes in the branch

.teamcity/_self/projects/WPComTests.kt
```diff
@@ -49,8 +49,17 @@ object WPComTests : Project({
 	// Editor Tracking Edge
 	buildType(editorTrackingBuildType("desktop", "c752ca9a-e77d-11ec-8fea-0242ac120002", atomic=false, edge=true));
 
-	buildType(jetpackPlaywrightBuildType("desktop", "68fe6336-5869-4244-b236-cca23ba03487"));
-	buildType(jetpackPlaywrightBuildType("mobile", "a80b5c10-1fef-4c7f-9e2c-5c5c30d637c8"));
+	// Jetpack for Connected Non-WPCOM Sites
+	buildType(jetpackPlaywrightBuildType("desktop", "68fe6336-5869-4244-b236-cca23ba03487", "not-wpcom"));
```
noahtallen: non-blocking: we could use named parameters here to make it a bit more readable. E.g. `jetpackTarget ="not-wpcom"`. (https://kotlinlang.org/docs/functions.html#named-arguments)

.teamcity/_self/projects/WPComTests.kt
```diff
@@ -188,14 +197,18 @@ fun jetpackPlaywrightBuildType( targetDevice: String, buildUuid: String): E2EBui
...
-				unchecked = "false"
+			select(
+				name = "env.JETPACK_TARGET",
+				value = jetpackTarget,
+				label = "Jetpack Target",
+				description = "Which Jetpack install are we targeting through Calypso?",
+				options = listOf(
+					"wpcom-production",
+					"wpcom-staging",
+					"not-wpcom"
```
noahtallen: Potentially better name could be `remote-site`?

packages/calypso-e2e/src/env-variables.ts
```diff
@@ -103,6 +107,18 @@ const castKnownEnvVariable = ( name: string, value: string ): EnvVariableValue =
 					`Invalid CALYPSO_BASE_URL value: ${ output }.\nYou must provide a valid URL.`
 				);
 			}
+			break;
+		}
+		case 'JETPACK_TARGET': {
+			const supportedValues: JetpackTarget[] = [ 'not-wpcom', 'wpcom-production', 'wpcom-staging' ];
+			if ( ! supportedValues.includes( output as JetpackTarget ) ) {
```
noahtallen, line 118: Technically, `! supportedValues.includes( output )` is testing that `output` is _not_ the type `JetpackTarget`. (so casting doesn't seem necessary)

noahtallen (commented):
This looks pretty solid to me. I think my main concern is just having a lot more builds in our queue. How often would we run these tests? If these will have to run frequently, we'll probably need to get more agents eventually

packages/calypso-e2e/src/env-variables.ts
dpasque: I very much agree! But the TypeScript linter doesn't seem to agree with us ☹️ 

Screen Shot 2023-02-16 at 5 10 02 PM

I figured it was just best to keep the linter happy, even though it feels silly. 🤷  What do you think? I could also cast it to any?

.teamcity/_self/projects/WPComTests.kt
dpasque: Oh, that's much better lololol

dpasque: My faux-enum writing skills never really graduated past the third grade 🤣

dpasque:
Thanks for the review @noahtallen 😀 

>I think my main concern is just having a lot more builds in our queue. How often would we run these tests? If these will have to run frequently, we'll probably need to get more agents eventually

Not very often at all! Each will probably be run only one time each week, literally! They're mostly just used during the Jetpack weekly upgrade. I reckon that shouldn't clog the queue too much?

Also, FYI I broke out several builds with the different parameters mostly to just make it extra simple for the Jetpack folks to be able to run the tests (just go right to the build and click run, as opposed to tweaking a bunch of args). If that's too verbose though, we can collapse these build down.

.teamcity/_self/projects/WPComTests.kt
dpasque: Fixed in [6acd928](https://github.com/Automattic/wp-calypso/pull/73457/commits/6acd928c6cfc133d195d5578dc07e44821eef125)

.teamcity/_self/projects/WPComTests.kt
dpasque: Fixed in [6acd928](https://github.com/Automattic/wp-calypso/pull/73457/commits/6acd928c6cfc133d195d5578dc07e44821eef125)

worldomonation:
> How often would we run these tests?

To add a bit more context, this is for pd5faL-qu-p2 and this particular build will be triggered only once a week or so, manually.

packages/calypso-e2e/src/env-variables.ts
noahtallen: I think we might as well let `supportedValues` be a plain string array unless having the type info is otherwise useful (for that variable)

noahtallen:
Cool!

test/e2e/specs/appearance/theme__details-preview-activate.ts
```diff
@@ -20,9 +20,10 @@ import { Browser, Page } from 'playwright';
 declare const browser: Browser;
 
 describe( DataHelper.createSuiteTitle( 'Theme: Preview and Activate' ), () => {
-	const accountName = envVariables.TEST_ON_JETPACK
-		? 'jetpackUser'
-		: getTestAccountByFeature( envToFeatureKey( envVariables ) );
+	const accountName =
+		envVariables.JETPACK_TARGET === 'remote-site'
+			? 'jetpackUser'
+			: getTestAccountByFeature( envToFeatureKey( envVariables ) );
```
worldomonation, line 26: **thought:non-blocking** - it occurred to me that we rely on this a lot, but do we actually have unit tests within `calypso-e2e` to verify this behavior?

worldomonation (approved):

packages/calypso-e2e/src/env-variables.ts
dpasque: > unless having the type info is otherwise useful (for that variable)

Ah, yeah, the main reason I type it as such is that I was hoping for transpile-time safety for that variable in case we update the options in the future. Otherwise, we could end up in a scenario where we change the typing, but that array sneaks through, and we accept (at runtime) options that aren't actually valid. 

The cast definitely looks weird, but my personal preference is currently to take the type safety at the expense of the weird-looking cast! But, definitely not married to that, so I'd be glad to swing the tradeoff back the other way!

dpasque: Also, I'm going to merge this to unblock a few things, but I'm glad to go back and update this too if we decide that is better! Just let me know 😄

Issue #72133, opened 2023-01-16
Stats: Extend horizontal bar component to display multiple items in one row


grzegorz-cp:
The `Tags & categories` component needs to display multiple tags in one row. 
This is currently implemented in the components without the horizontal bars, but the bars were originally designed to display only one label. Therefore, extend the horizontal bar component.

# Development
- [x] extend the horizontal bar component to accept multiple labels and match the design
- [ ] work with the designers to assess if this approach works well with a long bar background
- [x] work with the design to verify if a large number of labels should use fading effect on the right side or an ellipsis

# Design
![Image](https://user-images.githubusercontent.com/112354940/212594294-091bf817-8e90-4f96-8db0-2070f338cf62.png)

![Image](https://user-images.githubusercontent.com/112354940/212594303-0f59521d-5c66-472c-b198-091920e3e25f.png)

![Image](https://user-images.githubusercontent.com/112354940/212594314-6cd56fcc-dd80-4f0f-bca1-613ede5fbb7c.png)

grzegorz-cp:
This task will need a follow-up to determine if the expanded list of tags should have horizontal bars. I used a default option that showed the bars for children, but the design above is different. I'm creating a task addressing those kinds of differences on the entire Insights page [#73562](https://github.com/Automattic/wp-calypso/issues/73562).

PR #73462, opened 2023-02-16
Stats: Updating Tags module

grzegorz-cp:
Related to #72134, #72133

## Proposed Changes

* modernising `Tags & categories` module
* extending horizontal bar component to display an array of items for a label (resulting in a concatenated label with icons)
* refactoring the way the icons are generated for the left side of the labels
* accepting a function for the left side icon (to render it for child items)
* extending horizontal bar groups to allow placement of group chevron on either the left or right side of the label (previously only right side)
* removing obsolete value sorting for the Search terms module
* fixing label ellipsis for group labels
* if a child item doesn't have any value, apply the parent's value (grouped tags are grouped because of the value but don't have the value themselves) 

## Testing Instructions

* go to the live branch and the `Insights` page
* verify that the current `Tags & categories` renders correctly and you can expand collapsed groups
* apply feature flag to the URL `?flags=stats/insights-page-grid`
* verify that the new module with horizontal bars loads correctly and collapsing/expanding works
* check modules on the `Traffic` page for potential missing avatars or unnecessary spacing between items that haven't been there before

| Before | After |
| --- | --- |
| SCR-20230217-u0r | SCR-20230217-tzd |

## Pre-merge Checklist


keoshi (approved):
This looks great and works flawlessly in my testing!

keoshi:
@jeffgolenski @ederrengifo This PR made me look at tags and categories differently, and I'm kind of confused how it works and what it all means. It would be wonderful if we could challenge the current design with user interviews.

Adding a few examples below:

#### Lots of sites use either categories OR tags; should we let them toggle between them or even disable one of these taxonomies?

image


#### Why do some tags appear in groups and others dont?

image

#### When you expand a group, all the tags come up with the same amount of views

image

grzegorz-cp:
Good idea about challenging the design @keoshi. The tags are a bit confusing because they group items by value; this doesn't happen for any other widget. I also noticed on some posts that some tags are often grouped, and the same number of views reflects it, but I'm sure it's not the case for every grouped tag.
In the previous implementation, even when you expand a group, only the parent has a number of views. However, with horizontal bars, it's strange to have an expanded list without any background.

client/my-sites/stats/stats-list/stats-list-card.jsx
```diff
@@ -82,18 +83,43 @@ const StatsListCard = ( {
...
+		let leftSideItem; // undefined value avoids rendering an empty node if nothing generates the output
+
+		// left icon visible for avatars, contry flags or tags and categories.
+		if ( item?.countryCode ) {
+			leftSideItem = <StatsListCountryFlag countryCode={ item.countryCode } />;
+		} else if ( showLeftIcon && item?.icon ) {
+			leftSideItem = <StatsCardAvatar url={ item?.icon } altName={ item?.label } />;
+		}
+
+		if ( Array.isArray( item?.label ) ) {
```
kangzj: Should we `else if` here, otherwise the code here has the possibility of overriding the previous logic. Or is this intended?

client/my-sites/stats/stats-list/stats-list-card.jsx
```diff
@@ -82,18 +83,43 @@ const StatsListCard = ( {
...
+			// tags without children have one item in its label array;
+			// tags with children have them duplicated in this label array - chevron is added and label is constructed by concatenating items.
+			if ( item?.label?.length === 1 ) {
+				leftSideItem = getLabelIcon( item?.label[ 0 ]?.labelIcon );
```
kangzj: theoretically the first question mark could be saved, because it's already checked in the if condition.

client/my-sites/stats/stats-list/stats-list-card.jsx
```diff
@@ -82,18 +83,43 @@ const StatsListCard = ( {
...
 
+	const getLabelIcon = ( labelIconType ) => {
+		if ( labelIconType === 'folder' ) {
+			return <Icon className="stats-icon" icon={ file } size={ 22 } />;
+		} else if ( labelIconType === 'tag' ) {
+			return <Icon className="stats-icon" icon={ tag } size={ 22 } />;
+		}
+	};
+
+	const outputLeftItem = ( item ) => {
```
kangzj: Wondering if we could use a better name for the function, because it doesn't output anything IMO.

client/my-sites/stats/stats-list/stats-list-card.jsx
```diff
@@ -82,18 +83,43 @@ const StatsListCard = ( {
...
+
+		if ( Array.isArray( item?.label ) ) {
+			// tags without children have one item in its label array;
+			// tags with children have them duplicated in this label array - chevron is added and label is constructed by concatenating items.
+			if ( item?.label?.length === 1 ) {
+				leftSideItem = getLabelIcon( item?.label[ 0 ]?.labelIcon );
+			}
+			// else {} -> either unsupported icon or and an error with labels
+		} else if ( item?.labelIcon ) {
+			leftSideItem = getLabelIcon( item?.labelIcon );
```
kangzj: theoretically the question mark could be saved, because it's already checked in the if condition.

kangzj (approved):
It works as described. Left a couple of comments but no blockers. 👍

client/my-sites/stats/stats-list/stats-list-card.jsx
grzegorz-cp: It generates JSX for the left side of the label. I named it this way to match the right side. Do you think that `generateLeftItem` and `generateRightItem` would be better?

kangzj: `get` would do too. Either works for me.

kangzj (approved):
LGTM

PR #73464, opened 2023-02-16
Tags page: Add top section of public tags page

roo2:
Add a `/tags` page to replace the current page at wordpress.com/tags

see project thread: pe7F0s-Ed-p2

Only includes the page header and the first part of the page. I will add the alphabetic tags section in a follow up PR:

Screen Shot 2023-03-28 at 2 33 58 pm

This page is made available to search engines via SSR. see SSR docs 
https://github.com/Automattic/wp-calypso/blob/trunk/docs/isomorphic-routing.md
https://github.com/Automattic/wp-calypso/blob/trunk/docs/server-side-rendering.md

### Testing instructions
* The special page header should only be applied to the new `/tags` page. It should not affect headers on other pages such as `/read` (normal logged in masterbar) or `/themes` (when logged out).
* There is a logged out version and a logged in version of the page, with different headers. p1679967423624639-slack-C03NLNTPZ2T
* It is possible to see the server side rendered version of the page by doing _View Page Source_ and searching the generated HMTL


davemart-in:
Nice work! A couple of notes based on your screenshot:

- Let's cap that top section to the top 6 tags (2 columns of 3).
- Let's use a little bit of JS to ensure that these are title case (not uppercase). Note: Unfortunately CSS text-transform can't do title case.

roo2:
current state 
**Logged out:**
Screen Shot 2023-03-28 at 2 33 58 pm


**Logged in:**
Screen Shot 2023-03-28 at 2 36 24 pm

davemart-in:
Here's a list of the changes I made:

- Updated the BG color of the page to white
- Changed “Topics” to be “Tags”
- Updated padding in header
- Added header BG
- Updated line colors
- Added container padding
- Add “Trending” header
- Fixed font weight of tags
- Fixed padding between tag and count

Here's a preview:

CleanShot 2023-03-28 at 10 08 30

A couple of notes:

- @roo2, it looks like there is a lag before you’re hiding the logged out master bar and showing the logged in one where you see a flash of the logged out one.
- We likely need to add a `.tags` class to the body for this page and then prefix it to all of the CSS for this page so that these CSS styles don’t leak to other pages.
- @ollierozdarz, It doesn’t look like we are loading Recoleta font-weight 400. The current H1 is a tad chunkier than your designs.

roo2:
looking good, thank you @davemart-in! 👌

roo2:
I changed how data was loaded so that trending tags would be loaded when the page is rendered with SSR

roo2:
> it looks like there is a lag before you’re hiding the logged out master bar and showing the logged in one where you see a flash of the logged out one.

@davemart-in I looked in to this but at least when I run it locally, I see the same behavior on the /plugins and /themes pages. I think it's something we haven't been able to solve due to Server Side Rendering being enabled, in the server side rendering context, the user always appears to be logged out, so it first generated the logged out version of the page, then in the browser it gets re-rendered with a logged in user.

client/reader/tags-page/controller.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const tagsListing = ( context: PageJSContext, next: () => void ) => {
+	context.headerSection = renderHeaderSection();
+	context.primary = <TagsPage trendingTags={ context.params.trendingTags } />;
+	next();
+};
+
+function renderHeaderSection() {
+	return (
+		<div>
```
tyxla: Is the `div` necessary? If not, we could use a React.Fragment instead: `<></>`.

client/reader/tags-page/main.tsx
```diff
@@ -0,0 +1,18 @@
...
+import './style.scss';
+
+interface Props {
+	trendingTags: TagResult[];
+}
+
+export default function TagsPage( { trendingTags }: Props ) {
+	return (
+		<div className="tags-page">
+			<h4>Trending</h4>
```
tyxla: Are we intentionally skipping the localization of this string?

If we decide to localize it, it might need some additional context or comment to help the translators with what it refers to.

client/reader/tags-page/trending-tags.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+	// put the data into a format that can easily be rendered as a two column table.
+	const tagRows: TagResult[][] = [];
+	for ( let i = 0; i < trendingTags.length; i += 2 ) {
+		const tagOne = trendingTags[ i ];
+		const tagTwo = trendingTags[ i + 1 ];
+		tagRows.push( [ tagOne, tagTwo ] );
+	}
+
+	const toTitleCase = ( text: string ) => {
```
tyxla: Micro-optimization: it might be a good idea to move the generic function definitions out of this function, so they would be defined only once.

client/reader/tags-page/trending-tags.tsx
```diff
@@ -0,0 +1,54 @@
...
+	};
+
+	const abbreviateNumber = ( number: number ) => {
+		if ( number >= 1000000 ) {
+			return ( number / 1000000 ).toFixed( 0 ) + 'm';
+		} else if ( number >= 1000 ) {
+			return ( number / 1000 ).toFixed( 0 ) + 'k';
+		}
+		return number.toString();
+	};
```
tyxla: Does it make sense to use the existing `formatNumberCompact()`?

https://github.com/Automattic/wp-calypso/blob/0b0b3078d05bf7cf94be2cfc034db33539dbc59f/client/lib/format-number-compact/index.js#L11-L37

client/reader/tags-page/trending-tags.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+	const renderTagRow = ( tag: TagResult ) => (
```
tyxla: This could as well be extracted to a separate component that accepts props instead of a single argument.

client/reader/tags-page/trending-tags.tsx
```diff
@@ -0,0 +1,54 @@
...
+		<div className="trending-tags__column" key={ tag.tag.slug }>
```
tyxla: It might be a good opportunity for the component to accept `title`, `slug`, and `count` in order to simplify the signature of the function / component. Right now it expects a complex object, and it could be simplified to accept primitive values.

client/sections.js
```diff
@@ -336,7 +336,16 @@ const sections = [
 	},
 	{
 		name: 'reader',
-		paths: [ '/tags', '/tag' ],
+		paths: [ '/tags' ],
+		module: 'calypso/reader/tags-page',
```
tyxla: Nit: Is there a reason not to name the module just `tags`? The `-page` part feels redundant.

davemart-in:
> I looked in to this but at least when I run it locally, I see the same behavior on the /plugins and /themes pages. I think it's something we haven't been able to solve due to Server Side Rendering being enabled, in the server side rendering context, the user always appears to be logged out, so it first generated the logged out version of the page, then in the browser it gets re-rendered with a logged in user.

One possible idea. Could we apply `opacity: 0;` to the whole header on page load then after the nav changes (or doesn't change), change it to `opacity: 1` with `transition: opacity 100ms ease-out;` also applied?

client/reader/tags-page/trending-tags.tsx
roo2: perfect!

client/reader/tags-page/main.tsx
roo2: nope, just missed it, oops!

roo2:
@tyxla thank you for the reviews I addressed your feedback and simplified/improved the page

roo2:
@davemart-in I've fixed the padding of the header so that the page doesn't jump while the header is loading. I'm a bit hesitant to use the opacity fix just because it adds complexity and we don't use anything like that in other places.

Do you think this is a good enough experience?

https://user-images.githubusercontent.com/22446385/228732308-7927ebf3-b6d1-4903-8320-d0f201c95629.mov


Compare the /themes page

https://user-images.githubusercontent.com/22446385/228733063-8fc111cf-efea-40a3-816e-b706af4da6f5.mov

davemart-in:
> Do you think this is a good enough experience?

I don't think it's a blocker.

eoigal:
WFM, I just have a question about the header. When you scroll down the with icons get lost against the white background; 
Screenshot 2023-03-30 at 22 18 10

roo2:
good catch @eoigal I updated the menu it to no longer be fixed position so it stays in place on scroll

eoigal (approved):
Nice job @roo2 - I think I'd prefer if the logged in header kept the normal black admin bar and leave the logged out as it is here. This isn't a blocker for me though.

roo2:
thanks @eoigal I'll raise it with Ollie but merge this now

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7904154

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7904154).

PR #73476, opened 2023-02-17
Design picker: Use variation discs to change the theme thumbnail

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1658

## Proposed Changes

Changes the behavior of the style variations discs in the onboarding's design picker so they replace the theme thumbnail:

https://user-images.githubusercontent.com/1233880/220145265-771f2299-9c6a-4d81-9c87-95881d2ac904.mov

Some notes about the new behavior:

- The selected style variation is highlighted with a blue border
- The first disc is the default variation (free) for regular themes, and the preselected variation (paid) for virtual themes
- The "Free" / "Premium" badges are dynamically set based on the selected variation
- Clicking on the "+X" button (only visible when there are more than 4 variations) goes to the next step (design preview) with the default variation

## Testing Instructions

- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>&flags=virtual-themes/onboarding`
- For regular themes, make sure the first disc represents the default variation
- For virtual themes, make sure the first disc represents the preselected variation
- Use the discs to change the selected variation
- Make sure the thumbnails changes accordingly
- Make sure the "Free" / "Premium" badges change accordingly
- Click on the thumbnail
- Make sure the design preview keeps the same style variation selected
- Go back
- Click on the "+X" button on a theme with more than 4 variations
- Make sure the design preview is opened with the default style variation


davewhitley:
The animation is a bit slow switching between styles. I'd go with 300ms instead of 1.5s. Also, it seems like the thumbnails fade to white before transitioning to the next style. Does that make sense? So, after clicking on a disc, the thumbnail fades to white, then the new style fades in from white. I think the animation would look better if it didn't fade to white first and just went straight to the new image. You can slow the animation down to 3 seconds to see what I mean.

packages/design-picker/src/components/style-variation-badges/style.scss
```diff
@@ -42,3 +43,9 @@
 		box-shadow: none;
 	}
 }
+
+.style-variation__badge-is-selected {
+	> * {
+		outline: 2px solid var(--studio-blue-30);
+	}
+}
```
Copons, line 58: Should we remove the hover styles (fade, pointer cursor) from the selected disc to convey that it's "inactive"?

It's not the biggest of deals, but I wondered if clicking on a selected disc was supposed to do something (e.g. deselect it?).

packages/design-picker/src/components/style-variation-badges/index.tsx
```diff
@@ -10,23 +10,42 @@ interface BadgesProps {
...
+		return [
+			...( firstVariationToShow ? [ firstVariationToShow ] : [] ),
+			...( firstVariationToShow
+				? variations
+						.filter( ( variation ) => variation.slug !== firstVariationToShow.slug )
+						.slice( 0, maxVariationsToShow - 1 )
+				: [] ),
+			...( ! firstVariationToShow ? variations.slice( 0, maxVariationsToShow ) : [] ),
+		];
+	}, [ variations, maxVariationsToShow, firstVariationToShow ] );
```
Copons: I wonder if we could do this whole thing in `UnifiedDesignPicker` instead?

It would save us a prop (`firstVariationToShow`, which is a bit of a mouthful), but at the same time it makes sense to keep the logic in the `Badges`.

Copons: Also how do you feel about making this more readable by separating the two cases?
E.g.
```js
if ( firstVariationToShow ) {
  return [
    firstVariationToShow,
    ...variations
      .filter( ( variation ) => variation.slug !== firstVariationToShow.slug )
      .slice( 0, maxVariationsToShow - 1 )
  ];
}

return variations.slice( 0, maxVariationsToShow );
```

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -166,16 +169,24 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
...
-		is_premium: isPremium = false,
-		preselected_style_variation,
-		style_variations = [],
-	} = design;
-	const shouldUpgrade = isPremium && ! isPremiumThemeAvailable && ! hasPurchasedTheme;
-	const currentSiteCanInstallWoo = currentPlanFeatures?.includes( FEATURE_WOOP ) ?? false;
+	const [ _selectedStyleVariation, setSelectedStyleVariation ] = useState< StyleVariation >();
 
+	const selectedStyleVariation = _selectedStyleVariation ?? preselected_style_variation;
+	const currentSiteCanInstallWoo = currentPlanFeatures?.includes( FEATURE_WOOP ) ?? false;
```
Copons, line 178: Why is GH marking this as changeeeeeeeedddddddddd 😱 😭 💢

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -166,16 +169,24 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
...
 	const designIsBundledWithWoo = design.is_bundled_with_woo_commerce;
+	const title =
+		selectedStyleVariation && selectedStyleVariation.slug !== 'default'
+			? `${ design.title } – ${ selectedStyleVariation.title }`
+			: design.title;
+	let isPremium =
```
Copons: Uber nitpick: let's add an emtpy line above the `let isPremium` to visually separate it from `title`.

They are both slightly complicated conditionals and are already hard enough to read. 🙂

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -199,7 +210,10 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 			badge = <WooCommerceBundledBadge />;
 		} else if ( isPremium ) {
 			const toolTip =
-				shouldUpgrade && preselected_style_variation
+				shouldUpgrade &&
+				( is_virtual || ! design.is_premium ) &&
+				selectedStyleVariation &&
+				selectedStyleVariation?.slug !== 'default'
```
Copons: How do you feel about moving this out of here, to make it more readable?
E.g.

```js
const shouldShowTooltip =
  shouldUpgrade &&
  ( is_virtual || ! design.is_premium ) &&
  selectedStyleVariation &&
  selectedStyleVariation?.slug !== 'default';

// ...

const tooltip = shouldShowTooltip ? __( 'Unlock this style...' ) : undefined;
```

Copons (commented):
This looks really well @mmtr!

I've added a bunch of comments, for the most part about personal preferences, but otherwise this LGTM from a code standpoint.

I have a question about the behaviour of the selected disc (in a comment), and a more general doubt about how this affects style variation Virtual Themes, now that we are moving to A/B them.

It feels somewhat weird to me that the user can end up in a state with multiple cards showing the same theme-variation combination.

E.g. (look at TT3)
| Initial state | After the user changed selection |
| - | - |
| Screenshot 2023-02-17 at 18 07 02 | Screenshot 2023-02-17 at 18 07 07 |

I'm tempted to say this is not a big deal as it requires a user action that makes the UX clear anyway.
But I'm also not an expert on stores and how they handle product variations.

cc @davewhitley

packages/design-picker/src/components/style-variation-badges/style.scss
mmtr: Make sense! Done in the last update.

packages/design-picker/src/components/style-variation-badges/index.tsx
mmtr: > it makes sense to keep the logic in the `Badges`.

Yeah, I think so. The logic in `UnifiedDesignPicker` is already complex enough that I want to avoid making it even more complex if possible. And this logic fits very well in what `Badges` does.

> `firstVariationToShow`, which is a bit of a mouthful

I renamed it to `firstVariation` so it's simpler 🙂 

> Also how do you feel about making this more readable by separating the two cases?

Good suggestion! Done in the last update.

packages/design-picker/src/components/unified-design-picker.tsx
mmtr: I've done that in https://github.com/Automattic/wp-calypso/pull/73476/commits/aab4b8d71896db65cfe0338d49f6e78955788b3c and a few more changes to make all of this much more readable

packages/design-picker/src/components/unified-design-picker.tsx
mmtr: Another good call! Done in https://github.com/Automattic/wp-calypso/pull/73476/commits/65889c840422813619c058da3302712c2cec51a5

mmtr:
> The animation is a bit slow switching between styles. I'd go with 300ms instead of 1.5s.

Much better indeed! Done in https://github.com/Automattic/wp-calypso/pull/73476/commits/88ad0c9652f20d3e2d98d1083891df2eacd017d1

> I think the animation would look better if it didn't fade to white first and just went straight to the new image

I agree, but it's a bit tricky to implement because there are more scenarios to consider.

For instance, what if the new image takes a few seconds to load? The current fade to white is to handle precisely that, in order to give users a sense that the new preview is being loaded.

The problem is that we don't know how long would it take image to load, so we cannot decide beforehand whether we should fade or go straight to the new image, so I think it's better to always fade to white.

packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -198,10 +212,11 @@ const DesignButton: React.FC< DesignButtonProps > = ( {
 		if ( designIsBundledWithWoo ) {
 			badge = <WooCommerceBundledBadge />;
 		} else if ( isPremium ) {
-			const toolTip =
-				shouldUpgrade && preselected_style_variation
-					? __( 'Unlock this style, and tons of other features, by upgrading to a Premium plan.' )
-					: undefined;
+			const showTooltip =
```
dsas: Nitpick: This is perhaps a tiny bit misleading - a tooltip might still be shown even if showTooltip is false. It governs whether to use the custom style variation tooltip rather than whether to use a tooltip at all.

dsas (approved):
Works nicely for me!

packages/design-picker/src/components/unified-design-picker.tsx
mmtr: Good point – changed in https://github.com/Automattic/wp-calypso/pull/73476/commits/5165a6fa5fac9a9783331417a0354e6117c039ec

PR #73481, opened 2023-02-17
Shows the 2023 Plans Grid in selected pages

ddc22:
Fixes Automattic/martech#1523

### Show 2023 Plans on only
- Internal plans page `/plans/<site-slug>`
- Onboarding flow `/start/plans`

---

### Should not show the 2023 plans pages on any other flows or pages.

Eg: `/start/hosting/plans` should have the previous pricing grid

image

## Testing Instructions
* Try the  Internal plans page `/plans/<site-slug>`, Onboarding flow `/start/plans` and make sure the new plans grid is visible
* Try any other flow that includes the plans step and make sure they remain unchanged. i.e. `/start/hosting/`

## Pre-merge Checklist

### DIFF
--- a/packages/calypso-products/src/plans-utilities.ts
+++ b/packages/calypso-products/src/plans-utilities.ts
@@ -48,7 +48,8 @@ export function getTermDuration( term: string ): number | undefined {
 export const redirectCheckoutToWpAdmin = (): boolean => !! JETPACK_REDIRECT_CHECKOUT_TO_WPADMIN;
 
 /**
- * Returns if the 2023 Pricing grid optimized design should be visible or not
+ * Returns if the 2023 Pricing grid feature has been enabled.
+ * Currently this depends on the feature flag and on the locale the user is on
  *
  */
 export const is2023PricingGridEnabled = (): boolean => {
@@ -59,22 +60,31 @@ export const is2023PricingGridEnabled = (): boolean => {
 		return supportedLocales.includes( getLocaleSlug() ?? '' );
 	};
 
-	const isActivePage = (): boolean => {
-		let currentRoutePath = '';
-		currentRoutePath = window.location?.pathname ?? '';
+	return isFeatureFlagEnabled() && isLocaleSupported();
+};
 
-		// Is this the internal plans page /plans/<site-slug> ?
-		if ( currentRoutePath.startsWith( '/plans' ) ) {
-			return true;
-		}
+/**
+ * This function specifically checks if a given route path will display the pricing grid or not.
+ * However the pricing grid needs to be enabled in the first place for this function to return true.
+ *
+ * @param browserWindow Current browser window
+ * @returns true if the pricing grid maybe shown in a given page
+ */
+export const is2023PricingGridActivePage = (
+	browserWindow: Window & typeof globalThis
+): boolean => {
+	const currentRoutePath = browserWindow.location.pathname ?? '';
+	const isPricingGridEnabled = is2023PricingGridEnabled();
 
-		// Is this the onboarding flow?
-		if ( currentRoutePath.startsWith( '/start/plans' ) ) {
-			return true;
-		}
+	// Is this the internal plans page /plans/<site-slug> ?
+	if ( currentRoutePath.startsWith( '/plans' ) ) {
+		return isPricingGridEnabled;
+	}
 
-		return false;
-	};
+	// Is this the onboarding flow?
+	if ( currentRoutePath.startsWith( '/start/plans' ) ) {
+		return isPricingGridEnabled;
+	}
 
-	return isFeatureFlagEnabled() && isLocaleSupported() && isActivePage();
+	return false;
 };
--- a/packages/calypso-products/src/plans-utilities.ts
+++ b/packages/calypso-products/src/plans-utilities.ts
@@ -55,21 +55,15 @@ export const is2023PricingGridEnabled = (): boolean => {
 
 	const isActivePage = (): boolean => {
 		let currentRoutePath = '';
-		if ( typeof window === 'object' && window.location ) {
-			currentRoutePath = window.location.pathname;
-		}
-
-		const [ firstPathPart, secondPathPart ] = ( currentRoutePath || '' )
-			.split( '/' )
-			.filter( ( i ) => i );
+		currentRoutePath = globalThis.location?.pathname ?? '';
 
 		// Is this the internal plans page /plans/<site-slug> ?
-		if ( firstPathPart === 'plans' ) {
+		if ( currentRoutePath.startsWith( '/plans' ) ) {
 			return true;
 		}
 
-		// Is this the onboarding flow /start/plans?
-		if ( firstPathPart === 'start' && secondPathPart === 'plans' ) {
+		// Is this the onboarding flow?
+		if ( currentRoutePath.startsWith( '/start/plans' ) ) {
 			return true;
 		}
 
--- a/packages/calypso-products/src/plans-utilities.ts
+++ b/packages/calypso-products/src/plans-utilities.ts
@@ -13,6 +13,12 @@ import {
 	PLAN_TRIENNIAL_PERIOD,
 } from './constants';
 
+declare global {
+	interface Window {
+		location: Location;
+	}
+}
+
 export function isBestValue( plan: string ): boolean {
 	return ( BEST_VALUE_PLANS as ReadonlyArray< string > ).includes( plan );
 }
@@ -55,7 +61,7 @@ export const is2023PricingGridEnabled = (): boolean => {
 
 	const isActivePage = (): boolean => {
 		let currentRoutePath = '';
-		currentRoutePath = globalThis.location?.pathname ?? '';
+		currentRoutePath = window.location?.pathname ?? '';
 
 		// Is this the internal plans page /plans/<site-slug> ?
 		if ( currentRoutePath.startsWith( '/plans' ) ) {
--- a/packages/calypso-products/src/plans-utilities.ts
+++ b/packages/calypso-products/src/plans-utilities.ts
@@ -41,10 +41,15 @@ export function getTermDuration( term: string ): number | undefined {
 
 export const redirectCheckoutToWpAdmin = (): boolean => !! JETPACK_REDIRECT_CHECKOUT_TO_WPADMIN;
 
-export const is2023PricingGridEnabled = () => {
-	const supportedLocales = [ 'en', 'en-gb', 'es' ];
+export const is2023PricingGridEnabled = (): boolean => {
+	const isFeatureFlagEnabled = (): boolean => isEnabled( 'onboarding/2023-pricing-grid' );
 
-	const isPlans2023ActivePage = () => {
+	const isLocaleSupported = (): boolean => {
+		const supportedLocales = [ 'en', 'en-gb', 'es' ];
+		return supportedLocales.includes( getLocaleSlug() ?? '' );
+	};
+
+	const isActivePage = (): boolean => {
 		const currentRoutePath = window.location.pathname;
 
 		// Is this the internal plans page /plans/<site-slug> ?
@@ -64,9 +69,5 @@ export const is2023PricingGridEnabled = () => {
 		return false;
 	};
 
-	return (
-		isEnabled( 'onboarding/2023-pricing-grid' ) &&
-		supportedLocales.includes( getLocaleSlug() ?? '' ) &&
-		isPlans2023ActivePage()
-	);
+	return isFeatureFlagEnabled() && isLocaleSupported() && isActivePage();
 };
--- a/packages/calypso-products/src/plans-utilities.ts
+++ b/packages/calypso-products/src/plans-utilities.ts
@@ -43,8 +43,30 @@ export const redirectCheckoutToWpAdmin = (): boolean => !! JETPACK_REDIRECT_CHEC
 
 export const is2023PricingGridEnabled = () => {
 	const supportedLocales = [ 'en', 'en-gb', 'es' ];
+
+	const isPlans2023ActivePage = () => {
+		const currentRoutePath = window.location.pathname;
+
+		// Is this the internal plans page /plans/<site-slug> ?
+		if (
+			currentRoutePath.includes( 'plans' ) &&
+			! currentRoutePath.includes( 'start' ) &&
+			! currentRoutePath.includes( 'setup' )
+		) {
+			return true;
+		}
+
+		// Is this the onboarding flow ?
+		if ( currentRoutePath.includes( '/start/plans' ) ) {
+			return true;
+		}
+
+		return false;
+	};
+
 	return (
 		isEnabled( 'onboarding/2023-pricing-grid' ) &&
-		supportedLocales.includes( getLocaleSlug() ?? '' )
+		supportedLocales.includes( getLocaleSlug() ?? '' ) &&
+		isPlans2023ActivePage()
 	);
 };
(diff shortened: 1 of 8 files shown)


dzver (commented):
I tried booting calypso with this change and I see this:

Screenshot 2023-02-17 at 17 31 56

packages/calypso-products/src/plans-utilities.ts
```diff
@@ -41,10 +41,33 @@ export function getTermDuration( term: string ): number | undefined {
...
+		const currentRoutePath = window.location.pathname;
+
+		// Is this the internal plans page /plans/<site-slug> ?
+		if (
+			currentRoutePath.includes( 'plans' ) &&
+			! currentRoutePath.includes( 'start' ) &&
+			! currentRoutePath.includes( 'setup' )
+		) {
+			return true;
+		}
```
chriskmnds: If I'm reading this correctly, it assumes that a site slug cannot contain the words `start` or `setup`. It's a fairly ambiguous match using `includes` this way.

Can't a simple regex match work? simple as `/plans\/.+/g`. do we need the other exclusions?

packages/calypso-products/src/plans-utilities.ts
```diff
@@ -41,10 +41,37 @@ export function getTermDuration( term: string ): number | undefined {
...
+
+		// Is this the onboarding flow ?
+		if ( currentRoutePath.includes( '/start/plans' ) ) {
```
chriskmnds: Similarly, I'd opt for a regex pattern.

packages/calypso-products/src/plans-utilities.ts
ddc22: `/plans\/.+/g` will return match true for `/start/plans` right? I want to avoid this. Is it possible to avoid `start` or `setup` but match for any other string to match a siteSlug?

Maybe i should split path parts and then check them to be sure 🤔

chriskmnds: > /plans\/.+/g will return match true for /start/plans right?

You can mark the beginning of the string in your regex e.g. 
match: `'/plans/123'.match(/^\/plans\/.+/g)` 
no match: `'/start/plans/123'.match(/^\/plans\/.+/g)`

and of course you can have negations like you mentioned, but is it necessary? 🤔 

Please read up on regex. try https://regex101.com/

ddc22: ~> but is it necessary?~

~Yes the initial change was decided last Thursday to only be shipped to `/plans` and `/start/plans` since there are several flows that use this component and we want to reduce the area this change effects. Also many other pages were showing sketchy/broken UI similar to what happened in the add domain flow.~

~> Please read up on regex. try https://regex101.com/~

~I can handle most basic regex but need to read up on more advance uses~

~> You can mark the beginning of the string in your regex e.g.~

~So these are a few test cases to match in this specific instance~

~`'/plans'` - should match~
~`'/start/hosting/plans'` - should not match~
~`'/plans/yourgroovydomain.com` - should match~
~`'/plans/myhosting.com` - should match~

I just realised what you said and you are correct, that should be enough 💯 

However, I also added a more straightforward unambiguous way to arrive at the correct paths as well compared to what I did earlier.

https://github.com/Automattic/wp-calypso/blob/8c167ad57b82ad47061e93fc3cab0796020cd121/packages/calypso-products/src/plans-utilities.ts#L56-L77

I am ok to go ahead with any approach. Do let me know your thoughts! 👍

packages/calypso-products/src/plans-utilities.ts
```diff
@@ -41,10 +41,40 @@ export function getTermDuration( term: string ): number | undefined {
...
+		if ( firstPathPart === 'plans' ) {
+			return true;
+		}
+
+		// Is this the onboarding flow /start/plans?
+		if ( firstPathPart === 'start' && secondPathPart === 'plans' ) {
+			return true;
+		}
+
+		return false;
```
chriskmnds: This is unnecessarily complex for what you are describing. What is the point of splitting the string even? 

If regex seems scary, then I'd opt for some native method that does the string matching e.g. 

> Is this the internal plans page /plans/<site-slug> ?

`pathname.startsWith( '/plans' )`

> Is this the onboarding flow /start/plans?

`pathname.startsWith( '/start/plans' )`

But regex match would be my choice of method probably (and some form of caching on the context path)

packages/calypso-products/src/plans-utilities.ts
```diff
@@ -41,10 +41,40 @@ export function getTermDuration( term: string ): number | undefined {
...
+ */
+export const is2023PricingGridEnabled = (): boolean => {
+	const isFeatureFlagEnabled = (): boolean => isEnabled( 'onboarding/2023-pricing-grid' );
+
+	const isLocaleSupported = (): boolean => {
+		const supportedLocales = [ 'en', 'en-gb', 'es' ];
+		return supportedLocales.includes( getLocaleSlug() ?? '' );
+	};
+
+	const isActivePage = (): boolean => {
```
chriskmnds: Probably best to move this function definition to the outer/module scope. No need to be redefining these every time `is2023PricingGridEnabled` gets called. Same for `isLocaleSupported`.

chriskmnds: We are adding complexity to this function with string matching, etc. now. Ideally, we should be considering some form of caching/memoization, cause this will get called a lot on re-rerenders. Can we see what the right way of doing this outside of React? If nothing comes up, then we can memorize the edge/uses.

packages/calypso-products/src/plans-utilities.ts
```diff
@@ -41,10 +41,40 @@ export function getTermDuration( term: string ): number | undefined {
...
+		let currentRoutePath = '';
+		if ( typeof window === 'object' && window.location ) {
+			currentRoutePath = window.location.pathname;
+		}
```
chriskmnds: would this be the same as `window?.location?.pathname ?? ''`

p.s. if we were in all-React mode/usage, then we'd probably be looking at a different implementation, not relying explicitly on `window` (we have state selectors for grabbing the current path from context)

packages/calypso-products/src/plans-utilities.ts
chriskmnds: > I just realised what you said and you are correct, that should be enough 💯

Well, yes and no. `'/plans/123'.match(/^\/plans\/.+/g)` wouldn't match `/plans`. The cases you outline are quite simple, so there isn't a need for anything overly-engineered here, like manual string splitting and things like that. I'd still go for regex as anything regex-based can be easily adjusted to anything that may arise - but maybe just native string matching (like `startsWith` mentioned below) are more than sufficient (and unambiguous) for this case. Also potentially more performant as they probably won't traverse the whole string.

> Yes the initial change was decided last Thursday to only be shipped to /plans and /start/plans since there are several flows that use this component and we want to reduce the area this change effects. Also many other pages were showing sketchy/broken UI similar to what happened in the add domain flow.

I wonder if any caveats with the approach (to restrict the truthness of `is2023PricingGridEnabled` to only when the user is on one of these pages). There is no constrained usage of that method i.e. it can be called from anywhere in the app. Any chance that this may lead to ambiguity in the environment, in general?

packages/calypso-products/src/plans-utilities.ts
chriskmnds: > If regex seems scary

for regex to fully capture both cases you mentioned, this looks like the pattern: `/^\/(plans|start\/plans)(\/.*)?$/gm`. It's simple, just both patterns congested together. The `(\/.*)?` portion is the optional second part (trailing slash and everything after). These look like the cases you capture with the manual string splitting.

`/plans` match
`/plans/` match
`/plans/foo` match
`/plansfoo` ~no match~
`plansfoo` ~no match~

`/start/plans` match
`/start/plans/` match
`/start/plans/foo` match
`/start/plansfoo` ~no match~
`start/plansfoo` ~no match~

`/` ~no match~
`/foo` ~no match~
`/foo/bar` ~no match~

https://regex101.com/r/fgvoLz/2

ddc22: > If regex seems scary

Haha well not scary, but, There is just this nagging feeling I might have missed something when I put into regex!

Need to up my regex game 😎 

Thanks for the exact pattern Christos.

packages/calypso-products/src/plans-utilities.ts
ddc22: I have done this experiment before and moving these definitions inside the function seem to not have much of an impact. for 100_000 clicks the difference is in milliseconds. And most of the time the experiment has favourable time for having the functions inside. But I am not sure we can derive any meaningful performance stats at this minute difference level. At the very least it might not be that impactful. 

https://jsfiddle.net/jdc91xt/cu8s7gdk/40/

image


Consequently, I think it's better to have the sub functions inside this function for better readability. Do let me know if there is a flaw in the way I am measuring this information 👍

packages/calypso-products/src/plans-utilities.ts
ddc22: Actually the `window` object needs to be declared for optional chaining to work, but we can probably add optional chaining to the rest 👍 

> p.s. if we were in all-React mode/usage, then we'd probably be looking at a different implementation, not relying explicitly on window (we have state selectors for grabbing the current path from context)

Yes I thought of moving this into React but this is used locally inside this package so opted to not try to go down that rabbit hole right now.

packages/calypso-products/src/plans-utilities.ts
ddc22: > This is unnecessarily complex for what you are describing.

Wouldn't this complexity be warranted because this will be used frequently? When comparing to regex wouldn't a string split and a comparison be more performant, given that we know the exact use cases that we would be using a positive flag here? 

👍 for using the native method!

> pathname.startsWith( '/start/plans' )

packages/calypso-products/src/plans-utilities.ts
ddc22: ~On second thought I will use globalThis!~

Ok I switched to typescript global 😇 
https://github.com/Automattic/wp-calypso/blob/02edd3b3fbfbac7216ede3ff36a8d5c1ea7368e2/packages/calypso-products/src/plans-utilities.ts#L16-L20

packages/calypso-products/src/plans-utilities.ts
ddc22: >Can we see what the right way of doing this outside of React? If nothing comes up, then we can memorize the edge/uses.

Hey @chriskmnds is this something we can pick up immediately after the release?

I did come up with a solution, but we might need to spend some time on the change to make sure we are achieving the desired optimisation.

PR : #73532
~Commit : d319f4b9bf65d498b02db39365ed915bfecd79e0~ (Please check the PR)

jdc91: This is a good observation @chriskmnds, let me go back to the drawing board with this.

packages/calypso-products/src/plans-utilities.ts
ddc22: This is a good observation @chriskmnds, let me go back to the drawing board with this.

ddc22:
@chriskmnds After your valid concern here https://github.com/Automattic/wp-calypso/pull/73481#discussion_r1111321510 I did some changes to this approach.

The so `Visibility of the Grid` check was separated out, and explicitly named (`is2023PricingGridActivePage`), so that the underlying motivation for the function is clear. I am also expecting the window object as a parameter, making it even more apparent that the function deals with a page/url specific check.

https://github.com/Automattic/wp-calypso/blob/ce27e565cd1923c9126fedfe63e7ec41adc8ed71/packages/calypso-products/src/plans-utilities.ts#L66-L90

The other non page based, system wide condition remains as the previous function`is2023PricingGridEnabled`.
https://github.com/Automattic/wp-calypso/blob/ce27e565cd1923c9126fedfe63e7ec41adc8ed71/packages/calypso-products/src/plans-utilities.ts#L50-L64

The system wide check `is2023PricingGridEnabled`  will be utilised at places like the [feature list ](https://github.com/Automattic/wp-calypso/blob/08ed984494abfcf18855f303adf71e1e4d17d627/client/lib/plans/features-list.tsx#L233) and [plan list](https://github.com/Automattic/wp-calypso/blob/906098471d0227bad88b31f74bdedde775274f8a/packages/calypso-products/src/plans-list.tsx), which are system wide change irrespective of the page.  

https://github.com/Automattic/wp-calypso/blob/906098471d0227bad88b31f74bdedde775274f8a/packages/calypso-products/src/plans-list.tsx#L701-L702

The basic idea of this change is to evaluate system wide changes in `is2023PricingGridEnabled` and the grid visibility itself via `is2023PricingGridActivePage`.

Please do let me know if there is room for improvement :)

packages/calypso-products/src/plans-utilities.ts
chriskmnds: Not sure I find that more readable really, but as you like.

I think it's a question of good practice, drawing a line irrespective of minuscule or bigger gains. Unless memory is a concern, then the engine probably has to do additional work. It might also be a way to keep the code modular and give it proper names. For one, `isActivePage` is ambiguous, whether it lives here or outside (if outside, you'd probably be giving it a better name to account for the separation also - something like `isPathEnabledFor2023Grid` perhaps).

no strong feelings. I also do what you did many times, just probably should stop.

packages/calypso-products/src/plans-utilities.ts
ddc22: > Ok I switched to typescript global 😇

This is no longer applicable 😬

packages/calypso-products/src/plans-utilities.ts
chriskmnds: I think the bottom line is it shouldn't be used frequently. Not more than once per path change, so I don't know.

but yes, regex, I'd imagine would be more expensive. 

good thinking :-)

chriskmnds:
@jdc91 do you as you like with this, but let's be mindful how often utility functions like this run (cause this and that and the other eventually add up). outside of [this point](https://github.com/Automattic/wp-calypso/pull/73481#discussion_r1110227612), which is now resolved, the rest are fine.

packages/calypso-products/src/plans-utilities.ts
chriskmnds: I didn't mean for you to roll your own caching heuristic (also one that looks quite so involved). Maybe worth asking in the javascript channel in Slack for ideas first - there may be recommended ways/libs vs rolling your own. 

For something that tends to lead to over-engineering, maybe an alternative way exists that doesn't involve much complexity e.g. splitting this function apart, creating a memorized selector/wrapper/hook over it for cases where that's used in React, etc. (just rough thoughts)

chriskmnds: > Hey @chriskmnds is this something we can pick up immediately after the release?

yes, of course

southp (changes requested):
I've gone through various flows under `/start` and `/setup`, e.g.

* /start/hosting
* /start/
* /start/personal, /start/premium (these shouldn't be affected; just in case :) )
* /setup/link-in-bio
* /setup/newsletter
* /setup/videopress

The above works good. However, `/plans` looks broken, i.e. while logged-in, accessing `/plans` without the site slug. It now always redirects to https://wordpress.com/pricing no matter logged-in or logged-out. The behavior should be:

1. Logged-out, redirect to https://wordpress.com/pricing
2. Logged-in, redirect to /plans/<site_slug>

My impression is that this no-slug URL is used somewhere in the support doc and is sent sometimes during a livechat to direct users to the plan table, so let's fix this :)

southp (approved):
NM, it's the same behavior on `trunk` so unrelated. 👍🏼

Issue #73339, opened 2023-02-14
Newsletter accent colors do not reset "choose a plan" conditional task in launchpad


Addison-Stavlo:
Switching back to the default accent color does not get rid of the "choose a plan" requirement. (or more simply, it doesn't seem possible to actually switch back to the default color)

To Repro:
* Create a site in the newsletter flow - wordpress.com/setup/newsletter
* Select a different "favorite color" (like vivid red) in the first setup step.
![Screen Shot 2023-02-14 at 1 18 34 PM](https://user-images.githubusercontent.com/28742426/218823494-d7f39351-2747-4a23-a968-6e00f47d9e9a.png)
* continue through onboarding to the launchpad and notice the "choose a plan" task showed up due to our selection
![Screen Shot 2023-02-14 at 1 23 26 PM](https://user-images.githubusercontent.com/28742426/218824512-e6418723-c87f-4d0e-84c0-2a186299cf24.png)

* click on "Personalize Newsletter" to revisit the setup step.

### Problem 1 - the selection now loads with the original color, and not the color that is currently selected.

![Screen Shot 2023-02-14 at 1 24 41 PM](https://user-images.githubusercontent.com/28742426/218824744-a25f5ae1-808c-43a4-810e-c68b7addb6ac.png)

### Problem 2 - Continuing from here doesn't change the color.

It seems possible to change to other custom colors, but the default (free) option does not seem to work:
![newsletter-color-flow](https://user-images.githubusercontent.com/28742426/218825281-fa8204c5-bd8e-415c-bc45-a728a7a551ba.gif)

dsas:
You need a blog that limits global styles to reproduce this.

I had assumed that this was due to https://github.com/Automattic/wp-calypso/pull/70958 but actually, even after reverting the changes in that PR and in the phab diff, this is still reproducible. 

It looks like the problem is that the newsletter setup screen thinks the color is already blue and so doesn't make the XHR request to save it as blue. 

It think this because it hits the `/wpcom/v2/sites/215718860/global-styles-variation/site-accent-color` endpoint. This endpoint uses the `wpcom_get_site_accent_color` function, which eventually hits `WP_Theme_JSON_Resolver::get_user_data` which runs the `wp_theme_json_data_user` filter that we hook to implement global styles with `wpcom_block_global_styles_frontend`.

dsas:
Possible solutions:
 * Implement an exemption to using global styles
[...]

Issue #70822, opened 2022-12-06
[Bug]: Upgrade plan step text in launchpad can be misleading


dsas:
### Quick summary

When signing up for a site, in some limited circumstances you can come back into the launchpad to read text saying:

> Upgrade your plan to publish your color changes and unlock tons of other features

The text is activated by use of global styles but it doesn't check how global styles has been used - it assumes that it's due to a colour change as that's the most likely path (due to accent colours on newsletter flow).

### Steps to reproduce

1. Go to wordpress.com/setup/link-in-bio
2. Choose a free plan when prompted
3. Keep going through the steps until the 'Add links' step, which puts you in the site editor.
4. Add the `wpcom-limit-global-styles` sticker to your site
5. Refresh the site editor
6. You're in the site editor, feel free to add links but mainly, use the styles sidebar to tweak the custom styles, e.g. change a font, but do not change colours
7. Save your changes, note the warning about global styles when saving
8. Press back, you should end up back in the launchpad.

### What you expected to happen

The launchpad should say something like "Upgrade to a premium plan to publish your style changes and unlock tons of other features"

### What actually happened

Launchpad says "Upgrade your plan to publish your color changes and unlock tons of other features", even though no colours have been changed in the making of this site.

![image](https://user-images.githubusercontent.com/93301/205936750-61197302-c6ef-4627-be9b-74cf760b6e5f.png)

### Context

Noticed while developing https://github.com/Automattic/dotcom-forge/issues/1227

### Reproducibility

Consistent

### Severity

None

### Available workarounds?

No but the platform is still usable

### Workaround details

It's just confusing wording

Copons:
Thanks Dean, added to our board, with lower priority (not sure why GH-actions added `[Pri] High`?) as it's fundamentally a edge case.
I'll mention this during the next design walkthrough.

SaxonF:
"style changes" vs "color changes" makes sense to me

dsas:
It was pointed out in the pekYwv-OA-p2 (around 47 minute) the wording / presentation should be much more celebratory and not look like a warning - better wording but something more like "You have customized your site styles it's looking good, you need to upgrade for it to be used".

"It's a terrible UI, it looks like an alarm, we should celebrate and make people want to pay...The application isn't ideal but the intent is good, it gives you reassurance so you know what will happen next"

dsas:
See also https://github.com/Automattic/wp-calypso/issues/75728

dsas:
[...]

PR #73509, opened 2023-02-17
Bypass limited global styles when fetching accent colour

dsas:
Related to https://github.com/Automattic/wp-calypso/issues/73339

## Proposed Changes

The global-styles-variation/site-accent-color API end point is used to fetch the accent colour set by the user during the newsletter setup. If the site has limited global styles then it will always return the theme default colour, rather than the one set by the user.

The implication is that if the user chooses a non-default colour, they cannot later change their mind and revert back to the default colour during the newsletter set-up process.

This has been addressed by always setting the preview-global-styles GET var when calling the endpoint so that it returns the set value and not the limited value.

## Testing Instructions

1. Create a free site in the newsletter flow - wordpress.com/setup/newsletter
2. Select a non-default "favorite color" (like vivid red) in the first setup step.
3. Continue through onboarding to the launchpad and notice the "choose a plan" task showed up due to our selection
4. Click on "Personalize Newsletter" to revisit the setup step.
5. Check that the Favorite color is set to the non-default color you selected before
6. Change the Favourite color to the default color, press Continue
7. Check that the preview shows the Favorite color last selected.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/hooks/use-accent-color.ts
+++ b/client/landing/stepper/hooks/use-accent-color.ts
@@ -9,7 +9,7 @@ const useAccentColor = () => {
 	if ( ID ) {
 		wpcom.req
 			.get( {
-				path: `/sites/${ ID }/global-styles-variation/site-accent-color`,
+				path: `/sites/${ ID }/global-styles-variation/site-accent-color?preview-global-styles=true`,
 				apiNamespace: 'wpcom/v2',
 			} )
 			.then( ( color: string ) => {


mmtr (commented):
This works as advertised, but I think there is one more remaining issue.

After switching back to the default color, the preview correctly shows the default color, but the warning of the "choose a plan" task is still there:

Screenshot 2023-02-20 at 11 30 44

Is that supposed to be solved by this PR or will it be handled separately?

dsas:
> Is that supposed to be solved by this PR or will it be handled separately?

Oh sorry about that. The issue is that saving the default accent color still results in a global styles post being created or modified. There might be calypso changes necessary but this will need tackling in a back-end change too, otherwise we'll display things like the "preview" to users on the frontend.

There might be other GS changes in the post, which means we can't just delete the post https://github.com/Automattic/wp-calypso/issues/70822 is open for handling that case better.

mmtr:
> The issue is that saving the default accent color still results in a global styles post being created or modified.

Thanks for clarifying. IIRC, I think we handled a similar issue in the past by preventing default styles from being added to the global post for a particular scenario (but I don't remember which one), so I guess it's a matter of expanding that logic to also cover this scenario.

I'd recommend keeping https://github.com/Automattic/wp-calypso/issues/73339 open until we also fix that.

mmtr (approved):

PR #45040, opened 2020-08-19
Prevent duplicate sites in signup 

niranjan-uma-shankar:
#### Changes proposed in this Pull Request

* Detect if signup flow was loaded via back button, and in that case do not create a new site when the signup flow is complete.
* Check pau2Xa-1Io-p2#comment-6759 for an explanation on how this implementation works
* There's an edge case in which a user could end up seeing the launch flow UI version when going to wordpress.com/start to add a new site. But it's going to be an extremely rare case, and a few things have to align, so we can safely not bother about this issue:
    - Firstly, when the user is on checkout page from signup flow, they should navigate to a different part of Calypso on the _same tab_ without using the close button on checkout header, or empty cart action.
    - Next, on the _same tab_, they should navigate to wordpress.com/start by directly typing the URL in the address bar, or clicking a link that opens the page(on the same tab).
    - All of this should happen in < 1 day since a new site was created (signup destination cookie will expire in 1 day).
Important to note that this issue will not be a permanent problem - the user just needs to open wordpress.com/start in a new tab to see the regular signup flow again. Again, this is a super edge case and not something to worry about IMO.

#### Testing instructions

**Scenario 1** (Verify that going back from checkout page to signup, you can select a new domain/plan and complete purchase without creating a duplicate site).

1. Starting logged out of wpcom, go through the signup at wordpress.com/start and select a paid plan.
2. On the checkout page, hit browser back. Verify that you see the launch flow UI version:

Screenshot 2020-08-27 at 5 38 18 PM

3. Test the following two scenarios:
    (A) Selecting a free plan should take you to customer home and duplicate site should not be created. 
    (B) Select a custom domain and paid plan and complete payment. Verify that the new domain and plan got added to your site and that duplicate sites were not created.

**Scenario 2** (When on signup via browser back from checkout, page reload should not affect the UI)

1.  Starting logged out of wpcom, go through the signup at wordpress.com/start and select a paid plan.
2.  On the checkout page, hit browser back. Verify that you see the launch flow UI version (screen shot in Scenario 1)
3. Reload the page and verify that you still see the launch flow UI version.

**Scenario 3** (Add new site should work, and open the regular signup flow)

1.  Starting logged out of wpcom, go through the signup at wordpress.com/start and select a paid plan.
2.  On the checkout page, click the close button in the masterbar or empty all cart items.
3. Click "Add new site" from Calypso's sidebar, and verify that the domain step shows the regular signup flow UI(and not the launch flow UI)

**Scenario 4**
[...]

PR #73525, opened 2023-02-18
Signup: Handle case where window.sessionStorage does not exist

sirbrillig:
In checkout's review step (added in https://github.com/Automattic/wp-calypso/pull/72523), and many other places, there is code that calls a series of functions in the `client/signup` library like `getSignupCompleteFlowName()`. These functions (from https://github.com/Automattic/wp-calypso/pull/45040) all access `window.sessionStorage`. However, there are cases where that global object will not exist (I'm not 100% sure what those are but based on the errors it seems to be related to Safari). If this occurs, these functions throw a fatal error. 

This was noticed in checkout because we have more advanced error logging and reporting there, but it's likely causing crashes in other parts of calypso too.

Here's an example of the error: 

```
null is not an object (evaluating 'sessionStorage.getItem'); Stack: d@https:\/\/wordpress.com\/calypso\/evergreen\/90798.0e53b958e45ce8df970f.min.js:339:87368\nIo@https:\/\/wordpress.com\/calypso\/evergreen\/90798.0e53b958e45ce8df970f.min.js:106:3895\nai@https:\/\/wordpress.com\/calypso\/evergreen\/89564.0db68a9a80fa90a7f77c.min.js:2:168083...
```

For the alert, see p1676742304987539-slack-C096PD42U and p1676396706072749-slack-C096PD42U

## Proposed Changes

In this PR we use optional chaining to modify all the `window.sessionStorage` functions in the signup library so that they can be safely called when `window.sessionStorage` does not exist.

## Testing Instructions

None should be needed.

### DIFF
--- a/client/signup/storageUtils.js
+++ b/client/signup/storageUtils.js
@@ -21,16 +21,16 @@ export const clearSignupDestinationCookie = () => {
 };
 
 export const getSignupCompleteSlug = () =>
-	sessionStorage.getItem( 'wpcom_signup_complete_site_slug' );
+	sessionStorage?.getItem( 'wpcom_signup_complete_site_slug' );
 export const setSignupCompleteSlug = ( value ) =>
-	sessionStorage.setItem( 'wpcom_signup_complete_site_slug', value );
+	sessionStorage?.setItem( 'wpcom_signup_complete_site_slug', value );
 export const wasSignupCheckoutPageUnloaded = () =>
-	sessionStorage.getItem( 'was_signup_checkout_page_unloaded' );
+	sessionStorage?.getItem( 'was_signup_checkout_page_unloaded' );
 export const setSignupCheckoutPageUnloaded = ( value ) =>
-	sessionStorage.setItem( 'was_signup_checkout_page_unloaded', value );
+	sessionStorage?.setItem( 'was_signup_checkout_page_unloaded', value );
 export const getSignupCompleteFlowName = () =>
-	sessionStorage.getItem( 'wpcom_signup_complete_flow_name' );
+	sessionStorage?.getItem( 'wpcom_signup_complete_flow_name' );
 export const setSignupCompleteFlowName = ( value ) =>
-	sessionStorage.setItem( 'wpcom_signup_complete_flow_name', value );
+	sessionStorage?.setItem( 'wpcom_signup_complete_flow_name', value );
 export const clearSignupCompleteFlowName = () =>
-	sessionStorage.removeItem( 'wpcom_signup_complete_flow_name' );
+	sessionStorage?.removeItem( 'wpcom_signup_complete_flow_name' );


wojtekn (approved):

danielbachhuber:
> so that they can be safely called when `window.sessionStorage` does not exist.

It seems like TypeScript and/or linting should've caught this...

What else in the checkout flow breaks when `window.sessionStorage` isn't available?

sirbrillig:
That's a good question. It looks like some places in calypso are using optional chaining already like this:

https://github.com/Automattic/wp-calypso/blob/a736d6618aff94c53596edecb828b52d8bdb1776/client/components/domains/register-domain-step/index.jsx#L307

I generally like to use this approach because in the past certain browsers like Safari private browsing mode would make various `window` features unavailable (like `localStorage`). Also if a file is ever run in an server environment for server-side-rendering the `window` object itself won't exist.

However, I see that many other places in calypso call `window.sessionStorage` as though it were always available like this:

https://github.com/Automattic/wp-calypso/blob/a736d6618aff94c53596edecb828b52d8bdb1776/client/blocks/importer/wordpress/utils.js#L12

I don't know what the right policy is here, but in this case it seems easy enough to apply defensive coding principles and add a guard. I searched around and I don't see any other obvious places we rely on sessionStorage within checkout without guards so at least we can make sure that works.

danielbachhuber:
> I don't know what the right policy is here, but in this case it seems easy enough to apply defensive coding principles and add a guard.

@Automattic/team-calypso-platform Any recommendations you might be able to offer?

noahtallen:
I did a short amount of research, and it looks like this can happen in other browsers too. For example, this happens in Chrome: https://a8c.sentry.io/issues/3948311272 (`Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.`)

MDN recommends checking if local storage is available with a lot of extra checks for various browser-specific issues: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability

This makes me wonder if we should create a wrapper function which does nothing when storage isn't available for various reasons, and then add a linter rule to use that instead of directly accessing the API.

cc @tyxla or @jsnajdr for more thoughts

jsnajdr:
> This makes me wonder if we should create a wrapper function which does nothing when storage isn't available for various reasons, and then add a linter rule to use that instead of directly accessing the API.

Yes, we already have a module called [`local-storage-polyfill`](https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/local-storage-polyfill/index.js) which detects support for `window.localStorage` and polyfills it with memory-based implementation if it's missing or dysfunctional.

Currently that polyfill is [disabled](https://github.com/Automattic/wp-calypso/blob/trunk/client/webpack.config.js#L373-L380) for evergreen browsers (by aliasing it to `noop` in webpack config). We could update it, to detect contemporary browser bugs and omissions, instead of past ones, and extend it to polyfill `window.sessionStorage`, too.

danielbachhuber:
@jsnajdr Want to create a GitHub issue detailing your recommended fix?

PR #73557, opened 2023-02-20
Try parallel app processing

noahtallen:
This PR creates an all-in-one Calypso app artifact processor. In the previous approach, with each app processed in an individual TeamCity build, the artifact process script only had to care about one app.

The major change here is to parallelize and combine in the following ways:

1. In parallel, download the artifact for each app and compare it to the release. This gives us a `Promise.all` over functions which return boolean, which in turn gives us a list of apps which are changing.
2. Update the tag logic to do multiple tags in a build, based on the changing apps.
3. Update the notification logic to handle multiple apps. Instead of one GH comment per app, we'll have one comment which lists the app. (The slack endpoint will ultimately have to be updated too, but it's currently unused.)

All of this is re-written in NodeJS to be (hopefully!) easier to read than bash script in kotlin string! It's also easier to manage data and parallelization.

I've tested the following parts:
1. Does it detect a changing app vs previous tagged build: ✅
2. Does it add and remove GH comment: ✅
3. Does it authenticate with the Slack endpoint: ✅
4. Does it add tags based on what changes: ✅

What's missing: the previous artifact processor added a build_meta.txt file. I'd like to deprecate that in favor of build_meta.json. Some builds generate that with webpack, but we need some way to get it in all builds.

### DIFF
--- a/apps/happy-blocks/bin/prepArtifactForCI.sh
+++ b/apps/happy-blocks/bin/prepArtifactForCI.sh
@@ -4,15 +4,15 @@ set -o nounset
 set -o pipefail
 
 # Copy build directories to the release directory.
-find ./block-library/* -type d -name "*" -prune |\
+find ./block-library/* -type d -name "*" -prune ! -name "shared" |\
 while read -r block;
 do
-	mkdir -p ./release-files/${block//\.\.\//}; 
-	cp -r $block/build/* ./release-files/${block//\.\.\//}/; 
+	mkdir -p "./release-files/${block//\.\.\//}"; 
+	cp -r $block/build/* "./release-files/${block//\.\.\//}/"; 
 done
 
 # Add the index.php file
-cp ./index.php ./README.md ./dist/* ./release-files/
+cp ./index.php ./README.md ./release-files/
 cp ./translations-manifest.json ./release-files/
 
 printf "Finished configuration of the @automattic/happy-blocks plugin artifacts directory.\n"
--- a/apps/happy-blocks/bin/prepArtifactForCI.sh
+++ b/apps/happy-blocks/bin/prepArtifactForCI.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
-
-set -x
+set -o errexit
+set -o nounset
+set -o pipefail
 
 # Copy build directories to the release directory.
 find ./block-library/* -type d -name "*" -prune |\
--- a/apps/notifications/package.json
+++ b/apps/notifications/package.json
@@ -22,7 +22,7 @@
 		"start": "yarn run clean && yarn run build:notifications && yarn run dev-server",
 		"dev": "yarn run calypso-apps-builder --localPath dist --remotePath /home/wpcom/public_html/widgets.wp.com/notifications",
 		"build": "NODE_ENV=production yarn dev",
-		"build-ci": "yarn build && ./bin/prepArtifactForCI.sh"
+		"build-ci": "yarn build"
 	},
 	"dependencies": {
 		"@automattic/calypso-color-schemes": "workspace:^",
--- a/apps/odyssey-stats/webpack.config.js
+++ b/apps/odyssey-stats/webpack.config.js
@@ -41,9 +41,10 @@ const excludedPackages = [
 ];
 
 const excludedPackagePlugins = excludedPackages.map(
-	( package ) =>
+	// Note: apparently the word "package" is a reserved keyword here for some reason
+	( pkg ) =>
 		new webpack.NormalModuleReplacementPlugin(
-			package,
+			pkg,
 			path.resolve( __dirname, 'src/components/nothing' )
 		)
 );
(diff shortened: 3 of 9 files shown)


apps/notifications/package.json
```diff
@@ -22,7 +22,7 @@
 		"start": "yarn run clean && yarn run build:notifications && yarn run dev-server",
 		"dev": "yarn run calypso-apps-builder --localPath dist --remotePath /home/wpcom/public_html/widgets.wp.com/notifications",
 		"build": "NODE_ENV=production yarn dev",
-		"build-ci": "yarn build && ./bin/prepArtifactForCI.sh"
```
noahtallen, line 25: Prep artifact actually requires the "prev build" to be downloaded already, so it should be in the normalization script

worldomonation:
```
What's missing: the previous artifact processor added a build_meta.txt file. I'd like to deprecate that in favor of build_meta.json. Some builds generate that with webpack, but we need some way to get it in all builds.
```

I spun up a PR to hopefully address that. I didn't want to do it in the main stream of the CI optimization work, because (IIUC) the changeover to `build_meta.json` isn't absolutely necessary to complete the work.

worldomonation:
@noahtallen This PR is still in draft, but I think it's ready to review and possibly merge. What do you think?

noahtallen:
I'll take a look at the CI error and get it ready! You're probably right, we can test it out more after merging and getting more apps set up with it.

apps/odyssey-stats/webpack.config.js
```diff
@@ -41,9 +41,10 @@ const excludedPackages = [
 ];
 
 const excludedPackagePlugins = excludedPackages.map(
-	( package ) =>
+	// Note: apparently the word "package" is a reserved keyword here for some reason
+	( pkg ) =>
 		new webpack.NormalModuleReplacementPlugin(
-			package,
+			pkg,
```
noahtallen, line 47: Weird issue, but can consistently replicate when running the translation script in **happy blocks** of all places

apps/happy-blocks/bin/prepArtifactForCI.sh
```diff
@@ -1,17 +1,18 @@
...
 do
-	mkdir -p ./release-files/${block//\.\.\//}; 
-	cp -r $block/build/* ./release-files/${block//\.\.\//}/; 
+	mkdir -p "./release-files/${block//\.\.\//}"; 
+	cp -r $block/build/* "./release-files/${block//\.\.\//}/"; 
 done
 
 # Add the index.php file
-cp ./index.php ./README.md ./dist/* ./release-files/
+cp ./index.php ./README.md ./release-files/
```
noahtallen, line 15: dist isn't used anymore, which can be verified by checking the trunk artifacts

apps/happy-blocks/bin/prepArtifactForCI.sh
```diff
@@ -1,17 +1,18 @@
...
 
 # Copy build directories to the release directory.
-find ./block-library/* -type d -name "*" -prune |\
+find ./block-library/* -type d -name "*" -prune ! -name "shared" |\
 while read -r block;
 do
-	mkdir -p ./release-files/${block//\.\.\//}; 
-	cp -r $block/build/* ./release-files/${block//\.\.\//}/; 
+	mkdir -p "./release-files/${block//\.\.\//}"; 
+	cp -r $block/build/* "./release-files/${block//\.\.\//}/";
```
noahtallen, line 11: I have a feeling this code shouldn't be in this script. It seems necessary for the plugin to work normally, but it's in a CI-only script. So the normal `yarn build` script won't work

@worldomonation I wonder if this means we should move the CI related stuff back to .teamcity, so it's clearer that it should only be used for CI workflows, not for local builds 🤔

worldomonation: @noahtallen It's been close to a month so my recall is a bit hazy but I do think we can remove all `build-ci` yarn scripts, `prepArtifactForCI.sh` bash scripts and such from the repo for now. IIRC I tried to incorporate the processing into TeamCity in a follow-up branch I had.

worldomonation: Should I push to your branch to do that?

noahtallen: ok cool! I don't think we need to do it in this branch. If we merge this separately we can split that up into other PRs which are easier to look at individually

worldomonation: Gotcha. I'll rebase my WIP branch after we merge, and get rid of these files and scripts.

apps/odyssey-stats/webpack.config.js
worldomonation: Nice catch!

worldomonation (approved):

PR #73558, opened 2023-02-20
Update eslint

noahtallen:
## Proposed Changes
Updates eslint and related packages to the latest non-major version. The only two packages which _do_ have newer major versions are:

- eslint-plugin-jest
- eslint-plugin-jsdoc

I chose not to do those updates here since they'll probably entail more work.

This group of updates is inspired mostly by `fetch` now being available in NodeJS. In our repo, eslint says it's not defined, when it actually is. On more recent versions of eslint, this is fixed.

I also ran an eslint autofix (`yarn eslint --fix .`), which resulted in a few minor fixes. It looks like the import-order rule has gotten some bug fixes.

Previously, imports were allowed in this order:

```js
import bar from './foo-bar';
import foo from './foo';
```

Now, this order is required:

```js
import foo from './foo';
import bar from './foo-bar';
```

(Which is technically the correct ordering, but wasn't previously considered an issue.)

This was autofixed in this PR :) 

## Testing Instructions
- CI
- Local IDE eslint


worldomonation (approved):

tyxla:
Great work @noahtallen 🙌

PR #73572, opened 2023-02-21
Plugins: Add Action and Status Selectors to selectors-ts.

robertf4:
## Proposed Changes

This PR adds the selectors related to plugin actions and statuses to selectors-ts. It is a part of the ongoing improvements to the plugin selectors which began in https://github.com/Automattic/wp-calypso/pull/72363.

## Testing Instructions

Review the code and tests, and ensure the tests pass.

## Pre-merge Checklist

### DIFF
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -372,6 +372,7 @@ export function isPluginActionStatus(
 	state: AppState,
 	siteId: number,
 	pluginId: string,
+	// TODO: refactor to take only string and not string[] to reduce complexity.
 	action: string | string[],
 	status: string
 ) {
@@ -393,6 +394,7 @@ export function isPluginActionStatus(
  * @param  {string|Array} action   Action, or array of actions of interest
  * @returns {boolean}              True if one or more specified actions are in progress, false otherwise.
  */
+// TODO: remove isPluginActionInProgress and use isPluginActionStatus instead.
 export function isPluginActionInProgress(
 	state: AppState,
 	siteId: number,
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -351,8 +351,11 @@ export function getStatusForPlugin( state: AppState, siteId: number, pluginId: s
 		return undefined;
 	}
 
-	const status = state.plugins.installed.status[ siteId ][ pluginId ];
-	return Object.assign( {}, status, { siteId: siteId, pluginId: pluginId } );
+	return {
+		...state.plugins.installed.status[ siteId ][ pluginId ],
+		siteId,
+		pluginId,
+	};
 }
 
 /**
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -12,6 +12,7 @@ import type {
 	Plugin,
 	PluginFilter,
 	PluginSites,
+	PluginStatus,
 } from './types';
 import type { AppState } from 'calypso/types';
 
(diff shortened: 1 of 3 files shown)


client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
 	( state: AppState ) => [ getAllPluginsIndexedByPluginSlug( state ), getSitesItems( state ) ],
 	( state: AppState, siteIds: number[], pluginSlug: string ) => [ pluginSlug, ...siteIds ].join()
 );
+
+export function getStatusForPlugin( state: AppState, siteId: number, pluginId: string ) {
+	if ( typeof state.plugins.installed.status[ siteId ] === 'undefined' ) {
+		return false;
```
tyxla: A boolean is an unexpected value from this selector IMHO. Should it be `null` or `undefined` instead?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+	}
+	if ( typeof state.plugins.installed.status[ siteId ][ pluginId ] === 'undefined' ) {
+		return false;
+	}
```
tyxla: We could convert this to a single condition with some optional chaining:

```suggestion
	if ( typeof state.plugins.installed.status[ siteId ]?.[ pluginId ] === 'undefined' ) {
		return false;
	}
```

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+	const status = state.plugins.installed.status[ siteId ][ pluginId ];
+	return Object.assign( {}, status, { siteId: siteId, pluginId: pluginId } );
```
tyxla: We could shorten this and make it more comprehensible by avoiding declaration of a constant that is used just once, plus using spread, and object property shorthand:

```suggestion
	return {
		...state.plugins.installed.status[ siteId ][ pluginId ],
		siteId,
		pluginId
	};
```

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+ * @param  {string}       pluginId ID of the plugin
+ * @param  {string|Array} action   Action, or array of actions of interest
+ * @param  {string}       status   Status to check against
+ * @returns {boolean}              True if status is the specified one for one or more actions, false otherwise.
+ */
+export function isPluginActionStatus(
+	state: AppState,
+	siteId: number,
+	pluginId: string,
+	action: string | string[],
```
tyxla, line 376: Could we simplify the API by always accepting a single action? That would make the API, the selector, its tests, and its types simpler. It would also be more predictable. Then we can use this selector multiple times if necessary.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+	status: string
```
tyxla, line 377: It also felt weird that we can pass multiple actions, but just a single status. Is that intended?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+/**
+ * Whether the plugin's status for one or more recent actions is in progress.
+ *
+ * @param  {Object}       state    Global state tree
+ * @param  {number}       siteId   ID of the site
+ * @param  {string}       pluginId ID of the plugin
+ * @param  {string|Array} action   Action, or array of actions of interest
+ * @returns {boolean}              True if one or more specified actions are in progress, false otherwise.
+ */
+export function isPluginActionInProgress(
```
tyxla, line 398: I find this selector a bit unnecessary. The only thing it does is call another selector with one additional argument. Should we remove and always call `isPluginActionStatus` directly?

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -321,3 +322,89 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
+}
+
+/**
+ * Retrieve all plugin statuses of a certain type.
+ *
+ * @param  {Object} state    Global state tree
+ * @param  {string} status   Status of interest
+ * @returns {Array}          Array of plugin status objects
+ */
+export const getPluginStatusesByType = createSelector(
```
tyxla, line 414: Can you elaborate on this selector? I'm confused as to what it's doing and why it's necessary. What does "all plugin statuses of a certain type" mean? Since this is plugin information, is it data we could derive from other existing selectors?

client/state/plugins/installed/selectors-ts.ts
robertf4: I made it return `undefined`.

client/state/plugins/installed/selectors-ts.ts

client/state/plugins/installed/selectors-ts.ts
robertf4: We could, but it changes the contract of the API so I feel it is out of the scope of this PR.

client/state/plugins/installed/selectors-ts.ts
robertf4: It is just the contract as it existed in `selectors.js`, and for this development I was trying to not change any behavior or contracts.

client/state/plugins/installed/selectors-ts.ts
robertf4: I agree, but in this case `isPluginActionStatus` is not used anywhere, and `isPluginActionInProgress` is used in 5 components so I think it's best to leave things as they are for this PR.

client/state/plugins/installed/selectors-ts.ts
robertf4: This is another case where I am simply converting the selector from JavaScript to TypeScript along with the rest of the file, and this function is being used in several places so removing the selector is out of scope.

client/state/plugins/installed/selectors-ts.ts
tyxla: Worth adding a TODO comment in that case. It's one of those things that can add unnecessary complexity to a codebase, especially when it happens all over the place.

client/state/plugins/installed/selectors-ts.ts
tyxla: Same as above.

client/state/plugins/installed/selectors-ts.ts
tyxla: Adding a TODO comment could suffice in that case. I'm happy with keeping things contained.

client/state/plugins/installed/selectors-ts.ts
tyxla: Not sure how that answers my question, but that's fine. Would be good to get some understanding of what's going on and eventually improve in the future, even if we end up only adding TODO comments for the next person that works in here.

tyxla (approved):
I don't see anything blocking at this time. Would be ideal if we can add some TODO comments here and there and ideally, follow up in the future to improve the state of things.

PR #30569, opened 2019-02-04
Fix localStorage-bypass to work in Firefox

brandonpayton:
Today, to avoid conflicts when supporting multiple users in separate windows, our localStorage bypass tries to create properties on window.localStorage to shadow methods like setItem and getItem, but that doesn't work in Firefox. The assignments are silently ineffective. This is [according to spec](https://github.com/whatwg/html/issues/183#issuecomment-1429446050) because [the Storage interface](https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface) is not tagged with [OverrideBuiltins](https://heycam.github.io/webidl/#OverrideBuiltins).

This is needed by #30157 which supports Firefox.

#### Changes proposed in this Pull Request

Instead of trying to add properties on localStorage that shadow Storage methods, redefine `window.localStorage` to refer to an object that implements the same interface we rely upon. Note: Neither the current bypass nor this update fully implement the Storage interface which supports getting and setting props directly on `localStorage` (e.g., `localStorage['prop1'] = value`). For that, we'd need a different implementation that uses Proxy.

#### Testing instructions

* Open a tab for one supported user.
* Open a tab for a different supported user.
* Open the dev tools console for each.
* Run `localStorage.setItem( 'test-key', 'test-value' )` under one tab.
* Switch to console in the other tab, call `localStorage.getItem( 'test-key' )`, and verify it returns `null`.
* Some keys are still allowed to be set on real `localStorage`. One of those keys is 'debug'. Run `localStorage.setItem( 'debug', 'somevalue' )`, switch to the other tab, and verify that `localStorage.getItem( 'debug' )` returns 'somevalue'. Then call `localStorage.removeItem( 'debug' ) to clean up after this test.

PR #73584, opened 2023-02-21
Support user: pass correct args to localStorageBypass

jsnajdr:
While reviewing #73525 I looked at the `local-storage-bypass` module and noticed that the `support-user-interop` module has been calling its initialization function wrong, for 4 years, since #30569. Instead of passing the `allowedKeys` option, it's been passing effectively nothing, and the allowed keys (`boot_support_user` and `debug`) were not passed through to the native storage implementation.

This PR fixes that. Also does a drive-by modifications from `.indexOf` to `includes`, and gives a name to the default export of `local-storage-bypass`. That makes the function easier to grep.

### DIFF
--- a/client/lib/user/support-user-interop.js
+++ b/client/lib/user/support-user-interop.js
@@ -92,7 +92,7 @@ export async function supportUserBoot() {
 	// The following keys will not be bypassed as
 	// they are safe to share across user sessions.
 	const allowedKeys = [ STORAGE_KEY, 'debug' ];
-	localStorageBypass( allowedKeys );
+	localStorageBypass( { allowedKeys } );
 
 	wpcom.setSupportUserToken( user, token, onTokenError );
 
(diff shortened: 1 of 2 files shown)


client/lib/user/support-user-interop.js
```diff
@@ -92,7 +92,7 @@ export async function supportUserBoot() {
 	// The following keys will not be bypassed as
 	// they are safe to share across user sessions.
 	const allowedKeys = [ STORAGE_KEY, 'debug' ];
-	localStorageBypass( allowedKeys );
+	localStorageBypass( { allowedKeys } );
```
tyxla, line 95: 😅

tyxla (approved):
Nice fix 👍 Thanks @jsnajdr 🙌

noahtallen (approved):
Nice catch!

PR #73588, opened 2023-02-21
Design setup: Add "is_style_variation" property to Tracks events

mmtr:
## Proposed Changes

In order to A/B test the style variations as separate theme cards (https://github.com/Automattic/dotcom-forge/issues/1655), ExPlat needs a new metric to measure the number of designs with a style variation that have been selected.

The ExPlat metric can be configured to use any of the design setup Tracks events such as `calypso_signup_select_design`. However, these events don't indicate clearly whether a style variation is involved, since that information is encoded in the `slug` property (e.g. `slug: 'upsidedown-noir'` which refers to the Upsidedown theme and the Noir variation).

Since the ExPlat metrics can only perform an exact match of properties, this PR adds a new `is_style_variation` property so the metric can be built as `calypso_signup_select_design` events with `is_style_variation: true`.

## Testing Instructions

- Open the Network tab in the browser devtools
- Observe the `t.gif` requests
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>`
- Perform some actions such as:
  - Preview a design (with a default style variation)
  - Preview a style variation
  - Select a design (with a default style variation)
  - Select a style variation (after choosing the "Try out" option in the GS gating modal)
- Make sure the tracked events include a new `is_style_variation` property

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/analytics/record-design.ts
+++ b/client/landing/stepper/declarative-flow/internals/analytics/record-design.ts
@@ -91,15 +91,19 @@ export function getDesignEventProps( {
 }
 
 export function getVirtualDesignProps( design: Design, styleVariation?: StyleVariation ) {
+	let is_style_variation = false;
 	let variationSlugSuffix = '';
 	if ( styleVariation && styleVariation.slug !== 'default' ) {
+		is_style_variation = true;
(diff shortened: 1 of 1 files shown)


dsas (approved):
Everything I expected to have a is_style_variation property did, nice one.

PR #73608, opened 2023-02-21
Add group to My home domain upsell e2e tests

lupus2k:
## Proposed Changes

As [suggested here](https://github.com/Automattic/wp-calypso/pull/73391#issuecomment-1438798124), I added the `calypso-pr` group to the e2e tests

Related to https://github.com/Automattic/wp-calypso/pull/73391

### DIFF
--- a/packages/calypso-e2e/src/lib/components/navbar-cart-component.ts
+++ b/packages/calypso-e2e/src/lib/components/navbar-cart-component.ts
@@ -1,10 +1,11 @@
 import { Page } from 'playwright';
 
 const selectors = {
-	cartButton: 'masterbar-cart-button',
-	popover: '.popover',
+	cartButton: '.masterbar-cart-button',
+	popover: '.masterbar-cart-button__popover',
 	removeItemButton: 'text=Remove from cart',
-	removeDialogConfirmButton: 'text=Continue',
+	removeDialogConfirmButton: '.masterbar-cart-button__popover button:text("Continue")',
+	busyCheckoutButton: '.masterbar-cart-button__popover button:text("Checkout").is-busy',
 };
 /**
  * Component representing the navbar/masterbar at top of WPCOM.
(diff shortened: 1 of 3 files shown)


dpasque:
Btw, it looks like the cart removal isn't quite working correctly! I've been debugging this locally but have to jump on another call -- I'll push up my findings after that call!

lupus2k:
> Btw, it looks like the cart removal isn't quite working correctly! I've been debugging this locally but have to jump on another call -- I'll push up my findings after that call!

Hey @dpasque, thanks for taking a look, and yes, the `navbarCartComponent.emptyCart()` function is not working well, I mentioned in the [first PR](https://github.com/Automattic/wp-calypso/pull/73391#issuecomment-1433897997).

dpasque:
@lupus2k and @worldomonation -- let me know what you think of the cart changes I pushed up -- I ran them locally a lot of times, and they have been really stable.

The cart navbar button is just really tricky. It only appears if there are items in the cart, and it's very async in doing so (it is based on a web request and it's a `useEffect` hook calculated off of state that comes from the cart API endpoint). 

So, the most stable way to go about interacting with it is to use the cart endpoint response to determine if there's items in the cart, and if we know there are, we can just wait for the navbar icon to appear confidently.

We'll have to decide how we want to handle this in the quarantined test in the future: I didn't make an implementation there because the test has been quarantined for a while. We just need to either intercept the request or issue one on our own, based on whichever makes more sense.

lupus2k:
> @lupus2k and @worldomonation -- let me know what you think of the cart changes I pushed up -- I ran them locally a lot of times, and they have been really stable.

Hey, @dpasque, I tested, and sometimes it fails, maybe because of some delay loading the shopping cart on my location, I added `await` to `page.waitForResponse( /me\/shopping-cart/ )`, and it's working great now.

dpasque:
> Hey, @dpasque, I tested, and sometimes it fails, maybe because of some delay loading the shopping cart on my location, I added await to page.waitForResponse( /me\/shopping-cart/ ), and it's working great now.

@lupus2k thank you for looking into that! That's really interesting though! 🤔  Can you share more on how you were seeing it fail? Do you have an example log or video? (It would be in `./test/e2e/results`)

The reason being that I'm pretty sure we want to make sure we wait for that shopping cart response _after_ we start the navigation to the new site, right? Because I'm pretty sure the cart request is site-specific, so we want to make sure we're checking the cart response specific to the site we navigate to, which is why the await is after the navigation.

I can play around with it a little bit more too!

dpasque:
Ah, I just found that we actually already had a function on the test REST API client for clearing the cart directly with the REST API! That seems like that will be a lot more stable than trying to intercept things in the browser and using the navbar.

@lupus2k -- what do you think of the most recent changes? Do they pass for you locally in a more stable way?

dpasque (approved):
Also, if that works well for you, and you're good with the approach, we should be good to merge! So getting an approval on here 😄

lupus2k:
Thanks for your help, @dpasque, it works like a charm, and now I learned about using REST_API on the next e2e tests.

PR #73620, opened 2023-02-21
Universal Header: styling tweaks

arcangelini:
## Proposed Changes

While testing and getting ready to apply this to the block I noticed the UniversalNavbarHeader component is slightly off compared to the main WordPress.com page. This PR adjusts the font and WordPressWordmark to match closely to the main page.

### Goal
Markup 2023-02-21 at 23 10 54

### After
Markup 2023-02-21 at 23 06 44

### Before
Markup 2023-02-21 at 23 05 16


escapemanuele (approved):
I'd say the change makes it better 👍

Issue #73500, opened 2023-02-17
Task: Add post-purchase screen for users who purchase an eCommerce plan from the trial


daledupreez:
### Details

We need to implement a post-purchase screen for users who upgrade from the free trial to the eCommerce plan.

From an implementation perspective, detecting when an eCommerce purchase occurred on a site that was previously on the free trial may be really tricky in the Calypso context. It may be easier to specify a redirect URL when we send users to checkout, but that may be complicated if any upsells directly add products to the cart.

While building the page, it will also be worth confirming whether we should add some sort of escape hatch that doesn't require one of the next steps cards to be selected.

Design details:
 - peapX7-1yI-p2
 - jBwlERMS350NqhJNHZzAVF-fi-98%3A33334&t=xbjxJ2I9nnqtCPTA-0

### Checklist

- [ ] After I upgrade to the eCommerce plan from the free trial, I should see the post-checkout screen with next steps
- [ ] Add actions to the next-step cards - flows are catpured here: lD0gTB4IqGsmSmsnjZNaP6-fi-470%3A17348&t=oWgWy5zHSYbWIcS8-0
- [ ] Add dynamic confetti

that-mike-bal:
@daledupreez  why not just take them back to the plan page to show they are off the free trial and on the e-commerce plan or back to my home and their task list?

daledupreez:
That's as much a question for @veronicafasulo as it is for me. Any of the above could work, but we do need something other than the default thank you experience.

daledupreez:
With #74064 and #74093 being merged, we have now addressed all open items relating to the post-purchase UI, so I am going to close this issue.

PR #73627, opened 2023-02-22
WooExpress: Added trial upgrade confirmation page

nelsonec87:

## Proposed Changes

Figma: jBwlERMS350NqhJNHZzAVF-fi-98%3A33334

* Added a new route to place the confirmation page.
* Created a new component: `<TrialUpgradeConfirmation />` 
* Created a getter to hold the task data (copy from the designs): `upgrade-confirmation/confirmation-tasks.ts`
* Created basic structure/placeholders for the page content, including `<ConfirmationTask />`

## Outside of this PR's scope
* Full style definitions
* Mobile version
* Dynamic plan information
* Animations/confetti
* Illustrations

## Testing Instructions

   * The site slug is not being validated yet
* Please confirm that the approach to creating the new route (and the route name/path) is sensible.
* Note that the font family doesn't match the one from the designs. I'm not sure if we should change it.
   * The font used in the design is "SF Pro Text", but it's not currently loaded in the browser. Is this expected or do we need to load the font file dynamically?

image

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
@@ -1,5 +1,5 @@
 import { Card } from '@automattic/components';
-import { TranslateResult } from 'i18n-calypso';
+import type { TranslateResult } from 'i18n-calypso';
 
 import './style.scss';
 
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
@@ -1,10 +1,11 @@
 import { Card } from '@automattic/components';
+import { TranslateResult } from 'i18n-calypso';
 
 import './style.scss';
 
 interface ConfirmationTaskProps {
-	title: string;
-	subtitle: string;
+	title: TranslateResult;
+	subtitle: TranslateResult;
 	illustration: string;
 }
 
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
@@ -1,0 +1,23 @@
+import { Card } from '@automattic/components';
+
+import './style.scss';
+
+interface ConfirmationTaskProps {
+	title: string;
+	subtitle: string;
+	illustration: string;
+}
+
+const ConfirmationTask = ( props: ConfirmationTaskProps ) => {
+	const { title, subtitle } = props;
+
+	return (
+		<Card className="confirmation-task__card">
+			<span>{ title }</span>
+			<br />
+			<span>{ subtitle }</span>
+		</Card>
+	);
+};
+
+export default ConfirmationTask;
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
@@ -1,0 +1,33 @@
+import { useTranslate } from 'i18n-calypso';
+import Main from 'calypso/components/main';
+import './style.scss';
+import ConfirmationTask from './confirmation-task';
+import { getConfirmationTasks } from './confirmation-tasks';
+
+const TrialUpgradeConfirmation = () => {
+	const translate = useTranslate();
+
+	const tasks = getConfirmationTasks( { translate } );
+
+	return (
+		<Main wideLayout>
+			<div className="trial-upgrade-confirmation__header">
+				<h1 className="trial-upgrade-confirmation__title">
+					{ translate( 'Woo! Welcome to eCommerce' ) }
+				</h1>
+				<div className="trial-upgrade-confirmation__subtitle">
+					{ translate( 'Your purchase has been completed and you’re on the Commerce plan.' ) }
+					<br />
+					{ translate( "Now it's time to get creative. What would you like to do next?" ) }
+				</div>
+			</div>
+			<div className="trial-upgrade-confirmation__tasks">
+				{ tasks.map( ( task ) => (
+					<ConfirmationTask { ...task } />
+				) ) }
+			</div>
+		</Main>
+	);
+};
+
+export default TrialUpgradeConfirmation;
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
@@ -13,7 +13,7 @@ const TrialUpgradeConfirmation = () => {
 
 	return (
 		<>
-			<BodySectionCssClass bodyClass={ [ 'trial-upgraded' ] } />
+			<BodySectionCssClass bodyClass={ [ 'ecommerce-trial-upgraded' ] } />
 			<Main wideLayout>
 				<div className="trial-upgrade-confirmation__header">
 					<h1 className="trial-upgrade-confirmation__title">
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
@@ -17,10 +17,12 @@ const TrialUpgradeConfirmation = () => {
 			<Main wideLayout>
 				<div className="trial-upgrade-confirmation__header">
 					<h1 className="trial-upgrade-confirmation__title">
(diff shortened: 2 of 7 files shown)


client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { Card } from '@automattic/components';
+
+import './style.scss';
+
+interface ConfirmationTaskProps {
+	title: string;
+	subtitle: string;
```
daledupreez: If we want to get super fancy, we could type these as `TranslateResult` and not just `string`. But that's not blocking. (The main gotcha is that `translate()` returns a `TranslateResult` as the content may include some HTML tags - to work around that you need to pass in an option to type the return as a string.)
```suggestion
	title: TranslateResult;
	subtitle: TranslateResult;
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { useTranslate } from 'i18n-calypso';
+import Main from 'calypso/components/main';
+import './style.scss';
```
daledupreez: Nit: can we have the style import after the other imports?

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+const TrialUpgradeConfirmation = () => {
+	const translate = useTranslate();
+
+	const tasks = getConfirmationTasks( { translate } );
+
+	return (
+		<Main wideLayout>
+			<div className="trial-upgrade-confirmation__header">
+				<h1 className="trial-upgrade-confirmation__title">
+					{ translate( 'Woo! Welcome to eCommerce' ) }
```
daledupreez: This should be `Commerce` to match the design. More generally, I believe we're trying to move away from the `eCommerce` term, either to `ecommerce`, `Commerce`, or `Ecommerce` (when at the start of a sentence).
```suggestion
					{ translate( 'Woo! Welcome to Commerce' ) }
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+				</h1>
+				<div className="trial-upgrade-confirmation__subtitle">
+					{ translate( 'Your purchase has been completed and you’re on the Commerce plan.' ) }
```
daledupreez: I think we should use substitution for the plan name to start with.
```suggestion
					{ translate(
						'Your purchase has been completed and you’re on the %(planName)s plan.',
						{
							args: { planName: 'Commerce' } /* we should try to pull this from the plan object in calypso-products */
						}
					) }
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+					<br />
+					{ translate( "Now it's time to get creative. What would you like to do next?" ) }
```
daledupreez: We should use the same `'` or `’` character for these strings.
```suggestion
					{ translate( "Now it's time to get creative. What would you like to do next?" ) }
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/style.scss
```diff
@@ -0,0 +1,25 @@
+body {
```
daledupreez: This select needs to be more specific -- just using `body` as the selector means it will affect _every_ page after this page has been loaded, as they all have a `body` element.

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/style.scss
```diff
@@ -0,0 +1,25 @@
...
+		}
+	}
+
+	.trial-upgrade-confirmation__tasks {
+		display: grid;
+		grid-template-columns: repeat(3, 1fr);
+		grid-column-gap: 32px;
+		grid-row-gap: 32px;
+		margin-top: 35px;
+	}
```
daledupreez, line 27: This looks like it won't match the mobile display, which has things in a column format with tighter spacing. That wasn't called out in the PR description as something that wasn't implemented, so I want to flag it.

client/my-sites/plans/index.js
```diff
@@ -73,6 +74,11 @@ export default function () {
 		p2RedirectToHubPlans,
 		currentPlan
 	);
+	trackedPage(
+		'/plans/my-plan/trial-upgrade-confirmation/:domain',
```
daledupreez: In terms of the path, I think this matches our thinking, but not the user's. Maybe simplifying this to `my-plan/trial-upgraded/:domain` would be enough? That way the path reflects what just happened, rather than what the page is.

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/style.scss
daledupreez: After looking at the UI, I think we should also switch to a two-column display for some widths well before we switch to the single column for mobile.

daledupreez (commented):
Overall this is mostly working as described, but it might be clearer to identify exactly what wasn't included. However, there are some blockers here, especially the use of the raw `body` selector, and the translation items.

RE the fonts, could you please specify the font names that we are using and what we have in the designs?

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/style.scss
nelsonec87: Added a "not included" section in the first comment 👍

nelsonec87:
@daledupreez 
> RE the fonts, could you please specify the font names that we are using and what we have in the designs?

I've added as note in the 1st comment:
The font used in the design is "SF Pro Text", but it's not currently loaded in the browser. Is this expected or do we need to load the font file dynamically?

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/index.tsx
```diff
@@ -1,10 +1,11 @@
 import { Card } from '@automattic/components';
+import { TranslateResult } from 'i18n-calypso';
```
daledupreez: Nit: this should be `import type`.
```suggestion
import type { TranslateResult } from 'i18n-calypso';
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -1,32 +1,39 @@
...
-			<div className="trial-upgrade-confirmation__header">
-				<h1 className="trial-upgrade-confirmation__title">
-					{ translate( 'Woo! Welcome to eCommerce' ) }
-				</h1>
-				<div className="trial-upgrade-confirmation__subtitle">
-					{ translate( 'Your purchase has been completed and you’re on the Commerce plan.' ) }
-					<br />
-					{ translate( "Now it's time to get creative. What would you like to do next?" ) }
+		<>
+			<BodySectionCssClass bodyClass={ [ 'trial-upgraded' ] } />
```
daledupreez: Why not `trial-upgrade-confirmation` to match the class? We could also use `.ecommerce-trial-upgraded` to be a bit less likely to collide with another usage/context.
```suggestion
			<BodySectionCssClass bodyClass={ [ 'trial-upgrade-confirmation' ] } />
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/style.scss
```diff
@@ -1,5 +1,6 @@
-body {
+body.trial-upgraded {
 	background-color: var(--color-surface);
+	font-family: "SF Pro Text", $sans;
```
daledupreez: I think we could also pull this in from @automattic/components via the `$font-sf-pro-text` variable. See [_typography.scss](https://github.com/Automattic/wp-calypso/blob/4eb5bdd4705ef4b6447a34520afea4090eeac4f1/packages/components/src/styles/_typography.scss#L5) in that package.

As to loading the font, let's hold off on that for a little bit. It may be worth flagging that for review from the design team once we have the overall page in better shape.

daledupreez (approved):
LGTM! I have some non-blocking feedback on the PR, but I don't think this needs another round of review.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7829346

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7829346).

PR #73629, opened 2023-02-22
[ETK] Update `@wordpress/interface` to the latest published version

fullofcaffeine:
Update it to match the changes introduced by the same package in Gutenberg 15.2.

**IMPORTANT: This PR should be merged before releasing Gutenberg 15.2 to WPCOM.**

The mismatch between versions is causing a WSOD in simple sites (and probably in AT, too). Gutenberg 15.2 introduced some changes in this package (https://github.com/WordPress/gutenberg/pull/47498), but ETK ends up loading an older version without these changes. The ETK version is probably loaded (enqueued) after the GB version and then replaces most of it*, while the GB version still tries to import from the replaced module `selectors.js` and fails, since it contains the old code.

_*only some parts, probably because the package is code-split and ETK only loads some modules (like selectors.js)? I haven't looked deeply, this is just an assumption._

Related to https://github.com/WordPress/gutenberg/pull/47498.

Slack discussion: p1677020860615489-slack-CBTN58FTJ (also a good primer on why the `.yarmlrc.yml` changes were needed)

Sum up on Slack: p1677778450155879/1677777680.547219-slack-CBTN58FTJ

## Proposed Changes

Update the version of `@wordpress/interface` in ETK to match the one in Gutenberg 15.2.

## Testing Instructions

1. Apply this diff: D102317-code
to your sandbox, make sure your test sandboxed site has the `gutenberg-edge` sticker applied;
2. Try to create a new post
3. You'll get a WSOD
4. Sync a custom ETK build from this branch (`yarn && yarn dev --sync`)
5. Reload the page
6. It should load the editor without any fatal errors.


client/blocks/time-mismatch-warning/test/index.jsx
```diff
@@ -24,13 +24,21 @@ jest.mock( 'calypso/state/preferences/selectors', () => ( {
...
+		return 240;
+	}
+}
+
+const originalDate = global.Date;
+
 describe( 'TimeMismatchWarning', () => {
 	beforeAll( () => {
-		jest.spyOn( global, 'Date' ).mockImplementation( () => ( { getTimezoneOffset: () => 240 } ) );
+		global.Date = MockedDate;
```
fullofcaffeine, line 37: These tests were failing with:

```
   TypeError: Date.now is not a function                                                                                                                                                                           
                                                                                                                                                                                                                    
      at ../node_modules/requestidlecallback/index.js:57:21                                                                                                                                                         
      at MutationObserver.onInputorMutation (../node_modules/requestidlecallback/index.js:87:3)   
```

I haven't dug too deep, but looks like this is caused by the `requestidlecallback` package, which now uses `Date.now` . Before the package updates that are part of this PR, `Date.now` was not in any of the code paths for this test (apparently), so something changed after the update.

With the `mockedImplementation` approach, it discarded `Date`'s original implementation to only define the custom `getTimezoneOffset`. I've tried to use `spyOn` + `mockImplementation` to return a custom class based off `Date` but failed - it was much simpler to just override the global variable and re-assign the original object to it in the `afterAll` callback.

I wonder why this test triggered a portion of code that uses this package now while it didn't before the update. Maybe that package was already used, and it was recently updated to use `Date.now`. Provided tests pass, I think it's okay to consider this relatively safe to ship, but if you have more insights, let me know. Besides, in the case of globals like `Date`, I think it's safer to keep the original implementation accessible and mock the functions over it. The previous approach based on `mockImplementation` removed all other functions and properties (including the static `now`).

arthur791004:
Referring to https://github.com/Automattic/wp-calypso/pull/73078#issuecomment-1422373429, the `@wordpress/*` packages have their interconnected dependencies, I'm not sure we can only upgrade `@wordpress/interface` package. For example, there are lots of version changes in the yarn.lock, it might bring up unexpected side effects to Calypso

FYI @Automattic/team-calypso-frameworks  is working on [Try: Upgrade @wordpress packages to pre-React 18](https://github.com/Automattic/wp-calypso/pull/73155)

tyxla:
@arthur791004 the big difference between this PR and https://github.com/Automattic/wp-calypso/pull/73078 is that this is updating the `@wordpress/interface` dependency only in ETK, and not for the rest of Calypso monorepo. 

I've taken a look around the `yarn.lock` changes in this PR and I don't have any concerns. There are several additional changes caused by dependency deduplication, but they are fine because they resolve to versions of the packages that still use React 17. The only versions that use React 18 are packages that are specifically only used by ETK, so I don't see how the rest of Calypso could be affected.

package.json
```diff
@@ -300,7 +300,9 @@
 		"newspack-components/@wordpress/element": "4.7.0",
 		"newspack-components/@wordpress/i18n": "4.9.0",
 		"keytar@npm:7.7.0/node-addon-api": "3.1.0",
-		"lzma-native": "^8.0.5"
+		"lzma-native": "^8.0.5",
+		"@wordpress/primitives/@wordpress/element": "4.20.0",
```
tyxla: Should we pin this to a particular version of `@wordpress/primitives` to avoid this propagating to another version of `@wordpress/primitives` and accidentally into the rest of the monorepo?

client/blocks/time-mismatch-warning/test/index.jsx
tyxla: No concerns from my side about this change. We've seen similar issues in Gutenberg tests too. It's usually that something tries to access `Date` before the `jsdom` mock has been setup, or something inadvertently overrides it for the test.

arthur791004:
Got it! Looks like I misunderstood some changes in `yarn.lock` 😅

yarn.lock
```diff
@@ -7010,12 +7103,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/react-dom@npm:*, @types/react-dom@npm:>=16.9.0, @types/react-dom@npm:^17.0.11":
-  version: 17.0.13
-  resolution: "@types/react-dom@npm:17.0.13"
+"@types/react-dom@npm:*, @types/react-dom@npm:>=16.9.0, @types/react-dom@npm:^18.0.6":
+  version: 18.0.11
+  resolution: "@types/react-dom@npm:18.0.11"
```
tyxla: This is where things could eventually get thorny, but if there are no TS errors introduced, we're likely fine for now. We should keep an eye on this. Ideally, we should upgrade to the newest `@wordpress` packages and upgrade the repo to use React 18 as soon as possible.

yarn.lock
```diff
@@ -8138,14 +8240,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/a11y@npm:^3.10.0, @wordpress/a11y@npm:^3.13.0, @wordpress/a11y@npm:^3.9.0":
-  version: 3.13.0
-  resolution: "@wordpress/a11y@npm:3.13.0"
+"@wordpress/a11y@npm:^3.10.0, @wordpress/a11y@npm:^3.13.0, @wordpress/a11y@npm:^3.27.0, @wordpress/a11y@npm:^3.9.0":
+  version: 3.27.0
+  resolution: "@wordpress/a11y@npm:3.27.0"
```
tyxla: Seems like this is forcing the rest of Calypso to use `@wordpress/a11y@3.27.0` - is it intentional? I think we should refrain from affecting the rest of the monorepo.

yarn.lock
```diff
@@ -8510,14 +8691,40 @@ __metadata:
...
+    redux: ^4.1.2
+    turbo-combine-reducers: ^1.0.2
+    use-memo-one: ^1.1.1
+  peerDependencies:
+    react: ^18.0.0
+  checksum: e1b3c1e124ec6e540f192986348cc9fa553ad2aad8385a69f165125cc41ca4601f43e817e8a70c9c3254c4505c7f2dc522c14e4695499d7b2e11c0cbef65f84b
+  languageName: node
+  linkType: hard
+
+"@wordpress/date@npm:^4.13.0, @wordpress/date@npm:^4.27.0, @wordpress/date@npm:^4.9.0":
```
tyxla, line 8719: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8545,33 +8752,32 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/deprecated@npm:^3.13.0, @wordpress/deprecated@npm:^3.8.0, @wordpress/deprecated@npm:^3.9.0":
-  version: 3.13.0
-  resolution: "@wordpress/deprecated@npm:3.13.0"
+"@wordpress/deprecated@npm:^3.13.0, @wordpress/deprecated@npm:^3.27.0, @wordpress/deprecated@npm:^3.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8545,33 +8752,32 @@ __metadata:
...
-  checksum: 9a62783134dabac4210e8001c34d6b9237a9657552cb8a482501133080fb7f885273580abcfe69cd05bfe27110292b2c49192408e8c7f56d0ec6c346da3ab68f
+    "@wordpress/hooks": ^3.27.0
+  checksum: 1d95e20de156f354b013d8803b912b007e4625974c35784be47138e60447ac8fa43d508403e26d924e66397eaec8302f6e70205f29aec407b464bd97de4d9dc3
   languageName: node
   linkType: hard
 
-"@wordpress/dom-ready@npm:^3.13.0, @wordpress/dom-ready@npm:^3.9.0":
-  version: 3.13.0
-  resolution: "@wordpress/dom-ready@npm:3.13.0"
+"@wordpress/dom-ready@npm:^3.27.0, @wordpress/dom-ready@npm:^3.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8545,33 +8752,32 @@ __metadata:
...
     "@babel/runtime": ^7.16.0
-  checksum: 03b4b7db1b67115c3aef6c0f9bc3bca0a1e3d8445c449dd77d6703114ce132071b3ec87a0810cbe385947ebf4edbb043730d01f56e1937d0fda70075315b332c
+  checksum: 0ef2f2f0267f889712dbd92017432dd7c39ac26301f82ceda2c85720a174f948065f74b0c5c21191aa85de0ed935737dcbce81961605bf3bc464c520828d9d2a
   languageName: node
   linkType: hard
 
-"@wordpress/dom@npm:^3.13.0, @wordpress/dom@npm:^3.9.0":
-  version: 3.13.0
-  resolution: "@wordpress/dom@npm:3.13.0"
+"@wordpress/dom@npm:^3.13.0, @wordpress/dom@npm:^3.27.0, @wordpress/dom@npm:^3.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8704,6 +8910,22 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@wordpress/element@npm:4.20.0, @wordpress/element@npm:^4.1.1, @wordpress/element@npm:^4.11.0, @wordpress/element@npm:^4.19.0, @wordpress/element@npm:^4.5.0, @wordpress/element@npm:^4.7.0":
```
tyxla: Same as for other `@wordpress/a11y` above. I'm concerned that some of those packages could have unexpected effects on the rest of Calypso. It's nice that our e2e tests pass, but then they don't provide a 100% coverage. WDYT?

yarn.lock
```diff
@@ -8796,12 +9018,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/escape-html@npm:^2.13.0, @wordpress/escape-html@npm:^2.2.0, @wordpress/escape-html@npm:^2.21.0, @wordpress/escape-html@npm:^2.9.0":
-  version: 2.21.0
-  resolution: "@wordpress/escape-html@npm:2.21.0"
+"@wordpress/escape-html@npm:^2.13.0, @wordpress/escape-html@npm:^2.2.0, @wordpress/escape-html@npm:^2.22.0, @wordpress/escape-html@npm:^2.27.0, @wordpress/escape-html@npm:^2.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8863,21 +9085,21 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/hooks@npm:^3.13.0, @wordpress/hooks@npm:^3.21.0, @wordpress/hooks@npm:^3.9.0":
-  version: 3.21.0
-  resolution: "@wordpress/hooks@npm:3.21.0"
+"@wordpress/hooks@npm:^3.13.0, @wordpress/hooks@npm:^3.27.0, @wordpress/hooks@npm:^3.9.0":
```
tyxla, line 9088: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8863,21 +9085,21 @@ __metadata:
...
     "@babel/runtime": ^7.16.0
-  checksum: 3d63aef536dc382ce926d58877176a33c57976012c8e8a7aeaa54384da309fe7307f1d388f562d5584599db752091f9e3e4b24dcdeb63450a2a4fc9bdc7b94c6
+  checksum: f82c96f79a4ec2b9e5cc72e199a539823ed90b38b855577664fb66467dc9803d1387885d3bc2aa962656d6b2a91331e8b0db8b64ab5612a5ce075d0332b4bf2b
   languageName: node
   linkType: hard
 
-"@wordpress/html-entities@npm:^3.9.0":
-  version: 3.9.0
-  resolution: "@wordpress/html-entities@npm:3.9.0"
+"@wordpress/html-entities@npm:^3.27.0, @wordpress/html-entities@npm:^3.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8898,19 +9120,19 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/i18n@npm:^4.10.0, @wordpress/i18n@npm:^4.13.0, @wordpress/i18n@npm:^4.2.4, @wordpress/i18n@npm:^4.21.0, @wordpress/i18n@npm:^4.9.0":
-  version: 4.21.0
-  resolution: "@wordpress/i18n@npm:4.21.0"
+"@wordpress/i18n@npm:^4.10.0, @wordpress/i18n@npm:^4.13.0, @wordpress/i18n@npm:^4.2.4, @wordpress/i18n@npm:^4.21.0, @wordpress/i18n@npm:^4.27.0, @wordpress/i18n@npm:^4.9.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -8936,14 +9158,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/icons@npm:^9.0.0, @wordpress/icons@npm:^9.11.0, @wordpress/icons@npm:^9.4.0, @wordpress/icons@npm:^9.5.0":
-  version: 9.11.0
-  resolution: "@wordpress/icons@npm:9.11.0"
+"@wordpress/icons@npm:^9.0.0, @wordpress/icons@npm:^9.11.0, @wordpress/icons@npm:^9.18.0, @wordpress/icons@npm:^9.4.0, @wordpress/icons@npm:^9.5.0":
```
tyxla, line 9161: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -9028,14 +9274,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/keycodes@npm:^3.10.0, @wordpress/keycodes@npm:^3.13.0, @wordpress/keycodes@npm:^3.9.0":
-  version: 3.13.0
-  resolution: "@wordpress/keycodes@npm:3.13.0"
+"@wordpress/keycodes@npm:^3.10.0, @wordpress/keycodes@npm:^3.13.0, @wordpress/keycodes@npm:^3.27.0, @wordpress/keycodes@npm:^3.9.0":
```
tyxla, line 9277: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -9153,23 +9433,33 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/primitives@npm:^3.1.1, @wordpress/primitives@npm:^3.11.0, @wordpress/primitives@npm:^3.18.0, @wordpress/primitives@npm:^3.5.0, @wordpress/primitives@npm:^3.7.0, @wordpress/primitives@npm:^3.8.0":
-  version: 3.18.0
-  resolution: "@wordpress/primitives@npm:3.18.0"
+"@wordpress/primitives@npm:^3.1.1, @wordpress/primitives@npm:^3.11.0, @wordpress/primitives@npm:^3.25.0, @wordpress/primitives@npm:^3.5.0, @wordpress/primitives@npm:^3.7.0, @wordpress/primitives@npm:^3.8.0":
```
tyxla: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -9194,17 +9484,17 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wordpress/redux-routine@npm:^4.13.0":
-  version: 4.13.0
-  resolution: "@wordpress/redux-routine@npm:4.13.0"
+"@wordpress/redux-routine@npm:^4.13.0, @wordpress/redux-routine@npm:^4.27.0":
```
tyxla, line 9487: Same as for other `@wordpress/a11y` above.

yarn.lock
```diff
@@ -23739,10 +24088,10 @@ fsevents@^1.2.7:
   languageName: node
   linkType: hard
 
-"moment@npm:>= 2.9.0, moment@npm:>=1.6.0, moment@npm:^2.19.3, moment@npm:^2.22.1, moment@npm:^2.26.0":
-  version: 2.27.0
-  resolution: "moment@npm:2.27.0"
-  checksum: ea5b2fd73c8ad9070fc82c21ccb6226f7556c27a6a305e64371a3d1223729e9f9e75da59c3114160f21f9181e74f6d245534093179825eb20bf0f40ec7495d92
+"moment@npm:>= 2.9.0, moment@npm:>=1.6.0, moment@npm:^2.19.3, moment@npm:^2.22.1, moment@npm:^2.26.0, moment@npm:^2.29.4":
```
tyxla, line 24091: Are we sure this doesn't break something? Moment is quite broadly used throughout Calypso.

tyxla (commented):
I've taken a more extensive look over the `yarn.lock` updates and have some concerns - we this package upgrade, and specifically the deduplication might have unexpected effects on the rest of Calypso. I'm not sure we intended to upgrade all those `@wordpress` packages for the test of the monorepo.

tyxla:
> Got it! Looks like I misunderstood some changes in `yarn.lock` 😅

I've taken a more thorough look and I think I understand your concerns. Left some comments above to raise those concerns as well.

tyxla:
@fullofcaffeine I wonder if it's worth investing some time in finishing https://github.com/Automattic/wp-calypso/pull/73155 that upgrades all packages together to the pre-React 18 versions instead.

tyxla:
I've closed #73155 since it went far from what it should have been. I think we should proceed with upgrading to newer versions of the packages, but not with such a giant leap forward - rather in smaller chunks.

fullofcaffeine:
>  I'm not sure we intended to upgrade all those @wordpress packages for the test of the monorepo.

Hmm, not intentional, no; that was the (indirect?) result of either upgrading `@wordpress/interface` or the dedupes, maybe?

fullofcaffeine:
> Should we pin this to a particular version of @wordpress/primitives to avoid this propagating to another version of @wordpress/primitives and accidentally into the rest of the monorepo?

@tyxla Would you suggest reinstalling and not deduping? The problem is that the build might fail to ask us to dedupe. To make sure we're on the same page: when you say "pin", I assume you mean explicitly defining the version of these packages in the root `package.json`? Please advise.

fullofcaffeine:
>  the @wordpress/* packages have their interconnected dependencies, I'm not sure we can only upgrade @wordpress/interface package. For example, there are lots of version changes in the yarn.lock, it might bring up unexpected side effects to Calypso

That's true. The title for this PR might seem a bit misleading, but I wanted to point out that this was about this specific package, that originally caused the issue (and needed to be updated to match the version in Gutenberg). The alternative would be to explore extracting it using the Dependency Extraction plugin, but not sure if it'd have side effects in Gutenberg/WP (cc @gziolo)

tyxla:
> > Should we pin this to a particular version of @wordpress/primitives to avoid this propagating to another version of @wordpress/primitives and accidentally into the rest of the monorepo?
> 
> @tyxla Would you suggest reinstalling and not deduping? The problem is that the build might fail to ask us to dedupe. To make sure we're on the same page: when you say "pin", I assume you mean explicitly defining the version of these packages in the root `package.json`? Please advise.

Yes and yes - I think that would be the only way to ensure that we don't affect the rest of Calypso inadvertently. The downside is that we might and will end up shipping multiple versions of the same package, which is why `yarn` errors and suggests deduplication. But in this case, we don't want to deduplicate, right? IIRC, we want exactly the opposite.

fullofcaffeine:
Yeah, that's a very good point about not deduping, my bad. Deduping will tend to re-organize/DRY packages and that probably means moving them down the monorepo tree.

fullofcaffeine:
I remember one of the test builds failed once and asked me to dedupe, though, but I guess we can just skip that specific check.

fullofcaffeine:
One thing that worries me is that if someone else runs `dedupe` at a later point while updating/installing another package, we will have the same issue.

package.json
fullofcaffeine: Good point 👍🏻

fullofcaffeine: Hmm, after resetting the `yarn.lock` to the `trunk` version and re-installing the new version of `@wordpress/interface` in ETK without deduping, these resolution overrides don't seem to be needed anymore, the storybook task is working fine.

fullofcaffeine: I removed those entries.

tyxla:
The problem with the current state of the dependencies is that there are multiple dependencies that satisfy a set of requirements. In that case, `yarn` will always complain that it can deduplicate them. The only way to avoid that is to set more restrictive dependencies.

I've tried with a bit more strict dependencies in order to both satisfy the previous requirements and satisfy yarn to not think that deduplication is possible. I've reverted those changes (left them in place if you'd like to take a look) because there were still unnecessary updates to other packages.

My conclusion from working on this PR and other `@wordpress` package upgrade PRs recently is that we're too permissive with our dependencies. Being permissive, we allow for too much variation and that causes problems in these scenarios when we actually want to pin to specific versions. So even if we use `~` for all `@wordpress` dependencies, there are still unwanted updates affecting the rest of the codebase. Part of the problem is that we probably haven't been too diligent in deciding which change is a minor one and which is a major one in our `@wordpress` package releases, but there's not much we can do about that now.

It seems like we're blocked by this problem. So, what would I recommend? Exploring making the dependencies more strict in a separate PR, either by using `~` (which doesn't always seem to work well as my reverted commits indicated), or pinning to specific versions. This would make the dependencies a bit more predictable and upgrading one of them (and measuring/testing its impact) would be easier. 

The downside of that strict dependency approach would be that upgrades of `@wordpress` packages will have to occur more often. The other potential downside is that there might be more package duplication than we want. So it's not a situation we can remain in forever; rather, we can stay with strict dependencies until we get in line with the latest releases. At that point, once Renovate is enabled and we merge updates regularly, we could seek to make them less restrictive again. 

Let me know what you think.

gziolo:
> The title for this PR might seem a bit misleading, but I wanted to point out that this was about this specific package, that originally caused the issue (and needed to be updated to match the version in Gutenberg). The alternative would be to explore extracting it using the Dependency Extraction plugin, but not sure if it'd have side effects in Gutenberg/WP

Yes, it's excluded from the mechanics on purpose in the Gutenberg project and WordPress core to avoid creating an entry point and exposing it under `wp.interface` which would have backward compatibility implications.

fullofcaffeine:
Closed in favor of https://github.com/Automattic/wp-calypso/pull/73856.

PR #73651, opened 2023-02-22
Theme Showcase: Collapsible style variation previews

taipeicoder:
## Proposed Changes

As per the design spec in 0EogtxvlmxMGcnEEaMVECv-fi-2412%3A63031, and Slack thread p1676665429018439-slack-CRWCHQGUB, we want to improve the style variation preview UX in the Theme Detail page.

Screenshot 2023-02-22 at 4 14 35 PM

This PR implements the following behaviors:
- Always show "Default" for the default style variation.
- In mobile view, style variation previews scroll horizontally.
- In tablet/desktop view, style variation previews should be collapsible via a "Show all" button if they overflow the container width.
  - Show the first row of style variation previews by default.
  - Clicking the "Show all" button will expand to show all the remaining style variation previews.

## Testing Instructions

* Ensure that the style preview layout is as described above.
* Ensure that the "Show all" button works as described above.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/theme/theme-style-variations/index.tsx
+++ b/client/my-sites/theme/theme-style-variations/index.tsx
@@ -26,6 +26,7 @@ const ThemeStyleVariations = ( {
 	const [ collapsibleMaxHeight, setCollapsibleMaxHeight ] = useState< number | undefined >();
 	const [ isCollapsible, setIsCollapsible ] = useState< boolean >( false );
 	const [ isCollapsed, setIsCollapsed ] = useState< boolean >( false );
+	const isCollapsibleRef = useRef( false );
 
 	useLayoutEffect( () => {
 		if ( ! observerRef.current ) {
@@ -42,9 +43,11 @@ const ThemeStyleVariations = ( {
 
 			// Detect flex wrap.
 			const currentIsCollapsible = nodeLastChild.offsetTop > nodeFirstChild.offsetTop;
-			const shouldCollapse = isCollapsible !== currentIsCollapsible || isCollapsed;
+			const shouldCollapse = isCollapsibleRef.current !== currentIsCollapsible || isCollapsed;
 
 			setIsCollapsible( currentIsCollapsible );
+			isCollapsibleRef.current = currentIsCollapsible;
+
 			setIsCollapsed( shouldCollapse );
 			setCollapsibleMaxHeight( shouldCollapse ? nodeFirstChild.offsetHeight : node.scrollHeight );
 		} );
@@ -53,7 +56,7 @@ const ThemeStyleVariations = ( {
 		return () => {
 			resizeObserver.disconnect();
 		};
-	}, [ isCollapsible, isCollapsed ] );
+	}, [ isCollapsed ] );
 
 	const handleCollapseButtonClick = () => {
 		setIsCollapsed( ! isCollapsed );
--- a/client/my-sites/theme/theme-style-variations/index.tsx
+++ b/client/my-sites/theme/theme-style-variations/index.tsx
@@ -1,6 +1,6 @@
 import { Button } from '@automattic/components';
 import { PremiumBadge } from '@automattic/design-picker';
-import { useEffect, useLayoutEffect, useRef, useState } from '@wordpress/element';
+import { useLayoutEffect, useRef, useState } from '@wordpress/element';
 import classNames from 'classnames';
 import { translate } from 'i18n-calypso';
 import AsyncLoad from 'calypso/components/async-load';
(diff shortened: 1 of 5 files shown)


client/my-sites/theme/theme-style-variations/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+			setIsCollapsible( currentIsCollapsible );
+			setIsCollapsed( shouldCollapse );
+			setCollapsibleMaxHeight( shouldCollapse ? nodeFirstChildRect.height : node.scrollHeight );
+		} );
+
+		resizeObserver.observe( observerRef.current );
+		return () => {
+			resizeObserver.disconnect();
+		};
+	}, [ isCollapsible, isCollapsed ] );
```
arthur791004: We can use `useRef` to get the latest value in the callback of the resize observer

client/my-sites/theme/theme-style-variations/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+		const resizeObserver = new ResizeObserver( ( [ observerNode ] ) => {
+			const node = observerNode.target;
+			const nodeFirstChildRect = ( node.firstChild as HTMLElement )?.getBoundingClientRect();
+			const nodeLastChildRect = ( node.lastChild as HTMLElement )?.getBoundingClientRect();
+			if ( ! nodeFirstChildRect || ! nodeLastChildRect ) {
+				return;
+			}
+
+			// Detect flex wrap.
+			const currentIsCollapsible = nodeFirstChildRect.top !== nodeLastChildRect.top;
```
arthur791004: [nit] we can check the offsetTop of the last element to detect whether flex wrap and then we don't need to call `getBoundingClientRect` all the time.

```js
const currentIsCollapsible = node.lastChild.offsetTop > paddingTop
```

client/my-sites/theme/theme-style-variations/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+			const shouldCollapse = isCollapsible !== currentIsCollapsible || isCollapsed;
+
+			setIsCollapsible( currentIsCollapsible );
+			setIsCollapsed( shouldCollapse );
+			setCollapsibleMaxHeight( shouldCollapse ? nodeFirstChildRect.height : node.scrollHeight );
```
arthur791004: [nit] The collapsed height could be `node.firstChild.offsetHeight`

taipeicoder: While `offsetHeight` isn't exactly the same value as `getBoundingClientRect().height` (the former value seems rounded). I think it's close enough to justify the performance gain 🙂

client/my-sites/theme/theme-style-variations/index.tsx
taipeicoder: I'm not sure what you meant with `paddingTop` so I just changed it to compare the `offsetTop` of `firstChild` and `lastChild`. We still manage to avoid using `getBoundingClientRect()` this way!

client/my-sites/theme/theme-style-variations/index.tsx
taipeicoder: Could you explain more? Do you mean to use `useRef` instead of `useState` for `isCollapsible` and `isCollapsed`? Perhaps I'm missing something, but don't we need to update the state since we need to re-render the component (adding CSS classes, etc)? 🤔

arthur791004: I mean we can create a new variable and its value is from `useRef` so that we can avoid the resize observer disconnecting and reconnecting whenever the state changes. It's not a blocker for this PR 🙂

```js
const isCollapsibleRef = useRef();
const isCollapsedRef = useRef();

isCollapsibleRef.current = isCollapsible;
isCollapsedRef.current = isCollapsed;

useEffect( () => {
  ...
  // We don't need to list any deps here
}, [] );
```

What do you think?

client/my-sites/theme/theme-style-variations/index.tsx
arthur791004: Oh yes, that also makes senses 👍

arthur791004 (approved):

client/my-sites/theme/theme-style-variations/index.tsx
taipeicoder: Hmmm, I managed to use this approach for `isCollapsible` but not for `isCollapsed`. I think it doesn't work for `isCollapsed` since the `ResizeObserver` callback isn't called when the state changes. Whereas it works for `isCollapsible` since it only changes state when resizing, so the `ResizeObserver` callback is called.

Am I missing anything?

arthur791004: I think we should update the value of ref outside the `useEffect`

taipeicoder: @arthur791004 I've updated the code to what I _think_ is what you meant. It's not my usual coding approach, so let me know if it looks good to you!

client/my-sites/theme/theme-style-variations/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+const ThemeStyleVariations = ( {
+	description,
+	selectedVariation,
+	variations,
+	onClick,
+}: ThemeStyleVariationsProps ) => {
+	const observerRef = useRef< HTMLDivElement | null >( null );
+	const [ collapsibleMaxHeight, setCollapsibleMaxHeight ] = useState< number | undefined >();
```
arthur791004, line 26: [nit] Maybe we don't need to store the max height in the state 🤔

```js
const maxHeight = isCollapsed ? observerRef.current?.firstChild?.offsetHeight : observerRef.current?.scrollHeight;
```

arthur791004 (approved):
Tested and looks good 👍

client/my-sites/theme/theme-style-variations/index.tsx
taipeicoder: Hmmm I think we need to update the max height in the `ResizeObsever` callback as well since the style variation size dimensions might change according to the viewport 😞 We can optimize this further later 🙂

arthur791004: Got it. I forgot the style variation size dimensions might also change 😓

PR #73664, opened 2023-02-22
Backups will no longer be limited by the number of days in the policy

mavegaf:
The dates of the backups to be shown are limited by the number of days in the policy according to the plan. For example, it doesn't show backups older than 30 days.

This restriction is no longer valid since the plans offer a variable number of days of backups, and if the backups have stopped, the last backups will always be available, even if they are older than a number of days. For example, the backups in my site stopped a month ago, but my last 7 days of backups are still available. 

This change addresses this issue, allowing the user to access their old backups if the backups have stopped and they are older than the limit of the plan

## Testing Instructions

You need a site with old backups that are older than the limit in the policy. Using the link generated, you can verify that you can now access them in the backup section. 

*

## Pre-merge Checklist


client/my-sites/backup/backup-date-picker/hooks.ts
```diff
@@ -62,7 +51,7 @@ export const useCanGoToDate: CanGoToDateHook = (
 			// then everything's fine (this should never happen)
 			return true;
 		},
-		[ selectedDateIsVisible, selectedDate, today, oldestDateAvailable, hasNoBackups ]
+		[ true, selectedDate, today, oldestDateAvailable, hasNoBackups ]
```
Initsogar: Is `true` needed on this dependency array for `useCallback`? Given that we removed the `selectedDateIsVisible` logic inside this hook, we should only re-create it if the other dependencies change (`selectedDate`, `today`, `oldestDateAvailable`, `hasNoBackups`).

Initsogar (approved):
Just tested with a site that has valid backups from January 14th before and after the patch, and it is working as expected.

| Before the patch | After the patch |
|---|---|
| image | image |

Before the patch, it didn't display me backups before Jan 24th, but after the patch it does.

In addition I left a comment regarding the dependencies array of one of the useCallback(). Once that is resolved, we can proceed with the merge.

client/my-sites/backup/backup-date-picker/hooks.ts
mavegaf: You are right. Removed. Thanks!

PR #73666, opened 2023-02-22
Design picker: Fix selected disc style for Safari

mmtr:
## Proposed Changes

Looks like Safari cannot have an element with both `border-radius` and `outline`, which was causing the selected variation disc on the Design picker to have a square border.

This PR fixes that by implementing a workaround that uses `box-shadow`.

Before | After
--- | ---
Screenshot 2023-02-22 at 17 12 53 | Screenshot 2023-02-22 at 17 13 16
## Testing Instructions

- In Safari, go to the Calypso live link
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>`
- Make sure the selected variation discs have a rounded blue border

### DIFF
--- a/packages/design-picker/src/components/style-variation-badges/style.scss
+++ b/packages/design-picker/src/components/style-variation-badges/style.scss
@@ -47,7 +47,9 @@
 .style-variation__badge-is-selected {
 	> * {
 		cursor: default;
-		outline: 2px solid var(--studio-blue-30);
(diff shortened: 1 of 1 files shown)


BogdanUngureanu (approved):
Looks great in Safari! 🎉  I've also looked in Chrome and there're no regressions. :shipit:

PR #73673, opened 2023-02-22
WooExpress: Added trial upgrade confirmation card

nelsonec87:

Figma: jBwlERMS350NqhJNHZzAVF-fi-98%3A33334

## Proposed Changes

This PR builds on top of https://github.com/Automattic/wp-calypso/pull/73627 and 
* Adds the illustration files
* Adds the styles for the task cards

## Testing Instructions

* Check if cards' font sizes, colors, paddings, margins, hover effect, copy, and illustrations are correct

image


client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
```diff
@@ -1,4 +1,10 @@
 import { translate as i18nTranslate } from 'i18n-calypso';
+import customDomain from 'calypso/assets/images/plans/wpcom/custom-domain.png';
+import customize from 'calypso/assets/images/plans/wpcom/customize.png';
+import manage from 'calypso/assets/images/plans/wpcom/manage.png';
```
daledupreez: I can't comment on the image, so I am going to comment here: I think the image name should be changed to `manage-woocommerce.png` or `manage-woo.png` due to the presence of the "Woo" in the picture.

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/style.scss
```diff
@@ -1,4 +1,32 @@
 .confirmation-task__card {
+	--subtitle-font-size: 0.8125rem;
+	border-radius: 2px;
+	box-shadow: 0 0 0 1px var(--color-neutral-5);
+	cursor: pointer;
+
+	&:hover {
+		box-shadow: 0 0 0 2px var(--color-accent-50);
+	}
```
daledupreez: Might it be simpler to implement these as a `border` instead of `box-shadow`? (The Figma designs use `border`.)

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/style.scss
```diff
@@ -1,4 +1,32 @@
...
+
 	margin: 0;
 	padding: 40px 25px;
```
daledupreez, line 10: It doesn't look like you've implement the mobile styles, where the padding shifts to `32px` all round.

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/style.scss
```diff
@@ -1,4 +1,32 @@
 .confirmation-task__card {
+	--subtitle-font-size: 0.8125rem;
```
daledupreez, line 4: It feels like we should try to push this into a variable much higher up, possibly with a Woo-related name. But I think we can tackle that in a follow-up if you prefer. Having this be defined so locally is a recipe for longer-term trouble.

daledupreez (approved):
Overall, this is looking good. The only blocker is not having the different padding for mobile.

I also think we should tackle the font size variable ASAP, as we should make sure we have a clean way to specify the 13px font size across our WooCommerce-related PRs.

nelsonec87:
Rebased to fix conflicts before merge

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-task/style.scss
nelsonec87: I ended up using box-shadow here because the card component already uses it, and it also solves the box-sizing problem because we have 2 different border-widths.  
I'm merging as is since you also don't consider it blocking.

PR #73683, opened 2023-02-22
GitHub integration: soft release it for Automatticians

zaguiini:
## Proposed Changes

As part of this first iteration, let's soft-release the feature.

The idea is to gather feedback on the CfT, and usage from within Automattic, including Team 51, then use the feedback to hone the experience and functionalities. Once we achieve that state, we can release it to a wider public (a.k.a. everyone).

## Testing Instructions

Check that the feature shows up in WPCOM if you're logged in with your A8C WPCOM account.

Check that it doesn't show up if you log in with a different account that is not related to Automattic.


p-jackson (approved):
Great! 👍 
Card is hidden for a non-a8c account and no GitHub related network requests are made.

PR #73686, opened 2023-02-22
VAT: Change instances of 'VAT' to generic 'Business Tax ID' in Calypso

jjchrisdiehl:
## Proposed Changes

This is part one of fixing [this issue](https://github.com/Automattic/payments-shilling/issues/1400).

Since we are going to dynamically load B2B tax names depending on the customer's locale, the first step is to set up the fallback version of these tax names instead of defaulting to 'VAT'.

This series of changes aims to cover the cases noted by @michaeldcain in the link above.

Related to https://github.com/Automattic/payments-shilling/issues/1400

## Testing Instructions
Checkout up this branch and go to:
- http://calypso.localhost:3000/me/purchases/billing
- http://calypso.localhost:3000/me/purchases/vat-details
- http://calypso.localhost:3000/checkout/[YOUR SITE NAME]

Please cross reference the screenshots shown in the issue linked above and verify that the 'VAT' text is replaced with 'Business Tax ID'.

**Notes**
In some cases, Business Tax ID is a bit wordy and awkward, so not sure if we should use a truncated version (or if this presents perhaps legal greyarea), for instance:

> "We currently only provide Business Tax ID invoices to users who are properly listed in **Business Tax ID databases**. **Business Tax ID information** saved on this page will be applied to all of your account's receipts."

Could be:

> "We currently only provide Business Tax ID invoices to users who are properly listed in _tax ID databases_. _Tax ID information_ saved on this page will be applied to all of your account's receipts."

Also, not exactly sure how to handle the VAT text in checkout line items:

image

### DIFF
--- a/client/me/purchases/titles.js
+++ b/client/me/purchases/titles.js
@@ -56,7 +56,7 @@ Object.defineProperties( titles, {
 		get: () => i18n.translate( 'Payment Methods' ),
 	},
 	vatDetails: {
-		get: () => i18n.translate( 'VAT Details' ),
+		get: () => i18n.translate( 'Business Purchase ID Details' ),
 	},
 } );
 
--- a/client/me/purchases/titles.js
+++ b/client/me/purchases/titles.js
@@ -56,7 +56,7 @@ Object.defineProperties( titles, {
 		get: () => i18n.translate( 'Payment Methods' ),
 	},
 	vatDetails: {
-		get: () => i18n.translate( 'Business Purchase ID Details' ),
+		get: () => i18n.translate( 'Business Tax ID details' ),
 	},
 } );
 
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -28,7 +28,9 @@ export default function VatInfoPage() {
 	if ( fetchError ) {
 		return (
 			<div className="vat-info">
-				<CompactCard>{ translate( 'An error occurred while fetching VAT details.' ) }</CompactCard>
+				<CompactCard>
+					{ translate( 'An error occurred while fetching Business Tax ID details.' ) }
+				</CompactCard>
 			</div>
 		);
 	}
@@ -44,11 +46,11 @@ export default function VatInfoPage() {
 			<Column type="sidebar">
 				<Card className="vat-info__sidebar-card">
 					<CardHeading tagName="h1" size={ 16 } isBold={ true } className="vat-info__sidebar-title">
-						{ translate( 'VAT Information' ) }
+						{ translate( 'Business Tax ID Information' ) }
 					</CardHeading>
 					<p className="vat-info__sidebar-paragraph">
 						{ translate(
-							"We currently only provide VAT invoices to users who are properly listed in the VIES (VAT Information Exchange System) or the UK VAT databases. VAT information saved on this page will be applied to all of your account's receipts."
+							"We currently only provide Business Tax ID invoices to users who are properly listed in Business Tax ID databases. Business Tax ID information saved on this page will be applied to all of your account's receipts."
 						) }
 					</p>
 				</Card>
@@ -104,7 +106,7 @@ function VatForm() {
 				{ isVatAlreadySet && (
 					<FormSettingExplanation>
 						{ translate(
-							'To change your VAT number, {{contactSupportLink}}please contact support{{/contactSupportLink}}.',
+							'To change your Business Tax ID number, {{contactSupportLink}}please contact support{{/contactSupportLink}}.',
 							{
 								components: {
 									contactSupportLink: (
@@ -236,7 +238,9 @@ function useDisplayVatNotices( {
 			reduxDispatch( removeNotice( 'vat_info_notice' ) );
 			reduxDispatch(
 				errorNotice(
-					translate( 'Your VAT details are not valid. Please check each field and try again.' ),
+					translate(
+						'Your Business Tax ID details are not valid. Please check each field and try again.'
+					),
 					{ id: 'vat_info_notice' }
 				)
 			);
@@ -248,7 +252,7 @@ function useDisplayVatNotices( {
 			reduxDispatch(
 				errorNotice(
 					translate(
-						'An error occurred while updating your VAT details. Please try again or contact support.'
+						'An error occurred while updating your Business Tax ID details. Please try again or contact support.'
 					),
 					{
 						id: 'vat_info_notice',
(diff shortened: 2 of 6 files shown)


jjchrisdiehl:
In some places the usage of 'Business Tax ID' is a bit repetitive, could we perhaps change subsequent mentions to 'tax id' like so:

image

client/me/purchases/titles.js
```diff
@@ -56,7 +56,7 @@ Object.defineProperties( titles, {
 		get: () => i18n.translate( 'Payment Methods' ),
 	},
 	vatDetails: {
-		get: () => i18n.translate( 'VAT Details' ),
+		get: () => i18n.translate( 'Business Purchase ID Details' ),
```
sirbrillig: Should this be `Business Tax ID details` (or `Business Tax ID Details` maybe) to match the wording in the other text?

client/me/purchases/vat-info/index.tsx
```diff
@@ -44,11 +46,11 @@ export default function VatInfoPage() {
 			<Column type="sidebar">
 				<Card className="vat-info__sidebar-card">
 					<CardHeading tagName="h1" size={ 16 } isBold={ true } className="vat-info__sidebar-title">
-						{ translate( 'VAT Information' ) }
+						{ translate( 'Business Tax ID Information' ) }
```
sirbrillig: To save on translations, I think (?) this could probably be `Business Tax ID details` also (or whatever we end up using for the title string).

client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx
```diff
@@ -249,7 +249,7 @@ export function VatForm( {
 				<Field
 					id={ section + '-vat-address' }
 					type="text"
-					label={ String( translate( 'Address for VAT' ) ) }
+					label={ String( translate( 'Address for Business Tax ID' ) ) }
```
sirbrillig: As you suggested, I think this could be shortened to `Address for tax ID`.

client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx
```diff
@@ -221,7 +221,7 @@ export function VatForm( {
 				<Field
 					id={ section + '-vat-organization' }
 					type="text"
-					label={ String( translate( 'Organization for VAT' ) ) }
+					label={ String( translate( 'Organization for Business Tax ID' ) ) }
```
sirbrillig: And this could probably be `Organization for tax ID`.

client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx
```diff
@@ -204,15 +204,15 @@ export function VatForm( {
...
+					label={ translate( 'Add Business Tax ID details' ) }
 					disabled={ isDisabled || Boolean( vatDetailsFromServer.id ) }
 				/>
 				{ countryCode === 'GB' && (
 					<CheckboxControl
 						className="vat-form__expand-button"
 						checked={ vatDetailsInForm.country === 'XI' }
 						onChange={ toggleNorthernIreland }
-						label={ translate( 'Is the VAT for Northern Ireland?' ) }
+						label={ translate( 'Is the Business Tax ID for Northern Ireland?' ) }
```
sirbrillig: And this could be `Is the tax ID for Northern Ireland?`

sirbrillig (commented):
Overall I think this is great! I have some wording thoughts.

> Also, not exactly sure how to handle the VAT text in checkout line items

I think those labels come directly from Taxamo so we don't have any control over them (but they should already be named based on the actual tax name).

> In some places the usage of 'Business Tax ID' is a bit repetitive, could we perhaps change subsequent mentions to 'tax id' like so

🤔  I'm ok with either. I think you're right that we could use just "tax ID" for a few things. I'll leave inline suggestions.

sirbrillig:
Also you'll need to update the unit tests that are looking for text like "Add VAT details", etc. That's why they are failing.

jjchrisdiehl:
I think as long as the 'heading' is clearly "Business Tax ID" and that the subsequent fields are children of this heading, we could perhaps use the truncated 'tax ID'. Thoughts?

client/me/purchases/titles.js
jjchrisdiehl: I think lowercase makes more sense, since 'details' isn't really part of a title. I also don't know why I wrote 'purchase id' instead of tax id, good catch.

sirbrillig (approved):
Looks good to me! Nice work!

client/me/purchases/vat-info/index.tsx
```diff
@@ -236,7 +238,9 @@ function useDisplayVatNotices( {
 			reduxDispatch( removeNotice( 'vat_info_notice' ) );
 			reduxDispatch(
 				errorNotice(
-					translate( 'Your VAT details are not valid. Please check each field and try again.' ),
```
sirbrillig, line 239: I just noticed that there's another instance of this string in `client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx` that will need to be changed also!

client/me/purchases/vat-info/index.tsx
```diff
@@ -44,11 +46,11 @@ export default function VatInfoPage() {
 			<Column type="sidebar">
 				<Card className="vat-info__sidebar-card">
 					<CardHeading tagName="h1" size={ 16 } isBold={ true } className="vat-info__sidebar-title">
-						{ translate( 'VAT Information' ) }
+						{ translate( 'Business Tax ID details' ) }
 					</CardHeading>
 					<p className="vat-info__sidebar-paragraph">
 						{ translate(
-							"We currently only provide VAT invoices to users who are properly listed in the VIES (VAT Information Exchange System) or the UK VAT databases. VAT information saved on this page will be applied to all of your account's receipts."
+							"We currently only provide Business Tax ID invoices to users who are properly listed in Business Tax ID databases. Business Tax ID information saved on this page will be applied to all of your account's receipts."
```
jjchrisdiehl: @sirbrillig Thoughts on correcting this wording? The second and third usage of Business Tax ID could perhaps be simply tax ID

sirbrillig: 🤔 I don't know. I think it would be fine either way. Whatever you think is best.

sirbrillig (approved):

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7830896

Thank you @jjchrisdiehl for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7830896).

PR #73690, opened 2023-02-22
Plugin Bundle Flow: Redirect back to the activated theme page

### LINKED ISSUE #73661: WooExpress Trial: Update plugin-bundle flow redirect after theme activation
When we activate a bundled theme from a WooExpress trial site, we redirect back to `/home` and then the WooCommerce launch checklist.

Instead, we should redirect back to the theme page for the theme we just activated.

See #73595 for additional context.

markbiek:
Fixes #73661

Redirect to the theme page from the plugin-bundle flow

When exiting the plugin-bundle flow, we now redirect back to the theme page for the theme that was just activated. Previously we redirected to `/home`.

### WooExpress Trial

* Apply this branch.
* Visit http://calypso.localhost:3000/setup/wooexpress/ to create a new WooExpress trial site.
* After the trial site has been created, visit `http://calypso.localhost:3000/themes/:siteSlug?s=tazza`.
* Activate the Tazza theme.
* You should briefly enter the plugin-bundle flow but then exit without having to enter any information.
* After exiting the plugin-bundle flow, you should be redirected to the Tazza page.

### Business Site

* Apply this branch.
* Create a new site with a Business plan or use an existing one.
* Visit `http://calypso.localhost:3000/themes/:siteSlug?s=tazza`.
* Activate the Tazza theme.
* You should go through the plugin-bundle flow.
* After exiting the plugin-bundle flow, you should be redirected to the Tazza page.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
+++ b/client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
@@ -78,6 +78,7 @@ const pluginBundleFlow: Flow = {
 		const { setPendingAction, setStepProgress, resetOnboardStoreWithSkipFlags } =
 			useDispatch( ONBOARD_STORE );
 		const { setIntentOnSite, setGoalsOnSite, setThemeOnSite } = useDispatch( SITE_STORE );
+		const siteDetails = useSelect( ( select ) => site && select( SITE_STORE ).getSite( site.ID ) );
 		const dispatch = reduxDispatch();
 
 		// Since we're mimicking a subset of the site-setup-flow, we're safe to use the siteSetupProgress.
@@ -129,11 +130,16 @@ const pluginBundleFlow: Flow = {
 		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
 			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
 
+			const defaultExitDest = siteDetails?.options
+				? `/theme/${ siteDetails?.options.theme_slug }/${ siteSlug }`
+				: `/home/${ siteSlug }`;
+
 			switch ( currentStep ) {
 				case 'checkForWoo':
 					// If WooCommerce is already installed, we should exit the flow.
 					if ( providedDependencies?.hasWooCommerce ) {
-						return exitFlow( `/home/${ siteSlug }` );
+						// If we have the theme for the site, redirect to the theme page. Otherwise redirect to /home.
+						return exitFlow( defaultExitDest );
 					}
 
 					// Otherwise, we should continue to the next step.
@@ -188,7 +194,7 @@ const pluginBundleFlow: Flow = {
 						return exitFlow( `${ adminUrl }admin.php?page=wc-admin` );
 					}
 
-					return exitFlow( `/home/${ siteSlug }` );
+					return exitFlow( defaultExitDest );
 				}
 
 				case 'wooTransfer':
--- a/client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
+++ b/client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
@@ -130,7 +130,7 @@ const pluginBundleFlow: Flow = {
 		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
 			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
 
-			const defaultExitDest = siteDetails?.options
+			const defaultExitDest = siteDetails?.options?.theme_slug
 				? `/theme/${ siteDetails?.options.theme_slug }/${ siteSlug }`
 				: `/home/${ siteSlug }`;
 


daledupreez (commented):
This implementation feels like it's a bit too heavy for me, as well as having some unexpected behaviour because it is dependent on the _current_ site theme. More specifically, there are two primary contexts where these themes can be activated:
1. From the theme showcase, where no theme is selected -- this is done via the action menu for the theme
2. From the details page for a theme

As implemented, we will always send the user to `/theme/:themeId/:siteSlug` for the current theme, which may or may not be the theme they were trying to activate, depending on when we perform the redirect relative to when we get updates from the server.

Might it be simpler to have a single URL parameter that we stash in the data store from the incoming URL, and then redirect back accordingly? We could do that via a generic `originalPath` URL parameter that can redirect anywhere in the application via a `/`-prefixed value. We could also specify two parameters, e.g. `originalContext: 'theme-showcase' | 'theme-details'` and `themeId` where the second parameter is optional. This is fully expressive, but is less flexible in the long term.

markbiek:
@daledupreez That makes sense. I went back and forth trying to decide if a url parameter might be a better way to go. I agree that a url parameter is more lightweight.

> As implemented, we will always send the user to /theme/:themeId/:siteSlug for the current theme, which may or may not be the theme they were trying to activate, depending on when we perform the redirect relative to when we get updates from the server.

For my own knowledge, can you elaborate a little on this? I never ran into a situation where the theme being activated _wasn't_ the current theme but maybe that's a timing thing?

daledupreez:
> > As implemented, we will always send the user to /theme/:themeId/:siteSlug for the current theme, which may or may not be the theme they were trying to activate, depending on when we perform the redirect relative to when we get updates from the server.
> For my own knowledge, can you elaborate a little on this? I never ran into a situation where the theme being activated wasn't the current theme but maybe that's a timing thing?

I don't think I fully thought through the sequencing here, so the issue may be less pronounced in terms of which theme is involved, but I was thinking of this as follows:
1. Before we do anything we are in Calypso and we're using some other theme.
2. The user then requests activation of that theme - I think we currently wait for the theme to be fully activated before we trigger the bundle-plugin flow, but I am not 100% sure of that. If we're not waiting for confirmation of the theme update, then the old theme may still be active.
3. We then get to the bundle plugin flow, at which time we call `getSiteTheme()`, which will only return the new theme after we have made a request for current active theme, and if the theme switch is complete. I didn't check to make sure we're triggering a new fetch of the theme data, but there are some possible timing gaps where we'd redirect to somewhere unexpected (either `/home/:siteSlug` if we don't have a theme, or possibly `/theme/:oldTheme/:siteSlug`, though the second is less likely).

markbiek:
> We then get to the bundle plugin flow, at which time we call getSiteTheme(), which will only return the new theme after we have made a request for current active theme, and if the theme switch is complete. I didn't check to make sure we're triggering a new fetch of the theme data, but there are some possible timing gaps where we'd redirect to somewhere unexpected (either /home/:siteSlug if we don't have a theme, or possibly /theme/:oldTheme/:siteSlug, though the second is less likely).

This is a good point that I hadn't thought of.

At the moment, everything is fine because (currently) we're only calling `getSiteTheme` in the plugin-bundle flow in one spot in order to do the redirect. But it would definitely break if `getSiteTheme` got called at any other point.

I'll switch to using a query string parameter and put those data-store updates in a separate PR.

markbiek:
> I'll switch to using a query string parameter and put those data-store updates in a separate PR.

@daledupreez The query string thing is turning out to bit a little tricky. I did some digging though and I'm convinced that the correct `theme_slug` is set in the `site.options` of the `SITE_STORE`.

I'm using that value for the redirect instead of either the query string or making a new call to `/themes/mine`.

Let me know if you have any objections or concerns I'm not thinking of and we can discuss further.

client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
```diff
@@ -129,11 +130,16 @@ const pluginBundleFlow: Flow = {
 		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
 			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
 
+			const defaultExitDest = siteDetails?.options
+				? `/theme/${ siteDetails?.options.theme_slug }/${ siteSlug }`
```
daledupreez, line 134: Is `siteDetails?.options.theme_slug` guaranteed to be defined? The `?.` doesn't inspire confidence!

daledupreez (commented):
> The query string thing is turning out to bit a little tricky.

Which part? Handling it through the flow? Or finding the correct URL to redirect to? If it was the latter, it might be worth relying on the `getCurrentRoute()` selector to identify where you are in the application.

markbiek:
> Which part? Handling it through the flow? Or finding the correct URL to redirect to? If it was the latter, it might be worth relying on the getCurrentRoute() selector to identify where you are in the application.

At the moment, the only place we have any information about the pertinent theme is in the redux store (in the places we've been talking about).

I guess, if we went the query param route, we could pass the theme slug when we redirect [from the ThanksModal](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/themes/thanks-modal.jsx#L92) and then carry that value through the plugin-bundle flow.

> Is siteDetails?.options.theme_slug guaranteed to be defined? The ?. doesn't inspire confidence!

While there are some Stepper contexts where `siteDetails?.options` might not be defined, I'm still convinced that it _is_ guaranteed to be defined by the time we hit the plugin-bundle flow. Unless I'm mistaken, we seem to be waiting for the theme setup API call to finish before entering the plugin-bundle flow.

client/landing/stepper/declarative-flow/plugin-bundle-flow.ts
```diff
@@ -129,11 +130,16 @@ const pluginBundleFlow: Flow = {
 		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
 			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
 
+			const defaultExitDest = siteDetails?.options
```
daledupreez: It may be worth checking for the full property path to make sure we have a theme slug.
```suggestion
			const defaultExitDest = siteDetails?.options?.theme_slug
```

daledupreez (approved):
Apologies for the delayed review. This LGTM!

I spent a while getting stuck on what I think is a special case where the theme is already installed on the site -- that ends up redirecting to `/home/` for some reason.

But I think we should ship this change and iterate.

PR #73701, opened 2023-02-23
Fix issue causing reader to be constantly re-rendered.

roo2:
I noticed this issue when putting a console.log() in the render method of [the reader-main component ](https://github.com/Automattic/wp-calypso/blob/trunk/client/reader/components/reader-main/index.jsx#L45) and found it was being continuously rendered.

This is wasting a lot of resources and is likely to cause performance issues, though I haven't noticed symptoms before myself.

The issue is caused by the "with-dimensions" component, which is calling setState with each componentDidUpdate, which in turn triggers another componentDidUpdate, and another setState, essentially in an infinite loop.

The fix simply performs a check on whether the dimensions have actually changed since the last update.

#### Testing instructions
You can add a console.log to [the reader-main component ](https://github.com/Automattic/wp-calypso/blob/trunk/client/reader/components/reader-main/index.jsx#L45) as I did. Or you can use the chrome performance tool.
Wait until the page has finished loading and record a ~3 second snapshot. You should see that there is constant activity before and basically no activity after this patch is applied.
Also test that the with-dimensions component still works as it should, infinite scroll still works and resizing the window still works as before.

##### Before
Screen Shot 2023-02-23 at 2 18 42 pm

##### After 
Screen Shot 2023-02-23 at 2 19 42 pm

### DIFF
--- a/client/lib/with-dimensions/index.jsx
+++ b/client/lib/with-dimensions/index.jsx
@@ -39,7 +39,7 @@ export default ( EnhancedComponent ) =>
 			height: 0,
 		};
 
-		handleResize = afterLayoutFlush( () => {
+		handleResize = afterLayoutFlush( ( prevState ) => {
 			const domElement = this.props.domTarget || this.setRef || this.divRef;
 
 			if ( domElement ) {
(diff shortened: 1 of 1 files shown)


eoigal (approved):
WFM

tyxla (approved):
The code looks good and the changes test well. Thanks! 🚀

PR #73702, opened 2023-02-23
Make reader tag pages public

roo2:
See p2 pe7F0s-A6-p2 for requirements and p1676953886830299-slack-C03NLNTPZ2T for some more discussion.

This PR makes the `/tag/$tag_name` route public and makes the following changes if a user is logged out:

* Updates header bar to use the same header as /themes
* Removes left Nav
* "Follow tag" link redirects to `/start/account`
* "Like", "Comment" and "Follow" post actions buttons redirect to `/start/account`
* Removes post toggle menu
* Clicking on a feed now redirects you to `/start/account` instead of the login page.
* Clicking on a post now redirects you to `/start/account` instead of the login page.
* No longer makes a failing call to "sync reader follows"

Screen Shot 2023-02-23 at 3 09 26 pm

#### Testing instructions
Enable the `reader/public-tag-pages` feature flag via the URL e.g.: https://container-sharp-margulis.calypso.live/tag/fishing?flags=reader/public-tag-pages
Test the tag page for a logged out user
Test for a logged in user
Make sure (logged out)`/themes` page header still works as some shared CSS was changed.

### DIFF
--- a/client/blocks/follow-button/index.jsx
+++ b/client/blocks/follow-button/index.jsx
@@ -1,7 +1,11 @@
+import { getUrlParts } from '@automattic/calypso-url';
 import { omitBy } from 'lodash';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
 import { connect } from 'react-redux';
+import { navigate } from 'calypso/lib/navigate';
+import { createAccountUrl } from 'calypso/lib/paths';
+import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 import { follow, unfollow } from 'calypso/state/reader/follows/actions';
 import { isFollowing } from 'calypso/state/reader/follows/selectors';
 import FollowButton from './button';
@@ -26,6 +30,10 @@ class FollowButtonContainer extends Component {
 	};
 
 	handleFollowToggle = ( following ) => {
+		if ( ! this.props.isLoggedIn ) {
+			const { pathname } = getUrlParts( window.location.href );
+			return navigate( createAccountUrl( { redirectTo: pathname } ) );
+		}
 		if ( following ) {
 			const followData = omitBy(
 				{
@@ -63,6 +71,7 @@ class FollowButtonContainer extends Component {
 export default connect(
 	( state, ownProps ) => ( {
 		following: isFollowing( state, { feedUrl: ownProps.siteUrl } ),
+		isLoggedIn: isUserLoggedIn( state ),
 	} ),
 	{
 		follow,
--- a/client/blocks/like-button/button.jsx
+++ b/client/blocks/like-button/button.jsx
@@ -1,10 +1,14 @@
+import { getUrlParts } from '@automattic/calypso-url';
 import classNames from 'classnames';
 import { localize } from 'i18n-calypso';
 import { omitBy } from 'lodash';
 import PropTypes from 'prop-types';
 import { createElement, PureComponent } from 'react';
+import { connect } from 'react-redux';
+import { navigate } from 'calypso/lib/navigate';
+import { createAccountUrl } from 'calypso/lib/paths';
+import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 import LikeIcons from './icons';
-
 import './style.scss';
 
 class LikeButton extends PureComponent {
@@ -42,6 +46,10 @@ class LikeButton extends PureComponent {
 	}
 
 	toggleLiked( event ) {
+		if ( ! this.props.isLoggedIn ) {
+			const { pathname } = getUrlParts( window.location.href );
+			return navigate( createAccountUrl( { redirectTo: pathname } ) );
+		}
 		if ( event ) {
 			event.preventDefault();
 		}
@@ -102,4 +110,6 @@ class LikeButton extends PureComponent {
 	}
 }
 
-export default localize( LikeButton );
+export default connect( ( state ) => ( {
+	isLoggedIn: isUserLoggedIn( state ),
+} ) )( localize( LikeButton ) );
--- a/client/blocks/reader-post-options-menu/reader-post-ellipsis-menu.jsx
+++ b/client/blocks/reader-post-options-menu/reader-post-ellipsis-menu.jsx
@@ -16,6 +16,7 @@ import { READER_POST_OPTIONS_MENU } from 'calypso/reader/follow-sources';
 import { canBeMarkedAsSeen, isEligibleForUnseen } from 'calypso/reader/get-helpers';
 import { isAutomatticTeamMember } from 'calypso/reader/lib/teams';
 import * as stats from 'calypso/reader/stats';
+import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 import * as PostUtils from 'calypso/state/posts/utils';
 import { recordReaderTracksEvent } from 'calypso/state/reader/analytics/actions';
 import { hasReaderFollowOrganization } from 'calypso/state/reader/follows/selectors';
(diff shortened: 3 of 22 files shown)


eoigal:
Works as advertised however I was expecting the links to the post to go to the creators site.

Screenshot 2023-02-23 at 09 56 05

When I click a link I get redirected to the `/start/account` page.

davemart-in:
+1 to what @eoigal shared. I think we need to figure out a way to redirect these title links to the actual website vs. blocking them behind authentication. The following two links should take you to the same place:

CleanShot 2023-02-27 at 13 16 08

I'm testing the rest of this PR now...

davemart-in:
@roo2 did the content on https://wordpress.com/tags/ disappear with this PR? I'm no longer seeing the tag cloud there when viewing https://container-loving-edison.calypso.live/tags in an incognito window.

davemart-in:
@roo2 when I click the like button, I'm taken to the login screen as expected and when I click the back button I'm taken directly back to the tags page. 

That said, if I click on one of the usernames for someone who has liked a post, and try to click back after seeing the login screen, I'm unable to gat back to the tag page I was on previously:

https://user-images.githubusercontent.com/5634774/221649730-c17ce0ea-de83-4b04-a119-f82323f84981.mp4

If you look closely, it looks like there is a redirect that you hit prior to getting to the login page which is likely the culprit.

creativecoder:
@roo2 Do you see a way to put this behind a feature flag? It's clear from the feedback so far that there are a number of flows in the tag pages we need to be mindful of, but I'm worried a larger PR that handles all of these will be unwieldy. 

Seems like it would be better to merge this behind a flag, iterate, and then remove the flag when we're happy with all flows.

roo2:
Thanks for the feedback so far.

> I think we need to figure out a way to redirect these title links to the actual website vs. blocking them behind authentication. 

Ok, I'll work on this 👍 

> @roo2 did the content on https://wordpress.com/tags/ disappear with this PR? I'm no longer seeing the tag cloud there when viewing https://container-loving-edison.calypso.live/tags in an incognito window.

no, it's because the `/tags` is set up by the routing on wordpress.com, so it's not even part of calypso and won't be on the calypso.live containers.

> If I click on one of the usernames for someone who has liked a post, and try to click back after seeing the login screen, I'm unable to gat back to the tag page I was on previously

Thanks, I'll look into this 👀 

> Seems like it would be better to merge this behind a flag, iterate, and then remove the flag when we're happy with all flows.

hmm yes, I can do this 👍

roo2:
@davemart-in , @creativecoder I've updated to address the feedback and it's ready for another review.

It's now behind a feature toggle `reader/public-tag-pages` which is enabled everywhere but staging and production
I found that most of the code doesn't have to know about the feature toggle, because it only applies if the user is logged out.
I also found that since I can't access the feature toggles from `sections.js`, I had to leave it public there, and [add a middleware to the tag pages here](https://github.com/Automattic/wp-calypso/pull/73702/files#diff-5a4a686e51fe13d95f5fead304cff0efa25d6fe3fcf20c4ae51e3dd3403b2c1bR15).

client/blocks/follow-button/index.jsx
```diff
@@ -26,6 +30,10 @@ class FollowButtonContainer extends Component {
 	};
 
 	handleFollowToggle = ( following ) => {
+		if ( ! this.props.isLoggedIn ) {
+			const { pathname } = getUrlParts( window.location.href );
+			return navigate( createAccountUrl( { redirectTo: pathname } ) );
+		}
```
tyxla, line 36: I wanted to point out that this isn't ideal. This component likely needs little to no data when the user is logged out, as compared to when the user is logged in.

I'd recommend exploring separate components for logged-in users and logged-out users, where the logged-out one will end up being way simpler and smaller. This helps for better separation of concerns, on top of the performance and bundle size aspects.

That applies to the rest of the areas where we're suggesting support for logged-out users.

davemart-in:
Note for testing - I had to add the feature flag to the URL to get it to work: https://container-sharp-margulis.calypso.live/tag/fishing?flags=reader/public-tag-pages Just noting this for anyone else where this might not be obvious.

> I think we need to figure out a way to redirect these title links to the actual website vs. blocking them behind authentication.

Works as expected. Nicely done!

> no, it's because the /tags is set up by the routing on wordpress.com, so it's not even part of calypso and won't be on the calypso.live containers.

Ah! Cool. Makes sense.

> If I click on one of the usernames for someone who has liked a post, and try to click back after seeing the login screen, I'm unable to gat back to the tag page I was on previously

This looks fixed. 👍

davemart-in (approved):
Visually and functionally, everything looks good to me. Nice work!

roo2:
> Note for testing - I had to add the feature flag to the URL to get it to work: https://container-sharp-margulis.calypso.live/tag/fishing?flags=reader/public-tag-pages Just noting this for anyone else where this might not be obvious.

Thanks @davemart-in, I forgot that I'd set it to disabled by default in all of the calypso.live environments, I've changed it to be enabled everywhere but production and horizon

client/blocks/follow-button/index.jsx
roo2: hmm I understand the point that in the logged out state, we're loading a lot of code that can never be hit which makes the bundle bigger than it needs to be. 

When you talk about data, do you mean data being pulled in from `state` which is already loaded by the app, or new data by requested from the server? I think the problem would be if we're loading unnecessary data from the server, whereas passing existing data within the app doesn't seem like a problem.

I tried to make sure that no unnecessary requests are made (they would fail anyway if they require auth).

In investigating I actually just noticed that there are still calls being made to `/me/settings` and `/me/preferences`, so I'll track those down too 👀

If we want to reduce the bundle size with separate components, it would require a fair bit of duplication of code. So I think it's better not to unless we know that there are meaningful performance problems that we can improve by reducing the bundle size.

roo2: > In investigating I actually just noticed that there are still calls being made to /me/settings and /me/preferences, so I'll track those down too 👀

NVM it looks like those requests were coming from the environment badge injected on localhost

tyxla: I'm basically talking about anything that the component would require, but won't use for a logged-out user. That could include:

* Any component that we require, but don't render
* Any action creator that we include, but will not be using
* Any styles that we load, but we won't use
* Any component business logic that we don't use.

And it's not only the performance / bundle size aspect; it's also the developer experience aspect - why do I load all these things when I won't be using them? What will happen if we add a few more conditions like that? How will someone new to the codebase understand what's going on? What's the chance that these not-so-good practices will be spread throughout the codebase as someone is using that example as a base for their new work?

creativecoder: You can see an example of this problem when we [pulled the blogging prompts card into reader](https://github.com/Automattic/wp-calypso/pull/72923#issuecomment-1414148141). The reader bundle size increased by over 4%, just from adding a fairly simple card component and a little bit of code to determine where to render it.

The card component itself doesn't seem that large, so I'm guessing it's because of all the associated state code required by that component and its dependencies.

Given a primary goal of making these tags pages is SEO related, and bundle size/load speed is an important part of page ranking these days, I suspect it's very much worth investigating how we can reduce the bundle size for logged out pages.

Maybe we can continue with this PR behind a feature flag, and then go through the components one at a time to see where we can reduce the size?

roo2: ok, I agree, I think it makes sense to start with this as is, and then look at reducing the bundle size in follow up PRs.

PR #34949, opened 2019-07-29
Signup: don't resume from last step in flow after refresh

p-jackson:
##### The "I want to start over" story
start-again-master-smol
[Original video](https://cld.wthms.co/jmfDDO)

##### The "close laptop lid" story
back-refresh-master-smol
[Original video](https://cld.wthms.co/YEvHXd)

#### Changes proposed in this Pull Request

Removes code that redirected to the "last in-progress step" in the flow after `<Signup>` was first mounted. This means the location after refresh is determined by the url, as one would expect.

If any redirection happens, it's to the beginning of the flow e.g. a redirect to the beginning will happen if:
1. there's no data in `localStorage`. We need to go back to collect all those dependencies again.
2. you manually enter the url for a step that's _past_ the "last in-progress step".

One could argue that in the case of (2) that it should go back to the "last in-progress step". But I can't think of a user story for this. Whereas "staying on the same page after refresh" covers spoty interent, closing-reopening laptop lid, browser crash and restore, automattic OS updates.

Discussed in a number of places:
- Automattic/zelda-private#109
- #34299
- At team meetup

Fixes Automattic/zelda-private#109

#### Testing instructions

The expected behaviour for each of the test steps depends on whether the state was persisted or not. Just a reminder:
- With persisted state: `ENABLE_FEATURES=no-force-sympathy npm start`
- No persisted state: `ENABLE_FEATURES=force-sympathy npm start`

##### Start again
- `/start`, proceed through flow but don't complete
- Go back to `/start` using whatever method you like except clicking "back"
- You should have started the onboarding flow again (regardless of the local storage state)

##### Close laptop lid
- `/start`, proceed through flow but don't complete
- Click back
- Refresh (to simulate coming back later)
- You should have either:
  - stayed on the step you were just on (with local storage)
  - gone back to site type step, but still logged in (without local storage)

##### ~Hot linking past incomplete steps~
- ~`/start`, choose blog~
- ~You should be on the site topic step~
- ~Manually navigate to `/start/site-style-with-preview`~
- ~You should have been redirected to the start the onboarding flow again (regardless of the local storage state)~
_Edit: this wasn't straight forward to implement_

Manually test these changes trying to break it by:
- trying it both logged in and out
- having two tabs open that share the same local storage
- switching flows
- clearing or keeping localhost

PR #73711, opened 2023-02-23
Remove persistence for signup progress state

aneeshd16:
Related to p1677084219872039-slack-C040UAN6TPS

## Proposed Changes

* Removes persistence for signup progress state.

The original bug can be reproduced by the following steps:

- Clear out indexeddb for https://wpcalypso.wordpress.com
- Go to https://wpcalypso.wordpress.com/start/
- Search for any domain name and proceed to the plans page.
- Don’t select a plan. In the same tab, go to https://wpcalypso.wordpress.com/start/free
- You should now see the infinite loading screen.

## Testing Instructions

The assumption is that the stored progress state acts like a cache, and no signup flow relies on this storage. Testing would involve:
* Testing a few signup flows like `/start`, `/start/launch-site/`, `/start/free`, `/start/domain`, etc. More can be found here: 2f4ae-pb/#plain
* E2E tests.

## Pre-merge Checklist


chriskmnds (commented):
Thanks @aneeshd16 . This works as outlined on /start flow. Going through the steps in the description, then `/start/free` in the end will redirect me to `/start/user` (and not get stuck at the loading screen).

I'm not entirely following what is going on with the schema validation and persistence layer, but curious, was there no use for these progress steps being persisted? No redirection or anything else based on them (that we ought to investigate further)? 🤔

aneeshd16:
Thanks for the review @chriskmnds!

> was there no use for these progress steps being persisted? No redirection or anything else based on them

AFAICT, the redirection logic was removed long ago via https://github.com/Automattic/wp-calypso/pull/34949. I could not find any other usages, but I could be wrong too.

Hi @p-jackson, I was wondering if you had any additional context on whether the signup `progress` state was used for anything other than the redirection that was removed? It's been a while since that PR, so any help is appreciated! 😅

southp (dismissed):
I've audited if there is any potential flow that would rely on the persisted progress. For the flows on the classic signup framework, there isn't by default. For the flows on the Stepper framework, [/setup/newsletter seems concerning](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/newsletter.ts#L71), but ends up working well.

Together with what @chriskmnds confirmed above, I don't have any remaining concern for this. However, let's postpone the deployment of this on Monday in case we miss any nuances :)

p-jackson:
@aneeshd16 I think this is a fairly fundamental change to how the classic signup flows work. The signup framework intentionally keeps all the step state in between page loads. So if you are on the `domain` step, search for a domain, click refresh, the search term is persisted. But after this change it's not.

https://user-images.githubusercontent.com/1500769/223005106-732408ab-c3d3-484e-8686-72c4f050d3ad.mp4

This was more prominent back in the olden days. We had lots of other steps and we didn't think the user would want to lose all of their work if the closed their laptop lid/hit refresh/accidentally closed and reopened their tab with cmd+shift+t.

It's always been buggy though. It's hard to tell the difference between refreshing the page and intentionally bailing out of the flow and clicking "add new site" in Calypso at a later date. So I can see why you're proposing we remove it.

So long story short, this data store _is_ being used, but it's still ok to remove it if that's the product change you and the team want to make.

aneeshd16:
Thanks @p-jackson!

The domain search results are currently a part of the signup progress state.
image

The `domainForm` key is not in the type definition: https://github.com/Automattic/wp-calypso/blob/d8c953924d35b172b993e9c419c12f76bfae5076/client/state/signup/progress/schema.ts#L26

Maybe the `steps` store is a better fit to cache these results? But as you mentioned, this is old code, so it's tough to reason why something is the way it is right now. 🤔 

However, in my testing, the search query persisted after selecting a domain and refreshing on the plans step:

https://user-images.githubusercontent.com/5436027/223068672-f1e3cec8-5234-4d2e-abb2-7a63d517efad.mp4

p-jackson:
> the search query persisted after selecting a domain and refreshing on the plans step

Yes the query part persisted, but not your selection. You should be able to choose a domain, advance to the plan step, refresh, and you should still be on the plan step and see the text "The domain xyz.org is free for the first year" because it's remembered your selection.

I'm not sure about moving the domain data to the `steps` reducer. There's a lot of code that expects data to be in the `providedDependencies` object, so I think it'd be a hefty refactor.

southp:
Thanks for raising this important concern for us @p-jackson . What it shows is that it's a fundamental change that can affect things that's , e.g. I had audited the flows, but still didn't catch the issue you've found. 

It's hard to really pinpoint the impact of this issue, but I've created two funnels to see if we could get a rough feeling: 2f665-pb. One from seeing the processing screen to signup completion, and one from seeing the processing screen to the new site creation; the both funnels are set as 10 minutes. To my surprise, the former is only 87.9% and the latter is 71.9%. We can't say that the issue we are solving here is a major contributor for the drop, but it definitely worths a shot.

@aneeshd16 The problem seems to be that the persisted progress pollutes to the other flows when it shouldn't. Is there other safer way to fix this?

PR #73540, opened 2023-02-20
Fixing focus link/button states for domains step in onboarding flow

davemart-in:
## Description

The domain step within the onboarding flow is not great for keyboard navigation. Some focus styles are just completely missing.

### Before

domains-before

This example is a little hard to tell, but I'm tabbing around and the focus disappears completely from the page.

### After

https://user-images.githubusercontent.com/5634774/220118493-1088c09f-4d10-4c6b-b5cb-28a2f73c0813.mp4

## How to reproduce

- Go to create a new site
- On /start/domains tab around the page

PR #73712, opened 2023-02-23
Updating focus styles on domain step of onboarding flow

davemart-in:
## Description

The domain step within the onboarding flow is not great for keyboard navigation. Some focus styles are just completely missing.

### Before

domains-before

This example is a little hard to tell, but I'm tabbing around and the focus disappears completely from the page.

### After

https://user-images.githubusercontent.com/5634774/220118493-1088c09f-4d10-4c6b-b5cb-28a2f73c0813.mp4

## How to reproduce

- Go to create a new site
- On /start/domains tab around the page

## Previously deployed

This PR was previously deployed in #73540 and reverted in #73554 when it was discovered that these changes broke unit tests.

### DIFF
--- a/client/components/domains/domain-suggestion/style.scss
+++ b/client/components/domains/domain-suggestion/style.scss
@@ -462,6 +462,11 @@ body.is-section-signup.is-white-signup {
 					padding: 0.65em 2.8em;
 					width: auto;
 				}
+
+				&:focus {
+					border-color: var(--color-primary);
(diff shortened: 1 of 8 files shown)


dpasque (approved):
From the KitKat perspective, the pre-release tests look like they're passing great! 👍 You shouldn't hit any issues post merge!

worldomonation (commented):
Echoing @dpasque here, the changes to E2E framework looks good to me.

worldomonation (approved):

PR #73722, opened 2023-02-23
Implement tracking via GTM for woo trials upsells to ecommerce

gmovr:
Related to #

## Proposed Changes
This is still a WIP, but requesting an early review, to make sure we're on the right path. 

* Pass the SiteDetails object to tracking library, and use this to track users who currently have a Woo trial plan, and are upgrading to a paid ecommerce plan.
* Implement the Google Tag Manager pixel for Woo in certain scenarios, e.g. if a user upgrades from a Woo Express trial plan to a paid ecommerce plan
* Push a purchase event with meta data to the data layer.

This is just for testing for now, to see how to best implement this.

## Testing Instructions

* Need to add automated tests
* Sandbox store, allow ad-tracking flag.
* Upgrade a Woo Express trial site via the CTA in the /plans page
* When completing the payment, observe that the GTM container is added, and that the conversion data is pushed to the data layer for GTM ID `GTM-W64W8Q`

## Pre-merge Checklist


client/lib/analytics/ad-tracking/woo/is-wooexpress-upgrade.ts
```diff
@@ -0,0 +1,41 @@
...
+		return false;
+	}
+
+	// Does the cart contain an upsell we care about? I.e. is it an ecommerce plan?
+	const cartContainsValidUpsell = cart.products.some( ( product ) =>
+		isValidWooExpressUpsell( product )
+	);
+
+	// And does the current site have a Woo Express trial pre-upgrade?
+	return sitePlanSlug === PLAN_ECOMMERCE_TRIAL_MONTHLY && cartContainsValidUpsell;
```
daledupreez: Minor note: it might be simpler to perform the slug checks before looking at the products.
```suggestion
	// We don't want to track if we don't have a plan, or the plan isn't an eCommerce trial
	if ( ! sitePlanSlug ||  sitePlanSlug !== PLAN_ECOMMERCE_TRIAL_MONTHLY ) {
		return false;
	}

	// We only want to track when the cart contains an upsell we care about
	return cart.products.some( ( product ) =>
		isValidWooExpressUpsell( product )
	);
```

daledupreez (commented):
The basic logic that I understand seems reasonable to me, but I am not familiar with how we ought to pull in external tools like Google Tag Manager, so I think I'll need more detailed instructions to provide better feedback.

client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
```diff
@@ -1,3 +1,4 @@
+import { SiteDetails } from '@automattic/data-stores/src';
```
tyxla: Should be fine to import this from `@automattic/data-stores`. You could also use `import type` since this only imports a type.

client/lib/analytics/ad-tracking/woo/is-wooexpress-upgrade.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Checks if the cart contains an upgrade to a valid plan, and whether the site has a woo trial.
+ * If both are true, we'll return true. A valid upgrade assumes that the site currently (pre upgrade)
+ * has a Woo Experss trial, and that the upgrade is to a valid ecommerce plan as defined by the return
+ * value of `isEcommerce`.
+ *
+ * @param cart ResponseCart
+ * @param siteDetails SiteDetails
+ * @returns { boolean }
+ */
+export const isWooExpressUpgrade = ( cart: ResponseCart, siteDetails: SiteDetails ): boolean => {
```
tyxla: We seem to be needing only the product slug from `siteDetails`. I'd suggest passing only that to the function, and not the entire `siteDetails` object.

client/lib/analytics/ad-tracking/record-order.js
```diff
@@ -29,9 +30,10 @@ import './setup';
  *
  * @param {Object} cart - cart as `ResponseCart` object
  * @param {number} orderId - the order id
+ * @param {import('@automattic/data-stores/src').SiteDetails} selectedSiteData - site details
  * @returns {void}
  */
-export async function recordOrder( cart, orderId ) {
+export async function recordOrder( cart, orderId, selectedSiteData ) {
```
tyxla: As mentioned below, seems like we can only pass the product slug, and not the entire site details object.

client/lib/analytics/ad-tracking/woo/load-woo-gtm-container.ts
```diff
@@ -0,0 +1,46 @@
...
+	await loadScript( GOOGLE_GTM_SCRIPT_URL + TRACKING_IDS.wooGoogleTagManagerId );
+};
+
+/**
+ * Initializes the GTM container.
+ *
+ * @returns Promise<void>
+ */
+export const initWooGTMContainer = async (): Promise< void > => {
+	window.dataLayer = window.dataLayer || [];
```
tyxla, line 40: `dataLayer` sounds like an odd name for this, and this could be easily confused with the existing Calypso data layer:

https://github.com/Automattic/wp-calypso/blob/5f4eb761353006b94555609175101ac4f95d7ac1/client/state/data-layer/README.md#L1

tyxla (commented):
Hey @gmovr!

I've left some suggestions on the organization of the code.

I'm happy to take a look again - please let me know if you have anything more specific you'd like from my review.

client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
sirbrillig: I was going to say the same thing: we shouldn't import things from deep inside a package as such things are private implementation details and might change without warning.

client/lib/analytics/ad-tracking/woo/is-wooexpress-upgrade.ts
```diff
@@ -0,0 +1,41 @@
+import { PLAN_ECOMMERCE_TRIAL_MONTHLY, isEcommerce } from '@automattic/calypso-products';
+import { SiteDetails } from '@automattic/data-stores/src';
```
sirbrillig: Same comment here about importing from the package's public exports.

```suggestion
import type { SiteDetails } from '@automattic/data-stores';
```

client/lib/analytics/ad-tracking/record-order.js
```diff
@@ -540,6 +543,51 @@ function recordOrderInWPcomGA4( cart, orderId, wpcomJetpackCartInfo ) {
 	debug( 'recordOrderInWPcomGA4: Record WPcom Purchase in GA4' );
 }
 
+/**
+ * Sends a purchase event to Google Tag Manager for eligible Woo Express upgrades.
+ *
+ * @param {import('@automattic/shopping-cart').ResponseCart} cart
+ * @param {string} orderId
+ * @param {import('@automattic/data-stores/src').SiteDetails} selectedSiteData
```
sirbrillig: Same comment here about importing from a package's public exports rather than reaching inside the package.

```suggestion
 * @param {import('@automattic/data-stores').SiteDetails} selectedSiteData
```

client/lib/analytics/ad-tracking/record-order.js
```diff
@@ -29,9 +30,10 @@ import './setup';
  *
  * @param {Object} cart - cart as `ResponseCart` object
  * @param {number} orderId - the order id
+ * @param {import('@automattic/data-stores/src').SiteDetails} selectedSiteData - site details
```
sirbrillig: Same here.

```suggestion
 * @param {import('@automattic/data-stores').SiteDetails} selectedSiteData - site details
```

client/lib/analytics/ad-tracking/record-order.js
```diff
@@ -540,6 +543,51 @@ function recordOrderInWPcomGA4( cart, orderId, wpcomJetpackCartInfo ) {
...
+					transaction_id: orderId,
+					currency: 'USD',
+					items: cart.products.map( ( { product_id, product_name, cost, volume } ) => ( {
+						id: product_id.toString(),
+						name: product_name.toString(),
+						quantity: parseInt( volume ),
+						price: costToUSD( cost, cart.currency ) ?? 0,
+						billing_term: cart.billing?.interval_unit,
+					} ) ),
+					value: cart.total_cost_integer / 100,
```
sirbrillig: I don't fully understand what these stats are used for but not every currency can use this math safely. In USD, `cart.total_cost_integer / 100` will return dollars because USD has two decimal places, but JPY has no decimal places, which means this would produce a strange result that doesn't have much to do with the actual cost. Is that safe here?

sirbrillig (commented):
I left some minor feedback. Mostly I think this should be fine.

client/lib/analytics/ad-tracking/woo/load-woo-gtm-container.ts
robertsreberski: That's actually what Google bases their implementation on: https://developers.google.com/tag-platform/tag-manager/web/datalayer#installation

They require initialization of `dataLayer` property in `window` property as they access it from their scripts.

client/lib/analytics/ad-tracking/record-order.js
robertsreberski: I think this implementation is based on the existing code that has served us for e-commerce analytics for some time:
https://github.com/Automattic/wp-calypso/blob/add/hotjar-tracking-woo-express-plans-page/client/lib/analytics/utils/split-wpcom-jetpack-cart-info.ts#L23

However, when we track down the usage of this data - it appears we actually never use the raw value - and as you mentioned, probably this value in general is flawed.

The `value` e.g. in purchase event is converted to USD. I think we also need to use it similarly to get a valid output:
`costToUSD( cart.total_cost_integer / 100, cart.currency )`

client/lib/analytics/tracker-buckets.ts
```diff
@@ -52,6 +53,7 @@ export const AdTrackersBuckets: { [ key in AdTracker ]: Bucket | null } = {
 	bing: Bucket.ADVERTISING,
 	floodlight: Bucket.ADVERTISING,
 	googleAds: Bucket.ADVERTISING,
+	googleTagManager: Bucket.ADVERTISING,
```
robertsreberski, line 56: Just to make sure, that will require standalone consent given on checkout pages to load the script. Is it the way we want to proceed?

Will we still try cross-domain consent at a later stage?

client/lib/analytics/ad-tracking/record-order.js
sirbrillig: Cool. Just wanted to mention it. Probably `costToUSD( cost, cart.currency )` or something similar, but yes.

client/lib/analytics/ad-tracking/woo/load-woo-gtm-container.ts
tyxla: Makes sense. We could add some comment, indicating something like "not to be confused with the Calypso Data Layer".

robertsreberski: Right, that would help, indeed!

client/lib/analytics/tracker-buckets.ts
robertsreberski: I've asked the question in the discussion thread, and indeed we do not proceed with cross-domain consent at this time. It might involve more legal and technical adjustments in the near future.

Reference:
pau2Xa-4J9-p2#comment-13342

robertsreberski:
I've resolved all comments regarding the PR and tried to improve the code from the analytics side further.

All seems to work well now, except `sitePlanSlug` value. Even when my site has a trial plan, `sitePlanSlug = PLAN_ECOMMERCE_TRIAL_MONTHLY`. Because of that - the condition is not met and the GTM is not loaded as a result. Was it working for you before @gmovr ? I wonder if there is anything I'm doing wrong here 🤔

gmovr:
@robertsreberski  -- Thanks for all the updates! 

When you purchase the upgrade, do you use the sandbox? If not, the purchase wouldn't count for tracking due to the amount < $0.01 check. 

```js
	if ( cart.total_cost >= 0.01 ) {
``` 
in `record-purchase.js`


So to actually test it, if you sandbox (you'd also need to create the woo site sandboxed), you should have the right plan, and upgrading will enter the conditional.

To test a bit more realistic, I also tested to just remove the conditional, and adding a few `debugger;`s 

Screenshot 2023-03-16 at 17 13 33

And it seems the end result is that the GTM is appended to the DOM (and I can see it in the network tab with the correct ID of `GTM-W64W8Q`). 

I guess we should also double check that the conversion data coming through is of the expected format, and the GTM DataLayer has them available.

robertsreberski:
When following @gmovr suggestions, the PR tests well. We think it's ready to go, so I'll deploy it shortly.

Issue #69687, opened 2022-11-02
Deprecate Theme Switch modal because: "Preserve my homepage content" modal does not load preview for certain themes 


fushar:
# Description
[Updated 20 Jan 2023]
Over the past months we have had many discussions on how the visuals in  theme switch modal is "mostly" broken. It some very specific cases the live preview (the left side visual) works properly and applies the styles of the new theme to the user's existing content.  For the cases where this works properly we want to show the visuals.  Otherwise we want to hide the visuals and just show the copy in the modal.

See discussion in these P2s:
*  pekYwv-uV-p2
* [Auto loading homepage on theme switch issues] pbAok1-1Yh-p2

### Steps to reproduce

1. Create a new site (wordpress.com/start) with a theme whose homepage template is `home.html`
   - e.g.: Pendant / Cultivate / Archeo / Yuga
2. On My Home, open Appearances -> Themes.
3. Select a theme whose homepage template is not `home.html`
   - e.g.: Appleton / Dorna / Russell
4. Click "Activate this design"

### Expected

The previews for both column load

### What happened instead

The preview for the "Switch theme, preserving my homepage content." column does not load:

<img width="445" alt="image" src="https://user-images.githubusercontent.com/1525580/199427872-74b81fb4-f5ff-4484-bf7a-84f6b5ab2dc3.png">

# Tasks
- [ ] ~Show the modal for FSE -> FSE theme switch where XXXXX criteria to be decided~
- [ ] ~Hide the modal when the live preview visual is broken or shows something useless~
- [ ]  Hide the modal for all cases. See discussion  p1675816912909179-slack-C048CUFRGFQ.
- [ ] Inform  @ianstewart , @tanjoymor and HE before this ships to users p1674190501522309/1674172748.806669-slack-C048CUFRGFQ
- [ ] Write a short post on wpcom happy

@Automattic/lego

fushar:
Update on investigation: I've narrowed down the possible culprits; the preview loads if I commented out the following two lines 😅 

- fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sjc%2Qgurzr%2Qcerivrj.cuc%3Se%3Q09r44316%23124-og
- fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sjc%2Qgurzr%2Qcerivrj.cuc%3Se%3Q09r44316%23128-og

Will investigate further...

[...]

PR #65976, opened 2022-07-26
Homepage replacement on Atomic: Toggle modal based on flag

andres-blanco:
This PR enables the homepage modal on Atomic behind the `themes/atomic-homepage-replace` flag.

### Note about the previews
image
**The current behaviour of the preserve homepage option doesn't seem to be working** Please make a comment on this PR about this subject. I think that it's best to remove the previews on the modal on Atomic on a subsequent PR.

### Temporary steps until wpcomsh PR is merged
This is the `wpcomsh` PR that's needed in order for the modal to work: 1046-gh-Automattic/wpcomsh. If it's merged ignore the following steps and go directly to `Testing`.
1. Set up a WoA site. `/atomic/wpcom-dev-blogs/ ` on mc.
1. Checkout the `wpcomsh` PR on your dev environment. 1046-gh-Automattic/wpcomsh
2. run  `WPCOMSH_DEVMODE=1 make clean build`
3. run `rsync -avze ssh --delete build/wpcomsh [YOUR WOA SITE]@sftp.wp.com:htdocs/wp-content/mu-plugins`

### Testing 
1. Apply this diff.
2. Go to an Atomic site (or the WoA site you've created if the `wpcomsh` PR is not merged).
3. Navigate to `/themes/[YOUR_SITE]?flags=themes/atomic-homepage-replace`.
4. Select a theme and click on `Activate`.
5. Verify you see the modal.
6. Verify both options work as expected.
7. Navigate to `/themes/[YOUR_SITE]`.
8. Select a theme and click on `Activate`.
9. Verify there's no change from current Atomic behaviour and the homepage content is preserved.

Related to https://github.com/Automattic/wp-calypso/pull/65060 and https://github.com/Automattic/wp-calypso/issues/56869

PR #65846, opened 2022-07-21
Homepage replacement on AT: Fix default behaviour

andres-blanco:
The current behaviour on Atomic is to always preserve the home page. 
This PR sets the `dont_change_homepage` to `true` on the `themes/mine` POST request that is made when the `Activate` button is pressed. If this change is not added now, when this PR (1046-gh-Automattic/wpcomsh) is merged the default action would change on Atomic to "Always replace homepage" and we don't want that.

### Testing
1. Apply this PR.
2. Go to a Simple site.
3. Verify changing themes works as expected on its two flavors (replacing and preserving homepage).
4. Go to an Atomic site.
5. Verify changing themes preserves the homepage, the behaviour should be the same as WP.com.

PR #73726, opened 2023-02-23
Deprecate Theme Switch modal

miksansegundo:
Related to https://github.com/Automattic/wp-calypso/issues/69687
Site Onboarding Flows Homepage Setup > Switching to another theme PCYsg-Ln8-p2
Theme Switch User Research pekYwv-uV-p2
Theme Experience Walkthrough > Issues for preserving templates pekYwv-BQ-p2 
Persist template content on theme switch pbxlJb-2X4-p2

## Proposed Changes

* Remove the theme switch modal
* Pass `dont_change_homepage: true` in the request `POST /themes/mine`

## Description

In this PR, we always pass `dont_change_homepage: true` in the request for a theme switch, which is the same as choosing the "Switch theme, preserving my homepage content" option in the modal. 

We are forcing users to preserve their homepage content on theme switches, but we might have exceptions when testing.

**Where is the logic for `POST /themes/mine`?**
- Simple sites fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qnpgvir%2Qgurzr%2Qraqcbvag.cuc%3Se%3Q1p463405%2330-og
- Atomic sites fbhepr%2Skers%2Swrgcnpx%2Scebwrpgf%2Scyhtvaf%2Swrgcnpx%2Swfba%2Qraqcbvagf%2Swrgcnpx%2Spynff.wrgcnpx%2Qwfba%2Qncv%2Qgurzrf%2Qnpgvir%2Qraqcbvag.cuc%3Se%3Q3n3p4nqp%2336-og

### TODO
 - Add a modal to explain exactly what will happen, including what will happen to their "in-template" content of their current theme and a mention to their reading settings.
 - Update the field guide page: Site Onboarding Flows Homepage Setup > Switching to another theme PCYsg-Ln8-p2

### Known issues
- https://github.com/WordPress/gutenberg/issues/25071
- https://github.com/WordPress/gutenberg/issues/47420
- https://github.com/WordPress/gutenberg/issues/31954

## Testing Instructions

**The expected result after a theme switch is that the content on the homepage (front of the site), excluding template parts (e.g. header/footer), must be preserved.** 

Currently when you activate a new theme, any of the old theme templates that you have customised are retained. For this reason, even though template parts of the homepage template are not preserved after switching to another theme, they will be back if we switch back to the old theme.

The homepage outcome after a theme switch depends on the following:

- the previous homepage templates edited before the switch
- the site reading settings before the switch
- the theme you switch from
- the theme you switch to

### Test flows

**Flow 1 "Skip"**
  - Create a new site
  - Skip to dashboard 
  - Switch theme from `Appearances -> Themes` multiple times

**Flow 2 "Pick a design"** 
  - Create a new site
  - Pick a theme from the design picker
  - Switch theme from `Appearances -> Themes` multiple times

**Flow 3 "Assembler"** 
  - Create a new site
  - Click "Start designing" from the bottom of the design picker
    - Flow 3.1 
      - Add patterns
    - Flow 3.2 
      - Don't add patterns
    - Continue
  - Switch theme from `Appearances -> Themes` multiple times

### Test buttons

From the theme detail, click "Activate this design" 

Screenshot 2566-02-24 at 18 17 36

From the theme showcase, open the three dots menu and click "Activate"

Screenshot 2566-02-24 at 18 17 21

### Test themes switch by theme type
- Classic theme => Classic theme
- Classic theme => Block theme
- Block theme => Block theme
- Block theme => Classic theme

#### Classic themes
Classic themes with **posts** as homepage
  - e.g.: Dyad 2, Independent Publisher 2, Ixion 

Classic themes that uses a **page** as homepage 
 - e.g.: Alves, Twenty Seventeen

#### Block themes
Block themes with **posts** as homepage 
  - whose homepage template is "Index"
    - e.g.: Zoologist, Geologist, Rainfall, Masu
  - whose homepage template is "Home"
    - e.g.: Pendant / Archeo / Skatepark

Block themes that uses a **page** as homepage
  - whose homepage template is "Header and Footer Only"
    - e.g.: Appleton / Dorna / Russell
  - whose homepage template is "Footer Only"
    - e.g.: Ames

### Test sites

From `wordpress.com/themes/{ site slug }`
 - Simple (free)
 - Atomic (business)
 - Jetpack (jurassic.ninja + connection)

## Pre-merge Checklist


packages/calypso-e2e/src/lib/pages/themes-detail-page.ts
```diff
@@ -9,10 +9,6 @@ const selectors = {
 	activateDesignButton: 'button:text("Activate this design")',
 	customizeDesignButton: 'span:text("Customize site")',
 
-	// Activate modal
-	activateModal: '.themes__auto-loading-homepage-modal',
-	activateModalButton: '.dialog__action-buttons button:has-text("Activate")',
-
```
worldomonation, line 15: **praise** - thanks for not just removing the relevant part of the code but also removing the orphaned selectors.

client/my-sites/themes/thanks-modal.jsx
```diff
@@ -334,19 +330,6 @@ const ConnectedThanksModal = connect(
...
 
-		const isAtomic = isSiteAtomic( state, siteId );
-		const isJetpack = isJetpackSite( state, siteId );
-		const hasAutoLoadingHomepage = themeHasAutoLoadingHomepage( state, currentThemeId, siteId );
-
-		// Atomic & Jetpack do not have auto-loading-homepage behavior, so we trigger the layout picker for them.
-		const customizeUrl =
-			( isAtomic || isJetpack ) && hasAutoLoadingHomepage
-				? addQueryArgs(
-						{ 'new-homepage': true },
```
miksansegundo, line 345: I'm removing this part because the param `?new-homepage=true` is meant to open the modal in the Post editor to choose a starter page template for a new homepage, assuming Atomic and Jetpack sites don't support auto-loading-homepage behavior. However, Atomic sites support it. See https://github.com/Automattic/wp-calypso/pull/65976. 

In this PR, we always pass `dont_change_homepage: true`, which skips the auto-loading-homepage behavior and that works on Atomic & Jetpack. See https://github.com/Automattic/wp-calypso/pull/65846#issuecomment-1192650587 and fbhepr%2Skers%2Sjcpbzfu%2Ssrngher%2Qcyhtvaf%2Snhgbybnq%2Qubzrcntr%2Qercynprzrag.cuc%3Se%3Q923495oq%2320-og 

I'll test these scenarios again to confirm we are safely removing that param.

client/state/themes/actions/activate.js
```diff
@@ -29,33 +22,9 @@ export function activate(
 	siteId,
 	source = 'unknown',
 	purchased = false,
-	keepCurrentHomepage = false
+	keepCurrentHomepage = true
```
miksansegundo, line 25: I'm setting `keepCurrentHomepage: true` as default here, but it looks too fragile. We can also set `dontChangeHomepage: true` as default in the action `activateTheme`: https://github.com/Automattic/wp-calypso/blob/7bca130291db0f8a0af6cb7262bc4cdbedeca99f/client/state/themes/actions/activate-theme.js#L40 

What do you think?

okmttdhr, line 25: badge
Should we pass the `persist_homepage_template: true` always as well, or not?
fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qnpgvir%2Qgurzr%2Qraqcbvag.cuc%3Se%3Q1p463405%2333-og

miksansegundo: Thanks for pointing that out. 

I missed that `persist_homepage_template` is always `false` because the  `themes/theme-switch-persist-template` feature flag is `false` in all environments. 

That means the logic related to preserving the home template is not being used. 

See pbxlJb-2X4-p2#comment-2162 and the changes in that API fbhepr%2Suvfgbel%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qnpgvir%2Qgurzr%2Qraqcbvag.cuc-og

miksansegundo:
Some interesting conversations on the theme switch and the template persistence

- https://github.com/WordPress/gutenberg/issues/31954
- https://github.com/WordPress/gutenberg/issues/25071

Here some of the comments:

**User flows**

First we should outline user flows that we anticipate or want to make possible. For example:

1. Choosing a theme and wanting to switch everything to the new look (discard any user modifications in templates or styles).
2. Wanting to preserve some customizations such as colors or header navigation. Or conversely; wanting to replace styles and navigation but leave everything else intact.
3. Wanting to preserve everything (essentially incorporating a new theme's templates or patterns on demand).
4. There could be themes with different set of templates (say, a commerce theme that brings in new templates) that are additive.

What is clear is that we can no longer assume for users what they may want to do but we should choose a good default. 

It'd be reasonable to make the first option the default since it's what most are accustomed to, but we can discuss it. Semantic template parts are also a tool we have for managing some expectations since they can overlap between themes.

Theme switching should have the option of being a more granular operation in the site editor, walking users through some steps where they could choose to retain or discard specific templates (like a header) or global styles. Right now there's a lack of clarity upon switching themes about whether a specific part of the site will be replaced or not.

**User flow:**

I create a blank.html template that only renders the post content block (just like BlockBase). I've assigned it to display on a particular page, as I want no headers or footers — just my block content for this landing page. I change my theme to get a new look cause I feel like it, but now my landing page has a header, footer and an empty sidebar next to it.

To resolve this today, I need to create a blank custom template and assign it to the page once again. If I ever change my theme, I end up with the same issue, needed to create the blank template again (while deleting the old one).

**User next steps:**

While we can't assume that every time I change a theme, all my custom templates should stay, I do lean in the direction that that behavior should be the default experience.

Maybe we have a mechanism that let's me opt to resetting all templates as the theme's default templates, instead of what happens today where everything resets when a theme is changed.

In addition, I should also be able to use any template I've created on any theme. I'm not sure what the purpose behind locking templates to themes is, but its seems counter-intuitive.

**Block templates are theme specific but that is not obvious.**

I agree this deserve to be made clearer, maybe in the welcome guide? (Open to ideas here).

I wanted to detail the reasoning more here. The block templates are not necessarily theme specific, they can easily translate to other themes but theme CSS design can have impact and alter the result. In the future, we do want to allow them to be theme-agnostic but for the initial release we wanted to keep them theme specific because otherwise, there’s a lot of UX/UI work that need to happen on theme switching and elsewhere to take into account these new possibilities. 

It’s also a decision based on the initial version of FSE themes we want to ship. FSE themes rely on these templates as well (wp_template) and we don’t want to solve the cross-themes templates in the first release but we want to start exploring this at their own pace. Keeping the template editor theme specific allow us to simplify the mental model of these templates: These are just editable “Page Templates” (the custom page templates themes provide)

miksansegundo:
@okmttdhr for the Live Preview spike, you might find relevant the details about the Theme switch that is on the description and comments in this PR. Feel free to reach out to me to discuss anything related to it.

Here some interesting threads: 
p1677166541732979/1677165514.716649-slack-C048CUFRGFQ
p1677204510072609-slack-C048CUFRGFQ

The work on this PR has been stopped because removing the modal won't fix anything and the feedback from HEs was that the modal must explain exactly what will happen after the theme switch. The answer to that question is complex because there are several factors that can break the preview and the theme switch. So, in order to tell users what will happen we first must understand what is failing and that implies IMO understanding and fixing the preview and the theme switch.

okmttdhr:
Thank you for your sharing, @miksansegundo!
You are definitely one who is familiar with theme-switching. I will ask everything after I catch up!

PR #72697, opened 2023-01-27
Purchase Management: Add early return to block loading of '/cancel' endpoint

jjchrisdiehl:
#### Proposed Changes

Main thread https://github.com/Automattic/payments-shilling/issues/1321

This small PR adds an early return to the CancelPurchase component to ensure that `/cancel` redirects back to the purchase list if no site is found for the purchase.

#### Testing Instructions

- This will require two wpcom accounts and a self-hosted site, I recommend using https://jurassic.ninja/
- Have account A create a site and connect Jetpack
- Have account A invite account B as an admin from the Calypso dashboard
- Have account B purchase a Jetpack Complete plan
- Have account A remove account B
- From account B, go to /me/purchases and click on the Jetpack Complete plan
- From the manage purchase page, click on Cancel, this should redirect you back to the Purchases list
- Now try going to `http://calypso.localhost:3000/me/purchases/[SITE URL]/[SUBSCRIPTION ID]/cancel`, it should redirect back to the Purchases list

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to https://github.com/Automattic/payments-shilling/issues/1321

PR #73736, opened 2023-02-23
Revert "Purchase Management: Add early return to block loading of '/cancel' endpoint"

jjchrisdiehl:
Reverts Automattic/wp-calypso#72697

Issue: https://github.com/Automattic/payments-shilling/issues/1405


jjchrisdiehl:
Doesn't appear to cause any issues with the purchase management page loading.

PR #73749, opened 2023-02-24
Reader: make case consistent for tag header follow button

bluefuton:
In the Reader tag stream, the follow button currently has inconsistent capitalization. It's "Follow Tag" and then "Following tag" once you're following:

Screenshot 2023-02-24 at 14 56 45
Screenshot 2023-02-24 at 14 56 35

This PR standardizes on sentence case, so the labels become "Follow tag" and "Following tag" for consistency.

### Testing instructions
After switching to the PR's branch or visiting the calypso.live link:

1. Visit the Reader tag stream for a tag you do not follow (e.g. /tag/mountains)
2. Verify that the follow button label is "Follow tag"
3. Click on the follow button to follow the tag
4. Verify that the follow button label is "Following tag"

### DIFF
--- a/client/reader/tag-stream/header.jsx
+++ b/client/reader/tag-stream/header.jsx
@@ -96,7 +96,7 @@ class TagStreamHeader extends Component {
 				<div className="tag-stream__header-follow">
 					{ showFollow && (
 						<FollowButton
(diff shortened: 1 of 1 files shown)


davemart-in (approved):
looks good to me!

PR #73782, opened 2023-02-24
Checkout: Add free-form address field if required for country

sirbrillig:
## Proposed Changes

This adds a free-form "Address" field to the contact/billing details step in checkout if it is required for the tax calculations of that country.

This is separate from the "Address for VAT" field currently in the VAT form (which will override the "Address" field if set, the same way we work with the "Organization" field).

Without a domain in the cart:

Screenshot 2023-02-24 at 6 10 59 PM

With a domain in the cart (no change; it uses the existing "Address" field):

Screenshot 2023-02-24 at 6 10 47 PM

## Testing Instructions

Automated tests are included. For manual testing:

- First, you'll need to modify the countries endpoint on your sandbox so that it returns a country (eg: Switzerland) as having a required address. You can do that by applying D102799-code and sandboxing the API.
- Add a non-domain product to your cart and visit checkout. Edit the billing information step if it is not active.
- Verify that in the contact step you see an "Address" field.
- Open your browser's devtools to view network activity to the `/me/shopping-cart` endpoint.
- Fill out the form and submit it.
- Verify that the shopping cart endpoint received a `tax` field which contained the "Address" you entered in the form.
- Repeat the above steps with a domain in the cart.


jjchrisdiehl:
Functional test:

For plan:
image

For domain:
image

Looks to work as expected!

PR #73783, opened 2023-02-24
Developer UX: Scroll to top on mount Performance Settings and Hosting Configuration

sejas:
- Closes https://github.com/Automattic/dotcom-forge/issues/1635

## Proposed Changes

* Create a new component `ScrollToTopOnMount` and a new prop on `Main`
* Use this new prop on `Performance settings` and `Hosting configuration` to fix the old scroll position when we navigate from SMP to one of those pages.

## Testing Instructions

* Apply this branch and try it on localhost or wpcalypso.
* Go to SMP `/sites`
* Scroll down the page
* Click on the actions menu on a business site
* Click on `Performance settings` or `Hosting configuration`
* Observe you've navigated to the correct page and the scroll is on the top. Before this PR the scroll would remain in the middle of the page.

| **Before** | **After** |
|---|---|
| scroll-before | scroll-after |

## Pre-merge Checklist


danielbachhuber:
@sejas I don't seem to have the issue where I'm scrolled halfway down the page:

https://user-images.githubusercontent.com/36432/221570295-444d8ff8-80ae-444c-81d9-b17c363414f1.mp4

Is there some reproduction step that I'm missing?

sejas:
> @sejas I don't seem to have the issue where I'm scrolled halfway down the page:
> 
> Is there some reproduction step that I'm missing?

@danielbachhuber, to reproduce it, you need to scroll on SMP. Click on Hosting configuration only when SMP is "half way" scrolled. Not in your first site, try your 10th or 20th site.

danielbachhuber:
> to reproduce it, you need to scroll on SMP. Click on Hosting configuration only when SMP is "half way" scrolled. Not in your first site, try your 10th or 20th site.

@sejas Ah, I can reproduce it now.

It seems like this bug would affect other pages too? Should we make it default behavior for `<Main>`?

sejas:
> It seems like this bug would affect other pages too? Should we make it default behavior for <Main>?

@danielbachhuber , Yes! It could affect other pages too.

First, I was executing the hook by default to ship a general fix. But I'm not confident about testing all possible navigations to pages using `Main`. So I took this less intrusive path.
I've been able to reproduce the issue only in one scenario when the user navigates from SMP to the Dashboard, so I applied the hook to those pages that I'm aware of the error. If we find more pages, we can fix the error just by adding the prop `scrollToTopOnMount`.

The fact that It slipped our testing for over 7 months makes me think it's not critical.

danielbachhuber:
> The fact that It slipped our testing for over 7 months makes me think it's not critical.

@sejas Definitely not critical but it would be nice to fix holistically, instead of with a bandaid.

@Automattic/team-calypso-frameworks How might you recommend addressing this issue?

tyxla:
> @Automattic/team-calypso-frameworks How might you recommend addressing this issue?

This is not a new problem, and it has already been solved in other areas.

For example, if you click on any of the Settings submenu items in the sites sidebar (leading to a page in the app), it will scroll to the top and then navigate to the next route. Ideally, we should mirror that behavior. Note that in the existing behavior, we first scroll, and then navigate. You could borrow it from here:

https://github.com/Automattic/wp-calypso/blob/1e2a4a48e9ede45bfb0109dec8299fc2dfdec022/client/my-sites/sidebar/item.jsx#L38

and ideally, it would be abstracted in a place where it could be reused, so the next person working on such an issue would just be able to borrow it.

Hope that helps, but let me know if we can aid with anything else.

sejas:
@tyxla, thanks for sharing that solution.

I've created a new PR with that approach, but it doesn't work. It scrolls before navigating, but the destination page is not at the top. I also tried adding a big delay between the scroll and the navigation, and it's still failing.

- https://github.com/Automattic/wp-calypso/pull/73881

Here is the screencast:


https://user-images.githubusercontent.com/779993/221894914-288951b3-8154-4c8e-a59d-30d0dfb15099.mp4


@danielbachhuber, What do you think if we merge this PR until we find a better solution?

wojtekn:
@sejas I tested it, and it works as expected.

I thought about an alternative solution. What if we modified the `scrollToAnchor` function and scroll to the top if there is no `hash` in the request?

https://github.com/Automattic/wp-calypso/blob/5b2baa9f088e62111c72dd61b0d4b2a559925a35/client/lib/scroll-to-anchor/index.ts#L19-L36

We would need to improve 1660-gh-Automattic/dotcom-forge too, but that way, we would only need to add one line of code, and we would have scrolling handled by one component.

sejas:
> @sejas I tested it, and it works as expected.
> 
> I thought about an alternative solution. What if we modified the `scrollToAnchor` function and scroll to the top if there is no `hash` in the request?
> 
> We would need to improve [1660-gh-Automattic/dotcom-forge](https://href.li/?https://github.com/Automattic/dotcom-forge/issues/1660) too, but that way, we would only need to add one line of code, and we would have scrolling handled by one component.

That solution would be very similar to this PR. It fixes the problem on the destination page.
I agree that we can unify the code with `scrollToAnchor`. But I would prefer to move the logic to scrollToAnchor into `<ScrollToTopOnMount />` component.
`scrollToAnchor` is not a hook, and it can be difficult to use for function components.
I like the idea of fixing the issue just by adding a child component (`<ScrollToTopOnMount />`) wherever you need it. Or if you are using `<Main />`, you can add a prop, and that's all you need to fix the scroll.

tyxla:
> I've created a new PR with that approach, but it doesn't work. It scrolls before navigating, but the destination page is not at the top. I also tried adding a big delay between the scroll and the navigation, and it's still failing.

Have you explored why it doesn't work? I'm curious because that approach does work for the menu items in the "my sites" sidebar. I haven't explored any possible reason but I assume there may be something missing from the solution.

sejas:
> Have you explored why it doesn't work? I'm curious because that approach does work for the menu items in the "my sites" sidebar. I haven't explored any possible reason but I assume there may be something missing from the solution.

I'm taking a deeper look 🤞 comparing different destination pages.

sejas:
After investigating it, I cannot find the source of the issue. I've removed the sidebar and added a delay between the scroll and the navigation, but the issue is still there.
Even if we make it work. I doubt we like the solution. Because, in our case, we start with different content (SMP), and scrolling on the previous page feels a bit weird.

Here is a patch adding a delay of one second between the scroll and the navigation https://github.com/Automattic/wp-calypso/pull/73881#issuecomment-1451931111

**Screencast with 1s delay and without sidebar**

https://user-images.githubusercontent.com/779993/222453041-6c0451b4-25f7-4d8a-98a8-0fde0e288ffc.mp4

sejas:
@danielbachhuber , @tyxla , 
I've updated this PR https://github.com/Automattic/wp-calypso/pull/73783/commits/51dd7c5ddaf968bb80a8fa741eb4285b2f64d319 by applying a more holistic fix as [suggested](https://github.com/Automattic/wp-calypso/pull/73783#issuecomment-1447141998) by @danielbachhuber .
I'm also removing the [old scrollTo commented ](https://github.com/Automattic/wp-calypso/pull/73783#issuecomment-1447867952) by @tyxla.

With this PR we keep the current behavior, we fix our issue, and future issues where other parts of the application link to the Dashboard.

client/my-sites/site-settings/settings-performance/main.jsx
```diff
@@ -55,7 +55,10 @@ class SiteSettingsPerformance extends Component {
 		const siteIsJetpackNonAtomic = siteIsJetpack && ! siteIsAtomic;
 
 		return (
-			<Main className="settings-performance site-settings site-settings__performance-settings">
+			<Main
+				className="settings-performance site-settings site-settings__performance-settings"
+				scrollToTopOnMount
```
danielbachhuber: This isn't necessary anymore?

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react';
+
+export function ScrollToTopOnMount() {
+	useEffect( () => {
+		if ( ! window.location.hash ) {
```
danielbachhuber, line 5: Can we avoid scrolling if the user simply refreshes the page (and expects their scroll position to persist)?

danielbachhuber (commented):
Because `<Main>` is used on all views, this seems like it has the potential to accidentally break lots of things.

What's the best way to guard against this, and roll the change out safely?

client/my-sites/site-settings/settings-performance/main.jsx
sejas: True! Thanks! https://github.com/Automattic/wp-calypso/pull/73783/commits/9b48a9dd99c9abde4a111e3853f8da5ffd226bdf

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
sejas: > Can we avoid scrolling if the user simply refreshes the page (and expects their scroll position to persist)?

I'll investigate it!


> Because `<Main>` is used on all views, this seems like it has the potential to accidentally break lots of things.
> 
> What's the best way to guard against this, and roll the change out safely?

We can deploy it behind a feature flag on `stage` and deploy it to production after ~2 weeks.

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react';
+
+export function ScrollToTopOnMount() {
```
tyxla, line 3: It would be a bummer not to reuse this from somewhere else. It's been done many times, and here we're introducing a new one without attempting to reuse / refactor the original ones.

Here are a few examples of similar prior art I found on my first search:

https://github.com/Automattic/wp-calypso/blob/f2e6f88ae65c8571bbac21b284c29b48ec987598/client/reader/full-post/controller.js#L8

https://github.com/Automattic/wp-calypso/blob/f2e6f88ae65c8571bbac21b284c29b48ec987598/client/reader/full-post/controller.js#L8

https://github.com/Automattic/wp-calypso/blob/7e6d2b9c30257d9270c1c3aa8db0168945cec8a4/client/my-sites/plugins/controller.js#L141

https://github.com/Automattic/wp-calypso/blob/d648de9f1b6ebfa95d7729e1c38fb9f0ef61033e/client/my-sites/picker/picker.jsx#L74

I generally don't have very strong feelings about this, but just imagine if every time we need something so generic, we reinvent the wheel and create a new solution. We'll end with dozens of them, and the cumulative effect of that practice can impact Calypso negatively, from a maintenance, performance, and developer experience perspective.

tyxla, line 3: Could this have been a hook instead of a component?

client/components/main/index.jsx
```diff
@@ -18,6 +19,7 @@ export default function Main( {
 
 	return (
 		<main className={ classes } id={ id || null } role="main">
+			<ScrollToTopOnMount />
```
tyxla, line 22: I'm not convinced we want to apply this change everywhere where `<Main />` is used. 

For example, how can you be sure that `<Main />` won't be remounted while staying on the same screen and this would cause an unintended scroll to the top of the page?

client/components/main/README.md
```diff
@@ -16,3 +16,4 @@ function render() {
 
 - `className`: Add your own CSS class to the wrapper.
 - `wideLayout`: Use wide layout (larger `max-width`) for the section.
+- `scrollToTopOnMount`: Scroll to top when the component is mounted to avoid showing the previous scroll position.
```
tyxla: Remnant of the previous implementation?

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
tyxla, line 3: It's generally a good practice for every new `@automattic` component to have:

* A `docs/example`
* A README
* Tests
* Representation in `devdocs` and/or `storybook`.

sejas: Yes!, I can extract the hook I use inside 👌

client/components/main/README.md
sejas: Yes!, removed https://github.com/Automattic/wp-calypso/pull/73783/commits/a76772696e29e8d47fa618ef3906fd496dbaef2d
Thank you!

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
sejas: Yes, I like to use `storybook` because it has the fast refresh built in, and it's isolated from the rest of the UI, which makes your component more reusable. If I keep this component, I'll add the necessary documentation.

client/components/main/index.jsx
sejas: I agree that this PR was too ambitious and hard to test. The scenario where Main is re-mounted, and we will not want to scroll to the top, is very unlikely, but I agree that we need to ensure it won't cause any problems.
An example where Main is re-mounted is when the user clicks on a different page in the sidebar menu.

packages/components/src/scroll-to-top/scroll-to-top-on-mount.tsx
sejas: @tyxla, Thank you for your in-depth review. I really appreciate it.
Before continuing with this PR, I'll reconsider a new approach to fix this issue. I'll make sure to tag you when I have something solid in mind.
I'll reuse the existing resources and try to fix it holistically.
Thank you!

danielbachhuber:
@sejas I suggest we punt on this PR. I don't think it's worth investing more time in at this point, and I think Hosting Configuration might live somewhere else in the future anyway.

sejas:
Closing this PR, we couldn't find a general solution, and future changes will come to the Hosting Configuration page.

PR #73785, opened 2023-02-24
Rework the Jetpack E2E to work for WPCOM staging and remote sites

dpasque:
Related to #

### Background

We've been working on improving CI and test coverage for Jetpack's integration with WPCOM. We've been especially focused on adding test coverage during Jetpack deploys on WPCOM.

For more context, you can read this P2: pd5faL-rf-p2

In progressing that forward, we realized that the state of Jetpack testing in Calypso needed some working! We had a few old tests that targeted a remote site (Pressable) connected to Calypso. But that user had a lot of test data built up on them, the test selection wasn't particularly focused, and some were failing! We also didn't have a way to target different relevant aspects of Jetpack on WPCOM/Calypso.

### The New Paradigm

_As a quick reminder... In our E2E set up, we use `jest-groups` to determined which **tests scripts** to run, and ENV vars to determine which **conditions, users, or sites** to use in that run._

We now have three different conditions/users/sites for testing Jetpack's integration with WPCOM/Calyspo. We drive these with the `JETPACK_TARGET` ENV variable. There are unique TeamCity builds for each of these three cases.
1. `wpcom-production` - A WPCOM site that is running the production version of Jetpack on WPCOM. Put another way, this is what just about every WPCOM site is doing! So in most cases, these conditions just fall back to our existing test users.
2. `wpcom-staging` - A WPCOM site that always runs the staging version of Jetpack on WPCOM. We have a couple of hardcoded users and sites that always use the staging version of Jetpack. This allows testing new releases of Jetpack before releasing to the rest of WPCOM. This similar to what we do with new Gutenberg version and "Edge" sites!
3. `remote-site` A remote (hosted Pressable, not on WPCOM) site that has been Jetpack connected to WPCOM, so can be interacted with through Calypso. This is useful for testing Calypso as a client to interact with remote sites, and testing how specific Jetpack functionality works for remote, connected sites. 

We also have created two new `jest-group` test suites:
1. `jetpack-wpcom-integration`: for testing the Jetpack deployment on WPCOM. These tests focus on making sure new Jetpack versions play well with all the other code on WPCOM. These tests will always use a `JETPACK_TARGET` of `wpcom-production` and `wpcom-staging`.
2. `jetpack-remote-site`: for testing a remote, connected Jetpack site through the Calypso interface. These tests will always use a `JETPACK_TARGET` of `remote-site`

Why two different `jest-groups`? When you start mapping out relevant E2E tests, the suites for the different conditions end up being pretty different!  Consider Jetpack blocks, for example. We really care about testing those on WPCOM sites during Jetpack upgrades. Part of the Jetpack block code may conflict with our Gutenberg version, or other code running on our WPCOM infrastructure.  But Jetpack Blocks are not particularly interesting to test on a remote site! We just redirect to the wp-admin editor on that remote site. So at that point, you're not testing any code related to Calypso or WPCOM. 

### Tests in Each Suite

`jetpack-wpcom-integration`:
- `blocks__core-jetpack-extended`
- `blocks__jetpack-earn`
- `blocks__jetpack-grow`
- `blocks__jetpack-other`
- `blocks__media`
- `editor__page-basic-flow`
- `editor__post-basic-flow`
- `fse__temp-smoke-test`
- `media__edit`
- `media__upload`
- `settings__media`

`jetpack-remote-site`:
- `theme__details-preview-activate`
- `media__edit`
- `media__upload`
- `settings__media`
- `plugins__browse`
- `plugins__install-remove`
- `reader__view`  

### Other Changes
In doing all of this, there were a handful of other changes I had to do to make things play nicely together...

- I made a new user and Pressable site for testing the `remote-site` condition. The old one was really bloated and it was a time for a rotation.
- I made our media tests delete the files they upload (if they get to that point). I figured that is just best practice and minimizes the clean-up we need to do later. Especially for remote sites where we don't have as easy batch clean-up tools available.
- I had to do some clean-up around test users, and moved more tests over to determining the test user from the ENV variable configuration. Our media tests also looped through three hardcoded users within the same test, which doesn't fit well with our existing paradigm. **The main loss from that change is we now don't run our media tests for Atomic on PR builds.** We can easily add that back, but we should do it by making a dedicated Atomic PR TeamCity build that runs all the tests on Atomic!
- Our plugins tests needed some TLC. There were two browse ones that were effectively testing the same thing, so I collapsed them to one. I also took the the install/remove plugin and got it passing again!

## Testing Instructions

Lots of CI builds need to pass!

- [x] Regular PR desktop
- [x] Regular PR mobile
- [x] Jetpack remote-site desktop
- [x] Jetpack remote-site mobile
- [x] Jetpack wpcom-staging desktop
- [x] Jetpack wpcom-staging mobile
- [x] Jetpack wpcom-production desktop
- [x] Jetpack wpcom-production mobile

## Pre-merge Checklist


test/e2e/specs/media/settings__media.ts
```diff
@@ -0,0 +1,87 @@
...
+	const accountName = getTestAccountByFeature( envToFeatureKey( envVariables ) );
+
+	beforeAll( async function () {
+		page = await browser.newPage();
+
+		const testAccount = new TestAccount( accountName );
+		await testAccount.authenticate( page );
+	} );
+
+	if ( envVariables.JETPACK_TARGET === 'remote-site' ) {
```
worldomonation, line 35: **comment:non-blocking** - I am not the biggest fan of conditional logic at the block/step level, but it is also true that there is no cleaner way of doing this.

The most commonly cited way of achieving this is to use the `skipIf` implementation that I added some time ago. That in itself feels a bit like a kludge.

test/e2e/specs/media/media__edit.ts
```diff
@@ -19,49 +23,50 @@ declare const browser: Browser;
 describe( DataHelper.createSuiteTitle( 'Media: Edit Media' ), function () {
```
worldomonation, line 23: **comment:non-blocking** - unrelated to the scope of this, but we should rewrite this spec to upload and delete the image using the RestAPIClient in order to limit the scope of the feature under test to just the editing process.

worldomonation (approved):
The change here is large, but IIUC most of them are file name changes, addition of groups, or slight refactoring of specs themselves. 

I've also looked at the changes to the test framework itself and it registers as a safe change to me.

test/e2e/specs/media/media__edit.ts
dpasque: That's a great idea!! 🎉

test/e2e/specs/media/settings__media.ts
dpasque: Yeah, I agree! 😕 

> The most commonly cited way of achieving this is to use the skipIf implementation that I added some time ago. 

I can definitely switch to that here! I felt like since we were basically redirecting to two different text blocks, using the "if" clause read a little more simply (because it said what you **were** testing in each case. I'll probably merge for now, but can easily switch this back later!

PR #73786, opened 2023-02-24
Add Akismet checkout routes, allow for site-less cart

jboland88:
## Proposed Changes

* This PR adds new routes for `/akismet/checkout` and allows products at that route to be added to the cart without an associated site.
* This PR does not add the full purchase flow for the products (there is no thank you page post-purchase and there are some other improvements to add). The goal of this PR is to add the structure for the checkout flow enough to complete a purchase so that additional PRs can iterate in tandem.

## Testing Instructions

* Apply this patch to your local calypso Environment. Note that this patch is only configured to run in development and testing environments at the moment.
* You will need apply D102794-code on your sandbox to run this test. **Follow the setup instructions on that diff before proceeding here**.
* After setting up D102794-code, navigate to `calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1`. This should take you a checkout page with the Akismet Plus product in the cart. If this does not work, make sure you followed the setup steps in D102794-code.

Screen Shot 2023-03-08 at 4 49 58 PM

* Click to complete the purchase. You should be redirected to a thank you page `http://calypso.localhost:3000/checkout/akismet/thank-you/ak_plus_yearly_1`. This page does not exist yet, and will be blank.
* You should see the purchase in the list at `me/purchases`, however, you will not be able to view details or manage the purchase yet. This is how you confirm that the purchase was successful. (You might also get an emailed reciept)

Screen Shot 2023-03-08 at 5 25 36 PM

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -95,7 +95,6 @@ export default function CheckoutMain( {
 	disabledThankYouPage,
 	sitelessCheckoutType,
 	jetpackSiteSlug,
-	akismetSiteSlug,
 	jetpackPurchaseToken,
 	isUserComingFromLoginForm,
 	customizedPreviousPath,
@@ -124,7 +123,6 @@ export default function CheckoutMain( {
 	disabledThankYouPage?: boolean;
 	sitelessCheckoutType?: SitelessCheckoutType;
 	jetpackSiteSlug?: string;
-	akismetSiteSlug?: string;
 	jetpackPurchaseToken?: string;
 	isUserComingFromLoginForm?: boolean;
 	customizedPreviousPath?: string;
@@ -140,20 +138,16 @@ export default function CheckoutMain( {
 	const isSiteless = sitelessCheckoutType === 'jetpack' || sitelessCheckoutType === 'akismet';
 	const { stripe, stripeConfiguration, isStripeLoading, stripeLoadingError } = useStripe();
 	const createUserAndSiteBeforeTransaction =
-		Boolean( isLoggedOutCart || isNoSiteCart ) &&
-		sitelessCheckoutType !== 'akismet' &&
-		sitelessCheckoutType !== 'jetpack';
+		Boolean( isLoggedOutCart || isNoSiteCart ) && ! isSiteless;
 	const reduxDispatch = useDispatch();
 
-	const updatedSiteSlug = useCallback( () => {
+	const updatedSiteSlug = useMemo( () => {
 		if ( sitelessCheckoutType === 'jetpack' ) {
 			return jetpackSiteSlug;
 		}
-		if ( sitelessCheckoutType === 'akismet' ) {
-			return akismetSiteSlug;
-		}
+
 		return siteSlug;
-	}, [ akismetSiteSlug, jetpackSiteSlug, sitelessCheckoutType, siteSlug ] );
+	}, [ jetpackSiteSlug, sitelessCheckoutType, siteSlug ] );
 
 	const showErrorMessageBriefly = useCallback(
 		( error ) => {
@@ -188,7 +182,7 @@ export default function CheckoutMain( {
 		isInModal,
 		usesJetpackProducts: isJetpackNotAtomic,
 		isPrivate,
-		siteSlug: updatedSiteSlug(),
+		siteSlug: updatedSiteSlug,
 		sitelessCheckoutType,
 		isLoggedOutCart,
 		isNoSiteCart,
@@ -241,7 +235,7 @@ export default function CheckoutMain( {
 	// post-checkout page decided by `getThankYouUrl` and therefore must be
 	// passed the post-checkout URL before the transaction begins.
 	const getThankYouUrlBase = useGetThankYouUrl( {
-		siteSlug: updatedSiteSlug(),
+		siteSlug: updatedSiteSlug,
 		redirectTo,
 		purchaseId,
 		feature,
@@ -318,7 +312,7 @@ export default function CheckoutMain( {
 
 	const { isRemovingProductFromCart, removeProductFromCartAndMaybeRedirect } =
 		useRemoveFromCartAndRedirect(
-			updatedSiteSlug(),
+			updatedSiteSlug,
 			createUserAndSiteBeforeTransaction,
 			customizedPreviousPath
 		);
@@ -345,7 +339,7 @@ export default function CheckoutMain( {
 		stripeConfiguration,
 		stripe,
 		storedCards,
-		siteSlug: updatedSiteSlug(),
+		siteSlug: updatedSiteSlug,
 	} );
 	debug( 'created payment method objects', paymentMethodObjects );
 
@@ -379,7 +373,7 @@ export default function CheckoutMain( {
 	const { analyticsPath, analyticsProps } = getAnalyticsPath(
 		purchaseId,
 		productAliasFromUrl,
-		updatedSiteSlug(),
+		updatedSiteSlug,
 		feature,
 		plan,
 		sitelessCheckoutType,
@@ -431,7 +425,7 @@ export default function CheckoutMain( {
 			reduxDispatch,
 			responseCart,
 			siteId: updatedSiteId,
-			siteSlug: updatedSiteSlug(),
+			siteSlug: updatedSiteSlug,
 			stripeConfiguration,
 			stripe,
 			recaptchaClientId,
@@ -575,7 +569,7 @@ export default function CheckoutMain( {
 		isInModal,
 		isComingFromUpsell,
 		disabledThankYouPage,
-		siteSlug: updatedSiteSlug(),
+		siteSlug: updatedSiteSlug,
 		sitelessCheckoutType,
 		checkoutFlow,
 	} );
@@ -735,7 +729,7 @@ export default function CheckoutMain( {
 					removeProductFromCart={ removeProductFromCartAndMaybeRedirect }
 					showErrorMessageBriefly={ showErrorMessageBriefly }
 					siteId={ updatedSiteId }
-					siteUrl={ updatedSiteSlug() }
+					siteUrl={ updatedSiteSlug }
 				/>
 			</CheckoutProvider>
 		</Fragment>
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -148,6 +148,9 @@ export default function CheckoutMain( {
 			return jetpackSiteSlug;
 		}
 
+		// Currently, the `akismetSiteSlug` prop is not being passed to this component anywhere
+		// We are not doing any site specific things with akismet checkout, so this should always be undefined for now
+		// If this was not here to return `undefined`, the akismet routes would get messed with due to `siteSlug` returning "no-user" in akismet siteless checkout
 		if ( sitelessCheckoutType === 'akismet' ) {
 			return akismetSiteSlug;
 		}
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -59,12 +59,13 @@ import { StoredCard } from '../types/stored-cards';
 import WPCheckout from './wp-checkout';
 import type { WpcomCheckoutStoreSelectors as _WpcomCheckoutStoreSelectors } from '../hooks/wpcom-store';
 import type { PaymentProcessorOptions } from '../types/payment-processors';
-import type {
-	CheckoutPageErrorCallback,
-	SitelessCheckoutType,
-} from '@automattic/composite-checkout';
+import type { CheckoutPageErrorCallback } from '@automattic/composite-checkout';
 import type { MinimalRequestCartProduct } from '@automattic/shopping-cart';
-import type { CountryListItem, CheckoutPaymentMethodSlug } from '@automattic/wpcom-checkout';
+import type {
+	CountryListItem,
+	CheckoutPaymentMethodSlug,
+	SitelessCheckoutType,
+} from '@automattic/wpcom-checkout';
 
 type WpcomCheckoutStoreSelectors = _WpcomCheckoutStoreSelectors | undefined;
 
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -59,7 +59,10 @@ import { StoredCard } from '../types/stored-cards';
 import WPCheckout from './wp-checkout';
 import type { WpcomCheckoutStoreSelectors as _WpcomCheckoutStoreSelectors } from '../hooks/wpcom-store';
 import type { PaymentProcessorOptions } from '../types/payment-processors';
-import type { CheckoutPageErrorCallback, CheckoutType } from '@automattic/composite-checkout';
+import type {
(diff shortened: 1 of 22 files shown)


client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -130,9 +134,14 @@ export default function CheckoutMain( {
 	const isPrivate = useSelector( ( state ) => siteId && isPrivateSite( state, siteId ) ) || false;
 	const { stripe, stripeConfiguration, isStripeLoading, stripeLoadingError } = useStripe();
 	const createUserAndSiteBeforeTransaction =
-		Boolean( isLoggedOutCart || isNoSiteCart ) && ! isJetpackCheckout;
+		Boolean( isLoggedOutCart || isNoSiteCart ) &&
+		! isJetpackCheckout &&
+		! isAkismetSitelessCheckout;
 	const reduxDispatch = useDispatch();
-	const updatedSiteSlug = isJetpackCheckout ? jetpackSiteSlug : siteSlug;
+	let updatedSiteSlug = isJetpackCheckout ? jetpackSiteSlug : siteSlug;
```
sirbrillig: Since this is getting a little unwieldy (I'm sure that's the reason you split up the new option) what about changing this to an IIFE to improve clarity without resorting to multiple assignments? Something like: 

```js
const updatedSiteSlug = (() => {
  if (isJetpackCheckout) {
    return jetpackSiteSlug;
  }
  if (isAkismetCheckout) {
    return akismetSiteSlug;
  }
  return siteSlug;
})();
```

client/my-sites/checkout/get-thank-you-page-url/index.ts
```diff
@@ -234,6 +236,15 @@ export default function getThankYouPageUrl( {
 		);
 	}
 
+	// Asismet site-less checkout
+	if ( isAkismetSitelessCheckout ) {
+		// extract a product from the cart, in siteless checkout there should only be one
+		const productSlug = cart?.products[ 0 ]?.product_slug ?? 'no_product';
+
+		debug( 'redirecting to siteless Akismet thank you' );
+		return `/checkout/akismet/thank-you/${ productSlug }`;
```
sirbrillig, line 244: This function is quite complex and so has unit tests covering all possible outputs to prevent regressions. Can we add some tests for this case? See the comment at the top of the file for instructions.

client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
```diff
@@ -101,6 +103,7 @@ export default function useCreatePaymentCompleteCallback( {
 
 			// In the case of a Jetpack product site-less purchase, we need to include the blog ID of the
 			// created site in the Thank You page URL.
+			// TODO: It does not seem like this would be needed for Akismet, but marking to follow up
```
CodeyGuyDylan, line 107: Do you know why this is needed for jetpack siteless checkout? The page seems to work fine if I remove that param (for jetpack siteless)

Anyway, from what I understand so far, I also do not think that this is necessary for Akismet

client/my-sites/checkout/get-thank-you-page-url/index.ts
CodeyGuyDylan: I believe we have a task set for this here: 1204136657007202-as-1204136663449775/f

Since this PR will be feature flagged when deployed, I think it makes sense to do this separately

Correct me if I am wrong @jboland88

client/my-sites/checkout/get-thank-you-page-url/index.ts
```diff
@@ -234,6 +236,15 @@ export default function getThankYouPageUrl( {
 		);
 	}
 
+	// Asismet site-less checkout
+	if ( isAkismetSitelessCheckout ) {
+		// extract a product from the cart, in siteless checkout there should only be one
+		const productSlug = cart?.products[ 0 ]?.product_slug ?? 'no_product';
```
CodeyGuyDylan, line 241: We could potentially add a check to ensure that this `productSlug` is an Akismet product, do you think that would be useful, or unnecessary?

client/my-sites/checkout/controller.jsx
```diff
@@ -44,6 +44,14 @@ import { getProductSlugFromContext, isContextJetpackSitelessCheckout } from './u
...
 
 export function checkoutJetpackSiteless( context, next ) {
+	sitelessCheckout( context, next, { isJetpackCheckout: true } );
+}
+
+export function checkoutAkismetSiteless( context, next ) {
+	sitelessCheckout( context, next, { isAkismetSitelessCheckout: true } );
+}
+
+function sitelessCheckout( context, next, props ) {
```
CodeyGuyDylan: Nitpick, we could rename `props` to `extraProps` or something, since `context` and `next` are technically props too

CodeyGuyDylan (dismissed):
I took this review mostly digesting and trying to understand how it all works. I think I have a decently solid grasp now and left a couple comments/questions

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
jboland88: Much more readable :) Thanks for this recommendation!

client/my-sites/checkout/get-thank-you-page-url/index.ts
jboland88: Aha - yes, there is a separate task to address the thank you page.
I had added the route handling here initially so that the checkout would go to a known place where the Akismet thank you page will be in the future. 
Not opposed to scratching the thank you page in this PR and handling that in the separate task. OR if we can get the unit tests satisfied for the blank route - either seems OK IMO. WDYT @CodeyGuyDylan ?

client/my-sites/checkout/controller.jsx
IanRamosC: `context` and `next` are related to the route management itself, while `props` is an object containing props that are going to be passed to the component.

client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
jboland88: At a high level, it has to do with assigning the purchased Jetpack license from the thank you page - I think when you are actually assigning the license to one of your Jetpack sites, the temporary site ID from the purchase is also needed for some of the steps in that process.

Since the Akismet purchases won't have license management/ assignment steps, we shouldn't need the temporary site ID.

client/my-sites/checkout/controller.jsx
jboland88: Aha, good idea!

client/my-sites/checkout/get-thank-you-page-url/index.ts
jboland88: Likely redundant here - we can handle validation like this within the cart itself :)

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -114,8 +116,10 @@ export default function CheckoutMain( {
 	// `getThankYouUrl`.
 	onAfterPaymentComplete?: () => void;
 	disabledThankYouPage?: boolean;
+	isAkismetSitelessCheckout?: boolean;
 	isJetpackCheckout?: boolean;
 	jetpackSiteSlug?: string;
+	akismetSiteSlug?: string;
```
IanRamosC, line 126: It seems like this component is very dependent on boolean values. Just throwing some ideas here, what do you think if we do a minor refactor and use an enum of "checkout types" instead of boolean values?
That way we can simplify the mapping of some rules, while keeping the code readable when we have to extend it.
The idea would be to have certain rules in arrays, objects, and functions, so we could either map a property in an object/function, or check if an array includes the `checkoutType` we currently have. 
I think that might not add much value right now, but will avoid duplicate lines of code whenever we have to add a new type of checkout - like what we are doing right now.

config/development.json
```diff
@@ -27,6 +27,7 @@
 	"zendesk_presales_chat_key": "216bf91d-f10f-4f66-bf65-a0cba220cd38",
 	"features": {
 		"ad-tracking": false,
+		"akismet/siteless-checkout": true,
```
IanRamosC, line 30: Thanks for adding this behind a feature flag.

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
sirbrillig: From a Payments perspective, I think that would be great! Teams have been adding these booleans for a long time and I'd love to have fewer of them.

I would suggest using a simple string union for the options (like we did with [contact details type](https://github.com/Automattic/wp-calypso/blob/136443bd1dfb5792563475a4a4981e0ffded1750/packages/wpcom-checkout/src/types.ts#L245)) instead of a TS enum because the latter is a bit of a hack (TS normally can be erased but enums are actually code and this may create problems if JS implements enums in the future).

client/my-sites/checkout/get-thank-you-page-url/index.ts
CodeyGuyDylan: It couldn't hurt to add the tests here, since it should be too much work to do so. I think i'll go ahead and do that

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
CodeyGuyDylan: I like this idea as well, I'll see what I can get done 😄

packages/composite-checkout/src/types.ts
```diff
@@ -124,7 +124,7 @@ export interface CheckoutProviderProps {
 	children: React.ReactNode;
 }
 
-export type CheckoutType = 'jetpack' | 'akismet' | undefined;
+export type SitelessCheckoutType = 'jetpack' | 'akismet' | undefined;
```
sirbrillig: 💯 This is a great type, thanks! However, the `@automattic/composite-checkout` package is a general purpose UI framework (see its [README](https://github.com/Automattic/wp-calypso/blob/136443bd1dfb5792563475a4a4981e0ffded1750/packages/composite-checkout/README.md#L17) for more details) and doesn't know anything about WPCom, Jetpack, etc. It just creates a step UI and provides callbacks for payment methods. 

I think `SitelessCheckoutType` might be a good type to put into either calypso itself (the `client` directory somewhere) or the `@automattic/wpcom-checkout` package which is a tool chest for use by calypso checkout.

CodeyGuyDylan: Thank you for the extra context, i'll move this to the `wpcom-checkout` package

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -89,8 +93,9 @@ export default function CheckoutMain( {
 	isInModal,
 	onAfterPaymentComplete,
 	disabledThankYouPage,
-	isJetpackCheckout = false,
+	sitelessCheckoutType,
 	jetpackSiteSlug,
+	akismetSiteSlug,
```
jboland88, line 97: `akismetSiteSlug` is OK to have for completeness, but I don't know that we'll need an `akismetSiteSlug` in the end unless we implement site-specific checkout for Akismet products.

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -129,13 +135,25 @@ export default function CheckoutMain( {
...
-		} ) || isJetpackCheckout;
+		} ) || sitelessCheckoutType === 'jetpack';
 	const isPrivate = useSelector( ( state ) => siteId && isPrivateSite( state, siteId ) ) || false;
+	const isSiteless = sitelessCheckoutType === 'jetpack' || sitelessCheckoutType === 'akismet';
 	const { stripe, stripeConfiguration, isStripeLoading, stripeLoadingError } = useStripe();
 	const createUserAndSiteBeforeTransaction =
-		Boolean( isLoggedOutCart || isNoSiteCart ) && ! isJetpackCheckout;
+		Boolean( isLoggedOutCart || isNoSiteCart ) &&
+		sitelessCheckoutType !== 'akismet' &&
+		sitelessCheckoutType !== 'jetpack';
```
jboland88: Could this use the new `isSiteless` vairable defined just above?

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -129,13 +135,25 @@ export default function CheckoutMain( {
...
 	const reduxDispatch = useDispatch();
-	const updatedSiteSlug = isJetpackCheckout ? jetpackSiteSlug : siteSlug;
+
+	const updatedSiteSlug = useCallback( () => {
```
jboland88: Would it make sense to use `useMemo` here so that we don't need to call `updatedSiteSlug` as a function later on?

client/my-sites/checkout/controller.jsx
```diff
@@ -178,7 +190,10 @@ export function checkout( context, next ) {
 				redirectTo={ context.query.redirect_to }
 				isLoggedOutCart={ isLoggedOutCart }
 				isNoSiteCart={ isNoSiteCart }
-				isJetpackCheckout={ isJetpackCheckout }
+				// TODO: in theory, isJetpackCheckout should always be false here if it is indicating whether this is a siteless Jetpack purchase
+				// However, in this case, it's indicating that this checkout is a logged-out site purchase for Jetpack.
+				// This is creating some mixed use cases for the isJetpackCheckout prop
+				sitelessCheckoutType={ isJetpackCheckout ? 'jetpack' : undefined }
```
jboland88: I think the naming of `sitelessCheckoutType` could be a bit more misleading than `isJetpackCheckout` in this controller context. Since when `isJetpackCheckout` is true that just means this is a userless Jetpack checkout for a specific site.

I don't think separating this use case is in the scope of this PR :) 
Could we update the explainer comment here to reference the new `sitelessCheckoutType ` prop instead of the `isJetpackCheckout` prop that has been removed?

jboland88 (commented):
I can't issue the final approval since I'm the original author, but this LGTM 👍 
I added a few small non-blocking comments.

The `SitelessCheckoutType` is a nice improvement!
I tested both a siteless Jetpack and Akismet purchase and both went as expected.

We'll want to monitor this when it deploys since we are touching the Jetpack checkout flow here as well.

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
CodeyGuyDylan: Good point, I didn't realize this wasn't really being used ATM

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
CodeyGuyDylan: Yes! I had added the `isSiteless` variable and removed it more that once, so I forgot about this 😆

client/my-sites/checkout/controller.jsx
CodeyGuyDylan: Yeah I'll go ahead and update that comment 😄

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
CodeyGuyDylan: Yes that is better

CodeyGuyDylan (approved):
Approving because @jboland88 can't

sirbrillig (approved):
I didn't fully test but from a code perspective this looks good to me!

IanRamosC (approved):
Thanks for the improvements.

PR #73622, opened 2023-02-21
chore: remove backstopjs from dependencies.

worldomonation:
Related to https://github.com/Automattic/wp-calypso/issues/73601.

## Proposed Changes

This PR removes backstopjs, workarounds, scripts and files associated with the visual regression tests.

Context: p1677001849898539/1676996972.514049-slack-C02DQP0FP

Key changes:
- remove entire directory under `test/visual`.
- remove dependency on `backstopjs` which in turn removes its dependency on `puppeteer:5.5.0`, which was causing errors on Apple Silicon.
- remove scripts associated with visual regression.
- update documents where required.

## Testing Instructions

Ensure the following build configurations are passing:
  - [x] Unit Tests
  - [x] Any E2E test
  - [x] Code Style
  - [x] Docker image

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #73808, opened 2023-02-27
Fix config of ICFY job

jsnajdr:
Trying to fix the config of the ICFY GitHub action which stopped running after #73622 was merged. I suspect that the empty `env:` field is a kind of YAML syntax error that prevents the action from running.


jsnajdr:
The ICFY job really runs successfully after the patch. Let's review and merge.

tyxla (approved):
Thanks 🚀

PR #73845, opened 2023-02-27
Marketing Connections: Add Mastodon

fmfernandes:
#### Proposed Changes

* This adds the option to connect to a Mastodon instance.

image

#### Testing Instructions

* This should be tested alongside D93584-code (which also needs to be reviewed).
* Go to `/marketing/connections/yoursite.com`.
* Try to connect to a Mastodon instance.

Related discussion and more detailed testing instructions: pewaj1-3x-p2

#### Pre-merge Checklist


mpkelly (commented):
I have reviewed this and it looks fine. I get that you followed the existing code style, but we try to write new code in TypeScript where possible. So ideally we'd have `mastodon.tsx` with some types in that file. I also think `isValidUsername` is worth unit testing.  

One issue I found is that if you enter an invalid username and then a valid one, the error message is not cleared and it doesn't allow you to connect until you blur the input. Here is a video:

https://user-images.githubusercontent.com/6851384/221825804-e4670afc-446c-4975-b9e0-228bec49615f.mp4


It would also be good to show some kind of feedback that the username validation was in progress i.e. a spinner at the right side of the `input`. 

The popup window opens, and I see the login form. I don't have an account, so I will leave that for others to test. Everything happens quite slowly - it took over a minute to see the login form - and I am not sure how much of that is down to being sandboxed. I applied D93584 but didn't review that code. 

Screenshot 2566-02-28 at 17 25 42

fmfernandes:
Hey @mpkelly, thanks for your review! I've changed the component to use TypeScript instead and added some types for its Props.

> I also think isValidUsername is worth unit testing.

I believe I should create a `test` folder in `client/my-sites/marketing/connections` and then create tests for the `Mastodon` component then, is that correct?

> One issue I found is that if you enter an invalid username and then a valid one, the error message is not cleared and it doesn't allow you to connect until you blur the input. Here is a video:

From the video I assume you pasted in the valid username? I couldn't get `onChange` to work when pasting at first but doing so with `useEffect`/`useCallback` seems to do trick, though I'm not sure if that's the best way of doing it.

> It would also be good to show some kind of feedback that the username validation was in progress i.e. a spinner at the right side of the input.

Yup! I added a little `<Spinner />` whenever the error is shown.

> Everything happens quite slowly - it took over a minute to see the login form - and I am not sure how much of that is down to being sandboxed. I applied D93584 but didn't review that code.

I did notice that too! I'm still working on D93584 so I'll investigate that further.

mpkelly:
Hi @fmfernandes. Nice job reworking this PR. Everything is much faster now for me. I also created an account so I can test e2e. Here is some more feedback:

1. You shouldn't need `useEffect` or `useCallback` in `mastodon.tsx`. The top part can be reworked to this:

```tsx
	const [ instance, setInstance ] = useState( '' );
	const [ error, setError ] = useState( '' );

	/**
	 * Checks if username matches a valid Mastodon username.
	 */
	const isValidUsername = ( username: string ) =>
		/@?\b([A-Z0-9_]+)@([A-Z0-9.-]+\.[A-Z]{2,})\b/gi.test( username );

	const handleInstanceChange = ( e: ChangeEvent< HTMLInputElement > ) => {
		const instance = e.target.value.trim();
		setInstance( instance );
		if ( isValidUsername( instance ) || ! instance ) {
			setError( '' );
		} else {
			setError( translate( 'This username is not valid.' ) );
		}
	};
```
I tried the above locally and everything seems to work still. 

2. I don't know the Mastodon username format well, but I think we are accepting invalid values here:

image

3. I can now connect my test account. It tells me that "Jetpack" wants permission to access my account - is this correct, or should it be WordPress.com?

Screenshot 2566-03-02 at 11 51 49

4. I hit "authorize" and I get an error: "Missing access token". I have sandboxed the API and applied `D93584`. I am running your Calypso branch locally. Do I need to do anything else to test this?

image

Please tag me in any replies, and I will try and retest and approve ASAP.

fmfernandes:
Hey @mpkelly, thanks for the review!

1. Wow, thanks, guess I overly complicated some stuff 😅
2. The expected format should be `@username@instance.tld`, though for the API what matters is the `instance.tld` part. I guess we can change the input to ask specifically only for `Enter your instance domain` and try to validate it as such but it looks like asking for a username for a Mastodon instance in this format is expected for an user.
3. There's some discussion around that on D93584-code. Since this will be used on Publicize and the Twitter connection is also using Jetpack that seemed a good fit.
4. Nope, that should be it, I couldn't replicate this but I updated the priority in which the function runs on D93584-code and tested with a new instance as well. Could you please test it again?

mpkelly:
Hi, @fmfernandes, React hooks can be quite complicated. When they first arrived, I used `useCallback` everywhere, but it's not really needed much in practice. 

The code looks good now. I did some more testing and also created a second account. Here's what I found:

1. I was able to connect my `@a8ctest` account today and there was no error with the access token. 👍 As I had already authorised the connection yesterday, the popup window opened and then closed automatically. I was then shown this message in Calypso:

image

I hit connect and then I see my connection is listed below:

Screenshot 2566-03-03 at 13 42 50

I then created a new account and tried to add that. 

2. Using my new account, the popup opened and closed automatically. For a new account, I would expect to see the Mastodon "authorize" page but I didn't. On returning to Calypso I saw this toast message only:

image

But the second account I added is not listed and doesn't appear to be connected. 

3. My new WP posts don't appear in Mastodon. I checked both of my accounts, and nothing appears. I made one post before the site was launched and another after launching, but I don't see anything [in Mastodon](https://mastodon.social/@a8ctest). 

4. I tested disconnecting/reconnecting and it appears to work ok. 

5. I added a second user to my site as an editor, and I was able to see the connection. However, as the editor, I was able to uncheck the "Connection available to all administrators, editors, and authors" checkbox which feels a bit wrong. Maybe this checkbox should only be shown if you established the connection. 

One other thing, with this checkbox unchecked, the editor can go to **Marketing** > **Connections** and click connect but when I do that I see this error in the popup:

image

Side note: I think we need a similar checkbox for GH integration. We currently share the connection with other admins but this might be appropriate - so thanks for the idea.

fmfernandes:
> Using my new account, the popup opened and closed automatically. For a new account, I would expect to see the Mastodon "authorize" page but I didn't. On returning to Calypso I saw this toast message only:

Woohoo! Thanks for that @mpkelly, I didn't catch this one but I fixed it. Added `force_login=False` as per the [documentation](https://docs.joinmastodon.org/methods/oauth/#query-parameters) so it's now possible to add multiple accounts from the same instance.

> My new WP posts don't appear in Mastodon. I checked both of my accounts, and nothing appears. I made one post before the site was launched and another after launching, but I don't see anything [in Mastodon](https://mastodon.social/@a8ctest).

Oh, it'll not work unless the async jobs are sandboxed as well. As of now, I'm manually testing publicize on the API console with route `POST /sites/$site/post/$post_id/publicize`

> I added a second user to my site as an editor, and I was able to see the connection. However, as the editor, I was able to uncheck the "Connection available to all administrators, editors, and authors" checkbox which feels a bit wrong. Maybe this checkbox should only be shown if you established the connection.

Should we create a separate issue for this? I can replicate that and even on a Twitter connection this is the default behavior so I assume there's some reasoning behind it but I agree that this should be only shown for the one responsible for the connection 👍🏻 

> One other thing, with this checkbox unchecked, the editor can go to Marketing > Connections and click connect but when I do that I see this error in the popup:

Could you explain the steps you took to get there? The only way I can see that happening is if you tested that while sandboxing just the API on WordPress.com where the instance input will not be present.

mpkelly:
>Woohoo! Thanks for that @mpkelly, I didn't catch this one but I fixed it. Added force_login=False as per the [documentation](https://docs.joinmastodon.org/methods/oauth/#query-parameters) so it's now possible to add multiple accounts from the same instance.

I tried with the two accounts I used previously. I disconnected both of them and then tried to reconnect, but I never the authorisation page in the popup, which automatically closes. For the GH integration, we added an action that listens for the keyring connection being deleted and then deletes the app installation at GitHub. I think something like that would be appropriate here, too. You can see what we did in `Keyring_Service_Github_Deploy#keyring_token_deleted`. 

Even with the above behaviour, I cannot get two connections working for the same user on the same site. The UI suggests this should be supported. I can connect one Mastodon account, but when I connect the second, it doesn't get added - the popup closes but the second account is not listed:

https://user-images.githubusercontent.com/6851384/224254970-1401fb0f-7d20-4f8c-bb9e-0a56731b1ee3.mp4

The only way I can get multiple connections working is to use two accounts - my admin and the editor I invited. Both of these can establish a single connection and then check the "Connection available to all administrators, editors, and authors" checkbox to share their connections:

image

So the API can return multiple connections, but it looks like it still needs work for a single user to have multiple connections (assuming that is supported). 

>Should we create a separate issue for this?

Ok, I will create one. 

>Could you explain the steps you took to get there? The only way I can see that happening is if you tested that while sandboxing just the API on WordPress.com where the instance input will not be present.

Somehow I had jumped from calypso.local:3000 to wordpress.com during testing. So you can ignore this one.

fmfernandes:
Hi @mpkelly!

> For the GH integration, we added an action that listens for the keyring connection being deleted and then deletes the app installation at GitHub.

So, in this case, you think the popup was instantly disappearing because the application was already authorized before? In this case, similarly to what's done in the GitHub connection, you mean I should also revoke the token? Something like [this](https://docs.joinmastodon.org/methods/oauth/#revoke)?

I couldn't replicate this (I'm able to connect multiple users on the same instance; tried `hachyderm.io` and `mastodon.social`); I can provide a video as well if it's helpful but doing the above should be enough and necessary after all.

---

Just a little update: I've implemented that function (`keyring_token_deleted`) so as long as the token is not being used in any other connections it should be revoked from the instance, thus, asking for authorization again (that plus the `force_login=False` should be enough to allow for multiple users of the same instance). Could you please test it again? 🙂

mpkelly:
Hi @fmfernandes, I revoked access to Mastodon for both my accounts in the Mastodon account settings and then tried to add two accounts for the same WP user to the same site. Here's a video:

https://user-images.githubusercontent.com/6851384/225205278-babbb68d-bc7e-467c-8042-82df7496fd79.mp4

What I realised was even though the second account is `@kikemelly@mastodon.social`, it still tries to authorise under the first account because I am still signed into Mastodon with that account. So ideally, there would be a way to force a login for the second account. It does work if I log out of the first account and then try to connect `@kikemelly@mastodon.social`. 

On authorising the second account, the Calypso dialog is a bit confusing. It asks me to choose an account, but I can't change anything as there is a single radio button that is already selected:

image

>Just a little update: I've implemented that function (keyring_token_deleted) ... Could you please test it again? 

This is working now - I disconnect and am shown the authorise page at Mastodon. 👍 

If you want to push ahead and get this on trunk then let me know. I think it's already good to go for the main use case of a single account.

fmfernandes:
Hi @mpkelly 👋🏻 

I was playing around with how other connections handle multiple accounts. Here's an example of me trying to connect 2 Twitter accounts through the popup:

https://user-images.githubusercontent.com/24195829/225808627-ab150c9c-f92d-4a7c-bdae-d3dda0d2deef.mp4

Steps:

- 00:02 - Authorize user1.
- 00:38 - Authorize user1 again. nothing happens.
- 00:48 - Logout from twitter in popup, another tab opens, log in with user2, after Twitter log in popup and authorized user2, nothing happened on Calypso.
- 01:35 - Try to connect another account again, this time the authorization popup opens with user2 logged in, which is successfully connected after selecting the account (which looks like the same popup on your screenshot).

And trying to connect two accounts while changing which one is logged in (and cancelling connection of the first one):

https://user-images.githubusercontent.com/24195829/225809720-43fc312b-0475-4b62-ad2b-cfa7edc3b26d.mp4

Steps:

- 00:00 - Twitter user1 is logged in, and authorized but this time, cancel the connection.
- 00:18 - Before clicking on connect, I logged in on Twitter with user2 and authorized.
- 00:36 - Choose which account to connect to.
- 01:04 - The same popup is shown.

--- 

For Mastodon it looks way more fluid the connection process as you can change accounts in the popup itself:

https://user-images.githubusercontent.com/24195829/225810336-23d8dd8e-1084-414f-8ad4-8f9a995c7e0b.mp4

---

> What I realised was even though the second account is `@kikemelly@mastodon.social`, it still tries to authorise under the first account because I am still signed into Mastodon with that account. So ideally, there would be a way to force a login for the second account. It does work if I log out of the first account and then try to connect `@kikemelly@mastodon.social`.

I don't think it's possible to force a log in with the second account, at least how it was before (`force_login=true`), Mastodon would just skip the popup and jump to the account logged in (which is the first one). I believe the intended way to connect multiple accounts is to change those in the authorization popup, or as you said, you should log out and log in to the desired account before hitting Connect. _Twitter looks a bit odd in that the flow jumps to a new tab instead of allowing switching accounts in the popup._

Do you think that asking for _Enter your Mastodon **username**_ and expecting `@username@example.com` is adding up to the confusion? Back when we decided to have a text input instead of a list of instances, we had an example of a Twitter crossposter (something like moa.party) which also asks for username in that format, but you can type anything like `@blabla@example.com` and it'll connect you in to `example.com`. So, we could change the text to read `Enter your Mastodon instance` and let the popup show the user if that makes it clearer.

I can change the wording if you think that would be better, otherwise I believe we can merge this.

pablinos:
> which also asks for username in that format, but you can type anything like @blabla@example.com and it'll connect you in to example.com

I quite like that. It seems like people should talk about Mastodon adresses as it's more like an email address. We could ask for the instance, but accept the full format and ignore the username if it's provided.

Connecting multiple accounts could do with some work. Services like Facebook can only have one connected page per user account, and the messaging between the pop-up windows and Calypso sometimes gets broken, as I think you experienced with logging into a second twitter account. My point being that if the UX of connecting two Mastodon accounts isn't perfect, that could well be not unique to Mastodon.

mpkelly:
> Do you think that asking for _Enter your Mastodon **username**_ and expecting `@username@example.com` is adding up to the confusion?

Yes, this part is misleading, IMO. It seems like removing the input and just adding a connect button would make the most sense. This way the user is still expecting to enter the account they want to connect. We'd also have less code to maintain and wouldn't need to validate the input. 

I am going to approve this now anyway. The above is just a suggestion.

mpkelly (approved):

fmfernandes:
> It seems like removing the input and just adding a connect button would make the most sense.

We need the input to know where to get the access token from. Because Mastodon is decentralized, there's no single way to connect to any given instance, i.e. we need to know the domain for which [/oauth/authorize](https://docs.joinmastodon.org/methods/oauth/#authorize) will be called.

> We could ask for the instance, but accept the full format and ignore the username if it's provided.

That's what's being done actually. On the API side, only the instance is considered. The input and validation present on the PR is just a way to garantee some form of `user@instance.tld`. I'll go ahead and merge this tomorrow morning.

If anything, the feature is available only on `development` so we can further enhance it later on if necessary.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7880254

Thank you @fmfernandes for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7880254).

client/my-sites/marketing/connections/mastodon.tsx
```diff
@@ -0,0 +1,144 @@
...
+		}
+	};
+
+	/**
+	 * Set the instance to the connect URL so that it can be used to connect to Mastodon.
+	 */
+	const setInstanceToConnectURL = () => {
+		const url = new URL( service.connect_URL );
+		url.searchParams.set( 'instance', instance );
+		service.connect_URL = url.toString();
```
manzoorwanijk, line 95: We should not be mutating props here. CC @Automattic/team-calypso

tyxla: Agreed - this is worth a follow-up.

PR #73850, opened 2023-02-27
WooExpress: Added confetti component

nelsonec87:

## Proposed Changes

* This PRs adds the `<Confetti />` component from `@automattic/onbarding` to the trial post-checkout page.
* If the component is not good enough, we can add the confetti from Figma as a new svg.

## Testing Instructions

* See the Confetti

Figma:
image

HTML: 
image

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
@@ -1,3 +1,4 @@
+import { Confetti } from '@automattic/onboarding';
 import { useTranslate } from 'i18n-calypso';
 import Main from 'calypso/components/main';
 import BodySectionCssClass from 'calypso/layout/body-section-css-class';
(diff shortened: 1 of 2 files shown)


markbiek:
I definitely see the confetti so that's working properly.

But I notice that there's significantly less confetti here than in the design. Does a @verofasulo need to weigh in on this too?

verofasulo:
Hey @markbiek :) 

> But I notice that there's significantly less confetti here than in the design

Yes - It would be great to make it a bit more celebratory :)

daledupreez (approved):
I think this is a good first step, even if it's not quite what we were aiming for.

It also looks like building dynamic confetti is going to be time-consuming, so is not something we can tackle for a while.

@verofasulo, are you OK with us shipping this limited confetti as an interim step? Or would you prefer no confetti? We'll need to add the more dynamic confetti later, so the question is all about your preference for the experience in the meantime. Can you take a look via [this link](https://github.com/Automattic/wp-calypso/pull/73850#issuecomment-1447133584) and navigating to `/plans/my-plan/trial-upgraded/<site-slug>` for a site you own.

verofasulo:
> It also looks like building dynamic confetti is going to be time-consuming, so is not something we can tackle for a while.

Thanks for expanding, @daledupreez! If we end up with static confetti, please just remove it. It doesn't convey any sense of celebration, in my opinion. 😊 Thank you!

nelsonec87:
> Thanks for expanding, @daledupreez! If we end up with static confetti, please just remove it. It doesn't convey any sense of celebration, in my opinion. 😊 Thank you!

Closing this PR since we won't go with this implementation.

PR #73853, opened 2023-02-27
Launchpad: Fix typescript and eslint notices

edanzer:
### Proposed Changes

This PR fixes some small TypeScript errors by slightly modifying how we define and return values from a function.

I also took the opportunity to fix an eslint warning about a missing dependency in launchpad's step-content.tsx file. 

### Testing

**Review Time: Short**
**Testing Time: Short**

1) Checkout this branch and run yarn and yarn start if needed. 
2) Navigate to the following line of code in your editor, and confirm there are no longer TS errors. 
3) Create a new free site or pull up the Launchpad screen for an existing free site, and confirm that the domain field populates correctly with *.wordpress.com domains. 
4) Create a new site with a custom domain, or pull up the launchpad screen for an existing site with custom domain, and confirm the domain field populates correctly with the custom domain.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
@@ -30,11 +30,11 @@ function getUrlInfo( url: string ) {
 	const urlWithoutProtocol = url.replace( /^https?:\/\//, '' );
 
 	// Ex. mytest.wordpress.com matches mytest
-	const siteName = urlWithoutProtocol.match( /^[^.]*/ );
+	const siteName = urlWithoutProtocol.match( /^[^.]*/ )?.[ 0 ] || '';
 	// Ex. mytest.wordpress.com matches .wordpress.com
-	const topLevelDomain = urlWithoutProtocol.match( /\..*/ ) || [];
+	const topLevelDomain = urlWithoutProtocol.match( /\..*/ )?.[ 0 ] || '';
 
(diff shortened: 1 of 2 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
```diff
@@ -30,11 +30,11 @@ function getUrlInfo( url: string ) {
 	const urlWithoutProtocol = url.replace( /^https?:\/\//, '' );
 
 	// Ex. mytest.wordpress.com matches mytest
```
edanzer, line 32: Worth noting that in my view, the types should have been correct before the change in this PR, but TypeScript seems to be having a hard time inferring the types. 

`const siteName = urlWithoutProtocol.match( /^[^.]*/ );` returns a type of `RegExpMatchArray`. Technically, IF that array exists and you grab the first item of that array `(`siteName[0]`)`, it should always be a string. If not defined, it should have been an empty string. But TS was still reading it as potentially undefined. 

sidebar-ts-error

edanzer: As a small side note, we use let to define some vars in this file, and then update them later based on sidebarDomain. The fact that we define and then update them made me think we should be using useState. But this is technically just derived state. The values only changes based on changes in props. Even so, the way we're doing it seems a bit unusual. Normally, for derived state, you might just define the values once with const.

jeyip (approved):

agrullon95 (approved):

PR #73709, opened 2023-02-23
Staging Sites: Add Staging Sites UI in Hosting Configuration

wojtekn:
Related to 1696-gh-Automattic/dotcom-forge

## Proposed Changes

In this PR, I propose to add an initial UI that allows adding a staging site.

## Testing Instructions

1. Open the Atomic site
2. Navigate to Hosting Configuration
3. Confirm that the "Staging site" card is displayed
4. Click "Create staging site" to add a new staging site
5. Confirm that when the site is created, the UI changes and allows navigating to staging site

Screenshot 2023-02-23 at 14 35 33

Screenshot 2023-02-23 at 14 34 41

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #73870, opened 2023-02-28
Staging Sites: Prevent added notice from appearing on mount

danielbachhuber:
I introduced a buggy bug in https://github.com/Automattic/wp-calypso/pull/73709

## Proposed Changes

Prevents the 'Staging site added.' notice from appearing when loading the Hosting Configuration page if the site already has a staging site.

## Testing Instructions

2. Navigate to 'Hosting Configuration'.
3. Click 'Add staging site'.
4. Verify the 'Staging site added.' notice appears after the staging site is created.
5. Refresh the page.
6. Verify no 'Staging site added.' notice appears.


wojtekn (approved):
It fixes the issue.

PR #73882, opened 2023-02-28
HelpCenter: Add HelpCenter to the new WPSupport page

renancarvalho:
## Proposed Changes

Adds Help Center to the new WPCOM support page. More [context here](https://github.com/Automattic/wpsupport3/issues/232)

## Testing Instructions
 
- Pull and run this branch
- Navigate to `apps/editing-toolkit` and run `yarn dev --sync` to sync your sandbox
- Make sure to sandbox wordpress.com/support
- Navigate to `https://wordpress.com/support`
- Check if the Help Center is rendered on the admin bar
- Help Center should work fully and normally 

image


apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
```diff
@@ -78,7 +79,13 @@ public static function is_next_steps_tutorial_enabled() {
 	 * Enqueue Help Center assets.
 	 */
 	public function enqueue_script() {
-		$script_dependencies = $this->asset_file['dependencies'];
+		// Remove wp-preferences from the dependencies array because it is not available on the support page.
+		$script_dependencies = array_filter(
```
renancarvalho, line 83: I noticed that `wp-preferences` was breaking the dependencies for the support site, I believe we actually don't need it, and this is how I am ignoring it. I am open to suggestions on how improve it :)

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
```diff
@@ -127,7 +134,17 @@ public function enqueue_script() {
 	 * Get current site details.
 	 */
 	public function get_current_site() {
-		$site     = \Jetpack_Options::get_option( 'id' );
+		$is_support_site = defined( 'WPCOM_SUPPORT_BLOG_IDS' ) && in_array( get_current_blog_id(), WPCOM_SUPPORT_BLOG_IDS );
+
+		if ( $is_support_site ) {
```
renancarvalho, line 139: In this context, renders the primary user's  on Help Center instead of the support site.
image

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
```diff
@@ -199,8 +222,7 @@ public function enqueue_wp_admin_scripts() {
 		global $wp_admin_bar, $current_screen;
 
 		$is_site_editor = ( function_exists( 'gutenberg_is_edit_site_page' ) && gutenberg_is_edit_site_page( $current_screen->id ) );
-
-		if ( ! is_admin() || ! is_object( $wp_admin_bar ) || $is_site_editor || $current_screen->is_block_editor ) {
+		if ( ! is_object( $wp_admin_bar ) || $is_site_editor || $current_screen->is_block_editor ) {
```
alshakero: To avoid issues, can we add `! is_user_logged_in()` here?

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
alshakero: Hi @noahtallen! I remember you fought this issue before with react-i18n. Do you have any idea to make `wp-preferences` available?

alshakero (changes requested):
Pretty neat! The Help Center is apparently adding a class called `.search` that is breaking the site. 

wordpress com_support_ (6)

renancarvalho:
> The Help Center is apparently adding a class called .search that is breaking the site.


I had this on my sandbox and forgot to push. I think it would make more sense to make the theme selector more strict to the theme. https://github.com/Automattic/wpsupport3/pull/325

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
renancarvalho: sure thing 👍

alshakero:
I think Help Center is the offender here. It's modifying everything with this generic classname. I'd fix the Help Center.

renancarvalho:
> I think Help Center is the offender here. It's modifying everything with this generic classname. I'd fix the Help Center.

Yes, this is true in a certain way, I checked how Help Center is implementing that and turns out that HC is using a [generic search component](https://github.com/Automattic/wp-calypso/blob/trunk/client/components/search/index.jsx), this component has a "generic" class `.search`. So, HelpCenter is not the one 'breaking' it.
I am wondering if we should improve that since it is a generic component, also I am a bit afraid of the focus of this PR if we change that. 
The Support theme issue was [fixed here](https://github.com/Automattic/wpsupport3/pull/325).

WDYT?
Thank you for your attention on this PR :)

alshakero (approved):
You're right, the search component is stylized in many places. 

Tested with a user who has no sites and all worked as expected! Got the forums option.

Please make sure to deploy the theme with the search fix before deploying this.

PR #73906, opened 2023-02-28
Launchpad: Add required Verify Email task

edanzer:
### Proposed Changes

Based on feedback and testing, this PR adds a Verify Email task to Launchpad. It works as follows. I think many of these points might still be worth discussing, so consider this PR a demo of how this functionality works based on existing decisions. We can test and decide if we're comfortable with it as is, or would like to make further changes.
  - The Verify Email task appears on ALL flows.
  - This task is required. If a user does not have a verified email, we disable other tasks until email is verified. 
  - Once email is verified, we no longer show the task.
  - Because verification is handled via the user's email, the task does not link anywhere. 
  
Because the email verification requires the user to leave our interface, there is a reasonable chance this change may reduce the Launchpad conversion rate (ie, the rate at which users launch their sites). That is also worth discussing. 

### Testing Instructions

**Testing Time: Medium**
**Review Time: Medium**

These testing instructions involve testing on Free Flow only. If we release, we will release to all flows and will need to test all flows. But given there may be discussion and changes, I would recommend testing one flow for now. 

1. You can test via Calypso localhost or via the Calypso Live link below.
2. Logout of WordPress.com or use a browser where you are not logged into WordPress.com.
3. Create a temporary email account at https://temp-mail.org
4. Create a new site at http://calypso.localhost:3000/setup/free/intro. If you are testing via live testing link go to [CALYPSOLIVELINK]/setup/free/intro. Use the email you created in step 3.
5. Continue through flow until Launchpad. Confirm the Verify Email (Check Your Inbox) task shows as active, but is not clickable and has no chevron/arrow. Confirm all other tasks are disabled. Overall, it should look like this: 
verify-email

6. Go to your temp email https://temp-mail.org, open the WordPress confirmation email, click the link to verify your email.
7. You should be automatically redirected to Launchpad. Confirm the Verify Email no longer shows. All other tasks should now show in their usual/normal state. Overall, it should look like this: 

verify-email-2

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/style.scss
@@ -351,6 +351,10 @@
 	pointer-events: none;
 }
 
+.launchpad__checklist-item[data-task="verify_email"] {
+	color: var(--color-text);
+}
+
 .launchpad__checklist-item-checkmark-container,
 .launchpad__checklist-item-warning-container {
 	margin-right: 8px;
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
@@ -324,6 +324,11 @@ export function getEnhancedTasks(
 						badgeText: isPaidPlan ? '' : translate( 'Upgrade plan' ),
 					};
 					break;
+				case 'verify_email':
+					taskData = {
+						title: translate( 'Verify Email (Check Your Inbox)' ),
+					};
+					break;
 			}
 			enhancedTaskList.push( { ...task, ...taskData } );
 		} );
@@ -356,5 +361,22 @@ export function getArrayOfFilteredTasks( tasks: Task[], flow: string | null ) {
 			} );
 			return accumulator;
 		}, [] as Task[] )
+	);
+}
+
+// If email is verified, filter out verify_email task
+// If email is not verified, disable all other tasks
+export function updateTasksBasedOnEmailVerification(
+	enhancedTasks: Task[] | null,
+	isEmailVerified: boolean
+) {
+	if ( ! enhancedTasks ) {
+		return enhancedTasks;
+	}
+	if ( isEmailVerified ) {
+		return enhancedTasks.filter( ( task ) => task.id !== 'verify_email' );
+	}
+	return enhancedTasks.map( ( task ) =>
+		task.id === 'verify_email' ? task : { ...task, disabled: true }
 	);
 }
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/tasks.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/tasks.tsx
@@ -116,23 +116,37 @@ export const tasks: Task[] = [
 		disabled: false,
 		taskType: 'blog',
 	},
+	{
+		id: 'verify_email',
+		completed: false,
+		disabled: true,
+		taskType: 'blog',
+	},
 ];
 
 const linkInBioTaskList = [
 	'design_selected',
 	'setup_link_in_bio',
 	'plan_selected',
+	'verify_email',
 	'links_added',
 	'link_in_bio_launched',
 ];
 
 export const launchpadFlowTasks: LaunchpadFlowTaskList = {
-	newsletter: [ 'setup_newsletter', 'plan_selected', 'subscribers_added', 'first_post_published' ],
+	newsletter: [
+		'setup_newsletter',
+		'plan_selected',
+		'subscribers_added',
+		'verify_email',
+		'first_post_published',
+	],
 	[ LINK_IN_BIO_FLOW ]: linkInBioTaskList,
 	[ LINK_IN_BIO_TLD_FLOW ]: linkInBioTaskList,
 	free: [
 		'setup_free',
 		'design_selected',
+		'verify_email',
 		'domain_upsell',
 		'first_post_published',
 		'design_edited',
@@ -141,6 +155,7 @@ export const launchpadFlowTasks: LaunchpadFlowTaskList = {
 	build: [
 		'setup_general',
 		'design_selected',
+		'verify_email',
 		'domain_upsell',
 		'first_post_published',
 		'design_edited',
@@ -149,10 +164,17 @@ export const launchpadFlowTasks: LaunchpadFlowTaskList = {
 	write: [
 		'setup_write',
 		'design_selected',
+		'verify_email',
 		'domain_upsell',
 		'first_post_published',
 		'site_launched',
 	],
-	videopress: [ 'videopress_setup', 'plan_selected', 'videopress_upload', 'videopress_launched' ],
+	videopress: [
+		'videopress_setup',
+		'plan_selected',
+		'verify_email',
+		'videopress_upload',
+		'videopress_launched',
+	],
 	sensei: [ 'sensei_setup', 'plan_selected', 'sensei_publish_first_course' ],
 };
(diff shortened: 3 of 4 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/style.scss
```diff
@@ -351,6 +351,10 @@
 	pointer-events: none;
 }
```
edanzer, line 353: This very specific CSS is needed because this task is effectively disabled/non-clickable, but we want it to show as enabled.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/style.scss
```diff
@@ -351,6 +351,10 @@
...
 
+.launchpad__checklist-item[data-task="verify_email"] {
+	color: var(--color-text);
+}
+
```
edanzer, line 357: This very specific CSS is needed because this task is effectively disabled/non-clickable, but we want it to show as enabled.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -324,6 +324,11 @@ export function getEnhancedTasks(
 						badgeText: isPaidPlan ? '' : translate( 'Upgrade plan' ),
 					};
 					break;
+				case 'verify_email':
+					taskData = {
+						title: translate( 'Verify Email (Check Your Inbox)' ),
```
edanzer, line 329: Would be great to get feedback from @nuriapenya about the text that should appear here and if we need anything more than this.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/tasks.tsx
```diff
@@ -149,10 +164,17 @@ export const launchpadFlowTasks: LaunchpadFlowTaskList = {
...
 	],
-	videopress: [ 'videopress_setup', 'plan_selected', 'videopress_upload', 'videopress_launched' ],
+	videopress: [
+		'videopress_setup',
+		'plan_selected',
+		'verify_email',
+		'videopress_upload',
+		'videopress_launched',
+	],
 	sensei: [ 'sensei_setup', 'plan_selected', 'sensei_publish_first_course' ],
```
edanzer, line 179: I did not add the task to Sensei flow. It is a complex flow that handles a lot of logic separately and differently from other Launchpad flows. If we want to add this, it should likely be done as a follow up, and with strong input from the team working on this flow.

jeyip:
Since this seems like it could use more design review, I'll test more closely when things have settled.

This PR mostly tests well for me, but I had some hiccups in my local dev environment. Heads up for others using Calypso live:

- Signing up for a new account is wonky in Chrome incognito mode. I'd recommend logging out of WordPress.com instead.
- Firefox doesn't seem to want to confirm the email with the provided link ( not a problem I imagine that's caused by this PR ). Maybe stick to Chrome or Edge for testing

## Testing
### Requirements

**Free Flow**
- [x] Confirm the Verify Email (Check Your Inbox) task shows as active, but is not clickable and has no chevron/arrow.
- [x] Confirm all other tasks are disabled
- [x] Confirm the Verify Email no longer shows. All other tasks should now show in their usual/normal state.
- [ ] Redirect to Launchpad after verifying email ( doesn't seem to work in Firefox, but I doubt it's caused by this PR )

### Browsers
- [x] Chrome
- [x] Firefox

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -358,3 +363,20 @@ export function getArrayOfFilteredTasks( tasks: Task[], flow: string | null ) {
 		}, [] as Task[] )
 	);
 }
+
+// If email is verified, filter out verify_email task
+// If email is not verified, disable all other tasks
+export function updateTasksBasedOnEmailVerification(
```
jeyip, line 369: `NB`: 

Do we need to expose this method for public use? To me, this could probably be folded into `getArrayOfFilteredTasks` or `getEnhancedTasks` so other consumers don't have to remember to import this method

jeyip, line 369: `NB`:

Can we write a unit test for this functionality? It should be relatively easy and it can be in a follow up

agrullon95:
### Free flow testing

- [ ] Confirm `Verify Email (Check Your Inbox)` appears on the Launchpad checklist for a user with an unverified email.
- [ ] Confirm `Verify Email (Check Your Inbox)` is active, not clickable, and does not have a chevron icon.
- [ ] Confirm all other tasks are disabled if email is not verified
- [ ] Confirm email verification email redirects me to launchpad, displays email verification notification, removes `Verify Email (Check Your Inbox)` task and enables other tasks.

agrullon95 (commented):
We can also probably add CSS to disable the badge hover styling when a task that contains a badge is disabled. Not a blocker.

### Free flow testing
- [x] Confirm `Verify Email (Check Your Inbox)` task appears on the Launchpad checklist for a user with an unverified email.
- [x] Confirm `Verify Email (Check Your Inbox)` task is active, not clickable, and does not have a chevron icon.
- [x] Confirm all other tasks are disabled if the user's email is not verified
- [x] Confirm email verification email redirects me to the Launchpad, displays email verification notification, removes `Verify Email (Check Your Inbox)` task, and enables other tasks.

nuriapenya:
This might be a bug: Users can still click the "Edit design" button on the previewer, leading to the task disappearing from the Launchpad. To reproduce:

* Click `Edit Design` button on the previewer.
* Open the Editor, change something and click save.
* "Great progress" modal -> Click Next Steps
* Back to Launchpad: The verify email task disappeared

The copy of the task should be in sentence case, like the other tasks: `Verify email (check your inbox)` 

I'm still not 100% sure about the best way to name this task. Maybe @michaelpick could give us some advice? 🙏🏽  We want to ensure users check their inboxes to verify their email addresses. It's a mandatory task.

sharpirate:
### Testing
Before confirming my email
✔️ Confirm the Verify Email (Check Your Inbox) task shows as active, but is not clickable and has no chevron/arrow. Confirm all other tasks are disabled.

After confirming my email
✔️ You should be automatically redirected to Launchpad. Confirm the Verify Email no longer shows. All other tasks should now show in their usual/normal state

### Notes
I was able to reproduce the bug Nuria mentioned

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
edanzer: Actually, I initially put this in getArrayOfFilteredTasks(). I pulled it out thinking more about smaller, single-purpose, composable functions. But if we'd rather keep the logic there, I'm ok with that too.

edanzer: Yup, I was assuming I'd do this as a follow up.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
nuriapenya: @edanzer Upon checking the actual email users receive, I would suggest changing to the following copy:

`Confirm email (check your inbox)`

michaelpick: @nuriapenya apologies for the delay in getting to this. I think your suggestion above makes perfect sense.

obenland:
> Based on feedback and testing, this PR adds a Verify Email task to Launchpad
Could you add references to that feedback or any other discussion around it for context? It helps our future selves with tracing back decisions trees if necessary

edanzer:
I'm closing this issue for now. 

We added the Verify Email task separately to the newsletter flow here: https://github.com/Automattic/wp-calypso/pull/74135

And there is still some discussion about if/how it should be added to other flows: p1677754610259639/1677618619.287359-slack-CHN6J22MP

PR #73912, opened 2023-03-01
Happy Blocks: fix missing index in build

arcangelini:
## Proposed Changes

This adds the index.php to the build for each block AND removes the `/build` from the path. It allows the build package to work as well as the src files when in dev mode.

## Testing Instructions

Pull branch and run `cd apps/happy-blocks && yarn build`
Check the build directory includes index.php AND that the path for the register block is correct.


alshakero (approved):
Works great!

PR #73913, opened 2023-03-01
Blaze: Only query for page and posts in the Jetpack App

enejb:
## Proposed Changes

* This PR removes the product query from the blaze advertising view when the view is embedded in the Jetpack mobile app because products currently are not supported by the Jetpack app. This also makes the page load faster. 

## Testing Instructions

In order to test out this PR you will need a site with products. (Woo Store products)
One way you can do this is be creating a jurassic.ninja site (connecting jetpack) and then installing WooCommerce and adding a product to the store. 
After you have a site with a couple of products. You need to do the following. 

1. Visit the calypso live link.
2. Update the User agent to `Mozilla/5.0 (iPhone; CPU iPhone OS 14_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Mobile/15E148 wp-android` to mimic the app.
3. Navigate to /advertising/example.com for a site that has products. Open up the network tab. Notice that there are no more products being requested. 
4. Revert to a standard user agent (refresh of a page is required) notice that the request for products still happens. 

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/promote-post/main.tsx
+++ b/client/my-sites/promote-post/main.tsx
@@ -10,6 +10,7 @@ import InlineSupportLink from 'calypso/components/inline-support-link';
 import Main from 'calypso/components/main';
 import useCampaignsQuery from 'calypso/data/promote-post/use-promote-post-campaigns-query';
 import PageViewTracker from 'calypso/lib/analytics/page-view-tracker';
+import { isWpMobileApp } from 'calypso/lib/mobile-app';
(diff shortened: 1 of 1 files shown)


zwarm (approved):
@enejb - I was able to test this using my woo commerce site. First run without custom user agent shows products, second run with `wp-android` in the user agent does NOT show products. 👏

PR #73920, opened 2023-03-01
GitHub Deploy: show a message when no branches are returned; auto refresh on refocus

mpkelly:
## Proposed Changes

* Now shows a message "No branches found" if there are no branches for repo. 
* If you refocus the window, the branches are auto-refreshed

## Testing Instructions

* Create a new repo at GH and confirm the message "No branches found" is shown
* Add a branch and return to the window and confirm the new branch is shown

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/hosting/github/github-connect-card/search-branches.tsx
+++ b/client/my-sites/hosting/github/github-connect-card/search-branches.tsx
@@ -22,20 +22,31 @@ export const SearchBranches = ( {
 }: SearchBranchesProps ) => {
 	const { __ } = useI18n();
 
-	const { data: branches, isLoading: isLoading } = useGithubBranchesQuery(
-		siteId,
-		repoName,
-		connectionId,
-		{
-			onSuccess( branches ) {
-				if ( branches.length < 30 ) {
-					onSelect( branches[ 0 ] );
-				}
-			},
-		}
-	);
+	const { data: branches, isFetching } = useGithubBranchesQuery( siteId, repoName, connectionId, {
+		onSuccess( branches ) {
+			if ( branches.length < 30 ) {
+				onSelect( branches[ 0 ] );
+			}
+		},
+	} );
 
-	if ( branches && branches.length < 30 ) {
+	if ( ! repoName || isFetching ) {
+		return (
+			<div style={ { position: 'relative' } }>
+				<FormTextInput
+					id="branch"
+					disabled={ ! repoName || isFetching }
+					className="connect-branch__repository-field"
+					placeholder={ branches ? __( 'Type a branch' ) : __( 'Choose a branch' ) }
+					onChange={ ( e: ChangeEvent< HTMLInputElement > ) => onSelect( e.target.value ) }
+					value={ selectedBranch }
+				/>
+				{ isFetching && <Spinner className="connect-branch__loading" /> }
+			</div>
+		);
+	}
+
+	if ( branches && branches.length > 0 && branches.length < 30 ) {
 		return (
 			<FormSelect
 				id="branch"
@@ -54,16 +65,12 @@ export const SearchBranches = ( {
 	}
 
 	return (
-		<div style={ { position: 'relative' } }>
-			<FormTextInput
-				id="branch"
-				disabled={ ! repoName || isLoading }
-				className="connect-branch__repository-field"
-				placeholder={ branches ? __( 'Type a branch' ) : __( 'Choose a branch' ) }
-				onChange={ ( e: ChangeEvent< HTMLInputElement > ) => onSelect( e.target.value ) }
-				value={ selectedBranch }
-			/>
-			{ isLoading && <Spinner className="connect-branch__loading" /> }
-		</div>
+		<FormTextInput
+			id="branch"
+			disabled
+			className="connect-branch__repository-field"
+			placeholder={ __( 'No branches found' ) }
+			value=""
+		/>
 	);
 };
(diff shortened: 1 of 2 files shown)


client/my-sites/hosting/github/github-connect-card/search-branches.tsx
```diff
@@ -22,20 +22,31 @@ export const SearchBranches = ( {
...
+					placeholder={ branches ? __( 'Type a branch' ) : __( 'Choose a branch' ) }
+					onChange={ ( e: ChangeEvent< HTMLInputElement > ) => onSelect( e.target.value ) }
+					value={ selectedBranch }
+				/>
+				{ isFetching && <Spinner className="connect-branch__loading" /> }
+			</div>
+		);
+	}
+
+	if ( branches && branches.length > 0 && branches.length < 30 ) {
```
mpkelly: Not related directly to this PR: are we not supporting more than 30 branches, then? 😄

zaguiini: Showing more than 30 options on a select input would not be a very good experience 😄 

The branches are sorted by updated date and the default branch is always on top. If there are more than 30 branches we're forcing the user to type its name because GH API has no search functionality for branches, unfortunately.

client/my-sites/hosting/github/github-connect-card/use-github-branches-query.ts
```diff
@@ -32,7 +32,6 @@ export const useGithubBranchesQuery = (
 			meta: {
 				persist: false,
 			},
-			refetchOnWindowFocus: false,
```
zaguiini, line 35: The reason we didn't refetch on focus is that if the user switches tabs, they'll lose their current selection 😬

Try it:

1. select a branch that's not the default
2. switch tabs
3. check that the selection is gone and you're back to the default branch

zaguiini (commented):
I have forked `gutenberg` in the past. Using this patch, whenever I select it the repo, I see "No branches found"

We should:

1. show a free text input if there are more than 30 branches
2. show a select if there is more than one but less than 30
3. show the message if there are no branches

Also, I think we should make it more explicit that there is an invalid state. Maybe adding an exclamation as the icon on the right to the input? 🤔 Maybe not using a placeholder but explicitly adding the `<FormError />` element below the branch input? There are several options here. @javierarce might help.

client/my-sites/hosting/github/github-connect-card/search-branches.tsx
vykes-mac: @mpkelly could move that 30 into a constant in this PR.

client/my-sites/hosting/github/github-connect-card/use-github-branches-query.ts
p-jackson: I'm the changing the logic which auto selects the branch to check for an existing selection.

p-jackson:
> The reason we didn't refetch on focus is that if the user switches tabs, they'll lose their current selection 😬

I've rearranged how the initial selection is made so that changes to the ReactQuery data store don't override selections. It also means that if a branch gets deleted then after refocusing the tab we unselect that branch.

I've also refactored a little to handle the edge cases up front and exit early. It means there's a bit more duplication but I think it's easier to reason about.

> GH API has no search functionality for branches, unfortunately

I wonder if _this_ could be fixed by using the GraphQL API 🤔

p-jackson:
> I wonder if _this_ could be fixed by using the GraphQL API 🤔

Woot! This GraphQL query returns the first 10 branches in the Calypso repo that match the `foo` query string.

```
{
  repository(name: "wp-calypso", owner: "Automattic") {
    refs(refPrefix: "refs/heads/", first: 10, query: "foo") {
      edges {
        node {
          name
        }
      }
    }
  }
}
```

The `refs/heads/` bit differentiates a branch from a tag I believe.

zaguiini:
> > I wonder if _this_ could be fixed by using the GraphQL API 🤔
> 
> Woot! This GraphQL query returns the first 10 branches in the Calypso repo that match the `foo` query string.
> 
> ```
> {
>   repository(name: "wp-calypso", owner: "Automattic") {
>     refs(refPrefix: "refs/heads/", first: 10, query: "foo") {
>       edges {
>         node {
>           name
>         }
>       }
>     }
>   }
> }
> ```
> 
> The `refs/heads/` bit differentiates a branch from a tag I believe.

Great stuff!

client/my-sites/hosting/github/github-connect-card/use-github-branches-query.ts
javierarce: I like @zaguiini's idea of changing the type of input based on the number of items.

Regarding the error, I'm in favor of showing the error explicitly, too (especially, since our disabled states for input fields are not accessible). It would be nice to find a less redundant message than the one below, though. 

image

p-jackson: > It would be nice to find a less redundant message than the one below, though.

I think we can just leave the text field blank in this case.

A (!) red error is a bit more than I'd expect in this case. There's been no user error they should fix up, they haven't typed anything wrong. It feels more like an informational thing.

But I am definitely in favour of showing a more accessible message!

p-jackson:
I'm going to merge this change as a step in the right direction (we're now showing an explicit "no branches" message and are refetching when a branch is added), but it's also clear that there are further improvements to be made. [Like the improvement to the error message](https://github.com/Automattic/wp-calypso/pull/73920#discussion_r1127807088) and the ability to search for branches.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7847023

Hi @mpkelly, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include this string: `No branches found`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7847023).

PR #73944, opened 2023-03-01
Pattern assembler: Refactor submit logic into separate action

mmtr:
Part of https://github.com/Automattic/dotcom-forge/issues/1912

## Proposed Changes

Moves the logic that applies a design created in the pattern assembler to a separate action so it can be re-used when selecting a pattern-based theme (this will be required by virtual themes i2, as part of https://github.com/Automattic/dotcom-forge/issues/1912).

## Testing Instructions

- Go to `/start`
- Create a new site with a Free plan
- Select the "Write & publish" goal
- Proceed in the boarding until you see an option to "View designs"
- Click on it to access the Design picker
- Scroll to the bottom and click on the Pattern Assembler CTA
- Create your own design
- Once you're done, click on the Continue button
- Make sure the design has been applied successfully (it should match the same behavior as production)


packages/data-stores/src/site/constants.ts
```diff
@@ -1 +1,3 @@
 export const STORE_KEY = 'automattic/site';
+
+export const PLACEHOLDER_SITE_ID = 211865921; // blankcanvas3demo.wordpress.com
```
dsas: This constant is now duplicated in `client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/constants.ts` I think that copy is still used in `client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-assembler-container.tsx` but it would be nice to remove this duplication if possible

dsas (approved):
WFM, just one question about some two copies of a constant.

mmtr:
Looks like we finally don't need this, so I'm closing it in favor of D103329-code and https://github.com/Automattic/wp-calypso/pull/73994.

mmtr:
Reopening, since we noted that our current approach (that relies on how the backend handle recipes) modifies the Index template as opposed to creating a homepage template. See https://github.com/Automattic/dotcom-forge/issues/1912

dsas (approved):
Still works great for me

PR #73961, opened 2023-03-01
Add/calypso user resurrected event

ayebare:
Related to # This is from pau2Xa-4nn-p2

## Proposed Changes

Add New ‘Resurrection Event’ Request to Track Resurrected Users

## Testing Instructions

Setup Tracks Vigilante to log events.

The testing will be done in two parts.

### Part 1: Test that the code works.
(This part will need a local _setup.)_

1. Set up calypso locally.
2. switch to [add/calypso_user_resurrected_event](https://github.com/Automattic/wp-calypso/tree/add/calypso_user_resurrected_event) branch
3. in  [/client/lib/analytics/track-resurrections/index.jsx.](https://github.com/Automattic/wp-calypso/blob/add/calypso_user_resurrected_event/client/lib/analytics/track-resurrections/index.jsx#L17) change this line of code 
`return lastSeen < oneYearAgoInSeconds;`
to
`return lastSeen < Math.floor( ( Date.now() - 5 * 60 * 1000 ) / 1000 ); // return true if user was last seen 5mins ego` 

4. The change above will ensure that we don't have to wait for one year but 5 minutes to see a resurrection event.
5. Log out of Calypso.
6. Wait for 5 minutes or more.
7. log back in then immediately and visit these landing pages where we re-direct logged-in users.

- /home/${ primarySiteSlug }`;
- /read
- /sites 
- /stats/${ primarySiteSlug }

8. You should see a resurrection event logged in Vigilante. (Since local calypso setups don't automatically redirect to the paths above on login, ensure the landing pages above are visited immediately  as settings API calls will override the last_admin_activity_timestamp value after a short while; 

### Part 2: Test that the code will fire when a user logs in after a year.

1. Manually set the "last_admin_activity_timestamp"  field in WP me/settings API to '1647340800', a timestamp from a year ago, in seconds. 

see 
fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qzr%2Qfrggvatf%2Qi1%2Q1%2Qraqcbvag.cuc%3Se%3Qo5o2n480%231133-og
and 
p1681850208626219/1681844780.499219-slack-C03N25JPCE4  

2. Visit the following URLS while checking Vigilante logs and verify that the "calypso_user_resurrected" event is fired with the last seen timestamp from above.

https://container-lucid-hypatia.calypso.live/sites
https://container-lucid-hypatia.calypso.live/read
https://container-lucid-hypatia.calypso.live/home/{$slug}
https://container-lucid-hypatia.calypso.live/stats/day/{$slug}

3. Repeat with a timestamp **not** older than a year and verify that no resurrection event is fired.

Once this PR is approved, I will write unit tests.

*

## Pre-merge Checklist


client/boot/common.js
```diff
@@ -242,6 +252,16 @@ const configureReduxStore = ( currentUser, reduxStore ) => {
 			// If the browser is capable, registers a service worker & exposes the API
 			reduxStore.dispatch( pushNotificationsInit() );
 		}
+
+		currentUser.lastSeenDayInterval = 368; // @todo set the lastSeenDayInterval in User meta API in WP.
```
tyxla: I think we'll have to update that before shipping the PR.

client/boot/common.js
```diff
@@ -242,6 +252,16 @@ const configureReduxStore = ( currentUser, reduxStore ) => {
...
+
+		if ( isResurrectedUser( currentUser ) ) {
```
tyxla: Is this the best place to do this? 

Can we perform this later, and not at boot time? Ideally, we could add this into a component (similar to other `Query` components) and `AsyncLoad` it so it doesn't increase bundle size, since it's not critical functionality.

client/boot/common.js
```diff
@@ -220,6 +221,15 @@ const utils = () => {
 	Modal.setAppElement( document.getElementById( 'wpcom' ) );
 };
 
+const isLoginDateDiffAboveThreshold = ( currentUser, daysInterval ) => {
+	return currentUser.lastSeenDayInterval > daysInterval;
```
tyxla: One-liner functions are often redundant and can be inlined. I think that's the case here.

client/boot/common.js
```diff
@@ -220,6 +221,15 @@ const utils = () => {
...
+};
+
+const isResurrectedUser = ( currentUser ) => {
+	// Last login date diff > 1 year
+	return isLoginDateDiffAboveThreshold( currentUser, 365 );
```
tyxla: Same suggestion regarding the one-liner function.

client/boot/common.js
ayebare: Updated. Moved logic to separate component.

ayebare:
@tyxla  Thanks for helping me review this.
This is a work in progress. Following your recommendations, I have made a couple of adjustments.

Please take a look and let me know if there are things you would like me to change before I go further. I am almost at the end, anyway :) I want to start writing unit tests once I know this is in order.
Also, I need to figure out a way to test the code thoroughly. The current code uses new fields from the API  that need sandboxing to test.

cc @gmovr

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+		const previousLoginTime = previousLoginTimeArr.length
+			? Math.max( ...previousLoginTimeArr )
+			: loginTime;
+
+		if ( ! isResurrected( loginTime, previousLoginTime, yearInterval ) ) {
+			return;
+		}
+
+		// Only fire event if the difference between resurrection login time and current time is less 5 minutes.
+		const differenceInDays = ( Date.now() - loginTime ) / ( 1000 * 60 * 5 );
```
ayebare: @tyxla @gmovr  To ensure that the resurrection event doesn't fire indefinitely, I am allowing 
5 minutes in which the resurrection event can be fired after the user logs in.
- Question, would it be better to use local storage as a way of preventing events firing indefinitely i.e Once the event fires, we store a value locally to prevent further resurrection events.  Then increase this time gap in code from one 5 minutes to an hour?

client/lib/analytics/track-resurrections/README.md
```diff
@@ -0,0 +1,23 @@
+# Track User Resurrections
+
+`<TrackResurrections />` is a React component used to help track when a user has been resurrected.
+A resurrected user is a user who hasn't been active on the site for a very long period of time.
+
+For this particular implementation, we consider the time difference between the current and the previous login times as the yardstick to measure periods of inactivity.
+
+The time frame between logins for one to be considered resurrected defaults to One year. However this can be changed in the component properties.
```
tyxla:
```suggestion
The time frame between logins for one to be considered resurrected defaults to One year. However, this can be changed in the component props.
```

client/lib/analytics/track-resurrections/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+## Usage
+
+Render the component passing one or both of:
+
+- Interval: An `yearInterval` integer e.g 2 defaults to 1
```
tyxla: This part is a bit unclear, I'm not sure what it means by:

* Passing one or both of:
* An `yearInterval` integer e.g 2 defaults to 1

client/lib/analytics/track-resurrections/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+```jsx
+<ContainerComponent>
+	<TrackResurrections  yearInterval={ yearInterval } />
```
tyxla:
```suggestion
	<TrackResurrections yearInterval={ yearInterval } />
```

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
+import { useEffect } from '@wordpress/element';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import getUserSetting from 'calypso/state/selectors/get-user-setting';
+
+const TrackResurrections = ( { login_time, yearInterval, isFetchingUserSettings } ) => {
```
tyxla: Do we really need `yearInterval` to be a separate prop? I think it could easily be a constant somewhere. I doubt we'd want to use this with two or more different values, and if we do, we can always add such a prop. 

See the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
+import { useEffect } from '@wordpress/element';
```
tyxla: We could use that directly from `react`.

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import getUserSetting from 'calypso/state/selectors/get-user-setting';
+
+const TrackResurrections = ( { login_time, yearInterval, isFetchingUserSettings } ) => {
+	const isResurrected = ( loginTime, previousLoginTime ) => {
+		const differenceInMilliseconds = loginTime - previousLoginTime;
+		const differenceInYears = differenceInMilliseconds / ( 1000 * 60 * 60 * 24 * 365 );
```
tyxla: The calculation is ideal for a constant. We usually do them like this:

```
const YEAR_IN_MILLISECONDS = 1000 * 60 * 60 * 24 * 365;
```

and then we use it in the code.

It could greatly improve the readability of the code.

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+		return differenceInYears > yearInterval;
+	};
+
+	useEffect( () => {
+		if ( isFetchingUserSettings ) {
+			return null;
```
tyxla: Is there a need to return `null` here? Can we just `return` void?

client/lib/analytics/track-resurrections/index.jsx
tyxla: We likely should rename the constant, since it's not really "days".

Also, I don't understand the need for this additional check.  Given that we immediately update the login time of the user, and we make sure that it runs only once, why do we need that extra check?

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+		// Only fire event if the difference between resurrection login time and current time is less 5 minutes.
+		const differenceInDays = ( Date.now() - loginTime ) / ( 1000 * 60 * 5 );
+
+		if ( differenceInDays <= 1 ) {
+			recordTracksEvent( 'calypso_user_resurrected', {
+				loginTime,
+				previousLoginTime,
+			} );
+		}
+	}, [] ); // We want this to only fire once.
```
tyxla: Hm, I see it will run only once, but if it runs and then the user is still fetching, it will never re-run after the user has re-fetched. Thoughts?

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+	return null;
+};
+
+TrackResurrections.propTypes = {
+	yearInterval: PropTypes.number,
+	login_time: PropTypes.shape( {
```
tyxla: Feel free to ignore connected props here.

And if we end up removing `yearInterval` as I suggested above, we might remove the entire `propTypes` and `defaultProps` objects.

Have you considered using TypeScript? It would reduce the need for `propTypes` completely.

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+	yearInterval: 1,
+	login_time: { login_time: [], previous_login_time: [] },
+};
+
+const mapStateToProps = ( state ) => {
+	const login_time = getUserSetting( state, 'login_time' );
+
+	return {
+		login_time,
+		isFetchingUserSettings: state.userSettings.fetching,
```
tyxla: We already have a `isFetchingUserSettings()` selector that we should be using here.

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+	};
+};
+
+export default connect( mapStateToProps )( TrackResurrections );
```
tyxla: Given that this is a functional component, I'd suggest using the `useSelector()` hook instead of connecting the component.

client/my-sites/customer-home/main.jsx
```diff
@@ -96,14 +97,14 @@ const Home = ( {
 
 	const header = (
 		<div className="customer-home__heading">
+			<TrackResurrections />;
```
tyxla: Is this the place where we want to put this? Jetpack sites don't have a "My Home" page, so this just won't run for users whose primary (or only) site is a Jetpack site.

tyxla: I'd suggest async-loading this. It's not really essential/critical functionality and can definitely be delayed and loaded later.

tyxla (commented):
Hey, @ayebare thanks for working on this!

It would be great to see some unit tests, but I wouldn't call them necessary for a v1. 

I've left some more suggestions. Let me know what you think.

client/lib/analytics/track-resurrections/index.jsx
ayebare: Good catch, thanks updating ....

client/lib/analytics/track-resurrections/index.jsx
ayebare: @tyxla  void doesn't seem to be a valid child and throws an error
https://legacy.reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

client/my-sites/customer-home/main.jsx
ayebare: Noted thanks. Updating ...

ayebare: Going to test against Jetpack and relocate appropriately.

client/lib/analytics/track-resurrections/index.jsx
tyxla: Okay, two things:

1. Sorry for the confusion, I didn't mean literally `void` but rather an empty `return`:

```
if ( isFetchingUserSettings ) {
	return;
}
```

2. The link you sent applies to what we return when rendering, while the context we're talking about here is within a React `useEffect()` hook.

client/lib/analytics/track-resurrections/index.jsx
ayebare: This has been removed as its no longer necessary.

client/lib/analytics/track-resurrections/README.md
ayebare: The component doesn't have any more properties. This was fixed.

client/lib/analytics/track-resurrections/index.jsx
ayebare: Year interval was removed in the latest implementation.

client/lib/analytics/track-resurrections/index.jsx
ayebare: This has been updated to reflect the requested change.

client/lib/analytics/track-resurrections/index.jsx
ayebare: We are now using isFetchingUserSettings select as requested.

client/my-sites/customer-home/main.jsx
ayebare: This is now being loaded using async-loading

client/lib/analytics/track-resurrections/README.md
ayebare: The component doesn't have a yearInterval prop any more.

client/lib/analytics/track-resurrections/index.jsx
ayebare: The updated component doesn't have any props anymore.

client/lib/analytics/track-resurrections/index.jsx
ayebare: Updated to use the recommended useSelectore as this is a functional component.

client/lib/analytics/track-resurrections/README.md
ayebare: The yearInterval prop was removed in the changes and changes reflected in the README.md

client/my-sites/customer-home/main.jsx
ayebare: So I found the three other landing pages from the function getLoggedInLandingPage i.e
`/home/${ primarySiteSlug }`;
`/stats/${ primarySiteSlug }`;
and '/read';
And I would like to add the resurrection component to each of them.
However, I need guidance on the recommended way to include AsyncLoad to the files outside the client folder, e.g., in the HighlightCards component, as it's in the packages library outside the client library. This component is displayed on the landing page of jetpack sites
https://github.com/Automattic/wp-calypso/blob/2362c74ba7803cd595339cb16c80b2fbb6d06331/packages/components/src/highlight-cards/index.tsx

client/lib/analytics/track-resurrections/index.jsx
ayebare: Thanks, this was fixed.

client/my-sites/customer-home/main.jsx
tyxla: I wouldn't recommend putting `<TrackResurrections />` inside a generic component that can be used anywhere.

Instead, I'd recommend putting it alongside the `<HighlightCards />` instances, even if we have to insert it in multiple places.

client/lib/analytics/track-resurrections/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+	// Calculate the timestamp for one year ago in seconds
+	const oneYearAgoInSeconds = nowInSeconds - YEAR_IN_SECONDS;
+
+	// Compare the given timestamp with one year ago
+	return lastSeen < oneYearAgoInSeconds;
+};
+
+const TrackResurrections = () => {
+	const userSettings = useSelector( getUserSettings ) ?? {};
+	const lastSeen = userSettings?.last_admin_activity_timestamp || Date.now();
```
claudiucelfilip: Nitpick: do we need both `useSelector( getUserSettings ) ?? {}` and `userSettings?.`?

claudiucelfilip (approved):
I followed the steps and it works as described.
However, I don't think I have enough experience with these locations (ex. what runs in Jetpack), and might be worth getting a second opinion.

tyxla (approved):
Nothing further to add from my perspective.

Nice work @ayebare 👍 🚀

tyxla:
@ayebare actually the e2e test failures seem to persist. They seem unrelated, so perhaps a rebase should fix them?

Issue #73869, opened 2023-02-28
LaunchPad - domain upgrade flow.


Addison-Stavlo:
As discussed in team meeting yesterday - paYE8P-2zy-p2 - we want to make some updates to the domain upsell flow for when we are using it with the launchpad:

When the launchpad is active:
- [ ] Make initial `< Home` button lead to launchpad (next steps) instead of my home.
- [ ] Replace the final "All Ready to Go" pages "What's Next" section's content with info about / link to LaunchPad - we can cc @nuriapenya on the PR to help finalize descrption/copy.

Addison-Stavlo:
close this with the PRs linked above.

PR #73964, opened 2023-03-01
Launchpad: Change Back Button Text & URL for Domain Flow

sharpirate:
Related to https://github.com/Automattic/wp-calypso/issues/73869

### Time Estimate
Test: Short
Review: Short

### Proposed Changes
 Change the Back Button text to "**Next Steps**" and redirect back to Launchpad (when Launchpad is enabled) on the first step of the domain flow

### Testing
1. Checkout this branch
2. Create a new Launchpad site
3. Once you arrive at Launchpad, click the **Choose a domain** task
4. Confirm the back button in the top-left corner says "**Next Steps**" and redirects you back to Launchpad

https://user-images.githubusercontent.com/20927667/222249701-3e0ef8dc-21b4-4e51-8f1e-95185f8aaae9.mov

### DIFF
--- a/client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
+++ b/client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
@@ -1,6 +1,5 @@
 import { Button, Gridicon } from '@automattic/components';
 import { useTranslate } from 'i18n-calypso';
-import page from 'page';
 
 import './style.scss';
 
@@ -9,7 +8,7 @@ export default function DomainAndPlanPackageNavigation( props ) {
 
 	const goBack = () => {
 		if ( props.goBackLink ) {
-			page( props.goBackLink );
+			window.location.assign( props.goBackLink );
 		} else {
 			window.history.go( -1 );
 		}
--- a/client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
+++ b/client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
@@ -21,16 +21,15 @@ export default function DomainAndPlanPackageNavigation( props ) {
 		args: { currentStep: step, stepCount: 3 },
 	} );
 
+	const buttonText =
+		props.goBackText || ( props.step !== 1 ? translate( 'Back' ) : translate( 'Home' ) );
+
 	return (
 		<div className="domain-and-plan-package-navigation">
 			<div className="domain-and-plan-package-navigation__back">
 				<Button borderless="true" onClick={ goBack }>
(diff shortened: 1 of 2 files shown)


sharpirate:
### Notes
There's a weird bug right now where the domain screen flashes for a moment while the user is being redirected to Launchpad. I don't know why this is happening because the redirect is done by modifying the `href` prop of the Back button directly so technically this shouldn't be happening..

agrullon95 (commented):
You can replace `page()` with `window.location.assign()`. That should fix the redirection flicker issue. It may be flickering because of how page.js redirection is setup.

**Tests**

- [x] Verify the back button in the top-left corner of the domains page says "Next Steps" and redirects you back to Launchpad

agrullon95 (approved):

client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
```diff
@@ -9,7 +8,7 @@ export default function DomainAndPlanPackageNavigation( props ) {
 
 	const goBack = () => {
 		if ( props.goBackLink ) {
-			page( props.goBackLink );
```
edanzer, line 12: It looks like window.location.assign works here. But I do think page() and window.location.assign() are directly equal. The page() method seems to be an abstraction with some extra routing logic. I think it could theoretically accept arguments/values that would not work with window.location.assign(). I did a quick search to see what kinds of other values are passed as props.goBackLink. From what I can tell, they seem like they should all still work fine here. But still wanted to note this.

sharpirate: Another thing I can do is I can pass down a `isLaunchpad` prop and conditionally decide whether to use `page` or `location.assign` based on it

edanzer: As a side note - using TypeScript in this file and destructing 'props' might help provide clarity on what values and types are expected for goBackLink. It would also be more clear that goBackText is (I think) a new prop that's being introduced to this component in this PR. Obviously, this goes beyond the current PR though.

edanzer (approved):
This tests as expected. I added a small note about page() vs window.location.assign(), but I think that swap is fine here given how page() is being used.

client/my-sites/domains/components/domain-and-plan-package/navigation/index.jsx
edanzer: @Sharpirate - I think it's fine. I probably raised it just in case someone else wanted to confirm it's fine. But I think this seems fine to me here.

PR #73967, opened 2023-03-01
Launchpad: Fix URL state for Celebrate Launch Modal

sharpirate:
### Time Estimate
• Test: Short
• Review: Short

### Proposed Changes
Currently, when the user launches their site we are showing the Celebrate Launch modal and we're removing the `celebrateLaunch` URL param to avoid showing the modal more than once.

The problem is that the `celebrateLaunch` param comes first in the URL so removing `?celebrateLaunch=true` results in an invalid URL because subsequent params are appended with `&` and not `?`.

This PR uses `omitUrlParams` to guarantee that the order of the URL params does not matter

### Before

https://user-images.githubusercontent.com/20927667/222262941-c7057cb6-435b-4894-9d70-3dd187bd1a73.mov

### After

https://user-images.githubusercontent.com/20927667/222262886-d8356d5e-da16-4fd0-b6da-02c933f1b209.mov

#### Reproduce Bug
1. Go to https://wordpress.com/start and create a new Launchpad site
2. Once you arrive at Launchpad, click **Launch your site**
3. You should get redirected to **Home** with the Celebrate Launch modal open
4. Close the Celebrate Launch modal, you should see the URL **DOES** contain invalid params
5. Refresh the page - The URL fails to resolve and you **DO NOT SEE** the Home of your site

#### Test Fix
1. Checkout this branch
2. Go to http://calypso.localhost:3000/start and create a new Launchpad site
3. Once you arrive at Launchpad, click **Launch your site**
4. You should get redirected to **Home** with the Celebrate Launch modal open
5. Close the Celebrate Launch modal, you should see the URL **DOES NOT** contain invalid params
6. Refresh the page - The URL resolves successfully and you **DO SEE** the Home of your site

### DIFF
--- a/client/landing/stepper/declarative-flow/build.ts
+++ b/client/landing/stepper/declarative-flow/build.ts
@@ -47,7 +47,7 @@ const build: Flow = {
 				case 'processing':
 					if ( providedDependencies?.goToHome && providedDependencies?.siteSlug ) {
 						return window.location.replace(
-							`/home/${ providedDependencies?.siteSlug }?celebrateLaunch=true&launchpadComplete=true`
+							`/home/${ providedDependencies?.siteSlug }?launchpadComplete=true&celebrateLaunch=true`
 						);
 					}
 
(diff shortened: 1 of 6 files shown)


jeyip (approved):
## Testing
### Requirements
- [x] Close the Celebrate Launch modal, you should see the URL DOES NOT contain invalid params
Refresh the page
- [x] The URL resolves successfully and you DO SEE the Home of your site

### Browsers
- [x] Chrome

mattwiebe (approved):
This is a better way to do this 👍

PR #73979, opened 2023-03-02
Skip the "Editor canvas loads" test

p-jackson:
## Proposed Changes

The logic of the test is correct. It knows the sidebar is open by default and it tries to close it. The problem is that there's now no longer a a dedicated "close sidebar" button. We used to click the same "open nav" button which would toggle the state of the sidebar. The "open" button is now no longer available when the sidebar is open.


jkguidaven (approved):
LGTM 👍

PR #73663, opened 2023-02-22
Virtual themes i1: Implement A/B test

mmtr:
Part of https://github.com/Automattic/dotcom-forge/issues/1655
See pbxNRc-2n5-p2.

## Proposed Changes

Implements the logic to handle the Abacus experiment that will A/B test the virtual themes i1.

## Testing Instructions

- Go to the experiment details in Abacus (find a link in pbxNRc-2n5-p2) and click on the Overview tab.
- Navigate to the Audience section and hover over the `control` assignment group.
- Upon mouse hover, click on the link to be assigned that assignment group.
- Wait a minute for the changes to be applied.
- Use the Calypso live link below.
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>`.
- Make sure the virtual themes are not displayed.
- Now assign yourself to the `treatment` group.
- Wait a minute for the changes to be applied.
- Reload `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>`.
- Make sure the virtual themes are displayed now.

PR #73989, opened 2023-03-02
Prevent calypso_signup_design_scrolled_to_end from being tracked if designs have not been loaded

mmtr:
## Proposed Changes

Fixes a regression introduced by https://github.com/Automattic/wp-calypso/pull/73663 that caused the `calypso_signup_design_scrolled_to_end` event to be tracked without the user viewing all the designs.

The culprit is that we need to determine first the assignment group of the virtual themes i1 experiment and then fetch the designs based on that group. During that time, we render an empty page so the user technically has reached the end of the page and thus `calypso_signup_design_scrolled_to_end` gets tracked. However the event is intended to be triggered when a user has scrolled to the end of the page AFTER viewing all designs.

So this PR mitigates the issue by adding a safeguard to prevent the component from being rendered (effectively  blocking the `calypso_signup_design_scrolled_to_end` event) if we're still loading the A/B test group.

Props to @taipeicoder for finding out the culprit (and to @ianstewart for noting the spike).

Further context: pekYwv-O9-p2

## Testing Instructions

- Open a new incognito/private session (to prevent the cache from altering the results).
- Open the network tab in the browser devtools.
- Go to `wordpress.com/setup/site-setup/designSetup?siteSlug=<SITE_SLUG>`.
- Note how there is a `t.gif` request to track the `calypso_signup_design_scrolled_to_end` event upon loading the page.
- Repeat the same steps using the Calypso live link below.
- Make sure the `calypso_signup_design_scrolled_to_end` event is not tracked when loading the page.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/test/unified-design-picker.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/test/unified-design-picker.tsx
@@ -44,7 +44,7 @@ jest.mock( 'calypso/state/sites/hooks/use-premium-global-styles', () => ( {
 } ) );
 
 jest.mock( 'calypso/lib/explat', () => ( {
-	useExperiment: () => [ false, 'control' ],
+	useExperiment: () => [ false, null ],
 } ) );
 
 /**
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/test/unified-design-picker.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/test/unified-design-picker.tsx
@@ -43,6 +43,10 @@ jest.mock( 'calypso/state/sites/hooks/use-premium-global-styles', () => ( {
 	usePremiumGlobalStyles: () => ( { shouldLimitGlobalStyles: false } ),
 } ) );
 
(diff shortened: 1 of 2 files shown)


arthur791004:
Could we just add the check to [here](https://github.com/Automattic/wp-calypso/blob/8bc2c1b3819176ee69129d9bedc06f91e61e6192/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx#L584)? I think we should not render anything before the experiment is loaded. The `useStarterDesignsQuery` hook has the cache of the starter designs. If you land on the Design Picker multiple times, it has a chance to show the designs first and then become a white screen since the experiment is loaded (I guess). Hence, I think it would be better to avoid rendering anything before the experiment is loaded. What do you think?

mmtr:
Thanks @arthur791004, that makes sense! Added it in https://github.com/Automattic/wp-calypso/pull/73989/commits/05fbae5bfea302933dac3ca0962b66caab3d4a2f.

The `UnifiedDesignPickerStep` component is so big I didn't notice we're not rendering it under some circumstances 😅

arthur791004:
Yes...it grows bigger and bigger 😓

arthur791004 (approved):
LGTM 👍

arthur791004:
Looks like we still need to update the test case

PR #74007, opened 2023-03-02
VAT: Add VAT text fallback for Business Tax ID

sirbrillig:
## Proposed Changes

https://github.com/Automattic/wp-calypso/pull/73686 changed the labels for checkout and payments method management  so that instead of referencing "VAT" they referenced "Business Tax" to better support the wording used for taxes in different countries. However, that PR was deployed before the new translations were complete. This PR restores the old translations if the translations are not yet complete.

Fixes https://github.com/Automattic/payments-shilling/issues/1415

Before:

Screenshot 2023-03-02 at 11 47 04 AM

After:

Screenshot 2023-03-02 at 11 53 09 AM

## Testing Instructions

See the testing instructions on https://github.com/Automattic/wp-calypso/pull/73686 but view them in locales other than `en` to verify they are still translated. (This may actually be difficult because you'll need to find a language where the text has not yet been translated.)

### DIFF
--- a/client/me/purchases/billing-history/main.tsx
+++ b/client/me/purchases/billing-history/main.tsx
@@ -35,8 +35,8 @@ function BillingHistory() {
 	let editVatText = translate( 'Edit Business Tax ID details' );
 	let addVatText = translate( 'Add Business Tax ID details' );
 	if (
-		getLocaleSlug() !== 'en' ||
-		getLocaleSlug() !== 'en-gb' ||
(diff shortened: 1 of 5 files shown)


michaeldcain (approved):
This looks good to me.

michaeldcain:
I approved it, but it looks like [the strings](https://translate.wordpress.com/deliverables/overview/7830896/) from #73686 are complete, so we can probably just close this.

PR #69469, opened 2022-10-26
Remove `react-intersection-observer` usage from devdocs

p-jackson:
#### Proposed Changes

Removes the last use of the `react-intersection-observer` package and removes it from `package.json`.

The final use of `react-intersection-observer` was in `/devdocs`, chunking up the component galleries so they wouldn't try to render every component on page load (presumably that's quite slow).

This simplifies our dependency list, instead opting to factor out the code we'd already written into something reusable.
Discussion here https://github.com/Automattic/wp-calypso/pull/69119#issuecomment-1281714675
Interestingly the reason devdocs dropped its previous dependency (`react-lazily-render`) and moved to `react-intersection-observer` in #56154 was because `react-lazy-render` hadn't updated their peer dependency to React 17. That's exactly the sort of reason why I think we don't want too many React-related dependencies.

[Plus somehow it's managed to shave 3.5KB off the parsed size of some bundles 🤷‍♂️ ](https://github.com/Automattic/wp-calypso/pull/69469#issuecomment-1291717661)

#### Testing Instructions

* Test the component galleries at `/devdocs/blocks` and `/devdocs/design`
* Notice how the scroll bar shrinks as you scroll down the page, that's because components are rendering as you scroll
* You should still be able to find any component/block by searching
* Confirm that there are no changes in behaviour between this branch and production https://wpcalypso.wordpress.com/devdocs/design

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

PR #69119, opened 2022-10-17
Sites Management Page: Replace react-intersection-observer with IntersectionObserver

mpkelly:
Fixes https://github.com/Automattic/wp-calypso/issues/68625

#### Proposed Changes

* Remove `react-intersection-observer` and reuse some existing logic elsewhere in the codebase

#### Testing Instructions

This hook is used on the `/sites` page and at for a nudge at `/activity-log/:business-or-pro-site-with-jetpack`. 

* To test, go to `/sites` with one or more unlaunched sites. Note how many Tracks events fire with `calypso_sites_dashboard_site_launch_nag_inview` in both **list** and **grid** views. Repeat the process on the Calypso Live site and verify the new code produces the same output. The events should only ever fire one and you can verify this by scrolling the nag UI in and out of the screen. 
* Testing `/activity-log/:business-or-pro-site-with-jetpack` is a bit awkward. The easiest way to check is to modify the code to force the banner to display. You can do this like so:
    - go to `client/my-sites/activity/activity-log/index.jsx` and change line `551` so that `<VisibleDaysLimitUpsell/>` always get rendered. 
    - Next, go to `<VisibleDaysLimitUpsell/>` and change line `60` so it doesn't return `null`. Finally, go to `/activity-log/business-or-pro-site-with-jetpack` and scroll to the bottom and should see the banner. As before, ensure the Tracks event only fires one per page load after switching from `trunk` to `remove/react-intersection-observer-dep`

image

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Related to #

PR #74021, opened 2023-03-02
Sites Management Page: Use `react-intersection-observer` for `inView`

danielbachhuber:
See conversation in p1677514081942499-slack-C0347E545HR

Remaining use will be handled in https://github.com/Automattic/wp-calypso/issues/74022

## Proposed Changes

Switches to `react-intersection-observer` to manage `inView` state for various Sites Management Page components.

## Testing Instructions

1. Navigate to `/sites`.
2. Switch to list view.
3. Verify stats and the ellipsis menu appear as expected.
4. Verify `/rest/v1.1/sites/<site>/stats/insights` API requests are only made for the sites inview.
5. Scroll the page and verify the sites appear as expected while you scroll.
6. Switch to Grid view.
7. Verify the ellipsis menu appears as expected.
8. Scroll the page and verify the sites appear as expected.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -176,6 +176,7 @@
 		"react-day-picker": "^7.4.10",
 		"react-dom": "^17.0.2",
 		"react-element-to-jsx-string": "^14.3.2",
+		"react-intersection-observer": "^9.4.3",
 		"react-live": "^2.3.0",
 		"react-modal": "^3.14.3",
 		"react-query": "^3.32.1",
--- a/client/sites-dashboard/components/sites-grid-item.tsx
+++ b/client/sites-dashboard/components/sites-grid-item.tsx
@@ -2,9 +2,9 @@ import { useSiteLaunchStatusLabel, getSiteLaunchStatus } from '@automattic/sites
 import { css } from '@emotion/css';
 import styled from '@emotion/styled';
 import { useI18n } from '@wordpress/react-i18n';
-import { AnchorHTMLAttributes, memo, useState } from 'react';
+import { AnchorHTMLAttributes, memo } from 'react';
+import { useInView } from 'react-intersection-observer';
 import { SiteExcerptData } from 'calypso/data/sites/site-excerpt-types';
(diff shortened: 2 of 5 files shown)


wojtekn (approved):
It works as expected.

Side note - as @sejas mentioned, we are loading the stats even in mobile view, where that column is hidden. It works the same way in this PR and trunk, so probably we can fix it under another ticket, though.

sejas:
- Tagging for reference:
  - https://github.com/Automattic/wp-calypso/pull/69119
  - https://github.com/Automattic/wp-calypso/pull/69469

sejas:
I feel this PR goes against some effort we put into removing this library. I'm not able to reproduce the issue. Is there a simple way to reproduce it?

danielbachhuber:
> I'm not able to reproduce the issue. Is there simple way to reproduce it?

@sejas No, the issue reproduces intermittently.

> I feel this PR goes against some effort we put into removing this library.

See p1666729978883699/1666718787.803859-slack-C0347E545HR

We're spending a lot of effort re-discovering and re-inventing the wheel.

sejas:
@danielbachhuber, Is this PR trying to fix this same issue p1666729978883699/1666718787.803859-slack-C0347E545HR ?

danielbachhuber:
@sejas No, I think this is some other bug. By using an open source library, we will benefit from all of the bug fixes identified by other users, and can contribute our own fixes when we find bugs. When inventing our own solution, we have to identify, debug, and fix all of the bugs on our own.

sejas:
I'm just wondering how we can ensure this PR solves the issue?
Because it's not reproducible, I cannot say this PR fixes it. Maybe it comes from a different conflict/race condition.
@danielbachhuber, were you able to see the issue on production?

danielbachhuber:
> I'm just wondering how we can ensure this PR solves the issue?

@sejas We can't. If the issue still exists after we merge the PR, we can track down the bug and contribute the fix upstream so others can benefit. But, my hunch is that the random bug I'm experiencing was fixed at some point in the library.

Issue #73908, opened 2023-02-28
Domains: When removing a domain mapping record, the product seems wrong


ebinnion:
### Quick summary

When I attempt to remove a secondary domain mapping record for a site, at the top of the screen there's a mention of:

"Remove product eric.blog"

Which doesn't seem technically correct to me and kind of scared me. 😄 

<img width="1395" alt="Screenshot 2023-02-28 at 4 43 28 PM" src="https://user-images.githubusercontent.com/1126811/221999500-8acc1635-f936-4445-9fd2-3be645290fd6.png">

### Steps to reproduce

1. Have more than on domain for a site
2. View settings
3. Click "Delete"
4. Arrive on the page above

### What you expected to happen

For the domain that I'm deleting to be the domain mentioned.

### What actually happened

My primary domain was mentioned.

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

villanovachile:
Resolved with #74027

PR #74027, opened 2023-03-03
Cancel Purchase Form: Provide more context about product being canceled

villanovachile:
Resolves to #73908 

## Proposed Changes

* Currently the cancel purchase form displays the `site.slug` prop next to the "Remove product" or "Remove plan" heading at the top of the page. This means that regardless of the product being removed, whether it's a plan, domain connection for a secondary domain, or add-on product such as a site redirect or custom CSS, the only thing being displayed is the primary site address.

This PR seeks to provide additional context for the product/plan being removed. I decided to use switch statement here on the condition of `purchase.productId` matching a domain connection, site redirect, and a default description which should suffice for plans or other add-ons.

### Any non-plan Product Removal - Before

01 - before - removing any non-plan product

### Plan Removal - Before

02 - before - removing any plan

### Removing domain connection that is not the primary address
Uses the `purcachase.meta` prop to display the name of the actual domain connection

03 - after - removing domain connection

### Removing site redirect
Uses the `purcachase.meta` prop to display the name of the domain that the site redirect is pointing to
04 - after - removing site redirect

### Removing any non-plan product, such as custom CSS
05 - after - removing any other product

### Removing any plan

06 - after - removing any plan

## Testing Instructions

* Add a plan, domain connection, and add-ons including custom CSS and Site Redirect
* Remove  each plan and ensure that the context displayed at the top reflects the name of the product being removed and the appropriate address is being displayed that the removal is affecting
* Test on Simple & Atomic

## Pre-merge Checklist


villanovachile:
Heads up, FSE smoke tests are currently failing on trunk: p1677744583428479-slack-C1A1EKDGQ

client/components/marketing-survey/cancel-purchase-form/index.jsx
```diff
@@ -670,7 +671,20 @@ class CancelPurchaseForm extends Component {
 					<BlankCanvas className="cancel-purchase-form">
 						<BlankCanvas.Header onBackClick={ this.closeDialog }>
 							{ this.getHeaderTitle() }
-							<span className="cancel-purchase-form__site-slug">{ site.slug }</span>
+							<span className="cancel-purchase-form__site-slug">
+								{ ( () => {
+									switch ( purchase.productId ) {
+										// product.Id 5 is Domain Mapping
+										case 5:
```
fullofcaffeine: I don't know how the product data is created, but are we sure that these products will always keep these ids? I assume these are products that were injected as part of some data migration script, and that these ids will not change, but it's worth checking that, because this code might not work in a fresh instance of Calypso (though I'm not sure if this component is meant to be used outside of the WPCOM context).

villanovachile: This was definitely a point of concern for me as well. Initially I'd used the product name, but that seemed to be too volatile. Any idea what team we could check with to confirm?

client/components/marketing-survey/cancel-purchase-form/index.jsx
```diff
@@ -670,7 +671,20 @@ class CancelPurchaseForm extends Component {
 					<BlankCanvas className="cancel-purchase-form">
 						<BlankCanvas.Header onBackClick={ this.closeDialog }>
 							{ this.getHeaderTitle() }
-							<span className="cancel-purchase-form__site-slug">{ site.slug }</span>
+							<span className="cancel-purchase-form__site-slug">
+								{ ( () => {
```
fullofcaffeine: I'd avoid the IIFEE here and instead move the switch to a instance method of the component class, i.e:

```js
    // feel free the tweak the method name
	getCancelSubjectDescription() {
		switch ( purchase.productId ) {
			// product.Id 5 is Domain Mapping
			case 5:
				return purchase.productName + translate( ' for ' ) + purchase.meta;
			// product.Id 36 is a Site Redirect
			case 36:
				return purchase.productName + translate( ' to ' ) + purchase.meta;
			default:
				return purchase.productName + translate( ' for ' ) + site.slug;
		}
	}
```

And then in the JSX:

```jsx
(...)
							<span className="cancel-purchase-form__site-slug">
								{ this.getCancelSubjectDescription() }
							</span
(...)

```

This would also improve code comprehension as it's easier to figure out what that piece of code is doing by reading the function name.

client/components/marketing-survey/cancel-purchase-form/index.jsx
fullofcaffeine: I think you could check with @Automattic/shilling.

client/components/marketing-survey/cancel-purchase-form/index.jsx
villanovachile: That makes sense! This way it's something that is more modular as well, right? I'll implement this once we get some confirmation from Shilling about product IDs!

fullofcaffeine:
This is nice little UX improvement! I think that provided these are indeed "frozen" products in the DB, the current logic is pragmatic and fine as is (even though I'd probably move the switch over to a helper method, but up to you).

One other thing I think we should improve, though, is the actual presentation of the subject label; see the following screenshot:
Screenshot from 2023-03-03 18-16-41

And notice how the "Cancel plan" text has a poor contrast with the white background; the font is also very different from the text inside the span, making it look out of place. I think a simple solution would be to just move that prefix text inside the `<span>` so that everything has the same look:


Screenshot from 2023-03-03 18-33-50

But it might be interesting to ask someone with better UX/designer skills for insights here.

villanovachile:
> And notice how the "Cancel plan" text has a poor contrast with the white background; the font is also very different from the text inside the span, making it feel out of place. I think a simple solution would be to just move that prefix text inside the span so that everything is formatted the same

100% I think that looks much better!

client/components/marketing-survey/cancel-purchase-form/index.jsx
```diff
@@ -651,6 +651,19 @@ class CancelPurchaseForm extends Component {
 		return translate( 'Cancel product' );
 	}
 
+	getCanceledProduct() {
+		const { purchase, site, translate } = this.props;
+		switch ( purchase.productId ) {
+			// product.Id 5 is Domain Mapping
+			case 5:
+				return ' ' + purchase.productName + translate( ' for ' ) + purchase.meta;
```
tyxla: There are a couple of problems with this kind of localized string:

* As a translator, I'd be confused as to what to translate. The ` for ` string can be translated in different ways in some languages, depending on the context and what it applies to. To address this, I'd recommend improving the localized string to contain placeholders, and if needed, adding translator comments to clarify for someone without context. See [this example](https://github.com/Automattic/wp-calypso/blob/6b62dd76eed12916a92a8b64390c8492413181a0/client/login/magic-login/handle-emailed-link-form-jetpack-connect.tsx#L108) that adds an argument placeholder that translators should leave intact, and we replace automatically on the translated string. Finally, in an ideal world, translators should not be glancing at the code to understand what they're translating, so let's make sure we provide all the context. 😉 
* A leading or trailing space in a localized string can easily be lost when translating the string. I'd recommend avoiding those. If we add placeholders for the product name and meta at the beginning and end of the string (as suggested above), this problem will be solved too.

This applies to the rest of the new strings introduced in this file.

Hope that's helpful, but please let me know if I should provide further details.

tyxla: The leading space returned by this string feels like a code smell. I wouldn't expect a function that returns a string to return a leading space. This indicates that the place where we're rendering this function's result needs a space. This would also make sense if we want to reuse the result of that function somewhere else, where a leading space wouldn't be necessary.

client/components/marketing-survey/cancel-purchase-form/index.jsx
sirbrillig: `productId` should be stable and safe to use here. Another option is `productSlug`, neither of which are likely to change. (For reference, `purchase` does have TS types which you can find [here](https://github.com/Automattic/wp-calypso/blob/6ebce469050693a803502fc84490e876a0c674e3/client/lib/purchases/types.ts#L53-L55).)

villanovachile: Thanks for confirming @sirbrillig , really appreciate the insight here! Since you've confirmed `productSlug` should be stable as well, I'll opt to use this, as it'll allow me to do away with the comments explaining what each `productId` is, as `productSlug` is self explanatory in each switch case.

client/components/marketing-survey/cancel-purchase-form/index.jsx
villanovachile: Thanks so much for your input. Your rationale makes a lot of sense and I've gone ahead and refactored to include the entire message using placeholder text. Additionally as per the conversation [here](https://github.com/Automattic/wp-calypso/pull/74027#discussion_r1126724814), I've also opted to use the `productSlug` to identify the product for use in the switch statement. I believe this should rectify and resolve all of the above concerns!

villanovachile: Should be resolved by using a placeholders!

client/components/marketing-survey/cancel-purchase-form/index.jsx
```diff
@@ -651,6 +651,30 @@ class CancelPurchaseForm extends Component {
...
+			purchase: { productSlug, productName, meta },
+			site: { slug },
+			translate,
+		} = this.props;
+		const headerTitle = this.getHeaderTitle();
+		const args = { headerTitle, productName, purchaseMeta: meta, siteSlug: slug };
+		// Example output: 'Remove product: Domain Connection for mygroovydomain.com'
+		switch ( productSlug ) {
+			case 'domain_map':
+				return translate( '%(headerTitle)s: %(productName)s for %(purchaseMeta)s', {
```
tyxla, line 666: I would suggest adding a `translators: ` comment above each of those strings. Those will add the missing context for translators, since the current three strings may still be a bit confusing or lacking context. Those comments usually also add an example, too, which is extra useful for translators. Here's an example:

https://github.com/Automattic/wp-calypso/blob/94fe59f75d919c439f182764eed7d86f1f63da48/apps/editing-toolkit/editing-toolkit-plugin/editor-site-launch/src/launch-progress/index.tsx#L23

villanovachile: Thanks @tyxla ! I've tried to provide additional context, but hopefully in a manner that isn't too dense! Let me know if you think this last commit will work!

client/components/marketing-survey/cancel-purchase-form/index.jsx
```diff
@@ -651,6 +651,38 @@ class CancelPurchaseForm extends Component {
...
 	}
 
+	getCanceledProduct() {
+		const {
+			purchase: { productSlug, productName, meta },
+			site: { slug },
+			translate,
+		} = this.props;
+		const headerTitle = this.getHeaderTitle();
+		const args = { headerTitle, productName, purchaseMeta: meta, siteSlug: slug };
```
tyxla: One little nit: we'll provide all four args to the localized strings, but each of those strings needs only 3 of those args.

tyxla (commented):
I've left a non-blocking observation, but other than that, things look great from my perspective 👍 

Since I only intended to do a drop-by review, I'll leave the last review and approval to @fullofcaffeine.

Thanks @villanovachile 🙌

client/components/marketing-survey/cancel-purchase-form/index.jsx
villanovachile: Thanks, Marin! I went ahead and only provided the necessary args to the localized strings. I'll await @fullofcaffeine for that final approval!

tyxla: Looking great, thanks for taking care of that 👍

client/components/marketing-survey/cancel-purchase-form/index.jsx
fullofcaffeine: TIL about the `translators:` doc metadata in comments, interesting :)

fullofcaffeine: >  To address this, I'd recommend improving the localized string to contain placeholders, and if needed, adding translator comments to clarify for someone without context|

That's a very good point! Some languages might switch the order of the subjects / conjunctions or even change it in ways that are very alien to the structure/grammer of English. Best to make the whole thing "atomic" and let translators deal with how to organize the statement depending on the language.

fullofcaffeine (approved):
LGTM!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7918896

Thank you @villanovachile for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7918896).

PR #74032, opened 2023-03-03
Adding Hotjar tracking to the Woo Express specific plans page

gmovr:
Related to #

## Proposed Changes

* This will initialize Hotjar tracking on `/plans/woo-test-site.wpcomstaging.com`
* It uses the existing `loadTrackingTool`, which calls `addHotJarScript`. This function is already aware of user consent (GDPR/opt-out), so it's safe to call like this.

## Testing Instructions

* Via the browser console, run `localStorage.debug='calypso:analytics:hotjar' (for seeing debug messages in the console.
* Go to `/plans/woo-test-site.wpcomstaging.com` where `woo-test-site` is the subdomain of a site with Woo Express trial. If you don't have one, go ahead and create one.
* When the page has loaded, observe that you see a message like `calypso:analytics:hotjar Loading HotJar script +0ms` in the console, and you can also check that window.hj is defined.
* Make sure ad-tracking is enabled

## Pre-merge Checklist


gmovr:
@robertsreberski -- I think we should remove HotJar from `AdTrackersInitGuards` in tracker-buckets.ts, since it's not initialized in `load-tracking-scripts`, so `window.hj` won't be defined until the script is ran, but `mayWeTrackByTracker( 'hotjar' )` will still check these init guards. What do you think?

daledupreez:
@gmovr, I am not able to get HotJar to load - when I run via a local build it seems like it's getting to the `hj in window` check for the guard, and that's failing, but I also see the problem via the calypso.live branch.

> Make sure ad-tracking is enabled

I am not sure what else to try. Also, could you be more explicit about this instruction? It's not obvious to me that this means the Calypso `ad-tracking` feature needs to be enabled, e.g. by `?flags=ad-tracking` in the URL.

gmovr:
> @gmovr, I am not able to get HotJar to load - when I run via a local build it seems like it's getting to the `hj in window` check for the guard, and that's failing, but I also see the problem via the calypso.live branch.
> 
> > Make sure ad-tracking is enabled
> 
> I am not sure what else to try. Also, could you be more explicit about this instruction? It's not obvious to me that this means the Calypso `ad-tracking` feature needs to be enabled, e.g. by `?flags=ad-tracking` in the URL.

Hi, that's right -- the flag needs to be passed in via `?flags=ad-tracking`, or you can set `ad-tracking: true` in development.json. 

It appears the `window.hj` check is faulty though. I believe that's mostly in cases where you have a tracking pixel, e.g. Google Analytics, and you want to make sure the actual tracking script is loaded before calling one of the methods on the global object. In the case of Hotjar, we're loading the external scipt and there are no subsequent calls to this, so the check isn't needed.

I removed the check for hotjar for now -- so you should be able to see it both locally, and via calypso.live. If you're in GDPR, you also need to accept the cookie banner, as hotjar is categorized as an advertising tracker.

robertsreberski (approved):
The solution works for me! Indeed, the HotJar init guard doesn't seem necessary.

I've created a Woo Express page and can confirm the HotJar fires well on `/plans/<site>` - as described in the test plan.

Later in the console logs, I get second note about NOT loading HotJar: 
`calypso:analytics:hotjar Not loading HotJar script +758ms`

Is it because HotJar is already loaded? It's not blocking though, as it appears to work.

robertsreberski:
Together with @gmovr, we've investigated that the second log had meaning and reflected the logic issue.

The HotJar should already be pre-applied to all /plans pages in calypso. However, it looks like the init guard was preventing HotJar from loading altogether. This guard is indeed faulty and redundant so the goal of this change will be achieved just by simply removing it - that way, all HotJar tracking comes back to work, and the above-described test plan succeeds.

Issue #73759, opened 2023-02-24
Pattern Assembler - Remove the 8px padding from pattern previews


miksansegundo:
### Description

Let's remove the `8px` padding from the pattern previews (small and large). It's introduced by `.editor-styles-wrapper`.

![Image](https://user-images.githubusercontent.com/1881481/221131710-b6e3c941-4fa1-4072-8bff-f3e3fcfd8eba.png)

|Before|After|
|-|-|
|![Image](https://user-images.githubusercontent.com/1881481/221131620-5bcee952-eac2-4e68-8ac5-c9027fbac9b2.png)|![Image](https://user-images.githubusercontent.com/1881481/221131597-020d7ebd-b895-40db-a2e7-8e4c67807bfb.png)|

#### Why?
The previews look much better without it. Also, that padding is not visible in the editor or the front of the site, and the site preview in the assembler should look "like" the front of the site.

![Image](https://user-images.githubusercontent.com/1881481/221131757-2ef62caa-a3d0-4872-92a4-f3903a1eae3c.png)

arthur791004:
Closed it via https://github.com/Automattic/wp-calypso/pull/74033

PR #74033, opened 2023-03-03
Site Assembler: Fix the padding and background of the preview

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/73759

## Proposed Changes

* This PR is focusing on removing the outer padding on each preview. Also, setting the background color in the large preview so that it's consistent with the color variation

https://user-images.githubusercontent.com/13596067/222632328-a171de54-3069-44f9-8d47-20082e889288.mov

## Testing Instructions

* Go to `/setup?siteSlug=<your_site>&flags=pattern-assembler/sidebar-revamp,pattern-assembler/color-and-fonts`
  * Choose a color variation first and the background color is not white so that we're easy to notice where the padding is.
  * Browser header/sections/footer patterns and ensure the outer padding is gone
  * Select header, sections, and footer, and ensure
    * The outer padding is gone
    * There is padding between the header and sections, and the background color is the same as the selected color variation
    * There is padding between the footer and sections, and the background color is the same as the selected color variation

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -15,15 +15,13 @@
 	margin: 0;
 	list-style-type: none;
 	overflow: auto;
-	// @TODO: Use theme setting "styles.color.background"
-	background: var(--studio-white);
+	background: var(--pattern-large-preview-background);
 
-	// @TODO: Use theme setting "styles.spacing.blockGap"
 	.pattern-large-preview__pattern-header {
-		margin-block-end: 1.5rem;
+		margin-block-end: var(--pattern-large-preview-block-gap);
 	}
 	.pattern-large-preview__pattern-footer {
-		margin-block-start: 1.5rem;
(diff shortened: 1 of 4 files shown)


taipeicoder:
Code looks good 🙂 however, I noticed a couple of things:

The color transition doesn't look so smooth. I think it's because the background doesn't have transition:
ezgif-3-fc7b9eaa83

Some patterns probably need left/right padding, for instance:
Screenshot 2023-03-03 at 2 43 56 PM

Screenshot 2023-03-03 at 2 44 28 PM

arthur791004:
> Some patterns probably need left/right padding, for instance:

Currently, the behavior is the same as the site editor. If those patterns need horizontal padding, we have to update those patterns. cc @alaczek 

image

image

arthur791004:
> The color transition doesn't look so smooth. I think it's because the background doesn't have transition:

We don't have any transition for both. Let me think about how to fix it 🤔

arthur791004:
> The color transition doesn't look so smooth. I think it's because the background doesn't have transition:

The reason is the large preview will be updated first and then the selected patterns so that the transition looks weird. I made a change to delay updating the styles to make the transition smooth. Not sure whether there is any better idea 🤔

taipeicoder (approved):
Looks so much better now 🙂

PR #74052, opened 2023-03-03
I18N: Localize the labels (categories / description) into the user lo…

sdnunca:
## Context
When selecting "Add new page" in WP.com, the names of the categories/patterns appear in the site language, along with the actual pattern that's going to be inserted. It'd make more sense to use the user locale to display the pattern names/description, while still inserting the pattern in the site language.

image

Related to Automattic/i18n-issues#531

## Proposed Changes

* When the user locale differs from the blog locale, we also fetch the pattern templates in the user locale, and we replace the categories/description in the original data with the ones in the user locale. 
* This approach requires no API changes and should not interfere with any caching mechanism present on the end-point.

## Testing Instructions

* Select a simple site, update your user locale to be different from the site locale
* Trigger the Add new page pop-up and confirm that the categories and descriptions are shown in the user locale, while the content inserted by the templates is in the site locale. 

## Pre-merge Checklist

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/class-starter-page-templates.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/class-starter-page-templates.php
@@ -7,6 +7,8 @@
 
 namespace A8C\FSE;
 
+use function A8C\FSE\Common\get_iso_639_locale;
+
 /**
  * Class Starter_Page_Templates
  */
@@ -42,9 +44,11 @@ class Starter_Page_Templates {
 	/**
 	 * Gets the cache key for templates array, after setting it if it hasn't been set yet.
 	 *
+	 * @param string $locale The templates locale.
+	 *
 	 * @return string
 	 */
-	public function get_templates_cache_key() {
+	public function get_templates_cache_key( string $locale ) {
 		if ( empty( $this->templates_cache_key ) ) {
 			$this->templates_cache_key = implode(
 				'_',
@@ -52,11 +56,11 @@ class Starter_Page_Templates {
 					'starter_page_templates',
 					A8C_ETK_PLUGIN_VERSION,
 					get_option( 'site_vertical', 'default' ),
-					$this->get_verticals_locale(),
 				)
 			);
 		}
-		return $this->templates_cache_key;
+
+		return $this->templates_cache_key . '_' . $locale;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


yuliyan (commented):
Tested with one of my simple sites and a got a mix of categories in English, German (user locale), and Arabic (site locale):

CleanShot 2023-03-06 at 14 40 14

I noticed that on my testing site, [this condition](https://github.com/Automattic/wp-calypso/pull/74052/files#diff-07af893e807efa3387023fef4b6653614b0445b2715d52b2a41b951f99f25460R165) is not being fulfilled due to `description` property is empty for most template entries. After commenting it out, I'm seeing the expected translations:
CleanShot 2023-03-06 at 14 46 43

I'm not sure if I'm missing something, but it seems like we want to always overwrite the categories and description with the `$user_page_templates` values, and if that's the case, [the check for empty values](https://github.com/Automattic/wp-calypso/pull/74052/files#diff-07af893e807efa3387023fef4b6653614b0445b2715d52b2a41b951f99f25460R165) should probably be removed.

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/class-starter-page-templates.php
```diff
@@ -267,7 +289,7 @@ public function clear_sideloaded_image_cache( $id ) {
 	 * Deletes cached templates data when theme switches.
 	 */
 	public function clear_templates_cache() {
-		delete_transient( $this->get_templates_cache_key() );
+		delete_transient( $this->get_templates_cache_key( $this->get_verticals_locale() ) );
```
yuliyan, line 294: Should we also clear template data that has been cached for other locales?

sdnunca:
@yuliyan thanks a lot for catching that, I've updated the check!

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/class-starter-page-templates.php
sdnunca: The locale was always part of the cache key, so the previous behavior was only to delete the selected locale's transient. 

I didn't alter the default behavior as I'm not sure what the intention is here.

yuliyan (approved):
Tested again and it works as expected now. The failing e2e tests seem to be unrelated, LGTM 🚢

PR #74122, opened 2023-03-06
Help Center: Add Help Center on support sites

renancarvalho:
## Proposed changes

On [March 2nd](https://github.com/Automattic/wp-calypso/pull/73882), team vertex made the HelpCenter available on support pages, unfortunately that PR [caused issues](p58i-e0x-p2) in production. This PR fixes those issues.

## The fix

A. The Help Center was being wrongly loaded on all pages, we have fixed that by checking if the site is a support site.

B. We found that HelpCenter was relying on `@wordpress/interface` which uses `wp-preferences`, `wp-preferences` is not available or not loaded ( we are still not sure ) outside the editor (Guteberg) context. We have removed @wordpress/interface dependency thus wp-preferences

## How to test

- Pull and run this branch
- Navigate to apps/editing-toolkit and run yarn dev --sync to sync your sandbox
- Make sure to sandbox wordpress.com/support
- Navigate to https://wordpress.com/support
- Check if the Help Center is rendered on the admin bar
- Help Center should work fully and normally


apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
```diff
@@ -127,7 +127,17 @@ public function enqueue_script() {
 	 * Get current site details.
 	 */
 	public function get_current_site() {
-		$site     = \Jetpack_Options::get_option( 'id' );
+		$is_support_site = defined( 'WPCOM_SUPPORT_BLOG_IDS' ) && $this->is_support_site();
```
alshakero: This `defined` check should be part of `is_support_site` since that's the function's job.

apps/editing-toolkit/editing-toolkit-plugin/help-center/src/help-center.js
```diff
@@ -14,6 +13,10 @@ import { getSectionName } from 'calypso/state/ui/selectors';
 import { whatsNewQueryClient } from '../../common/what-new-query-client';
 import CalypsoStateProvider from './CalypsoStateProvider';
 
+function PinnedItems( { scope, ...props } ) {
```
alshakero, line 20: This deservers a comment explaining why we avoid using interface. Because it's a bit of a hack.

I'd say: Implement `PinnedItems` to avoid importing `@wordpress/interface`. Because `@wordpress/interface` depends on `@wordpress/preferences` which is not always available outside the editor, causing the script to not be enqueued due to the missing dependency.

alshakero (commented):
Beautiful. Left some comments.

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-help-center.php
renancarvalho: Absolutely 👍

apps/editing-toolkit/editing-toolkit-plugin/help-center/src/help-center.js
renancarvalho: ✅

heavyweight (changes requested):
We have a regression in wp-admin. The hc is not showing for sites that are not support site.

Screenshot 2023-03-09 at 15 59 46

renancarvalho:
> We have a regression in wp-admin. The hc is not showing for sites that are not support site.
> 
> Screenshot 2023-03-09 at 15 59 46

Great catch @heavyweight  I am enqueueing for both hooks, `admin_enqueue_scripts` and `wp_enqueue_scripts` which fixes the problem :) Thank you!

alshakero (approved):
Really clean now!

Tested in:

- Editor, all works as expected
- wp-admin, all works as expected.
- support site, all works as expected.
- fieldguide, sandbox, all works as expected (no Help Center).

heavyweight (approved):
Tested with my a8c user, free user and no user
https://wordpress.com/support/
https://wordpress.com/blog/
https://wordpress.com/go/

Tested the admin bar
Pew9g2-qx-p2.php

Tested the editor
Pew9g2-lO-p2.php?post=3&action=edit

Tested the site editor
Pew9g2-N51-p2.php?

Didn't test on atomic since both of my sites are broken and I need to fix them :D

Issue #73838, opened 2023-02-27
Display Thank You page when purchasing a Marketplace theme


epeicher:
Parent task: 
- https://github.com/Automattic/wp-calypso/issues/72703

WBerredo:
On the cart step we will need two additional information on the `extra` object of marketplace themes: 
* `is_marketplace_product ` set to `true`
* the theme slug to mount the thank you page URL

---

Currently when we receive the plugin information on cart we use two infos from the extra object: `is_marketplace_product`
and `plugin_slug`
We use it to identify if we should redirect to the marketplace thank you page and to mount the url with the proper slug

WBerredo:
On the cart step we will need two additional information on the `extra` object of marketplace themes: 
* `is_marketplace_product ` set to `true`
* the theme slug to mount the thank you page URL

---

Currently when we receive the plugin information on cart we have this info on the extra object: `is_marketplace_product`
and `plugin_slug`
We use it to identify if we should redirect to the marketplace thank you page and to mount the url with the proper slug

---

discussion:  p1678098925961309-slack-C02JPCHEKDY

epeicher:
This task will be closed in favor of the following:
- https://github.com/Automattic/wp-calypso/issues/73995
- https://github.com/Automattic/wp-calypso/issues/74295
- https://github.com/Automattic/wp-calypso/issues/74371

PR #74125, opened 2023-03-06
Marketplace: Retrieve themes data on the Thank you page

WBerredo:
Related to #73838

## Proposed Changes

* Retrieve themes data when a theme slug is passed on the Marketplace Thank You page URL. 

* A new selector was added to get data from more than 1 theme.

## Testing Instructions
* This PR shouldn't change any behavior from the prod version
* Install and purchase plugins to make sure everything still working as expected
* Update the URL to add a theme slug and check if still working as expected. From `/marketplace/thank-you/:plugin-slug/:site` to `/marketplace/thank-you/:plugin-slug,:theme-slug/:site`. Ex: `/marketplace/thank-you/woocommerce-subscriptions,yuna/www.test.com.br`

PS: The marketplace theme purchase is not redirecting to the Marketplace Thank you page yet and cannot be tested.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
+++ b/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
@@ -101,7 +101,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 	// Retrieve theme information
 	const wpComThemes = useSelector( ( state ) => getThemes( state, 'wpcom', productSlugs ) );
 
-	const isAllDataFetched =
+	const areAllProductsFetched =
 		!! pluginsOnSite.length &&
 		pluginsOnSite.every( ( pluginOnSite, index ) => !! pluginOnSite || !! wpComThemes[ index ] );
 
@@ -167,7 +167,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 		}
 
 		// Update the menu after the plugin has been installed, since that might change some menu items.
-		if ( isAllDataFetched ) {
+		if ( areAllProductsFetched ) {
 			dispatch( requestAdminMenu( siteId ) );
 			return;
 		}
@@ -177,7 +177,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 		}
 	}, [
 		isRequestingPlugins,
-		isAllDataFetched,
+		areAllProductsFetched,
 		dispatch,
 		siteId,
 		transferStatus,
@@ -191,7 +191,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 			return;
 		}
 
-		setShowProgressBar( ! isAllDataFetched );
+		setShowProgressBar( ! areAllProductsFetched );
 
 		// Sites already transferred to Atomic or self-hosted Jetpack sites no longer need to change the current step.
 		if ( isJetpack ) {
@@ -207,7 +207,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 		} else if ( transferStatus === transferStates.COMPLETE ) {
 			setCurrentStep( 3 );
 		}
-	}, [ transferStatus, isAllDataFetched, showProgressBar, isJetpack ] );
+	}, [ transferStatus, areAllProductsFetched, showProgressBar, isJetpack ] );
 
 	const steps = useMemo(
 		() =>
@@ -323,7 +323,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 			<MasterbarStyled
 				onClick={ () => page( `/plugins/${ siteSlug }` ) }
 				backText={ translate( 'Back to plugins' ) }
-				canGoBack={ isAllDataFetched }
+				canGoBack={ areAllProductsFetched }
 			/>
 			{ productSlugs.map( ( productSlug, index ) => (
 				<QueryTheme key={ 'query-theme-' + index } siteId="wpcom" themeId={ productSlug } />
--- a/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
+++ b/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
@@ -6,6 +6,7 @@ import { useTranslate } from 'i18n-calypso';
 import page from 'page';
 import { useEffect, useState, useMemo, useCallback } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
+import QueryTheme from 'calypso/components/data/query-theme';
 import { ThankYou } from 'calypso/components/thank-you';
 import { ThankYouSectionProps } from 'calypso/components/thank-you/types';
 import { useWPCOMPlugins } from 'calypso/data/marketplace/use-wpcom-plugins-query';
@@ -31,6 +32,7 @@ import { fetchPluginData as wporgFetchPluginData } from 'calypso/state/plugins/w
 import { areFetched, areFetching, getPlugins } from 'calypso/state/plugins/wporg/selectors';
 import isSiteAutomatedTransfer from 'calypso/state/selectors/is-site-automated-transfer';
 import { isJetpackSite } from 'calypso/state/sites/selectors';
+import { getThemes } from 'calypso/state/themes/selectors';
 import { getSelectedSiteId, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 import { ThankYouPluginSection } from './marketplace-thank-you-plugin-section';
 import './style.scss';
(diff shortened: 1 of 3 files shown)


client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -143,7 +151,7 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
 		}
 
 		// Update the menu after the plugin has been installed, since that might change some menu items.
-		if ( areAllPluginsOnSite ) {
+		if ( isAllDataFetched ) {
```
gavande1: Nitpick: Should we use `products` as terminology everywhere as above instead of using `data`?

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -89,24 +91,30 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
...
+	const isAllDataFetched =
+		!! pluginsOnSite.length &&
+		pluginsOnSite.every( ( pluginOnSite, index ) => !! pluginOnSite || !! wpComThemes[ index ] );
 
 	// Consolidate the plugin information from the .org and .com sources in a single list
-	const pluginInformationList = useMemo( () => {
+	const productInformationList = useMemo( () => {
 		return pluginsOnSite.reduce(
 			( pluginsList: Array< any >, pluginOnSite: Plugin, index: number ) => {
 				pluginsList.push( {
```
cpapazoglou: Should we rename `pluginsList` to `productsList`?

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -89,24 +91,30 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
...
 	);
 
-	const areAllPluginsOnSite =
-		!! pluginsOnSite.length && pluginsOnSite.every( ( pluginOnSite ) => !! pluginOnSite );
+	// Retrieve theme information
+	const wpComThemes = useSelector( ( state ) => getThemes( state, 'wpcom', productSlugs ) );
+
+	const isAllDataFetched =
+		!! pluginsOnSite.length &&
+		pluginsOnSite.every( ( pluginOnSite, index ) => !! pluginOnSite || !! wpComThemes[ index ] );
```
cpapazoglou: Do `pluginsOnSite` also return themes? I am not sure why you have added `!! wpComThemes[ index ]` here

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -89,24 +91,30 @@ const MarketplaceThankYou = ( { productSlug }: { productSlug: string } ) => {
...
-	const pluginInformationList = useMemo( () => {
+	const productInformationList = useMemo( () => {
 		return pluginsOnSite.reduce(
 			( pluginsList: Array< any >, pluginOnSite: Plugin, index: number ) => {
 				pluginsList.push( {
 					...wpComPluginsData[ index ],
 					...wporgPlugins[ index ],
 					...pluginOnSite,
+					...wpComThemes[ index ],
+					isTheme: !! wpComThemes[ index ],
```
cpapazoglou: instead of `isTheme` boolean, would it be more useful having a `ProductType` enumeration?

cpapazoglou:
I tested it this and it doesn't seem to introduce any regressions.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: `pluginOnSite` will only have the installed plugins, while `wpComThemes will be the source of truth for themes. 

That said, they preserve the order they were passed via URL, and we need to check both sources to verify if a given position is filled. 

For example:
URL: `/marketplace/thank-you/woocommerce-subscriptions,yuna,wordpress-seo/www.test.com.br`

Slugs: [woocommerce-subscriptions, yuna ,wordpress-seo]
pluginsOnSite: [{woocommerce-subscriptions object}, `undefined`, {wordpress-seo object}]
wpComThemes: [`undefined`, {yuna object}, `undefined`]

---

Does that make sense? Do you have any suggestions to make this logic more explicit or even change it completely?

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: Done https://github.com/Automattic/wp-calypso/pull/74125/commits/9af806d1cc524adee6f3fed98f9e0af52d2b105b

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: Done https://github.com/Automattic/wp-calypso/pull/74125/commits/c089ea9c689002788c5eb26580b92407d99307b7

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
cpapazoglou: Would moving the purchase items (maybe the whole cart before it gets truncated?) to state be helpful?

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: Updated here https://github.com/Automattic/wp-calypso/pull/74125/commits/c69433cc985af19d544643499090efbcd55c7d80

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: We are currently working with this page being stateless, this way it could be refreshed/shared and the user would still be able to see the same results. 

This page is also used for not purchased items (ex: installed plugins) and we handle it the same way as the purchased ones, if we start to work with the purchase objects it would need to be handled as an edge case.

cpapazoglou: > refreshed

We could persist the data in the browser.

> shared

I am not sure we have such a business case?

> This page is also used for not purchased items (ex: installed plugins) and we handle it the same way as the purchased ones

That's a fair point. 

I wonder how we can implement the changes in p9Jlb4-6qr-p2 in the future if we have all products in the URL.

Currently, this means:
- parsing the product slugs
- querying to find out their type
- querying to find out more data based on their type

Am I correct?

WBerredo: > We could persist the data in the browser.

Good point, I haven't considered it.

> I am not sure we have such a business case?

I can't think of one, I just pointed it out.

> Am I correct?

I think its a bit worse request-wise, but not time-wise.
We actually send all requests in parallel and  check what results we get.

Using the example above of 3 slugs, we would send:
* 3 requests to a wporg search
* 3 requests to a wpcom search
* 3 requests for themes

Once we get all of them from at least one source we display the actual thank you page.  

Considering we'll have even more product types in the future, I agree its a good idea to try to find a better alternative to retrieve it.

cpapazoglou: What do you think about using a query string like `?themes=slug1,slug2&plugins=slug1,slug2&plan=slug`?

gcsecsey: +1 to this, I think it'd be great to be able to reduce the number of queries by indicating the product type in the URL

gcsecsey (commented):
I also didn't find any regressions compared to the prod version when testing this.

However, I did find a strange issue on the plugin page, the plugins query is sent repeatedly, and the plugin name is changing back and forth:
CleanShot 2023-03-08 at 11 30 00

~~I'll add an issue to investigate this further.~~ Added https://github.com/Automattic/wp-calypso/issues/74200

WBerredo:
Thanks @gcsecsey for pointing out this issue. I'll check it later but its probably caused by receiving the info from one source and it getting overridden by another source, like `pluginOnSite` and `wpOrgPlugins`.

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
WBerredo: > What do you think about using a query string like `?themes=slug1,slug2&plugins=slug1,slug2&plan=slug`?

I think we can work with it. I can't see another way to better define the type from the URL. 

---

That said, I think this PR can be merged for retrieving the theme data and I'll create a new PR for updating how we receive the info from the URL.

cpapazoglou: Ok, let's iterate on it.

gcsecsey (approved):
Latest changes LGTM, and work as described 🎉 🚢 

I've tested with a WP.org plugin, a WC.com plugin, and a theme slug, and the plugins are shown on the thank you page just as on trunk ✅ 
`/woocommerce-subscriptions,contact-form-7,yuna/gcsecseyatomic.business`
CleanShot 2023-03-14 at 13 14 19@2x
On the Network tab I could also check that the theme data is being fetched ✅

PR #74138, opened 2023-03-06
Add ref for reader signup link

roo2:
As part of making reader tag pages public, we want to track what impact this has on wordpress.com signups.

> Figure out a way to create a funnel report of folks coming from logged out tag pages that then lead to signup. Could be tricky since initially they’ll be logged out.

It will work with this tracks funnel pe7F0s-A6-p2#comment-684

## Proposed Changes

Adds `?ref=lp-reader` to redirects for logged out reader.
`?ref=lp-reader` was chosen because this is already appended to the signup link in the header. It stands for "landing-page-reader" and is the same convention used by `/themes` and other logged out pages.

## Testing Instructions

The following situations should redirect you to /start/account/user?redirect_to=/tag/fishing`&ref=reader-lp`
* Clicking action buttons
* Clicking follow tag
* Clicking a users profile to go to their feed
* The signup button in the header

### DIFF
--- a/client/lib/paths/index.js
+++ b/client/lib/paths/index.js
@@ -22,6 +22,6 @@ export function newPost( site ) {
 	return '/post' + sitePath;
 }
 
-export function createAccountUrl( { redirectTo } ) {
-	return `/start/account?redirect_to=${ redirectTo }`;
+export function createAccountUrl( { redirectTo, ref } ) {
+	return `/start/account?redirect_to=${ redirectTo }&ref=${ ref }`;
 }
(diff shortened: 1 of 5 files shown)


client/lib/paths/index.js
```diff
@@ -22,6 +22,6 @@ export function newPost( site ) {
 	return '/post' + sitePath;
 }
 
-export function createAccountUrl( { redirectTo } ) {
-	return `/start/account?redirect_to=${ redirectTo }`;
+export function createAccountUrl( { redirectTo, ref } ) {
```
tyxla, line 25: I've confirmed that this PR updates all `createAccountUrl` usages to add a `ref` argument 👍

tyxla (approved):
Looks good from my perspective 👍

eoigal (approved):
lgtm

Issue #72545, opened 2023-01-25
Stats: Fix indexing error in inline help


a8ck3n:
The inline help on the Stats page throws an indexing/key error.

The issue is that two of the links in the help overlay have the same underlying URL and this is used as the key for React rendering. See the following link for the line of code:

https://github.com/Automattic/wp-calypso/blob/trunk/client/blocks/inline-help/inline-help-search-results.jsx#L145

Should review and update to use unique keys.

### Steps to Reproduce

1. Test locally. The error message isn't printed in production.
2. Visit any P2 page starting on My Home.
3. Open the dev console.
4. Switch to the Stats page.
5. Confirm you see the following in the console…

![Image](https://user-images.githubusercontent.com/40267301/214482168-660e86d6-8943-40e1-9650-93a0b41db453.png)

The two help items involved are shown in this screenshot…

![Image](https://user-images.githubusercontent.com/40267301/214482279-ebeb47e0-6e62-4fa6-8775-519fa1af8a1a.png)

kisankumavat85:
Hi @a8ck3n, After changing `key={ link ?? title }` to `key={ title ?? link }` solved the issue so can I create a pull request for that?

a8ck3n:
Hey @kisankumavat85, thanks for having a look. I think your proposal sounds reasonable!

kisankumavat85:
@a8ck3n, When I am committing my change I get this error. I don't know what this means is, Actually I am new to open source contribution so it will be helpful if you can guide me.

![image](https://user-images.githubusercontent.com/76833891/215544963-c0a5dc4e-71cd-4a7c-a3ed-7f5d858270c2.png)

![image](https://user-images.githubusercontent.com/76833891/215544988-8a2f32f2-0a76-4b50-a3d2-a2f92ec9f123.png)

a8ck3n:
Looks like an error with your VS Code setup. Specifically, Prettier is failing per this part of the error output:

> Prettier formatting staged file: […]
> '.' is not recognized…

I'm not running Windows so I can't speak to a fix. I'd recommend hitting Google or [Stack Overflow](https://stackoverflow.com/questions/48321639/error-is-not-recognized-as-an-internal-or-external-command-operable-program-o) for environment issues like this. Copying the error into Google is probably your best bet!

PR #74154, opened 2023-03-07
Inline Help: Prefer title over link value for React keys.

a8ck3n:
## Proposed Changes

Prefer the title value over the link value when setting up results listing so as to avoid collisions when the URL is the same.

This fixes a console error on the Stats page due to two of the surfaced help articles having the same URL. Testing across top-level pages doesn’t reveal any further key collisions.

See #72545 for details on reproducing the error.

## Testing Instructions

1. Launch the Calypso Live link. (The error message isn't printed in production.)
2. Visit any P2 page starting on My Home.
3. Open the dev console.
4. Switch to the Stats page.
5. Confirm the React error is no longer present in the console.

## Pre-merge Checklist

### DIFF
--- a/client/blocks/inline-help/inline-help-search-results.jsx
+++ b/client/blocks/inline-help/inline-help-search-results.jsx
@@ -142,7 +142,7 @@ function HelpSearchResults( {
 		};
 
 		return (
-			<Fragment key={ link ?? title }>
+			<Fragment key={ title ?? link }>
 				<li className="inline-help__results-item">
 					<div className="inline-help__results-cell">
 						<a


flootr (approved):
Hey! I think this works as a quick fix.

Things I've noticed:
- the `result` attribute also exposes `post_id` which might be suited even better for a `key`
- I wonder why there are two links with the same article in the first place. Also it seems like the second link doesn't load for me (the request comes back with a `500` for me, can you reproduce that?) – might be worth checking.

a8ck3n:
Thanks for the quick review @flootr!

> the result attribute also exposes post_id which might be suited even better for a key

I saw that as well. For some reason, my initial thought is we'd be more likely to see a collision there. Neither option is great if the API isn't enforcing any uniqueness in the results though. Oddly, despite both posts existing on the target site, the API is returning the same URL for both sets of results.

- Post 3307 — https://wordpress.com/support/?page_id=3307
- Post 132186 — https://wordpress.com/support/?page_id=132186

> I wonder why there are two links with the same article in the first place. Also it seems like the second link doesn't load for me (the request comes back with a 500 for me, can you reproduce that?) – might be worth checking.

I wasn't seeing this in my testing. I tried again from a non-a11n account and do see the error now. I'll do a little more digging to see what's happening here.

**Update:** Further digging shows that the second post has a redirect in place to the first post so it was only visible via direct access if you had admin privileges on the site. Seems the API for fetching these results is not reflecting what visitors to the support site would actually see if attempting to view that specific URL.

I set the post to private and then back to public and that appears to have fixed the errors in the inline help viewer.

flootr:
Thanks @a8ck3n for elaborating!

PR #74164, opened 2023-03-07
Jetpack Checkout: add purchase warning for JP Social overlapping plan in cart

TheSteveK:
Asana task: 1202411927350212-as-1203885046675411/f

## Proposed Changes

* In Calypso checkout for Jetpack products, there's a component that checks if the plan in the user's cart has features the owner already used  and displays a message if so.
* It was discovered Social did not display the message.
* There were two issues causing that, first the jetpack_complete plan does not return Jetpack Social (or Boost FWIW) as a plan feature.  Second, the logic in the cart warning component was hardcoded for each individual overlapping product, and those weren't added.
* This PR does two things:
* Refactor the component code so that rather than handpicking what features trigger the message, any match between the plan and the site will trigger the warning.
* Next because the complete product wasn't returning items that are products and not featured, I needed to get that data to come along with the feature data.  There is a get allFeaturesForPlan that returns the plan data, and Products is a viable part of the underlying object, but that function did not call features.
* Rather than just adding that to th existing function, I added a new function to main.ts that returns the Products only so it can be used in conjunction with the Features but not impact any other code.
* To be safe I added a Unit test for the new function so we'll be able to find out if something drops out of the product list.

Screenshot 2023-03-14 at 12 40 02 PM

## Testing Instructions

* You'll want a JN or other site that already has Social (that's what this PR addresses but it would also apply to boost) already.
* Add Jetpack Complete to your cart, and notice the lack of a warning
* Head on over to the Calypso Live Link (or roll your own with this PR) and append the slug /checkout/[your site here] and you should see the warning at the top of the cart
* Bonus round you can play with any other combination, like social and complete, akismet and security and so on.  Any combo should work and show the banner if there's overlap in the site products and plan.
* If you feel like running the unit test `yarn test-packages packages/calypso-products/test/get-all-products-for-plan.ts`

## Pre-merge Checklist


packages/calypso-products/src/main.ts
```diff
@@ -199,6 +200,9 @@ export function getAllFeaturesForPlan( plan: Plan | string ): TranslateResult[]
 		...( 'getIncludedFeatures' in planObj && planObj.getIncludedFeatures
 			? planObj.getIncludedFeatures()
 			: [] ),
+		...( 'getProductsIncluded' in planObj && planObj.getProductsIncluded
+			? planObj.getProductsIncluded()
+			: [] ),
```
elliottprogrammer: Hmm..  This feels like a scary change here to me...  This function (`getAllFeaturesForPlan(...)`) is used by the function `planHasFeature(..)` & `planHasAtLeastOneFeature(..)` which is utilized all throughout Calypso in various places ( like approx. used over 60 different times spread throughout over ~53 different files).
It definitely gives me uncertainty that this would not cause any undesired bugs in some of the other places where `planHasFeature(..)` is used throughout Calypso.
Possibly rather than changing this function (`getAllFeaturesForPlan(..)` here in the shared `calypso-products` package, do you think we could retrieve the same results by creating another custom version of the function in some utils directory (or whatever) that merges the results of `PlanHasFeature(..)` and `planObj.getProductsIncluded()`, and use that instead?

elliottprogrammer (changes requested):
I tested this and it definitely works! Nice work @TheSteveK ! 🙂  This could definitely prevent some HE tickets.
BTW, is there an Asana task or gh issue associated to this fix/PR?

Markup 2023-03-09 at 10 13 39


However I left a comment about the possible consequences of changing a vital function within the `calypso-products` _package_. I'm thinking we could create a function to accomplish the results you need, but without changing the `getAllFeaturesForPlan(..)` function within the shared calypso-products package. (More details in code comment below).

Also, although this PR works and does what it's intended to do,  I think this might be better accomplished using the **shopping cart validator endpoint** ( fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Snqzva%2Qcyhtvaf%2Sjcpbz%2Qovyyvat%2Sqrsnhyg%2Qfubccvat%2Qpneg%2Qinyvqngbe.cuc%3Se%3Qso85922o%2322-og )  in the backend.  For better context about this, I feel it would be important for you to read some previous discussion here: p1HpG7-fLo-p2#comment-54066

And as an example, here is a diff: ( D96776-code ) that Viknesh did, using the shopping cart validator, that prevents a wide spectrum of overlapping jetpack products (although I don't think Social was a product at that time yet, so it was not added in at that time). 👍

However, doing it this way on the front-end works for now, and if there is some priority on this fix, maybe it would be ok to go ahead with it and ship. (Sorry if I'm sending crossed/confusing answer/message here, the truth is I'm not completely sure what would be better, I'm leaning toward the shopping cart validator on the back-end which also has much more cart modification functionality.

I know @sirbrillig is always very busy, but it would be **really great** to get his opinion here.

.

packages/calypso-products/src/main.ts
TheSteveK: Yeah that's for sure one of the things I was unsure about and wanted advice.  I'll take a look at that approach as well as seeing what I can get out of the backend.  Thanks for taking a minute to take a look.

TheSteveK:
Discarding what I have, will open a new PR

TheSteveK:
I've re-opened this PR.  There was a slack discussion on how to approach fetching a product that's part of a plan (and also the revelation that there's a bug in the pricing page because of products included in plans) p1678717256991079-slack-C03V4PXKUGP

TheSteveK:
pushed a refactor based on the previous comment.

One question I need help with, why the heck is my linter complaining about getAllProductsForPlan not being a valid export? the new function is exported in main.ts and the index file just exports * from all of the files in @automattic/calypso-products

sirbrillig:
> why the heck is my linter complaining about getAllProductsForPlan not being a valid export? the new function is exported in main.ts and the index file just exports * from all of the files in @automattic/calypso-products

I'm not sure but I know that sometime when I change something in a package I have to run `yarn run build-packages` and then restart my language server (or just restart my editor) to get it to notice the new exports.

packages/calypso-products/src/main.ts
```diff
@@ -174,6 +174,7 @@ export function planHasAtLeastOneFeature( plan: string | Plan, features: string[
  */
 export function getAllFeaturesForPlan( plan: Plan | string ): TranslateResult[] {
 	const planObj = getPlan( plan );
+	// console.log( 'planObj: Products Included', planObj.getProductsIncluded() );
```
sirbrillig: I'm sure you plan to but so that we don't forget: we probably want to remove this and the other console.logs in this file.

TheSteveK: Always a good reminder!  The code's not ready for prime time yet, but hopefully I'll remember to scrub those when I'm done with them.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -72,35 +71,41 @@ const PrePurchaseNotices = () => {
 		return products.filter( ( p ) => ! p.expired );
 	} );
 
-	/**
-	 * The active product on the current site that overlaps/conflicts with the plan currently in the cart.
-	 */
-	const siteProductThatOverlapsCartPlan = useSelector( ( state ) => {
+	const siteProductThatOverlapsCartPlan = useSelector( () => {
```
monsieur-z: Since we're not using the `state` anymore, I believe we can get rid of `useSelector`, and maybe use `useCallback` if we want to avoid re-creating the function on each render.

packages/calypso-products/test/get-all-products-for-plan.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	test( 'returns an empty array for Jetpack free', () => {
+		const plan = 'jetpack_free';
+		const result = getAllProductsForPlan( plan );
+		expect( result ).toEqual( [] );
+	} );
+	test( 'Checks for products that should exist in "jetpack_complete" plan', () => {
+		const plan = 'jetpack_complete';
+		const result = getAllProductsForPlan( plan );
+		expect( result ).toContain( `jetpack_backup_t2_yearly` );
```
monsieur-z: I've got a slight concern that the test might break if we ever update the content of the Complete plan. Maybe write an assertion for one or two products only?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -1,13 +1,13 @@
...
-	isJetpackAntiSpam,
 	isJetpackAntiSpamSlug,
-	isJetpackBackup,
 	isJetpackBackupSlug,
 	isJetpackPlanSlug,
-	isJetpackScan,
 	isJetpackScanSlug,
-	planHasFeature,
+	getAllFeaturesForPlan,
+	// eslint-disable-next-line import/named
```
monsieur-z: I think this line will become irrelevant once the PR is merged and the package built. You could make the package changes in a different PR, or delete that line later.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -72,35 +71,41 @@ const PrePurchaseNotices = () => {
...
+			} );
+
+			// Filter the site products to only include those in the plan items
+			const matchingProducts = siteProducts.filter( ( product ) =>
+				planItems.includes( product.productSlug )
+			);
+
+			//special handling for sites with any social product
+			if ( socialProductInSite ) {
+				matchingProducts.push( socialProductInSite );
```
monsieur-z: If I'm understanding things correctly, when a user owns Social and has Complete in the cart, `matchingProducts` should already include the Social product.

It seems here that whenever a site has Social or a plan that includes it, we'll show the notice even though it's not Social that's overlapping. I did a test with a site with both Social and Backup, tried to purchase Security, and saw the Social notice in the checkout page, though the overlap is Backup in that case.

monsieur-z (changes requested):
I've added several inline comments. The only blocker I believe is the logic inside `getMatchingProducts` not being quite right.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: fixed

packages/calypso-products/test/get-all-products-for-plan.ts
TheSteveK: fixed

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: yeah it's ok after I rebuilt earlier

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: So as far as the first part goes, it's like that because social products may not be an exact match as the site could have social basic which would be overlapped by social advanced in complete.  There might be a more elegant way to handle that. 

I do see where the logic is broken and it will display social overlapping the Security plan, so I'll work on that.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -1,13 +1,12 @@
...
-	isJetpackAntiSpam,
 	isJetpackAntiSpamSlug,
-	isJetpackBackup,
 	isJetpackBackupSlug,
 	isJetpackPlanSlug,
-	isJetpackScan,
 	isJetpackScanSlug,
-	planHasFeature,
+	getAllFeaturesForPlan,
+	getAllProductsForPlan,
```
sirbrillig: Also worth noting: `import/named` is not applied to TypeScript files, since TS is much better at finding out what actually exists. Maybe too much for this PR (?) but if this were converted to TS then import issues would go away.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: or not....  So I found when I'm not doing useSelector, it triggers the notice without a product (even with the logic fixed on the Security plan) I've tried useMemo and useCallback and nothing at all, nothing passes the correct overlapping product to the notification.  I might mean that I need to pass that some other way by adjusting the prepurchasenotice component itself but that doesn't feel right.  Can you provide some additional insight to the right approach?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -72,35 +70,46 @@ const PrePurchaseNotices = () => {
...
+			// Combine the plan features and products into a single array
+			const planItems = [ ...planFeatures, ...planProducts ];
+
+			// Filter the site products to only include those in the plan items
+			const matchingProducts = siteProducts.filter( ( product ) =>
+				planItems.includes( product.productSlug )
+			);
+
+			//get all possible social slugs from the site
+			const socialProductInSite = currentSiteProducts.find( ( product ) => {
```
sirbrillig: `currentSiteProducts` can be null so this will need to handle that case.

sirbrillig (commented):
When you're working on JS files, a helpful tool is to temporarily add `// @ts-check` to the top of the file. If your editor supports TypeScript, this will cause it to highlight type errors in the file. While most of the errors found this way can be safely ignored, it can help to find actual potential bugs like the one I mentioned above about `currentSiteProducts`.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
monsieur-z: >  I've tried useMemo and useCallback and nothing at all, nothing passes the correct overlapping product to the notification.

Just to clear that out, did you pass `currentSiteProducts` and  `planSlugInCart` as depencies to `useCallback`?

TheSteveK:
It looks like @mattgawarecki has a PR for review out https://github.com/Automattic/wp-calypso/pull/74574 that will properly bring social and boost feature status.  Therefore I'll need to refactor this to not pull the stuff from Jetpack Products after that's landed.

TheSteveK:
Alright, I refactored based on Matt's PR, I also simplified the comparison by using planHasSuperiorFeature to completely do away with custom checks for specific plans, and I fixed the useSelector() bit on the checks to useEffect() with some state to make sure the component refreshes when there are changes.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: Yeah, but I forgot that useCallback turns that into a function, so I had to 

```
if ( siteProductThatOverlapsCartPlan ) {
		return (
			<CartPlanOverlapsOwnedProductNotice
				product={ siteProductThatOverlapsCartPlan() }
				selectedSite={ selectedSite }
			/>
		);
	}
```
	
Now it's all happy.

mattgawarecki: @TheSteveK In general, the rule of thumb I use for determining when to use `useCallback` and when to use `useMemo` is whether or not I'll need to supply an argument that I'll only know later. For example, if I have a variable that'll need to pull a value from an input field, I can create a callback that'll take that value as a parameter. For data I can figure out with no outside intervention, `useMemo` is easier and more performant (fun fact: function calls actually cost a little more processor time).

In your case, I'd suggest converting to `useMemo` so that `siteProductThatOverlapsCartPlan` becomes a variable and eliminates an extra function call.

TheSteveK: You're right, for some reason I thought I had unsuccessfully tried that, but it works just fine.

packages/calypso-products/src/constants/features.ts
```diff
@@ -191,6 +191,8 @@ export const FEATURE_JETPACK_SOCIAL_BASIC = PRODUCT_JETPACK_SOCIAL_BASIC;
 export const FEATURE_JETPACK_SOCIAL_BASIC_MONTHLY = PRODUCT_JETPACK_SOCIAL_BASIC_MONTHLY;
 export const FEATURE_SOCIAL_SHARES_1000 = 'social-shares-1000';
 export const FEATURE_SOCIAL_ENHANCED_PUBLISHING = 'social-enhanced-publishing';
+export const FEATURE_SOCIAL_ADVANCED = PRODUCT_JETPACK_SOCIAL_ADVANCED;
+export const FEATURE_SOCIAL_ADVANCED_MONTHLY = PRODUCT_JETPACK_SOCIAL_ADVANCED_MONTHLY;
```
mattgawarecki: These seem to be duplicated from lines 188-189, don't they?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -72,39 +69,33 @@ const PrePurchaseNotices = () => {
...
-			return currentSiteProducts.find( isJetpackScan );
+			return matchingProducts;
+		};
+
+		const matchingProducts = getMatchingProducts( currentSiteProducts, planSlugInCart );
+		if ( matchingProducts.length ) {
+			return matchingProducts[ 0 ];
 		}
 
 		return null;
```
mattgawarecki: Non-blocking: there's a shorter way to write this 🙂

```suggestion
		return matchingProducts?.[ 0 ];
```

packages/calypso-products/src/constants/features.ts
TheSteveK: 👍

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
TheSteveK: 👍

mattgawarecki:
The code changes make sense in their current state, and I can confirm I saw a warning at checkout when I added Complete to my cart for a site with an existing Jetpack Social subscription.

NOTE: I see here that the warning says "You currently own Social." I would have expected this to say "Jetpack Social," but that's a question for a different PR.

image

mattgawarecki (approved):

PR #74168, opened 2023-03-07
Launchpad: Remove unused Sensei code

edanzer:
### Proposed Changes

The Sensei flow no longer uses the launchpad screen. This PR simply removed unused, Sensei-related code from Launchpad files. 

### Testing

**Review Time: Short**
**Testing Time: Short**

There should be no differences in Launchpad behavior before and after this PR. 

For review, confirm that Sensei related code has been removed cleanly, in a way that will not impact other flows. 

For testing, pick 1-2 flows (link in bio, newsletter, free, videopress, write, build, etc), and go through them to smoke test and confirm no issues.


mattwiebe (approved):
A beautiful red sea! Test fine. 

I was wondering why we didn't remove `SENSEI_FLOW` from `packages/onboarding/src/utils/flows.ts` but it appears that is still being used by the Sensei folks elsewhere. I'm still able to access `/setup/sensei` with this running.

agrullon95 (approved):
Thanks for cleaning this up. I tested the free flow and it's still working as expected.

PR #74171, opened 2023-03-07
Cancel Purchase: Make refund currency more accurate

sirbrillig:
## Proposed Changes

The cancel purchases page uses `formatCurrency()` to format its refund price, however, it assumes that the refund amount is in the same currency as the purchase's currency. This can be inaccurate.

In this diff we modify the displayed price to use the refund amount's currency instead.

Requires D103872-code

Fixes https://github.com/Automattic/payments-shilling/issues/1411

Before:

Screenshot 2023-03-07 at 2 54 57 PM

After:

Screenshot 2023-03-07 at 2 51 26 PM

## Testing Instructions

- You'll need a recent, refundable purchase in calypso. If you do not have one, make a purchase of a plan.
- Switch your user's account to a different currency using Store Admin.
- Visit the `/me/purchases` page and click on that purchase.
- Click the "Cancel purchase" button at the bottom.
- Verify that you see the amount to be refunded displayed at the bottom of the page.
- Verify that the refunded amount's currency symbol matches the ORIGINAL currency your purchase was made in, not the current currency.


michaeldcain (approved):
This works well for me. Thanks for the quick fix!

PR #74026, opened 2023-03-02
Gutenberg: Avoid redirecting to home from within the iframe

mattwiebe:
- The Site Editor back button selector has changed and is broken (when iframed)
- We update the selector for the site editor back button. This allows us to identify when it's clicked, pass a message from the iframe to the top frame of the browser, and redirect properly to my home

## Screenshots
https://user-images.githubusercontent.com/5414230/222586586-a785066f-440c-4f8f-a872-32fe2826b6d2.mp4

## Testing
- Check out branch
- Create site through free onboarding flow wordpress.com/setup/free/intro
- When at the launchpad, click on the "Edit site design" task
- Open the site editor navigation sidebar 
- Click on the back button
- Verify that the url is correct ( does not have `site-editor` in the path )

PR #74178, opened 2023-03-07
Site Editor: fix template sidebar navigation

mattwiebe:
The fix in #74026 introduced a bug where no matter the level of nesting in the Site Editor nav sidebar, clicking the arrow button returns us to Calypso:

https://user-images.githubusercontent.com/195089/223567740-d20b8217-8edf-44c8-a253-4268616e37b0.mov

## Proposed Changes

* better handling of link detection
* get rid of old outdated forms of link detection
* unrelated fix I saw while testing in 9213231abc9eb0c563dfde99cfda7139c175f70a

## Testing Instructions

0. sandbox `widgets.wp.com` and the site whose Site Editor you will test
1. Run `install-plugin.sh wbe fix/site-editor-template-sidebar-navigation` on your sandbox.
2. Click into the nav sidebar as in the video above. Only the top-level link should take you back to the Calypso dashboard.

## Pre-merge Checklist

### DIFF
--- a/apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
+++ b/apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
@@ -698,7 +698,7 @@ async function openLinksInParentFrame( calypsoPort ) {
 		}
 	} );
 	const popoverSlotElem = document.querySelector( '.interface-interface-skeleton ~ .popover-slot' );
-	popoverSlotObserver.observe( popoverSlotElem, { childList: true } );
+	popoverSlotElem && popoverSlotObserver.observe( popoverSlotElem, { childList: true } );
 
 	// Sidebar might already be open before this script is executed.
 	// post and site editors
--- a/apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
+++ b/apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
@@ -1037,7 +1037,7 @@ function handleSiteEditorBackButton( calypsoPort ) {
 	// have to do this event delegation style because the Editor isn't fully initialized yet.
 	document.getElementById( 'wpwrap' ).addEventListener( 'click', ( event ) => {
 		const dashboardLink = select( 'core/edit-site' )?.getSettings?.().__experimentalDashboardLink;
-		// The link has changed. Pray it doesn't change again.
+		// The link has changed. Pray it doesn't change any further.
 		// This is how to find it in Gutenberg 15.2.
 		const isDashboardLink = traverseToFindLink( event.target, dashboardLink );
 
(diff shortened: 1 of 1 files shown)


apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
```diff
@@ -1020,40 +1020,28 @@ function handleInlineHelpButton( calypsoPort ) {
  * @param {MessagePort} calypsoPort Port used for communication with parent frame.
  */
 function handleSiteEditorBackButton( calypsoPort ) {
+	// We use the traversal helper because the target element may be the SVG or an SVG element inside.
```
edanzer, line 1023: Some interesting recursion going on here. I think I understand how/why we're doing this, but I wouldn't mind getting one extra pair of eyes on this before merging. 

Conversely, I know this is a bug fix, so maybe we release first to resolve this issue, but then I still wouldn't mind separate eyes on this after the fact to see if anyone else can anticipate any issues with this method.

apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
```diff
@@ -1020,40 +1020,28 @@ function handleInlineHelpButton( calypsoPort ) {
...
+			depth--;
+		}
+		return foundLink;
+	}
+
 	// have to do this event delegation style because the Editor isn't fully initialized yet.
 	document.getElementById( 'wpwrap' ).addEventListener( 'click', ( event ) => {
-		const clickedElement = event.target;
-
-		const isOldDashboardButton =
```
edanzer, line 1027: I know we've had the references to OldDashBoard button in this file for a while. I did wonder if this was still needed as a backwards compatibility measure for some reason. Sounds like we do not think this is needed any more.

edanzer (approved):

apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
mattwiebe: The recursion saves me from having to manually check the same thing over and over again via `parentElement` as [below](#diff-3cd9fa865f1d9d2531b85d0a08341a655b497a3304dfbb78ff787e34ff80556bL1052-L1053)

apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js
jeyip: As far as performance goes, this linked list traversal doesn’t look super concerning, especially because it’s limited to, at most, two cycles ( depth = 2 ). There's also nothing computationally expensive happening here.

From what I'm reading, the `traverseToFindLink` is identifying an element that has been clicked and checking if it's the back to dashboard button. If not, it checks two levels up to two levels of the ancestral tree ( parent / grandparent ) to see if those might be the dashboard button too.

The overall implementation is still dependent on the location of the `a` tag ( what if it's not a parent _or_ grandparent element, but a great grandparent instead? ), but at this point in time, it's the nature of working with open source Gutenberg.

Thanks for taking care of this @mattwiebe!

PR #74179, opened 2023-03-07
Checkout: Move extra tax field tests to own file

sirbrillig:
## Proposed Changes

As a follow-up to https://github.com/Automattic/wp-calypso/pull/73513, this further splits up the checkout contact details tests to speed things up. The VAT details tests in particular are often timing out so something needs to be improved. In this case we move the tests which have to do with the "extra" tax contact details required by some countries.

In my anecdotal testing on my local machine, running the full set of `checkout-vat-form` tests took around 70 seconds prior to this diff and now running `checkout-vat-form` and the new `checkout-extra-tax-fields` together takes about 40 seconds (due mostly to parallelization; the tests still take about the same amount of time).

## Testing Instructions

Do automated tests pass?


noahtallen (approved):
Worth trying!

PR #74198, opened 2023-03-08
Contact form: change the query sent to the Q&A endpoint

yoavf:
## Proposed Changes

* In the support contact form: don't send the message subject as part of the q&a search.
Results seem to be worse than when only the message itself is sent (see pbvpgB-2MS-p2)

## Testing Instructions
* Visit /help/contact
* Fill in the subject and message
* Check the network request going to the `help/qanda` endpoint only contains the message string.

## Pre-merge Checklist

### DIFF
--- a/client/me/help/help-contact-form/index.jsx
+++ b/client/me/help/help-contact-form/index.jsx
@@ -151,7 +151,7 @@ export class HelpContactForm extends PureComponent {
 		} );
 	};
 
-	getSibylQuery = () => ( this.state.subject + ' ' + this.state.message ).trim();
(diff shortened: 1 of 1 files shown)


klimeryk (approved):
Tested and verified the request only sends the message now. Tried how it behaves with empty input, only whitespace, etc. - all good!

I wonder if a more flexible approach would be to send explicitly subject, message, etc. like `{ subject: "", message: "" }`, so that on the backend we could decide which part we use and in what way (without changes on the frontend). But that might be worth looking into if we switch the backend engine for something smarter (like based on GPT).

PR #74131, opened 2023-03-06
Creating conversion experiment on paid media signup flow (biannual option)

donlair:
Related: #74207 

## Proposed Changes

* Replaces the test setup in #72968 with a followup test that replaces the Monthly option with Bi-annually, and removes the Free plan link.
* The experiment name is `paid_media_signup_2023_03_biannual_toggle_hide_free`
* This experiment affects only the paid media signup flow (`onboarding-pm`)

Here's are screenshots showing the page for both tab options:

CleanShot 2023-03-09 at 13 36 11@2x

## Testing Instructions

* Checkout this PR and start Calypso locally.
* Navigate to the paid media signup flow (`/start/onboarding-pm`)
* Proceed to the Plans step of the signup flow without assigning yourself to a variant.
* Confirm that you see that Plans step with Annual and Monthly toggle options as well as the Free link (note: `onboarding-pm` doesn't use the 2023 design)
* Confirm that you can select a plan and make it to the checkout page as expected.
* Assign yourself to the treatment variant of this experiment and repeat the flow.
* Confirm that you see 1 year and 2 years tab options.
* Click the tabs and confirm that you see the correct prices.
* Click choose a variety of plans and confirm that the correct plan appears on the checkout page.
* Test the `onboarding` flow at `/start` and confirm that the updates to the shared tab code hasn't affected the other flows.

PR #74207, opened 2023-03-08
Implementing new term length toggle option for the plans step

donlair:
Related to #74131 

## Proposed Changes

* This implements the changes discussed at pdgrnI-28f-p2.
* Creates a second set of term length tab options - the choices are now either `[ Monthly, Yearly ]` or `[ 1 Year, 2 Years]`
* This PR does not implement maxDiscount calculations for the new two year option, so it doesn't display that modal for the new set.

Screen shots:

Screenshot 2023-03-08 at 8 11 11 AM

Screenshot 2023-03-08 at 8 06 30 AM

## Testing Instructions

* Checkout this PR
* Force a plans step to show the two year option by adding `shortestTermOption="1year"` as a prop to the `PlanTypeSelector` component (e.g., [here](https://github.com/Automattic/wp-calypso/blob/8b72626d660c1cf6566fc38796da1261aa59323c/client/my-sites/plans-features-main/index.jsx#L623))
* Confirm that the default is still [Monthly, Yearly]

### DIFF
--- a/client/my-sites/plans-features-main/plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/plan-type-selector.tsx
@@ -35,7 +35,7 @@ export type PlanTypeSelectorProps = {
 	siteSlug?: string | null;
 	selectedPlan?: string;
 	selectedFeature?: string;
-	shortestTermOption?: 'monthly' | '1year';
+	showBiannualToggle?: boolean;
 	isInSignup: boolean;
 	plans: string[];
 	eligibleForWpcomMonthlyPlans?: boolean;
@@ -124,7 +124,7 @@ export type IntervalTypeProps = Pick<
 	| 'isPlansInsideStepper'
 	| 'hideDiscountLabel'
 	| 'redirectTo'
-	| 'shortestTermOption'
+	| 'showBiannualToggle'
 >;
 
 export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > = ( props ) => {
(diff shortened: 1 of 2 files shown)


client/my-sites/plans-features-main/plan-type-selector.tsx
```diff
@@ -146,49 +154,46 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > =
 	const isDomainAndPlanPackageFlow = new URLSearchParams( window.location.search ).get(
 		'domainAndPlanPackage'
 	);
+	const shouldHideMonthly = shortestTermOption === '1year';
```
chriskmnds: I think this is creating some additional knowledge here that isn't necessary, at least not currently i.e. that the "shortest term option" being "1year" implies that "monthly" should be hidden. It feels like unnecessary complexity. Why not just keep it simple for now e.g. pass a boolean `showBiannualToggle`? 

I can envision something like `shortestTermOption` being useful for showing more than 2 values e.g. everything from `shortestTermOption` and above, but this is not the case with the rest of the logic. 🤔

chriskmnds (approved):
Tests out. Thanks for putting this together :shipit:

client/my-sites/plans-features-main/plan-type-selector.tsx
donlair: @chriskmnds good call, I'll switch it to that approach. Thanks for the review!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7850250

Thank you @donlair for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7850250).

PR #74208, opened 2023-03-08
Staging Sites: Hide module on current staging sites

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/1921

## Proposed Changes

Hides the 'Staging site' module on 'Hosting Configuration' if the site is a staging site.

We made eventually add some UI for this with https://github.com/Automattic/dotcom-forge/issues/1929 but better to hide for now.

## Testing Instructions

1. Navigate to Hosting Configuration for a production site.
2. Verify the 'Staging site' module is still present.
3. Navigate to Hosting Configuration for a staging site.
4. Verify 'Staging site' module is absent.


client/my-sites/hosting/main.js
```diff
@@ -233,6 +238,7 @@ export const clickActivate = () =>
 export default connect(
 	( state ) => {
 		const siteId = getSelectedSiteId( state );
+		const site = getSelectedSite( state );
```
sejas, line 243: I think that getting the selected site state is making a new request and clearing the state for that site which makes the UI show the Staging card for a few seconds.

We probably need to avoid making that extra request or waiting to be loaded before showing the Staging card

sejas (changes requested):
@katinthehatsite and I have tested this PR and we've found a weird behaviour.

In the screencast the card is visible for few seconds when you click `Manage staging site` or when you refresh the page.

https://user-images.githubusercontent.com/779993/223745332-c7224706-4ae8-4fc5-86de-a3ce1b64484c.mp4

## Steps to reproduce the issue
1. Create a staging site from a new atomic website on Settings > Hosting configuration
2. Click on `Manage staging site` button
3. Go to Hosting configuration in that staging site
4. Observe the Staging card is visible
5. It eventually can disappear after few seconds


@danielbachhuber , should we worry about disabling the backend for staging sites too?
The request would fail, but the user ends up with a new free site created.

client/my-sites/hosting/main.js
sejas, line 243: Probably this can fix the issue.

```suggestion
		const site = getSite( state, siteId );
```

client/my-sites/hosting/main.js
```diff
@@ -174,7 +179,7 @@ class Hosting extends Component {
 		const getContent = () => {
 			const isGithubIntegrationEnabled = isAutomatticTeamMember( teams );
 			const WrapperComponent = isDisabled || isTransferring ? FeatureExample : Fragment;
-			const isStagingSiteEnabled = isEnabled( 'yolo/staging-sites-i1' );
+			const isStagingSiteEnabled = ! isStagingSite && isEnabled( 'yolo/staging-sites-i1' );
```
wojtekn: Would it make more sense to add `isStagingSite` where we check for `isStagingSiteEnabled` before displaying the component? In another case, it can become weird if we remove the `isEnabled` check.

danielbachhuber:
> should we worry about disabling the backend for staging sites too?

@sejas Yep, definitely. I created D104018-code

client/my-sites/hosting/main.js
wojtekn: I see in other places we also do:

```
const site = getSelectedSite( state );
const siteId = get( site, 'ID' );
```

danielbachhuber: I wasn't able to reproduce that issue, but I experienced this behavior:

https://user-images.githubusercontent.com/36432/223862148-65b964db-0d2b-4747-b5b9-4bd58a1b5a07.mp4

```diff
diff --git a/client/my-sites/hosting/staging-site-card/index.js b/client/my-sites/hosting/staging-site-card/index.js
index 0a63c86fb6..49697bce9c 100644
--- a/client/my-sites/hosting/staging-site-card/index.js
+++ b/client/my-sites/hosting/staging-site-card/index.js
@@ -125,7 +125,7 @@ const StagingSiteCard = ( { disabled, siteId, translate } ) => {
 						},
 					} ) }
 				</p>
-				<Button primary href={ `/home/${ stagingSiteSlug }` } disabled={ disabled }>
+				<Button primary href={ `/hosting-config/${ stagingSiteSlug }` } disabled={ disabled }>
 					<span>{ translate( 'Manage staging site' ) }</span>
 				</Button>
 			</>

```

I think what you're describing is likely a problem across this entire page...

client/my-sites/hosting/main.js
danielbachhuber: Sure, did that with 7efdf46b80

PR #74219, opened 2023-03-08
Add GR to the lists of EU countries and countries supporting VAT

martian77:
Related to Automattic/payments-shilling#1436

## Proposed Changes

It turns out that there are two two letter country codes in use for Greece: GR (from  ISO 3166) and EL ([used in the EU](https://publications.europa.eu/code/pdf/370000en.htm)). EL is used in various places to whether to display e.g. the inline vat form, but the checkout code supplies GR instead. 

This adds Greece's 2 letter country code (GR rather than EL) to the list of countries that show the inline VAT details form in checkout. It also adds it to the list of EU countries for displaying the Automattic VAT details. 

## Testing Instructions

* This will need D103796-code and D104012-code applied to a wpcom sandbox to work completely. 
* Go to checkout and set the billing country to Greece. You should see the inline form to add vat details appear. 
* With the two backend diffs applied, adding a valid Greek VAT number here should reduce the tax to zero.
(Without the two backend diffs the form will still show, but there will be no tax displayed and you will be unable to save a Greek VAT number.)
Screenshot 2023-03-08 at 16 54 43

## Pre-merge Checklist

### DIFF
--- a/client/me/purchases/is-country-in-eu.ts
+++ b/client/me/purchases/is-country-in-eu.ts
@@ -14,6 +14,7 @@ export function isCountryInEu( country: string, date?: string ): boolean {
 		'ES',
 		'FI',
 		'FR',
+		'GR',
 		'HR',
 		'HU',
 		'IE',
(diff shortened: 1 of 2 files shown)


sirbrillig (approved):
Works for me. I wonder if we should eventually make these lists come from an endpoint so we don't have to change them in so many places... anyway, a future project.

martian77:
> I wonder if we should eventually make these lists come from an endpoint so we don't have to change them in so many places... anyway, a future project.

I was wondering about that too. Feels like the kind of thing that could do with being controlled from a single place. But as you say, a future project.

PR #74230, opened 2023-03-08
Checkout: Do not validate VAT details in checkout if unchanged

sirbrillig:
## Proposed Changes

In checkout, if you fill out the VAT details form (for a country that supports VAT), the details are always validated and saved to the server via the `/me/vat-info` endpoint. However, you're only allowed to save those details once, and they are always validated before saving. So there's really no need to validate them on every use within checkout. (In fact, if validation fails a second time this presents quite a UX problem since the user cannot edit them after that point!)

In this PR we modify the billing step validation so that it does not submit the details for validation if they are unchanged.

Screenshot 2023-02-15 at 5 46 27 PM

## Testing Instructions

- Open the network panel of your browser's devtools.
- Save a VAT ID to your test account (either using `/me/purchases/vat-details` or Payments Admin or checkout itself).
- Edit the billing details step if it's not already active.
- Make sure you select a country which supports VAT.
- Make sure the "Add Business Tax ID details" checkbox is checked and the VAT details are pre-filled.
- Click "Continue" to complete the step. We need to do this once first to prepare for the test.

Now for the actual test...

- Edit the step again.
- Click "Continue" to complete the step.
- Verify that there is **no POST request** made to the `/me/vat-info` endpoint.
- Edit the step again and change one of the VAT fields like "Address for tax ID" and click "Continue".
- Verify that there is **a successful POST request** made to the `/me/vat-info` endpoint with the updated address.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/test/checkout-extra-tax-fields.tsx
+++ b/client/my-sites/checkout/composite-checkout/test/checkout-extra-tax-fields.tsx
@@ -35,6 +35,13 @@ jest.mock( 'calypso/state/products-list/selectors/is-marketplace-product' );
 jest.mock( 'calypso/lib/navigate' );
 
 describe( 'Checkout contact step extra tax fields', () => {
+	// These tests seem to be particularly slow (it might be because of using
+	// it.each; it's not clear but the timeout might apply to the whole loop
+	// rather that each iteration?), so we need to increase the timeout for their
+	// operation. The standard timeout (at the time of writing) is 5 seconds so
+	// we are increasing this to 8 seconds.
+	jest.setTimeout( 8000 );
+
 	const mainCartKey: CartKey = 'foo.com' as CartKey;
 	const initialCart = getBasicCart();
 	const defaultPropsForMockCheckout = {
--- a/client/my-sites/checkout/composite-checkout/test/checkout-extra-tax-fields.tsx
+++ b/client/my-sites/checkout/composite-checkout/test/checkout-extra-tax-fields.tsx
@@ -34,14 +34,14 @@ jest.mock( 'calypso/lib/analytics/utils/refresh-country-code-cookie-gdpr' );
 jest.mock( 'calypso/state/products-list/selectors/is-marketplace-product' );
 jest.mock( 'calypso/lib/navigate' );
 
-describe( 'Checkout contact step extra tax fields', () => {
-	// These tests seem to be particularly slow (it might be because of using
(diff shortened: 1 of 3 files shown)


michaeldcain (approved):
The code LGTM and this works as described.

pottedmeat:
Code looks good. I'll defer to Cain's testing.

client/my-sites/checkout/composite-checkout/test/checkout-extra-tax-fields.tsx
```diff
@@ -35,6 +35,13 @@ jest.mock( 'calypso/state/products-list/selectors/is-marketplace-product' );
 jest.mock( 'calypso/lib/navigate' );
 
 describe( 'Checkout contact step extra tax fields', () => {
+	// These tests seem to be particularly slow (it might be because of using
+	// it.each; it's not clear but the timeout might apply to the whole loop
+	// rather that each iteration?), so we need to increase the timeout for their
+	// operation. The standard timeout (at the time of writing) is 5 seconds so
+	// we are increasing this to 8 seconds.
+	jest.setTimeout( 8000 );
```
noahtallen: Since this impacts everything in the file, could we set it in the top-level outside of the describe block?

sirbrillig: No problem.

noahtallen: 🙇 thanks!

cintacks:
Hi y'all - can we get some eyes on why this is still failing for this user? 6026786-zd-woothemes

To replicate:
- Attempt checkout with BP included in cart
- Empty cart and try again, issue persists
- VAT details have been checked and validated 

I can confirm that the issue is still present. Please advise - thanks!

sirbrillig:
@cintacks I don't have access to that ticket right now but this makes me wonder about several things 🤔 

Firstly, if the user is getting an error that their VAT details are invalid during checkout, then that means that the validation is running, which in turn means that they've changed something about their VAT details within checkout. For example, if they edit their address or organization name that would cause validation to run. It would also run validation if they changed their country code. Something must be different than what is saved when they visit https://wordpress.com/me/purchases/vat-details. 

The second thing this makes me wonder is: last week we believed that the validation issue was some sort of downtime in the validator, but it seems unlikely that's still going on. If there's a validation failure now, something else must be going on where perhaps invalid VAT details are being submitted by the form. 

In order to research this further, I've created https://github.com/Automattic/payments-shilling/issues/1451

PR #74267, opened 2023-03-09
Fix event recording for `calypso_plan_features_upgrade_click`

aneeshd16:
Related to p1678294085915499-slack-CFFF01Q4V

## Proposed Changes

* The `calypso_plan_features_upgrade_click` event is not recorded correctly. This is because the `recordTracksEvent` function imported from `calypso/state/analytics/actions` is an action generator meant to be used along with `dispatch`. This PR changes the import so that the function is imported from `calypso/lib/analytics/tracks`, which does not need to be wrapped in `dispatch`.
* While working on this, I noticed that the `current_plan` event prop was passed as `null`. This change also replaces `null` with the actual current plan slug.

This change is a "band-aid" solution and does not fix the root problem - it is easy to use the `recordTracksEvent` from the wrong file. The proper fix is to probably rename the `recordTracksEvent` from `calypso/state/analytics/actions`.

## Testing Instructions

* Go to `/plans/<site slug>` where the selected site is on any plan lower than the eCommerce plan. (free is also okay)
* Open the network inspector and switch to the "Images/Img" tab.
* Click on the "Upgrade" button for any plan.
* Confirm that you see a network request to `pixel.wp.com`.
* Open the network request and confirm that the payload contains the correct event props.
image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/actions.tsx
+++ b/client/my-sites/plan-features-2023-grid/actions.tsx
@@ -14,7 +14,7 @@ import classNames from 'classnames';
 import { localize, TranslateResult, useTranslate } from 'i18n-calypso';
 import { useSelector } from 'react-redux';
 import ExternalLinkWithTracking from 'calypso/components/external-link/with-tracking';
-import { recordTracksEvent } from 'calypso/state/analytics/actions';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import getDomainFromHomeUpsellInQuery from 'calypso/state/selectors/get-domain-from-home-upsell-in-query';
 import { Plans2023Tooltip } from './plans-2023-tooltip';
 import { WPCOM_PLANS_UI_STORE } from './store';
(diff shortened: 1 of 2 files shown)


client/my-sites/plan-features-2023-grid/actions.tsx
```diff
@@ -278,7 +278,7 @@ const PlanFeaturesActionsButton: React.FC< PlanFeaturesActionsButtonProps > = (
 
 		if ( ! freePlan ) {
 			recordTracksEvent( 'calypso_plan_features_upgrade_click', {
-				current_plan: null,
+				current_plan: currentSitePlanSlug,
```
chriskmnds, line 281: I guess we could also add this in the old comparison grid action [below](https://github.com/Automattic/wp-calypso/pull/74267/files#diff-353909e3a7048cba30b26145d5207238876cd67d255732af7b69ed3c49ee2d39R42). but maybe no need to mess with that right now.

chriskmnds (approved):
Tests out as per the instructions 🚢 
Thanks for the detailed description and instructions! 🦸

client/my-sites/plan-features-2023-grid/actions.tsx
aneeshd16: Oh yes, I missed mentioning this in the PR. I skipped that change because the old grid is no longer in use.

aneeshd16:
The E2E test failure is unrelated and is being discussed in p1678363478775119-slack-C02DQP0FP.

PR #74279, opened 2023-03-09
Server Bundler: Enable cache for assets on the development env

arthur791004:
Related to #

## Proposed Changes

* Enable cache for file assets so that those assets won't request every time. This is especially for the `gridicon` since the calypso will request svg and it leads the icon might become white when you are back and forth between different pages.

image

## Testing Instructions

* Go to `/setup/site-setup?siteSlug=<your_site>`
* Back and forth between different steps, and ensure the back icon, `<`, on the top-left corner won't be flashing

## Pre-merge Checklist

### DIFF
--- a/client/server/bundler/index.js
+++ b/client/server/bundler/index.js
@@ -100,7 +100,15 @@ function middleware( app ) {
 	}
 
 	app.use( waitForCompiler );
-	app.use( webpackMiddleware( compiler ) );
+	app.use(
+		webpackMiddleware( compiler, {
+			headers: ( req, res ) => {
+				if ( /\.(?:gif|jpg|jpeg|png|svg)$/i.test( req.originalUrl ) ) {
+					res.setHeader( 'Cache-Control', 'max-age=3600' );
+				}
+			},
+		} )
+	);
 }
 
 module.exports = middleware;


client/server/bundler/index.js
```diff
@@ -100,7 +100,15 @@ function middleware( app ) {
 	}
 
 	app.use( waitForCompiler );
-	app.use( webpackMiddleware( compiler ) );
+	app.use(
+		webpackMiddleware( compiler, {
+			headers: ( req, res ) => {
+				if ( /\.(?:gif|jpg|jpeg|png|svg)$/i.test( req.originalUrl ) ) {
+					res.setHeader( 'Cache-Control', 'max-age=3600' );
```
noahtallen, line 107: Would a shorter cache-control length still work? (e.g. would like 5-10s solve the problem?)

I think we do generally want to avoid caching these in dev mode so that people can test changes more easily

arthur791004: I think it's fine for the file assets as we set the hash to their name. Hence, when we update the file, the cache will be busting automatically. Does it make sense?

tyxla: Another option could be to have a flag to enable this. Then folks who would like to develop with file asset cache enabled could just run Calypso development with that flag enabled.

tyxla (commented):
I'm not completely sure I understand the motivation behind this PR. Anything that contains "enabling cache" and "development env" in the same sentence raises a red flag. Sometimes it's fine, like when it saves minutes of babel transpilation and webpack compilation time, but in this case, I don't see a justification.

I see this as the rationale:

> calypso will request svg and it leads the icon might become white when you are back and forth between different pages.

however, I'd be totally fine to have this behavior on development and have the benefit of seeing changes immediately, rather than having unexpected cache confuse developers as to why their changes aren't considered. 

Perhaps we could explore @noahtallen's suggestion?

I've also suggested another alternative - adding a flag to enable the cache, essentially making that cache opt-in.

arthur791004:
Thanks for your feedback! I plan to close this one and re-think whether we still need this feature in the future 🙂

PR #74301, opened 2023-03-09
Fix the FSE template edit e2e test

noahtallen:
Noticed this test permafailing on trunk. Looks like the editor loads differently on the new gutenberg version, where it's zoomed into the index template instead of starting more zoomed out. As a result, to edit the "page" template, we need to click the "back" button first.

Note that on the e2e test site, it seems to start zoomed into the "navigation" submenu, which is kinda weird. I'm not sure how sustainable this test will be, but it's breaking everywhere right now so a quick fix seems prudent.

## Testing Instructions
e2e should pass


noahtallen:
Hm, still failing, but not sure why. The little "<" button has an aria-label of `Back`. Inspecting it with the accessibility debugger, it has the name "Back", which means we should be able to locate it like this...

noahtallen:
The test logs indicate that it successfully clicked the back button, but the screenshot and video show that nothing happened

youbrokesomething (approved):
LGTM 🚢

PR #74304, opened 2023-03-09
Checkout: Increase timeout to 12s for all checkout tests using `it.each`

sirbrillig:
## Proposed Changes

There are several checkout tests which use Jest's `it.each()` (aka `test.each()`) architecture to iterate over test cases. It's not totally clear but it's possible that the default timeout for a test applies to the whole loop of tests and not a single iteration. In any case, these tests seem to time out frequently, so this PR increases the timeout to 12 seconds for those test files.

## Testing Instructions

Do automated tests pass?


noahtallen (approved):
Let's try it 👍

PR #73503, opened 2023-02-17
Flaky E2E: disable shared memory on Chromium.

worldomonation:
Related to https://github.com/Automattic/wp-calypso/issues/70674.

## Proposed Changes

This PR opts to disable the usage of shared memory for Chromium entirely.

Despite multiple attempts to eliminate the page crashing from occurring on some suites (documented in https://github.com/Automattic/wp-calypso/issues/70674), this issue persists.

image

On upstream repo, it appears most hits for the query `shm` comes from the `--disable-dev-shm-usage` flag as part of the user's command. Seeing how prevalent it is, I think it's time for us to go back to not using shared memory in the first place to see if this has a positive impact on browser crash rates.

## Testing Instructions

Any E2E build should run and pass.

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #74309, opened 2023-03-09
Revert "Flaky E2E: disable shared memory on Chromium."

worldomonation:
Reverts Automattic/wp-calypso#73503

I don't think the previous change above did much to address the crashes; in fact, I suspect the incidence has increased. Take a look at the trend recently below:

image


youbrokesomething (approved):
LGTM 🚢 

Too bad it didn't work :(

PR #74311, opened 2023-03-10
Billing History: Add Card displaying link to Taxamo receipt

spraveenitpro:
### Summary:

We need to add a link to the Taxamo tax invoice to the bottom of a user's individual receipt in billing history. We will use the external link card component  for this purpose.

Related to [#](https://github.com/Automattic/payments-shilling/issues/1341)

## Proposed Changes

* Add a separate card component that will have the link to the taxamo link from the Billing page.
* Requires D104748-code to attach tax_external_id to receipts
* Important: ensure that the order object does have the ` [tax_external_id]` set up as only then it will display the card component with the link to the external receipt page. You can confirm this by viewing the order object for that receipt from payments admin.

## Testing Instructions
1. Before applying the patch, you should not see a card component with the link to the tax receipt like this 
2. Apply this Patch along with the backend diff D104748-code
1. Visit a specific receipt page from `/me/purchases/billing/{receipt}` for a receipt that has vat id saved and for a country that is eligible for taxes.
3. You should now see a `View tax invoice` card component like this:   
4. Clicking that view tax invoice link should open an external tax receipt that looks like this

### DIFF
--- a/client/state/billing-transactions/types.ts
+++ b/client/state/billing-transactions/types.ts
@@ -14,7 +14,7 @@ export interface BillingTransaction {
 	address: string;
 	amount: string;
 	tax_country_code: string;
-	tax_external_id: string;
+	tax_external_id?: string;
 	cc_email: string;
 	cc_name: string;
 	cc_num: string;
--- a/client/state/billing-transactions/types.ts
+++ b/client/state/billing-transactions/types.ts
@@ -14,6 +14,7 @@ export interface BillingTransaction {
 	address: string;
 	amount: string;
 	tax_country_code: string;
+	tax_external_id: string;
 	cc_email: string;
 	cc_name: string;
 	cc_num: string;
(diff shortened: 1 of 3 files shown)


client/state/billing-transactions/types.ts
```diff
@@ -14,6 +14,7 @@ export interface BillingTransaction {
 	address: string;
 	amount: string;
 	tax_country_code: string;
+	tax_external_id: string;
```
sirbrillig: Because the server is not quite yet providing `tax_external_id` (and because it can be null currently; I've left a comment on the patch), it might help if we set this type to be optional so that TypeScript can protect us in case it's not set. You're already correctly guarding for it with the code `transaction.tax_external_id && transaction.tax_external_id.length > 0` but this way TS will also help us out.

```suggestion
	tax_external_id?: string;
```

spraveenitpro: Thank you, great catch! 🙌🏻

sirbrillig (approved):
Looks good to me! Since the new property is optional, this can be merged before D104748-code.

PR #74314, opened 2023-03-10
Storage Upgrade: update offer

monsieur-z:
### Changes proposed in this Pull Request

This PR updates the products that are shown on the storage upgrade page to show Backup addons only (instead of plans).

Fixes 1182633034225700-as-1204031357952912.

### Implementation notes

It's kind of hard to comment on all changes since the code required a fair amount of changes to account for add-ons. Here are some pointer that I hope will help:
- Defined upgrade paths in the `UPGRADES`  map `use-get-storage-upgrade-products.ts`.
- Added new hooks in `packages/calypso-products/src/translations.tsx` to facilitate translations of storage amount.
- Removed the duration check from the storage page, since the default never changed anyway (the route doesn't consider duration). Upgrade relationships are defined in the `UPGRADES` map instead.
- Moved all copy in the `useGetProductCardData` hook.
- Removed the mentions of storage amount and backup days in the heading and subheading since they now depend on the add-ons being shown. This could be added back in a subsequent PR.

### Testing instructions

- Make sure you have a Jetpack site with a monthly Jetpack Backup 10 GB subscription
- Spin up Calypso by running this branch locally, or clicking on a live link below
- Visit `plans/storage/:site`
- Check that you see the 10GB, 100GB, and 1TB addons upgrades
- Verify that the buttons behave as expected
- Pick a couple of monthly products in the table below, and check that you see the correct upgrades (one should not be able to upgrade from a monthly to a yearly plan, nor buy the add-on they already own)

| Owned product | Upgrades |
|-----------------|-----------|
| Jetpack VaultPress Backup 10GB (monthly or yearly) | Add-on 10GB, Add-on 100GB, Add-on 1TB |
| Jetpack VaultPress Backup 1TB (monthly or yearly) |	Add-on 1TB, Add-on 3TB, Add-on 5TB |
| Add-on Storage 10GB |	 Add-on 100GB, Add-on 1TB |
| Add-on Storage 100GB | Add-on 1TB, Add-on 3TB |
| Add-on Storage 1TB |Add-on 3TB, Add-on 5TB |
| Add-on Storage 3TB |Add-on 5TB |
| Add-on Storage 5TB | – |
| Security 10GB (monthly or yearly) | Add-on 10GB, Add-on 100GB, Add-on 1TB |
| Security 1TB (monthly or yearly) |	Add-on 1TB, Add-on 3TB, Add-on 5TB |

_Note: there may be some styling issues. This is due to styles not being loaded in the same order locally and in staging. Fixing this is not in the scope of this PR._

### Screenshots

_Before_
Screenshot 2023-03-13 at 3 46 03 PM

_After_

Screenshot 2023-03-16 at 8 26 50 AMScreenshot 2023-03-16 at 8 26 57 AM


packages/calypso-products/src/translations.tsx
```diff
@@ -1013,25 +1013,63 @@ export const useJetpack1TbStorageAmountText = (): TranslateResult => {
...
+	const FIVE_TERABYTE = useJetpackTbStorageAmountText( 5 );
+
+	return useCallback(
+		( productSlug ) =>
 			( {
 				[ PRODUCT_JETPACK_BACKUP_T1_MONTHLY ]: TEN_GIGABYTES,
 				[ PRODUCT_JETPACK_BACKUP_T1_YEARLY ]: TEN_GIGABYTES,
 				[ PRODUCT_JETPACK_BACKUP_T2_MONTHLY ]: ONE_TERABYTE,
 				[ PRODUCT_JETPACK_BACKUP_T2_YEARLY ]: ONE_TERABYTE,
+				[ PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_10GB_MONTHLY ]: TEN_GIGABYTES,
```
monsieur-z, line 1063: There's no yearly addon.

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-storage-upgrades-to-display.ts
```diff
@@ -1,36 +1,28 @@
...
-			TIER_2_SLUGS.includes( productSlug )
+		// Always show from the lowest to the highest tier
+		availableUpgrades.sort(
+			( a, b ) => UPGRADES_ORDER.indexOf( a.productSlug ) - UPGRADES_ORDER.indexOf( b.productSlug )
 		);
 
 		const upgradesToDisplay = [
 			...purchasedUpgrades,
 			// Don't show tier 1 if tier 2 has already been purchased
-			...( hasTier2Purchase
```
monsieur-z, line 25: Relationships are defined in `UPGRADES` now.

client/my-sites/plans/jetpack-plans/get-purchased-storage-subscriptions.ts
```diff
@@ -1,10 +1,13 @@
+import { JETPACK_BACKUP_ADDON_PRODUCTS } from '@automattic/calypso-products';
 import { getSitePurchases } from 'calypso/state/purchases/selectors/get-site-purchases';
 import { TIER_1_SLUGS, TIER_2_SLUGS } from './constants';
 import type { Purchase } from 'calypso/lib/purchases/types';
 import type { AppState } from 'calypso/types';
 
 const hasUpgradeableStorage = ( slug: string ) =>
```
monsieur-z, line 7: It's safe to update `hasUpgradeableStorage` here.

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-product-card-data.ts
```diff
@@ -0,0 +1,205 @@
...
+	PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_10GB_MONTHLY,
+	PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_100GB_MONTHLY,
+];
+
+const YEARLY_BACKUPS_PRODUCTS = [
+	PRODUCT_JETPACK_BACKUP_T2_MONTHLY,
+	PRODUCT_JETPACK_BACKUP_T2_YEARLY,
+	PLAN_JETPACK_SECURITY_T2_MONTHLY,
+	PLAN_JETPACK_SECURITY_T2_YEARLY,
+	PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_1TB_MONTHLY,
```
singerb, line 45: These 3 should be in the other array, no?

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-storage-upgrade-products.ts
```diff
@@ -1,166 +1,65 @@
...
+		PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_3TB_MONTHLY,
+	],
+	[ PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_1TB_MONTHLY ]: [
+		PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_3TB_MONTHLY,
+		PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_5TB_MONTHLY,
+	],
+	[ PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_3TB_MONTHLY ]: [
+		PRODUCT_JETPACK_BACKUP_ADDON_STORAGE_5TB_MONTHLY,
+	],
+	[ PLAN_JETPACK_SECURITY_T1_MONTHLY ]: [ PLAN_JETPACK_SECURITY_T2_MONTHLY ],
```
singerb: Don't we want to get rid of these two as well?

singerb (commented):
Left a couple questions; we clearly need to revisit this whole flow so that users can purchase any addon for any plan (when it makes sense) but that's out of scope here.

andrii-lysenko:
This is definitely a tricky one as there are a lot of dependencies and add-ons were added pretty quickly. I haven't yet looked deep into code but there are some problems I run into when testing.
In my case: I initially purchased **Backup 10GB Monthly**
- When first entering the page, I see `Your Product` which is `Backup 10GB Monthly` and a button `Upgrade to Yearly`. 
Screenshot 2023-03-13 at 16 17 34
- If I click it, it will add `1Tb addon to cart on checkout`
Screenshot 2023-03-13 at 16 17 46
- If I checkout `10GB addon` it will also suggest me upgrading to Yearly, though, the button will be disabled. Maybe we could remove button in that case.
- After that, If I purchase `100GB` addon, because we can have multiple, I will have both. And then in 'plans/storage' I will see duplicated items.

Maybe, if it's an overkill to tackle everything in terms of one task we could split somehow, or, even transform in a project, also modifying the UI to be inline with a new pricing page? What do you think?

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-product-card-data.ts
monsieur-z: These are products that have a one-year activity log and restore period. It's confusing indeed, I'll update the name 👍

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-storage-upgrade-products.ts
monsieur-z: I don't know about that. If you confirm that we do, I'll remove it.

monsieur-z:
> we clearly need to revisit this whole flow so that users can purchase any addon for any plan (when it makes sense) but that's out of scope here.

Just tagging you for visibility @jboland88, if that's something that falls under the umbrella of MarTech in the future.

monsieur-z:
Thanks for the review @andrii-lysenko!

> When first entering the page, I see Your Product which is Backup 10GB Monthly and a button Upgrade to Yearly. If I click it, it will add 1Tb addon to cart on checkout.

Good catch! I'll try to fix that or remove the CTA altogether for now.

> After that, If I purchase 100GB addon, because we can have multiple, I will have both. And then in 'plans/storage' I will see duplicated items.

Not ideal indeed :) I'll make sure we show each product only once.

> Maybe, if it's an overkill to tackle everything in terms of one task we could split somehow, or, even transform in a project, also modifying the UI to be inline with a new pricing page? What do you think?

This PR is very much a fix for the issue that was raised in p7pQDF-8dJ-p2. The storage upgrade page could definitely benefit from some more attention. Feel free to take part in the discussion in that post.

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-storage-upgrade-products.ts
singerb: Yeah, the pricing page is similarly only showing 10GB for Security, so we shouldn't do the upgrade to 1TB Security here either; they should just buy addon storage.

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-product-card-data.ts
```diff
@@ -0,0 +1,234 @@
...
+	const backupStorageFaq = document.getElementById( backupStorageFaqId );
+
+	if ( backupStorageFaq ) {
+		return `#${ backupStorageFaqId }`;
+	}
+
+	return calypsoEnv === 'development'
+		? `http://jetpack.cloud.localhost:3000/pricing#${ backupStorageFaqId }`
+		: `https://cloud.jetpack.com/pricing#${ backupStorageFaqId }`;
+};
```
TheSteveK, line 66: Not blocking... more of a question, what happens if the environment is something else, like "staging"? Would it make sense to build some sort of reusable selector similar to client/state/selectors/get-jetpack-wp-admin-url.js?

TheSteveK:
in testing, I saw that when I added 100gb, 10gb was still purchasable, is that ok?

Screenshot 2023-03-20 at 10 04 27 AM

Also, a note, I couldn't get this to work in calypso live, it'd just redirect me to the site's plan page, so I had to run it locally to test.

TheSteveK (commented):
Also an inline comment/question

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-product-card-data.ts
mattgawarecki: If this page is only visible from the Jetpack Cloud environment, I would suggest we leave out the domain entirely and go with simply `/pricing#${ backupStorageFaqId }`. If it's visible from WordPress.com as well, we can do a conditional like so:

```js
if ( isJetpackCloud() ) {
	return `/pricing#${ backupStorageFaqId }`;
}

return `https://cloud.jetpack.com/pricing#${ backupStorageFaqId }`;
```

mattgawarecki:
> in testing, I saw that when I added 100gb, 10gb was still purchasable, is that ok?

I saw the same after purchasing 1TB of add-on storage: 10GB and 100GB were still available. I also noticed the same issue @andrii-lysenko mentioned when I revisit the page (replacing 10GB with 1TB):

> If I checkout 10GB addon it will also suggest me upgrading to Yearly, though, the button will be disabled. Maybe we could remove button in that case.

My earlier understanding was that the 3TB and 5TB options were intentionally not exposed through the UI, but giving people more options to pay Jetpack seems like a good thing 🙂

client/my-sites/plans/jetpack-plans/storage-tier-upgrade/hooks/use-get-product-card-data.ts
monsieur-z: I copied this function over and didn't give it more thought. Your suggestion makes sense Matt, I'll update it.

monsieur-z:
> in testing, I saw that when I added 100gb, 10gb was still purchasable, is that ok?

Thanks for testing this Steve. I've pushed a fix for the issue you mentioned.

> Also, a note, I couldn't get this to work in calypso live, it'd just redirect me to the site's plan page, so I had to run it locally to test.

That was maybe a one-off. It works for me.

monsieur-z:
> I also noticed the same issue @andrii-lysenko mentioned when I revisit the page (replacing 10GB with 1TB):

Thanks Matt, that was fixed as well.

> My earlier understanding was that the 3TB and 5TB options were intentionally not exposed through the UI, but giving people more options to pay Jetpack seems like a good thing 🙂

Yes it seems from the comment thread of the [original issue](p7pQDF-8dJ-p2) that people are on board with it.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7879955

Thank you @monsieur-z for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7879955).

PR #74315, opened 2023-03-10
Comments: add buttons to empty spam and trash

bluefuton:
## Proposed Changes

It's currently possible to bulk edit comments in Calypso, but not to empty spam or trash with a single action (like you can in wp-admin). This was spotted by Matt in a Jetpack walkthrough recently - see pedxs5-7f-p2.

This PR adds an "empty" button to the spam and trash tabs in Calypso comment management.

Screenshot 2023-03-20 at 3 20 09 PM

## Testing Instructions

1. Ensure there is an 'Empty spam' button on http://calypso.localhost:3000/comments/spam and an 'Empty trash' button on http://calypso.localhost:3000/comments/trash when you have comments in Spam or Trash.
2. Try the button and ensure that it empties all comments from the current section successfully, and gives you adequate information while it does so. The comment counts should be accurate after the emptying has happened.

## Pre-merge Checklist


bluefuton:
Todo: 
- [x] fix apiVersion (always sends v1.1 even though v1 is specified)
- [x] add Redux bits to remove emptied comments from store (wait for API request to complete before removal?)
- [x] update comment counts to 0 in Redux if empty is successful
- [x] check where post counts are used to ensure we're not causing problems by not updating them
- [x] add tests where appropriate
- [x] hide the empty button for single-post view because the API endpoint does not appear to support that (e.g. /comments/trash/example.com/6)

stephdau:
I've tried this out on [Calypso Live](https://calypso.live/?image=registry.a8c.com/calypso/app:build-60017), and it worked as advertized. 🙂

This said, I'm not sure what the best practice (or option) is within the Calypso framework for longer-lasting processes, but we might want to consider some kind of status indicator when emptying the spam or trash, if possible.

As it stands now, I emptied spam on a site with 95 spam comments, which wasn't quite instant, and I wasn't sure anything was happening (but it did succeed in the end).

Screenshots, for the record:

Emptying the trash:
Screenshot 2023-03-14 at 14 28 50

Emptying the spam:
Screenshot 2023-03-14 at 14 25 17

bluefuton:
> I've tried this out on [Calypso Live](https://calypso.live/?image=registry.a8c.com/calypso/app:build-60017), and it worked as advertized. 🙂

Thank you @stephdau - I'm encouraged that it worked! 

I don't quite have the full process connected up yet, so the removed comments are not deleted from state and the comment counts are not updated.

If the process feels too slow after I've finished all that, we may need to do an optimistic removal of comments, which we then reverse if the API call fails. This has its own complications however 😅

bluefuton:
@tyxla thanks for taking a peek at this for me! I would love some advice in the following areas:

* The API request can take a while if there are lots of comments to empty (as @stephdau [notes above](https://github.com/Automattic/wp-calypso/pull/74315#issuecomment-1468229599)). Is there a lightweight way to make the change optimistically? Currently we're waiting for the API to return the deleted comment IDs so we can remove them from state.
* Any feedback on modernizing the data layer bits?

client/state/comments/reducer.js
```diff
@@ -509,6 +530,26 @@ export const counts = ( state = {}, action ) => {
...
+
+			if ( ! siteId || ! state[ siteId ] || ! status ) {
+				return state;
+			}
+			const { site: siteCounts } = state[ siteId ];
+
+			const emptiedCommentsCount = commentIds?.length || 0;
+			const newSiteCounts = updateCount( siteCounts, status, -emptiedCommentsCount );
+
+			// Post counts can't be updated here because we don't know the post ID.
```
bluefuton, line 543: My findings on this are:

* Counts in `state.comments.counts` are stored per-site, and per-post. We can successfully update the per-site ones, but we can't update the per-post ones after an empty comments action.
* Comment counts are only used by Comment Management (`getSiteCommentCounts` selector).
* Comment Management requests fresh counts via a query component on each load of the comment list (via `QuerySiteCommentCounts`).
* Spam and trash comments are not included in the total comment count that is displayed on the post list (e.g. /posts/example.com).

...so there seem to be no adverse effects of having stale spam/trash post counts in state, even if that doesn't feel particularly nice.

client/my-sites/comments/comment-navigation/index.jsx
```diff
@@ -304,9 +319,24 @@ export class CommentNavigation extends Component {
 					) }
 
 					{ hasComments && (
-						<Button compact onClick={ toggleBulkMode }>
-							{ translate( 'Bulk edit' ) }
-						</Button>
+						<Fragment>
```
tyxla: We could use the short syntax:
```suggestion
						<>
```

client/my-sites/comments/comment-navigation/index.jsx
```diff
@@ -304,9 +319,24 @@ export class CommentNavigation extends Component {
...
+									scary
+									onClick={ this.emptyPermanently }
+									className="comment-navigation__button-empty"
+								>
+									{ this.props.status === 'spam'
+										? translate( 'Empty spam' )
+										: translate( 'Empty trash' ) }
+								</Button>
+							) }
+						</Fragment>
```
tyxla:
```suggestion
						</>
```

client/my-sites/comments/comment-navigation/index.jsx
```diff
@@ -65,6 +66,19 @@ export class CommentNavigation extends Component {
 		}
 	};
 
+	emptyPermanently = () => {
+		const { status, translate } = this.props;
+		if (
+			window?.confirm(
```
tyxla: Why is the `window` optional chaining necessary? Does that run in SSR context?

client/state/action-types.ts
```diff
@@ -93,6 +93,8 @@ export const COMMENTS_COUNT_INCREMENT = 'COMMENTS_COUNT_INCREMENT';
 export const COMMENTS_COUNT_RECEIVE = 'COMMENTS_COUNT_RECEIVE';
 export const COMMENTS_DELETE = 'COMMENTS_DELETE';
 export const COMMENTS_EDIT = 'COMMENTS_EDIT';
+export const COMMENTS_EMPTY = 'COMMENTS_EMPTY';
+export const COMMENTS_EMPTY_RECEIVE = 'COMMENTS_EMPTY_RECEIVE';
```
tyxla: Should this be called `COMMENTS_EMPTY_SUCCESS`?

client/state/comments/reducer.js
```diff
@@ -158,6 +163,22 @@ export function items( state = {}, action ) {
...
+			const { commentIds } = action;
+
+			const commentKeys = Object.keys( state );
+			let newState = { ...state };
+			commentKeys.forEach( ( key ) => {
+				newState = {
+					...newState,
+					[ key ]: state[ key ].filter( ( comment ) => ! commentIds.includes( comment.ID ) ),
+				};
+			} );
```
tyxla: Could be more straightforward to do a `Object.entries().map()` here.

client/state/comments/reducer.js
tyxla: We could trigger some re-fetching of comment counts on successful spam cleanup, can't we?

client/state/comments/reducer.js
```diff
@@ -509,6 +530,26 @@ export const counts = ( state = {}, action ) => {
...
+			const newSiteCounts = updateCount( siteCounts, status, -emptiedCommentsCount );
+
+			// Post counts can't be updated here because we don't know the post ID.
+
+			const newTotalSiteCounts = Object.assign(
+				{},
+				state[ siteId ],
+				newSiteCounts && { site: newSiteCounts }
+			);
+			return Object.assign( {}, state, { [ siteId ]: newTotalSiteCounts } );
```
tyxla: Nit: object spread could simplify `Object.assign()` calls a bit:

```suggestion
			const newTotalSiteCounts = Object.assign(
				{},
				state[ siteId ],
				newSiteCounts && { site: newSiteCounts }
			);
			return { ...state, ...{ [ siteId ]: newTotalSiteCounts } };
```

client/state/data-layer/wpcom/comments/index.js
```diff
@@ -188,6 +191,68 @@ export const announceDeleteFailure = ( action ) => {
...
+				apiVersion: '1',
+				body: {
+					empty_status: status,
+				},
+				method: 'POST',
+				path: `/sites/${ siteId }/comments/delete`,
+			},
+			{
+				...action,
+			}
```
tyxla: Perhaps we could just pass `action` here? Is there a good reason why we create a new object?

```suggestion
			action
```

client/state/data-layer/wpcom/comments/index.js
```diff
@@ -188,6 +191,68 @@ export const announceDeleteFailure = ( action ) => {
...
+
+export const announceEmptyFailure = ( action ) => {
+	const { status } = action;
+
+	const error = errorNotice(
+		status === 'spam'
+			? translate( 'Could not empty spam.' )
+			: translate( 'Could not empty trash.' ),
+		{
+			duration: DEFAULT_NOTICE_DURATION,
```
tyxla: Does this also need the same `id` as the success notice?

Perhaps we could reuse the same `noticeOptions` that we're passing to both notices.

client/state/data-layer/wpcom/comments/index.js
```diff
@@ -188,6 +191,68 @@ export const announceDeleteFailure = ( action ) => {
...
+					id: 'comment-notice',
+					isPersistent: true,
+				}
+			),
+		!! refreshCommentListQuery && requestCommentsList( refreshCommentListQuery ),
+		{
+			type: COMMENTS_EMPTY_RECEIVE,
+			siteId,
+			status,
+			commentIds: apiResponse?.results?.map( ( x ) => +x ), // convert to number
```
tyxla: In which scenario would `apiResponse` contain no results? Should we default to an empty array in this case?

```suggestion
			commentIds: apiResponse?.results?.map( ( x ) => +x ), // convert to number
```

tyxla (commented):
Thanks for working on that issue, @bluefuton! Nice work 👍 

I've left some suggestions. Also, replying to your prior questions: 

> Is there a lightweight way to make the change optimistically? Currently we're waiting for the API to return the deleted comment IDs so we can remove them from state.

I don't think we have any precedent for that in Calypso. It could be done in multiple ways, though, some of which are not necessarily very straightforward.

A - AKA the more complex one. We could use an async job to delete spam comments. Then, we could have a flag (option or blog meta for example) to mark that we're currently in the process of emptying all spam. That flag would be used to modify the UI to display that we're currently deleting spam. That could be as a global notice, or a notice that is displayed on the comments section. Once the async job has finished, it would empty the flag. Of course, we would also need a "status" endpoint and we'd have to request it every few seconds, to check if the job had finished already. The benefit about this approach is that it would work well both for the user and for us, regardless of the spam count - it wouldn't block any UI, it would still be optimistic and any server-side resources used would be exhausted behind the scenes. We could also deal with millions of spam comments that way, by deleting them in chunks and triggering another job to deal with the next batch. This would work for millions and billions of spam comments.

B - AKA the simpler one - we could expand what this PR is suggesting, by adding a notice that the spam deletion is in progress. So on top of the current notices on success and failure, we'd have an info/warning notice, indicating that spam deletion is in progress. We'd then clean the notice when the request is finished. This would work okay, but likely would cause OOM errors and long requests with millions/billions of spam comments. The downside is that on a page refresh, everything would be lost (including the "currently emptying spam" state), and results may turn to be inconsistent (some spam comments deleted, some not, etc). That could be improved by implementing the flag mentioned above, but there's still no guarantee that the aborted deletion request would finish properly.

We could try B) first, and keep an eye on the error logs. If there are timeouts or OOMs, then we'd have to consider the more complicated A) approach.

I think the "in progress" notice is the bare minimum needed here. 

> Any feedback on modernizing the data layer bits?

I think the data layer can be very helpful in complex situations like this. I took a look and I didn't see anything blocking that needs our attention, but I offered some suggestions you could consider.

bluefuton:
> We could try B) first, and keep an eye on the error logs. If there are timeouts or OOMs, then we'd have to consider the more complicated A) approach.
> we could expand what this PR is suggesting, by adding a notice that the spam deletion is in progress. So on top of the current notices on success and failure, we'd have an info/warning notice, indicating that spam deletion is in progress.

That sounds smart to me 👍 I'll implement the extra notice and we can see how that goes.

Thanks for taking the time to review this @tyxla! 🙌

client/my-sites/comments/comment-navigation/index.jsx
bluefuton: I was being cautious because the equivalent function for bulk deleting comments in the same component does this:

```
if (
	typeof window === 'undefined' ||
	window.confirm( translate( 'Delete these comments permanently?' ) )
) {
```

What's the best way to find out if a section runs in the SSR context?

I wonder if that code was accounting for some other situation where `window` wasn't available - tests perhaps?

client/state/action-types.ts
bluefuton: That sounds better 👍

client/state/comments/reducer.js
bluefuton: Thanks :) I was reusing old code here. Much neater with the spread operator.

client/state/data-layer/wpcom/comments/index.js
bluefuton: It turns out - never:

Screenshot 2023-03-17 at 3 05 52 PM

This was an API request where there were no comments to clean up.

So no need to be as defensive here.

client/state/comments/reducer.js
bluefuton: Sounds like a plan. What's the cleanest way to do that? via an additional prop to the query component?

bluefuton:
We now show an info notice while emptying is in progress:

Screenshot 2023-03-17 at 3 54 54 PM

...which changes to a success notice once complete.

Screenshot 2023-03-17 at 3 54 58 PM

client/state/comments/reducer.js
tyxla: If you want this to unconditionally happen, you can even add a `COMMENTS_REQUEST` action to the data layer, somewhere at the end of `handleEmptySuccess()`.

client/my-sites/comments/comment-navigation/index.jsx
tyxla: FWIW I'm fine with keeping the optional chaining, especially if there's a good reason for it.

> What's the best way to find out if a section runs in the SSR context?

Basically, tracing the `CommentNavigation` component up to its corresponding section:

https://github.com/Automattic/wp-calypso/blob/bb00cff9a4e21a8beed8cfa95c17e09bf428bda4/client/sections.js#L391-L396

and checking if `isomorphic` is set to `true` for that section. If it's not then, it doesn't support SSR.

> I wonder if that code was accounting for some other situation where window wasn't available - tests perhaps?

Perhaps that's why! In any case, we should be able to mock what's missing, or enable `jsdom` testing environment for the tests, which supports most of the browser APIs.

client/state/data-layer/wpcom/comments/index.js
tyxla: Nice 👍

client/my-sites/comments/comment-navigation/index.jsx
```diff
@@ -364,6 +394,17 @@ const mapDispatchToProps = ( dispatch, { siteId, commentsListQuery } ) => ( {
 				deleteComment( siteId, postId, commentId, { showSuccessNotice: true }, commentsListQuery )
 			)
 		),
+	// Empty all comments (from spam or trash only)
+	emptyPermanently: ( status ) =>
+		dispatch(
+			withAnalytics(
+				composeAnalytics(
+					recordTracksEvent( 'calypso_comment_management_empty' ),
+					bumpStat( 'calypso_comment_management', 'comments_emptied' )
```
tyxla, line 403: Nit: this bumps stats before the emptying has finished, right? So if emptying has failed, stat will be bumped regardless, and that sounds a bit imprecise.

tyxla (approved):
This tests great 👍 

Code also looks good to me. 💯 

While I've suggested a more elaborate and scalable solution, I think this is much better than not having it at all, and it is a huge improvement to be able to just empty all spam and trash with a single click. I'd say let's try it that way, and if need arises for scalability, we can scale it.

Thanks @bluefuton 🚀

client/my-sites/comments/comment-navigation/index.jsx
bluefuton: Thanks for filling me in on `isomorphic: true` :) Doesn't seem like there's a compelling reason to keep it. Removed in [260f745](https://github.com/Automattic/wp-calypso/pull/74315/commits/260f745ebb798039c11c56950fe501638c48d0c2).

client/my-sites/comments/comment-navigation/index.jsx
bluefuton: I also noticed this, but other actions in the same component behave in the same way (like `deletePermanently` and `changeStatus`) - the stat is bumped when the user takes the action, and does not take into consideration whether or not the action is successful. I thought it made sense to stick with the existing convention, even if it is a bit imprecise.

client/state/comments/reducer.js
bluefuton: Good idea. I had a play with firing a `COMMENT_COUNTS_REQUEST` from the data layer but I think that just gives us the same site comment counts we already have. We also re-fetch comments (filtered for the current comment list) when the emptying is finished via `refreshCommentListQuery`:

https://github.com/Automattic/wp-calypso/blob/260f745ebb798039c11c56950fe501638c48d0c2/client/state/data-layer/wpcom/comments/index.js#L239

I think that should be enough. 🤞

bluefuton:
Thanks for all your help @tyxla 🙂

> I'd say let's try it that way, and if need arises for scalability, we can scale it.

Would it be sensible to launch it for internal users only first, or would you ship it to everyone and monitor it?

This is a bit unexpected as I'm not touching anything editing-related - is it possible to easily find out what in the PR overlaps with `editing-toolkit`?

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7877829

Thank you @bluefuton for including a screenshot in the description! This is really helpful for our translators.

tyxla:
> Would it be sensible to launch it for internal users only first, or would you ship it to everyone and monitor it?

Unless you're planning a call for testing internally, I'd suggest launching to the public and monitoring for errors. I doubt this would be used too much internally, so you might be luckier to gather better usage data out in the public.

> This is a bit unexpected as I'm not touching anything editing-related - is it possible to easily find out what in the PR overlaps with editing-toolkit?

This is expected because ETK uses a bit of Calypso's state, and since you touched `client/state/action-types.ts`, and it uses that, it will warn you that. Since this feature is not exposed on ETK, feel free to ship and disregard the warning. 

One separate thing I wanted to highlight was i18n: you might want to wait until all strings are translated before shipping this. That will guarantee that the feature is available in all 16 most used languages.

bluefuton:
> One separate thing I wanted to highlight was i18n: you might want to wait until all strings are translated before shipping this. That will guarantee that the feature is available in all 16 most used languages.

Thank you! I've applied the string freeze label and will wait until that completes before launch 🚀

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7877829).

stephdau:
Great work. :)

PR #73890, opened 2023-02-28
Try #2: Upgrade @wordpress packages to pre-React 18

tyxla:
This PR aims to upgrade the `@wordpress` packages to versions before upgrading to React 18.

The second take of #73155.

Does a lot of dependency and TS changes. There are some dependency compromises and specific pinning, but this PR aims to make a step forward towards updating the `@wordpress` packages to the latest version and restarting their regular periodic upgrades.

I think it requires thorough testing of all major flows.

PR #74346, opened 2023-03-10
Migrate spt store from registerStore to register

noahtallen:
## Proposed Changes
Experiments with migrating a store from `registerStore` to `createReduxStore` + `register`. The first reason to do it is because `registerStore` is deprecated. The main benefit we get is built-in type inference without any extra work, which ultimately simplifies existing code.

This depends on the `@wordpress/data` update in #73890 being merged first.

## Testing Instructions
1. install this branch's ETK build on your sandbox
2. Sandbox a test site and create a new page
3. verify the page layout store works
4. TBD: verify interactions with other parts of ETK, like the welcome modals.

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts
+++ b/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts
@@ -1,4 +1,4 @@
-import { registerStore } from '@wordpress/data';
+import { register, createReduxStore } from '@wordpress/data';
 import type { Reducer } from 'redux';
 
 type OpenState = 'CLOSED' | 'OPEN_FROM_ADD_PAGE' | 'OPEN_FOR_BLANK_CANVAS';
@@ -19,14 +19,9 @@ export const selectors = {
 	isPatternPicker: ( state: OpenState ): boolean => 'OPEN_FOR_BLANK_CANVAS' === state,
 };
 
-const STORE_KEY = 'automattic/starter-page-layouts';
-
-registerStore( STORE_KEY, {
-	// In reality the store can dispatch any action, however `reducer` has a
-	// strongly typed action type to make the typings inside the function
-	// easier to work with.
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	reducer: reducer as any,
+export const pageLayoutStore = createReduxStore( 'automattic/starter-page-layouts', {
+	reducer,
 	actions,
 	selectors,
 } );
+register( pageLayoutStore );
--- a/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/index.tsx
+++ b/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/index.tsx
@@ -2,8 +2,7 @@ import { initializeTracksWithIdentity, PatternDefinition } from '@automattic/pag
 import { dispatch } from '@wordpress/data';
 import { registerPlugin } from '@wordpress/plugins';
 import { PagePatternsPlugin } from './page-patterns-plugin';
-
-import './store';
+import { pageLayoutStore } from './store';
 import './index.scss';
 
 declare global {
@@ -29,7 +28,7 @@ if ( tracksUserData ) {
 
 // Open plugin only if we are creating new page.
 if ( screenAction === 'add' ) {
-	dispatch( 'automattic/starter-page-layouts' ).setOpenState( 'OPEN_FROM_ADD_PAGE' );
+	dispatch( pageLayoutStore ).setOpenState( 'OPEN_FROM_ADD_PAGE' );
 }
 
 // Always register ability to open from document sidebar.
--- a/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/page-patterns-plugin.tsx
+++ b/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/page-patterns-plugin.tsx
@@ -3,8 +3,7 @@ import { useSelect, useDispatch } from '@wordpress/data';
 import { useCallback } from '@wordpress/element';
 import { addFilter, removeFilter } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
-import { selectors as starterPageTemplatesSelectors } from '../starter-page-templates/store';
-import type { SelectFromMap } from '@automattic/data-stores';
+import { pageLayoutStore } from './store';
 import '@wordpress/nux';
 
 const INSERTING_HOOK_NAME = 'isInsertingPagePattern';
@@ -13,7 +12,6 @@ const INSERTING_HOOK_NAMESPACE = 'automattic/full-site-editing/inserting-pattern
 interface PagePatternsPluginProps {
 	patterns: PatternDefinition[];
 }
-type StarterPageTemplatesSelectors = SelectFromMap< typeof starterPageTemplatesSelectors >;
 type CoreEditorPlaceholder = {
 	getBlocks: ( ...args: unknown[] ) => Array< { name: string; clientId: string } >;
 	getEditedPostAttribute: ( ...args: unknown[] ) => unknown;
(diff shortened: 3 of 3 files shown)


apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/block-editor-nux.js
```diff
@@ -26,19 +27,18 @@ function WelcomeTour() {
 	const { show, isLoaded, variant, isManuallyOpened, isNewPageLayoutModalOpen } = useSelect(
 		( select ) => {
 			const welcomeGuideStoreSelect = select( 'automattic/wpcom-welcome-guide' );
-			const starterPageLayoutsStoreSelect = select( 'automattic/starter-page-layouts' );
 			return {
 				show: welcomeGuideStoreSelect.isWelcomeGuideShown(),
 				isLoaded: welcomeGuideStoreSelect.isWelcomeGuideStatusLoaded(),
 				variant: welcomeGuideStoreSelect.getWelcomeGuideVariant(),
 				isManuallyOpened: welcomeGuideStoreSelect.isWelcomeGuideManuallyOpened(),
-				isNewPageLayoutModalOpen: starterPageLayoutsStoreSelect?.isOpen(), // Handle the case where SPT is not initalized.
```
noahtallen: This is the main thing we need to test. The purpose of the SPT store is to track whether the "new page template" screen is open or closed. Since that feature isn't always loaded, it used to be impossible for the store to be defined.

However, now that we directly import the store from the file, it seems more likely that the store stuff would at least be defined.

But on top of that, I'm not sure how it works when the store is imported in multiple different plugins that are technically part of different apps/scripts. ETK isn't one unified application -- instead, it's multiple smaller features which are ultimately individual scripts that get registered in PHP based on whether that feature should be loaded on the page in question.

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts
```diff
@@ -21,12 +21,10 @@ export const selectors = {
...
-	actions,
-	selectors,
-} );
+export const store = createReduxStore( STORE_KEY, { reducer, actions, selectors } );
+
+// TODO: Why is this needed?
+if ( ! starterPageTemplatesConfig.didRegisterStore ) {
+	register( store );
+	starterPageTemplatesConfig.didRegisterStore = true;
+}
```
noahtallen: @jsnajdr I'm curious if you have any thoughts about the best way to avoid registering the store twice.

From what I can tell, the `@wordpress/data` package overwrites the existing store every time you call `register`: https://github.com/WordPress/gutenberg/blob/605aeb0f4f7d2225120e498f95ae27b9f56d77a3/packages/data/src/registry.js#L217-L265

(It doesn't check if the store name was already registered, and just writes it to the stores object under the store's unique key).

This isn't normally an issue, if a file is only executed once when importing it.

But what if there are multiple different bundles trying to access the same store? In ETK, each sub-feature is its own webpack bundle. So starter page templates has its own bundle which is conditionally loaded when applicable, and a different feature like "wpcom block editor nux" has a different bundle which is conditionally loaded when needed. (When a bundle is needed depends on different factors like what page you visit.)

So we end up with this:

```js
// file1, which is in one bundle:
import { store } from './store';

// file2, which is in a different bundle
import { store as templateStore } from '../../starter-page-templates/store';
```

When the bundle associated with file1 is loaded, it sets up some initial store based on some conditionals: https://github.com/Automattic/wp-calypso/blob/2656f2beeca506ac256b71be878ac05f42b17832/apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/index.tsx#L29-L44

But when the bundle associated with file2 is loaded, register is called again, which overwrites the store which clears out that change.

As far as I can tell, after this point, dispatch/select will be using the same store from any bundle. But it'd be nice to avoid clearing out the existing store the second time it gets registered!

I'm wondering if there's a better way to do this other than relying on globals, because I think we'll want similar behavior for some other stores.

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts
```diff
@@ -21,12 +21,10 @@ export const selectors = {
...
-	// easier to work with.
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	reducer: reducer as any,
-	actions,
-	selectors,
-} );
+export const store = createReduxStore( STORE_KEY, { reducer, actions, selectors } );
+
+// TODO: Why is this needed?
+if ( ! starterPageTemplatesConfig.didRegisterStore ) {
```
tyxla: I don't think that registered check is necessary:

https://github.com/WordPress/gutenberg/blob/1b6b4b8d8439b7cc0b0d90334a3b7cccaa665b45/packages/block-directory/src/store/index.js#L40

There are a bunch of core stores registered that same way, and I haven't noticed that practice there.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/block-editor-nux.js
tyxla: The store gets registered during import, so I can't figure out how can it be not initialized at this time. I'm not sure I share any concern about this change - perhaps I misunderstood something?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/welcome-tour/tour-launch.tsx
```diff
@@ -34,10 +33,7 @@ function LaunchWpcomWelcomeTour() {
 			show: (
 				select( 'automattic/wpcom-welcome-guide' ) as WpcomWelcomeGuideSelectors
 			 ).isWelcomeGuideShown(),
-			// Handle the case where the new page pattern modal is initialized and open
```
tyxla: We could keep the comment I guess.

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts
tyxla: Removing the check should also address the TS error that the `type_check_etk` CI task is failing with.

noahtallen:
@tyxla Unfortunately, it is needed or we end up in a situation where the store is registered twice. Mainly because `src/block-editor-nux.js` is in a different webpack bundle -- so two get different webpack bundles importing the same store file, which ends up calling the registration code twice. This ultimately clears out data updated in the first instance of the store which breaks the feature :( (If you test it out on a branch without this workaround, the SPT feature will never load)

I tried to explain it in my [comment above](https://github.com/Automattic/wp-calypso/pull/74346#discussion_r1132879384), but touching on your feedback:

> There are a bunch of core stores registered that same way, and I haven't noticed that practice there.

Right, but the reason these stores don't have the same problem is that we access them through dependency extraction. So the end result is `useSelect( select => select( wp.editor.store ) )` (fake example). This ultimately doesn't go through webpack import, which means the code registering the store wouldn't be executed a second time -- we're just accessing an existing global variable instead of importing anything.

> The store gets registered during import, so I can't figure out how can it be not initialized at this time

This would be the case after the migration, but wasn't the case using the old approach. With the old approach, `src/block-editor-nux.js` never imported the store -- as a result, it was subject to race conditions with script load order. (All these ETK bundles are ultimately added via wp script enqueue). But this won't be an issue any more!

tyxla:
That makes sense, thanks for the explanation @noahtallen 👍 

In that case, we'll need to make sure `didRegisterStore` is an expected property of `starterPageTemplatesConfig` since right now we'll see type errors.

noahtallen:
I'll go back to the string-based store selectors for now, and add a follow-up item for once core data has a fix (hopefully!)

noahtallen:
So essentially, internally, the page layout store can be used with the new approach, and the other webpack bundles continue doing what they were doing before (use the string-based store id)

tyxla:
> So essentially, internally, the page layout store can be used with the new approach, and the other webpack bundles continue doing what they were doing before (use the string-based store id)

How does that work? Are we registering the store multiple times, or whichever gets to register the store first would register it?

noahtallen:
No, it only gets registered once in the SPT module. This is how it worked before -- the only reason the other module consumes the SPT store is to make sure its modal doesn't conflict with SPT. SPT is only loaded in some scenarios (only in the page editor, not in the post editor), so the absence of the store is a scenario in which SPT does not conflict with this other module.

Again, this is how it has been working for the past several years (as wonky as it is :/)

tyxla (approved):
Looks great to me 👍 🚀

PR #74351, opened 2023-03-10
format-currency: Show `US$` for USD outside of US

### LINKED ISSUE #73496: Argentina set to en-US misleadingly shows USD with $ currency symbol
### Quick summary

Many countries (eg: Canada, Argentina) use `$` as their currency symbol in addition to the US. If a user is in Argentina, are paying in USD, and have their locale set to `en-US`, they will see prices like `$100.00`. This is misleading because they might assume it's in ARS and not USD. (This is not an issue if the user has `en-GB` or `en-CA` where the currency symbol will appear as `US $`; but we cannot use that inside the US because US users would be confused.)

This problem was resolved in https://github.com/Automattic/wp-calypso/issues/46205 primarily by using the IP address to determine if we should show `$` or `US $`. However, https://github.com/Automattic/wp-calypso/pull/71387 moved all currency formatting to the frontend and https://github.com/Automattic/wp-calypso/pull/71895 made it rely only on the user's locale, so this issue has returned.

Props to @daledupreez who noticed this in https://github.com/Automattic/wp-calypso/pull/71387#issuecomment-1434628152

sirbrillig:
In the US, users will expect to see USD prices rendered with the currency symbol `$`. However, there are many other currencies which use `$` as their currency symbol (eg: `CAD`). The `@automattic/format-currency` functions try to prevent confusion between these symbols by using an international version of the symbol when the locale does not match the currency. So if your locale is `en-CA`, USD prices will be rendered with the symbol `US$`. 

However, this relies on the user having set their interface language to something other than `en-US`/`en`, and many English-speaking non-US users still have that interface language (eg: there's no English locale available in our settings for Argentinian English so such users would probably still have `en`). As a result, those users will see a price with `$` and could be misled about what currency is being displayed.

## Proposed Changes

In this diff, we fetch the user's geolocation (using the `/geo` API endpoint which does not require authentication) and use it to override the currency symbol for USD. If the user is in the US, they will see USD with the `$` symbol, otherwise they will see the `US$` symbol.

The fetching should occur as early as possible during calypso's loading process and should be attempted for both logged-in and logged-out users (eg: so that the logged-out version of the Plans page shows the correct currency symbol). To do this, we make the API call inside the "layout" component. The call is a non-blocking side effect so it's entirely possible for it to fail without any harmful effects.

Fixes https://github.com/Automattic/wp-calypso/issues/73496

## Screenshots

Inside US:

Screenshot 2023-03-10 at 4 26 09 PM

Outside US:

Screenshot 2023-03-10 at 4 26 20 PM

## Testing Instructions

- View calypso with a user that has `en` or `en-US` as their locale and `USD` as their currency.
- Visit a page in calypso that has at least one price displayed (eg: `/plans` or `/checkout`).
- Make your geolocation inside the US (using a VPN if necessary) and verify that the prices are displayed with `$`.
- Make your geolocation outside the US (using a VPN if necessary) and verify that the prices are displayed with `US$`.

### DIFF
--- a/client/layout/index.jsx
+++ b/client/layout/index.jsx
@@ -1,6 +1,6 @@
 import config from '@automattic/calypso-config';
 import { HelpCenter } from '@automattic/data-stores';
-import { detectGeolocation } from '@automattic/format-currency';
+import { geolocateCurrencySymbol } from '@automattic/format-currency';
 import { shouldShowHelpCenterToUser, shouldLoadInlineHelp } from '@automattic/help-center';
 import { isWithinBreakpoint } from '@automattic/viewport';
 import { useBreakpoint } from '@automattic/viewport-react';
@@ -153,7 +153,7 @@ class Layout extends Component {
 	};
 
 	componentDidMount() {
-		detectGeolocation();
+		geolocateCurrencySymbol();
 
 		if ( ! config.isEnabled( 'me/account/color-scheme-picker' ) ) {
 			return;
--- a/client/layout/index.jsx
+++ b/client/layout/index.jsx
@@ -1,5 +1,6 @@
 import config from '@automattic/calypso-config';
 import { HelpCenter } from '@automattic/data-stores';
+import { detectGeolocation } from '@automattic/format-currency';
 import { shouldShowHelpCenterToUser, shouldLoadInlineHelp } from '@automattic/help-center';
 import { isWithinBreakpoint } from '@automattic/viewport';
 import { useBreakpoint } from '@automattic/viewport-react';
@@ -152,6 +153,8 @@ class Layout extends Component {
 	};
 
 	componentDidMount() {
+		detectGeolocation();
+
 		if ( ! config.isEnabled( 'me/account/color-scheme-picker' ) ) {
 			return;
 		}
--- a/client/layout/index.jsx
+++ b/client/layout/index.jsx
@@ -1,6 +1,5 @@
 import config from '@automattic/calypso-config';
 import { HelpCenter } from '@automattic/data-stores';
-import { geolocateCurrencySymbol } from '@automattic/format-currency';
 import { shouldShowHelpCenterToUser, shouldLoadInlineHelp } from '@automattic/help-center';
 import { isWithinBreakpoint } from '@automattic/viewport';
 import { useBreakpoint } from '@automattic/viewport-react';
@@ -153,8 +152,6 @@ class Layout extends Component {
 	};
 
 	componentDidMount() {
-		geolocateCurrencySymbol();
-
 		if ( ! config.isEnabled( 'me/account/color-scheme-picker' ) ) {
 			return;
 		}
--- a/client/layout/logged-out.jsx
+++ b/client/layout/logged-out.jsx
@@ -1,9 +1,11 @@
 import config from '@automattic/calypso-config';
+import { detectGeolocation } from '@automattic/format-currency';
 import { useLocalizeUrl, removeLocaleFromPathLocaleInFront } from '@automattic/i18n-utils';
 import { UniversalNavbarHeader, UniversalNavbarFooter } from '@automattic/wpcom-template-parts';
 import classNames from 'classnames';
 import { localize } from 'i18n-calypso';
 import PropTypes from 'prop-types';
+import { useEffect } from 'react';
 import { connect, useSelector } from 'react-redux';
 import { CookieBannerContainerSSR } from 'calypso/blocks/cookie-banner';
 import AsyncLoad from 'calypso/components/async-load';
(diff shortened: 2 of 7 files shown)


packages/format-currency/src/index.ts
```diff
@@ -11,11 +11,29 @@ export * from './types';
...
 const fallbackCurrency = 'USD';
+const geolocationEndpointUrl = 'https://public-api.wordpress.com/geo/';
 
 export function createFormatter(): CurrencyFormatter {
 	const currencyOverrides: Record< string, { symbol?: string | undefined } > = {};
 	let defaultLocale: string | undefined = undefined;
 
+	async function setUsdCurrencySymbolBasedOnGeolocation(): Promise< void > {
+		const geoData = await globalThis
+			.fetch?.( geolocationEndpointUrl )
```
tyxla, line 23: Might make sense to also have a `catch()` in case of a failed request.

packages/format-currency/src/index.ts
```diff
@@ -223,10 +261,17 @@ export function createFormatter(): CurrencyFormatter {
 		defaultLocale = locale;
 	}
 
+	// If the user is inside the US using USD, they should only see `$` and not `US$`.
+	async function detectGeolocation() {
```
tyxla: I think naming here is a bit confusing. 

When I see `@automattic/format-currency` exposing a `detectGeolocation()` function, it immediately feels weird - what does geo location detection has to do with currency formatting?

Once we read the function's contents, it all clicks. However, the function does something more on top of what it claims to do - it changes the currency symbol also.

So I'd suggest renaming this function to something more explicit, like `geolocateCurrencySymbol` for example. It could be a bit verbose, but it would remove the confusion I described above.

packages/format-currency/src/index.ts
```diff
@@ -223,10 +261,17 @@ export function createFormatter(): CurrencyFormatter {
...
+		return setUsdCurrencySymbolBasedOnGeolocation();
```
tyxla: I don't see a good reason for the indirection - `setUsdCurrencySymbolBasedOnGeolocation() could be inlined here and we'd have one less defined function.

packages/format-currency/src/index.ts
```diff
@@ -372,8 +417,20 @@ function getSmallestUnitDivisor( precision: number ): number {
 	return 10 ** precision;
 }
 
+interface GeoLocationResponse {
+	country_short: string;
+}
+
+function isGeolocationResponse( response: unknown ): response is GeoLocationResponse {
```
tyxla: Should this be named `containsGeolocationCountry` or something similar? Being explicit wins IMHO. Same for the `GeoLocationResponse` type guard, since it only contains a `country_short` anyway, and the geolocation response has a bunch more fields we don't need to think about here.

client/layout/logged-out.jsx
```diff
@@ -83,6 +85,10 @@ const LayoutLoggedOut = ( {
 		'is-p2-login': isP2Login,
 	};
 
+	useEffect( () => {
+		detectGeolocation();
+	}, [] );
```
tyxla: I've tested this on a few logged-out pages and it works as expected 👍

client/layout/index.jsx
```diff
@@ -152,6 +153,8 @@ class Layout extends Component {
 	};
 
 	componentDidMount() {
+		detectGeolocation();
```
tyxla: I'd have suggested doing this in a component like the query components we have, but then calling it as early as possible makes sense too. 

Alternatively, we could do that even earlier at boot time:

https://github.com/Automattic/wp-calypso/blob/6a8c4cc88dad56a331def849f5acea515cc37f53/client/boot/common.js#L421-L450

which would also have a good side benefit of not having to repeat it in both layout components.

tyxla (commented):
This is looking good to me overall 👍 

I've left a few suggestions and questions, let me know what you think.

packages/format-currency/src/index.ts
sirbrillig: 👍

packages/format-currency/src/index.ts
sirbrillig: Good idea! Naming is hard.

packages/format-currency/src/index.ts
sirbrillig: 👍

packages/format-currency/src/index.ts
sirbrillig: 👍

client/layout/index.jsx
sirbrillig: Ah, nice idea! I've moved the call into the `boot` function. Let me know if that looks right to you.

tyxla (approved):
The code looks great and this tests well.

Thanks @sirbrillig 🚀

PR #74362, opened 2023-03-13
VAT: add AU, CH, and JP tax vendor info

michaeldcain:
This add the tax vendor info for Switzerland, Australia, and Japan. It also abstracts the Aut O'Mattic address into a variable so that it only needs to be updated in a single place.

See: https://github.com/Automattic/payments-shilling/issues/1448

To test:
- with Store sandboxed
- visit Checkout and make a purchase using the Swiss VAT details in D104290-code
- visit Purchases > Billing History and view the receipt for the most recent purchase
- verify that the correct Swiss VAT number (`CHE-259.584.214 MWST`) is shown with the Aut O'Mattic Irish vendor address

### DIFF
--- a/client/me/purchases/billing-history/vat-vendor-details.tsx
+++ b/client/me/purchases/billing-history/vat-vendor-details.tsx
@@ -44,10 +44,10 @@ function getVatVendorInfo(
 		};
 	}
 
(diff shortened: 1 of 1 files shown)


sirbrillig (approved):
👍

Issue #73830, opened 2023-02-27
Launchpad: Atomic not syncing correctly/quickly enough


mattwiebe:
### Quick summary

We are relying on Sync Site data to deliver our Launchpad-related options. This is not fast enough, particularly with a freshly-transferred site that has not yet completed its initial Sync to the now-empty wpcom Simple site.

These issues were discovered while trying to debug #73377 

### Steps to reproduce

1. Create a site that enables Launchpad
2. In Launchpad, "Skip to Admin"
3. Transfer the site to Atomic (upload a plugin or theme, go to Settings -> Hosting Config)
4. Visit My Home
5. Notice that you are no longer redirected to Launchpad from My Home (this does not always happen, but it happens regularly)

### What you expected to happen

Launchpad behaviour maintained after Atomic transfer

### What actually happened

Launchpad behaviour is not maintained after Atomic transfer

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

karenroldan:
📌 ACTIONS
- Due to the age of this issue, I've downgraded the priority from `High` to `Normal` to better reflect its status in the larger context of open bugs in this repo.

mmtr:
I think this is issue is no longer applicable, because (in my tests) My Home no longer redirects to the Launchpad when I click on the "Skip for now" button.

Feel free to re-open this or report a new issue if I missed anything.

PR #74365, opened 2023-03-13
Launchpad.- Update mayberedirect launchpad option retrieval

agrullon95:
Related to https://github.com/Automattic/wp-calypso/issues/73830

## Time Estimate

Review: Short
Test: Short

## Proposed Changes

* Add useLaunchpad hook to maybeRedirect controller function. This hook will use the launchpad endpoint to fetch the latest launchpad to determine if the user needs to be redirected to the launchpad screen or `/home`

## Testing Instructions

* Checkout this branch
* Create or use an existing launchpad enabled site (Test on an atomic site as well)
* Navigate to `/home/{siteSlug}` and verify that you are redirected to the launchpad screen
* Navigate to `/home/{siteSlug}?launchpadComplete=true` and verify you are not redirected to the launchpad screen
* Disable the launchpad (change site option `launchpad_screen` to `off` or FALSE) or use a site that is not launchpad enable and navigate to `/home`. Verify that you are not navigated to the launchpad screen.

## Pre-merge Checklist


client/data/sites/use-launchpad.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const fetchLaunchpad = ( siteSlug: string ) => {
+	const slug = encodeURIComponent( siteSlug as string );
+	return wpcom.req.get( {
+		path: `/sites/${ slug }/launchpad`,
+		apiNamespace: 'wpcom/v2',
+	} );
+};
+
+export const useLaunchpad = ( siteSlug: string, cache = true ) => {
```
edanzer: Been a while since I used React Query, and I know it has a range of different options and settings around caching. Do we need to worry about caching this value, and not updating quickly after the setting has been updated on the backend?

client/my-sites/customer-home/controller.jsx
```diff
@@ -38,24 +37,10 @@ export async function maybeRedirect( context, next ) {
...
-		getQueryArgs()?.showLaunchpad === 'true'
-	) {
-		await context.store.dispatch( requestSite( siteId ) );
-	}
-
-	const refetchedOptions = getSiteOptions( context.store.getState(), siteId );
 	const shouldRedirectToLaunchpad =
-		refetchedOptions?.launchpad_screen === 'full' &&
+		launchpad_screen === 'full' &&
 		// Temporary hack/band-aid to resolve a stale issue with atomic that the requestSite
```
edanzer: Wondering if we need the 'temporary band-aid for atomic' part of this given the new endpoint?

client/data/sites/use-launchpad.ts
edanzer: Worth noting I tested this - by completing Launchpad, which turns the launchpad_screen option to off, and then immediately navigating to My Home. Seems to work fine. If I recall, RQ may automatticaly refetch data on window refreshes. Not sure how that interacts with the meta -> persist property here, but in any case, seems to be working just fine - no stale data when arriving on My Home.

edanzer:
This is testing as expected for me. Added a few comments. I'll also hold on approving as it's still in draft, and there's a convo happening about React Query vs using Redux data stores. Also some failing e2e tests.

It's nice to be able to delete all that extra code to handle stale state situations.

client/my-sites/customer-home/controller.jsx
agrullon95: We can remove the `shouldRedirectToLaunchpad` logic since `const { launchpad_screen, site_intent } = await fetchLaunchpad( slug );` should always retrieve the latest data.

client/data/sites/use-launchpad.ts
agrullon95: Looking at the React Query docs, it seems like the query will refetch when its mounted again or the user refocuses on the window by default: https://react-query-v3.tanstack.com/guides/caching

For now we may be fine with the react query default values

edanzer (approved):
Latest code changes look good. I tested this multiple times on Simple sites with different flows. I then tested multiple times on Atomic via different flows and upgrade paths. I also tested both prelaunch (when the home-to-launchpad redirect should work) and post-launch (when the redirect should not be active). In all cases, the My Home redirect worked when it should, and did not fire when it shouldn't. 

Only small issue I had was unrelated to the PR. On two separate atomic sites, I'm noticing if you upgrade to an Ecommerce plan, we always seem to load the wp-admin version of My Home, not the calypso version. In such cases, obviously the redirect doesn't work. I would force the issue by manually loading the calypso home url, and the redirect would then work. 

I usually test atomic via business plans (and force the atomic transfer by activating a plugin). 
I can't recall testing an ecommerce plan. I only did so here because I wanted to force the atomic transfer right away without extra steps. But because I haven't tested this before, it could be that loading the wp-admin version of the dashboard is standard behavior for ecommerce plan sites. But to the extent that's expected, our redirect in the my home controller won't fire on such sites.

This is nothing we need to address in this PR. If we did somehow want to address it, I think we would need to discuss this and do it via a very different PR.

edanzer:
Some e2e tests had been failing. I just re-ran them, and looks like they're good now.

sharpirate:
### Testing
✔️ Navigate to /home/{siteSlug} and verify that you are redirected to the launchpad screen
✔️ Navigate to /home/{siteSlug}?launchpadComplete=true and verify you are not redirected to the launchpad screen
✔️ Disable the launchpad (change site option launchpad_screen to off or FALSE) or use a site that is not launchpad enable and navigate to /home. Verify that you are not navigated to the launchpad screen.

sharpirate (approved):

PR #74383, opened 2023-03-13
VAT Info: Strip leading CHE from Swiss VAT ID

sirbrillig:
## Proposed Changes

It looks like the default VAT number format for Switzerland includes 'CHE-' at the front. It's optional, but if people are going to try it by default we should probably validate numbers with it there.

This PR modifies the code that submits VAT ID numbers to also strip a leading `CHE-` (in addition to the existing behavior which is to strip a leading `CH`, or whatever the country code is). Without this change, we actually will submit the ID with a leading `E-` (what's left after stripping the `CH`).

## Testing Instructions

Automated tests are included.


martian77:
👍 LGTM.

martian77 (approved):

PR #74384, opened 2023-03-13
Post Editor: ditch the iframe?

mattwiebe:
This will need to be updated after #75183 and whatever follows it to enable the Site Editor un-iframing in all envs.

----
This will redirect the Editors to the URL directly rather than trying to load it in the iframe.

Enabled in dev, horizon, wpcalypso, and calypso.live environments.

Related to #74413

## Proposed Changes

* This will redirect the `/page/SITE/PAGE_ID`, `/post/SITE/POST_ID`, and `/site-editor/SITE` URLs directly to the site in question rather than iframing them. 

## Testing Instructions

* Smoke test it!
* Especially check about using an incognito window to use the various redirects to an Atomic site. Are you logged-in to the Atomic site automatically in the redirect?

## Ramifications

There are many further steps to complete this migration. The first and biggest one is testing, as **I bet that this breaks a whole bunch of e2e tests** once we turn it on for everyone.

Other things:

1. We'll want to switch to classic URLs directly in the admin menu sidebar itself. Remove for [Site Editor](https://github.com/Automattic/jetpack/blob/512b38b57af5d416043d5cfa02677f9a84a75db5/projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php#L430-L448) and posts and pages are in that class too.
2. We can break these rewrites down to something that runs purely in the `/site-editor`, `post`, and `page` routes. Are there others to consider?
3. We need to fix the back button integration from the Post Editor - it's currently taking us to `wp-admin/edit.php`

## Pre-merge Checklist


tyxla (commented):
This change appears to introduce some regressions - specifically, the removal of features. The removal could be intentional or not, but we have to understand and assess it in any case. 

According to my understanding, by redirecting to wp-admin, we lose some functionality that we were adding on top of the iframe, for example:

- Media Library modal
- Revisions modal
- Editor checkout modal

Are we fine with losing that? Do we have a backup plan for getting those features into the admin and closing the feature gap?

Addison-Stavlo:
@tyxla - Definitely agreed!

We are going to open this as a CFT and identify the tradeoffs and weigh these options in P2, this change will initially be limited to staging environments. My understanding from previous discussions and plans over the last year+ is that this has been long planned for (ex. Help Center as a means to provide support without the iframe) and many of the remaining items that will be losing may be an acceptable tradeoff, but we aim to identify and discuss these tradeoffs and plans more thoroughly before we release this in production.

mattwiebe:
Thanks for looking @tyxla !

Let's run down the list of regressions:

> Media Library modal

It works. This was a prior blocker to iframe removal.

> Revisions modal

Users will just use the WP core revisions viewer. Do we miss out on anything substantive there?

> Editor checkout modal

Can you tell me more about this?

tyxla:
@mattwiebe, I'd suggest auditing for other missing features and feature gaps. I listed a few, but there might be more that I've missed. 

> > Media Library modal

> It works. This was a prior blocker to iframe removal.

It works, but it's a different one. In Calypso, we use the Calypso media modal, and in wp-admin, we use the core WordPress media modal. They are entirely different interfaces, and the Calypso one tends to be better for many flavors. 

Calypso:
Screenshot 2023-03-20 at 11 25 35

wp-admin:
Screenshot 2023-03-20 at 11 26 00

There are many differences and removing the Calypso one in favor of wp-admin one could leave some feature gaps. I think we should analyze those and take them into consideration.

There have been plans to extract the Calypso media editor as a separate app and reuse it in wp-admin, but that never materialized. cc @flootr who I believe was part of the preparatory work of that project, and cc @griffbrad for awareness and any additional context.

> > Revisions modal

> Users will just use the WP core revisions viewer. Do we miss out on anything substantive there?

It's worth auditing that part for feature gaps as well. I haven't been involved in the development of it, so I might be missing some nuances. In any case, it might be a better user experience to stay on the same page while viewing revisions, instead of having to go to a separate page like wp-admin does it.

> > Editor checkout modal

> Can you tell me more about this?

I haven't worked on it either, but I see @sirbrillig has touched recently, so maybe he can help with adding some context?

sirbrillig:
> Editor checkout modal

Surprisingly I actually don't know a ton about that project, although since the team that implemented it doesn't appear to be around any more (its [project milestone](https://github.com/Automattic/wp-calypso/milestone/322) looks abandoned) maybe it's become another thing dropped for @Automattic/shilling to pick up. It was implemented (I believe) by @Automattic/create (based on this thread: pbAok1-1zo-p2). @razvanpapadopol may have more information. 

FWIW, I don't know how to trigger the modal. I thought it occurred when you tried to purchase a premium block, but when I do that now I'm redirected to checkout.

If the checkout modal is unmaintained, I personally have no concerns about disabling it. iframes are never great, especially for checkout which often involves redirecting the user to external sites and back.

mattwiebe:
Thanks so much again @tyxla - you are correct in all you say there. We have nicer UIs in the Calypso media library and revisions interface, but maintained by a continually vexing iframe setup. It might be that we want to migrate one or both to wp-admin, before or after the iframe bridge is removed.

The point right now isn't that we're for sure going to do it, but we'd like to, and this current exercise is about taking stock of the tradeoffs, and seeing if there's any actual blockers to doing it. I wouldn't call those blockers, but maybe now somebody will get on the media library conversion to working directly in the Editor.

All that to say, this is precisely the first step of the audit you're rightly calling for, seeing what shakes out. Your input is super helpful, thank you!

mattwiebe:
@sirbrillig many thanks for all that context! It looks like we're still using that but the upgrade flow still works without it, just not as well.

yansern:
> FWIW, I don't know how to trigger the modal. I thought it occurred when you tried to purchase a premium block, but when I do that now I'm redirected to checkout.

So when you press "Upgrade" button on a premium block, it will call the Calypso parent that renders a checkout modal that covers the iframe editor underneath it.

The idea was that you could upgrade to premium plan without leaving editor. That's how I remembered it worked back then.

But now when I tried it again, even though the modal still shows, the subsequent checkout process seems to redirect the user out from the editor, and then back to the editor again, which defeats the idea of upgrading without leaving the editor.

> If the checkout modal is unmaintained, I personally have no concerns about disabling it. iframes are never great, especially for checkout which often involves redirecting the user to external sites and back.

How does the "Upgrade" button work on a non-iframed editor currently? Would be interesting to compare the flow and weigh the pros and cons and decide what work needs to be done.

mattwiebe:
@yansern Thanks for weighing in, I noticed the same thing. You can see the checkout experience with or without the iframe in my draft post at paYE8P-2CL-p2

Addison-Stavlo:
Note - I think there may be some ad-hoc URLs in places that dont respect the "ShouldLoadGutenframe" checks.  I ran into one potentially when I was testing this PR in combination with https://github.com/Automattic/wp-calypso/pull/74801

* Loading the post editor from the launchpad "write your first post" task first directs towards the iframe URL, then eventually loads into wp-admin.

I imagine launchpad tasks are just hardcoding the iframe URLs as opposed to using any other checks there, and I wonder if there are other places in the codebase that do similar.

Addison-Stavlo:
Another thing we will need to keep track of with this change:

With the iframe - the site editor exits directly to the launchpad. On this branch, it exits to /home and has to go through a flashing redirect.

mattwiebe:
> I imagine launchpad tasks are just hardcoding the iframe URLs as opposed to using any other checks there, and I wonder if there are other places in the codebase that do similar.

I think this should be fine, so long as the `redirect` function in the controller translates for us, which `should` happen just fine (and means we're not doing ad-hoc editor URL construction within components)

mattwiebe:
I will leave this open for now in the hopes that we will also un-iframe the Post Editor, but for now we are just un-iframing the Site Editor in #75183

mattwiebe:
Closing this, might circle back to it when we revisit un-iframing the Post Editor

PR #74394, opened 2023-03-13
Migrate legacy global styles to register from registerStore

noahtallen:
See #74399 for more info

## Proposed Changes
Very simple PR which migrates global styles from registerStore to createReduxStore. The store is only used in one file, as far as I can tell.

## Testing Instructions
1. Sandbox a test site and sync the ETK build to your sandbox
2. Activate a theme with "jetpack-global-styles" support, such as Leven, on that site.
3. Access the editor -- Global Styles can be found by the typography icon in the top toolbar.
4. Verify that changing the fonts changes the fonts in the editor.
5. Verify that publishing the changes changes the fonts in both the editor and on the front end of the site.

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/global-styles/index.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/global-styles/index.js
@@ -20,25 +20,25 @@ import './editor.scss';
 // Global variable.
 const { PLUGIN_NAME, STORE_NAME, REST_PATH } = JETPACK_GLOBAL_STYLES_EDITOR_CONSTANTS; // eslint-disable-line no-undef
 
-registerStore( STORE_NAME, REST_PATH );
-registerDOMUpdater( [ FONT_BASE, FONT_HEADINGS ], select( STORE_NAME ).getOption );
+const globalStylesStore = registerStore( STORE_NAME, REST_PATH );
+registerDOMUpdater( [ FONT_BASE, FONT_HEADINGS ], select( globalStylesStore ).getOption );
 
 registerPlugin( PLUGIN_NAME, {
 	render: compose(
 		withSelect( ( getSelectors ) => ( {
-			siteName: getSelectors( STORE_NAME ).getOption( SITE_NAME ),
-			fontHeadings: getSelectors( STORE_NAME ).getOption( FONT_HEADINGS ),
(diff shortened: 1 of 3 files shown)


apps/editing-toolkit/editing-toolkit-plugin/global-styles/index.js
```diff
@@ -20,25 +20,25 @@ import './editor.scss';
 // Global variable.
 const { PLUGIN_NAME, STORE_NAME, REST_PATH } = JETPACK_GLOBAL_STYLES_EDITOR_CONSTANTS; // eslint-disable-line no-undef
 
-registerStore( STORE_NAME, REST_PATH );
-registerDOMUpdater( [ FONT_BASE, FONT_HEADINGS ], select( STORE_NAME ).getOption );
+const globalStylesStore = registerStore( STORE_NAME, REST_PATH );
```
tyxla: I find this pretty confusing: one would think we are using the `@wordpress/data` `registerStore`, while this is our own data store registration function. I'd suggest refactoring that while we're here.

apps/editing-toolkit/editing-toolkit-plugin/global-styles/src/store.js
```diff
@@ -52,7 +52,7 @@ const actions = {
  * @param {string} optionsPath REST path used to interact with the options API.
  */
 export default ( storeName, optionsPath ) => {
-	registerStore( storeName, {
+	const store = createReduxStore( storeName, {
```
tyxla: I'd just declare `store` globally for the module and export it, then call `register()` just like we did for other PRs already. 

No need for the custom registration function IMHO.

noahtallen:
Done! Thanks for the push to keep it consistent with our other migrations 👍

apps/editing-toolkit/editing-toolkit-plugin/global-styles/src/store.js
```diff
@@ -34,86 +37,69 @@ const actions = {
 	},
 };
 
-/**
```
noahtallen, line 37: I removed this comment because it doesn't provide anything beyond what the type hints give us now.

tyxla (commented):
Code-wise this looks good. 👍 Thanks @noahtallen 

Still needs to be tested before shipping though.

noahtallen:
I updated the testing instructions and verified this is working well 👍

PR #74395, opened 2023-03-13
Remove unused data store

noahtallen:
See #74399 for more info

## Proposed Changes

As I was working on migrating some data stores to `createReduxStore`, I found that this data store (automattic/whats-new) is never used anywhere. It appears we migrated to using inline state instead. 

## Testing Instructions

- sync ETK to your sandbox and sandbox a test site.
- When loading the editor verify that the "what's new" feature works (it's available in the three-dot dropdown menu in the top-right.)


tyxla (approved):
LGTM 👍 

FWIW the `@automattic/whats-new` package is responsible for that functionality, and it now uses react query so even if that store was considered to be used, I don't think it will be. 

🚀

Issue #74399, opened 2023-03-13
Task: Migrate wp-data stores from registerStore to createReduxStore


noahtallen:
### Details

The main reason we should do this is that `registerStore` is deprecated. But the main benefit we get is the new approach to using stores gives us much better typescript integration. Previously, one would have to declare types to override the entire wp data package.

This is actually impossible to do now that the package provides its own typedefs, so we've had to rely on the temporary solution of casting to selector types like `(select( 'automattic/foo' ) as FooSelectors).selector()`. This is very clunky, and we also lost type info for `dispatch` calls. (See [this comment](https://github.com/Automattic/wp-calypso/pull/73890#discussion_r1131702449) for more info.)

As a result, we should migrate stores to use the "new" format (it's been available for a long time actually, though it wasn't supported by the WordPress versions we support at first.) And ideally, also start using stores via imports instead of via string names!

This will restore type data we lost in the upgrade, and also make types much easier to work with for WordPress data.

This is all made possible by our recent WordPress package updates done in #73890.

### Checklist

A store is registered in each of these files. See PRs for example migrations.

#### Editing Toolkit Plugin
- [x] `editing-toolkit-plugin/global-styles/src/store.js` @noahtallen #74394
- [x] `apps/editing-toolkit/editing-toolkit-plugin/whats-new/src/store.js` @noahtallen #74395 
- [x] `apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nav-sidebar/src/store.ts` @noahtallen #74396
- [x] `apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/store.ts` @noahtallen #74346

#### Calypso data-stores
- [x] Update documentation about how to use stores, which references the previous register approach. @noahtallen #74602
- [x] `packages/data-stores/src/analyzer/index.ts` @tyxla #74414
- [x] `packages/data-stores/src/auth/index.ts` @tyxla #74418
- [x] `packages/data-stores/src/automated-transfer-eligibility/index.ts` @tyxla #100472
- [x] `packages/data-stores/src/domain-suggestions/index.ts` @noahtallen #74551
- [x] `packages/data-stores/src/i18n/index.ts` @tyxla #74427
- [ ] `packages/data-stores/src/plans/index.ts`
- [x] `packages/data-stores/src/products-list/index.ts` @tyxla #74566
- [x] `packages/data-stores/src/reader/index.ts` @tyxla #74425
[...]

PR #74414, opened 2023-03-14
Data Stores: Migrate Analyzer store to `createReduxStore()`

tyxla:
## Proposed Changes

This PR migrates the Analyzer store to use `createReduxStore()` and `register()` instead of `registerStore()`.

Part of #74399. A follow-up to #73890.

## Testing Instructions

* The only step that uses this store is not currently enabled, so we have to manually enable it with a feature flag.
* Navigate to `/setup/site-setup/importLight?siteSlug=YOURSITE&flags=onboarding/import-light` where `YOURSITE` is the slug of one of your sites.
* If you land on the goals step, check the checkbox that you want to import your content into your site and push the button to continue forward. Retry the step above.
* Verify import process still works well.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/stores.ts
+++ b/client/landing/stepper/stores.ts
@@ -26,4 +26,4 @@ export const USER_STORE = User.register( {
 
 export const AUTOMATED_ELIGIBILITY_STORE = AutomatedTransferEligibility.register();
 
-export const ANALYZER_STORE = Analyzer.register();
+export const ANALYZER_STORE = Analyzer.store;
--- a/client/landing/stepper/stores.ts
+++ b/client/landing/stepper/stores.ts
@@ -6,7 +6,6 @@ import {
 	User,
 	AutomatedTransferEligibility,
 	StepperInternal,
-	Analyzer,
 } from '@automattic/data-stores';
 
 export const ONBOARD_STORE = Onboard.register();
@@ -25,5 +24,3 @@ export const USER_STORE = User.register( {
 } );
 
 export const AUTOMATED_ELIGIBILITY_STORE = AutomatedTransferEligibility.register();
-
-export const ANALYZER_STORE = Analyzer.store;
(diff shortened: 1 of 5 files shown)


flootr (approved):
Appears to be working well 🚀

client/landing/stepper/stores.ts
```diff
@@ -26,4 +26,4 @@ export const USER_STORE = User.register( {
 
 export const AUTOMATED_ELIGIBILITY_STORE = AutomatedTransferEligibility.register();
 
-export const ANALYZER_STORE = Analyzer.register();
+export const ANALYZER_STORE = Analyzer.store;
```
noahtallen: General thought for all Automattic data stores -- instead of this, should we just export the store object as `Analyzer`?

Then, when we're using anything in automattic/data-stores, it becomes simpler to use:

```ts
// or export it as analyzerStore, which is a bit clearer and follows our naming conventions
import { Analyzer } from '@automattic/data-stires';

useSelect( select => select( Analyzer ) )
```

(potentially avoiding the need to re-export the store from client/landing/stepper/stores too)

tyxla: I think for some stores we export additional modules and types, so it might be necessary to export some more on top of the store. For example:

https://github.com/Automattic/wp-calypso/blob/9651d4a39d5075bcb10cf1bd3c98f705766eb519/packages/data-stores/src/onboard/index.ts#L13-L15

Those are then re-exported by the `@automattic/data-stores` package and reused somewhere. 

I'd suggest approaching those in another PR, since they might require additional work, and I don't see that as something related to refactoring to `createReduxStore()`.

tyxla: I've removed the store export indirection in https://github.com/Automattic/wp-calypso/pull/74414/commits/1ce6df95120576cda20bad62721a450e0cdc496a

bogiii (approved):
Looks good to me! 🚢

client/landing/stepper/stores.ts
noahtallen: Fair enough! 👍

PR #74436, opened 2023-03-14
Staging Sites: Clear site excerpt query cache when new site is created

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/1966

## Proposed Changes

Clears the site excerpt query cache when a new staging site is created. Otherwise, the site selector search can be stale.

Redux cache doesn't need to be updated because the transfer process already fetches the new site object. https://github.com/Automattic/jetpack/pull/29453 was sufficient to resolve that part of the problem.

## Testing Instructions

2. Navigate to Hosting Configuration.
3. Click 'Add staging site'.
4. After the staging site is created, verify you can find it in the site switcher.


wojtekn (approved):
Works as expected.

PR #74443, opened 2023-03-14
Add more perf marks for logged in routes

noahtallen:
Related to #

## Proposed Changes
We don't quite have enough information to track down some of the recent response time spikes. This adds more data for the logged in route setup, which is currently a blind spot.

Not much needed for testing -- just verify calypso.live continues to work correctly.


lynnbgriffith (approved):
👍 Looks good. Curious to see how these turn out.

tyxla:
Thanks Noah!

PR #74447, opened 2023-03-14
Only show option to update contact info when appropriate

delputnam:
We should only show the "Manage contact info" option for domains that are registered with us and in a state where the contact info can be updated.

Cursor_and_Domains_‹_A8C_Test_Site_—_WordPress_com

Domains_‹_A8C_Test_Site_—_WordPress_com

### Test

- On a site with a mapped domain, make sure that this option does not appear in the list of actions for action  menu.
- Make sure that the option does appear for domains that are registered with us and in an updateable state.

### DIFF
--- a/client/my-sites/domains/domain-management/list/domain-row.jsx
+++ b/client/my-sites/domains/domain-management/list/domain-row.jsx
@@ -16,6 +16,8 @@ import PopoverMenuItem from 'calypso/components/popover-menu/item';
 import {
 	canCurrentUserAddEmail,
 	getDomainTypeText,
+	isDomainInGracePeriod,
+	isDomainUpdateable,
 	resolveDomainStatus,
 } from 'calypso/lib/domains';
 import { type as domainTypes, domainInfoContext } from 'calypso/lib/domains/constants';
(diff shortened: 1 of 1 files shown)


jeyip:
Will take a look tomorrow morning!

jeyip:
## Testing
### Requirements
- [x] On a site with a mapped domain, make sure that this option does not appear in the list of actions for action menu.
- [x] Make sure that the option does appear for domains that are registered with us and in an updateable state.

### Browsers
- [x] Chrome

jeyip (approved):

client/my-sites/domains/domain-management/list/domain-row.jsx
```diff
@@ -390,9 +392,13 @@ class DomainRow extends PureComponent {
 					<PopoverMenuItem icon="info-outline" onClick={ this.goToDNSManagement }>
 						{ translate( 'Manage DNS' ) }
 					</PopoverMenuItem>
-					<PopoverMenuItem icon="book" onClick={ this.goToEditContactInfo }>
-						{ translate( 'Manage Contact Info' ) }
-					</PopoverMenuItem>
+
+					{ domain.type === domainTypes.REGISTERED &&
+						( isDomainUpdateable( domain ) || isDomainInGracePeriod( domain ) ) && (
```
jeyip, line 397: `NB:`

Just to clarify, we want contact info to be editable if a domain isn't updateable but it _is_ in a grace period?

delputnam: It has to be both:
- registered with us
- either updatable or in the grace period.

leonardost (approved):
LGTM, thanks for this update!

PR #74452, opened 2023-03-14
Checkout: Migrate checkout data store to createReduxStore

sirbrillig:
## Proposed Changes

The `registerStore()` function in the `@wordpress/data` package is deprecated. This PR replaces its use in the checkout data store with the recommended `createReduxStore()` function. This will give us better TypeScript support amongst other things.

The most notable difference is that we must pass the store itself to `useSelect()` and `useDispatch()` which means that we can no longer create the store on-the-fly with a React hook inside checkout. Instead we must initialize the store as a side-effect of importing the store file. To preserve some of the original behavior, this adds a new `reset()` action to the store which will reset the store's data to its original values. However, since the store was already a singleton, this PR does not call the `reset()` method anywhere except tests. Hopefully that won't cause any issues.

Part of https://github.com/Automattic/wp-calypso/issues/74399

## Testing Instructions

Automated tests should cover everything. However, it's probably a good idea to make sure that the store updates as expected. For the most part the store is used to hold contact/billing fields inside checkout, so you can edit those fields and make sure that they change as you type. It may also be a good idea to add a domain to your cart and verify that the domain fields also work.

Screenshot 2023-03-15 at 10 47 30 AM

Screenshot 2023-03-15 at 10 50 15 AM


client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -361,13 +356,9 @@ export default function CheckoutMain( {
 		// Only wait for stored cards to load if we are using cards
 		( allowedPaymentMethods.includes( 'card' ) && isLoadingStoredCards );
 
-	const contactDetails = useSelect(
-		( select ) => ( select( 'wpcom-checkout' ) as WpcomCheckoutStoreSelectors )?.getContactInfo(),
-		[]
-	);
+	const contactDetails = useSelect( ( select ) => select( CHECKOUT_STORE )?.getContactInfo(), [] );
```
tyxla: With the new structure, we'll always have the store registered, no? When would `select( CHECKOUT_STORE )` actually return `undefined`?

This question applies to all pre-existing areas that use optional chaining on the `select()` result.

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
tyxla, line 1: This appears to be the wrong directory for the store registration file now. This is no longer a hook.

client/my-sites/checkout/composite-checkout/hooks/use-cached-domain-contact-details.ts
```diff
@@ -57,8 +52,7 @@ function useCachedContactDetailsForCheckoutForm(
 			? getCountryPostalCodeSupport( countriesList, cachedContactDetails.countryCode )
 			: false;
 
-	const checkoutStoreActions: CheckoutStoreMappedActions | undefined =
-		useDispatch( 'wpcom-checkout' );
+	const checkoutStoreActions = useDispatch( CHECKOUT_STORE );
```
tyxla, line 55: Noting that this intentionally introduces some type specificity regression, which will be resolved in the future when we upgrade to a newer version of `@wordpress/data`. We've explained that here: https://github.com/Automattic/wp-calypso/issues/74399#issuecomment-1469812339

client/my-sites/checkout/composite-checkout/test/checkout-main.tsx
```diff
@@ -47,6 +49,7 @@ describe( 'CheckoutMain', () => {
 	let MyCheckout;
 
 	beforeEach( () => {
+		dispatch( CHECKOUT_STORE ).reset();
```
tyxla, line 52: Alternatively, if we only have to do this for tests, we could use `jest.resetModules()` and re-import the store before every test, as demonstrated here: https://stackoverflow.com/a/48990084/3936800

I wonder why this is necessary, though. We should be able to mock specific selectors or actions in the store per-test if necessary, no?

client/my-sites/checkout/composite-checkout/test/use-cached-domain-contact-details.tsx
```diff
@@ -59,12 +59,11 @@ function MyTestWrapper( {
 }
 
 function MyTestContent( { countries }: { countries: CountryListItem[] } ) {
-	useWpcomStore();
 	const { responseCart, reloadFromServer, updateLocation } = useShoppingCart(
 		initialCart.cart_key
 	);
 	useCachedDomainContactDetails( () => null, countries );
-	const contactInfo = useSelect( ( select ) => select( 'wpcom-checkout' ).getContactInfo() );
+	const contactInfo = useSelect( ( select ) => select( CHECKOUT_STORE ).getContactInfo(), [] );
```
tyxla, line 66: It's interesting that TS missed that! Likely because it's in a test and nothing specifically imports it as a module.

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
```diff
@@ -125,14 +121,7 @@ const selectors = {
...
 
-export function useWpcomStore(): WpcomCheckoutStore | undefined {
-	// Only register once
-	const registerIsComplete = useRef< boolean >( false );
-	if ( registerIsComplete.current ) {
-		return undefined;
-	}
-	registerIsComplete.current = true;
-
+function createStore() {
```
tyxla: I'd suggest getting rid of the function. If we're going to call it just once, it may just inline its contents.

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
```diff
@@ -188,17 +177,24 @@ export function useWpcomStore(): WpcomCheckoutStore | undefined {
...
 		actions,
 		selectors,
 	} );
 }
+
+const store = createStore();
+
+register( store );
+
+export const CHECKOUT_STORE = store;
```
tyxla, line 196: That's a nice way to enforce the store export to have a particular unique name, cc @noahtallen

tyxla (commented):
This looks great, thank you @sirbrillig! 

I have only a few comments, mostly minor / reorg suggestions. 

Let me know what you think.

client/my-sites/checkout/composite-checkout/test/use-cached-domain-contact-details.tsx
sirbrillig: Yeah, TS does not scan the tests because they're not imported. I often find TS errors in them when I go to do something else.

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
sirbrillig: Good point. I've moved it.

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
sirbrillig: You're right! I've removed the guards.

client/my-sites/checkout/composite-checkout/test/checkout-main.tsx
sirbrillig: > we could use jest.resetModules() and re-import the store before every test

🤔  Interesting... I wonder if that might cause challenges based on the other things being imported/mocked in these tests though. An explicit reset of the store feels a little more clear to me than needing to re-mock a ton of selectors.

> I wonder why this is necessary, though. We should be able to mock specific selectors or actions in the store per-test if necessary, no?

These tests are integration tests so they are designed to operate as close to "real behavior" as possible to cover a lot of functionality. If we mock parts of the store there's a risk that a regression could occur in the store itself without being caught by these tests.

In most cases when checkout is loaded, this store will be empty (users do not often visit checkout, fill out the forms, then leave checkout, then come back) so these tests are designed to start from that premise. That way we can test behavior like auto-fill without needing to clear specific data added by other tests (or to put it another way, this is clearing that specific data).

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
sirbrillig: That's a good point.

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
sirbrillig: Thanks. I thought I copied this pattern from another one of the diffs off of https://github.com/Automattic/wp-calypso/issues/74399 but... now I'm not sure if maybe I invented it accidentally?

noahtallen: True, we could even still use `export const CHECKOUT_STORE = createReduxStore( ... )` etc!

I wonder if we should decide on a naming pattern to use. I've been using a format like `import { store as checkoutStore }`

noahtallen (approved):
This tests nicely for me! Thanks for the cleanup :)

client/my-sites/checkout/composite-checkout/hooks/wpcom-store.ts
tyxla: I think we should go with the `checkoutStore` style, since `All_CAPS` style assumes a constant, but this is technically no longer a constant.

PR #74454, opened 2023-03-15
Migrate block editor nux to createReduxStore

noahtallen:
see #74399 for more info

## Proposed Changes
Migrates the block editor nux (new user experience) ETK feature to `createReduxStore`

While there are other references to `automattic/wpcom-welcome-guide` in the codebase, they're in different webpack bundles, so I continued using the string-based store identifier approach.

## Testing Instructions

TBD... you definitely start by syncing ETK. Create a new account and sandbox the new test site. At that point it might show up?

Added a couple people who recently touched these files for tips and maybe a review ;)

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/store.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/store.js
@@ -1,5 +1,5 @@
 import apiFetch from '@wordpress/api-fetch';
-import { combineReducers, registerStore } from '@wordpress/data';
+import { combineReducers, createReduxStore, register } from '@wordpress/data';
 import { apiFetch as apiFetchControls, controls } from '@wordpress/data-controls';
 
 import 'a8c-fse-common-data-stores';
@@ -138,12 +138,12 @@ export const selectors = {
 	getShouldShowFirstPostPublishedModal: ( state ) => state.shouldShowFirstPostPublishedModal,
 };
 
-export function register() {
-	return registerStore( 'automattic/wpcom-welcome-guide', {
-		reducer,
-		actions,
-		selectors,
-		controls,
-		persist: true,
-	} );
-}
+export const store = createReduxStore( 'automattic/wpcom-welcome-guide', {
+	reducer,
+	actions,
+	selectors,
+	controls,
+	persist: true, // TODO @noahtallen: persist plugin probably doesn't work here.
+} );
+
+register( store );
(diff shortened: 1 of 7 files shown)


apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/welcome-tour/tour-launch.tsx
```diff
@@ -153,16 +147,10 @@ function WelcomeTour( { siteIntent }: { siteIntent?: string } ) {
 			tourRating: {
 				enabled: true,
 				useTourRating: () => {
```
noahtallen, line 157: this use of hooks is interesting 🤔

p-jackson: They told us React was "just JavaScript", so why not hooks as config 😄

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/store.js
```diff
@@ -141,12 +141,12 @@ export const selectors = {
...
-		controls,
-		persist: true,
-	} );
-}
+export const store = createReduxStore( 'automattic/wpcom-welcome-guide', {
+	reducer,
+	actions,
+	selectors,
+	controls,
+	persist: true, // TODO @noahtallen: persist plugin probably doesn't work here.
```
p-jackson, line 146: We do persist store data. This is it what the store looks like without the ETK patch.

CleanShot 2023-03-15 at 21 06 09@2x

And we do rely on it for caching so we're not always fetching the "should they see the welcome tour" flag from the server forever into the future.

And you're right it no longer appears to work with this patch applied.

p-jackson (commented):
> Create a new account and sandbox the new test site. At that point it might show up?

You can also choose "Welcome Guide" from the kebab menu to make it reappear.

Everything else is working as expected

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/welcome-tour/tour-launch.tsx
noahtallen: haha, fair enough! It stood out since I hadn't seen hooks used like this before :P

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/store.js
noahtallen: Unfortunate! I think the persist plugin may not have been migrated to work with the new store registration, so that may require more work before we can migrate this one.

Thanks for taking a look!

noahtallen: for context, the persistence plugin definitely isn't compatible with the more modern registration methods :( See https://github.com/WordPress/gutenberg/issues/11449

Another question is whether we could use the editor preferences layer instead, which does support browser persistence

noahtallen:
Labeling as blocked until we get more info on what to do about the persistence plugin

tyxla:
With ETK no longer a thing, this can be closed.

PR #74455, opened 2023-03-15
Update/hide GitHub on sites page for users with no access

p-jackson:
## Proposed Changes

The new developer sub-menu has launched to all users pesfPa-3t-p2#comment-41, but the "Deploy from GitHub" feature has not, see: https://github.com/Automattic/wp-calypso/blob/9a2605e41d93ad5e5ddf2f24e0959a1bb19484f9/client/my-sites/hosting/main.js#L180

This PR

* Refactors `<QueryReaderTeams>` so we can use it's logic in hooks
* Hides the "Deploy from GitHub" menu item unless the user is an a11n

## Testing Instructions

Test on a12n and non-a12n accounts. Confirm nothing has changed for non-atomic and non-business sites.

## Pre-merge Checklist

### DIFF
--- a/client/components/data/query-reader-teams/index.jsx
+++ b/client/components/data/query-reader-teams/index.jsx
@@ -2,11 +2,14 @@ import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 import { requestTeams } from 'calypso/state/teams/actions';
 
-export default function QueryReaderTeams() {
+export function useQueryReaderTeams() {
 	const dispatch = useDispatch();
 	useEffect( () => {
 		dispatch( requestTeams() );
 	}, [ dispatch ] );
+}
 
+export default function QueryReaderTeams() {
+	useQueryReaderTeams();
 	return null;
 }
--- a/client/sites-dashboard/components/sites-ellipsis-menu.tsx
+++ b/client/sites-dashboard/components/sites-ellipsis-menu.tsx
@@ -18,13 +18,16 @@ import { useI18n } from '@wordpress/react-i18n';
 import { addQueryArgs } from '@wordpress/url';
 import { ComponentType, useEffect, useMemo, useState } from 'react';
 import { useDispatch as useReduxDispatch, useSelector } from 'react-redux';
+import { useQueryReaderTeams } from 'calypso/components/data/query-reader-teams';
 import SitePreviewLink from 'calypso/components/site-preview-link';
 import { useSiteCopy } from 'calypso/landing/stepper/hooks/use-site-copy';
 import TrackComponentView from 'calypso/lib/analytics/track-component-view';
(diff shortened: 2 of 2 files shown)


client/components/data/query-reader-teams/index.jsx
```diff
@@ -2,11 +2,14 @@ import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 import { requestTeams } from 'calypso/state/teams/actions';
 
-export default function QueryReaderTeams() {
+export function useQueryReaderTeams() {
 	const dispatch = useDispatch();
```
mpkelly, line 6: I think we ditch the `useEffect` and go with this: 

```js
dispatch( requestTeams() );
```

mpkelly (approved):
I tested this successfully. For a non-a12n Business account, the GH option is now hidden (see below). The rest works as before. Nice work catching this. 

image

client/components/data/query-reader-teams/index.jsx
p-jackson: I don't think so, wouldn't this re-request every render?

mpkelly: I think it will do that already as `dispatch` is not stable and will change every render.

mpkelly: Yeah, this is how `useDispatch` is created:

```js
export function createDispatchHook(context) {
  if (context === void 0) {
    context = ReactReduxContext;
  }

  var useStore = context === ReactReduxContext ? useDefaultStore : createStoreHook(context);
  return function useDispatch() {
    var store = useStore();
    return store.dispatch;
  };
}
```
So I guess `useEffect` should be removed or should have an empty dependencies array.

p-jackson: I read that code and to me it looks like the value of `dispatch` will never change 😅 

> The `dispatch` function reference will be stable as long as the same store instance is being passed to the `<Provider>`. Normally, that store instance never changes in an application.

https://react-redux.js.org/api/hooks#usedispatch

The other reason we want to run it in an effect is it moves it out of the high-priority part of the React lifecycle and into the low-priority part, see https://github.com/Automattic/wp-calypso/pull/72868#discussion_r1098113617

p-jackson: > should have an empty dependencies array.

I should also say that this is correct, we don't actually need `dispatch` in the dependencies array because it makes no difference in the way it behaves. It's only there to silence the linter :) Fortunately the linter is clever enough to know that some functions (like a setter returned by `useState()`) doesn't need to go in the dependencies.

mpkelly: Yeah, I did read the code closely enough. We create a new function for each render, but its return value is stable.

PR #74463, opened 2023-03-15
Hosting Configuration: Add margin between SSH keys and 'Add' button

danielbachhuber:
## Proposed Changes

Adds a bit of margin between the listed SSH keys and the 'Attach SSH key to this site' button.

### No SSH keys

image

### Other user has a SSH key

image

### Two SSH keys

image

## Testing Instructions

2. Add another administrator to your site and sign into their account via incognito window or otherwise.
3. Play around with different variations of attached SSH keys.

### DIFF
--- a/client/my-sites/hosting/sftp-card/ssh-keys.scss
+++ b/client/my-sites/hosting/sftp-card/ssh-keys.scss
@@ -2,7 +2,9 @@
 	max-width: calc(100vw - 32px);
 }
 
-.ssh-keys__form {
+.ssh-keys__form.form-fieldset {
+	margin-top: 16px;
+	margin-bottom: 0;
 	display: flex;
 	flex-direction: column;
 	align-items: flex-start;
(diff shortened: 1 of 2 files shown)


wojtekn (approved):
Looks good, works as expected.

katinthehatsite (commented):
Hi @danielbachhuber  - it worked fine for me with the changes that you implemented both on mobile and on desktop versions. 

One thing that I noticed and this might not be in the scope of this PR is that the SSH Keys box does not size well on mobile so the `Detach` button is not visible:

Screenshot 2023-03-17 at 10 16 27 AM

danielbachhuber:
> One thing that I noticed and this might not be in the scope of this PR is that the SSH Keys box does not size well on mobile so the `Detach` button is not visible:

@katinthehatsite Great catch! Want to open a new issue for that and work on it when you have a moment?

katinthehatsite:
Sure!

PR #74411, opened 2023-03-14
Tailored Flows: Use stepper's domain and plans for NL flow

renancarvalho:
## Proposed Changes

Newsletter flow depends on /start for domains and plans step, last year team Vertex moved [domains and plans into stepper](pdDR7T-nj-p2) and initially applied this change only on Link in Bio flow. This PR applies it on Newsletter flow.

## Testing Instructions

- Pull and run this branch or use calypso live links
- Navigate to `/setup/newsletter/intro`
- Complete the Newsletter flow
- Note that, on Domains and Plans step, the user should never be redirect to `/start` besides for login IF logged out
	- Domains step URL should be similar to `/setup/newsletter/domains...`  and not `start/newsletter/domains....`
	- Plans step URL should be similar to `/setup/newsletter/plans...` and not `start/newsletter/plans-newsletter...`

- Some cases to double check: 
	- Logged out user 
	- Free plan, free domain and/or both

- Bonus test ❤️  :
 	- Test Link In Bio flow should be intact.

Fixes #74378

PR #74466, opened 2023-03-15
Tailored Flows: /start clean up

renancarvalho:
## Proposed Changes

The first release of Newsletter and LIB flow was using `/start`, [now it is not anymore](https://github.com/Automattic/wp-calypso/pull/74411). This is a cleanup PR removing LIB and mainly Newsletter code from `/start` since it is not being used anymore.

## Testing Instructions
This is a bit tricky since it is hard to keep track of all the flows using `/start`

I have in mind the following

Case 1:

- Test NL and/or LiB flow won't make much sense since it is not using /start anymore, but worths a second check, although I am not expecting issues there.

Case 2: Test `/start` which means create a flow from scratch.

- Pull and run this branch or use calypso live links
- Navigate to `/start` for instance `{calypsoLiveLink or Localhost}/start`, that should render the domains step since there's no specific flow.
- Try to complete the flow.
- There should be no console erros
- There should be no Newsletter or LiB copies on domains and plans step, copies should be more "generic"


client/signup/tailored-flow-processing-screen/index.jsx
```diff
@@ -14,21 +12,7 @@ const useSteps = ( flowName ) => {
...
-			];
-			break;
-		case NEWSLETTER_FLOW:
-			steps = [
-				{ title: __( 'Excellent choices. Nearly there!' ) },
-				{ title: __( 'Smoothing down the stationery' ) },
-				{ title: __( 'Embossing all the envelopes' ) },
-			];
-			break;
+		// perhaps this can be removed in the future.
```
heavyweight: I think this component is no longer used and we should remove it

heavyweight:
Tested the newsletter and link in bio for regressions and they look good to me

client/signup/main.jsx
```diff
@@ -727,10 +723,6 @@ class Signup extends Component {
 			return <P2SignupProcessingScreen signupSiteName={ this.state.signupSiteName } />;
 		}
 
-		if ( isNewsletterOrLinkInBioFlow( this.props.flowName ) ) {
```
heavyweight, line 730: Here is where it was it's single usage

renancarvalho: Thank you @heavyweight 
Component removed :)

escapemanuele (changes requested):
So glad when we remove things!

Could we remove [these styles](https://github.com/Automattic/wp-calypso/blob/tailored_flows_cleanup/client/signup/signup-header/style.scss#L22-L24) as well? ✌️

renancarvalho:
> So glad when we remove things!
> 
> Could we remove [these styles](https://github.com/Automattic/wp-calypso/blob/tailored_flows_cleanup/client/signup/signup-header/style.scss?rgh-link-date=2023-03-16T16%3A23%3A58Z#L22-L24) as well? ✌️

Sure thing! thank you for checking :)

escapemanuele (approved):
Thanks for addressing my feedback! Good to go!

PR #74469, opened 2023-03-15
Tailored Flows: Fix uppercase label style

renancarvalho:
## Context

Some flows are using TextControl from `@wordpress/components`, these components have a css rule that sets the label text-transformation property to uppercase.

## Proposed Changes

Fixes uppercase site name on flows.

Before:

image

After: 

image

## Testing Instructions

- Pull and run this branch or use calypso live link
- Navigate to `/setup/link-in-bio/` flow.
- On Personalize step the site name should not be uppercase
- This change is applied to all flows.


heavyweight (approved):
Thanks for fixing this!!

PR #74474, opened 2023-03-15
Add tracks to plans block

unDemian:
Related to [#](323-gh-Automattic/lighthouse-forums)

## Proposed Changes

Fixes 323-gh-Automattic/lighthouse-forums by adding a track record for upgrade block page views.

## Testing Instructions

There should be a tracks request for `calypso_happyblocks_upgrade_plan_view` only for forum posts that have the upgrade block.
Markup on 2023-03-15 at 4:32:17 PM

*

## Pre-merge Checklist

### DIFF
--- a/apps/happy-blocks/block-library/pricing-plans/view.tsx
+++ b/apps/happy-blocks/block-library/pricing-plans/view.tsx
@@ -1,4 +1,5 @@
 import './view.scss';
+import { recordTracksEvent } from '@automattic/calypso-analytics';
 import domReady from '@wordpress/dom-ready';
 import { useState, render } from '@wordpress/element';
 import useI18nCalypsoTranslations from '../shared/use-i18n-calypso-translations';
@@ -31,6 +32,10 @@ function renderPricingPlansBlock( element: HTMLElement | Element ) {
 }
 
 domReady( () => {
+	recordTracksEvent( 'calypso_happyblocks_upgrade_plan_view', {
+		location: window?.location?.href,
+	} );
+
 	document
 		.querySelectorAll( '.a8c-happy-tools-pricing-plans-block-placeholder' )
 		.forEach( renderPricingPlansBlock );


apps/happy-blocks/block-library/pricing-plans/view.tsx
```diff
@@ -31,6 +32,10 @@ function renderPricingPlansBlock( element: HTMLElement | Element ) {
 }
 
 domReady( () => {
+	recordTracksEvent( 'calypso_happyblocks_upgrade_plan_view', {
```
msurdi-a8c, line 35: I think doing this here will record the event every single time this file is loaded, despite the block is used or not.

To confirm my understanding, can you try this branch on a page which doesn't have any instance of this block? If I'm correct the event will be firing anyways, despite there's no usage of it.


If it is fine to record N events, if there are N usages of the block in the page (suppose the block is used on a Topic's body and then again in the replies) then you could move this to `renderPricingPlansBlock `. That function won't run if there are no usages.

unDemian: > I think doing this here will record the event every single time this file is loaded, despite the block is used or not.
> To confirm my understanding, can you try this branch on a page which doesn't have any instance of this block? If I'm correct the event will be firing anyways, despite there's no usage of it.

I have tried it while developing and was also part of the testing instructions, and it is not the case, the [documentation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/#view-script) is a bit vague, but I think it makes sense to not call all blocks' view scripts all the time.

> If it is fine to record N events, if there are N usages of the block in the page (suppose the block is used on a Topic's body and then again in the replies) then you could move this to renderPricingPlansBlock . That function won't run if there are no usages.

I thought about this a bit, my thinking was more in the context of page views so I went with counting a single time per page no matter how many blocks are there, but if it makes more sense we can track per block, I can update it cc: @naxoc

msurdi-a8c: Makes sense then! Thanks for double checking it.

msurdi-a8c (approved):

PR #55772, opened 2021-08-26
Domains: Fix bug when switching sites while on the domain search page

leonardost:
#### Changes proposed in this Pull Request

When you were on the domain search page and switched to another site, the navigation would break. It seems there are some reference issues with the `SearchWithTyper` component, so we'll fix this for now by removing it temporarily.

Video showing the bug:

https://user-images.githubusercontent.com/5324818/131029125-d0af440c-eb04-46dc-a3f1-7fdd08886a9d.mp4

#### Testing instructions

Open the live Calypso link and navigate to the domain search page on a site. Switch to another site and ensure nothing breaks and that no reference warnings appear on the dev console of your browser.

PR #74476, opened 2023-03-15
Domains: Remove unused `SearchWithTyper` component

tyxla:
While I was working on removing the unused Gutenboarding code in #74475, I spotted that the only usage of the `useTyper` hook is in an unused `SearchWithTyper` component. It hasn't been in use for over 18 months (since #55772 landed).

## Proposed Changes

Remove the unused component. We can always re-add it, but given how long it's been unused, chances are low. And even if we need to bring it back, we can always use the code history.

## Testing Instructions

Verify the removed code is not in use.


leonardost (approved):
LGTM! Thanks for the cleanup!

PR #74477, opened 2023-03-15
Checkout: Pre-check Northern Ireland checkbox if saved VAT info is for XI

### LINKED ISSUE #73407: VAT details in checkout does not prefill XI checkbox
When pre-filling the contact details step in checkout, if there is saved VAT information, the "Northern Ireland" checkbox does not get pre-checked, which causes the form to fail validation until it is checked manually.

Screenshot 2023-02-15 at 5 46 42 PM

sirbrillig:
## Proposed Changes

When pre-filling the billing information step's VAT section in checkout, if there is already saved VAT information and the saved country code is `XI` (Northern Ireland's fake country code for tax purposes), the "Northern Ireland" checkbox should be pre-checked. This diff makes that happen.

Fixes https://github.com/Automattic/wp-calypso/issues/73407

Before:

Screenshot 2023-03-15 at 1 08 44 PM

After:

Screenshot 2023-03-15 at 1 09 33 PM

## Testing Instructions

Automated tests are included but you can try this manually.

- Use an account without VAT info saved. Use Payments Admin to remove VAT info from the account if it is already set.
- Click to edit the billing information step and change the country to "United Kingdom" and enter a valid postal code like `GU16 7HF`.
- Press "Continue" to save the address.
- Find a valid Northern Ireland VAT ID and save it to your account using Payments Admin or https://wordpress.com/me/purchases/vat-details. You can find a valid one by searching the database for VAT info for the country `XI`.
- Reload checkout.
- Click to edit the billing information step.
- Verify that the checkbox labeled "Is the tax ID for Northern Ireland?" is pre-checked.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/vat-form/index.tsx
@@ -106,8 +106,8 @@ export function VatForm( {
 				...vatDetailsFromServer,
 				// Initialize the VAT country in this form data to match the country in
 				// the parent form, which may differ from the country in the saved VAT
-				// details.
-				country: countryCode,
+				// details unless Northern Ireland is saved.
+				country: vatDetailsFromServer.country === 'XI' ? vatDetailsFromServer.country : countryCode,
 			} );
 			// Pre-check the checkbox to show the form when the country is supported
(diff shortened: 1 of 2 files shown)


martian77 (approved):
Works as described and makes sense to me.

PR #74386, opened 2023-03-13
Remove verify email banner from launchpad

edanzer:
Related Issue: https://github.com/Automattic/wp-calypso/issues/73868
Related PR: https://github.com/Automattic/wp-calypso/pull/74135

### Proposed Changes

This PR removes the Email Verification Banner from Launchpad. We've added a Verify Email task to the Launchpad tasks for Newsletters ([see PR](https://github.com/Automattic/wp-calypso/pull/74135)), and no longer need the banner. 
  - REMOVE COMPONENT FILE. The EmailVerificationBanner is not used anywhere else, so I've completely removed the file. 
  - REMOVE TEST FILE. Tests for step-content.tsx were entirely focused on testing this banner, so I've removed that test file completely. We might consider additional tests for that file, but that can be handled in a separate PR.
 
email verification banner
  
### Testing 

**Review Time: Short**
**Testing Time: Medium**

1. Check out this brand and run yarn and yarn start if needed. 
2. Log out of WordPress.com or use a fresh browser where you are not logged in. 
3. Set up a temporary email at https://temp-mail.org/
4. Create a new newsletter site at http://calypso.localhost:3000/setup/newsletter/intro. You'll need to create a new account. Use the email setup in Step 3. 
5. Click through to Launchpad. 
6. TEST: Confirm the email verification banner no longer shows, even when your email is not verified.
7. TEST: Smoke test other aspects of Launchpad to confirm it continues to work as expected. Consider verifying your email, and coming back to Launchpad, and again smoke test to confirm everything works as expected. 
8. TEST: Confirm Launchpad unit tests still pass by running `yarn run test-client client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/` from your local repository root.

PR #74489, opened 2023-03-15
Launchpad: Add basic tests for step-content.tsx file

edanzer:
### Proposed Changes

In a [recent PR](https://github.com/Automattic/wp-calypso/pull/74386), we removed tests for this file because they were focused on a single component (EmailVerificationBanner) that was removed. I noted I would follow up and add at least some basic test coverage for this file. 

This PR adds basic rendering tests for the Newsletter flow. If we think it's worth adding other flows and/or adding interactivity, I can do so. Note that tests in this file are somewhat redundant with tests we have in place for child components (sidebar, checklist, checklist item). So the goal here is just to cast a wide net and ensure the general content and behavior we expect for Launchpad as a whole is there. 

### Testing Instructions

**Review Time: Short**
**Testing Time: Short**

To test, checkout this branch locally, run the following line, and confirm all tests pass (1 test suite, 5 tests):
`yarn run test-client client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/test/step-content.tsx`

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/test/step-content.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/test/step-content.tsx
@@ -74,6 +74,8 @@ function renderStepContent( emailVerified = false, flow: string ) {
 }
 
 describe( 'StepContent', () => {
+	// To get things started, test basic rendering for Newsletter flow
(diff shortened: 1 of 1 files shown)


sharpirate (approved):
✔️ The tests are passing and the code makes sense

agrullon95 (approved):

Issue #74008, opened 2023-03-02
Stats: Create the "Most popular post" card for the Insights page


dognose24:
# Development
Tackle the `all-time` most popular post data via updating or creating specific endpoints. The data query may come across a few DB tables.

- [x] Apply data to the [existing MostPopularPostCard component](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/stats/all-time-highlights-section/most-popular-card.tsx).
( https://github.com/Automattic/wp-calypso/pull/74495 )
- [x] When the latest post is also the most popular, just display the most popular at full width
( https://github.com/Automattic/wp-calypso/pull/76775 )
- [x] Add mobile support using the dotPager component.
( https://github.com/Automattic/wp-calypso/pull/74495 )

--- Updated ---
- [x] [Add a placeholder/loading visual for the Most popular post card](https://github.com/Automattic/wp-calypso/issues/74008#issuecomment-1476639537) during data load instead of having the Latest post card expand to take up the entire width.

# Design
Create the most popular post that is not tied to any specific author.

### PC
<img width="644" alt="截圖 2023-03-03 上午12 50 23" src="https://user-images.githubusercontent.com/6869813/222497338-1029d10e-9ab7-4484-8f86-6badb3789a8b.png">

### Mobile
<img width="606" alt="截圖 2023-03-03 上午1 01 45" src="https://user-images.githubusercontent.com/6869813/222499651-80e7c17b-8196-4474-9bb4-dc6042212841.png">

jsnmoon:
Note that it's currently not possible to obtain an all-time most popular post.

dognose24:
The data period to determine the most popular post was proposed in the discussion: `pejTkB-nM-p2`.

dognose24:
[Feedback from the PR comment](https://github.com/Automattic/wp-calypso/pull/74495#pullrequestreview-1346847427) of @jsnmoon:

> In the future, it'd be nice to see a placeholder/loading visual for the `Most popular post` card during data load instead of having the `Latest post` card expand to take up the entire width:

While Loading|While Loading (desired)|After Loading
-|-|-
[...]

PR #74495, opened 2023-03-15
Stats: Restore the most popular post with the data of the last 365 days

dognose24:
Related to #74008 partially

## Proposed Changes

* Change the `top posts` data period from `7 days` to `1 year` as proposed in the related discussion (`pejTkB-nM-p2`).
* Merge `LatestPostCard` and `MostPopularPostCard` into `PostCardsGroup` for applying the `DotPager` component.
* Remove feature flags `stats/latest-post-stats` and `stats/most-popular-post`.

## Testing Instructions

* Spin up the change with the Live Branch.
* Navigate to the Stats `Insights` page.
* Ensure the card of the `Most popular post` is displayed with the data of the last 365 days.
* Ensure the mobile layout of cards aligns with the design.

|PC|Mobile|
|-|-|
|截圖 2023-03-17 上午1 29 50|截圖 2023-03-17 上午1 30 26|

## Pre-merge Checklist


kangzj:
I think we should probably change the title to `Most popular post in the past year` instead?

-----UPDATE-------
I also notice when the home page is the most popular post, the section would be hidden. Do you think we should exclude the home page?

dognose24:
> I think we should probably change the title to Most popular post in the past year instead?

Nice catch! @kangzj 

> I also notice when the home page is the most popular post, the section would be hidden. Do you think we should exclude the home page?

Yeah, the most popular `post` should exclude the home page. What do you think? @jeffgolenski @keoshi

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
```diff
@@ -100,10 +102,48 @@ export default function MostPopularPostCard( {
...
+					likeCount={ mostPopularPostData?.likeCount }
+					post={ mostPopularPostData }
+					viewCount={ mostPopularPostData?.viewCount }
+					commentCount={ mostPopularPostData?.commentCount }
+					titleLink={ `/stats/post/${ mostPopularPost.ID }/${ siteSlug }` }
+					uploadHref={
+						! isOdysseyStats ? `/post/${ siteSlug }/${ mostPopularPost.ID }` : undefined
+					}
+				/>
+			) }
```
jsnmoon: It might be good to refactor this out of the `mobileCardsGroup` and `cardsGroup` declarations to avoid regressions in the future (e.g., where we update one block but forget to update the other).

Something like this:

```js
	const cards = (
		<>
			<LatestPostCard siteId={ siteId } siteSlug={ siteSlug } />
			{ ! isTopViewedPostRequesting && mostPopularPost && (
				<PostStatsCard
					heading={ translate( 'Most popular post in the past year' ) }
					likeCount={ mostPopularPostData?.likeCount }
					post={ mostPopularPostData }
					viewCount={ mostPopularPostData?.viewCount }
					commentCount={ mostPopularPostData?.commentCount }
					titleLink={ `/stats/post/${ mostPopularPost.ID }/${ siteSlug }` }
					uploadHref={
						! isOdysseyStats ? `/post/${ siteSlug }/${ mostPopularPost.ID }` : undefined
					}
				/>
			) }
		</>
	);

	const mobileCardsGroup = (
		<DotPager>
			{ cards }
		</DotPager>
	);

	const cardsGroup = (
			{ cards }
		</div>
	);
```

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
```diff
@@ -53,40 +54,41 @@ const getStatsQueries = createSelector(
 );
 
 const getStatsData = createSelector(
-	( state, siteId, topPostsQuery ) => {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	( state, siteId, topPostsQuery, isTopViewedPostRequesting ) => {
```
jsnmoon, line 59:
```suggestion
	( state, siteId, topPostsQuery ) => {
```

It should be fine to exclude `isTopViewedPostRequesting` from the selector function since we're not using it.

`isTopViewedPostRequesting` will still be available for the cache-busting function since `getStatsData` is invoked with it as its fourth parameter on line 91.

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
```diff
@@ -53,40 +54,41 @@ const getStatsQueries = createSelector(
...
 		return {
 			topPost,
 			topPage,
 		};
 	},
-	( state, siteId, topPostsQuery ) => [ topPostsQuery ]
+	( state, siteId, topPostsQuery, isTopViewedPostRequesting ) => [
+		topPostsQuery,
+		isTopViewedPostRequesting,
+	]
```
jsnmoon, line 71: It might be nice to move the "Refresh memoized data from `createSelector` with the loading flag" code comment to annotate this block.

jsnmoon (commented):
Looks good overall; I left some code quality feedback inline. This is ready to ship after addressing any design feedback. 👍 

I also agree with Jasper in that we should exclude the home page from consideration when determining the "Most popular post in the past year".

jeffgolenski:
@dognose24 You're a genius. This is looking wonderful

Screen Shot 2023-03-16 at 3 39 32 PM

Screen Shot 2023-03-16 at 3 43 05 PM

Screen Shot 2023-03-16 at 3 44 03 PM

Screen Shot 2023-03-16 at 3 42 10 PM

Just a few items I noticed:

- [ ] As Jasper mentioned, we definitely want to exclude the homepage otherwise that would dominate the view all the time
- [ ] Card with no featured images: Let's change the button size so the "Add featured image" is on one line. Simply add `is-compact` to the button classes and we should be good there!
- [ ] Mobile: The featured image seems to get cut off in the top right (see above). Can we just fix that so it's 100% in view?

Thanks my friend!

keoshi (commented):
This looks absolutely fantastic, @dognose24!

I've spotted a couple of things that I'd like your thoughts on before approving the PR:

- [x] **Incorrect data?**

Reading "Most popular post in the past year" and knowing that this module is looking at the past 365, the data I'm getting doesn't look right. Take this site for instance:

image

It tells me that this post has had a total of 63 views for the past 365 days. But rather than being the aggregate number for the past year (past 365 days), it seems to be taking account only the current 2023 year (19 + 31 + 13). Notice the data I'm getting on the chart:

January | February | March
--|--|--
19 | 31 | 13
image | image | image

- [x] **Timestamp**

It was a bit odd to see this data format associated with my site: `Published 7/23/2012` — seems to be using the US notation of `MM/DD/YYYY`.

It would be wonderful if we could use the site's own setting for `Data format` or, at least, reuse the same that's in other places in Calypso, like the Post/Pages listing:

image

- [x] **Counts alignment**

When the post titles in "Latest post" and “Most popular post” differ in size it's not unusual for the counts section to be misaligned between them like so:

image

This tweak seems to do the trick:

```diff
.post-stats-card__counts {
+  align-self: flex-end;
}
```

- [x] **Latest post summary visible**

I remember this happening before so not sure if it's just me again or not, but the old “Latest post summary” is still visible at the end of the page:

image

Thank you for pushing this out and for always being accommodating of our feedback. 🙏

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
dognose24: Yeah, I do definitely want to exclude it. In fact, it is [intended to be left](https://github.com/Automattic/wp-calypso/pull/74495/commits/5896bf076be7a6cde8c0ff50fe54cbab2778bd9a) to avoid another warning for the second argument that would break the unit tests. Would you happen to have any idea about the warning? @jsnmoon

截圖 2023-03-18 上午12 08 46

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
dognose24: Yes, I did miss it! 😅

dognose24: Yes! I did miss it. 😅 The DotPager seems only to accept the components array. [7b7442d](https://github.com/Automattic/wp-calypso/pull/74495/commits/7b7442dd1970dfbb353716ec163fe9d61908fd18)

dognose24:
Thanks for the detailed reviews! @jsnmoon @jeffgolenski @keoshi

About the home page, in fact, the [`post_type` of the most popular post is always `post`](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/stats/lists/selectors.js#L286). In other words, the home page cannot be the most popular post since its `post_type` is `homepage`.

Those issues mentioned above should be all good now, which means this PR can use another round of review! 🙇🏼‍♂️

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
jsnmoon: I see; that's undesirable.

It'd be great to fix this, but fixing the typing error is more trouble than it's worth. Let's ignore this for now.

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
```diff
@@ -100,10 +104,29 @@ export default function MostPopularPostCard( {
...
+				likeCount={ mostPopularPostData?.likeCount }
+				post={ mostPopularPostData }
+				viewCount={ mostPopularPostData?.viewCount }
+				commentCount={ mostPopularPostData?.commentCount }
+				titleLink={ `/stats/post/${ mostPopularPost.ID }/${ siteSlug }` }
+				uploadHref={ ! isOdysseyStats ? `/post/${ siteSlug }/${ mostPopularPost.ID }` : undefined }
+				locale={ userLocale }
+			/>
+		);
+	}
```
jsnmoon: You'll want to define a `key` property to each element for use within `cardsGroup`. Otherwise, you'll get these errors in local development:

image

jsnmoon (commented):
Looks good overall.

In the future, it'd be nice to see a placeholder/loading visual for the `Most popular post` card during data load instead of having the `Latest post` card expand to take up the entire width:

While Loading|While Loading (desired)|After Loading
-|-|-
image|image|image

keoshi (approved):
Thanks for addressing all the feedback. The module behaves as expected and looks amazing! 🚢

client/my-sites/stats/all-time-highlights-section/post-cards-group.tsx
dognose24: Resolved in [b112018](https://github.com/Automattic/wp-calypso/pull/74495/commits/b112018005e3a5a84dcaebfbcfac768738862ea3).

jsnmoon (approved):

jeffgolenski:
Looking great, thanks for the quick changes! @dognose24 as @jsnmoon  mentioned, perhaps we should open up an immediate follow up PR to build in that placeholder. On data heavy sites, like Jetpack.com, I'm noticing a several second delay before the full width Latest post snaps to 50% and the Most popular post card displays?

jsnmoon:
Let's ship this and add the loading cards in a follow-up PR.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7882687

Thank you @dognose24 for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7882687).

PR #74398, opened 2023-03-13
Migrate wpcom global styles to createReduxStore

noahtallen:
See #74399 for more info

## Proposed Changes

Migrates wpcom global styles from `registerStore` to `createReduxStore`. (The former is deprecated and the latter provides better typescript integration).

## Testing Instructions

TBD -- suggestions on the best way to test?

PR #74500, opened 2023-03-16
Migrate wpcom global styles to createReduxStore

noahtallen:
See #74399 for more info.

I had to revert the initial change (#74398), because it didn't account for the persistence plugin not working :(

That makes this blocked by this issue in Gutenberg: https://github.com/WordPress/gutenberg/issues/11449#issuecomment-1470877989

Hoping a good workaround gets suggested :)

## Proposed Changes

Migrates wpcom global styles from `registerStore` to `createReduxStore`. (The former is deprecated and the latter provides better typescript integration).

## Testing Instructions

TBD -- suggestions on the best way to test?


tyxla:
With ETK no longer a thing, this can be closed.

PR #74472, opened 2023-03-15
Added Woo Express Plans page

nelsonec87:
Closes https://github.com/Automattic/wp-calypso/issues/74126

Designs: jBwlERMS350NqhJNHZzAVF-fi-98%3A33334

## Proposed Changes

* Added new Plans/My Plan page for WX Medium plans

## Testing Instructions

Desktop

* Get a site with the WX Medium plan
   * Start with a free site
   * Go to `/checkout/<site-slug>/wooexpress-medium-bundle-monthly`
   * Finish checkout
* Go to the Plans page (`/plans/<site-slug>`)
* With the feature flag enabled (`/plans/<site-slug>?flags=plans/wooexpress-medium`) you should see the new page.

Figma:
image

HTML:
image

Mobile
* Change the view to mobile mode

Figma: 
image

HTML:
image

PR #74501, opened 2023-03-16
Fix duplicate calypso live comments

noahtallen:
## Proposed Changes
Recently, I’ve noticed some PRs get duplicate calypso.live links. (example: #74472)

We have two different places which can add calypso.live links. Firstly, in TeamCity, after the image is generated, a script will add a comment. Secondly, in GitHub actions, a script will add a comment.

Why do we have two? Well, there is an edge case where the TeamCity job will execute before you have a PR created (because it triggers on commit push.) In this scenario, you’d never get a comment about calypso.live. Additionally, it can be helpful for GitHub actions to add the comment immediately, when the PR is opened, rather than waiting several minutes for the TeamCity job to finish.

But this only impacts PR creation. After that, the TeamCity job will maintain a single comment, keeping it updated. This job _also_ deletes any existing comment from GitHub, to avoid having duplicates.

The issue with duplicate commits happens when the commit gets pushed, and TeamCity starts. You start writing your PR. You post it. Right afterwards, TeamCity gets to the point where it posts the comment, and does so. (It doesn't find an existing comment to delete either). Immediately after, the GitHub action posts its comment, because it did this after PR creation.

Essentially, we need to check in the GitHub action if a comment exists already, and if it does, skip.

## Testing Instructions

I tested that it does not post a comment if one exists. (By changing the trigger event to something else.) I think this is good enough for this PR 👍

### DIFF
--- a/.github/workflows/calypso-live.yml
+++ b/.github/workflows/calypso-live.yml
@@ -2,7 +2,7 @@ name: Calypso Live
 
 on:
   pull_request:
-    types: ['opened']
+    types: ['synchronize']
 
 jobs:
   calypso-live:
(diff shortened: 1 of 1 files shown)


.github/workflows/calypso-live.yml
```diff
@@ -12,18 +12,34 @@ jobs:
...
+              repo: context.repo.repo,
+            } );
+            const existingComment = comments?.find( comment => comment?.body?.contains( "calypso-live-watermark:apr@v1" ) );
+            return existingComment ? "true" : "false";
       - name: Build calypso.live link.
         run: |
-          echo '::set-output name=LINK::https://calypso.live?image=registry.a8c.com/calypso/app:commit-${{ github.event.pull_request.head.sha }}'
-          echo '::set-output name=QR_CALYPSO::https://chart.googleapis.com/chart?chs=150x150&cht=qr&chl=https%3A%2F%2Fcalypso.live%3Fimage%3Dregistry.a8c.com%2Fcalypso%2Fapp%3Acommit-${{ github.event.pull_request.head.sha }}%26flags%3Doauth&choe=UTF-8'
-          echo '::set-output name=QR_JETPACK::https://chart.googleapis.com/chart?chs=150x150&cht=qr&chl=https%3A%2F%2Fcalypso.live%3Fimage%3Dregistry.a8c.com%2Fcalypso%2Fapp%3Acommit-${{ github.event.pull_request.head.sha }}%26env%3Djetpack%26flags%3Doauth&choe=UTF-8'
+          echo 'name=LINK::https://calypso.live?image=registry.a8c.com/calypso/app:commit-${{ github.event.pull_request.head.sha }}' >> $GITHUB_STATE
```
noahtallen: The old ::set-output syntax was deprecated, so I switched it over. (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

.github/workflows/calypso-live.yml
```diff
@@ -12,18 +12,34 @@ jobs:
...
       - name: Post comment on PR
-        uses: actions/github-script@v3
+        uses: actions/github-script@v6
         with:
+          # Skip adding a comment if we already have one.
+          if: steps.has_comment.outputs.result == 'false'
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            github.issues.createComment({
+            github.rest.issues.createComment({
```
noahtallen, line 46: We have to access it under the rest namespace in newer versions of the action.

tyxla:
Thanks for this @noahtallen, I've been seeing that behavior in many of my recent PRs as well.

Do we have a better way to test this, other than merging it and trying with a new PR?

noahtallen:
You might be able to manually run the workflow, but when I did so, it didn't use the changes from this branch. I did test the workflow myself and verified the behavior when comments already exist. I think merging would be the best way to continue, since it's lower impact

.github/workflows/calypso-live.yml
```diff
@@ -12,18 +12,38 @@ jobs:
...
+            const comments = await github.rest.issues.listComments( {
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            } );
+            // For debugging, if needed!
+            console.log( comments );
+            const existingComment = comments.data.find( comment => comment.body.includes( "calypso-live-watermark:apr@v1" ) );
+            const hasComment = existingComment ? "true" : "false";
+            console.log( `Has existing comment: ${ hasComment }` );
```
tyxla, line 31: Should we clean those `console.log()` calls up before shipping?

tyxla (approved):
Alright, let's try this out 👍 🚀

.github/workflows/calypso-live.yml
noahtallen: I wanted to keep them because they provide helpful info when a build fails. It's only in a dev build log, so the only reason you'd check it is if you wanted to get information about what the build was doing. So I think it makes sense to keep them here. (Otherwise, debugging issues gets harder because you have to spin up a new branch with your own logs)

Issue #74422, opened 2023-03-14
`/subscriptions/settings`: Create client-side data interface for retrieving subscription settings.


mashikag:
Since this is going to a new API endpoint I would encourage using react-query for integration (see related conversation: p1678807696397609-slack-C7YPUHBB2).

The call to the new API should be integrated into [@automattic/data-stores](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores).

TimBroddin:
@mashikag am I making the right assumptions here?

- we should add a new store to [@automattic/data-stores](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores) 
- the hooks to call React-Query's useQuery and populate the store shoud live in our `subscription-manager`, in useQuery's onSuccess method we dispatch to the store (like they do here: p9oQ9f-1qT-p2)

mashikag:
> we should add a new store to [@automattic/data-stores](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores)

It depends on what you mean by adding a new store. For now, we will just add a new subdirectory with our queries. Similarly to [`starter-designs-queries`](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores/src/starter-designs-queries), or to [`support-queries`](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores/src/support-queries).

mashikag:
> in useQuery's onSuccess method we dispatch to the store (like they do here: p9oQ9f-1qT-p2)

For now, I do not see a point of storing this user state in a Redux store, especially when it comes to our external subscription management portal.

PR #74504, opened 2023-03-16
Subscription management: Add UserSettings container component

### LINKED ISSUE #74420: `/subscriptions/settings`: Create `UserSettings` container component
As per the agreed `SubscriptionManager` compound component structure proposal, the component is to be exported as: 
```jsx
SubscriptionManager.UserSettings
``` 
**Note:** This is a suggestion. The name is not strictly binding.

The component should be designed in mind that most likely, on the calypso end, it will be imported and integrated with a HOC similar to [`withFormBase`](https://github.com/Automattic/wp-calypso/blob/f2e6f88ae65c8571bbac21b284c29b48ec987598/client/me/form-base/with-form-base.jsx). 

## Acceptance criteria:
- the component is exported, from the `@automattic/subscription-manager` package, as a subcomponent of `SubscriptionManager` compound component
- the component should be rendered on the page when the **Settings** tab is selected, and/or when the path is `/subscriptions/settings`
- the component should accept at least the following props:
```typescript
type UserSettingsProps = {
  value?: UserSettingsValue;
  onChange?: ( value: UserSettingsValue ) => void;
  loading?: boolean;
  onError?: ( error: string ) => void;
}
```

mashikag:
Resolves  https://github.com/Automattic/wp-calypso/issues/74420

Screen Shot 2023-03-16 at 11 59 29

## Proposed Changes

* Introduce `UserSettings` container component with its roughly defined props
* Adjust bottom margin on `TabsSwitcher` as per Figma design

## Testing Instructions

* Go to http://calypso.localhost:3000/subscriptions/settings

## Pre-merge Checklist

### DIFF
--- a/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
+++ b/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
@@ -1,8 +1,9 @@
 @import "@automattic/color-studio/dist/color-variables";
+@import "@wordpress/base-styles/breakpoints";
 
 .section-nav {
 	box-shadow: none !important;
-	border-bottom: 32px;
+	margin-bottom: 32px !important;
 
 	.section-nav-tabs {
 		display: flex;
@@ -58,5 +59,11 @@
 				}
 			}
 		}
+	}
+}
+
+@media screen and ( max-width: $break-medium ) {
+	.section-nav {
+		margin-bottom: 22px !important;
 	}
 }
--- a/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
+++ b/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
@@ -1,9 +1,9 @@
 @import "@automattic/color-studio/dist/color-variables";
 @import "@wordpress/base-styles/breakpoints";
 
-.section-nav {
-	box-shadow: none !important;
-	margin-bottom: 32px !important;
+.section-nav.subscription-manager-tab-switcher {
+	box-shadow: none;
+	margin-bottom: 32px;
 
 	.section-nav-tabs {
 		display: flex;
@@ -63,7 +63,7 @@
 }
 
 @media screen and ( max-width: $break-medium ) {
-	.section-nav {
-		margin-bottom: 22px !important;
+	.section-nav.subscription-manager-tab-switcher {
+		margin-bottom: 22px;
 	}
 }
--- a/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
+++ b/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
@@ -1,18 +1,18 @@
 type SubscriptionUserSettings = Partial< {
-	subscription_delivery_mail_option: 'html' | 'text';
-	subscription_delivery_day: number; // 0-6, 0 is Sunday
-	subscription_delivery_hour: number; // 0-23, 0 is midnight
-	subscription_delivery_email_blocked: boolean;
+	mail_option: 'html' | 'text';
+	delivery_day: number; // 0-6, 0 is Sunday
+	delivery_hour: number; // 0-23, 0 is midnight
+	blocked: boolean;
 } >;
 
 type UserSettingsProps = {
-	value?: SubscriptionUserSettings;
+	value: SubscriptionUserSettings;
 	onChange?: ( value: SubscriptionUserSettings ) => void;
-	loading?: boolean;
+	loading: boolean;
 };
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- until we start using any of these props
-const UserSettings = ( props: UserSettingsProps ) => (
+const UserSettings = ( { value = {}, loading = false }: UserSettingsProps ) => (
 	<div className="user-settings">User settings will go here</div>
 );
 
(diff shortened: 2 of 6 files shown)


packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -0,0 +1,19 @@
+type SubscriptionUserSettings = Partial< {
+	subscription_delivery_mail_option: 'html' | 'text';
+	subscription_delivery_day: number; // 0-6, 0 is Sunday
+	subscription_delivery_hour: number; // 0-23, 0 is midnight
+	subscription_delivery_email_blocked: boolean;
+} >;
```
TimBroddin, line 6: One thing I'm thinking here is that the values returned from the API are much less verbose: `mail_option`, `delivery_day`, `delivery_hour`, `blocked`. Since it's already in a type containing Subscription in it's name, I would lose the prefixes. I think this would be more readable.

packages/subscription-manager/src/components/TabsSwitcher/styles.scss
TimBroddin, line 1: Is there a way to make these selectors more unique and thus losing the `!important` rule. I tried wrapping it all in `.subscription-manager-container` and that works, another option would be to wrap `<Tabs>` inside a wrapper div.

The reason for this is twofold: !important rules can make CSS specificity hard, and this stylesheet would get included in every bundle that includes `TabSwitcher.tsx `, thus overriding the default `.section-nav ` margin there.

Currently it's only included on the Subscription management - but just imagine what would happen if a developer starts looking around the codebase and thinks: this is a nice Tab component, I'll use it on my other Calypso page. Boom ... the margin of `.section-navs` would be 32px everywhere!

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type UserSettingsProps = {
+	value?: SubscriptionUserSettings;
+	onChange?: ( value: SubscriptionUserSettings ) => void;
+	loading?: boolean;
```
TimBroddin, line 11: Small one: but I think loading can be required, and be set a default value

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -0,0 +1,19 @@
+type SubscriptionUserSettings = Partial< {
+	subscription_delivery_mail_option: 'html' | 'text';
+	subscription_delivery_day: number; // 0-6, 0 is Sunday
+	subscription_delivery_hour: number; // 0-23, 0 is midnight
+	subscription_delivery_email_blocked: boolean;
+} >;
+
+type UserSettingsProps = {
+	value?: SubscriptionUserSettings;
```
TimBroddin, line 9: Could this ever be null? Since this is coming from Redux, the initialState (before data fetch) could be an empty object - this would satisfy the Partial<> (SubscriptionUserSettings).

TimBroddin (commented):
I left some small comments 👋

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
mashikag: Nice! I forgot we have the API already there. I was basing the naming on the current implemetnation in https://wordpress.com/me/notifications/subscriptions. 🤦‍♂️ Thank you for spotting this, and let me correct it.

JuanLucha: Regarding this, is it worth it (for the sake of variable naming convention) to transform to camel case when we receive the settings from the API? For example, I mean storing the value of `subscription_delivery_mail_option` into `subscriptionDeliveryMailOption` (or just `mailOption` as suggested by @TimBroddin ).

packages/subscription-manager/src/components/TabsSwitcher/styles.scss
mashikag: > Currently it's only included on the Subscription management - but just imagine what would happen if a developer starts looking around the codebase and thinks: this is a nice Tab component, I'll use it on my other Calypso page. Boom ... the margin of .section-navs would be 32px everywhere!

Good point! 🎯 

> Is there a way to make these selectors more unique and thus losing the !important rule. I tried wrapping it all in .subscription-manager-container and that works, another option would be to wrap <Tabs> inside a wrapper div.

Thanks for providing the solution. 🙇‍♂️ Let me implement this.

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
mashikag: Sure, sounds good to me. 👍

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
mashikag: I believe we may not use Redux for the external portal, and we will use `react-query` only. With the current requirements, there is no need to store it in global UI state.

mashikag: Nevertheless, we can make it required and pass an empty object by default. I have no objections to that.

TimBroddin: Ok, in that case we wouldn't need to use @automattic/data-stores I think. We could just create hooks that wrap useQuery. 

Referencing this ticket: https://github.com/Automattic/wp-calypso/issues/74422

TimBroddin (approved):
Approved! 🎊

PR #74510, opened 2023-03-16
Add a GS limit notice in the Global Styles sidebar

BogdanUngureanu:
Just like the other Global Styles limit notices, the new notice is shown only for free and personal plans. The implementation is based on the existing noticed by refactoring the logic we used in the pre-save GS notice.

Related to https://github.com/Automattic/dotcom-forge/issues/1233

## Proposed Changes

* Add a new notice in the Global Styles sidebar based on the `ComplementaryArea/core/edit-site` slot.

## Testing Instructions

* Use ETK build from this branch on your Sandbox
* On a free Simple Site, go to site editor and click on the GS button
* Notice that the upgrade nudge is not displayed
* Make a GS change
* Go to Global Styles sidebar and notice that we now display an upgrade nudge on all pages. 

## Pre-merge Checklist


apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/screen-root-notice.tsx
```diff
@@ -0,0 +1,41 @@
...
+export function GlobalStylesScreenRootNotice() {
+	const { isVisible } = useGlobalStylesConfig();
+
+	useEffect( () => {
+		if ( isVisible ) {
+			recordUpgradeNoticeShow( GLOBAL_STYLES_SCREEN_ROOT_CONTEXT );
+		}
+	}, [ isVisible ] );
+
+	if ( ! isVisible ) {
```
BogdanUngureanu: This prevents the notice to be shown when the user doesn't have any custom styles. However, this means it might not be visible enough for users:

- When a user first click on the GS button, the notice is NOT shown.
- They go to a specific menu and make a change.
- They close the menu since they are satisfied with the changes.

In order for them to see the notice, they'll have to click again on the GS panel; but since they don't plan to make any further changes, the's a smaller chance for them to notice that the changes they've made are actually limited (seeing the notice).

BogdanUngureanu: I've left it this way since the notice needs a design anyway and removing that behavior is simple.

BogdanUngureanu:
There's a TS error because it fails to recognize `__experimentalGetCurrentGlobalStylesId`, `getEditedEntityRecord`and `__experimentalGetDirtyEntityRecords`.


TypeScript fails me :(. I'll make the adjustments after the code review.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/screen-root-notice.tsx
dsas: Maybe we'll want to display different messages depending on if you have any GS vs not having any GS anyway 👍

BogdanUngureanu: Fair enough! I'll leave it as it is for now.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+	useEffect( () => {
+		if ( isVisible && isGlobalStylesSidebar ) {
+			recordUpgradeNoticeShow( GLOBAL_STYLES_SIDEBAR );
+		}
+	}, [ isVisible, isGlobalStylesSidebar ] );
+
+	return (
+		<Fill name="ComplementaryArea/core/edit-site">
+			{ !! isVisible && isGlobalStylesSidebar && (
```
BogdanUngureanu: We'll need to do the condition here because if we are doing an early return, the fill will be introduced at the bottom of the page, which means some additional CSS magic needs to be done.

This way, the notice is added at the top of the sidebar.

dsas:
This change means that the sidebar will always be taller than the page - we're displaying content above the global styles sidebar which has a css rule of `.edit-site-global-styles-sidebar { min-height:100% }`.

In turn this means that due to an existing GB bug, a horizontal scrollbar will always be displayed for some combinations of browsers and scrollbar settings (see https://github.com/WordPress/ gutenberg/issues/29912#issuecomment-900000689).

Here's a video of it in action: 
https://user-images.githubusercontent.com/93301/227530057-1d7b82a6-8dfb-4e95-923b-eb1e553ade9c.mp4

It'd be nice if we could so something about this but I don't think it's a blocker. The same thing happens when looking at the global styles block list.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
dsas: This would be super useful as a code comment

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
mmtr, line 1: Does this really need to be a `.tsx` file?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+			recordUpgradeNoticeShow( GLOBAL_STYLES_SIDEBAR );
+		}
+	}, [ isVisible, isGlobalStylesSidebar ] );
+
+	return (
+		<Fill name="ComplementaryArea/core/edit-site">
+			{ /*
+			We'll need to do the condition here because if we are doing an early return, the fill will be introduced at the bottom of the page, which means some additional CSS magic needs to be done.
+			*/ }
+			{ !! isVisible && isGlobalStylesSidebar && (
```
mmtr: Isn't `isVisible` already a boolean? If so, you shouldn't need to cast it with `!!`.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+		[]
+	);
+
+	const isGlobalStylesSidebar = GLOBAL_STYLES_SIDEBAR === area;
+
+	const isVisible = useGlobalStylesConfig().isVisible;
+
+	useEffect( () => {
+		if ( isVisible && isGlobalStylesSidebar ) {
+			recordUpgradeNoticeShow( GLOBAL_STYLES_SIDEBAR );
```
mmtr: The parameter is used as a context for the event which should be `site-editor` since this is being tracked in the site editor.

If you want to differentiate it from the other event tracked when the pre-save notice shows up, you'd probably want a new event name.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/use-global-styles-config.js
```diff
@@ -0,0 +1,32 @@
...
+			? __experimentalGetCurrentGlobalStylesId()
+			: null;
+		const globalStylesRecord = getEditedEntityRecord( 'root', 'globalStyles', _globalStylesId );
+
+		const globalStylesConfig = {
+			styles: globalStylesRecord?.styles ?? {},
+			settings: globalStylesRecord?.settings ?? {},
+		};
+
+		// Do not show the notice if the use is trying to save the default styles.
```
mmtr: This comment should be more generic, since `useGlobalStylesConfig` doesn't know whether `isVisible` is being used for a notice.

Actually, I think this property should be called `globalStylesInUse` or something like that, since that's really the information we're returning here.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/use-global-styles-config.js
```diff
@@ -0,0 +1,32 @@
...
+		const isVisible =
+			Object.keys( globalStylesConfig.styles ).length ||
+			Object.keys( globalStylesConfig.settings ).length;
+
+		return {
+			globalStylesConfig,
```
mmtr: `globalStylesConfig` is never used externally, so we can probably avoid returning it here.

mmtr (commented):
This work well!

It'd be great if the notice wouldn't have the top margin, so it's right at top the sidebar.

I also consider that the current pre-save notice is no longer relevant, so I'd probably get rid of it (and possibly re-use the same copy in the new permanent notice since it looks a bit clearer).

Screenshot 2023-03-30 at 16 00 29

> This change means that the sidebar will always be taller than the page

IMO, I think we should fix this before shipping it, and ensure that the sidebar doesn't have the scrollbar if it is not needed:


https://user-images.githubusercontent.com/1233880/228861468-ec47240d-8fcf-4d11-a9ac-9a33fdd2de93.mov

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/use-global-styles-config.js
BogdanUngureanu: Good idea, done! :)

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: Added new events and also added a blog_id property. :)

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: It needed to be typecasted in the use function. This should simplify a bit the condition.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: Personal preference. :) I've seen that more and more folks are migrating to TS. :)

mmtr: Let's solve all the TS issues then and ensure the file doesn't have any `@ts-ignore` comment.

BogdanUngureanu:
> It'd be great if the notice wouldn't have the top margin, so it's right at top the sidebar.

Fixed it! :) 

> I also consider that the current pre-save notice is no longer relevant, so I'd probably get rid of it (and possibly re-use the same copy in the new permanent notice since it looks a bit clearer).

I think there's still value in showing the notice in the PrePublish panel too. We've seen that upgrade notices in the editor historically don't convert well; one theory we had was that the user might not want to interrupt their work on their site to upgrade to a paid plan.

> IMO, I think we should fix this before shipping it, and ensure that the sidebar doesn't have the scrollbar if it is not needed:

I've tried again to remove the scrollbar but without much luck (since the changes would require at least overriding the existing styling from Gutenberg core - which will make the implementation less robust). 

IMO I think this styling should be handled in Core because the slot we are using (which is a public API) doesn't accommodate multiple UI elements in that section.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ExternalLink, Fill, Notice } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { createInterpolateElement, useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { recordUpgradeNoticeSidebarShow, recordUpgradeSidebarNoticeClick } from './tracks-events';
+import { useGlobalStylesConfig } from './use-global-styles-config';
+
+const GLOBAL_STYLES_SIDEBAR = 'edit-site/global-styles';
+
+export function GlobalStylesSidebarNotice() {
```
mmtr: Ideally, the returned value by this function should be typed.

Additionally, is there any reason why the component is not exported as `default` similarly to the other [two](https://github.com/Automattic/wp-calypso/blob/cc032392b4bbac06203b0e7cdcb3392bda7ca7cc/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/modal.js) [components](https://github.com/Automattic/wp-calypso/blob/cc032392b4bbac06203b0e7cdcb3392bda7ca7cc/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.js)?

BogdanUngureanu:
Looks like the tests are failing :(

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: I've added the ts-ignore because I'm getting some errors even though that shouldn't happen:
```
TS2339: Property 'getActiveComplementaryArea' does not exist on type 'never'.
```

Which is strange... because I don't think I'm doing anything weird to trigger that error. I've had the same issues with `__experimentalGetCurrentGlobalStylesId` functions too, which is why I kept using a JS file for `use-global-styles-config.js`.

mmtr: Where are you getting that error? I tried removing the `ts-ignore` rule and everything looks good according the IDE and the pre-commit hooks.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,55 @@
+declare const wpcomGlobalStyles: { upgradeUrl: string; blogId: string };
+
+import { ExternalLink, Fill, Notice } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { createInterpolateElement, useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { recordUpgradeNoticeSidebarShow, recordUpgradeSidebarNoticeClick } from './tracks-events';
+import { useGlobalStylesConfig } from './use-global-styles-config';
+
+const GLOBAL_STYLES_SIDEBAR = 'edit-site/global-styles';
```
mmtr, line 10: Nitpick (non-blocking): This is only used once now, so we don't really need a reusable const anymore.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+	return (
+		<Fill name="ComplementaryArea/core/edit-site">
+			{ /*
+			We'll need to do the condition here because if we are doing an early return, the fill will be introduced at the bottom of the page, which means some additional CSS magic needs to be done.
+			*/ }
+			{ globalStylesInUse && isGlobalStylesSidebar && (
+				<div className="interface-complementary-area">
+					<Notice status="warning" isDismissible={ false } className="wpcom-global-styles-notice">
+						{ createInterpolateElement(
+							__( '<a>Upgrade your plan</a> to keep your style changes.', 'full-site-editing' ),
```
mmtr: I'm unsure about saying "keep" here, since the styles are never lost – they are just not visible on the frontend, so they are technically kept even if you don't upgrade. 

Could we maybe reuse the same wording used in the other notice to say something like `Your changes include customized styles that will only be visible once you <a>upgrade to a Premium plan</a>`?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: That's strange. I'm getting the error in the IDE.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/tracks-events.ts
```diff
@@ -0,0 +1,43 @@
+import { recordTracksEvent } from '@automattic/calypso-analytics';
+
+declare const wpcomGlobalStyles: { upgradeUrl: string; wpcomBlogId: string | null };
+
+/**
+ * Record an event when a user clicks on the notice from the PrePublish panel.
+ */
+export function recordUpgradePrePublishNoticeClick(): void {
```
mmtr: Nitpick (non-blocking): The site editor uses the concept of saving instead of publishing, so it might be good to stay consistent here.

```suggestion
 * Record an event when a user clicks on the notice from the pre-save panel.
 */
export function recordUpgradePreSaveNoticeClick(): void {
```

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
mmtr: Do you get the error if you run `yarn typecheck`? If you don't, you can probably ignore the IDE error – it's maybe misconfigured.

mmtr: Actually, you should run `yarn run tsc --project apps/editing-toolkit`.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu: It's a personal preference to use constants instead of magic strings.

BogdanUngureanu:
Looks like the merge is not blocked anymore, but I still have two issues:
The WPCOM_Global_Styles_Test::test_wpcom_block_global_styles_frontend is failing on line 30:
```1) WPCOM_Global_Styles_Test::test_wpcom_block_global_styles_frontend
Failed asserting that null matches expected 'hotpink'.

/var/www/html/wp-content/plugins/editing-toolkit-plugin/wpcom-global-styles/test/class-wpcom-global-styles-test.php:30
```

~There's a JS error in the console when you click on a GS menu item:~ Looks like it's unrelated. It's happening for Atomic sites on production.

```
caught TypeError: Cannot read properties of undefined (reading 'find')
    at Object.trackGlobalStylesMenuSelected [as handler] (wpcom-block-editor-global-styles-menu-selected.js:18:1)
    at delegate-event-tracking.js:116:1
    at Array.forEach (<anonymous>)
    at __webpack_modules__../src/wpcom/features/tracking/delegate-event-tracking.js.__webpack_exports__.default (delegate-event-tracking.js:111:1)
    at HTMLDocument.delegateCaptureListener (tracking.js:865:1)
```

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -60,6 +60,26 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 	return true;
 }
 
+/**
+ * Get the WPCOM blog id of the current site for tracking purposes.
+ */
+function wpcom_global_styles_get_wpcom_current_blog_id() {
```
mmtr, line 66: Can we update `wpcom_should_limit_global_styles` to use this new function and avoid having duplicated code?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -60,6 +60,26 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
...
+		 * Atomic sites have the WP.com blog ID stored as a Jetpack option. This code deliberately
+		 * doesn't use `Jetpack_Options::get_option` so it works even when Jetpack has not been loaded.
+		 */
+		$jetpack_options = get_option( 'jetpack_options' );
+		if ( is_array( $jetpack_options ) && isset( $jetpack_options['id'] ) ) {
+			return (int) $jetpack_options['id'];
+		}
+	}
+
+		return null;
```
mmtr: Minor linting issue.
```suggestion
	return null;
```

mmtr:
I found a regression, not sure if that is related with the test failure.

Previously, you could switch between style variations and the GS panel remained open:

https://user-images.githubusercontent.com/1233880/229159564-750e9d82-2ae6-47bd-be3b-d3f356cccea1.mov



However, with these changes, the GS panel is closed automatically after selecting a style variation:


https://user-images.githubusercontent.com/1233880/229159753-7d155807-be87-4a93-a8e0-3faab9537449.mov

EDIT: The same happens if you try to change any other style such as a color or a typography.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
BogdanUngureanu: I initially thought about it too, however, this function has a different behavior for disconnected Atomic sites: instead of passing the id = 1, it returns null (since that's the expected value for tracking purposes). In the end, I opted for simplicity and duplicated the code.

I guess we could have a condition where it checks if the id returned is `1`. If it's 1, then we can assume that the blogId returned should be passed as `null`. 

I can make the change if you want, but I think it would give double meaning to the blogId variable (in the current state): sometimes is the WPCOM blog id(most of the time), but sometimes is `1`, the local id of the site (for the disconnected sites).

mmtr: Let's leave it for a follow-up then.

BogdanUngureanu:
> I found a regression, not sure if that is related with the test failure.

Yeah, it's a known issue with the approach (more like a limitation)... it was already discussed about it somewhere. The problem is that when the slots change, the slot-fill generic component does a forceUpdate on all fills, which means the local state is reset for all fills and the user is moved back to RootScreen. 

This happens in two scenarios, as you've noticed:
- When the notice is shown the first time
- When the notice is removed (e.g. when you revert back to default state, non-gated, like you did)

Sadly I haven't found a solution for this (and I doubt we can do something about it) :(

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+	useEffect( () => {
+		if ( globalStylesInUse && isGlobalStylesSidebar ) {
+			recordUpgradeNoticeSidebarShow();
+		}
+	}, [ globalStylesInUse, isGlobalStylesSidebar ] );
+
+	return (
+		<Fill name="ComplementaryArea/core/edit-site">
+			<>
```
mmtr, line 43: It'd be nice to add comment explaining why we need this apparently useless fragment 🙂

BogdanUngureanu: Good point! :)

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
```diff
@@ -0,0 +1,71 @@
...
+								__(
+									'Your changes include customized styles that will only be visible once you <a>upgrade to a Premium plan</a>.',
+									'full-site-editing'
+								),
+								{
+									a: (
+										<ExternalLink
+											href={ wpcomGlobalStyles.upgradeUrl }
+											target="_blank"
+											onClick={ () => recordUpgradeSidebarNoticeClick() }
```
mmtr: Nitpick (non-blocking): The arrow function is not needed, so we could simplify this a little bit.
```suggestion
											onClick={ recordUpgradeSidebarNoticeClick }
```

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.js
```diff
@@ -73,11 +51,7 @@ function GlobalStylesNoticeComponent() {
 						<ExternalLink
 							href={ wpcomGlobalStyles.upgradeUrl }
 							target="_blank"
-							onClick={ () =>
-								recordTracksEvent( 'calypso_global_styles_gating_notice_upgrade_click', {
-									context: 'site-editor',
-								} )
-							}
+							onClick={ () => recordUpgradePreSaveNoticeClick() }
```
mmtr: Nitpick (non-blocking): The arrow function is not needed, so we could simplify this a little bit.
```suggestion
							onClick={ recordUpgradePreSaveNoticeClick }
```

mmtr (approved):
Nice, it works great now!

I noted that on mobile viewports the notice is partially duplicated but we can address that if needed on a follow-up (I still think we can/should remove the pre-save notice).

Screenshot 2023-04-03 at 11 27 21

PR #74381, opened 2023-03-13
chore: upgrade `react-query` (v3) to `@tanstack/react-query` (v4)

flootr:
## Proposed Changes

* Upgrade the `react-query@3` package to `@tanstack/react-query@4`
* Include any adjustments necessary because of breaking changes

## Testing Instructions

[Overview of changes between v3 and v4.](https://tanstack.com/query/latest/docs/react/guides/migrating-to-react-query-4)

**Please note**, this PR touches almost all queries and mutations for which we use the `react-query` library (now `@tanstack/react-query`) as well as a few other areas in the Calypso codebase (except for a few `/apps` directories which are not imported or don't import related code from Calypso or other packages.

* Verify Calypso builds well and does not contain any new type errors or warnings
* Depending on the query or mutation, verify data is fetched as expected (compare with prod) and no errors are thrown during runtime
* Verify the local persistence plugin still works as expected

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #74515, opened 2023-03-16
`useP2GuestsQuery`: fix `useQuery` usage & prepare for v4

flootr:
This PR is split from #74381 and part of the effort to upgrade `react-query` to version 4.

## Proposed Changes

* Move check whether request is necessary to `enabled` option
* Make sure `queryFn` doesn't return an empty function which isn't serializable

## Testing Instructions

* Go through these on a P2 and a regular site:
  * Go to `/people/team/<slug>` and check your Browser Developer Tools **Network** tab
  * Make sure there's a request to `/p2/users/guests` depending on whether you're on a P2 or not

Please consider p1678958587546719-slack-C010KDAPG49 when testing on P2s with a large number of users.


tyxla (approved):
Nice one. Tests well, too. Thanks @flootr 🚀

PR #74517, opened 2023-03-16
Reader: Add label to search results when feed detected

### LINKED ISSUE #74506: Reader: Tidy up RSS UI
### Details

**Actual**
We don't do a good job of letting people know that they can add external RSS feeds to reader and if they go to add one, we don't make it easy to add. We show a black text link with no underline and no indication that it's a link.

**Expected**
We should add a line of text above this field to add RSS feeds that explicitly states that you can add external RSS, then we should turn that link into a proper button, or at least make it look like a link.

CleanShot 2023-03-06 at 12 08 04

### Checklist

- [x] Create PR to tidy up UI 

### Related

_No response_

eoigal:
The goal of this PR is to improve the UX when searching for RSS feeds on Reader.

At the moment, when you lookup a RSS feed, you get a `Follow {search URL}` link below the search text area and the search results come back with no results found. This is a confusing experience for the user.

**Search Page**
Screenshot 2023-04-06 at 22 43 45

**Manage Follow**
Screenshot 2023-04-06 at 22 47 43

### Updates

The PR works along with D107200-code which updates the search endpoints to look up feeds based on the search RSS URL entered.

This PR creates a new `SearchFollowButton` component to be used on both pages

Now when you look up a RSS feed URL, if that feed is already subscribed to by any user, the component will show a notice `Click follow to add feed to Reader` and the Follow title updates to the feed name.

**Search Page**
Screenshot 2023-04-06 at 21 45 04

**Manage Follow**
Screenshot 2023-04-06 at 21 43 09

If the feed is not subscribed to by any user yet, then the component will show the notice `Potential new feed found, click follow to add feed to Reader`

**Search Page**
Screenshot 2023-04-06 at 21 44 29

**Manage Follow**
Screenshot 2023-04-06 at 21 43 26

This PR also includes style changes.

### Testing

* Apply patch code-D107200 
* checkout PR branch
* use feeds from https://blog.feedspot.com/world_news_rss_feeds/ to test if search functionality works as expected

### DIFF
--- a/client/blocks/reader-subscription-list-item/style.scss
+++ b/client/blocks/reader-subscription-list-item/style.scss
@@ -115,7 +115,7 @@
 	display: flex;
 	flex-direction: row;
 	flex-wrap: wrap;
-	max-height: 16px * 3.2;
+	max-height: 16px * 6.2;
 	min-width: 100%;
 
 	@include breakpoint-deprecated( "<960px" ) {
--- a/client/blocks/reader-subscription-list-item/style.scss
+++ b/client/blocks/reader-subscription-list-item/style.scss
@@ -81,7 +81,7 @@
 .reader-subscription-list-item__site-excerpt,
 .reader-subscription-list-item__site-url-timestamp {
 	display: block;
-	max-height: 16px * 2.5;
+	max-height: 16px * 3;
 	overflow: hidden;
 	text-overflow: ellipsis;
 	position: relative;
--- a/client/reader/search-stream/search-follow-button.jsx
+++ b/client/reader/search-stream/search-follow-button.jsx
@@ -36,12 +36,10 @@ class SearchFollowButton extends Component {
 			return null;
 		}
 
-		let isNewFeed = true;
 		let isFollowing = false;
 		let followTitle = withoutHttp( query );
 
 		if ( feed && feed.name?.length ) {
-			isNewFeed = false;
 			isFollowing = feed.is_following;
 			followTitle = feed.name;
 		}
@@ -55,11 +53,7 @@ class SearchFollowButton extends Component {
 			<div className="search-stream__url-follow">
 				<p>
 					<Gridicon icon="info" size="16" />
-					<strong>
-						{ isNewFeed
-							? translate( 'Potential new feed found, click follow to add this feed to Reader' )
-							: translate( 'Click follow to add this feed to Reader' ) }
-					</strong>
+					<strong>{ translate( 'Click below to add this site to your Reader feed:' ) }</strong>
 				</p>
 				<FollowButton
 					followLabel={ translate( 'Follow %s', {
--- a/client/reader/search-stream/search-follow-button.jsx
+++ b/client/reader/search-stream/search-follow-button.jsx
@@ -1,0 +1,88 @@
+import { Gridicon } from '@automattic/components';
+import { localize } from 'i18n-calypso';
+import { some } from 'lodash';
+import PropTypes from 'prop-types';
+import { Component } from 'react';
+import { connect } from 'react-redux';
+import { resemblesUrl, withoutHttp, addSchemeIfMissing } from 'calypso/lib/url';
+import ReaderFollowFeedIcon from 'calypso/reader/components/icons/follow-feed-icon';
+import ReaderFollowingFeedIcon from 'calypso/reader/components/icons/following-feed-icon';
+import FollowButton from 'calypso/reader/follow-button';
+import { SEARCH_RESULTS_URL_INPUT } from 'calypso/reader/follow-sources';
+import { getReaderAliasedFollowFeedUrl } from 'calypso/state/reader/follows/selectors';
+import { commonExtensions } from 'calypso/state/reader/follows/selectors/get-reader-aliased-follow-feed-url';
+import './style.scss';
+
+class SearchFollowButton extends Component {
+	static propTypes = {
+		query: PropTypes.string,
+		feed: PropTypes.object,
+	};
+
+	render() {
+		const { query, translate, readerAliasedFollowFeedUrl, feed } = this.props;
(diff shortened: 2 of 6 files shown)


eoigal:
Asked @ollierozdarz for some design ideas at p1678976392595179-slack-C03NLNTPZ2T;

Screenshot 2023-03-17 at 8 04 01 am
Screenshot 2023-03-17 at 8 03 51 am
Screenshot 2023-03-17 at 8 03 02 am

eoigal:
I feel like the search results should either populate the sites or posts area when we look up a valid RSS feed. 

Screenshot 2023-03-21 at 10 55 32

davemart-in:
First test didn't end well. I've got D107200 applied, public-api.wordpress.com sandboxed, and this PR applied on the Calypso side of things. When I tried searching for cnn.com, this error happened:

https://user-images.githubusercontent.com/5634774/230956638-959a3205-b729-4104-a4b7-251a026192f8.mp4

Searching for https://blog.feedspot.com/world_news_rss_feeds/ resulted in the following:

CleanShot 2023-04-10 at 13 28 24@2x

davemart-in:
Made a couple of minor updates in: https://github.com/Automattic/wp-calypso/pull/74517/commits/463495afd7d3c11d1a720ac3200511b680b18fa8

- Updated helper text wording
- Updated style of follow link to make it look more like a link

Here's a preview:

CleanShot 2023-04-10 at 13 42 39

client/blocks/reader-subscription-list-item/style.scss
```diff
@@ -81,7 +81,7 @@
 .reader-subscription-list-item__site-excerpt,
 .reader-subscription-list-item__site-url-timestamp {
 	display: block;
-	max-height: 16px * 2.5;
+	max-height: 16px * 3;
```
roo2, line 84: nit: I think it's a bit strange to specify height this way, I think "48px" would be clearer

creativecoder:
Can we get the title property of the feed and display the link as "Follow NY Times", instead of "Follow https://rss.nytimes.com/services/xml/rss/nyt/NYRegion.xml"?

I think that would a clearer action link, as well as easier on the eyes!

client/reader/search-stream/search-follow-button.jsx
```diff
@@ -0,0 +1,82 @@
...
+	static propTypes = {
+		query: PropTypes.string,
+		feed: PropTypes.object,
+	};
+
+	render() {
+		const { query, translate, readerAliasedFollowFeedUrl, feed } = this.props;
+		let isPotentialFeedUrl = false;
+		if ( resemblesUrl( query ) ) {
+			const parsedUrl = new URL( query );
```
roo2: this throws an error if query has a URL without a protocol e.g. "`wordpress.com`"

roo2:
the UI doesn't handle 404s from the API correctly. It just stays in the searching state wheras I think it should go to a no results found page. e.g. 
* If I search for `https://wordpress.com/` it goes into a "no results found state" but 
* if I search for `https://wordpress.com/a` is goes into searching state and stays there.

roo2:
I commented in D107200-code. I don't think it's ideal that we only return results if the feed already exists in reader, is it possible to have it look up the feed or even create it if the feed doesn't exist in the FeedBag yet but it is a valid RSS feed endpoint?

eoigal:
I've updated code-D107200 to fix the issues you experienced while testing this PR.

eoigal:
> I commented in D107200-code. I don't think it's ideal that we only return results if the feed already exists in reader, is it possible to have it look up the feed or even create it if the feed doesn't exist in the FeedBag yet but it is a valid RSS feed endpoint?

Odd, I don't see that comment on Phab.
The whole process for looking up feeds is well established and is done in an async nature as things are. One of the requirements for a feed to be parsed is that someone subscribes to the feed first. In this case, if we look up a feed and nobody has already subscribed to it - it will not return any posts for that feed - however if they subscribe to it, that feed will be parsed async and the next time someone searches for that feed URL, it will return posts and site data.

There probably is some room for improvement with how this works but it feels out of scope for this fix so I've left it.

eoigal:
> Can we get the title property of the feed and display the link as "Follow NY Times",

If the feed already exists on our end, then we can pull out the feed name. If the feed doesn't exist yet, we handle the fetching of feeds async and knowing the feed title isn't something we have access.

We could change this to fetch the feed synchronously but I think that is outside the scope of this change.

client/reader/search-stream/search-follow-button.jsx
eoigal: Should be fixed now

davemart-in:
@eoigal when I click the link to follow, I'm now seeing the following error:

CleanShot 2023-04-11 at 10 33 28

I see the following JS error:

CleanShot 2023-04-11 at 10 34 55

Oddly, I don't see any network requests going out. Maybe due to the JS error?

The cnn.com link bug appears to be fixed.

eoigal:
@davemart-in I don't think this is a bug. The URL you entered isn't a valid RSS feed URL, so when you try to follow that URL it will fail.

This happens now without the patch applied;


https://user-images.githubusercontent.com/5560595/231202710-11761e6a-970a-4d5f-82d7-db420a394c6e.mov

davemart-in (approved):
Ah! Sorry. I misread your testing instructions. Thought that was the RSS feed. I just tested again with http://feeds.bbci.co.uk/news/world/rss.xml and it works as expected.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7920218

roo2:
> Odd, I don't see that comment on Phab.

Oops comment was unsubmitted sorry!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7920218).

PR #74534, opened 2023-03-16
Marketplace Thank You: Update the page moving each product type logic into their own hook

### LINKED ISSUE #74300: Marketplace: Update Thank You page component to split the logic for different product types
The [Marketplace Thank You](https://github.com/Automattic/wp-calypso/blob/3d1006d1780d61012d7a1f1572abd6660cbb6a7d/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx) component handles all its logic inside it, as show/hide the progress bar, retrieve .com and .org plugins and themes.

With p9Jlb4-6qr-p2, the Thank You page should support multiple types and it would be increasingly hard to keep up the development on this component the way it is. 

We should change this component's architecture in order to split the logic for different product types in multiple components.

WBerredo:
Fixes [#74300](https://github.com/Automattic/wp-calypso/issues/74300)

## Proposed Changes

* Create a hook to provide data for each product type: plugins and themes.
* Move the logic from the main Marketplace Thank You component to its hooks
* Create a hook to provide the default footer sections
* Add `useQueryThemes` to allow query for a list of ids instead of a single id

PS: Currently this page is only being used for plugins, when actually adding themes the strings will be changed to be more generalized.

## Testing Instructions
* Install free and paid plugins 
* Install more than 1 paid plugin
* The Thank You page shown should be the same as the current version in production. NO VISUAL CHANGES.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
+++ b/client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
@@ -1,48 +1,28 @@
-import { recordTracksEvent } from '@automattic/calypso-analytics';
 import { ConfettiAnimation } from '@automattic/components';
 import { ThemeProvider, Global, css } from '@emotion/react';
-import { Button } from '@wordpress/components';
 import { useTranslate } from 'i18n-calypso';
 import page from 'page';
-import { useEffect, useState, useMemo, useCallback } from 'react';
+import { useEffect, useState, useMemo } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import QueryTheme from 'calypso/components/data/query-theme';
 import { ThankYou } from 'calypso/components/thank-you';
-import { ThankYouSectionProps } from 'calypso/components/thank-you/types';
-import { useWPCOMPlugins } from 'calypso/data/marketplace/use-wpcom-plugins-query';
 import PageViewTracker from 'calypso/lib/analytics/page-view-tracker';
 import MasterbarStyled from 'calypso/my-sites/marketplace/components/masterbar-styled';
 import MarketplaceProgressBar from 'calypso/my-sites/marketplace/components/progressbar';
 import useMarketplaceAdditionalSteps from 'calypso/my-sites/marketplace/pages/marketplace-plugin-install/use-marketplace-additional-steps';
 import theme from 'calypso/my-sites/marketplace/theme';
-import { waitFor } from 'calypso/my-sites/marketplace/util';
 import { requestAdminMenu } from 'calypso/state/admin-menu/actions';
-import { fetchAutomatedTransferStatus } from 'calypso/state/automated-transfer/actions';
 import { transferStates } from 'calypso/state/automated-transfer/constants';
-import {
-	getAutomatedTransferStatus,
-	isFetchingAutomatedTransferStatus,
-} from 'calypso/state/automated-transfer/selectors';
+import { getAutomatedTransferStatus } from 'calypso/state/automated-transfer/selectors';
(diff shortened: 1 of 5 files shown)


client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -58,41 +38,14 @@ const MarketplaceThankYou = ( {
...
 
-	// retrieve WPCom plugin data
-	const wpComPluginsDataResults = useWPCOMPlugins( pluginSlugs );
-	const wpComPluginsData: Array< any > = wpComPluginsDataResults.map(
-		( wpComPluginData ) => wpComPluginData.data
-	);
-	const softwareSlugs = wpComPluginsData.map( ( wpComPluginData, i ) =>
-		wpComPluginData ? wpComPluginData.software_slug || wpComPluginData.org_slug : pluginSlugs[ i ]
-	);
+	const allSlugs = useMemo( () => [ ...pluginSlugs, ...themeSlugs ], [ pluginSlugs, themeSlugs ] );
```
epeicher, line 40: Nit: Concatenating two arrays that will have, in a worst-case scenario, around 10 elements is not expensive, so I would not use `useMemo` here for better readability, although it is not doing any harm (as [docs mention](https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere)), so I will leave it up to you, not blocking at all.

epeicher (approved):
This is so much better, thank you very much for the refactoring. Great work here! 🎉 

I have tested it and works as advertised. LGTM! 🚢

gcsecsey (approved):
Thanks for tackling this @WBerredo, awesome work! 🙌 

Changes LGTM and the thank you page is still working as before! 👍 🚢

client/my-sites/checkout/checkout-thank-you/marketplace/marketplace-thank-you.tsx
```diff
@@ -37,7 +37,7 @@ const MarketplaceThankYou = ( {
 	const currentUser = useSelector( getCurrentUser );
 	const isRequestingPlugins = useSelector( ( state ) => isRequesting( state, siteId ) );
 
-	const allSlugs = useMemo( () => [ ...pluginSlugs, ...themeSlugs ], [ pluginSlugs, themeSlugs ] );
+	const allSlugs = [ ...pluginSlugs, ...themeSlugs ];
```
epeicher: Hi @WBerredo , sorry in advance for what I am going to suggest 🙏 . [My previous comment](https://github.com/Automattic/wp-calypso/pull/74534#discussion_r1139898235) suggested removing the `useMemo` to improve readability but I have realized that `allSlugs` is the dependency of the other custom hooks and it is better to avoid executing inner Effects when not required (i.e. `allSlugs` being the same). 
I have tested and confirmed that `pluginSlugs` and `themeSlugs` do not change between renders, so using `useMemo` here would return the same `allSlugs` between renders while concatenating the arrays without `useMemo` will return a new array.
In summary, _memoizing_ `allSlugs` is a good optimization here
```
const allSlugs = useMemo( () => [ ...pluginSlugs, ...themeSlugs ], [ pluginSlugs, themeSlugs ] );
```
Not a blocking suggestion, but I wanted to identify my previous misleading comment 😅 I think your original implementation is better.

WBerredo: Yeah, exactly, I decided to remove it because in the future we plan to remove most of those `useEffects`. But as we still have it, I'm going to readd the memoization.

gcsecsey (approved):
I re-tested the changes with memoization and they still work as expected. 🚢

PR #74537, opened 2023-03-16
Launchpad: Avoid showing Launchpad if tasks are complete

edanzer:
### Proposed Changes

We periodically have cases where a user has gone through all launchpad steps and/or a site has already been launched, but `launchpad_screen` is still set to to 'full'. We're not exactly sure how such cases arise - theoretically, once tasks are complete, launchpad_screen should be off. 

Regardless of how the situation arises, because we only update `launchpad_screen` during actual site launch - on the `wpcom_site_launched` action - there is no way for a normal end user to turn off launchpad once their site is in this condition. That means they'll keep getting redirected from My Home to Launchpad indefinitely. 

This adds an check to Launchpad and too My Home. We use site_intent to determine the last launchpad task for each site. In most cases, that will be `site_launched`, but it may different for others (ie, for newsletter, it is `first_post_published`). We then confirm if that last task has been completed. If so, we do not redirect from My Home to Launchpad, and if the user accesses Launchpad directly, we turn launchpad_screen off and redirect to My Home. 

### Testing

**Review Time: Medium**
**Testing Time: Medium**

To start, check out this branch and run yarn and yarn start if needed. 

**Test Free Flow** (could also test write/build/link in bio)
1. Start a new free site at http://calypso.localhost:3000/setup/free/intro, and go through to Launchpad.
2. First, test that Launchpad works as expected under normal conditions. 
   - TEST: Confirm Launchpad screen loads as expected when you first arrive. 
   - TEST: Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad. 
   - TEST: Click the Launch button to launch the site, go to My Home, and confirm you are not redirected to Launchpad.
3. Second, test the specific situation we're fixing. 
   - Go to [YOURSITEURL]/wp-admin/options.php and update the launchpad_screen option to full. This duplicates the bug situation: the site is fully launched with all required launchpad tasks complete, but launchpad_screen is still `full`. 
   - TEST: Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are **NOT** redirected to My Home even though launchpad_screen === full. Prior to this PR, you would be. 
   - TEST: Go to launchpad at http://calypso.localhost:3000/setup/free/launchpad?siteSlug=YOURSITESLUG, and confirm you **ARE** redirected to launchpad even though launchpad_screen === full. 
   - TEST: Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'. 
   
**Test Newsletter Flow** (this is different because sites are created in 'launched state and last task if publishing a post)
1. Start a newsletter site at http://calypso.localhost:3000/setup/newsletter/intro, and go through to Launchpad.
2. First, test that Launchpad works as expected under normal conditions. 
   - TEST: Confirm Launchpad screen loads as expected when you first arrive. 
   - TEST: Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad. 
   - TEST: Click the Publish Post button, publish a post, go to My Home, and confirm you are not redirected to Launchpad.
3. Second, test the specific situation we're fixing. 
   - Go to [YOURSITEURL]/wp-admin/options.php and update the launchpad_screen option to full. This duplicates the bug situation: the site is fully launched with all required launchpad tasks complete, but launchpad_screen is still `full`. 
   - TEST: Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are **NOT** redirected to launchpad even though launchpad_screen === full. Prior to this PR, you would be. 
   - TEST: Go to launchpad at http://calypso.localhost:3000/setup/newsletter/launchpad?siteSlug=YOURSITESLUG, and confirm you **ARE** redirected to My Home even though launchpad_screen === full. 
   - TEST: Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
@@ -2,7 +2,6 @@ import { isNewsletterFlow, StepContainer } from '@automattic/onboarding';
 import { useSelect } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
 import { useTranslate } from 'i18n-calypso';
-import { useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import DocumentHead from 'calypso/components/data/document-head';
 import FormattedHeader from 'calypso/components/formatted-header';
@@ -46,21 +45,6 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 		[]
 	);
 
-	const areLaunchpadTasksCompleted = useCallback(
-		( flow: string | null ) => {
-			if ( isNewsletterFlow( flow ) ) {
-				return Boolean( checklist_statuses?.first_post_published );
-			}
-			return isSiteLaunched || Boolean( checklist_statuses?.site_launched );
-		},
-		[ checklist_statuses?.first_post_published, checklist_statuses?.site_launched, isSiteLaunched ]
-	);
-
-	const redirectToSiteHome = useCallback( ( siteSlug: string | null, flow: string | null ) => {
-		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
-		window.location.replace( `/home/${ siteSlug }` );
-	}, [] );
-
 	if (
 		! isLoggedIn ||
 		launchpadScreenOption === 'off' ||
@@ -72,6 +56,18 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 
 	if ( ! siteSlug || fetchingSiteError?.error ) {
 		window.location.replace( '/home' );
+	}
+
+	function areLaunchpadTasksCompleted( flow: string | null ) {
+		if ( isNewsletterFlow( flow ) ) {
+			return Boolean( checklist_statuses?.first_post_published );
+		}
+		return isSiteLaunched || Boolean( checklist_statuses?.site_launched );
+	}
+
+	function redirectToSiteHome( siteSlug: string | null, flow: string | null ) {
+		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
+		window.location.replace( `/home/${ siteSlug }` );
 	}
 
 	useEffect( () => {
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
@@ -2,6 +2,7 @@ import { isNewsletterFlow, StepContainer } from '@automattic/onboarding';
 import { useSelect } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
 import { useTranslate } from 'i18n-calypso';
+import { useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import DocumentHead from 'calypso/components/data/document-head';
 import FormattedHeader from 'calypso/components/formatted-header';
@@ -46,24 +47,27 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 		[]
 	);
 
+	const redirectToSiteHome = useCallback( ( siteSlug: string | null, flow: string | null ) => {
+		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
+		window.location.replace( `/home/${ siteSlug }` );
+	}, [] );
+
+	const areLaunchpadTasksCompleted = useCallback(
+		( flow: string | null ) => {
+			if ( isNewsletterFlow( flow ) ) {
+				return checklist_statuses?.first_post_published;
+			}
+			return isSiteLaunched || checklist_statuses?.site_launched;
+		},
+		[ checklist_statuses?.first_post_published, checklist_statuses?.site_launched, isSiteLaunched ]
+	);
+
 	if ( ! isLoggedIn ) {
-		redirectToSiteHome( siteSlug );
+		redirectToSiteHome( siteSlug, flow );
 	}
 
 	if ( ! siteSlug || fetchingSiteError?.error ) {
 		window.location.replace( '/home' );
-	}
-
-	function redirectToSiteHome( siteSlug: string | null ): void {
-		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
-		window.location.replace( `/home/${ siteSlug }` );
-	}
-
-	function areLaunchpadTasksCompleted( flow: string | null ): boolean {
-		if ( isNewsletterFlow( flow ) ) {
-			return checklist_statuses?.first_post_published;
-		}
-		return isSiteLaunched || checklist_statuses?.site_launched;
 	}
 
 	useEffect( () => {
@@ -85,12 +89,12 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 				launchpadScreenOption === 'off' ||
 				( launchpadScreenOption === false && 'videopress' !== flow )
 			) {
-				redirectToSiteHome( siteSlug );
+				redirectToSiteHome( siteSlug, flow );
 			}
 
 			if ( areLaunchpadTasksCompleted( flow ) ) {
 				saveSiteSettings( site.ID, { launchpad_screen: 'off' } );
-				redirectToSiteHome( siteSlug );
+				redirectToSiteHome( siteSlug, flow );
 			}
 		}
 		recordTracksEvent( 'calypso_launchpad_loaded', { flow: flow } );
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
@@ -17,6 +17,7 @@ import { successNotice } from 'calypso/state/notices/actions';
 import { useQuery } from '../../../../hooks/use-query';
 import StepContent from './step-content';
 import { areLaunchpadTasksCompleted } from './task-helper';
+import { launchpadFlowTasks } from './tasks';
 import type { Step } from '../../types';
 import type { SiteSelect } from '@automattic/data-stores';
 
@@ -49,7 +50,12 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 	if (
 		! isLoggedIn ||
 		launchpadScreenOption === 'off' ||
-		areLaunchpadTasksCompleted( site_intent, checklist_statuses, isSiteLaunched ) ||
+		areLaunchpadTasksCompleted(
+			site_intent,
+			launchpadFlowTasks,
+			checklist_statuses,
+			isSiteLaunched
+		) ||
 		( launchpadScreenOption === false && 'videopress' !== flow )
 	) {
 		redirectToSiteHome( siteSlug, flow );
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
@@ -1,4 +1,4 @@
-import { StepContainer } from '@automattic/onboarding';
+import { isNewsletterFlow, StepContainer } from '@automattic/onboarding';
 import { useSelect } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
 import { useTranslate } from 'i18n-calypso';
@@ -14,6 +14,7 @@ import { SITE_STORE } from 'calypso/landing/stepper/stores';
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { isUserLoggedIn } from 'calypso/state/current-user/selectors';
 import { successNotice } from 'calypso/state/notices/actions';
+import { saveSiteSettings } from 'calypso/state/site-settings/actions';
 import { useQuery } from '../../../../hooks/use-query';
 import StepContent from './step-content';
 import type { Step } from '../../types';
@@ -33,8 +34,9 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 	const verifiedParam = useQuery().get( 'verified' );
 	const site = useSite();
 	const {
-		data: { launchpad_screen: launchpadScreenOption },
+		data: { launchpad_screen: launchpadScreenOption, checklist_statuses },
 	} = useLaunchpad( siteSlug );
+	const isSiteLaunched = site?.launch_status === 'launched' || false;
 	const recordSignupComplete = useRecordSignupComplete( flow );
 	const dispatch = useDispatch();
 	const isLoggedIn = useSelector( isUserLoggedIn );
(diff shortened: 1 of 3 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -42,13 +46,25 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
+		window.location.replace( `/home/${ siteSlug }` );
+	}
+
+	function areLaunchpadTasksCompleted( flow: string | null ): boolean {
+		if ( isNewsletterFlow( flow ) ) {
+			return isFirstPostPublished;
+		}
+		return isSiteLaunched;
+	}
+
```
edanzer, line 67: The real change here is the addition and use of this method. Note that for non-newsletter sites, I'm not actually checking if *launchpad tasks are complete*, but rather if site is launched. This captures a wider array of situations - it's theoretically possible to have a site end up in 'launched' state without all launchpad tasks being complete. But in any case, if a site launched (other than newsletter), then the current full screen launchpad should not be shown. 

We could also check actual task status by checking against `site?.options?.launchpad_checklist_tasks_statuses?.site_launched`.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -61,21 +77,23 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
-		if ( launchpadScreenOption !== undefined ) {
-			// The screen option returns false for sites that have never set the option
+		// Site is null for new users/sites during onboarding and for existing sites before data loads.
+		// In either case, we do not want to redirect from Launchpad until we can check site data.
+		if ( site ) {
 			if (
-				( 'videopress' !== flow && launchpadScreenOption === false ) ||
-				launchpadScreenOption === 'off'
+				launchpadScreenOption === 'off' ||
+				( launchpadScreenOption === false && 'videopress' !== flow )
```
edanzer, line 85: Does anyone know the original reason for the inclusion of this videopress exception? I'm wondering if we can remove that now as part of this PR? Technically, launchpad_screen should not be false in any case.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -44,8 +47,23 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 		[]
 	);
 
-	if ( ! isLoggedIn ) {
+	const redirectToSiteHome = useCallback( ( siteSlug: string | null, flow: string | null ) => {
+		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
 		window.location.replace( `/home/${ siteSlug }` );
+	}, [] );
+
```
edanzer, line 54: Adding redirectToSiteHome() to reduce duplication and ensure consistency. We hadn't been logging the calypso_launchpad_redirect_to_home event when redirecting on line 47 for the isLoggedIn check.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -44,8 +47,23 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
+	const areLaunchpadTasksCompleted = useCallback(
+		( flow: string | null ) => {
+			if ( isNewsletterFlow( flow ) ) {
+				return checklist_statuses?.first_post_published;
+			}
+			return isSiteLaunched || checklist_statuses?.site_launched;
+		},
+		[ checklist_statuses?.first_post_published, checklist_statuses?.site_launched, isSiteLaunched ]
+	);
+
```
edanzer, line 64: The addition and use of areLaunchpadTasksCompleted() is the main change in this PR. For newsletter sites, if first_post_published is done (the last launchpad task for that flow), then Launchpad should already be complete/already. Likewise, for all other flows, if site_launched is done, or the site is actually already launched, then launchpad should be complete/off already.

So we check those here, and below if areLaunchpadTasksCompleted(), we update launchpad_screen to 'off' and redirect to home.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -63,23 +85,6 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
 		}
 	}, [ verifiedParam, translate, dispatch ] );
 
-	useEffect( () => {
```
edanzer, line 66: Best I can tell, none of the logic here needs to be in a useEffect, so I've removed it and shifted it above. In addition to removing the unneeded useEffect, this also has the effect of triggering the code faster on the frontend. There's still a flicker of the launchpad before redirecting, but it's much faster without the useEffect.

client/my-sites/customer-home/controller.jsx
```diff
@@ -37,9 +41,18 @@ export async function maybeRedirect( context, next ) {
...
+	const site = getSelectedSite( state );
+	const isSiteLaunched = site?.launch_status === 'launched' || false;
+	const { launchpad_screen, checklist_statuses, site_intent } = await fetchLaunchpad( slug );
+
+	function areLaunchpadTasksCompleted() {
+		if ( 'newsletter' === site_intent ) {
+			return Boolean( checklist_statuses?.first_post_published );
+		}
+		return isSiteLaunched || Boolean( checklist_statuses?.site_launched );
+	}
```
edanzer: Worth noting the two areLaunchpadTasksComplete() methods used in the two place here are almost, but not quite, identical. May be worth trying to consolidate those to a single method that is imported and used in both places.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -44,14 +45,31 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
 	);
 
-	if ( ! isLoggedIn ) {
-		window.location.replace( `/home/${ siteSlug }` );
+	if (
+		! isLoggedIn ||
+		launchpadScreenOption === 'off' ||
+		areLaunchpadTasksCompleted( flow ) ||
+		( launchpadScreenOption === false && 'videopress' !== flow )
+	) {
```
edanzer, line 53: Does anyone know why that videopress exception was originally added? I wonder if we could get rid of that now. We also do not set launchpad_screen to false any more. Would be nice to clean that line out.

client/my-sites/customer-home/controller.jsx
Addison-Stavlo: 1. Why can't they be identical?  This version seems to check the `site_intent` while the other seems to check the flow. We should ensure the flow matches the site intent - the site intent is the main source of truth. The flow is just info from the URL iirc? Right now we can load any flow via the URL, but perhaps those should always be redirecting to the launchpad checklist corresponding to actual `site_intent` (yes, its a bit of an edge case).

2. We are still doing a blanket generalization that all intents other than newsletter mark completion via site launch. this might be ok for now, but will cause bugs and need to be updated as soon as anyone adds any other checklist that doesn't - we should probably be more specific about checking the intent/flow before raising conditional logic based on launch status.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -44,14 +45,31 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
+		redirectToSiteHome( siteSlug, flow );
```
Addison-Stavlo, line 68: We should probably turn `launchpadScreen` to `off` here if `areLaunchpadTasksCompleted` is true. However, we need to ensure the flow/checklist actually corresponds to their given site intent before doing so.

More thoughts in p1679057706850659/1678732852.469439-slack-CHN6J22MP

edanzer: @Addison-Stavlo - Agreed. Would be OK if I handled turning off Launchpad via separate PR?

edanzer: I've updated this so that launchpad_screen to off if areLaunchpadTasksCompleted() returns true.

There is a bit of a question about how to handle matching flow/site_intent/checklist. We fetch the list of tasks dynamically when launchpad is loaded based largely on the flow, which is provided / parsed from the url by stepper. 

That means it is indeed possible to show the wrong set of tasks on launchpad for a given site. I can start a new free site, and when I get to launchpad, update the url from 'free' to 'newsletter', and I will get newsletter tasks even though the site_intent for the site is 'free'. 

In short, if flow does not match site_intent, we'll get the wrong list of tasks. 

That said, I don't think we need to match these to do what's needed for this PR. Ultimately, the site_intent is the source of truth about which tasks should show. If we know site_intent === free, then we already know the last task is site_launched, and if the site is launched, then we no longer need to show Launchpad. We're able to determine all that without checking that site_intent matches flow. 

All that said, what you're pointing to here does seems like a broader issue with launchpad that we should fix. If we are displaying launchpad, we probably should not allow the wrong tasks to show. That would mean adding a check that flow is correct based site_intent, and if not, redirecting to the correct stepper url for launchpad. 

Given that it's a good change but not technically needed for this PR, would it be OK if I handle this via a follow up?

client/my-sites/customer-home/controller.jsx
edanzer: I've consolidated to a single method for these - so they're now identical. I've also updated the logic for checking tasks to be more robust, but I'm still not checking site_intent vs flow. I've added a separate comment/thought on that [here](https://github.com/Automattic/wp-calypso/pull/74537#discussion_r1147893482).

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
```diff
@@ -33,25 +35,48 @@ const Launchpad: Step = ( { navigation, flow }: LaunchpadProps ) => {
...
 
 	if ( ! siteSlug || fetchingSiteError?.error ) {
 		window.location.replace( '/home' );
 	}
 
+	function redirectToSiteHome( siteSlug: string | null, flow: string | null ) {
+		recordTracksEvent( 'calypso_launchpad_redirect_to_home', { flow: flow } );
+		window.location.replace( `/home/${ siteSlug }` );
+	}
+
```
edanzer, line 79: I've added redirectToSiteHome() to reduce duplication and ensure consistency. We hadn't been logging the calypso_launchpad_redirect_to_home event when redirecting on line 47 for the isLoggedIn check.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx
Addison-Stavlo: > That said, I don't think we need to match these to do what's needed for this PR. Ultimately, the site_intent is the source of truth about which tasks should show. 

Makes sense to me!

sharpirate:
### Testing
**Test Free Flow** (could also test write/build/link in bio)
1. Start a new free site at http://calypso.localhost:3000/setup/free/intro, and go through to Launchpad.
2. First, test that Launchpad works as expected under normal conditions. 
   ✔️ Confirm Launchpad screen loads as expected when you first arrive. 
   ✔️ Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad. 
   ✔️ Click the Launch button to launch the site, go to My Home, and confirm you are not redirected to Launchpad.
3. Second, test the specific situation we're fixing. 
   - Go to [YOURSITEURL]/wp-admin/options.php and update the launchpad_screen option to full. This duplicates the bug situation: the site is fully launched with all required launchpad tasks complete, but launchpad_screen is still `full`. 
   ✔️ Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are **NOT** redirected to My Home even though launchpad_screen === full. Prior to this PR, you would be. 
   ✔️ Go to launchpad at http://calypso.localhost:3000/setup/free/launchpad?siteSlug=YOURSITESLUG, and confirm you **ARE** redirected to launchpad even though launchpad_screen === full. 
   ✔️ Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'. 
   
**Test Newsletter Flow** (this is different because sites are created in 'launched state and last task if publishing a post)
1. Start a newsletter site at http://calypso.localhost:3000/setup/newsletter/intro, and go through to Launchpad.
2. First, test that Launchpad works as expected under normal conditions. 
   ✔️ Confirm Launchpad screen loads as expected when you first arrive. 
   ✔️ Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad. 
   ✔️ Click the Publish Post button, publish a post, go to My Home, and confirm you are not redirected to Launchpad.
3. Second, test the specific situation we're fixing. 
   - Go to [YOURSITEURL]/wp-admin/options.php and update the launchpad_screen option to full. This duplicates the bug situation: the site is fully launched with all required launchpad tasks complete, but launchpad_screen is still `full`. 
   ✔️ Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are **NOT** redirected to launchpad even though launchpad_screen === full. Prior to this PR, you would be. 
   ✔️ Go to launchpad at http://calypso.localhost:3000/setup/newsletter/launchpad?siteSlug=YOURSITESLUG, and confirm you **ARE** redirected to My Home even though launchpad_screen === full. 
   ✔️ Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'.

sharpirate (approved):

agrullon95:
## Tests:

Free site:

- [x] TEST: Confirm Launchpad screen loads as expected when you first arrive.
- [x] TEST: Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad.
- [x] TEST: Click the Launch button to launch the site, go to My Home, and confirm you are not redirected to Launchpad.
- [x] TEST: Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are NOT redirected to My Home even though launchpad_screen === full. Prior to this PR, you would be.
- [x] TEST: Go to launchpad at http://calypso.localhost:3000/setup/free/launchpad?siteSlug=YOURSITESLUG, and confirm you ARE redirected to launchpad even though launchpad_screen === full.
- [x] TEST: Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'.

Newsletter site:

- [x] TEST: Confirm Launchpad screen loads as expected when you first arrive.
- [x] TEST: Click 'Skip to Dashboard', navigate to My Home, and confirm you are redirected back to Launchpad.
- [x] TEST: Click the Publish Post button, publish a post, go to My Home, and confirm you are not redirected to Launchpad.
- [x] TEST: Go to My Home (be sure you are on localhost, not wordpress.com) and confirm you are NOT redirected to launchpad even though launchpad_screen === full. Prior to this PR, you would be.
- [x] TEST: Go to launchpad at http://calypso.localhost:3000/setup/newsletter/launchpad?siteSlug=YOURSITESLUG, and confirm you ARE redirected to My Home even though launchpad_screen === full.
- [x] TEST: Finally go back to [YOURSITEURL]/wp-admin/options.php and confirm launchpad_screen is now 'off'.

agrullon95 (approved):

PR #74538, opened 2023-03-16
Checkout: Use string interpolation for tax names in VAT form

sirbrillig:
## Proposed Changes

The checkout billing info/contact info form sometimes collects VAT information depending on the country chosen. Because the name of VAT changes based on country, it would be better to use a localized term. In https://github.com/Automattic/wp-calypso/pull/73686 we modified the labels for the VAT form use use the more generic term "Business Tax", but this is still a little unclear.

In this PR we switch each of these labels to replace "Business Tax" with the selected country's term for VAT (eg: "GST", "CT", etc.).

**NOTE:** this adds a good number of new strings. Make sure to translate these before merging!

**NOTE 2:** this does not alter the labels of other pages that interact with VAT, like `/me/purchases/vat-details`, because on those pages it's not as easy to tell what country we should use to determine the label. Updating those (if possible) will be the subject of a different PR.

Before:

Screenshot 2023-03-16 at 4 05 16 PM

After:

Screenshot 2023-03-16 at 4 04 27 PM

Screenshot 2023-03-16 at 4 07 49 PM

## Testing Instructions

- Click to edit the billing details step if it is not already active.
- Select "United Kingdom" as the country.
- Check the VAT checkbox.
- Verify that the form mentions "VAT".
- Change the country to "Switzerland".
- Check the GST checkbox.
- Verify that the form mentions "GST".

### DIFF
--- a/client/me/purchases/billing-history/vat-vendor-details.tsx
+++ b/client/me/purchases/billing-history/vat-vendor-details.tsx
@@ -24,9 +24,19 @@ interface VatVendorInfo {
 	vatId: string;
 }
 
-function getVatVendorInfo(
+export function getVatVendorInfo(
+	/**
+	 * Two-letter country code.
+	 */
 	country: string,
+
+	/**
+	 * A string version of date that can be read by Date.parse.
+	 *
+	 * Can be 'now'.
+	 */
 	date: string,
+
 	translate: ReturnType< typeof useTranslate >
 ): VatVendorInfo | undefined {
 	const automatticVatAddress = [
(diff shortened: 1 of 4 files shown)


michaeldcain (approved):
This works well. Let's `String Freeze` it.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7871432

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7871432).

PR #55292, opened 2021-08-09
Tooling/Building CalypsO: Bump `sharp` dependency version to 0.28 to allow building of Calypso on M1 macs

bisko:
The 0.25.3 version of `sharp` seems to require `libvips` that has no build for M1 macs. This prevents Calypso from completing the `yarn install` step and subsequently is unable to start because of that.

Seems that version 0.28 of `sharp` solves this - https://github.com/lovell/sharp/issues/2460

This only affects native M1 flows, where the terminal/yarn/homebrew is not being ran through Rosetta.

#### Changes proposed in this Pull Request

* Bump `sharp` dependency version to 0.28

#### Testing instructions

* Apply PR
* Build Calypso
* Make sure it builds successfully

PR #74539, opened 2023-03-16
Remove unused gutenboarding script

noahtallen:
## Proposed Changes
I think this script has been unused for a while. It was added for some aspects of gutenboarding, which is no longer used (#74475). This lets us remove a dependency on `sharp`, which has been problematic in the past (#55292). And I noticed it causing issues with the latest typescript update, so it'd be nice to remove it.

## Testing Instructions
None; CI should pass. Also verify `sharp` isn't imported elsewhere.


bin/install-gutenboarding-design-thumbnails-dependencies.js
```diff
@@ -1,48 +0,0 @@
...
-	process.exit( installResult.status );
-}
-
-/*
- * Add screenshot dependencies
- * I didn't want to add these to dependencies just for this script since these are kinda heavy ones)
- * https://www.npmjs.com/package/capture-website
- * https://www.npmjs.com/package/sharp
- */
-const installDependencies = spawnSync( 'yarn', [ 'add', '-W', 'capture-website', 'sharp' ], {
```
tyxla, line 38: While we're here, I think we can remove the `capture-website` dependency, too. This seems to be its sole usage.

tyxla (approved):
Looking great! 🚀 

I'd only suggest removing another unused package before shipping.

test/e2e/docs/troubleshooting.md
```diff
@@ -40,24 +40,3 @@ yarn install
 ```
 
 Once complete, running `git commit` should no longer trigger the git pre-commit hook error.
-
-## Package 'lcms2', required by 'vips', not found
```
noahtallen, line 44: This won't be an issue any more :D

bin/install-gutenboarding-design-thumbnails-dependencies.js
noahtallen: done!

PR #74540, opened 2023-03-16
Update to typescript 5.1

noahtallen:
## Proposed Changes
This PR starts the update to Typescript 5, which was just released. I mostly wanted to see how much extra work this would require (if any). So far, a couple issues have come up, which we can track here.

See also: https://github.com/WordPress/gutenberg/pull/52621

## Testing Instructions
- CI should pass (no type errors)
- Smoke test all of calypso

## To Do:
- [x] Fix an issue with generic inference related to `treeSelect` and `createSelector`. This was hard to track down, but it might be a TS bug: https://github.com/microsoft/TypeScript/issues/55192. This was fixed by duplicating the generics which had multiple candidates, extending from the generic we want TS to pick.
- [x] `cloneElement` and `isValidElement` have some new issues where types won't narrow in the same way as before if no generics are passed to props.
- [x] The sharp dependency is causing issues. This can be removed here. (#74539)
- [x] Some equality checks in calypso-e2e/src/rest-api-client.ts are incorrect (#74741)
- [x] PopperJS build types have issues. Fixed by updating to a newer minor version of Popper. (Side note, this package was renamed to something else.)
- [x] There is type issue in the [`ButtonBar` component](https://github.com/Automattic/wp-calypso/blob/94fe59f75d919c439f182764eed7d86f1f63da48/packages/components/src/dialog/button-bar.tsx). Typescript doesn't think that `ReactElement` includes the `className` property, along with several other properties. This was caused by an issue with `isValidElement` not narrowing types correctly: https://github.com/microsoft/TypeScript/issues/53178#issuecomment-1659301034
- [x] There are a handful of `useContext` calls which rely on a `GlobalStylesContext` import from `@wordpress/edit-site`. It now thinks the type is `unknown`, but previously it didn't complain about any issues here. ([Example issue](https://github.com/Automattic/wp-calypso/blob/13ac561fce9096a7e08b912f16708ea10b1116f2/packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts#L7)) These global styles issues were fixed by that team.
- [x] `@wordpress/data`'s build types cause errors in ETK. Appears to be resolved when updating to newer wordpress data versions. (Potentially blocking the PR on React 18)
- [x] `@wordpress/data-control`'s types cause errors in ETK. This might require a DT update. ([See here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/wordpress__data-controls))


tyxla (commented):
Thanks for starting work on this 👍 

While I haven't thoroughly read through the changelog, I'm assuming that the new version could have a bunch of fixes which make tests more precise.

One example I noticed is the group of errors related to unsupported props. I think that error actually makes sense and here's an example how I made it work locally:

```
diff --git a/packages/components/src/dialog/button-bar.tsx b/packages/components/src/dialog/button-bar.tsx
index 22b852296c..777cf2f166 100644
--- a/packages/components/src/dialog/button-bar.tsx
+++ b/packages/components/src/dialog/button-bar.tsx
@@ -1,7 +1,7 @@
 import classnames from 'classnames';
 import { isValidElement, cloneElement } from 'react';
 import Button from '../button';
-import type { ReactElement, ReactNode, FunctionComponent } from 'react';
+import type { ReactNode, FunctionComponent } from 'react';

 export type BaseButton = {
        action: string;
@@ -15,10 +15,8 @@ export type BaseButton = {
        target?: string;
 };

-export type Button = ReactElement | BaseButton;
-
 type Props = {
-       buttons?: Button[];
+       buttons?: BaseButton[];
        baseClassName: string;
        onButtonClick: ( button: BaseButton ) => void;
 };
diff --git a/packages/components/src/dialog/index.tsx b/packages/components/src/dialog/index.tsx
index 819dc4a221..a62ae3f88b 100644
--- a/packages/components/src/dialog/index.tsx
+++ b/packages/components/src/dialog/index.tsx
@@ -3,7 +3,7 @@ import { useCallback } from 'react';
 import Modal from 'react-modal';
 import Gridicon from '../gridicon';
 import ButtonBar from './button-bar';
-import type { Button, BaseButton } from './button-bar';
+import type { BaseButton } from './button-bar';
 import type { PropsWithChildren } from 'react';

 import './style.scss';
@@ -12,7 +12,7 @@ type Props = {
        additionalClassNames?: Parameters< typeof classnames >[ 0 ];
        additionalOverlayClassNames?: Parameters< typeof classnames >[ 0 ];
        baseClassName?: string;
-       buttons?: Button[];
+       buttons?: BaseButton[];
        className?: string;
        isBackdropVisible?: boolean;
        isFullScreen?: boolean;
```

That being said, I wonder if there was a good reason it was typed that way before 🤔 

Those new "Cannot find module '@automattic/components' or its corresponding type declarations." errors are also interesting. Any idea what changed with regards to module resolution?

noahtallen:
> Those new "Cannot find module '@automattic/components' or its corresponding type declarations." errors are also interesting. Any idea what changed with regards to module resolution?

I think this happens when there's a type issue compiling an internal dependency. Since other packages depend on it, they now fail since the dependency on components fails. 

Thanks for looking into that; I'll try to integrate it!

noahtallen:
I like your fix for the buttons! I think the reason we allowed the `Button` type is because this logic exists:

https://github.com/Automattic/wp-calypso/blob/b96434e6b089f14db0ad0faaeec8d07e62d4e3ad/packages/components/src/dialog/button-bar.tsx#L36-L38

But it looks like just using `BaseButton` still allows normal buttons to be passed in

packages/components/src/list-tile/index.tsx
```diff
@@ -40,7 +40,7 @@ export const ListTile = ( {
 		typeof trailing === 'string' ? (
 			<div className="list-tile__trailing">{ trailing }</div>
 		) : isValidElement( trailing ) ? (
-			cloneElement( trailing, {
+			cloneElement( trailing as React.ReactElement, {
```
noahtallen: From what I can tell, this is still correct -- leading can be either `ReactNode` or `ReactElement`. The `isValidElement` check means we know it's an element here, which means we can access its props directly below.

Though, this seems to indicate that the type restriction from `isValidElement` isn't working out of the box any more 🤔

packages/global-styles/src/components/global-styles-provider/index.tsx
```diff
@@ -17,7 +17,11 @@ const cleanEmptyObject = ( object: any ) => {
 	return isEmpty( cleanedNestedObjects ) ? undefined : cleanedNestedObjects;
 };
 
-const useGlobalStylesUserConfig = () => {
+const useGlobalStylesUserConfig = (): [
+	boolean,
+	GlobalStylesObject,
+	( config: object ) => unknown
+] => {
```
noahtallen, line 24: Typescript doesn't infer tuples very well unless you give it some help, so this is mostly a drive-by improvement for types in the file.

packages/global-styles/src/hooks/use-get-global-styles-base-config.ts
```diff
@@ -3,7 +3,7 @@ import wpcomRequest from 'wpcom-proxy-request';
 import type { GlobalStylesObject } from '../types';
 
 const useGetGlobalStylesBaseConfig = ( siteId: number | string, stylesheet: string ) => {
-	return useQuery< any, unknown, unknown >(
+	return useQuery(
```
noahtallen: This was preventing type inference from working -- the type parameter to `wpcomRequest` below was actually enough to get the return type inferred correctly. (But that didn't work with any/unknown type params to useQuery)

packages/calypso-e2e/src/rest-api-client.ts
noahtallen, line 1: Changes in this file are temporary to unblock other work. Will remove them when #74741 is merged

packages/global-styles/src/global.d.ts
```diff
@@ -16,8 +16,14 @@ declare module '@wordpress/components' {
 	export const __experimentalVStack: React.ComponentType< Props >;
 }
 
-declare module '@wordpress/edit-site/build-module/components/global-styles/context';
-declare module '@wordpress/edit-site/build-module/components/global-styles/global-styles-provider';
+// Note: edit-site doesn't have anything on DefinitelyTyped yet :(
+declare module '@wordpress/edit-site/build-module/components/global-styles/context' {
+	export const GlobalStylesContext: React.Context< T >;
```
noahtallen: I'd like to provide more specific types for this ideally, but I wasn't sure how to import anything from the `./types` file, and declaring them in this file meant I couldn't export them... but this fixes the issue 🤷

packages/block-renderer/src/components/block-renderer-container.tsx
```diff
@@ -110,7 +110,7 @@ const ScaledBlockRendererContainer = ( {
 				transform: `scale(${ scale })`,
 				height: scaledHeight,
 				maxHeight:
-					maxHeight !== 'none' && ( contentHeight as number ) > maxHeight
+					maxHeight !== 'none' && ( contentHeight as number ) > ( maxHeight as number )
```
noahtallen, line 113: I commented in another PR about this. Currently, the component only accepts "none" and maxHeight, but types maxHeight as string|number.

But beyond that, `maxHeight` itself can accept several things -- like px values, and this logic wouldn't work with that.

packages/block-renderer/src/components/block-renderer-container.tsx
```diff
@@ -19,7 +19,7 @@ interface BlockRendererContainerProps {
 	inlineCss?: string;
 	viewportWidth?: number;
 	viewportHeight?: number;
-	maxHeight?: string | number;
```
noahtallen: Consumers of the component only pass numbers or "none", and the component implementation only handles none and number types. (E.g. its logic doesn't work with a value like "100px") So the string type wasn't specific enough which caused some issues in the component implementation.

packages/calypso-e2e/src/rest-api-client.ts
noahtallen, line 1: These changes will be removed after #74741 is merged. Basically just to unblock the PR until the real fix is merged.

noahtallen:
Pretty much everything is fixed now, but there are some issues with _internal_ WordPress data and data control types when compiling ETK. I don't understand why those aren't problematic for the other packages, but they do seem to be resolved when I temporarily update wp data to a newer version. (So this might be blocked on React 18.)

Unless this is an internal problem, I think we may need to update the DT definitions for WP data controls as well.

tyxla:
> Pretty much everything is fixed now, but there are some issues with _internal_ WordPress data and data control types when compiling ETK. I don't understand why those aren't problematic for the other packages, but they do seem to be resolved when I temporarily update wp data to a newer version. (So this might be blocked on React 18.)
> 
> Unless this is an internal problem, I think we may need to update the DT definitions for WP data controls as well.

I'd also expect that most of the data controls usage could be refactored to simpler means. The generators are no longer in use in Gutenberg and we should likely follow suite.

packages/block-renderer/src/components/block-renderer-container.tsx
miksansegundo: Addressed in https://github.com/Automattic/wp-calypso/pull/74843

packages/components/src/list-tile/index.tsx
jsnajdr: The trouble here is that TS doesn't know if `className` is a valid prop for the `trailing` element. `isValidElement` assures that it's of type `ReactElement<unknown>`, while asserting it as `ReactElement` makes it `ReactElement<any>`.

I'm confused by the six overloads of `cloneElement`s type, but apparently the difference between `unknown` and `any` matters.

I was able to fix the type error by calling:
```ts
isValidElement< { className: string } >( trailing )
```

Ideally we should specify a more specific type on `leading` and `trailing`, saying that it must be something on what we can put a `className` prop.

noahtallen: Added the < { className: string } > fix

noahtallen:
Rebased and updated to Typescript 5.1. Will then get this finished after #78711!

packages/browser-data-collector/src/report.ts
```diff
@@ -12,12 +12,19 @@ import {
...
 	id: string;
+	url: string;
 	beginning!: number;
 	end?: number;
 	data: ReportData = new Map();
 	startCollectors!: Collector[];
 	stopCollectors!: Collector[];
 
+	// Stubs that are necessary to implement but are not used.
+	body = null;
```
noahtallen, line 23: Typescript's definitions now require us to implement these, but we don't seem to use them anywhere.

packages/global-styles/src/global.d.ts
noahtallen: Will likely remove this after #78711

noahtallen:
Alright, this is almost ready to go. Working through a few remaining type issues in the client:

1. BaseButton types are still broken, potentially related to the rebase.
2. There are a handful of selectors with the type error "expected 1 argument, not 2/3". This seems to be related to `createSelector` selectors implemented in JS but used in TS. When I move those selectors to a .ts file, the error goes away. So I think the JSX type parsing is either less smart, or it was previously very permissible. Unfortunately, moving to .ts means fixing minor type errors in those selectors. Hopefully not a big deal.
3. a handful of other one-off issues

noahtallen:
I was digging into this selector issue and wanted to share what I found. Say you have a simple selector like this defined in a JS file:

```js
export const getImporterStatusForSiteId = createSelector(
	( state, siteId ) => {
		return Object.values( state.imports.importerStatus ).filter(
			( importer ) => importer.site.ID === siteId
		);
	},
	( state ) => [ state.imports.importerStatus ]
);
```

When using this selector -- `getImporterStatusForSiteId(site, siteId)` -- you get the error "Expected 1 arguments, but got 2" When inspecting the IDE type hints for `getImporterStatusForSiteId`, it says the return type is `(state: any) => any[]`. Completely omitting `siteId`!

It took a while to figure out why this was happening, and I think it has to do with how TS is inferring the `TProps` argument for `createSelector`:

```ts
type Dependant< TState, TProps extends any[], TDependency > = (
	state: TState,
	...props: TProps
) => TDependency;

export default function createSelector< TState, TProps extends any[], TDerivedState >(
	selector: ( state: TState, ...props: TProps ) => TDerivedState,
	getDependants:
		| Dependant< TState, TProps, any >
		| Dependant< TState, TProps, any >[] = DEFAULT_GET_DEPENDANTS,
	getCacheKey: ( state: TState, ...props: TProps ) => string = DEFAULT_GET_CACHE_KEY
): ( state: TState, ...props: TProps ) => TDerivedState {
 ...
```

For TS to infer `TProps`, it's going to look at all the arguments -- _including_ `getDependants`! In this case, it's noticing that our dependency function does not include `siteId`, and so it doesn't include it in `TProps`. 

You can see that happen here:
```js
export const getImporterStatusForSiteId = createSelector(
	( state, siteId ) => {
		return 1;
	},
	( state, siteId, fooBar ) => [ state.imports.importerStatus ]
);
```

When we add siteId and fooBar, and type of `getImporterStatusForSiteId` becomes `(state: any, siteId?: any) => number` (the correct type). Notice that it does _not_ include fooBar, and that the return type is now `number`.

I'm not sure how to fix this though. The type error does seem to go away when moving to a ts file, but that introduces a lot of other work since we have to ignore or type untyped stores.

sirreal:
Nice analysis @noahtallen. Something seems to have changed in this version of TypeScript with how it infers the type argument. There are different candidates to choose from, `createSelector`'s `TProp` type argument is used in multiple parameters and it looks like it used to prefer the selector arg but now prefers the dependents arg.

I added `Partial<TProps>` in the dependants signature and that seems to fix the issue. It also seems accurate looking at some of the uses.

noahtallen:
Thanks @sirreal! Unfortunately, I think this might cause some other issues. The problem is that `TProps` is an array, and I think `Partial` mostly works best with objects... so what seems to happen is that now the other arguments to `getDependent` are possibly undefined. Which isn't correct unfortunately! So we get this error in some Typescript selectors now:

```sh
Argument of type '(state: IAppState, id: number) => ...' is not assignable to parameter of type 'Dependant<IAppState, [id?: number | undefined], any> | Dependant<IAppState, [id?: number | undefined], any>[] | undefined'.
...
    Types of parameters 'id' and 'id' are incompatible.
      Type 'number | undefined' is not assignable to type 'number'.
        Type 'undefined' is not assignable to type 'number'. ts(2345)
```

noahtallen:
I created a bug report in TS to get broader feedback: https://github.com/microsoft/TypeScript/issues/55192

sirreal:
Good spot, I reverted the `Partial` change. I don't think `Partial` is important, but using a mapped type seems to discourage that type from being used for inference. Some kind of array identity type seems to work (`type ArrayId< T extends any[] > = Array< T[ number ] >`). I pushed a change using `ReadonlyArray` to do the same thing.

noahtallen:
Very interesting! That seems to be working a lot better, thank you :)

noahtallen:
I think I found another TS bug: https://github.com/microsoft/TypeScript/issues/53178#issuecomment-1659301034

noahtallen:
Another error in the selector component:

```ts
  Type '(state: AppState, siteId: number, pluginSlugs: string[]) => never[]' is not assignable to type 'Dependant<any, readonly (number | string[])[], any>'.
    Types of parameters 'siteId' and 'props' are incompatible.
      Type 'number | string[]' is not assignable to type 'number'.
        Type 'string[]' is not assignable to type 'number'.ts(2345)
```

here:
https://github.com/Automattic/wp-calypso/blob/18d427456ae953751e089be353bf86b9a3559c8f/client/state/plugins/installed/selectors-ts.ts#L286-L294

I think `ReadonlyArray` makes every array _element_ a union of the possible element types, which means every prop can be a lot of things. Super annoying.

noahtallen:
I think we can work around it with an extra generic: 

`export default function createSelector< TState, TProps extends any[], TDepProps extends TProps, TDerivedState >`. This doesn't seem to have other side effects, fixes the root cause, and still prevents `getDependants` from using different args than `selector`.

packages/components/src/list-tile/index.tsx
noahtallen: Added this in one other place, but I noticed another issue where `isValidElement` won't narrow to `ReactElement` whatsoever, even when we don't care about the type of the props. See https://github.com/microsoft/TypeScript/issues/53178#issuecomment-1659301034

In that case, it was easy to fix with our own wrapper function:

```ts
function isElement( el ): el is ReactElement {
  return isValidElement( el );
}
```

This was the root cause of issues with some base button types where you could pass either a react element or an object

packages/components/src/dialog/button-bar.tsx
```diff
@@ -15,10 +15,8 @@ export type BaseButton = {
 	target?: string;
 };
 
-export type Button = ReactElement | BaseButton;
```
noahtallen, line 18: I suppose removing this isn't strictly necessary any more, but I didn't love having yet another plain `Button` type that isn't very meaningful, so I just inlined the types in the few places it was used.

worldomonation:
I've triggered Pre-Release Tests against the calypso.live image and it passed.

worldomonation (approved):

tyxla (approved):
Great work @noahtallen 🚀

PR #74550, opened 2023-03-16
Migrate atomic transfer eligibility store to createReduxStore

noahtallen:
See #74399 for more details.

## Proposed Changes
Migrates the atomic transfer eligibility store to `createReduxStore`.

Along the way, I combined all the `useSelect` calls into a single one. Thoughts on this? In theory, it seems better because we're calling functions fewer times.

## Testing Instructions
TBD -- thoughts on how to test?


tyxla (commented):
There are also some custom registry selectors as part of this store that still use the `TransferSelectFn` type:

https://github.com/Automattic/wp-calypso/blob/9651d4a39d5075bcb10cf1bd3c98f705766eb519/packages/data-stores/src/automated-transfer-eligibility/selectors.ts#L21-L94

Anything we can do about those? Could we use the same type inference there?

noahtallen:
Hm, good point. I don't know if we can improve the types on the registry selector, but I wonder if registry selection is actually needed here. As far as I know, we only need that if a selector accesses other stores than its own -- which isn't the case here.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -118,31 +100,23 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
-			message: '',
-		} ) &&
+		! transferringBlockers.includes( eligibilityHoldsConstants.BLOCKED_ATOMIC_TRANSFER ) &&
 		( isBlockByTransferStatus || isTransferStuck )
 	) {
-		transferringBlockers?.push( {
-			code: eligibilityHoldsConstants.BLOCKED_ATOMIC_TRANSFER,
-			message: '',
-		} );
+		transferringBlockers.push( eligibilityHoldsConstants.BLOCKED_ATOMIC_TRANSFER );
```
noahtallen, line 115: There is a logic error here. Previously, when it was possible for `transferringBlockers` to be null, this error would not have been pushed to the array, even though it probably should be. The ultimate change was to make sure that `transferringBlockers` was initialized to an array at least.

packages/data-stores/src/automated-transfer-eligibility/selectors.ts
```diff
@@ -18,77 +16,73 @@ export const getAutomatedTransferEligibility = (
...
-			const warning = eligibilityWarnings?.find( ( { id } ) => id === 'wordpress_subdomain' );
-
-			return warning || null;
-		}
-);
+export const getEligibilityHolds = (
+	state: State,
+	siteId: number | null
+): TransferEligibilityError[] | null => {
+	const transferEligibility = getAutomatedTransferEligibility( state, siteId );
```
noahtallen: This is an internal selector in the same file, so the registry selector pattern isn't needed here. (Same for the others!)

packages/data-stores/src/automated-transfer-eligibility/constants.ts
```diff
@@ -1,44 +1,23 @@
...
-	no_vip_sites: eligibilityHolds.NO_VIP_SITES,
-	site_private: eligibilityHolds.SITE_PRIVATE,
-	site_graylisted: eligibilityHolds.SITE_GRAYLISTED,
-	non_admin_user: eligibilityHolds.NON_ADMIN_USER,
-	not_resolving_to_wpcom: eligibilityHolds.NOT_RESOLVING_TO_WPCOM,
-	no_ssl_certificate: eligibilityHolds.NO_SSL_CERTIFICATE,
-	email_unverified: eligibilityHolds.EMAIL_UNVERIFIED,
-	excessive_disk_space: eligibilityHolds.EXCESSIVE_DISK_SPACE,
-};
+export const statusMapping = {
```
noahtallen, line 10: The duplication and manually types here don't give us anything when we can use `as const`

packages/data-stores/src/automated-transfer-eligibility/selectors.ts
```diff
@@ -18,77 +16,58 @@ export const getAutomatedTransferEligibility = (
...
+	state: State,
+	siteId: number | null
+): TransferEligibilityWarning[] | null => {
+	const transferEligibility = getAutomatedTransferEligibility( state, siteId );
 
-export const getNonSubdomainWarnings = createRegistrySelector(
-	( select: TransferSelectFn ) =>
-		( state: State, siteId: number | null ): TransferEligibilityWarning[] | null => {
-			const eligibilityWarnings = select( STORE_KEY ).getEligibilityWarnings( siteId );
+	if ( ! transferEligibility?.warnings ) {
```
noahtallen: I simplified these null checks along the way

packages/data-stores/src/automated-transfer-eligibility/selectors.ts
```diff
@@ -18,77 +16,58 @@ export const getAutomatedTransferEligibility = (
...
-			const { warnings } = transferEligibility;
-
-			if ( ! warnings ) {
-				return null;
-			}
-
-			// combine plugin and theme warnings into one list
-			const combined = Object.values( warnings ).flat();
+		return errs;
+	}, [] );
```
noahtallen: This function was completely incorrect before, and I wonder how many bugs it caused. Previously, the function was _actually_ returning an array of strings, while claiming that it returned `TransferEligibilityError[]`, which is an object type with the shape `{ code, message }`. And the one place that consumed it interacted with this array as if it contained objects. 

**In other words, this selector was previously not doing anything at all.**

Since this is only consumed in a single place, and everywhere that consumes it only needed to interact with the `code`, I changed the type to be much more specific. It's now restricted to values of the statusMapping object. I also used `reduce` to avoid putting empty strings in the array.

So while it looks like a bigger change, this is basically just correctly encoding the previous behavior -- but with correct and specific types now.

Above, we updated the consumer of this selector to handle the types correctly.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -119,24 +102,18 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
 	const isTransferStuck = latestAtomicTransfer?.is_stuck || false;
 	const isBlockByTransferStatus = latestAtomicTransferError || false;
 
 	// Add blocked-transfer-hold when something is wrong in the transfer status.
 	if (
-		! transferringBlockers?.includes( {
-			code: eligibilityHoldsConstants.BLOCKED_ATOMIC_TRANSFER,
-			message: '',
-		} ) &&
+		! transferringBlockers?.includes( 'BLOCKED_ATOMIC_TRANSFER' ) &&
```
noahtallen, line 114: Typescript typechecks this now that `transferringBlockers` has specific types. As a result, `includes` can only be passed values of the `statusMapping` object. I think this is enough to avoid needing the constant here. (If we change it to something random, there'd be a typescript error.) 

But the side problem here is that there are two identical lists of constants stored in different places 🙃

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -119,24 +102,18 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
 		( isBlockByTransferStatus || isTransferStuck )
 	) {
-		transferringBlockers?.push( {
-			code: eligibilityHoldsConstants.BLOCKED_ATOMIC_TRANSFER,
-			message: '',
-		} );
+		transferringBlockers?.push( 'BLOCKED_ATOMIC_TRANSFER' );
```
noahtallen, line 117: I feel like this is a logic error -- when `transferringBlockers` is null (e.g. when there are no errors at all), then this check will never add this error. I don't know if this is intentional, and the logic in this file is complicated enough I didn't want to mess with it too much, beyond fixing the obviously wrong type mismatch.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -119,24 +102,18 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
 
 	// Filter the Woop transferring blockers.
 	const transferringBlockers = eligibilityHolds?.filter(
-		( hold: TransferEligibilityError ) => ! TRANSFERRING_NOT_BLOCKERS.includes( hold.code )
+		( hold: TransferEligibilityHold ) => ! TRANSFERRING_NOT_BLOCKERS.includes( hold )
```
noahtallen: See below for a bigger explanation -- but the selector providing `eligibilityHolds` actually returns an array of strings, not objects. (So this has been broken for who knows how long.)

I updated the types to be more specific, and fixed the main type issues in this file.

packages/data-stores/src/automated-transfer-eligibility/types.ts
```diff
@@ -29,17 +31,11 @@ export interface TransferEligibilityWarningsType {
...
+	errors: InternalTransferEligibilityError[];
 	is_eligible: boolean;
 	warnings: TransferEligibilityWarningsType;
 }
 
 export type State = {
 	[ key: number ]: TransferEligibility;
 };
-
-export interface StatusMapping {
```
noahtallen, line 41: This type is useless since `statusMapping` itself provides a lot of detail as a const object.

noahtallen:
@tyxla, I think removing the registry selectors works fine. But when I did that, I uncovered a serious type/logic issue, so I went ahead and fixed that as well.

packages/data-stores/src/automated-transfer-eligibility/selectors.ts
tyxla: Nice catch, @noahtallen!

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -64,49 +52,44 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
+				getSiteLatestAtomicTransfer,
+				getSiteLatestAtomicTransferError,
+				requiresUpgrade,
+				isSiteAtomic,
+			} = select( SITE_STORE ) as SiteSelect;
+			return {
+				eligibilityHolds: getEligibilityHolds( siteId ),
+				eligibilityWarnings: getEligibilityWarnings( siteId ),
+				wpcomSubdomainWarning: getWpcomSubdomainWarning( siteId ),
+				warnings: getNonSubdomainWarnings( siteId ),
```
jsnajdr: IMO the only selector that really makes sense is `getAutomatedTransferEligibility`. All others are just helper functions that extract information from the eligibility object. They could be exported from the `AutomatedTransferEligibility` module alongside the `store` object.

A very straightforward followup to this PR would be replacing the store with one simple React Query call 🙂

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -64,49 +52,44 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
+				getEligibilityWarnings,
+				getWpcomSubdomainWarning,
+				getNonSubdomainWarnings,
+			} = select( AutomatedTransferEligibility.store );
+			const {
+				getSiteLatestAtomicTransfer,
+				getSiteLatestAtomicTransferError,
+				requiresUpgrade,
+				isSiteAtomic,
+			} = select( SITE_STORE ) as SiteSelect;
```
jsnajdr: I'd personally select from separate stores in separate `useSelect` calls. Especially when they don't depend on each other, like result of one `select` being an input to another.

It breaks a big blob of code into several smaller ones, and also avoids calling selectors for store A when there's and update only in store B.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -64,49 +52,44 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
 		requestLatestAtomicTransfer( siteId );
 	}, [ requestLatestAtomicTransfer, siteId ] );
```
jsnajdr, line 59: This would be also sooo elegant as a React Query 🙂 Fetch from the `/latest` endpoint, but specify we don't want to cache the data and always want to fetch fresh info.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
tyxla: > A very straightforward followup to this PR would be replacing the store with one simple React Query call 🙂

💯 I strongly agree with this. It applies to most of those stores that are actually a client-side representation of server-side data with little to no interactivity involved.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
tyxla: This does bring a couple of `@wordpress/data` potential improvement questions, though: 

* Should we do some work in `@wordpress/data` to optimize for both scenarios?
* If the answer to the first question is "no", then should we update the docs to suggest one store per one `useSelect()` call for optimal performance?

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -10,20 +11,15 @@ import FormattedHeader from 'calypso/components/formatted-header';
 import { LoadingEllipsis } from 'calypso/components/loading-ellipsis';
 import WarningCard from 'calypso/components/warning-card';
 import { useSite } from 'calypso/landing/stepper/hooks/use-site';
-import {
-	AUTOMATED_ELIGIBILITY_STORE,
-	SITE_STORE,
-	ONBOARD_STORE,
-	PRODUCTS_LIST_STORE,
-} from 'calypso/landing/stepper/stores';
+import { SITE_STORE, ONBOARD_STORE, PRODUCTS_LIST_STORE } from 'calypso/landing/stepper/stores';
```
tyxla: The products list PR was merged, so this is going to need a rebase.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -10,20 +11,15 @@ import FormattedHeader from 'calypso/components/formatted-header';
...
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { addQueryArgs } from 'calypso/lib/url';
 import { ActionSection, StyledNextButton } from 'calypso/signup/steps/woocommerce-install';
 import { eligibilityHolds as eligibilityHoldsConstants } from 'calypso/state/automated-transfer/constants';
 import SupportCard from '../store-address/support-card';
 import type { Step } from '../../types';
 import type { ProductsListSelect, OnboardSelect, SiteSelect } from '@automattic/data-stores';
-import type { TransferEligibilityError } from '@automattic/data-stores/src/automated-transfer-eligibility/types';
+import type { TransferEligibilityHold } from '@automattic/data-stores/src/automated-transfer-eligibility/types';
```
tyxla: This is definitely odd. I wonder if we can export the type as part of the `@automattic/data-stores` package and import it from there as well, without the extra sub-path?

packages/data-stores/src/automated-transfer-eligibility/constants.ts
tyxla: It's a bummer that we're still duplicating those, though:

https://github.com/Automattic/wp-calypso/blob/a14b6727ffc1ed68b951c95f8a43ca6e1cb15ccc/client/state/data-layer/wpcom/sites/automated-transfer/eligibility/index.js#L18-L32

How do you feel about unifying them?

tyxla (commented):
Thanks for taking care of the `createRegistrySelector` selectors 👍 

I've added a couple more suggestions on top of @jsnajdr's excellent review.

This grew a bit larger than expected, but I think it's a wonderful improvement!

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
noahtallen: Cool, that's good to know.

noahtallen:
I started looking at deduping these, and I found one of the lists defined SIX different times across the repo! 

#### Transfer constants:
https://github.com/Automattic/wp-calypso/blob/94f955231c4e275f76d132bf04a7c81089fb40eb/client/state/automated-transfer/constants.ts#L1-L45

https://github.com/Automattic/wp-calypso/blob/9651d4a39d5075bcb10cf1bd3c98f705766eb519/client/landing/stepper/declarative-flow/internals/steps-repository/automated-copy-site/index.tsx#L12-L26

https://github.com/Automattic/wp-calypso/blob/9651d4a39d5075bcb10cf1bd3c98f705766eb519/client/landing/stepper/declarative-flow/internals/steps-repository/wait-for-atomic/index.tsx#L18-L32

https://github.com/Automattic/wp-calypso/blob/9651d4a39d5075bcb10cf1bd3c98f705766eb519/client/landing/stepper/declarative-flow/internals/steps-repository/woo-transfer/index.tsx#L18-L32

https://github.com/Automattic/wp-calypso/blob/6dd5aad11ebdb7af312d854dc7a7c71d0456304e/client/state/atomic/transfers/constants.ts#L1-L15

https://github.com/Automattic/wp-calypso/blob/f2c9d99b07e14ea938829e9b2e5b31b6fb388a90/client/state/atomic-transfer/constants.js#L1-L7

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
noahtallen: Turns out we didn't even need it since type inference worked where it was used

noahtallen:
I refactored the selectors such that they aren't selectors any more, and migrated everything to use that approach.

I started looking into de-duplicating the constants, but I think it'd be worth splitting out separately.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
```diff
@@ -46,68 +49,51 @@ const WooConfirm: Step = function WooCommerceConfirm( { navigation } ) {
...
+
+	const { eligibilityHolds, eligibilityWarnings, wpcomSubdomainWarning, warnings } =
+		useMemo( () => {
+			return {
+				eligibilityHolds: getEligibilityHolds( transferEligibility ),
+				eligibilityWarnings: getEligibilityWarnings( transferEligibility ),
+				wpcomSubdomainWarning: getWpcomSubdomainWarning( transferEligibility ),
+				warnings: getNonSubdomainWarnings( transferEligibility ),
+			};
+		}, [ transferEligibility ] );
```
tyxla: Interesting call to pull those out. While it looks like a good optimization, I wonder if you've done any comparison between having those inside the `useSelect()` and outside of it in a `useMemo()`.

packages/data-stores/src/automated-transfer-eligibility/utilities.ts
```diff
@@ -0,0 +1,52 @@
...
+	if ( ! transferEligibility?.errors ) {
+		return null;
+	}
+
+	return transferEligibility.errors.reduce< TransferEligibilityHold[] >( ( errs, { code } ) => {
+		if ( code in statusMapping ) {
+			// At this point, we know code is a key of statusMapping, so we help TS understand that.
+			errs.push( statusMapping[ code as keyof typeof statusMapping ] );
+		}
+		return errs;
```
tyxla, line 16: Nice simplification to hold only the statuses since we didn't really need the messages.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
tyxla: Fair enough. I've double-checked that we're keeping the logic as it was, which is what we should be after in this PR essentially.

packages/data-stores/src/automated-transfer-eligibility/index.ts
```diff
@@ -1,25 +1,22 @@
...
-	}
-	return STORE_KEY;
-}
+export const store = createReduxStore( STORE_KEY, {
+	actions,
+	reducer,
+	resolvers,
+	selectors: {
+		getAutomatedTransferEligibility: ( state: State, siteId: number | null ) =>
+			siteId ? state[ siteId ] : null,
```
tyxla: Could theoretically this return `undefined` if `state` doesn't have a `siteId` key yet? Do we want to return `null` in that scenario?

tyxla (commented):
The code is looking great! I've left some suggestions and questions, and it might be a good call to follow up with some further deduplication.

I haven't tested though since I didn't really know how to. Would be nice if we can test this thoroughly before we ship it, because we changed a bunch of things on top of the `createReduxStore` refactor.

client/landing/stepper/declarative-flow/internals/steps-repository/woo-confirm/index.tsx
noahtallen: That's a good point, I hadn't considered putting them inside the useSelect call. I agree that would have the same effect.

packages/data-stores/src/automated-transfer-eligibility/index.ts
noahtallen: Good point, I'll update that.

tyxla:
We might need to do some updates with regard to #74775 (the reason for the failing code style check)

tyxla:
This one won't be necessary once we merge #100472 that deletes the entire store altogether.

PR #74555, opened 2023-03-17
Checkout: Send correct tax data from one-click upsell to cart

sirbrillig:
## Proposed Changes

This PR modifies the one-click upsell dialog to correctly send tax data from the saved card to the shopping cart.

Screenshot 2023-03-20 at 2 01 09 PM

Requires D105292-code

Fixes https://github.com/Automattic/payments-shilling/issues/1471

## Testing Instructions

Make sure that D105292-code is applied.

- Use an account that has at least one saved credit card that includes tax info (any card that has a recent purchase in a taxable jurisdiction should work).
- With this PR running, visit `/checkout/example.com/offer-plan-upgrade/business/12345` but replace `example.com` with your test site.
- In your browser's JS console run `localStorage.setItem('debug', 'calypso:upsell-nudge')` and reload the page to view debug info.
- Click to agree to the upsell.
- In your browser's JS console, verify that you see `updating cart with contact info and product` and that it includes some objects that contain contact information; that contact information comes from the card data and will be used to construct the tax data for the cart.
- In your browser's network devtools, go to the most recent POST request to the `/me/shopping-cart` endpoint and verify that it includes the same info in the `tax` field.

Screenshot 2023-03-20 at 3 21 53 PM

### DIFF
--- a/client/lib/checkout/payment-methods.tsx
+++ b/client/lib/checkout/payment-methods.tsx
@@ -28,6 +28,18 @@ export interface PaymentMethod {
 	remember: '1' | '0';
 	stored_details_id: string;
 	user_id: string;
+	tax_location: StoredPaymentMethodTaxLocation | null;
+}
+
+export interface StoredPaymentMethodTaxLocation {
+	country_code?: string;
+	postal_code?: string;
+	subdivision_code?: string;
+	ip_address?: string;
+	vat_id?: string;
+	organization?: string;
+	address?: string;
+	city?: string;
 }
 
 export interface PaymentMethodMeta {
--- a/client/my-sites/checkout/composite-checkout/lib/contact-validation.tsx
+++ b/client/my-sites/checkout/composite-checkout/lib/contact-validation.tsx
@@ -294,7 +294,7 @@ function convertValidationResponse( rawResponse: unknown ): DomainContactValidat
 	};
 }
 
-export async function wpcomValidateTaxContactInformation(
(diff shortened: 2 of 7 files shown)


sirbrillig:
Holy crap, there's a bunch of type errors in the code that TS is not picking up. Like this one: 

https://github.com/Automattic/wp-calypso/blob/e4621d3d2498140e2f3a35284954aaa5bd7ad979/client/my-sites/checkout/upsell-nudge/purchase-modal/content.tsx#L217

`product` must not be undefined but you can see from the optional chaining that it can be (and it frequently is in my testing). Something is broken with TypeScript.

martian77:
I tested this, but I'm not 100% sure if I'm seeing the right things or not. 
- In my JS console I see `updating cart with contact info and product` rather than `updating the cart with contact info`, and there's nothing that looks like a tax object in that call. 
- But the network POST to `/me/shopping-cart` does have the tax location as described. Taxes are also shown and are correct for the card details. 
Screenshot 2023-03-21 at 11 04 42

sirbrillig:
> In my JS console I see updating cart with contact info and product rather than updating the cart with contact info, and there's nothing that looks like a tax object in that call.

Yes, sorry, I changed the log message to add the product info after writing the test instructions 🤦 so you're seeing the right thing, but thanks for commenting on it. That log message does not have the tax object; rather it has the contact info and vat info that will be used to construct the tax object. I've updated the instructions to be more clear but it sounds like this is working!

PR #74597, opened 2023-03-17
Update Akismet(site-less) purchase in me/purchases so that it can be viewed and managed.

elliottprogrammer:
This PR allows for Akismet (site-less) purchases to be viewed on the `me/purchases` page in Calypso and can click the purchase to view the purchase detail page at `/me/purchases/:site/:purchaseId`. 

Prior to this PR, clicking the Akismet purchase would do nothing, when it is supposed to take you to the purchase detail page, where you can view additional details as well as cancel or renew the subscription.
We also don't need to display the site name on the purchase when managing (siteless.akismet.com).

**Important Note:  The ability to _Cancel_ and _Renew_ the subscription purchase will need to be done in a follow-up PR's/Patches.

- See "**Before**" screenshot here:

Markup 2023-03-17 at 15 18 02

- See "**After**" Screenshot(s) here:

Screenshot on 2023-03-21 at 12-56-44

<hr/>

Markup 2023-03-19 at 15 48 14

Asana Task: 1201210512993648-as-1204136657007224/f

Additional notes:
Much of all main files & components in the `client/me/purchases/` directory are still built using React classes.  Ideally we should refactor all these to functional components, however I think it would be best to do this refactoring as a maintenance task in the future because it would be quite an undertaking as these files are all quite large and complex, and they should be refactored and tested without introducing additional changes.

## Testing Instructions

- Checkout this PR/branch and spin up Calypso.
    - Run `git fetch && git checkout add/akismet-me-purchases && yarn start`
- Go to `calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1`. and purchase Askimet (siteless).  Once you complete the purchase you should be redirected to a thank you page `http://calypso.localhost:3000/checkout/akismet/thank-you/ak_plus_yearly_1` which will be blank. The page does not exist yet (coming soon).
- Navigate to `/me/purchases` and look for your "**Akismet Plus (10K requests/month)**" purchase. (see screenshots above).
- Click into the "**Akismet Plus (10K requests/month)**" purchase, and verify you are directed to the purchase detail page (`/me/purchases/siteless.akismet.com/{:purchaseId}`). Verify the "Renew now" and "Cancel subscription" buttons are showing (Although the ability to renew and cancel siteless Akismet purchases are not working yet.) 

## Pre-merge Checklist


CodeyGuyDylan:
> **Important Note: The ability to Cancel and Renew the subscription purchase will need to be done in a follow-up PR's/Patches.

I have added a task to address this to the board: 1204136657007202-as-1204210417785429/f

client/me/purchases/purchase-item/index.jsx
```diff
@@ -77,20 +72,26 @@ class PurchaseItem extends Component {
...
 						<span className="purchase-item__is-error">
-							{ translate( 'Activate your product license key' ) }
+							{ translate( 'Activate your Jetpack product license key', {
+								components: {
+									br: <br />,
+									// TODO: These anchor links are causing React console warnings,
+									// "Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>."
+									// Because the <CompactCard> component that renders this also us surrounded by an anchor link.
+									// See: <Card> General Guidelines: https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/card#general-guidelines
+									// TLDR: Don't display more than one primary button or action in a single card. (in which the card itself if a primary action/link in this case)
```
CodeyGuyDylan: If I understand this correctly, you are removing the "Learn More" link for now until we can come back and fix this later correct?

If so, I know there are React warnings, but I feel as though we may want to keep this link here for now and leave this TODO comment so we can fix the warnings in the future. I think removing the link and leaving a TODO comment could leave it in limbo for a while, as I don't know when or who will get to this in the near future? Let me know what you think

client/me/purchases/purchases-site/header.jsx
```diff
@@ -6,23 +6,23 @@ import Site from 'calypso/blocks/site';
 import SitePlaceholder from 'calypso/blocks/site/placeholder';
 import QuerySites from 'calypso/components/data/query-sites';
 import { getSite } from 'calypso/state/sites/selectors';
-import { isJetpackTemporarySitePurchase } from '../utils';
+import { isTemporarySitePurchase } from '../utils';
```
CodeyGuyDylan, line 9: If `isTemporarySitePurchase` is being passed to this component as a Prop, do we need to import the function here as well?

CodeyGuyDylan (changes requested):
About the red icon here:
image

I wonder if we should render a different Icon for Akismet siteless subscriptions, since this icon indicates that the subscription is "disconnected." Since the Akismet site connection works differently than Jetpack, would it make more sense to use a less "error" type icon here?

Correct me if I am wrong but I think depending on how the customer uses the API key, this might never change like the Jetpack ones do when you connect a site

I had just a few other comments in addition to this 😄

client/me/purchases/purchase-item/index.jsx
elliottprogrammer: Oh my apologies, I probably should have added more context here by including a GH comment here with some additional explanation when I first posted this PR. 😉 .

No, the link is still there and showing. It looks the same, and is pointing to the same location. It was throwing a console warning before this PR just as it is still doing now. The only difference here is that originally the "Learn more text" was not being translated and it was being rendered by a separate component `<ExternalLink>{ children}</ExternalLink>`. So initially I thought maybe including the "Learn more" link in the same function call to`transate(..., {...} )` might fix the console warning, but it did not.  But still, it is translated now, so I just left it like that.

I could however put it back to the way it was, using `<ExternalLink>`, but make sure to `translate()` the "Learn more" text.. 
In fact, looking closer now, I think I will do that, because still, the way I have it now, I forgot to add `target="_blank"` and `rel="noopener noreferrer"`. Additionally, ExternalLink also adds some type of accessibility advantages for screen readers, I believe, too. Thanks for catching this, I'll get it fixed up. 👍

Oh, and as for the console warning, yes, the warning was already there before, and the fix, I feel is out of scope of this PR and would be best resolved in it's own separate PR.

client/me/purchases/purchases-site/header.jsx
elliottprogrammer: Thanks for catching this.  Actually `isTemporarySitePurchase` is not being passed in as a prop, so it needs to be removed from the `propTypes`. 👍

elliottprogrammer:
> About the red icon here: image
> 
> I wonder if we should render a different Icon for Akismet siteless subscriptions, since this icon indicates that the subscription is "disconnected." Since the Akismet site connection works differently than Jetpack, would it make more sense to use a less "error" type icon here?
> 
> Correct me if I am wrong but I think depending on how the customer uses the API key, this might never change like the Jetpack ones do when you connect a site

Yes good point. I'm not completely sure how the API key activation process is going to work exactly, and if there is going to be a case when Akismet is inactive and still needs the user to activate it using the API key, but either way, we should still have the Akismet icon here, IMO.

Also it would be nice/ideal to display the Akismet API key on the purchases 'detail' page. I'm trying to look into that right now, but we could proceed to merge this PR and then add the API key in follow-up PR, in order to reduce blockers and keep the flow going. 😉

client/me/purchases/manage-purchase/index.jsx
```diff
@@ -1081,12 +1083,12 @@ class ManagePurchase extends Component {
 						{ ! preventRenewal && ! renderMonthlyRenewalOption && this.renderRenewNowNavItem() }
 						{ ! preventRenewal && renderMonthlyRenewalOption && this.renderRenewAnnuallyNavItem() }
 						{ ! preventRenewal && renderMonthlyRenewalOption && this.renderRenewMonthlyNavItem() }
-						{ ! isJetpackTemporarySite && this.renderUpgradeNavItem() }
+						{ ! isTemporarySitePurchase( purchase ) && this.renderUpgradeNavItem() }
```
jboland88: I think we will want to allow for upgrades and removal of Akismet purchases from here, wondering if these checks should remain as `! isJetpackTemporarySite `.  However, since upgrades and removal are outside the scope of allowing the purchases to be viewed in the management page, maybe we can leave the new checks for now and add comments here indicating that these actions are pending for Akismet purchases?

client/me/purchases/purchase-item/index.jsx
```diff
@@ -77,19 +74,24 @@ class PurchaseItem extends Component {
...
+						"Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>."
+						Because the <CompactCard> component that renders this also us surrounded by an anchor link.
+						See: <Card> General Guidelines: https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/card#general-guidelines
+						TLDR: Don't display more than one primary button or action in a single card. (in which the card itself if a primary action/link in this case) */ }
 						<ExternalLink
 							className="purchase-item__link"
 							href="https://jetpack.com/support/install-jetpack-and-connect-your-new-plan/#how-can-i-activate-my-license-key-in-my-jetpack-installation"
 						>
-							Learn more
+							{ translate( 'Learn more' ) }
```
jboland88, line 94: Nice improvement!

client/me/purchases/purchase-item/index.jsx
```diff
@@ -77,19 +74,24 @@ class PurchaseItem extends Component {
 			} );
 		}
 
-		if ( isDisconnectedSite ) {
-			if ( isJetpackTemporarySite ) {
+		if ( isDisconnectedSite && ! isAkismetTemporarySitePurchase( purchase ) ) {
```
jboland88, line 77: Nitpick, would it make sense to add another conditional inside `if( isDisconnectedSite  )` like this:

```
if( isDisconnectedSite  ) {
    if( isAkismetTemporarySitePurchase( purchase ) ) {
        return null;
    }
}

```

That seems like it might be a bit easier to parse?

jboland88 (approved):
No blockers on this one for me - it tested well, I was able to view the purchase management screen for my test Akismet purchases and my existing Jetpack purchases are still showing as I expect them to.

I left a few small feedback items inline.

client/me/purchases/manage-purchase/index.jsx
elliottprogrammer: Yes good point, that makes sense. Thanks, will do! 🙂

client/me/purchases/purchase-item/index.jsx
elliottprogrammer: > would it make sense to add another conditional inside if( isDisconnectedSite  ) like this: ...

Yes, I see what you're saying, and yes, you would think so...  But `getStatus() {` is looong function(method) with a lot of checks (conditionals), and `if ( isDisconnectedSite ) {` is just one of them, and if we returned `null` right there, we will be returning too early, and returning a status of `null`, which is not what we want.

jboland88: Ahhh, I see what you mean. Thanks for that additional context - makes sense why this approach is needed 👍

PR #74602, opened 2023-03-17
Update data stores docs

noahtallen:
See #74399 for more details.

## Proposed Changes

Updates the documentation for `@automattic/data-stores` to reflect the latest usage. I kept in a section for the legacy approach since we haven't migrated everything yet, and changed the main example to something we did migrate already.

## Testing Instructions
None


tyxla (approved):
LGTM, thanks @noahtallen!

We might need some more updates if we end up changing the way we import data stores (as I suggested [here](https://github.com/Automattic/wp-calypso/issues/74399#issuecomment-1476016414), but those README updates could be done separately.

🚀

PR #74611, opened 2023-03-18
Changes to maintain the tabswitch aspect on mobile

### LINKED ISSUE #74523: `/subscriptions/settings`: Never display `TabsSwitcher` as a select-dropdown
As per our Figma design we should not turn our "Sites", "Comments" and "Settings" tabs into a dropdown for small screens.

The Figma design for mobile has a width of `414px`. When we downsize our screen to that width with our current implementation, the `TabsSwitcher` is displayed as a select-dropdown.

Image

vs Figma design

Image

Ideally, we should never turn our `TabsSwitcher` into a select-dropdown.

JuanLucha:
## Proposed Changes

The changes proposed in this PR make the tab switcher in Subscriptions Manager maintain the same aspect in mobile as in the desktop version.

This is the desktop version:
Screenshot 2023-03-18 at 19 37 35

This is the mobile version:
Screenshot 2023-03-18 at 19 38 22

## Testing Instructions

1. Apply this patch.
2. Go to http://calypso.localhost:3000/subscriptions/settings in your desktop browser.
3. You should see the desktop version as the desktop image above.
4. Shrink the browser window or use the IOS Simulator on your computer (if you use Mac).
5. You should see the mobile version as the mobile image above.

Fixes https://github.com/Automattic/wp-calypso/issues/74523

### DIFF
--- a/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
+++ b/packages/subscription-manager/src/components/TabsSwitcher/styles.scss
@@ -67,3 +67,43 @@
 		margin-bottom: 22px;
 	}
 }
+
+@media screen and ( max-width: $break-medium ) {
+	.section-nav.subscription-manager-tab-switcher {
+		.section-nav__mobile-header {
+			display: none;
+		}
+
+		.section-nav-group {
+			display: flex;
+		}
+
+		.section-nav__panel {
+			display: flex;
+			align-items: center;
+		}
+
+		.section-nav-tab__text {
+			display: inline;
+			width: auto;
+			text-align: center;
+			flex-grow: 0;
+			font-weight: 400;
+		}
+
+		.section-nav-tab__link {
+			display: block;
+			font-weight: 400;
+			background-color: transparent;
+		}
+
+		.section-nav-tabs {
+			flex-grow: 1;
+			flex-shrink: 0;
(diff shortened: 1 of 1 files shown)


packages/subscription-manager/src/components/TabsSwitcher/styles.scss
```diff
@@ -67,3 +67,43 @@
 		margin-bottom: 22px;
 	}
 }
+
+@media screen and ( max-width: $break-medium ) {
```
phcp, line 71: There is a property on `SectionNav` called `allowDropdown` (file: `packages/subscription-manager/src/components/TabsSwitcher/Tabs.tsx`), that may be a built-in solution to disable the dropdown on mobile.

JuanLucha: I tried that, but I think that wasn't the expected output when that prop was created, as it doesn't work like that.

phcp: Got it, thanks for clarifying!

phcp (approved):
Looks good and work as expected on both desktop and mobile views (even smaller than 400px).
Screenshot 2023-03-20 at 12 52 54

PR #74646, opened 2023-03-20
Staging Sites: Fix incorrect plan display under sites page

kozer:
Fixes https://github.com/Automattic/dotcom-forge/issues/1953

## Proposed Changes

In this pr, I propose fixing the staging incorrect plan display under the sites page, by using the plan from the associated production site.

**Some notes for this pr:**

- Initially, I tried to implement this on the server side. This had some problems, as shown [here](p1679310173536649-slack-C04GESRBWKW)
- Under the `/sites` page we use the `useQuery` hook. Using this hook, the fetched doesn't end up in the main reducer that the site uses (`client/state/sites/reducer.js`). We end up fetching the sites twice. **We may need to revisit that.** I created [this issue ](https://github.com/Automattic/wp-calypso/issues/74647) for  having further discussions.
- As a result, this introduced some discrepancies. `useQuery` was fetching `wpcom_production_blog_id` and `wpcom_staging_blog_ids` but the "main" reducer didn't. I added those fields there as well. Also, please note, that I didn't handle the plans in the "main" reducer as well for safety reasons. `/sites` is a "self-contained" page. But changing the plans in the main reducer that is used everywhere, may introduce bugs, and will make things harder to test for this relatively simple pr.

## Testing Instructions

1. Make sure you have a staging site. If not you have to create one.
2. Visit `/sites`
3. Ensure that your staging site has the same plan as it's production counterpart.

stage_prod_same_plan

*

## Pre-merge Checklist


client/data/sites/use-site-excerpts-query.ts
```diff
@@ -26,11 +26,31 @@ const fetchSites = (): Promise< { sites: SiteExcerptNetworkData[] } > => {
...
+	// Create a map once, and use that,instead of use find every time, in the map below.
+	const sitesById = sites.reduce( ( allSites, site: SiteExcerptNetworkData ) => {
+		allSites[ site.ID ] = site;
+		return allSites;
+	}, {} as { [ id: number ]: SiteExcerptNetworkData } );
+	return sites.map( ( site: SiteExcerptNetworkData ) => {
+		if ( site?.options?.wpcom_production_blog_id ) {
+			const productionSitePlan = sitesById[ site?.options?.wpcom_production_blog_id ]?.plan;
+			if ( productionSitePlan ) {
+				site.plan = productionSitePlan;
```
danielbachhuber: Rather than injecting this into the React query data, could we instead use a much higher level helper function for the label?

Here's some pseudo-code to describe:

```
export const getSitePlanLabel = ( siteId ) => {
    const site = getSite( siteId );
    if ( site?.options?.wpcom_production_blog_id ) {
        return getSitePlanLabel( site.options.wpcom_production_blog_id ).label
   }
   return site.plan.label;
}
```

I'm a little fearful that the current approach will have unexpected negative side effects. This approach will limit the surface area of the change.

kozer: Yes, we can do that. I'm concerned about that as well. Whatever path we choose, at its core, the plan is free, and not Business. 
Thankfully, that query is used only by the dashboard to my understanding, so it doesn't seem that will have many implications.

wojtekn: I also think having a plan label function that is used only in this particular place would be safer.

We could also use the `isStagingSite` selector there, and only if it's true, display the parent site plan. Or say "Staging", as technically, the staging site doesn't have a plan?

kozer: Update:
- I moved the same function higher and change it a bit, to alter only the plan label, at the `SitesDashboard` level.
- Avoid using recursion for safety reasons.

wojtekn: What do you think about moving it even higher, by using a function similar to that proposed by @danielbachhuber and using it in the `<SitePlan>` component?

kozer: > What do you think about moving it even higher, by using a function similar to that proposed by @danielbachhuber 

What do you mean by higher? This is the place the sites that are used inside the `/sites` are fetched. Higher in the hierarchy, those sites are not available ( This is related to the issue I have created, we fetch the sites twice and there could be an inconsistency)
Also, I don't think using recursion is a very good idea, especially in js. Most probably it will not create any stack overflows, but I want to be as safe as I can and eliminate that possibility......

> and using it in the <SitePlan> component?

...... however, the logic can be rewritten, so I'm thinking to create a container component that is connected to the state and render the appropriate label. What do you think? 

More specifically, I'm thinking of replacing this:

```
				<SitePlan>
					{ isNotAtomicJetpack( site ) && ! site.plan?.expired && (
						<SitePlanIcon>
							<JetpackLogo size={ 16 } />
						</SitePlanIcon>
					) }
					{ site.plan?.expired ? (
						<PlanRenewNagContainer>
							<PlanRenewNag
								plan={ site.plan }
								isSiteOwner={ site?.site_owner === userId }
								checkoutUrl={ `/checkout/${ site.slug }/${ site.plan?.product_slug }` }
							/>
						</PlanRenewNagContainer>
					) : (
						site.plan?.product_name_short
					) }
				</SitePlan>
```
with a connected component.

Another problem, I think of, is that as I see here that there is a chance that:

- The production site has an expired business plan
- For staging, we show "Business" but the "expired" property is false, as we replace only the label (staging's plan is free)
- We incorrectly show a "Business" label without the expiration notice.

How do we want to handle this?

**Proposals:**
- Add a "Staging" label instead of "Business"
- We don't care and leave it as is.
- We also use the expired property of the production site.

cc: @wojtekn

wojtekn: > What do you mean by higher?

I see it's not really higher, I meant to use a function that we would call in `<SitePlan>`.

> More specifically, I'm thinking of replacing this:
> with a connected component.

Agreed, this styled component has grown a bit, it got some business logic, so it makes sense to extract it to its own component. If we use `productionSitePlan` there, we could base all the logic on the parent site, so even renew nag would work.

My (and @danielbachhuber 's ) general thought was to use a helper function that runs late in the flow and replaces the plan before it's displayed, instead of modifying data for all sites upfront using `map` earlier in the flow.

kozer: > My (and @danielbachhuber 's ) general thought was to use a helper function that runs late in the flow and replaces the plan before it's displayed, instead of modifying data for all sites upfront using a map earlier in the flow.

Actually, the modified code will do exactly the same, all the sites will go through this check and only those that are staging would be modified, the same as now. But it will be more clear for anyone that visits the code in the future to immediately see the change, it makes sense.

kozer: You are correct, we will not modify anything by going this way. Also, it will make it more apparent for anyone who visits the code what is going on, and why.

kozer: cc: @wojtekn @danielbachhuber , I updated as discussed.

client/sites-dashboard/components/sites-site-plan.tsx
```diff
@@ -0,0 +1,59 @@
...
+			{ site.plan?.expired ? (
+				<PlanRenewNagContainer>
+					<PlanRenewNag
+						plan={ site.plan }
+						isSiteOwner={ site?.site_owner === userId }
+						checkoutUrl={ `/checkout/${ site.slug }/${ site.plan?.product_slug }` }
+					/>
+				</PlanRenewNagContainer>
+			) : (
+				site.plan?.product_name_short
```
wojtekn: We still use the current `site` plan instead of one loaded for the parent site. Should we use `plan` here?

client/sites-dashboard/components/sites-site-plan.tsx
```diff
@@ -0,0 +1,59 @@
...
+	} );
+
+	return (
+		<SitePlanContainer>
+			{ isNotAtomicJetpack( site ) && ! plan?.expired && (
+				<SitePlanIcon>
+					<JetpackLogo size={ 16 } />
+				</SitePlanIcon>
+			) }
+			{ site.plan?.expired ? (
```
wojtekn: Should we explicitly hide this nag for staging sites?

wojtekn (changes requested):
@kozer it's great that we refactor this to a separate component.

While testing it, I noticed one more thing. I have access to one staging site that was shared with me by another user. I don't have access to the corresponding production site, and the plan shows as empty. This is expected, as there is no plan data for that parent site in the store.

This looks like a common use case, as the site owner can create staging from the existing production site and share only the staging site with the developer. It seems we are not able to display the parent site plan in all cases. Thoughts? Should we go back to the fallback idea and display just "Staging"?

client/sites-dashboard/components/sites-site-plan.tsx
kozer: I believe that deciding which plan to show, it's the component's responsibility, and by doing that we don't add extra logic to the parent component. Thanks for your comment!

client/sites-dashboard/components/sites-site-plan.tsx
kozer: > Another problem, I think of, is that as I see here that there is a chance that:
> The production site has an expired business plan
> For staging, we show "Business" but the "expired" property is false, as we replace only the label (staging's plan is free)
> We incorrectly show a "Business" label without the expiration notice.
> How do we want to handle this?

I asked this above, and I couldn't decide what to do. I thought that if the user is aware of this, that the production and staging share the same plan, it makes sense, and this was my original thought on that. If he is aware, not showing the nag for the staging but showing it for the production is inconsistent.

Is this going to be communicated to the user?

client/sites-dashboard/components/sites-site-plan.tsx
wojtekn: I meant that this code currently doesn't work for me - it shows "Free" for staging sites. We load the parent site's plan into `plan` constant but still use `site.plan` below.

client/sites-dashboard/components/sites-site-plan.tsx
wojtekn: We could display nag pointing to the parent site, however, this wouldn't make sense if the user doesn't have access to the parent site. Maybe we should also add an explicit `! is_ staging` check here?

kozer: I ended up changing the function a little bit. I only use the label instead of the plan. I also used is staging to determine if the nag will be shown or not.

wojtekn:
@kozer per slack discussion, let's modify this to not use parent site data, to display "Staging" for the staging site and the old logic otherwise. Let's keep refactor as this is a good move.

kozer:
Already on it

kozer:
@wojtekn , updated.

client/sites-dashboard/components/sites-site-plan.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface SitePlanProps {
+	site: SiteExcerptData;
+	userId: number | null;
+}
+
+const STAGING_PLAN_LABEL = 'Staging';
+
+export const SitePlan = ( { site, userId }: SitePlanProps ) => {
+	const isStaging = useSelector( ( state ) => isSiteStaging( state, site.ID ) );
+	const hasProductionBlog = isStaging && site?.options?.wpcom_production_blog_id;
```
wojtekn: I think that `isStaging` should be enough.

wojtekn (approved):

client/sites-dashboard/components/sites-site-plan.tsx
```diff
@@ -0,0 +1,60 @@
...
+					{ site.plan?.expired && (
+						<PlanRenewNagContainer>
+							<PlanRenewNag
+								plan={ site.plan }
+								isSiteOwner={ site?.site_owner === userId }
+								checkoutUrl={ `/checkout/${ site.slug }/${ site.plan?.product_slug }` }
+							/>
+						</PlanRenewNagContainer>
+					) }
+					{ site.plan?.product_name_short }
```
danielbachhuber, line 53: @kozer @wojtekn FYI - this is buggy logic https://github.com/Automattic/wp-calypso/pull/75590

PR #74657, opened 2023-03-20
Draft: Inject the subkey cookie in window._subscriptionManagementSubkey

TimBroddin:
⚠️ This is a WIP, to get some code discussion going. ⚠️

## Proposed Changes

This PR takes the subkey cookie and passes it into `window._subscriptionManagementSubkey`. 

## Testing Instructions

TBD

## Pre-merge Checklist


yansern:
Could we possibly limit the rendering of the subkey script to only when the page is `/subscriptions`?

tyxla (commented):
Is this something we could use an existing global variable for?

One that comes to mind is [`configData`](https://github.com/Automattic/wp-calypso/blob/69e8b25831419281b7c0afe860c219dcabfcec90/client/document/index.jsx#L81).

It already combines data from a few different sources - server (secrets), client (config values), additional custom ones (env), so why not add another one to those?

TimBroddin:
Nice! Thanks for the pointer @tyxla !
I'll update this PR.

yansern:
That's very helpful! Thanks @tyxla!

TimBroddin:
Hey @tyxla 

I changed it to store it inside configData. One question: how would this work with caching? Is the sever-rendered output of Calypso stored on a CDN? And if so: is there a way to opt out of it for certain routes?

🙏

tyxla:
> how would this work with caching? Is the sever-rendered output of Calypso stored on a CDN? And if so: is there a way to opt out of it for certain routes?

AFAIK it is never cached. This should be easily confirmed by the document caching headers when testing with a logged-out user. 

I'd like to understand why we're not using cookies directly, though. What do you think?

TimBroddin:
> > how would this work with caching? Is the sever-rendered output of Calypso stored on a CDN? And if so: is there a way to opt out of it for certain routes?
> 
> AFAIK it is never cached. This should be easily confirmed by the document caching headers when testing with a logged-out user.
> 
> I'd like to understand why we're not using cookies directly, though. What do you think?

The cookie we're using has the `httpOnly` attribute set to true, so it cannot be read from JavaScript. We've disabled this attribute for now here: D105375-code . 

Exposing it via a global variable seemed like a `cleaner` solution, since the cookie is valid for all` .wordpress.com` subdomains. This way, we can still keep the httpOnly attribute, and prevent it from being read by JavaScript outside of Calypso.

tyxla:
I see, thanks for clarifying!

Now, another question: I wonder if we need this that early at boot time, or if we can retrieve it later with a REST API request?

TimBroddin:
> I see, thanks for clarifying!
> 
> Now, another question: I wonder if we need this that early at boot time, or if we can retrieve it later with a REST API request?

I tried a lot of things, but I ran into CORS-limitations. For example, passing `{ "credentials": "include" }` to a fetch request triggers a CORS error, and the `wpcomRequestProxy` doesn't seem to pass along cookies as well.

tyxla:
> > I see, thanks for clarifying!
> > Now, another question: I wonder if we need this that early at boot time, or if we can retrieve it later with a REST API request?
> 
> I tried a lot of things, but I ran into CORS-limitations. For example, passing `{ "credentials": "include" }` to a fetch request triggers a CORS error, and the `wpcomRequestProxy` doesn't seem to pass along cookies as well.

I was referring more to the scenario where we hit an endpoint, then that endpoint retrieves the `$_COOKIE` value and returns it in its response body, not in the response headers. CORS shouldn't be an issue there IIUC?

yansern:
> I wonder if we need this that early at boot time, or if we can retrieve it later with a REST API request?

> I was referring more to the scenario where we hit an endpoint, then that endpoint retrieves the `$_COOKIE` value and returns it in its response body, not in the response headers. CORS shouldn't be an issue there IIUC?

That is certainly possible. We could create an endpoint that returns the cookie auth data.

But of course, to get back to the earlier question, if there's a way to get this data earlier without calling an endpoint, we'd very much prefer that.

At the same time, by using an API endpoint, we could possibly return more useful data in the API endpoint that could offset the tradeoff of that one additional API request, e.g. including data like is there an WP account linked to this email on the cookie, etc.

yansern:
> That is certainly possible. We could create an endpoint that returns the cookie auth data.

Wait, need to double check.
@TimBroddin When the API request is made through `wpcomRequestProxy`, is the cookie available/retrievable on the server-side?

TimBroddin:
I think it wasn't when I checked it. Otherwise we wouldn't have been jumping through all this hoops.

Will double check tomorrow.

yansern:
> I think it wasn't when I checked it. Otherwise we wouldn't have been jumping through all this hoops.
> Will double check tomorrow.

If that is the case, then yes, that's why we needed to add this to `configData` @tyxla.

tyxla:
> > I think it wasn't when I checked it. Otherwise we wouldn't have been jumping through all this hoops.
> > Will double check tomorrow.
> 
> If that is the case, then yes, that's why we needed to add this to `configData` @tyxla.

Let me know when you check it.

Anyway, suppose it ends up being like that, assuming those cookies are necessary only for logged-in users. In that case, one alternative we could explore is adding the cookie value to the bootstrapped user data and the current user object. I recently suggested a similar solution here if that could help: p1678696245577889/1678473246.782839-slack-C02DQP0FP.

TimBroddin:
Since the merge with trunk went a bit crazy, I'll open a new PR for this.

PR #74639, opened 2023-03-20
Create new Explat experiment on Plans step in the paid media flow

donlair:
## Proposed Changes

* This PR implements Explat experiment `paid_media_signup_2023_03_legacy_free_presentation`
* The treatment variation changes two things:
1) Replaces the Monthly term length option with a Two year term length option
2) Restores the prior control behavior for the Free plan link (the link is not shown if the person selects a paid domain or the skip option on the domains step).

CleanShot 2023-03-20 at 08 09 53@2x

## Testing Instructions

* Assign yourself to the Treatment variation of `paid_media_signup_2023_03_legacy_free_presentation` in Abacus.
* Checkout this PR and start it locally.
* Navigate to `/start/onboarding-pm`
* Proceed through the domain step with a free domain option. Confirm that the Free link is shown.
* Go back to the domain step and select a paid domain. Confirm that the Free link is not shown.
* Go back to the domain step and select the skip option on the right. Confirm that the Free link is not shown.
* Confirm that the 1 year and 2 years term-length toggles are present on all experiences.

PR #74665, opened 2023-03-20
Changing discount description if the user is viewing two-year plans

donlair:
## Proposed Changes

* @gmovr [noticed](https://github.com/Automattic/wp-calypso/pull/74639#pullrequestreview-1348637475) a bug in the two-year plans that mentioned paying annually on the plans page even if the user was viewing two-year plans.
* This PR adds a check for two year plans and updates the language.

## Testing Instructions

* Check out this PR and start Calypso locally.
* Assign yourself to the Explat experiment as described in [this PR's](https://github.com/Automattic/wp-calypso/pull/74639) test instructions.
* Confirm that the discount language now refers to bi-annual payments.

### DIFF
--- a/client/my-sites/plan-features-comparison/header.jsx
+++ b/client/my-sites/plan-features-comparison/header.jsx
@@ -64,6 +64,7 @@ export class PlanFeaturesComparisonHeader extends Component {
 			translate,
 			annualPricePerMonth,
 			isMonthlyPlan,
+			planType,
 		} = this.props;
 
 		if ( isMonthlyPlan && annualPricePerMonth < rawPrice ) {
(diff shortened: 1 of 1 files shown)


client/my-sites/plan-features-comparison/header.jsx
```diff
@@ -84,17 +84,28 @@ export class PlanFeaturesComparisonHeader extends Component {
...
-				args: { discountRate },
-			} );
+
+			const annualDiscountText = [
+				'personal-bundle-2y',
+				'value_bundle-2y',
+				'business-bundle-2y',
+				'ecommerce-bundle-2y',
+			].includes( planType )
+				? translate( `You're saving %(discountRate)s%% by paying bi-annually`, {
```
gmovr: I _think_ it's **biennial** here (and no hyphen). As **biannual** is something that happens twice in a year, while **biennial** happens once every two years. 

Not sure which one we use across Calypso/WPcom though, but biennial should be the correct one. 

If the Abacus test is running outside of language=EN, we also need to "string freeze" this, e.g. if it's running in the US (as it also has Spanish and many other language users). 

Approving this. The logic works. So you can just do a quick push with the grammar change, and add string freeze if needed depending on the test config. 👍

gmovr (approved):

client/my-sites/plan-features-comparison/header.jsx
donlair: At least for the US, I think biannual makes the most sense here. I researched a little and seems like technically "biannual" is ambiguous and can mean either twice per year or once every other year. I think "every other year" is the common interpretation.

Since this PR is just being seen by US audiences, I'm going with biannual but we should have a broader conversation about the overall language if/when we deploy this more widely.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7880501

Hi @donlair, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `billed as %(price)s biannually`
- `You're saving %(discountRate)s%% by paying biannually`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7880501).

PR #74666, opened 2023-03-20
Newsletter/LiB: Signup form, change button background

### LINKED ISSUE #74659: Newsletter: Unify blue hues
[Context](https://wp.me/puPv3-e72#comment-25266)

The CTA blue hues are slightly different between the setup and the launchpad (#117AC9 vs #0675C4)

The Launchpad #0675C4 is the correct one, which corresponds to studio-blue-50

image

escapemanuele:
The CTA blue hues are slightly different between the setup and the launchpad (#117AC9 vs #0675C4)

This PR changes the CTA blu in the signup form for `newsletter`and `link-in-bio` flows.

## Testing
1. Check live link
2. Navigate in incognito to reach the signup form
3. Check that the cta blue is `--color-primary` = `--studio-blue-50`

Fixes https://github.com/Automattic/wp-calypso/issues/74659

### DIFF
--- a/client/blocks/signup-form/index.jsx
+++ b/client/blocks/signup-form/index.jsx
@@ -962,11 +962,17 @@ class SignupForm extends Component {
 			);
 		}
 
+		const params = new URLSearchParams( window.location.search );
+		const variationName = params.get( 'variationName' );
+
 		return (
 			<LoggedOutFormFooter isBlended={ this.props.isSocialSignupEnabled }>
 				{ this.termsOfServiceLink() }
 				<FormButton
-					className="signup-form__submit"
+					className={ classNames(
+						'signup-form__submit',
+						variationName && `${ variationName }-signup-form`
+					) }
 					disabled={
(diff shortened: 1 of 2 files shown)


client/blocks/signup-form/index.jsx
```diff
@@ -962,11 +962,17 @@ class SignupForm extends Component {
 			);
 		}
 
+		const params = new URLSearchParams( window.location.search );
```
renancarvalho, line 965: I believe we could get this information from `this.props.flowname`, WDYT ?

renancarvalho (changes requested):

client/blocks/signup-form/index.jsx
escapemanuele: It would be great, but sadly flowname returns `account`. User signup is not in Stepper... yet 👍

renancarvalho (approved):

PR #74673, opened 2023-03-20
Fix ftme__sell e2e spec

youbrokesomething:
## Proposed Changes

* Fix for ftme__sell spec that fails due to incorrect button text. Changes include a new step to click on a specific button locator.
  - The logic for the theme picker in the start site flow is: themes with style variations have a 'Continue' button, and themes without style variations have 'Start with {themeName}.' The theme currently used in the spec has style variations.
  - The theme preview screen can have one of two different continue buttons depending on the viewport size, so two new selectors were added. 

## Testing Instructions

* The ftme__sell e2e spec should pass.

## Pre-merge Checklist

### DIFF
--- a/packages/calypso-e2e/src/lib/flows/start-site-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/start-site-flow.ts
@@ -21,6 +21,8 @@ const selectors = {
 
 	// Themes
 	individualThemeContainer: ( name: string ) => `.design-button-container:has-text("${ name }")`,
+	themeContinueButton: '.design-preview__sidebar-action-buttons:has-text("Continue")',
+	themeContinueButtonSmall: '.step-container__navigation:has-text("Continue")',
 
 	// Goals
 	goalButton: ( goal: string ) => `.select-card__container:has-text("${ goal.toLowerCase() }")`,
@@ -166,5 +168,17 @@ export class StartSiteFlow {
 	 */
 	async selectTheme( themeName: string ): Promise< void > {
 		await this.page.getByRole( 'button', { name: themeName } ).click();
+	}
+
+	/**
+	 * Confirms Theme selection by clicking the correct Continue button depending on viewport size.
+	 * themeContinueButton will be visible at viewports above 1080px.
+	 * themeContinueButtonSmall will be visible at viewports below 1080px.
+	 */
+	async confirmThemeSelection(): Promise< void > {
+		await Promise.race( [
+			this.page.click( selectors.themeContinueButton ),
+			this.page.click( selectors.themeContinueButtonSmall ),
+		] );
 	}
 }
--- a/test/e2e/specs/onboarding/ftme__sell.ts
+++ b/test/e2e/specs/onboarding/ftme__sell.ts
@@ -178,7 +178,10 @@ describe( DataHelper.createSuiteTitle( 'FTME: Sell' ), function () {
 
 		it( 'Select theme', async function () {
 			await startSiteFlow.selectTheme( themeName );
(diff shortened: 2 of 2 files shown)


youbrokesomething:
I have just the two continue button selectors currently. I could add the 'Start with ${ themeName}' option to the race condition to allow the test to continue passing if we change the theme to a theme that doesn't have style variations.

Thoughts/opinions?

packages/calypso-e2e/src/lib/flows/start-site-flow.ts
```diff
@@ -167,4 +169,16 @@ export class StartSiteFlow {
 	async selectTheme( themeName: string ): Promise< void > {
 		await this.page.getByRole( 'button', { name: themeName } ).click();
 	}
+
+	/**
+	 * Confirms Theme selection by clicking the correct Continue button depending on viewport size.
+	 * themeContinueButton will be visible at viewports above 1080px.
+	 * themeContinueButtonSmall will be visible at viewports below 1080px.
```
worldomonation, line 176: **comment:blocking** - turns out, the assumption here is not entirely correct!

As you can see here as an example, the Winkel theme shows the same `Continue` button regardless of the horizontal viewport width.

Desktop
image

Mobile
image
----

Instead, the actual difference is whether the theme is FSE-enabled or not. You can see the difference with Vivre below, which is **not** a FSE theme.
image

So in summary, we have two situations:
- theme is FSE: `Continue` button is shown to confirm the selected theme.
- theme is not FSE: `Start with ${themeName}` is button is shown to confirm the selected theme.

### So what does this mean?

This has implications for the added selectors in `StartSiteFlow` - we don't actually need to use any new selectors!

packages/calypso-e2e/src/lib/flows/start-site-flow.ts
```diff
@@ -21,6 +21,8 @@ const selectors = {
 
 	// Themes
 	individualThemeContainer: ( name: string ) => `.design-button-container:has-text("${ name }")`,
+	themeContinueButton: '.design-preview__sidebar-action-buttons:has-text("Continue")',
+	themeContinueButtonSmall: '.step-container__navigation:has-text("Continue")',
```
worldomonation, line 25: **issue:blocking** - because the distinction is between FSE and non-FSE themes, this selector is not actually going to work if we opt to test a theme that is not FSE. That's because for such themes, the button to confirm theme selection reads "Start with ${themeName}` instead.

I should have been clearer that we do not need to worry about mobile viewports for the time being.

worldomonation: **issue:continuing** - Furthermore with Playwright adding support for obtaining locators by one of the many a11y attributes, it would be preferable to begin deprecating the use of selector objects as time goes on. As an added bonus, it makes the code easier to follow.

An example of what Playwright can do:

```typescript
await this.page.getByRole( 'button', { name: "Continue" } ).click();
```

worldomonation (changes requested):
The structure of the changes are good - adding an explicit step to confirm the theme selection is a good idea! 

That said, the initial assumption we worked from turned out to be unfortunately incorrect, so let's pivot to not worry about Mobile viewport for the time being and support both FSE and non-FSE theme selections.

worldomonation:
Also, **issue:blocking** - please update the title to follow the convention used by KitKat.

worldomonation:
@youbrokesomething Checking up on the progress of this PR. If you would like some guidance and/or for me to take over the PR, let me know.

worldomonation:
Superseded by https://github.com/Automattic/wp-calypso/pull/76659.

PR #74676, opened 2023-03-20
Replace storedCards Redux state with useStoredPaymentMethods hook

sirbrillig:
There are several existing API endpoints for fetching stored payment methods: `/me/stored-cards` and `/me/payment-methods`. The former (which came first) only returns saved credit cards and only ones that are not expired. The latter supports cards, (PayPal) payment agreements, or both, and allows returning either unexpired methods or all of them. (There are also a series of related endpoints which allow adding methods, deleting methods, fetching and changing tax information for methods, and fetching and changing backup status for methods.)

Calypso uses both of these endpoints in various places: primarily the payment method list pages and checkout. Although they have slightly different APIs and result sets, they return information in the same format. You can best see that format in the [calypso data type](https://github.com/Automattic/wp-calypso/blob/ad025008afff8f1f3f7132cf800835fe59bb0f7a/client/lib/checkout/payment-methods.tsx#L15-L31) that represents it.

However, creating that format does a lot of DB work in order to find out when the payment method was added, when it was last used, and to attach the payment method's meta values to the data. Most of this information is not needed by calypso or its format requires [considerable hoops](https://github.com/Automattic/wp-calypso/blob/ad025008afff8f1f3f7132cf800835fe59bb0f7a/client/me/purchases/manage-purchase/payment-info-block.tsx#L31) to parse. In addition, there is already a normalized object type on the backend for payment methods which would be less work to create and more useful to the consumers of the API endpoints.

D105098-code adds a new version of the `/me/payment-methods` API endpoint which returns an array of serialized `Stored_Payment_Method` objects instead of the custom form used by its predecessors. We had to use version 1.2 because calypso already uses version 1.1 of the endpoint.

## Proposed Changes

This PR creates a new React hook, `useStoredPaymentMethods` (and an HOC version `withStoredPaymentMethods`), which uses the 1.2 version of the `/me/payment-methods` endpoint. We then replace all instances of the other two endpoints with the new one.

Since the hook uses React Query for state management, this PR removes the `storedCards` state from Redux and all its selectors, actions, and reducers.

This PR looks pretty big but most of it is adding new tests and removing Redux state.

Requires D105098-code.

## Testing Instructions

This PR makes sure there are basic tests covering the loading of stored payment methods in:

- Checkout. (`/checkout`)
- Payment method list (and deletion). (`/me/purchases/payment-methods`)
- Add/change payment method. (start at `me/purchases`, click a purchase, click "Add a payment method" or "Change payment method")
- One-click checkout. (see https://github.com/Automattic/wp-calypso/pull/74555)

You can manually view each of those pages and verify that the stored payment methods appear as expected.


client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
```diff
@@ -0,0 +1,171 @@
...
+	error: string | null;
+	deletePaymentMethod: ( id: StoredPaymentMethod[ 'stored_details_id' ] ) => Promise< void >;
+}
+
+export interface WithStoredPaymentMethodsProps {
+	paymentMethodsState: StoredPaymentMethodsState;
+}
+
+export function withStoredPaymentMethods< P >(
+	Component: React.ComponentType< P >,
```
pottedmeat: import React?

pottedmeat (approved):
Code looks good. Tested with no issues.

client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
sirbrillig: Good catch. Updated.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7893200

Hi @sirbrillig, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `Delete the "%s" payment method`
- `There was a problem deleting that payment method.`
- `There was a problem loading your stored payment methods.`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7893200).

Issue #74605, opened 2023-03-18
Plans: formatCurrency should be used in a couple places


jjchrisdiehl:
### Quick summary

The new 2023 plans grid doesn't use `formatCurrency` in the grid header billing info section, this leads to conflicting currency symbol positioning:

![image](https://user-images.githubusercontent.com/16580129/226078489-f5d808ca-2597-4bb0-8e56-83339ec6a239.png)

This can be resolved by using `formatCurrency` rather than `getCurrencyObject`. The latter is useful when you wish to construct a price from its constituent parts and need a bit more control, but that doesn't really appear to be the case in the [code](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plan-features-2023-grid/billing-timeframe.jsx#L25). Using `formatCurrency` will manage the symbol placement for you.

**Resolved:**
<img width="603" alt="image" src="https://user-images.githubusercontent.com/16580129/226078437-197b5b29-f986-4bd9-bccc-ceb2b16f58a7.png">

An additional issue (though similar) is that it appears we hardcode the `US$25,000` text for Enterprise plans. This falls into the sticky issue of using 'US' for currencies in the "US", which is an issue we tried to steer away from.

We can do this dynamically by using `formatCurrency` here as well, since `formatCurrency` automatically checks if the customer is in the US and adds/removes the `US$` in favor of its alternate `$`. This can be useful since it allows US customers to see a plain dollar sign, but, doesn't confuse other locales that use a $ for their currency, such as Argentinian Pesos.

**formatCurrency in action from the USA:**
![image](https://user-images.githubusercontent.com/16580129/226079382-b73b10c2-5143-4bd2-a02e-295bac556756.png)

**formatCurrency in action from Argentina:**
![image](https://user-images.githubusercontent.com/16580129/226079029-5312b79a-bce6-4ed1-abdf-01a0c2f81945.png)

### Steps to reproduce

1. Go to /me and set your interface language to `fr-CA`
2. Go to the Plans page
3. Observe the incorrect formatting, in `fr-CA` the symbol should be right aligned, yet in the billing heading info section it is still left aligned

### What you expected to happen

I expected to see the symbol align left or right depending on the locale that is set, but at this time those specified prices only show as left aligned regardless of locale.

I also expect to not see the `US$` symbol while in the USA, you may need to use a VPN to test this on your own.

### What actually happened

Symbol positioning was not implemented based on locale.

### Impact

All

### Available workarounds?

[...]

PR #74683, opened 2023-03-21
Plans: Implement formatCurrency for header-billing-info section

jjchrisdiehl:
This PR implements formatCurrency in the 2023 plans grid "header-billing-info" section:

image

## Testing Instructions

* Set your locale from `/me` to `fr-CA`
* Go to `/plans/[your-site-url]`
* You should now see the currency symbol right aligned

## Pre-merge Checklist

Related to #74605


client/my-sites/plan-features-2023-grid/billing-timeframe.jsx
```diff
@@ -22,11 +22,10 @@ function getPerMonthDescription( {
 	}
 
 	if ( ! isMonthlyPlan ) {
-		const annualPriceObj = getCurrencyObject( rawPriceAnnual, currencyCode );
-		const annualPriceText = `${ annualPriceObj?.symbol }${ annualPriceObj?.integer }`;
+		const annualPrice = formatCurrency( rawPriceAnnual, currencyCode );
 
-		return translate( 'per month, %(annualPriceText)s billed annually', {
-			args: { annualPriceText },
+		return translate( 'per month, %(annualPrice)s billed annually', {
```
sirbrillig: We can maintain the same translated string here and just change how the variable is generated (otherwise it will require a new translation).

sirbrillig (commented):
This looks great! Thank you! I think it just needs a rebase.

jjchrisdiehl:
I noticed that these strings are not fully translated, for instance while the string works in English and `fr-CA`:

image

The interpolated `fullTermPriceText` does not render in Espanol:

image

sirbrillig:
🤔 could the condition in the translation line, `currencyCode && null !== maybeDiscountedFullTermPrice` be returning false for the spanish version? You should probably add some code here to handle what should be rendered if the price does not exist. I didn't notice before that there's a case where that can happen. That bug was already present, but as for _why_ there's no price or currency, you'll have to debug the code to figure that out.

jjchrisdiehl:
I see that this string is [not translated](https://translate.wordpress.com/projects/wpcom/es/default?filters%5Bterm%5D=billed+every+two+years) for Spanish yet, @sirbrillig could this be related?

sirbrillig:
Maybe. I think that the translation you're looking for is [this one](https://translate.wordpress.com/projects/wpcom/es/default?filters%5Bterm%5D=per+month%2C+%25%28fullTermPriceText%29s+billed+annually&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filters%5Bview%5D=&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc), which looks like it's translated as `al mes, %(annualPriceText)s con facturación anual`. If the price text is an empty string, that could produce the output we see in your screenshot. Maybe since the placeholder is different in the translated version, it's not doing the interpolation correctly? I've never translated anything but that string was fuzzy translated so maybe we just need to replace it with a real translation.

In any case, you can figure out what's going on by trying three things:

1. Log the value of `fullTermPriceText` when you see the poor Spanish translation.
2. Try the page with a different language to see if the same problem exists.
3. View the page in production in Spanish to see if it was caused by your PR or if it's true for everybody.

jjchrisdiehl:
The string in production isn't translated into Spanish at all (yet appears to be translated in other languages like French Canadian). 

I [asked](p1679529086444509-slack-C02AED43D) in i18n if they can update the translation in production.

Otherwise, I attempted to handle the case where fullTermPriceText returns null, and while it returns null it ends up being ignored in favor of `billingTimeFrame` when rendered in the `PlanFeatures2023GridBillingTimeFrame` component. 

https://github.com/Automattic/wp-calypso/blob/f3ab7ff3a73023fffcf6e18f15764ae4e3c11050/client/my-sites/plan-features-2023-grid/billing-timeframe.tsx#L77

It looks like `billingTimeFrame` effectively uses [this translation fallback](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-products/src/plans-list.tsx#L350).

Arguably, this isn't necessarily wrong, all it does is omit the full 12mo price of the plan which isn't necessarily confusing or incorrect.

I think at this point we could merge these changes since we're just focusing on updating the price to use formatCurrency _when it is actually displayed_.

sirbrillig (approved):
Looks good!

PR #74686, opened 2023-03-21
format-currency: Use i18n USD currency symbol if geolocation is US but locale is not en

sirbrillig:
## Proposed Changes

This is a follow-up to https://github.com/Automattic/wp-calypso/pull/74351 which added code to use geolocation to determine the USD currency symbol of the functions of the `@automattic/format-currency` package (eg: `formatCurrency()` and `getCurrencyObject()`). That PR missed something subtle but important: if the currency is USD and the location is `US` and the locale starts with `en`, we want the currency symbol to be `$` **but** if the locale is not `en` we still want to use the default i18n currency symbol. This PR corrects that oversight.

Props to @jjchrisdiehl who noticed this.

Before:

Screenshot 2023-03-21 at 10 53 15 AM

After:

Screenshot 2023-03-21 at 10 53 41 AM

## Testing Instructions

Automated tests should cover it. 

To manually test this:

- Set your currency to USD.
- Geolocate yourself in the US (using a VPN if necessary).
- Set your user interface language to a non-English one with right-aligned currency symbols (eg: `fr`).
- View a page that displays prices (eg: `/plans`).
- Verify that the currency symbol for the prices is `$US`.

### DIFF
--- a/packages/format-currency/src/index.ts
+++ b/packages/format-currency/src/index.ts
@@ -16,6 +16,7 @@ const geolocationEndpointUrl = 'https://public-api.wordpress.com/geo/';
 export function createFormatter(): CurrencyFormatter {
 	const currencyOverrides: Record< string, { symbol?: string | undefined } > = {};
 	let defaultLocale: string | undefined = undefined;
+	let geoLocation = '';
 
 	// If the user is inside the US using USD, they should only see `$` and not `US$`.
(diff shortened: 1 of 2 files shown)


jjchrisdiehl:
Looks good to me, 

Cur: USD,  Locale: English, Location: Washington

image

Cur: USD,  Locale: French Canadian, Location: Washington

image

PR #74697, opened 2023-03-21
Newsletter: Hover effect on color dropdown

### LINKED ISSUE #74661: Newsletter: Add hover state to colour dropdown
The colour selection dropdown does not have a hover state. Here’s the comparison with the Calypso dropdown

image

escapemanuele:
## What
image
 Color dropdown was missing an hover effect, now it has it.

## Testing
1. Calypso Live link
2. Check `setup/newsletter/newsletterSetup` page
3. Verify that the hover effect is there

Fixes https://github.com/Automattic/wp-calypso/issues/74661


renancarvalho (approved):
Pretty! 🚢

PR #74718, opened 2023-03-21
Create client-side data interface for updating subscription settings.

### LINKED ISSUE #74423: `/subscriptions/settings`: Create client-side data interface for updating subscription settings.
Since this is going to a new API endpoint I would encourage using react-query for integration (see related conversation: p1678807696397609-slack-C7YPUHBB2).

The call to the new API should be integrated into [@automattic/data-stores](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores).

TimBroddin:
Fixes #74423

## Proposed Changes

This PR adds the `useSubscriptionManagerUserSettingsMutation`, which can be used to update the email settings.
It also improves upon the typings of the React Query hooks.

## Testing Instructions

1. Apply this PR
2. Since this PR does only include the hook, you'll have to add it to a component to test. You could add an empty component, but it's quicker to just temporary add it to an existing component such as `SubscriptionManagerContainer.tsx`

- Add the needed imports:
```js
import { useEffect } from "@wordpress/element";
import { Reader } from '@automattic/data-stores';
```

- Add the test code inside the component:
```js
	const { mutate, data, error } = Reader.useSubscriptionManagerUserSettingsMutation();

	useEffect(() => {
		setTimeout(() => {
			mutate({
				mail_option: 'html',
				delivery_day: 0, // 0-6, 0 is Sunday
				delivery_hour: 17, // 0-23, 0 is midnight
				blocked: false,
			});
		}, 10000);

	}, []);

	useEffect(() => {
		console.log(data, error);
	}, [data, error]);
```
This will initiate the mutation, call it once and print data/errors.

⁉️ Why the setTimeout? We only know that we're logged in after a few seconds (because the User datastore has to fetch our user first). In a real world scenario, we wouldn't call the mutation before the data is loaded, but here we have to simulate it.

3. Test as a WP.com user:

- Visit http://calypso.localhost:3000/subscriptions
- After 10 seconds you should see the request going out in the Network panel of devtools, and you should see the data coming in.

| | |
|-|-|
| CleanShot 2023-03-21 at 15 10 21@2x | CleanShot 2023-03-21 at 15 10 34@2x |

4. Test as an external user

- Open an incognito window
- Set a subkey cookie and reload (paste this in your console: `document.cookie = 'subkey=8ee6a7e013b168ad8750dfeaac078298.jefpober%2Bxt%40gmail.com'; window.location.reload();`)
- After 10 seconds you should see the request going out in the Network panel of devtools, and you should see the data coming in.
- Verify that the subkey is being set as well

| | |
|-|-|
| CleanShot 2023-03-21 at 15 13 04@2x | CleanShot 2023-03-21 at 15 13 26@2x |

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/reader/helpers/index.ts
+++ b/packages/data-stores/src/reader/helpers/index.ts
@@ -32,6 +32,10 @@ async function callApi< ReturnType >( {
 		?.find( ( c ) => c.startsWith( 'subkey=' ) )
 		?.split( '=' )[ 1 ];
 
+	if ( ! subkey ) {
+		throw new Error( 'Subkey not found' );
+	}
+
 	return apiFetch( {
 		global: true,
 		path: `https://public-api.wordpress.com/rest/v1.1${ path }`,
--- a/packages/data-stores/src/reader/helpers/index.ts
+++ b/packages/data-stores/src/reader/helpers/index.ts
@@ -1,19 +1,20 @@
 import apiFetch, { APIFetchOptions } from '@wordpress/api-fetch';
 import wpcomRequest from 'wpcom-proxy-request';
 
-type FetchFromApiParams = {
+type callApiParams = {
 	path: string;
 	method?: 'GET' | 'POST';
 	body?: object;
 	isLoggedIn?: boolean;
 };
+
 // Helper function for fetching from subkey authenticated API. Subkey authentication process is only applied in case of logged-out users.
-async function fetchFromApi< ReturnType >( {
+async function callApi< ReturnType >( {
 	path,
 	method = 'GET',
-	body = {},
+	body,
 	isLoggedIn = false,
-}: FetchFromApiParams ): Promise< ReturnType > {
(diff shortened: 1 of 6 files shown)


JuanLucha (approved):
It works as expected. I changed the params to non-valid ones, like `delivery_hour: 40`, and the error is correct: `invalid_input : [ "The delivery_hour value provided is invalid." ]`

ivan-ottinger (approved):
Looks great and works as expected (for both WPCOM and non-WPCOM authenticated user). 👍🏼👍🏼

- Request headers related the non-WPCOM user include the correct subkey authorization
- I did not notice any JS/TS errors nor warnings

TimBroddin:
> It works as expected. I changed the params to non-valid ones, like `delivery_hour: 40`, and the error is correct: `invalid_input : [ "The delivery_hour value provided is invalid." ]`

Thanks for testing that as well!

PR #74720, opened 2023-03-21
data: always use arrays for `queryKey` in `use(Query|Mutation)`

flootr:
Related to #74381

This PR is part of the effort to upgrade `react-query` to the next version. `queryKey` now always has to be an array and doesn't accept a string anymore.

## Proposed Changes

* Convert all string `queryKey`s to arrays.

## Testing Instructions

* This should have no effect on the production build but smoke-testing is probably a good idea
* All checks should be green

## Pre-merge Checklist


client/components/language-picker/site-language-picker.jsx
```diff
@@ -35,7 +35,7 @@ const SiteLanguagePicker = ( { languages: origLanguages, ...restProps } ) => {
 		error,
 		isLoading,
 	} = useQuery(
-		'wporg-translations-' + wpVersion,
+		[ 'wporg-translations-' + wpVersion ],
```
tyxla: Noting that now when we always use an array for query keys, we could have the library actually take care of [hashing the key](https://tanstack.com/query/v4/docs/react/guides/query-keys#query-keys-are-hashed-deterministically). Meaning, we could alter this to:

```suggestion
		[ 'wporg-translations', wpVersion ],
```

That applies to all query keys that we're currently manually hashing.

client/data/marketplace/utils.ts
```diff
@@ -18,5 +18,5 @@ export const getPluginsListKey = (
 		options.tag && ! options.searchTerm ? options.tag : '',
 	];
 
-	return [ PLUGINS_CACHE_VERSION.toString(), key, ...keyParams ];
+	return [ PLUGINS_CACHE_VERSION.toString(), ...key, ...keyParams ];
```
tyxla: I know this was here before, but I wonder if the `key` part should be at the beginning and the cache version would be afterwards.

tyxla (approved):
LGTM 👍 

I'd personally suggest for query keys with a dynamic part that the dynamic part would be a separate array item, but that's optional and could be followed up with in another PR.

Thanks 🚀

client/data/geo/use-geolocation-query.js
```diff
@@ -1,7 +1,7 @@
 import { useQuery } from 'react-query';
 
 export const useGeoLocationQuery = () =>
-	useQuery( 'geo', () =>
+	useQuery( [ 'geo' ], () =>
```
sirbrillig: I'll take care of this in https://github.com/Automattic/wp-calypso/pull/75401

client/components/language-picker/site-language-picker.jsx
flootr: Agreed 👍🏻 I'll add a commit which takes care of these.

client/data/marketplace/utils.ts
flootr: Good point, I'll double-check and add a commit.

tyxla: Sounds good. Can be done in another PR if you don't want to clutter this one.

client/components/language-picker/site-language-picker.jsx
tyxla: That totally makes sense to be another PR, FWIW.

flootr: Added here https://github.com/Automattic/wp-calypso/pull/74720/commits/01d47ac5b2b5d68cca4ae4f05437774fd94779b4

client/data/marketplace/utils.ts
flootr: See https://github.com/Automattic/wp-calypso/pull/74720/commits/d8ddad71b60b687e64b1dc92f1d7f2ecdabc30c0

tyxla (approved):
LGTM, thanks 🚀

PR #74721, opened 2023-03-21
User Invites: Wrap long domain on accept screen

### LINKED ISSUE #74359: User Invites: Accept screen doesn't properly wrap long domain
image

From p1677684822767929-slack-C0347E545HR

**Done is:**
* We either wrap or clip the domain, depending on what's done elsewhere in the app.

katinthehatsite:
* Close https://github.com/Automattic/wp-calypso/issues/74359

## Proposed Changes

This PR fixes the display of the long domain on accept screen on user invites.

## Testing Instructions

* Create a site with a very long domain (35-40 characters, does not have to be a custom domain name)
* Make sure that the `Site Title` field under `/settings/general/{site}` is empty
* Invite an administrator to your site from `/people/team/{site}` by clicking `Add a team member`
* Click an email link for the administrator invitation (you can copy a part of the link that starts with `accept...` from the email and append it to Calypso Live so it would be something like this: `https://container-competent-dewdney.calypso.live/accept-invite/216949641/98f5d9ffaacac36d2e0ed9e34949704c_r7Mm1RbNXk5BjJ07F3OZY6Fp`)
* Check if the domain name is going on a new line:

Screenshot 2023-03-21 at 2 52 51 PM

* Switch to mobile view (you can use Chrome dev tools) and check if the domain fits in the container:

Screenshot 2023-03-21 at 3 15 16 PM

* You can also invite yourself as a viewer (follow the same steps as above by choose "Viewer" instead of "Administrator) and check the domain:

Screenshot 2023-03-21 at 2 56 30 PM

**Other Notes**

* This is a bit of an edge case because as long as the user has `Site Title` set under `/settings/general/{site}` , it will be the site title that appears instead of the domain
* Another option that we can try is to set `overflow: hidden` which will cut off the part of the domain that does not fit in the box. This is what we already do on the site switcher:

Screenshot 2023-03-21 at 3 19 40 PM

* This is what it would look like on the user invite:

Screenshot 2023-03-21 at 2 57 08 PM

* This would be the view from mobile:

Screenshot 2023-03-21 at 3 22 34 PM

My main reasoning behind leaving a full domain would be that the user might want to see the exact site to which they accept invitations so I considered it to be important information. However, this is open to feedback!

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/invites/invite-form-header/style.scss
+++ b/client/my-sites/invites/invite-form-header/style.scss
@@ -4,6 +4,7 @@
 	font-size: $font-title-small;
 	line-height: 24px;
 	margin-bottom: 16px;
+	word-break: break-word;
 }
 
 .invite-form-header__explanation {
@@ -12,4 +13,5 @@
 	font-size: $font-body-small;
 	line-height: 21px;
 	margin-bottom: 16px;
+	word-break: break-word;
 }


wojtekn (approved):
The code looks good, it fixes the issue.

I think `break-word` is better here than overflow, because that way user has full information about the site he is going to accept invitations for.

sejas (approved):
Nice!,
Good catch also with the explanation 👌 .
Works well for me.


Screenshot 2023-03-22 at 09 49 12

PR #74725, opened 2023-03-21
 Redirect from cart to marketplace thank you page

### LINKED ISSUE #74295: Marketplace Themes: Cart redirect to Marketplace Thank you page
Marketplace Themes, after the purchase, should be redirected to the Marketplace Thank you page. See pekYwv-Pz-p2.
The steps to achieve this are: 
 - [ ] Remove the `redirect_to` parameter from the cart URL for Marketplace themes.
 - [ ] Make Marketplace themes enter in the following condition (see https://github.com/Automattic/wp-calypso/issues/73838#issuecomment-1458339666): 

 
https://github.com/Automattic/wp-calypso/blob/3d1006d1780d61012d7a1f1572abd6660cbb6a7d/client/my-sites/checkout/get-thank-you-page-url/index.ts#L526-L535

gcsecsey:
Closes #74295

## Proposed Changes

- Remove the `redirect_to` param from the cart URL for marketplace themes
- update the cart fallback URL logic to get theme and plugin slugs from the `extra` object

#### With an Atomic site
- Put a marketplace theme and a marketplace product into the cart
- Complete checkout
- Check that after checkout, the site redirects to the marketplace thank you page
- Check that in the URL, both the theme and plugin slugs are passed like `?plugins=slug1&themes=slug2`

CleanShot 2023-03-21 at 15 52 29@2x

#### With a non-atomic site
- Put a marketplace theme in the cart
- In checkout, check that both the theme and a plan is added to the cart
- Complete checkout
- Check that the site redirects back to the theme page and the CTA is fetching the Atomic transfer status as implemented in https://github.com/Automattic/wp-calypso/pull/72837

CleanShot 2023-03-21 at 16 00 03@2x

## Pre-merge Checklist


packages/shopping-cart/src/types.ts
```diff
@@ -564,7 +564,8 @@ export interface ResponseCartProductExtra {
 
 	// Marketplace properties
 	is_marketplace_product?: boolean;
-	plugin_slug?: boolean;
+	product_slug?: boolean;
+	product_type?: string;
```
sirbrillig: Could we add a comment above this to explain what `product_type` and `product_slug` are for and when they're likely to be set? I know there's a lot of uncommented properties here but I'm trying to add comments for new ones to help us in the future.

packages/shopping-cart/src/types.ts
```diff
@@ -564,7 +564,8 @@ export interface ResponseCartProductExtra {
 
 	// Marketplace properties
 	is_marketplace_product?: boolean;
-	plugin_slug?: boolean;
+	product_slug?: boolean;
```
WBerredo:
```suggestion
	product_slug?: string;
```

This should be string, right? I think the current version is wrong.

client/my-sites/checkout/get-thank-you-page-url/index.ts
```diff
@@ -533,14 +533,23 @@ function getFallbackDestination( {
...
+		cart?.products?.filter( ( product ) => product?.extra?.is_marketplace_product ) || [];
 
-	if ( pluginSlugs.length > 0 ) {
+	const marketplacePluginSlugs = marketplaceProducts
+		?.filter( ( { extra } ) => extra.product_type === 'marketplace_plugin' )
+		.map( ( { extra } ) => extra.product_slug );
+
+	const marketplaceThemeSlugs = marketplaceProducts
+		?.filter( ( { extra } ) => extra.product_type === 'marketplace_theme' )
+		.map( ( { extra } ) => extra.product_slug );
```
WBerredo:
```suggestion
	const marketplacePluginSlugs = marketplaceProducts
		.filter( ( { extra } ) => extra.product_type === 'marketplace_plugin' )
		.map( ( { extra } ) => extra.product_slug );

	const marketplaceThemeSlugs = marketplaceProducts
		.filter( ( { extra } ) => extra.product_type === 'marketplace_theme' )
		.map( ( { extra } ) => extra.product_slug );
```

Nit: As `marketplaceProducts` will always be an array (maybe empty but still) I think you can remove the conditional check (`?`)

WBerredo (commented):
For the non-atomic case, I think we must turn the site into atomic on the Thank You page as we do for plugins. (I just saw your comment on the ticket later, we said the same thing 😀)

WBerredo:
I added the `DO NOT MERGE` tag as this can only be merged when all the work on the Thank You page is done to support Themes.

packages/shopping-cart/src/types.ts
gcsecsey: Ooops, yes it should, I didn't notice that one!

client/my-sites/checkout/get-thank-you-page-url/index.ts
gcsecsey: Thanks for pointing this out, I was planning to remove this at one point but received a TS error, went back to fix it, then forgot to remove the optional chaining.

packages/shopping-cart/src/types.ts
gcsecsey: Thanks @sirbrillig, it's a good idea to make new properties easier to understand. I've updated the types to make them more representative and added comments to them.

gcsecsey:
> For the non-atomic case, I think we must turn the site into atomic on the Thank You page as we do for plugins. (I just saw your comment on the ticket later, we said the same thing 😀)

Thanks for the confirmation @WBerredo. In that case, I'll remove the `redirect` parameter from the checkout entirely and update the test steps.

gcsecsey:
I've removed the parameter, now all sites are redirected to the new thank you page.

epeicher (approved):
I have tested it, and it works as described in the comments. Both Atomic and non-Atomic sites are redirected to the Thank You page. LGTM!

PR #74729, opened 2023-03-21
Purchases: Add AU to countries supporting VAT

sirbrillig:
## Proposed Changes

This PR adds Australia to the list of countries for which we support collecting VAT info.

Part of https://github.com/Automattic/payments-shilling/issues/1475

## Testing Instructions

- Visit `/me/purchases/vat-details` and verify that `AU` is listed in the list of countries.
- Click to edit the billing details step of checkout if it is not already active.
- Select "Australia" from the list of countries.
- Verify that you see a "Add Business Tax ID details" checkbox.

### DIFF
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -166,6 +166,7 @@ function CountryCodeInput( {
 } ) {
 	const countries = [
 		'AT',
+		'AU',
 		'BE',
 		'BG',
 		'CH',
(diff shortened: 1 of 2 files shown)


michaeldcain (approved):
This LGTM. Since we can use the fallback Taxamo validator, there's no need to wait on merging.

michaeldcain:
Fixes: https://github.com/Automattic/payments-shilling/issues/1490

PR #74733, opened 2023-03-21
Subscription Manager: Add an entry point for the external subscription-manager portal

### LINKED ISSUE #74732: `/subscriptions`: Add a dedicated entry point for the external subscription management portal in calypso

mashikag:
Closes #74732

## Proposed Changes

* Create a dedicated webpack entry point for the external subscription management portal. 
* Server the `landing/subscriptions` bundle for `/subscriptions` path. 

## Testing Instructions

* Go to http://calypso.localhost:3000/subscriptions/settings

## Pre-merge Checklist


TimBroddin:
My review is in progress, but while doing so I found some errors when trying to build the module from scratch:

```bash
packages/subscription-manager/src/app/app.tsx:2:33 - error TS2307: Cannot find module '@automattic/subscription-manager' or its corresponding type declarations.

2 import SubscriptionManager from '@automattic/subscription-manager';
```

Fixing this gave a few other issues. I'll fix them and make a commit.

packages/subscription-manager/src/app/app.tsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable no-restricted-imports */
+import '@automattic/calypso-polyfills';
+import { Reader } from '@automattic/data-stores';
+import { useLocale } from '@automattic/i18n-utils';
+import { UniversalNavbarHeader } from '@automattic/wpcom-template-parts';
```
noahtallen: Essentially, I'd expect types for this component to work out of the box. (See below)

packages/subscription-manager/src/types.d.ts
```diff
@@ -0,0 +1,13 @@
...
+declare module '@automattic/wpcom-template-parts' {
+	export interface HeaderProps {
+		className?: string;
+		isLoggedIn: boolean;
+		sectionName?: string;
+		logoColor?: string;
+		variant?: 'default' | 'minimal';
+	}
+
+	export function UniversalNavbarHeader( props: HeaderProps ): JSX.Element;
```
noahtallen: This module declaration definitely shouldn't be needed 🤔 

My reasoning is that both subscription-manager and wpcom-template-parts are TS packages we control, so we should be able to get type interoperability without having to override things manually. Plus, we want to avoid duplicating types, because that makes future updates/refactors harder!

If there's an issue with internal type resolution like this, it normally indicates a deeper issue. If I had to guess, it's because `packages/subscription-manager/package.json` should declare every dependency it uses, even internal ones. That lets the compiler + package manager resolve all the types correctly :) So that probably needs to reference wpcom-template-parts as a yarn workspace dependency

On top of that, I think generally we try to avoid having packages depend on the Calypso application itself -- in other words, we probably wouldn't want to be declaring dependencies on calypso/state. (The calypso app is a really big dependency to have!) We try to have things go the other way, where the calypso app depends on packages, but not the other way around.

If I remember correctly, other standalone entrypoints are implemented more directly under client/. They can rely on packages, but packages should be designed to be more standalone. So you might put a component or data store in packages, but you might not put a full app/section definition there. Otherwise, it kinda defeats the purpose of packages!

In other words, I would probably refactor this so that any _internal calypso dependencies_ are only referenced from `client/landing/subscriptions`. You could then pass props to a component in `packages/subscription-manager`, and then the package itself would only depend on other packages.

@Automattic/team-calypso-frameworks might be able to provide even more context here :)

client/components/calypso-i18n-provider/index.tsx
```diff
@@ -6,7 +6,7 @@ import { useState, useEffect } from 'react';
 import type { I18N } from 'i18n-calypso';
 import type { FunctionComponent } from 'react';
 
-const CalypsoI18nProvider: FunctionComponent< { i18n: I18N } > = ( {
+const CalypsoI18nProvider: FunctionComponent< { i18n?: I18N } > = ( {
```
tyxla, line 9: How is this component expected to work with an undefined `i18n` property? This seems problematic. Perhaps we should just not use the provider at all in that case?

client/landing/subscriptions/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { SubscriptionManagerApp } from '@automattic/subscription-manager';
+import ReactDom from 'react-dom';
+import { initializeCurrentUser } from 'calypso/lib/user/shared-utils';
+
+declare const window: Window & {
+	AppBoot?: () => void;
+};
+
+window.AppBoot = async () => {
+	const user = await initializeCurrentUser();
```
tyxla: Are we planning to use the bootstrapped user here on production?

packages/data-stores/src/reader/context/index.tsx
```diff
@@ -0,0 +1,30 @@
+import { createContext } from '@wordpress/element';
+import type { ReactNode } from 'react';
+
+export type UserType = object | null;
```
tyxla: These types are a bit vague. Should we use the existing `UserData` type somehow?

packages/subscription-manager/src/components/MainContent/MainContent.tsx
tyxla, line 1: It's worth mentioning that ideally, we should be following the kebab case for file name structure like the rest of Calypso.

packages/subscription-manager/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,20 @@
...
+import './styles.scss';
+
+type TabsProps = {
+	children: ReactNode;
+};
+
+const Tabs = ( { children }: TabsProps ) => (
+	<SectionNav className="subscription-manager-tabs">
+		<SectionNavTabs>{ children }</SectionNavTabs>
+	</SectionNav>
```
tyxla: This seems to be a straightforward wrapper that doesn't add anything but an extra class name. Seems like it could be avoided in favor of using the original components, no?

packages/subscription-manager/src/components/index.d.ts
```diff
@@ -113,3 +114,20 @@ declare module 'calypso/components/forms/form-label' {
 declare module 'calypso/components/localized-moment' {
 	export const useLocalizedMoment: () => import('moment');
 }
+
+declare module 'calypso/components/localized-moment/provider' {
+	const MomentProvider: FC< { currentLocale: string } >;
+	export default MomentProvider;
+}
+
+declare module 'calypso/state' {
```
tyxla: From my understanding, it doesn't make much sense for a package to depend on the client. It should be the other way around. Therefore, it shouldn't make sense for this package to declare modules that live in the `calypso` client.

Furthermore, declaring them here could create problems and/or conflicts with other packages or the client one that also declare them or rely on them.

client/components/calypso-i18n-provider/index.tsx
mashikag: > How is this component expected to work with an undefined i18n property? 

A default value is provided, see one line below. From what I understand, this is exactly what we want to pass in.

packages/subscription-manager/src/types.d.ts
mashikag: > If there's an issue with internal type resolution like this, it normally indicates a deeper issue. If I had to guess, it's because packages/subscription-manager/package.json should declare every dependency it uses, even internal ones. That lets the compiler + package manager resolve all the types correctly :) So that probably needs to reference wpcom-template-parts as a yarn workspace dependency

Spot on. Thanks for noticing this one. 🙇‍♂️

mashikag: Thank you so much for the in depth feedback. It makes me now better understand the architectural vision behind calypso. I will now take a step back, I will mark the PR as draft, and I will complete the task keeping the points you made in mind. Once done, I will re-activate this PR.  🙇‍♂️

packages/subscription-manager/src/components/MainContent/MainContent.tsx
mashikag: Ops, this is something I did not really pay attention to up until now. I am very used to using pascal case names for such files. I will have to do a cleanup PR to address my previously created files.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
tyxla, line 1: Nit: just wanted to note that we use kebab case both for file names and dir names. There seem to be some pre-existing files here that we could consider updating.

client/landing/subscriptions/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const initialState = getInitialState( initialReducer, user.ID );
+	const reduxStore = createReduxStore( initialState, initialReducer );
+	setStore( reduxStore, getStateFromCache( user.ID ) );
+	setupLocale( user, reduxStore );
+	if ( user ) {
+		// Initialize calypso user store
+		config.isEnabled( 'signup/inline-help' ) &&
+			reduxStore.dispatch( requestHappychatEligibility() as unknown as AnyAction );
+		reduxStore.dispatch( setCurrentUser( user ) as AnyAction );
+		reduxStore.dispatch( requestSites() as unknown as AnyAction );
```
tyxla: Same as above in terms of reusing code.

client/landing/subscriptions/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+	site_id_label: 'wpcom',
+	client: config( 'client_slug' ),
+} );
+
+const setupQueryClient = async ( userId: number ) => {
+	await loadPersistedState();
+	const queryClient = createQueryClient();
+	await hydrateBrowserState( queryClient, userId );
+	return queryClient;
+};
```
tyxla: Looks like we might be doing the same that we are doing in the stepper entry point. Would be great to have it managed in a single place and not duplicated.

client/landing/subscriptions/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { requestSites } from 'calypso/state/sites/actions';
+import SubscriptionManagemer from './subscription-manager/subscription-manager';
+import './styles.scss';
+
+const tracksSuperProps = () => ( {
+	environment: process.env.NODE_ENV,
+	environment_id: config( 'env_id' ),
+	site_id_label: 'wpcom',
+	client: config( 'client_slug' ),
+} );
```
tyxla: We have this in the stepper entry point, count be a good opportunity to reuse.

client/landing/subscriptions/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const user = ( await initializeCurrentUser() ) as unknown as CurrentUser;
+	const queryClient = await setupQueryClient( user.ID );
+	const reduxStore = setupReduxStore( user );
+	initializeAnalytics( user, tracksSuperProps );
+
+	ReactDom.render(
+		<CalypsoI18nProvider>
+			<Provider store={ reduxStore }>
+				<QueryClientProvider client={ queryClient }>
+					<MomentProvider>
```
tyxla, line 57: Do we really need all these providers? 

Are we going to use `moment` for example?

client/landing/subscriptions/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+	font-family: $sans;
+	text-rendering: optimizelegibility;
+	-moz-osx-font-smoothing: grayscale;
+	-webkit-font-smoothing: antialiased;
+
+	// Basic stylings
+	margin: 0;
+	line-height: 1.5;
+
+	h1,
```
tyxla, line 22: Headings and paragraphs in a button? Is that intentional?

client/landing/subscriptions/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+		margin: 0;
+	}
+
+	a,
+	a:visited {
+		text-decoration: none;
+		color: var(--color-link);
+	}
+
+	button {
```
tyxla, line 37: Button inside a button? 😅

client/landing/subscriptions/styles.scss
```diff
@@ -0,0 +1,49 @@
+/*
```
tyxla, line 1: This stylesheet duplicates a lot of what is in the stepper global stylesheet. Can we abstract and reuse instead of duplicating?

client/landing/subscriptions/subscription-manager/subscription-manager.tsx
```diff
@@ -0,0 +1,42 @@
...
+const SubscriptionManagementPage = () => {
+	const translate = useTranslate();
+	const { data: counts } = Reader.useSubscriptionManagerSubscriptionsCountQuery();
+
+	return (
+		<SubscriptionManager>
+			<SubscriptionManager.TabsSwitcher
+				baseRoute="subscriptions"
+				defaultTab="sites"
+				tabs={ [
```
tyxla: I haven't checked thoroughly yet, but I assume these might trigger updates on every re-render since they create a new array with new objects every time. We could put them in a hook and/or memoize them.

client/landing/subscriptions/subscription-manager/subscription-manager.tsx
```diff
@@ -0,0 +1,42 @@
...
+const SitesView = () => <span>Sites View</span>;
+const CommentsView = () => <span>Comments View</span>;
+
+const SubscriptionManagementPage = () => {
+	const translate = useTranslate();
+	const { data: counts } = Reader.useSubscriptionManagerSubscriptionsCountQuery();
+
+	return (
+		<SubscriptionManager>
+			<SubscriptionManager.TabsSwitcher
```
tyxla, line 38: Is the namespacing necessary? Can we just import the module names?

packages/subscription-manager/package.json
```diff
@@ -28,13 +28,23 @@
 		"prepack": "yarn run clean && yarn run build",
 		"watch": "tsc --build ./tsconfig.json --watch"
 	},
+	"dependencies": {
+		"@automattic/data-stores": "workspace:^",
+		"@automattic/typography": "workspace:^",
+		"@automattic/wpcom-template-parts": "workspace:^",
+		"i18n-calypso": "workspace:^",
+		"react": "^17.0.2",
+		"react-dom": "^17.0.2",
```
tyxla: What do we need `react-dom` for?

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -16,18 +17,25 @@ const SubscriptionManagerContainer = ( { children }: SubscriptionManagerContaine
...
+				variant="minimal"
+				isLoggedIn={ false }
 			/>
-			{ children }
-		</Main>
+			<Main className="subscription-manager-container">
+				<DocumentHead title="Subscriptions" />
+				<FormattedHeader
+					brandFont
+					headerText={ translate( 'Subscription management' ) }
```
tyxla, line 54: Are these strings finalized? They'll be sent for translation, so if we don't want to translate them just yet because they aren't final, we can set `translate()` to be an identity `( x ) => x` function while they are being finalized.

client/landing/subscriptions/index.tsx
mashikag: #### `CalypsoI18nProvider`
- Needed for translations
#### `Provider`
- Needed for storing current user data, also unfortunately many client components rely on a provider being present
#### `QueryClientProvider`
- We use `react-query` for querying and mutating subscriptions and user related data
#### `MomentProvider`
- We use it for email delivery window input (see [here](https://github.com/Automattic/wp-calypso/blob/3bd82489247e2e6e481933719604092970b5101d/packages/subscription-manager/src/components/fields/DeliveryWindowInput/DeliveryWindowInput.tsx#L28))

client/landing/subscriptions/subscription-manager/subscription-manager.tsx
mashikag: Good catch, thanks. 🙇‍♂️

packages/subscription-manager/src/app/app.tsx
mashikag: I added the to [package.json](https://github.com/Automattic/wp-calypso/pull/74733/files#diff-93159aa5e50501d8d23636b4f196e96c2d6dd371cff4a4775c7eb465b3bf133eR43), but building for unit tests seems to still fail the type check for `UniversalNavbarHeader` (see [here](https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Run_All_Unit_Tests/9716616?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=9716616_6999_135.596.6996&logFilter=debug&logView=flowAware)). I am trying to find out what may be causing the issue.

packages/subscription-manager/package.json
mashikag: You are right, we do not need. Thanks for catching this out. 🙇‍♂️

packages/subscription-manager/src/app/app.tsx
mashikag: > but building for unit tests seems to still fail the type check for UniversalNavbarHeader (see [here](https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Run_All_Unit_Tests/9716616?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=9716616_6999_135.596.6996&logFilter=debug&logView=flowAware)).

I got it working. I was missing [the entry](https://github.com/Automattic/wp-calypso/pull/74733/files#diff-f4d0aca639aee846827a3f6b275ab6903ef17fa222680329a652ef7375ce4bdfR10) in tsconfig.

client/landing/subscriptions/index.tsx
mashikag: Yes, but only for checking whether the current user is a logged-in user. 

However, I think that ideally we would like to have a check in our express server before we decide to load the entry point. Then, in case we are dealing with a logged-in user, we would redirect them to the reader part of calypso. This subscription-manager portal is meant only for external (not logged-in) users.

mashikag: However, I would like to address this in a follow-up PR, if you do not mind. There is ongoing work on the subscription manager parallel to my PR. The sooner we merge it, the fewer conflicts will be caused. I will create a task on our board for every issue that we discuss on this PR, and that is not completed as part of this PR. I hope you are OK with this @tyxla . 🙂

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
mashikag: > Are these strings finalized?

Yes, the following are the designs: inDLaEQV8jJ21O4WXawIsJ-fi-712%3A13475

>  if we don't want to translate them just yet because they aren't final, we can set translate() to be an identity ( x ) => x function while they are being finalized.

Thank you, @tyxla ! This is something I did not really think about, but this is a great solution for such case.  I will keep this in mind for future development in Calypso.

client/landing/subscriptions/subscription-manager/subscription-manager.tsx
mashikag: Now, after all your recent feedback, that I understand the architectural idea of packages in calypso better, I realize it was not a good idea to create this subscription-manager package in the first place. Mostly due to the high dependency on the client components. Following on this, creating the `SubscriptionManager` as compound component does not make much sense anymore too.

mashikag: However, I believe it would be better to address this in a separate PR.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
mashikag: > There seem to be some pre-existing files here that we could consider updating.

Yes, definitely. However, I will create another PR for this. Just not to introduce any more noise into this one, but either way, thank you for keeping an eye on this. 🙇‍♂️

client/landing/subscriptions/index.tsx
mashikag: I created the following task for the above: https://github.com/Automattic/wp-calypso/issues/74886

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
mashikag: I created the following task for the above: https://github.com/Automattic/wp-calypso/issues/74884

client/landing/subscriptions/subscription-manager/subscription-manager.tsx
mashikag: This will most likely be addressed as part of the: https://github.com/Automattic/wp-calypso/issues/74887

client/landing/subscriptions/index.tsx
mashikag: This code has changed since the comment was made. We no longer use 3 out of those 4 lines.

client/landing/subscriptions/index.tsx
mashikag: Resolved with [this commit](https://github.com/Automattic/wp-calypso/pull/74733/commits/6ed46ec6c087abff433f057234d9b97383262f34).

client/landing/subscriptions/index.tsx
mashikag: Resolved with the following commits:
1.  [Resolve code duplication for creating tracks super props getter, and for creating query client with applying persistence](https://github.com/Automattic/wp-calypso/pull/74733/commits/6ed46ec6c087abff433f057234d9b97383262f34)
2.  [Include hydrateBrowserState as part of createQueryClient](https://github.com/Automattic/wp-calypso/pull/74733/commits/deb9c97e45c865f253e6945ca812b18e19c16e45)

client/landing/subscriptions/styles.scss
mashikag: I do not see a suitable place where I could put this code in at the moment. I do not think that it would be a good idea for me to place it in either the `landing/stepper`, or `landing/subscription-manager`, and then to import by the other one, because it would be too easy for someone not to realize the stylesheet may be used by another entry point when making changes to it. 

If you do not mind, I will leave those 20 lines of duplicated `.scss` for the time being of this PR, and I will follow up with another PR to resolve this duplication. I am really eager to get this one in, because the parallel ongoing work keeps on creating conflicts.

mashikag: I created the following task for the above: https://github.com/Automattic/wp-calypso/issues/74888

client/landing/subscriptions/styles.scss
mashikag: I only noticed this comment now. Yes, you are totally right, it looks like a mistake. It was a copy & paste for me. I will fix it now.

client/boot/common.js
```diff
@@ -422,10 +420,8 @@ function renderLayout( reduxStore, reactQueryClient ) {
 const boot = async ( currentUser, registerRoutes ) => {
 	saveOauthFlags();
 	utils();
-	await loadPersistedState();
```
TimBroddin, line 425: You probably have a good reason to do so, but why are you removing the loading/persisting of the state of react-query?

client/boot/common.js
```diff
@@ -422,10 +420,8 @@ function renderLayout( reduxStore, reactQueryClient ) {
...
-	const queryClient = createQueryClient();
 
-	await hydrateBrowserState( queryClient, currentUser?.ID );
+	const queryClient = await createQueryClient( currentUser?.ID );
 	const initialQueryState = getInitialQueryState();
```
yansern, line 425: Context: https://github.com/Automattic/wp-calypso/pull/74733#discussion_r1146562136

client/landing/stepper/index.tsx
```diff
@@ -114,14 +104,12 @@ window.AppBoot = async () => {
...
-	await loadPersistedState();
 	const user = ( await initializeCurrentUser() ) as unknown;
 	const userId = ( user as CurrentUser ).ID;
 
-	const queryClient = createQueryClient();
-	await hydrateBrowserState( queryClient, userId );
+	const queryClient = await createQueryClient( userId );
 
-	initializeAnalytics( user, generateGetSuperProps() );
+	initializeAnalytics( user, getGenericSuperPropsGetter( config ) );
```
yansern, line 112: Note to self: `generateGetSuperProps()` was moved to `getGenericSuperPropsGetter()` under `@automattic/calypso-analytics` in this same PR.

TimBroddin:
I did a code review and tested the changes locally. All looks good to me.

One question, though. Did you test the additional packages/landing pages you've touched? Stepper and Calypso analytics come to mind.

yansern (approved):
Let's get this in to unblock work. It doesn't break existing Calypso quick smoke test. Tests are passing too. Any further changes / fixes can be a follow-up PRs.

yansern:
> Did you test the additional packages/landing pages you've touched? Stepper and Calypso analytics come to mind.

Tested Stepper in local instance, the flow works. Also, stepper has E2E tests written for it and seems to be passing.

A new function was introduced & exported from Calypso analytics. So that's fine.

TimBroddin (approved):

tyxla:
Follow-ups sound good to me 👍  Feel free to ping me and the rest of the team if you need anything. 

Note that we're at a meetup this week, so we might be a bit slower to respond than usual.

PR #74734, opened 2023-03-21
AT E2E: fixes to editor and blocks specs.

worldomonation:
Related to #

## Proposed Changes

This PR fixes the permanent failure encountered in the `Blocks: Core` spec by narrowing the selector.

Key changes:
- match only on visible selectors.
- skip the use of block inserter sidebar search field for faster test execution.

## Testing Instructions

  - [ ] Gutenberg E2E (desktop)
  - [ ] Gutenberg E2E (mobile)
  - [ ] Gutenberg E2E AT (desktop)
  - [ ] Gutenberg E2E AT (mobile)

## Pre-merge Checklist

### DIFF
--- a/packages/calypso-e2e/src/lib/blocks/block-flows/layout-grid.ts
+++ b/packages/calypso-e2e/src/lib/blocks/block-flows/layout-grid.ts
@@ -115,14 +115,12 @@ export class LayoutGridBlockFlow implements BlockFlow {
 	 * @param {PublishedPostContext} context The current context for the published post at the point of test execution.
 	 */
 	async validateAfterPublish( context: PublishedPostContext ): Promise< void > {
-		const expectedLeftColumnTextLocator = context.page.locator(
-			`text=${ this.configurationData.leftColumnText }`
-		);
-		await expectedLeftColumnTextLocator.waitFor();
+		await context.page
+			.locator( `:text("${ this.configurationData.leftColumnText }"):visible` )
+			.waitFor();
 
-		const expectedRightColumnTextLocator = context.page.locator(
-			`text=${ this.configurationData.rightColumnText }`
-		);
-		await expectedRightColumnTextLocator.waitFor();
+		await context.page
+			.locator( `:text("${ this.configurationData.rightColumnText }"):visible` )
+			.waitFor();
 	}
 }
--- a/packages/calypso-e2e/src/lib/components/editor-gutenberg-component.ts
+++ b/packages/calypso-e2e/src/lib/components/editor-gutenberg-component.ts
@@ -1,6 +1,5 @@
 import { Page, ElementHandle, Locator } from 'playwright';
 
-const editorPane = 'div.edit-post-visual-editor__content-area';
 const selectors = {
 	// Title
 	title: '.editor-post-title__input',
@@ -147,7 +146,7 @@ export class EditorGutenbergComponent {
 	async getSelectedBlockElementHandle( blockEditorSelector: string ): Promise< ElementHandle > {
 		// Note the :is() selector. This is to support both the block API v1 and V2.
 		const locator = this.editorCanvas.locator(
-			`:is(${ editorPane } ${ blockEditorSelector }.is-selected, ${ editorPane } ${ blockEditorSelector }.has-child-selected)`
+			`:is(${ blockEditorSelector }.is-selected, ${ blockEditorSelector }.has-child-selected)`
 		);
 		await locator.waitFor();
 		return ( await locator.elementHandle() ) as ElementHandle;
--- a/packages/calypso-e2e/src/lib/pages/revisions-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/revisions-page.ts
@@ -1,4 +1,5 @@
 import { Page } from 'playwright';
+import { ElementHelper } from '../..';
 
 /**
  * Represents the Revisions page used on Atomic sites.
@@ -38,8 +39,15 @@ export class RevisionsPage {
 		// slider width.
 		const clickPositionX = ( sliderWidth * ( index - 1 ) ) / ( revisionCount - 1 );
 
-		await this.page.waitForLoadState( 'networkidle' );
-		await slider.click( { position: { x: clickPositionX, y: 1 } } );
+		await Promise.all( [
+			// We need to rely on waiting for mutations here because neither
+			// checking the ".loading" status nor the auto-waiting are enough to
+			// indicate that the diff has been loaded and is ready to be
+			// interacted with. This might be related to the fact that the diff
+			// is rendered entirely from a fetched HTML string.
+			ElementHelper.waitForMutations( this.page, '.revisions-diff-frame' ),
+			slider.click( { position: { x: clickPositionX, y: 1 } } ),
+		] );
 	}
 
 	/**
--- a/packages/calypso-e2e/src/lib/pages/revisions-page.ts
+++ b/packages/calypso-e2e/src/lib/pages/revisions-page.ts
@@ -1,5 +1,4 @@
 import { Page } from 'playwright';
-import { ElementHelper } from '../..';
 
 /**
  * Represents the Revisions page used on Atomic sites.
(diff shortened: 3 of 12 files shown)


packages/calypso-e2e/src/lib/blocks/block-flows/layout-grid.ts
```diff
@@ -115,14 +115,12 @@ export class LayoutGridBlockFlow implements BlockFlow {
...
+			.locator( `:text("${ this.configurationData.leftColumnText }"):visible` )
+			.waitFor();
 
-		const expectedRightColumnTextLocator = context.page.locator(
-			`text=${ this.configurationData.rightColumnText }`
-		);
-		await expectedRightColumnTextLocator.waitFor();
+		await context.page
+			.locator( `:text("${ this.configurationData.rightColumnText }"):visible` )
+			.waitFor();
```
worldomonation, line 124: The previous selector ended up matching on more than 1 element on the page, but one of the element was always hidden.

By specifying the `:visible` pseudo-CSS class, we are able to limit the match to visible elements.

packages/calypso-e2e/src/lib/components/editor-gutenberg-component.ts
```diff
@@ -147,7 +146,7 @@ export class EditorGutenbergComponent {
 	async getSelectedBlockElementHandle( blockEditorSelector: string ): Promise< ElementHandle > {
 		// Note the :is() selector. This is to support both the block API v1 and V2.
 		const locator = this.editorCanvas.locator(
-			`:is(${ editorPane } ${ blockEditorSelector }.is-selected, ${ editorPane } ${ blockEditorSelector }.has-child-selected)`
+			`:is(${ blockEditorSelector }.is-selected, ${ blockEditorSelector }.has-child-selected)`
```
worldomonation, line 149: These selectors used to have the `editorPane` to narrow the target, but it ended up causing problems here as the DOM structure of the editor canvas has been in flux (block-based theme, etc).

packages/calypso-e2e/src/lib/pages/revisions-page.ts
```diff
@@ -39,30 +38,28 @@ export class RevisionsPage {
...
-			// We need to rely on waiting for mutations here because neither
-			// checking the ".loading" status nor the auto-waiting are enough to
-			// indicate that the diff has been loaded and is ready to be
-			// interacted with. This might be related to the fact that the diff
-			// is rendered entirely from a fetched HTML string.
-			ElementHelper.waitForMutations( this.page, '.revisions-diff-frame' ),
-			slider.click( { position: { x: clickPositionX, y: 1 } } ),
-		] );
+		await this.page.waitForLoadState( 'networkidle' );
+		await slider.click( { position: { x: clickPositionX, y: 1 } } );
```
worldomonation: This is my workaround to the complicated `waitForMutation` call - wait until the network activity has ceased.

`networkidle` has its own set of issues such as when pages inadvertently take longer than expected to load. On the other hand, it's simple to understand and follow along. Any timeouts we can address by adding an appropriate timeout.

packages/calypso-e2e/src/lib/pages/revisions-page.ts
```diff
@@ -39,30 +38,28 @@ export class RevisionsPage {
...
 
-		await Promise.all( [
-			this.page.waitForNavigation( { waitUntil: 'networkidle' } ),
-			restoreButton.click(),
-		] );
+		await restoreButton.click();
+
+		// If the spec is using RevisionsPage, this implies the
+		// account is using WP-Admin.
+		await this.page.waitForURL( /wp-admin\/post.php/, { timeout: 20 * 1000 } );
```
worldomonation, line 71: This bakes in the assumption that navigation initiated from WPAdmin should lead to other WPAdmin pages.

test/e2e/specs/blocks/blocks__coblocks__blocks.ts
```diff
@@ -27,7 +26,11 @@ const features = envToFeatureKey( {
 	COBLOCKS_EDGE: envVariables.TEST_ON_ATOMIC || envVariables.COBLOCKS_EDGE,
 } );
 
-describe( DataHelper.createSuiteTitle( 'CoBlocks: Blocks' ), () => {
+/**
+ * This spec requires the following:
+ * 	- theme: a non-block-based theme (eg. Twenty-Twenty One)
+ */
+describe( 'CoBlocks: Blocks', function () {
```
worldomonation, line 33: This is just part of the ongoing `DataHelper.createrSuiteTitle` removal, plus additional notes required for the spec to function.

test/e2e/specs/blocks/shared/block-smoke-testing.ts
```diff
@@ -53,7 +53,8 @@ export function createBlockTests( specName: string, blockFlows: BlockFlow[] ): v
 				it( `${ blockFlow.blockSidebarName }: Add the block from the sidebar`, async function () {
 					await editorPage.addBlockFromSidebar(
 						blockFlow.blockSidebarName,
-						blockFlow.blockEditorSelector
+						blockFlow.blockEditorSelector,
+						{ noSearch: true }
```
worldomonation, line 57: The sidebar block inserter is very slow, especially when searching.

As there exists other specs covering the search and add portion, I've decided that speeding up this spec was preferable.

packages/calypso-e2e/src/lib/pages/revisions-page.ts
dpasque: **Question/Comment (blocking)**: could you explain this change a little bit more? Was the `waitForMutation` call being flaky?

I could be misunderstanding, but I'm not sure this is a comparable change, and I'm not sure this will get us what we want over time. 🤔 

Previously, the `Promise.all` check, combined with the `waitForMutations` call, made sure that we knew the revisions view had finished changing _after_ we used the slider.

Now, we're just waiting for networkidle here _before_ we use the slider. So interacting with the revision panel right after this is still prone to race conditions!

Also, I'm not sure checking for `networkidle` this long after a document load is going to do anything. I'm pretty sure that Playwright looks back to see if that event ever fired for this URL/document, so this may just immediately carry on?

dpasque (changes requested):
Thanks for making these fixes @worldomonation! 🎉 

Everything looks good to me, except I'm a little confused by the waitForMutations change. Can you explain that a bit more?

packages/calypso-e2e/src/lib/pages/revisions-page.ts
worldomonation: My initial understanding was that `waitForMutation` is checking the stability of the page _before_ the revision was loaded. 

With the corrected spec but without removing the `waitForMutation`, the following is seen because it is leaving behind open handles for Jest to clean up.

```
Ran all test suites matching /test\/e2e\/specs\/editor\/editor__revisions.ts/i.
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Perhaps I can pose a few questions that will help lead to a decision:
- is there a need to "wait" until the revision is ready to interact, as long as the "load revision" button is ready to interact?
- instead of the complex `waitForMutation`, could we stand in a wait for string we know is part of the revision?

dpasque: > With the corrected spec but without removing the waitForMutation, the following is seen because it is leaving behind open handles for Jest to clean up.

Hmm, weird... 🤔 

Here's what my local testing found:
- Running the test without the `waitForMutations` (as in the current change) actually failed! It looks like it hit the race condition the `waitForMutations` was trying to help with -- the revision content didn't change fast enough, so the validations of that new content ended up failing.
- Running with the `waitForMutations` passed and didn't leave any open handles. 

I wonder if maybe I'm missing something else from your local set-up? Here was my command:
`TEST_ON_ATOMIC=true yarn jest specs/editor/editor__revisions.ts`

> is there a need to "wait" until the revision is ready to interact, as long as the "load revision" button is ready to interact?

"Ready to interact" may be confusing wording here... I think the idea was that we don't want to do validations of the _content of the revisions panel_ until we were sure the new revision content loaded.

> Instead of the complex waitForMutation, could we stand in a wait for string we know is part of the revision?

I think that's a totally valid approach! 👍  I personally don't mind the `waitForMutation` function, and I like that it "future-proofs" a little bit, but we can totally design around this by focusing on specific text content and playwright-waiting for that (instead of getting the `innerText` and doing string assertions against that).

I'm fine either way! 😁  I just think the current approach in the PR might introduce some flakiness.

dpasque (approved):
Also, going to approve to unblock you as I'm a bit more async today! 😄 I think we should get the Atomic revisions script stable again before merge, but I'm open to however you want to do that!

packages/calypso-e2e/src/lib/pages/revisions-page.ts
worldomonation: @dpasque I'll revisit or eliminate the `waitForMutation` changes. It was on the table only because Jest reliably (on my machine) threw the open handle warning, but it wasn't strictly necessary to fix the `Editor: Revision` spec.

PR #74746, opened 2023-03-22
e2e test for purchasing multiple plugins

### LINKED ISSUE #74327: e2e test for purchasing multiple plugins
As part of our testing efforts (p1678135347895659-slack-CKZHG0QCR), we'd like to explicitly break this into an issue to account for the time needed.

Acceptance criteria:

- [ ] Add e2e test for purchasing multiple plugins

lupus2k:
Fixes https://github.com/Automattic/wp-calypso/issues/74327

## Proposed Changes

Create e2e test for purchases of multiple plugins.
As on normal flow, you can't purchase multiple plugins, we have to use a free site, and when choosing to upgrade, we need to go back to plugins and add another one. Then, the cart has two plugins + the new plan.

## Testing Instructions

Run `yarn workspace wp-e2e-tests test -- plugins/plugins__purchase.ts `

## Pre-merge Checklist

- [ ] ~~Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)~~
- [ ] ~~Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?~~
- [ ] ~~For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~~


lupus2k:
Hey, @worldomonation, @dpasque, I have problems finishing the flow of this e2e test. As I mentioned in the description, there is a special flow for this one, and all the flow works fine, but when I add [this test](https://github.com/Automattic/wp-calypso/blob/78316ed9b492b49ec375921f705155260375def5/test/e2e/specs/plugins/plugins__purchase.ts#L111-L113), I get a timeout error because it takes too long (it never finished the process).
Can you take a look and see if everything is ok with this?

lupus2k:
Btw, it's also happening another error on `specs/domain-upsell/domain-upsell__my-home.ts` > `Check only one item on checkout` and could be related to the empty cart problem you [fixed here](https://github.com/Automattic/wp-calypso/pull/73608#issuecomment-1440390545) using the REST API.

I have to test the purchase of a single plugin. Is there a user with a credit card and an Atomic testing site? (I tried with atomicUser but it has no payment methods)

worldomonation:
> Hey, @worldomonation, @dpasque, I have problems finishing the flow of this e2e test. As I mentioned in the description, there is a special flow for this one, and all the flow works fine, but when I add [this test](https://github.com/Automattic/wp-calypso/blob/78316ed9b492b49ec375921f705155260375def5/test/e2e/specs/plugins/plugins__purchase.ts#L111-L113), I get a timeout error because it takes too long (it never finished the process). Can you take a look and see if everything is ok with this?

It is probably timing out because of the 300 second wait to finish purchase. That is likely too long for Jest test runner itself, which is why you're seeing a timeout.

worldomonation:
> As on normal flow, you can't purchase multiple plugins,

**question:blocking** - if this is not a "normal" flow, as in something that a sizeable number of users will be going through, I'm not so certain that it is worth adding a E2E test for it. 

[I go into depth on some of the criteria to use when evaluating the feasibility of an E2E test](p6fDka-Ld-p2) and the checklist probably applies here. At first glance, it seems to be a pretty niche flow, and my concern is that it will be neglected soon after release.

test/e2e/specs/plugins/plugins__purchase.ts
```diff
@@ -0,0 +1,136 @@
...
+
+			// Launch browser.
+			page = await browser.newPage();
+
+			// Authenticate as simpleSiteFreePlanUser.
+			const testAccount = new TestAccount( 'simpleSiteFreePlanUser' );
+			await testAccount.authenticate( page );
+			siteURL = newSiteDetails.blog_details.site_slug as string;
+			pluginsPage = new PluginsPage( page );
+			await pluginsPage.visit( siteURL );
```
worldomonation: **question:blocking** - [if the spec is to be kept](https://github.com/Automattic/wp-calypso/pull/74746#issuecomment-1478823150), does it _need_ to be run against a fresh site? 

Nothing is ever really deleted, so we strive to be very careful with the data we generate and put into the database when testing. That is why all the "expensive" tests database-wise is reserved for the Pre-Release Tests, which are only run ~40 times a day.

packages/calypso-e2e/src/lib/pages/plugins-page.ts
```diff
@@ -34,11 +34,17 @@ const selectors = {
...
 	deactivateButton: 'button:text("Deactivate")',
 	activateButton: 'button:text("Activate")',
 	openRemoveMenuButton: '.plugin-details-cta__manage-plugin-menu button[title="Toggle menu"]',
 	removeButton: '.popover__menu button:has-text("Remove")',
 
+	// Elegibility warnings
+	eligibilityWarning:
+		'.eligibility-warnings__title:has-text("Upgrade your plan to install plugins")',
+	eligibilityWarningContinueButton: 'button:text("Upgrade and activate plugin")',
+
```
worldomonation: **comment:non-blocking** - Playwright has implemented new locator methods that eliminate the need  to keep these "selector objects" in the POMs. 

Methods like [`getByText`](https://playwright.dev/docs/locators#locate-by-text) can stand in for the selectors and make the class overall easier to follow and read.

worldomonation (changes requested):
I've had a look over the whole PR but I had some overarching concerns with the long-term maintainability of the proposed E2E spec. If we agree that it is worth keeping, I can do a more in-depth review.

packages/calypso-e2e/src/lib/pages/plugins-page.ts
lupus2k: Thanks, @worldomonation, I will use the new methods from now on!

test/e2e/specs/plugins/plugins__purchase.ts
lupus2k: Currently, there is no way to install >1 plugin or plugin+theme using a standard flow, and I think we can avoid this test until we have a specific flow for this.

lupus2k:
Thanks, @worldomonation, for your detailed answer. I will go for testing only for one plugin purchase. I commented about it [here](https://github.com/Automattic/wp-calypso/pull/74746#issuecomment-1478817337), could you take a look at this?

worldomonation:
> Btw, it's also happening another error on `specs/domain-upsell/domain-upsell__my-home.ts` > `Check only one item on checkout` and could be related to the empty cart problem you [fixed here](https://github.com/Automattic/wp-calypso/pull/73608#issuecomment-1440390545) using the REST API.
> 
> I have to test the purchase of a single plugin. Is there a user with a credit card and an Atomic testing site? (I tried with atomicUser but it has no payment methods)

I don't think we have anything set up payments for an AT user yet. I can get that done.

Keep in mind though, all purchases are in the sandboxed site - which _should_ be a good enough stand-in, but it isn't 100% identical.

lupus2k:
> I don't think we have anything set up payments for an AT user yet. I can get that done.

Thanks, it will be good to have it. 
For now, I will add some validations on the free plugins tests that run on [here](https://github.com/Automattic/wp-calypso/blob/trunk/test/e2e/specs/plugins/plugins__install-remove.ts)

lupus2k:
We only tested adding multiple plugins to the cart until we had an atomic user to test it. Follow up [here](https://github.com/Automattic/dotcom-forge/issues/2032)

DustyReagan (approved):
LGTM!

test/e2e/specs/plugins/plugins__purchase.ts
```diff
@@ -0,0 +1,102 @@
...
+		// Authenticate as simpleSiteFreePlanUser.
+		const testAccount = new TestAccount( 'simpleSiteFreePlanUser' );
+		await testAccount.authenticate( page );
+		siteURL = credentials.testSites?.primary.url as string;
+		pluginsPage = new PluginsPage( page );
+		await pluginsPage.visit( siteURL );
+	} );
+
+	describe( 'Plugin: Purchase', function () {
+		jest.setTimeout( 1800000 );
```
worldomonation: **comment:non-blocking** - As far as I can tell, Playwright is still operating on the default timeout of 10 seconds for all waits, so I'm not sure if extending Jest's timeout is going to have any effect.

packages/calypso-e2e/src/lib/pages/plugins-page.ts
```diff
@@ -286,6 +290,26 @@ export class PluginsPage {
...
+			.getByText( 'Upgrade your plan to install plugins' )
+			.waitFor();
+	}
+
+	/**
+	 * Clicks continue button on the Elegibility warning.
+	 */
+	async clickContinueElebigilityWarning(): Promise< void > {
+		const locator = this.page.getByText( 'Upgrade and activate plugin' );
+		await Promise.all( [ locator.click() ] );
```
worldomonation: **suggestion:non-blocking** - this works, but it isn't necessary to wrap a single call in `Promise.all` pattern. 

This would do the same:

```suggestion
		await locator.click();
```

packages/calypso-e2e/src/lib/pages/plugins-page.ts
```diff
@@ -286,6 +290,26 @@ export class PluginsPage {
 		await Promise.all( [ this.page.waitForResponse( /.*install\?.*/ ), locator.click() ] );
 	}
 
+	/**
+	 * Clicks on the `Purchase and Activate` button and get Elegibility warning.
+	 */
+	async clickPurchasePlugin(): Promise< void > {
+		const locator = this.page.locator( selectors.purchaseButton );
+		await locator.click();
```
worldomonation: **suggestion:non-blocking** - these two lines can likely be replaced with a `getByText` selector then a click:

```suggestion
		await this.page.getByText( "Purchase and activate" ).click();
```

and then the change in Line 37 can be removed if this suggestion is taken.

packages/calypso-e2e/src/lib/pages/plugins-page.ts
```diff
@@ -286,6 +290,26 @@ export class PluginsPage {
...
+		await this.page
+			.locator( selectors.eligibilityWarning )
+			.getByText( 'Upgrade your plan to install plugins' )
+			.waitFor();
```
worldomonation: **question:non-blocking** - so I'm in two minds about this here.

On one hand, it's good safe practice to ensure the state is as expected after performing an action. This is in contrast to the "fire-and-forget" approach where the method performs an action and that's all it does.

On the other hand though, this bakes in the assumption that when one clicks to purchase a plugin, they're **_always_** met with an eligibility warning. Is this the case, even for first-party plugins made by A8C?

If the eligibility warning is shown for _all_ plugins, then I think we should keep this check. In addition, perhaps even combine this method with the `clickContinueElebigilityWarning` method so that it achieves the end result, which is to install a plugin.

test/e2e/specs/plugins/plugins__purchase.ts
```diff
@@ -0,0 +1,102 @@
...
+	PluginsPage,
+} from '@automattic/calypso-e2e';
+import { Page, Browser } from 'playwright';
+
+declare const browser: Browser;
+
+describe( DataHelper.createSuiteTitle( 'Plugins: Add two plugins to the cart' ), function () {
+	const planName = 'Business';
+	const pluginName =
+		envVariables.VIEWPORT_NAME === 'desktop' ? 'WooCommerce Bookings' : 'WooCommerce Subscriptions';
```
worldomonation: **question:blocking** - if we're keeping the two-plugin approach here, could we be deterministic as to which is plugin1 and which is plugin2? I think it would help make the flow clearer to follow.

packages/calypso-e2e/src/lib/pages/cart-checkout-page.ts
```diff
@@ -304,4 +308,27 @@ export class CartCheckoutPage {
...
+	/**
+	 * Complete the purchase accepting third party developer check and clicking on the 'Pay' button.
+	 */
+	async purchaseWithPlugin( { timeout }: { timeout?: number } = {} ): Promise< void > {
+		await this.page.getByLabel( selectors.thirdPartyDeveloperCheckboxLabel ).check();
+		await Promise.all( [
+			this.page.waitForResponse( /.*marketplace\/thank-you.*/, { timeout: timeout } ),
+			this.page.click( selectors.purchaseButton ),
+		] );
+	}
```
worldomonation: **suggestion:non-blocking** - this is also partly motivated by style reasons, but hear me out.

I walked through the steps to verify my logic and it looks like the difference in checkout flow between a third-party plugin and a plan is minimal, just the checkbox.

I think either one of the following approaches would be more maintainable going forward:
- a method to tick the third-party plugin agreement box to be called prior to the purchase, or;
- add an optional parameter to the existing `purchase`, named something like `checkThirdPartyAgree`, then add an `if` clause prior to checking out.

Option 2 example:

```
async purchase( { checkThirdPartyAgree, timeout }: { checkThirdPartyAgree?: boolean, timeout?: number } = {} ): Promise< void > {
        if (checkThirdPartyAgree) {
                await this.page.getByLabel( target ).check();
        }

		await Promise.all( [
			this.page.waitForResponse( /.*me\/transactions.*/, { timeout: timeout } ),
			this.page.click( selectors.purchaseButton ),
		] );
	}
```

image

worldomonation (commented):
Generally, I think this looks good. I did have some questions and suggestions to reduce the number of lines of code that needs to be added while keeping the page objects tightly focused.

One thing I noticed when I went through the flow manually - after checking out a Business plan upgrade and 2 WooCommerce plugins, the page was stuck on the "Installing Plugin..." screen. I haven't tried replicating it on another site. From what I can tell, the plugins were never actually installed.

test/e2e/specs/plugins/plugins__purchase.ts
lupus2k: Thanks, I'm going to remove the wait time here anyway because since we're not doing the "pay" step, there's nothing that can take a while

packages/calypso-e2e/src/lib/pages/plugins-page.ts
lupus2k: Not all the plugins show the eligibility warning. This depends on the plan or the site type (Simple sites show it always)
I will separate the clickPurchase and ClickEligibility methods.

packages/calypso-e2e/src/lib/pages/cart-checkout-page.ts
lupus2k: [I checked the code](https://github.com/Automattic/wp-calypso/blob/ccac18fca861c5f812a1c0503528f295023bab28/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx#L482) and this checkbox appears when the cart has a Marketplace product, and in this case the `waitForResponse` URL is different. 
I will use your `Option 2` but instead of `checkThirdPartyAgree`, I will add `hasMarketplaceProduct`

lupus2k: Btw, I will not use the purchase method until I can test it purchase a plugin (after my AFK)

test/e2e/specs/plugins/plugins__purchase.ts
lupus2k: I updated the names to plugin1Name and plugin2Name, and also updated the description of some tests.

packages/calypso-e2e/src/lib/pages/cart-checkout-page.ts
lupus2k: I created a follow-up for the purchase process here: https://github.com/Automattic/dotcom-forge/issues/2032

lupus2k:
@worldomonation I'm still having some failed tests, but I think is because of parallel tests, what do you think about it?

packages/calypso-e2e/src/lib/pages/cart-checkout-page.ts
```diff
@@ -298,10 +302,32 @@ export class CartCheckoutPage {
...
+		const URLtoValidate = hasMarketplaceProduct
+			? /.*marketplace\/thank-you.*/
+			: /.*me\/transactions.*/;
+
+		if ( hasMarketplaceProduct ) {
+			await this.page.getByLabel( selectors.thirdPartyDeveloperCheckboxLabel ).check();
+		}
 		await Promise.all( [
-			this.page.waitForResponse( /.*me\/transactions.*/, { timeout: timeout } ),
+			this.page.waitForResponse( URLtoValidate, { timeout: timeout } ),
```
worldomonation: **comment:non-blocking** - if you'd like, you can simplify this wait on multiple URLs or values like this

```typescript
/(marketplace\/thank-you|me\/transactions)/
```

test/e2e/specs/plugins/plugins__purchase.ts
```diff
@@ -0,0 +1,104 @@
...
+	BrowserManager,
+	SecretsManager,
+	envVariables,
+	PluginsPage,
+} from '@automattic/calypso-e2e';
+import { Page, Browser } from 'playwright';
+
+declare const browser: Browser;
+
+describe( DataHelper.createSuiteTitle( 'Plugins: Add two plugins to the cart' ), function () {
```
worldomonation: **suggestion:non-blocking** - this is a more recent development and not recorded anywhere (yet) but we're moving away from the `DataHelper.createSuiteTitle`. You don't have to, but if you'd like you can remove it and just define the title directly like this:

```
describe( 'Plugins: Add two plugins to the cart' , function () {
```

test/e2e/specs/plugins/plugins__purchase.ts
```diff
@@ -0,0 +1,104 @@
...
+		it( `Purchase ${ plugin2Name } and show Elegibility warning`, async function () {
+			await pluginsPage.visitPage( plugin2Name.replace( ' ', '-' ).toLowerCase(), siteURL );
+			await pluginsPage.clickPurchasePlugin();
+		} );
+
+		it( `Validate Elegibility warning for ${ plugin2Name } and continue`, async function () {
+			await pluginsPage.validateAndContinueElebigilityWarning();
+		} );
+
+		it( `WordPress.com ${ planName } is still to cart`, async function () {
```
worldomonation: **nit:blocking** - "is still in cart" I think is the title you're looking for! Also applies to  subsequent steps.

worldomonation (approved):

test/e2e/specs/plugins/plugins__purchase.ts
lupus2k: You're right, thanks!

PR #74747, opened 2023-03-22
Plans: Update Enterprise hardcoded pricing to use formatCurrency

jjchrisdiehl:
This will pass the hardcoded `$25,000` price point for Enterprise plans through formatCurrency to ensure the formatting is localized and displayed correctly across languages/locales.

Related to #74605

## Testing Instructions

* Set your user currency to EUR (or any non-USD currency)
* Set locale to `fr-CA` from `/me` for French Canadian (Espanol works too)
* Go to `/plans/[your-site-name]`

If you're within the US, you should see the following:
image

If you're outside the US, you should see:
image

Note that the currencies are showing as right aligned, this is how formatCurrency aligns for `fr-CA` and `es` locales. Also note that when outside of the US you'll see `US$` right aligned, this is expected and correct. 

You may also run into some weird fringe cases if you're in Canada or Mexico where the currency is showing as `$`, this is because you're actually hitting our US servers, in those cases you may need to use a VPN to test this.


sirbrillig (approved):
Looks good to me! We should do a String Freeze and get translations prepared before merging this, though.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7883693

Thank you @jjchrisdiehl for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7883693).

sirbrillig:
Still looks good to me!

PR #74748, opened 2023-03-22
Checkout: add logged out checkout for Akismet

IanRamosC:
Related to 1201210512993648-as-1204136663449777

## Proposed Changes

* Add specific checkout flow for Akismet users who are not logged in. It allows users to complete the checkout process without a WordPress.com account

## Testing Instructions
* Follow the instructions in D105505-code
* Checkout this branch locally
* `yarn start`
* Go to http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1 in a browser that doesn't have an account logged in (it can be the incognito mode)
* Add your email (e.g. your.email+test@gmail.com)
* Complete the purchase using a [test card](https://stripe.com/docs/testing#cards), and make sure you get the receipt and a welcome email

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
@@ -17,17 +17,19 @@ export async function createWpcomAccountBeforeTransaction(
 		( product ) => product.extra.isGiftPurchase
 	);
 
-	let signupFlowName;
+	const signupFlowName = ( () => {
+		if ( isJetpackUserLessCheckout ) {
+			return 'jetpack-userless-checkout';
+		}
+		if ( isAkismetUserLessCheckout ) {
+			return 'akismet-userless-checkout';
+		}
+		if ( isGiftingCheckout ) {
+			return 'gifting-userless-checkout';
+		}
 
-	if ( isJetpackUserLessCheckout ) {
-		signupFlowName = 'jetpack-userless-checkout';
-	} else if ( isAkismetUserLessCheckout ) {
-		signupFlowName = 'akismet-userless-checkout';
-	} else if ( isGiftingCheckout ) {
-		signupFlowName = 'gifting-userless-checkout';
-	} else {
-		signupFlowName = 'onboarding-registrationless';
-	}
+		return 'onboarding-registrationless';
+	} )();
 
 	/*
 	 * We treat Gifting as jetpack-userless-checkout to create and verify the user
--- a/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
@@ -10,21 +10,31 @@ export async function createWpcomAccountBeforeTransaction(
 	const isJetpackUserLessCheckout = transactionCart.products.some(
 		( product ) => product.extra.isJetpackCheckout
 	);
+	const isAkismetUserLessCheckout = transactionCart.products.some(
+		( product ) => product.extra.isAkismetSitelessCheckout
+	);
 	const isGiftingCheckout = transactionCart.products.some(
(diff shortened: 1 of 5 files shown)


client/my-sites/checkout/composite-checkout/lib/submit-wpcom-transaction.ts
```diff
@@ -25,11 +25,12 @@ export default async function submitWpcomTransaction(
 	payload: WPCOMTransactionEndpointRequestPayload,
 	transactionOptions: PaymentProcessorOptions
 ): Promise< WPCOMTransactionEndpointResponse > {
-	const isJetpackUserLessCheckout =
-		payload.cart.products.some( ( product ) => product.extra.isJetpackCheckout ) &&
-		payload.cart.cart_key === 'no-user';
+	const isUserLessCheckout =
+		payload.cart.products.some( ( product ) => {
+			return product.extra.isJetpackCheckout || product.extra.isAkismetSitelessCheckout;
+		} ) && payload.cart.cart_key === 'no-user';
```
IanRamosC, line 31: The change here is a little bit similar to the one in the file above, but I decided not to make a helper function because this is very specific and might vary from other places where we would want only one check or the other (i.e. if it has the product.extra properties, or the cart_key, or even just one of the properties).

client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
```diff
@@ -10,21 +10,31 @@ export async function createWpcomAccountBeforeTransaction(
...
 		( product ) => product.extra.isJetpackCheckout
 	);
+	const isAkismetUserLessCheckout = transactionCart.products.some(
+		( product ) => product.extra.isAkismetSitelessCheckout
+	);
 	const isGiftingCheckout = transactionCart.products.some(
 		( product ) => product.extra.isGiftPurchase
 	);
 
+	let signupFlowName;
```
sirbrillig: Thanks! This is much better than the growing ternary it replaces. I think the following block is pretty readable right now, but in my experience, over time if/else code like this will get muddied and mixed up with other conditions. Could we use an IIFE instead? Something like this:

```suggestion
	const signupFlowName = ( () => {
	if ( isJetpackUserLessCheckout ) {
		return 'jetpack-userless-checkout';
	}
	if ( isAkismetUserLessCheckout ) {
		return 'akismet-userless-checkout';
	}
	if ( isGiftingCheckout ) {
		return 'gifting-userless-checkout';
	}
	return 'onboarding-registrationless';
} )();
```

This way the priority of each value is clear (the conditions do not overlap right now but I am sure they will eventually), assignments will clearly be made in one place (often when a variable can be modified folks will start making assignments in different areas over time) and the whole thing could be moved to its own function if needed.

sirbrillig (commented):
This code looks good to me!

However, when I tried to follow the testing instructions (and I'm probably doing something wrong) I get an error message:

Screenshot 2023-03-22 at 11 07 39 AM

Any advice?

IanRamosC:
> However, when I tried to follow the testing instructions (and I'm probably doing something wrong) I get an error message:

@sirbrillig, you're not doing anything wrong. I moved this PR back to In Progress yesterday because right before I pushed created the WPCOM diff, I rebased the branch I was working on - and trunk had code that made this PR stop working. I've just updated the WPCOM diff so this PR should work now.

client/my-sites/checkout/composite-checkout/lib/create-wpcom-account-before-transaction.ts
IanRamosC: Sure, that makes more sense to me as well. :)

jboland88 (approved):
Code changes look good and this tested well 👍 

I was redirected to a login screen after completing the checkout - it seems like this may be a result of the test setup. It looks like the expected behavior is to log the new user in 🤔 but we can test this more and follow up if needed.

Issue #73688, opened 2023-02-22
Can't hide Fiverr card in Calypso


edequalsawesome:
### Quick summary

The Fiverr card in the My Home area of Calypso can't be hidden, it just gets a spinning "loading" icon that never clears

### Steps to reproduce

1. Go to the My Home page in Calypso
2. Click on the `Hide > Forever` option in the Fiverr card

### What you expected to happen

The card would hide like all the other ones

### What actually happened

The Loading icon keeps going in a loop (but _only_ for the Fiverr card!)

![2023-02-22 16 18 23](https://user-images.githubusercontent.com/4000628/220762571-9e2ec5a1-6a4d-4459-a4b8-08bf926fa77b.gif)

### Impact

All

### Available workarounds?

No but the platform is still usable

### Platform (Simple and/or Atomic)

Simple, Atomic

### Logs or notes

- Tested on Firefox, Edge, and Chrome
- Originally reported in #38525816-hc, follow-up in #5983915-zen

cuemarie:
📌 SCRUBBING : RESULT - *Replicated / Could Not Replicate / Uncertain*

- Tested on Simple: Replicated
- Tested on AT: Replicated

📌 FINDINGS/SCREENSHOTS/VIDEO
- Replicated as described! Even when going out of order, hiding the Fiverr logo design card fails to hide. 

**AT Example**

https://user-images.githubusercontent.com/27249804/221082955-8def7666-40e2-4ae7-9e8b-2e513f1fafbc.mov

**Simple Example**

https://user-images.githubusercontent.com/27249804/221082961-4100c909-7888-4211-8d2e-88c6062ddd8b.mov


📌 ACTIONS
- Marked as Triaged for Quality Squad review

mgozdis:
I also reproduced this on a test site after a user reported here: 5990448-zd-woothemes

aleone89:
There's another report of this here: https://wordpress.com/forums/topic/cant-remove-fiverr-box-on-home-page/?view=all

kingsleyinfo:
There is another report of this here: 
6022827-zd-woothemes

aisajib:
Another user said they couldn't dismiss the Fiverr box. https://wordpress.com/forums/topic/cant-remove-fiverr-box-on-home-page/?view=all#post-3932295

edwinho89:
6039997-zd-woothemes

cuemarie:
[...]

PR #74757, opened 2023-03-22
My Home: Fix Fiverr card hide button and hide Promote post slide

BogdanUngureanu:
Fix the hide button for Fiverr card in My Home and hide the Promote Post slide when the Promote Post card is hidden.

Related to #73688

## Proposed Changes

* Fix the Fiverr card button
* Hide the Promote Post slide when the card is hidden

## Testing Instructions

* Use the live link and go to My home on a Free Site
* Try to dismiss the `Fiverr` card - now it should work
* Notice the `Promote Post` is now hidden and there's no longer an empty slide in the slideshow (if you have a non-en locale)

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/customer-home/cards/tasks/fiverr/index.jsx
+++ b/client/my-sites/customer-home/cards/tasks/fiverr/index.jsx
@@ -1,7 +1,7 @@
 import { useTranslate } from 'i18n-calypso';
 import fiverrIllustration from 'calypso/assets/images/customer-home/illustration--task-fiverr.svg';
 import { preventWidows } from 'calypso/lib/formatting';
-import { TASK_FIND_DOMAIN } from 'calypso/my-sites/customer-home/cards/constants';
+import { TASK_FIVERR } from 'calypso/my-sites/customer-home/cards/constants';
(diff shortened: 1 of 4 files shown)


BogdanUngureanu:
Even though I have a `en` locale on my WP.com account, it looks like I still don't have access to `Promote Posts`. For my account, the Promote Posts slide should always be hidden.

dsas (approved):
It seems odd that the home/layout endpoint (fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Serfg%2Qncv%2Qcyhtvaf%2Sraqcbvagf%2Subzr.cuc%3Se%3Qn90p2nqo-og) would return the card and then calypso would overrule that. Ultimately they both look at the same user option to decide what to do. 

If I changed calypso so that `showPromotePost` was hardcoded to false (and server returning `approved: true`) then I could repro your problem and this change does fix it.

My guess is that this is a stale data problem.

I think it'd be better fixed by just removing the `showPromotePost` check, but I could be wrong.

I'm going to approve because this fixes two problems, but I'm ccing @andres-blanco to get his thoughts on whether a follow up could just remove the showPromotePost variable and uses.

PR #74763, opened 2023-03-22
Subscription Management settings page: Get the current values from the API

### LINKED ISSUE #74764: `/subscriptions/settings`: Display current settings values in the form
## Description of the task

The goal of this task is to load the current values from the API to all form fields on the Settings view:

Markup on 2023-03-22 at 15:24:14

To accomplish the goal, let's use `useSubscriptionManagerUserSettingsQuery()`.

## Acceptance Criteria

- all the form fields will have their values loaded from the API

ivan-ottinger:
Fixes https://github.com/Automattic/wp-calypso/issues/74764.

The goal of this PR is to make sure the Settings form fields get the current values from the API:

Markup on 2023-03-22 at 15:24:14

## Proposed Changes

* provide current value to all Settings view form fields from the API (via `useSubscriptionManagerUserSettingsQuery()` callback)
* move `DeliveryWindowDayType` and `DeliveryWindowHourType` types definition to `packages/data-stores/src/reader/types/index.ts`
* provide default `false` value for the `BlockEmailsSetting` component (this was necessary because of "A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen." warning)

## Testing Instructions

1. Check out the branch of this PR.
2. Navigate to https://wordpress.com/me/notifications/subscriptions and adjust the settings the way you like.
3. Open http://calypso.localhost:3000/subscriptions/settings. The changes you saved on the previous step should be reflected on the new Settings view.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/subscriptions/index.tsx
+++ b/client/my-sites/subscriptions/index.tsx
@@ -10,11 +10,11 @@ import { makeLayout, render } from 'calypso/controller';
 
 const SitesView = () => <span>Sites View</span>;
 const CommentsView = () => <span>Comments View</span>;
-const SettingsView = () => <SubscriptionManager.UserSettings />;
 
 const SubscriptionManagementPage = () => {
 	const translate = useTranslate();
 	const { data: counts } = Reader.useSubscriptionManagerSubscriptionsCountQuery();
+	const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
 	const locale = useLocale();
 
 	return (
@@ -39,7 +39,7 @@ const SubscriptionManagementPage = () => {
 						{
 							label: translate( 'Settings' ),
 							path: 'settings',
-							view: SettingsView,
+							view: () => <SubscriptionManager.UserSettings value={ settings } />,
 						},
 					] }
 				/>
--- a/client/my-sites/subscriptions/index.tsx
+++ b/client/my-sites/subscriptions/index.tsx
@@ -14,10 +14,12 @@ const CommentsView = () => <span>Comments View</span>;
 const SubscriptionManagementPage = () => {
 	const translate = useTranslate();
 	const { data: counts } = Reader.useSubscriptionManagerSubscriptionsCountQuery();
-	const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
 	const locale = useLocale();
 
-	const SettingsView = () => <SubscriptionManager.UserSettings value={ settings } />;
+	const SettingsView = () => {
+		const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
+		return <SubscriptionManager.UserSettings value={ settings } />;
+	};
 
 	return (
 		<MomentProvider currentLocale={ locale }>
--- a/client/my-sites/subscriptions/index.tsx
+++ b/client/my-sites/subscriptions/index.tsx
@@ -17,6 +17,8 @@ const SubscriptionManagementPage = () => {
 	const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
 	const locale = useLocale();
 
+	const SettingsView = () => <SubscriptionManager.UserSettings value={ settings } />;
+
 	return (
 		<MomentProvider currentLocale={ locale }>
 			<SubscriptionManager>
@@ -39,7 +41,7 @@ const SubscriptionManagementPage = () => {
 						{
 							label: translate( 'Settings' ),
 							path: 'settings',
-							view: () => <SubscriptionManager.UserSettings value={ settings } />,
+							view: SettingsView,
 						},
 					] }
 				/>
--- a/packages/data-stores/src/reader/types/index.ts
+++ b/packages/data-stores/src/reader/types/index.ts
@@ -1,9 +1,12 @@
 export type EmailFormatType = 'html' | 'text';
 
+export type DeliveryWindowDayType = 0 | 1 | 2 | 3 | 4 | 5 | 6;
+export type DeliveryWindowHourType = 0 | 2 | 4 | 6 | 8 | 10 | 12 | 14 | 16 | 18 | 20 | 22;
+
 export type SubscriptionManagerUserSettings = Partial< {
 	mail_option: EmailFormatType;
-	delivery_day: number; // 0-6, 0 is Sunday
-	delivery_hour: number; // 0-23, 0 is midnight
+	delivery_day: DeliveryWindowDayType; // 0-6, 0 is Sunday
+	delivery_hour: DeliveryWindowHourType; // 0-23, 0 is midnight
 	blocked: boolean;
 	email: string;
 } >;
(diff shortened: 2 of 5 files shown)


packages/data-stores/src/reader/types/index.ts
```diff
@@ -1,9 +1,12 @@
 export type EmailFormatType = 'html' | 'text';
 
+export type DeliveryWindowDayType = 0 | 1 | 2 | 3 | 4 | 5 | 6;
+export type DeliveryWindowHourType = 0 | 2 | 4 | 6 | 8 | 10 | 12 | 14 | 16 | 18 | 20 | 22;
```
ivan-ottinger, line 4: We already have these two types defined inside the `subscription-manager` package: `packages/subscription-manager/src/components/fields/DeliveryWindowInput/DeliveryWindowInput.tsx`.

What is not clear to me is whether it is better practice to:
  A) have the types defined in both places
  B) keep the types defined in `subscription-manager` package and just import it to `data-stores`
  C) keep the types defined in `data-stores` package and just import it to `subscription-manager`
  D) some other solution

TimBroddin: I would say C, because I would expect the data-stores to be the source of truth. 

@mashikag what do you think?

client/my-sites/subscriptions/index.tsx
```diff
@@ -10,13 +10,15 @@ import { makeLayout, render } from 'calypso/controller';
 
 const SitesView = () => <span>Sites View</span>;
 const CommentsView = () => <span>Comments View</span>;
-const SettingsView = () => <SubscriptionManager.UserSettings />;
 
 const SubscriptionManagementPage = () => {
 	const translate = useTranslate();
 	const { data: counts } = Reader.useSubscriptionManagerSubscriptionsCountQuery();
+	const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
```
TimBroddin: I know that the initial design of `<SubscriptionManager.UserSettings /> ` has the value & onChange props. But wouldn't it actually be better to do the fetching and mutation inside the component itself?

Currently, all data would be loaded when you visit *any* subscription management page. Which seems like a bit of overkill. 

@mashikag is there a reason you made the component was designed like this? Do we intend to reuse it? In that case it might be better to create a wrapper I think.

mashikag: Oh no, this was not my intention. Let me type in my code suggestion. I will link it here.

client/my-sites/subscriptions/index.tsx
```diff
@@ -10,13 +10,15 @@ import { makeLayout, render } from 'calypso/controller';
 
 const SitesView = () => <span>Sites View</span>;
 const CommentsView = () => <span>Comments View</span>;
-const SettingsView = () => <SubscriptionManager.UserSettings />;
```
mashikag, line 13:
```suggestion
const SettingsView = () => {
    const { data: settings } = Reader.useSubscriptionManagerUserSettingsQuery();
    return <SubscriptionManager.UserSettings value={ settings } />;
}

```
However, if we will be able to use the exact same query and mutation hook for querying and mutating the data for a logged in user, then we may implement it within the `UserSettings` component, as suggested by @TimBroddin.

client/my-sites/subscriptions/index.tsx
mashikag: https://github.com/Automattic/wp-calypso/pull/74763/files/6a3166050091099ce96cd6c50db45c04af1dc69e#r1144717742

packages/data-stores/src/reader/types/index.ts
mashikag: I would also go with C for the same reason as @TimBroddin mentioned above.

client/my-sites/subscriptions/index.tsx
TimBroddin: Yes, it's exactly the same :)

ivan-ottinger: Thanks for the feedback Tim & Maciej! I will applied the suggested changes.

client/my-sites/subscriptions/index.tsx
ivan-ottinger: [Suggested code](https://github.com/Automattic/wp-calypso/pull/74763#discussion_r1144717742)

packages/data-stores/src/reader/types/index.ts
ivan-ottinger: ~~Do you have any thoughts on the types definition above as well, @TimBroddin? Thanks! 🙂~~

ivan-ottinger: Oh, your messages just appeared. Scratch the last message! 🙂 

Thanks for your feedback.

client/my-sites/subscriptions/index.tsx
ivan-ottinger: Addressed in the following commit: https://github.com/Automattic/wp-calypso/pull/74763/commits/aebfe81d3dc02b349251c79200e2554ddb91eb5e.

packages/data-stores/src/reader/types/index.ts
ivan-ottinger: Addressed in https://github.com/Automattic/wp-calypso/pull/74763/commits/a85ff4e5d40be2789286774f38b9ea0f4b4e0e42.

packages/data-stores/src/reader/types/index.ts
```diff
@@ -1,9 +1,12 @@
...
+
 export type SubscriptionManagerUserSettings = Partial< {
 	mail_option: EmailFormatType;
-	delivery_day: number; // 0-6, 0 is Sunday
-	delivery_hour: number; // 0-23, 0 is midnight
+	delivery_day: DeliveryWindowDayType; // 0-6, 0 is Sunday
```
mashikag:
```suggestion
	delivery_day: DeliveryWindowDayType;
```

packages/data-stores/src/reader/types/index.ts
```diff
@@ -1,9 +1,12 @@
...
+	delivery_hour: DeliveryWindowHourType; // 0-23, 0 is midnight
```
mashikag: Now that we have the types, we can remove the comment
```suggestion
	delivery_hour: DeliveryWindowHourType;

ivan-ottinger: That's true! Great catch. Thanks! Going to fix it now.

ivan-ottinger: Addressed in https://github.com/Automattic/wp-calypso/pull/74763/commits/38055d058bec70ae3cd771d41d949174f3781359.

packages/data-stores/src/reader/types/index.ts
ivan-ottinger: Addressed in https://github.com/Automattic/wp-calypso/pull/74763/commits/38055d058bec70ae3cd771d41d949174f3781359.

mashikag:
The code looks good, and it works as expected too. However, as I played with changing the setting values for an external user (via the old portal), the API started misbehaving. As if it stopped applying my changes. I would hit the Save changes button, no error would be returned, but when the page reloaded it was back to old values. It is something we should keep an eye on.

mashikag (approved):

PR #74766, opened 2023-03-22
Staging Sites: custom my home for staging sites

kozer:
Fixes https://github.com/Automattic/dotcom-forge/issues/1900

## Proposed Changes

Some My-home elements are not relevant for staging sites. This pr introduces a custom my-home for those sites.

### Test new my home

1. Makes sure you have applied D105402-code to your sandbox.
2. Create a staging site
3. Visit `/home/<your site>`
4. Ensure that you see something like the following:

home-card

## Pre-merge Checklist


client/my-sites/customer-home/locations/primary/index.jsx
```diff
@@ -115,13 +118,14 @@ const Primary = ( { cards, trackCard } ) => {
 	if ( cards.length === 1 ) {
 		isUrgent = urgentTasks.includes( cards[ 0 ] );
 	}
+	const showControlLabels = cards.length !== 1 || isUrgent;
```
wojtekn: Why do we want to show dots when there is one urgent task?

wojtekn (commented):
Looks good to me, and it looks well both on mobile and desktop view.

I noticed two issues:
- I tested it with the staging site that was created from a production I don't have access to, and it doesn't work for me in this case. I see a loading indicator all the time
- when I change the staging site title to empty, I see the loading indicator all the time

The second one looks like a small bug, but the first one is more challenging. Maybe, to avoid more rabbit holes, we could remove "for Site Title" from the first iteration? @javierarce thoughts?

client/my-sites/customer-home/locations/primary/index.jsx
kozer: I don't know! But this was the case before. And this is why I kept it like that. The property was true, whatever the case was, so I kept it like that. I only changed it to match my case.

wojtekn: I'm not sure if I understand it correctly. Before, `showControlLabels` was always true. Now, it will be true if we have more than one card (which makes sense), or if we have one card which is urgent.

wojtekn:
@kozer per Slack discussion, let's remove the parts that access production site data in this PR, and we will do this under a separate ticket if we have time.

> Maybe, to avoid more rabbit holes, we could remove "for Site Title" from the first iteration?

kozer:
Already on it

client/my-sites/customer-home/locations/primary/index.jsx
kozer: > I'm not sure if I understand it correctly. Before, showControlLabels was always true. Now, it will be true if we have more than one card (which makes sense), or if we have one card which is urgent.

Yes, as you said before it was always true. Either we have an urgent ( which implies that `cards.length === 1` or not (more than one card).

I don't know if it was on purpose, or not, so I left it as is.

kozer:
@wojtekn Updated.

wojtekn (approved):

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7884762

Thank you @kozer for including a screenshot in the description! This is really helpful for our translators.

wojtekn:
Thanks for landing this @kozer , looks great!

kozer:
You are welcome @wojtekn !

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7884762).

PR #74722, opened 2023-03-21
Plans 2023: Update action buttons to Upgrade when site on shorter term plan

chriskmnds:
Fixes https://github.com/Automattic/wp-calypso/issues/74637

## Proposed Changes

Updates the plan action buttons to direct towards an "Upgrade" when the site plan is on a shorter term. This applies to all `n-yearly` grid variations:  `/plans/yearly`, `/plans/2yearly`, `/plans/3yearly`. Button texts should read "Upgrade to Yearly", "Upgrade to Biannual", and "Upgrade to Triannual", given a respective site with a lower-term plan.

### Media

- `/plans/yearly/[SITE]` with site on Premium **monthly**:

Screenshot 2023-03-22 at 3 39 14 PM

- `/plans/2yearly/[SITE]` with site on Premium **monthly**:

Screenshot 2023-03-22 at 3 39 32 PM

- `/plans/3yearly/[SITE]` with site on Premium **monthly**:

Screenshot 2023-03-22 at 3 39 45 PM

- `/plans/yearly/[SITE]` with site on Business **yearly**:

Screenshot 2023-03-22 at 3 40 00 PM

## Testing Instructions

* Go to each of `/plans/monthly`, `/plans/yearly`, `/plans/2yearly`, `/plans/3yearly` with a variation of site plans (monthly, yearly, biannual, triannual)
* Confirm the updated button texts for the respective longer-term plan variant, as in the media ("Upgrade to ...")
* Confirm the other buttons display and work as expected (downgrade, upgrade, manage-plan, etc.)
* Confirm clicking on "Upgrade to ..." goes to checkout with the respective plan in the cart

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
[...]

PR #74778, opened 2023-03-22
Plans 2023: Always show the biannual toggle. free/paid site plans

### LINKED ISSUE #74636: Plans 2023: Enable the 1yr/2yr toggle when on an annual/biennial plan
As of merging the biennial toggle in https://github.com/Automattic/wp-calypso/pull/74206:

> Visiting /plans, the 1y/2y toggle is missing when I'm on an annual plan. Currently it is hide because it's problematic to downgrade from an annual term to a monthly one. Now that we are showing a longer term option, it makes sense to show the toggle . It would also open up the question regarding what we want to show for the monthly plan owner, though. We can discuss that further separately. 

_Originally posted by @southp in https://github.com/Automattic/wp-calypso/issues/74206#issuecomment-1475758211_

chriskmnds:
Fixes https://github.com/Automattic/wp-calypso/issues/74636
Based on https://github.com/Automattic/wp-calypso/pull/74722

## Proposed Changes

This will show the 1yr/2yr toggle on both paid/free site plans (irrespective of billing term) when visiting `/plans/[SITE]` with the `plans/biannual-toggle` flag enabled. 

Previously, the toggle was only visible on free sites and on `/plans/monthly/[SITE]` with site on a monthly plan due to the "monthly" option and monthly restrictions (downgrading from yearly to monthly not being permissible).

### Media

Screenshot 2023-03-22 at 4 04 28 PM

## Testing Instructions

* Go to `plans/[SITE]` with both free/paid sites and feature flag enabled `?flags=plans/biannual-toggle`
* The 1yr/2yr toggle should always be visible/active

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plans-features-main/plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/plan-type-selector.tsx
@@ -152,15 +152,17 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > = 
 		return currentSitePlanSlug ? getPlanBillPeriod( state, currentSitePlanSlug ) : null;
 	} );
 
-	if ( ! showBiannualToggle ) {
-		if ( ! eligibleForWpcomMonthlyPlans ) {
+	if ( showBiannualToggle ) {
+		// skip showing toggle if current plan's term is higher than 1 year
+		if ( currentPlanBillingPeriod && PLAN_ANNUAL_PERIOD < currentPlanBillingPeriod ) {
 			return null;
 		}
 	}
 
-	// skip showing toggle if current plan's term is higher than 1 year
-	if ( currentPlanBillingPeriod && PLAN_ANNUAL_PERIOD < currentPlanBillingPeriod ) {
-		return null;
+	if ( ! showBiannualToggle ) {
+		if ( ! eligibleForWpcomMonthlyPlans ) {
+			return null;
+		}
 	}
 
 	const additionalPathProps = props.redirectTo ? { redirect_to: props.redirectTo } : {};
--- a/client/my-sites/plans-features-main/plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/plan-type-selector.tsx
@@ -4,6 +4,7 @@ import {
 	isWpComPlan,
 	plansLink,
 	isMonthly,
+	PLAN_ANNUAL_PERIOD,
 } from '@automattic/calypso-products';
 import { Popover } from '@automattic/components';
 import styled from '@emotion/styled';
@@ -16,13 +17,15 @@ import { useSelector } from 'react-redux';
 import CSSTransition from 'react-transition-group/CSSTransition';
 import { Primitive } from 'utility-types';
 import SegmentedControl from 'calypso/components/segmented-control';
-import { ProvideExperimentData } from 'calypso/lib/explat';
 import { addQueryArgs } from 'calypso/lib/url';
 import {
 	getPlanBySlug,
 	getPlanRawPrice,
 	getDiscountedRawPrice,
+	getPlanBillPeriod,
 } from 'calypso/state/plans/selectors';
+import { getSitePlanSlug } from 'calypso/state/sites/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import type { IAppState } from 'calypso/state/types';
 
 export type PlanTypeSelectorProps = {
@@ -143,11 +146,20 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > = 
 	} );
 	const popupIsVisible = Boolean( intervalType === 'monthly' && isInSignup && props.plans.length );
 	const maxDiscount = useMaxDiscount( props.plans );
+	const currentPlanBillingPeriod = useSelector( ( state ) => {
+		const currentSitePlanSlug = getSitePlanSlug( state, getSelectedSiteId( state ) );
+		return currentSitePlanSlug ? getPlanBillPeriod( state, currentSitePlanSlug ) : null;
+	} );
 
 	if ( ! showBiannualToggle ) {
 		if ( ! eligibleForWpcomMonthlyPlans ) {
 			return null;
 		}
+	}
+
+	// skip showing toggle if current plan's term is higher than 1 year
+	if ( currentPlanBillingPeriod && PLAN_ANNUAL_PERIOD < currentPlanBillingPeriod ) {
+		return null;
 	}
 
 	const additionalPathProps = props.redirectTo ? { redirect_to: props.redirectTo } : {};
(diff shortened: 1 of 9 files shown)


aneeshd16 (commented):
I tested this with a site on a monthly plan and noticed that while the toggle was visible, none of the toggle options were in the active state. Also, the plans grid showed monthly prices. 
image

I can't think of any alternate solutions for this other than having a 3-way toggle with Pay monthly/Pay 1 year/Pay 2 years. 🤔

chriskmnds:
> I tested this with a site on a monthly plan and noticed that while the toggle was visible, none of the toggle options were in the active state. Also, the plans grid showed monthly prices.

Good catch @aneeshd16.  This happens only on `/plans/[SITE]` with a site on monthly, right?

> I can't think of any alternative solutions for this other than having a 3-way toggle with Pay monthly/Pay 1 year/Pay 2 years. 🤔

Do you recon there would be any problems with redirecting to `/plans/yearly` in that case? You can test this behavior from the latest commit.

aneeshd16:
> This happens only on /plans/[SITE] with a site on monthly, right?

That's right. 👍 

> Do you recon there would be any problems with redirecting to /plans/yearly in that case? You can test this behavior from the latest commit.

I tried with the latest commit:
1. The "Manage Plan" CTA and the "Your plan" badge are shown when I switch to the 2-yearly view (the site is on a monthly Premium plan)
2. The "Personal" plan CTA should be an "Upgrade" CTA instead of the "Downgrade" one.
image

Even if we fix the above two points, this will lead to a product change:
1. Users won't see the "Manage Plan" CTA and the "Your plan" badge.
2. Users won't be able to upgrade to a higher monthly plan.

chriskmnds:
Thanks @aneeshd16 

> The "Manage Plan" CTA and the "Your plan" badge are shown when I switch to the 2-yearly view (the site is on a monthly Premium plan)

This should be fixed in https://github.com/Automattic/wp-calypso/pull/74722. From my end when visiting `/plans/2yearly/[MONTHLY_PREMIUM]` (on https://github.com/Automattic/wp-calypso/pull/74722):

Screenshot 2023-03-24 at 2 01 35 PM

> The "Personal" plan CTA should be an "Upgrade" CTA instead of the "Downgrade" one.

This is also the case in trunk right now when visiting `/plans/yearly/[MONTHLY]`. It's not something that is being addressed anywhere AFAIK. Also not sure what/ if any caveats with "upgrading" to a lower tier plan for longer-term 🤔  From my end with monthly premium, on trunk:

Screenshot 2023-03-24 at 2 07 47 PM

> Users won't see the "Manage Plan" CTA and the "Your plan" badge.

I think we settled on not treating longer-term plans as "current plan" in the grids for shorter-term ones. Same for looking at 2-yearly with a yearly plan, etc. This is what is covered by in https://github.com/Automattic/wp-calypso/pull/74722 (and the reason it exists)

> Users won't be able to upgrade to a higher monthly plan.

Indeed. I think this is the main deviation (and potential blocker) with having the redirection (`/plans/monthly` is still accessible/usable, but not linked from the toggle). I'll get back to this and see if any workaround.

southp (changes requested):
I've confirmed that:

* When the 1y/2y toggle is on, the behavior of /plans for the monthly plan owners match with what we've agreed on: the toggle looks untoggled to either side, and the monthly plan grid shows by default.
* When the 1y/2y toggle is off, /plans works the way it is.
* For an annual plan owner account, everything works either the 1y/2y toggle is on or off.

However, regarding this:
> Also not sure what/ if any caveats with "upgrading" to a lower tier plan for longer-term 
It is a forbidden behavior from the backend. Thus, when I attempt to "upgrade" from a 2y Premium plan to a 1y Business plan, the following error will show:
截圖 2023-03-31 下午4 36 53

The simplest solution I can think of now is to, once again, hide the toggle when a user is on a 2y plan. I'm wondering if we'd need to generalize it a bit, e.g. if the highest term the toggle poses is equal or lower to the user's plan, don't show the toggle. I'll leave that to your best judgement, of course; just thinking out loud here :)

chriskmnds:
Thanks, @southp.

> Also not sure what/ if any caveats with "upgrading" to a lower tier plan for longer-term
> It is a forbidden behavior from the backend. Thus, when I attempt to "upgrade" from a 2y Premium plan to a 1y Business plan, the following error will show:

(Indeed i thought about it but I guess my brain too limited to connect it to the 2y part 😆 )

> The simplest solution I can think of now is to, once again, hide the toggle when a user is on a 2y plan. I'm wondering if we'd need to generalize it a bit, e.g. if the highest term the toggle poses is equal or lower to the user's plan, don't show the toggle. I'll leave that to your best judgement, of course; just thinking out loud here :)

This might be the simplest way out indeed. However, also thinking about whether it makes sense to instead give users more visibility and a unified experience with this e.g. show the toggle to all, but change the active states of buttons (and their respective text - maybe to "contact support"?). so 1y, 2y, 3y users will see the same, just not be able to proceed with the "forbidden" resource.

client/my-sites/plans-features-main/plan-type-selector.tsx
```diff
@@ -199,26 +213,6 @@ export const IntervalTypeToggle: React.FunctionComponent< IntervalTypeProps > =
...
-				if ( 'treatment' !== experimentAssignment?.variationName ) {
-					return <></>;
-				}
-
-				return <IntervalTypeToggle { ...props } />;
-			} }
-		</ProvideExperimentData>
-	);
-};
-
```
chriskmnds: robots 😆

chriskmnds:
Thanks again for reviewing/testing. 😎 

> The simplest solution I can think of now is to, once again, hide the toggle when a user is on a 2y plan. I'm wondering if we'd need to generalize it a bit, e.g. if the highest term the toggle poses is equal or lower to the user's plan, don't show the toggle. I'll leave that to your best judgement, of course; just thinking out loud here :)

This should hopefully be handled now. Updated to do two things:

- Not show the biannual toggle when site has a plan > 1 year.
- Show a disabled "Contact support" button (instead of an active "Upgrade" button) when site is on a lower tier but higher term plan e.g. current site plan 2Y Premium and grid plan 1Y Business - Only relevant when visiting the paths directly e.g. `/plans/yearly/[SITE_2Y_PLAN]`.

Rebased onto https://github.com/Automattic/wp-calypso/pull/74722 to make use of the action button changes.

The proposed changes are also temporary until we have a decision on the toggle. The "showBiannualToggle" prop is restrictive, but refactoring now will mean touching code with the running experiment and redoing the new experiment code. I will create a follow-up issue to come back to this and make things more flexible to incorporate 3Y term in the toggle more easily.

cc @Automattic/martech-decepticons

southp (approved):
Thanks for the fix, @chriskmnds . I've re-confirmed all the combinations and they all work as expected. Please feel free to ship this after resolving the code conflicts :)

PR #74780, opened 2023-03-22
Newsletter: Change focus on subscribers and setup

### LINKED ISSUE #74583: Newsletter onboarding: Update input focus
Focus state is different in _Let's get started_ and subsequent pages with inputs.
The first one seems to be the right one, at least considering the [Figma file](eqlepL7MHifK8Bgr1OWqLr-fi-5807%3A8019).

escapemanuele:
On the newsletter flow, Focus state is different in user signup page and subsequent pages with inputs.

This updates the focus state on setup and subscriber steps.

Fixes https://github.com/Automattic/wp-calypso/issues/74583

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/components/setup-form/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/components/setup-form/style.scss
@@ -38,9 +38,9 @@
 		background-repeat: no-repeat;
 		background-position: 95%;
 		&:focus {
-			border-color: var(--wp-admin-theme-color);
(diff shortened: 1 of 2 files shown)


alshakero (changes requested):
Looks exactly as spec'd!

The error state in subscribers step looks strange. I'll ask Cris for feedback.

image

escapemanuele:
Maybe it should follow what is on the signup step that we took as an example for the focus state 🤔 

image

alshakero:
Yes I agree!

escapemanuele:
Ok @alshakero, should be better now!

image

alshakero (approved):
Super ✨

PR #74798, opened 2023-03-22
READER - Add reader notifications page

eoigal:
This PR aims to add the notifications panel, typically viewed from the bell icon in the admin bar, into a page on Reader.
Screenshot 2023-03-23 at 22 07 37

The existing notifications are rendered in an iframe that is hosted on a different domain widgets.wp.com. There is an issue with third party cookies working and as such, on some browsers like Safari, the notifications panel does not render any notes.

The workaround is to find a home for the notifications outside of the iframe and this PR will add them to reader.

This page will be accessible at https://wordpress.com/read/notifications

This PR loads the existing Notification component and modifies the styling to work within a reader page, rather than from a panel that works from the right side of page, sliding to the left.

Screenshot 2023-03-23 at 22 09 41

### Testing

* Apply patch
* Run `yarn & yarn start`
* Go to http://calypso.localhost:3000/read/notifications and confirm notifications load and work as expected


davemart-in:
If you are on /read/notifications should we disable the regular right sidebar notifications panel to avoid this:

CleanShot 2023-03-24 at 08 26 19

davemart-in:
I dropped a note in Slack about the line width of the notification bell icon looking wider than the rest of these icons.

Other than those two things, this is looking sharp!

eoigal:
> If you are on /read/notifications should we disable the regular right sidebar notifications panel to avoid this:
>

This sounds like a good idea to me

ollierozdarz:
@eoigal @davemart-in  👋 I've recreated our WP.com filled Bell icon but in a stroke format. I dropped the SVG in our Slack thread.

**Reference:**
Screenshot 2023-03-27 at 9 28 12 am

PR #74805, opened 2023-03-22
WoA Logging i1: Add tabs to the site logs page 1st pass

vykes-mac:
Closes https://github.com/Automattic/dotcom-forge/issues/1992

## Proposed Changes

* Add tab panel to choose between php and webserver logs
* Fetch data based on the selected tab

## Testing Instructions

- Verify tabs are rendered with `PHP Logs` selected
- Verify refreshing the browser maintains the current tab.

image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/site-logs/components/site-logs-tab-panel.tsx
+++ b/client/my-sites/site-logs/components/site-logs-tab-panel.tsx
@@ -8,11 +8,13 @@ export const tabs = [
 	{ name: 'web', title: __( 'Webserver Logs' ) },
 ];
 
+export type SiteLogsTab = 'php' | 'web';
+
 interface SiteLogsTabPanelProps {
 	children( tab: TabPanel.Tab ): JSX.Element;
+	selectedTab?: SiteLogsTab;
 	className?: string;
-	onSelected?: ( tab: string ) => void;
-	initialTab?: TabPanel.Tab;
+	onSelected?: ( tabName: string ) => void;
 }
 
 const LogsTabPanel = styled( TabPanel )`
@@ -38,15 +40,15 @@ const LogsTabPanel = styled( TabPanel )`
 
 export const SiteLogsTabPanel = ( {
 	children: renderContents,
+	selectedTab = 'php',
 	className,
 	onSelected,
-	initialTab = tabs[ 0 ],
 }: SiteLogsTabPanelProps ) => {
 	return (
 		<LogsTabPanel
+			initialTabName={ selectedTab }
 			className={ className }
 			tabs={ tabs }
-			initialTabName={ initialTab.name }
 			onSelect={ ( tabName ) => {
 				onTabSelected( tabName );
 				onSelected?.( tabName );
--- a/client/my-sites/site-logs/components/site-logs-tab-panel.tsx
+++ b/client/my-sites/site-logs/components/site-logs-tab-panel.tsx
@@ -1,0 +1,64 @@
+import styled from '@emotion/styled';
+import { TabPanel } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import page from 'page';
+
+export const tabs = [
+	{ name: 'php', title: __( 'PHP Logs' ) },
+	{ name: 'web', title: __( 'Webserver Logs' ) },
+];
+
+interface SiteLogsTabPanelProps {
+	children( tab: TabPanel.Tab ): JSX.Element;
+	className?: string;
+	onSelected?: ( tab: string ) => void;
+	initialTab?: TabPanel.Tab;
+}
+
+const LogsTabPanel = styled( TabPanel )`
+	.components-tab-panel__tabs {
+		overflow-x: auto;
+	}
+	.components-tab-panel__tabs-item {
+		--wp-admin-theme-color: var( --studio-gray-100 );
+		color: var( --studio-gray-60 );
+		padding: 0;
+		margin-right: 24px;
+		font-size: 16px;
+		flex-shrink: 0;
+	}
+	.components-tab-panel__tabs-item:hover,
+	.components-tab-panel__tabs-item.is-active,
+	.components-tab-panel__tabs-item.is-active:focus,
+	.components-tab-panel__tabs-item:focus:not( :disabled ) {
+		box-shadow: inset 0 -2px 0 0 var( --wp-admin-theme-color );
+		color: var( --studio-gray-100 );
+	}
+`;
+
+export const SiteLogsTabPanel = ( {
+	children: renderContents,
+	className,
+	onSelected,
+	initialTab = tabs[ 0 ],
+}: SiteLogsTabPanelProps ) => {
+	return (
+		<LogsTabPanel
+			className={ className }
+			tabs={ tabs }
+			initialTabName={ initialTab.name }
+			onSelect={ ( tabName ) => {
+				onTabSelected( tabName );
+				onSelected?.( tabName );
+			} }
+		>
+			{ ( tab ) => renderContents( tab ) }
+		</LogsTabPanel>
+	);
+};
+
+function onTabSelected( tabName: string ) {
+	const url = new URL( window.location.href );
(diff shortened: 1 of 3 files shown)


client/my-sites/site-logs/components/site-logs-tab-panel.tsx
```diff
@@ -0,0 +1,64 @@
...
+		>
+			{ ( tab ) => renderContents( tab ) }
+		</LogsTabPanel>
+	);
+};
+
+function onTabSelected( tabName: string ) {
+	const url = new URL( window.location.href );
+	url.searchParams.set( 'logType', tabName );
+	page.replace( url.pathname + url.search );
```
p-jackson, line 65: This line is what's causing the unit tests to fail. Which is a bit mysterious because we're running the tests with jsdom so it shouldn't have an issue 🤔 

If you add this right below the `import` statements in `site-logs/test/main.tsx` then it will turn the call to `page.replace` into a no-op for now.
```ts
// PageJS breaks the tests due to failing to read the document title.
// Mock it for now to avoid the error.
jest.mock( 'page' );
```

client/my-sites/site-logs/components/site-logs-tab-panel.tsx
```diff
@@ -0,0 +1,64 @@
...
+		box-shadow: inset 0 -2px 0 0 var( --wp-admin-theme-color );
+		color: var( --studio-gray-100 );
+	}
+`;
+
+export const SiteLogsTabPanel = ( {
+	children: renderContents,
+	className,
+	onSelected,
+	initialTab = tabs[ 0 ],
```
p-jackson: I don't think we need to call this sort of prop "initial". Initial makes me think that after receiving an initial prop the `<SiteLogsTabPanel>` will take over and own the state. But the state is fully controlled by the parent, so I'd just call this prop `selectedTab` or `currentSelection`.

I'd also just take a `tabName` as the prop type rather than a whole object. the `onSelected` callback just takes the tab name, and the state inside `<SiteLogs>` is just the tab name. I think it's more symmetrical to just use the `'php' | 'web'` string type for all of the things.

client/my-sites/site-logs/main.tsx
```diff
@@ -16,14 +17,24 @@ export function SiteLogs() {
...
 		start: startTime,
 		end: endTime,
 		sort_order: 'desc',
 		page_size: 10,
 	} );
 
+	const handleTabSelected = useCallback(
+		( tabName ) => {
+			setLogType( tabName );
+			refetch();
```
p-jackson: Is the explicit refetch necessary? When the value of `logType` changes it gets passed into `useSiteLogsQuery` and that should be enough to kick off a fresh request.

But maybe it makes sense to reset the `startTime` and `endTime` values each time a tab is selected?

client/my-sites/site-logs/main.tsx
```diff
@@ -16,14 +17,24 @@ export function SiteLogs() {
 	const [ startTime ] = useState( moment().subtract( 7, 'd' ).unix() );
 	const [ endTime ] = useState( moment().unix() );
 
-	const { data } = useSiteLogsQuery( siteId, {
-		logType: 'web',
+	const [ logType, setLogType ] = useState< 'php' | 'web' >( 'php' );
```
p-jackson: The query uses this `logType` state as the source of truth. The tab panel uses the query parameter as the source of truth. But I think they should both be using the same source of truth.

One option would be to have both the query and tab panel use the `logType` state, and the query param is only used on initial render as a way to look up what the initial selection should be.

```suggestion
	const [ logType, setLogType ] = useState< 'php' | 'web' >( () => {
		const queryParam = new URL( window.location.href ).searchParams.get( 'logType' );
		return [ 'php', 'web' ].includes( queryParam ) ? queryParam : 'php';
	 } );
```

client/my-sites/site-logs/main.tsx
```diff
@@ -52,8 +65,20 @@ export function SiteLogs() {
...
 				align="left"
 			/>
-			<h2>Web logs</h2>
-			<pre>{ formattedLogs }</pre>
+
+			<SiteLogsTabPanel
+				initialTab={ tab }
+				onSelected={ ( tabName ) => {
+					handleTabSelected( tabName );
+				} }
```
p-jackson: By passing an anonymous function to the `onSelected` prop we lose the advantage of `useCallback` because the prop value will change each render anyway. It's not a big deal because this probably isn't a performance bottleneck anyway, but we should either remove the `useCallback` or just pass `handleTabSelected` directly to `onSelected`

```suggestion
				onSelected={ handleTabSelected }
```

client/my-sites/site-logs/main.tsx
```diff
@@ -42,6 +53,8 @@ export function SiteLogs() {
 		.join( '\n' );
 
 	const titleHeader = __( 'Site Logs' );
+	const tabName = new URL( window.location.href ).searchParams.get( 'logType' ) ?? 'php';
```
p-jackson: Just a nitpick, but I'm a fan of using only lower case in the URL. Maybe `?log-type=php`?

p-jackson (changes requested):
Nice! I had actually imagined that the URL would look something like `/site-logs/php/:site` and `/site-logs/web/:site`, because that's how the tabs in the Marketing section work for example. But the query parameter approach could actually be good for us. Since the sidebar menu is defined in Jetpack, changing the link from `/site-logs` to `/site-logs/php` would be annoying.

Something I don't like so much is that the tabs aren't `<a>` tags. Since we're changing the URL it feels like they should be. But that looks like a limitation of the Gutenberg `<TabPanel>`, and it's probably best that we use the component from Core.

I've turned a bit anti-Emotion so I'm not sure whether we should actually use it in this project. Case-in-point, you're having to rely on nested selectors in this case anyway. But we do seem to use Emotion a bit in Build 🤷 so it's not a rule.

Unit tests need to be fixed before merging, and I also think it'd be good to use the `'php' | 'web'` data type everywhere instead of passing the complete tab object as a prop.

client/my-sites/site-logs/components/site-logs-tab-panel.tsx
vykes-mac: did try the `jest.mock` but it didn't seem to work.

vykes-mac: > I've turned a bit anti-Emotion so I'm not sure whether we should actually use it in this project. Case-in-point, you're having to rely on nested selectors in this case anyway. But we do seem to use Emotion a bit in Build 🤷 so it's not a rule.

I'm neither here nor there on this. We used it heavily in the Sites Management Page so it seemed were going that direction for any new components we adding to the codebase.

p-jackson: Don't feel like you have to change it for merging. But good to know you don't mind either way, means I might convert it to vanilla later if I have time 😈 IMO it was more of an experiment adding it to SMP. But I do see it in some new code.

p-jackson: > did try the jest.mock but it didn't seem to work.

Weird. I committed it myself to see what was going on but it seems to have fixed the tests.

vykes-mac: It seems issue was where I added it, I added it in the `describe` block.

client/my-sites/site-logs/main.tsx
vykes-mac: yeah this make sense.

p-jackson (approved):
Changes look great @vykes-mac! Re-tested and everything is working as expected including keyboard nav 😎 

Ready to merge!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7884039

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7884039).

PR #74830, opened 2023-03-23
Add user's email address to the Subscription Management portal

### LINKED ISSUE #74822: Display the user's email address for the non-WPCOM (external) user
## Description

The goal of this task is to display the user's email address in the subheading of the Subscription management portal for non-WPCOM users only.

Image

## Acceptance Criteria

- email address is displayed in the subheading based on the provided design - for non-WPCOM users only

---

Design: inDLaEQV8jJ21O4WXawIsJ-fi-712-29014
Related discussion: p1679501025614859/1679474258.545539-slack-C02TCEHP3HA

ivan-ottinger:
Fixes https://github.com/Automattic/wp-calypso/issues/74822.

The proposed changes add user's email address to the subheading of the Subscription Management portal for non-WPCOM users:

Markup on 2023-03-23 at 12:10:42

## Proposed Changes

* create helper `getSubkey` function
* introduce `useEmailAddress` hook that extracts user's email address from the `subkey` cookie
* introduce `useSubHeaderText` hook that outputs the subheading text based on whether the email address is available

## Testing Instructions

Checkout the branch related to this PR and build Calypso locally.

**Non-WPCOM user**

1. Trigger the "subscription management link" email by submitting the form on https://wordpress.com/email-subscriptions/.
2. Open the link in an Incognito browser window:
Markup on 2023-03-20 at 16:03:35
3. Open the browser inspector and copy the `subkey` cookie value as it is (without decoding it):
Markup on 2023-03-23 at 17:00:11
4. Navigate to http://calypso.localhost:3000/subscriptions/settings.
5. Open console and set the cookie by `document.cookie = "subkey=PASTED_COOKIE_VALUE";`.
6. Reload http://calypso.localhost:3000/subscriptions/settings → the correct email address should display in the subheading.

**WPCOM logged-in user**

If you navigate to http://calypso.localhost:3000/subscriptions/settings as a logged-in WPCOM user, the `subkey` cookie is not available and hence the subheading will be as follows: "Manage your WordPress.com newsletter and blog subscriptions. (the original subheading copy)":

Markup on 2023-03-23 at 17:09:54

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/reader/helpers/index.ts
+++ b/packages/data-stores/src/reader/helpers/index.ts
@@ -1,4 +1,5 @@
 import apiFetch, { APIFetchOptions } from '@wordpress/api-fetch';
+import cookie from 'cookie';
 import wpcomRequest from 'wpcom-proxy-request';
 
 type callApiParams = {
@@ -10,11 +11,7 @@ type callApiParams = {
 
 // Get cookie named subkey
 const getSubkey = () => {
-	const subkey = document.cookie
-		?.split( ';' )
-		?.map( ( c ) => c.trim() )
-		?.find( ( c ) => c.startsWith( 'subkey=' ) )
-		?.split( '=' )[ 1 ];
+	const subkey = cookie.parse( document.cookie )?.subkey;
 	return subkey;
 };
 
--- a/packages/data-stores/src/reader/helpers/index.ts
+++ b/packages/data-stores/src/reader/helpers/index.ts
@@ -8,6 +8,16 @@ type callApiParams = {
 	isLoggedIn?: boolean;
 };
 
+// Get cookie named subkey
+const getSubkey = () => {
+	const subkey = document.cookie
+		?.split( ';' )
+		?.map( ( c ) => c.trim() )
+		?.find( ( c ) => c.startsWith( 'subkey=' ) )
+		?.split( '=' )[ 1 ];
+	return subkey;
+};
+
 // Helper function for fetching from subkey authenticated API. Subkey authentication process is only applied in case of logged-out users.
 async function callApi< ReturnType >( {
 	path,
@@ -25,12 +35,7 @@ async function callApi< ReturnType >( {
 		return res as ReturnType;
 	}
 
-	// get cookie named subkey
-	const subkey = document.cookie
-		?.split( ';' )
-		?.map( ( c ) => c.trim() )
-		?.find( ( c ) => c.startsWith( 'subkey=' ) )
-		?.split( '=' )[ 1 ];
+	const subkey = getSubkey();
 
 	if ( ! subkey ) {
 		throw new Error( 'Subkey not found' );
@@ -50,4 +55,4 @@ async function callApi< ReturnType >( {
 	} as APIFetchOptions );
 }
 
-export { callApi };
+export { callApi, getSubkey };
--- a/packages/data-stores/src/reader/hooks/index.ts
+++ b/packages/data-stores/src/reader/hooks/index.ts
@@ -1,5 +1,7 @@
 import { useSelect } from '@wordpress/data';
+import { useMemo } from '@wordpress/element';
 import { register as registerUserStore } from '../../user';
+import { getSubkey } from '../helpers';
 import type { UserSelect } from '../../user';
 
 const USER_STORE = registerUserStore( { client_id: '', client_secret: '' } );
@@ -18,4 +20,25 @@ export const useIsQueryEnabled = () => {
 	}
 
 	return false;
+};
+
+// Get subscriber's email address based on the subkey cookie
+export const useSubscriberEmailAddress = () => {
+	return useMemo( () => {
+		const subkey = getSubkey();
+
+		if ( ! subkey ) {
+			return null;
+		}
+
+		const decodedSubkeyValue = decodeURIComponent( subkey );
+
+		const firstPeriodIndex = decodedSubkeyValue.indexOf( '.' );
+		if ( firstPeriodIndex === -1 ) {
+			return null;
+		}
+
+		const emailAddress = decodedSubkeyValue.slice( firstPeriodIndex + 1 );
+		return emailAddress;
+	}, [] );
 };
--- a/packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
+++ b/packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
@@ -2,7 +2,7 @@
 /**
  * External dependencies
  */
-import { useTranslate } from 'i18n-calypso';
+import { useTranslate, LocalizeProps } from 'i18n-calypso';
 import DocumentHead from 'calypso/components/data/document-head';
 import FormattedHeader from 'calypso/components/formatted-header';
 import Main from 'calypso/components/main';
@@ -34,27 +34,27 @@ const getEmailAddress = () => {
 	return emailAddress;
 };
 
+const getSubHeaderText = ( translate: LocalizeProps[ 'translate' ] ) => {
+	const emailAddress = getEmailAddress();
+	if ( emailAddress ) {
+		return translate(
+			"Manage the WordPress.com newsletter and blogs you've subscribed to with {{span}}%(emailAddress)s{{/span}}.",
+			{
+				args: {
+					emailAddress: emailAddress,
+				},
+				components: {
+					span: <span className="email-address" />,
+				},
+			}
+		);
+	}
+
+	return translate( 'Manage your WordPress.com newsletter and blog subscriptions.' );
+};
+
 const SubscriptionManagerContainer = ( { children }: SubscriptionManagerContainerProps ) => {
 	const translate = useTranslate();
-
-	const getSubHeaderText = () => {
-		const emailAddress = getEmailAddress();
-		if ( emailAddress ) {
-			return translate(
-				"Manage the WordPress.com newsletter and blogs you've subscribed to with {{span}}%(emailAddress)s{{/span}}.",
-				{
-					args: {
-						emailAddress: emailAddress,
-					},
-					components: {
-						span: <span className="email-address" />,
-					},
-				}
-			);
-		}
-
-		return translate( 'Manage your WordPress.com newsletter and blog subscriptions.' );
-	};
 
 	return (
 		<Main className="subscription-manager-container">
@@ -62,7 +62,7 @@ const SubscriptionManagerContainer = ( { children }: SubscriptionManagerContaine
 			<FormattedHeader
 				brandFont
 				headerText={ translate( 'Subscription management' ) }
-				subHeaderText={ getSubHeaderText() }
+				subHeaderText={ getSubHeaderText( translate ) }
 				align="left"
 			/>
 			{ children }
--- a/packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
+++ b/packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
@@ -3,7 +3,8 @@
  * External dependencies
  */
 import { getSubkey } from '@automattic/data-stores/src/reader/helpers';
-import { useTranslate, LocalizeProps } from 'i18n-calypso';
+import { useTranslate } from 'i18n-calypso';
+import { useMemo } from 'react';
 import DocumentHead from 'calypso/components/data/document-head';
 import FormattedHeader from 'calypso/components/formatted-header';
 import Main from 'calypso/components/main';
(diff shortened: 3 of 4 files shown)


packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,18 +12,57 @@ export type SubscriptionManagerContainerProps = {
 	children?: React.ReactNode;
 };
 
+const getEmailAddress = () => {
+	const subkey: string | undefined = document.cookie
```
ivan-ottinger: Would you extract out the type of subkey here?

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,18 +12,57 @@ export type SubscriptionManagerContainerProps = {
...
+	}
+
+	const emailAddress = decodedSubkeyValue.slice( firstPeriodIndex + 1 );
+	return emailAddress;
+};
+
 const SubscriptionManagerContainer = ( { children }: SubscriptionManagerContainerProps ) => {
 	const translate = useTranslate();
 
+	const getSubHeaderText = () => {
```
ivan-ottinger: Would you turn the `getSubHeaderText` into a callback?

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,18 +12,57 @@ export type SubscriptionManagerContainerProps = {
 	children?: React.ReactNode;
 };
 
+const getEmailAddress = () => {
+	const subkey: string | undefined = document.cookie
+		?.split( ';' )
+		?.map( ( c ) => c.trim() )
+		?.find( ( c ) => c.startsWith( 'subkey=' ) )
+		?.split( '=' )[ 1 ];
```
ivan-ottinger: I am using the same logic as @TimBroddin introduced in `packages/data-stores/src/reader/helpers/index.ts`.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,18 +12,57 @@ export type SubscriptionManagerContainerProps = {
 	children?: React.ReactNode;
 };
 
+const getEmailAddress = () => {
```
ivan-ottinger: This is currently getting email from the cookie only, ignoring whether the user is logged in our not.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
TimBroddin: I might move this out of the component, since it doesn't depend on any props or state from inside.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
TimBroddin: We might want to split this off into its own helper function.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
TimBroddin: Yes, indeed. One note: there's no need to type this variable, TypeScript can infer the type from the usage:

CleanShot 2023-03-23 at 12 28 37@2x

ivan-ottinger: Thanks, Tim! Addressed in https://github.com/Automattic/wp-calypso/pull/74830/commits/282d369bffb0104b17a42c61b4d5bf9bb85937e6.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
ivan-ottinger: Makes sense. I have moved it out of the component in https://github.com/Automattic/wp-calypso/pull/74830/commits/c6282ae8da5186e657faa5a9458864dfd5d92b82.

saygunnyc:
Not sure if this has already been resolved, but I wanted to mention again that this UI is only meant for folks who do not have a WordPress.com account. The Reader version of this interface does not have an email address, as the user is already logged into their WP.COM account, and if they are willing to change their email, that's something they should do on their account level.

Let me know if this doesn't make sense and if there are any remaining questions!

ivan-ottinger:
> Not sure if this has already been resolved, but I wanted to mention again that this UI is only meant for folks who do not have a WordPress.com account. The Reader version of this interface does not have an email address, as the user is already logged into their WP.COM account, and if they are willing to change their email, that's something they should do on their account level.

> Let me know if this doesn't make sense and if there are any remaining questions!

Thank you for sharing your point of view, @saygunnyc.

I know we discussed the actual email form field yesterday - that it should only display to the non-WPCOM users. What I wasn't 100% sure in the p1679563253901539-slack-C02TCEHP3HA discussion though was whether this applies to the statically displayed email address as well:

Markup on 2023-03-23 at 12:10:42

The way I understand it is that this updated subheading should be visible to the non-WPCOM users only. Is that correct?

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
ivan-ottinger: Thank you, Tim! I have created a new helped function in https://github.com/Automattic/wp-calypso/pull/74830/commits/da0a43c81ca383e8fcf812da22cd5e5432019793.

saygunnyc:
> The way I understand it is that this updated subheading should be visible to the non-WPCOM users only. Is that correct?

Yes, because this interface shouldn't exist for WPCOM users. It should be within the Reader _– Manage followed sites._ where the user already has profile/account settings for such reasons. Does it make sense?

Screenshot 2023-03-23 at 10 57 23 AM

ivan-ottinger:
> Yes, because this interface shouldn't exist for WPCOM users. It should be within the Reader – Manage followed sites. where the user already has profile/account settings for such reasons. Does it make sense?

Yes, it does make sense. Thanks! It now also clarifies that part of the discussion we had yesterday. 🙂

ivan-ottinger:
Based on this discussion (p1679586336703769/1679584642.397649-slack-C02TCEHP3HA) we may want to _temporarily_ let the logged-in WPCOM users into the "interim" portal (that includes the Settings tab).

Would you say, @saygunnyc, that it is alright to just display the original subheading copy to these users?

Markup on 2023-03-23 at 17:09:54

(of course, non-WPCOM users will still see their email address)

@yansern in case you would like to join the discussion. 🙂 

Thank you folks!

saygunnyc:
Yes, 💯 !

ivan-ottinger:
This PR is now ready for review.

packages/data-stores/src/reader/helpers/index.ts
```diff
@@ -8,6 +8,16 @@ type callApiParams = {
 	isLoggedIn?: boolean;
 };
 
+// Get cookie named subkey
+const getSubkey = () => {
+	const subkey = document.cookie
```
phcp: This could be changed to use the cookie library, something like:
```
const cookies = cookie.parse( document.cookie );
return cookies.subkey;
```

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,6 +13,43 @@ export type SubscriptionManagerContainerProps = {
...
+	const decodedSubkeyValue = decodeURIComponent( subkey );
+
+	const firstPeriodIndex = decodedSubkeyValue.indexOf( '.' );
+	if ( firstPeriodIndex === -1 ) {
+		return null;
+	}
+
+	const emailAddress = decodedSubkeyValue.slice( firstPeriodIndex + 1 );
+	return emailAddress;
+};
```
phcp: I see this function as a hook like `useSubscriberEmailAddress`, using useMemo to avoid processing the string on every render. It could be either placed on this file or on  a hooks.ts file (in this case it could be reused by other components if needed). What do you think?

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
```diff
@@ -12,6 +13,43 @@ export type SubscriptionManagerContainerProps = {
...
+
+const getSubHeaderText = ( translate: LocalizeProps[ 'translate' ] ) => {
+	const emailAddress = getEmailAddress();
```
phcp: This one could also be converted to a hook, in this case it would be able to call both `useSubscriberEmailAddress` and `useTranslate` (so no need to pass translate as a parameter). It could take advantage of useMemo as well, listening to changes on emailAddress.

phcp (commented):
PR work as expected, just added a few suggestions to the code.

image

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
TimBroddin: Good point!

I added some additional thoughts on an `useExternalUser` hook [here](https://github.com/Automattic/wp-calypso/pull/74861#issuecomment-1482360662)

packages/data-stores/src/reader/helpers/index.ts
ivan-ottinger: Thanks for the suggestion, Paulo! I will implement the change.

ivan-ottinger: Addressed in https://github.com/Automattic/wp-calypso/pull/74830/commits/cd016de255b8afd160a4a1c68e9c792347beea59.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
ivan-ottinger: Thanks for the suggestions folks!

> [...] [refactor](https://github.com/Automattic/wp-calypso/pull/74733) will bring in a SubscriptionManagementContext which we can use to store application state. I suggest we create a hook that returns the email address together with a verified key, which by default is false, but switches to true after the first API request succeeds.

Since the refactor is not ready yet, I would convert both `getSubHeaderText` and `getEmailAddress` into hooks - so we have the email address display ready for the Monday launch and then this logic can be later replaced by the data from the context (once it is ready). How does that sound to you?

ivan-ottinger: Both functions are now hooks.

packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx
ivan-ottinger: Thanks, Paulo. That makes sense. I have applied the changes.

packages/data-stores/src/reader/hooks/index.ts
```diff
@@ -1,5 +1,7 @@
 import { useSelect } from '@wordpress/data';
+import { useMemo } from '@wordpress/element';
```
ivan-ottinger, line 2: Does that make any difference that we import useMemo from `@wordpress/element` as opposed to `react`?

TimBroddin: I think it's considered best practice to import from `@wordpress/element`. It's just a thin abstraction layer that re-exports the React exports. Should Calypso choose to migrate to another library that's compatible with the React API (such as preact), this would smoothen the migration.

ivan-ottinger: Thank you, Tim.

JuanLucha (approved):
Nice job 🚀

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7884827

Thank you @ivan-ottinger for including a screenshot in the description! This is really helpful for our translators.

ivan-ottinger:
Looks like `packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx` isn't included in the watched paths for translation.

→ Asked i18n if they could add it there (p1679903418201549-slack-C02AED43D). 🙂

ivan-ottinger:
✅ `packages/subscription-manager/src/components/SubscriptionManagerContainer/SubscriptionManagerContainer.tsx` is now included in the watched paths.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7884827).

PR #65414, opened 2022-07-08
Add checkbox to accept third party developers account creation

epeicher:
#### Proposed Changes

* This PR adds a required checkbox to the checkout process that is displayed if a _Marketplace_ product is in the cart. In that case, the Submit button does not submit the purchase until the user checks the checkbox.

Further context can be found here: pdt2If-e4-p2 and in the following comment: pdt2If-e4-p2#comment-150 and the final result in this comment pdt2If-e4-p2#comment-160

|Before Changes | After Changes Un-checked|After Changes Un-checked clicking the `Pay` button|  After Changes Checked|
|-------|------|------|-------|
|CleanShot 2022-07-20 at 11 31 50@2x|    CleanShot 2022-07-26 at 08 42 25@2x       |      CleanShot 2022-07-26 at 08 46 33@2x    |     CleanShot 2022-07-26 at 08 47 53@2x     |

##### Test 1

1. Navigate to the plugins page of a site that allows buying Plugins (Pro, Business or eCommerce)
2. Click on a Marketplace plugin, for example `/plugins/automatewoo/:siteId`
3. Click on `Purchase and activate`
4. Ensure there is a checkbox on top of the `Pay` button
5. Click on the `Pay` button
7. Ensure the checkout process is not completed, the checkbox above is highlighted in blue (a blue border is added to the checkbox) and there is an error message indicating that `The terms above need to be accepted`
8. Check the checkbox
9. Ensure the blue highlighting border is removed and the error message disappears.
10. Un-check the checkbox again
11. Step 7 again, highlighting in blue and error message is displayed
12. Check the checkbox 
13. Step 9 again, highlighting and error message disappear.
14. Click on `Pay` and make sure the plugin is installed as expected.

##### Test 2

1. Navigate to a Site and Upgrade the Plan, for example from Business to eCommerce, e.g. `plans/:siteId`
2. Ensure there is not any checkbox and the `Pay` button is enabled when the required data is populated
3. Go back or select the top-left X button but when prompted select `Leave items` to leave the plan upgrade in the cart
4. Repeat **Test 1** above. Ensure that with both products, a Plan Upgrade and a Marketplace item in the Cart, the checkbox is working as expected.
5. Remove the Marketplace plugin from the cart.
6. Ensure the checkbox is **not** displayed and the `Pay` button works as expected.

---

Please @Automattic/editorial , could you please review the wording of the messages? 

The following messages have been added:

In the checkbox label:
[...]

PR #74850, opened 2023-03-23
composite-checkout: Make sure that the return type of submit button onClick is always a Promise

sirbrillig:
Every payment method object passed into a `CheckoutProvider` must supply a submit button component. That component is then rendered by the `CheckoutSubmitButton` which is rendered as part of every form that uses the provider. `CheckoutSubmitButton` injects a `onClick` callback prop into each submit button component which hooks that button into the transaction flow that the `@automattic/composite-checkout` package provides (specifically, it calls the payment processor function that the provider has assigned to that payment method). The return value of that callback is the same as the return value of the hook that creates it, `useProcessPayment()`, which is `Promise<PaymentProcessorResponse>`.

However, the `onClick` prop is wrapped in another function before it is injected which allows `CheckoutSubmitButton` to run a validation callback before it calls the payment processor. That wrapper has two cases where it can return `undefined` instead of a Promise. This was overlooked because most payment methods do not use the return value of `onClick`, but `WeChatPayButton` does, and this results in a fatal error (see p1679556359011019-slack-C04U5A26MJB).

This issue was created by https://github.com/Automattic/wp-calypso/pull/65414

## Proposed Changes

In this PR we modify `CheckoutSubmitButton` so that its return value is the same as the return value of the `onClick` method it wraps.

## Testing Instructions

First, modify calypso so that the checkout submit button's validation callback will always return false:

```diff
diff --git a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
index a9c6a07dc0..9b15f82882 100644
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -244,10 +244,7 @@ export default function WPCheckout( {
 
 	const validateForm = async () => {
 		setIsSubmitted( true );
-		if ( hasMarketplaceProduct && ! is3PDAccountConsentAccepted ) {
-			return false;
-		}
-		return true;
+		return false;
 	};
 
 	if ( ! checkoutActions ) {
```

Next, add a product to your cart and visit checkout.

Complete the checkout form and press the submit button.

Verify that nothing happens when the button is clicked. (If this were a real failure a different mechanism would display the error; see https://github.com/Automattic/wp-calypso/pull/65414.)

### DIFF
--- a/client/my-sites/checkout/composite-checkout/payment-methods/wechat/index.tsx
+++ b/client/my-sites/checkout/composite-checkout/payment-methods/wechat/index.tsx
@@ -21,7 +21,11 @@ import {
 } from 'calypso/my-sites/checkout/composite-checkout/components/summary-details';
 import useCartKey from 'calypso/my-sites/checkout/use-cart-key';
 import WeChatPaymentQRcodeUnstyled from './wechat-payment-qrcode';
-import type { LineItem, ProcessPayment } from '@automattic/composite-checkout';
+import type {
+	LineItem,
+	PaymentMethod,
+	PaymentMethodSubmitButtonProps,
+} from '@automattic/composite-checkout';
 import type {
 	PaymentMethodStore,
 	StoreSelectors,
@@ -79,7 +83,7 @@ export function createWeChatMethod( {
 }: {
 	store: WeChatStore;
 	siteSlug?: string;
-} ) {
+} ): PaymentMethod {
 	return {
 		id: 'wechat',
 		paymentProcessorId: 'wechat',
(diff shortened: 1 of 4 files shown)


spraveenitpro:
LGTM

<details>
<summary>Functional testing:</summary>

1. As expected the pay button did not work, it behaved like this 
</details>

spraveenitpro (approved):
I did some static analysis, from what I am seeing it looks good to me, nothing specifically stood out.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7884977

Hi @sirbrillig, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `This payment method is not currently available.`
- `The information requried by this payment method is not valid.`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7884977).

PR #74852, opened 2023-03-23
Fix longreads avatar issues

davemart-in:
## Description

Within the reader, whenever Peter publishes a Longreads post his personal avatar is showing up. It would be preferable if the longreads site logo always showed.

### Before

before

before-post

### After

after

after-post

## Testing Instructions

- Use the Calypso preview link (or apply this PR)
- Visit http://calypso.localhost:3000/read/feeds/138734090
- Visit http://calypso.localhost:3000/read/feeds/138734090/posts/4617747083

### DIFF
--- a/client/blocks/author-compact-profile/index.jsx
+++ b/client/blocks/author-compact-profile/index.jsx
@@ -53,6 +53,7 @@ class AuthorCompactProfile extends Component {
 			'has-author-icon': siteIcon || feedIcon || author.has_avatar,
 		} );
 		const streamUrl = getStreamUrl( feedId, siteId );
+		const isLongreads = siteId === 211646052;
 
 		// If we have a feed URL, use that for the follow button in preference to the site URL
 		const followUrl = feedUrl || siteUrl;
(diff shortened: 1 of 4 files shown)


eoigal (commented):
I don't think it should be necessary to create a new `isLongreads` property to handle this - it looks excessive.

davemart-in:
@eoigal any suggestions?

creativecoder:
I think a better fix for this would be forcing `null` for the author of each post for this blog. But perhaps we don't need to? (pe7F0s-Hp-p2#comment-753)

creativecoder:
> null for the author of each post for this blog

To be clear, I mean doing that server-side.

roo2:
It looks like this is resolved by Peter disconnecting his wpcom account, I think it was only an issue because all of the other longreads authors did not have connected wpcom accounts so he believed that there was a problem with his account showing its avatar. pe7F0s-Hp-p2#comment-753

But in general it seems like the current behavior is correct. Unless we want to change it for all sites.

roo2:
I'll close this if no objections

davemart-in:
I'm just glad it is resolved. Thanks for closing this one @roo2!

PR #74861, opened 2023-03-23
Unauthorized users should be redirected to subscriptions landing page.

### LINKED ISSUE #74849: Unauthorized users should be redirected from `/subscriptions/settings` to `/email-subscriptions`
Unauthorized users that go to `wordpress.com/subscriptions/settings` are presented with this screen:

Image

That screen is located in Calypso, with a `loading` spinning circle in an infinite loop.

The correct behavior is to redirect the unauthorized  to the landing page `wordpress.com/email-subscriptions` so they can ask for the logging URL and continue with the Subscriptions flow:

Image

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/74849

## Proposed Changes

If an unauthorized user goes to `http://calypso.localhost:3000/subscriptions/settings` for any reason (bookmark, for example), they will see an infinite spinner like this:

Screenshot 2023-03-23 at 17 07 56

That's because that settings page lives in Calypso, and the unauthorized user doesn't have a cookie set for Calypso to work.

The correct behavior is the user being redirected to the initial screen in the Subscription Management flow:

Screenshot 2023-03-23 at 17 16 11

### Non-WPCOM user, without verifying
1. Apply this branch.
2. In an incognito window in your browser, go to `http://calypso.localhost:3000/subscriptions/settings`.
3. You should be redirected to `http://calypso.localhost:3000/email-subscriptions`. That page will show the following:
Screenshot 2023-03-23 at 23 31 24
This is because `/email-subscriptions` doesn't leave in Calypso. In order to check that the redirection shows the correct screen, replace `http://calypso.localhost:3000` with `https://wordpress.com`

### Non-WPCOM user, verified
1. Trigger the "subscription management link" email by submitting the form at https://wordpress.com/email-subscriptions/ using an incognito window.
2. Open the link in an Incognito browser window:
image
3. Open the browser inspector and copy the subkey cookie value as it is (without decoding it):
image
4. Navigate to http://calypso.localhost:3000/subscriptions/settings.
5. Open the console and set the cookie by document.cookie = "subkey=PASTED_COOKIE_VALUE";.
6. Reload http://calypso.localhost:3000/subscriptions/settings → the correct email address should display in the subheading.

### WPCOM logged-in user

If you navigate to `http://calypso.localhost:3000/subscriptions/settings` as a logged-in WPCOM user, you will be redirected to `http://calypso.localhost:3000/read`, which is the user interface used for that kind of user.

### DIFF
--- a/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
+++ b/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable no-restricted-imports */
 import { Spinner } from '@automattic/components';
 import { Reader } from '@automattic/data-stores';
+import { useSubscriberEmailAddress } from '@automattic/data-stores/src/reader/hooks';
 import { useState, useCallback, useEffect } from '@wordpress/element';
 import { translate } from 'i18n-calypso';
 import page from 'page';
@@ -42,12 +43,15 @@ const UserSettings = ( { value = {}, loading, userId }: UserSettingsProps ) => {
 	const { mutate, isLoading, isSuccess, error } =
 		Reader.useSubscriptionManagerUserSettingsMutation();
 	const [ notice, setNotice ] = useState< NoticeProps | null >( null );
+	const subscriberEmailAddress = useSubscriberEmailAddress();
 
 	useEffect( () => {
-		if ( ! userId ) {
+		if ( userId ) {
+			page.redirect( '/read' );
+		} else if ( ! subscriberEmailAddress ) {
 			page.redirect( '/email-subscriptions' );
 		}
-	}, [ userId ] );
+	}, [ userId, subscriberEmailAddress ] );
 
 	useEffect( () => {
 		// check if formState is empty object
--- a/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
+++ b/packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
@@ -1,8 +1,12 @@
+/* eslint-disable no-restricted-imports */
 import { Spinner } from '@automattic/components';
 import { Reader } from '@automattic/data-stores';
 import { useState, useCallback, useEffect } from '@wordpress/element';
 import { translate } from 'i18n-calypso';
+import page from 'page';
 import { FormEvent } from 'react';
+import { connect } from 'react-redux';
+import { getCurrentUserId } from 'calypso/state/current-user/selectors';
 import { Button } from '../Button';
 import { Notice } from '../Notice';
 import { BlockEmailsSetting } from '../fields/BlockEmailsSetting';
@@ -24,6 +28,7 @@ type SubscriptionUserSettings = Partial< {
 type UserSettingsProps = {
 	value?: SubscriptionUserSettings;
 	loading: boolean;
+	userId: number;
 };
 
 type NoticeProps = {
@@ -32,11 +37,17 @@ type NoticeProps = {
 };
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- until we start using any of these props
-const UserSettings = ( { value = {}, loading }: UserSettingsProps ) => {
+const UserSettings = ( { value = {}, loading, userId }: UserSettingsProps ) => {
 	const [ formState, setFormState ] = useState< SubscriptionUserSettings >( value );
 	const { mutate, isLoading, isSuccess, error } =
 		Reader.useSubscriptionManagerUserSettingsMutation();
 	const [ notice, setNotice ] = useState< NoticeProps | null >( null );
+
+	useEffect( () => {
+		if ( ! userId ) {
+			page.redirect( '/email-subscriptions' );
+		}
+	}, [ userId ] );
 
 	useEffect( () => {
 		// check if formState is empty object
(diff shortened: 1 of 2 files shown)


yansern:
Unauthorized not external users should be redirected to the landing page, meaning users who:
- do not have a subkey cookie AND
- is not logged in to wordpress

I think you can use `getSubkey()` to check if the cookie exists.

yansern:
@TimBroddin Do you think a `useExternalUser` hook would be useful? 

I see @phcp suggested something similar like `useSubscriberEmailAddress` in [this other PR](https://github.com/Automattic/wp-calypso/pull/74830/files#r1146552329) but I would interpret that hook as an abstraction hook that returns email for both external & logged-in users.

TimBroddin:
> @TimBroddin Do you think a `useExternalUser` hook would be useful?
> 
> I see @phcp suggested something similar like `useSubscriberEmailAddress` in [this other PR](https://github.com/Automattic/wp-calypso/pull/74830/files#r1146552329) but I would interpret that hook as an abstraction hook that returns email for both external & logged-in users.

No, it's not an abstraction. it would only return it for external users, because the only place we need to show this is the external portal. 

As for the hook, I've been thinking about it, and the problem right now is that it just takes the value from the cookie and displays it. If you put `stevejobs@apple.com` in your cookie, the header will tell you that you are indeed Steve Jobs. It's not a huge problem in this use case (displaying a simple help text), but might become one if other people start consuming this hook without knowking how it's implemented.

@mashikag's [refactor](https://github.com/Automattic/wp-calypso/pull/74733) will bring in a `SubscriptionManagementContext` which we can use to store application state. I suggest we create a hook that returns the email address together with a verified key, which by default is false, but switches to true after the first API request succeeds.

useexternaluser

 If this is the approach we want to use, I'll create an issue for it, and I can implement it when @mashikag 's PR lands

yansern:
> @mashikag's https://github.com/Automattic/wp-calypso/pull/74733 will bring in a SubscriptionManagementContext which we can use to store application state. I suggest we create a hook that returns the email address together with a verified key, which by default is false, but switches to true after the first API request succeeds.

It's very subtle, but I can see from your drawing what you meant by "first API request", in your illustration it's the count API request. So I get entirely that you mean.

And as discussed on the Slack thread p1679651957226219/1679651639.332759-slack-C02TCEHP3HA, we'll get a version out that just parses the cookie and return a usable object.

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -32,11 +39,20 @@ type NoticeProps = {
...
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- until we start using any of these props
-const UserSettings = ( { value = {}, loading }: UserSettingsProps ) => {
+const UserSettings = ( { value = {}, loading, userId }: UserSettingsProps ) => {
 	const [ formState, setFormState ] = useState< SubscriptionUserSettings >( value );
 	const { mutate, isLoading, isSuccess, error } =
 		Reader.useSubscriptionManagerUserSettingsMutation();
 	const [ notice, setNotice ] = useState< NoticeProps | null >( null );
+	const subscriberEmailAddress = useSubscriberEmailAddress();
+
+	useEffect( () => {
```
phcp, line 49: This redirect rule seems to be related with the Subscription Manager flow, not only to the UserSettings, in this case maybe SubscriptionManagementPage would be a better place to keep this redirect logic. What do you think?

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -32,11 +39,20 @@ type NoticeProps = {
...
+		if ( userId ) {
```
phcp, line 50: The `userId` could be replaced by `isUserLoggedIn` selector, it has similar implementation and it's easier to consume.

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -32,11 +39,20 @@ type NoticeProps = {
...
+			page.redirect( '/read' );
+		} else if ( ! subscriberEmailAddress ) {
+			page.redirect( '/email-subscriptions' );
```
phcp, line 53: Looks like the page.redirect is just performed after the end of the function, with that the UserSettings component is rendered before the redirect. I'd suggest to do an early return with an empty fragment to avoid the first render. 

```
	if ( userId ) {
		page.redirect( '/read' );
		return <></>;
	} else if ( ! subscriberEmailAddress ) {
		page.redirect( '/email-subscriptions' );
		return <></>;
	}
```

phcp (commented):
The redirect works as intended, just added a few suggestions.

packages/subscription-manager/src/components/UserSettings/UserSettings.tsx
```diff
@@ -32,11 +39,20 @@ type NoticeProps = {
...
+const UserSettings = ( { value = {}, loading, userId }: UserSettingsProps ) => {
 	const [ formState, setFormState ] = useState< SubscriptionUserSettings >( value );
 	const { mutate, isLoading, isSuccess, error } =
 		Reader.useSubscriptionManagerUserSettingsMutation();
 	const [ notice, setNotice ] = useState< NoticeProps | null >( null );
+	const subscriberEmailAddress = useSubscriberEmailAddress();
+
+	useEffect( () => {
+		if ( userId ) {
+			page.redirect( '/read' );
```
yansern, line 51: I'm not so sure if we should redirect logged-in users to reader yet. Let's just keep ourselves in the mind frame of the old Subscriptions Management first.

yansern:
Closing this as we'll work on this here. https://github.com/Automattic/wp-calypso/issues/74886

PR #74862, opened 2023-03-23
Logging: Render site log data in a table

p-jackson:
Fixes Automattic/dotcom-forge#1989

## Proposed Changes

* Renders log entries in a table
* Scrolls horizontally
* Changes the Sites Log section to a "full width" layout
* `date`/`timestamp` column (depending on the log type) appears as first column
* The first column always appears in the site's configured timezone

CleanShot 2023-03-24 at 18 11 05@2x

I decided to use the site's timezone (rather than the current user's timezone) because I thought it might be nicer to have consistency between team members. For example if an HE is helping a user track down a bug, it might be better for everyone to be using the same frame of reference when talking.
Whichever timezone is used personally I prefer when the timezone is explicitly displayed too. To avoid confusion.

## Testing Instructions

* Check out locally
* Head to `/site-logs/:siteSlug` for a WoA site
* Take a look at the table
* Acceptance criteria in Automattic/dotcom-forge#1989 should be complete
* You can change the site's timezone at `/settings/general/:siteSlug`

## Pre-merge Checklist


client/data/hosting/use-site-logs-query.ts
```diff
@@ -5,12 +5,12 @@ interface SiteLogsAPIResponse {
 	message: string;
 	data: {
 		total_results: number;
-		logs: object[];
+		logs: Record< string, any >[];
```
mpkelly, line 8: It's always a `string` or a `number`, so maybe can narrow this down to `Record< string, string | number >` here

mpkelly (approved):
Tables now!? It's really getting fancy. What next, white panel backgrounds? Icons?

I think it's good. The time zone updated according to my settings 👍  

I think a more compact time format and sans-serif font would be a good change as this column eats a quite a bit of space: 

image

Also, you could show the local time in a tooltip, e.g. `<td title={localTime}>...</td>`

client/data/hosting/use-site-logs-query.ts
p-jackson: I wanted to protect against any future API changes in `renderCell()` by making sure we fell back back to something like `JSON.stringify` if we encountered something unexpected. We only get strings and numbers for now but it doesn't really document what the type is going to be.

But I probably should have gone with `unknown` instead of `any`. Since `unknown` will make sure we do checking before we use the value.

PR #74868, opened 2023-03-24
Logging: Move Site Logs tab type definition to `client/data/hosting`

p-jackson:
## Proposed Changes

Refactors the `'php' | 'web'` data type, which is going to be used by a lot of our logging components, to the `client/data/hosting` folder. This is the lowest-common-denominator for places that use this type, so everyone should import the type from here.

## Testing Instructions

* Pure refactor, there should be no change


mpkelly (approved):
All good. LGTM

PR #74896, opened 2023-03-24
Woo Express: The state didn't update after upgrading an expired woo trial site

### LINKED ISSUE #74785: Woo Express Performance: Upgrade after trial expiration shows trial expiration page
The first time upgrading an expired free trial shows the trial expiration page. It should show the post-upgrade task screen instead.

P2: pdtkmj-1k4-p2#comment-2397

1. Upgrade an expired trial site from the Plans page.
2. Complete the purchase.
3. You'll land on the free trial expiration page 
image

The user should land on the post-upgrade task screen instead.
image

valterlorran:
## Proposed Changes

* The state was not updated after upgrading an expired site.
* This would cause the user to be redirected back to the Trial Expired page, instead of the Trial Upgraded page.

## Testing Instructions

* Navigate to `/plans/my-plan/trial-expired/{YOUR-SITE}` on an expired woo trial site
* Upgrade your site by clicking on the "Upgrade now" button.
* Go through the checkout flow.
* You should no longer be redirected to the `plans/my-plan/trial-expired/{YOUR-SITE}`, you should see the Trial Upgraded page.

Closes #74785


markbiek (approved):
Working as expected!

client/my-sites/controller.js
```diff
@@ -553,9 +553,15 @@ export function siteSelection( context, next ) {
 		// onSelectedSiteAvailable might render an error page about domain-only sites or redirect
 		// to wp-admin. In that case, don't continue handling the route.
 		dispatch( setSelectedSiteId( siteId ) );
-		if ( onSelectedSiteAvailable( context ) ) {
-			next();
-		}
+
+		// This will fetch the site and update the state after the plan is upgraded
+		dispatch( requestSite( siteFragment ) )
```
daledupreez: I think this is not the right place to add this request, as it will trigger a new API request _every time a user navigates to a site-specific page_.

I think we need to work out if we can invalidate the site cache when a plan purchase occurs.

daledupreez (changes requested):

client/my-sites/controller.js
markbiek: Thanks for catching that! I wasn't aware this was triggered on every site-specific navigation.

client/my-sites/controller.js
```diff
@@ -548,11 +548,26 @@ export function siteSelection( context, next ) {
 	const isUnlinkedCheckoutNotBoost =
 		isUnlinkedCheckout && ! context.pathname.includes( 'jetpack_boost' );
 
+	const currentPlanSlug = getSitePlanSlug( getState(), siteId );
+	const shouldUpdateStateAfterUpgrade =
+		context.pathname.includes( 'plans/my-plan/trial-upgraded/' ) &&
```
tyxla: Noting that this could be made more precise. Right now, it would return `true` for `my-plans/my-plan/trial-upgraded/` so maybe we want to change the `.includes()` with a `.startsWith()`. You might need to add a leading forward slash then.

client/my-sites/controller.js
```diff
@@ -548,11 +548,26 @@ export function siteSelection( context, next ) {
...
+		// This will fetch the site and update the state after the plan is upgraded
+		if ( shouldUpdateStateAfterUpgrade ) {
+			dispatch( requestSite( siteFragment ) )
+				.catch( () => null )
+				.then( () => {
+					if ( onSelectedSiteAvailable( context ) ) {
+						next();
+					}
+				} );
+		} else if ( onSelectedSiteAvailable( context ) ) {
```
tyxla: This adds a bit of duplication, which indicates a potential code smell. Could we improve so we don't call the `onSelectedSiteAvailable( context )` condition twice?

valterlorran:
@tyxla Thanks for the comments. I applied both of your suggestions!

client/my-sites/controller.js
```diff
@@ -552,9 +552,21 @@ export function siteSelection( context, next ) {
...
-		}
+
+		const currentPlanSlug = getSitePlanSlug( getState(), siteId );
+		const shouldUpdateStateAfterUpgrade =
+			context.pathname.startsWith( '/plans/my-plan/trial-upgraded/' ) &&
+			[ PLAN_FREE, PLAN_JETPACK_FREE ].includes( currentPlanSlug );
+
+		// This will fetch the site and update the state after the plan is upgraded if the site is on the trial-upgraded flow.
+		const promise = shouldUpdateStateAfterUpgrade
+			? dispatch( requestSite( siteId ) )
```
tyxla: What will happen if the site request fails? Should we be handling it somehow?

valterlorran: The Trial Upgraded page will be rendered as it should. However, it may be worth invalidating the cache for the upgraded site.

Screen Shot 2023-03-29 at 11 47 09

tyxla: Sounds good. We might also catch the error just in case.

client/my-sites/controller.js
```diff
@@ -552,9 +552,24 @@ export function siteSelection( context, next ) {
...
+
+		const currentPlanSlug = getSitePlanSlug( getState(), siteId );
+		const shouldUpdateStateAfterUpgrade =
+			context.pathname.startsWith( '/plans/my-plan/trial-upgraded/' ) &&
+			[ PLAN_FREE, PLAN_JETPACK_FREE ].includes( currentPlanSlug );
+
+		// This will fetch the site and update the state after the plan is upgraded if the site is on the trial-upgraded flow.
+		const promise = shouldUpdateStateAfterUpgrade
+			? dispatch( requestSite( siteId ) ).catch( () => {
+					dispatch( removeSite( siteId ) );
```
tyxla: On second thought, I'm not 100% sure that this makes sense. What if the site is just there but the request just failed? Why would we remove the known site object if we can just keep it and use it for the time being?

valterlorran: I was concerned that the user could navigate to another page and see the free plan instead of the performance plan in case an exception was thrown. But I made several tests here, and I don't think it's necessary.

daledupreez (approved):
LGTM!

Issue #74416, opened 2023-03-14
LaunchPad - create a "choose domain" task flow.


Addison-Stavlo:
We want to update the "choose domain" task flow to not be reliant on calypso and have a better cohesive feel with the onboarding flows it is used with.

For this flow to be considered complete:
- [x] Create a stepper flow to use with the "choose domain" launchpad task.
- [x] The first step of the flow is domains selection, and allows the user for a clear way to select that they are happy with their current domain to return to launchpad with the step completed.
- [x] The second step (if choosing a domain) should be the plans and pricing grid. Note - we can show the "free" plan here IFF the domain they chose was a free domain.
- [x] After this we should leverage the existing checkout step, and completion should return to the launchpad.

Some visual and design representation of this flow can be found in p9Jlb4-6mr-p2 (2nd video) 

When this is completed, lets move the below follow up tasks to a new issue to be actioned:
- <s> update the launchpad congrats/launch modal to lean harder into the domain upsell language (@nuriapenya is looking into this) </s> moved this to a separate issue - https://github.com/Automattic/wp-calypso/issues/74712 
- [ ] after that is complete (per p9Jlb4-6mr-p2) we will also want to remove the "choose domain" task from flows that already have domain selection steps (pretty much all main flows except "free"). - when this is being done we should notify @paulbonahora as he has requested a ping once this goes into effect.

jeyip:
@agrullon95 started on this https://github.com/Automattic/wp-calypso/pull/74313

PR #74897, opened 2023-03-24
Hide free plan for domain upsell flow

agrullon95:
Related to #74416

## Testing / Review Estimation

Test: short
Review: short

## Proposed Changes

* Hides free plan on the plan selection screen for domain upsell flow

## Testing Instructions

* Checkout branch
* Create or use an existing launchpad-enabled site (Write/build/free flow)
* Click the "Choose a domain" task
* Select a random domain and verify the plan selection screen does not display a free plan.
* Go back to the domain selection screen. Click "Use a domain I own" and go through the steps until you reach the plans screen again. Verify the plan selection screen does not display a free plan.

CleanShot 2023-03-24 at 11 18 38@2x

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
@@ -94,15 +94,7 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
 			} );
 			dispatch( submitDomainStepSelection( suggestion, getAnalyticsSection() ) );
 
-			let hideFreePlan;
-
-			if ( flow === DOMAIN_UPSELL_FLOW ) {
-				hideFreePlan = true;
-			} else {
-				hideFreePlan = Boolean( suggestion.product_slug ) || shouldHideFreePlan;
-			}
-
-			setHideFreePlan( hideFreePlan );
+			setHideFreePlan( Boolean( suggestion.product_slug ) || shouldHideFreePlan );
 			setDomainCartItem( domainCartItem );
 		}
 
--- a/client/landing/stepper/declarative-flow/domain-upsell.ts
+++ b/client/landing/stepper/declarative-flow/domain-upsell.ts
@@ -1,4 +1,4 @@
-import { useSelect } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
 import { translate } from 'i18n-calypso';
 import { OnboardSelect } from 'calypso/../packages/data-stores/src';
 import {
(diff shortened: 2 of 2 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
```diff
@@ -94,15 +94,7 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
...
-			let hideFreePlan;
-
-			if ( flow === DOMAIN_UPSELL_FLOW ) {
-				hideFreePlan = true;
-			} else {
-				hideFreePlan = Boolean( suggestion.product_slug ) || shouldHideFreePlan;
-			}
-
-			setHideFreePlan( hideFreePlan );
+			setHideFreePlan( Boolean( suggestion.product_slug ) || shouldHideFreePlan );
```
edanzer, line 97: Just want to clarify what's happening here. Looks like we're taking out special logic for the DOMAIN_UPSELL_FLOW. I think we can do that here - without affecting other flows - because we're invoking `setHideFreePlan( true );` on line 51 in [domain-upsell.ts](https://github.com/Automattic/wp-calypso/pull/74897/files#diff-528d0329c8ad2a1886e8ceeacf51f5b4f639a264bae1ea37826c1b5080846a43). 

So basically, we just update that value in state when working within the domain-upsell flow, and the rest of the logic just works down stream. 

Assuming that's right, I wonder whether there would be any unanticipated effects of updating that state value? For example, once we invoke `setHideFreePlan( true );` that means as long as that state value persist, we'd never show the free plans. Is it possible that a user would end up back on the domains page via some other route/flow where they SHOULD see the free plan, but not see it because we've updated that state value? 

Worth noting I can't think of one, but I just want to be sure we don't introduce unexpected changes to the domains page when not intended.

sharpirate:
This tests well, I'll hold off on approval due to Erick's comment

client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
agrullon95: We're moving the "hide free plan" logic into the domain-upsell flow itself, instead of having that logic inside the domains step. It's essentially the same thing, but this change allows us to ensure that the Free plan is always hidden for the domain-upsell flow without modifying other code in the plans or domain step (until we decide we want to show the free plan for specific scenarios within the domain-upsell flow).

> Assuming that's right, I wonder whether there would be any unanticipated effects of updating that state value? For example, once we invoke setHideFreePlan( true ); that means as long as that state value persist, we'd never show the free plans. Is it possible that a user would end up back on the domains page via some other route/flow where they SHOULD see the free plan, but not see it because we've updated that state value?

This is definitely possible if the same redux state value is retrieved/read in another plans-related component, but that will happen regardless of these changes. The only reference for `getHideFreePlan` (to retrieve the "hide free plan" value from redux) in stepper `plans/plans-wrapper.tsx`. This plans component is only used in Link-in-bio, newsletter, and the domain-upsell flow. The link-in-bio and newsletter will not be affected by these changes since they use the "old" plans component design (vs the new plans grid design we see on the /plans screen when going the domain-upsell flow)

sharpirate (approved):

edanzer (approved):

PR #74903, opened 2023-03-24
Upsell higher Akismet tier in Calypso (i1)

andrii-lysenko:
Related to # 1201210512993648-as-1204165400433577/f

## Proposed Changes
In order to provide an Upgrade path for Akismet products we should enable Upgrade button in Calypso Purchases page. As a first iteration, we propose to directly navigate user to the next available plan at checkout, to avoid lengthy implementation and design.

## Testing Instructions
- Apply the branch locally and spin `yarn start` OR use Live link provided below
- Navigate to `/checkout/akismet/test` (if using live link provide a Feature flag param `?flags=akismet/siteless-checkout`) and purchase 10k plan
- Navigate to `/me/purchases` and select the new Purchased Akismet plan
TOGIT
- Hit the`Upgrade` button. You should be redirected to checkout for 20K plan. 
- You can repeat until the 40K plan, after which, it will redirect you to `https://akismet.com/enterprise`. Note, that in current Akismet site there's no such page, but you could update the hosts file with new one to test. 

## Pre-merge Checklist


client/me/purchases/manage-purchase/index.jsx
```diff
@@ -277,7 +280,9 @@ class ManagePurchase extends Component {
 			! isComplete( purchase ) &&
 			! isP2Plus( purchase );
 		const isUpgradeableProduct =
-			! isPlan( purchase ) && JETPACK_BACKUP_T1_PRODUCTS.includes( purchase.productSlug );
+			! isPlan( purchase ) &&
+			( JETPACK_BACKUP_T1_PRODUCTS.includes( purchase.productSlug ) ||
```
CodeyGuyDylan: Isn't required here since it's from prior implementation, but I noticed in the `calypso-products` package, there is a `isJetpackBackupT1Slug` function that could be used here

client/me/purchases/manage-purchase/index.jsx
```diff
@@ -277,7 +280,9 @@ class ManagePurchase extends Component {
...
+				isAkismetProduct( { product_slug: purchase.productSlug } ) );
```
CodeyGuyDylan: Just curious, what is the reason for calling the function this way instead of `isAkismetProduct( purchase.productSlug )`? Readability maybe?

client/me/purchases/manage-purchase/index.jsx
```diff
@@ -373,13 +378,42 @@ class ManagePurchase extends Component {
...
+		if ( yearlyIndex === AKISMET_PLUS_YEARLY_PRODUCTS.length - 1 ) {
+			return 'https://akismet.com/enterprise';
+		} else if ( yearlyIndex !== -1 ) {
+			const nextPlan = AKISMET_PLUS_YEARLY_PRODUCTS[ yearlyIndex + 1 ];
+			return `/checkout/akismet/${ nextPlan }`;
+		}
+
+		return 'https://akismet.com/enterprise';
+	}
+
```
CodeyGuyDylan: This does seem pretty hacky, but if we are planning on sending them to an upgrade page in the future this will do for now

CodeyGuyDylan (approved):
Couple small notes, none of them blocking

client/me/purchases/manage-purchase/index.jsx
andrii-lysenko: Good catch! I think I was writing this too tired:)

client/me/purchases/manage-purchase/index.jsx
andrii-lysenko: Yes, I think it worth updating) Thanks

client/me/purchases/manage-purchase/index.jsx
andrii-lysenko: Definitely! And overall, I think this logic will be replaced by a proper upsell page in the next iteration.

client/me/purchases/manage-purchase/index.jsx
```diff
@@ -373,13 +378,42 @@ class ManagePurchase extends Component {
...
 
+	// For the first Iteration of Calypso Akismet checkout we are only suggesting
+	// for immediate upgrades to the next plan. We will change this in the future
+	// with appropriate page.
+	getAkismetUpgradeUrl() {
+		const { purchase } = this.props;
+
+		const monthlyIndex = AKISTMET_PLUS_MONTHLY_PRODUCTS.indexOf( purchase.productSlug );
+		if ( monthlyIndex === AKISTMET_PLUS_MONTHLY_PRODUCTS.length - 1 ) {
+			return 'https://akismet.com/enterprise';
```
jboland88: Since this value is re-used, can it be declared as a variable?

client/me/purchases/manage-purchase/index.jsx
```diff
@@ -373,13 +378,42 @@ class ManagePurchase extends Component {
...
+		} else if ( monthlyIndex !== -1 ) {
+			const nextPlan = AKISTMET_PLUS_MONTHLY_PRODUCTS[ monthlyIndex + 1 ];
+			return `/checkout/akismet/${ nextPlan }`;
+		}
+
+		const yearlyIndex = AKISMET_PLUS_YEARLY_PRODUCTS.indexOf( purchase.productSlug );
+		if ( yearlyIndex === AKISMET_PLUS_YEARLY_PRODUCTS.length - 1 ) {
+			return 'https://akismet.com/enterprise';
+		} else if ( yearlyIndex !== -1 ) {
+			const nextPlan = AKISMET_PLUS_YEARLY_PRODUCTS[ yearlyIndex + 1 ];
```
jboland88: This approach works as long as we can assume the order of `AKISMET_PLUS_YEARLY_PRODUCTS ` matches our desired upgrade paths. Would there ever be an instance where that is not true?

An upgrade map would be much more verbose, but perhaps easier to parse? WDYT?

jboland88 (commented):
What happens for the personal plans? It looks like the current upgrade flow might point them to the enterprise URL?

client/me/purchases/manage-purchase/index.jsx
andrii-lysenko: Yes, I could define the Upgrade map. Initially I've done this way to use current logic as my assumption that we would want to create a proper Upsell page within month or two, and then the we don't need the mapping

jboland88 (approved):
Thanks for the updates! I think this is ready to ship.

We may need to tinker with the upgrade URL from free -> personal, IIRC there's an additional parameter that controls the price for the personal plan. But we can iterate on that.

packages/calypso-products/src/constants/akismet.ts
```diff
@@ -80,3 +80,17 @@ export const AKISMET_PRODUCTS_LIST = <const>[
 	...AKISMET_PLUS_PRODUCTS,
 	...AKISMET_ENTERPRISE_PRODUCTS,
 ];
+
+export const AKISMET_UPGADES_PRODUCTS_MAP = {
```
CodeyGuyDylan, line 84: It's a bit late for this, but I just noticed that there is a typo here `UPGADES`

Doesn't keep it from working, but noticed randomly when working on a file that imported this

PR #74911, opened 2023-03-24
Hosting onboarding: don't show free plan alternative

zaguiini:
Closes Automattic/dotcom-forge#2113

## Proposed Changes

We're not displaying the Free plan on the grid, so let's remove the free plan link from the subheading too:

image

## Testing Instructions

- Check out this branch and go to `/start/hosting`
- Verify that the subheading does not give the option to "start with a free site"

image

### DIFF
--- a/client/signup/config/steps-pure.js
+++ b/client/signup/config/steps-pure.js
@@ -137,6 +137,24 @@ export function generateSteps( {
 			},
 		},
 
+		'user-hosting': {
+			stepName: 'user-hosting',
+			apiRequestFunction: createAccount,
+			providesToken: true,
+			providesDependencies: [
+				'bearer_token',
+				'username',
+				'marketing_price_group',
+				'plans_reorder_abtest_variation',
+				'redirect',
+			],
+			optionalDependencies: [ 'plans_reorder_abtest_variation', 'redirect' ],
+			props: {
+				isSocialSignupEnabled: config.isEnabled( 'signup/social' ),
+				isPasswordless: true,
+			},
+		},
+
 		'user-new': {
 			stepName: 'user-new',
 			apiRequestFunction: createAccount,
--- a/client/signup/config/steps-pure.js
+++ b/client/signup/config/steps-pure.js
@@ -218,6 +218,9 @@ export function generateSteps( {
 			fulfilledStepCallback: isPlanFulfilled,
 			props: {
 				hideFreePlan: true,
+				hidePremiumPlan: true,
+				hidePersonalPlan: true,
+				shouldHideNavButtons: true,
 			},
 		},
 
--- a/client/signup/config/steps-pure.js
+++ b/client/signup/config/steps-pure.js
@@ -209,6 +209,18 @@ export function generateSteps( {
 			fulfilledStepCallback: isPlanFulfilled,
 		},
 
+		'plans-hosting': {
+			stepName: 'plans',
+			apiRequestFunction: addPlanToCart,
+			dependencies: [ 'siteSlug' ],
+			optionalDependencies: [ 'emailItem', 'themeSlugWithRepo' ],
+			providesDependencies: [ 'cartItem', 'themeSlugWithRepo' ],
+			fulfilledStepCallback: isPlanFulfilled,
+			props: {
+				hideFreePlan: true,
+			},
+		},
+
 		'plans-pm': {
 			stepName: 'plans-pm',
 			apiRequestFunction: addPlanToCart,
--- a/client/signup/main.jsx
+++ b/client/signup/main.jsx
@@ -72,7 +72,7 @@ import {
 } from 'calypso/state/sites/selectors';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import flows from './config/flows';
-import { getStepComponent } from './config/step-components';
+import { getStepComponent, getStepModuleName } from './config/step-components';
 import steps from './config/steps';
 import { addP2SignupClassName } from './controller';
 import {
@@ -785,9 +785,12 @@ class Signup extends Component {
 			};
 		}
 
+		let stepName = getStepModuleName( this.props.stepName );
+		stepName = stepName === 'user' ? 'user' : stepName;
+
 		return (
 			<div className="signup__step" key={ stepKey }>
-				<div className={ `signup__step is-${ kebabCase( this.props.stepName ) }` }>
+				<div className={ `signup__step is-${ stepName }` }>
 					{ shouldRenderLocaleSuggestions && (
 						<LocaleSuggestions path={ this.props.path } locale={ this.props.locale } />
 					) }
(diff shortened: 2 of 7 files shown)


client/signup/config/steps-pure.js
```diff
@@ -209,6 +209,18 @@ export function generateSteps( {
...
+		'plans-hosting': {
+			stepName: 'plans',
+			apiRequestFunction: addPlanToCart,
+			dependencies: [ 'siteSlug' ],
+			optionalDependencies: [ 'emailItem', 'themeSlugWithRepo' ],
+			providesDependencies: [ 'cartItem', 'themeSlugWithRepo' ],
+			fulfilledStepCallback: isPlanFulfilled,
+			props: {
+				hideFreePlan: true,
+			},
```
zaguiini: I think this is a much better solution than hooking into the `flowName` inside `plans/index.jsx`.

The downside is that it renames the plans page from `/start/hosting/plans` to `/start/hosting/plans-hosting`. But I guess this is a valid trade-off.

client/signup/steps/plans/index.jsx
```diff
@@ -226,6 +226,10 @@ export class PlansStep extends Component {
 		}
 
 		if ( useEmailOnboardingSubheader ) {
+			if ( hideFreePlan ) {
+				return translate( 'Add more features to your professional website with a plan. ' );
+			}
```
p-jackson: `useEmailOnboardingSubheader` comes from a custom step prop so there's no way for us to go down this branch where `useEmailOnboardingSubheader` and `hideFreePlan` are both true. So rather than add a new bit of copy that isn't reachable perhaps we just have one of the flags invalidate the other like:

```suggestion
		if ( useEmailOnboardingSubheader && ! hideFreePlan ) {
```

client/signup/config/steps-pure.js
```diff
@@ -209,6 +209,21 @@ export function generateSteps( {
...
 
+		'plans-hosting': {
+			stepName: 'plans',
+			apiRequestFunction: addPlanToCart,
+			dependencies: [ 'siteSlug' ],
+			optionalDependencies: [ 'emailItem', 'themeSlugWithRepo' ],
+			providesDependencies: [ 'cartItem', 'themeSlugWithRepo' ],
+			fulfilledStepCallback: isPlanFulfilled,
+			props: {
+				hideFreePlan: true,
```
p-jackson, line 238: Note that usually the free plan is hidden based on dependency data. So this step prop is overriding the work that's usually done here:
https://github.com/Automattic/wp-calypso/blob/2669d8e11a270e97501a128685c12f83270c5561/client/signup/main.jsx#L767

It works because custom step props take precedence over any other props:
https://github.com/Automattic/wp-calypso/blob/2669d8e11a270e97501a128685c12f83270c5561/client/signup/main.jsx#L808

p-jackson (approved):
I like using the custom step props instead of `isHostingLPFlow()` 👍 

You should be aware that the new plans grid doesn't support hiding the free plan. Looks like it was something to do with working around a Jetpack logo issue? https://github.com/Automattic/martech/issues/1464 🤷 

The `/start/hosting` flow doesn't use the new plans grid so we're fine for now. But we might run into trouble if martech rolls it out everywhere (by changing this function https://github.com/Automattic/wp-calypso/blob/2669d8e11a270e97501a128685c12f83270c5561/packages/calypso-products/src/plans-utilities.ts#L61

> We're not displaying the Free plan on the grid, so let's remove the free plan link from the subheading too

Perhaps there's additional context for why we're making this change that I'm missing. But I don't think the above justification makes sense. We never show the free plan in the old plan grid. It's always in the subheading.

Personally I worry that removing the free plan will turn builders off, because I assume builders want to try before they buy. On the other hand the free plan doesn't really let them try the features they'll be interested in. Ultimately I think we need to somehow give builders the ability to try out the Atomic platform for free.

zaguiini:
Thanks @p-jackson. Yeah it's indeed not removing the free plan on the new grid. Let's hold it and find a solution that works for both grids.

>Perhaps there's additional context for why we're making this change that I'm missing.

Sorry, there you go: pet6gk-jJ-p2

zaguiini:
~This is blocked by https://github.com/Automattic/wp-calypso/pull/75424. Do not merge this PR.~ The new pricing grid now supports hiding the free plan, and because of that we're good to go 👍

client/signup/config/steps-pure.js
zaguiini: Yeah. The reason we're doing this is since the plans section is coming first, there is no way we can pass this dependency, so we need to explicitly tell it to the plans step.

client/signup/main.jsx
```diff
@@ -785,9 +785,12 @@ class Signup extends Component {
 			};
 		}
 
+		let stepName = getStepModuleName( this.props.stepName );
+		stepName = stepName === 'user' ? 'user' : stepName;
+
 		return (
 			<div className="signup__step" key={ stepKey }>
-				<div className={ `signup__step is-${ kebabCase( this.props.stepName ) }` }>
+				<div className={ `signup__step is-${ stepName }` }>
```
zaguiini, line 793: We're displaying the vertically-split login page if the `.is-user` class matches. But the thing is the step name must match `user`, otherwise, it's not applied. Since the `user-hosting` step is using the `user` component, we must extract it from the step name and pass it to the class name to style the step properly.

mpkelly (approved):
It's working as advertised. LGTM.

client/signup/main.jsx
aneeshd16: Hi @zaguiini! This change caused some styling issues with a few signup steps, as there are CSS styles with selectors like `signup__step.is-<step name>`.
https://github.com/search?q=repo%3AAutomattic%2Fwp-calypso%20.signup__step.is&type=code

One example is the page picker step in the BBE flow:
| Before | After |
|--------|--------|
|image|image |

What do you think is the best step forward? How about something like this in L789?
```
stepName = stepName === 'user' ? 'user' : this.props.stepName;
```

I think having this distinction between step name and step module name is important because two different steps might have the same step module, and need to have different styles.

p-jackson: @aneeshd16 It looks like we're the ones who wanted to do something special for the `user-hosting` step here, so rather than changing the behaviour for every flow in the signup framework we should have only made the change for our new step.

I'm thinking of something more like
```js
const stepName = this.props.stepName === 'user-hosting' ? 'user' : this.props.stepName
```

The `.is-xyz` classes have always been used for styling steps, not components. If there are some CSS rules that would be more appropriately tied to the React component so they get shared by everyone who uses the component regardless of what step it is (which I think is what @zaguiini was thinking) then it'd be better to migrate those CSS rules to the component's stylesheet.
To me, using `getStepModuleName()` is sort of like trying to associate something with the React component.

zaguiini: Yeah, that was exactly my intention. Sorry for the regression -- I'll fix it.

zaguiini: @aneeshd16 here you go: https://github.com/Automattic/wp-calypso/pull/75611

PR #74914, opened 2023-03-24
WoA Logging i1: Add starter toolbar with date range picker

vykes-mac:
Closes https://github.com/Automattic/dotcom-forge/issues/1994

## Proposed Changes
This PR is a starter to push the project forward, we will refactor as more things are added to make it better. As it stands this PR:

* Adds date range picker to the toolbar.
* Fetches new logs based on the selected date range.

## Testing Instructions

- Verify default date range is `now - 7days`
- Choose a date range and verify new logs are fetched for that range (_given your site has logs for that date_)

**Note** The logs are paginated so will not see all the logs for the selected dates.

image

## Pre-merge Checklist


zaguiini:
It does work, functionally. I don't like how it looks, though. That single field looks a out of place there:

image

Something like that would be much better:

image

The filtered date would be applied to both types of logs, so that the user can quickly skim through the logs without having to sync the date between the two tabs 👍

vykes-mac:
@zaguiini we can probably sync dates by adding it to the url, that way we keep it also if user refreshes the page. As for the UI we can have that discussion further with @p-jackson.

zaguiini:
>we can probably sync dates by adding it to the url, that way we keep it also if user refreshes the page.

It'd definitely be a good thing to add the dates to the URL as query parameters, but I do think we should move the date selection above the tabs so the user can be sure it's synced across tabs.

>As for the UI we can have that discussion further with @p-jackson.

Of course. I think the version I proposed is much better for a few reasons:

- it's immediately visible that the user can filter by a date range;
- space is better filled (there is no weird blank space);
- we're displaying the full information without needing an additional click;
- it looks better too 🤷‍♂️

p-jackson:
> we can probably sync dates by adding it to the url

Keeping the date in sync between the two tabs sound good. I'm not sure whether putting it in the URL is the best thing, but I'm kinda of ambivalent about it. So I'm ok with it.

@zaguiini A big part of using the single drop down component for selecting a time range is because that's what we already use for the log download UI. When we include the download UI in this view would it be strange to use one control in one place and another control in another place?

> space is better filled (there is no weird blank space);

There are more controls coming in future PRs, so won't be blank for long. Also there's the possibility of including other filter controls too. I think it'd be weird to have the time range filtering above the tabs and then the other filters (which can't all be shared between PHP and Web logs) below the tabs.

> it's immediately visible that the user can filter by a date range;

I'm not sure what you mean by this. I think in both variations it's obvious you can filter by date range.

> we're displaying the full information without needing an additional click;

Same thing here, the separate and combined date controls both show a date range 🤔 

> it looks better too 🤷‍♂️

I disagree I guess 😅 But on top of that I think that the IA I was going for was definitely tabs first, and then once you've selected a tab then you do your filtering.

I wonder if the issue is actually the styling of the tabs? What if the tabs looked more like this?

CleanShot 2023-03-27 at 15 37 58@2x

This is what other sections with the modern tab design look like, where the tabs are clearly in the section header. I think it'd be peculiar for the time range filtering to be in the section header. At least with the IA I proposed in the PT.

client/my-sites/site-logs/components/toolbar/index.tsx
```diff
@@ -0,0 +1,33 @@
+import DateRange from 'calypso/components/date-range';
+import { useLocalizedMoment } from 'calypso/components/localized-moment';
+import './style.scss';
+
+type toolbarProps = {
+	startDateRange: number;
```
p-jackson: I don't think the props needs `Range` in the name. Just `startDate` and `endDate`. Or even `start` or `startDateTime`, since the unix timestamp includes time-of-day too.

client/my-sites/site-logs/main.tsx
```diff
@@ -9,20 +9,21 @@ import { SiteLogsTab, useSiteLogsQuery } from 'calypso/data/hosting/use-site-log
...
-	const [ endTime ] = useState( moment().unix() );
+	const [ startTime, setStartTime ] = useState( moment().subtract( 7, 'd' ).unix() );
+	const [ endTime, setEndTime ] = useState( moment().unix() );
 
 	const [ logType, setLogType ] = useState< SiteLogsTab >( () => {
 		const queryParam = new URL( window.location.href ).searchParams.get( 'log-type' );
 		return (
 			queryParam && [ 'php', 'web' ].includes( queryParam ) ? queryParam : 'php'
-		) as SiteLogsTab;
+		 ) as SiteLogsTab;
```
p-jackson: This rogue space has come back to break the build somehow 😅 Do you have VS Code set up with eslint formatting? p4TIVU-9Lo-p2

p-jackson (commented):
I think an issue here is that users should be able to specify not just the date, but also the time they want logs within. Otherwise for a high traffic site, if the log entry they're looking for is at 11pm then it could be on page 100.

Edit: I don't know how I missed this but I thought the download UI supported filtering by time, but I see now it only supports filtering by day. I don't think that's going to be good enough for our use case. I missed this sorry. I've added this requirement to the task's acceptance criteria.

client/my-sites/site-logs/main.tsx
vykes-mac: Haha I saw it in the linter but somehow it missed me

zaguiini:
Cool 👍 I didn't know we were adding more stuff to the toolbar.

I think it'd be essential to have a label like "Date range" above the input (and linked to it so clicking opens the range picker).

One thing that's bugging me is that the picker input jumps on click. It also loses the background color:

https://user-images.githubusercontent.com/26530524/228041165-cea6a6d7-47ed-41ad-adeb-134a04181f68.mov

client/my-sites/site-logs/main.tsx
```diff
@@ -18,8 +18,8 @@ export function SiteLogs() {
 	const moment = useLocalizedMoment();
 
 	const getDateRange = () => {
-		const startTime = moment().subtract( 7, 'd' ).unix();
-		const endTime = moment().unix();
+		const startTime = moment().subtract( 7, 'd' ).startOf( 'day' ).utc().unix();
```
p-jackson: I don't think it'll cause any issues, but is there a reason we want the start of the time range to be set to the start of the day rather than a time range that's precisely 7 days long? This is the source of the unit test failure. But if there's a reason it's better to have the time range start at midnight then we can update the unit test to match the new behaviour.

vykes-mac: This is to be refactored to be 7 days with the current time. I'll make the changes

client/components/date-range/index.js
```diff
@@ -93,10 +94,12 @@ export class DateRange extends Component {
 		// Build initial state
 		this.state = {
 			popoverVisible: false,
-			staleStartDate: null,
-			staleEndDate: null,
+			staleStartDate: startDate,
+			staleEndDate: endDate,
```
p-jackson, line 98: Why did we need to change these initial "stale" state values? I'm not 100% sure what they're for, but it seems like they were intentionally left `null` and only set the first time the dates are changed https://github.com/Automattic/wp-calypso/blob/cd52b1acc86ad780227ec8f3400b25beeca672b1/client/components/date-range/index.js#L340

client/components/date-range/index.js
```diff
@@ -176,7 +180,7 @@ export class DateRange extends Component {
 	isValidDate( date ) {
 		const { firstSelectableDate, lastSelectableDate } = this.props;
 
-		const epoch = this.props.moment( '01/01/1970', this.getLocaleDateFormat() );
+		const epoch = this.props.moment( '01/01/1970' );
```
p-jackson: In the interests of changing as little code as possible, we could hardcode this to

```suggestion
		const epoch = this.props.moment( '01/01/1970', this.props.moment.localeData().longDateFormat( 'L' ) );
```

p-jackson (approved):
Lets get this merged.

The `<DateRange>` component is only used in a couple other places: log downloading in `/hosting-config`, and the Jetpack activity log. So I was able to test those two areas and feel fairly confident that things are working fine.

This does introduce some bugs with the refresh button, but I think we can address that as a bug fix once it's merged.

One of the other changes on `trunk` has been that the date range state has changed from unix time stamps to `Moment` objects. Which might also be causing issues.

client/components/date-range/index.js
vykes-mac: I set staleDates initially because when you open the dates and the click cancel without changing anything it empty the dates field and display the date format instead. 

This is because even when cancel is clicked it set state with staleDate which would be null if anything wasn't set.

```js
revertDates = () => {
		this.setState(
			( previousState ) => {
				const startDate = previousState.staleStartDate;
				const endDate = previousState.staleEndDate;
				const newState = {
					staleDatesSaved: false,
					startDate: startDate,
					endDate: endDate,
					textInputStartDate: this.toDateString( startDate ),
					textInputEndDate: this.toDateString( endDate ),
				};

				return newState;
			},
			() => {
				this.props.onDateCommit( this.state.startDate, this.state.endDate );
			}
		);
	};
```

image

PR #74915, opened 2023-03-24
Hosting onboarding: allow skipping domain step

zaguiini:
Part of https://github.com/Automattic/dotcom-forge/issues/2122

## Proposed Changes

This PR allows the user to skip the domains step and use a free, generated blog URL when checking out.

The user shouldn't need to pick a domain when signing up through `/start/hosting`.

The idea for that flow is to showcase and highlight hosting features, unrelated to domain selection. I don't think we should remove the domains step because the user _might_ want to choose a domain from the get-go if they're confident about moving to WPCOM.

image

## Testing Instructions

- Check out this branch;
- Visit `/start/hosting`;
- Pick a plan;
- Verify that it's possible to skip the domain selection;
- Checkout works correctly and you can buy the business plan.

### DIFF
--- a/client/signup/steps/domains/index.jsx
+++ b/client/signup/steps/domains/index.jsx
@@ -656,11 +656,9 @@ class DomainsStep extends Component {
 		if ( HOSTING_LP_FLOW === flowName ) {
 			const components = {
 				span: (
-					// eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus
-					<span
-						role="button"
+					<button
 						className="tailored-flow-subtitle__cta-text"
-						style={ { fontWeight: 500 } }
+						style={ { fontWeight: 500, fontSize: '1em', display: 'inline' } }
 						onClick={ () => this.handleSkip( undefined, true ) }
 					/>
 				),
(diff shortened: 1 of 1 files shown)


client/signup/steps/domains/index.jsx
```diff
@@ -653,6 +653,25 @@ class DomainsStep extends Component {
...
+		if ( HOSTING_LP_FLOW === flowName ) {
+			const components = {
+				span: (
+					// eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus
+					<span
+						role="button"
+						className="tailored-flow-subtitle__cta-text"
+						style={ { fontWeight: 500 } }
+						onClick={ () => this.handleSkip( undefined, true ) }
+					/>
```
p-jackson: We can just address the a11y issues by using a button

```suggestion
					<button
						className="tailored-flow-subtitle__cta-text"
						style={ { fontWeight: 500, fontSize: '1em', display: 'inline' } }
						onClick={ () => this.handleSkip( undefined, true ) }
					/>
```

p-jackson (approved):
Good to get us started.

I think this step could be changed a lot to be better for the builder audience. "Enter some descriptive keywords", "not sure yet", "the best names are short and memorable" do not feel right for our audience.

zaguiini:
>I think this step could be changed a lot to be better for the builder audience. "Enter some descriptive keywords", "not sure yet", "the best names are short and memorable" do not feel right for our audience.

Yeah, I think we can improve it as part of pet6gk-jJ-p2.

Thanks for approving the PR! ~I'll ask for design input before merging it though.~ Asked for input: p1679949297238489-slack-C041RHH38NQ

zaguiini:
I'm merging this so we can allow skipping the step. We can create a follow-up PR if we decide to change how the user might do it.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7919827

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7919827).

PR #74918, opened 2023-03-24
Remove unused launchpad taskType property

edanzer:
### Proposed Changes

This PR removes the 'taskType' property from launchpad tasks. We do not use this property. It was originally added to keep data consistency with the older Checklist API, which does use that property. But at this point, we're unlikely to refactor to use the Checklist API, and it seems worth cleaning this out to avoid confusion about why it's there. 

### Testing

**Review Time: Short**
**Testing Time: Short**

There should be zero change in functionality or behavior before and after this PR. 

1. To test, run through any launchpad flow to smoke test and confirm no issues. 

2. Because I removed a reference to this property in our unit tests fixtures, make sure Launchpad tests still pass by running `yarn run test-client client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/ `.

3. Bonus: Consider searching the code base for references to 'taskType' for launchpad to confirm this property is not being used. I've already done so exhaustively, but an extra set of eyes would be good. Note that I've also checked our backend code in launchpad.php to confirm no references to this property. Main thing is that we want to be sure we're not actually using this property in some location.


agrullon95:
I didn't find other references to `taskTypes`. 
The launchpad tests are passing, but I did notice a warning for the new launchpad API. it's unrelated to the changes in this PR. We can fix this in another PR using nock: https://github.com/nock/nock

CleanShot 2023-03-27 at 09 49 51@2x


## Tests

- [x] Smoke-tested Link-in-bio flow
- [x] Make sure Launchpad tests still pass

agrullon95 (approved):

edanzer:
> I didn't find other references to `taskTypes`. The launchpad tests are passing, but I did notice a warning for the new launchpad API. it's unrelated to the changes in this PR. 

I missed this notice when running tests (was just looking at them passing), but you're right and I can confirm the notice above shows even when running the launchpad tests against latest trunk. Makes me wonder how long that's been there.

PR #74920, opened 2023-03-24
VAT: Use countries list from server to determine VAT eligibility

sirbrillig:
## Proposed Changes

This PR replaces the hard-coded lists of countries supporting VAT with a list from the `/me/transactions/supported-countries` endpoint.

Requires D106018-code

## Testing Instructions

This affects two forms which must be tested:

#### VatInfo

Screenshot 2023-03-24 at 5 51 22 PM

- Verify that the country dropdown menu contains a list of countries; make sure that it contains "Northern Ireland" (XI) and "Greece" (EL).

#### VatForm

Screenshot 2023-03-25 at 12 16 04 AM

- In the billing info (aka: "contact details") step, select a country for which we support VAT collection (eg: United Kingdom).
- Verify that the "Add VAT details" (aka: "Add Business Tax ID details") checkbox appears.

### DIFF
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -194,7 +194,7 @@ function CountryCodeInput( {
 				const name = country.code === 'XI' ? translate( 'Northern Ireland' ) : country.name;
 				return country.tax_country_codes.map( ( countryCode ) => (
 					<option key={ countryCode } value={ countryCode }>
-						{ name }
+						{ countryCode } - { name }
 					</option>
 				) );
 			} ) }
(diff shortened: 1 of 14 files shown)


client/me/purchases/vat-info/index.tsx
```diff
@@ -153,6 +158,20 @@ function VatForm() {
 	);
 }
 
+const isVatSupported = ( country: CountryListItem ): country is CountryListItemWithVat =>
```
pottedmeat: Would it be worth DRYing the `isVatSupported`?

pottedmeat (approved):
All testing went as expected:

image

Code looks great. Love the With/WithoutVat.

client/me/purchases/vat-info/index.tsx
sirbrillig: Good idea. There's many places I could put it, but for now I've just moved it to the `useCountryList` file since it mostly relates to its results.

Issue #67012, opened 2022-08-25
Cart action rejections are not all handled


sirbrillig:
In https://github.com/Automattic/wp-calypso/pull/58834 we began rejecting the Promise returned by code using the @automattic/shopping-cart package if the /me/shopping-cart endpoint returns its payload with messages in the errors array. 

Messages returned in the errors array are displayed by the `CartMessages` component regardless of what action caused the endpoint to return the errors. This is in addition to the rejected Promise.

Because the Promise rejects, any code that was calling an action (eg: addProductsToCart()) can assume that the operation failed. However, some calling code does not handle the rejection, because there is no action needed. This causes the error to bubble up to the console, and more importantly, to our error logging systems. Recording errors like "This coupon is not valid" is not useful.

We should make sure that every action's Promise includes a `try/catch` (if using `await`) or a `.catch()`. (Make sure to double check that rejections are not already being handled at a higher level, like a parent `async` function.)

sirbrillig:
One example of resolving this: https://github.com/Automattic/wp-calypso/pull/66540

sirbrillig:
https://github.com/Automattic/wp-calypso/pull/67020 starts this process for one of the actions.

https://github.com/Automattic/wp-calypso/pull/67061 does it for all the other actions I think are likely causing failures. 

Once those are merged we can check the Sentry logs to see if any more errors are being recorded.

sirbrillig:
https://github.com/Automattic/wp-calypso/pull/74930 covers more of these.

PR #58834, opened 2021-12-04
Shopping Cart: Reject action promises if an error occurs

sirbrillig:
#### Changes proposed in this Pull Request

The shopping cart manager in the `@automattic/shopping-cart` package returns a Promise when calling any of its actions. The Promise resolves, not when the action is complete, but when the cart next becomes "valid"; this means that several actions may be batched together and the Promise won't resolve until all of them complete. However, the Promise never rejects, which can be confusing because it breaks the typical Promise contract. 

An action can certainly fail, but it's not always easy to know what "failure" is since many errors still return a valid cart (with error messages included), and if there are multiple batched actions, which action failed? The original intention was that consumers would check for errors on the response themselves, but I believe that this will only create bugs (eg: https://github.com/Automattic/wp-calypso/pull/58428).

This PR changes the action Promise behavior so that it will now reject if one of two things occurs:

1. If any currently batched action experiences a connection issue, setting `loadingError` to a string. In this case, the `loadingError` and `loadingErrorType` will be passed as the rejection argument as an instance of `CartActionConnectionError`.
2. If the cart becomes "valid" after any number of batched actions, and the response contains a non-zero number of values in `responseCart.messages.errors`. The first error in that array will be passed as the rejection argument as an instance of `CartActionResponseError`.

In most cases this will not have any notable effect because most places that use cart actions already display errors from the cart to the user (via `CalypsoShoppingCartProvider` which includes `CartMessages`). The biggest difference will be that if an action fails before, say, a redirect to checkout, the error message will appear on the source page rather than the destination page.

#### Testing instructions

Unit tests are included.

Since this obviates the need for additional guards in most cases, this also reverts the guards added in https://github.com/Automattic/wp-calypso/pull/58428. Therefore, we should test that the functionality of that PR remains unchanged:

- You'll need to modify the shopping-cart endpoint to always return an error; D70551-code should do it.
- Visit `/start` and proceed through the signup flow, adding a paid plan to the cart.
- Verify that you are returned to the plan step.
- Verify that you see an error message displayed.

PR #74930, opened 2023-03-25
shopping-cart: Handle errors from calling reloadFromServer

sirbrillig:
In https://github.com/Automattic/wp-calypso/pull/58834 we began rejecting the Promise returned by code using the `@automattic/shopping-cart` package if the `/me/shopping-cart` endpoint returns its payload with messages in the errors array or if there is a connection error.

Messages returned in this way are displayed by the `CartMessages` component regardless of what action caused the endpoint to return the errors. This is in addition to the rejected Promise.

Because the Promise rejects, any code that was calling an action can assume that the operation failed. However, some calling code does not handle the rejection, because there is no action needed. This causes the error to bubble up to the console, and more importantly, to our error logging systems. Recording errors like these are not useful.

We should make sure that every action's Promise includes a `try/catch` (if using `await`) or a `.catch()`. See https://github.com/Automattic/wp-calypso/issues/67012 for more info.

## Proposed Changes

This PR adds `catch()` calls to every call to the `reloadFromServer` action. This will stop errors from being logged for these events like p1679769139086579-slack-C04U5A26MJB which was caused by the Masterbar cart having connectivity issues reaching the public-api server.

## Screenshots

Before this PR:

Screenshot 2023-03-25 at 4 53 23 PM

## Testing Instructions

To verify that this silences the errors, first you need to break the shopping-cart endpoint. The easiest way to do that is to sandbox the API and then throw an exception inside the endpoint code (eg D50045-code).

Next, load a page in calypso that loads the Masterbar cart. Any site-level page should do, like the customer home page or the sites page. Leave the JavaScript console open while it loads.

Verify that there are no "Uncaught Promise" errors that refer to the shopping-cart endpoint failing.

### DIFF
--- a/client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
+++ b/client/layout/masterbar/masterbar-cart/masterbar-cart-button.tsx
@@ -72,7 +72,13 @@ export function MasterbarCartButton( {
 		setIsActive( ( active ) => {
 			if ( ! active ) {
 				// This is to prevent an error in updating the component in the same event loop
-				setTimeout( reloadFromServer, 0 ); // Refresh the cart whenever the popup is made visible.
+				setTimeout(
+					() =>
+						reloadFromServer().catch( () => {
+							// No need to do anything here. CartMessages will report this error to the user.
+						} ),
+					0
+				); // Refresh the cart whenever the popup is made visible.
 				reduxDispatch( recordTracksEvent( 'calypso_masterbar_cart_open' ) );
 			}
 			return ! active;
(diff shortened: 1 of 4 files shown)


spraveenitpro:
<details>
<summary>Before Applying the Patch</summary>


</details>


<details>
<summary>After Applying the Patch</summary>

1. I confirm I applied the exception to the backend code 
2. Applied the frontend patch you have provided.
3. Refreshed on various pages loading the mastercart
4. And still got 
</details>

sirbrillig:
@spraveenitpro Thanks! It looks like that's working as intended. Yes, you'll always see the browser's error on the failed GET but there's no failed Promise any more, which is what this PR fixes.

PR #74935, opened 2023-03-27
Display pagination controls below site log table

p-jackson:
Closes Automattic/dotcom-forge#1991

## Proposed Changes

Adds pagination to the `/site-logs` section.

CleanShot 2023-03-28 at 11 09 59@2x

* Change log page size from 10 to 50.
* Display pagination UI at the bottom of the table. The logging API doesn't allow jumping to specific pages so we only show the **Previous** and **Next** buttons.
* Store logs in the ReactQuery cache by page index, not by scroll ID. The scroll ID represents a server state that is shared between multiple requests, so it's not an appropriate way to cache pages.
* Deal with undocumented API behaviour where `total_results` is sometimes an object
* Changed log entry values from `any` to `unknown`. My intention was to ensure unexpected values from the server are handled, which is better modelled by `unknown`. See https://github.com/Automattic/wp-calypso/pull/74862#discussion_r1147384905
* Caches the previous page of logs in the component so that they can continue to be shown while we're loading the next page. The main benefit is that the table stays the same height while loading the next page and the layout doesn't have to shift.

## Testing Instructions

* Head to `/site-logs` for a WoA site
* Click forwards and backwards through the pages
* You could adjust the page size locally if you don't have enough logs on your test site.

## Pre-merge Checklist


mpkelly:
It mostly works well for me. I added a commit to reset the page number when changing tabs. 

I also discovered an issue:

1. Go to http://calypso.localhost:3000/site-logs/atomicsite66.wpcomstaging.com?log-type=php and observe the total results count
2. Next, go to http://calypso.localhost:3000/site-logs/atomicsite66.wpcomstaging.com?log-type=web and hard refresh
3. Again, observe the total results count, and see that it is different: 1811 vs 9387 for me. 
4. Now change tabs without refreshing and observe that the total results don't change. After checking the API responses, it appears this problem is on the backend, which always returns the same `total_results` object for both `/logs` and `/error-logs` endpoints. 

As the problem seems to be on the backend, I think we should merge this to avoid backing-up too many PRs. I created https://github.com/Automattic/dotcom-forge/issues/2065 for this bug.

mpkelly (approved):
It works well. There is one issue but that can be handled separately.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7892698

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7892698).

PR #74960, opened 2023-03-27
Plans 2023: header-price refactors and unit tests

chriskmnds:
Related to https://github.com/Automattic/wp-calypso/issues/74261

## Proposed Changes

This is a follow-up to https://github.com/Automattic/wp-calypso/pull/74375 (which we have put on hold) to compile the relevant refactors that we'd like to merge into `trunk`, along with some unit/UI tests for the header-price component (should also cover usage of the new hook). We are introducing a `use-plan-prices` hook under `/plans`, something that we'll be able to repurpose into other refactors and work pricing-related.

Changes here should allow us to quickly pick up developments with  https://github.com/Automattic/wp-calypso/issues/74261 once we have a clearer idea of cost-savings.

## Testing Instructions

* Code review
* Unit tests
* Go to `/plans/[SITE]` with a variety of sites (e.g. on free plan, on paid plan, with pro-rated credits, etc.) and confirm prices are displayed as expected (with cross-out prices when discounts exist)
* Confirm with different currencies e.g. IRN to ensure those discounts are also visible in the prices

p.s. This is a foundation PR that will also be tested more thoroughly in a follow-up PR that will be based on this one to update the price displays.

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #74961, opened 2023-03-27
Plans 2023: Update discounted price styling

aneeshd16:
Related to Automattic/martech#1472

## Proposed Changes

* This PR updates the styling of discounts according to xkhhUlDowF13dRoXJqlRDM-fi-1872-54424 . It builds on top of #74960 so that it can utilize the refactored `PlanFeatures2023GridHeaderPrice` component.

## Testing Instructions

* Confirm that the discount styling matches the design in the following cases:
  * On `/start/plans` for a currency with a first-year discount.
image
image

  * On `/plans/<site slug>` where pro-rated discounts are applicable.
 
image

  * Both the above cases after switching to the IDR currency.
  
image
 
image

* Also check for mobile/tablet resolutions:
image
image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/billing-timeframe.tsx
+++ b/client/my-sites/plan-features-2023-grid/billing-timeframe.tsx
@@ -50,7 +50,7 @@ function usePerMonthDescription( {
 	if ( ! isMonthlyPlan ) {
 		const fullTermPriceText =
 			currencyCode && null !== maybeDiscountedFullTermPrice
-				? formatCurrency( maybeDiscountedFullTermPrice, currencyCode )
+				? formatCurrency( maybeDiscountedFullTermPrice, currencyCode, { stripZeros: true } )
 				: null;
 
 		if ( fullTermPriceText ) {
--- a/client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
+++ b/client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
@@ -296,6 +296,7 @@ const PlanComparisonGridHeaderCell: React.FunctionComponent<
 		planProperties: PlanProperties;
 		allVisible: boolean;
 		isLastInRow: boolean;
+		isLargeCurrency: boolean;
 	}
 > = ( {
 	planProperties,
@@ -312,6 +313,7 @@ const PlanComparisonGridHeaderCell: React.FunctionComponent<
 	isLaunchPage,
 	flowName,
 	selectedSiteSlug,
+	isLargeCurrency,
 	onUpgradeClick,
 } ) => {
 	const { planName, planConstantObj, availableForPurchase, current, ...planPropertiesObj } =
@@ -331,7 +333,6 @@ const PlanComparisonGridHeaderCell: React.FunctionComponent<
 		'is-current-plan': current,
 	} );
 	const rawPrice = planPropertiesObj.rawPrice;
-	const isLargeCurrency = rawPrice ? rawPrice > 99000 : false;
 	const showPlanSelect = ! allVisible && ! current;
 
 	return (
@@ -424,6 +425,10 @@ const PlanComparisonGridHeader: React.FC< PlanComparisonGridHeaderProps > = ( {
 } ) => {
 	const allVisible = visiblePlansProperties.length === displayedPlansProperties.length;
 
+	const isLargeCurrency = displayedPlansProperties.some(
+		( properties ) => properties?.rawPrice && properties?.rawPrice > 99000
+	);
+
 	return (
 		<PlanRow>
 			<RowTitleCell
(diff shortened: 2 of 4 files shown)


client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
```diff
@@ -323,7 +325,6 @@ const PlanComparisonGridHeaderCell: React.FunctionComponent<
 		'is-current-plan': current,
 	} );
 	const rawPrice = planPropertiesObj.rawPrice;
-	const isLargeCurrency = rawPrice ? rawPrice > 99000 : false;
```
aneeshd16, line 334: This implementation is incorrect as it calculates the value of `isLargeCurrency` for each plan separately. So, it is possible that the value is `false` for the Personal plan, but for the Business plan, the value is `true`. This causes the `is-large-currency` class to apply to only some of the plans displayed, leading to inconsistent styling.

client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -323,6 +323,10 @@
 			font-weight: 600;
 		}
 
+		&.is-bottom-aligned {
+			vertical-align: bottom;
+		}
+
```
aneeshd16, line 329: The `is-bottom-aligned` class was already applied, but this CSS rule was missing.

client/my-sites/plan-features-2023-grid/billing-timeframe.tsx
```diff
@@ -50,7 +50,7 @@ function usePerMonthDescription( {
 	if ( ! isMonthlyPlan ) {
 		const fullTermPriceText =
 			currencyCode && null !== maybeDiscountedFullTermPrice
-				? formatCurrency( maybeDiscountedFullTermPrice, currencyCode )
+				? formatCurrency( maybeDiscountedFullTermPrice, currencyCode, { stripZeros: true } )
```
aneeshd16, line 53: Small change.
Before:
image

After:
image

client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -336,27 +340,25 @@
 
 	.plan-features-2023-grid__pricing {
```
chriskmnds: WDYT about consolidating all the `plan-features-2023-grid__pricing` styles into one location while at it. So either move everything from here into `header-price` (which most look like overrides of `PlanPrice`) or move the [new styles](https://github.com/Automattic/wp-calypso/pull/74961/files#diff-0b8c173504dad73d4da3beb1c69d587a74c61893c99961bc0544b5bd506eb1b7R15-R22) here?

chriskmnds (approved):
Thanks @aneeshd16. Tests out. I left a suggestion for a refactor, which I'm also happy to do it, and push the commit.

The only somewhat minor inconsistency I notice is the display and cross-outs when having decimals in the price e.g.

Screenshot 2023-03-28 at 1 24 16 PM

Screenshot 2023-03-28 at 1 39 14 PM


- the alignment is not identical between the two
- the double strikethrough feels a little odd (which looks like something that was previously solved with the diagonal line?)

client/my-sites/plan-features-2023-grid/style.scss
aneeshd16: Agreed!

aneeshd16:
> The only somewhat minor inconsistency I notice is the display and cross-outs when having decimals in the price e.g.
> * the alignment is not identical between the two
> * the double strikethrough feels a little odd (which looks like something that was previously solved with the diagonal line?)

Good catch! I'll figure out a way to replicate decimal pricing here.

chriskmnds:
> Good catch!

By accident :D

aneeshd16:
Hi @chriskmnds, I have consolidated all styles with the component via https://github.com/Automattic/wp-calypso/pull/74961/commits/8c335502fee873a0930d0175c9473322e3075af5. This does not address the issue with the decimal prices, I have asked for clarification in p1680077449183959-slack-C040UAN6TPS

aneeshd16:
Since the `isLargeCurrency` prop must be computed at a higher level, it would require more refactoring. I'll tackle that separately in a follow-up PR. This PR is good to go from my end. 👍

chriskmnds:
Sounds good, @aneeshd16. Thanks for addressing the feedback 👍 

p.s. it needs a rebase. I guess we'll ship the base PR first and then this, or can it go separately into `trunk`? Obviously, as you prefer. This also served well for testing the underlying refactors. I plan on shipping https://github.com/Automattic/wp-calypso/pull/74960 a little later or tomorrow - replied to your comments there in case you have other suggestions :-)

aneeshd16:
Thanks @chriskmnds! Let's ship the base PR first, and I'll rebase this one after that. I'll have a look at https://github.com/Automattic/wp-calypso/pull/74960 soon!

chriskmnds:
LGTM after rebase

PR #74975, opened 2023-03-27
Payments: Convert BillingHistoryList to TypeScript

sirbrillig:
## Proposed Changes

This PR cleans up the `BillingHistoryList` component and its dependencies, and converts them all to TypeScript.

This was not a straightforward conversion because the component and its functions used a number of techniques that were highly dynamic. In order to complete the conversion, I had to simplify many of these operations.

Screenshot 2023-03-27 at 3 15 16 PM

## Testing Instructions

Use an account which has a purchase history.

View `/me/purchases/billing` with and without this PR and make sure that they show the same results.

Select some of the filters (Month and App) at the top of the page and verify that they work as expected.


client/me/purchases/billing-history/billing-history-list.tsx
```diff
@@ -50,11 +79,7 @@ class BillingHistoryList extends Component {
 		);
 	}
 
-	serviceName = ( transaction ) => {
-		if ( ! transaction.items ) {
```
sirbrillig, line 54: I don't think this condition is ever hit because all past transactions should include a list of items. And that's good because there's no way that `serviceNameDescription()` would have operated on the transaction itself without fatal errors.

client/me/purchases/billing-history/billing-history-list.tsx
```diff
@@ -68,56 +93,50 @@ class BillingHistoryList extends Component {
 			return transactionItem.product;
 		}
 
-		return this.serviceNameDescription( {
-			...transactionItem,
-			plan: capitalPDangit( transactionItem.variation ),
```
sirbrillig, line 73: I moved this derived variable into `serviceNameDescription()` itself.

client/me/purchases/billing-history/billing-history-list.tsx
```diff
@@ -242,44 +261,109 @@ class BillingHistoryList extends Component {
...
+
+	return transactions.filter( ( transaction: BillingTransaction ) =>
+		getSearchableStrings( transaction ).some( ( val ) => {
+			const haystack = val.toString().toLowerCase();
+			return haystack.includes( needle );
+		} )
+	);
+}
+
+function filterTransactions(
```
sirbrillig, line 312: This function and its dependencies were moved here from `get-billing-transaction-filters`, which was overly complex and only ever used here.

client/me/purchases/billing-history/filter-transactions.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	return transactions.filter( ( transaction: BillingTransaction ) =>
+		getSearchableStrings( transaction ).some( ( val ) => {
+			const haystack = val.toString().toLowerCase();
+			return haystack.includes( needle );
+		} )
+	);
+}
+
+export function filterTransactions(
```
sirbrillig, line 49: This file and its test were moved here from `getFilteredBillingTransactions` because they are only used here and don't need to be a selector.

client/me/purchases/billing-history/test/filter-transactions.ts
```diff
@@ -0,0 +1,471 @@
...
+const state = {
+	billingTransactions: {
+		items: {
+			past,
+		},
+		ui: { past: undefined, upcoming: undefined },
+	},
+};
+
+describe( 'filterTransactions()', () => {
```
sirbrillig, line 235: Despite looking quite different now, these tests are functionally the same as they were when they were in `get-filtered-billing-transactions`.

spraveenitpro:
### Functional Testing:

<details>
<summary>Before applying the patch: </summary>


</details>


<details>
<summary>After applying the patch: </summary>


</details>

spraveenitpro (approved):
Static analysis looks good, great opportunity to revise some TS concepts!

PR #74991, opened 2023-03-27
Add couple of tweaks to reader notifications

eoigal:
This PR does 2 things;

* Adds page title to the reader notifications page - https://calypso.localhost/read/notifications/
* Forces the bell/notification icon to active to highlight the fact that these notifications are the same as the iframed notifications

Screenshot 2023-03-28 at 10 17 01

### DIFF
--- a/client/layout/masterbar/masterbar-notifications/notifications-button.jsx
+++ b/client/layout/masterbar/masterbar-notifications/notifications-button.jsx
@@ -30,6 +30,7 @@ class MasterbarItemNotifications extends Component {
 	state = {
 		animationState: 0,
 		newNote: this.props.hasUnseenNotifications,
+		forceIsActive: window.location.pathname === '/read/notifications',
 	};
 
 	componentDidUpdate( prevProps ) {
@@ -110,7 +111,7 @@ class MasterbarItemNotifications extends Component {
 
 	render() {
 		const classes = classNames( this.props.className, 'masterbar-notifications', {
-			'is-active': this.props.isNotificationsOpen,
+			'is-active': this.props.isNotificationsOpen || this.state.forceIsActive,
 			'has-unread': this.state.newNote,
 			'is-initial-load': this.state.animationState === -1,
 		} );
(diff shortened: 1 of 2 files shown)


client/layout/masterbar/masterbar-notifications/notifications-button.jsx
```diff
@@ -110,7 +111,7 @@ class MasterbarItemNotifications extends Component {
 
 	render() {
 		const classes = classNames( this.props.className, 'masterbar-notifications', {
-			'is-active': this.props.isNotificationsOpen,
+			'is-active': this.props.isNotificationsOpen || this.state.forceIsActive,
```
tyxla: I don't see a good reason for this to be in the component state. Can't we just inline `window.location.pathname === '/read/notifications'`?

client/reader/notifications/controller.jsx
```diff
@@ -11,15 +12,18 @@ export function notifications( context, next ) {
...
-			<AsyncLoad
-				require="calypso/notifications"
-				isShowing={ true }
-				checkToggle={ () => {} }
-				setIndicator={ () => {} }
-				placeholder={ null }
-			/>
-		</div>
+		<>
+			<DocumentHead title="Notifications" />
```
tyxla: Is there a good reason not to put the `DocumentHead` in the `Notifications` component?

tyxla: This "Notifications" string isn't localized. I think we should localize it.

eoigal: I'll update that. The other documenthead's I saw in Reader didn't localize either so assumed this was intentional.

eoigal: The `DocumentHead` isn't shown in the iframe and I want to avoid making unnecessary changes to `Notifications` component if possible.

client/layout/masterbar/masterbar-notifications/notifications-button.jsx
eoigal: Good idea, I'll make this change.

client/reader/notifications/controller.jsx
tyxla: > The other documenthead's I saw in Reader didn't localize either so assumed this was intentional.

I think this needs to be updated, otherwise the Reader titles won't be properly translated when using a different UI language.

PR #74992, opened 2023-03-27
Hide variant picker for Akismet Personal plans

CodeyGuyDylan:
## Proposed Changes

This diff removes the product variant picker for Akismet Personal plans, as we are only planning on selling yearly plans for now

Notes:
 * There is a bug right now where a yearly plan is pro-rated due to wpcom detecting the first Akismet Personal subscription ever purchased and reducing the price based on that. A fix for that is in progress
 * The variant picker will still currently show for Akismet Personal Monthly plans, but this product will automatically be removed from the cart when some new validation is added soon: 1201210512993648-as-1204264517688090/f

## Testing Instructions

1. Checkout this branch
2. Sandbox `public-api.wordpress.com`
3. Make sure you are using the Store Sandbox
4. Head to http://calypso.localhost:3000/checkout/akismet/ak_personal_yearly:-q-36
5. Make sure the display of the pricing is not confusing and makes sense (this is where the pro-rated discount mentioned in the `Proposed Changes` section will show up, you can ignore this aspect)
image
6. Make sure the variant picker does not show up
7. Go to http://calypso.localhost:3000/me/purchases and find your new subscription
8. Open the manage purchase page associated with it and make sure everything looks good
image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/test/util/index.ts
+++ b/client/my-sites/checkout/composite-checkout/test/util/index.ts
@@ -127,6 +127,24 @@ export const stateList = {
 
 export const siteId = 13579;
 
+export const akismetPersonalProduct: ResponseCartProduct = {
+	...getEmptyResponseCartProduct(),
+	product_name: 'Akismet Personal (Paid)',
+	product_slug: 'ak_personal_yearly',
+	currency: 'USD',
+	extra: {
+		isAkismetSitelessCheckout: true,
+	},
+	meta: 'foo.com',
+	product_id: 2310,
+	volume: 30,
+	is_domain_registration: false,
+	item_original_cost_integer: 30000,
+	item_subtotal_integer: 30000,
+	bill_period: '365',
+	months_per_bill_period: 12,
+};
+
 export const domainProduct: ResponseCartProduct = {
 	...getEmptyResponseCartProduct(),
 	product_name: '.cash Domain',
--- a/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
@@ -1,4 +1,8 @@
-import { isAkismetProduct, isJetpackPurchasableItem } from '@automattic/calypso-products';
+import {
+	isAkismetPersonalProduct,
+	isAkismetProduct,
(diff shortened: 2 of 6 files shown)


sirbrillig (approved):
I haven't testing (I leave that to your team) but from a Payments perspective this looks good to me! Thanks for adding a test!

client/my-sites/checkout/composite-checkout/test/util/index.ts
```diff
@@ -127,6 +127,24 @@ export const stateList = {
 
 export const siteId = 13579;
 
+export const akismetPersonalProduct: ResponseCartProduct = {
```
robertsreberski, line 130: Should we also add a test case for `product_slug: 'ak_personal_monthy'`?

It's possible to enter checkout also with this product (or am I wrong)? Should we actually offer this product in the first place, if we don't want to sell it?

DavidRothstein (changes requested):
>as we are only planning on selling yearly plans for now

@michaeldcain pointed out that the monthly version of this product is already marked as not available to be sold in the server-side code.  So the bug here is actually on the server.  We should never be offering someone a variant in the variant picker that can't actually be purchased, because if they try to buy it, it will fail.

I'll create an issue to fix this server-side.  I don't think you will need any Calypso changes here at all.

DavidRothstein:
>I'll create an issue to fix this server-side.

Issue created here (should be a very easy fix, I think): https://github.com/Automattic/payments-shilling/issues/1508

robertsreberski (commented):
The PR tests well, but I would address first disabling `ak_personal_monthly` altogether

client/my-sites/checkout/composite-checkout/test/util/index.ts
CodeyGuyDylan: I actually just added validation to remove this product on the server side, as well as a test there so I don't think that is necessary here. This whole PR might actually be unnecessary 😄

sirbrillig:
> Issue created here (should be a very easy fix, I think): https://github.com/Automattic/payments-shilling/issues/1508

D106310-code may make this PR unnecessary.

CodeyGuyDylan:
Closing this PR as it is probably not needed. Will reopen if necessary

PR #74996, opened 2023-03-27
Shopping Cart: Remove create_new_blog property

sirbrillig:
## Proposed Changes

A shopping cart request can contain the `create_new_blog` property. The intent of this property was for the transactions endpoint to create a new site for a user as part of the transaction. However, it's just a flag to represent a cart with no `blog_id` and we can more reliably use that instead.. Most of its functionality was already removed by https://github.com/Automattic/wp-calypso/pull/69837 but some still remains and it causes a lot of confusion.

This PR removes all mention of `create_new_blog` from calypso. The server-side use of `create_new_blog` was removed in D106477-code.

Part of the work in https://github.com/Automattic/payments-shilling/issues/759

## Testing Instructions

Reviewers could verify that removing the property has no effect on siteless checkout:

- Visit `/start/domain/domain-only` and complete the flow to purchase a new domain without a site.
- Verify that checkout completes successfully and that you are redirected to a post-checkout page for that domain.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-create-payment-complete-callback.tsx
@@ -107,7 +107,11 @@ export default function useCreatePaymentCompleteCallback( {
 			// created site in the Thank You page URL.
 			// TODO: It does not seem like this would be needed for Akismet, but marking to follow up
 			let jetpackTemporarySiteId;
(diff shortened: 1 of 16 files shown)


sirbrillig:
Pinging @jboland88 for a review as well because his work on a backend diff reminded me of this and will probably be affected if this gets merged.

sirbrillig:
We just need to confirm that the logging that proves this is not being used is correct (or can safely be replaced). See D106290-code

sirbrillig:
Before we merge this we should change the backend to remove create_new_blog there: D106477-code

Issue #74968, opened 2023-03-27
`/subscriptions/sites`: Make sure the list of sites the user is subscribed to has a proper mobile view


ivan-ottinger:
🔴 Blocked: [This PR](https://github.com/Automattic/wp-calypso/pull/75064) has been merged with mocked data. When we retrieve real user data we can finish this task.

## Acceptance Criteria

Make sure the "Sites" list has a proper mobile view:

![Image](https://user-images.githubusercontent.com/25105483/227983852-edaa3fb6-5e8b-46e4-848a-867bd727a27d.png)

Related design: inDLaEQV8jJ21O4WXawIsJ-fi-712-13475.

Please ignore the search field, site type drop-down and "..." menu in each row. These will be covered in a separate task.

JuanLucha:
Like the desktop version, this task should be closed when the data shown is not mocked. Opening the issue again.

mashikag:
> Like the desktop version, this task should be closed when the data shown is not mocked. Opening the issue again.

@JuanLucha, as far as I understand, after re-opening the issue, now its status should be to-do. Could you please change the status if that is the case? 🙂

PR #75014, opened 2023-03-28
Add/ List of sites in Subscription Manager (Desktop version)

### LINKED ISSUE #74966: `/subscriptions/sites`: Display all sites user is subscribed to - use mock data.
The goal of this task is to introduce the following list of sites the user is subscribed to (using mock data - as the API is not ready yet):

Markup on 2023-03-27 at 17:00:12

## Acceptance Criteria

**✅ Changes the task covers**

Looking at the design (inDLaEQV8jJ21O4WXawIsJ-fi-712-15534), display the list of all sites the user is subscribed to, including the following columns:

- Subscribed site
  - site icon (circular mask should be applied; we need to handle situation when there's no icon)
  - site title (we need to handle situation when there's no title; long title should be addressed with ellipsis)
  - site URL (site URL should be properly formatted - as can be seen in the design;  long URL should be addressed with ellipsis)
- Since (the date needs to be properly formatted - maybe respecting user's locale if possible) 
- Email Frequency

Please use mock data. Focus on the desktop view and check the RTL version.

What should happen when user clicks on the site title, site icon or the URL is currently being discussed.

**❌ Changes this task does not cover**

Some of the changes will be handled in separate tasks that are linked below.

- "New Posts" column
- "New Comments" column
- "..." (three dots) that will later open a pop-up: https://github.com/Automattic/wp-calypso/issues/74973
- Mobile view: https://github.com/Automattic/wp-calypso/issues/74968
- Retrieving and displaying user's real data: https://github.com/Automattic/wp-calypso/issues/75068
- Blue badge / tag for P2 sites: https://github.com/Automattic/wp-calypso/issues/75078

We are neither implementing the filter, search, nor site type drop-down as a part of this task.

JuanLucha:
Closes https://github.com/Automattic/wp-calypso/issues/74966

## Proposed Changes

As part of the Subscription Manager in Cmustinitiative, we need to implement the list of sites to the Subscription Manager. This PR adds it to the desktop version.

This is the result after these changes:

Screenshot 2023-03-28 at 11 45 13

## Testing Instructions

- Go to http://calypso.localhost:3000/subscriptions/sites
- Check the design matches the one in Figma: inDLaEQV8jJ21O4WXawIsJ-fi-712-15534

## Notes

- The mobile version of this list is outside the scope of this PR, and the issue for that is here: https://github.com/Automattic/wp-calypso/issues/74968


packages/subscription-manager/src/components/TabsSwitcher/TabsSwitcher.tsx
```diff
@@ -23,7 +23,7 @@ type TabsSwitcherProps = {
 const getRoute = ( baseRoute: string, path: string ): string => {
 	// This is a temporary redirect to the re-skinned Subscription Management portal.
 	// Once the new "Commnets" and "Sites" views are ready, this will be removed.
-	const temporaryRedirect = [ 'comments', 'sites' ];
+	const temporaryRedirect = [ 'comments' ];
```
ivan-ottinger: I think we should keep this out of the PR for now - since we don't want to expose the unfinished Sites view to our users yet (unless I missed something).

Instead, we could add the change to the PR test instructions.

JuanLucha: That's a good point. I'll do it 👍

mashikag: > Instead, we could add the change to the PR test instructions.

The code change should not be required to test. We can use the direct path: http://calypso.localhost:3000/subscriptions/sites

client/landing/subscriptions/tab-views/sites.tsx
```diff
@@ -0,0 +1,58 @@
...
+				<li className="row" key={ site.id }>
+					<span className="title-box">
+						<img className="icon" src={ site.icon } alt={ site.name } />
+						<span className="title-column">
+							<span className="name">{ site.name }</span>
+							<span className="url">{ site.url }</span>
+						</span>
+					</span>
+					<span className="date">{ site.date.toDateString() }</span>
+					<span className="post-notifications">{ site.newPostNotifications }</span>
```
mashikag: The post notifications and comment notifications were meant to be left out for now. See pdDOJh-1H7-p2

client/landing/subscriptions/tab-views/sites.tsx
```diff
@@ -0,0 +1,58 @@
...
+		<ul className="site-list">
+			<li className="row header">
+				<span className="title-box">{ translate( 'Subscribed site' ) }</span>
+				<span className="date">{ translate( 'Since' ) }</span>
+				<span className="post-notifications">{ translate( 'New posts' ) }</span>
+				<span className="comment-notifications">{ translate( 'New comments' ) }</span>
+				<span className="email-frequency">{ translate( 'Email frequency' ) }</span>
+			</li>
+			{ sites.map( ( site ) => (
+				<li className="row" key={ site.id }>
```
mashikag: If it was me, I would create a `Site`, or `SiteRow`, component out of this block.

client/landing/subscriptions/tab-views/sites.tsx
```diff
@@ -0,0 +1,58 @@
...
+		} );
+	}
+	return sites;
+};
+
+export default function SitesView() {
+	const translate = useTranslate();
+	const sites = useSites();
+
+	const commentYes = <Gridicon icon="checkmark" size={ 18 } className="green" />;
```
mashikag: The `commentYes` and `commentNo` should be defined as a component outside the render of the `SitesView` component.

client/landing/subscriptions/tab-views/sites.tsx
```diff
@@ -0,0 +1,58 @@
...
+	const commentNo = <Gridicon icon="cross" size={ 18 } className="red" />;
+
+	return (
+		<ul className="site-list">
```
mashikag: I know the structure of the whole list may change when the infinite scroll is applied, but I think we may already include the appropriate `role` attributes for headers, list items, cells etc. What do you think? Maybe to reduce the scope for yourself, you can concentrate on building the `Site` component on its own for this PR?

packages/subscription-manager/src/components/TabsSwitcher/TabsSwitcher.tsx
JuanLucha: Yes, that's how I have the test instructions.

client/landing/subscriptions/tab-views/sites.tsx
JuanLucha: Very true 🤦‍♂️
Changing it. Thanks for the heads up!

client/landing/subscriptions/tab-views/sites.tsx
JuanLucha: Yeah, nice. I'll do it.

client/landing/subscriptions/tab-views/sites.scss
```diff
@@ -0,0 +1,73 @@
+@import "@automattic/color-studio/dist/color-variables";
+
+$max-list-width: 1300px;
+
+.site-list {
```
phcp: Maybe switch to more specific classes would be interesting to avoid collisions with generic names like row and header. Something like `subscriptions-manager__sites-table`, `subscriptions-manager__sites-table-header`, `subscriptions-manager__sites-row`. What do you think?

client/landing/subscriptions/tab-views/sites.tsx
JuanLucha: Not required now.

client/landing/subscriptions/tab-views/sites.tsx
JuanLucha: Indeed.

client/landing/subscriptions/tab-views/sites.scss
JuanLucha: I'm taking advantage of the compound nature of SCSS. I understand long and composed class names like `subscriptions-manager__sites-table` are suitable for CSS but overlap with SCSS.

phcp: Got your point. My concern is about external CSS that may be loaded on the same page. If there is no other CSS loaded on the page than ours, it should be fine to have classes like this. But if we have external components with similar selectors, they may affect your components. If you feel this is not an issue we can close this thread.

JuanLucha: Good point. As everything is inside `.site-list`, changing that class for a more specific one should change everything inside it after transpilation. Thanks!

client/landing/subscriptions/tab-views/Site.tsx
```diff
@@ -0,0 +1,30 @@
+import { useTranslate } from 'i18n-calypso';
+import SiteRow from './SiteRow';
+import { SiteType } from './site-types';
+
+type SiteProps = {
+	sites: SiteType[];
+};
+
+export default function Site( { sites }: SiteProps ) {
```
mashikag: Hmm, `Site` does not sound right to me as a name for the component. The component is responsible for displaying a list of sites. What do you think yourself, @JuanLucha? 🤔

client/landing/subscriptions/tab-views/sites.scss
```diff
@@ -0,0 +1,74 @@
+@import "@automattic/color-studio/dist/color-variables";
```
mashikag: We tend to call style files as `styles.scss` within our entry point. I am not sure what is the calypso standard, though. 🙈 Do you mind checking what is the tendency within the repo? 🙏

client/landing/subscriptions/tab-views/SiteRow.tsx
```diff
@@ -0,0 +1,21 @@
+import { SiteType } from './site-types';
```
mashikag: I know in the past I got you to use the Pascal case for file names, but we should actually be using kebab_case instead, as the rest of the calypso. 🙈

client/landing/subscriptions/tab-views/Site.tsx
```diff
@@ -0,0 +1,30 @@
+import { useTranslate } from 'i18n-calypso';
```
mashikag: Same here it should be kebab case, hence the filename should start with lower case.

client/landing/subscriptions/tab-views/Site.tsx
JuanLucha: True, `Sites` or `SiteList` sounds better. In fact, having the child named `SiteRow`, `SiteList` sounds good to me.

client/landing/subscriptions/tab-views/sites.scss
JuanLucha: While there are some examples of style files called differently, you are right regarding the tendency being `styles.scss`. I called this file my component because we have different components in the same folder. But I guess there is no problem using the same file for all of them 👍

client/landing/subscriptions/tab-views/SiteRow.tsx
JuanLucha: Done 👍

client/landing/subscriptions/tab-views/Site.tsx
JuanLucha: Done 👍

client/landing/subscriptions/tab-views/sites.tsx
```diff
@@ -0,0 +1,24 @@
+import SiteList from './site-list';
```
phcp: What do you think about moving `site-list`, `site-row`, `site-types`, and `styles` to a new folder `client/landing/subscriptions/site-list`? The tab-views folder seems to hold only the entry points for each tab, like on UserSettings.

client/landing/subscriptions/tab-views/site-types.tsx
```diff
@@ -0,0 +1,9 @@
+// typo de dato con estos valores: id, name, icon, url, date, emailFrequency
```
phcp: Could you delete this comment? Not necessary for this type

phcp (commented):
Overall looks good, just added two minor suggestions. Let me know what you think.

client/landing/subscriptions/tab-views/site-types.tsx
JuanLucha: Sure!

client/landing/subscriptions/tab-views/sites.tsx
JuanLucha: Good point. Done 👍

client/landing/subscriptions/styles.scss
```diff
@@ -42,4 +42,14 @@ body {
 	ul {
 		padding: 0;
 	}
+
```
mashikag: I do not see the `.green` and `.red` classes being used anywhere? 🤔

mashikag (approved):
All looks good to me! 👍

client/landing/subscriptions/styles.scss
JuanLucha: True. Removed 👍

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7893948

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7893948).

PR #75015, opened 2023-03-28
Refactor away from Lodash functionality in bin/audit-svg-colors

yashitamittal11:
## Description

This change is for a utility that audits .svg files. I have modified the code to remove the Lodash functions from this component.

## Proposed Changes

* Replaced the following Lodash functions with the native JS code
1. _.pickBy
2. _.sortBy
3. _.forInRight
4. _.startsWith

Why?
Lodash is known to inflate the bundle size of packages unnecessarily, and in most cases, it can be replaced with native language functionality. 

## Testing Instructions

- Open the file bin/audit-svg-colors.js
- Place the breakpoints at all the lines where the Lodash functions are replaced with the native JS code.
- Click on Run and Debug
- Ensure the values returned by the new functionality are the same as those returned with the Lodash functions.

## Checklist
Verify all tests still pass.


bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
+const sortBy = () => {
+	return ( objA, objB ) => {
+		if ( objA.to.name > objB.to.name ) {
+			return 1;
+		} else if ( objB.to.name > objA.to.name ) {
+			return -1;
+		}
+		return 0;
+	};
+};
```
tyxla: I think we can unwrap this function since we'll always need to invoke it to get the actual sorting function as a result:

```suggestion
const compareByName = ( objA, objB ) => {
	if ( objA.to.name > objB.to.name ) {
		return 1;
	} else if ( objB.to.name > objA.to.name ) {
		return -1;
	}
	return 0;
};
```

As you can see, we also renamed the function, since it's now a comparator function.

Alternatively, we could use [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) since it is supported in Node.js.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
 
 /**
  * Palette color subsets
  */
 
+/**
+ * pickBy function is a replacement for the Lodash ._pickby
+ *
+ * @param {*} obj a palette of colors
```
tyxla: If the object is always a palette of colors, should we rename the argument to `palette`?

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
+ * @param {*} fn This function returns truthy or falsy value to filter the colors
+ * @returns objext with filtered color values
+ * Used Object.keys() and Array.prototype.filter()to remove the keys for which fn returns a falsy value.
```
tyxla: Feel free to remove this line, I believe it should be pretty straightforward what it does by reading the code.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
+ * Used Array.prototype.reduce() to convert the filtered keys back to an object with the corresponding key-value pairs.
```
tyxla: Feel free to remove this line as well.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
-const PALETTE_ILLUSTRATION_COLORS = _.pickBy( PALETTE.colors, ( colorValue, colorName ) => {
-	// Avoid using pure black
-	if ( colorValue === '#000' ) {
-		return;
-	}
-	// Avoid specific colors for illustration use
-	return ! _.startsWith( colorName, 'Simplenote Blue' );
-} );
+const PALETTE_ILLUSTRATION_COLORS = PALETTE?.colors
+	? pickBy( PALETTE.colors, ( colorValue, colorName ) => {
```
tyxla: You could avoid the ternary here by making `pickBy()` a bit more flexible, and changing:

```
Object.keys( obj )
```

to:

```
Object.keys( obj ?? {} )
```

so `obj` will always be an object.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
-	}
-	// Don’t use brand colors for any WordPress.com app images
-	return ! (
-		_.startsWith( colorName, 'Simplenote Blue' ) ||
-		_.startsWith( colorName, 'WooCommerce Purple' ) ||
-		_.startsWith( colorName, 'WordPress Blue' )
-	);
-} );
+const PALETTE_APP_COLORS = PALETTE?.colors
+	? pickBy( PALETTE.colors, ( colorValue, colorName ) => {
```
tyxla: Same as above on the ternary.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
 // The subset of palette colors allowed in illustrations
-const PALETTE_ILLUSTRATION_COLORS = _.pickBy( PALETTE.colors, ( colorValue, colorName ) => {
-	// Avoid using pure black
-	if ( colorValue === '#000' ) {
-		return;
-	}
-	// Avoid specific colors for illustration use
-	return ! _.startsWith( colorName, 'Simplenote Blue' );
-} );
+const PALETTE_ILLUSTRATION_COLORS = PALETTE?.colors
```
tyxla: Are we ever going to have a use case where `PALETTE` won't be an object? I think it will always be an object and therefore the optional chaining might be unnecessary.

bin/audit-svg-colors.js
```diff
@@ -257,16 +290,14 @@ function findClosestColor( value, targetValues ) {
...
-		if ( paletteValue === value ) {
-			name = paletteColorName;
-			return false;
-		}
-	} );
-
+	const name = PALETTE?.colors
+		? Object.keys( PALETTE.colors )
+				.reverse()
+				.find( ( paletteColorName ) => {
```
tyxla: Instead of calling `.reverse().find()` we could just use [`Array.prototype.findLast()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast).

bin/audit-svg-colors.js
```diff
@@ -257,16 +290,14 @@ function findClosestColor( value, targetValues ) {
...
 	// Iterating from right to make sure color name aliases aren’t caught
-	_.forInRight( PALETTE.colors, ( paletteValue, paletteColorName ) => {
-		if ( paletteValue === value ) {
-			name = paletteColorName;
-			return false;
-		}
-	} );
-
+	const name = PALETTE?.colors
+		? Object.keys( PALETTE.colors )
```
tyxla: Same on the ternary as above.

bin/audit-svg-colors.js
```diff
@@ -257,16 +290,14 @@ function findClosestColor( value, targetValues ) {
...
-
 	// Iterating from right to make sure color name aliases aren’t caught
-	_.forInRight( PALETTE.colors, ( paletteValue, paletteColorName ) => {
-		if ( paletteValue === value ) {
-			name = paletteColorName;
-			return false;
-		}
-	} );
-
+	const name = PALETTE?.colors
```
tyxla: Same on the `PALETTE` optional chaining as above.

bin/audit-svg-colors.js
```diff
@@ -16,36 +16,69 @@ const { readFileSync } = require( 'fs' );
...
-		return;
-	}
-	// Don’t use brand colors for any WordPress.com app images
-	return ! (
-		_.startsWith( colorName, 'Simplenote Blue' ) ||
-		_.startsWith( colorName, 'WooCommerce Purple' ) ||
-		_.startsWith( colorName, 'WordPress Blue' )
-	);
-} );
+const PALETTE_APP_COLORS = PALETTE?.colors
```
tyxla: Same as above on the optional chaining.

bin/audit-svg-colors.js
```diff
@@ -293,12 +324,19 @@ function printReplacementRules( replacementObjects ) {
 }
 
 function formatReplacementRules( rules ) {
-	return _.sortBy( rules, 'to.name' ).map( ( rule ) => {
-		const valueFrom = rule.from.value.padEnd( 7 );
-		const valueTo = rule.to.value.padEnd( 7 );
-
-		return `${ valueFrom } → ${ valueTo } (${ rule.to.name })`;
-	} );
+	if ( rules && rules.length ) {
```
tyxla, line 312: Previously, this function would always return an array. Now if `rules` is an empty array or not an array, it will return `undefined`. Sounds like we should take care of this.

bin/audit-svg-colors.js
```diff
@@ -293,12 +324,19 @@ function printReplacementRules( replacementObjects ) {
...
+		/*The native sort modifies the array in place, so we use `.concat()`
+    	to copy the array, then sort.*/
+		return rules
+			.concat()
```
tyxla: Feel free to use `[ ...rules ]` as well, instead of `.concat()` to create a copy of the array.

bin/audit-svg-colors.js
```diff
@@ -293,12 +324,19 @@ function printReplacementRules( replacementObjects ) {
...
 function formatReplacementRules( rules ) {
-	return _.sortBy( rules, 'to.name' ).map( ( rule ) => {
-		const valueFrom = rule.from.value.padEnd( 7 );
-		const valueTo = rule.to.value.padEnd( 7 );
-
-		return `${ valueFrom } → ${ valueTo } (${ rule.to.name })`;
-	} );
+	if ( rules && rules.length ) {
+		/*The native sort modifies the array in place, so we use `.concat()`
+    	to copy the array, then sort.*/
```
tyxla: Feel free to remove that comment, especially if you change it to use Array spread like I suggest below.

tyxla (commented):
Nice work! Thanks for working on Lodash removal 😍 

I've left a few suggestions, let me know what you think.

bin/audit-svg-colors.js
yashitamittal11: Yes, it will clarify things since we are not using pickBy anywhere else.

bin/audit-svg-colors.js
yashitamittal11: I have returned an empty array, so undefined is not returned from the function.

bin/audit-svg-colors.js
```diff
@@ -16,34 +16,56 @@ const { readFileSync } = require( 'fs' );
...
+	}
+	return 0;
+};
 
 /**
  * Palette color subsets
  */
 
+/**
+ * pickBy function is a replacement for the Lodash ._pickby
```
tyxla:
```suggestion
 * pickBy function is a replacement for the Lodash _.pickBy
```

bin/audit-svg-colors.js
```diff
@@ -16,34 +16,56 @@ const { readFileSync } = require( 'fs' );
 const path = require( 'path' );
 const PALETTE = require( '@automattic/color-studio' );
 const chroma = require( 'chroma-js' );
-const _ = require( 'lodash' );
+
+/**
+ * Native Sort - replacement of ._sortBy Lodash function
```
tyxla: I assume you meant `_.sortBy` since that's the way we invoke a Lodash function when it's globally declared.

```suggestion
 * Native Sort - replacement of _.sortBy Lodash function
```

bin/audit-svg-colors.js
```diff
@@ -257,16 +279,12 @@ function findClosestColor( value, targetValues ) {
...
-	_.forInRight( PALETTE.colors, ( paletteValue, paletteColorName ) => {
-		if ( paletteValue === value ) {
-			name = paletteColorName;
-			return false;
-		}
-	} );
-
+	const name = Object.keys( PALETTE.colors ?? {} )
+		.reverse()
+		.find( ( paletteColorName ) => {
```
tyxla: I think we can still use `findLast` instead of having to call both `reverse` and `find`?

tyxla (approved):
Works well - I ran `node ./bin/audit-svg-colors.js` locally and got the same result as on trunk 👍 

Code also looks good, thanks!

I've left a couple of optional suggestions, feel free to address them or not before shipping.

🚀

yashitamittal11:
> I've left a couple of optional suggestions, feel free to address them or not before shipping.

I have implemented all the suggestions including findLast(). It worked when I checked again. Thanks for the review and feedback. 🙏

tyxla:
Thanks, @yashitamittal11, keep up the great work! 🚀

PR #75030, opened 2023-03-28
Logging: invalidate query and reset scroll ID when switching tabs

mpkelly:
Fixes https://github.com/Automattic/dotcom-forge/issues/2065

## Proposed Changes

* Fixes the bug described in https://github.com/Automattic/dotcom-forge/issues/2065 which is caused by both log queries sharing the same `scroll_id`. The app now resets the `scroll_id` and invalidates the old query to free up memory. 

## Testing Instructions

* On trunk, reproduce the issue by following the steps at https://github.com/Automattic/dotcom-forge/issues/2065
* Switch to this branch and navigate to `/site-logs/:my-atomic-site`
* Verify the issue is fixed and that the total records count now updates for each tab. 

## Pre-merge Checklist

### DIFF
--- a/client/data/hosting/use-site-logs-query.ts
+++ b/client/data/hosting/use-site-logs-query.ts
@@ -37,6 +37,22 @@ export function useSiteLogsQuery(
 	const [ scrollId, setScrollId ] = useState< string | undefined >();
 	const [ isFinishedPaging, setIsFinishedPaging ] = useState< boolean >( false );
 
+	// The scroll ID represents the state of a particular set of filtering arguments. If any of
+	// those filter arguments change we throw out the scroll ID so we can start over.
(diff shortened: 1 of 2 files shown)


p-jackson:
Thanks for fixing @mpkelly! I thought there was something going on with the server too, but it was actually my code 😅 

I've updated how the queries get invalidated. Rather than the UI explicitly calling an "invalidate" function when it thinks it needs to, the `useSiteLogsQuery` hook now deals with it's own internal state itself. When it sees the scroll ID is no longer valid (which doesn't just happen when the log type changes, it's also when the time range or page size changes) it throws it out.

I also switched from `queryClient.invalidateQueries()` to `queryClient.removeQueries()`. I tested this in the dev console to see which was actually removing the data from memory and it looks like `removeQueries` is what we want.

p-jackson (approved):

PR #75038, opened 2023-03-28
Add useSubscriptionManagerSiteSubscriptionsQuery()

### LINKED ISSUE #74963: `/subscriptions/sites`: Add useSiteSubscriptionsQuery for getting subscribed sites
When #74762 is in place, create a `useSiteSubscriptionsQuery` that returns the subscriptions the current user is subscribed to.

## Acceptance Criteria

**✅ Changes the task covers**

- Hook calls the API with page size of 200, and makes API calls to get all subscriptions
- Hook returns data in the shape the component #74762 expects
- Hook provides optional `filter`  & `sort` callbacks that allow for data filtering & sorting down the line.

```js

const { data } = useSiteSubscriptionsQuery({
  filter: (obj) => ob?.title.includes(filterValue),
  sort: (a, b) => a['sortField'] > b['sortField']
}
```

**❌ Changes this task does not cover**

Any filtering or sort logic, the data is just passed to the JavaScript `.filter()` and `.sort()` methods.

TimBroddin:
Closes [#74963](https://github.com/Automattic/wp-calypso/issues/74963)

## Proposed Changes

This PR adds the `useSubscriptionManagerSiteSubscriptionsQuery()`. This hook recursively calls the `/reader/following/mine` endpoint to fetch the user's site subscriptions. 

It allows passing of an object consisting of `filter` and `sort` keys to return the filtered data:

```js
	const { data } = Reader.useSubscriptionManagerSiteSubscriptionsQuery( {
		filter: ( subscription ) => subscription.name.includes( 'Broddin' ),
		sort: ( a, b ) => ( a.name > b.name ? 1 : -1 ),
	} );

```

## Testing Instructions

⚠️ Requirements ⚠️
- This PR needs this patch applied to your sandbox: D106264-code
- This PR needs a subkey set. You can find my subkey in the patch above. Navigate to `http://calypso.localhost:3000` and set it using the Application tab from Chrome Devtools.

To be able to test this hook you need to add it somewhere to your code. I suggest adding it to `client/landing/subscriptions/user-settings/user-settings.tsx` and calling the hook from there:

```js
	const { data } = Reader.useSubscriptionManagerSiteSubscriptionsQuery();
	console.log(data);
```

You can also test filtering and sorting:

```js
	const { data } = Reader.useSubscriptionManagerSiteSubscriptionsQuery( {
		filter: ( subscription ) => subscription.name.includes( 'Broddin' ),
		sort: ( a, b ) => ( a.name > b.name ? 1 : -1 ),
	} );
	console.log(data);

```

To test the pagination, you can change the limit it fetches at once:

```js
	const { data } = Reader.useSubscriptionManagerSiteSubscriptionsQuery( {
		limit: 1,
	} );
	console.log(data);
```

## Pre-merge Checklist


ivan-ottinger:
I know this is still in draft and needs a rebase, but I will give a look already.

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
```diff
@@ -0,0 +1,71 @@
...
+	return data;
+};
+
+const defaultFilter = () => true;
+const defaultSort = () => 0;
+
+const useSubscriptionManagerSiteSubscriptionsQuery = ( {
+	filter,
+	sort,
+	limit = 200,
```
ivan-ottinger: I wonder if the limit of `200` is necessary here - since the API itself has a hard limit set to `100` on the server side (related code reference: 2fb09-pb and patch: D106264#2142123-code).

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
```diff
@@ -0,0 +1,71 @@
...
+	total_subscriptions: number;
+};
+
+type SubscriptionManagerSiteSubscriptionsQueryProps = {
+	filter?: ( item?: SiteSubscription ) => boolean;
+	sort?: ( a?: SiteSubscription, b?: SiteSubscription ) => number;
+	limit?: number;
+};
+
+const callFollowingEndPoint = async (
```
ivan-ottinger: I ❤️ the readability and cleanliness of this function.

ivan-ottinger (commented):
I have given the PR a first look, but haven't tested it yet. Left just two comments so far. 🙂

ivan-ottinger:
Btw, I have created additional task: https://github.com/Automattic/wp-calypso/issues/75068 which will use your hook. 🙂 

I can work on it once this PR is merged.

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
TimBroddin: Thank you very much  🙏🙏🙏

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
TimBroddin: You're absolutely right, this makes no sense 💯  I'll fix it!

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const defaultFilter = () => true;
+const defaultSort = () => 0;
+
+const useSubscriptionManagerSiteSubscriptionsQuery = ( {
+	filter,
+	sort,
+	limit = 100,
+}: SubscriptionManagerSiteSubscriptionsQueryProps ) => {
```
ivan-ottinger: I would say that if we want to let the callback be used without any parameters as well, I think we will need to set the default values here:
```suggestion
const useSubscriptionManagerSiteSubscriptionsQuery = ( {
	filter = defaultFilter,
	sort = defaultSort,
	limit = 100,
}: SubscriptionManagerSiteSubscriptionsQueryProps = {} ) => {
```

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
```diff
@@ -0,0 +1,71 @@
...
+			return await callFollowingEndPoint( 1, limit, isLoggedIn );
+		},
+		{
+			enabled,
+			refetchOnWindowFocus: false,
+		}
+	);
+
+	return {
+		data: data?.filter( filter || defaultFilter ).sort( sort || defaultSort ),
```
ivan-ottinger: Related to the previous comment:

```suggestion
		data: data?.filter( filter ).sort( sort ),
```

ivan-ottinger (commented):
I have noticed two more things:

### 1. default values for the callback
→ more in the two comments below

### 2. adding callback export 
I think we would also want to add the following export to `packages/data-stores/src/reader/index.ts`:

```js
export { useSubscriptionManagerSiteSubscriptionsQuery } from './queries/use-subscription-manager-site-subscriptions-query';
```

This will allow us to import the query with `import { Reader } from '@automattic/data-stores';
`.

TimBroddin:
> I have noticed two more things:
> 
> ### 1. default values for the callback
> → more in the two comments below
> 
> ### 2. adding callback export
> I think we would also want to add the following export to `packages/data-stores/src/reader/index.ts`:
> 
> ```js
> export { useSubscriptionManagerSiteSubscriptionsQuery } from './queries/use-subscription-manager-site-subscriptions-query';
> ```
> 
> This will allow us to import the query with `import { Reader } from '@automattic/data-stores'; `.

Thanks for the suggestions! I don't know where the export went, I think a merge with trunk ate it 😁 

I also committed your default values for callback changes, thank you very much! 🙏

TimBroddin:
@ivan-ottinger I've changed this to a real PR, since D106264-code has landed.

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const callFollowingEndPoint = async (
+	page: number,
+	limit: number,
+	isLoggedIn: boolean
+): Promise< SiteSubscription[] > => {
+	const data = [];
+
+	const incoming = await callApi< SubscriptionManagerSiteSubscriptions >( {
+		path: `/read/following/mine?limit=${ limit }&page=${ page }`,
```
ivan-ottinger: While testing the query limit, I found one more thing. When looking at the API (2fb19-pb), it is expecting the `number` parameter instead of the `limit`. If we use `limit` only, the pagination won't work as expected. 🙂 

If the `number` parameter isn't provided, `$subs_query['limit']` will be set to `self::PER_PAGE_MAX` and all the records will be fetched at once (with the limit of `100` of course).

ivan-ottinger (commented):
Alright, I have tested everything now. 🙂 

- I did not observe any regressions, nor warnings or errors
- sorting and filtering works as expected
- all the changes from the previous review rounds have been implemented

Found just one more 🐛 . Once that is fixed, I think we are good to go with the merge. 🙂

packages/data-stores/src/reader/queries/use-subscription-manager-site-subscriptions-query.ts
TimBroddin: Wow, that's a huge oversight on my behalf. Let me fix it.

TimBroddin:
I replaced limit with number, and it works as intended:

CleanShot 2023-03-30 at 09 16 34@2x


I also merged with @mashikag's PR: https://github.com/Automattic/wp-calypso/pull/75085
I'll hold off merging this one until that PR lands.

ivan-ottinger (approved):
Looks good, also with the Maciej's changes applied.

I did not notice any more issues and the query is run as expected. 🚀 🙂

mashikag:
@TimBroddin I think the call to this API should be implemented with [useInfiniteQuery](https://tanstack.com/query/v4/docs/react/reference/useInfiniteQuery) instead of useQuery.

TimBroddin:
> @TimBroddin I think the call to this API should be implemented with [useInfiniteQuery](https://tanstack.com/query/v4/docs/react/reference/useInfiniteQuery) instead of useQuery.

Are we planning on implementing this is an infinite scroll (where data is loaded as we scroll)? I thought we agreed on loading everything at once, like Reader does at the moment.

Happy to turn this to an useInfiniteQuery if that's the better option.

mashikag:
> Are we planning on implementing this is an infinite scroll

Yes
> (where data is loaded as we scroll)? 

No 😅

> I thought we agreed on loading everything at once, like Reader does at the moment.

Correct. However, I remember looking into the reader implementation and the API allows for a fetch of maximum 200 sites followed at a time. They have a maximum of sites to fetch set to `2000` on the frontend. If a user has more than 200 sites, they send multiple queries to fetch all the data. This is still done on render, i.e. all the queries are executed on mount.

mashikag:
> They have a maximum of sites to fetch set to 2000 on the frontend.

You may notice a flaw there, i.e. if a user has more than 2000 sites followed they will never get to see all their subscriptions. This limitation introduced by the implementation in reader is real, and it does exist.

TimBroddin:
> > Are we planning on implementing this is an infinite scroll
> 
> Yes
> 
> > (where data is loaded as we scroll)?
> 
> No 😅
> 
> > I thought we agreed on loading everything at once, like Reader does at the moment.
> 
> Correct. However, I remember looking into the reader implementation and the API allows for a fetch of maximum 200 sites followed at a time. They have a maximum of sites to fetch set to `2000` on the frontend. If a user has more than 200 sites, they send multiple queries to fetch all the data. This is still done on render, i.e. all the queries are executed on mount.

I think we might need to distinguish between virtualized list and infinite scroll here 😀 Virtualized list is the pattern we're using to reuse table cells while scrolling, while infinite scroll is the pattern where you load data asynchronously as you scroll.

What I implemented here should be enough for the consumer of the hook. Data is being fetched in pages of 100 until we have it all, then it's returned. If I use `useInfiniteQuery` here, it would also mean that calling `fetchNextPage ` would be up to the consumer of the hook. This is fine if we want real infinite scroll, but feels like adding complexity to the render component when it's for our use case.

One downside of the current implementation is that `isFetching` will be true until all data pages are loaded (instead of already showing data after the first API call). But since this API endpoint is really fast, and we won't have that many users with hundreds of subscriptions, I think the impact of this is rather low.

(I think there's also something wrong with the Reader implementation, because the maximum amount the API returns per call is 100 and not 200, as @ivan-ottinger discovered. But that's another thing 😁)

mashikag:
@TimBroddin thanks for looking more into this for me. The points you make justify your implementation in my opinion. Let's go with it then (just the conflicts need to be resolved). 🙂

mashikag (approved):

Issue #74403, opened 2023-03-14
Stats: Add mock data for subscriber count


grzegorz-cp:
Check in with Squad 1 and come up with a contract between the back-end and front-end regarding the data that will be used for the subscribers' chart.

Document the structure here and propose a place to store the mock data before implementation.
It might be a good idea to keep it in a `connect()` function that will be replaced with proper action and reducer in the future.

Tasks:
- [x] Document data structure
- [x] Provide mock data allowing us to create a chart

annacmc:
Notes so far:

* Squad 1 are still finalising their mock API and finalising the data structure. In the meantime, we expect that it will likely take the format of the other stats endpoint returns. CC @kangzj here for more details about the specific data being used & confirmation of the structure once it's available. 
* I think that we could create a separate mock data store in Redux (`client/state/stats`) to use temporarily until real data is available for us. I have been playing around with some approaches at how to do this today. 
* We can then use the `connect()` function along with a `mapStateToProps()` function to map the mocked data object to the component's props and use as needed in our new section.
* Once the real data is ready, we can replace the `connect()` function and remove the mock data store in redux.

annacmc:
Noting that the [following patch ](D105106-code) contains sample data structure for the API endpoint return

grzegorz-cp:
The team has discussed adding Redux with `react-query` but decided to opt for an easier implementation, using `react-query` only. At this point, this is sufficient. If we need Redux in the future, we can reference commits related to this GH issue and add Redux code.

PR #75045, opened 2023-03-28
Stats Subscribers: Use react-queries to get data 

annacmc:
Related to [#74403](https://github.com/Automattic/wp-calypso/issues/74403)

## Proposed Changes

* Extends redux store and uses react-queries to get the data
* Currently, it is pulling through mock data. future PRs will swap this to using the mock API, then eventually the real API once it's ready. 

## Testing Instructions

* Open the Calypso Live Branch. Navigate to the main stats dashboard page
* Append the following to the URL: `?flags=stats/subscribers-section`
* Scroll down to the card next to Videos and check that subscriber data is being pulled through

image

## Pre-merge Checklist


client/state/stats/subscribers/actions.js
```diff
@@ -0,0 +1,23 @@
+import {
+	STATS_SUBSCRIBERS_RECEIVE,
+	STATS_SUBSCRIBERS_REQUEST,
+	STATS_SUBSCRIBERS_REQUEST_SUCCESS,
+	STATS_SUBSCRIBERS_REQUEST_FAILURE,
+} from 'calypso/state/action-types';
+import 'calypso/state/stats/init';
+
+export function requestStatSubscribers( siteId ) {
```
kangzj: from my understanding, if we use react-query, the action wouldn't be necessary anymore or am I missing anything?

client/state/stats/subscribers/hooks/use-subscribers-query.js
```diff
@@ -0,0 +1,104 @@
...
+			} ),
+		};
+	} );
+}
+
+export default function useSubscribersQuery( siteId ) {
+	const translate = useTranslate();
+	const dispatch = useDispatch();
+
+	return useQuery( [ 'stats', 'subscribers', siteId ], querySubscribers, {
```
kangzj: I understand this is only a PoC and also I'm not super familiar with react-query - would we need to add other parameters to the `queryKey`? like `period`, `startDate`, `num` etc?

client/state/stats/subscribers/actions.js
kangzj: Another question is, if we are not going to cross reference the data, would it be easier just use `react-query` meaning we could skip Redux all together?

client/state/stats/subscribers/hooks/use-subscribers-query.js
jsnmoon: I believe you're correct, @kangzj. This should be the [`queryKey`](https://tanstack.com/query/latest/docs/react/guides/query-keys) used to deterministically cache the result of our [query function `querySubscribers`](https://tanstack.com/query/latest/docs/react/guides/query-functions).

It's something we could note in a code comment and implement in the future, though.

```suggestion
	// TODO: Account for other query parameters before release.
	return useQuery( [ 'stats', 'subscribers', siteId ], querySubscribers, {
```

client/state/stats/subscribers/hooks/use-subscribers-query.js
```diff
@@ -0,0 +1,104 @@
...
+				errorNotice(
+					translate(
+						'We were unable to retrieve your latest product details. Please try again later.'
+					),
+					{
+						id: 'partner-portal-product-families-failure',
+					}
+				)
+			);
+		},
```
jsnmoon: Could we move these error/success handling out of the hook (which handles the data request layer) and into the component (which handles visuals and UX)?

Having side effects (like error notice rendering) within the hook makes it significantly more complicated.

client/state/stats/subscribers/actions.js
jsnmoon: @kangzj brings up a good question. If we're not planning on reusing this data elsewhere, we can reduce the complexity of our approach by eliminating our additions to Redux in this PR (actions/reducer/selectors).

I have two open-ended questions: 
1. Why are we using react-query?
2. Why are we using Redux?

Cc @grzegorz-cp, who I think recommended this approach.

client/state/stats/subscribers/hooks/use-subscribers-query.js
```diff
@@ -0,0 +1,104 @@
+import { useTranslate } from 'i18n-calypso';
```
jsnmoon: Could we relocate this into a `client/my-sites/stats/hooks` folder?

(Hooks are not part of our Redux state and don't need to be placed here.)

client/my-sites/stats/site.jsx
```diff
@@ -329,6 +330,7 @@ class StatsSite extends Component {
 								/>
 							)
 						}
+						<StatsSubscribers siteId={ siteId } />
```
jsnmoon: Blocking: Could we wrap this in a feature flag, as we do in line 336?

```suggestion
					{ config.isEnabled( 'stats/subscribers-section' ) && (
						<StatsSubscribers siteId={ siteId } />
					) }
```

client/state/stats/subscribers/actions.js
grzegorz-cp: I was also wondering about skipping Redux, but using it gives us more flexibility and performance. I think it helps when navigating to another page, like a summary page and coming back or swapping periods, and most likely also when handling multiple websites. It also aligns with the way we handle other Stats data.
If we decide to extend the section with other modules for subscribers, such as a world map, it will probably be easier to handle compared to passing props.

It would work both ways, with or without Redux, so it's not obvious what to pick, and I'm open to discussion.

client/state/stats/subscribers/hooks/use-subscribers-query.js
grzegorz-cp: Yes, I think we'll have to extend it at least with a date once we start testing different periods, and the store structure will probably have to change a little as well.

jsnmoon (changes requested):
This change needs to be placed behind a feature flag. Also, it'd be good to investigate the trade-offs involved in using both react-query and Redux for subscriber counts; we can continue that discussion in the inline conversation Jasper started.

client/my-sites/stats/site.jsx
annacmc: Pushed this change just as you commented 😄 All resolved now :)

client/state/stats/subscribers/actions.js
kangzj: >  I think it helps when navigating to another page, like a summary page and coming back or swapping periods, and most likely also when handling multiple websites

These could be handled by the `queryKey`, also the results could be cached for some time (default 5 min) during the lifetime of the app. So IMHO, if we are not yet to use subscribers data  with for example visitor data (cross reference in several places), then `react-query` would work beautifully.

> It also aligns with the way we handle other Stats data.
If we decide to extend the section with other modules for subscribers, such as a world map, it will probably be easier to handle compared to passing props.

This makes sense. However I guess it wouldn't hurt if we only start with `react-query`, because for one, it does the job; for two, we can always add the states for Redux easily later when we need them. 

I'm new to the `react-query` approach too and happy to discuss more. 😄

kangzj: As a result, we would be able to get rid of the lot of tricky data handling in components, for example:

https://github.com/Automattic/wp-calypso/blob/6e6636e9633456a68032d28fafd6d734e08db379/client/my-sites/stats/stats-chart-tabs/index.jsx#L85

https://github.com/Automattic/wp-calypso/blob/6e6636e9633456a68032d28fafd6d734e08db379/client/my-sites/stats/stats-chart-tabs/index.jsx#L160

https://github.com/Automattic/wp-calypso/blob/6e6636e9633456a68032d28fafd6d734e08db379/client/my-sites/stats/stats-chart-tabs/index.jsx#L62

grzegorz-cp: Taking Redux out would simplify the implementation; simplicity is always better.
I agree that we can always add it if we need to and we've got reference code in previous commits.

jsnmoon: @annacmc: Let's go ahead and remove the Redux implementation introduced by this PR.

grzegorz-cp: I pushed a commit removing all Redux elements.

jsnmoon (approved):
Looks good. Looking forward to the next iteration!

PR #75052, opened 2023-03-29
Add overlaping Akismet product notice in checkout

robertsreberski:
## Proposed Changes

* Add a notice about old (lower tier) Akismet product when user is about to purchase a higher-tier one

## Testing Instructions

* Sandbox `public-api.wordpress.com`, enable store sandbox on your sandbox
* Purchase Akismet Personal plan (http://calypso.localhost:3000/checkout/akismet/ak_personal_yearly)
* Enter checkout with the Akismet Plus plan (http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1)
* Verify that you see the notice encouraging to deactivate Akismet Personal plan:
CleanShot 2023-03-28 at 21 19 06@2x

* Verify that the link in the notice forwards to the plan management page in `/me/purchases` dashboard

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -684,7 +684,7 @@ export default function CheckoutMain( {
 			<QueryJetpackSaleCoupon />
 			<QuerySitePlans siteId={ updatedSiteId } />
 			<QuerySitePurchases siteId={ updatedSiteId } />
-			<QueryUserPurchases />
+			{ isSiteless && <QueryUserPurchases /> }
 			<QueryPlans />
 			<QueryProducts />
 			<QueryContactDetailsCache />
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -16,6 +16,7 @@ import QueryPostCounts from 'calypso/components/data/query-post-counts';
 import QueryProducts from 'calypso/components/data/query-products-list';
 import QuerySitePlans from 'calypso/components/data/query-site-plans';
 import QuerySitePurchases from 'calypso/components/data/query-site-purchases';
+import QueryUserPurchases from 'calypso/components/data/query-user-purchases';
 import { recordAddEvent } from 'calypso/lib/analytics/cart';
 import PageViewTracker from 'calypso/lib/analytics/page-view-tracker';
 import useSiteDomains from 'calypso/my-sites/checkout/composite-checkout/hooks/use-site-domains';
@@ -683,6 +684,7 @@ export default function CheckoutMain( {
 			<QueryJetpackSaleCoupon />
 			<QuerySitePlans siteId={ updatedSiteId } />
 			<QuerySitePurchases siteId={ updatedSiteId } />
+			<QueryUserPurchases />
 			<QueryPlans />
 			<QueryProducts />
 			<QueryContactDetailsCache />
--- a/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
@@ -1,6 +1,6 @@
 import { useTranslate } from 'i18n-calypso';
-import { FunctionComponent } from 'react';
-import { Purchase } from 'calypso/lib/purchases/types';
+import type { FunctionComponent } from 'react';
+import type { Purchase } from 'calypso/lib/purchases/types';
 import PrePurchaseNotice from './prepurchase-notice';
 
 import './style.scss';
--- a/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
@@ -11,10 +11,7 @@ type Props = {
 
 const AkismetProductOverlapsOwnedProductNotice: FunctionComponent< Props > = ( { purchase } ) => {
 	const translate = useTranslate();
-	const purchaseId = purchase?.id;
-	const subscriptionUrl = purchaseId
-		? `/me/purchases/siteless.akismet.com/${ purchaseId }`
-		: '/me/purchases/';
+	const subscriptionUrl = `/me/purchases/siteless.akismet.com/${ purchase.id }`;
 
 	const message = translate(
 		'The plan you are about to purchase contains all the same features of your current plan. Consider {{link}}removing your %(product)s subscription{{/link}} to avoid paying for both plans.',
--- a/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
@@ -1,44 +1,30 @@
 import { useTranslate } from 'i18n-calypso';
 import { FunctionComponent } from 'react';
-import { useSelector } from 'react-redux';
-import { getProductBySlug } from 'calypso/state/products-list/selectors';
-import { getUserPurchases } from 'calypso/state/purchases/selectors';
+import { Purchase } from 'calypso/lib/purchases/types';
 import PrePurchaseNotice from './prepurchase-notice';
 
 import './style.scss';
 
 type Props = {
-	overlapingProductSlug: string;
-	cartProductSlug: string;
+	purchase: Purchase;
 };
 
-const AkismetProductOverlapsOwnedProductNotice: FunctionComponent< Props > = ( {
-	overlapingProductSlug,
-	cartProductSlug,
-} ) => {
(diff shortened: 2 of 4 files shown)


client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
```diff
@@ -683,6 +684,7 @@ export default function CheckoutMain( {
 			<QueryJetpackSaleCoupon />
 			<QuerySitePlans siteId={ updatedSiteId } />
 			<QuerySitePurchases siteId={ updatedSiteId } />
+			<QueryUserPurchases />
```
jboland88: Is there a way to query this conditionally? Perhaps only when the cart key is `no-site`?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+	const purchase = Array.isArray( purchases )
+		? purchases.find( ( p ) => p.productSlug === products.overlaping?.product_slug )
+		: null;
+	const purchaseId = purchase?.id;
+	const subscriptionUrl = purchaseId
+		? `/me/purchases/siteless.akismet.com/${ purchaseId }`
+		: '/me/purchases/';
+
+	const message = translate(
+		'You current own {{oldProduct/}}. The plan you are about to purchase {{newProduct/}} contains all the same functionality and benefits of your current plan. Consider {{link}}removing your {{oldProduct/}} subscription{{/link}} to avoid paying for both plans.',
```
jboland88: As you mentioned elsewhere, this text is a bit lengthy 😅 

WDYT about:
"The plan you are about to purchase contains all the same features of your current plan. Consider removing your (Akismet Old) subscription to avoid paying for both plans."

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -69,6 +76,31 @@ const PrePurchaseNotices = () => {
...
+		// Continue only if the cart includes Akismet product
+		if ( ! productSlugInCart ) {
+			return null;
+		}
+
+		const lowerTierProducts = akismetPurchases.filter(
+			( { productSlug } ) =>
+				AKISMET_PRODUCTS_LIST.indexOf( productSlug ) -
+					AKISMET_PRODUCTS_LIST.indexOf( productSlugInCart ) <
+				0
```
jboland88: Could this be simplified to 
`AKISMET_PRODUCTS_LIST.indexOf( productSlugInCart ) > AKISMET_PRODUCTS_LIST.indexOf( productSlug )`

I think this works if we assume the product list array will always be in order of progressive tiers. Would there ever be a case where that is not true?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
sirbrillig: I wonder if we should use string interpolation instead of component interpolation for these variables. Using `components` implies that each variable is a React component or HTML tag, but these just look like strings, which means we can use `args` instead (see [the documentation](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/README.md#string-substitution) for using variables in `translate`).

Also there's a typo here. It should be "You currentLY own..."
```suggestion
		'You currently own %(oldProduct)s. The plan you are about to purchase %(newProduct)s contains all the same functionality and benefits of your current plan. Consider {{link}}removing your %(oldProduct)s subscription{{/link}} to avoid paying for both plans.',
```

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
```diff
@@ -0,0 +1,55 @@
...
+		{
+			comment:
+				'The `oldProduct` variable refers to the product the customer owns already. The `newProduct` variable refers to the product they are about to purchase.',
+			components: {
+				link: <a href={ subscriptionUrl } />,
+				oldProduct: <>{ products.overlaping?.product_name }</>,
```
sirbrillig: If there is no old product (if `products.overlaping` is undefined), do we still want to show this notice? Maybe this component should return `null` in that case?

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -69,6 +76,31 @@ const PrePurchaseNotices = () => {
 		return products.filter( ( p ) => ! p.expired );
 	} );
 
+	const akismetProductSlugThatOverlapsCartProduct = useMemo( () => {
+		const productSlugInCart = cartItemSlugs.find( ( productSlug ) =>
+			isAkismetProduct( { productSlug } )
+		);
+		const akismetPurchases = userActivePurchases.filter( isAkismetProduct );
```
jboland88: Wondering if we have a way to determine if the purchase is a "site-less" purchase when filtering. If it's possible to get only "site-less" Akismet purchases, that would be preferable, but not a blocker if not - this would be some edge case handling.

jboland88 (commented):
This LGTM overall - I had just a few small comments. I think making `<QueryUserPurchases>` run conditionally would be a nice update.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
robertsreberski: Yes, we can use `domain` property from Purchase object

client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
robertsreberski: Yes, we can use `isSiteless` property in that component, thanks for the idea!

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
robertsreberski: Thanks, it makes sense to change it from component to arg. I've also reworked the text in general as per @jboland88 suggestions.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
robertsreberski: I've refactored the Notice code to actually don't even fetch Product - but rather work on `purchase` object. That simplifies the case here.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
robertsreberski: It looks similar to the default Jetpack notice - I don't mind using any of them. But the version you suggests now works for me as well, I think it looks fair and clear.

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
```diff
@@ -0,0 +1,38 @@
...
+	return (
+		<PrePurchaseNotice
+			message={ message }
+			linkUrl={ subscriptionUrl }
+			linkText={ translate( 'Manage subscription' ) }
+		/>
+	);
+};
+
+export default AkismetProductOverlapsOwnedProductNotice;
```
CodeyGuyDylan, line 38: IMO this looks a bit iffy on mobile, the Manage Subscription link is very small. Maybe we could consult somebody on the design team to make sure this looks as good as we can make it?
image

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
```diff
@@ -0,0 +1,38 @@
+import { useTranslate } from 'i18n-calypso';
+import { FunctionComponent } from 'react';
+import { Purchase } from 'calypso/lib/purchases/types';
```
CodeyGuyDylan: These could both be imported as a type instead
```suggestion
import type { FunctionComponent } from 'react';
import type { Purchase } from 'calypso/lib/purchases/types';
```

client/my-sites/checkout/composite-checkout/components/prepurchase-notices/akismet-product-overlaps-owned-product-notice.tsx
CodeyGuyDylan: I see though that you aren't updating any styles, this is probably outside the scope of this diff

CodeyGuyDylan (approved):
Left a comment about the design that is outside the scope of this diff, but wanted to make note of. Other than that, just a small comment about importing `type` and the rest looks good and tests well 😄

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7896762

Thank you @robertsreberski for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7896762).

PR #75056, opened 2023-03-29
Data Stores: Fix persistence data might be incorrect

arthur791004:
Related to #

## Proposed Changes

We use the `persistence` plugin to persist the store to the local storage on demand. However, if we register the plugin multiple times, the result will be incorrect and some of them will be overwritten with outdated data. Hence, this PR is proposing a change to wrap the registration in the single function to avoid the duplication.

## Testing Instructions

* Open the Devtool > Application > LocalStorage, and watch `WPCOM_7_DAYS_PERSISTENCE`
* When you back and forth between steps, the value of the `automattic/onboard` store should be updated.
  * For example, when you submit the selected goals, both goals, and intent will be updated to the selected values.
  * Note that the `verticalId` won't be updated since the vertical step doesn't set the value to the store. I tried to remove the `verticalId` from the store but looks like it's still used in the `storeProfiler` step

## Pre-merge Checklist


packages/data-stores/src/launch/index.ts
```diff
@@ -10,7 +10,7 @@ export type { State };
 export type { LaunchStepType } from './types';
 export { STORE_KEY };
 
-use( plugins.persistence, persistOptions );
+registerPlugins();
```
taipeicoder, line 13: So many different ways this is implemented 🥲

taipeicoder (approved):
LGTM

packages/data-stores/src/launch/index.ts
arthur791004: Yep...I also feel confused 🫠

okmttdhr (approved):
LGTM✅

Are there any other plugins we have implemented that might cause similar problems?

arthur791004:
> Are there any other plugins we have implemented that might cause similar problems?

Looks like this is the only plugin we registered multiple times 🤔

Issue #74946, opened 2023-03-27
Assembler: improving colors & fonts and style gating integration (aka Unlock this style modal)


autumnfjeld:
# Description

![image](https://user-images.githubusercontent.com/5665959/227901140-5303b462-7898-45c6-a235-c51a8216fd4d.png)

We've been discussing how the copy & UX of the style gating "Unlock this style" modal doesn't fit with the Assembler context. 

**Recent discussions:**
*  @dsas says "you can fix it, go for it" 😄  p1679478751843249/1679394673.172269-slack-C048CUFRGFQ
*  Discussion in Lego p1679459482050959-slack-CRWCHQGUB
* @miksansegundo idea pbxlJb-3d0-p2#comment-2444
* p1679459482050959-slack-CRWCHQGU
* Colors & Fonts walkthrough: pbxlJb-3EX-p2

## Proposed solution
I've made a [quick FigJam](cWkuJQiObZafHS02lDmxs6-fi-0%3A1) to outline the solution.  Ideally we'd  like @davewhitley  to mock up the final design, but perhaps we can make due with my fabulous FigJam.  These updates are for the Assembler context only, the modal shouldn't change for other contexts unless stated in the Dev Tasks items below.

## Design Tasks
- [ ] Approve the FigJam ideas.  Ideally it would be great to get an updated Figma design (but might be hard with Dave being sick. 
- [ ] Get approval on copy from Pablo.  However if he doesn't have time to review copy by the time the dev work on this is done then,  we will ship this update.  We can incorporate any Pablo-copy-updates later.

## Dev Tasks & Acceptance Criteria

- [x] https://github.com/Automattic/wp-calypso/pull/75062
  - [x] Update copy in the modal (see the FigJam cWkuJQiObZafHS02lDmxs6-fi-1%3A212)
     - [x]  First sentence: dynamic text: You’ve selected a premium { font and color | font | color}.   OR with static text "You've selected a premium color or font."
     - [x] Last sentence "You can also continue with your selected colors and font and upgrade later."
  - [x] The title on the right side of the modal should be updated to "Included with your Premium plan" .  This change can be made for all instances of the modal.  
  - [x] Update button copy in the modal to [Upgrade later] and have it close the modal - it should not go to the editor.
  - [x] When the modal closes the button in the side nav should change to [Continue]
- [x] https://github.com/Automattic/wp-calypso/pull/75057
[...]

PR #75057, opened 2023-03-29
Site Assembler: Preserve the selected layout and styles after the checkout

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/74946, p1679976634888699-slack-CRWCHQGUB

## Proposed Changes

Preserve the selected header, sections, footer, colors, and fonts after the checkout

### How

Here are possible ways to persist the data
1. Add the selected data to the query string of the returned URL
2. Add the selected data to the recipe since the recipe supports header, sections, and footer already.
3. Use the checkout modal (it worked before but it's broken after migrating to the stepper framework), and we don't need to persist any data since the user is still in the pattern assembler screen.

This PR uses the (2) option to persist the data since we might need to support editing the virtual theme in the PA screen. The virtual theme uses the `pattern_ids` in the recipe to display the patterns. Hence, we can use the same mechanism to store the selected header, sections, and footer. However, the recipe doesn't support color & font variation before, so I introduce new properties to the recipe, called `color_variation_title` and `font_variation_title`. If you have any concerns about new properties, I can just create a new store to keep this information but I guess it's fine because the virtual theme might need to support something like style variation in the future. Thus, I assume it's good to introduce them to the recipe.

Next, when the user enters the Pattern Assembler screen, we will restore the preserved data from the recipe to initialize the selected header, sections, footer, color, and font. While the user heads to the checkout, we will call the `snapshotRecipe` function to persist the selected data

Any thoughts?

### Demo

https://user-images.githubusercontent.com/13596067/228460284-33898e83-b1cc-45cf-8c96-fdf97d482407.mov

## Testing Instructions

* Go to `/setup/site-setup?siteSlug=<your_site>&flags=no-force-sympathy`
  * Note that the site should be the site with a free plan
  * Select a header
  * Select sections
  * Select a footer
  * Select a color variation
  * Select a font variation
  * Click on `Unlock this style` > `Upgrade plan`
* On the checkout screen
  * Use the browser back button to go back to the Pattern Assembler screen, ensure the selected data is preserved
  * Finish the checkout, and ensure you're redirected back to the Pattern Assembler screen the selected data is preserved as well

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
@@ -1,0 +1,130 @@
+import { useColorPaletteVariations, useFontPairingVariations } from '@automattic/global-styles';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { useState, useEffect, useRef, useCallback } from 'react';
+import { ONBOARD_STORE } from '../../../../../stores';
+import type { Pattern } from '../types';
+import type { OnboardSelect } from '@automattic/data-stores';
+import type { Design } from '@automattic/design-picker/src/types';
+import type { GlobalStylesObject } from '@automattic/global-styles';
+
+const useRecipe = ( siteId = 0, patterns: Pattern[] ) => {
+	const incrementIndexRef = useRef( 0 );
+	const [ header, setHeader ] = useState< Pattern | null >( null );
+	const [ footer, setFooter ] = useState< Pattern | null >( null );
+	const [ sections, setSections ] = useState< Pattern[] >( [] );
+	const [ colorVariation, setColorVariation ] = useState< GlobalStylesObject | null >( null );
+	const [ fontVariation, setFontVariation ] = useState< GlobalStylesObject | null >( null );
+	const selectedDesign = useSelect(
+		( select ) => ( select( ONBOARD_STORE ) as OnboardSelect ).getSelectedDesign(),
+		[]
+	);
+
+	const {
+		stylesheet = '',
+		header_pattern_ids = [],
+		footer_pattern_ids = [],
+		pattern_ids = [],
+		color_variation_title = '',
+		font_variation_title = '',
+	} = selectedDesign?.recipe || {};
+
+	const colorVariations = useColorPaletteVariations( siteId, stylesheet );
+
+	const fontVariations = useFontPairingVariations( siteId, stylesheet );
+
+	const { setSelectedDesign } = useDispatch( ONBOARD_STORE );
+
+	const selectedDesignRef = useRef< Design | undefined >( selectedDesign );
+
+	selectedDesignRef.current = {
+		...selectedDesign,
+		recipe: {
+			...selectedDesign?.recipe,
+			header_pattern_ids: header ? [ header.id ] : [],
+			footer_pattern_ids: footer ? [ footer.id ] : [],
+			pattern_ids: sections.map( ( section ) => section.id ),
+			color_variation_title: colorVariation?.title,
+			font_variation_title: fontVariation?.title,
+		},
+	} as Design;
+
+	const snapshotRecipe = useCallback( () => setSelectedDesign( selectedDesignRef.current ), [] );
+
+	/**
+	 * Initialize the default value from the recipe of the selected design when the patterns are ready
+	 */
+	useEffect( () => {
+		if ( patterns.length === 0 || ! selectedDesign?.recipe ) {
+			return;
+		}
+
+		const patternsById: { [ key: string ]: Pattern } = patterns.reduce(
+			( acc, pattern ) => ( {
+				...acc,
+				[ pattern.id ]: pattern,
+			} ),
+			{}
+		);
+
+		if ( patternsById[ header_pattern_ids[ 0 ] ] ) {
+			setHeader( patternsById[ header_pattern_ids[ 0 ] ] );
+		}
+
+		if ( patternsById[ footer_pattern_ids[ 0 ] ] ) {
+			setFooter( patternsById[ footer_pattern_ids[ 0 ] ] );
+		}
+
+		setSections(
+			pattern_ids.map( ( patternId: string | number ) => {
+				const pattern = patternsById[ patternId ];
+				return {
+					...pattern,
+					key: `${ incrementIndexRef.current }-${ pattern.id }`,
+				};
+			} )
+		);
+	}, [ patterns.length ] );
+
+	useEffect( () => {
+		if ( ! colorVariations || ! color_variation_title ) {
+			return;
+		}
+
+		const initialColorVariation = colorVariations.find(
+			( { title } ) => title === color_variation_title
+		);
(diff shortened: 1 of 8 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
```diff
@@ -0,0 +1,130 @@
...
+		},
+	} as Design;
+
+	const snapshotRecipe = useCallback( () => setSelectedDesign( selectedDesignRef.current ), [] );
+
+	/**
+	 * Initialize the default value from the recipe of the selected design when the patterns are ready
+	 */
+	useEffect( () => {
+		if ( patterns.length === 0 || ! selectedDesign?.recipe ) {
```
arthur791004: We have to wait until the patterns are ready as we will fetch the patterns from the remote server instead of hard-coding them in calypso.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
```diff
@@ -0,0 +1,130 @@
...
+		header_pattern_ids = [],
+		footer_pattern_ids = [],
+		pattern_ids = [],
+		color_variation_title = '',
+		font_variation_title = '',
+	} = selectedDesign?.recipe || {};
+
+	const colorVariations = useColorPaletteVariations( siteId, stylesheet );
+
+	const fontVariations = useFontPairingVariations( siteId, stylesheet );
```
arthur791004, line 34: We have to load colors & fonts here as they will be loaded only when the user goes to the colors or fonts screen. We can check the values of `color_variation_title` and `font_variation_title` to determine whether to load the data. But the endpoint is fast enough so I think it's fine to load the variations at the beginning.

taipeicoder (commented):
The implementation looks good! Nice reuse of the design recipe 🙂
I noticed two things:

1) After going back to the PA, the pattern names show as undefined:
Screenshot 2023-03-30 at 11 19 20 AM

2) Probably unrelated to this PR, selecting default color/font triggers this error:
Screenshot 2023-03-30 at 11 09 32 AM

arthur791004:
> After going back to the PA, the pattern names show as undefined:

Oops...I didn't notice we assign the category manually 😓 Now I pick the first one of the pattern categories to display on the list since we're not able to know the user selected this pattern by which category after checkout.

> Probably unrelated to this PR, selecting default color/font triggers this error:

Yep, this was introduced by adding track events, and I used the incorrect type definition as the variation might be `null` for the default one.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
```diff
@@ -1,13 +1,14 @@
 import { useColorPaletteVariations, useFontPairingVariations } from '@automattic/global-styles';
 import { useDispatch, useSelect } from '@wordpress/data';
+import keyBy from 'lodash/keyBy';
```
taipeicoder: Non-blocking, but it would be nice if we take this opportunity and avoiding using `lodash`, or even add a replacement for `keyBy` in `@automattic/js-utils`. As per https://github.com/Automattic/wp-calypso/blob/trunk/docs/guide/tech-behind-calypso.md#other-technologies-used-in-calypso-that-are-now-deprecated

From https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_keyBy
```js
// Lodash
console.log(_.keyBy(['a', 'b', 'c']))
// output: { a: 'a', b: 'b', c: 'c' }
console.log(_.keyBy([{ id: 'a1', title: 'abc' }, { id: 'b2', title: 'def' }], 'id'))
// output: { a1: { id: 'a1', title: 'abc' }, b2: { id: 'b2', title: 'def' } }
console.log(_.keyBy({ data: { id: 'a1', title: 'abc' }}, 'id'))
// output: { a1: { id: 'a1', title: 'abc' }}

// keyBy for array only
const keyBy = (array, key) => (array || []).reduce((r, x) => ({ ...r, [key ? x[key] : x]: x }), {});

// Native
console.log(keyBy(['a', 'b', 'c']))
// output: { a: 'a', b: 'b', c: 'c' }
console.log(keyBy([{ id: 'a1', title: 'abc' }, { id: 'b2', title: 'def' }], 'id'))
// output: { a1: { id: 'a1', title: 'abc' }, b2: { id: 'b2', title: 'def' } }
console.log(keyBy(Object.values({ data: { id: 'a1', title: 'abc' }}), 'id'))
// output: { a1: { id: 'a1', title: 'abc' }}

// keyBy for array and object
const collectionKeyBy = (collection, key) => {
  const c = collection || {};
  return c.isArray() ? keyBy(c, key) : keyBy(Object.values(c), key);
}
```

taipeicoder (approved):
Works as intended 👍

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
```diff
@@ -0,0 +1,133 @@
...
+import { keyBy } from '@automattic/js-utils';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { useState, useEffect, useRef, useCallback } from 'react';
+import { ONBOARD_STORE } from '../../../../../stores';
+import type { Pattern, Category } from '../types';
+import type { OnboardSelect } from '@automattic/data-stores';
+import type { Design } from '@automattic/design-picker/src/types';
+import type { GlobalStylesObject } from '@automattic/global-styles';
+
+const useRecipe = ( siteId = 0, patterns: Pattern[], categories: Category[] ) => {
```
okmttdhr, line 11: badge

Can I ask what `recipe` mean in this PR? Is there something called `recipe` about onboarding?

arthur791004: Maybe PCYsg-HDF-p2. It's just a pre-defined page layout.

okmttdhr: Thank you!
I've looked a bit at how `state.selectedDesign (recipe)` is used, and I see that we use it to hold the selected design during onboarding (in my understanding).
And it's stored here: https://github.com/Automattic/wp-calypso/blob/57dfae5ad23b1f78b5eb15992f1de6f532d1be48/packages/data-stores/src/onboard/index.ts#L48

okmttdhr (approved):
✅

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/hooks/use-recipe.ts
arthur791004: Added!

taipeicoder: Thanks for going the extra mile 🙂

Issue #74264, opened 2023-03-09
Plans 2023: biannual toggle - implement AB test


chriskmnds:
Part of addressing pdgrnI-28f-p2 

- Decide if an AB test should run for this
    - See relevant discussion and confirm with @southp pdgrnI-28f-p2#comment-3561
- Once https://github.com/Automattic/wp-calypso/issues/74260, https://github.com/Automattic/wp-calypso/issues/74262, and https://github.com/Automattic/wp-calypso/issues/74261 have shipped

southp:
Assigning myself to it since it's an ongoing negotiation effort :)

southp:
I've updated the issue title from "Implement AB test(?" to "implement AB test" since we have the green light to do so now: pdgrnI-28f-p2#comment-3738 .

* It will be a calypso-side A/B test
* We can likely just use the experiment spec @aaronyan wrote for us: pbxNRc-2qr-p2

southp:
Memo: 

* The PR where hiding the shorter terms comparing to the picked one was first introduced: https://github.com/Automattic/wp-calypso/pull/69643

chriskmnds:
@southp this will probably involve adding the flag to the configs right? 

we've been calling it `plans/biannual-toggle` so far, but it really should be something else - `plans/biennial-toggle` or `plans/2y-toggle`, etc. props to @aneeshd16 who actually checked what "biannual" means :-)

southp:
@chriskmnds Good catch. TIL that `biannual` is a totally different term 🙃
My PR ends up replacing the config flag by the experiment assignment though. If we need the flag again, we should surely update it :)

PR #75063, opened 2023-03-29
Plans 2023: The 1y/2y term presentation test

southp:
Related to #74264 

## Proposed Changes

This PR implements the A/B/n test proposed in p2-pbxNRc-2qr. There are two variations introduced:

* `toggle_only`: The term toggle shows 1y/2y in `/start/plans` and `/plans`
* `toggle_and_checkout`: The term toggle shows 1y/2y in `/start/plans` and `/plans`. In the checkout, the terms lower than the selected one show. e.g. picking an 2yr plan, both the yearly and the monthly options will show in the checkout.

Note that I only preload the experiment for the onboarding flow, but not for cases when users are already logged-in. We want to make sure that the checkout experience is only altered when the users are from either the onboarding flow or `/plans`. While it's easy to restrict and to preload the experiment within the onboarding flow, there are countless other places that lead to the checkout page once outside the flow. Thus I found it hard to find a safe place to preload outside the onboarding flow. Chances are that I've overlooked something, of course 🙃 

## Testing Instructions

In general, use ExPlat to enforce variation. The instance is linked in p2-pbxNRc-2qr.

1.  For both variations, the 1y/2y term toggle should show for `/start/plans` and `/plans`, while the default variation shows the monthly/annually one.  
2. For `toggle_and_checkout` variation, make sure the lower terms comparing to what you have picked are shown. e.g. if you pick a 2yearly plan, the monthly option. For the `toggle_only` and the default variation, it stays the way it works currently.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
@@ -14,6 +14,7 @@ import {
 } from '@automattic/wpcom-checkout';
 import styled from '@emotion/styled';
 import { useState } from 'react';
+import { dangerouslyGetExperimentAssignment } from 'calypso/lib/explat';
 import { isWcMobileApp } from 'calypso/lib/mobile-app';
 import { useGetProductVariants } from 'calypso/my-sites/checkout/composite-checkout/hooks/product-variants';
 import { getSignupCompleteFlowName } from 'calypso/signup/storageUtils';
@@ -201,6 +202,21 @@ function LineItemWrapper( {
 		if ( isJetpack || isAkismet ) {
 			return true;
 		}
+
+		try {
+			const { variationName } = dangerouslyGetExperimentAssignment( 'calypso_plans_2yr_toggle' );
+
+			if ( variationName === 'toggle_and_checkout' ) {
+				return true;
+			}
+		} catch {
+			// The error is intentionally ignore here. For this particular experiment,
+			// the experience should start from the 2023 version of plans page to the checkout.
+			// Thus, for any other flow that leads to the checkout, they shouldn't be affected.
+			// That also means chances are that we can't load or preload the experiment earlier
+			// so that `dangerouslyGetExperimentAssignment` doesn't throw.
+		}
+
 		return variant.termIntervalInMonths >= initialVariantTerm;
 	} );
 
--- a/client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
+++ b/client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
@@ -1,4 +1,3 @@
-import config from '@automattic/calypso-config';
 import {
 	applyTestFiltersToPlansList,
 	FeatureGroup,
@@ -18,9 +17,8 @@ import { useTranslate } from 'i18n-calypso';
 import { useState, useCallback, useEffect, ChangeEvent } from 'react';
 import JetpackLogo from 'calypso/components/jetpack-logo';
 import { FeatureObject, getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
-import PlanTypeSelector, {
-	PlanTypeSelectorProps,
-} from 'calypso/my-sites/plans-features-main/plan-type-selector';
+import { PlanTypeSelectorProps } from 'calypso/my-sites/plans-features-main/plan-type-selector';
+import TermExperimentPlanTypeSelector from 'calypso/my-sites/plans-features-main/term-experiment-plan-type-selector';
 import PlanFeatures2023GridActions from './actions';
 import PlanFeatures2023GridBillingTimeframe from './billing-timeframe';
 import PopularBadge from './components/popular-badge';
@@ -756,20 +754,11 @@ export const PlanComparisonGrid: React.FC< PlanComparisonGridProps > = ( {
 			<PlanComparisonHeader className="wp-brand-font">
 				{ translate( 'Compare our plans and find yours' ) }
 			</PlanComparisonHeader>
-			<PlanTypeSelector
+			<TermExperimentPlanTypeSelector
+				isEligible={ true }
 				kind="interval"
 				plans={ displayedPlansProperties.map( ( { planName } ) => planName ) }
-				isInSignup={ planTypeSelectorProps.isInSignup }
-				isStepperUpgradeFlow={ planTypeSelectorProps.isStepperUpgradeFlow }
-				eligibleForWpcomMonthlyPlans={ planTypeSelectorProps.eligibleForWpcomMonthlyPlans }
-				isPlansInsideStepper={ planTypeSelectorProps.isPlansInsideStepper }
-				intervalType={ planTypeSelectorProps.intervalType }
-				customerType={ planTypeSelectorProps.customerType }
-				hidePersonalPlan={ planTypeSelectorProps.hidePersonalPlan }
-				basePlansPath={ planTypeSelectorProps.basePlansPath }
-				siteSlug={ planTypeSelectorProps.siteSlug }
-				hideDiscountLabel={ false }
-				showBiannualToggle={ config.isEnabled( 'plans/biannual-toggle' ) }
+				planTypeSelectorProps={ planTypeSelectorProps }
 			/>
 			<Grid isInSignup={ isInSignup }>
 				<PlanComparisonGridHeader
--- a/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
@@ -3,10 +3,10 @@ import PlanTypeSelector, { PlanTypeSelectorProps } from './plan-type-selector';
 
 const TermExperimentPlanTypeSelector: React.FunctionComponent< {
 	isEligible: boolean;
-	selectorKind: string;
+	kind: string;
 	plans: string[];
 	planTypeSelectorProps: PlanTypeSelectorProps;
-} > = ( { isEligible, selectorKind, plans, planTypeSelectorProps } ) => (
+} > = ( { isEligible, kind, plans, planTypeSelectorProps } ) => (
 	<ProvideExperimentData name="calypso_plans_2yr_toggle" options={ { isEligible } }>
 		{ ( isLoading, experimentData ) => {
 			if ( isLoading ) {
@@ -21,9 +21,7 @@ const TermExperimentPlanTypeSelector: React.FunctionComponent< {
 					variationName === 'toggle_and_checkout' || variationName === 'toggle_only',
 			};
 
-			return (
-				<PlanTypeSelector { ...propsWithBiannualToggle } kind={ selectorKind } plans={ plans } />
-			);
+			return <PlanTypeSelector { ...propsWithBiannualToggle } kind={ kind } plans={ plans } />;
 		} }
 	</ProvideExperimentData>
 );
--- a/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
@@ -10,7 +10,7 @@ const PlanSelectorLoadingEllipsis = styled( LoadingEllipsis )`
 
 const TermExperimentPlanTypeSelector: React.FunctionComponent< {
 	isEligible: boolean;
-	kind: string;
+	kind: 'interval' | 'customer';
 	plans: string[];
 	planTypeSelectorProps: PlanTypeSelectorProps;
 } > = ( { isEligible, kind, plans, planTypeSelectorProps } ) => (
--- a/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
@@ -1,3 +1,4 @@
+import { LoadingEllipsis } from 'calypso/components/loading-ellipsis';
 import { ProvideExperimentData } from 'calypso/lib/explat';
 import PlanTypeSelector, { PlanTypeSelectorProps } from './plan-type-selector';
 
@@ -10,7 +11,7 @@ const TermExperimentPlanTypeSelector: React.FunctionComponent< {
 	<ProvideExperimentData name="calypso_plans_2yr_toggle" options={ { isEligible } }>
 		{ ( isLoading, experimentData ) => {
 			if ( isLoading ) {
-				return <></>;
+				return <LoadingEllipsis />;
 			}
 
 			const { variationName } = experimentData;
--- a/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
+++ b/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
@@ -1,0 +1,31 @@
+import { ProvideExperimentData } from 'calypso/lib/explat';
+import PlanTypeSelector, { PlanTypeSelectorProps } from './plan-type-selector';
+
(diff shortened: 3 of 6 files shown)


client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx
```diff
@@ -0,0 +1,30 @@
...
+const TermExperimentPlanTypeSelector: React.FunctionComponent< {
+	isEligible: boolean;
+	kind: string;
+	plans: string[];
+	planTypeSelectorProps: PlanTypeSelectorProps;
+} > = ( { isEligible, kind, plans, planTypeSelectorProps } ) => (
+	<ProvideExperimentData name="calypso_plans_2yr_toggle" options={ { isEligible } }>
+		{ ( isLoading, experimentData ) => {
+			if ( isLoading ) {
+				return <LoadingEllipsis />;
```
southp, line 14: This is ugly. Need a better styling.

aaronyan (commented):
LGTM, @southp! I'd like for @razvanpapadopol and @jessie-ross to approve too though since they will be more knowledgeable about the code.

Just to confirm my understanding, it looks like the experiment assignment is called in these three places:
- Assign the variation on the signup page (users will eventually hit `path: /start/plans`)
- Assign the variation on the site management plans page (`path: /plans/:site`)
- Retrieve variation on the checkout page to show plan variants in the dropdown

I was able to test the logged-in experience on the site management plans page and checkout page - all three variations looked good to me! How do you test the user signup flow? Whenever I try to register a new user, I can't get past the first step of creating an account when running the branch from my local machine.

southp:
@aaronyan Thanks for taking a look!

> How do you test the user signup flow? Whenever I try to register a new user, I can't get past the first step of creating an account when running the branch from my local machine.

Could you elaborate more on this, e.g. a screencast, any warning/error in the consoles? It doesn't happen to me, but there might be something going wrong in the code.

In my case, I first signup for an account, assign that account to a specfic variation, refresh, and proceed. I also did use the bookmarklet at `/start` to anonymously assign a variation before signing up. However, that also need a few anecdotal refresh to see the correct UI showing up. My assumption is that the experiment fetch result was cached when I first entered `/start`, but it should work fine in production where a specific variation will be given there.

client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
```diff
@@ -764,20 +762,11 @@ export const PlanComparisonGrid: React.FC< PlanComparisonGridProps > = ( {
...
-				isInSignup={ planTypeSelectorProps.isInSignup }
-				isStepperUpgradeFlow={ planTypeSelectorProps.isStepperUpgradeFlow }
-				eligibleForWpcomMonthlyPlans={ planTypeSelectorProps.eligibleForWpcomMonthlyPlans }
-				isPlansInsideStepper={ planTypeSelectorProps.isPlansInsideStepper }
-				intervalType={ planTypeSelectorProps.intervalType }
-				customerType={ planTypeSelectorProps.customerType }
-				hidePersonalPlan={ planTypeSelectorProps.hidePersonalPlan }
-				basePlansPath={ planTypeSelectorProps.basePlansPath }
-				siteSlug={ planTypeSelectorProps.siteSlug }
-				hideDiscountLabel={ false }
```
aneeshd16, line 771: The `hideDiscountLabel` prop is not passed in the refactored `TermExperimentPlanTypeSelector`, but this should be okay since we don't explicitly check for the `false` value. No action is required here, just adding this for the benefit of other reviewers. 🙂 
https://github.com/Automattic/wp-calypso/blob/abaa8eee9298e2c39623bb0e50c5cee4e306f212/client/my-sites/plans-features-main/plan-type-selector.tsx#L184

aneeshd16 (approved):
LGTM! 

I am curious if there was any specific reason to use the HOC instead of the `useExperiment` hook? But it probably doesn't matter because the HOC uses the hook under the hood.

client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
```diff
@@ -764,20 +762,11 @@ export const PlanComparisonGrid: React.FC< PlanComparisonGridProps > = ( {
 			<PlanComparisonHeader className="wp-brand-font">
 				{ translate( 'Compare our plans and find yours' ) }
 			</PlanComparisonHeader>
-			<PlanTypeSelector
+			<TermExperimentPlanTypeSelector
+				isEligible={ true }
```
jessie-ross, line 758: If possible, it would be prudent to put `is2023PricingGridVisible` in here as well.

client/signup/main.jsx
```diff
@@ -207,6 +208,11 @@ class Signup extends Component {
 				)
 			);
 		}
+
+		// preload the experiment to minimize the perceived loading time
+		if ( this.props.flowName === flows.defaultFlowName ) {
+			loadExperimentAssignment( 'calypso_plans_2yr_toggle' );
```
jessie-ross, line 214: It is necessary to have this conditional on `is2023PricingGridVisible`.

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
```diff
@@ -201,6 +202,21 @@ function LineItemWrapper( {
 		if ( isJetpack || isAkismet ) {
 			return true;
 		}
+
+		try {
+			const { variationName } = dangerouslyGetExperimentAssignment( 'calypso_plans_2yr_toggle' );
```
jessie-ross, line 207: Are we sure that we are pre-loading the experiment with enough time for this point? Do we have an estimate?

jessie-ross (changes requested):
See comments, but I could be incorrect on some things as I don't know this code too well.

client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
southp: Good point. However, I didn't put it in because this component is only used in the 2023 pricing grid, and that's also why that property isn't available here. If I added it, it would be removed in the near future when we either concluding this experiment or exposing this new pricing grid to all the flows, so I chose to not. Does that make sense to you?

client/signup/main.jsx
southp: Since both that prop and its related function [is2023PricingGridActivePage()](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-products/src/plans-utilities.ts#L61) basically check the URL paths matching `/start/plans` or `/plans`, both of them can't be used here because it will be `/start/` at this point. I originally think it's find to preload here since there are exposing events for filtering out unwanted traffic. If it's not safe, maybe I should just not preload it?

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
southp: Good question. [The doc](https://github.com/Automattic/wp-calypso/blob/trunk/packages/explat-client/README.md#L50) says:

> minimum 2 seconds looking at perf data, 5-10 seconds is best

If we include the preload in the previous comment, for the signup case we'd load the experiment from `/start` so I think it should be safe anecdotally. However, for the alread logged-in case, we'd load the experiment only from `/plans`, which means they do have a chance to access `/plans` and click on any upgrade button to get to the checkout page sooner. One potential solution is to block loading the pricing grid so it's not possible, even though that's something I intentionally avoid here. 

Could you advice how we could estimate the time?

client/signup/main.jsx
jessie-ross: > I originally think it's find to preload here since there are exposing events for filtering out unwanted traffic.

This isn't ideal, but it is ok.

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
jessie-ross: > Could you advice how we could estimate the time?

I think the easiest way is to track timing using developer tools as you run through the experience.
You can see when the API call to `/experiments` occurs, and you can see what time you hit this code with a `console.log`.

> One potential solution is to block loading the pricing grid so it's not possible, even though that's something I intentionally avoid here.

Yeah it's a hard decision to make, but I'd recommend it if the timing is too tight, and there is no other way around that.

client/signup/main.jsx
razvanpapadopol: > I originally think it's find to preload here since there are exposing events for filtering out unwanted traffic. If it's not safe, maybe I should just not preload it?

I'd say it's safe to preload like this. Then, using the exposure event with `path: /start/plans` will filter out all users that were assigned and dropped before reaching that step.

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
razvanpapadopol: >  for the signup case we'd load the experiment from /start so I think it should be safe anecdotally

Yes, the assignment is definitely done by the time a `/start` user reaches checkout from loading the signup controller as there are at least 2 steps in between.

> they do have a chance to access /plans and click on any upgrade button to get to the checkout page sooner.

That chance is pretty slim as I see we're actually [displaying a spinner while the assignment is loading](https://github.com/Automattic/wp-calypso/pull/75063/files#diff-55518f6f3437bf6669939fefd8df0377540924ae6a53133830805b30d028f04cR19-R21). It's true this is only for the `<PlanTypeSelector />` but still, it gives the impression that the grid is loading.

@jessie-ross , I wonder if a solution to be 100% sure that all analyzed users were correctly exposed, without adding extra loading time, is to fire a Tracks event when rendering the `<PlanTypeSelector />` so it will say something like `calypso_plans_plan_type_selector_view` and use that as exposure event.

jessie-ross: > I wonder if a solution to be 100% sure that all analyzed users were correctly exposed, without adding extra loading time, is to fire a Tracks event when rendering the <PlanTypeSelector /> so it will say something like calypso_plans_plan_type_selector_view and use that as exposure event.

Yeah, an exposure event we know only occurs once the experiment has loaded should be unbiased and help out - it would be useful to have in the event things aren't working.

aaronyan: @southp, is there a Tracks event that already fires when rendering the `<PlanTypeSelector />`? If not, it sounds like one should be added to be used as an exposure event in this experiment. Let me know what the exposure event ends up being, and I can add it to the experiment model in Abacus!

client/my-sites/plan-features-2023-grid/plan-comparison-grid.tsx
southp: Thanks for noting this, @aneeshd16 . I should have written an inline comment so there was no confusion :)

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
southp: Thanks for the in-depth review and discussion, @razvanpapadopol @jessie-ross @aaronyan . I've added `calypso_plans_plan_type_selector_view` in https://github.com/Automattic/wp-calypso/pull/75063/commits/6e99cf40b4c1eb23340fb8dbb1b78069311ec133. Could you help me check if it suits for the requested change? I'd be grateful :) 
Once we agree on it, I'll then register it on the event registration tool.

southp:
@aneeshd16 Regarding:

> I am curious if there was any specific reason to use the HOC instead of the useExperiment hook? But it probably doesn't matter because the HOC uses the hook under the hood.

No specific reason; `useExperiment` hook will work just fine here. I'm just old school and picked that HOC by instinct :)

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
jessie-ross: So that this discussion isn't a blocker, this is a summary of the recommendations:
- Test the timing out.
- If the timing is too tight, block loading.
- If blocking loading is too complex, try adding an exposure event that fires only if users were exposed (hit the experience after loading), although note that this is complex too.
- If the above is still too complex, see if we can avoid testing the area somehow.

southp: @jessie-ross https://github.com/Automattic/wp-calypso/commit/6e99cf40b4c1eb23340fb8dbb1b78069311ec133 should be following the 3rd option:

> If blocking loading is too complex, try adding an exposure event that fires only if users were exposed (hit the experience after loading), although note that this is complex too.

which is a direct implementation of @razvanpapadopol 's [idea](https://github.com/Automattic/wp-calypso/pull/75063#discussion_r1157855919). Rethinking about it, maybe what's needed here is to add an event that fires when `<TermExperimentPlanTypeSelector/>` correctly loads the experiment and renders the underlying `<PlanTypeSelector/>`, rather than when `<PlanTypeSelector/>` is rendered? Could you help me confirm?

ddc22: @southp not sure if exposure events should depend on experiment events such as the successful loading of an experiment. An exposure event generally is based on a meaningful `app event` in the flow of a given user journey. This distinction is important to understand the result of the experiment and maintaining clarity of what exact events were filtered out.
 In other words the current implementation LGTM.
 
WDYT @jessie-ross @razvanpapadopol ?

razvanpapadopol: Hi, @southp and thank you for implementing the new exposure event! 

> maybe what's needed here is to add an event that fires when TermExperimentPlanTypeSelector correctly loads the experiment

Same as @jdc91, I think the current implementation is correct and it also doesn't need to be cleaned up after the experiment. Other experiments or Tracks funnels can use the event in the future if needed.

As a bonus, we'll have now 1 exposure event instead of 2 in the experiment model in Abacus 😁 

> An exposure event generally is based on a meaningful app event in the flow of a given user journey. This distinction is important to understand the result of the experiment and maintaining clarity of what exact events were filtered out.

Great summary, @jdc91 ! In this case, viewing the PlanTypeSelector is the app event that we want to be considered the start of the forked experience. 

As a side note, we are also using the exposure event to "reset the clock" of the attribution window so every time the user will see the PlanTypeSelector, a new 72 hours timeframe will start for a conversion to be recorded ( PCYsg-FqB-p2#6-get-comfortable-with-metric-attribution-and-refund-windows ).

razvanpapadopol (approved):
The experiment implementation looks good! 🚀

client/my-sites/checkout/composite-checkout/components/wp-order-review-line-items.tsx
```diff
@@ -201,6 +202,21 @@ function LineItemWrapper( {
...
+			if ( variationName === 'toggle_and_checkout' ) {
+				return true;
+			}
+		} catch {
+			// The error is intentionally ignore here. For this particular experiment,
+			// the experience should start from the 2023 version of plans page to the checkout.
+			// Thus, for any other flow that leads to the checkout, they shouldn't be affected.
+			// That also means chances are that we can't load or preload the experiment earlier
+			// so that `dangerouslyGetExperimentAssignment` doesn't throw.
+		}
```
razvanpapadopol, line 218: @southp, there is no error being thrown here and the explat-client function was logging directly the error as reported in p1680892941597569-slack-C02DQP0FP. We're disabling the error logging for the purpose of this experiment in https://github.com/Automattic/wp-calypso/pull/75442.

PR #75074, opened 2023-03-29
Add useSubscriptionManagerSiteDeliveryFrequencyMutation

### LINKED ISSUE #74965: `/subscriptions/sites`: Add a useSiteSubscriptionsDeliveryFrequencyMutation
Add a `useSiteSubscriptionsDeliveryFrequencyMutation` for updating a site subscription's delivery frequency. 

 It calls `POST /read/sites/%s/post_email_subscriptions/update`

## Acceptance criteria
**✅ Changes the task covers**

After calling `mutate`with 2 arguments (blog_id & frequency) it updates the delivery frequency for that user.

TimBroddin:
Closes [#74965](https://github.com/Automattic/wp-calypso/issues/74965)

## Proposed Changes

This PR adds the `useSubscriptionManagerSiteDeliveryFrequencyMutation()` hook. It changes the delivery frequency for a site subscription.

## Testing Instructions

⚠️ This PR needs a subkey cookie set. You can find my subkey in the patch above. Navigate to `http://calypso.localhost:3000` and set it using the Application tab from Chrome Devtools.  ⚠️

To be able to test this hook you need to add it somewhere to your code. I suggest adding it to `client/landing/subscriptions/user-settings/user-settings.tsx` and calling the hook from there:

```js
	const { mutate: mutateDeliveryFrequency } =
		Reader.useSubscriptionManagerSiteDeliveryFrequencyMutation();
	useEffect( () => {
		mutateDeliveryFrequency( { blog_id: '105013038', delivery_frequency: 'daily' } );
	}, [ mutateDeliveryFrequency ] );

```

Replace 105013038 to the `blogId` of a blog your external user is subscribed to. 

In Devtools > Network you should see a call to the endpoint

CleanShot 2023-03-29 at 15 03 43@2x

If you were subscribed to the blog, you should see success in the response:

CleanShot 2023-03-29 at 15 05 05@2x

This PR also contains a simple test, to check if the mutation calls the API with the expected parameters. This test file will get more integration tests, to see if this mutation invalidates the right query, in a future PR

## Pre-merge Checklist


packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
```diff
@@ -0,0 +1,69 @@
...
+	success: boolean;
+	subscribed: boolean;
+	subscription: SubscriptionResponse | null;
+};
+
+const useSubscriptionManagerSiteDeliveryFrequencyMutation = () => {
+	const isLoggedIn = useIsLoggedIn();
+	const queryClient = useQueryClient();
+	return useMutation<
+		SiteSubscriptionDeliveryFrequencyResponse,
```
mashikag: Not an issue at all, but it is not necessary to explicitly define the types here. They will be inferred. It was probably complaining when it was work in progress, and you did not have well typed params. 🙂

packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
```diff
@@ -0,0 +1,69 @@
...
+		Error,
+		SiteSubscriptionDeliveryFrequencyParams
+	>(
+		async ( params: SiteSubscriptionDeliveryFrequencyParams ) => {
+			if ( ! params.blog_id || ! params.delivery_frequency ) {
+				throw new Error(
+					translate( 'Something went wrong while saving the delivery frequency.', {
```
mashikag: The fact that we are translating this here means, that we will use this string to display it to a user, right?

TimBroddin: Would it be better to leave it to the client and return something like 'missing_params' here?

packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
TimBroddin: You're absolutely right. It was nagging, I added them, but it seems like they are not needed anymore 👍 
I've removed them.

packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
mashikag: Oh, I am not sure. I am just asking because if we do not handle this, then it will only be displayed in console. Are we planning on handling this at all? I do not think we considered this yet? Right?

TimBroddin: No, I put it in there in case we might want to handle it in the future.

mashikag: Maybe then, for now we could just leave a comment above the line as a reminder to translate the string if we decide to display it to the user.

TimBroddin: I added one. But won't it be sent for translation automatically? Or am I overassuming things? 😁

mashikag: > But won't it be sent for translation automatically?

Oh, it will. I meant that we put the comment, and we do remove the call to translate.

mashikag:
```typescript
// reminder: translate this string when we add it to the UI
throw new Error( 'Something went wrong while saving the delivery frequency.' );
```

TimBroddin: Ah ofcourse :D

mashikag: You addressed it in one place but forgot to here ([line 32](https://github.com/Automattic/wp-calypso/pull/75074/files#diff-19fa8b334602d00ee32b78e9de33f1963311b4e81f1eb264b3d152c8e7b6e2d3R32)). 😅

TimBroddin: Haha oops. Fixed now.

mashikag (approved):

packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const useSubscriptionManagerSiteDeliveryFrequencyMutation = () => {
+	const isLoggedIn = useIsLoggedIn();
+	const queryClient = useQueryClient();
+	return useMutation(
+		async ( params: SiteSubscriptionDeliveryFrequencyParams ) => {
+			if ( ! params.blog_id || ! params.delivery_frequency ) {
+				throw new Error(
+					// reminder: translate this string when we add it to the UI
+					'Something went wrong while saving the delivery frequency.'
```
obenland: Is there a more helpful error message we could show here and below?

TimBroddin:
I merged with this PR, so I'll hold off merging until that one lands.

https://github.com/Automattic/wp-calypso/pull/75085

packages/data-stores/src/reader/mutations/use-subscription-manager-site-delivery-frequency-mutation.ts
TimBroddin: I don't know. We don't get much back from the backend in this case (`success: false)`, and when one of the paramaters is missing it's an implementation error and not a user error. We also don't handle the display of errors in the frontend (something I'm planning on picking up at our next design sync).

PR #63292, opened 2022-05-04
Checkout: Remove products list guards when adding new products

sirbrillig:
#### Changes proposed in this Pull Request

Checkout has a React hook called `usePrepareProductsForCart()` which takes product slugs from the URL or from localStorage and creates cart items from them that will later be added to the cart via another hook. This system waits for the list of products from the products endpoint to load before operating and then validates each product slug. The original reason for this is that the shopping cart endpoint used to require a product ID for each product so we needed to be able to map from the slug to the ID. Since https://github.com/Automattic/wp-calypso/pull/59365, this is no longer used and we send the slug directly.

Aside from being unnecessary, it seems that the data from the products endpoint (which is stored in Redux) may sometimes be inaccurate. There seems to be an intermittent bug where the products endpoint returns data for a Jetpack site and then that data remains even when operating on a non-Jetpack site. As a result, the validation performed when adding cart items can fail, displaying an error to the user even though the product slug is correct.

This PR removes the dependency on the products data from `usePrepareProductsForCart()`.

The only thing that this removes which might be important is setting the `privacy` property on a cart item (this is different from the `privacy_available` property!). Prior to this PR that property was set based on the `is_privacy_protection_product_purchase_allowed` property of the data from the products endpoint, but this seems unnecessary to me for three reasons:

1. I can't find any use of that property on the backend (unless it gets transferred to the payment object). Also, this PR removes the property from the cart TypeScript type entirely and there are no TypeScript errors, so I'm pretty sure it's not used by the frontend either.
2. If we do need it, why not get it directly on the backend inside the cart endpoint rather than requiring the frontend to fetch it from a different endpoint?
3. That property was not set until https://github.com/Automattic/wp-calypso/pull/60308 which was for gutenboarding and I believe gutenboarding is deprecated. Prior to that, it was not being set, at least since the cart was refactored several years ago.

Fixes https://github.com/Automattic/wp-calypso/issues/63291

#### Testing instructions

To test this, we'll need to try adding products to the cart via URL. Here's some URL paths that should correctly add a product to your cart (replace example.com with a real site):

- `/checkout/example.com/pro`
- `/checkout/example.com/personal`
- `/checkout/example.com/premium`
- `/checkout/example.com/domain_reg:lajdlakdjlkjdalkfdgegrgrsdjasldjs.com`
- `/checkout/example.com/domain-mapping:ajdlakdjlkjdalksdjasldjsadadsads.com`

Verify that the following path does not add a product to your cart and that you see an error message to that effect:

[...]

PR #40725, opened 2020-04-02
composite-checkout: support adding renewal products to cart via url

sirbrillig:
#### Changes proposed in this Pull Request

Renewal URLs contain both product slugs and purchase Ids and must be handled in a particular way; notably, unlike "new product" URLs, they use real product slugs. They were originally added in #15043. This PR adds support for such URLs to composite checkout.

Renewal URLs look something like `checkout/dotblog_domain:example.com,domain_map:example.com/renew/12345,54321/example.com`

#### Testing instructions

- Verify that your cart is empty.
- Use the flag `composite-checkout-testing` to ensure you are in the AB test.
- Click a "renew" link in Calypso for a product and visit checkout.
- Verify that you see composite checkout.
- Verify that the renewed product is in your cart.
- Check out and verify that the product is successfully renewed.
- Repeat the above steps with different types of products (eg: plans, domains, domain mapping, etc.).
- Repeat the above steps but instead of clicking a renew link, manually visit a URL with multiple product slugs separated by commas (eg: `domain_map:example.com,domain_reg:example.com`) and the purchase Ids also separated by commas (eg: `12345,54321`). Verify that all products are added to your cart.

PR #75079, opened 2023-03-29
Checkout: Remove useFetchProductsIfNotLoaded

sirbrillig:
## Proposed Changes

In https://github.com/Automattic/wp-calypso/pull/63292 we modified checkout so that it no longer requires a list of products to add products to the cart. However, there's still a hook, `useFetchProductsIfNotLoaded`, which is run by checkout to fetch those products. That hook is removed by this PR.

## Testing Instructions

The hook was added to help add products from the URL in https://github.com/Automattic/wp-calypso/pull/40725 so we can verify that still works:

- Visit checkout with a product in the URL like `/checkout/personal`.
- Verify that the product is added successfully.


michaeldcain (approved):
This works well and LGTM. Tested with a number of different products that would've required the legacy mapping to product_id.

PR #75091, opened 2023-03-29
Subscription Management: Create NotificationSettings dropdown

### LINKED ISSUE #74973: `/subscriptions/sites`: Display the site subscription settings dropdown
## What needs to be done
When the user is in the Site list, each row has a meatball menu at the end. When the user clicks that meatball icon, the site subscription dropdown should appear, like this:

Image

## Acceptance criteria
- The look and feel of the dropdown should match the design here: inDLaEQV8jJ21O4WXawIsJ-fi-712-16898
- The dropdown should appear by clicking the meatball icon.
- The dropdown should disappear when clicking the meatballs again.
- At this point, a simple placeholder text is ok as the content for the dropdown.

phcp:
Resolves #74973

## Proposed Changes

* Creates `NotificationSettings` dropdown component
  * Uses `EllipsisMenu` calypso component
  * Adds a mock item to the menu body
* Move `site-types` to `types` on the Subscriptions project root
* Apply the `NotificationSettings` component to the Sites tab

## Testing Instructions

1. Run this PR on your local env
2. Go to http://calypso.localhost:3000/subscriptions/sites (you will need a valid subkey cookie on a non-logged session)
3. Verify if the page is rendered as expected
4. Click on ellipsis icon and verify if the dropdown is rendered as expected

image
## Pre-merge Checklist

### DIFF
--- a/client/landing/subscriptions/site-list/site-list.tsx
+++ b/client/landing/subscriptions/site-list/site-list.tsx
@@ -22,6 +22,7 @@ export default function SiteList( { sites }: SiteProps ) {
 				<span className="email-frequency" role="columnheader">
 					{ translate( 'Email frequency' ) }
 				</span>
+				<span className="actions" role="columnheader" />
 			</li>
 			{ sites.map( ( site ) => (
 				<SiteRow key={ site.id } { ...site } />
--- a/client/landing/subscriptions/notification-settings/index.tsx
+++ b/client/landing/subscriptions/notification-settings/index.tsx
@@ -1,0 +1,1 @@
+export { default as NotificationSettings } from './notification-settings';
--- a/client/landing/subscriptions/notification-settings/notification-settings.tsx
+++ b/client/landing/subscriptions/notification-settings/notification-settings.tsx
@@ -1,0 +1,13 @@
+import EllipsisMenu from 'calypso/components/ellipsis-menu';
+import PopoverMenuItem from 'calypso/components/popover-menu/item';
+import './styles.scss';
+
+const NotificationSettings = () => {
+	return (
+		<EllipsisMenu popoverClassName="notification-settings__popover">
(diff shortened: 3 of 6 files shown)


client/landing/subscriptions/site-list/site-list.tsx
```diff
@@ -22,6 +22,7 @@ export default function SiteList( { sites }: SiteProps ) {
 				<span className="email-frequency" role="columnheader">
 					{ translate( 'Email frequency' ) }
 				</span>
+				<span className="actions" role="columnheader" />
```
mashikag, line 25: I am a bit unclear about the class name we should define for here based on [our guidelines](https://github.com/Automattic/wp-calypso/blob/trunk/docs/coding-guidelines/css.md).

I realize that we define our components outside of `client/components`. Main reason being they are specific to our entry point. Therefore, we decided to add the `subscription-manager__` prefix to all our class names. This is not strictly as per guideline, but the guideline does not take into consideration separate apps (entry points) within calypso, and what we do makes sense to me.

For this specific example of `actions` class name it still feels to me like it should be prepended with the component name it belongs to, i.e. `subscription-manager__site-list__actions`. What do you think @phcp ? 

I realize you were going by what is here already. All the other already defined cells, they all have such class names without the prefix applied. Maybe we can merge this in as is, and then have a separate PR to address our class names, to make sure they all are prefixed with their component name. What do you think?

mashikag (approved):
I approve with one small comment, which does not have to be addressed as part of this PR.

client/landing/subscriptions/site-list/site-list.tsx
phcp: Yes, totally agree. I just kept the same format as the other classes on the component. But sure, I can fix it on a follow-up PR.

PR #75098, opened 2023-03-30
Logging: Switch from `<DateRange>` component to `<input type="datetime-local">`

p-jackson:
Closes https://github.com/Automattic/dotcom-forge/issues/2080
Closes https://github.com/Automattic/dotcom-forge/issues/2074

## Proposed Changes

Replaces the `<DatePicker>` component with two standard `<input type="datetime-local">` elements.

CleanShot 2023-03-30 at 20 39 21@2x

#### `datetime-local`

This is the first time I've used this input type but it seems to work quite well, especially on Chromium browsers. It doesn't understand timezones, but from the docs this seems like it was intentionally left out. The element's `value` attribute is a `string` which must be written in a very specific format. The browser is then responsible for rendering the control in a locale appropriate way, so just because `value` is specified as `YYYYMMDD` the user doesn't see it that way.

Chromium has a nice UI for specifying the date _and_ the time. Firefox and Safari only have nice UIs for the date and users have to type the time. But this is no worse than the `<DateRange>`, so on average users will have a better experience with the `<input>`.

All evergreen browsers support `datetime-local`, but it was designed to gracefully fallback. If the browser doesn't support `datetime-local` then it falls back to a `<input type="text">` and the user isn't totally lost because they can type in the timestamp they want.

#### Time zones and time stamps

The strategy I landed on for dealing with timezones was to represent timestamps everywhere as `Moment` objects and not worry about what timezone they happened to be in while in memory. They are transformed into the required timezone at the last minute right before the interface with some external API.

- Query params: timestamps are converted to UTC unix timestamps before writing them to the URL. I'm using UTC in the URL to make the links more shareable.
- `<input>` elements: the internal representation of `startTime` and `endTime` Moment objects are switched to the site's timezone immediately before being formatted for the `<input>` elements
- API request param: the logging endpoints accept time ranges in unix time, so the Moment objects are converted to unix timestamps immediately before being passed to the `useSiteLogsQuery` hook

Something that tripped me up for a while was that I didn't realise is that "unix time" implies UTC. So when you use the `.unix()` method, you don't need to do any timezone transformations first. It's always UTC.

Another Moment.js thing to beware: some of the methods take unix timestamps in seconds and some in milliseconds 😬 

#### Query params

I've moved the query param handling to a separate `site-logs-filter-params.ts` module. We can do all the careful serialisation and deserialisation in there.

I separated the time time range into two separate query params rather than a json encoded object. That way there's fewer `%` symbols escaping all the special characters in the URL.

#### Validation

There's no validation of the input boxes yet. Fingers crossed we don't need too much 🤞. You can set a `min` and `max` time on the `datetime-local` controls which prevent users from selecting invalid times, which means I think I've made it impossible to pick a start time that's greater than the end time in the first place. Could do with more testing though.

Because it's a standard HTML element you can probably style invalid states using the `input:invalid` selector.

## Testing Instructions

* Test at `/site-logs`
* Check refreshing
* Check timezones (things should appear in the site's timezone)
* Try to make invalid ranges
* Layout at smaller screen sizes

## Pre-merge Checklist

### DIFF
--- a/client/data/hosting/use-site-logs-query.ts
+++ b/client/data/hosting/use-site-logs-query.ts
@@ -70,7 +70,7 @@ export function useSiteLogsQuery(
 			);
 		},
 		{
-			enabled: !! siteId,
+			enabled: !! siteId && params.start <= params.end,
 			staleTime: Infinity, // The logs within a specified time range never change.
 			select( { data } ) {
 				return {
--- a/client/my-sites/site-logs/components/site-logs-tab-panel/index.tsx
+++ b/client/my-sites/site-logs/components/site-logs-tab-panel/index.tsx
@@ -1,7 +1,7 @@
 import { TabPanel } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import classnames from 'classnames';
-import page from 'page';
+import { updateLogTypeQueryParam } from '../../site-logs-filter-params';
 import type { SiteLogsTab } from 'calypso/data/hosting/use-site-logs-query';
 import './style.scss';
 
@@ -28,8 +28,8 @@ export const SiteLogsTabPanel = ( {
 			initialTabName={ selectedTab }
 			className={ classnames( 'site-logs-tab-panel', className ) }
 			tabs={ tabs }
-			onSelect={ ( tabName ) => {
-				onTabSelected( tabName );
+			onSelect={ ( tabName: SiteLogsTab ) => {
+				updateLogTypeQueryParam( tabName );
 				onSelected?.( tabName as SiteLogsTab );
 			} }
 		>
@@ -37,9 +37,3 @@ export const SiteLogsTabPanel = ( {
 		</TabPanel>
 	);
 };
-
-function onTabSelected( tabName: string ) {
-	const url = new URL( window.location.href );
-	url.searchParams.set( 'log-type', tabName );
-	page.replace( url.pathname + url.search );
-}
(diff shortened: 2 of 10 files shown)


zaguiini:
Looks cool 👍 Can we trigger the date time picker modal when clicking the input? This is how we can do it: [`showPicker` method on click](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker)

zaguiini:
I'm pretty much able to set the `From` date to be after `To`, though 😅 

image

p-jackson:
> Looks cool 👍 Can we trigger the date time picker modal when clicking the input?

This might be a good idea on Chromium, but might be annoying on other browsers where the popup doesn't have any controls for setting the time. In which case I don't want the modal popping up while I'm just trying to move my cursor around in the editing field and using my keyboard.

vykes-mac:
I think we have to allow typing just to be safe, showPicker is cool but might restrict some users

vykes-mac:
@p-jackson if everything looks workable we can probably merge this.

p-jackson (commented):
Thanks for your work getting validation and error stats working @vykes-mac! I'm going to merge it now.

I had to adjust the way invalid dates work a little bit. By having the `<DateTimePicker>` control it's own date state and return `null` when the dates were invalid, this made the component "uncontrollable". You could no longer tell the input boxes to show a different date by passing it a prop. This is needed when the user checks the auto-refresh toggle. We update the date range state, but the input boxes didn't update because they were ignoring prop changes.

[The new React docs talk about this specific thing](https://react.dev/learn/choosing-the-state-structure#don-t-mirror-props-in-state).

vykes-mac:
Thanks @p-jackson

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7897513

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7897513).

PR #75116, opened 2023-03-30
Limited Global Styles: Get blog ID of Atomic site from blog option

mmtr:
See p9F6qB-byy-p2

## Proposed Changes

Changes the logic that checks whether Global Styles are available to ensure that the expected WP.com blog ID is passed to the `wpcom_site_has_feature` function when it runs on an Atomic site with Jetpack deactivated.

## Testing Instructions

- Apply these changes to a WoA dev site.
- Deactivate Jetpack.
- Try to reactivate Jetpack.
- Make sure you don't get any fatal and Jetpack is correctly activated.

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
@@ -26,9 +26,16 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			$blog_id = get_current_blog_id();
 		} elseif ( defined( 'IS_ATOMIC' ) && IS_ATOMIC ) {
-			$blog_id = method_exists( 'Jetpack_Options', 'get_option' )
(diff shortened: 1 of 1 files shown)


hideandgeek404 (approved):
I have tested this on my own test site and it works perfectly. This is the same implementation as in other Atomic specific areas and checking the return value I now see the correct blog ID where previously I saw blog ID being returned as 1.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -26,9 +26,16 @@ function wpcom_should_limit_global_styles( $blog_id = 0 ) {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			$blog_id = get_current_blog_id();
 		} elseif ( defined( 'IS_ATOMIC' ) && IS_ATOMIC ) {
-			$blog_id = method_exists( 'Jetpack_Options', 'get_option' )
-				? (int) Jetpack_Options::get_option( 'id' )
-				: get_current_blog_id();
+			/*
```
dsas, line 29: Non-blocking: It feels like this should really be a reusable function somewhere

PR #75136, opened 2023-03-30
Apply responsive styles to dedicated notifications page

davemart-in:
## Description

We just deployed the new dedicated notifications page, but failed to add responsive styles. This PR should remedy that.

### Before

before

### After

after

### DIFF
--- a/client/notifications/style.scss
+++ b/client/notifications/style.scss
@@ -216,4 +216,18 @@ div.reader-notifications__panel {
 		}
 	}
 }
-
+@media only screen and (max-width: 783px) {
+	div.reader-notifications__panel #wpnc-panel.wpnc__main {
+		&.wpnt-open {
+			left: 0;
+			min-width: 100%;
+		}
+		.wpnc__list-view .wpnc__selected-note {
+			box-shadow: none;
+		}
+		.wpnc__single-view {
+			animation-name: wpnc__slideIn;
+			left: 0;
+		}
+	}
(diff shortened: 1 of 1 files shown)


noahtallen:
Nice! This is working a lot better. I noticed this issue right when it loads in though:


https://user-images.githubusercontent.com/6265975/228879071-1769c293-592c-46e8-ac28-ce18626f6aaa.mov

client/notifications/style.scss
```diff
@@ -216,4 +216,18 @@ div.reader-notifications__panel {
 		}
 	}
 }
-
+@media only screen and (max-width: 783px) {
```
noahtallen, line 219: do we have existing sass mixins/etc for breakpoints?

davemart-in: We do, but none of them seem to be accessible to this scss file. 783px seems to be hard coded in a number of different spots. Not sure I feel comfortable refactoring stuff in this PR.

davemart-in:
> I noticed this issue right when it loads in though:

Good eye! I believe this should be fixed in: https://github.com/Automattic/wp-calypso/pull/75136/commits/538578434634ceccb868164609f4b005089cfdb3

client/notifications/style.scss
noahtallen: cool 👍

noahtallen (approved):

PR #75137, opened 2023-03-30
One-click Checkout: Wait for validation to finish before rendering

sirbrillig:
## Proposed Changes

https://github.com/Automattic/wp-calypso/pull/74676 added an automated test for the one-click checkout modal. However, the test could be flaky because the upsell button that triggered the modal could be clicked before the contact validation was complete, which meant that the page considered the validation invalid and did not display the modal.

This PR changes the page that displays the upsell button so that it does not render if validation is in-progress.

## Testing Instructions

Automated tests should cover this, but you can try it manually as well by following the test instructions in https://github.com/Automattic/wp-calypso/pull/74555


pottedmeat (approved):
Code looks good.

PR #75138, opened 2023-03-30
Fix possibly undefined content window

noahtallen:
Related to #

## Proposed Changes

Fixes a sentry error. We are listening to window message events, which trigger `onmessage`, but this isn't directly related to the iframe. As a result, we need to do a quick check that the iframe actually loaded before doing anything.

## Testing Instructions
Verify code

### DIFF
--- a/packages/wpcom-proxy-request/src/index.js
+++ b/packages/wpcom-proxy-request/src/index.js
@@ -358,6 +358,10 @@ function onload() {
  */
 
 function onmessage( e ) {
+	// If the iframe was never loaded, this message might be unrelated.
+	if ( ! iframe?.contentWindow ) {
+		return;
+	}
 	debug( 'onmessage' );
 
 	// Filter out messages from different origins


tyxla:
Looks good, thanks!

jsnajdr:
Hi @noahtallen and @tyxla :wave: I'm getting regular Sentry notifications about a related error, at this line in the `wpcom-proxy-request` package:
```js
iframe.contentWindow.postMessage( postStrings ? JSON.stringify( params ) : params, proxyOrigin );
```
where the error is
```
Cannot read properties of null (reading 'postMessage')
```
I have no idea how it could possibly happen. The `submitRequest` function is called only after the `iframe` got loaded, i.e., after the frame has sent us a `ready` message. Before it's loaded, and after it's unloaded, the outgoing requests are buffered in the `buffered` array and `submitRequest` can't be called.

And yet we have a situation where the `iframe` exists, but doesn't have a `contentWindow`. But it has successfully loaded and the script inside it has sent us a message. How can both be true at the same time?

Maybe it's some race condition inside `reloadProxy()`, but I don't see it either. On `uninstall()`, everything gets property uninstalled: the `onmessage` listener removed, iframe removed from document, the `loaded` global variable set to `false`...

I'm also wondering which error exactly this PR intended to fix. The PR description says:

> We are listening to window message events, which trigger `onmessage`, but this isn't directly related to the iframe. 

But the `onmessage` handler was _already_ checking where the message is coming from -- by doing an `event.source === iframe.contentWindow` check. So, I don't see the possible effect the additional check could have. Was it the `iframe` variable itself that was `null`?

jsnajdr:
Update: after reading through the HTML standard, I think I know who might be stealing our `iframe.contentWindow`: the `contentWindow` object is destroyed and nullified when the `iframe` element is removed from the document. Any script can do this, by calling `document.body.removeChild( iframeEl )`. The `wpcom-proxy-request` package will be completely unaware of that. The `iframe` element still exists, it has a local variable pointing to it, but `iframe.contentWindow` will be gone.

There's probably some code that's mistakenly removing iframes owned by someone else.

tyxla:
Interesting, @jsnajdr! 

But if an iframe is no longer in the document, should we continue communicating with it? My gut sense tells me that we shouldn't.

jsnajdr:
> But if an iframe is no longer in the document, should we continue communicating with it?

Yes, in that case, we should detect the `iframe.contentWindow === null` condition and `submitRequest` should throw an error. "The request can't be sent because someone stole our proxy iframe from us." The `wpcom.req.get()` call should return a rejected promise, it should not fail silently.

But it would be also nice to figure out who is stealing the proxy iframe. If that's what happens and causes this error.

noahtallen:
Could a browser extension remove the iframe (potentially thinking it's related to ad tracking or something like that)? Since the error is reported through the Calypso client Sentry project, it's probably not something in wp-admin, which should help narrow the issue

jsnajdr:
> Could a browser extension remove the iframe (potentially thinking it's related to ad tracking or something like that)?

Yes, it could definitely be some extensions, possibly some ad blocker etc.

By the way, I'm fixing the discussed Sentry error with `iframe.contentWindow` in this new PR: #79825.

PR #75140, opened 2023-03-30
State: Convert getDiscountedRawPrice to TypeScript

sirbrillig:
## Proposed Changes

This converts the `getDiscountedRawPrice()` function (and the associated `getPlanDiscountedRawPrice` and `getSitePlanRawPrice` and `getPlanRawDiscount`) to TypeScript and clarifies the meaning of its third parameter that converts the price to a monthly one (renaming it from `isMonthly` to `returnMonthly`). 

The monthly conversion operation is inherently imprecise because it performs math on floating point numbers and will certainly result in rounding errors in some currencies.

## Testing Instructions

Verify that there are no places still using the `isMonthly` option when calling any function (there's a standalone _function_ called `isMonthly`, but that's separate).

You can also verify that the "per month" prices on the `/plans` page for annual plans look the same both before and after this change:

after-rename


client/state/sites/plans/selectors/get-site-plan-raw-price.ts
```diff
@@ -0,0 +1,35 @@
...
+	const plan = getSitePlan( state, siteId, productSlug );
+	if ( ! plan ) {
+		return null;
+	}
+
+	if ( ( plan.rawPrice ?? -1 ) < 0 ) {
+		return null;
+	}
+
+	const price = plan.rawPrice + parseFloat( plan.rawDiscount );
```
sirbrillig, line 32: I had to add a `parseFloat` here to convert this string to a number (unless the type is wrong and it's actually a number?)

client/state/sites/plans/selectors/get-plan-raw-discount.ts
```diff
@@ -1,25 +1,38 @@
...
+	}
+
 	if ( ! isSitePlanDiscounted( state, siteId, productSlug ) ) {
 		return null;
 	}
 
-	return isMonthly
-		? calculateMonthlyPriceForPlan( productSlug, plan.rawDiscount )
+	return returnMonthly
+		? calculateMonthlyPriceForPlan( productSlug, parseFloat( plan.rawDiscount ) )
```
sirbrillig, line 36: Here too, I had to call `parseFloat` to convert the string.

Issue #75050, opened 2023-03-28
Handle errors from Launchpad endpoint


edanzer:
**Initial Report**. We have been seeing some sentry errors related to our Launchpad endpoint. See this slack thread for the original issue: p1680023761269689-slack-CHN6J22MP

**Where to Find Relevant Code**. All the related code for the discussion below is in [launchpad/index.tsx](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx) and [use-launchpad.tsx](https://github.com/Automattic/wp-calypso/blob/trunk/client/data/sites/use-launchpad.ts).

**The Core Issue.** The short summary is that sometimes the fetchLaunchpad() method returns unexpected values (undefined or null or error). When that happens, the following line of code generates a type an error `Cannot read properties of undefined (reading 'launchpad_screen')' because are trying to destructure launchpad_screen from a non-existing value. 

https://github.com/Automattic/wp-calypso/blob/454b8b56c3ba6f15213f3b9e7778da5c649ffe5f/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx#L38

**Temporary Fix**. We prepped a temporary fix here: https://github.com/Automattic/wp-calypso/pull/75049 just to stop the sentry errors. But this fix is not ideal. It simply 'hides' the error by returning some default values when an error occurs. 

**Ideas for Better Fixes**. We (@jeyip and @edanzer) paired for a while to work though possible fixes, but each ran into some secondary concerns. 

The best fix - assuming we're going to continue to use React Query - probably involves using useQuery's built in logic to add error handling. So you can do something like `const { data, isError, isLoading } = useLaunchpad( siteSlug );`. This avoids immediately destructing the output of 'data' (in case it's undefined). You can check if there's an error or check if data is defined. If error or no data, you can redirect to Home or take other actions. If all is good, you can then destructure data do all the usual things. 

We started implementing this but found a separate issue. When you first arrive on Launchpad after creating a new site, Launchpad would initially load based on useLaunchpad's placeholderData. Then the *actual* request would return a "Invalid site" response (and error). If we redirect on error, launchpad would redirect and never load. 

[...]

PR #75049, opened 2023-03-28
Launchpad: Update fetch launchpad options error handling

jeyip:
Related to https://github.com/Automattic/wp-calypso/issues/75050

## Context
* We were spotting an error `Cannot read properties of undefined (reading 'launchpad_screen')`. 
* According to our logging service, it was being generated by [this line](https://github.com/Automattic/wp-calypso/blob/2c1934f2a5b128511ffa0d6d25991c156423d342/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/index.tsx#L38) in `launchpad/index.tsx` ( it invokes the hook for our new endpoint and returns the `launchpad_screen` option )
* We're still unsure of the root cause, and for the time being, we're simply catching errors and returning placeholder data so that runtime errors stop occurring

## Proposed Changes

* Catch errors being returned by fetch launchpad and return placeholder data instead

## Testing Instructions

We know that the error Cannot read properties of undefined (reading 'launchpad_screen') is generated when a falsey siteSlug value is passed to useLaunchpad.

- Comment out references to `fetchingSiteError` ( unsure why this is necessary to emulate the error, but we think it has to do with race conditions )
- Add the `cacheTime: 0` setting to the `useQuery` call in https://github.com/Automattic/wp-calypso/blob/trunk/client/data/sites/use-launchpad.ts#L14
- Force siteSlug to be an `undefined` value in the `useLaunchpad` hook
- Use the free flow to create a new site calypso.localhost:3000/setup/free/intro
- Open chrome dev tools
- Finish all steps and get to the Launchpad
- Ensure that the `Cannot read properties of undefined (reading 'launchpad_screen')` error does not appear in the dev console

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
[...]

PR #75145, opened 2023-03-30
Launchpad: Update useQuery implementation for fetching data

edanzer:
### Proposed Changes

Addresses this issue: https://github.com/Automattic/wp-calypso/issues/75050

This PR tweaks our useLaunchpad hook, which uses react query's useQuery to fetch data from our launchpad endpoint. Specifically, it makes the following changes: 
   - Remove the catch statement from fetchLaunchpad() that was returning data in case of an error. This was added in [this PR](https://github.com/Automattic/wp-calypso/pull/75049) as a quick fix to stop some sentry errors. Those should be resolved more robustly now. 
   - Provide default data via the initialData option rather than placeholderData option. The difference is that if the request fails, the initialData is still be available to our component whereas placeholderData is replaced with undefined in the same situation. Data being undefined is what was triggering the type error in the issue above, because we're trying to immediately destructure that data from the hook. 
   - Added a retry option of 3. This will insure that if our first call to launchpad fails, we'll try to fetch the data twice more before react query returns an error. When we first tried updating this, we found that the first request was sometimes failing the first time you land on the launchpad screen. I'm not 100% sure the reason for the intermittent initial failures, but retrying before sending an error back resolves that. 
   - Set staleTime to zero and refetchOnMount to true. Technically refetchOnMount defaults to true, but it's important enough here that I think it's worth being explicit on that option. These together will just ensure that fresh launchpad data is loaded each time someone comes back to the launchpad screen.
   
Note: one other concern I had with this implementation is that with staleTime: 0, we would be constantly running data fetches on every component re-render. But React Query is smart enough to not do that. There are not excessive data requests with this solution. 
   
### Testing Instructions

**Review Time: Short**
**Testing Time: Short**

1. Checkout this branch and run yarn and yarn start if needed. 
2. Go through a couple of Launchpad flows and confirm that the launchpad screen loads correctly, and that task statuses update correctly as you go through the steps. 
3. Confirm that if you remove the siteSlug from the Launchpad ie, `http://calypso.localhost:3000/setup/free/launchpad?siteSlug=`, you are redirected to /home and no type errors show in the console.
4. Confirm that if you provide a malformed siteSlug, `http://calypso.localhost:3000/setup/free/launchpad?siteSlug=randomstring`,  as part of the Launchpad url, you are redirected to /home and no type errors show in the console.

### DIFF
--- a/client/data/sites/use-launchpad.ts
+++ b/client/data/sites/use-launchpad.ts
@@ -4,34 +4,18 @@ import wpcom from 'calypso/lib/wp';
 export const fetchLaunchpad = ( siteSlug: string | null ) => {
 	const slug = encodeURIComponent( siteSlug as string );
 
-	return (
-		wpcom.req
-			.get( {
-				path: `/sites/${ slug }/launchpad`,
-				apiNamespace: 'wpcom/v2',
-			} )
-			// There are unidentified situations where the fetch launchpad
-			// get request fail. We introduce error handling that returns
-			// placeholder data as a temporary stopgap to prevent noisy sentry
-			// logs from being generated.
-			.catch( () => {
-				return Promise.resolve( {
-					checklist_statuses: [],
-					launchpad_screen: undefined,
-					site_intent: '',
-				} );
-			} )
-	);
+	return wpcom.req.get( {
+		path: `/sites/${ slug }/launchpad`,
+		apiNamespace: 'wpcom/v2',
+	} );
 };
 
-export const useLaunchpad = ( siteSlug: string | null, cache = true ) => {
+export const useLaunchpad = ( siteSlug: string | null ) => {
 	const key = [ 'launchpad', siteSlug ];
-	return useQuery( key, () => siteSlug && fetchLaunchpad( siteSlug ), {
-		meta: {
-			persist: cache,
-		},
-		enabled: true,
-		placeholderData: {
+	return useQuery( key, () => fetchLaunchpad( siteSlug ), {
+		cacheTime: 0,
+		retry: 3,
+		initialData: {
 			checklist_statuses: [],
 			launchpad_screen: undefined,
 			site_intent: '',
--- a/client/data/sites/use-launchpad.ts
+++ b/client/data/sites/use-launchpad.ts
@@ -13,8 +13,8 @@ export const fetchLaunchpad = ( siteSlug: string | null ) => {
 export const useLaunchpad = ( siteSlug: string | null ) => {
 	const key = [ 'launchpad', siteSlug ];
 	return useQuery( key, () => fetchLaunchpad( siteSlug ), {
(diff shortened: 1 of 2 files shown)


jeyip (approved):

jeyip:
## Testing
### Requirements
- [x] Confirm that if you remove the siteSlug from the Launchpad ie, http://calypso.localhost:3000/setup/free/launchpad?siteSlug=, you are redirected to /home and no type errors show in the console.
- [x] Confirm that if you provide a malformed siteSlug, http://calypso.localhost:3000/setup/free/launchpad?siteSlug=randomstring, as part of the Launchpad url, you are redirected to /home and no type errors show in the console.

### Browsers
- [x] Chrome

sharpirate:
### Testing
✔️ Confirm that the launchpad screen loads correctly, and that task statuses update correctly as you go through the steps.
✔️ Confirm that if you remove the siteSlug from the Launchpad ie, `http://calypso.localhost:3000/setup/free/launchpad?siteSlug=`, you are redirected to /home and no type errors show in the console.
✔️ Confirm that if you provide a malformed siteSlug, `http://calypso.localhost:3000/setup/free/launchpad?siteSlug=randomstring`,  as part of the Launchpad url, you are redirected to /home and no type errors show in the console.

sharpirate (approved):

client/data/sites/use-launchpad.ts
```diff
@@ -4,34 +4,19 @@ import wpcom from 'calypso/lib/wp';
...
 	const key = [ 'launchpad', siteSlug ];
-	return useQuery( key, () => siteSlug && fetchLaunchpad( siteSlug ), {
-		meta: {
-			persist: cache,
-		},
-		enabled: true,
-		placeholderData: {
+	return useQuery( key, () => fetchLaunchpad( siteSlug ), {
+		refetchOnMount: true,
+		staleTime: 0,
```
agrullon95, line 17: We can remove `staleTime: 0` since the React Query default value for staleTime is 0

client/data/sites/use-launchpad.ts
```diff
@@ -4,34 +4,19 @@ import wpcom from 'calypso/lib/wp';
...
+export const useLaunchpad = ( siteSlug: string | null ) => {
 	const key = [ 'launchpad', siteSlug ];
-	return useQuery( key, () => siteSlug && fetchLaunchpad( siteSlug ), {
-		meta: {
-			persist: cache,
-		},
-		enabled: true,
-		placeholderData: {
+	return useQuery( key, () => fetchLaunchpad( siteSlug ), {
+		refetchOnMount: true,
```
agrullon95, line 16: The default value for refetchOnMount is true. This can be removed.

agrullon95 (approved):

Issue #74712, opened 2023-03-21
LaunchPad Celebration modal - redesign to harness more prominent upsell opportunity.


Addison-Stavlo:
We want to update the launchpad celebration launch modal to make upsells more prominent. Currently our "Choose a domain" link at the bottom is not very prominent and we would like to increase this since we may be removing upsell opportunities from earlier in the flow.

While the design is not 100% final on the copies, the layout and base idea seems to be solidified to start working on. And we can always update the copies later.

The modal can support 3 different copies based on the state of the site and plan:

- [x] user has no paid plan / custom domain
- [x] user has a paid plan but has not claimed their custom domain
- [x] user has a paid plan and custom domain set -> we upsell something else like Blaze.

Base version (user has no plan):
![Screenshot 2023-03-21 at 8 32 05 AM](https://user-images.githubusercontent.com/28742426/226606827-b7962d4f-1e23-460e-b761-d47e04c4fa99.png)

If users already purchased a plan but didn’t get the domain yet, we could adapt the upsell part:
![Screenshot 2023-03-21 at 8 33 04 AM](https://user-images.githubusercontent.com/28742426/226607058-4c1757b5-67b4-479e-a7e6-0bf87e5e0a34.png)

And if the user already has a paid domain + plan, the upsell part could display a different (but still relevant) feature:
![Screenshot 2023-03-21 at 8 34 14 AM](https://user-images.githubusercontent.com/28742426/226607273-6b095897-9273-466f-a6c1-99e70f7b35de.png)

some slack context - p1678883754247339-slack-CHN6J22MP

jeyip:
I’m out from tomorrow until next week, so if someone could shepherd it, that’d be awesome. If folks are too busy though, no worries, I’ll pick it up next week.

### Next Steps:
- Decide what we want to upsell when the user already has a paid plan and a custom domain. We currently just have placeholder text when that condition is triggered.
- Address any feedback from @nuriapenya  or the design team
- Code review and Test the PR
- Bonus points — figure out if we can find cached data about the blog’s custom domains without making a new network request https://github.com/Automattic/wp-calypso/pull/74744/files#diff-b2a3bdf3b4cec65b339e0803b39186ea50d15861cfa347b97294688a3b89e58eR14-R19

PR #75148, opened 2023-03-30
Launchpad: Wait for data to fetch before rendering celebration modal

jeyip:
Related to https://github.com/Automattic/wp-calypso/issues/74712

## Proposed Changes

* Prevent Launchpad celebration modal from rendering until requisite data has been fetched

## Testing Instructions

* Check out this branch
* `yarn start`
* Create a launchpad site through any flow that requires the site to be launched ( [write](wordpress.com/start), [build](wordpress.com/start), `link-in-bio`, etc. )
* Launch the site from the Launchpad
* Confirm that the domain upsell is rendered
* Verify that the upsell text is correct and that the upsell button behaves as expected

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/customer-home/main.jsx
+++ b/client/my-sites/customer-home/main.jsx
@@ -57,7 +57,11 @@ const Home = ( {
 
 	const { data: layout, isLoading } = useHomeLayoutQuery( siteId );
 
-	const { data: allDomains = [], isSuccess } = useGetDomainsQuery( site?.ID ?? null, {
+	const {
+		data: allDomains = [],
+		isSuccess,
+		isFetching,
+	} = useGetDomainsQuery( site?.ID ?? null, {
 		retry: false,
 	} );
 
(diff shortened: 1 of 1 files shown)


client/my-sites/customer-home/main.jsx
```diff
@@ -83,10 +87,10 @@ const Home = ( {
 	}, [ detectedCountryCode, sitePlan, isNew7DUser ] );
 
 	useEffect( () => {
-		if ( getQueryArgs().celebrateLaunch === 'true' && isSuccess ) {
+		if ( getQueryArgs().celebrateLaunch === 'true' && ! isFetching && isSuccess ) {
```
Addison-Stavlo, line 90: This makes it seem like `isSuccess` is getting prematurely marked as `true` ?  Did you find that this was being set before `isFetching` had completed? Or is it a situation with a subsequent request?

jeyip: > Did you find that this was being set before isFetching had completed? Or is it a situation with a subsequent request?

My hypothesis is that, if a value has been cached, `isSuccess` will always return true. I'm wondering if, in production, somehow it thinks that the request is cached?

We use a cache key of site id, which should be unique for each request, but there might be a period of time, during data hydration, where site id is `null`. Because we rely on `isSuccess` for the celebration modal, I then wonder if that's why we're loading the celebration modal prematurely.

jeyip:
Merging into production to test theory about whether or not this resolves the production only bug

PR #75174, opened 2023-03-31
Newsletter Flow: Plan step to use new grid

### LINKED ISSUE #75060: Newsletter flow: Update plans copy
### What

We need to update the copy of the Plans step to be more relevant to the Newsletter flow.
[Figma](HSljkKnYECEEegL3DqP2xC-fi-7779-72182)
[Google Docs](https://docs.google.com/document/d/1s73a6xbK2CSAFgyI1lpm9pWQlSBfkQ9uTcHuCAFX6yA/edit#heading=h.vc7dgcwchwm8)

escapemanuele:
[Context](p9Jlb4-6L9-p2)

This PR updates the Newsletter flow to use a new design for the Plans step:

image

## Testing

1. Checkout the branch and run `yarn start` or use the Live Link
2. Visit `setup/newsletter` and reach the plans step
3. Verify that the step works correctly. Remember to check the plans comparison at the end of the step.

Fixes https://github.com/Automattic/wp-calypso/issues/75060


escapemanuele:
Also tagging @chriskmnds @Automattic/martech-decepticons , as I saw them working on the 2023 grid 👋 

We're using the new grid for the Newsletter flow Plan step, and so we need to have some customisation for the copy. Please tell me if there is a way to move around the code better :)

packages/calypso-products/src/plans-list.tsx
```diff
@@ -424,6 +427,15 @@ const getPlanFreeDetails = (): IncompleteWPcomPlan => ( {
 	getStoreAudience: () => i18n.translate( 'Best for students' ),
 	getPlanTagline: () =>
 		i18n.translate( 'Get a taste of the world’s most popular CMS & blogging software.' ),
+	getCustomTagline: ( flowName ) => {
```
chriskmnds: Do we need a new method for this? Can we use the existing `getPlanTagLine` and pass optional `flowName`?

escapemanuele: That was my first thought, but then didn't want to change too much working things. I'll expand `getPlanTagline` 👍

client/lib/plans/features-list.tsx
```diff
@@ -1612,7 +1616,7 @@ export const FEATURES_LIST: FeatureList = {
 	},
 	[ FEATURE_UNLIMITED_SUBSCRIBERS ]: {
 		getSlug: () => FEATURE_UNLIMITED_SUBSCRIBERS,
-		getTitle: () => i18n.translate( 'Unlimited subscribers' ),
+		getTitle: () => i18n.translate( 'Import unlimited subscribers' ),
```
chriskmnds, line 1651: We may need to add the "string freeze" label to the PR. Otherwise, I've seen us do "fallback" translations for similar cases (where updating a previously translated string) e.g. [here](https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/editing-toolkit-plugin/editor-site-launch/src/launch-steps/domain-step/index.tsx#L40-L51) using `hasTranslation` and locale slug. Do you think this is relevant and/or possible to do here? 🤔

chriskmnds (commented):
Thanks, @escapemanuele for working on this. I tested the flow and it works. 🎉 

I noticed some minor design inconsistencies that we may need to address.

1. In both features and comparison tables, the borders around the highlighted plan are missing a radius:

**Top table:**

Screenshot 2023-04-06 at 12 46 09 PM


**Comparison:**

Screenshot 2023-04-06 at 12 47 03 PM


Highlighting is handled via an "adjacency matrix/table" that is used to determine which highlighted plans are sitting next to each other - and hence determine the radius at the top. In this case, I think the Business plan is still marked as "highlighted" and the Premium plan as "left-of-highlight" (so it doesn't get a `border-radius` at the top right). I think you will need to modify `useHighlightIndices` to not include the Business plan when in the respective flow. Let's see if that fixes it and go from there.

2. In tablet view the Business plan at the bottom extends across the full width of the grid. I think this may be a larger issue, but not sure if we'd want to ship it this way. Can we get a designer to take a look and tell us how to approach this case? @lucasmendes-design perhaps?


Screenshot 2023-04-06 at 12 49 00 PM


p.s. we should probably _always_ add testing instructions to PRs. they are missing entirely here 🦸

chriskmnds (changes requested):
For the previous review :-)

packages/calypso-products/src/plans-list.tsx
chriskmnds: Thanks for addressing :-)

client/lib/plans/features-list.tsx
escapemanuele: Added the `string freeze` label, we can wait for translations 👍

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7913065

Thank you @escapemanuele for including a screenshot in the description! This is really helpful for our translators.

escapemanuele:
Thank you @chriskmnds, I think I was able to implement your suggestion regarding the highlight indices!

For the tablet view, I propose a change to the [renderTabletView](https://github.com/Automattic/wp-calypso/blob/update/newsletter-plans-copy/client/my-sites/plan-features-2023-grid/index.tsx#L399), as 4 appears to me to be a good number of plans to show on one line on the Newsletter flow.

image

I'd like to check with @lucasmendes-design and @crisbusquets if this tablet view is ok for them 😄

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -394,19 +397,33 @@ export class PlanFeatures2023Grid extends Component<
...
-					TYPE_ECOMMERCE,
-					TYPE_ENTERPRISE_GRID_WPCOM,
-			  ];
-		const topRowPlans = plansToShow.slice( 0, 3 );
-		const bottomRowPlans = plansToShow.slice( 3, 6 );
+		const { planProperties, hideFreePlan, flowName } = this.props;
+
+		let plansToShow = [];
+		let indexOfLastPlanToShowOnTop = 3;
+		const indexOfLastPlanToShowOnBottom = 6;
```
chriskmnds: I find these a little hard to follow, I think they are also misleading e.g. 

- the semantics of `indexOfLastPlanToShowOnTop` is not correct as used for the `slice` method. `3` in the case of `[ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM, TYPE_BUSINESS, TYPE_ECOMMERCE, TYPE_ENTERPRISE_GRID_WPCOM ]` would be `TYPE_BUSINESS` (indices starts at `0`) and that's not shown on the top, but the bottom.
- `indexOfLastPlanToShowOnBottom` - there is no plan with index `6` in any of the plan combinations.

Let's maybe rework these with a variable of the kind "number of items to show on the top", and figure out the second value (`indexOfLastPlanToShowOnBottom`) dynamically from the array length maybe

client/lib/plans/features-list.tsx
```diff
@@ -1667,6 +1671,22 @@ export const FEATURES_LIST: FeatureList = {
...
+		getTitle: () => i18n.translate( '8% transaction fee for payments + Stripe fees' ),
+	},
+	[ FEATURE_PAYMENT_TRANSACTION_FEES_4 ]: {
+		getSlug: () => FEATURE_PAYMENT_TRANSACTION_FEES_4,
+		getTitle: () => i18n.translate( '4% transaction fee for payments + Stripe fees' ),
+	},
+	[ FEATURE_PAYMENT_TRANSACTION_FEES_2 ]: {
+		getSlug: () => FEATURE_PAYMENT_TRANSACTION_FEES_2,
+		getTitle: () => i18n.translate( '2% transaction fee for payments + Stripe fees' ),
+	},
```
chriskmnds, line 1721: Should these features (or some variation of them) also appear in the comparison table at the bottom?

chriskmnds (commented):
Thanks for addressing the feedback @escapemanuele. The borders are fixed now. I left some follow-up comments.

> For the tablet view, I propose a change to the [renderTabletView](https://github.com/Automattic/wp-calypso/blob/update/newsletter-plans-copy/client/my-sites/plan-features-2023-grid/index.tsx#L399), as 4 appears to me to be a good number of plans to show on one line on the Newsletter flow.

Just to note that 4 plans force the columns to be a little too narrow e.g. with large currencies (try switching to IDR or INR), the prices will overflow between 780-1025px: 

Screenshot 2023-04-10 at 12 09 23 PM

However, the currency is at least partly being addressed in https://github.com/Automattic/wp-calypso/pull/75210, which reduces the size of large currencies (> 6 chars). Not sure about the rest though (alignment of the highlight badge seems a little off in the img above, but seems like a minor issue). This is what it would look like with the changes from https://github.com/Automattic/wp-calypso/pull/75210:

Screenshot 2023-04-10 at 12 55 32 PM

@lucasmendes-design

client/lib/plans/features-list.tsx
southp: cc @lucasmendes-design What features should the plans comparison grid listing for the newsletter flow?

southp (changes requested):
Thanks for pioneering into this, @escapemanuele :)

Overall my impression is that the PR introduces too many Newsletter flow specific things in the component itself by `flowName === NEWSLETTER_FLOW`. That means:

1. All the other flows and the plans admin page would pick up code that they never need.
2. The component itself would have dependency to the context it's being used, which I don't think is a good code structure. (Yep, we already have several places doing that, but I'd rather us to not add more.) 
3. When we do the same thing to the other tailored flow, e.g. Link in Bio, the same level of bloat would be expected, making maintenance harder.

Let's look into ways of making what needs to be done here configurable so we don't need to hardcode all these within the component. Our team will be happy to assist :)

client/lib/plans/features-list.tsx
escapemanuele: I'd also like to tag @crisbusquets here, as she is working on improving the Newsletter flow. Changes to the Plans step on the Newsletter flow come from [this Figma file](HSljkKnYECEEegL3DqP2xC-fi-7779_72182)

client/my-sites/plan-features-2023-grid/index.tsx
escapemanuele: Totally right! Changed it to reflect your suggestion 👍

escapemanuele:
Thank you @southp, and also great work you/the team on the new design!

I have the same feeling, and I do not like to have too many flow specific things. I'd love to improve this, as I am really new to plans and their structure I didn't want to change the code too much.

Do we have some idea on how to improve the configurability?

lucasmendes-design:
**Business plans in Newsletters**
I was on AFK so I may miss the latest on this. Pablo asked to remove the Business plan for [Newsletters flow (P2)](https://storytellersp2.wordpress.com/2022/11/18/new-landing-page-creation-newsletter/#comment-893). I also [updated here (GitHub)](https://github.com/Automattic/martech/issues/1533#issuecomment-1494934081). Did we add the business plan back again? 

**Responsiveness for 4 plans** 
I've created a GH for this case here ([GitHub](https://github.com/Automattic/martech/issues/1584#issuecomment-1468598354)). With 4 plans, we need to change the responsiveness otherwise it will be broken on small screens. We'll break down the plans grid into two rolls 2x2.   

[Figma file](https://www.figma.com/file/xkhhUlDowF13dRoXJqlRDM?node-id=1:95#392843704)

**Plans Comparison grid**
Currently, I don't think it makes sense to have the comparison plans grid for the tailored flows since we don't have enough contextual features to present and compare. We can iterate on that for the next version. 

The only flow I think is beneficial to have that it is the blog flow since this audience could be from high to low-tech savviness. 

**Font size for currencies**
Recently, I iterated with @aneeshd16. We can reduce the font size when the price has a huge width dimension. ([GitHub](https://github.com/Automattic/martech/issues/1653#issuecomment-1491839508)). If it doesn't fit, I would reduce the font size even more. This edge case is very specific and will not affect conversions pretty much so this solution may solve this issue (smaller font sizes). 

**Alignment of the highlight badge** 
I agreed with @chriskmnds. It seems that we didn't left-aligned. By default, it's left-aligned. 

Let me know if I missed something here :) 

cc: @crisbusquets @escapemanuele @chriskmnds @southp

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
```diff
@@ -5,16 +5,21 @@ import {
 } from '@automattic/calypso-products';
 import { useTranslate } from 'i18n-calypso';
 import { useSelector } from 'react-redux';
+import { NEWSLETTER_FLOW } from 'calypso/../packages/onboarding/src';
```
chriskmnds: we may need to update the import here to `@automattic/onboarding`

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
```diff
@@ -5,16 +5,21 @@ import {
...
 
-const useHighlightLabel = ( planName: string ) => {
+const useHighlightLabel = ( planName: string, flowName?: string ) => {
 	const translate = useTranslate();
 	const selectedSiteId = useSelector( getSelectedSiteId );
 	const currentPlan = useSelector( ( state ) => getCurrentPlan( state, selectedSiteId ) );
 	const isCurrentPlan = currentPlan?.productSlug === planName;
 
-	if ( isCurrentPlan ) {
+	if ( flowName === NEWSLETTER_FLOW ) {
```
chriskmnds: I know the processing here may change (per James' review), but for such conditions, I think we have respective selectors right? e.g. `isNewsletterFlow`

southp:
> Thank you @southp, and also great work you/the team on the new design!
> 
> I have the same feeling, and I do not like to have too many flow specific things. I'd love to improve this, as I am really new to plans and their structure I didn't want to change the code too much.
> 
> Do we have some idea on how to improve the configurability?

Thanks for being open to feedback and for pioneering into this, @escapemanuele :) 
As you already have gained the 1st hand experience, this component is not yet designed with the configurability that we need to reuse it everywhere in calypso. And this pioneering PR has provided us with valuable clues on how needs to be done.

In general, the places where we've put `flowName === NEWSLETTERS_FLOW` or `isNewsletterFlow( flowName)` are where should be configurable via props. The end result would be that the Newsletter flow would pass in the required customization props somewhere, e.g. from the step wrapper component,  and then the plans component should do what we want to achieve here. 

Here is a rough idea of what should be configurable via props, and a few related things that would need to accompany:

1. A prop that hides the feature comparison grid according to @lucasmendes-design 's [design feedback](https://github.com/Automattic/wp-calypso/pull/75174#issuecomment-1503400548). ( `hidePlansFeatureComparison` already does that, but worth mentioning here :) )
2. A list of the plans to be shown.
3. The responsiveness for showing fewer plans. It doesn't have to be super general; just for the needs of this particular case will be enough.
4. The highlighted plan and the highlight label.
5. The feature list to be shown per each plan.
6. The plan tag lines.

However, I understand that doing all these within one PR is not feasible. It would be already a great foundation in-place if we can get to the first 3 items and then iterate from there. Does that make sense to you?

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7913065).

escapemanuele:
Hey @southp, I just sent some changes to pass the plans and to hide the features comparison directly from the Newsletter flow 😄

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
```diff
@@ -12,15 +13,21 @@ interface HighlightAdjacencyMatrix {
...
 	const currentPlan = useSelector( ( state ) => getCurrentPlan( state, selectedSiteId ) );
 
 	return visiblePlans.reduce< number[] >( ( acc, { planName }, index ) => {
-		const isHighlight =
-			isBusinessPlan( planName ) ||
-			isPremiumPlan( planName ) ||
-			currentPlan?.productSlug === planName;
+		let isHighlight = false;
+
+		if ( flowName === NEWSLETTER_FLOW ) {
```
chriskmnds: another case for `isNewsletterFlow`

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -126,6 +140,8 @@ const PlansWrapper: React.FC< Props > = ( props ) => {
 					isInVerticalScrollingPlansExperiment={ isInVerticalScrollingPlansExperiment }
 					shouldShowPlansFeatureComparison={ isDesktop } // Show feature comparison layout in signup flow and desktop resolutions
 					isReskinned={ isReskinned }
+					hidePlansFeatureComparison={ hidePlansFeatureComparison }
```
southp, line 160: This works but I feel this is a bit of overkill. From the current use case, there wouldn't be a case we need a dynamic state like this. Plus, it adds up the dependency between the domains step and the plans step, since the action that set this flag is dispatched in the domains step. 

Could we simply pass a `true` here?

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -126,6 +140,8 @@ const PlansWrapper: React.FC< Props > = ( props ) => {
...
+					plansToShow={ plansToShow }
```
southp: This doesn't work since it doesn't take the terms, i.e. monthly or annually, into account, which is what the `getPlansFromProps()` function should be doing in `PlansFesaturesMain`, btw. However, it is a bit broken right now.

I'm fixing it now in https://github.com/Automattic/wp-calypso/pull/75949 . Once that's done, you'll be able to just write `planTypes={ [ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM] }` here. I'll circle back when it's done.

client/my-sites/plans-features-main/index.jsx
```diff
@@ -362,14 +362,17 @@ export class PlansFeaturesMain extends Component {
 			showTreatmentPlansReorderTest,
 			flowName,
 			is2023PricingGridVisible,
+			plansToShow,
```
chriskmnds: `getPlansFromProps` above tries to get the plans from `planTypes` first. It's an old piece of code, and it looks like it achieves the same thing. Can we use that prop instead (or remove/rename that one if nothing links to it)? maybe better to remove, cause `plansToShow` has a better semantic than `planTypes`, I think.

southp (changes requested):
Thanks for further polishing the PR, @escapemanuele :) 
The grid is currently broken upon toggling to the monthly one:
image

As my inline comment says, I'm currently working on a fix for it from the component level in https://github.com/Automattic/wp-calypso/pull/75949. I'll be back for further review and collaboration once that's done 🙇🏼‍♂️

packages/data-stores/src/onboard/actions.ts
```diff
@@ -455,6 +455,11 @@ export const setHideFreePlan = ( hideFreePlan: boolean ) => ( {
 	hideFreePlan,
 } );
 
+export const setHidePlansFeatureComparison = ( hidePlansFeatureComparison: boolean ) => ( {
+	type: 'SET_HIDE_PLANS_FEATURE_COMPARISON' as const,
+	hidePlansFeatureComparison,
+} );
```
chriskmnds, line 457: Is the `onboard` store something like a general data store for most signup needs? It's also missing a README. Is @Automattic/vertex mostly owning this one?

There's also a `plans-ui` store, which may be another candidate for this property. It could ultimately be a more general concern whether to show or hide the comparison grid 🤔

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
escapemanuele: Yes, I was following @southp `if we can get to the first 3 items and then iterate from there` 😄

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: I agree, but FWIW the less "prop drilling" we introduce the better, in general. Querying a store or calling a hook is far better than passing things as props. You can think of hooks (which may query stores) as globals that are acceptable.

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
escapemanuele: Thank you @southp, I need some architectural help on the Plans side of things ✌️

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: ^ I was just merely looking for the opportunity to say this, cause we've been steering things on occasion towards prop usage (pass a prop for this or that). and similarly, I'm equally looking for the opportunity to "defeat" some recommendations for inheritance (creating subclasses) I've read at a few places 🤣

packages/data-stores/src/onboard/actions.ts
escapemanuele: Yes, we created it to manage onboarding in the new flows. Added [this issue](https://github.com/Automattic/wp-calypso/issues/75954) to add a README.

I feel like onboard is a better candidate to this, as it manages all the things related to the experience of going through the flow!

client/my-sites/plans-features-main/index.jsx
escapemanuele: Ok, good idea, I can iterate on that ✌️

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
escapemanuele: After adding the dispatch in the domains step, I changed it to have it as a [side effect of the flow](https://github.com/Automattic/wp-calypso/blob/update/newsletter-plans-copy/client/landing/stepper/declarative-flow/newsletter.ts#L53-L58). So it is no more dependant on the domains step :)

escapemanuele: ` I was just merely looking for the opportunity to say this` hehe 😆

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
chriskmnds: I think these hooks here would need some refactoring to be pulled out of the components and into some more central location (and what that "location" would be is also subject to discussion I think). Ideally, I would think the plan properties for the UI would be coming from a selector or a hook which can be customized to return things as they pertain to a certain flow. 

I would also think that having the "flowName" property used throughout these components is not a problem (if anything, `n-1` of cases where the grid is rendered is within some flow). So every component that needs access to the plan properties could query `usePlanProperties( flowName )`. From there, we can iterate on hoisting `flowName` into context  (or whatever location) so as to be queryable by the hook that fetches the plan properties  (definitely something for a follow-up to having things fetched from `usePlanProperties( flowName )` first).

So I don't know if correct right now to devilize `flowName` too much. Rough thinking 🤷

chriskmnds: > if correct right now to devilize `flowName` too much

but please follow along this path. it's just my rough thinking, and I don't mean to detract from the direction. we have to think backward with these components (extract a design from a ton of underlying complexity)

packages/data-stores/src/onboard/actions.ts
chriskmnds: Cool. No objection.

I think it might be a little weird if `PlanFeaturesMain` ends up querying the onboard store to figure out whether to render a comparison grid or not - if, say, `hidePlansFeatureComparison` is no longer a prop but a hook (something like `useGridVisibilities()`) that queries a store to determine which grids to show. but it's not the case now, and not sure if it will be

client/my-sites/plans-features-main/index.jsx
chriskmnds: > `plansToShow` has a better semantic than `planTypes`

I'm not too sure about this though! Naming is important and should be self-conveying @southp maybe for a vote :-)

southp: Yeah, fixing `planTypes` so we don't have to introduce a new prop here is what I'm targeting at https://github.com/Automattic/wp-calypso/pull/75949. It is called `planTypes` now since it was orignated as an early attempt to decouple "plan names" and "billing terms". Thus, what it accepts are `TYPE_FREE`, `TYPE_PERSONAL`, etc. Thus `planTypes` does make some sense but it's so not readable, while `plansToShow` is very readable yet lack of some precisions.

I'll see if I can combine the both into something while working that PR out.

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
southp: Cool. I have no objection keeping it this way then especially now that it's set at the flow level :)

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
southp: > I was following @southp if we can get to the first 3 items and then iterate from there 😄

Yeah, I'm worrying to put in too much architectural change into this PR so proposed that we should iterate it later on. Before we decide on a way to either fully decoupling from the flow name or keeping it, using `isNewsletterFlow()` will be a better choice in the name of [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself). 

Let's update it here :)

client/my-sites/plans-features-main/index.jsx
escapemanuele: Great to hear that @southp, what will you combine the two props to? Something like `planTypesToShow` ? Possibilities are endless here

southp: I've opened another PR for fixing and some refactoring in https://github.com/Automattic/wp-calypso/pull/76070 instead. Note that I ended up not updating the prop name to limit the scope, since there are other parts in calypso depending on that prop name. We can iterate on the prop name as a follow-up :)

chriskmnds: Thanks for explaining @southp . Happy with the direction. I guess we prioritize a review on #76070 as a next step and rebase here

packages/calypso-products/src/constants/features.ts
```diff
@@ -280,6 +280,10 @@ export const FEATURE_CUSTOMIZE_THEMES_BUTTONS_COLORS = 'customize-themes-buttons
 export const FEATURE_TRACK_VIEWS_CLICKS = 'track-views-clicks';
 export const FEATURE_COLLECT_PAYMENTS_LINK_IN_BIO = 'collect-payments-link-in-bio';
 export const FEATURE_LINK_IN_BIO_THEMES_CUSTOMIZATION = 'link-in-bio-themes-customization';
+export const FEATURE_NEWSLETTER_IMPORT_SUBSCRIBERS_FREE = 'newsletter-import-subscribers-free';
+export const FEATURE_PAYMENT_TRANSACTION_FEES_8 = 'payment-transaction-fees-8';
+export const FEATURE_PAYMENT_TRANSACTION_FEES_4 = 'payment-transaction-fees-4';
+export const FEATURE_PAYMENT_TRANSACTION_FEES_2 = 'payment-transaction-fees-2';
```
chriskmnds, line 286: I wonder if we should update the plans details (features) endpoint with the new features added here. I think we may not have been doing this as consistently as we should, but introducing these slugs (along with titles and descriptions) without updating the endpoint will make the two that much more out of sync. Or have these been added already and I'm just missing something?

The plans data store (that we plan to introduce and use more) produces a unified image of plans and features (it consumes and normalizes things from both `/plans` and `/plans/details` endpoints - [here](https://github.com/Automattic/wp-calypso/blob/trunk/packages/data-stores/src/plans/resolvers.ts#L163-L177)). Ultimately, I t think this is the direction we'd want to go with the plans -> to bring things directly from the endpoints - while keeping account only of slugs and IDs on the front end, for the most part (I think?).

If we haven't been updating the endpoints, then this is something that we'll have to do at some point for everything, so it might make sense to ignore here too. So not sure. WDYT about investigating a little more to confirm (and we'll do the same for link-in-bio next).  @southp? 🤔 

some more details:

This is the API that's consumed by the plans data store: `/wpcom/v2/plans/details?locale=en`

This is the shape of the data returned from there:

Screenshot 2023-04-22 at 7 22 12 PM

I haven't looked at the code yet. It looks like we'll need to add the feature objects to `features` (I think the ID would be the slug from here), decide on the `type`/category (might just resolve to "general"), and associate the plans with those features.

packages/calypso-products/src/plans-list.tsx
```diff
@@ -1194,8 +1209,14 @@ const getPlanBusinessDetails = (): IncompleteWPcomPlan => ( {
 	getBlogAudience: () => i18n.translate( 'Best for small businesses' ),
 	getPortfolioAudience: () => i18n.translate( 'Best for small businesses' ),
 	getStoreAudience: () => i18n.translate( 'The plan for small businesses' ),
-	getPlanTagline: () =>
-		i18n.translate( 'Unlock the power of WordPress with plugins and cloud tools.' ),
+	getPlanTagline: ( flowName ) => {
```
chriskmnds: I'm having some second thoughts about this change. Sorry! :/ 

Let me come back to this with a clear mind, but it might be the case we'll be better off creating a separate method, something like `getNewsletterPlanTagline`. This is how it's done for other things here, and it may be easier to port to API-fetched properties later too. WDYT? 🤷

packages/calypso-products/src/constants/features.ts
southp: > If we haven't been updating the endpoints, then this is something that we'll have to do at some point for everything, so it might make sense to ignore here too. So not sure. WDYT about investigating a little more to confirm (and we'll do the same for link-in-bio next

You are right. I believe there are more than a handful of cases like this that we need to confirm before moving the data into the endpoint. When we do it, I suppose we'd also have a great chances to consolidate the features data a bit better, e.g. [these shouldn't be features](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-products/src/constants/cancellation-features.ts#L4).

For now, let's keep it this way so at least it's consistent with the rest :)

packages/calypso-products/src/plans-list.tsx
southp: Interesting. We already have `getNewsletterSignupFeatures()` so this might make the call sites more consitent. I don't have a strong opinion on this though, so I'll defer to @escapemanuele to decide :)

escapemanuele: Another solution could be to have a more generic `getFlowPlanTagline` that accepts a `flowName` property. I was thinking at when we'll want to do the same thing for the Link In Bio flow and others.
But `getNewsletterPlanTagline` is totally fine for me, and it solves our current problem. 

Also, where should the call to `getNewsletterPlanTagline` be? I imagine not in the Plans structure, as to not have the `isNewsletterFlow` function there

chriskmnds: My thinking was it might be easier to migrate individual/dumb methods to API-fetched values. But, like you said, the conditioning will change, and we'll likely end up refactoring more work there. So, as you like, and happy to leave as-is.

chriskmnds: My thinking was it might be easier to migrate individual/dumb methods to API-fetched values. But, like you said, the conditioning will change, and we'll likely end up refactoring more work there. So, as you like, and happy to leave as-is.

packages/calypso-products/src/constants/features.ts
chriskmnds: Makes sense. We'll come back to these in a more focused session.

southp:
@escapemanuele [The planTypes fix](https://github.com/Automattic/wp-calypso/pull/76070) has been deployed. Rebasing this branch, you should be able to just pass `[TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM]` to `planTypes` to curate plans that we need in this flow. Thanks for bearing with me 🙇🏼

escapemanuele:
Thank you @southp, I rebased (it was a long process) and managed to use planTypes 👍

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
southp: @escapemanuele ^ in case you missed this :)

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -105,6 +115,10 @@ const PlansWrapper: React.FC< Props > = ( props ) => {
 			return renderLoading();
 		}
 
+		if ( isNewsletterFlow( flowName ) ) {
+			planTypes = [ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM ];
```
southp: Just to confirm: I know I suggested this [in my lastest comment](https://github.com/Automattic/wp-calypso/pull/75174/#issuecomment-1524849986). However, are we sure we don't need to put `TYPE_BUSINESS` in? I'm asking because I just realized that in the screenshot of the PR description there is actually the Business plan, so I'd like to confirm.

southp:
Thanks for the hard work, @escapemanuele . Overall it looks great already. I've sneaked in one commit https://github.com/Automattic/wp-calypso/pull/75174/commits/3b28d48350388cb3d3f4f3d6a6b7a89a8641c338 and left a few more small inline comments. The major blocker right now is how the unit tests are failing in a way that doesn't look like relating to this PR: p1682652206934719-slack-C02DQP0FP . Hopefully it will be resolved soon 🤞🏼

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
escapemanuele: Yes, Lucas said that `Pablo pushed back the pricing page with 4 plans` [here](https://github.com/Automattic/martech/issues/1533#issuecomment-1494934081) :) Also, updated the PR image!

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
escapemanuele: Totally missed this! Updated the code with [a42be0f](https://github.com/Automattic/wp-calypso/pull/75174/commits/a42be0f7da23e307f4b658462eaece3ad557f862). Also passed down `flowName` in `PlanComparisonGridFeatureGroupRowCell`

chriskmnds:
Hey @escapemanuele . I rebased the PR and fixed some conflicts. Can you take this for a final spin to make sure nothing is broken? I've asked for some help in Slack concerning the failing E2E tests (although not sure if due to the just a rebase needed). Let's sort out the tests tomorrow and move this forward. :-)

packages/calypso-products/package.json
```diff
@@ -36,6 +36,7 @@
 	"homepage": "https://github.com/Automattic/wp-calypso/tree/HEAD/packages/calypso-products#readme",
 	"dependencies": {
 		"@automattic/calypso-config": "workspace:^",
+		"@automattic/onboarding": "workspace:^",
```
chriskmnds: @escapemanuele this might be causing the unit test failures. let's see if removing this dependency fixes the tests. we might have to go the other way with this (what we [discussed earlier](https://github.com/Automattic/wp-calypso/pull/75174#discussion_r1174437990) about creating `getNewsletterTagLine`, etc. VS having the smarter functions that do the conditioning inside `calypso-products`).

client/my-sites/plan-features-2023-grid/components/plan-comparison-grid.tsx
```diff
@@ -493,10 +494,18 @@ const PlanComparisonGridFeatureGroupRowCell: React.FunctionComponent< {
...
+	restructuredFeatures,
+	planName,
+	isStorageFeature,
+	flowName,
+} ) => {
 	const translate = useTranslate();
-	const highlightAdjacencyMatrix = useHighlightAdjacencyMatrix( visiblePlansProperties );
-	const highlightLabel = useHighlightLabel( planName );
+	const highlightAdjacencyMatrix = useHighlightAdjacencyMatrix( visiblePlansProperties, flowName );
+	const highlightLabel = useHighlightLabel( planName, null );
```
chriskmnds: Should this be `const highlightLabel = useHighlightLabel( planName, flowName );` ? 🤔

chriskmnds (approved):
LGTM. Tested and works as advertised. I haven't noticed any regressions in `/start` or `/plans`. Thanks for pulling this through :-)

Issue #74645, opened 2023-03-20
Reader: Remove unnecessary comment hiding under "Conversations" section


davemart-in:
## Description

A user reported that comments were being needlessly hidden within the Reader "Conversations" section. Here is the screenshot they shared:

![image](https://user-images.githubusercontent.com/5634774/226350417-72f2546b-0556-467a-847d-9afe310c28c0.png)

### Expected

Newer comments and comments without too much of a thread should just be shown by default.

### Actual

Many comment threads are truncated by default, even when the comment thread is shallow, or the comment is new.

### Testing Instructions

- Go to https://wordpress.com/read/conversations
- If you don't see posts with comments truncated, try adding https://philosophyandfiction.com to your reader feed

hngdny:
5990448-zd-woothemes

eoigal:
Should be fixed with https://github.com/Automattic/wp-calypso/pull/75187

PR #75187, opened 2023-03-31
Update comment list in conversations

eoigal:
This PR updates how comments are displayed in Reader.

The PR will fix the page title to match what we use on other pages on reader;
**Before**
Screenshot 2023-04-04 at 12 07 21
**After**
Screenshot 2023-04-04 at 12 12 25

We used to only pull in 3 comments, now we will pull in 20; we will also bypass the filtering of parent comments to make them more readable and not necessary to expand all comments.
**Before** 
Screenshot 2023-04-04 at 12 08 34
Screenshot 2023-04-04 at 12 07 43
**After**
Screenshot 2023-04-04 at 12 14 01
Screenshot 2023-04-04 at 12 13 02

I looked into re-ordering the comments, it is possible but the comment stream isn't really setup for it well without some major refactoring. I don't think its worth the effort so decided to leave this out of this PR.

Ref - https://github.com/Automattic/wp-calypso/issues/74645

### DIFF
--- a/client/blocks/reader-post-card/conversation-post.jsx
+++ b/client/blocks/reader-post-card/conversation-post.jsx
@@ -13,7 +13,11 @@ class ConversationPost extends Component {
 		return (
 			<div className="reader-post-card__conversation-post">
 				<CompactPostCard { ...this.props } />
-				<ConversationPostList post={ this.props.post } commentIds={ this.props.commentIds } />
+				<ConversationPostList
+					post={ this.props.post }
+					commentIds={ this.props.commentIds }
+					filterParents={ false }
+				/>
 			</div>
 		);
 	}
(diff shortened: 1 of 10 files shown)


davemart-in (approved):
I can't speak to the code side of things, but I reviewed the functionality and compared it to the existing functionality. This unfurling of comments feels better IMO. No JS errors. Looks good to me!

client/blocks/reader-post-card/conversation-post.jsx
```diff
@@ -13,7 +13,11 @@ class ConversationPost extends Component {
 		return (
 			<div className="reader-post-card__conversation-post">
 				<CompactPostCard { ...this.props } />
-				<ConversationPostList post={ this.props.post } commentIds={ this.props.commentIds } />
+				<ConversationPostList
+					post={ this.props.post }
+					commentIds={ this.props.commentIds }
+					filterParents={ false }
```
roo2, line 19: this seems to be the only place that ConversationPostList is used, so I'd probably just make `filterParents=false` the default and not add the property

roo2 (approved):
LGTM either way but as far as I can tell we don't actually need to add the filterParents property since the component is only called once.

PR #75188, opened 2023-03-31
WoA Logging i1: Add the ability to expand row to show logs vertical layout.

vykes-mac:
Closes https://github.com/Automattic/dotcom-forge/issues/1997

## Proposed Changes

*Add ability to expand rows and view logs vertically

## Testing Instructions

- Get logs for a date range
- Verify multiple rows can be expanded

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
+++ b/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
@@ -1,0 +1,20 @@
+import './style.scss';
+
+interface Props {
+	log: Record< string, unknown >;
+}
+
+export default function SiteLogsExpandedContent( { log }: Props ) {
+	return (
+		<div className="site-logs-table__expanded-content">
+			{ Object.keys( log ).map( ( key ) => (
+				<tr key={ `expanded-row-${ key }` }>
+					<td>{ key }</td>
+					<td>
+						<div className="site-logs-table__expanded-content-info">{ log[ key ] }</div>
+					</td>
+				</tr>
+			) ) }
+		</div>
+	);
+}
--- a/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
+++ b/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
@@ -1,7 +1,8 @@
+import { SiteLogsData } from 'calypso/data/hosting/use-site-logs-query';
 import './style.scss';
 
 interface Props {
-	log: Record< string, unknown >;
+	log: SiteLogsData[ 'logs' ][ 0 ];
 }
 
 export default function SiteLogsExpandedContent( { log }: Props ) {
@@ -10,7 +11,7 @@ export default function SiteLogsExpandedContent( { log }: Props ) {
 			{ Object.keys( log ).map( ( key ) => (
 				<tr key={ `expanded-row-${ key }` }>
 					<td>
-						<strong>{ key }</strong>{ ' ' }
+						<strong>{ key }</strong>
 					</td>
 					<td>
 						<div className="site-logs-table__expanded-content-info">{ log[ key ] || '-' }</div>
--- a/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
+++ b/client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
@@ -9,9 +9,11 @@ export default function SiteLogsExpandedContent( { log }: Props ) {
 		<div className="site-logs-table__expanded-content">
 			{ Object.keys( log ).map( ( key ) => (
 				<tr key={ `expanded-row-${ key }` }>
-					<td>{ key }</td>
 					<td>
-						<div className="site-logs-table__expanded-content-info">{ log[ key ] }</div>
+						<strong>{ key }</strong>{ ' ' }
+					</td>
+					<td>
+						<div className="site-logs-table__expanded-content-info">{ log[ key ] || '-' }</div>
 					</td>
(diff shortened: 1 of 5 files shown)


zaguiini:
We need an empty `<th />` element as the first child of the `thead` node, otherwise the columns will not be aligned:

image

Adding an empty `<th></th>` element to `<thead>`:

image

We could reduce the horizontal padding specifically for the first column, that way it won't take much space.

---

Two things in the details:

- align the heading (left) text on top, vertically, and make it bold;
- add the `-` to the empty values.

Like this:

image

Also, I think it'd be good to move the the timestamp/date as the first item in that list to keep consistency with the table 👍

client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default function SiteLogsExpandedContent( { log }: Props ) {
+	return (
+		<div className="site-logs-table__expanded-content">
+			{ Object.keys( log ).map( ( key ) => (
+				<tr key={ `expanded-row-${ key }` }>
+					<td>
+						<strong>{ key }</strong>
+					</td>
+					<td>
+						<div className="site-logs-table__expanded-content-info">{ log[ key ] || '-' }</div>
```
p-jackson: The type error here is what's causing the build to fail. The values in the log are intentionally typed as `unknown` so that we don't make any assumptions about what's coming back from the server. Take a look at the `renderCell()` function used by `<SiteLogsTableRow>` to see how the table cells handle the `unknown` values.

One of the issues that the error is alerting you to is that values of `0` and `false` are going to be rendered with a dash, which I don't think is what we want.

client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface Props {
+	log: SiteLogsData[ 'logs' ][ 0 ];
+}
+
+export default function SiteLogsExpandedContent( { log }: Props ) {
+	return (
+		<div className="site-logs-table__expanded-content">
+			{ Object.keys( log ).map( ( key ) => (
+				<tr key={ `expanded-row-${ key }` }>
```
p-jackson: Just a note that the React keys don't have to be unique across multiple components or anything

```suggestion
				<tr key={ key }>
```

client/my-sites/site-logs/components/site-logs-table/site-logs-expanded-content.tsx
```diff
@@ -0,0 +1,23 @@
+import { SiteLogsData } from 'calypso/data/hosting/use-site-logs-query';
+import './style.scss';
+
+interface Props {
+	log: SiteLogsData[ 'logs' ][ 0 ];
+}
+
+export default function SiteLogsExpandedContent( { log }: Props ) {
+	return (
+		<div className="site-logs-table__expanded-content">
```
p-jackson, line 12: We also really shouldn't be using `<tr>` elements when they're not children of a `<table>` or `<tbody>`. You could replace the `.site-logs-table__expanded-content` element with a `<table>`. Or the other option could be to use CSS grid to lay out the expanded content. e.g.

```tsx
		<div className="site-logs-table__expanded-content" style={{
			display: 'grid',
			gridTemplateColumns: 'auto 1fr',
			gap: '0.6em 1em'
		}}>
			{ Object.keys( log ).map( ( key ) => (
				<Fragment key={ key }>
					<div>
						<strong>{ key }</strong>
					</div>
					<div>
						<div className="site-logs-table__expanded-content-info">{ log[ key ] || '-' }</div>
					</div>
				</Fragment>
			) ) }
		</div>
```

(inline styling for demonstration purposes only 😄)

client/my-sites/site-logs/components/site-logs-table/style.scss
```diff
@@ -41,3 +42,27 @@ $border-color: #d3dae6;
...
+	font-size: 0.875rem;
+	tr {
+		border: none !important;
+	}
+}
+
+.site-logs-table__expanded-content-info {
+	word-break: break-word;
+	word-wrap: break-word;
+	max-width: 70%;
```
p-jackson: Hmmm, tricky. the whole point of expanding the row is so that you don't have to use the horizontal scrollbar. So we have to have some sort of width specified. The downside of this particular width is that as the screen size gets smaller even the expanded content needs the horizontal scrollbar.

I found another method where, rather than adding a `max-width` on the `.site-logs-table__expanded-content-info` we add a width to the `.site-logs-table__expanded-content` element. It looks something like this:

```scss
.site-logs-table__expanded-content {
	// This makes the expanded content fit inside the viewport
	// The 440px takes into account the Calypso sidebar and the necessary table padding
	width: calc(100vw - 440px);
}

body.is-sidebar-collapsed .site-logs-table__expanded-content {
	// We'd need another rule for when the sidebar is collapsed
	width: calc(100vw - 200px);
}

// And then I guess other rules for mobile breakpoints?
```

mpkelly: I've used this as a base for a solution for all viewport sizes. On mobile, I made it a single column, as there isn't enough space for two columns.

Screenshot 2566-04-10 at 15 34 21

mpkelly:
I implemented changes based on @p-jackson's feedback. I also made some further tweaks so it looks ok on mobile. I have merged it as it's an improvement over what we have. Any rework can be done in another PR.

worldomonation:
@mpkelly Just a drive-by comment that I wrote a P2 on how to avoid review bombs: p7H4VZ-3Oh-p2

PR #75191, opened 2023-03-31
Add Akismet API key to thank you page

jboland88:
## Proposed Changes

* This diff adds the Akismet API key to the thank you page following an Akismet purchase

## Testing Instructions

* Navigate to `http://calypso.localhost:3000/checkout/akismet/thank-you/ak_plus_yearly_1` while logged in
* You should see your API key on the thank you page with a button to copy it

## Pre-merge Checklist


client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
```diff
@@ -72,13 +77,18 @@ const AkismetCheckoutThankYou: FunctionComponent< AkismetCheckoutThankYouProps >
...
 
+				<p className="akismet-checkout-thank-you__api-key-header">
+					{ __( 'Your Akismet API Key is:', 'akismet-thank-you' ) }
+				</p>
+				<ThankYouAPIKeyClipboard />
+
 				<Button
 					primary
 					busy={ isLoading }
 					href="https://akismet.com/account/"
```
CodeyGuyDylan, line 86: Was the plan still to link to the Calypso purchase here? 0-as-1204210417785446/1204285311458080/f

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
```diff
@@ -87,4 +97,48 @@ const AkismetCheckoutThankYou: FunctionComponent< AkismetCheckoutThankYouProps >
 	);
 };
 
+function ThankYouAPIKeyClipboard() {
+	const translate = useTranslate();
+	const [ isCopied, setCopied ] = useState( false );
```
CodeyGuyDylan: I think by convention, the setting function should be named `setIsCopied`

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
```diff
@@ -87,4 +97,48 @@ const AkismetCheckoutThankYou: FunctionComponent< AkismetCheckoutThankYouProps >
...
+	if ( isError ) {
+		return null;
+	}
+
+	const akismetApiKey = data ?? '';
+	const keyInputSize = akismetApiKey ? akismetApiKey.length + 5 : 0;
+
+	return (
+		<div
+			className={ 'akismet-checkout-thank-you__key-clipboard' + ( isLoading ? ' loading' : '' ) }
```
CodeyGuyDylan: For multiple classnames I like importing
```tsx
import classNames from 'classnames';
```
So they can be added in a comma separated list. Not required by any stretch though

client/my-sites/checkout/checkout-thank-you/style.scss
```diff
@@ -881,6 +881,46 @@ main.akismet-checkout-thank-you {
...
+			color: var(--color-text);
+			border: none;
+		}
+
+		.button {
+			min-width: 100px;
+			font-size: 1.25rem;
+		}
+
+		@include breakpoint-deprecated( ">480px" ) {
```
CodeyGuyDylan: I see that the deprecated breakpoint is being used in this file quite a bit, but it looks like this value is the same as if you did this
```scss
@include break-mobile;
```

CodeyGuyDylan (changes requested):
I left three minor non-blockers, the main question was whether or not we will be linking to Calypso purchase instead of `/account`

and1can:
I am testing in localhost and I do not see an API key. I am not saying the issue is caused by the pull request. I am wondering if there is any configuration that needs to happen on my end. I am an OSS Citizen. I do see a logged error that says `Unauthorized Error`. 

checkout_akismet
unauthorized_error

CodeyGuyDylan:
> I am testing in localhost and I do not see an API key. I am not saying the issue is caused by the pull request. I am wondering if there is any configuration that needs to happen on my end. I am an OSS Citizen. I do see a logged error that says Unauthorized Error.

Were you logged in during checkout? If not, I can see being logged out and going straight to the thank you page causing this. If you are logged out during checkout, it creates a temporary user for you once you complete the checkout, and in that case I would think you would avoid getting that error 🤔 

If that is the case, we could probably handle an authentication error better by just not displaying the whole API key section

If you were logged in, I am not entirely sure why that would happen. Maybe Josh has more insight

CodeyGuyDylan:
I can confirm I got the same error when not logged in @jboland88. When I went through the userless checkout instead of going straight to the page, the API key loaded correctly. So I think the main thing to do here is just not render the entire API key section if the key does not load.

I also noticed the placeholder looks a bit off when the API key is loading, should be some space between it and the CTA
image

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
jboland88: I think that would be ideal, however, I was running into the issue of not knowing the correct purchase ID in this context. I wasn't finding much luck in obtaining the purchase ID without having to modify the route and how/ when it gets called after the purchase :) TLDR; I think it's too complex for the current iteration.

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
jboland88: TIL about `classnames`!

client/my-sites/checkout/checkout-thank-you/style.scss
jboland88: Good catch - thanks!

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
```diff
@@ -87,4 +95,56 @@ const AkismetCheckoutThankYou: FunctionComponent< AkismetCheckoutThankYouProps >
...
+
+	return (
+		<>
+			<p className="akismet-checkout-thank-you__api-key-header">
+				{ __( 'Your Akismet API Key is:', 'akismet-thank-you' ) }
+			</p>
+			<div
+				className={ classNames(
+					'akismet-checkout-thank-you__key-clipboard',
+					isLoading ? ' loading' : ''
```
CodeyGuyDylan: I don't think it makes a difference but you shouldn't need the space here anymore 😄 
```suggestion
					isLoading ? 'loading' : ''
```

jboland88:
Thanks for testing and noting an issue @and1can 👏 

@CodeyGuyDylan 
I think we can assume the thank you page should only show when logged in since the CTA is directed to the account page. I've updated the route. Additionally, I've updated the output so that the whole API key section is not shown if there is some other kind of loading error.

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
jboland88: yep yep - applied!

CodeyGuyDylan (approved):
Tests great now! Thanks for the changes 😄

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7905158

Hi @jboland88, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include this string: `Your Akismet API Key is:`

Thank you in advance!

client/my-sites/checkout/checkout-thank-you/akismet-checkout-thank-you.tsx
```diff
@@ -87,4 +95,56 @@ const AkismetCheckoutThankYou: FunctionComponent< AkismetCheckoutThankYouProps >
 	);
 };
 
+function ThankYouAPIKeyClipboard() {
+	const translate = useTranslate();
+	const [ isCopied, setIsCopied ] = useState( false );
+	const { data, isError, isLoading } = useAkismetKeyQuery();
+
+	useEffect( () => {
```
and1can, line 103: note to self: `setTimeout` serves to allow notification to display for ~4 seconds and then remove notification

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7905158).

PR #75210, opened 2023-04-03
Plans grid: Fix styling for decimal prices

aneeshd16:
Fixes Automattic/martech#1653
Fixes Automattic/martech#1674

## Proposed Changes

* This PR displays prices in a reduced font size when any of the displayed plan prices exceed a certain character limit. If the string representation of a price (without currency symbol) is greater than 5 chars, it's classified as a "large" currency. Examples: `10,000`, `100.50`, `100000`. This check cannot be applied independently for each price since it will result in inconsistent styling.
* Introduces a new hook, `useIsLargeCurrency( planSlugs )`, that returns true if any of the prices (discounted/undiscounted) of any of the `planSlugs` is greater than 5 characters.
* Extracted a selector function(`getPlanPrices`) from the existing `usePlanPrices` hook so that it can be used in the `useIsLargeCurrency` hook.
* Changed the font size of "large" currencies to 30px. (Ref: https://github.com/Automattic/martech/issues/1653#issuecomment-1492401794)

## Testing Instructions

* Check `/plans/yearly/<site slug>` where <site slug> is on a monthly plan and has pro-rated credits available in a few different currencies.
* IDR
image

* USD
image

* INR
image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/index.tsx
+++ b/client/my-sites/plan-features-2023-grid/index.tsx
@@ -80,6 +80,7 @@ import {
 	isJetpackSite,
 } from 'calypso/state/sites/selectors';
 import CalypsoShoppingCartProvider from '../checkout/calypso-shopping-cart-provider';
+import useIsLargeCurrency from '../plans/hooks/use-is-large-currency';
 import { getManagePurchaseUrlFor } from '../purchases/paths';
 import PlanFeatures2023GridActions from './actions';
 import PlanFeatures2023GridBillingTimeframe from './billing-timeframe';
@@ -242,6 +243,51 @@ const PlanLogo: React.FunctionComponent< {
 				) }
 			</header>
 		</Container>
+	);
+};
+
+const PlanPricesRow = function ( {
+	planPropertiesObj,
+	options,
+	isReskinned,
+}: {
+	planPropertiesObj: PlanProperties[];
+	options?: PlanRowOptions;
+	isReskinned: boolean;
+} ) {
+	const isLargeCurrency = useIsLargeCurrency(
+		planPropertiesObj.map( ( properties ) => properties.planName as PlanSlug )
+	);
+
+	return (
+		<>
+			{ planPropertiesObj
+				.filter( ( { isVisible } ) => isVisible )
+				.map( ( properties ) => {
+					const { planName, rawPrice } = properties;
+					const classes = classNames( 'plan-features-2023-grid__table-item', 'is-bottom-aligned', {
+						'has-border-top': ! isReskinned,
+					} );
+					const hasNoPrice = rawPrice === undefined || rawPrice === null;
+
+					return (
+						<Container
+							scope="col"
+							key={ planName }
+							className={ classes }
+							isMobile={ options?.isMobile }
+						>
+							{ ! hasNoPrice && (
+								<PlanFeatures2023GridHeaderPrice
+									planProperties={ properties }
+									is2023OnboardingPricingGrid={ true }
+									isLargeCurrency={ isLargeCurrency }
+								/>
+							) }
+						</Container>
+					);
+				} ) }
+		</>
 	);
 };
 
@@ -489,38 +535,15 @@ export class PlanFeatures2023Grid extends Component<
 	}
 
 	renderPlanPrice( planPropertiesObj: PlanProperties[], options?: PlanRowOptions ) {
-		const { isReskinned, is2023OnboardingPricingGrid } = this.props;
+		const { isReskinned } = this.props;
 
-		const isLargeCurrency = planPropertiesObj.some(
-			( properties ) => properties?.rawPrice && properties?.rawPrice > 99000
+		return (
+			<PlanPricesRow
+				planPropertiesObj={ planPropertiesObj }
+				isReskinned={ isReskinned }
+				options={ options }
+			/>
 		);
-
-		return planPropertiesObj
-			.filter( ( { isVisible } ) => isVisible )
-			.map( ( properties ) => {
-				const { planName, rawPrice } = properties;
-				const classes = classNames( 'plan-features-2023-grid__table-item', 'is-bottom-aligned', {
-					'has-border-top': ! isReskinned,
-				} );
-				const hasNoPrice = rawPrice === undefined || rawPrice === null;
-
-				return (
-					<Container
-						scope="col"
-						key={ planName }
-						className={ classes }
-						isMobile={ options?.isMobile }
-					>
-						{ ! hasNoPrice && (
-							<PlanFeatures2023GridHeaderPrice
-								planProperties={ properties }
-								is2023OnboardingPricingGrid={ is2023OnboardingPricingGrid }
-								isLargeCurrency={ isLargeCurrency }
-							/>
-						) }
-					</Container>
-				);
-			} );
 	}
 
 	renderBillingTimeframe( planPropertiesObj: PlanProperties[], options?: PlanRowOptions ) {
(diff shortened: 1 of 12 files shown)


client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -493,36 +537,15 @@ export class PlanFeatures2023Grid extends Component<
...
+		const { isReskinned } = this.props;
 
-		const isLargeCurrency = planPropertiesObj.some(
-			( properties ) => properties?.rawPrice && properties?.rawPrice > 99000
+		return (
+			<PlanPricesRow
+				planPropertiesObj={ planPropertiesObj }
+				isReskinned={ isReskinned }
+				options={ options }
+			/>
```
aneeshd16: Extracted this into a functional component to use the `useIsLargeCurrency` hook.

client/my-sites/plans-comparison/use-plan-prices.tsx
```diff
@@ -31,6 +31,9 @@ function toMonthlyPrice( plan: WPComPlan ) {
 	};
 }
 
+/**
+ * @deprecated
+ */
```
aneeshd16, line 38: This hook has the same name as the one added recently. It seems to be having similar functionality, and I found only 1 usage. @chriskmnds, wondering if you had noticed this one and if it can be replaced by the new hook? For now, I added the `@deprecated` comment so that it is not used by mistake.

chriskmnds: I did notice, and I think recall using it in the previous plans overhaul. It's an earlier version that's doing the monthly conversions the selectors handle natively now. `@deprecated` is fine. we can replace it, but might be subject to throw-away instead along with any previous grids not in use

client/my-sites/plans/hooks/use-is-large-currency.ts
```diff
@@ -0,0 +1,28 @@
...
+ * This is primarily used for lowering the font-size of "large" display prices.
+ */
+export default function useIsLargeCurrency( planSlugs: PlanSlug[] ): boolean {
+	return useSelector( ( state: IAppState ) => {
+		const siteId = getSelectedSiteId( state ) ?? null;
+		return planSlugs.some( ( planSlug ) => {
+			const { discountedRawPrice, planDiscountedRawPrice, rawPrice } = getPlanPrices( state, {
+				planSlug,
+				siteId,
+				returnMonthly: true,
```
chriskmnds: I think we should pass this through as a parameter rather than hardcode it here. we can set `true` as the default value.

client/my-sites/plans/hooks/use-is-large-currency.ts
```diff
@@ -0,0 +1,28 @@
...
+			} );
+			return [ rawPrice, discountedRawPrice, planDiscountedRawPrice ].some(
+				( price ) => price?.toString().length > LARGE_CURRENCY_CHARS
```
chriskmnds: NIT - I think, semantically, the comparison would make more sense if the constant was `SMALL_CURRENCY_CHARS` (or some similar name) i.e. the condition to read like "if greater than small currency chars". Right now it reads "if greater than large currency chars", which feels incorrect

client/my-sites/plans/hooks/use-plan-prices.ts
chriskmnds, line 1: We don't really need a custom hook anymore if it's only to call `getPlanPrices`. Unless there's intention to add more logic to it, then I would opt for removing it. We can just use `useSelector` in its place instead.

client/state/plans/selectors/get-plan-prices.ts
chriskmnds, line 1: Should we add a test for the new selector? A simple one akin to what we have for `usePlanPrices` might do.

client/my-sites/plans/test/use-is-large-currency.ts
chriskmnds, line 1: Nice!

chriskmnds (changes requested):
Thanks, @aneeshd16. Tests out and the refactors look great. I appreciate our focus on tests. 😎 

I left some comments below to hopefully tune things a little more. and just a minor inconsistency with the grid in tablet and mobile views:

Screenshot 2023-04-07 at 2 21 56 PM

client/my-sites/plans/hooks/use-plan-prices.ts
aneeshd16: Agreed! But I am seeing that the hook is already being used in a few in-progress PRs. What do you think about making this change after the "dust settles"? 🙂

client/my-sites/plans/hooks/use-is-large-currency.ts
aneeshd16: Good call out! I renamed this to `LARGE_CURRENCY_CHAR_THRESHOLD` so the condition reads as "length of price greater than large currency character threshold".

client/my-sites/plans/hooks/use-is-large-currency.ts
aneeshd16: Done!

client/state/plans/selectors/get-plan-prices.ts
aneeshd16: Done! Added.

aneeshd16:
Thanks @chriskmnds, for the review! I have addressed your comments and it is ready for another round.

southp:
@aneeshd16 It looks like solving Automattic/martech#1674 as well. Is that correct? 😎

aneeshd16:
> @aneeshd16 It looks like solving [Automattic/martech#1674](https://github.com/Automattic/martech/issues/1674) as well. Is that correct? 😎

It does! Adding this to the PR description.

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1008,6 +1014,15 @@ export class PlanFeatures2023Grid extends Component<
...
 }
 
+const withIsLargeCurrency = ( Component: LocalizedComponent< typeof PlanFeatures2023Grid > ) => {
+	return function ( props: PlanFeatures2023GridType ) {
+		const isLargeCurrency = useIsLargeCurrency(
+			( props.planProperties || [] ).map( ( properties ) => properties.planName as PlanSlug )
+		);
+		return <Component { ...props } isLargeCurrency={ isLargeCurrency } />;
+	};
+};
```
chriskmnds, line 1024: was there no way to avoid the HOC? It should be doable if we refactor renderPlanPrice into a PlanPrice component, right?

p.s. I foresee we'll be doing a lot less prop drilling and more component extractions and hook usage on additional refactors.

aneeshd16: > was there no way to avoid the HOC? It should be doable if we refactor renderPlanPrice into a PlanPrice component, right?

That's what I had done earlier :) I had to revert that change and add the HOC via https://github.com/Automattic/wp-calypso/pull/75210/commits/8b539b9acc8401a39556966e4d8b1591642d3882

While investigating [this](https://github.com/Automattic/wp-calypso/pull/75210#pullrequestreview-1376074288) bug, I noticed that tablet view renders two separate tables:
https://github.com/Automattic/wp-calypso/blob/478ceeb1cb32c600c915733238765945f670f20f/client/my-sites/plan-features-2023-grid/index.tsx#L403-L412

So effectively, for tablet, the value would have to be calculated in the `renderTabletView` function and then passed down to `renderTable`. Similarly, for desktop, the value would have to be calculated in the `render` function itself.

chriskmnds: Gotcha. Thanks for explaining. It sounds like a pretty big refactor otherwise (the whole thing moving to a functional component pretty much), and something for another time.

chriskmnds (approved):
Tests out to me. Nice job. Thanks for addressing the feedback. :shipit:

client/my-sites/plans/hooks/use-plan-prices.ts
chriskmnds: Thanks. I thought, was sure, I replied here 🤔 

Sounds good. We can come back to this once things settle.

PR #75213, opened 2023-04-03
Add/ useSiteUnfollowMutation tests

### LINKED ISSUE #75212: Create tests for useSiteUnfollowMutation

JuanLucha:
Closes https://github.com/Automattic/wp-calypso/issues/75212

## Proposed Changes

This PR adds tests for useSiteUnfollowMutation.

Screenshot 2023-04-03 at 11 11 14

## Testing Instructions

1. Apply these changes.
2. Run the tests with `yarn jest packages/data-stores/src/reader/test/mutations.test.tsx`. They should pass:
Screenshot 2023-04-03 at 11 36 53
3. In order to avoid false positives, break them. Some suggestions are:
  - change the `apiVersion` in line 96 to any value other than `1.2`, like `5`.
  - change the `success` in line 113 to `true`.
  - change the `blog_id` in line 132 to `123`.
If they all break after these changes, they weren't false positives.

Thanks for the review! 🤖 💘 

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/jest.config.js
+++ b/packages/data-stores/jest.config.js
@@ -1,3 +1,10 @@
+const jestPreset = require( '../../test/packages/jest-preset.js' );
+
 module.exports = {
-	preset: '../../test/packages/jest-preset.js',
+	...jestPreset,
+	testPathIgnorePatterns: [
+		...( jestPreset.testPathIgnorePatterns || [] ), // Spread existing ignore patterns (if any)
+		// Ignore test utils files
+		'[a-zA-Z/]*test/utils.ts[x]?',
+	],
 };
--- a/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
+++ b/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
@@ -21,7 +21,7 @@ const useUserSettingsMutation = () => {
 				} else {
 					throw new Error(
 						translate( 'Something went wrong.', {
-							context: 'Something went wrong will saving the user settings',
+							context: 'Something went wrong with saving the user settings',
 						} ) as string
 					);
 				}
--- a/packages/data-stores/src/reader/test/use-site-delivery-frequency-mutation.test.tsx
+++ b/packages/data-stores/src/reader/test/use-site-delivery-frequency-mutation.test.tsx
@@ -2,39 +2,32 @@
  * @jest-environment jsdom
  */
 import { render, waitFor } from '@testing-library/react';
-import { useEffect } from 'react';
-import { QueryClient, QueryClientProvider } from 'react-query';
+import React from 'react';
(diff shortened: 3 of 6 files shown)


packages/data-stores/src/reader/test/mutations.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+	return Skeleton;
+};
+
+beforeEach( () => {
+	jest.resetAllMocks();
+
+	( useIsLoggedIn as jest.Mock ).mockReturnValue( true );
+} );
+
+describe( 'useSubscriptionManagerSiteDeliveryFrequencyMutation', () => {
```
phcp: It should follow the new function name: `useSiteDeliveryFrequencyMutation`

packages/data-stores/src/reader/test/mutations.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+					delivery_frequency: 'daily',
+				},
+				isLoggedIn: true,
+				method: 'POST',
+			} )
+		);
+	} );
+} );
+
+describe( 'useSiteUnfollowMutation', () => {
```
phcp: I think it would be more readable if we split this into multiple files, with the same name as the file to tested. And the reusable part (for now just the `buildSkeleton` function) could be moved to a `utils.ts` file inside the test folder.

TimBroddin:
It would be nice to also have tests to see if the invalidated queries are re-run. I looked into it earlier, and it should be possible to mock the hooks and see how many times they're called with `toHaveBeenCalledTimes`. 

https://github.com/TanStack/query/discussions/1745

packages/data-stores/jest.config.js
```diff
@@ -1,3 +1,10 @@
+const jestPreset = require( '../../test/packages/jest-preset.js' );
+
 module.exports = {
-	preset: '../../test/packages/jest-preset.js',
+	...jestPreset,
+	testPathIgnorePatterns: [
+		...( jestPreset.testPathIgnorePatterns || [] ), // Spread existing ignore patterns (if any)
+		// Ignore test utils files
+		'[a-zA-Z/]*test/utils.ts[x]?',
+	],
```
JuanLucha, line 9: The change in this file is required for the utils file to live inside the test folder and jest not take it as a test file.

phcp: Based on `test/packages/jest.config.js` implementation, seems like we can have our own `jest.config.js` (inside reader folder) file with our specific implementation. I'm ok with both, but if you prefer to keep on the higher level  make sure there is no impact on other tests.

phcp:
@JuanLucha I'm getting the error below while running the tests on my local. Seems that there is still some setup to do on the jest config.
`yarn jest packages/data-stores/src/reader/test`

image

PR #75223, opened 2023-04-03
Woo Express: Added Essential plan to trials Plans page

nelsonec87:
Added Essential plan to trial's Plans page:

*This PR is not updating the features' copy. This is gonna be handled in a separate PR.

image

### Testing
* Clicking the Upgrade button should added the Plan to the cart, associated to the current site.


daledupreez (approved):
This matches the described intent, but I think there are some gaps that are worth calling out, even if we're likely to tackle them in follow-up PRs:
* We don't have a monthly vs annual selector
* The CTAs for both the Essential and Performance plans are primary -- only the Performance plan should have the primary colour

PR #75227, opened 2023-04-03
Tailored Flow: Fix decide later on domain

### LINKED ISSUE #75214: Tailored Flow: Not working with selecting Decide Later on Domains step
If we select `Decide Later`on Newsletter/Link-InBio, on the Domains step, siteSlug become undefined.

 That is caused by a call to [site creation](https://github.com/Automattic/wp-calypso/blob/trunk/packages/onboarding/src/cart/index.tsx#L130), when [useAutoGeneratedBlogName](https://github.com/Automattic/wp-calypso/blob/trunk/packages/onboarding/src/cart/index.tsx#L62) is true. 

The bug happens when we use the same siteTitle more times.

Error returned is `managed_subdomain_not_available` and message: `This address is already taken. Please choose another one.`

escapemanuele:
##  What
`Decide Later` does not set the domain cart item on the store, so if we click that and still have something in the store from a previous test, it will use that and cause troubles. Now it deletes the store value.

Also, when we added Plans and Domains to the Newsletter flow, we missed clearing the destination cookie upon start, causing problems during the site creation.

## Testing
1. Live Link
2. Go through the Newsletter Flow and use `Decide Later` on the Domains step
3. Try with paid and free plans.
4. Test for regression on free or paid domains

Fixes https://github.com/Automattic/wp-calypso/issues/75214

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/index.tsx
@@ -100,6 +100,8 @@ const DomainsStep: Step = function DomainsStep( { navigation, flow } ) {
 
 			setHideFreePlan( Boolean( suggestion.product_slug ) || shouldHideFreePlan );
 			setDomainCartItem( domainCartItem );
+		} else {
+			setDomainCartItem( undefined );
 		}
 
 		submit?.();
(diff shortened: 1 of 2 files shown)


alshakero (approved):
Beautiful! LGTM

PR #75231, opened 2023-04-03
LiB Domain: Remove plans step from lib connect

escapemanuele:
## What
Context: pcYYhz-13z-p2#comment-1121

We now just add the personal plan to the cart and skip the plan step.

## Testing
1. Calypso Live link
2. `/setup/link-in-bio-domain?domain=yourdomain.com&provider=test-provider`
3. No Plans step should be visible and on checkout you should see a Personal Plan added


alshakero (approved):
Works like a charm.

PR #75232, opened 2023-04-03
Woo Express: Added interval toggle to Woo Express plans

nelsonec87:
Added Interval toggle to trial's Plans page:

*This PR is not updating the features' copy. This is going to be handled in a separate PR.

### Testing
* You should see the toggle, and using it should update the prices.

Monthly:
image

Yearly:
image


daledupreez (approved):
LGTM!

One area of follow-up to note is that we're using the WordPress.com brand font for the currency and plan names, which we'll want to correct later on.

PR #75235, opened 2023-04-03
Woo Express: Added Woo Express styles to CtA

nelsonec87:
* Added Woo Express styles to the plans buttons
* Removed the Woo logo

*This PR is not updating the features' copy. This is going to be handled in a separate PR.

### Testing

Before:
image

After:
image

### DIFF
--- a/client/my-sites/plan-features-2023-grid/style.scss
+++ b/client/my-sites/plan-features-2023-grid/style.scss
@@ -56,6 +56,23 @@
 				}
 			}
 
+			&.is-wooexpress-medium-plan {
+				background-color: var(--studio-pink-50);
+				color: var(--studio-white);
+				font-weight: 400;
+
+				&:focus {
+					box-shadow: 0 0 0 2px var(--studio-white), 0 0 0 4px var(--studio-pink-50);
+				}
+			}
+
+			&.is-wooexpress-small-plan {
+				background-color: var(--studio-white);
+				color: var(--studio-pink-50);
+				border: 1px solid  var(--studio-pink-50);
+				font-weight: 400;
+			}
+
 			&.is-ecommerce-plan {
 				background-color: #55347d;
 				color: var(--studio-purple-0);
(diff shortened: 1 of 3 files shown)


client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -56,6 +56,23 @@
 				}
 			}
 
+			&.is-wooexpress-medium-plan {
+				background-color: var(--studio-pink-50);
```
sixhours: I think instances of `--studio-pink-50` here should be `--color-primary` so they change with user-specified color schemes, otherwise you get weirdness like this (blue color scheme with pink buttons):

Screen Shot 2023-04-03 at 11 31 25 AM

sixhours (changes requested):
This is working as expected in my testing. I added a comment about updating the color values to respect user color schemes.

client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -56,6 +56,23 @@
...
+				&:focus {
+					box-shadow: 0 0 0 2px var(--studio-white), 0 0 0 4px var(--studio-pink-50);
+				}
+			}
+
+			&.is-wooexpress-small-plan {
+				background-color: var(--studio-white);
+				color: var(--studio-pink-50);
+				border: 1px solid  var(--studio-pink-50);
+				font-weight: 400;
```
daledupreez: I think we should the same values as we do for standard buttons.
```suggestion
			&.is-wooexpress-medium-plan {
				background-color: var(--color-accent);
				color: var(--color-text-inverted);
				font-weight: 400;

				&:focus {
					box-shadow: 0 0 0 2px var(--color-surface), 0 0 0 4px var(--color-accent);
				}
			}

			&.is-wooexpress-small-plan {
				background-color: var(--color-surface);
				color: var(--color-accent);
				border: 1px solid  var(--color-accent);
				font-weight: 400;
```

sixhours: Whoops, Dale's right, it's the accent color you want. Good catch :D

sixhours (approved):
Screen Shot 2023-04-03 at 11 55 53 AM

LGTM!

PR #75269, opened 2023-04-04
Packages: Remove unused `@automattic/launch` package and launch store

tyxla:
## Proposed Changes

This PR removes the unused `automattic/launch` store and the unused `@automattic/launch` package.

Part of #74399. A follow-up to #73890.

## Testing Instructions

* Verify that the removed store is not in use.
* Verify the removed package is not in use.
* All checks should be green and Calypso should build well.


tyxla:
There are a few `@automattic/plans-grid` components that are no longer in use once this lands, but we can address them separately to keep this PR from growing indefinitely.

noahtallen (approved):
Tests well and couldn't find any other usage of the code.

daledupreez:
@tyxla, I just merged a PR and was notified that ETK was modified, but when I went to deploy ETK, it looks like this PR hasn't been deployed yet. Could you take a look at D107274-code and maybe help me test?

I think my changes are effectively a no-op.

noahtallen:
@daledupreez I took a look at that diff and it looks fine. I agree, it looks like a noop. (Mostly because we removed where that store was used in a separate PR/diff, so importing it there isn't really doing anything)

tyxla:
I second this, it's a noop, we were just removing dead / unused code.

PR #75279, opened 2023-04-04
Data Stores: Remove unused i18n store

tyxla:
## Proposed Changes

This PR removes the `i18n` store that we are no longer using since #74475.

## Testing Instructions

* Verify the removed store is no longer in use.
* Verify all checks are green.


noahtallen (approved):
Nice!

yuliyan (approved):
LGTM 👍

PR #75293, opened 2023-04-04
Checkout: Add errorMessage to callback in CheckoutPaymentMethods error boundary

sirbrillig:
## Proposed Changes

The `CheckoutErrorBoundary` component is a React error boundary that does two things if there is an error during render: it renders its `errorMessage` prop to the user and it (optionally) calls its `onError` callback prop, passing the error itself as an argument. The callback typically logs this message for debugging purposes.

Most of the time this is fine, as the error contains all the information needed to figure it out. However, sometimes the `errorMessage` prop contains contextual information that would further help debug the issue.

In this PR, we add the `errorMessage` (iff it's a string) to the error passed to the callback so that it can be logged also.

**NOTE**: in order to have TypeScript allow us to use `error.cause`, we have to update its config to allow using `ES2022`. Apparently we already allow ES2022 inside the `client` directory (although I can't understand how), but not in `packages` so this should be safe, but please speak up in review if you think it's a bad idea.

## Testing Instructions

We need to cause an error and verify that it gets logged with both `error.message` and `error.cause`. To do that, add an error to checkout in a locally running version of calypso on this branch like the following:

```diff
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -203,6 +203,8 @@ export default function WPCheckout( {
        const [ isSummaryVisible, setIsSummaryVisible ] = useState( false );
        const { formStatus } = useFormStatus();

+       throw new Error( 'this is a test' );
+
        const onReviewError = useCallback(
                ( error: Error ) =>
                        onPageLoadError( 'step_load', error, {
```

Then, visit checkout with your browser's network devtools open. Look for a connection to the `logstash` endpoint with a body that includes both the message and the test error as a `Cause`. For example:

Screenshot 2023-04-04 at 3 45 53 PM

### DIFF
--- a/client/me/purchases/manage-purchase/payment-method-selector/index.tsx
+++ b/client/me/purchases/manage-purchase/payment-method-selector/index.tsx
@@ -43,6 +43,13 @@ const { colors } = colorStudio;
 const jetpackColors = isJetpackCloud() ? { highlight: colors[ 'Jetpack Green 50' ] } : {};
 const theme = { ...checkoutTheme, colors: { ...checkoutTheme.colors, ...jetpackColors } };
 
+function convertErrorToString( error: Error ): string {
+	if ( error.cause ) {
+		return `${ error.message }; ${ error.cause }`;
+	}
+	return error.message;
+}
+
 function useLogError( message: string ): CheckoutPageErrorCallback {
 	return useCallback(
 		( errorType, errorMessage ) => {
@@ -53,7 +60,7 @@ function useLogError( message: string ): CheckoutPageErrorCallback {
 				extra: {
 					env: config( 'env_id' ),
 					type: 'payment_method_selector',
-					message: String( errorMessage ),
+					message: convertErrorToString( errorMessage ),
 					errorType,
 				},
 			} );
--- a/client/me/purchases/manage-purchase/payment-method-selector/index.tsx
+++ b/client/me/purchases/manage-purchase/payment-method-selector/index.tsx
@@ -45,14 +45,14 @@ const theme = { ...checkoutTheme, colors: { ...checkoutTheme.colors, ...jetpackC
 
 function convertErrorToString( error: Error ): string {
 	if ( error.cause ) {
-		return `${ error.message }; ${ error.cause }`;
+		return `${ error.message }; Cause: ${ error.cause }`;
 	}
 	return error.message;
 }
 
 function useLogError( message: string ): CheckoutPageErrorCallback {
 	return useCallback(
-		( errorType, errorMessage ) => {
+		( errorType, error ) => {
 			logToLogstash( {
 				feature: 'calypso_client',
 				message,
@@ -60,7 +60,7 @@ function useLogError( message: string ): CheckoutPageErrorCallback {
 				extra: {
 					env: config( 'env_id' ),
 					type: 'payment_method_selector',
-					message: convertErrorToString( errorMessage ),
+					message: convertErrorToString( error ),
 					errorType,
 				},
 			} );
--- a/packages/calypso-typescript-config/js-package.json
+++ b/packages/calypso-typescript-config/js-package.json
@@ -7,9 +7,9 @@
 
 		"noEmit": true,
 
-		"lib": [ "ES2020", "DOM", "DOM.Iterable" ],
-		"module": "ES2020",
-		"target": "ES2020",
+		"lib": [ "ES2022", "DOM", "DOM.Iterable" ],
+		"module": "ES2022",
+		"target": "ES2022",
 		"esModuleInterop": true,
 		"jsx": "react-jsx",
 		"allowJs": true,
--- a/packages/calypso-typescript-config/ts-package.json
+++ b/packages/calypso-typescript-config/ts-package.json
@@ -13,7 +13,7 @@
 		"noEmitHelpers": true,
 		"skipLibCheck": true,
 
-		"lib": [ "ES2020", "DOM", "DOM.Iterable" ],
+		"lib": [ "ES2022", "DOM", "DOM.Iterable" ],
 		"module": "ES2020",
 		"target": "ES2020",
 		"esModuleInterop": true,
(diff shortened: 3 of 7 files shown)


packages/calypso-typescript-config/ts-package.json
noahtallen, line 1: We might want to update the js-package file while we're here too.

noahtallen (commented):
I think the client might use ES2020 as well. `client/tsconfig.json` seems to extend the mixed config, which in turn extends the main ts config we're updating here

packages/calypso-typescript-config/ts-package.json
```diff
@@ -13,7 +13,7 @@
 		"noEmitHelpers": true,
 		"skipLibCheck": true,
 
-		"lib": [ "ES2020", "DOM", "DOM.Iterable" ],
+		"lib": [ "ES2022", "DOM", "DOM.Iterable" ],
 		"module": "ES2020",
 		"target": "ES2020",
```
noahtallen: I think it'd be ok to update this to ES2022 as well

sirbrillig:
> I think the client might use ES2020 as well. client/tsconfig.json seems to extend the mixed config, which in turn extends the main ts config we're updating here

That's what I thought I understood too but I've checked and there are no TS errors if you try to do `throw new Error('something', { cause: new Error('other') })` inside `client` even without this change 🤷 .

noahtallen (commented):
Another side note -- I'm unsure how big an impact changing the target and lib options would have. I know that we run typechecks separately in CI, and IIRC, babel ultimately transforms the typescript files without any type checks. I imagine that follows babel's rules rather than tsconfig

packages/calypso-typescript-config/ts-package.json
sirbrillig: I don't totally understand the subtleties between `lib`, `module`, and `target`, so I had done only the minimum required. However, that's easy to do. I've updated all the other mentions of `ES2020` in the TS/JS configs to `ES2022`.

noahtallen (approved):
I'm happy with these changes, but probably a good idea to get more of @Automattic/team-calypso-frameworks involved

client/me/purchases/manage-purchase/payment-method-selector/index.tsx
```diff
@@ -53,7 +60,7 @@ function useLogError( message: string ): CheckoutPageErrorCallback {
 				extra: {
 					env: config( 'env_id' ),
 					type: 'payment_method_selector',
-					message: String( errorMessage ),
+					message: convertErrorToString( errorMessage ),
```
tyxla: Should we rename the `errorMessage` argument? Feels like an unexpected name for an error object.

packages/composite-checkout/README.md
```diff
@@ -121,7 +121,7 @@ It has the following props.
 - `onPaymentComplete?: ({paymentMethodId: string | null, transactionLastResponse: unknown }) => null`. A function to call for non-redirect payment methods when payment is successful. Passed the current payment method id and the transaction response as set by the payment processor function.
 - `onPaymentRedirect?: ({paymentMethodId: string | null, transactionLastResponse: unknown }) => null`. A function to call for redirect payment methods when payment begins to redirect. Passed the current payment method id and the transaction response as set by the payment processor function.
 - `onPaymentError?: ({paymentMethodId: string | null, transactionError: string | null }) => null`. A function to call for payment methods when payment is not successful.
-- `onPageLoadError?: ( errorType: string, errorMessage: string, errorData?: Record< string, string | number | undefined > ) => void`. A function to call when an internal error boundary triggered.
+- `onPageLoadError?: ( errorType: string, errorMessage: Error, errorData?: Record< string, string | number | undefined > ) => void`. A function to call when an internal error boundary triggered.
```
tyxla: Same thought on renaming.

packages/calypso-typescript-config/js-package.json
```diff
@@ -7,9 +7,9 @@
 
 		"noEmit": true,
 
-		"lib": [ "ES2020", "DOM", "DOM.Iterable" ],
-		"module": "ES2020",
-		"target": "ES2020",
+		"lib": [ "ES2022", "DOM", "DOM.Iterable" ],
+		"module": "ES2022",
+		"target": "ES2022",
```
tyxla, line 12: Looks good to me 👍 We're already using ES2022 features, and our browser support policy is to support the latest versions, so I don't have concerns about this.

packages/composite-checkout/src/components/checkout-error-boundary.tsx
```diff
@@ -20,6 +20,9 @@ export default class CheckoutErrorBoundary extends Component< CheckoutErrorBound
 
 	componentDidCatch( error: Error ): void {
 		if ( this.props.onError ) {
+			const errorContext =
+				typeof this.props.errorMessage === 'string' ? this.props.errorMessage : error.message;
+			error = new Error( errorContext, { cause: error } );
 			this.props.onError( error );
```
tyxla:
```suggestion
			const errorWithCause = new Error( errorContext, { cause: error } );
			this.props.onError( errorWithCause );
```

tyxla (approved):
LGTM 👍 

I've left a few minor suggestions, but feel free to ship regardless of whether you'll decide to address them.

packages/calypso-typescript-config/ts-package.json
tyxla: FWIW the difference is well explained here:

https://stackoverflow.com/questions/67371787/what-typescript-configuration-produces-output-closest-to-node-js-16-capabilities/67371788#67371788

client/me/purchases/manage-purchase/payment-method-selector/index.tsx
sirbrillig: Good idea.

packages/composite-checkout/src/components/checkout-error-boundary.tsx
sirbrillig: Good idea. Better than reassignment.

PR #75303, opened 2023-04-04
Components: Update CircularProgressBar Prop Handling

### LINKED ISSUE #72789: Launchpad: Handle edge cases for CircularProgressBar component
We've recently worked on a [CircularProgressBar](https://github.com/Automattic/wp-calypso/blob/trunk/packages/components/src/circular-progress-bar/index.tsx#L3) component as part of the cohesive Launchpad Experience.

We want to make sure we are handling some edge cases gracefully. We are using the `currentStep` and `numberOfSteps` props to calculate the progress arc. We want to make sure the following cases look good:
1. `currentStep` = `0` and `numberOfSteps` = `0`
2. `currentStep` = `0` and `numberOfSteps` = `other than 0`

sharpirate:
Fixes https://github.com/Automattic/wp-calypso/issues/72789

### Time Estimate
Testing: Short
Review: Short

### Summary
This PR updates the **CircularProgressBar** component to improve the behavior with invalid props:
1. If the `currentStep` prop is set to `0`, we display an empty circle. Previously the progress circle was still showing some small amount of progress
2. Move the invalid prop handling inside the component. Previously we had to check if the props are valid before rendering the component but now the component has its own built in logic for handling invalid props. I believe this is a better practice for this particular component because we can provide consistent graceful fallback built into the component. For now we are not rendering the component (returning `null`) if the props are invalid (`numberOfSteps` is `undefined` or `0`, `currentStep` is `undefined`)

### Testing
1. Checkout this branch
2. Arrive at Launchpad on a Launchpad-enabled site
3. Use the `React DevTools` to manually adjust the props of the `CircularProgressBar` component:
3.1. Set the `currentStep` to `0`. Verify the circle progress circle is empty.

currentStep

3.2. Set the `numberOfSteps` to `0`. Verify the `CircularProgressBar` component doesn't render.

numberOfSteps

4. Observe the `CircularProgressBar` as you refresh the page. The component would not render until its props are initialized.
5. Checkout trunk. Repeat the above steps and observe the difference in behavior.
6. Verify the `CircularProgressBar` continues to behave properly as you're going through the Launchpad tasks

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx
@@ -117,16 +117,14 @@ const Sidebar = ( { sidebarDomain, siteSlug, submit, goNext, goToStep, flow }: S
 				<span className="launchpad__sidebar-header-flow-name">{ flowName }</span>
 			</div>
 			<div className="launchpad__sidebar-content-container">
-				{ currentTask && enhancedTasks?.length && (
-					<div className="launchpad__progress-bar-container">
-						<CircularProgressBar
-							size={ 40 }
-							enableDesktopScaling
-							currentStep={ currentTask }
-							numberOfSteps={ enhancedTasks?.length }
-						/>
-					</div>
-				) }
+				<div className="launchpad__progress-bar-container">
+					<CircularProgressBar
+						size={ 40 }
+						enableDesktopScaling
+						currentStep={ currentTask }
+						numberOfSteps={ enhancedTasks?.length }
+					/>
+				</div>
 				{ /* eslint-disable-next-line wpcalypso/jsx-classname-namespace*/ }
 				<h1 className="launchpad__sidebar-h1">
 					{ showLaunchTitle && launchTitle ? launchTitle : title }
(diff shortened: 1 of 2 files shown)


agrullon95:
## Tests:

- [x] Verify setting `currentStep` prop to 0 displays an empty circle
- [x] Verify setting `numberOfSteps` prop to 0 hides the CircularProgressBar component

agrullon95 (approved):

edanzer (approved):
Code changes look straightforward and this tests as expected for me. 
- [x] Set the currentStep to 0 via react dev tools and confirm the circle progress circle is empty.
- [x] Set the numberOfSteps to 0 via react dev tools and confirm CircularProgressBar component doesn't render.
- [x] Verify the progress icon updates as expected as we move through launchpad steps.

UPDATE: Some tests are failing on the PR. I re-ran those and they failed again. So I took a look and they look like they are legit. It seems we have a TS error in the test build logs: 

`client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/sidebar.tsx(124,7): error TS2322: Type 'number | null' is not assignable to type 'number | undefined'.`

Not sure why this isn't being caught in the editor at build time when developing, but seems like we may need to resolve that.

PR #74473, opened 2023-03-15
Data Stores: Refactor `WpcomFeatures` store to `createReduxStore()`

tyxla:
## Proposed Changes

This PR migrates the `WpcomFeatures` store to use `createReduxStore()` and `register()` instead of `registerStore()`.

Part of #74399. A follow-up to #73890.

## Testing Instructions

* This is used in ETK - @noahtallen - any ideas on how to test it?
* Verify all checks are green.

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #75304, opened 2023-04-04
Update wordpress/editor store calls to import approach

noahtallen:
## Proposed Changes
**Note: typecheck is currently failing, but the relevant package will be removed soon. (#75269)**

Related to #74473. The benefits to importing stores directly are:

1. Explicit dependencies, not implicit dependencies.
2. Better types (where applicable).

So this makes sure we reference the 'core/editor` store via an import instead of a string.

The main motivation for this PR was to fix some deprecation warnings on trunk. (Some selectors were moved to core/block-editor instead, so those were also updated.)

## Testing Instructions
Essentially, smoke test as many ETK features as you can, after sandboxing a test site and syncing ETK to your sandbox. This is pretty straightforward, so you can also verify I didn't make any typos :P.


apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/page-patterns-plugin.tsx
```diff
@@ -53,8 +54,8 @@ export function PagePatternsPlugin( props: PagePatternsPluginProps ) {
 
 	const { getMeta, postContentBlock } = useSelect( ( select ) => {
 		const getMeta = () =>
-			( select( 'core/editor' ) as CoreEditorPlaceholder ).getEditedPostAttribute( 'meta' );
-		const currentBlocks = ( select( 'core/editor' ) as CoreEditorPlaceholder ).getBlocks();
+			( select( editorStore ) as CoreEditorPlaceholder ).getEditedPostAttribute( 'meta' );
+		const currentBlocks = ( select( 'core/block-editor' ) as CoreEditorPlaceholder ).getBlocks();
```
noahtallen, line 58: Basically, selectors and actions related to blocks were moved to core/block-editor. In the future, we should also do this for core/block-editor

.yarnrc.yml
```diff
@@ -37,6 +37,10 @@ logFilters:
     pattern: "@automattic/wpcom-editing-toolkit@workspace:apps/editing-toolkit provides react (*) with version 17.0.2, which doesn't satisfy what @wordpress/block-editor and some of its descendants request"
   - level: discard
     pattern: "@automattic/wpcom-editing-toolkit@workspace:apps/editing-toolkit provides react-dom (*) with version 17.0.2, which doesn't satisfy what @wordpress/block-editor and some of its descendants request"
+  - level: discard
```
noahtallen, line 40: can't wait until we don't get these any more

apps/editing-toolkit/package.json
```diff
@@ -80,6 +80,7 @@
 		"@wordpress/dependency-extraction-webpack-plugin": "^4.5.0",
 		"@wordpress/dom-ready": "^3.22.0",
 		"@wordpress/edit-post": "^6.19.0",
+		"@wordpress/editor": "^13.7.0",
```
noahtallen, line 84: Explicit dependencies ftw!

yarn.lock
```diff
@@ -8459,6 +8714,65 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@wordpress/components@npm:^23.7.0":
```
noahtallen: uh oh

tyxla:
> Note: typecheck is currently failing, but the relevant package will be removed soon. (https://github.com/Automattic/wp-calypso/pull/75269)

Rebased after landing #75269.

apps/editing-toolkit/package.json
```diff
@@ -126,6 +127,7 @@
 		"@testing-library/jest-dom": "^5.16.2",
 		"@testing-library/react": "^12.1.3",
 		"@types/node": "^18.11.18",
+		"@types/wordpress__editor": "^13.6.0",
```
tyxla, line 130: Note that those types are still not loaded due to a forced noop package resolution here:

https://github.com/Automattic/wp-calypso/blob/8c09604f5f03362894e1ac1a947117192453d13e/package.json#L293

tyxla:
I rebased and deduped, however, it seems like there are still failures. I won't have the chance to take a look at them today.

apps/editing-toolkit/editing-toolkit-plugin/starter-page-templates/page-patterns-plugin.tsx
tyxla: I think we should do that change in a separate PR; I'd say it's too easy to break something that way. I'd suggest keeping references to the previously used store, and then if we change to a different store, that would be done in a separate PR with less noise. WDYT?

noahtallen: That's reasonable 👍

PR #75309, opened 2023-04-04
Checkout: Show helpful error message for renewal URLs with incorrect site

### LINKED ISSUE #75240: Display a helpful message when users are redirected to the purchase management page because they don't have access to a checkout URL
### What

When you visit a checkout URL you don't have access to (for example, if you are logged in as User A but the checkout URL came from a renewal email for User B's subscription) you get redirected to https://wordpress.com/me/purchases instead.

I think the redirect makes sense, but we should also display a message on the screen in this case explaining e.g. that you don't have access to renew the subscription that you were trying to renew and that the reason is probably because it belongs to a different user account.

### Why

This would inform users of why they can't actually renew the subscription they were trying to renew, rather than giving them no feedback (which can cause them to get confused and start trying to buy or renew something else on the wrong account).

Background: https://github.com/Automattic/wp-calypso/issues/74310

### How

_No response_

sirbrillig:
## Proposed Changes

When loading a checkout renewal URL where the site does not belong to the logged-in user, currently the user is redirected to their own `/me/purchases` page without an explanation. This is confusing and can cause problems if the user then tries to renew something they did not intend to.

In this PR we modify checkout so that renewal URLs (that are not gifts or siteless checkouts) with an invalid site (or with no site) will instead reach checkout and display a specific error message.

In addition, this alters checkout so that if it attempts to add renewal products from the URL and no valid renewal products exist (as is the case with an invalid site URL) it will clear the cart. This is because while the error message is being displayed, checkout will be showing the current cart for the `'no-site'` cart key, which could potentially have something in it from some previous operation. If that's the case, we don't want the user to see something in the cart that could imply their renewal URL was successful.

Fixes https://github.com/Automattic/wp-calypso/issues/75240

Screenshot 2023-04-26 at 4 46 36 PM

## Testing Instructions

Automated tests are included but you can test this manually.

- Visit an invalid renewal URL like `/checkout/personal-bundle/renew/12345/example.com`.
- Verify that you see an error message that tells you that the renewal is invalid and that the cart is empty.

- Add a gift purchase to your cart (instructions for gift purchases can be found in D92011-code). The URL will look something like `/checkout/personal-bundle/gift/12345`.
- Verify that checkout loads without an error message and that you see the "Gift" badge next to the product in the cart.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-main.tsx
@@ -179,6 +179,7 @@ export default function CheckoutMain( {
 		productsForCart,
 		isLoading: areCartProductsPreparing,
 		error: cartProductPrepError,
+		addingRenewals,
 	} = usePrepareProductsForCart( {
 		productAliasFromUrl,
 		purchaseId,
@@ -218,6 +219,7 @@ export default function CheckoutMain( {
 		couponCodeFromUrl,
 		applyCoupon,
 		addProductsToCart,
+		addingRenewals,
 	} );
 
 	useRecordCartLoaded( {
--- a/client/my-sites/checkout/composite-checkout/hooks/use-add-products-from-url.ts
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-add-products-from-url.ts
@@ -95,6 +95,7 @@ export default function useAddProductsFromUrl( {
 			// renewals prepared. This is because if a renewal was requested by URL
 			// and was invalid, we don't want them to see what may have been in their
 			// cart previously to avoid the appearance that the URL succeeded.
+			debug( 'clearing the cart due to a renewal request with no products' );
 			cartPromises.push( replaceProductsInCart( [] ) );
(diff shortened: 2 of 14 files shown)


sirbrillig:
AFAIK the only type of no-site renewal URL is Gifting, but I'm asking in Slack to make sure that this is safe for Jetpack also: p1680789433453519-slack-C03V4PXKUGP

sirbrillig:
Akismet URLs also can have no-site renewals so I've just modified this to allow any siteless checkout renewals.

sirbrillig:
While this includes a new error string, IMO we don't really need to wait for it to be translated because it should already be more clear than no error message at all.

pottedmeat (approved):
Code looks good and tests were confirmed.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7966379

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7966379).

PR #75319, opened 2023-04-05
Signup: Attempt to handle redirect after Apple sign in

ebinnion:
There were several reports that using Apple sign in would break flows. The root issue here seems to be that we would default to redirecting the user back to `/start/user` after the social login was complete. While the user would be successfully logged in, this meant that the user was now in the main onboarding flow as opposed to whichever flow the user had entered from.

The biggest change here is to send the user back to the current flow and step that they're on after social signup, from which we then submit the signup step, which then allows us to continue on the flow. This is at least my best understanding.

One other change that is made in this PR is to persist the `redirect_to` query arg. This is useful in the case of the `/start/account/user` step and is used to allow us to redirect within WordPress.com after the user creates an account.

## Proposed Changes

* Support redirect after sign in with Apple.

## Testing Instructions

* Checkout PR
* Download ProxyMan app for Mac
* Enable MapRemote tool and set it up with the following arguments
Screenshot 2023-04-03 at 8 25 05 PM
- Choose a tailored case to work through. This has been tested with Link in Bio, Newsletter, and VideoPress, eCommerce, p2, and a few other flows so far.

## Pre-merge Checklist

### DIFF
--- a/client/blocks/signup-form/social.jsx
+++ b/client/blocks/signup-form/social.jsx
@@ -23,6 +23,7 @@ class SocialSignupForm extends Component {
 		socialServiceResponse: PropTypes.object,
 		disableTosText: PropTypes.bool,
 		flowName: PropTypes.string,
+		redirectToAfterLoginUrl: PropTypes.string,
 	};
 
 	static defaultProps = {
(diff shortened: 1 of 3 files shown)


jgcaruso:
I've tested with the `/setup/videopress/intro` and `/setup/newsletter` flows and it is working well for me. The user gets logged in. I took the flow all the way to checkout (didn't attempt to complete the purchase though).

Not very familiar with the code, but the changes look reasonable to me.

client/blocks/signup-form/social.jsx
```diff
@@ -110,7 +117,7 @@ class SocialSignupForm extends Component {
 							responseHandler={ this.handleGoogleResponse }
 							uxMode={ uxMode }
 							redirectUri={ this.getRedirectUri( 'google' ) }
-							onClick={ () => this.trackSocialSignup( 'google' ) }
+							onClick={ this.trackLoginAndRememberRedirect.bind( null, 'google' ) }
```
zaguiini: Is this change necessary? I thought that inlining an arrow function would bind `this` to the lexical scope 🤔  The same thing for the `onClick` prop below.

zaguiini (approved):
This is working well. It took a while for me to understand that it's not the sign-in process that is broken but rather that the flow was broken after that step 😅 But all good, I managed to test both the `trunk` and this branch's code through the LIB flow and the problem is fixed.

~I wonder if this also happens on Google when we redirect instead of showing the popup...~ It won't, Google signup only runs in redirect mode if we're already in a popup or logging in to Jetpack: 
image

Issue #72910, opened 2023-02-02
Plans Grid 2023: Review architecture/design considerations for mobile, tablet, and desktop grid views


chriskmnds:
#### What

Let's review some of the design choices on the plan comparison and features tables - gather reasoning behind any underlying design considerations, concerns, etc.

Some of my thoughts and observations so far in the comments below:
- [comparison grid](https://github.com/Automattic/wp-calypso/issues/72910#issuecomment-1450211794)
- [features table](https://github.com/Automattic/wp-calypso/issues/72910#issuecomment-1451904062)

chriskmnds:
The comparison grid. 

I believe this carries a bit of unwarranted complexity in places. It took rounds of refactors to get simple changes to the badges and grid borders to get them aligned - yet none of that helped improve the underlying foundation, which is likely the root of various issues. I find the foundation somewhat on the opinionated side not relying on a more standardized grid-like structure to render the table view. I'm sure it's not a simple question, but why couldn't this be a `N x M` CSS grid? 🤔 

Some rough thoughts/observations, from what I remember off the top of my head pretty much:

- The structure is based on a "row" and "cell" relation - whilst rows can act both as containers of cells and individual content rows of their own. This results in overlapping content that's near-impossible to manage without creating coupling across these two forms - extend the min-width of cells in one row to cover the content rendered directly into the other row. 

<img width="500" alt="Screenshot 2023-03-01 at 3 48 06 PM" src="https://user-images.githubusercontent.com/1705499/222157009-1041088b-3d12-44d2-b0d1-2bd37e9d412d.png">

- Inner-grid borders of highlighted "columns" (no such thing as a column here) are `:before` and `:after` pseudo-elements on cells with explicit 140px height to cover and span into the previous rows (so one cell's presumed left border actually covers and renders the borders of the previous two cells).

<img width="500" alt="Screenshot 2023-03-01 at 3 36 39 PM" src="https://user-images.githubusercontent.com/1705499/222157743-54f73a64-3158-4fec-b3f5-6b3031d06b67.png">

- The top highlight badges come in the form of cell overlays that are repositioned several pixels above to hide (and override) the grid borders at the top. This was both difficult to grasp and quite hard to tune. It's relying on pixel-perfect quantities to perfectly align, creating coupling. Again, not sure if all of that is warranted - couldn't badges be a first/top layer in the grid, rendering the top borders when visible?

[...]

PR #75333, opened 2023-04-05
Plans 2023: Notice reskin

chriskmnds:
Closes Automattic/martech#1528

## Proposed Changes

- Updates the Notice bar to more closely match the design in Figma (xkhhUlDowF13dRoXJqlRDM-fi-1_95).
- Uses the Gridicons from previously, so the icons aren't being updated.
- Adds the changes behind a `isReskinned` property that we can enable/test gradually outside of the plans page. Alternatively, we can refactor the base styles directly and enable them across the whole.

### Before

Screenshot 2023-04-06 at 1 23 50 PM

### After

Screenshot 2023-04-06 at 1 24 32 PM

## Testing Instructions

* Go to `/plans/[SITE]` with a site on a paid plan and pro-rated credits available
* Confirm the Notice bar matches the media above
* (maybe) Confirm a Notice bar elsewhere and ensure it renders the old style

## Pre-merge Checklist

### DIFF
--- a/client/components/notice/style.scss
+++ b/client/components/notice/style.scss
@@ -345,6 +345,11 @@ a.notice__action {
 			color: var(--studio-red-40);
 		}
 	}
+	&.is-plain {
+		.notice__icon-wrapper {
+			color: inherit;
+		}
+	}
 
 	// "X" for dismissing a notice
 	.notice__dismiss {
--- a/client/components/notice/style.scss
+++ b/client/components/notice/style.scss
@@ -316,11 +316,15 @@ a.notice__action {
 	color: inherit;
 	border-radius: 4px;
 
+	.notice__content {
+		padding: 13px 0;
+	}
+
 	.notice__icon-wrapper {
 		border-top-left-radius: 4px;
 		border-bottom-left-radius: 4px;
 		background: var(--studio-gray-0);
-		padding: 13px;
+		padding: 16px;
 		align-items: flex-start;
 		width: unset;
 	}
@@ -346,7 +350,7 @@ a.notice__action {
 	.notice__dismiss {
 		display: flex;
 		flex-shrink: 0;
-		padding: 13px;
+		padding: 16px;
 		cursor: pointer;
 
 		.gridicon {
--- a/client/components/notice/style.scss
+++ b/client/components/notice/style.scss
@@ -333,21 +333,37 @@ a.notice__action {
 	&.is-success {
 		.notice__icon-wrapper {
 			color: var(--studio-green-40);
+			background: var(--studio-gray-0);
+			.notice__icon-wrapper-drop {
+				background: var(--color-text-inverted);
+			}
 		}
 	}
 	&.is-warning {
 		.notice__icon-wrapper {
 			color: var(--studio-orange-40);
+			background: var(--studio-gray-0);
+			.notice__icon-wrapper-drop {
+				background: var(--color-text-inverted);
+			}
 		}
 	}
 	&.is-error {
 		.notice__icon-wrapper {
 			color: var(--studio-red-40);
+			background: var(--studio-gray-0);
+			.notice__icon-wrapper-drop {
+				background: var(--color-text-inverted);
+			}
 		}
 	}
 	&.is-plain {
 		.notice__icon-wrapper {
 			color: inherit;
+			background: var(--studio-gray-0);
+			.notice__icon-wrapper-drop {
+				background: var(--color-text-inverted);
+			}
 		}
 	}
 
--- a/client/components/notice/style.scss
+++ b/client/components/notice/style.scss
@@ -310,3 +310,48 @@ a.notice__action {
 		}
 	}
 }
+
+.notice.is-reskinned {
+	background: var(--studio-gray-0);
+	color: inherit;
+	border-radius: 4px;
+
+	.notice__icon-wrapper {
+		border-top-left-radius: 4px;
+		border-bottom-left-radius: 4px;
+		background: var(--studio-gray-0);
+		padding: 13px;
+		align-items: flex-start;
+		width: unset;
+	}
+
+	&.is-info,
+	&.is-success {
+		.notice__icon-wrapper {
+			color: var(--studio-green-40);
+		}
+	}
+	&.is-warning {
+		.notice__icon-wrapper {
+			color: var(--studio-orange-40);
+		}
+	}
+	&.is-error {
+		.notice__icon-wrapper {
+			color: var(--studio-red-40);
+		}
+	}
+
+	// "X" for dismissing a notice
+	.notice__dismiss {
+		display: flex;
+		flex-shrink: 0;
+		padding: 13px;
+		cursor: pointer;
+
+		.gridicon {
+			width: 18px;
+			height: 18px;
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


chriskmnds:
@lucasmendes-design can you pls test this and confirm if close to what's expected? I was also not entirely sure if the `x` button and the icon should be top-aligned, as implemented here.

southp:
@lucasmendes-design pinging you here in case you missed the ping above ^

lucasmendes-design:
Sorry for the delay, folks. I've missed this one :( thanks for reminding me, @southp. 

The styling is good 💯. The only detail I would change is the spacing between the icon and copy. It should be 16px (pink rectangle). 
image

Another detail is the spacing between the header, notice, and plans grid (24px)
image

lucasmendes-design:
> I was also not entirely sure if the x button and the icon should be top-aligned, as implemented here.

It's good 👍🏼

chriskmnds:
Thanks for reviewing! 🙌 Paddings/margins updated to match.

@southp you'll notice that the updated styles are behind a`isReskinned` param that defaults to `false`, so currently the reskin will only apply to the plans. How do you propose we enable this everywhere? leave for case by case, or go for all? I have a feeling it might not sit well with every context e.g. the dev notices will look a little "non-dev", so to speak 😆 

Screenshot 2023-04-28 at 4 27 52 PM

southp:
> You'll notice that the updated styles are behind aisReskinned param that defaults to false, so currently the reskin will only apply to the plans. How do you propose we enable this everywhere? leave for case by case, or go for all? I have a feeling it might not sit well with every context e.g. the dev notices will look a little "non-dev", so to speak 😆

Good question, @chriskmnds . Let's leave it case by case for now and iteratively enable it. Even if we want to go bold enablingly it everywhere, it should be a follow-up PR.

Also, should we lift this into a ready-for-review PR rather than a draft one? :)

chriskmnds:
> Let's leave it case by case for now and iteratively enable it. Even if we want to go bold enablingly it everywhere, it should be a follow-up PR.

Thanks @southp . Sound. will create an issue once we merge. (sounds like a good dev onboarding task - "touch all the files in Calypso") 😆 

> Also, should we lift this into a ready-for-review PR rather than a draft one? :)

👍

client/components/notice/style.scss
```diff
@@ -310,3 +310,57 @@ a.notice__action {
 		}
 	}
 }
+
+.notice.is-reskinned {
+	background: var(--studio-gray-0);
+	color: inherit;
```
chriskmnds, line 316: Not sure about "inherit" here. It works but it might be best to set it explicitly 🤔 
same for icon [below](https://github.com/Automattic/wp-calypso/pull/75333/files#diff-cb8e8eca0c2990d3d8994dcbc991f202c49257238de2438ba6795c383cbdb298R350)

southp (changes requested):
On the tablet resolution, there is a gap between the grid borderline on the right hand side. On the mobile resolution, there is also this gap while the left hand side is attached to the screen. 

Tablet:
image

Mobile:
image

I think we can just do what the current notice component do: 

* On the tablet resolution, the left and right border of the notice aligns to the grid border lines.
* On the mobile resolution, the left and right border of the notice attaches to the screen border.

southp:
@aneeshd16 adding you to this since I won't be available to revisit the code in the following days :)

chriskmnds:
> On the tablet resolution, there is a gap between the grid borderline on the right hand side. On the mobile resolution, there is also this gap while the left hand side is attached to the screen.

This is a little weird, as I can't reproduce it locally at all. I do notice it on the live image and in production though, but that also goes away after reloading. I wasn't expecting any change to the notice width either with these changes. 

This is with Chrome/Firefox:

https://user-images.githubusercontent.com/1705499/236688741-f390f206-d462-4640-bdc4-7f2474541d58.mov

This is in production (that goes away after reloading):

Screenshot 2023-05-08 at 1 05 23 PM

@aneeshd16 can you help me reproduce what @southp is seeing above and if anything to address here? Otheriwse, the PR is ready for review. Gridicons are not updated either, they come from the standard set.

aneeshd16:
Hi @chriskmnds, I could not reproduce the first issue that James mentioned, but I could reproduce the second one:

The notice touches the edges of the screen at <780px screen width.
image

However, this is the same as production.
image

I think applying a 20px margin using the `plans-2023-break-small()` mixin should fix the issue.

chriskmnds:
Thanks @aneeshd16 . 

> I think applying a 20px margin using the plans-2023-break-small() mixin should fix the issue.

Do you think we need to fix that, or just leave as-is in production right now? (so, touching the screen borders) 🤔 

I think this is what @southp suggested in his comment too:

> I think we can just do what the current notice component do:
> 
> - On the tablet resolution, the left and right border of the notice aligns to the grid border lines.
> - On the mobile resolution, the left and right border of the notice attaches to the screen border.

aneeshd16 (approved):
> Do you think we need to fix that, or just leave as-is in production right now? (so, touching the screen borders) 🤔

My vote would go for fixing it, although it can be argued that a full-width notice is intentional, design-wise. I'll defer to @lucasmendes-design for this one. 🙂 

In any case, it can be tackled as a follow-up, and I think this PR is good to go!

ddc22:
I could reproduce the issue raised by James, but it went away after a refresh! I think we have a deep issue related to race conditions when loading SASS files. Upon some investigation I noticed that this bug appears whenever the SASS files wins and overrides the width property (Unrelated to this PR)https://github.com/Automattic/wp-calypso/blob/d2f075e3a9ab2b11c902f0d4eb7334381ccf9ce8/client/my-sites/plans/jetpack-plans/jetpack-complete-page/style.scss#L29-L31

Not entirely sure how to fix it though!

And my OCD picked up this NIT too :D When the notice text is on a single line the text does not seem to be vertically alligned!

image

chriskmnds:
> My vote would go for fixing it, although it can be argued that a full-width notice is intentional, design-wise. I'll defer to @lucasmendes-design for this one. 🙂

Sound. I think this was just not handled when the rest of the grid was updated to not touch the sides (I recall there was an intentional fix for it). I will leave this for another day to get a design check. For reference, this adds the margins around the notice on mobile, if we choose to go with it:

```
.plan-features-main__notice {
	margin: 0 20px;
	width: unset;
	@include plans-section-custom-mobile-breakpoint {
		margin: 0;
		width: 100%;
	}
}
```

> And my OCD picked up this NIT too :D When the notice text is on a single line the text does not seem to be vertically aligned!

It's maybe not NIT. It doesn't look like it's supposed to be vertically aligned, so it covers multi-line comments better. I mentioned this above for design review (which is what you are referring to):

>  I was also not entirely sure if the x button and the icon should be top-aligned, as implemented here.

lucasmendes-design:
> Do you think we need to fix that, or just leave as-is in production right now? (so, touching the screen borders) 🤔

If we could fix that would be amazing, what we have in production is not great so I would appreciate an improvement there :)

chriskmnds:
Updated the top/bottom and side margins. They should be aligned at 24px and 20px resp with/out the plan type selector. 

These are not perfect, and they feel prone to error/misalignment with future updates. I think we are kinda back to the [same grid concerns from earlier](https://github.com/Automattic/wp-calypso/issues/72910#issuecomment-1450211794) whenever touching top/bottom margins.

A couple of states:

### Notice + plan type selector

Screenshot 2023-05-12 at 12 14 05 PM

### Notice only

Screenshot 2023-05-12 at 12 14 23 PM

### Plan type selector only

Screenshot 2023-05-12 at 12 18 16 PM

### Grid only

Screenshot 2023-05-12 at 12 47 38 PM

The last (grid only) now has a ~48px spacing from header. Easy to trim it down to 24px, it just looked a little squashed to me. I believe that will also never be a state after the experiment cleanup.

PR #75340, opened 2023-04-05
One-click checkout: Only bail on email upsell if there are actually cart errors

sirbrillig:
In https://github.com/Automattic/wp-calypso/pull/74555 we modified a number of lines on the one-click checkout page to correct how it sends data to the cart. However, there was a bug introduced: if the user had no saved cards, the Professional Email upsell would silently fail to redirect to checkout after the upsell was added to the cart. This is because we were checking for `cart.messages.errors`, which is an array, and in JavaScript an empty array is truthy. As a result, the code thought that there were errors adding the product to the cart and stopped its processing (assuming those errors would be displayed by the `CartMessages` component). Since there were no actual errors, the user received no information.

## Proposed Changes

In this PR we change the condition to be `cart.messages.errors.length > 0` instead.

I'm going to add some unit tests to prevent this from happening in the future but I'll do those in another PR so that we can get this fix out faster.

## Screenshots

Upsell page:

Screenshot 2023-04-05 at 11 51 45 AM

## Testing Instructions

- First make sure you're using account that has a purchased domain registration on WPcom.
- Second, you'll need to have no saved cards. You can simulate that by applying the following diff on top of these changes:

```diff
--- a/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
@@ -163,7 +163,7 @@ export function useStoredPaymentMethods( {
        } )();

        return {
-               paymentMethods,
+               paymentMethods: [],
                isLoading,
                isDeleting: mutation.isLoading,
                error: errorMessage,
```

- Then visit `/checkout/offer-professional-email/example.com/78805399/example.com`, replacing both `example.com` instances with your registered domain name.
- Click the "Add Professional Email" button.
- Verify that you are redirected to checkout with the email product in your cart.

### DIFF
--- a/client/my-sites/checkout/upsell-nudge/index.tsx
+++ b/client/my-sites/checkout/upsell-nudge/index.tsx
@@ -402,10 +402,7 @@ export class UpsellNudge extends Component< UpsellNudgeProps, UpsellNudgeState >
 				? this.props.paymentMethodsState.paymentMethods[ 0 ]
 				: undefined;
 		if ( this.isEligibleForOneClickUpsell( buttonAction ) && productToAdd && storedCard ) {
-			if ( ! storedCard ) {
-				return;
-			}
-
+			debug( 'accept upsell allows one-click, has a product, and a stored card' );
 			this.setState( {
 				showPurchaseModal: true,
 			} );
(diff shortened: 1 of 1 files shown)


sirbrillig:
Adding tests to guard against the regression this fixes in https://github.com/Automattic/wp-calypso/pull/75348

PR #75358, opened 2023-04-05
Fix untranslated text on /stats/ads/ admin page

d-alleyne:
Related to 539-gh-Automattic/i18n-issues

## Proposed Changes

This is a combination of two problems:
- The 'Ads' navigational item always appeared in English
- Ad chart tabs containing text like 'Ads Served' and 'Average CPM' can appear untranslated.

The first issue was fixed by wrapping the text, and getters with `translate()` calls.

The second issue is more benign. Although the strings are wrapped in `translate()` calls, they can appear untranslated and require page reloading to fix. Using Javascript getters allow the text to appear translated at the time the label is queried by the UI.

## Testing Instructions

1.  Change your user settings to one of the 16 supported languages in the user interface.
2. Navigate to wordpress.com/stats/ads on a Business Account.
3. Verify that the 'Ads' navigational menu item, along with the Chart buttons ('Ads Served', 'Average CPM', etc.) are translated.

**Before**
Before
Multiple reloads later
After multiple page reloads

**After**

After implementing the fix

## Pre-merge Checklist

### DIFF
--- a/client/blocks/stats-navigation/constants.js
+++ b/client/blocks/stats-navigation/constants.js
@@ -37,7 +37,7 @@ const store = {
 	showIntervals: true,
 };
 const wordads = {
-	label: 'Ads',
+	label: translate( 'Ads' ),
 	path: '/stats/ads',
 	showIntervals: true,
 };
@@ -77,7 +77,7 @@ Object.defineProperty( year, 'label', { get: () => translate( 'Years' ) } );
 Object.defineProperty( traffic, 'label', { get: () => translate( 'Traffic' ) } );
 Object.defineProperty( insights, 'label', { get: () => translate( 'Insights' ) } );
 Object.defineProperty( store, 'label', { get: () => translate( 'Store' ) } );
-Object.defineProperty( wordads, 'label', { get: () => 'Ads' } );
+Object.defineProperty( wordads, 'label', { get: () => translate( 'Ads' ) } );
 Object.defineProperty( googleMyBusiness, 'label', {
 	get: () => translate( 'Google My Business' ),
 } );
--- a/client/my-sites/stats/wordads/index.jsx
+++ b/client/my-sites/stats/wordads/index.jsx
@@ -12,6 +12,7 @@ import { connect } from 'react-redux';
 import titlecase from 'to-title-case';
 import illustration404 from 'calypso/assets/images/illustrations/illustration-404.svg';
 import StatsNavigation from 'calypso/blocks/stats-navigation';
+import { navItems } from 'calypso/blocks/stats-navigation/constants';
 import Intervals from 'calypso/blocks/stats-navigation/intervals';
 import DocumentHead from 'calypso/components/data/document-head';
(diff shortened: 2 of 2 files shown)


yuliyan (approved):
Works as expected, LGTM!

CleanShot 2023-04-06 at 15 37 09

jsnmoon:
> Although the strings are wrapped in `translate()` calls, they can appear untranslated and require page reloading to fix. Using Javascript getters allow the text to appear translated at the time the label is queried by the UI. (@d-alleyne)

Can you help me understand how this works? I thought `translate` looked up strings from locale files generated via GlotPress using [static code analysis](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/README.md#strings-only). Are the locale files not available on the first page render? Why does a page reload (or using a getter) fix the issue?

d-alleyne:
@jsnmoon 
> Are the locale files not available on the first page render?

A race condition exists. The `const` assignment of variables may occur before the [locale files](https://github.com/Automattic/wp-calypso/blob/809fe3b8ee3b046fcaf569828dee313eefb053ea/client/lib/i18n-utils/switch-locale.js#L43) have been fully loaded and cached. Using the [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) is a workaround to ensure that the `translate()` call occurs during the rendering of the component. 

See PCYsg-o2s-p2 for more about the Calypso translation loading process.

jsnmoon:
> A race condition exists. The `const` assignment of variables may occur before the [locale files](https://github.com/Automattic/wp-calypso/blob/809fe3b8ee3b046fcaf569828dee313eefb053ea/client/lib/i18n-utils/switch-locale.js#L43) have been fully loaded and cached. Using the [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) is a workaround to ensure that the `translate()` call occurs during the rendering of the component. (@d-alleyne)

Interesting! Thank you for the detailed explanation. I mistakenly thought we awaited for locales to finish loading before booting Calypso, but [that's apparently not the case](https://github.com/Automattic/wp-calypso/blob/584e9db1018fa268eb5742d85b5d55053d008971/client/boot/common.js#L432). (Aside: I suppose the `useTranslate` hook works as well as it does due to React performing multiple re-renders.)

Should we consider invoking `translate` at the topmost scope (like we did in `wordads/index.jsx`) to be an anti-pattern going forward?

d-alleyne:
@jsnmoon 
> I suppose the useTranslate hook works as well as it does due to React performing multiple re-renders

Correct.

> Should we consider invoking translate at the topmost scope (like we did in `wordads/index.jsx`) to be an anti-pattern going forward?

We'd go one step further, and discourage the use of static `translate()` calls _anywhere_ in favour of the `useTranslate()` hook or `this.props.translate()` in [localized components](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-calypso/README.md#localize).

PR #75359, opened 2023-04-05
Media Library: Show Video Upload Progress

roundhill:
Adds an event emitter from the Tus upload script that the media library item listens to to show upload progress.

video-upload-progress-calypso

**To Test**
* Upload a video to the media library, you should see upload progress.
* Other media upload types should behave as before with the indeterminate spinner.

### DIFF
--- a/client/assets/stylesheets/shared/functions/_z-index.scss
+++ b/client/assets/stylesheets/shared/functions/_z-index.scss
@@ -222,6 +222,7 @@ $z-layers: (
 		".media-library__list-item.is-transient .media-library__list-item-figure::after": 10,
 		".media-library__list-item-selected-icon .gridicon": 20,
 		".media-library__list-item-spinner": 20,
+		".media-library__list-item-progress": 20,
 	),
 	".dialog__backdrop": (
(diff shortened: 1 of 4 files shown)


roundhill:
@evilluendas approved the design via slack ;)

jgcaruso (approved):
Works great!

I noticed the same issue mentioned earlier with 2 of the same video at the same time, but it doesn't seem like a huge issue. The individual files do end up as 2 distinct files, so its just progress that seems a bit off... they do seem to complete independently though, so the 2nd upload should eventually be consistent with its actual progress once the 1st completes.

One thing on colours, it seems like in Prod and in `calypso.live` we get a pink border, not blue. I wonder if you can match the progress bar with the border colour
image

roundhill:
> One thing on colours, it seems like in Prod and in `calypso.live` we get a pink border, not blue. I wonder if you can match the progress bar with the border colour 

Good call, I added the `--color-accent` CSS var which adjusts it to the Calypso theme:

Screenshot 2023-04-06 at 3 46 53 PM

PR #75391, opened 2023-04-06
Hosting onboarding: use passwordless mode on signup step

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2123.

## Proposed Changes

Let's not ask for a password when creating an account exclusively on the hosting flow to improve conversions for that step. More info on [the issue](https://github.com/Automattic/dotcom-forge/issues/2123).

## Testing Instructions

- Check out this branch;
- Open `/start/hosting`;
- Check that the signup form includes only an e-mail input;
- Input an e-mail;
- Complete the purchase;
- Verify that you're authorized and the site has been created.

Also, try signing in with the same e-mail, and verify that the signup link arrives in your inbox, and the link successfully logs you in.

image

### DIFF
--- a/client/signup/steps/user/index.jsx
+++ b/client/signup/steps/user/index.jsx
@@ -1,6 +1,6 @@
 import config from '@automattic/calypso-config';
 import { localizeUrl } from '@automattic/i18n-utils';
-import { isNewsletterFlow } from '@automattic/onboarding';
+import { isHostingLPFlow, isNewsletterFlow } from '@automattic/onboarding';
 import { isMobile } from '@automattic/viewport';
 import classNames from 'classnames';
 import { localize } from 'i18n-calypso';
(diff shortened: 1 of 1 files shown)


mpkelly (approved):
Tested and working. I was able to use my email to login via the link that gets emailed. And I confirmed that other flows work as before. LGTM. 

image

simison:
Looking at https://wordpress.com/start/hosting/user

Screenshot 2023-04-07 at 22 51 45

Some notes on UX that could be good follow-ups and probably affect things negatively as-is:

- There's no "login" link available anywhere for those who already know they have an account

- There's no "use a password instead" action. Here is an example from Substack:

    image

- When using email that already has an account, we show an error when we could just welcome them and send them a login link:

    Screenshot 2023-04-07 at 22 49 23

Worth testing how others do passwordless features and comparing if ours is worse or better.

zaguiini:
Thanks @simison, good thing we're scoping it to the hosting flow 😅 Ideally, this flow should be used by non-WPCOM users so the points don't really affect our use case. But if we have the bandwidth we're definitely adding it to Hosting onboarding i1.

PR #75396, opened 2023-04-06
Add additional props in the calypso_signup_step_start event for the with-theme flow

BogdanUngureanu:
Add the theme, theme_type and style_variation props to the calypso_signup_step_start event for the with-theme flow. 

With these new props will get more insights about what themes, their type and which style variation the user selects when their register from the with-theme flow. 

## Proposed Changes

* Add new tracks props to the with-theme flow.

## Testing Instructions

* Use the calypso live link and open an incognito window
* Go to /themes and click on a theme
* Open DevTools and filter the network requests by "t.gif"
* Click on the "Pick this design" button and you'll be redirected to signup page
* Notice that a new `calypso_signup_step_start` event was triggered 
* Make sure that the event contains the `theme`, `theme_type` and `style_variation` props (if you selected a style variation)
* Repeat this step for each type of theme (free, premium, WooCommerce, paid and DotORG). 
* Examples of themes (TT3, Alonso, Astra, Scentina)

## Pre-merge Checklist


mmtr (commented):
I'm getting these errors when testing this branch:

```
Tracks: Event `calypso_signup_start` will be rejected because property name `siteId` does not match /^[a-z_][a-z0-9_]*$/. Please use a compliant property name.
Tracks: Event `calypso_signup_step_start` will be rejected because property name `siteId` does not match /^[a-z_][a-z0-9_]*$/. Please use a compliant property name.
```

They don't happen on production, so I guess they are related to this PR?

client/signup/main.jsx
```diff
@@ -314,9 +315,23 @@ class Signup extends Component {
 
 	getRecordProps() {
 		const { signupDependencies } = this.props;
+		let theme = get( signupDependencies, 'selectedDesign.theme' );
+
+		if ( ! theme && signupDependencies.themeParameter ) {
+			theme = signupDependencies.themeParameter;
+		}
+
+		const deps = this.getCurrentFlowSupportedQueryParams();
```
mmtr: Any reason why this does not execute `this.getDependenciesInQuery` directly? That'd would likely remove the need of converting the deps to snake case. Apparently, `getCurrentFlowSupportedQueryParams` only adds data about the site, but I think it's always empty at this point of the signup, so this data is probably useless?

client/signup/main.jsx
```diff
@@ -314,9 +315,23 @@ class Signup extends Component {
 
 	getRecordProps() {
 		const { signupDependencies } = this.props;
+		let theme = get( signupDependencies, 'selectedDesign.theme' );
+
+		if ( ! theme && signupDependencies.themeParameter ) {
+			theme = signupDependencies.themeParameter;
+		}
```
mmtr: Under which circumstances can `theme` be empty? In my tests is always defined.

client/signup/main.jsx
```diff
@@ -598,9 +613,13 @@ class Signup extends Component {
...
-		const dependenciesInQuery =
-			flows.getFlow( this.props.flowName, this.props.isLoggedIn )?.providesDependenciesInQuery ??
-			[];
+
+		const flow = flows.getFlow( this.props.flowName, this.props.isLoggedIn );
+
+		const requiredDependenciesInQuery = flow?.providesDependenciesInQuery ?? [];
+		const optionalDependenciesInQuery = flow?.optionalDependenciesInQuery ?? [];
+
+		const dependenciesInQuery = [ ...requiredDependenciesInQuery, ...optionalDependenciesInQuery ];
```
mmtr: Not sure if we should be concerned about this, but there is existing code that uses `getDependenciesInQuery` which, according to the var name, seems to assume that it only returns dependencies defined in `providesDependenciesInQuery`:

https://github.com/Automattic/wp-calypso/blob/5c038a94ecccd9ceaab320ccf975fc9256b18e96/client/signup/main.jsx#L157

Would the addition of `optionalDependenciesInQuery` cause any unexpected side effect?

client/state/themes/selectors/get-theme-signup-url.js
```diff
@@ -20,5 +22,47 @@ export function getThemeSignupUrl( state, themeId ) {
...
+ * @param  {Object}  state   Global state tree
+ * @param  {string}  themeId Theme ID
+ * @returns {string}         theme type
+ */
+function getThemeTypeUrlParameter( state, themeId ) {
+	/**
+	 * Is Marketplace theme.
+	 */
+	if ( isExternallyManagedTheme( state, themeId ) ) {
+		return 'managed-externally';
```
mmtr, line 43: Maybe `marketplace` as theme type is clearer?

mmtr (approved):
This works nicely for me! I left a few questions that I'd love to clarify before shipping this but this looks good otherwise.

client/signup/main.jsx
BogdanUngureanu: I'm not sure. Keep in mind that these changes are applied to all flows, not just ours. I thought that I should pass them just in case those props are passed. If their values are null (like in the with-theme flow), the props are omitted anyway.

client/signup/main.jsx
BogdanUngureanu: The method is called for all signup flows, not just the with-theme flow. Also, in the controller, we already check if it's null or not https://github.com/Automattic/wp-calypso/blob/11fcd29e2d573a5e54e7b7022149b343f45a0dd0/client/signup/controller.js#L290

client/state/themes/selectors/get-theme-signup-url.js
BogdanUngureanu: I wanted to use the internal name, for consistency reasons. :)

client/signup/main.jsx
BogdanUngureanu: As for the need to convert them to snake case, I think we still need that because there are flows that define camelCase variables.

client/signup/main.jsx
BogdanUngureanu: Let's play it safe. I'll move the logic in the getRecordProps.

client/signup/main.jsx
mmtr: > I thought that I should pass them just in case those props are passed

I think that, in that case, the flow ideally should define the site data as dependency. Anyway, not a blocker.

mmtr (approved):

PR #75397, opened 2023-04-06
Stepper: Generalise removal of grey progress bar

### LINKED ISSUE #75375: Stepper: Grey progress bar on signup step
[Context](p9Jlb4-6L9-p2#comment-7458)

We fixed it for Newsletter and Link In Bio in [this PR](https://github.com/Automattic/wp-calypso/pull/75039), but it would be better to have a more generalised approach.

image

escapemanuele:
## What
We have CSS that removes the grey progress bar on the signup step while on some of the flows in Stepper. This needs to be generalised to not happen again

## Testing
1. Checkout branch or Live Link
2. Try the LiB (/setup/link-in-bio) or Newsletter flows (/setup/newsletter) while on incognito
3. Check that the grey progress bar on the signup step is gone

Fixes https://github.com/Automattic/wp-calypso/issues/75375

### DIFF
--- a/client/signup/signup-header/index.tsx
+++ b/client/signup/signup-header/index.tsx
@@ -46,13 +46,16 @@ const SignupHeader = ( {
 
 	return (
 		<div className="signup-header" role="banner" aria-label="banner">
-			{ flowProgress && ! shouldShowLoadingScreen && showProgressBar && (
-				<ProgressBar
-					className={ variationName ? variationName : progressBar.flowName }
-					value={ flowProgress.progress }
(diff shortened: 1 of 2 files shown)


renancarvalho (approved):
I could not reproduce this issue, but I could not reproduce on trunk. 🤔 maybe I did something wrong locally ?

It looks good to me in any case.

escapemanuele:
Thanks @renancarvalho, yeah the issue was fixed on trunk but with CSS that would have require maintenance for future flows 😄

PR #75413, opened 2023-04-06
Remove unused type packages

noahtallen:
Related to #74540 

## Proposed Changes

As part of work on Typescript and types with packages, I've been getting into a rabbit hole related to WordPress type definitions. This PR removes three DT packages since these types are provided by the packages themselves now.

## Testing Instructions
TSC


noahtallen:
Hm, here's the error:

```
node_modules/@types/wordpress__data-controls/index.d.ts(8,10): error TS2305: Module '"@wordpress/data"' has no exported member 'Action'.
```

It must be trying to resolve types from the newer wp data package version which it doesn't actually support.

package.json
```diff
@@ -293,6 +290,7 @@
 		"@types/wordpress__editor": "npm:noop-package@1.0.0",
 		"@types/wordpress__notices": "npm:noop-package@1.0.0",
 		"@types/wordpress__rich-text": "npm:noop-package@1.0.0",
+		"@types/wordpress__data-controls/@wordpress/data": "npm:noop-package@1.0.0",
```
noahtallen, line 292: This is ok because `@wordpress/data` is only a dev dependency of the type package. Unfortunately DT doesn't allow dev dependencies, and using `@wordpress/data` for the test file is important. However, the dependency doesn't impact the actual type definitions. (I know this because I had to updated the data-controls package yesterday, after [removing the DT definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65008) for `@wordpress/data`)

package.json
```diff
@@ -166,12 +166,8 @@
 		"@types/uuid": "^8.3.4",
 		"@types/validator": "^13.7.1",
 		"@types/webpack-env": "^1.16.3",
-		"@types/wordpress__api-fetch": "^3.2.4",
-		"@types/wordpress__block-library": "^2.6.1",
```
noahtallen, line 170: block-library isn't used in calypso. While the other packages are, they bundle their own types (for quite a while actually)

package.json
tyxla: Sounds good. I'd love to see all those hacks removed once all packages expose their own types.

tyxla (approved):
Nice cleanup, thanks 🚀

PR #75414, opened 2023-04-06
Fix: update styles for push notification screen

enejb:
Fixes the styling of wordpress.com/login/push page to make the button align better with the rest of the page. 

## Proposed Changes

Before:
Screenshot 2023-04-06 at 1 59 48 PM

After:
Screenshot 2023-04-12 at 5 25 29 PM

## Testing Instructions

In a incognito or logged out state open the calypso live link.
1. Try login in with an account that has 2FA enabled. 
2. Notice the App Notification screen button looks more align then before.

## Pre-merge Checklist

### DIFF
--- a/client/blocks/login/two-factor-authentication/two-factor-actions.scss
+++ b/client/blocks/login/two-factor-authentication/two-factor-actions.scss
@@ -1,12 +1,8 @@
 .two-factor-authentication__actions.card {
 	margin-bottom: 0;
+	padding: 24px 0;
 
-	.button {
-		height: auto;
-		width: 100%;
-
-		&:not(:first-child) {
-			margin-top: 20px;
-		}
+	.button:not(:first-child) {
+		margin-top: 20px;
 	}
 }
--- a/client/blocks/login/two-factor-authentication/two-factor-actions.scss
+++ b/client/blocks/login/two-factor-authentication/two-factor-actions.scss
@@ -1,6 +1,6 @@
 .two-factor-authentication__actions.card {
 	margin-bottom: 0;
-	padding: 24px 0;
+	padding: 0 0 24px;
 
 	.button:not(:first-child) {
 		margin-top: 20px;
--- a/client/blocks/login/two-factor-authentication/two-factor-actions.jsx
+++ b/client/blocks/login/two-factor-authentication/two-factor-actions.jsx
@@ -72,7 +72,7 @@ class TwoFactorActions extends Component {
 				{ this.props.isWoo && ! this.props.isPartnerSignup && (
 					<Divider>{ this.props.translate( 'or' ) }</Divider>
 				) }
-				<Card className="two-factor-authentication__actions">
+				<Card className="two-factor-authentication__actions wp-login__links">
 					{ isSecurityKeyAvailable && (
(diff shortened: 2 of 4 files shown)


client/blocks/login/two-factor-authentication/two-factor-actions.scss
```diff
@@ -1,9 +1,27 @@
 .two-factor-authentication__actions.card {
 	margin-bottom: 0;
+	max-width: 400px;
+	padding: 24px 0;
 
 	.button {
-		height: auto;
 		width: 100%;
+		margin: 0;
```
tyxla: Does it make sense to reuse as much as possible from the `.wp-login__links button` styles? Meaning, literally declaring additional selectors in that same stylesheet:

https://github.com/Automattic/wp-calypso/blob/c29ba2e5d238cc0c2207e56a7636bc987b9cf4b2/client/login/wp-login/style.scss#L61-L63

and

https://github.com/Automattic/wp-calypso/blob/c29ba2e5d238cc0c2207e56a7636bc987b9cf4b2/client/login/wp-login/style.scss#L355-L357

and

https://github.com/Automattic/wp-calypso/blob/c29ba2e5d238cc0c2207e56a7636bc987b9cf4b2/client/login/wp-login/style.scss#L390

That way if they end up changed, they'll change together. Otherwise, they might still get in a weird state if we change it only in one place.

client/blocks/login/two-factor-authentication/two-factor-actions.scss
```diff
@@ -1,9 +1,27 @@
 .two-factor-authentication__actions.card {
 	margin-bottom: 0;
+	max-width: 400px;
```
tyxla: I think it might make sense to have this declared at the same place where we declare it for the login main container:

https://github.com/Automattic/wp-calypso/blob/c29ba2e5d238cc0c2207e56a7636bc987b9cf4b2/client/login/wp-login/style.scss#L33

That way they'll always be equal and changed together.

enejb:
Thanks for the review @tyxla. 
I think I addressed your issue in slightly different way. 
Do you think I should create a different class that apply the styling? across both elements?

tyxla (approved):
Looks good from my perspective 👍 

Would be good to check with design folks before shipping.

kellychoffman:
It was using a wonky style before but due to its size and appearance, it better emphasized itself as the (secondary) desired action. Now its very close in location, size, and appearance to the `Create a new account` button, which isn't of equal importance.

I'd deemphasize `Create new account` by making it a link instead of a button. Like:

Screenshot 2023-04-12 at 4 45 17 PM

enejb:
@kellychoffman  I updated the code so that the new page looks like the screenshot above. 
Do you think this is ready to land now? 

Screenshot 2023-04-12 at 5 25 29 PM

nuriapenya:
I also did a quick test and everything looks great from the design side 💎 

Only a more general question: Is there any reason why are we not showing the `Verification code` input right away on the screen? I think it would be more clear for the users what to do if they see the two options right away:

image
+
Screenshot 2023-04-13 at 09 52 03

(Apologies if it's a dummy question, probably I lack some context) 🙏🏽

Issue #65969, opened 2022-07-26
[Bug]: Annual discount percentage is inconsistent between Plans page and checkout


DavidRothstein:
### Quick summary

When you go through the process of buying a plan, the Plans page shows you an annual purchase discount that is slightly different from the number shown later during checkout (for example, 43% on the Plans page vs. 42% during checkout, for the Premium plan).

**Plans page:**

![plans-page](https://user-images.githubusercontent.com/235183/181011179-07307973-7a12-4dcc-9273-cd493e456d95.png)

**Checkout:**

![checkout](https://user-images.githubusercontent.com/235183/181011177-dc1ab5a6-9ab6-426d-b482-158b7473d8f5.png)

### Steps to reproduce

1. Either sign up for a new site and choose a plan (e.g. Premium), or purchase a plan for an existing site starting at https://wordpress.com/plans.
2. Continue through to checkout.

### What you expected to happen

The quoted percentage discounts to be consistent.

### What actually happened

The quoted percentage discounts were not consistent.

### Reproducibility

Consistent

michaeldcain:
From planning:

We should use `floor()` and round down, since we're technically not providing them 43%.

spraveenitpro:
The premium plan says that it is saving 56% by paying annually:

![LHAOoJ.png](https://user-images.githubusercontent.com/552016/230482364-d672c2fb-8ac5-49bc-b17c-024eaab15d8f.png)

But the annual payment for the premium plan is $216, which means:

![LPqpkx.png](https://user-images.githubusercontent.com/552016/230484041-c39a1199-b649-406c-8d03-04ae188eef2a.png)

This means it is 55% that is saved by paying annually, rounding down from 55.5%

Similarly for personal plan:

![HYA8Tm.png](https://user-images.githubusercontent.com/552016/230486053-b94ecb52-6005-49e7-a1d1-331d3d384214.png)

The annual payment for personal plan being $48 which means:

![it4GtK.png](https://user-images.githubusercontent.com/552016/230486765-953e897a-6452-4f86-9c16-6d3bd898ea41.png)

For Biz plan:

![40kdQ8.png](https://user-images.githubusercontent.com/552016/230487968-bc3468b2-84bb-4da6-8d98-b557252b4dc8.png)

For Business plan it is 37.5% but shown as 38%:

[...]

PR #65990, opened 2022-07-26
Plans Grids: show correct discount rounding

michaeldcain:
Checkout uses `Math.floor()` to show annual (and biannual) discount rates, but everywhere else `Math.round()` to show the same rate. This causes the variant picker in Checkout to show a slightly lower discount rate.

This PR updates all instances to use `Math.floor()` since we're technically not giving the user the full percentage point of discount rate (e.g 42.85% is not actually 43%).

Fixes: #65969

calypso plans
signup plans

**To test:**

To test the plans layout in Calypso:
- on a new site, or site with access to legacy plans (personal, premium, business)
- visit /plans page - it should match the first screenshot above
- note the discount rate for each plan (e.g. "You're saving 42% by paying annually")
- add a plan to your cart and visit Checkout
- verify that the discount rate shown in the variation picker dropdown matches the rate from the plans page

To test the plans layout in Signup (uses the `PlansFeatureComparison` component for Desktop views):
- on desktop, visit /start and proceed through to the plans step
- the Plans step should match the second screenshot above
- note the discount rate for each plan (e.g. "You're saving 42% by paying annually")
- add a plan to your cart and visit Checkout
- verify that the discount rate shown in the variation picker dropdown matches the rate from the plans page

PR #75416, opened 2023-04-06
Display correct saved amount percentage when switching to annual plan

spraveenitpro:
When you go through the process of buying a plan, the Plans page [monthly] shows you an annual purchase discount that is slight more than the actual discount. For example choosing the premium plan will provide a discount of 55.5% but it claims to be 56%:

Same applies to the other plans, so we will floor the discount rate rather than rounding it so it shows as:

Related to #65969

## Proposed Changes

* Display correct saving percentage by using floor rather than rounding.

## Testing Instructions

1. Before applying the patch, visit the plans page [http://calypso.localhost:3000/plans/monthly/siteName] while toggled to the monthly plans: we should see 
2. Note that the amount saved by the customer is informed for premium plan as  `Save 56% by paying annually`, similarly it is also mentioned the saving amounts for other plans.
3. Apply the patch.
4. Notice a change/reduction in discount percentage to this


michaeldcain (approved):
This works as described and LGTM. You'll need to rebase before merging.

Screenshot 2023-04-13 at 9 43 41 AM

As a follow-up, can you check [the Checkout issue Niranjan mentioned](https://github.com/Automattic/wp-calypso/pull/65990#pullrequestreview-1154695683) in the PR for the old plans page and see if we need to address anything in Checkout?

PR #75417, opened 2023-04-06
Shopping Cart: Make the cart blog_id always a number and always exist

sirbrillig:
## Proposed Changes

This PR modifies the shopping cart to specify that the cart's `blog_id`, both in a cart request and a cart response, must always be a number and must always exist (although it can be 0). Previously the types allowed for a `blog_id` which was a string (I don't think that ever happens in a cart response and shouldn't be necessary for a cart request) and sometimes did not exist on the request (this is risky because it would make the cart's `blog_id` default to the current blog ID during the API request, which might not belong to the cart's owner and also probably never happened if TypeScript can be believed since we did not have to change anything other than casting by making it required).

This PR should not actually change any behavior other than to submit `blog_id` to the cart and transactions endpoints as a number rather than a string. Since the shopping cart on the backend always expects `blog_id` to be a number (or null), this should just prevent casting the value inside the endpoint. By requiring a number we will also protect against the possibility of using an arbitrary string somehow which could cause undefined behavior when submitted.

## Testing Instructions

If the tests pass, the main things to try here are:

1. A regular purchase for an existing site. Verify that the purchase succeeds and that a receipt appears under `/me/purchases` with the correct data for the upgrade.
2. A domain-only purchase. You can visit `/start/domain/domain-only` to begin this flow. Verify that the purchase succeeds and that a receipt appears under `/me/purchases` with the correct data for the upgrade.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
@@ -69,13 +69,12 @@ export default async function existingCardProcessor(
 	const formattedTransactionData = createTransactionEndpointRequestPayload( {
 		...transactionData,
 		name: transactionData.name ?? '',
-		siteId: dataForProcessor.siteId ? String( dataForProcessor.siteId ) : undefined,
 		country: contactDetails?.countryCode?.value ?? '',
 		postalCode: getPostalCode( contactDetails ),
 		subdivisionCode: contactDetails?.state?.value,
 		domainDetails,
 		cart: createTransactionEndpointCartFromResponseCart( {
-			siteId: dataForProcessor.siteId ? String( dataForProcessor.siteId ) : undefined,
+			siteId: dataForProcessor.siteId,
 			contactDetails: domainDetails ?? null,
 			responseCart,
 		} ),
--- a/client/my-sites/checkout/composite-checkout/lib/free-purchase-processor.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/free-purchase-processor.ts
@@ -25,7 +25,6 @@ export default async function freePurchaseProcessor(
 	transactionOptions: PaymentProcessorOptions
 ): Promise< PaymentProcessorResponse > {
 	const {
-		siteId,
 		responseCart,
 		includeDomainDetails,
 		includeGSuiteDetails,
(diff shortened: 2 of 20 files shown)


JessBoctor:
LGTM 👍 

I did a functional review and both receipts appeared as expected, I also reviewed the code and it is pretty clear where you removed the option for the siteId to be a string.

client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
```diff
@@ -69,13 +69,12 @@ export default async function existingCardProcessor(
 	const formattedTransactionData = createTransactionEndpointRequestPayload( {
 		...transactionData,
 		name: transactionData.name ?? '',
-		siteId: dataForProcessor.siteId ? String( dataForProcessor.siteId ) : undefined,
```
sirbrillig, line 72: This was never used for anything apparently.

client/my-sites/checkout/composite-checkout/lib/free-purchase-processor.ts
```diff
@@ -52,7 +51,6 @@ export default async function freePurchaseProcessor(
 		{
 			name: '',
 			couponId: responseCart.coupon,
-			siteId: siteId ? String( siteId ) : '',
```
sirbrillig, line 55: This was never used for anything either.

client/my-sites/checkout/composite-checkout/lib/generic-redirect-processor.ts
```diff
@@ -65,7 +64,6 @@ export default async function genericRedirectProcessor(
 			country: contactDetails?.countryCode?.value ?? '',
 			postalCode: getPostalCode( contactDetails ),
 			subdivisionCode: contactDetails?.state?.value,
-			siteId: siteId ? String( siteId ) : '',
```
sirbrillig, line 68: Same here: this was never used.

client/my-sites/checkout/composite-checkout/lib/multi-partner-card-processor.ts
```diff
@@ -201,15 +200,14 @@ async function ebanxCardProcessor(
 		...submitData,
 		couponId: responseCart.coupon,
 		country: submitData.countryCode,
-		siteId: siteId ? String( siteId ) : undefined,
```
sirbrillig, line 204: Same here: this was never used.

PR #75424, opened 2023-04-07
Fix: Filtered plans comparison grid features 

claudiucelfilip:
Related to [#1464](https://github.com/Automattic/martech/issues/1464)

## Proposed Changes

The features for each plan in the Comparison grid found in the Onboarding Flows and the Plans pages are constructed by appending to the previous plan features. If a lower-tier plan is filtered-out, then its features never get inherited by the other plans.

Screenshot 2023-04-07 at 13 32 33

## Testing Instructions

* Checkout and run this branch locally
* Go to `/start/plans` and `/plans/{SITE_ID}`
* Try out multiple screen widths
* The plans grid and comparison table below should be unchanged

* Go to this line and manipulate the `visiblePlans` array to show fewer plans (ex. `visiblePlans.splice( 0, 1 )` )
* The plans grid and comparison table below should only show the remaining plans
* Make sure the correct features are shown in the comparison table

Screenshot 2023-04-07 at 13 53 40

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #75446, opened 2023-04-07
Onboarding: allow hiding free plan on 2023 pricing grid

zaguiini:
On https://github.com/Automattic/wp-calypso/pull/75424, we're fixing the bug that prevented the free plan from being hidden. However, we missed removing the gate behind the `hideFreenPlan` variable. This PR opens the gate since the issue is fixed.

### DIFF
--- a/client/signup/main.jsx
+++ b/client/signup/main.jsx
@@ -3,7 +3,6 @@ import {
 	isDomainRegistration,
 	isDomainTransfer,
 	isDomainMapping,
-	is2023PricingGridActivePage,
 } from '@automattic/calypso-products';
 import { isBlankCanvasDesign } from '@automattic/design-picker';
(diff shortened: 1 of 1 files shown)


claudiucelfilip:
@zaguiini Thanks for catching this! However, this change will remove the Free plan if the user selects `Choose my domain later` on the domains step.
This behavior (to keep the Free plan) has been introduced [here](https://github.com/Automattic/wp-calypso/pull/72929) because of the [missing features issue](https://github.com/Automattic/martech/issues/1464), but it's also an intended behavior, with the main onboarding flow at least.
The situation is being tackled in this thread: p9Jlb4-6HH-p2.

zaguiini:
Oh, I see, @claudiucelfilip, so that's actually intended behavior. Let me close this PR then!

PR #65327, opened 2022-07-06
Checkout: Update PayPal processor to use "pending" page on success

sirbrillig:
#### Proposed Changes

In this PR we try to organize some of the post-checkout redirect behavior by making the PayPal payment method redirect to the "pending" page after checkout completes like we do with Stripe redirect payment methods (eg: 3DS, Bancontact, iDeal). The "pending" page will then send the browser to the final redirect location. Previously the user's browser was redirected immediately after returning from PayPal to the final destination, but this is inconsistent with the other payment methods and means that post-checkout code has to be implemented in two or even three locations.

Along with https://github.com/Automattic/wp-calypso/pull/65273, this will make it so we have a single location inside calypso where we can run functions after a purchase is complete. Currently many parts of the code assume (quite understandably) that the `onPaymentComplete` prop of the `CheckoutProvider` component will run when the payment is complete, but it does not run for redirect payment methods because the transaction hasn't been completed (nor even started) by the time we redirect to the payment partner.

This is part of implementing https://github.com/Automattic/wp-calypso/issues/53356

Requires D83820-code

#### Screenshots

Here's an example of the "pending" page and final page for a Starter Plan:

pending-page

post-checkout

#### Testing Instructions

- Add a plan to your cart and visit checkout using this branch.
- Complete the checkout form and select PayPal as the payment method. 
- Complete the purchase and confirm it at PayPal.
- Verify that you briefly see the "Please wait…" pending page in calypso.
- Verify that you are then redirected to a final post-checkout page (this usually will be something like `/checkout/thank-you/:site/:receiptId` but it depends on a lot of factors) that is not the generic thank-you page (`/checkout/thank-you/:site/pending` or `/checkout/thank-you/:site/unknown`).

PR #53345, opened 2021-06-02
Jetpack Checkout: Set post purchase destination to wp-admin recommendations page.

elliottprogrammer:
This PR updates the post-purchase destination (for non-atomic Jetpack purchases) to the site's wp-admin Jetpack dashboard recommendations assistant.

Previously these purchases were being redirected to the Calypso blue product checklist (guided tour).

Related to: 1164141197617539-as-1200403218057318

#### Changes proposed in this Pull Request

* Renamed the flag `redirectCloudCheckoutToWpAdmin` to `redirectCheckoutToWpAdmin`
* And now use this flag for both Cloud and non-Cloud (Calypso) checkout to determine the post-checkout destination.
* As of this PR, `redirectCheckoutToWpAdmin` is `true` and set to `/wp-admin/admin.php?page=jetpack#/recommendations` for all non-atomic Jetpack purchases originating from **_both_** Calypso checkout and Cloud checkout. (however Cloud checkout is not currently in use at this time anyway, but if/when it is enabled someday, it is also set to the Recommendations assistant page.)

#### Testing instructions

1. Checkout this PR & run `yarn start`
2. Create a JN site and connect it (Or use any self-hosted Jetpack site already connected to your account)
3. Select the site and go to the /plans page: `http://calypso.localhost:3000/plans/:site`
4. Select and purchase any product or plan.
5. Verify that after purchase completed, you are redirected to the site's recommendation assistant.
    ( `/wp-admin/admin.php?page=jetpack#/recommendations` )

**Verify all unit tests pass:**
Run: `yarn test-client client/my-sites/checkout/composite-checkout/test/composite-checkout-thank-you.js`

PR #75471, opened 2023-04-10
Copy site: Fix copy site when using various payment methods

kozer:
Fixes https://github.com/Automattic/dotcom-forge/issues/1958#issue-1621351950

## Proposed Changes

When a user tries to copy a site using payment methods other than credits, the copy site doesn't complete successfully.
This pr fix that.

### Test 1

1.  Make sure you have a test user account in place.
2. Ensure that you have followed the instructions in the fieldguide (search for Testing the Store. I've not pasted the URL, as the script that handles the URLs is not working for the fieldguide for me)
3. Make sure that you visit the secret store and find the credentials with the following name: PayPal Checkout
4. Copy a site.
5. Use PayPal as a payment method, and checkout.
6. Ensure that the copy site process completes successfully.

### Test 2

1.  Make sure you have a test user account in place.
2. Ensure that you have disabled the cookie you set up in the previous test
4. Copy a site using credits.
6. Ensure that the copy site process completes successfully.

## Pre-merge Checklist


client/my-sites/checkout/composite-checkout/hooks/use-get-thank-you-url/index.tsx
```diff
@@ -62,7 +62,12 @@ export default function useGetThankYouUrl( {
 		const url = getThankYouPageUrl( getThankYouPageUrlArguments );
 		debug( 'getThankYouUrl returned', url );
 
-		return url;
+		try {
+			return encodeURIComponent( url );
```
danielbachhuber, line 72: Interesting! I'm surprised this wasn't broken for more flows. What is the full Copy Site thank you URL, and what is it on other flows?

sirbrillig:
Since this hook is just a wrapper used for some post-checkout redirects and not others, I wonder if maybe it would be more clear to escape its output closer to what is causing the bug. Eg: sometimes this url is used as a redirect right away but other times it is encoded into a url query (sometimes two levels deep).

client/my-sites/checkout/composite-checkout/hooks/use-get-thank-you-url/index.tsx
kozer: I searched all other flows inside `/stepper/declarative-flow`. There are no other URLs that redirect to the checkout that have `&` in them. I think this is the reason this issue wasn't apparent till now.

The `redirectTo` copy site URL is `/setup/copy-site/automated-copy?sourceSlug=<site slug>&siteSlug=<site slug>`

kozer:
> Since this hook is just a wrapper used for some post-checkout redirects and not others, I wonder if maybe it would be more clear to escape its output closer to what is causing the bug. Eg: sometimes this url is used as a redirect right away but other times it is encoded into a url query (sometimes two levels deep).

I think the problem occurs when it's used in a URL query. I assume that this is not used when the user uses credits, as the problem doesn't exist there. Can you provide more info on where you believe it's the right place to add this change?

My rational was to add it there to cover all the "PayPal like" flows.

sirbrillig:
Sorry for the delayed response 😅 I was on vacation for the past few weeks.

>  I assume that this is not used when the user uses credits, as the problem doesn't exist there. 

As far as I know, the same functions are used to generate the post-checkout URL if the user uses credits or if they do not use credits. But since you're seeing different behavior, that implies that whatever is causing the bug is happening somewhere where credits are a concern. I don't know off the top of my head where the bug might be but if this fixes it for you I bet you can make incremental changes and follow the code flow to to track it down.

> Can you provide more info on where you believe it's the right place to add this change?

I have several things I'll need to know to help:

- Where in calypso does the copy site functionality activate?
- What happens when this PR is not applied that prevents the copy site functionality from operating?

(Both of those are probably good things to include in the PR description as well.)

kozer:
> As far as I know, the same functions are used to generate the post-checkout URL if the user uses credits or if they do not use credits. But since you're seeing different behavior, that implies that whatever is causing the bug is happening somewhere where credits are a concern. I don't know off the top of my head where the bug might be but if this fixes it for you I bet you can make incremental changes and follow the code flow to to track it down.

@sirbrillig , I did an investigation on why this is happening, and where. 

Those are my findings:

First of all, you are right, everyone uses the `getThankYouPageUrl`.

When we use credits, we end up calling `addUrlToPendingPageRedirect`, which takes the `getThankYouPageUrl` URL and adds it to the checkout redirect URL, using `searchParams.set( 'redirectTo', url )`. This does what it should. It encodes the `redirectTo` url.
When we use a method like PayPal, `onPaymentComplete`, is not called, as when we use credits, however, the same function, `addUrlToPendingPageRedirect` is called, via PayPal processor.

After that, all end up to `/checkout/thank-you/:site/pending/:orderId`. The `redirectTo` URL exists as a param, but in one case is encoded and in the other is not.

It seems that on the PayPal use case, the `redirectTo` param is fetched from the cookies, (stored in `wpcom_signup_complete_destination`) but on the other hand, this cookie doesn't seem to exist using credits, and I strongly believe it has to do with that and the way we populate the context:

`const redirectTo = context.query.redirectTo;`

and that is initiated inside in `client/boot/common.js` , in `setupContextMiddleware`, and more specifically the following lines:

```
// page.js url parsing is broken so we had to disable it with `decodeURLComponents: false`
const parsed = getUrlParts( context.canonicalPath );
const path = parsed.pathname + parsed.search || null;
context.prevPath = path === context.path ? false : path;
context.query = Object.fromEntries( parsed.searchParams.entries() );
```

Finally, all the above end up in the following:
- When a user uses credits, the `redirectTo` param is ended up being encoded in the url. ( ampersand is properly escaped)
- When a user uses PayPal (and the same goes with other payment processors as well), the URL on redirect seems to be just fine.

to replicate please navigate to `client/boot/common.js`
and under `context.query = Object.fromEntries...` try to log the `context.query`. Using PayPal the `redirectTo` is not the one it should, but using credits it is.

The copy site functionality is happening when the user navigates back to `/setup/copy-site/automated-copy` and its constructed at `copy-site.tsx` file

sirbrillig:
Thank you for the research!

Ok, this is starting to make more sense to me. For the following, you might want to refer my post on redirect payment method flows: p4jECJ-1G2-p2 and specifically [this PR](https://github.com/Automattic/wp-calypso/pull/65327) which updated PayPal to operate more like our other flows.

Most payment methods, like credits and standard cards, only have two redirects after checkout: they take the result of `getThankYouPageUrl` and encode it for the pending page, **redirecting** the user to that page. That page then decodes `redirectTo` and **redirects** the user to that URL.

PayPal, however, uses four redirects. First it encodes the result of `getThankYouPageUrl` and adds it to the URL for the pending page in the same way as other payment methods. Then it encodes _that_ URL into the PayPal URL query string and **redirects** the user to PayPal. When PayPal finishes, it **redirects** the user back to a pseudo-endpoint which decodes the second URL and **redirects** the user to the pending page, which finally decodes `redirectTo` and **redirects** the user to that URL.

It sounds like there's a bug where one of the additional two redirects in the PayPal flow is losing the encoding of the final URL. I'm still trying to grok the rest of the information you found but I'm hopeful that there might be a way to write a test to prove this bug and its fix in (perhaps) the [paypal processor test](https://github.com/Automattic/wp-calypso/blob/6f8d219edc00fd63d8f635f6996a8849e55f93c4/client/my-sites/checkout/composite-checkout/test/paypal-express-processor.ts#L303).

sirbrillig:
Would encoding the URL [inside the PayPal processor](https://github.com/Automattic/wp-calypso/blob/6f8d219edc00fd63d8f635f6996a8849e55f93c4/client/my-sites/checkout/composite-checkout/lib/paypal-express-processor.ts#L52) fix the issue instead of this? That feels much safer to me as it's targeted directly to the bug.

sirbrillig:
I did some digging to reproduce the bug and I think I discovered some helpful information.

First of all, this PR does not describe how to copy a site, so I'll start with that (you should not assume your reviewer knows how to do something specific in the app):

* Start with a site that has a Business plan that also has "Hosting Features" enabled (this seems to mean that it's a site that has a `wpstaging.com` URL, which I have but I forget how I got it).
* Visit `/settings/general/example.com` where `example.com` is that site's URL.
* Scroll to the bottom of the settings and press "Copy site".
* Select a domain name.
* This brings you to checkout with a new Business plan, a new domain registration, and a new WooCommerce Subscriptions product in the cart.
* Remember to cancel and delete the domain you purchased after testing!! Domains purchased in test mode are still real domains that we are charged for.

Next, we'll examine what happens.

Checkout loads with a URL like this: `https://wordpress.com/checkout/ahdliaudhiuahdauifiaulfhaiuwf.wordpress.com?redirect_to=%2Fsetup%2Fcopy-site%2Fautomated-copy%3FsourceSlug%3Dahdasdhuahda8odiawd.wpcomstaging.com%26siteSlug%3Dahdliaudhiuahdauifiaulfhaiuwf.wordpress.com&signup=1`

Note that the URL has a `redirect_to` query parameter which contains its own query string with the params `sourceSlug` and `siteSlug`.

When checkout renders, the `redirectTo` prop contains the already-decoded query param: `/setup/copy-site/automated-copy?sourceSlug=ahdasdhuahda8odiawd.wpcomstaging.com&siteSlug=ahdliaudhiuahdauifiaulfhaiuwf.wordpress.com`

When we click to submit checkout with PayPal we send the transactions endpoint (for PayPal this is `https://public-api.wordpress.com/rest/v1.2/me/paypal-express-url`) the `success_url` as `"https://wordpress.com/checkout/thank-you/ahdliaudhiuahdauifiaulfhaiuwf.wordpress.com/pending/:orderId?redirectTo=%2Fsetup%2Fcopy-site%2Fautomated-copy%3FsourceSlug%3Dahdasdhuahda8odiawd.wpcomstaging.com%26siteSlug%3Dahdliaudhiuahdauifiaulfhaiuwf.wordpress.com&receiptId=%3AreceiptId"`

You can see that the `redirectTo` query param has been encoded again.

When we return from PayPal, we are redirected to the pseudo-endpoint with a URL like this:
https://public-api.wordpress.com/rest/v1/me/transactions/paypal-express?order_id=584608&success_url=https%3A%2F%2Fwordpress.com%2Fcheckout%2Fthank-you%2Fahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com%2Fpending%2F%3AorderId%3FredirectTo%3D%252Fsetup%252Fcopy-site%252Fautomated-copy%253FsourceSlug%253Dahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com%2526siteSlug%253Dahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com%26receiptId%3D%253AreceiptId&cancel_url=https%3A%2F%2Fwordpress.com%2Fcheckout%2Fahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com%3Fredirect_to%3D%252Fsetup%252Fcopy-site%252Fautomated-copy%253FsourceSlug%253Dahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com%2526siteSlug%253Dahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com%26signup%3D1&token=EC-XXX&PayerID=XXX

That does a 302 redirect to:
https://wordpress.com/checkout/thank-you/ahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com/pending/584608?redirectTo=/setup/copy-site/automated-copy?sourceSlug=ahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com&siteSlug=ahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com&receiptId=20075772

Note that the `redirectTo` query param is no longer encoded! 

So the problem is happening when the pseudo-endpoint sends us back to calypso. It is decoding the `redirectTo` query param too far. If you examine the pseudo-endpoint URL, you can see the `success_url` of:
https://wordpress.com/checkout/thank-you/ahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com/pending/:orderId?redirectTo=%2Fsetup%2Fcopy-site%2Fautomated-copy%3FsourceSlug%3Dahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com%26siteSlug%3Dahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com&receiptId=%3AreceiptId

This is still encoded correctly. 

The 302 redirect happens inside the PayPal pseudo-endpoint `https://public-api.wordpress.com/rest/v1/me/transactions/paypal-express`. You can look at the code there and see that before it performs its redirect, it calls `urldecode()` on the `success_url` param. Since `redirectTo` is only encoded once at this point, it decodes it along with the rest of the pending page URL, creating an invalid URL that is the source of this bug.

Your solution here is to double-encode the URL that ends up in `redirectTo` so that when it is decoded inside the pseudo-endpoint it goes from `https://wordpress.com/checkout/thank-you/ahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com/pending/:orderId?redirectTo=%252Fsetup%252Fcopy-site%252Fautomated-copy%253FsourceSlug%253Dahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com%2526siteSlug%253Dahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com&receiptId=%3AreceiptId` to `https://wordpress.com/checkout/thank-you/ahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com/pending/:orderId?redirectTo=%2Fsetup%2Fcopy-site%2Fautomated-copy%3FsourceSlug%3Dahdliaudhiuahdauifiaulfhaiuwf.wpcomstaging.com%26siteSlug%3Dahdliaudhiuahdauifiaulfhaiuwflive.wordpress.com&receiptId=:receiptId`.

However, in this PR you're encoding it for every payment method, not just PayPal, and I think that will break something since  payment methods other than PayPal and Stripe redirect methods (eg: Bancontact) do not use the pseudo-endpoints and won't perform the double-decoding.

The first question to ask is: why is the pseudo-endpoint decoding the URL at all? We can see from the blame that this behavior was added in D62649-code. That diff was made because some `redirectTo` URLs include a `#` symbol and [that symbol was getting conflated with the query params](https://github.com/Automattic/wp-calypso/pull/53345). This is a related problem. 

So it seems like we can solve this in one of two ways:

1. We can double-encode the redirect URL inside the payment processor function [as I suggested above](https://github.com/Automattic/wp-calypso/pull/75471#issuecomment-1535295255). This will require testing to make sure that it does not cause any issues, particularly with [Jetpack flows](https://github.com/Automattic/wp-calypso/pull/53345).
2. We can alter the pseudo-endpoint to remove its URL decoding. I wonder if it is no longer required now that the PayPal flow has been updated to use the pending page; while the pseudo-endpoint's URL includes an encoded `success_url`, it appears that it might already be decoded by the time it reaches the redirect and it already includes a triple-encoded `redirectTo`. This will also require testing Jetpack flows but feels more correct.

I'll experiment with number 2 tomorrow if you or someone else doesn't get to it first.

sirbrillig:
D109962-code may fix this issue without any changes to calypso. If you have a chance, please let me know.

sirbrillig:
Closing this since I believe the problem has been fixed on the backend.

PR #75495, opened 2023-04-10
Packages: Remove unused domain utils package

tyxla:
## Proposed Changes

While working on #74399 I realized that the `getDomainSearch()` selector is the only instance where the `@automattic/domain-utils` package is used, but that selector was unused, too (since #75269).

So, this PR removes that selector, and the `@automattic/domain-utils` package altogether.

## Testing Instructions

Verify the removed selector and package are not in use.


noahtallen (approved):
Verified it doesn't appear elsewhere 👍

Issue #74022, opened 2023-03-02
Replace `calypso/lib/use-in-view` with `react-intersection-observer`


danielbachhuber:
`calypso/lib/use-in-view` is buggy, and it's not worth fixing the bugs when we have a well-tested alternative.

See conversation in p1677514081942499-slack-C0347E545HR and p1666718787803859-slack-C0347E545HR

First pass: https://github.com/Automattic/wp-calypso/pull/74021

**Done is:**
* `calypso/lib/use-in-view` is no longer in use, and has been removed from the codebase.

PR #75512, opened 2023-04-10
Activity Log: Restore `react-intersection-observer` usage

danielbachhuber:
Previously https://github.com/Automattic/wp-calypso/pull/69119

## Proposed Changes

Restores use of `react-intersection-observer` in the Activity Log because `calypso/lib/use-in-view` is buggy and it doesn't make sense to spend time reinventing the wheel.

## Testing Instructions

1. Apply the following patch to have the banner appear:

```diff
diff --git a/client/components/activity-card-list/visible-days-limit-upsell/index.tsx b/client/components/activity-card-list/visible-days-limit-upsell/index.tsx
index 03c6b5639a..9712dccd52 100644
--- a/client/components/activity-card-list/visible-days-limit-upsell/index.tsx
+++ b/client/components/activity-card-list/visible-days-limit-upsell/index.tsx
@@ -56,10 +56,6 @@ const VisibleDaysLimitUpsell: React.FC< OwnProps > = ( { cardClassName } ) => {
 
 	const upsellRef = useTrackUpsellView( siteId );
 
-	if ( ! Number.isInteger( visibleDays ) ) {
-		return null;
-	}
-
 	const card =
 		isJetpackCloud() || isEnabled( 'activity-log/v2' ) ? (
 			<ActivityCard className={ cardClassName } activity={ PLACEHOLDER_ACTIVITY } />
diff --git a/client/my-sites/activity/activity-log/index.jsx b/client/my-sites/activity/activity-log/index.jsx
index 2bad1c2a88..5134631420 100644
--- a/client/my-sites/activity/activity-log/index.jsx
+++ b/client/my-sites/activity/activity-log/index.jsx
@@ -547,9 +547,7 @@ class ActivityLog extends Component {
 								)
 							) }
 						</section>
-						{ showVisibleDaysLimitUpsell && (
-							<VisibleDaysLimitUpsell cardClassName="activity-log-item__card" />
-						) }
+						<VisibleDaysLimitUpsell cardClassName="activity-log-item__card" />
 						{ ! hasFullActivityLog && ! isIntroDismissed && <UpgradeBanner siteId={ siteId } /> }
 						<Pagination
 							compact={ isMobile() }

```

image

2. On a WoA Business site, navigate to 'Activity Log'
3. Scroll to the bottom of the page and verify the `calypso_activitylog_visiblelimit_upsell_view` event only appears once.


mpkelly:
I think the problem with this usage for the upsell is the dependency array is missing. The `siteId` needs to be included as a dependency here:

image

Otherwise, this observer will be added once and always record the first value of `siteId` which can be `null`. 

If our hand-rolled `useInView` is buggy, we'd want to replace it everywhere:

image

But it seems ok to me. I can't seem to break the UI, as @danielbachhuber showed in Slack, where some stats don't get rendered. I've tried changing filters, searching, and toggling between grid and list, and it always works flawlessly. I also tried Firefox. I should add that I wasn't able to reproduce an earlier bug in our hand-rolled component which did require fixing - probably because it's more evident on certain hardware/networks.

danielbachhuber:
> I think the problem with this usage for the upsell is the dependency array is missing. The `siteId` needs to be included as a dependency here:

Hm, the code you're referring to was removed in my pull request?

> If our hand-rolled `useInView` is buggy, we'd want to replace it everywhere:

Yep, that's what I'm doing with https://github.com/Automattic/wp-calypso/issues/74022

> I should add that I wasn't able to reproduce an earlier bug in our hand-rolled component which did require fixing - probably because it's more evident on certain hardware/networks.

Yeah, it was an intermittent issue that was difficult to reproduce (and all the more reason to use someone else's well-tested library, instead of finding and fixing these issues on our own).

mpkelly:
>Hm, the code you're referring to was removed in my pull request?

Yeah, but I guess you went with this usage first because it was buggy. I was pointing out the fix was simpler than replacing the library. 

>Yeah, it was an intermittent issue that was difficult to reproduce (and all the more reason to use someone else's well-tested library, instead of finding and fixing these issues on our own).

I think replacing a library in several places should be justified by a reproducible bug.

danielbachhuber:
> I think replacing a library in several places should be justified by a reproducible bug.

@mpkelly Prior to my changes in https://github.com/Automattic/wp-calypso/pull/75591, all of the Tracks events in that PR were firing intermittently.

> I was pointing out the fix was simpler than replacing the library.

Ah, thanks for clarifying 😊

PR #75365, opened 2023-04-06
Focus on the first pattern in the pattern panel

okmttdhr:
Related to https://github.com/Automattic/wp-calypso/issues/74829

## Proposed Changes

* Focus on the first pattern when we open the pattern panel.

## Testing Instructions

- Create a new site and `Continue` until landing on the Design Picker.
- Go to the site assembler by clicking `Start designing` from the bottom.
- Click `Sections` and open the pattern panel, see it focuses on the first element.

https://user-images.githubusercontent.com/5287479/230270462-623bc13f-2da4-43e6-a646-ceced845e0f1.mov

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [x] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #75514, opened 2023-04-10
Pattern Assembler: Replace `calypso/lib/use-in-view` with `react-intersection-observer`

danielbachhuber:
Previously https://github.com/Automattic/wp-calypso/pull/75365

## Proposed Changes

Switches to `react-intersection-observer` in the Pattern Assembler because `calypso/lib/use-in-view` is buggy and it doesn't make sense to spend time reinventing the wheel.

## Testing Instructions

- Click on the Calypso Live link in the first comment.
- Create a new site and `Continue` until landing on the Design Picker.
- Go to the site assembler by clicking `Start designing` from the bottom.
- Click `Sections`, then `Add patterns` and explore patterns by clicking on the categories and scrolling the list of pattens
- Verify that patterns load properly when you scroll the list

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-list-renderer.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-list-renderer.tsx
@@ -1,8 +1,8 @@
 import { PatternRenderer } from '@automattic/block-renderer';
 import { Button } from '@automattic/components';
 import classnames from 'classnames';
-import { useEffect, useState } from 'react';
-import { useInView } from 'calypso/lib/use-in-view';
+import { useEffect, useCallback, useRef } from 'react';
+import { useInView } from 'react-intersection-observer';
 import EmptyPattern from './empty-pattern';
 import { encodePatternId } from './utils';
 import type { Pattern } from './types';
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@okmttdhr @taipeicoder I couldn't figure out how to actually access this code path 🙃

Would one of you be able to review, and fill in the testing instructions when you do? Thanks!

miksansegundo (approved):
LGTM 👍 Thanks @danielbachhuber! 

Tested all pattern lists scrolling up and down.

https://user-images.githubusercontent.com/1881481/231144175-47c40023-3e7f-4d3e-bd4c-8ea71a553482.mov

danielbachhuber:
Thanks @miksansegundo !

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-list-renderer.tsx
```diff
@@ -35,11 +35,21 @@ const PatternListItem = ( {
...
-		threshold: [ 0 ],
+	const ref = useRef< HTMLButtonElement >();
+	const { ref: inViewRef, inView: inViewOnce } = useInView( {
+		triggerOnce: true,
 	} );
 
+	const setRefs = useCallback(
+		( node?: Element | null | undefined ) => {
+			if ( node ) {
+				ref.current;
```
arthur791004, line 46: we have to assign the node to `ref.current`, right?

miksansegundo: I think you are right. I think it should be something like this:

```jsx
const setRefs = useCallback(
  ( node?: Element | null | undefined ) => {
    if ( node ) {
      ref.current = node as HTMLButtonElement;
      inViewRef( node );
    }
  },
  [ inViewRef ]
);
```

miksansegundo: Here is a [PR](https://github.com/Automattic/wp-calypso/pull/75613) for that.

PR #74877, opened 2023-03-24
Clarify the domain registration & transfer policy in the domain management screen.

southp:
Fixes #74521 

## Proposed Changes

This PR updates the explanation text of "Get your domain" dialog in `/domains/manage/`, clarifying that it's annual plan only since a monthly plan is a paid plan yet doesn't have that.

image

Note: instead of "any annual paid plan", I've put in "any annual plan" since I feel "paid" is a bit redundant. an annual plan should imply "paid" without ambiguity.  

## Testing Instructions

Logging-in as a user without any annual plan, go `/domains/manage` and confirm that the copy is updated accordingly.

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #75526, opened 2023-04-11
Signup: Add free plan + paid domain modal

aneeshd16:
Fixes Automattic/martech#1649

## Proposed Changes
The full spec is in Automattic/martech#1649.

#### Domains Step
* Change in the free domain explainer copy:
image

#### Plans Step
* Change in the domain feature copy for paid plans:
image

* Change in the domain feature copy for the free plan. The `.wordpress.com` subdomain is fetched via a new hook `useGetWordPressSubdomain`.
image

* The free plan + paid domain modal when a paid domain has been selected, and the user clicks on the Free plan CTA. This uses the same hook, and the value is cached for 5 minutes, so in most cases, a separate network is not required to fetch the subdomain.
image

## Testing Instructions

1. Go to `/start`.
1. On the domains step, search for some domains. Confirm that the free-domain explainer text (in green) matches the spec.
1. Select a paid domain.
1. On the plans step, confirm that the domain feature for paid plans matches the spec.
1. Confirm that a `wordpress.com` subdomain is displayed under the crossed-out domain name in the feature list of the free plan.
1. Click on the `Start with Free` CTA.
1. Confirm that the modal is shown, and the subdomain matches the one shown on the plans grid.
1. Confirm that the price shown for the Personal plan is correct.
1. Select `Get Personal`, and confirm that you are taken to the checkout page with the selected paid domain and the Personal plan in the cart.
1. Repeat steps 1-8.
1. Select "Continue with Free".
1. Confirm that a free site is created with free subdomain shown on the modal. 

## Pre-merge Checklist


wensco:
Quick note that the free domain credit is only available with paid _annual_ plans.

client/components/domains/domain-product-price/index.jsx
```diff
@@ -68,7 +68,7 @@ class DomainProductPrice extends Component {
 			return domainPriceElement( translate( 'Included in paid plans' ) );
 		}
 
-		const message = translate( '{{span}}Free for the first year{{/span}}', {
+		const message = translate( '{{span}}Free for the first year with paid plans{{/span}}', {
```
southp: To be precise, this should be "annual paid plans". e.g. here is a fix for a similar copy: https://github.com/Automattic/wp-calypso/pull/74877

client/my-sites/plan-features-2023-grid/features.tsx
```diff
@@ -13,29 +44,38 @@ const PlanFeatures2023GridFeatures: React.FC< {
 	return (
 		<>
 			{ features.map( ( currentFeature, featureIndex ) => {
+				const key = `${ currentFeature.getSlug() }-${ featureIndex }`;
+
+				const isFreePlanAndCustomDomainFeature =
```
southp: Non-blocker: this condition means that the each feature no longer binds to one single component using the same logic. By putting this in here, it means that all the feature item will have to go through this computation while most of them doesn't need it.

We don't have to address it now, but we should look for a better way in the future :)

client/my-sites/plan-features-2023-grid/features.tsx
```diff
@@ -1,10 +1,41 @@
...
+	const {
+		data: wordPressSubdomainSuggestion,
+		isLoading,
+		isError,
+	} = useGetWordPressSubdomain( domainName );
+
+	return (
+		<SubdomainSuggestion>
+			<div className="is-domain-name">{ domainName }</div>
+			{ isLoading && <LoadingPlaceHolder className="async-load__placeholder" /> }
```
southp: Non-blocker: it's intriguing to see that this "loading placeholder" has been used this way across calypso, e.g. https://github.com/Automattic/wp-calypso/blob/trunk/Users/southp/a8c/wp-calypso/client/components/async-load/index.jsx#L14 seems to be the origin.

Should we consider to extract it into a component, say `<AsyncLoadPlaceholder />` under `/component/async-load/`?

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
```diff
@@ -0,0 +1,230 @@
...
+						border-radius: 4px;
+						width: 639px;
+					}
+				` }
+			/>
+			<DialogContainer>
+				<Heading>{ translate( 'A paid plan is required for your domain.' ) }</Heading>
+				<SubHeading>
+					{ translate(
+						'Custom domains are only available with a paid plan. And they are free for the first year.'
```
southp: To be very precise, we should emphasize that they are free for the first year with annual paid plans. e.g. `Custom domains are only available with a paid plan. And they are free for the first year with an annual paid plan.` It's super mouthful but I don't have a better idea for now. Also, there is a related issue that I'm outlining above :)

southp (changes requested):
Overall it already works pretty well. Thanks for the great work :) 
A critical issue is that when I toggle to the monthly plans, clicking "Get Personal" in the dialog will crash:
image

I think this also reveals the question of which term of the Personal plan we want to advocate here. Should we advocate according to the picked term, or it should be always be the annual one? What if I've picked the 2y term?

client/components/domains/domain-product-price/index.jsx
aneeshd16: Done!

client/my-sites/plan-features-2023-grid/features.tsx
aneeshd16: > By putting this in here, it means that all the feature item will have to go through this computation while most of them doesn't need it.

Can you please explain this a bit more? Do you mean this computation?
`const isFreePlanAndCustomDomainFeature = currentFeature.getSlug() === FEATURE_CUSTOM_DOMAIN && isFreePlan( planName );`

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
aneeshd16: How about this?
`Custom domains are only available with an annual paid plan. And they are free for the first year.`

client/my-sites/plan-features-2023-grid/features.tsx
aneeshd16: That's a good point, the app has many different loading components, but it is challenging to choose the correct one without documentation. On second thoughts, I don't think we should use this one since it relies on the class name being present, which is not guaranteed. I'll replace this with something else.

aneeshd16:
> A critical issue is that when I toggle to the monthly plans, clicking "Get Personal" in the dialog will crash:

Thanks for the review! I have resolved this issue. As mentioned [here](https://github.com/Automattic/wp-calypso/pull/75526#discussion_r1166765048), I'll also replace the placeholder with something else.

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
southp: An annual paid plan is not required for using the custom domains as primary; just that they are only free for the first year with an annual paid plan. We still need to massage the copy here a bit :)

client/my-sites/plan-features-2023-grid/features.tsx
southp: Yes, it's done inside `features.map()`. Thus, when the anonymous function is being called on each item in `features`, it will be computed even if we know that only the custom domain feature item needs it. 

Fwiw this is where an object-oriented language would solve by subclassing. e.g. each item is a subclass of the parent class `FeatureItem` with an abstract method `renderComponent()`. However, I'm not quite sure what's the idiomatic React counterpart of this concept :)

southp:
> Thanks for the review! I have resolved this issue.

I can confirm that it's not crashing anymore 👍🏼

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
aneeshd16: You're right! I'm changing it to `Custom domains are only available with a paid plan. And they are free for the first year with an annual paid plan.` as per your suggestion, so that the copy is accurate. We can iterate on the wordiness later. 🙂 

`Custom domains are only available with a paid plan.` -> this statement is not entirely true either, since users can buy a paid domain later on in the flow.

southp (approved):
I've tested it according to the following test plan:

For each term, i.e. monthly, yearly, 2-yearly:
* Pick a free domain and then a free or a paid plan -> works as it is.
* Pick a paid domain and then a paid plan -> works as it is.
* Pick a paid domain and then a free plan -> the modal shows
  * Pick the personal plan -> the checkout page has the domain and the annual personal plan in cart.
  * Continue as free -> the processing screen shows and proceeds 

LGTM 🚢

aneeshd16:
Thanks @southp for the review! Since the pricing grid is used in a few different places, I thought it best to be a bit defensive, so I added an extra check to restrict the changes to the `onboarding` flow:
https://github.com/Automattic/wp-calypso/pull/75526/commits/e4c6591abaaed8ec45962346d6d22516fd2470fb

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7931457

chriskmnds (changes requested):
Thanks @aneeshd16 . I am curious if domain suggestions aren't implemented already in the [domain-suggestions data store](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores/src/domain-suggestions). It seems to be resolving an endpoint (same?) and populating the state tree. What functionality is essential here that makes recreating all of the data fetching in a custom hook necessary? 

It would be great if we can follow up here with a P2 post outlining why we chose one over the other (if that's the direction). AFAIC we shouldn't be diverging from things established in the code base without addressing the need to do so in a way to raise awareness. It may well be a valid and necessary divergence, but I think let's establish and communicate the basis for doing so.

aneeshd16:
> What functionality is essential here that makes recreating all of the data fetching in a custom hook necessary?

Hi @chriskmnds, appreciate your review on this. The new hook is a convenience wrapper around the `wpcom.domains().suggestions` function, which is [used currently to fetch suggestions](https://github.com/Automattic/wp-calypso/blob/95a95c9c2825abf379695395792aaff51b0731d6/client/components/domains/register-domain-step/index.jsx#L1049-L1051) on the domains step in signup. Using react-query here allows us to leverage inbuilt functions like caching, loading/error states and globally available state without the boilerplate required to implement this. Apart from these features, the new hook encapsulates the parameters passed to `wpcom.domains().suggestions` so that it is consistent across the two usages in this PR.

> It would be great if we can follow up here with a P2 post outlining why we chose one over the other (if that's the direction).

Anecdotally speaking, I have seen increased usage of `react-query` in Calypso. Still, when I tried to find P2 references for the shift, I could not find one, and a recommended standard has yet to be communicated. I agree with you completely on communicating this, and so to kick off this discussion, I've started a post here: p4TIVU-auD-p2

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { useQueryClient } from 'react-query';
+import { useSelector } from 'react-redux';
+import { getCurrentUserCurrencyCode } from 'calypso/state/currency-code/selectors';
+import usePlanPrices from '../plans/hooks/use-plan-prices';
+import { LoadingPlaceHolder } from './components/loading-placeholder';
+import {
+	USE_GET_WORDPRESS_SUBDOMAIN_QUERY_KEY,
+	useGetWordPressSubdomain,
+} from './hooks/use-get-wordpress-subdomain';
+import type { DomainSuggestion } from '@automattic/data-stores';
```
chriskmnds, line 14: This import is what triggered my curiosity here. 

Are the domain suggestions fetched via the react-query hook something that we can already retrieve via the [getDomainSuggestions](https://github.com/Automattic/wp-calypso/blob/95a95c9c2825abf379695395792aaff51b0731d6/packages/data-stores/src/domain-suggestions/selectors.ts#L26) selector?

client/my-sites/plan-features-2023-grid/hooks/use-get-wordpress-subdomain.ts
```diff
@@ -0,0 +1,31 @@
+import { useQuery } from 'react-query';
+import wpcom from 'calypso/lib/wp';
```
chriskmnds, line 2: How do we foresee porting this into a package that can live outside of Calypso (npm)?

client/my-sites/plan-features-2023-grid/hooks/use-get-wordpress-subdomain.ts
```diff
@@ -0,0 +1,31 @@
...
+			} ),
+		{
+			enabled: !! domainName,
+			select: ( data ) => {
+				return data[ 0 ];
+			},
+			staleTime: 1000 * 60 * 5, // 5 minutes
+		}
+	);
+}
```
chriskmnds: Let's expose this from the domain-suggestions data store, as it is a fairly generic function that serves what looks like the same purpose as the respective selector from there. We can import here and use it instead. WDYT?

chriskmnds (commented):
Thanks, @aneeshd16.  I left some comments/suggestions below following up from previously :-)

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
aneeshd16: We can use the selector, although we would have to write a wrapper around it that would expose error and loading states also.

southp:
Thanks for the decent technical discussion here, @aneeshd16 @chriskmnds. Since this change is important to stop the main stream of Free plan + paid domain site and we've promised to launch it this week (p2-p7DVsv-h8q#comment-45404), let's ship it as soon as the translations are done and do the remaining refactor as a follow-up.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7931457).

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
chriskmnds: > expose error and loading states also.

These should be available via the core selectors that every store gets (`isResolving`/`hasFinishedResolution` should serve for loading states). you can see example usage in the [plans store PR](https://github.com/Automattic/wp-calypso/pull/76024/files#diff-e268185bf1c246017f6796f5f74ff3c0ae49b12e2acf525d8074859449573a19R1236). dunno about error states though.

In this case, you'd have something like (pseudo):
```
useSelect( ( select ) => ( {
    suggestions: select( STORE ).getDomainSuggestions( query ),
    resolved: select( STORE ).hasFinishedResolution( 'getDomainSuggestions', query ),
} ) )
```

client/my-sites/plan-features-2023-grid/hooks/use-get-wordpress-subdomain.ts
chriskmnds: This would be the main thing to refactor next. Although, I'm not sure if it's indeed an issue and not a design "flaw" here. I haven't looked at the code much closer, but the modal right now seems to be rendered [directly in the grid component](https://github.com/Automattic/wp-calypso/pull/75526/files#diff-e268185bf1c246017f6796f5f74ff3c0ae49b12e2acf525d8074859449573a19R330), introducing this and other dependencies (that we are so dearly looking to uncouple from). Why is the modal even a concern to the grid? One solution, in that case, would be for the modal to sit outside the component and be triggered as a side effect here.

The latter is probably how I'd approach this scenario e.g. 

- move `FreePlanPaidDomainDialog` out of the 2023 grid component and into the plans step (or wherever it's relevant)
- pass a callback through to the grid that would trigger the visibility of the modal (e.g. you'd be calling it from [here](https://github.com/Automattic/wp-calypso/pull/75526/files#diff-e268185bf1c246017f6796f5f74ff3c0ae49b12e2acf525d8074859449573a19R667))

I haven't looked at the code closer, but we seem to have gone the opposite direction here (render the dialog inside the grid, and pass whatever props the step needs instead). I'm late in the game, but I would have flagged this earlier if this is indeed the case. :/

This deserves a follow-up (more like an immediate follow-up to confirm - IMO). @southp (per your suggestion):

> Thanks for the decent technical discussion here, @aneeshd16 @chriskmnds. Since this change is important to stop the main stream of Free plan + paid domain site and we've promised to launch it this week (p2-p7DVsv-h8q#comment-45404), let's ship it as soon as the translations are done and do the remaining refactor as a follow-up.

client/my-sites/plan-features-2023-grid/free-plan-paid-domain-dialog.tsx
aneeshd16: TIL! Thanks, will do this when I pick up the refactor.

client/my-sites/plan-features-2023-grid/hooks/use-get-wordpress-subdomain.ts
aneeshd16: Thanks for the feedback, @chriskmnds! 

> How do we foresee porting this into a package that can live outside of Calypso (npm)?

If I'm not wrong, I think this is a Calypso wrapper over the existing wpcom package: https://www.npmjs.com/package/wpcom

> Why is the modal even a concern to the grid?

The idea was to make this flow work, irrespective of where the 2023 grid is used. While currently, this is enabled only for the `onboarding` flow, it could also be enabled for other flows which use the domains+plans step. One example is the Stepper-based newsletter flow. (`/setup/newsletter`). My line of thought was that the plans grid knows how to display the free domain suggestion and the modal. However, it does not know how to change the selected domain from paid -> free, and therefore the `replacePaidDomainWithFreeDomain` prop needs to be passed down from the parent.

>  One solution, in that case, would be for the modal to sit outside the component and be triggered as a side effect here.

I can also see your point, and it makes sense to me. In fact, I suggested something similar [here](https://github.com/Automattic/wp-calypso/pull/76100#discussion_r1179068707). 😄 The only thing doubtful was that we would also need to render the modal in the Stepper flows. (but that might not be that big an issue 🤷 )

TBH, I started working on this before our discussions on decoupling the grid, so I was in a different frame of mind, and my approach was to add new changes contained within the "plans 2023 grid" boundary.

> This deserves a follow-up (more like an immediate follow-up to confirm - IMO)

Since we're committing to data stores, I can move the hook to the domain suggestions store as an immediate follow-up. Please let me know what you think about my reasoning above, and based on that, we can refactor the modal as well.

aneeshd16:
Reported E2E failures here: p1682667641426689-slack-C02DQP0FP

EDIT: Test has been muted.

client/my-sites/plan-features-2023-grid/hooks/use-get-wordpress-subdomain.ts
southp: I can see how it is structurally-weird that a modal component is embedded inside the grid component. However, this behavior has to be consistent across all the flows that can incur Free plan + paid domain combinations. I'm not sure what's the best option here, but the bottomline is that the call sites that using the pricing grid component won't have to bind the modal and the pricing grid every time :)

chriskmnds: Hi @aneeshd16 . Thanks for the follow-up. I replied in Slack, will just paste it here:

> I haven't looked at the code closer yet, but my instinct tells me the modal should not be rendered within the grid. a design principle i'd follow is for such things that are not the immediate concern of rendering a grid, to happen as side effects. this is one such case - where "signup" (whether stepper or start) needs to direct user into some sort of domain suggestion. importing a store and querying an endpoint is one more side-effect that we can "release" the grid from
> 
> within calypso, the grid is imported into PlanFeaturesMain and that's what stepper and start import and use. I'd move the modal closer to there and pass the callback into it. The less the grid knows about context the easier it will be to manage.
> 
> I'd need to look at the code closer if the above don't make sense offhand :slightly_smiling_face:

> stuff under my-sites/plan-features-2023-grid are what we want to pull out into a package. and the "dumber" these are (as close to just rendering a grid), the easier, and better. most everything else are side effects that we'd want to detach.
> 
> even a question at some point will be "should the grid even need network access" - but I feel that's for later; for now, probably best to assume it will, for plans (as that part needs to be fleshed out for reuse anyhow)
> 

If you feel my thinking/instinct is "incorrect", then happy to look at the code closer and understand why we need to keep that modal within the "grid".

client/lib/plans/features-list.tsx
```diff
@@ -610,7 +610,7 @@ export const FEATURES_LIST: FeatureList = {
 		getSlug: () => FEATURE_CUSTOM_DOMAIN,
 		getTitle: ( domainName?: string ) => {
 			if ( domainName ) {
-				return i18n.translate( 'The domain %(domainName)s is free for the first year', {
+				return i18n.translate( '%(domainName)s is included', {
```
ddc22, line 613: Hi @aneeshd16 just a random thought regarding this regarding the UX here. This feature will show the same `is included` even when a domain is mapped. Shouldn't this be `mapping` is included in the instance a domain is being mapped?

image

PR #75539, opened 2023-04-11
LeanDomainSearch: customize domain step

thedebian:
Related to #

## Proposed Changes

This PR adds customizations to the domain step, displaying the selected domain, for LeanDomainSearch landing.
* 1663-gh-Automattic/greenhouse

## Testing Instructions
* Apply PR
* `yarn start`
* Go to `/start/domain/domain-only?ref=leandomainsearch&new=PierreSiteRandom.com`
* ✅ You should see the domain in the domain step

### DIFF
--- a/client/signup/steps/site-or-domain/index.jsx
+++ b/client/signup/steps/site-or-domain/index.jsx
@@ -38,18 +38,19 @@ class SiteOrDomain extends Component {
 
 	isLeanDomainSearch() {
 		const { signupDependencies } = this.props;
-		return (
-			'leandomainsearch' === signupDependencies?.refParameter
-		);
+		return 'leandomainsearch' === signupDependencies?.refParameter;
 	}
 
 	getChoices() {
 		const { translate, isReskinned, isLoggedIn, siteCount } = this.props;
 
 		const domainName = this.getDomainName();
-		const buyADomainTitle = this.isLeanDomainSearch()
-			? translate( 'Just buy %s', { args: [ domainName ] } )
-			: translate( 'Just buy a domain' );
+		let buyADomainTitle = translate( 'Just buy a domain' );
+
+		if ( this.isLeanDomainSearch() && domainName ) {
+			// translators: %s is a domain name
+			buyADomainTitle = translate( 'Just buy %s', { args: [ domainName ] } );
+		}
 
 		const choices = [];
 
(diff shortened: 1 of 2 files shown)


client/signup/steps/site-or-domain/index.jsx
```diff
@@ -36,15 +36,29 @@ class SiteOrDomain extends Component {
 		return isValidDomain && domain;
 	}
 
+	isLeanDomainSearch() {
+		const { signupDependencies } = this.props;
+		return (
+			!! signupDependencies &&
+			!! signupDependencies.refParameter &&
+			'leandomainsearch' === signupDependencies.refParameter
```
jgcaruso: Not a blocker: This could be simplified using optional chaining.
```suggestion
			'leandomainsearch' === signupDependencies?.refParameter
```

client/signup/steps/site-or-domain/index.jsx
```diff
@@ -36,15 +36,29 @@ class SiteOrDomain extends Component {
...
+		);
+	}
+
 	getChoices() {
 		const { translate, isReskinned, isLoggedIn, siteCount } = this.props;
 
+		const domainName = this.getDomainName();
+		const buyADomainTitle = this.isLeanDomainSearch()
+			? translate( 'Just buy %s', { args: [ domainName ] } )
```
jgcaruso: Might be helpful to add a `// translators: %s is the domain name` comment here to specify what the string is, since "Just buy" doesn't have much context.

I'm just not sure where it needs to be added since the `translate` call is in a ternary. There is an example here, I'm assuming it is correctly formatted but 🤷 https://github.com/Automattic/wp-calypso/blob/c80dfe9149491db66b490cdc71d364c85cf6747c/client/login/wp-login/login-links.jsx#L154-L158

jgcaruso (commented):
Leaving a comment instead of requesting changes since I'll be AFK and don't want to be a blocker.

The page eventually rendered as expected, but initially I saw the word `false` in the place of the domain placeholders instead of the default text being rendered. Eventually the page re-rendered and it looked correct. It could have just been a slow initial render before the container warmed up though, I wasn't able to reproduce it (screenshot was manually made for demonstration only, not from when I reproduced it).
image


image

Not sure what can be done about long domains, but breaking `.com` onto 2 lines doesn't look great. Not sure what the solution is, other than maybe moving the domain out into the very large whitespace above all the text instead of inserting it into that sentence?

image


Other than that, it worked well and I was able to go all the way to checkout, so the flow seems to still work as expected. Removing the `ref=leandomainsearch` from the query string renders the page as expected.

roundhill:
I saw the same thing as @jgcaruso with the `false` showing, I guess there should be a check in the code if the domain doesn't exist yet to show the old `Choose how to use your domain` and `Just buy a domain` text if so?

roundhill (approved):
🚀

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7922119

Hi @thedebian, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `Just buy %s`
- `Choose how to use %s`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7922119).

PR #75547, opened 2023-04-11
Hosting onboarding flow: use new plans grid

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2126.

## Proposed Changes

This PR enables the new plan grid for the `/start/hosting` flow. It narrows the grid a bit, because there are only two plans and it got too wide.

It also hides the "Enterprise" plan on demand. Do you think it's OK to do that? Should we present the Enterprise plan even though it bails the hosting flow? @vindl would like your input here.

@claudiucelfilip is there any better way of narrowing the grid?

image

## Testing Instructions

1. Clone this branch;
2. Go to `/start/hosting`;
3. Verify that the new grid is displayed according to the screenshot;
4. Verify that the plan comparison still stretches to the full width of the page.

### DIFF
--- a/client/my-sites/plan-features-2023-grid/style.scss
+++ b/client/my-sites/plan-features-2023-grid/style.scss
@@ -1041,3 +1041,11 @@ body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup_
 .formatted-header {
 	margin-bottom: 0;
 }
+
+
+/* Hosting onboarding flow */
+.is-hosting div.plan-features-2023-grid__content {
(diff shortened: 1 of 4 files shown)


mpkelly (approved):
Tested and working. Looks good at all breakpoints.

vindl:
> It also hides the "Enterprise" plan on demand. Do you think it's OK to do that? Should we present the Enterprise plan even though it bails the hosting flow? @vindl would like your input here.

@zaguiini They can pick the Enterprise plan on the LP, so there is no need to show it here again IMO.

PR #75554, opened 2023-04-11
Update Reader tag header to use MasterbarLoggedOut header when logged…

eoigal:
As part of the Public tags pages, we want to use the `MasterbarLoggedOut` header component when viewing a public tag page while logged out.

At the moment we are showing the blue `UniversalNavbarHeader` header component.

This PR checks to see if we are viewing a tag page and renders the `MasterbarLoggedOut` header component instead.

**Before:**
Screenshot 2023-04-11 at 14 20 45

**After:**
Screenshot 2023-04-11 at 14 19 45

Ref: pe7F0s-IS-p2


davemart-in (approved):
LGTM!

PR #75562, opened 2023-04-11
Remove reader tag image from header

eoigal:
We are restyling the tag pages.

As part of this we are removing the tag image from the header for a simpler layout.

This PR removes the image from the tag header and adds the h1 tag for the title and a follow button.

**Before:**
Screenshot 2023-04-11 at 15 10 51

**After:**
Screenshot 2023-04-11 at 15 08 47

### DIFF
--- a/client/reader/tag-stream/header.jsx
+++ b/client/reader/tag-stream/header.jsx
@@ -1,24 +1,11 @@
-import { Gridicon } from '@automattic/components';
 import classnames from 'classnames';
 import { localize } from 'i18n-calypso';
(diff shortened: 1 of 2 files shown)


davemart-in (approved):
LGTM! Ship it!

bluefuton:
Nice improvement! 👍 

If the change is permanent, it could well be worth cleaning up the associated code as well - the `QueryReaderTagImages` data component and all the Redux plumbing associated with tag images. Probably also worth removing the tag images API endpoint, unless it's used in the mobile apps.

PR #74210, opened 2023-03-08
Virtual themes i2: Show description and Pattern Assembler CTA

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/1908
Requires D103960-code

## Proposed Changes

Changes the `designPreview` step for pattern-based virtual themes to show a theme description that sets better expectations for users about which theme will actually be activated.

It also displays a CTA to start the Pattern Assembler.

Screenshot 2023-03-08 at 13 33 03

## Testing Instructions

- Use the Calypso live link below
- Go to `/setup/site-setup/designSetup?siteSlug=<SITE_SLUG>&flags=virtual-themes/onboarding`
- Select a pattern-based virtual theme
- Make sure you see a design description in the sidebar. Make sure it's clear and doesn't have typos.
  - If you don't see it, you might need to clear the `useQuery` cache by applying this change and running the PR locally:
```diff
diff --git a/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts b/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
index 6a056f8777..e895422ef3 100644
--- a/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
+++ b/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
@@ -55,7 +55,7 @@ export function useStarterDesignsQuery(
 	queryParams: StarterDesignsQueryParams,
 	{ select, shouldLimitGlobalStyles, ...queryOptions }: Options = {}
 ): UseQueryResult< StarterDesigns > {
-	return useQuery( [ 'starter-designs', queryParams ], () => fetchStarterDesigns( queryParams ), {
+	return useQuery( [ 'starter-designs*', queryParams ], () => fetchStarterDesigns( queryParams ), {
 		select: ( response: StarterDesignsResponse ) => {
 			const allDesigns = {
 				generated: {
```
- Make sure you see a CTA to open the Pattern Assembler
- Make sure neither the description nor the CTA are visible on smaller screens
- Make sure the CTA starts the Pattern Assembler

PR #75567, opened 2023-04-11
Virtual themes: Remove description

mmtr:
## Proposed Changes

In https://github.com/Automattic/wp-calypso/pull/74210, we started showing a description for virtual themes in order to give users an expectation about what theme will be activated in the site.

This is now clearly visible in the design title now that virtual themes include the theme name (D107549-code), so we no longer need this description.

Before | After
--- | ---
Screenshot 2023-04-11 at 17 23 57 | Screenshot 2023-04-11 at 17 24 34

This also makes the preview screen more consistent since now all themes without style variations (virtual and non-virtual) won't display the description sidebar.

Someone could argue that it'd be nice to keep the Pattern Assembler CTA but it won't have much value once we can start sending users to PA to further customize assembled designs.

## Testing Instructions

- Select a virtual design.
- Make sure there is no sidebar.


BogdanUngureanu (approved):
LGTM! I haven't noticed any regressions after removing this code.

There are some failing tests (probably unrelated), but other than that, we should ship it 🚢

PR #75569, opened 2023-04-11
Virtual themes: Remove feature flag

mmtr:
## Proposed Changes

Removes the `virtual-themes/onboarding` feature flag since Virtual Themes have been already released to all environments and have been running for a few days.


client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -140,7 +140,7 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow } ) => {
 			intent,
 			seed: siteSlugOrId || undefined,
 			_locale: locale,
-			include_pattern_virtual_designs: isEnabled( 'virtual-themes/onboarding' ),
+			include_pattern_virtual_designs: true,
```
BogdanUngureanu, line 143: Since we no longer need this param, shouldn't we remove it from the API endpoint and discard it entirely from here? :)

mmtr: I think we still need it, since there are other places currently fetching data from the `starter-designs` endpoint which might not expect virtual themes:

https://github.com/Automattic/wp-calypso/blob/90c79fa6313c44dbce94acca5b385490e42892fc/packages/design-carousel/src/index.tsx#L24-L28

BogdanUngureanu: Hmm... Good point!

BogdanUngureanu (approved):
LGTM! 🚢

Issue #75504, opened 2023-04-10
Tracking Task: Moving Launchpad Logic to Backend


edanzer:
## Goal

Refactor the launchpad checklist API and migrate checklist definitions into backend. Launchpad should now fetch checklists via backend endpoint, rather than have those hard coded in calypso. This will make retrieving checklist info platform agnostic.

## All Remaining Work
- [x] https://github.com/Automattic/wp-calypso/pull/76644

**Deleting unused code**
- [x] Dependency: Once https://github.com/Automattic/wp-calypso/issues/76058 is done and **deployed to atomic**
- [x] Dependency: Also before deleting these, we should double check old launchpad methods aren't being used anywhere.
  - [x] https://github.com/Automattic/wp-calypso/issues/76473
  - [x] https://github.com/Automattic/wp-calypso/issues/76059
- [ ] Dependency: Once the **Frontend Batch** above is updated to use data-stores/useLaunchpad 
  - [ ] https://github.com/Automattic/wp-calypso/issues/76364
  - [ ] https://github.com/Automattic/wp-calypso/issues/76368
  - [x] https://github.com/Automattic/wp-calypso/issues/76597

**Misc follow ups**
- [ ] https://github.com/Automattic/wp-calypso/issues/76060
- [x] Remove unnecessary `require` of the launchpad feature in the jetpack-mu-wpcom checklist API endpoint ( it's already being `required` in the index https://github.com/Automattic/jetpack/pull/30856
- [x] Ensure that in the main launchpad endpoint, the defintion and validation of possible args is generated from the checklist registry rather than manually. See https://github.com/Automattic/jetpack/pull/30227#discussion_r1179291097
- [x] https://github.com/Automattic/wp-calypso/issues/76056
- [ ] Figure out how to get the _doing_it_wrong checklist registry error to display in the error log (see https://github.com/Automattic/jetpack/pull/30211#discussion_r1173998786) 
- [ ] https://github.com/Automattic/jetpack/pull/30369
 
## Completed Since First Milestone
- [x] https://github.com/Automattic/wp-calypso/issues/76057
- [x] https://github.com/Automattic/wp-calypso/issues/75846
- [x] https://github.com/Automattic/jetpack/pull/30211
- [x] https://github.com/Automattic/wp-calypso/pull/76164
- [x] https://github.com/Automattic/wp-calypso/pull/76154
- [x] https://github.com/Automattic/wp-calypso/pull/76510
[...]

PR #75575, opened 2023-04-11
Help Center: Update Launchpad icon to use real data

edanzer:
### Proposed Changes

**This PR is a draft. It depends on the backend Jetpack PR below, which is still in development. Because the backend endpoint is in development and may return incomplete or unpredictable data, this PR may not render the launchpad icon with the correct progress level for a given site. It simply uses whatever data is returned from the endpoint.**

Related Issue: https://github.com/Automattic/wp-calypso/issues/75504
Requires Backend Jetpack PR: https://github.com/Automattic/jetpack/pull/30023

Before this PR the Launchpad icon in the help center was using hard coded placeholder numbers to determine how much of the progress bar was completed (1 out of 4 steps or 25%). This PR use our new /launchpad/checklist endpoint to fetch the site checklist render the launchpad icon based on actual completed / total steps. 

launchpad icon

### Testing Instructions

1. You'll need to set up the required backend diff. You can do that by running `bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-checklist-refactor-scaffolding` on your sandbox. Also sandbox the _url of the test site_ you'll be working with
2. Checkout this branch and run yarn and yarn start. This should build the help center package fresh for testing in addition to starting your local dev environment. Note that the relevant changes here are in a package. I sometimes needed to run `yarn build-packages` to force the help center package to rebuild with latest changes. 
3. You will need to create or test with a Launchpad enabled site. 
4. Go to http://calypso.localhost:3000/post/YOURSITESLUG
4. TEST: Click on the help / question mark icon in the upper right icon. You should see the Launchpad icon. Historically, this has always rendered with the progress bar at 25% complete. It should now be dynamically based on the total and completed site steps. See screenshot above.

### DIFF
--- a/packages/help-center/src/components/help-center-launchpad.tsx
+++ b/packages/help-center/src/components/help-center-launchpad.tsx
@@ -44,7 +44,7 @@ export const HelpCenterLaunchpad = () => {
 		siteSlug = window?.location?.host;
 	}
 
-	const { data } = useLaunchpadChecklist( siteSlug );
+	const { data } = useLaunchpadChecklist( siteSlug, siteIntent || '' );
 	const totalLaunchpadSteps = data?.checklist?.length || 4;
 	const completeLaunchpadSteps =
 		data?.checklist?.filter( ( checklistItem ) => checklistItem.completed ).length || 1;
--- a/packages/help-center/src/components/help-center-launchpad.tsx
+++ b/packages/help-center/src/components/help-center-launchpad.tsx
@@ -8,6 +8,7 @@ import { chevronRight, Icon } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { useSelector } from 'react-redux';
 import { getSectionName, getSelectedSiteId } from 'calypso/state/ui/selectors';
+import { useLaunchpadChecklist } from '../hooks/use-launchpad';
 import { SITE_STORE } from '../stores';
 import type { SiteSelect } from '@automattic/data-stores';
 
@@ -43,6 +44,11 @@ export const HelpCenterLaunchpad = () => {
 		siteSlug = window?.location?.host;
 	}
 
+	const { data } = useLaunchpadChecklist( siteSlug );
+	const totalLaunchpadSteps = data?.checklist?.length || 4;
+	const completeLaunchpadSteps =
+		data?.checklist?.filter( ( checklistItem ) => checklistItem.completed ).length || 1;
+
 	const launchpadURL = `${ getEnvironmentHostname() }/setup/${ siteIntent }/launchpad?siteSlug=${ siteSlug }`;
 	const sectionName = useSelector( ( state ) => getSectionName( state ) );
 	const handleLaunchpadHelpLinkClick = () => {
@@ -69,7 +75,11 @@ export const HelpCenterLaunchpad = () => {
 					handleLaunchpadHelpLinkClick();
 				} }
 			>
-				<CircularProgressBar size={ 32 } currentStep={ 1 } numberOfSteps={ 4 } />
+				<CircularProgressBar
+					size={ 32 }
+					currentStep={ completeLaunchpadSteps }
+					numberOfSteps={ totalLaunchpadSteps }
+				/>
 				<span className="inline-help-launchpad-link-text">
 					{ __( 'Continue setting up your site with these next steps.' ) }
 				</span>
--- a/packages/help-center/src/hooks/use-launchpad.tsx
+++ b/packages/help-center/src/hooks/use-launchpad.tsx
@@ -23,24 +23,27 @@ interface LaunchpadTasks {
 	checklist: LaunchpadTask[];
 }
 
-export const fetchLaunchpadChecklist = ( siteSlug: string | null ): Promise< LaunchpadTasks > => {
+export const fetchLaunchpadChecklist = (
+	siteSlug: string | null,
+	siteIntent: string
+): Promise< LaunchpadTasks > => {
 	const slug = encodeURIComponent( siteSlug as string );
 
 	return canAccessWpcomApis()
 		? wpcomRequest( {
-				path: `sites/${ slug }/launchpad/checklist?checklist_slug=newsletter`,
+				path: `sites/${ slug }/launchpad/checklist?checklist_slug=${ siteIntent }`,
 				apiNamespace: 'wpcom/v2/',
 				apiVersion: '2',
 		  } )
 		: apiFetch( {
 				global: true,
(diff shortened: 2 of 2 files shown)


packages/help-center/src/hooks/use-launchpad.tsx
```diff
@@ -0,0 +1,52 @@
...
+		? wpcomRequest( {
+				path: `sites/${ slug }/launchpad/checklist?checklist_slug=${ siteIntent }`,
+				apiNamespace: 'wpcom/v2/',
+				apiVersion: '2',
+		  } )
+		: apiFetch( {
+				global: true,
+				path: `sites/${ slug }/launchpad/checklist?checklist_slug=${ siteIntent }`,
+		  } as APIFetchOptions );
+};
```
edanzer, line 42: The method above uses siteIntent as the checklist_slug. That may work for now. Question 1: should we use anything else? Question 2: If siteIntent is an empty string here, our endpoint returns an empty array, which works fine with how I digest and use the data. But still worth asking what we want to return in case of no siteIntent or checklist_slug?

packages/help-center/src/hooks/use-launchpad.tsx
```diff
@@ -0,0 +1,52 @@
+import apiFetch from '@wordpress/api-fetch';
```
edanzer, line 1: The approach I take in this hook - including the use of react-query, and the conditional use of wpcomRequest and apiFetch is following existing approaches/files in help center. See this file for example: 

https://github.com/Automattic/wp-calypso/blob/trunk/packages/help-center/src/hooks/use-help-search-query.ts

edanzer:
This continues to test as expected for me following latest changes. 

Approving, but just for the github record: Approving knowing the endpoint and associated logic is incomplete still. The endpoint is not being used, and the goal is to get this merged as scaffolding so the full team can being to work on separate parts of the logic. So the bar for approval here is lower than for many other PRs. We want to get some basics in place while doing no harm.

packages/help-center/src/hooks/use-launchpad.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const fetchLaunchpadChecklist = (
+	siteSlug: string | null,
+	siteIntent: string
+): Promise< LaunchpadTasks > => {
+	const slug = encodeURIComponent( siteSlug as string );
+
+	return canAccessWpcomApis()
+		? wpcomRequest( {
```
jeyip, line 33: More info about `wpcomRequest` vs. `apiFetch` in p1681177419999259-slack-CHN6J22MP

packages/help-center/src/hooks/use-launchpad.tsx
jeyip: `Question 1: should we use anything else`

I think we should rename the parameter from `siteIntent` to `checklistSlug`. For the time being, when we consume the hook, we can pass `useLaunchpadChecklist` the value for `siteIntent`, but with the updated parameter naming, it's clearer to future consumers of the hook that they're not limited to _only_ `siteIntent`

`Question 2: If siteIntent is an empty string here, our endpoint returns an empty array, which works fine with how I digest and use the data. But still worth asking what we want to return in case of no siteIntent or checklist_slug?`

I think empty array is fine for now. I generally prefer returning consistent types, so there's less to consider when manipulating the data ( can call array methods on an empty array, but not null ), but not a strong opinion.

packages/help-center/src/hooks/use-launchpad.tsx
```diff
@@ -0,0 +1,52 @@
...
+		  } )
+		: apiFetch( {
+				global: true,
+				path: `sites/${ slug }/launchpad/checklist?checklist_slug=${ siteIntent }`,
+		  } as APIFetchOptions );
+};
+
+export const useLaunchpadChecklist = ( siteSlug: string | null, siteIntent: string ) => {
+	const key = [ 'launchpad', siteSlug ];
+	return useQuery( key, () => fetchLaunchpadChecklist( siteSlug, siteIntent ), {
```
jeyip, line 46: Just leaving a note here to make sure we consider what kind of caching strategy we want ( if we want one at all ).

I'm assuming we'd want to make data stale immediately and refetch on every call

jeyip (commented):
Everything generally looks good to me. Thanks for taking care of this 😄 

I'll test things tomorrow

packages/help-center/src/components/help-center-launchpad.tsx
```diff
@@ -43,6 +44,11 @@ export const HelpCenterLaunchpad = () => {
 		siteSlug = window?.location?.host;
 	}
 
+	const { data } = useLaunchpadChecklist( siteSlug, siteIntent || '' );
+	const totalLaunchpadSteps = data?.checklist?.length || 4;
```
sharpirate, line 48: Curious why we're providing a default value of `4` here? I assume it's to avoid flashing styles when the data is still being fetched?

packages/help-center/src/components/help-center-launchpad.tsx
```diff
@@ -43,6 +44,11 @@ export const HelpCenterLaunchpad = () => {
 		siteSlug = window?.location?.host;
 	}
 
+	const { data } = useLaunchpadChecklist( siteSlug, siteIntent || '' );
```
sharpirate, line 47: Can we have the siteIntent's empty string fallback `|| ''` built into the hook itself. The drawback right now is we have to remember to pass a fallback for that value each time we reuse the hook

jeyip:
I'm unable to get this to work. Wondering if I'm doing something wrong @edanzer 

- sandboxing public API and new launchpad enabled site
- `bin/jetpack-downloader test jetpack-mu-wpcom-plugin update/launchpad-build-flow-task-definitions` ( one of the more recent checklist API PRs )
- check out this PR
- yarn start
- Go to http://calypso.localhost:3000/post/YOURSITESLUG

**Note:** I'm running into TSC errors when I run `yarn build-packages`. I add the `// @ts-nocheck` to the top of each problematic file, at which point tsc compiles, but I'm not sure if this is how you handled the issue

sharpirate:
Just retested this and still works well

image

packages/help-center/src/components/help-center-launchpad.tsx
edanzer: @Sharpirate - I kept the same default values that had been hard coded here before this PR (1 out of 4 steps, or 25%). So if we don't get reliable complete / total steps data, we'd just render the icon the same as we did in the past.

jeyip (approved):

PR #75585, opened 2023-04-11
Staging Sites: Add explanatory tooltip to plan storage indicator

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/2044

## Proposed Changes

Updates the plan storage component to show a helpful tooltip on both staging and production sites:

image

## Testing Instructions

1. Navigate to `/media/<domain>`.
2. On a staging site, the tooltip should appear on hover and the upgrade link should no longer appear.
3. On a Business site with a staging site, the tooltip should appear on hover.
4. On a Business site without a staging site, the tooltip should not appear on hover.
5. On a Personal site, an upgrade link should appear with an upgrade nag in the tooltip.

### DIFF
--- a/client/blocks/plan-storage/index.jsx
+++ b/client/blocks/plan-storage/index.jsx
@@ -109,7 +109,9 @@ export function PlanStorage( { children, className, siteId } ) {
 				title={ translate( 'Storage quota is shared between production and staging.' ) }
 				className="plan-storage__tooltip"
 			>
-				<div className={ classNames( className, 'plan-storage' ) }>{ planStorageComponents }</div>
+				<div className={ classNames( className, 'plan-storage plan-storage__shared_quota' ) }>
+					{ planStorageComponents }
+				</div>
 			</Tooltip>
 		);
 	}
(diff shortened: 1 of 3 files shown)


wojtekn (approved):
The code change looks good and it works as expected.

I'm wondering if "Storage quota is shared between production and staging." is clear enough to say that the displayed quota is already a result of a split and not the one that is shared between the sites. What do you think about making it more clear in the tooltip message e.g. by saying "Storage quota of 200GB is shared between production and staging."?

javierarce:
I couldn't find other examples of that tooltip in use, but I think this one is more common (and also documented): https://wpcalypso.wordpress.com/devdocs/design/info-popover

danielbachhuber:
> I'm wondering if "Storage quota is shared between production and staging." is clear enough to say that the displayed quota is already a result of a split and not the one that is shared between the sites. What do you think about making it more clear in the tooltip message e.g. by saying "Storage quota of 200GB is shared between production and staging."?

@wojtekn Good point. I'm not sure I want to hardcode a value into the tooltip, though.

I gave it a little thought but I can't think of a better phrasing than what I've already implemented.

@javierarce Any ideas on your end?

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7921696

Thank you @danielbachhuber for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7921696).

PR #75590, opened 2023-04-11
Sites Page: Prevent duplicate 'Business' from appearing when expired

danielbachhuber:
Bug introduced in https://github.com/Automattic/wp-calypso/pull/74646

## Proposed Changes

Prevents duplicate 'Business' from appearing when plan is expired:

image

These two conditionals aren't equivalent:
* https://github.com/Automattic/wp-calypso/pull/74646/files#diff-be436b2097d20ffc385e8ef2a2b5cdc160b528ece9475d22f9ca291ecde650f8R44-R53
* https://github.com/Automattic/wp-calypso/pull/74646/files#diff-0ab55e20b31ca28d1f43bea293ab55c2e8b414ec7be2334cc4ff22461bd1cb8eL153-L163

## Testing Instructions

1. View `/sites` with a variety of sites and verify the plans appear as expected.


client/sites-dashboard/components/test/sites-site-plan.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const siteOwnerId = 1;
+const otherAdminId = 2;
+
+const stagingSite = {
+	ID: 1,
+	site_owner: siteOwnerId,
+	title: 'Staging Site',
+	is_wpcom_staging_site: true,
+};
```
danielbachhuber, line 15: @p-jackson @Automattic/build Where might be a good place to put some standard mocks that conform to `SiteExcerptData` ?

client/sites-dashboard/components/sites-site-plan.tsx
```diff
@@ -32,7 +30,7 @@ interface SitePlanProps {
 const STAGING_PLAN_LABEL = 'Staging';
 
 export const SitePlan = ( { site, userId }: SitePlanProps ) => {
-	const isWpcomStagingSite = useSelector( ( state ) => isSiteWpcomStaging( state, site.ID ) );
+	const isWpcomStagingSite = site?.is_wpcom_staging_site ?? false;
```
p-jackson, line 33: I'm guessing this was changed to make it easier to write the unit tests? I think it's a good change because why get the data from the Redux store when we've already been passed all the state we need via props.

But in cases where you _don't_ want to change the code to suit the test, just wanted to make you aware of `renderWithProvider` https://github.com/Automattic/wp-calypso/blob/9aa1faee6367d357e7cd2e477c787abdce0af9b3/client/test-helpers/testing-library/index.js#L8

Used when testing a component that needs access to the Redux store.

client/sites-dashboard/components/test/sites-site-plan.tsx
p-jackson: Controversial opinion: I'm in favour of copy/paste heavy code for tests. Unlike production code, I think it's good for tests to have basically zero abstractions so that it's super easy to see all the inputs and outputs. (I know, I know, `@testing-library` itself is an abstraction, but ignore that part 😅). I guess the downside is that if we ever added a new mandatory field to `SiteExcerptData` then there'd be lots of places to update. But given that's an easy change to make, compared to reading a test failure in code you're unfamiliar with, I'd make that trade off.

Anyway to answer the question, a pattern I'm familiar with in Calypso is the `client/state/selectors/test/fixtures/` folder. It's filled with objects that you can use to easily mock out a Redux store.

For `SiteExcerptData` I'd do something like `client/data/sites/test/fixtures/site-excerpt-data.ts`, putting the fixtures close to the hook that produces objects of that type. That way it'd be sitting alongside hypothetical tests in `client/data/sites/test/use-site-excerpts-query.ts`

client/sites-dashboard/components/test/sites-site-plan.tsx
```diff
@@ -0,0 +1,56 @@
...
+			<SitePlan site={ activeBusinessSite } userId={ otherAdminId } />
+		);
+		expect( container.textContent ).toBe( 'Business' );
+	} );
+
+	test( 'shows "Business" as label for a site with an expired Business plan', () => {
+		const { container } = render(
+			<SitePlan site={ expiredBusinessSite } userId={ otherAdminId } />
+		);
+		expect( container.textContent ).toBe( 'Business - Expired' );
```
p-jackson: Here checking `.textContent` directly works because there's nothing else within the component, but with `@testing-library` you usually search for the element using text/label/aria role/etc. so that the test is more resilient.

For example you can't test the case where the plan has expired and the current user is the site owner. But you could with `getByRole` and `getByText`

```ts
		const { getByText, getByRole } = render(
			<SitePlan site={ expiredBusinessSite } userId={ siteOwnerId } />
		);
		expect( getByText( 'Business - Expired' ) ).toBeInTheDocument();
		expect( getByRole( 'link', { name: 'Renew plan' } ) ).toHaveAttribute(
			'href',
			expect.stringMatching( /\/checkout\// )
		);
```

p-jackson (approved):
I've given it a test and it's working for me 👍

wojtekn (approved):

client/sites-dashboard/components/sites-site-plan.tsx
danielbachhuber: > I'm guessing this was changed to make it easier to write the unit tests? I think it's a good change because why get the data from the Redux store when we've already been passed all the state we need via props.

@p-jackson Yes to both.

client/sites-dashboard/components/test/sites-site-plan.tsx
danielbachhuber: @p-jackson Thanks! For the first couple of scenarios, I want to assert _exactly_ the text I expect.

For the second two scenarios, I applied your suggestions e02fb9be98

client/sites-dashboard/components/test/sites-site-plan.tsx
danielbachhuber: > Controversial opinion: I'm in favour of copy/paste heavy code for tests.

@p-jackson Fair enough! I'll leave it as-is for now.

PR #75606, opened 2023-04-12
Site Assembler: Prevent the Continue button mis-clicking

arthur791004:
Related to p1681263218769429-slack-CRWCHQGUB

## Proposed Changes

* The user easily misclicks the Continue button to proceed to the next step when they come back to the sub-screen. So, this PR is proposing to delay the Continue button being clickable by about 300ms to reduce this kind of accident.
* Also, adding a new event to track mis-clicking
  image

## Testing Instructions

* Go to /setup?siteSlug=<your_free_site>
* On the Design Picker screen, scroll to the bottom and select "Start designing"
  * Go to any sub-screen
  * Double click the `Save` button
  * When you're back to the main screen, ensure you're not proceeding to the next step

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-main.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-main.tsx
@@ -4,6 +4,7 @@ import { __experimentalHStack as HStack } from '@wordpress/components';
 import { header, footer, layout, color, typography } from '@wordpress/icons';
 import { useTranslate } from 'i18n-calypso';
 import { useState, useEffect } from 'react';
+import { PATTERN_ASSEMBLER_EVENTS } from './events';
 import { NavigationButtonAsItem } from './navigator-buttons';
 import NavigatorHeader from './navigator-header';
 import { NavigatorItemGroup } from './navigator-item-group';
@@ -14,6 +15,7 @@ interface Props {
 	isDismissedGlobalStylesUpgradeModal?: boolean;
 	onSelect: ( name: string ) => void;
 	onContinueClick: () => void;
+	recordTracksEvent: ( name: string, eventProperties?: any ) => void;
 }
 
 const ScreenMain = ( {
@@ -21,6 +23,7 @@ const ScreenMain = ( {
 	isDismissedGlobalStylesUpgradeModal,
 	onSelect,
 	onContinueClick,
+	recordTracksEvent,
 }: Props ) => {
 	const translate = useTranslate();
 	const [ disabled, setDisabled ] = useState( true );
@@ -42,6 +45,7 @@ const ScreenMain = ( {
 	const handleMouseDown = ( event: MouseEvent< HTMLButtonElement > ) => {
 		if ( disabled ) {
 			event.preventDefault();
+			recordTracksEvent( PATTERN_ASSEMBLER_EVENTS.CONTINUE_MISCLICK );
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-main.tsx
```diff
@@ -34,6 +37,24 @@ const ScreenMain = ( {
 		return translate( "You've selected a premium color or font for your site." );
 	};
 
+	// Use the mousedown event to prevent either the button focusing or text selection
+	const handleMouseDown = ( event: MouseEvent< HTMLButtonElement > ) => {
+		if ( disabled ) {
+			event.preventDefault();
```
miksansegundo, line 47: Can we add a tracks event to know how many mis-clicks we avoid with this fix?

arthur791004: Sounds good!

miksansegundo (approved):
Tested and LGTM 👍 

Adding a tracks event will help to measure the impact of this issue.

I believe it would be better to improve the UX by removing the `Save` buttons or something else like a little modal before redirecting to the editor.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-main.tsx
arthur791004: Added 👍

taipeicoder (approved):

okmttdhr (approved):

autumnfjeld:
@miksansegundo  @arthur791004   Good you added the tracks event.  My guess is that the % of users double clicking super fast will be very very low.  I think it is something that happens to us as devs/testers, but that users act slower.    Will be interesting to see the data. 

@miksansegundo Feel free to create an issue with the idea to add the "Are you sure modal" (Or whatever we want to call it) and label it with "[Needs Design](https://github.com/orgs/Automattic/projects/338/views/40)", then we have an item open for discussion to share with Dave.

autumnfjeld:
I just tested this in production (is it live in production?) and my fast double click landed me in the editor.  Tested twice.

* Go to Assembler
* Click on subscreen
* Click Save
* Immediately click Continue
* Land in Editor

Instead I expected my "Continue" click to do nothing.

arthur791004:
> @miksansegundo @arthur791004 Good you added the tracks event. My guess is that the % of users double clicking super fast will be very very low. I think it is something that happens to us as devs/testers, but that users act slower. Will be interesting to see the data

Yes, it looks like the users act slower as expected as we just got a several events last 4 days

image

worldomonation:
On the other hand, this change is causing a big headache for the E2E for this flow in https://github.com/Automattic/wp-calypso/pull/75813. 

Playwright sees the button, passes all actionability checks (because the button is not denoted with an [aria-enabled="false"]) and so clicks on it thinking it's ready for interaction, when it is not.

If the button is disabled, should it not have the appropriate attributes to denote such, especially in light of a11y compliance?

arthur791004:
> Playwright sees the button, passes all actionability checks (because the button is not denoted with an [aria-enabled="false"]) and so clicks on it thinking it's ready for interaction, when it is not.

I'll add the `aria-enabled` to that button via https://github.com/Automattic/wp-calypso/pull/75865 when it's unavailable. Thanks for the feedback 👍

worldomonation:
> > Playwright sees the button, passes all actionability checks (because the button is not denoted with an [aria-enabled="false"]) and so clicks on it thinking it's ready for interaction, when it is not.
> 
> I'll add the `aria-enabled` to that button via #75865 when it's unavailable. Thanks for the feedback 👍

Appreciate it @arthur791004! Thank you so for being so receptive.

PR #75609, opened 2023-04-12
Add extended header section to logged out tags page.

roo2:
Part of pe7F0s-IS-p2

Adds the header section for logged out users. Header section for logged in users will stay the same for now as per pe7F0s-IS-p2#comment-799

Before:
Screen Shot 2023-04-12 at 3 02 44 pm
After:
Screen Shot 2023-04-12 at 3 02 12 pm

This also enables the wordpress logo on mobile for anywhere logged out masterbar is used.
I couldn't find examples of where the logged out masterbar is used except for when it flashes on screen before the normal logged out header loads, for example: previously the wordpress logo in the black masterbar would not be shown on mobile.

https://user-images.githubusercontent.com/22446385/231354561-a92d513d-f797-4e58-8559-37ddce04b021.mov

### DIFF
--- a/client/reader/tag-stream/style.scss
+++ b/client/reader/tag-stream/style.scss
@@ -43,3 +43,19 @@
 .is-reader-page .tag-stream__main.main {
 	max-width: 800px;
 }
+
+// specific rule to override existing rules
+.layout.has-header-section.is-section-reader .layout__header-section-content .tag-stream__page-header {
+	text-align: center;
+
+	h2 {
+		font-size: 18px; /* stylelint-disable-line declaration-property-unit-allowed-list */
+		font-weight: 500;
+		margin-bottom: 10px;
+	}
+
+	h1 {
+		font-size: 40px; /* stylelint-disable-line declaration-property-unit-allowed-list */
+		font-weight: 500;
+	}
+}
--- a/client/layout/masterbar/logged-out.jsx
+++ b/client/layout/masterbar/logged-out.jsx
@@ -150,7 +150,7 @@ class MasterbarLoggedOut extends Component {
 
 		return (
 			<Masterbar>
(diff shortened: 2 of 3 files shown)


eoigal:
Going by designs in p9Jlb4-6CW-p2 - we need to include a minimized version of this header if the user has viewed more than 1 tag. Could be done as a follow up, just wondering what do you think?

roo2:
ahh let's just do it in a follow up PR I think. It seems like a standalone thing that's not essential to the functionality working

eoigal (approved):
WFM

client/reader/tag-stream/style.scss
```diff
@@ -43,3 +43,19 @@
 .is-reader-page .tag-stream__main.main {
 	max-width: 800px;
 }
+
```
eoigal, line 46:
```suggestion
.layout.has-header-section.is-section-reader.has-header-section .layout__header-section {
	background-image: none;

	.tag-stream__page-header {
		text-align: center;

		h2 {
			font-size: 18px; /* stylelint-disable-line declaration-property-unit-allowed-list */
			font-weight: 500;
			margin-bottom: 10px;
		}

		h1 {
			font-size: 40px; /* stylelint-disable-line declaration-property-unit-allowed-list */
			font-weight: 500;
		}
	}
}
```

eoigal: If we want to have a black (no image) background for logged out users, we can use this styling

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7924462

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7924462).

PR #75611, opened 2023-04-12
Hosting onboarding: only reuse is-user class name when visiting the user-hosting step on signup

zaguiini:
Related to https://github.com/Automattic/wp-calypso/pull/74911

## Proposed Changes

Fixes a regression introduced in https://github.com/Automattic/wp-calypso/pull/74911 where other flows were affected by the step name to component name mapping. We're only treating `user-hosting` (and renaming it as `user`) moving forward.

## Testing Instructions

1. Checkout this branch;
2. Visit `/start/hosting` and check that the passwordless flow is styled just like `trunk`;
3. Visit `/start/do-it-for-me-store` and check that the last step is styled just like the "after" column in [this comment](https://github.com/Automattic/wp-calypso/pull/74911#discussion_r1163643335).

### DIFF
--- a/client/signup/main.jsx
+++ b/client/signup/main.jsx
@@ -73,7 +73,7 @@ import {
 } from 'calypso/state/sites/selectors';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import flows from './config/flows';
-import { getStepComponent, getStepModuleName } from './config/step-components';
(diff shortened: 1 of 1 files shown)


p-jackson (approved):

aneeshd16:
Thank you for fixing this so quickly! 🎉

PR #75628, opened 2023-04-12
Revert "Hosting onboarding flow: use new plans grid"

vindl:
Reverts Automattic/wp-calypso#75547

The Enterprise plan was removed from the plans grid in general, and not just from the Hosting onboarding flow.


aneeshd16 (approved):
Confirmed that the Enterprise plan is shown on this branch:
image

I'm not sure if any hosting flow feature is dependent on the contents of the original PR though.

PR #75639, opened 2023-04-12
Updates theme detection of full site editing to use block-template-parts

### LINKED ISSUE #62755: Atomic: "Edit Site" button doesn't update after theme switch to/from FSE
### Quick summary

When someone installs a theme that requires a site builder(e.g. Elementor) like Hello Elementor, the "Edit site" button will link to this:

https://SITENAME/wp-admin/themes.php?page=gutenberg-edit-site

This will render this message:

`"Sorry, you do not have permission to access this page."
`

### Steps to reproduce

1. Go to https://wordpress.com/home/sitename
2. Click on "Edit site" on the right sidebar
image

Be forwarded to the Pages section at least or wp-admin/Pages

### What actually happened

I saw this:

image

### Context

4946414-zen

### Simple, Atomic or both?

_No response_

### Theme-specific issue?

This appears to be dependent on the theme installed, but only those that require a site builder.(Elementor, Astra, etc)

### Browser, operating system and other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Go through wp-admin instead

creativecoder:
## Proposed Changes

WordPress 6.1 (https://github.com/WordPress/wordpress-develop/commit/98cfa29f8d312e1aa8f3979387c75e26b5bf1734) introduced the `block-template-parts` theme support and uses that to determine support for the Site editor.

This change updates Calypso logic to do the same.

## Testing Instructions

- The easiest way to make sure this works correctly is to enable a block based theme, and then a classic theme.
- Check that the link at the top of the Quick links section of My Home says either "Edit site" (for themes that support the Site editor) or "Write blog post" (classic themes)

| **Classic** | **Block** |
| - | - |
| image | image |

### DIFF
--- a/client/data/themes/with-is-fse-active.js
+++ b/client/data/themes/with-is-fse-active.js
@@ -9,7 +9,7 @@ const withIsFSEActive = createHigherOrderComponent(
 		const siteId = useSelector( getSelectedSiteId );
 		const userLoggedIn = useSelector( isUserLoggedIn );
 		const { data, isLoading } = useActiveThemeQuery( siteId, userLoggedIn );
-		const isFSEActive = data?.[ 0 ]?.theme_supports?.[ 'block-templates' ] ?? false;
+		const isFSEActive = data?.[ 0 ]?.theme_supports?.[ 'block-template-parts' ] ?? false;
 
 		return <Wrapped { ...props } isFSEActiveLoading={ isLoading } isFSEActive={ isFSEActive } />;
 	},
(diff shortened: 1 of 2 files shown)


paulopmt1:
Hey @creativecoder, tested the feature with classical themes and we don't get `Edit site` anymore, which is great. After switching to a FSE theme, like `Twenty Twenty-Two` I still can't see it. It's weird, because it seems we're not receiving the correct information from our backend. Am I missing something? Can you confirm that?

image

creativecoder:
@paulopmt1 You're totally right, it looks like the `block-template-parts` theme support [isn't added automatically](https://github.com/WordPress/gutenberg/blob/trunk/docs/how-to-guides/themes/theme-support.md#block-based-template-parts) ([like `block-templates` is](https://github.com/WordPress/wordpress-develop/blob/ce28deee43c222e6c9ff27da215b24248e371b3a/src/wp-includes/theme-templates.php#L211)). So the result won't be consistent--it'll depend on if the theme author knew to properly declare support since that was introduced late last year.

What we're missing is the ability to access the value of `wp_is_block_theme()` from the API, which detects Site editor support by looking for the actual template files in the theme. We used to use the `wpcom/v2/sites/%s/block-editor` endpoint to get this value, but the endpoint isn't in Jetpack, which caused some issues with Atomic sites.

Ideally Core would surface this value in the themes endpoint. I'll make an issue, but it'll take a while to get this through.

creativecoder:
Closing, as this isn't a proper solution.

Issue #75172, opened 2023-03-31
Bug: Update post-upgrade screen to use the correct payments task


daledupreez:
### Details

We show a post-upgrade screen for users who have upgraded to a Woo Express plan. This page is shown at `/plans/my-plan/trial-upgraded/:siteSlug`. The screen includes a link to set up payment methods. However, the link is for a generic payments setup task, which may not be active when WooCommerce Payments has been installed.

We need to ensure that the task links to a valid, available task.

### Checklist

- [ ] Ensure that the link loads a valid task when WooCommerce Payments is installed

sixhours:
This has been fixed in #75642!

PR #75642, opened 2023-04-12
Woo Express: Update post-purchase payments link to point to payments page

sixhours:
Related to #75172

## Proposed Changes

* Rather than pointing to a non-existent task, check to see if the user has WC Payments installed before pointing to the WC Payments page; otherwise, go to the Home screen.

## Testing Instructions

* Switch to this PR
* Create a new test site from `/setup/wooexpress`
* Upgrade the site to Essential or Performance
* After checkout, click on the "Provide a way to pay" box

Screen Shot 2023-04-12 at 12 49 20 PM

* If in a country that supports WC Payments, you should be brought to the Payments screen:

Screen Shot 2023-04-12 at 12 48 17 PM

* If in a country that does not support WC Payments, you'll be brought to the home screen:

Screen Shot 2023-04-25 at 12 07 26 PM

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
@@ -1,13 +1,17 @@
 import { translate as i18nTranslate } from 'i18n-calypso';
+import { useSelector } from 'react-redux';
 import customDomain from 'calypso/assets/images/plans/wpcom/custom-domain.png';
 import customize from 'calypso/assets/images/plans/wpcom/customize.png';
 import launch from 'calypso/assets/images/plans/wpcom/launch-store.png';
 import manageWooCommerce from 'calypso/assets/images/plans/wpcom/manage-woocommerce.png';
 import promote from 'calypso/assets/images/plans/wpcom/promote.png';
 import wayToPay from 'calypso/assets/images/plans/wpcom/way-to-pay.png';
+import isPluginActive from 'calypso/state/selectors/is-plugin-active';
+import { getSiteId } from 'calypso/state/sites/selectors';
 
 type ConfirmationTasksProps = {
 	translate: typeof i18nTranslate;
+	siteSlug?: string | number;
 };
 
 export interface GetActionUrlProps {
@@ -17,7 +21,12 @@ export interface GetActionUrlProps {
 	wpAdminUrl: string;
 }
 
-export const getConfirmationTasks = ( { translate }: ConfirmationTasksProps ) => {
+export const GetConfirmationTasks = ( { translate, siteSlug }: ConfirmationTasksProps ) => {
+	const siteId = useSelector( ( state ) => siteSlug && getSiteId( state, siteSlug ) );
+	const hasWCPay = useSelector(
+		( state ) => siteId && isPluginActive( state, siteId, 'woocommerce-payments' )
+	);
+
 	return [
 		{
 			id: 'launch-store',
@@ -41,7 +50,7 @@ export const getConfirmationTasks = ( { translate }: ConfirmationTasksProps ) =>
 				'Set up one or more payment methods to make it easy for your customers to pay.'
 			),
 			getActionUrl: ( { wooAdminUrl }: GetActionUrlProps ) =>
-				`${ wooAdminUrl }&path=${ encodeURIComponent( '/payments/connect' ) }`,
+				`${ wooAdminUrl }${ hasWCPay ? '&path=' + encodeURIComponent( '/payments/connect' ) : '' }`,
 		},
 		{
 			id: 'custom-domain',
--- a/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
+++ b/client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
@@ -41,7 +41,7 @@ export const getConfirmationTasks = ( { translate }: ConfirmationTasksProps ) =>
 				'Set up one or more payment methods to make it easy for your customers to pay.'
 			),
 			getActionUrl: ( { wooAdminUrl }: GetActionUrlProps ) =>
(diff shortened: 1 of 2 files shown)


daledupreez:
I don't think this is the right approach for multiple reasons:
1. We can't assume the user is eligible to use WC Pay, as it is not available in a large number of countries
2. We can't assume WC Pay is active

Without thinking about this too deeply, it may be worth identifying whether WC Pay is installed and active, and then rendering the appropriate payments task?

markbiek (approved):
LGTM!

sixhours:
> Without thinking about this too deeply, it may be worth identifying whether WC Pay is installed and active, and then rendering the appropriate payments task?

Ahh, I didn't realize that page was specific to WC Pay and not a generic "Payments" type admin page. 🙃 

I'll check the logic for the checklist and see what approach we take there, provided it also accounts for potential differences.

sixhours:
A check for `woocommerce-payments` has been implemented, but it's unclear what we should do in the event someone doesn't have access to that plugin. The onboarding tasks don't seem to offer an alternative to WC Pay, so right now I'm just linking to the Home screen. Thoughts? /cc @daledupreez

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
```diff
@@ -17,7 +21,12 @@ export interface GetActionUrlProps {
 	wpAdminUrl: string;
 }
 
-export const getConfirmationTasks = ( { translate }: ConfirmationTasksProps ) => {
+export const GetConfirmationTasks = ( { translate, siteSlug }: ConfirmationTasksProps ) => {
```
daledupreez: Nit: I don't like the case change here, as this is not a component, but the changed name suggests it is.

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -1,22 +1,23 @@
...
 
 import './style.scss';
 
 const TrialUpgradeConfirmation = () => {
 	const selectedSite = useSelector( getSelectedSite );
 	const translate = useTranslate();
-
-	const tasks = getConfirmationTasks( { translate } );
+	const siteSlug = selectedSite?.slug;
+	const tasks = GetConfirmationTasks( { translate, siteSlug } );
```
daledupreez: Can we pass the `siteId` in directly? That simplifies the logic in the function so it doesn't need to resolve the slug to a siteId when we already have the ID.
```suggestion
	const siteId = selectedSite?.ID;
	const tasks = GetConfirmationTasks( { translate, siteId } );
```

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
```diff
@@ -40,7 +49,8 @@ export const getConfirmationTasks = ( { translate }: ConfirmationTasksProps ) =>
 			subtitle: translate(
 				'Set up one or more payment methods to make it easy for your customers to pay.'
 			),
-			getActionUrl: ( { wooAdminUrl }: GetActionUrlProps ) => `${ wooAdminUrl }&task=payments`,
+			getActionUrl: ( { wooAdminUrl }: GetActionUrlProps ) =>
+				`${ wooAdminUrl }${ hasWCPay ? '&path=' + encodeURIComponent( '/payments/connect' ) : '' }`,
```
daledupreez: I think this logic needs to be slightly different based on whether WooCommerce Payments is installed and active, because we should fall back on the old URL when it is not active. I believe Woo uses the original task by default, but we remove or override it when WooCommerce payments is installed.

The complication here is that the default state depends on which country your store is in, and whether WC Pay is available in that country.

```suggestion
			getActionUrl: ( { wooAdminUrl }: GetActionUrlProps ) => {
				// WC Pay overrides the default task, so use this task when WC Pay is active.
				if ( hasWCPay ) {
					return `${ wooAdminUrl }&path=${ encodeURIComponent( '/payments/connect' ) }`;
				}

				return `${ wooAdminUrl }&task=payments`;
```

daledupreez (commented):
Argh! I took a look at some point but forgot to submit my review. Apologies for the delay, @sixhours!

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/confirmation-tasks.ts
sixhours: Ahh yeah. I was trying to be sneaky because we can't use React hooks in non-components, but I can probably pass that data to getConfirmationTasks from the parent component instead...I'll give that a try!

markbiek:
It's been a minute since I first reviewed this but I wanted to mention it's still working as expected!

client/my-sites/plans/ecommerce-trial/upgrade-confirmation/index.tsx
```diff
@@ -15,8 +17,11 @@ import './style.scss';
 const TrialUpgradeConfirmation = () => {
 	const selectedSite = useSelector( getSelectedSite );
 	const translate = useTranslate();
-
-	const tasks = getConfirmationTasks( { translate } );
+	const siteId = selectedSite?.ID;
+	const hasWCPay = useSelector(
+		( state ) => siteId && isPluginActive( state, siteId, 'woocommerce-payments' )
+	) as boolean;
+	const tasks = getConfirmationTasks( { translate, hasWCPay } );
```
daledupreez, line 24: Non-blocking thought: this might be worth wrapping in `useMemo()` so it only gets recomputed when `translate` or `hasWCPay` change.
```suggestion
	const tasks = useMemo( () => getConfirmationTasks( { translate, hasWCPay } ), [ hasWCPay, translate ] );
```

daledupreez (approved):
Tested with and without WooCommerce Payments active on a store, and the tasks work as expected.

🚢

PR #75645, opened 2023-04-12
Hosting onboarding flow: use new plans grid

zaguiini:
Related to https://github.com/Automattic/dotcom-forge/issues/2126.

## Proposed Changes

This PR brings back the new plan grid on hosting after a [revert PR](https://github.com/Automattic/wp-calypso/pull/75628) was merged.

The fix was to forward the prop from the plans step container and also I fixed the conditional. It was "working" on the hosting flow because the conditional was flipped, resulting in buggy behavior.

## Testing Instructions

1. Checkout this branch;
2. Verify that `/start/plans` include the Enterprise plan;
3. Verify that `/start/hosting/plans-hosting` does not include the Enterprise plan.

Of course, make sure to have the `onboarding/2023-pricing-grid` feature flag turned on.


p-jackson (approved):
I've tested existing flows and the new hosting flow and the enterprise plan is hidden and appears at the correct time.

I've gone ahead and added `hideEnterprisePlan` to the component's prop types.

I'm a little bit suspicious of how the enterprise plan is removed in `getPlansForPlanFeatures()`, given that right below where we exclude the enterprise plan there's a comment that says

```js
/*
 * We need to pass all the plans in order to show the correct features in the plan comparison table.
 * Pleas use the getVisiblePlansForPlanFeatures selector to filter out the plans that should not be visible.
 */
```

So it would seem that the plan should actually be excluded in the `getVisiblePlansForPlanFeatures` function instead. However the way it's working now was existing, so I'm not going to change it.

p-jackson:

I'm guessing this matticbot message was triggered by the change to `packages/calypso-products/src/plans-utilities.ts`. However when I checked out the ETK changes in the sandbox it doesn't look like the `plans-utilities.ts` changes are included anywhere D107738-code. So I haven't deployed ETK.

PR #75275, opened 2023-04-04
Launchpad: Update Copy Icon & Fix Tooltip Z-index

sharpirate:
### Time Estimate
Review: Short
Test: Short

### Summary
• Update the clipboard button's icon to use `@wordpress/icons`
• Fix tooltip's Z index inside the celebration modal (It was previously invisible)

image
image

### Testing
1. Checkout this branch
2. Create a new Launchpad site
3. Once you arrive at Launchpad, hover over the domain container and verify the copy icon is updated. Verify the copy functionality continues to work as expected
4. Launch your site
5. Once you see the celebration modal, hover over the domain container and verify the copy icon is updated. Verify the copy functionality continues to work as expected. Verify you can see the tooltip when you copy the domain

PR #75648, opened 2023-04-12
Launchpad: Remove unscoped css style from Celebrate Launch modal

sharpirate:
This is a follow up to this PR https://github.com/Automattic/wp-calypso/pull/75275#discussion_r1160660752

### Time Estimate
Review: Short
Test: Short

### Summary
This PR removes unscoped styles from `celebrate-launch.scss` to the tooltip component's own stylesheet to avoid unexpected style bleed.

### Testing
1. Checkout this branch
2. Create a new Launchpad site and proceed to launch it
3. Once you get redirected to My Home you will see the celebrate launch modal appear
4. Click the Copy site button and confirm you still see the tooltip **Copied to clipboard!**

image


jeyip:
@nuriapenya Do you happen to know who else we'd need to talk to about making a change to the shared tooltip component z index style?

jeyip:
From the technical side of things,  things look fine. We'll want to smoke test other instances of tooltips though

client/components/tooltip/style.scss
```diff
@@ -1,6 +1,10 @@
 $large-tooltip-box-shadow-color: rgba(0, 0, 0, 0.05);
 
 .tooltip.popover {
+
+	// always display the popover on top even inside a modal with a high z index
+	z-index: 100000;
```
simison: I remember we used to have some sass functions to manage z-indexes ([similar to Gutenberg](https://github.com/WordPress/gutenberg/blob/308f3ee5ce02192d6129e0764d62ecf5c5ebb6e1/packages/components/src/tooltip/style.scss#L2)).

It makes it easier to manage different z-indexes in different components collectively.

Can you look up if that exists and use it here? Thanks!

simison:
Looks good — I haven't tested but I'd test modals, dropdowns, "lightboxes" or "[scrims](https://m2.material.io/design/environment/surfaces.html#attributes)" to ensure it works as expected; above modals and dropdowns but below scrims for example.

I left technical nitpick which can help manage above.

Also asked Team Calypso to review but I don't think that has to be blocking merging if you're confident this works. Up to you!

client/assets/stylesheets/shared/functions/_z-index.scss
```diff
@@ -172,6 +172,7 @@ $z-layers: (
 
 		".is-section-woocommerce .global-notices": 999999,
 		".webpack-build-monitor": 99999999,
+		".tooltip.popover": 100000,
```
simison: I think you'd put this somewhere higher on the list, they seem to be in numerical order?

sharpirate: Thanks, updated!

sharpirate:
> Looks good — I haven't tested but I'd test modals, dropdowns, "lightboxes" or "[scrims](https://m2.material.io/design/environment/surfaces.html#attributes)" to ensure it works as expected; above modals and dropdowns but below scrims for example.

Hey @simison, do you know how I can find instances of all of those to test with the popup? Not sure if there are specific classnames corresponding to each one

simison:
I'd probably look from code with component name, or create dummy page with all sorts of components mixed to ensure it works all together. Playing with devdocs might also help, since all components should be there in one page https://wpcalypso.wordpress.com/devdocs — they're just not combined together, e.g. tooltip in a modal etc.

sharpirate:
I couldn't find any obvious issues when testing, so I'll go ahead and deploy this. If anything pops up I can address it in a follow up

PR #75654, opened 2023-04-12
Redirect to `/sites` after deleting site, depending on user prefs

p-jackson:
## Proposed Changes

Just a papercut I felt while testing.

After deleting a site in Calypso the current site selection is cleared and you're taken to `/stats/day`. But that doesn't seem like the natural place to go for someone who's accustomed to using the `/sites` page.

The "Admin home" preference in `/me/account` is used to determine what view the user wants to see by default after logging in.

CleanShot 2023-04-13 at 10 53 17@2x

We can use this same preference as a hint for what view the user would expect to see after deleting a site.

Another way of thinking about it is after deleting a site I expect to be taken to the top of the IA, and for users with the "Admin home" preference set, the top of the IA is `/sites`.

One could make a case that after deleting a site we should _always_ go to `/sites`, but I only thought of it because I have the setting enabled, and it's easier to justify making the change if we depend on the setting I think.

## Testing Instructions

* Ensure `/sites` _isn't_ the home page and delete a site. The behaviour should be the same as before—you are taken to an all-sites stats view.
* Set `/sites` as the home page and delete a site. After successfully deleting the site you should be taken to `/sites`
* Create a brand new user with only one site. Set `/sites` as the home page. Delete your one and only site. You should be taken to `/sites` and see the "create your first site" experience. IMO this feels like the right experience for someone who has deleted all of their sites.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/site-settings/delete-site/index.jsx
+++ b/client/my-sites/site-settings/delete-site/index.jsx
@@ -21,6 +21,7 @@ import hasCancelableSitePurchases from 'calypso/state/selectors/has-cancelable-s
 import isSiteAutomatedTransfer from 'calypso/state/selectors/is-site-automated-transfer';
 import { deleteSite } from 'calypso/state/sites/actions';
 import { getSite, getSiteDomain } from 'calypso/state/sites/selectors';
+import { hasSitesAsLandingPage } from 'calypso/state/sites/selectors/has-sites-as-landing-page';
 import { setSelectedSiteId } from 'calypso/state/ui/actions';
 import { getSelectedSiteId, getSelectedSiteSlug } from 'calypso/state/ui/selectors';
 
@@ -93,11 +94,15 @@ class DeleteSite extends Component {
 	};
 
 	componentDidUpdate( prevProps ) {
-		const { siteId, siteExists } = this.props;
+		const { siteId, siteExists, useSitesAsLandingPage } = this.props;
 
 		if ( siteId && prevProps.siteExists && ! siteExists ) {
 			this.props.setSelectedSiteId( null );
-			page.redirect( '/stats' );
+			if ( useSitesAsLandingPage ) {
+				page.redirect( '/sites' );
+			} else {
+				page.redirect( '/stats' );
+			}
 		}
 	}
 
@@ -364,6 +369,7 @@ export default connect(
 			siteSlug,
 			siteExists: !! getSite( state, siteId ),
 			hasCancelablePurchases: hasCancelableSitePurchases( state, siteId ),
+			useSitesAsLandingPage: hasSitesAsLandingPage( state ),
 		};
 	},
 	{


p-jackson:
I think we should also do something like this so that users get sent to `/sites` after disconnecting a Jetpack site too. However it needs a bit more work so I left it out of this PR.

The problem is that the Jetpack disconnect flow navigates directly to `/sites`, even before the deletion is completed. That means the Jetpack site still appears in the sites management page. What we want is something to invalidate the `useSiteExcerptsQuery` hook once the Jetpack site has successfully disconnected so that it automatically disappears.

Regular WordPress.com sites don't have this issue because the site deletion flow doesn't navigate to `/sites` until the deletion process is complete.

```diff
 .../my-sites/site-settings/disconnect-site/confirm.jsx   | 9 ++++++++-
 1 file changed, 8 insertions(+), 1 deletion(-)

diff --git a/client/my-sites/site-settings/disconnect-site/confirm.jsx b/client/my-sites/site-settings/disconnect-site/confirm.jsx
index 3a29192279..c29be670fa 100644
--- a/client/my-sites/site-settings/disconnect-site/confirm.jsx
+++ b/client/my-sites/site-settings/disconnect-site/confirm.jsx
@@ -11,6 +11,7 @@ import NavigationLink from 'calypso/components/wizard/navigation-link';
 import { submitSurvey } from 'calypso/lib/purchases/actions';
 import redirectNonJetpack from 'calypso/my-sites/site-settings/redirect-non-jetpack';
 import { getCurrentPlan } from 'calypso/state/sites/plans/selectors';
+import { hasSitesAsLandingPage } from 'calypso/state/sites/selectors/has-sites-as-landing-page';
 import {
 	getSelectedSite,
 	getSelectedSiteId,
@@ -29,6 +30,7 @@ class ConfirmDisconnection extends Component {
 		siteId: PropTypes.number,
 		siteSlug: PropTypes.string,
 		translate: PropTypes.func,
+		useSitesAsLandingPage: PropTypes.bool,
 	};
 
 	static allowedReasons = [
@@ -61,6 +63,10 @@ class ConfirmDisconnection extends Component {
 		);
 	};
 
+	getDefaultDisconnectHref = () => {
+		return this.props.useSitesAsLandingPage ? '/sites' : '/stats';
+	};
+
 	render() {
 		const { disconnectHref, siteId, siteSlug, stayConnectedHref, translate, type } = this.props;
 
@@ -78,7 +84,7 @@ class ConfirmDisconnection extends Component {
 					) }
 				/>
 				<DisconnectJetpack
-					disconnectHref={ disconnectHref ?? '/stats' }
+					disconnectHref={ disconnectHref ?? this.getDefaultDisconnectHref() }
 					isBroken={ false }
 					onDisconnectClick={ this.submitSurvey }
 					showTitle={ false }
@@ -101,6 +107,7 @@ const connectComponent = connect(
 			site: getSelectedSite( state ),
 			siteId,
 			siteSlug: getSelectedSiteSlug( state ),
+			useSitesAsLandingPage: hasSitesAsLandingPage( state ),
 		};
 	},
 	{ submitSurvey }
-- 
2.39.2 (Apple Git-143)

```

mpkelly (approved):
Good idea. Tested and working. LGTM

PR #75679, opened 2023-04-13
Hosting onboarding: Customising the "included with your purchase" list of the hosting flow

p-jackson:
Closes Automattic/dotcom-forge#2125. Closes https://github.com/Automattic/dotcom-forge/issues/2121.

## Proposed Changes

Updates the "Included with your purchase" list in checkout for users who are going through the `/start/hosting` onboarding flow so that it includes more technical focused features. Like they would have seen on the hosting LOHP.

**Business plan**
bus

**Commerce plan**
comm

Usually the items in the list are based on what's in the cart and so can be different based on whether for example there's a domain in the cart. However the ability to customise the list based on onboarding flow had already been added for the "tailored onboarding flows". We just need to update the value of `shouldUseFlowFeatureList` so it's also true for the hosting flow.

One difference between the tailored onboarding flows and the regular flows is that the flow-based feature list didn't include the refund window. But that's something we're trying to emphasise in the hosting flow. So I made sure that it's included in our flow.

One downside of basing the list on the flow and not the cart contents is that we don't highlight the fact that the user is purchasing a domain. Usually a business plan would see a list like this:
biz-with-domain

I don't think this is a huge deal. But worth pointing out.

## Testing Instructions

* Test the `/setup/link-in-bio` flow and ensure the checkout experience hasn't changed
* Test the regular `/start/onboarding` flow and ensure the checkout experience hasn't changed
* Test the `/start/hosting` flow
   * Test with and without domains
   * Test with business plan
   * Test with commerce plan
* Test the [etk](https://github.com/Automattic/wp-calypso/pull/75679#issuecomment-1506558301) and [happy-blocks](https://github.com/Automattic/wp-calypso/pull/75679#issuecomment-1506558898) builds


p-jackson:
What do you think about the proposed feature lists @javierarce @zaguiini? They're pretty long, especially the Commerce one. But I could really decide which my favourite hosting features were 😊
Maybe the long list helps with sticker shock anyway.

The Commerce one is so long because I felt like we had to at least mention the features they're paying the premium price for. I don't think we're expecting `/start/hosting` users to select Commerce, so maybe we could also just leave it unchanged?

I decided against it, but we also have the option of **bolding** specific items.

javierarce:
> Maybe the long list helps with sticker shock anyway.

I agree, you can see immediately that you get a ton of features.

zaguiini:
I noticed that the free domain is on the list even for monthly plans. I raised this issue (p1681477905718389-slack-C0117V2PCAE) because this behavior is observed in the Link-in-bio flow too.

For now, I'm pushing a solution to fix the problem on our onboarding flow.

mpkelly:
Redeployed as https://github.com/Automattic/wp-calypso/pull/77736

PR #75681, opened 2023-04-13
Jetpack Pro Dashboard: Fix issue with assigning incompatible license key to multisite.

### LINKED ISSUE #74689: Jetpack Cloud: Assign licenses page shouldn't allow to assign Jetpack Scan or Jetpack Backup to a multisite since they not supported
### Summary

The assign license page in the Jetpack Cloud shouldn't allow users to select a multisite when assigning a license(Scan or Backup) since they are not supported.

### Steps to reproduce

- Add a multisite
- Go to the [Issue New License](http://jetpack.cloud.localhost:3001/partner-portal/issue-license) page
- Select Jetpack Scan or Backup and click on Select License
- On the Assign license page, it is allowed to select a multisite even though it is not supported.

If it is not supported, I shouldn't be allowed to assign licenses at all that are not supported.

### What actually happened

I was able to assign a license that is not supported to a multisite

Image

jkguidaven:
The assign license page in the Jetpack Cloud shouldn't allow users to select a multisite when assigning a Backup or Scan license since they are not supported. This PR fixes this issue by ensuring multisites are non-selectable when these licenses are selected.
226523218-7b62f664-effd-47e8-b88b-37659cb6c2a2

Related to #74689

## Proposed Changes

* Update the assigned License page to check first if the currently selected licenses are compatible with Multisites. If not, the page will filter all Multisites from the selectable sites.

## Testing Instructions

* Spin up a JN multisite and connect it with your agency account.
* Use the Jetpack cloud live link and go to `/partner-portal/licenses`
* Issue a Backup or Scan license and click **Assign license**
* In the Assign License page, make sure there are no multisites in the list.
Screen Shot 2023-04-13 at 4 26 36 PM

* Go back to `/partner-portal/licenses` and issue another license. (This time it should not be **Backup** or **Scan**)
* In the Assign License page, make sure all sites are listed including multisites.
Screen Shot 2023-04-13 at 4 29 23 PM

## Pre-merge Checklist

### DIFF
--- a/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
@@ -9,7 +9,7 @@ import SearchCard from 'calypso/components/search-card';
 import { SITE_CARDS_PER_PAGE } from 'calypso/jetpack-cloud/sections/partner-portal/assign-license-form/constants';
 import { useAssignMultipleLicenses } from 'calypso/jetpack-cloud/sections/partner-portal/hooks';
 import { addQueryArgs } from 'calypso/lib/url';
-import { areLicenseKeysMultisiteCompatible } from '../utils';
+import { areLicenseKeysAssignableToMultisite } from '../utils';
 import './style.scss';
 
 function setPage( pageNumber: number ): void {
@@ -53,7 +53,8 @@ export default function AssignLicenseForm( {
 	};
 	const [ assignLicenses, isLoading ] = useAssignMultipleLicenses( licenseKeysArray, selectedSite );
 
-	let results = areLicenseKeysMultisiteCompatible( licenseKeysArray )
+	// We need to filter out multisites if the licenses are not assignable to a multisite.
+	let results = areLicenseKeysAssignableToMultisite( licenseKeysArray )
 		? sites
 		: sites.filter( ( site: any ) => ! site.is_multisite );
 
--- a/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
@@ -9,7 +9,7 @@ import SearchCard from 'calypso/components/search-card';
 import { SITE_CARDS_PER_PAGE } from 'calypso/jetpack-cloud/sections/partner-portal/assign-license-form/constants';
 import { useAssignMultipleLicenses } from 'calypso/jetpack-cloud/sections/partner-portal/hooks';
 import { addQueryArgs } from 'calypso/lib/url';
-import { isMultisiteCompatibleLicenseKeys } from '../utils';
+import { areLicenseKeysMultisiteCompatible } from '../utils';
 import './style.scss';
 
 function setPage( pageNumber: number ): void {
@@ -53,7 +53,7 @@ export default function AssignLicenseForm( {
 	};
 	const [ assignLicenses, isLoading ] = useAssignMultipleLicenses( licenseKeysArray, selectedSite );
 
-	let results = isMultisiteCompatibleLicenseKeys( licenseKeysArray )
+	let results = areLicenseKeysMultisiteCompatible( licenseKeysArray )
 		? sites
 		: sites.filter( ( site: any ) => ! site.is_multisite );
 
--- a/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
@@ -9,6 +9,7 @@ import SearchCard from 'calypso/components/search-card';
 import { SITE_CARDS_PER_PAGE } from 'calypso/jetpack-cloud/sections/partner-portal/assign-license-form/constants';
 import { useAssignMultipleLicenses } from 'calypso/jetpack-cloud/sections/partner-portal/hooks';
 import { addQueryArgs } from 'calypso/lib/url';
+import { isMultisiteCompatibleLicenseKeys } from '../utils';
 import './style.scss';
 
 function setPage( pageNumber: number ): void {
@@ -52,7 +53,9 @@ export default function AssignLicenseForm( {
 	};
 	const [ assignLicenses, isLoading ] = useAssignMultipleLicenses( licenseKeysArray, selectedSite );
 
-	let results = sites;
+	let results = isMultisiteCompatibleLicenseKeys( licenseKeysArray )
+		? sites
+		: sites.filter( ( site: any ) => ! site.is_multisite );
 
 	if ( search ) {
 		results = results.filter( ( site: any ) => site.domain.search( search ) !== -1 );
--- a/client/jetpack-cloud/sections/partner-portal/utils.ts
+++ b/client/jetpack-cloud/sections/partner-portal/utils.ts
@@ -203,12 +203,13 @@ export function isJetpackBundle( product: APIProductFamilyProduct | string ) {
 }
 
 /**
- * Whether the license keys are compatible with WP Multisite.
+ * Whether the licenses are assignable to WP multisite. This function uses key prefix to determine
+ * if the license is compatible with multisite.
  *
  * @param {Array<string>} licenseKeys
- * @returns {boolean} indicating if the license keys are compatible with multisite
+ * @returns {boolean} indicating if the license keys are assignable to multisite
  */
-export function areLicenseKeysMultisiteCompatible( licenseKeys: Array< string > ): boolean {
-	// Jetpack Backup and Scan are not compatible with WP Multisite.
-	return licenseKeys.every( ( key ) => ! /^jetpack-(backup|scan)/.test( key ) );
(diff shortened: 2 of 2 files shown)


client/jetpack-cloud/sections/partner-portal/utils.ts
```diff
@@ -201,3 +201,18 @@ export function isJetpackBundle( product: APIProductFamilyProduct | string ) {
...
 	return product.family_slug === 'jetpack-packs';
 }
+
+/**
+ *	Checks if the license keys are compatible with multisite.
+ *
+ * @param licenseKeys
+ * @returns boolean indicating if the license keys are compatible with multisite
+ */
+export function isMultisiteCompatibleLicenseKeys( licenseKeys: string[] ): boolean {
```
jkguidaven: Ideally, we would get the license info from the key and match it by product slug, but this will cost us to delay the UI from rendering the site selections until we resolve the license info by fetching the data from the backend.

This function instead uses the license key prefix to determine what type of product we are trying to match.

manzoorwanijk: I think a better name for this function would be `areLicenseKeysMultisiteCompatible`.

client/jetpack-cloud/sections/partner-portal/utils.ts
```diff
@@ -201,3 +201,18 @@ export function isJetpackBundle( product: APIProductFamilyProduct | string ) {
...
+	const INCOMPATIBLE_LICENSE_KEY_PREFIXES = [ 'jetpack-backup', 'jetpack-scan' ];
+
+	const isIncompatibleLicenseKey = ( licenseKey: string ) =>
+		!! INCOMPATIBLE_LICENSE_KEY_PREFIXES.find( ( prefix ) => licenseKey?.startsWith( prefix ) );
+
+	return ! licenseKeys.find( ( licenseKey ) => isIncompatibleLicenseKey( licenseKey ) );
```
manzoorwanijk: I think it will be faster and more readable if we use [`.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) or [`.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) here instead of [`.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), because `.some()` and `.every()` return early if the condition matches. Also, we won't have to use double negation (`!!`) for a boolean value.

So, we can re-write it like this:

```ts
/**
 * Whether the license keys are compatible with WP Multisite.
 *
 * @param {Array<string>} licenseKeys
 * @returns {boolean} indicating if the license keys are compatible with multisite
 */
export function areLicenseKeysMultisiteCompatible( licenseKeys: Array< string > ): boolean {
	const INCOMPATIBLE_LICENSE_KEY_PREFIXES = [ 'jetpack-backup', 'jetpack-scan' ];

	return licenseKeys.every( ( licenseKey ) => {
		return INCOMPATIBLE_LICENSE_KEY_PREFIXES.every(
			( prefix ) => ! licenseKey?.startsWith( prefix )
		);
	} );
}
```

Or, if you like Regex, we can make it a one-liner

```ts
/**
 * Whether the license keys are compatible with WP Multisite.
 *
 * @param {Array<string>} licenseKeys
 * @returns {boolean} indicating if the license keys are compatible with multisite
 */
export function areLicenseKeysMultisiteCompatible( licenseKeys: Array< string > ): boolean {
	return licenseKeys.every( ( key ) => ! /^jetpack-(backup|scan)/.test( key ) );
}
```

manzoorwanijk: IMHO, the regex one is more readable from left to right:

`.every` licence `key` should pass the `.test` i.e. it does not `!` start with `^` `jetpack-backup` or `jetpack-scan`

jkguidaven: @manzoorwanijk I initially was planning to use regex but thought the list might grow in the future, but I guess it makes sense, for now, to use it for readability. 👍 thanks for the suggestion.

client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
```diff
@@ -52,7 +53,9 @@ export default function AssignLicenseForm( {
 	};
 	const [ assignLicenses, isLoading ] = useAssignMultipleLicenses( licenseKeysArray, selectedSite );
 
-	let results = sites;
+	let results = isMultisiteCompatibleLicenseKeys( licenseKeysArray )
```
manzoorwanijk:
```suggestion
	let results = areLicenseKeysMultisiteCompatible( licenseKeysArray )
```

client/jetpack-cloud/sections/partner-portal/assign-license-form/index.tsx
```diff
@@ -9,6 +9,7 @@ import SearchCard from 'calypso/components/search-card';
 import { SITE_CARDS_PER_PAGE } from 'calypso/jetpack-cloud/sections/partner-portal/assign-license-form/constants';
 import { useAssignMultipleLicenses } from 'calypso/jetpack-cloud/sections/partner-portal/hooks';
 import { addQueryArgs } from 'calypso/lib/url';
+import { isMultisiteCompatibleLicenseKeys } from '../utils';
```
manzoorwanijk:
```suggestion
import { areLicenseKeysMultisiteCompatible } from '../utils';
```

jkguidaven: Thought I had pushed the file but this is fixed. 👍

client/jetpack-cloud/sections/partner-portal/utils.ts
jkguidaven: Yes that sounds better 👍

manzoorwanijk (approved):
I did not test it but the code looks good to me now.

yashwin (commented):
Everything looks good to me.

I have 2 concerns here.

1) This won't work well when multiple licenses are selected(one backup/scan and another other than backup/scan). It'll not show the multisite in this case.
2) Should we disable the multisite selection instead of filtering it, as it might cause some confusion to the user as to why the sites are missing?

We could check with the design team. @ederrengifo, any thoughts here?

jkguidaven:
> 1. This won't work well when multiple licenses are selected(one backup/scan and another other than backup/scan). It'll not show the multisite in this case.

 I assumed it was safer to block multisite selection when at least one of the licenses is either Backup or Scan. In the case of the multiple licenses scenario, does the compatible license only get applied and ignores Backup and Scan? would it make more sense if we add some text instead why was multisite cannot selected?
 
> 2. Should we disable the multisite selection instead of filtering it, as it might cause some confusion to the user as to why the sites are missing?
> 
> We could check with the design team. @ederrengifo, any thoughts here?

This is a good point. It may be confusing to the users. I agree disabling it might be the better approach here instead of hiding. But yes, we can refer to the design team and how to deal with it better.

yashwin:
> I assumed it was safer to block multisite selection when at least one of the licenses is either Backup or Scan. In the case of the multiple licenses scenario, does the compatible license only get applied and ignores Backup and Scan? would it make more sense if we add some text instead why was multisite cannot selected?

I think adding a help text will solve the problem. We could just filter the licenses which are not compatible on a multisite and assign licenses that are only compatible.

> This is a good point. It may be confusing to the users. I agree disabling it might be the better approach here instead of hiding. But yes, we can refer to the design team and how to deal with it better.

Yes 👍

ederrengifo:
@yashwin @jkguidaven I agree. This creates a conflict with multiple licenses. I remember we mapped the case, but as always, we push it "for later" 😅  Maybe it's worth addressing it now, but it would require doing it in a project. I don't think this PR is the best place to do it since we would need to do some research and exploration. 

Besides that, this PR looks good. Is there a way to skip this when you selected multiple licenses? So it doesn't become a blocker

ederrengifo (approved):
LGTM

jkguidaven:
@ederrengifo @yashwin I'm not sure if I understand it correctly. Do we want to close this PR and address the issue as a project, or can we go ahead with the implementation for now and come back later with a proper solution?

yashwin:
@ederrengifo Makes sense 👍 

@jkguidaven - Eder is asking if we can skip this implementation when multiple licenses are selected and implement it only when one license is selected. 

Basically, we could see if the length of licenseKeysArray is equal to 1 and then apply this implementation

jkguidaven:
> @ederrengifo Makes sense 👍
> 
> @jkguidaven - Eder is asking if we can skip this implementation when multiple licenses are selected and implement it only when one license is selected.
> 
> Basically, we could see if the length of licenseKeysArray is equal to 1 and then apply this implementation

Thanks for clarifying it, @yashwin. Since our main concern is on mixed licenses, I slightly modified the implementation to skip filtering when we have mixed licenses. The filtering will still apply if all of the selected licenses are non-compatible. wdyt? 👍

yashwin (approved):
LGTM. Thanks for making these changes.

PR #75716, opened 2023-04-13
Billing History: Show Loading State

pottedmeat:
When updating the [filtered transactions function](https://github.com/Automattic/wp-calypso/pull/74975/commits/89ead7aebc04605605628611ded7e3894a2a0611#diff-d4a9af5c7314b149cffd7b0cc6c1d2fdc538be3e9a6559aa422470fed78db261L68), a regression was introduced where the previous version would return a null value if passed a null value and the updated version will return an empty array. The billing history list has logic that [shows the loading state if the passed transactions property is null](https://github.com/Automattic/wp-calypso/blob/e5fbc3172d9c192f289b47cf1f59101b4c778a7b/client/me/purchases/billing-history/billing-history-list.tsx#L207-L209).

Fixes Automattic/payments-shilling#1516

## Proposed Changes

* Restores the behavior of passing a null value to the transactions property if it received a null value.

## Testing Instructions

* Visit [`/me/purchases/billing`](https://container-mystifying-lederberg.calypso.live/me/purchases/billing) (preferably with an account with a few purchases)
* Observe a loading state
	* Billing history loading state
* You may need to try again with an empty cache/incognito window/deleted Redux store

### DIFF
--- a/client/me/purchases/billing-history/billing-history-list.tsx
+++ b/client/me/purchases/billing-history/billing-history-list.tsx
@@ -271,12 +271,9 @@ export default connect(
 		const transactions = getPastBillingTransactions( state );
 		const filter = getBillingTransactionFilters( state, 'past' );
 		const pageSize = 5;
-		const filteredTransactions = filterTransactions( transactions, filter, siteId );
-		const paginatedTransactions = paginateTransactions(
-			filteredTransactions,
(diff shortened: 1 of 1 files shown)


michaeldcain (approved):
This works well for me.

PR #72952, opened 2023-02-03
Plans 2023: Add notices on the /plans page

claudiucelfilip:
#### Proposed Changes

This restores Notice banners on /plans by copying and integrating the previous functionality.

#### Testing Instructions

* Go to `/plans/{SITE_NAME}?flags=onboarding/onboarding/2023-pricing-grid` and make sure the following banners are visible when the conditions are met:

**Pro-rated credits**
This should show when you already have an annual paid plan:

Screenshot 2023-02-03 at 13 51 44

**Upgrade disabled**
This should show if you're not the owner of the current paid plan site:

Screenshot 2023-02-03 at 13 12 42

**Marketing**
You'll need to sandbox the public API and follow the instructions in `2efed-pb`  to trigger it:

Screenshot 2023-02-03 at 13 09 34

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to https://github.com/Automattic/martech/issues/1462

PR #75725, opened 2023-04-13
Plans 2023 : Plan purchase credits messaging tweak

ddc22:
~Blocked by #76100~
Related to pdgrnI-2iW-p2
### Proposed Changes and test cases

* Fixes following use case bugs which were discussed @ pdgrnI-2iW-p2

The notice has been updated to following and this update is not reflected in the table below.
image

### Use Cases

| # | Use Case      | Existing  | New  |
| -- | ----------- | -----------  | -----------  | 
| 1 | Purchase any monthly plan other than Woocommerce  and look at yearly plans   |  image   |   image   |
| 2 | Purchase any monthly plan other than Woocommerce  and look at monthly plans   |  image    | image |
| 3 | On a free plan change the currency to INR and look at the yearly plans. This has a flat discount for the first year for select currencies.   |  image    | image | 
| 4 | In the [biennial experiment variation](https://github.com/Automattic/wp-calypso/blob/c0aba3a809b43ecce0d6ec109397e2a591cb0ed4/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx#L27-L28) purchase a non WooCommerce monthly plan and look at yearly plans   |  image  | image |
| 5 | In the [biennial experiment variation](https://github.com/Automattic/wp-calypso/blob/c0aba3a809b43ecce0d6ec109397e2a591cb0ed4/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx#L27-L28) purchase a non WooCommerce monthly plan and look at biennial plans   |  image    | image |
| 6 | On the signup flow |  image | image |

## Pre-merge Checklist

Fixes Automattic/martech#1678

### DIFF
--- a/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
+++ b/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
@@ -22,9 +22,9 @@ interface Props {
 	isMonthlyPlan: boolean;
 }
 
-const StrikethroughText = styled.span`
-	color: 'var(--studio-gray-20)';
-	text-decoration: 'line-through';
+export const StrikethroughText = styled.span`
+	color: var( --studio-gray-20 );
+	text-decoration: line-through;
 `;
 
 function usePerMonthDescription( {
--- a/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
+++ b/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
@@ -8,6 +8,7 @@ import {
 	TERM_ANNUALLY,
 } from '@automattic/calypso-products';
 import { formatCurrency } from '@automattic/format-currency';
+import styled from '@emotion/styled';
 import { localize, TranslateResult, useTranslate } from 'i18n-calypso';
 import { FunctionComponent } from 'react';
 import { useSelector } from 'react-redux';
@@ -20,6 +21,11 @@ interface Props {
 	billingPeriod: number | null | undefined;
 	isMonthlyPlan: boolean;
 }
+
+const StrikethroughText = styled.span`
+	color: 'var(--studio-gray-20)';
+	text-decoration: 'line-through';
+`;
 
 function usePerMonthDescription( {
 	isMonthlyPlan,
@@ -63,23 +69,51 @@ function usePerMonthDescription( {
 	}
 
 	if ( ! isMonthlyPlan ) {
-		const maybeDiscountedPrice =
-			planPrices.planDiscountedRawPrice || planPrices.discountedRawPrice || planPrices.rawPrice;
-		const fullTermPriceText =
-			currencyCode && maybeDiscountedPrice
-				? formatCurrency( maybeDiscountedPrice, currencyCode, { stripZeros: true } )
+		const discountedPrice = planPrices.planDiscountedRawPrice || planPrices.discountedRawPrice;
+		const fullTermDiscountedPriceText =
+			currencyCode && discountedPrice
+				? formatCurrency( discountedPrice, currencyCode, { stripZeros: true } )
 				: null;
-
-		if ( fullTermPriceText ) {
+		const rawPrice =
+			currencyCode && planPrices.rawPrice
+				? formatCurrency( planPrices.rawPrice, currencyCode, { stripZeros: true } )
+				: null;
+		if ( fullTermDiscountedPriceText ) {
 			if ( PLAN_ANNUAL_PERIOD === billingPeriod ) {
-				return translate( 'per month, %(fullTermPriceText)s billed annually', {
-					args: { fullTermPriceText },
+				//per month, $96 billed annually $84 for the first year
+
+				return translate(
+					'per month, {{discount}} %(rawPrice)s billed annually{{/discount}} %(fullTermDiscountedPriceText)s for the first year',
+					{
+						args: { fullTermDiscountedPriceText, rawPrice },
+						components: {
+							discount: <StrikethroughText />,
+						},
+					}
+				);
+			}
+
+			if ( PLAN_BIENNIAL_PERIOD === billingPeriod ) {
+				return translate(
+					'per month, {{discount}} %(rawPrice)s billed annually{{/discount}} %(fullTermDiscountedPriceText)s for the first year',
+					{
+						args: { fullTermDiscountedPriceText, rawPrice },
+						components: {
+							discount: <StrikethroughText />,
+						},
+					}
+				);
+			}
+		} else if ( rawPrice ) {
+			if ( PLAN_ANNUAL_PERIOD === billingPeriod ) {
+				return translate( 'per month, %(rawPrice)s billed annually', {
+					args: { rawPrice },
 				} );
 			}
 
 			if ( PLAN_BIENNIAL_PERIOD === billingPeriod ) {
-				return translate( 'per month, %(fullTermPriceText)s billed every two years', {
-					args: { fullTermPriceText },
+				return translate( 'per month, %(rawPrice)s billed every two years.', {
+					args: { rawPrice },
 				} );
 			}
 		}
@@ -97,7 +131,7 @@ const PlanFeatures2023GridBillingTimeframe: FunctionComponent< Props > = ( props
 	if ( isWpcomEnterpriseGridPlan( planName ) ) {
 		return (
 			<div className="plan-features-2023-grid__vip-price">
-				{ translate( 'Starts at {{b}}%(price)s{{/b}} yearly.', {
+				{ translate( 'Starts at {{b}}%(price)s{{/b}} yearly', {
 					args: { price },
 					components: { b: <b /> },
 					comment: 'Translators: the price is in US dollars for all users (US$25,000)',
--- a/client/my-sites/plan-features-2023-grid/index.tsx
+++ b/client/my-sites/plan-features-2023-grid/index.tsx
@@ -51,6 +51,7 @@ import { retargetViewPlans } from 'calypso/lib/analytics/ad-tracking';
 import { planItem as getCartItemForPlan } from 'calypso/lib/cart-values/cart-items';
 import { FeatureObject, getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
 import scrollIntoViewport from 'calypso/lib/scroll-into-viewport';
+import { useIsPlanUpgradeCreditVisible } from 'calypso/my-sites/plan-features-2023-grid/hooks/use-is-plan-upgrade-credit-visible';
 import { PlanTypeSelectorProps } from 'calypso/my-sites/plans-features-main/components/plan-type-selector';
 import { recordTracksEvent } from 'calypso/state/analytics/actions';
 import { getCurrentUserCurrencyCode } from 'calypso/state/currency-code/selectors';
@@ -132,6 +133,7 @@ type PlanFeatures2023GridConnectedProps = {
 	planTypeSelectorProps: PlanTypeSelectorProps;
 	manageHref: string;
 	selectedSiteSlug: string | null;
+	isPlanUpgradeCreditEligible: boolean;
 };
 
 type PlanFeatures2023GridType = PlanFeatures2023GridProps &
@@ -332,6 +334,7 @@ export class PlanFeatures2023Grid extends Component<
 						className="plan-features-2023-grid__plan-comparison-grid-container"
 					>
 						<PlanComparisonGrid
+							siteId={ siteId }
 							planTypeSelectorProps={ planTypeSelectorProps }
 							planProperties={ planProperties }
 							intervalType={ intervalType }
@@ -494,7 +497,7 @@ export class PlanFeatures2023Grid extends Component<
 	}
 
 	renderPlanPrice( planPropertiesObj: PlanProperties[], options?: PlanRowOptions ) {
-		const { isReskinned, isLargeCurrency, translate } = this.props;
+		const { isReskinned, isLargeCurrency, translate, isPlanUpgradeCreditEligible } = this.props;
 
 		return planPropertiesObj
 			.filter( ( { isVisible } ) => isVisible )
@@ -515,6 +518,7 @@ export class PlanFeatures2023Grid extends Component<
 					>
 						{ ! hasNoPrice && (
 							<PlanFeatures2023GridHeaderPrice
+								isPlanUpgradeCreditEligible={ isPlanUpgradeCreditEligible }
 								planProperties={ properties }
 								is2023OnboardingPricingGrid={ true }
 								isLargeCurrency={ isLargeCurrency }
@@ -901,7 +905,7 @@ const ConnectedPlanFeatures2023Grid = connect(
 		const purchaseId = getCurrentPlanPurchaseId( state, siteId );
 		const selectedSiteSlug = getSiteSlug( state, siteId );
 
-		const planProperties: PlanProperties[] = plans.map( ( plan: string ) => {
+		const planProperties: PlanProperties[] = plans.map( ( plan: PlanSlug ) => {
 			let isPlaceholder = false;
 			const planConstantObj = applyTestFiltersToPlansList( plan, undefined );
 			const planProductId = planConstantObj.getProductId();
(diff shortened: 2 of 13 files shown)


client/my-sites/plan-features-2023-grid/components/badge.tsx
chriskmnds, line 1: I think the name is too generic, in contrast to the component that is too specific for the specific use case (being green, bold, and the other styling pretty much) e.g. we also have the highlight "badges" in the same grid, which are not rendered by this component. I would prefer if we can generalize the component and serve both, or maybe not create a separate component file and keep this closer to the source

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1180,6 +1148,47 @@ const ConnectedPlanFeatures2023Grid = connect(
...
+
+		const higherPlanAvailable = () => {
+			const currentPlan = currentSitePlan?.productSlug;
+			const highestPurchasablePlan = planProperties
+				.filter( ( { planName } ) => planName !== PLAN_ENTERPRISE_GRID_WPCOM )
+				.filter( ( { availableForPurchase } ) => availableForPurchase )
+				.pop();
+			const { planName: highestPurchasablePlanName } = highestPurchasablePlan ?? {};
+			return currentPlan !== highestPurchasablePlanName;
+		};
```
chriskmnds: Not sure I follow why we moved these into the loop. was it necessary? 🤔

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1180,6 +1148,47 @@ const ConnectedPlanFeatures2023Grid = connect(
...
+
+		const isDiscountNoticeVisible = hasDiscountNotice();
+		const isFreePlanActive = isFreePlan( sitePlan?.product_slug ?? '' );
+		const isWPPlansPage = sitePlan && ! isJetpackNotAtomic && ! isInSignup;
+		const isHigherPlanAvailable = higherPlanAvailable();
+		const bannerContainer = getBannerContainer();
+
+		const hasPlaceholdersFlag = hasPlaceholders( planProperties );
```
chriskmnds: does this need to be put into a variable? it's only used right below

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1180,6 +1148,47 @@ const ConnectedPlanFeatures2023Grid = connect(
...
+			const highestPurchasablePlan = planProperties
+				.filter( ( { planName } ) => planName !== PLAN_ENTERPRISE_GRID_WPCOM )
+				.filter( ( { availableForPurchase } ) => availableForPurchase )
+				.pop();
+			const { planName: highestPurchasablePlanName } = highestPurchasablePlan ?? {};
+			return currentPlan !== highestPurchasablePlanName;
+		};
+
+		const isDiscountNoticeVisible = hasDiscountNotice();
+		const isFreePlanActive = isFreePlan( sitePlan?.product_slug ?? '' );
```
chriskmnds: the name `isFreePlanActive` is not clear to me. what does it mean (that the site is on a free plan?)

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1180,6 +1148,47 @@ const ConnectedPlanFeatures2023Grid = connect(
...
+		const isWPPlansPage = sitePlan && ! isJetpackNotAtomic && ! isInSignup;
```
chriskmnds: I think this is another name/variable that tries to convey too much. it takes me a few attempts to understand, hence find indirect for the most part. can we split this apart?

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1180,6 +1148,47 @@ const ConnectedPlanFeatures2023Grid = connect(
...
+		const hasPlaceholdersFlag = hasPlaceholders( planProperties );
+		const hideCreditNotice =
+			! isWPPlansPage ||
+			isFreePlanActive ||
+			isDiscountNoticeVisible ||
+			hasPlaceholdersFlag ||
+			! canUserPurchasePlan ||
+			! bannerContainer ||
+			! planCredits ||
+			! isHigherPlanAvailable;
```
chriskmnds: This is quite a long expression that I'd rather if it lives somewhere else or that we break apart. the interplay between "hiding the credit notice" (`hideCreditNotice`) and "the discount notice being visible" (`isDiscountNoticeVisible`) is also hidden within, and we are also passing these along as separate props. 

I don't have solutions, but it's hard to digest all this logic. Maybe see if anything can be cleaned up and we can re-review. 🙇

chriskmnds (commented):
Thanks @jdc91 . left some code comments, but haven't tested anything yet :-)

client/my-sites/plan-features-2023-grid/components/badge.tsx
ddc22: Ah yes this makes sense @chriskmnds I actually attempted to genaralize this with popular-badge but it was built on top of plan pill and felt like it would require more work to change it. I might just move this component closer to the source as you suggested. I will try to pick it up later for generalising!

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: Ah yes let me rename that to `isCurrentSiteOnFreeplan`

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: I did this to reduce some complexity from the enormous expression below 😁.

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: I guess you mean the connect function here?
I wanted lift the logic required to deduce the state of `hideCreditNotice` to the connect function. This is because this state needed to be used at 2 different parts of the component. I know I could have just moved it into a function inside the class and called it. However I was planning on doing a refactor of possibly getting rid of this React.portal implementation and consolidating the notice logic higher up in the tree into it's own component. And the logic related to it will be close by instead of in multiple functions inside the above class. Sorry I guess I could have been a bit more clear in my pull request.

Do you recommend reverting to doing all of this inside the class? 🤔

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: This is actually cleaning up a loosely defined set of conditions from before.
The initial set of flags were taken from below.

https://github.com/Automattic/wp-calypso/blob/2bd7e469b071c51d3ddc0f69805439e0b29e795f/client/my-sites/plan-features-2023-grid/index.tsx#L1003-L1012

The `isShowPlanCreditsApplied` above further expands into another set of flags from below

https://github.com/Automattic/wp-calypso/blob/2bd7e469b071c51d3ddc0f69805439e0b29e795f/client/my-sites/plan-features-2023-grid/index.tsx#L1003-L1012

I do see a cleanup for this which involves moving all of this notice logic into it's own component. Some complexity also arises from using `React.portal` where we are directly mutating the DOM. Cleaning this up can potentially eliminate `bannerContainer` which does a DOM check to see whether the mounting element exists. 
https://github.com/Automattic/wp-calypso/blob/2bd7e469b071c51d3ddc0f69805439e0b29e795f/client/my-sites/plan-features-2023-grid/index.tsx#L874-L876

I was initially worried that tampering too much will break something unrelated. But looks like this is a pretty isolated change done at #72952 (Which was essentially copied from the existing plans page) and the code volume does seem to be manageable. Let me take another stab at this refactor!

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -55,6 +55,7 @@ import { planItem as getCartItemForPlan } from 'calypso/lib/cart-values/cart-ite
 import { getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
 import scrollIntoViewport from 'calypso/lib/scroll-into-viewport';
 import { PlanTypeSelectorProps } from 'calypso/my-sites/plans-features-main/components/plan-type-selector';
+import { usePlanUpgradeCreditsDisplay } from 'calypso/my-sites/plans-features-main/hooks/use-plan-upgrade-credits-display';
```
chriskmnds: Hey @jdc91. I think let's move `usePlanUpgradeCreditsDisplay` under `my-sites/plan-features-2023-grid/hooks`

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7963339

chriskmnds (commented):
Looks great code-wise. There's a lot to test though, which is probably what you need help with :-) 

I will carry on Monday and give a final confidence check

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -63,23 +63,59 @@ function usePerMonthDescription( {
...
+
+				return translate(
+					'per month, {{discount}} %(rawPrice)s billed annually{{/discount}} %(fullTermDiscountedPriceText)s for the first year',
+					{
+						args: { fullTermDiscountedPriceText, rawPrice },
+						components: {
+							discount: (
+								<span
+									style={ { color: 'var(--studio-gray-20)', textDecoration: 'line-through' } }
+								/>
```
aneeshd16: Nit: Maybe this can be a small styled component? The same component can be used in L101 as well.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -63,23 +63,59 @@ function usePerMonthDescription( {
...
+							),
+						},
+					}
+				);
+			}
+		} else if ( rawPrice ) {
 			if ( PLAN_ANNUAL_PERIOD === billingPeriod ) {
-				return translate( 'per month, %(fullTermPriceText)s billed annually', {
-					args: { fullTermPriceText },
+				return translate( 'per month, %(rawPrice)s billed annually.', {
```
aneeshd16: I'm not sure if there should be a period at the end of this string since we are not using periods anywhere in the billing timeframes.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -63,23 +63,59 @@ function usePerMonthDescription( {
...
+					args: { rawPrice },
 				} );
 			}
 
 			if ( PLAN_BIENNIAL_PERIOD === billingPeriod ) {
-				return translate( 'per month, %(fullTermPriceText)s billed every two years', {
-					args: { fullTermPriceText },
+				return translate( 'per month, %(rawPrice)s billed every two years.', {
```
aneeshd16, line 115: Same as above.

aneeshd16 (commented):
Tested and works well! I left some code feedback below:

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -51,6 +51,7 @@ import { retargetViewPlans } from 'calypso/lib/analytics/ad-tracking';
 import { planItem as getCartItemForPlan } from 'calypso/lib/cart-values/cart-items';
 import { getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
 import scrollIntoViewport from 'calypso/lib/scroll-into-viewport';
+import { usePlanUpgradeCreditsDisplay } from 'calypso/my-sites/plan-features-2023-grid/hooks/use-plan-upgrade-credits-display';
```
chriskmnds: Just noticed that `usePlanUpgradeCreditsDisplay` is missing any documentation. we probably want to add something, a brief description maybe. 

The discrepancy between that and `usePlanUpgradeCredits` I think should be clearer. In fact, considering that `usePlanUpgradeCredits` is only used in `usePlanUpgradeCreditsDisplay`, then I'm not sure if it should even be living and exported separately. It's a little confusing with both living side by side. Maybe just a single `usePlanUpgradeCredits` might do. anw - probably NIT.

client/my-sites/plans-features-main/components/plan-notice.tsx
```diff
@@ -83,14 +84,21 @@ export default function PlanNotice( {
 				status="is-success"
 			>
 				{ translate(
-					'You have {{b}}%(amountInCurrency)s{{/b}} of pro-rated credits available from your current plan. ' +
-						'Apply those credits towards an upgrade before they expire!',
+					'We’ve applied the {{b}}%(amountInCurrency)s{{/b}} {{a}}upgrade credit{{/a}} from your current plan as a deduction to your new plan, below. This remaining credit will be applied at checkout if you upgrade today!',
```
chriskmnds, line 86: `We’ve` to `We have` maybe? a little less informal 🤔

chriskmnds (approved):
Giving this a go from my end as it tests out on the scenarios outlined. Left some comments, and also respecting @aneeshd16 's current review in progress. So trust we address whatever needs to be addressed :-)

chriskmnds:
> Tested and works well! I left some code feedback below:

Sorry, @aneeshd16 - didn't mean to intervene with the ongoing review. Gave this a go as I tested it out, but trust we'll address the comments :-)

aneeshd16:
> Sorry, @aneeshd16 - didn't mean to intervene with the ongoing review. Gave this a go as I tested it out, but trust we'll address the comments :-)

No worries! My suggestions are only string changes. 🙂

client/my-sites/plans-features-main/components/plan-notice.tsx
jeyip: @chriskmnds There are a number of examples where the tone WordPress uses is more conversational -- do we typically prefer more formal language for our plans pricing optimizations?

Screenshot 2023-05-10 at 7 29 02 AM

Screenshot 2023-05-10 at 7 32 39 AM

Screenshot 2023-05-10 at 7 31 52 AM

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7963339).

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: Yes you are correct, this is indeed confusing 😅 .
I think this mainly stems from the fact that my attempt here was to unify hooks and subsequently resulting bad naming on my part.
`usePlanUpgradeCredits` is used in several places and the main purpose of that hook is to calculate the plan upgrade credit. But `usePlanUpgradeCreditsDisplay` is complex UI logic that decides if the credits should be displayed. Maybe dividing these two hooks instead of unifying them and giving a proper name and some explanation should fix the confusion.

client/my-sites/plans-features-main/components/plan-notice.tsx
ddc22: Yes I think the objective here was to be conversational. Can you weigh in here @tanjoymor  ? 🙂

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: I have changed the names to be more meaningful and added a comment :)
https://github.com/Automattic/wp-calypso/blob/50a0b38f7bc047b728db47d9b580d2130f4c4bb1/client/my-sites/plans-features-main/components/plan-notice.tsx#L10-L11

client/my-sites/plans-features-main/components/plan-notice.tsx
chriskmnds: Informal is fine in any case :-)

chriskmnds:
Final confidence check. Everything tests out from the description, except:

> In the [biennial experiment variation](https://github.com/Automattic/wp-calypso/blob/c0aba3a809b43ecce0d6ec109397e2a591cb0ed4/client/my-sites/plans-features-main/term-experiment-plan-type-selector.tsx#L27-L28) purchase a non WooCommerce monthly plan and look at yearly plans

I do not see the "one time discount" badge in `/start` or `/plans` when on any plan (free/yearly/monthly). Not sure if I'm missing something though or lost myself in all these test cases.

p.s. seeing how many cases we have to verify manually, it does sound like a suite of unit tests would do that better for us. Similar to what we ended up doing for the upgrade actions that were too many to account for. Maybe something for a follow-up if it makes sense

ddc22:
@chriskmnds Thanks for the final confidence check. Since the multi year plans are currently not active let's ignore those and pick them up later 👍 

Issue tracked here Automattic/martech#1742

jeyip:
> I do not see the "one time discount" badge in /start or /plans when on any plan (free/yearly/monthly). Not sure if I'm missing something though or lost myself in all these test cases.

Could you elaborate on when we expect to see the `One time discount` badge @chriskmnds? Currently, in production, the one time discount badge ( I'm using the INR currency ) is only displayed for yearly plans.



In my local dev environment when I enable the current 1yr/2yr toggle logic, I _also_ see the `One time discount` badge  displayed only for 1yr plans. Do we expect 2yr plans to be currency discounted?



> p.s. seeing how many cases we have to verify manually, it does sound like a suite of unit tests would do that better for us. Similar to what we ended up doing for the upgrade actions that were too many to account for. Maybe something for a follow-up if it makes sense

I'm happy to draft an issue -- to clarify, we'd like to test the variations of the billing timeframe and badges we'll see given different inputs? I'm also happy to write up said tests for the billing timeframe and header-price components depending on prioritization of work this week 🙂

PR #75747, opened 2023-04-14
Site Assembler: Update colors and fonts track events

arthur791004:
Related to https://github.com/Automattic/dotcom-forge/issues/2159

## Proposed Changes

* Change the `title` prop to `color_variation_title` or `font_variation_title` to make it clear. I think we don't need to know the previously selected variation when the user clicks other variations. But if we want, we can just rename it to something like `next_color_variation_title`.
* Add the props called `color_variation_type` and `font_variation_type` to track whether the selected variation is free or premium
* Add a new event to track the `adding a header pattern` button

Screenshot 2023-04-14 at 2 14 20 PM

Screenshot 2023-04-14 at 2 14 59 PM

image

## Testing Instructions

* Go to /setup?siteSlug=<your_free_site>
* On the Design Picker screen, scroll to the bottom and select "Start designing"
  * Select `Colors`
  * Click the `adding a header pattern` text on the large preview, and verify `calypso_signup_pattern_assembler_large_preview_add_header_button_click` is fired
  * Select any color variation
  * Verify the track event, `calypso_signup_pattern_assembler_screen_colors_preview_click`, is fired with both `color_variation_title` and `color_variation_type`
  * Go back to select `Fonts`
  * Select any font variation
  * Verify the track event, `calypso_signup_pattern_assembler_screen_fonts_preview_click`, is fired with both `font_variation_title` and `font_variation_type`

## Pre-merge Checklist

### DIFF
--- a/packages/global-styles/src/constants.ts
+++ b/packages/global-styles/src/constants.ts
@@ -2,6 +2,6 @@ export const STYLE_PREVIEW_WIDTH = 248;
 export const STYLE_PREVIEW_HEIGHT = 152;
 export const STYLE_PREVIEW_COLOR_SWATCH_SIZE = 32;
 
-export const DEFAULT_VARIATION_TITLE = 'default';
+export const DEFAULT_GLOBAL_STYLES_VARIATION_TITLE = 'default';
 export const DEFAULT_GLOBAL_STYLES_VARIATION_SLUG = 'default';
 export const SYSTEM_FONT_SLUG = 'system-font';
--- a/packages/global-styles/src/constants.ts
+++ b/packages/global-styles/src/constants.ts
@@ -2,5 +2,6 @@ export const STYLE_PREVIEW_WIDTH = 248;
 export const STYLE_PREVIEW_HEIGHT = 152;
 export const STYLE_PREVIEW_COLOR_SWATCH_SIZE = 32;
 
+export const DEFAULT_VARIATION_TITLE = 'default';
 export const DEFAULT_GLOBAL_STYLES_VARIATION_SLUG = 'default';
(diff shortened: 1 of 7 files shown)


packages/global-styles/src/constants.ts
```diff
@@ -2,5 +2,6 @@ export const STYLE_PREVIEW_WIDTH = 248;
 export const STYLE_PREVIEW_HEIGHT = 152;
 export const STYLE_PREVIEW_COLOR_SWATCH_SIZE = 32;
 
+export const DEFAULT_VARIATION_TITLE = 'default';
```
taipeicoder: [nit] Could we make `DEFAULT_VARIATION_TITLE` and `DEFAULT_GLOBAL_STYLES_VARIATION_SLUG` naming consistent? Either one is fine with me!

taipeicoder (approved):
A minor naming suggestion, but otherwise it looks good 👍

packages/global-styles/src/constants.ts
arthur791004: Sure!

arthur791004: Updated! I renamed it to `DEFAULT_GLOBAL_STYLES_VARIATION_TITLE`

autumnfjeld:
>  I think we don't need to know the previously selected variation when the user clicks other variations. 

@arthur791004   I agree.   I can't think of anything we'd do with that date.  It's too minutia. 

It's only the final selection that is interesting.   And maybe it would be interesting to know if users click  say only one or two  variation or if most users click say 80-90% of the color swatches.   So this  would be a count value.   But no need to do this now.  I'm just thinking about if it is useful.

PR #75766, opened 2023-04-14
Subscription manager: add `useCacheKey` hook

TimBroddin:
## Proposed Changes

This PR adds an useCacheKey hook, that returns the cache key entered, concatenated with 

```js
const useCacheKey = ( key: string[] ) => {
	const { id, isLoggedIn } = useIsLoggedIn();

	return [ ...key, isLoggedIn ? 'logged-in' : 'not-logged-in', id ? id : '' ];
};
```

It also changes the return signature for `useIsLoggedIn` from `boolan` to `{id: number|undefined, isLoggedIn: boolean }`.

Finally, it also fixes a bug where the invalidation of the Site subscriptions list would fail. useInfiniteQuery stores this list in the query cache in it's raw paged form. To make optimistic updates easier on this paged cache results, this PR also introduces a `applyCallbackToPages` function.

## Testing Instructions

1. Apply this PR and make sure a valid subkey is set
2. Visit all three tabs
3. Change delivery frequency of a site subscription, change should be instant
4. Delete a site subscription, change should be instant in both tab count & site list
5. Do a good code review because this PR touches a lot.

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
+++ b/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
@@ -1,7 +1,7 @@
 import { translate } from 'i18n-calypso';
 import { useMutation, useQueryClient } from 'react-query';
 import { callApi } from '../helpers';
-import { useIsLoggedIn } from '../hooks';
+import { useCacheKey, useIsLoggedIn } from '../hooks';
 import type { SubscriptionManagerUserSettings, EmailSettingsAPIResponse } from '../types';
 
 type MutationContext = {
@@ -11,7 +11,7 @@ type MutationContext = {
 const useUserSettingsMutation = () => {
 	const { isLoggedIn } = useIsLoggedIn();
 	const queryClient = useQueryClient();
-	const emailSettingsCacheKey = [ 'read', 'email-settings' ];
+	const emailSettingsCacheKey = useCacheKey( [ 'read', 'email-settings' ] );
 	return useMutation< SubscriptionManagerUserSettings, Error, SubscriptionManagerUserSettings >(
 		async ( data: SubscriptionManagerUserSettings ) => {
 			const { settings } = await callApi< EmailSettingsAPIResponse >( {
--- a/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
+++ b/packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
@@ -39,7 +39,7 @@ const useUserSettingsMutation = () => {
 				const previousSettings =
 					queryClient.getQueryData< SubscriptionManagerUserSettings >( emailSettingsCacheKey );
 
-				queryClient.setQueryData< SubscriptionManagerUserSettings >( [ 'read', 'email-settings' ], {
+				queryClient.setQueryData< SubscriptionManagerUserSettings >( emailSettingsCacheKey, {
 					...previousSettings,
 					...data,
 				} );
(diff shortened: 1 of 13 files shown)


packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
```diff
@@ -34,31 +35,29 @@ const useUserSettingsMutation = () => {
...
-				] );
+				await queryClient.cancelQueries( emailSettingsCacheKey );
+				const previousSettings =
+					queryClient.getQueryData< SubscriptionManagerUserSettings >( emailSettingsCacheKey );
 
-				queryClient.setQueryData< SubscriptionManagerUserSettings >(
-					[ 'read', 'email-settings', isLoggedIn ],
-					{ ...previousSettings, ...data }
-				);
+				queryClient.setQueryData< SubscriptionManagerUserSettings >( [ 'read', 'email-settings' ], {
```
mashikag: I think you meant to use the `emailSettingsCacheKey` instead of `[ 'read', 'email-settings' ]` as the key in this cache update? 🤔

packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
```diff
@@ -9,8 +9,9 @@ type MutationContext = {
 };
 
 const useUserSettingsMutation = () => {
-	const isLoggedIn = useIsLoggedIn();
+	const { isLoggedIn } = useIsLoggedIn();
 	const queryClient = useQueryClient();
+	const emailSettingsCacheKey = [ 'read', 'email-settings' ];
```
mashikag: Don't we want to use the `useCacheKey` hook for this too?

packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
TimBroddin: Yes, nice catch!

packages/data-stores/src/reader/mutations/use-user-settings-mutation.ts
TimBroddin: Indeed, I already updated it in the previous commit 😀

packages/data-stores/src/reader/types/index.ts
```diff
@@ -44,6 +44,13 @@ type SiteSubscriptionDeliveryMethods = {
 	};
 };
 
+export type PagedQueryResult< TDataType, TKey extends string > = {
```
mashikag, line 47: I suppose we could use `InfiniteData` from `react-query` but yours is a bit more precise to our case. 👍

packages/data-stores/src/reader/queries/use-site-subscriptions-query.ts
```diff
@@ -56,15 +68,13 @@ const useSiteSubscriptionsQuery = ( {
...
 	// Flatten all the pages into a single array containing all subscriptions
 	const flattenedData = data?.pages?.map( ( page ) => page.subscriptions ).flat();
-	// Transform the dates into Date objects
-	const transformedData = flattenedData?.map( ( subscription ) => ( {
-		...subscription,
-		last_updated: new Date( subscription.last_updated ),
-		date_subscribed: new Date( subscription.date_subscribed ),
-	} ) );
 
 	return {
```
mashikag, line 72: In case we notice some performance issues, we should keep in mind that the results here are not memoized. We iterate over all items in `flattenedData` 3 times (once above, when we initialize the const itself). I think it is good for now, though.

mashikag (approved):
The code looks good, and fixes the addressed issues. 👌

PR #75777, opened 2023-04-14
Checkout: remove annual features from the "what's included" list when picking a monthly plan on tailored flows

zaguiini:
Related to https://github.com/Automattic/wp-calypso/pull/75679.

## Proposed Changes

The tailored flows' checkout sidebar currently includes annual plan features even though the picked plan is monthly-termed. This PR fixes that.

## Testing Instructions

Test the Link-in-bio flow (`/setup/link-in-bio`), Hosting onboarding flow (`/start/hosting`), and regular onboarding flow (`/start`):

1. Start the flow and verify that switching to the monthly plans view strikes the annual features (ss 1);
2. Proceed to checkout and check that the annual features (e.g. free domain included) are not on the sidebar list (ss 2);
3. Start the flow again and check that the annual features appear on the list if you select an annual plan.

## Screenshots

| 1 | 2 |
| - | - |
| image | image |

### DIFF
--- a/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
@@ -39,7 +39,7 @@ export default function getFlowPlanFeatures(
 	}
 
 	const highlightedFeatures = getHighlightedFeatures( flowName, planConstantObj );
-	let featuresObject = getPlanFeaturesObject( featureAccessor );
+	let featuresObject = getPlanFeaturesObject( featureAccessor() );
 
 	if ( isMonthly( planConstantObj.getStoreSlug() ) ) {
 		const annualOnlyFeatures = planConstantObj.getAnnualPlansOnlyFeatures?.() ?? [];
--- a/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
@@ -1,4 +1,4 @@
-import { getPlan } from '@automattic/calypso-products';
+import { applyTestFiltersToPlansList, getPlan, isMonthly } from '@automattic/calypso-products';
 import { ResponseCartProduct } from '@automattic/shopping-cart';
 import { getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
 import {
@@ -8,15 +8,21 @@ import {
 
 export default function getFlowPlanFeatures(
 	flowName: string,
-	plan: ResponseCartProduct | undefined
+	product: ResponseCartProduct | undefined
 ) {
-	const productSlug = plan?.product_slug;
+	const productSlug = product?.product_slug;
 
 	if ( ! productSlug ) {
 		return [];
 	}
 
-	const planConstantObj = getPlan( productSlug );
+	const plan = getPlan( productSlug );
+
+	if ( ! plan ) {
+		return [];
+	}
+
+	const planConstantObj = applyTestFiltersToPlansList( plan, undefined );
 
 	if ( ! planConstantObj ) {
 		return [];
@@ -33,7 +39,19 @@ export default function getFlowPlanFeatures(
 	}
 
 	const highlightedFeatures = getHighlightedFeatures( flowName, planConstantObj );
-	return getPlanFeaturesObject( featureAccessor ).map( ( feature ) => {
+	let featuresObject = getPlanFeaturesObject( featureAccessor );
+
+	if ( isMonthly( planConstantObj.getStoreSlug() ) ) {
+		const annualOnlyFeatures = planConstantObj.getAnnualPlansOnlyFeatures?.() ?? [];
+
+		if ( annualOnlyFeatures.length > 0 ) {
+			featuresObject = featuresObject.filter( ( feature ) => {
+				return ! annualOnlyFeatures.includes( feature.getSlug() );
+			} );
+		}
+	}
+
+	return featuresObject.map( ( feature ) => {
 		return {
 			...feature,
 			isHighlightedFeature: highlightedFeatures.includes( feature.getSlug() ),
--- a/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
@@ -1,4 +1,8 @@
-import { applyTestFiltersToPlansList, getPlan, isMonthly } from '@automattic/calypso-products';
+import {
+	applyTestFiltersToPlansList,
+	getPlan,
+	FEATURE_CUSTOM_DOMAIN,
+} from '@automattic/calypso-products';
 import { ResponseCartProduct } from '@automattic/shopping-cart';
 import { getPlanFeaturesObject } from 'calypso/lib/plans/features-list';
 import {
@@ -8,7 +12,10 @@ import {
 
 export default function getFlowPlanFeatures(
 	flowName: string,
-	product: ResponseCartProduct | undefined
+	product: ResponseCartProduct | undefined,
+	hasDomainsInCart: boolean,
+	hasRenewalInCart: boolean,
+	nextDomainIsFree: boolean
 ) {
 	const productSlug = product?.product_slug;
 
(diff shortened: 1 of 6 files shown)


DavidRothstein:
Thanks for catching this - it definitely seems like a major problem!

This fix may work, but I looked into how this functionality all works and it seems like these custom signup flows use a forked version of the checkout feature list (originally added in https://github.com/Automattic/wp-calypso/pull/67211): https://github.com/Automattic/wp-calypso/blob/cefd7b1eb96bfad6521339241aa8a150b0932a99/client/my-sites/checkout/composite-checkout/components/wp-checkout-order-summary.tsx#L177-L181

And you can see how the standard list of checkout features uses that `nextDomainIsFree` property (as well as some other things) to ultimately decide whether to display the "Free domain for one year" text here: https://github.com/Automattic/wp-calypso/blob/cefd7b1eb96bfad6521339241aa8a150b0932a99/client/my-sites/checkout/composite-checkout/lib/get-plan-features.ts#L22

But the separate version of the checkout feature list for these signup flows doesn't.

So I think a better fix would reuse as much of the existing code as possible -- at the very least, making use of the `nextDomainIsFree` property, but ideally even more of it.  For example, when you have an **annual** plan in the cart and also a domain, regular checkout looks like this and highlights your specific free domain:

free-domain

But the custom signup flows still just say "Free domain for one year".

I'm going to add @Automattic/vertex here too, since I think they might be the ones who maintain these flows.

mpkelly:
> So I think a better fix would reuse as much of the existing code as possible -- at the very least, making use of the `nextDomainIsFree` property, but ideally even more of it.

I agree with this, but I think it's a different refactoring PR, whereas this is a bug-fix PR. I went to implement your changes and realised the different flows have already diverged quite a bit. There are two separate functions, `getPlanFeatures` and `getFlowPlanFeatures`, plus some different components for rendering them. So there is a bit more work to do than passing in `nextDomainIsFree`. 

Would you be ok to approve this and we can create a separate ticket for the refactoring? @DavidRothstein

DavidRothstein:
@mpkelly In my opinion using `nextDomainIsFree` is **part** of the bug fix, since it's the property that is guaranteed to correctly tell you whether or not the domain is free.  Anything else is kind of just guessing.

I agree that any refactoring beyond that could be deferred to a separate issue, though.

Wouldn't something like this work, as a limited fix? https://github.com/Automattic/wp-calypso/pull/76391 (it seemed to handle everything correctly in my brief testing).

To be clear, that only handles the domain situation specifically (not other annual-only features), so probably it needs to be merged in with the fix in this pull request, and I think they should complement each other.

I won't be available to work on this more myself, but feel free to take that code and run with it.

mpkelly:
I have updated this based on @DavidRothstein's PR https://github.com/Automattic/wp-calypso/pull/76391. 

| Annual with domain  | Annual no domain | Monthly with domain  | Monthly no domain |
| ------------- | ------------- |  ------------- | ------------- |
| Screenshot 2566-04-28 at 18 09 07  | Screenshot 2566-04-28 at 18 09 42  | Screenshot 2566-04-28 at 18 10 35 | 
Screenshot 2566-04-28 at 18 12 03 |

Please can someone review and retest? It should look like the above for `setup/link-in-bio/intro` and `setup/newsletter/newsletterSetup`, and be unchanged elsewhere.

zaguiini:
@Automattic/vertex could you please help us test this PR? Thanks!

pottedmeat (approved):
Code looks good and tests function as expected.

I had a hard time drawing the line from the removal of the term check in `hostingFeatures` until I realized that `hostingFeatures` are only used in the case of a hosting flow which will always route through the `getFlowPlanFeatures` function.

Something to note is that the link in bio workflow strikes out the domain name even in the case of a non-custom domain:

image

zaguiini:
Thanks @pottedmeat. This looks like a different issue, though... For some reason, `domainName` is being passed to `FEATURE_CUSTOM_DOMAIN.getTitle()` despite not being a custom domain. https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/plans/features-list.tsx#LL617C41-L617C41

If we want to call this function even with the free domain, we might want to check whether the TLD is `.wordpress.com` and display the "Free domain for one year" copy instead.

I'm thinking of tackling this in a follow-up PR.

client/my-sites/checkout/composite-checkout/lib/get-flow-plan-features.ts
```diff
@@ -33,10 +46,15 @@ export default function getFlowPlanFeatures(
...
 
 	const highlightedFeatures = getHighlightedFeatures( flowName, planConstantObj );
-	return getPlanFeaturesObject( featureAccessor() ).map( ( feature ) => {
-		return {
-			...feature,
-			isHighlightedFeature: highlightedFeatures.includes( feature.getSlug() ),
-		};
-	} );
+	const showFreeDomainFeature = ! hasDomainsInCart && ! hasRenewalInCart && nextDomainIsFree;
+	return getPlanFeaturesObject( ( featureAccessor as () => string[] )() )
```
sirbrillig, line 50: I believe this line caused the fatal errors that resulted in [the revert](https://github.com/Automattic/wp-calypso/pull/77317) of this PR.

`featureAccessor` comes from calling `getPlanFeatureAccessor()`, but the return type of that function is very complicated and I think it might be buggy.

Currently, it looks like this:

```ts
	return [
		newsletterFeatures( flowName, plan ),
		linkInBioFeatures( flowName, plan ),
		hostingFeatures( flowName, plan ),
		signupFlowDefaultFeatures( flowName, plan, isInVerticalScrollingPlansExperiment ),
	].find( ( accessor ) => {
		return accessor instanceof Function;
	} );
```

The way it's used here the code expects the return value to be `)() => string[])|undefined`, but there's two problems with that assumption which you can see more easily if we add a return type to the function itself:

```ts
export const getPlanFeatureAccessor = ( {
	flowName = '',
	isInVerticalScrollingPlansExperiment = false,
	plan,
}: {
	flowName?: string;
	isInVerticalScrollingPlansExperiment?: boolean;
	plan: IncompleteWPcomPlan;
} ): undefined | ( () => string[] ) => {
	return [
		newsletterFeatures( flowName, plan ),
		linkInBioFeatures( flowName, plan ),
		hostingFeatures( flowName, plan ),
		signupFlowDefaultFeatures( flowName, plan, isInVerticalScrollingPlansExperiment ),
	].find( ( accessor ) => {
		return accessor instanceof Function;
	} );
};
```

1. The first is more of a TS issue. Each of the sub-functions in `getPlanFeatureAccessor` can return `false`, so in order to properly operate on the ones that don't, we need to convince TS that the falsy ones have been removed. You can do that with `isValueTruthy` like this:

```ts
	return [
		newsletterFeatures( flowName, plan ),
		linkInBioFeatures( flowName, plan ),
		hostingFeatures( flowName, plan ),
		signupFlowDefaultFeatures( flowName, plan, isInVerticalScrollingPlansExperiment ),
	]
		.filter( isValueTruthy )
		.find( ( accessor ) => {
			return accessor instanceof Function;
		} );
```

2. Unlike the others, the `hostingFeatures` sub-function does not return `() => string[]`. It returns `(term?: string) => () => string[]`. This makes `getPlanFeatureAccessor` difficult to use since it sometimes returns a function that returns strings and sometimes returns a function that returns another function. If you remove `hostingFeatures`, you'll see that the types now work correctly and you don't need `as` here:

```ts
export const getPlanFeatureAccessor = ( {
	flowName = '',
	isInVerticalScrollingPlansExperiment = false,
	plan,
}: {
	flowName?: string;
	isInVerticalScrollingPlansExperiment?: boolean;
	plan: IncompleteWPcomPlan;
} ): undefined | ( () => string[] ) => {
	return [
		newsletterFeatures( flowName, plan ),
		linkInBioFeatures( flowName, plan ),
		signupFlowDefaultFeatures( flowName, plan, isInVerticalScrollingPlansExperiment ),
	]
		.filter( isValueTruthy )
		.find( ( accessor ) => {
			return accessor instanceof Function;
		} );
};
```

You could try to guard against that return value, but I don't know what the best solution is because I don't know how `hostingFeatures` is intended to be used. 

My guess is that when `getPlanFeaturesObject` was passed a function instead of an array of strings, the fatal error `planFeaturesList.map is not a function` was thrown.

Any time you find yourself reaching for the `as` keyword to fix a TypeScript error, please consider that there might be something else going on!

pottedmeat: To build on this, the `hostingFeatures` function used to call `plan.getHostingSignupFeatures` and now it does not. That explains the transition from `() => string[]` to `(term?: string) => () => string[]`

sirbrillig: Oh, yeah, good eye! This PR changes it, but it's not totally clear why.

pottedmeat: If I remember correctly, when I reviewed it `getHostingSignupFeatures` had the `(term) => () => string` wrapper removed, so removing the function call made sense at the time. At some point, it was moved back to its previous syntax, but restoring the call with `term` wasn't.

mpkelly: Thanks for the feedback, @sirbrillig & @pottedmeat. I have tried in vain to reproduce the issue by buying monthly/annual plans, adding custom domains, buying add-ons, buying just a domain, and using the tailored flows. The checkout summary page always loads. Were any of you able to reproduce the issue?

sirbrillig: I was not, no. But we had a large number of fatal errors so we know it happens. I think we don't really need to reproduce it; this is the kind of error that TypeScript will prevent. If you can get the types to pass without using `as` or `any`, then things should be fine.

PR #75809, opened 2023-04-17
Checkout: change log severity for non-error logs

michaeldcain:
Searching Kibana for Checkout errors (`feature:calypso_client, severity:error`) shows a ton of messages for adding a product to Checkout from the URL (the primary way we add products in Checkout) and generating a thank you page URL. Neither of these things are errors and bury real errors.

This changes the log severity for non-error logstash calls to be 'info' in production. They'll remain 'debug' outside of production.

**To test:**
- a visual check should suffice. `logStashEvent`'s severity argument should default to `error` in production and be set to info for non-error logs (changing a payment method, generating a thank you url, adding a product from the url)
- you can run Calypso with `CALYPSO_ENV=production` to verify that those events don't log as errors


martian77 (approved):
👍 LGTM.

PR #75821, opened 2023-04-17
Newsletter: fix email domain on subscriber step

### LINKED ISSUE #75687: Newsletter onboarding: fix placeholder emails
Small bug: the first email here should be `sue@email.com` instead of `sue@example.com`.

Image

escapemanuele:
## What
The email domain of the placeholder example on the subscriber step should be sue@email.com instead of sue@example.com.

Image

## Testing
1. Checkout branch and `yarn start` OR visit the Live Link
2. Reach `/setup/newslettter/newsletterSetup`
3. Check the header and subheader copy

Fixes https://github.com/Automattic/wp-calypso/issues/75687

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
@@ -52,7 +52,7 @@ const Subscribers: Step = function ( { navigation } ) {
 							) }
 							showSubtitle={ true }
 							emailPlaceholders={ [
-								translate( 'sue@example.com' ),
+								translate( 'sue@email.com' ),
 								translate( 'thomaswhigginson@email.com' ),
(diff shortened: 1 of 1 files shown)


simison:
@crisbusquets @escapemanuele I like the sentiment (`example.com` is so dry) but we try not to use domains owned by others, see PCYsg-eI0-p2

Maybe `yourgroovydomain.com`? We can also register something new and fun.

escapemanuele:
Nice catch @simison!

simison (approved):
Oh and since this is just aligning example with others that are already `@email.com`, free free to deploy. We can follow up with another domain separately.

Also not sure about putting these through translate function.

escapemanuele:
Sorry @simison, merged before realizing the `translate` thing.

Opened a PR for that [here](https://github.com/Automattic/wp-calypso/pull/75830).

PR #75838, opened 2023-04-17
Help Center: Add GPT-generated response

artpi:
This adds a GPT-generated response in the Calypso help center, limited to dev, wp-calypso, and staging environments.

Screen Recording 2023-05-08 at 16 46 08

## Testing Instructions

* Navigate to Help Center
* Click to contact via email (or chat if available)
* Type your message
* Pretend to send it
* In the "last chance" screen, you will see the GPT-generated message

## Still to do 

- [ ] Proper ABtesting split so we only show this in some scenarios
- [x] Events
- [x] Proper loading indicator
- [x] Maybe feed the links quicker so that we can at least click articles
- [x] Make help articles clickable inside this window
- [x] Add required text mentioning the suggestion is AI-generated.

## Pre-merge Checklist


andrewspittle:
Thanks! A few ideas and comments.

I dropped Kelly and Pablo a note to see who within Design we can get some time from to help polish this up. I'd love their input on how we keep the interface / interaction feeling fun and whimsical. We want to avoid the customer experience of AI feeling too clinical or free of personality.

The other place we should integrate this is into the search results for queries in the "Search for help" box. And that box in particular is something we may want to make more conversational.

Agreed that if we can return documentation results faster we should. Those are also a good way to gracefully degrade in case OpenAI is down or something.

We definitely want some more text formatting in the response. Paragraph breaks, as you note, are needed.

We should think about (not necessarily for launch but perhaps as a v2) how to capture what the AI result was and pass that along to Zendesk any time a customer chooses to contact us anyway. That way we can understand a little bit about what they asked and what we already tried serving them as an answer.

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -428,6 +430,34 @@ export const HelpCenterContactForm = () => {
 		return isSubmitting ? formTitles.buttonSubmittingLabel : formTitles.buttonLabel;
 	};
 
+	// TODO: Figure out in which environments this makes sense.
+	const showGPTResponse = true;
```
yoavf: The PR mentions adding `show-results=true` to the URL - I assume you're planning to use a feature flag with `isEnabled`?

pablohoneyhoney:
So I have a few immediate comments to the above. 

- I'm gathering that we are speeding up the reactions/responses. It does need it. 

- We should probably nuance this interface to be less _send us an email_ as it sets a specific expectation. And more like: how can we help or alike. 

Screenshot 2023-04-18 at 7 17 01 PM

- Can we capture the subject from the query itself? So we reduce drastically the interface to be more GPT-esque or conversational? It also favors the loading/searching time/state to be less disruptive from the input state. 

- This header also feels off, we can simplify.

Screenshot 2023-04-18 at 7 09 12 PM

- I'd also favor this response state to have a follow up question rather than email. Which would mean an open field _or_ still email us. 

Screenshot 2023-04-18 at 7 21 23 PM

I can try to work on some quick mocks before AFK next week, or have a live session to solve them.

kriskarkoski:
> We should probably nuance this interface to be less send us an email as it sets a specific expectation. And more like: how can we help or alike.

I think it's this way because we ask them to choose a channel at the start, but in the interest of it being more conversational I think both Search and Still need Help can lead with a text box along the lines of "What are you looking to do" or "What can we help you with" and only show the support options if/when they proceed to contact us. 

> Can we capture the subject from the query itself? So we reduce drastically the interface to be more GPT-esque or conversational? It also favors the loading/searching time/state to be less disruptive from the input state.

We don't need to require a subject, I think it just helps with skimming the queue, but we could either use a default or generate it with the first X characters if we can't summarize it with GPT, but a GPT summary might be neat to experiment with.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}, [ isFetching, setLoadingState, message ] );
+
+	return (
+		<div className="help-center-sibyl-articles__container">
+			<h3 id="help-center--contextual_help" className="help-center__section-title">
+				{ __( 'Quick response', __i18n_text_domain__ ) }
+			</h3>
+			<div className="help-center-gpt-response">
+				{ data?.response && stripHTML( data.response ) }
```
yoavf: On the backend we start with Markdown, and transform to HTML just before returning the response.

The main options I see for displaying the actual formatting:
- use `dangerouslySetInnerHTML` - I think it's reasonably safe here but dangerous if the backend changes enough to introduce the risk of malicious HTML. 
- Return markdown and use something like  [react-markdown](https://github.com/remarkjs/react-markdown) to render it.

@artpi Any preference?

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -441,7 +441,7 @@ export const HelpCenterContactForm = () => {
 					<Button
 						disabled={ gptFetching ? true : isCTADisabled() }
 						onClick={ handleCTA }
-						primary
+						primary={ true }
```
yoavf: Was trying to figure out an earlier version of the following error from the unit tests:
```
packages/help-center/src/components/help-center-contact-form.tsx(444,7): error TS2769: No overload matches this call.
    Overload 1 of 2, '(props: Props, context?: any): ReactElement<any, any> | Component<Props, any, any> | null', gave the following error.
      Type '{ children: string | undefined; disabled: boolean; onClick: () => void; primary: boolean; className: string; }' is not assignable to type 'IntrinsicAttributes & Props'.
        Property 'primary' does not exist on type 'IntrinsicAttributes & Props'. Did you mean 'isPrimary'?
```
But My IDE tells me the opposite.

Screenshot 2023-04-24 at 14 30 50


Edit: needed a rebase.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -0,0 +1,32 @@
...
+export type JetpackSearchAIResult = {
+	response: string;
+	step: string;
+	urls: AIResponseURL[];
+	terms: string[];
+};
+
+export function useJetpackSearchAIQuery( siteId: number | string, query: string, stopAt: string ) {
+	return useQuery< JetpackSearchAIResult >(
+		[ query ],
```
alshakero: We should prefix this with something uniquish, it's highly unlikely, but if someone for examples searches for ["happychat-available"](https://github.com/Automattic/wp-calypso/blob/fix/proper-rq-config/packages/happychat-connection/src/use-happychat-available.ts#L59) an incorrect value will be cached and Happychat auth will break.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -0,0 +1,32 @@
...
+		async () =>
+			await wpcomRequest( {
+				path: `sites/${ siteId }/jetpack-search/ai/search`,
+				apiNamespace: 'wpcom/v2/',
+				apiVersion: '2',
+				query: `query=${ encodeURIComponent( query ) }&stop_at=${ stopAt }`,
+			} ),
```
alshakero: This won't work on Atomic sites. If we already have a Jetpack endpoint for search, we can use it like [here](https://github.com/Automattic/wp-calypso/blob/trunk/packages/help-center/src/hooks/use-help-search-query.ts#L21-L31).

packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -0,0 +1,32 @@
...
+		{
+			refetchOnWindowFocus: false,
+			keepPreviousData: false,
```
alshakero, line 41: Why not? This query is expensive, if someone searches for "A", then "B", then "A", I think they should get a cached responsive for "A".

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -428,6 +430,34 @@ export const HelpCenterContactForm = () => {
 		return isSubmitting ? formTitles.buttonSubmittingLabel : formTitles.buttonLabel;
 	};
 
+	// TODO: Figure out in which environments this makes sense.
+	const showGPTResponse = true;
+	if ( showGPTResponse && showingSibylResults ) {
+		return (
+			<div className="help-center__sibyl-articles-page">
+				<BackButton />
+				<HelpCenterGPT message={ message } setLoadingState={ setGPTFetching } />
```
alshakero: `HelpCenterGPT` can query the store directly for `message`.

```js
const message = useSelect ( select => select ( HELP_CENTER_STORE ) ).getMessage();
```

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -428,6 +430,34 @@ export const HelpCenterContactForm = () => {
...
+				<section className="contact-form-submit">
+					<Button
+						disabled={ gptFetching ? true : isCTADisabled() }
```
alshakero: Is this the same as this?
```suggestion
						disabled={ gptFetching || isCTADisabled() }
```

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -428,6 +430,34 @@ export const HelpCenterContactForm = () => {
...
+						onClick={ handleCTA }
+						primary
+						className="help-center-contact-form__site-picker-cta"
+					>
+						{ gptFetching
+							? __( 'Gathering quick response.', __i18n_text_domain__ )
+							: getCTALabel() }
```
alshakero: I think `getCTALabel` should take care of this case as well.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,86 @@
...
+type Props = {
+	message: string | undefined;
+	setLoadingState: CallableFunction;
+};
+
+const GPTResponsePlaceholder = styled( LoadingPlaceholder )< { width?: string } >`
+	:not( :last-child ) {
+		margin-block-end: 0.5rem;
+	}
+	width: ${ ( props ) => ( props.width ? props.width : '100%' ) };
```
alshakero: Nit
```suggestion
	width: ${ ( { width = '100%' } ) => width };
```

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -154,6 +155,7 @@ export const HelpCenterContactForm = () => {
 	const overflow = params.get( 'overflow' ) === 'true';
 	const navigate = useNavigate();
 	const [ hideSiteInfo, setHideSiteInfo ] = useState( false );
+	const [ gptFetching, setGPTFetching ] = useState( 'loading' );
```
alshakero: You don't need your own state, you can just duplicate the query and RQ will take care of syncing everything for you (it won't fetch twice).

`const { isFetching } = useJetpackSearchAIQuery( '9619154', message, 'response' )`

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,86 @@
...
+export function HelpCenterGPT( { message = '', setLoadingState }: Props ) {
+	const { __ } = useI18n();
+
+	const { data, isFetching } = useJetpackSearchAIQuery( '9619154', message, 'response' );
+	// Report loading state up.
+	useEffect( () => {
+		if ( setLoadingState ) {
+			setLoadingState( isFetching );
+		}
+	}, [ isFetching, setLoadingState, message ] );
```
alshakero: If you agree with the above:
```suggestion
```

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+	return (
+		<div className="help-center-sibyl-articles__container">
+			<h3 id="help-center--contextual_help" className="help-center__section-title">
+				{ __( 'Quick response', __i18n_text_domain__ ) }
+			</h3>
+			<div className="help-center-gpt-response">
+				{ ! data?.response && <Placeholders /> }
```
alshakero: I think we should handle failure. Luckily, RQ gives you `isError` boolean.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,86 @@
...
+				{ data?.response && stripHTML( data.response ) }
+				<GPTResponseDisclaimer>Generated by WordPress.com's Support AI</GPTResponseDisclaimer>
```
alshakero:
```suggestion
				<GPTResponseDisclaimer>__( 'Generated by WordPress.com's Support AI', __i18n_text_domain__ )</GPTResponseDisclaimer>
```

alshakero (changes requested):
Hey neighbor! Wow, incredible job on this one!!

My review is entirely technical, since others took care of UX. Happy to help with anything on this.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
artpi: @alshakero this endpoint is entirely unauthenticated and it's a GET request , are you sure this won't work?

packages/data-stores/src/queries/use-jetpack-search-ai.ts
artpi: The endpoint is a bit peculiar:

1. It caches by itself, quite heavily
2. Each new query is "moving the steps" a little further. In order to reduce lag, we can get early results by querying the endpoint with different ?stop_at until we have a result.

In the light of this I decided against multiple layers of caching

packages/help-center/src/components/help-center-contact-form.tsx
artpi: NICE!

packages/help-center/src/components/help-center-gpt.tsx
artpi: I was searching for a component similar to `kses` that would do `dangerouslySetInnerHTML` but only permit certain tags/attrs - it sounds like there should be something like this.
I am leaning towards HTML but I can also introduce a parameter for the endpoint to return markdown.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
alshakero: Yes sir, `wpcomRequest` actively blocks requests from Atomic sites. If this API is public, you can use `apiFetch` instead, which will work everywhere if CORS is allowed.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
alshakero: Thanks for the context, really appreciated.

packages/help-center/src/components/help-center-gpt.tsx
alshakero: You can use [`stripTags`](https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/formatting/strip-html.js#L10). It allows you to whitelist tags.

alshakero (changes requested):
Left a few more comments. If this too far out your comfort zone, we can take it from here if you like. Just let me know 👍🏼

artpi:
> If this too far out your comfort zone, we can take it from here if you like. Just let me know

Thank you for all the comments and pointers @alshakero !
I am going on sabbatical on Friday so I have a lot other work to wrap up as well :D. I'm going to try to make some progress on all of this tomorrow and we'll see how far can I get :)

packages/help-center/src/components/help-center-contact-form.tsx
yoavf: @alshakero when trying to implement this, I'm getting the following error:
```
Uncaught Error: No QueryClient set, use QueryClientProvider to set one
    useQueryClient app:///../node_modules/react-query/es/react/QueryClientProvider.js:32
    useBaseQuery app:///../node_modules/react-query/es/react/useBaseQuery.js:21
    useQuery app:///../node_modules/react-query/es/react/useQuery.js:14
    useJetpackSearchAIQuery app:///../packages/data-stores/src/queries/use-jetpack-search-ai.ts:10
    HelpCenterContactForm app:///../packages/help-center/src/components/help-center-contact-form.tsx:428
``` 

Do I need to implement a QueryClient on `useJetpackSearchAIQuery`?

**Edit**: this is odd - it's also happening to me with the current state of the PR.

alshakero: This error means there was no `QueryClientProvider` up the tree of React components. And I'm sure it's there because [this](https://github.com/Automattic/wp-calypso/blob/gpt-help-center/packages/help-center/src/components/help-center-contact-form.tsx#L166) works. 

It's also strange that it works inside `HelpCenterGPT` component but not here. 

Did you try this in the editor or in Calypso?

artpi:
It seems like @yoavf is making some very good progress here, and I will have to leave this in his and @gravityrail hands.
I am missing about three more days before going on Sabbatical, but here we are :)

I implemented a bunch of improvements in  the backend, all showcased in PeGwbA-2O-p2

**What I think should be done here:**

1. We should of course show the documents returned from the endpoint early, and break the calls into two steps ( `?stop_at=urls` and `stop_at=response`  ) how it's showcased in PeGwbA-2O-p2
2. We can collect feedback for responses via the rating endpoint - also showcased in PeGwbA-2O-p2 . Informing people about where the responses come from and giving them options to rate them will catch some grievances and frustrations in case of a really bad suggestion.
3. Since the responses are now broken into paragraphs with html, we can use `stripTags` and dangerouslySetInnerHTML to output them
4. We need to implement @alshakero  feedback
5. Apart from AI feedback, we need to also add tracks and stats to see how much % of support requests this is deflecting. This is a very important number to be able to pull up
6. I hardcoded this appearing with the `show-results=true` and hardcoding this instead of Sybil. We can do an ABtest there or just roll it out.

**From the design feedback:**

From @andrewspittle :

> The other place we should integrate this is into the search results for queries in the "Search for help" box. And that box in particular is something we may want to make more conversational.

I would do this in a follow-up PR. I tried to limit scope to a "last-ditch deflection", but we may very much change that direction.
However the **endpoint is not conversational yet.**

From @pablohoneyhoney :

> I'm gathering that we are speeding up the reactions/responses. It does need it.

I made some improvements and now we are as fast as we probably can be. We could be passing fewer documents to gpt4 which would reduce prompt size and make it faster.

> We should probably nuance this interface to be less send us an email as it sets a specific expectation. And more like: how can we help or alike.

This goes with the specific purpose of this PR: I just replaced Sybil  with GPT4 for a last-ditch deflection.

I would rephrase this in an earlier step.

> This header also feels off, we can simplify.

Yes definitely, we should

> I'd also favor this response state to have a follow up question rather than email. Which would mean an open field or still email us.

I gather this would require the whole thing to be more conversational still?

packages/help-center/src/components/help-center-contact-form.tsx
yoavf: That was happening because the import in `use-jetpack-search-ai.ts` was
`import { useQuery } from 'react-query';`
 instead of 
`import { useQuery } from '@tanstack/react-query';`

Fixed in 7d8aa74efe8148d027f4614ac5ede6402aea8c35

packages/data-stores/src/queries/use-jetpack-search-ai.ts
yoavf: Implemented, not tested yet.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -0,0 +1,45 @@
...
+	urls: AIResponseURL[];
+	terms: string[];
+};
+
+export function useJetpackSearchAIQuery( siteId: number | string, query: string, stopAt: string ) {
+	return useQuery< JetpackSearchAIResult >(
+		[ 'aiQuery', query, stopAt ],
+		async () =>
+			canAccessWpcomApis()
+				? await wpcomRequest( {
```
alshakero: This creates a promise that wraps a promise. 

```suggestion
	 () =>
			canAccessWpcomApis()
				? wpcomRequest( {
```

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -247,16 +249,35 @@ export const HelpCenterContactForm = () => {
 		Boolean( supportSite?.is_wpcom_atomic )
 	);
 
+	const enableGPTResponse = config.isEnabled( 'help/gpt-response' );
```
alshakero, line 252: This will only enable this feature in Calypso (which is probably OK), if you want it to work in the editor, you'll have to add the flag [here](https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js#L0-L1). Note that in the editor, there is no concept of "dev", I'm only sharing this info for the future when you want to deploy to prod.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,156 @@
+/* eslint-disable no-restricted-imports */
```
alshakero: I don't think you need this. All imports look unrestricted.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { HelpCenterSelect, useJetpackSearchAIQuery } from '@automattic/data-stores';
+import styled from '@emotion/styled';
+import { useSelect } from '@wordpress/data';
+import { useI18n } from '@wordpress/react-i18n';
+import { useEffect, useState } from 'react';
+import stripTags from 'striptags';
+import './help-center-article-content.scss';
+import { HELP_CENTER_STORE } from '../stores';
+
+export const SITE_STORE = 'automattic/site' as const;
```
alshakero: Important: this should be imported from the site store. We're moving away from string keys and hard-coding this would make migration harder.

[Here's](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/stores.ts#L14) the right way.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,156 @@
...
+				} );
+			}, 5000 );
+		}
+
+		return () => {
+			if ( intervalId ) {
+				clearInterval( intervalId );
+			}
+		};
+	}, [ isFetchingLinks, isFetchingResponse ] );
```
alshakero: 100% optional: but you may want to use this now-deleted [hook](https://github.com/Automattic/wp-calypso/blob/fb6e2dd2eccb8add87877fd38849132aabcf659d/client/landing/gutenboarding/hooks/use-typer.ts) from Gutenboarding. You'll have to copy-paste it though. 

Then you'd just do `const phrase = useTyper(loadingMessages)`;

It makes for a beautiful effect.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+	useEffect( () => {
+		if ( data?.response ) {
+			recordTracksEvent( 'calypso_helpcenter_show_gpt_response', {
+				location: 'help-center',
+			} );
```
alshakero, line 81: Optional: you may want to add an `onSuccess` handler in the `useJetpackSearchAIQuery` query instead. It doesn't seem to be in the scope of this component tasks.

alshakero (changes requested):
Wow, this really shaped up now! Left mostly optional comments, only the store issue is important.

packages/help-center/src/components/help-center-gpt.tsx
yoavf: Thanks! But I'm wondering, what is the purpose of the SITE_STORE export in this file? Could it be a leftover?

alshakero: This component was added in this PR, so I think the line was copied and pasted from the store. In any case, don't worry about the export, TS will complain if something needs it.

packages/help-center/src/components/help-center-gpt.tsx
yoavf: Wow, thanks so much! I added `useTyper` with a minor changes (cycling through the messages in order rather than randomly), and it works great. So much cleaner now :)

packages/help-center/src/components/help-center-gpt.tsx
yoavf: Thanks - I'll pass on this one because I'd like to keep `useJetpackSearchAIQuery` generic, and the event is help-center specific

yoavf:
@pablohoneyhoney As @artpi said, for this first iteration this isn't going to be conversational: we just load a GPT generated response in the hope to deflect some of the easier support requests.

What I'm struggling with UI / design wise is the flow after the response has been loaded:
Screenshot 2023-05-10 at 13 12 56

At this point, the user may be satisfied with the response, and close the window (do we need another pointer or button for that?). If not, then they can click the "Still email us" button to reach us. 

In the next iteration, we'd also want to let the user rate the response.

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -0,0 +1,157 @@
...
+		__( 'Processing the information found…', __i18n_text_domain__ ),
+		__( "I'm still writing, thank you for your patience!", __i18n_text_domain__ ),
+		__( 'Any minute now…', __i18n_text_domain__ ),
+		__( 'Really, any minute now…', __i18n_text_domain__ ),
+	];
+
+	const loadingMessage = useTyper( loadingMessages, true, {
+		delayBetweenCharacters: 80,
+		delayBetweenWords: 1000,
+	} );
```
alshakero, line 98: It would be nice to do `loadingMessage || '&nbsp;'` or change the hook to return nbsp instead of nothing. Currently, the UI is jumping.

https://github.com/Automattic/wp-calypso/assets/17054134/7b830dd5-6bad-4559-97a0-86025145e963

packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -0,0 +1,45 @@
...
+						apiNamespace: 'wpcom/v2/',
+						apiVersion: '2',
+						query: `query=${ encodeURIComponent( query ) }&stop_at=${ stopAt }`,
+				  } )
+				: apiFetch( {
+						global: true,
+						path: `/sites/${ siteId }/jetpack-search/ai/search?query=${ encodeURIComponent(
+							query
+						) }&stop_at=${ stopAt }`,
+				  } as APIFetchOptions ),
```
alshakero, line 38: This means we have to test in Atomic, it's a bit of a hassle, but not too bad. Do you need help testing? You'd have to download ETK from Teamcity build of this PR (called Editing Toolkit), upload it as a plugin to an Atomic site after deleting the original, then opening the Help Center.

alshakero (approved):
Beautiful! Really great work on this.

Left a tiny UX nit, other than that, I love it.

I'm almost certain it will work in Atomic, but if we ship as-is, it some code will ship untested.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
yoavf: Tried for a bit, but for some reason when testing I just get the spinner after clicking on "Still need help", but nothing happens. No errors that I can see. I'll try again tomorrow.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7980920

Thank you @artpi for including a screenshot in the description! This is really helpful for our translators.

klimeryk:
I've tested and it looks very promising! While the answer takes a while to arrive, I like that the UI keeps updating and hopefully keeps the user engaged and reassured that things are gonna load.
One nitpick for UI I have would be to switch the CTA button from `disabled` state to the `busy` one (see in action on https://wpcalypso.wordpress.com/devdocs/design/button):
Screenshot 2023-05-11 at 16 46 29

That way it's also clear something is loading and before it finishes it won't be possible to proceed. Versus the disabled status that might imply that the user needs to take action somewhere (fill something out?) to proceed.

alshakero:
Re: Busy button.

Be wary of using the Button from `automattic/components` that has this feature. It has a very generic class name that breaks random stuff. See: pdDR7T-Jm-p2.

packages/data-stores/src/queries/use-jetpack-search-ai.ts
yoavf: Confirming I tested in atomic (with the feature intentionally not enabled for now). All working as expected.

yoavf:
> Re: Busy button.
> 
> Be wary of using the Button from `automattic/components` that has this feature. It has a very generic class name that breaks random stuff. See: pdDR7T-Jm-p2.

Seems like the `wordpress/components` button also has that, so I think we're good

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7980920).

klimeryk:
> Re: Busy button.
> 
> Be wary of using the Button from `automattic/components` that has this feature. It has a very generic class name that breaks random stuff. See: pdDR7T-Jm-p2.

Appreciate the insight, @alshakero 🙇 That's the exact issues I'd expect you to catch, given the trials you went through to get Help Center working in different envs 😅 Good to know the `@wordpress/component`'s Button has this style too (and should be used in Help Center already). I linked to the Calypso's Button, but I completely forgot we have these two separate Button components 🤦

PR #75839, opened 2023-04-17
Update/allow utf 8 in domain contact form

delputnam:
## Proposed Changes

We want to allow UTF-8 characters in some contact fields.

Enabled by backend changes in D107942-code

## Testing Instructions

Make sure that it's possible to use accented characters in first name, last name, organization, city, state, address 1, address 2 fields for the domain contacts.


delputnam:
Accidentally mixed two sets of changes here.

PR #75841, opened 2023-04-17
Allow utf-8 characters in some domain contact fields

delputnam:
## Proposed Changes

We want to allow UTF-8 characters in some contact fields.

Enabled by backend changes in D107942-code

## Testing Instructions

Make sure that it's possible to use accented characters in first name, last name, organization, city, state, address 1, address 2 fields for the domain contacts.


leonardost (approved):
LGTM! Tested with D107942-code and works great, I can write my address with correct accents now 😄

PR #75845, opened 2023-04-17
Re-add Akismet overlapping products message

CodeyGuyDylan:
This PR undo's [this PR](https://github.com/Automattic/wp-calypso/pull/75184) and re-enables the overlapping akismet notice that tells the user when they have a lesser plan they might want to cancel 

## Proposed Changes

* Add the overlapping akismet notice back to the checkout page

## Testing Instructions

1. Checkout these changes
2. Sandbox `public-api.wordpress.com`
3. In your 0-sandbox.php file, add the following line
```
lang=php
define( 'USE_STORE_SANDBOX', true );
```
4. If you do not already have an Akismet product, go to http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1 and complete the checkout
5. Next, try added a better plan to the cart, in the case above, that would be http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_2
6. You should now see the notice letting the user know they have a lesser plan they might want to cancel manually
image
7. Make sure the link to the plan management page works

NOTE: This only shows up when trying to buy a *better* plan than the one you currently have. If it is the same plan or less, it does not show up because sometimes people buy 2 10K plans on purpose, or if they already have a 60K plan (business), they'll add a 10K plan because they don't want to upgrade to Enterprise. In other words, this notice is just to let users know their lower plan will not be auto-cancelled when "upgrading" 

Bonus: You can try the following combinations to ensure it is working correctly

1. If you have an Akismet Free plan and are purchasing an Akismet Plus Yearly plan, the notice should not show
2. If you have an Akismet Free plan and an Akismet Plus 10K Yearly plan, and are purchasing an Akismet Plus 20K or higher plan, the notice should show, referring to and linking to the Akismet Plus Yearly plan
3. You can tests all these plans with the "monthly" versions as well to make sure they behave normally

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
+++ b/client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
@@ -6,12 +6,15 @@ import {
 	isJetpackScanSlug,
 	getAllFeaturesForPlan,
 	planHasSuperiorFeature,
+	isAkismetProduct,
+	AKISMET_PRODUCTS_LIST,
 } from '@automattic/calypso-products';
 import { useShoppingCart } from '@automattic/shopping-cart';
 import { useEffect, useMemo } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import Notice from 'calypso/components/notice';
 import useCartKey from 'calypso/my-sites/checkout/use-cart-key';
+import { getUserPurchases } from 'calypso/state/purchases/selectors';
 import { requestRewindCapabilities } from 'calypso/state/rewind/capabilities/actions';
 import {
 	getSitePlan,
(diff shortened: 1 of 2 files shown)


client/my-sites/checkout/composite-checkout/components/prepurchase-notices/index.jsx
```diff
@@ -80,6 +87,41 @@ const PrePurchaseNotices = () => {
...
+			return null;
+		}
+
+		const highestTierOverlappedProduct = lowerTierProducts[ lowerTierProducts.length - 1 ];
+
+		// If the highest tiered product is Akismet Free, return null
+		// We don't need to display a notice about multiple paid plans in this case
+		if ( highestTierOverlappedProduct.productSlug === 'ak_free_yearly' ) {
+			return null;
+		}
```
CodeyGuyDylan, line 119: This section is the only difference from the original logic, it makes it so the notice will not display if the overlapping plan is "Free" as the messaging doesn't really make sense for a Free plan

elliottprogrammer (approved):
Tested and works great as described.  Code looks all good to me, I don't see any issues.
LGTM!! 👍

PR #75847, opened 2023-04-17
Display marketplace product type on purchases page

### LINKED ISSUE #75840: Marketplace: Display product type on purchases page
Purchases of marketplace products listed on the purchases page are not displaying the correct product type. In the case of SaaS products, the billing plan id is displayed since that value is coming in the purchase meta.

Image

JulianBustamante:
Related to #75840 

## Proposed Changes

* Returns null for marketplace_plugins and saas_plugin productTypes

## Testing Instructions

* Checkout this branch on your local environment
* Go to /plugins and purchase a Marketplace plugin
* Got to Upgrades -> Purchases
* Check the product type is not displayed for marketplace_plugins and saas_plugin

- To check with a SaaS product follow these testing instructions D105222-code

Before
Screenshot 2023-04-17 at 1 45 35 PM
After

## Pre-merge Checklist

### DIFF
--- a/client/lib/purchases/index.ts
+++ b/client/lib/purchases/index.ts
@@ -867,7 +867,7 @@ export function purchaseType( purchase: Purchase ) {
 	}
 
 	if ( purchase.productType === 'marketplace_plugin' || purchase.productType === 'saas_plugin' ) {
-		return i18n.translate( 'Marketplace Product' );
+		return null;
 	}
 
 	if ( purchase.meta ) {
--- a/client/lib/purchases/index.ts
+++ b/client/lib/purchases/index.ts
@@ -867,7 +867,7 @@ export function purchaseType( purchase: Purchase ) {
 	}
 
 	if ( purchase.productType === 'marketplace_plugin' || purchase.productType === 'saas_plugin' ) {
-		return null;
+		return i18n.translate( 'Plugin' );
 	}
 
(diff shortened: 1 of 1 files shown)


JulianBustamante:
@cpapazoglou Can we agree on the product type that should be displayed? I set "Marketplace Product" to create this PR. Also, we might want to avoid displaying the product type below the product name for marketplace products.

cpapazoglou:
> @cpapazoglou Can we agree on the product type that should be displayed? I set "Marketplace Product" to create this PR. Also, we might want to avoid displaying the product type below the product name for marketplace products.

Seems we already go with just the Product Name for plugins. I would keep it the same for consistency.

CleanShot 2023-04-19 at 15 21 28@2x

client/lib/purchases/index.ts
```diff
@@ -866,6 +866,10 @@ export function purchaseType( purchase: Purchase ) {
 		} );
 	}
 
+	if ( purchase.productType === 'marketplace_plugin' || purchase.productType === 'saas_plugin' ) {
+		return null;
```
cpapazoglou:
```suggestion
		return i18n.translate( 'Plugin' );
```

cpapazoglou (approved):
Eventually, I think that returning `Plugin` is a better idea and matches the condition.

PR #75874, opened 2023-04-18
VideoPress onboarding: poll video status instead of relying on playback event

thedebian:
## Proposed Changes

This PR polls the player on VideoPress onboarding intro instead of relying on status change, in case the player is already playing before we attach the iframe API to it.
It also adds a slight opacity transition to make the still image to video transition a bit smoother.

## Testing Instructions

* Apply PR
* Go to /setup/videopress
* ✅ The video should play at some point
* ✅ The transition between the still frame and the video should be less choppy

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/intro/styles.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/intro/styles.scss
@@ -127,7 +127,7 @@
 		}
 
 		.intro__video.loading-frame {
-			transition: opacity 0.3s ease-in-out;
+			transition: opacity 0.31s ease-in-out;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/intro/styles.scss
```diff
@@ -125,6 +125,10 @@
 			margin-top: 32px;
 			font-size: 1.25rem;
 		}
+
+		.intro__video.loading-frame {
+			transition: opacity 0.3s ease-in-out;
```
roundhill, line 130: Nice!

roundhill (approved):
Tested well for me!

PR #75888, opened 2023-04-18
Removed Lodash from sort-imports codemod

yashitamittal11:
## Description

This change is for the sort-imports.js codemods script( that adds import comment blocks and sorts them as necessary). I have modified the code to remove the Lodash functions from this script.

## Proposed Changes

Replaced the following Lodash functions with the native JS code

1. _.sortBy
2. _.isEmpty

## Testing Instructions

- Open the file
- Click on Run and Debug
Ensure the values returned by the new functionality are the same as those returned with the Lodash functions.

## Checklist

Verify all tests still pass.

### DIFF
--- a/packages/calypso-codemods/transforms/sort-imports.js
+++ b/packages/calypso-codemods/transforms/sort-imports.js
@@ -97,7 +97,9 @@ const addNewlineBeforeDocBlock = ( str ) => str.replace( /(import.*\n)(\/\*\*)/,
  * @returns {Array} the sorted set of import nodes
  */
 const sortImports = ( importNodes ) =>
-	importNodes.sort( ( nodeA, nodeB ) => nodeA.source.value.localeCompare( nodeB.source.value ) );
+	[ ...importNodes ].sort( ( nodeA, nodeB ) =>
+		nodeA.source.value.localeCompare( nodeB.source.value )
+	);
 
 module.exports = function ( file, api ) {
 	const j = api.jscodeshift;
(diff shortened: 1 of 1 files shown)


packages/calypso-codemods/transforms/sort-imports.js
```diff
@@ -112,7 +112,7 @@ module.exports = function ( file, api ) {
 		externalDependenciesSet.has( importNode.source.value.split( '/' )[ 0 ] );
 
 	// if there are no deps at all, then return early.
-	if ( _.isEmpty( declarations.nodes() ) ) {
+	if ( ! declarations.nodes() || declarations.nodes().length === 0 ) {
```
tyxla: Nit: Should we assign `declarations.nodes()` to a constant so it's invoked just once?

packages/calypso-codemods/transforms/sort-imports.js
```diff
@@ -97,7 +96,8 @@ const addNewlineBeforeDocBlock = ( str ) => str.replace( /(import.*\n)(\/\*\*)/,
  * @param {Array} importNodes the import nodes to sort
  * @returns {Array} the sorted set of import nodes
  */
-const sortImports = ( importNodes ) => _.sortBy( importNodes, ( node ) => node.source.value );
+const sortImports = ( importNodes ) =>
+	importNodes.sort( ( nodeA, nodeB ) => nodeA.source.value.localeCompare( nodeB.source.value ) );
```
tyxla: `_.sortBy()` won't mutate the input array, while this one will (since `Array.prototype.sort()` sorts the input array in-place, and returning a reference to the array is a bonus). Should we create a copy of the array before sorting it?

yashitamittal11: Ohh. I missed it. I have created a copy of the array now.

packages/calypso-codemods/transforms/sort-imports.js
yashitamittal11: Yes. done

tyxla (approved):
LGTM, thank you! 🚀

yashitamittal11:
@tyxla Thanks for the review!

PR #75898, opened 2023-04-18
Virtual themes: Track recipe slug

### LINKED ISSUE #75490: Virtual themes: Track pattern slug
We currently track the pattern ID of all events involving a virtual theme in the `virtual_theme_pattern` property, but's far from being human-readable.

Similary to how we change the tracked `slug` prop for non-virtual themes to `<THEME_SLUG>-<STYLE_VARIATION_SLUG>` when a style variation is selected, it'd be great if the tracked `slug` prop for virtual themes changes to `<THEME_SLUG>-<PATTERN_SLUG>`.

Original request: p1681136519074649/1681135345.566989-slack-C04DZ8M0GHW

mmtr:
Fixes https://github.com/Automattic/wp-calypso/issues/75490
Fixes https://github.com/Automattic/dotcom-forge/issues/2099
Requires D108161-code

## Proposed changes

Uses the new `recipe.slug` property added in D108161-code as the tracked `slug` of the `calypso_signup_design_preview_select` and `calypso_signup_select_design` events, so we can identify with a human-readable string what pattern has been used to build the virtual theme.

## Testing instructions

- Open the Network tab of the browser devtools.
- Click on a virtual theme to preview it.
- Make sure that the tracked `calypso_signup_design_preview_select` event uses the `recipe.slug` value (`<THEME_SLUG>-<PATTERN_SLUG>`) in the `slug` property.
- Proceed with the design selection.
- Make sure that the tracked `calypso_signup_select_design` event uses the `recipe.slug` value (`<THEME_SLUG>-<PATTERN_SLUG>`) in the `slug` property.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/analytics/record-design.ts
+++ b/client/landing/stepper/declarative-flow/internals/analytics/record-design.ts
@@ -92,14 +92,8 @@ export function getDesignEventProps( {
 }
 
 export function getVirtualDesignProps( design: Design ) {
-	/**
-	 * If the design is virtual, and it has a recipe with pattern_ids,
-	 * then we assume that it's a pattern based virtual theme.
-	 */
-	const virtual_theme_pattern = design.is_virtual ? design.recipe?.pattern_ids?.[ 0 ] : null;
-
 	return {
 		is_virtual: design.is_virtual,
-		virtual_theme_pattern,
+		slug: design.is_virtual ? design.recipe?.slug : design.slug,
 	};
 }
--- a/packages/design-picker/src/types.ts
+++ b/packages/design-picker/src/types.ts
@@ -62,6 +62,7 @@ export interface DesignRecipe {
 	footer_pattern_ids?: number[] | string[];
 	color_variation_title?: string;
 	font_variation_title?: string;
+	slug?: string;
 }
 
 export interface SoftwareSet {


client/landing/stepper/declarative-flow/internals/analytics/record-design.ts
```diff
@@ -92,14 +92,8 @@ export function getDesignEventProps( {
...
-	/**
-	 * If the design is virtual, and it has a recipe with pattern_ids,
-	 * then we assume that it's a pattern based virtual theme.
-	 */
-	const virtual_theme_pattern = design.is_virtual ? design.recipe?.pattern_ids?.[ 0 ] : null;
-
 	return {
 		is_virtual: design.is_virtual,
-		virtual_theme_pattern,
+		slug: design.is_virtual ? design.recipe?.slug : design.slug,
```
BogdanUngureanu, line 97: Non-blocking: imo, we shouldn't have the slug used for both patterns and themes because it represents two different types of information. It would make it harder to determine which theme is actually used, especially since we'll support multiple themes for virtual themes - we'll change the existing data format. 

I think we should use a new property, like `virtual_theme_pattern`.

mmtr: I wondered about that too, but ultimately decided to stay consistent with how the `slug` property is currently tracked when a style variation or a vertical-generated design is selected.

@ianstewart also suggested if we could have something similar for virtual themes and was open to keep using the same `slug` prop: p1681136519074649/1681135345.566989-slack-C04DZ8M0GHW.

BogdanUngureanu: Oh, I see. Then if that was the request, then it's fine with me :) I thought that prop should always contain a valid theme and might confuse folks that monitor that event. :)

BogdanUngureanu (approved):
LGTM and works as expected! 🎉

PR #75904, opened 2023-04-18
Stop tracking virtual themes properties when switching style variations

mmtr:
## Proposed Changes

Virtual themes no longer have style variations, so we don't need to add virtual themes properties to the `calypso_signup_design_picker_style_variation_button_click` event tracked when changing the previewed style variation.

## Testing Instructions

- Open the Network tab of the browser devtools.
- Use the discs to change the previewed style variation of a design with style variations.
- Make sure the tracked `calypso_signup_design_picker_style_variation_button_click` doesn't have the `is_virtual`/`virtual_theme_pattern` properties.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
@@ -40,7 +40,6 @@ import {
 	getDesignTypeProps,
 	recordPreviewedDesign,
 	recordSelectedDesign,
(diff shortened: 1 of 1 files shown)


BogdanUngureanu:
Hmm... Right now vThemes use BC3; however, we do plan to use multiple themes in the near future. Maybe those themes would support style variations. 

Are there any undesired side-effects with these props we have right now? :)

mmtr:
Happy to keep the code around! I just don't like having useless code if it's not strictly needed, but this doesn't cause any undesired side-effect as you pointed out and might be useful in the future.

PR #75913, opened 2023-04-18
Sidebar Heading: fix React error

monsieur-z:
## Proposed Changes

This PR fixes the following error triggered by the `SidebarHeading` component. Invalid HTML attributes are passed to a `a` tag.

Screenshot 2023-04-18 at 11 39 15 AM

## Implementation Notes

A proper solution would be not to pass `...props`, but only the desired attributes. Since I'm not aware of all the use cases for this component and what `props` might hold, I'm being conservative with this fix.

## Testing Instructions

- Spin up Calypso, using a live link below, or by running the branch locally
- Open the developer console
- Visit `/posts`, for instance
- Check there's no error in the console

## Pre-merge Checklist

### DIFF
--- a/client/layout/sidebar/heading.jsx
+++ b/client/layout/sidebar/heading.jsx
@@ -13,6 +13,12 @@ const SidebarHeading = ( { children, onClick, ...props } ) => {
 		};
 	}
 
+	const linkAttrs = { ...props };
+
+	// These are not valid HTML attribute for a <a> tag
+	delete linkAttrs.navigationLabel;
+	delete linkAttrs.url;
+
 	/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 	return (
 		<li>
@@ -21,7 +27,7 @@ const SidebarHeading = ( { children, onClick, ...props } ) => {
 				className="sidebar__heading"
 				onKeyDown={ onKeyDown }
 				onClick={ onClick }
-				{ ...props }
+				{ ...linkAttrs }
 			>
 				{ children }
 			</a>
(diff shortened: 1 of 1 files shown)


client/layout/sidebar/heading.jsx
```diff
@@ -13,6 +13,12 @@ const SidebarHeading = ( { children, onClick, ...props } ) => {
 		};
 	}
 
+	const linkAttrs = { ...props };
+
+	// These are not valid HTML attribute for a <a> tag
+	delete linkAttrs.navigationLabel;
+	delete linkAttrs.url;
```
manzoorwanijk, line 20: It's nice that you cloned the object before deleting to avoid mutating the props. 👍

manzoorwanijk (approved):
Great

client/layout/sidebar/heading.jsx
jsnajdr: There's another elegant way to do exactly the same thing, with destructuring:
```js
const { navigationLabel, url, ...linkAttrs } = props;
```
This creates `linkAttrs` from `props`, omitting the `navigationLabel` and `url` fields. The `no-unused-vars` ESLint rule doesn't mind that the destructured variables are not used, because it recognizes the pattern, and supports it through the `ignoreRestSiblings` option, which Calypso has configured to `true`.

tyxla: I'd second Jarda's suggestion, since `delete` is effectively a mutation, and mutations may sometimes be inadvertently carried over as we refactor code. By creating `linkAttrs` object without those props by design, we keep code immutable.

Issue #75927, opened 2023-04-18
No option to change payment method on an Akismet subscription


DavidRothstein:
### Quick summary

When you buy an Akismet subscription (now managed on WordPress.com) you can't change the payment method attached to the subscription.

Normally the link would be somewhere in here, at the bottom of the purchase management page:

![manage-payment-methods](https://user-images.githubusercontent.com/235183/232913868-b4f72b3f-5f31-4aa1-a1e3-4ed845d5f38e.png)

We wonder if this has something to do with siteless checkout.

### Steps to reproduce

1. Start at https://akismet.com/
2. Buy a subscription
3. Find it at https://wordpress.com/me/purchases and click to manage it

### What you expected to happen

There should be a link to change the payment method

### What actually happened

There is no link to change the payment method

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

DavidRothstein:
https://github.com/Automattic/wp-calypso/pull/75929 might fix this, once it's ready.

DavidRothstein:
This is fixed by the above issue.

PR #75929, opened 2023-04-18
Remove site selection for payment method management in me/purchases

jboland88:
Related to #75927

## Proposed Changes

* Adds a way to add or change payment methods for a siteless purchase (including Akismet)

## Testing Instructions

* Purchase a product from Akismet.com
* Navigate to /me/purchases and find your purchase in the list - click on it to go to the purchase management screen
* From the purchase management screen, you should now have the button/ action to change your payment method

Screenshot 2023-04-18 at 6 02 10 PM

* Test Jetpack and WordPress.com purchases as well from /me/purchases to make sure managing payment methods works as expected
* Test that payment management works as expected when accessed from the /my-sites/ routes (managing a purchase with a specific site selected in the site selector)

## Pre-merge Checklist

### DIFF
--- a/client/me/purchases/manage-purchase/change-payment-method/index.tsx
+++ b/client/me/purchases/manage-purchase/change-payment-method/index.tsx
@@ -24,7 +24,6 @@ import {
 	hasLoadedUserPurchasesFromServer,
 } from 'calypso/state/purchases/selectors';
 import { isRequestingSites } from 'calypso/state/sites/selectors';
-import { getSelectedSite } from 'calypso/state/ui/selectors';
 import PaymentMethodSelector from '../payment-method-selector';
 import getPaymentMethodIdFromPayment from '../payment-method-selector/get-payment-method-id-from-payment';
 import useCreateAssignablePaymentMethods from './use-create-assignable-payment-methods';
@@ -46,14 +45,13 @@ function ChangePaymentMethod( {
 	const hasLoadedUserPurchases = useSelector( hasLoadedUserPurchasesFromServer );
 	const purchase = useSelector( ( state ) => getByPurchaseId( state, purchaseId ) );
 	const payment = useSelector( ( state ) => getByPurchaseId( state, purchaseId )?.payment );
-	const selectedSite = useSelector( getSelectedSite );
 	const { isLoading: isLoadingStoredCards } = useStoredPaymentMethods( { type: 'card' } );
 
 	const { isStripeLoading } = useStripe();
 
 	const isDataLoading =
 		! hasLoadedSites || ! hasLoadedUserPurchases || isLoadingStoredCards || isStripeLoading;
-	const isDataValid = purchase && selectedSite;
+	const isDataValid = purchase;
 
 	useEffect( () => {
 		if ( ! isDataLoading && ! isDataValid ) {
--- a/client/me/purchases/manage-purchase/change-payment-method/index.tsx
+++ b/client/me/purchases/manage-purchase/change-payment-method/index.tsx
@@ -51,21 +51,20 @@ function ChangePaymentMethod( {
 
 	const isDataLoading =
 		! hasLoadedSites || ! hasLoadedUserPurchases || isLoadingStoredCards || isStripeLoading;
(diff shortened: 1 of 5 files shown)


JessBoctor (commented):
It looks good to me. I was able to edit the payment method with this PR applied: 

GH75929-AkismetPaymentMethods-Screenshot 2023-04-18 at 3 30 15 PM

I also checked for Jetpack and WordPress.com purchase management and they worked as I expected.

The only question I had about the code was if removing `siteSelected` would be a problem for renewals, or which payments are listed under each payment method. Since renewal information is stored in the subscription tables and there is no screen to see which subscriptions are listed with the payment method, I think this is all good. 

Thanks for the quick response to this issue!

DavidRothstein:
Thanks for working on this!

It worked fine for me when switching to an **existing** payment method, but when trying to switch to a new credit card on that screen I just got "User or Token does not have access to specified site" which comes from the `upgrades/[subscription-id]/update-credit-card` endpoint.  Can anyone else reproduce that (using cards from https://stripe.com/docs/testing in store sandbox)?

As far as I can see, that error only happens if the user does not have access to the site on which the subscription exists.  But why would the user not have access to their own Akismet Siteless Checkout holding site?

IanRamosC:
> Can anyone else reproduce that (using cards from https://stripe.com/docs/ in store sandbox)?

It worked fine for me both when changing to an existing payment method and changing to a new one - it showed the success message "Your payment method has been set."

I also tried cancelling all my active Akismet subscriptions, so the next purchase would trigger the creation of a new holding site, and it also worked for the new purchase (with the new holding site).

@DavidRothstein is there anything else that could be triggering that message to you?

IanRamosC (approved):
LGTM! Let's see what's causing what was mentioned in the other comments and merge it if it's ready.

jboland88:
@DavidRothstein @IanRamosC - I was able to reproduce the issue mentioned [here](https://github.com/Automattic/wp-calypso/pull/75929#issuecomment-1514096784) where adding a new card was declined. I could reproduce this with a test user.

I have a fix for that in D108321-code which will need to get deployed ahead of this PR. Could you retest with this diff applied?

client/me/purchases/manage-purchase/change-payment-method/index.tsx
```diff
@@ -46,14 +45,13 @@ function ChangePaymentMethod( {
...
 	const payment = useSelector( ( state ) => getByPurchaseId( state, purchaseId )?.payment );
-	const selectedSite = useSelector( getSelectedSite );
 	const { isLoading: isLoadingStoredCards } = useStoredPaymentMethods( { type: 'card' } );
 
 	const { isStripeLoading } = useStripe();
 
 	const isDataLoading =
 		! hasLoadedSites || ! hasLoadedUserPurchases || isLoadingStoredCards || isStripeLoading;
-	const isDataValid = purchase && selectedSite;
+	const isDataValid = purchase;
```
CodeyGuyDylan: This variable seems kinda irrelevant now. You could just use `! purchase` below and it should do the same thing, and likewise anywhere else this variable is used 😄

CodeyGuyDylan (approved):
Changes look good! Made one note about an unneeded variable but the test plan worked well as well as changing the payment card with a non-a11n user account

DavidRothstein:
>I have a fix for that in D108321-code which will need to get deployed ahead of this PR. Could you retest with this diff applied?

I tested and confirmed it seems to be working well with that -- thanks!

client/me/purchases/manage-purchase/change-payment-method/index.tsx
jboland88: Good point - had to fix another bit of logic, so updated this as well.

elliottprogrammer (approved):
Tested this with D108321-code applied. Tested changing & adding payment method for Akismet purchase and works great, as described. Tested changing payment method for Jetpack and WordPress.com purchases too. Can successfully turn auto-renew ON and OFF.
Everything works well for me. Code looks good, I don't see any issues. Unit tests passing too.
LGTM! 👍

PR #75945, opened 2023-04-19
Improve visual style on /tags page

roo2:
Addresses some feedback on the /tags page from @ollierozdarz
 
> * I think give the contents a max-width, same as in the design (1140px). On a big monitor it's far too stretched out
> * The idea was to make the A-Z list position: sticky once it hits the top of the viewport. Would be cool to see that working.
> * We should give the tags themselves a hover state (especially all of those in the A-Z). A simple border-bottom: 1px solid grey100; (or similar)
> * The large Letter at the start of each tag section should be font-weight: medium;
> * Content marketing gave us a new title for the page: All the Tags and subtitle For every one of your interests, there's a  tag on [WordPress.com](http://wordpress.com/).

#### Testing instructions

The only interesting change I see is making the sticky panel test for whether the header is fixed position or not. I manually tested that change on the /tags page by setting the header back to `position:fixed`. In code I found references to stickypanel on the themes and media library components, but I couldn't see it in use on wordpress.com

### DIFF
--- a/client/components/sticky-panel/index.jsx
+++ b/client/components/sticky-panel/index.jsx
@@ -21,9 +21,12 @@ const commonDefaultProps = {
 };
 
 export function calculateOffset() {
-	// Offset to account for Masterbar
 	const headerEl = document.getElementById( 'header' );
-	return headerEl ? headerEl.getBoundingClientRect().height : 0;
+	// Offset to account for Masterbar if it is fixed position
+	if ( headerEl && getComputedStyle( headerEl ).position === 'fixed' ) {
+		return headerEl.getBoundingClientRect().height;
+	}
+	return 0;
 }
 
 export function getBlockStyle( state ) {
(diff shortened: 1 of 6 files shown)


davemart-in (approved):
LGTM!

roo2:
Also including a few more fixes 

> 1. Let's kill the padding on the main 1140px container - it's making it super condensed.
> 2. We should include the header headline and subheading in its own 1140px container to be consistent with the other marketing pages.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7945200

Hi @roo2, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `All the Tags`
- `For every one of your interests, there's a tag on WordPress.com.`

Thank you in advance!

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7945200).

Issue #75855, opened 2023-04-18
Stats: Subscribers page - add Subscribers highlights


grzegorz-cp:
Replace the placeholder `Annual highlights` (aka YYYY in review) from the `Subscribers` page with the new highlights section.

# Designs
* Widescreen
![Image](https://user-images.githubusercontent.com/112354940/232655784-1b26009f-b6ca-45c9-b259-c73a092fc4e8.png)

* Mobile
![Image](https://user-images.githubusercontent.com/112354940/232655834-161427c7-f747-4289-905c-53eb03eaf382.png)

# Development
* Verify if all data points are already available - e.g. free and paid email subscribers
* Create new components to display data and place it in Storybook with the shared component (use Storybook with test data set)
* Add the new component to the Subscribers page
* Verify with the designers if the tooltip content is correct

Before starting the task, consider a further breakdown of the task and which step could result in a partial pull request.

jsnmoon:
I believe this is complete!

PR #75956, opened 2023-04-19
Stats: Add new SubscribersHighlightSection component to Subscribers page

a8ck3n:
Related to #75855.

## Proposed Changes

Replaces the `AnnualHighlightsSection` component (from the Traffic page) with a new `SubscribersHighlightSection` component specific to this page.

- Mostly lines up with the designs in 75855 except for the number formatting. Currently using the same formatting rules as the Traffic page for consistency.
- Not wired to any real data yet. Just using hard-coded data per the initial designs.

Will address the following (as needed) in follow-up PRs:

1. Tooltip content for "i" icon.
2. Do we want tooltips on the individual values? Currently enabled.
3. Do we want icons for each card?

## Testing Instructions

* Open the Live branch.
* Visit Jetpack → Stats.
* Append `?flags=stats/subscribers-section` to the URL and load.
* Visit the new Subscribers page.
* Confirm the new "All time stats" section is visible and behaves as expected.

SCR-20230419-qvga

## Pre-merge Checklist


client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+function SubscriberHighlightsHeader() {
+	const [ isTooltipVisible, setTooltipVisible ] = useState( false );
+	const infoReferenceElement = useRef( null );
+	const localizedTitle = translate( 'All time stats', {
+		comment: 'Heading for Subscribers page highlights section',
+	} );
+	const showNotices = true;
+	const tooltipContent = <p>Tooltip content goes here.</p>;
+	return (
+		<h1 className="highlight-cards-heading">
```
jsnmoon: Could we add a subscriber-section-specific class name here?

client/my-sites/stats/stats-subscribers-page/index.jsx
```diff
@@ -56,8 +56,7 @@ const StatsSubscribersPage = ( { period } ) => {
 					align="left"
 				/>
 				<StatsNavigation selectedItem="subscribers" siteId={ siteId } slug={ siteSlug } />
-				{ /* TODO: replace annual highlight */ }
-				<AnnualHighlightsSection siteId={ siteId } />
+				<SubscriberHighlights />
```
jsnmoon: We'll probably still want to pass `siteId` into the new highlight section 🙂

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
+import { HighlightCard, Popover } from '@automattic/components';
```
jsnmoon: This file name has a typo; it should probably be `subscriber-highlights.jsx`.

jsnmoon: We could also try making this a TypeScript file if you'd be interested in the challenge 🙂

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+import { Icon, info } from '@wordpress/icons';
+import { translate } from 'i18n-calypso';
+import { useRef, useState } from 'react';
+import '../wordads/highlights-section.scss';
```
jsnmoon: Could we use a `subscriber-section.scss` instead of using styling from `wordads`? This import creates an unnecessary dependency chain between subscriber and wordads sections.

jsnmoon (approved):
Looks good; let's ship and iterate!

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
grzegorz-cp: Let's also make it plural because we always name the elements on this page as `subscribers`.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
grzegorz-cp: Great catch. If we need to share the same styles, we can create a mixin and import it the same way we did for all of the grids in `grid-layout` or `modernized-mixins`. Mixins can also accept parameters and are very flexible.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+					showValueTooltip
+				/>
+			) ) }
+		</div>
+	);
+}
+
+export default function SubscriberHighlights() {
+	return (
+		<div className="highlight-cards wordads has-odyssey-stats-bg-color">
```
grzegorz-cp: Similar to what Jason mentioned for the heading, this element could use a class specific to subscribers highlights because it's the parent.

It looks like `wordads has-odyssey-stats-bg-color`are redundant.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+					>
+						<div className="highlight-card-tooltip-content">{ tooltipContent }</div>
+					</Popover>
+				</>
+			) }
+		</h1>
+	);
+}
+
+function SubscriberHighlightsListing() {
```
grzegorz-cp: All of the functions here are separate components. Let's use arrow functions while defining them.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+	return highlights;
+}
+
+function SubscriberHighlightsHeader() {
+	const [ isTooltipVisible, setTooltipVisible ] = useState( false );
+	const infoReferenceElement = useRef( null );
+	const localizedTitle = translate( 'All time stats', {
+		comment: 'Heading for Subscribers page highlights section',
+	} );
+	const showNotices = true;
```
grzegorz-cp: `showNotices` should either be a property, or the tooltip should always be visible. Since the designs contain a tooltip, let's always show it and remove the variable.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
...
+	);
+}
+
+function SubscriberHighlightsListing() {
+	const highlights = getSubscriberHighlights();
+
+	return (
+		<div className="highlight-cards-list">
+			{ highlights.map( ( highlight ) => (
+				<HighlightCard
```
grzegorz-cp: It would be great if we extended the card to avoid showing an icon if that property is not passed to the component.

grzegorz-cp (approved):
Looks good. I left a few ideas for future changes.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
```diff
@@ -0,0 +1,97 @@
+import { HighlightCard, Popover } from '@automattic/components';
+import { Icon, info } from '@wordpress/icons';
+import { translate } from 'i18n-calypso';
+import { useRef, useState } from 'react';
+import '../wordads/highlights-section.scss';
+
+function getSubscriberHighlights() {
+	const highlights = [
```
grzegorz-cp: 7-day highlight from the Traffic page and Annual highlight from the Insights page are in Storybook. It's also a great place to use with mock data for testing. In the next PRs let's move the component there using the same pattern from Traffic and Insights pages.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: I actually thought about naming the component `SubscribersHighlights` but felt it read awkwardly. I think you're right though, as it makes for a more consistent naming pattern within the page. I'll address the typo as well.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: I can definitely do this. Is there a technical reason for this suggestion or is this just a preference?

In my reading, I'm not finding anything specific to React that would argue for this as a best practice. My understanding of the potentially relevant issues here are as follows:

1. Arrow functions automatically bind `this` to the calling scope. My understanding is this is only relevant in Class-based components.
2. Arrow functions are allocated at run-time each time they are called. I'm assuming this means they are copied to the stack and addressed by their memory address. This means they could actually introduce performance concerns if called repeatedly in a tight loop. Probably not relevant to our usage.
3. Related to the above, arrow functions are not "comparison safe". So passing an arrow function as a prop would always result in a new render even though the callback itself has not changed.

I'm guessing our build tools might also impact this if they are rewriting our arrow functions to some baseline version for compatibility.

Anyway, mostly just asking to make sure I'm clear on this.

cc: @grzegorz-cp

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: Oh my! I meant to remove this. I was playing with refactoring the heading bits and left this in. I'll clean this up for now though.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: Jason is currently doing some refactoring in the various Highlight Cards so I'll circle back to this one.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: Sounds good! I was looking at this again today to make sure I have the right level of separation. I think it makes sense to have all data configuration in the Stats code and only visual components in the Storybook to keep things from being coupled too tightly. I'll review the highlights we currently have in Storybook and address that in a follow-up PR.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: Will do. I was looking at the various section headers across the Traffic, Insights, and Ads pages and they all use this same CSS class except for the chart header and the "Posting activity" header. I think it makes sense to review this and pull it into a mix-in or standalone "section header" component.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: Definitely, I'll give this a go!

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: I've copied the necessary styles into a new CSS file for now. I want to try a small refactoring next which should eliminate the need for a mixin.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
a8ck3n: The `wordads` is necessary there to get the info icon vertically aligned and properly colored. I've updated it to `subscribers-page` now that the CSS lives in a new file.

As for `has-odyssey-stats-bg-color`, I believe it's only defined when running in Odyssey stats. It's still defined in the code and present on the other highlight sections on the Traffic and Insights page. I'll take a closer look at this and open a follow-up PR if needed though.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
grzegorz-cp: In this circumstance, it was just a suggestion and preference.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
jsnmoon: Note that `.subscribers-page` may lead to styling leaks like `.promo-card` ([issue](https://github.com/Automattic/wp-calypso/issues/76032)). I recommend we use something more specific to this component, like `.stats-subscriber-highlights`.

client/my-sites/stats/stats-subscribers-page/subsriber-highlights.jsx
kangzj: I normally would use normal functions for components. I'd like to know more about this too 😄 

But @a8ck3n I have some questions regarding the 'potentially relevant issues' regarding arrow functions,

> Arrow functions automatically bind this to the calling scope. My understanding is this is only relevant in Class-based components.

I'm not too sure about this statement. As I see it, the best thing about arrow functions is that they do NOT automatically bind to the calling scope.

> Arrow functions are allocated at run-time each time they are called.

Can I ask for reference regarding the statement? From my understanding, arrow functions are exactly the same as functions, they don't get allocated any way at run-time, which is only related to where you define it.

For example, if you define the function outside components, where they could just stay what they are and not re-allocated at all on re-rendering.

So I think it's not related to arrow functions themselves, but how you use them. For normal functions this could happen too.



> Related to the above, arrow functions are not "comparison safe".

The same thing applies here too.

```
// These are good - they don't get 're-allocated' on every rendering, and they are "comparison safe".
const f1 = function(){};
const a1 = ()=> {};

function TheComponent() {
   // these get 're-allocated' on every rendering, so they are not "comparison safe".
   const f2 = function(){};
   const a2 = ()=> {};

   f1(); f2(); a1(); a2();

   return <div>Blaaaablaaa..</div>;
}
```

That said tho. I'm in line with Kevin. I don't see normal functions or arrow ones have advantages over the other. Let me know if I missed stuff.

PR #61849, opened 2022-03-11
HackWeek: Show a banner when the site has 0 content

AllTerrainDeveloper:
#### Changes proposed in this Pull Request

As part of the hackweek we want to add a banner in Stats page when the user has not created any content yet. This code depends on backend changes added by the patch D77480-code

#### Testing instructions

Use calypso live branch or checkout this branch (add/banner-in-stats-when-site-has-0-content)
~~if D77480-code is deployed, or sandbox your public-api before perform this tests.~~

Test Events:
Ensure the following events are being sent to Tracks (An easy way to test this is to have [Tracks Vigilante](https://github.com/Automattic/tracks-chrome-extension) monitoring the two next tests)

1. "calypso_stats_no_content_banner_click"
2. "calypso_stats_no_content_banner_dismiss"
3. "calypso_stats_no_content_banner_view"

Test 1:

1. Have a new site where you have not created any content (page or post)
2. Go to Stats page
3. Check that you see the banner
4. Edit (or create if the blog does not have) a page and repeat the test
5. Check that you don't see the banner
6. Delete the page
7. Repeat request with include_pages=true EG: /sites/203903727/site-has-never-published-post?include_pages=true
8. Check that you see the banner
9. Create a blog post
10. Check that you don't see the banner

Test 2:

1. Have a new site where you have not created any content (page or post)
2. Click on the CTA shown by the banner (Write a post)
3. Check that it takes you to write a post (don't create anything)
4. Go to Stats again
5. Click on the dismissal icon
6. Check reload the page and check that you are no longer presented with the banner

*Banner:
image

image

PR #75963, opened 2023-04-19
wpcom-proxy-request: don't default a falsy response

jsnajdr:
Fixes a little bug in `wpcom-proxy-request` that I discovered when testing the React Query upgrade in #74381, namely the `useHasNeverPublishedPost` query (introduced in #61849). The endpoint the query uses returns a body that is a plain boolean, not wrapped in an object or anything:

Screenshot 2023-04-18 at 15 44 20

Then there is a `e.response || xhr.response` that incorrrectly defaults a `false` response to `xhr.response`, which is an empty string -- because the `xhr` object was never used for sending a request.

Then the hook returns `''` instead of `false`. That's still a falsy value, so it's not that bad, but it's still an incorrect value that also doesn't satisfy the declared type.

Fixing this by using the `??` operator instead. That should be fine until we start receiving `body: null` responses 🙂 Hopefully that never happens.

### DIFF
--- a/packages/wpcom-proxy-request/History.md
+++ b/packages/wpcom-proxy-request/History.md
@@ -2,6 +2,8 @@
 
 ## Next / TBD
 
+- Don't modify a falsy boolean JSON response body by defaulting it.
+
 ## 6.0.0 / 2021-03-19
 
 - Add "support" for streamed responses and `onStreamRecord`. Doesn't implement, just tolerates the callback.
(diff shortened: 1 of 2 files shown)


noahtallen (approved):
Nice, this change makes sense to me 👍 

Potentially a good scenario for a unit test?

tyxla (approved):
Makes sense, thanks for the fix 🚀

jsnajdr:
> Potentially a good scenario for a unit test?

The `wpcom-proxy-request` package is unfortunately not very amenable to unit testing. Internally it creates the proxy iframe and loads the proxy script into it, relying on a real server that exists.

PR #75982, opened 2023-04-19
Billing History: Load past billing history separately from upcoming charges

pottedmeat:
Allows billing transactions to be limited to just past or upcoming transactions.

Related to Automattic/payments-shilling#1562

Depends on D108409-code

## Proposed Changes

* Update `QueryBillingTransactions` to be a functional component
* Pass an optional `transactionType` to `QueryBillingTransactions`
* Pass `transactionType` to a variant of `/me/billing-history` (D108409-code)
* Adjust reducer to reflect optional responses
* This code should only change what is loaded by `QueryBillingTransactions` for the [`BillingHistory` component](https://github.com/Automattic/wp-calypso/blob/a4d3eb3d0318546e2532d3fb7759d7826e1b129e/client/me/purchases/billing-history/main.tsx#L73)
* The `BillingHistory` component [only uses the 'past' data](https://github.com/Automattic/wp-calypso/blob/2f091b6c63dcbf55ed6527c632863d1d68088bcd/client/me/purchases/billing-history/billing-history-list.tsx#L67) loaded by `QueryBillingTransactions`

## Testing Instructions

* Verify my assertion that `BillingHistory` only needs the 'past' key which was sent through the reducer by the API
* Follow testing instructions from D108409-code using the Calypso Live instance created for this PR


JessBoctor:
I didn't realize just how slow `/me/billing` was until I tried it with this patch 😂 

This is a great improvements and works as described. I was able to verify that the api endpoint was changed from `/billing-history` to `/billing-past`.

client/state/billing-transactions/test/actions.js
```diff
@@ -89,7 +89,7 @@ describe( 'actions', () => {
 			} );
 
 			test( 'should dispatch request failure action when request fails', () => {
-				return requestBillingTransactions( 87654321 )( spy ).then( () => {
+				return requestBillingTransactions()( spy ).then( () => {
```
sirbrillig, line 92: Nice find!

client/state/billing-transactions/actions.ts
```diff
@@ -10,19 +10,21 @@ import {
...
-export const requestBillingTransactions = () => {
+export const requestBillingTransactions = ( transactionType?: BillingTransactionsType ) => {
 	return ( dispatch: CalypsoDispatch ) => {
 		dispatch( {
 			type: BILLING_TRANSACTIONS_REQUEST,
 		} );
 
 		return wp.req
-			.get( '/me/billing-history', { apiVersion: '1.3' } )
+			.get( `/me/billing-history` + ( transactionType ? `/${ transactionType }` : '' ), {
```
sirbrillig, line 25: Good job here!

sirbrillig (approved):
Works for me too. 👏

PR #75933, opened 2023-04-19
Help Center: Comment out Launchpad api request

edanzer:
### Proposed Changes

Related to https://github.com/Automattic/wp-calypso/pull/75575

The above commit made the Help Center Launchpad Icon dynamic based on data fetched from a new endpoint. That endpoint has been merged to Jetpack for a week, but is not deployed yet to WordPress.com. While the request fails gracefully and the component renders correctly, we've decided to comment out the api request until the associated endpoint is deployed to avoid potential noise on the failed api request in Sentry or other places. 

### Testing Instructions

Review Time: Short
Testing Time: Short

1. Checkout this branch and run `yarn` and `yarn start` if needed. 
2. Go to any Launchpad-enabled site and go to the posts page: `http://calypso.localhost:3000/posts/SITESLUG`
3. Click on the Help Center (question mark) icon in the upper right, and confirm a) that the help center loads correctly, b) that the launchpad icon shows with 25% progress (hard coded default), and c) that there are no errors or failed requests in the dev tools > network tab.

PR #75987, opened 2023-04-19
Reenable help center launchpad endpoint

edanzer:
### Proposed Changes

Related to: https://github.com/Automattic/wp-calypso/pull/75575 and https://github.com/Automattic/wp-calypso/pull/75933

This PR re-enables the new Launchpad endpoint for the Launchpad icon in Help Center, and sets the progress bar on the icon dynamically based on completed / total steps. This had been commented out in one of the PRs above until the endpoint was deployed as part of the jetpack-mu-plugin. 

launchpad icon

### Testing Instructions

1. Checkout this branch and run yarn and yarn start. 
   - These changes are part of a package. Yarn commands should build the help center package fresh for testing. If for some reason you need to rebuild the package, run `yarn build-packages`.
3. Create or use a launchpad enabled site.
4. Go to `http://calypso.localhost:3000/post/YOURSITESLUG`
5. TEST: Click on the help / question mark icon in the upper right icon and confirm the Launchpad icon shows. 
6. TEST: Confirm the progress status on the icon is being dynamically set (before it was always set to 25%). 
   - Ideally, the percentage complete should directly match launchpad. You can check that by clicking Home on the dashboard, which will redirect you to the launchpad screen. However, note that some of our data may not be accurate, as we're still adding the logic needed to set completion status correctly in the new endpoint. For now, it's probably enough to confirm the progress bar is being set dynamically.

### DIFF
--- a/packages/help-center/src/components/help-center-launchpad.tsx
+++ b/packages/help-center/src/components/help-center-launchpad.tsx
@@ -8,7 +8,7 @@ import { chevronRight, Icon } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { useSelector } from 'react-redux';
 import { getSectionName, getSelectedSiteId } from 'calypso/state/ui/selectors';
-// import { useLaunchpadChecklist } from '../hooks/use-launchpad';
+import { useLaunchpadChecklist } from '../hooks/use-launchpad';
 import { SITE_STORE } from '../stores';
 import type { SiteSelect } from '@automattic/data-stores';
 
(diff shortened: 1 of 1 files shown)


packages/help-center/src/components/help-center-launchpad.tsx
```diff
@@ -8,7 +8,7 @@ import { chevronRight, Icon } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
 import { useSelector } from 'react-redux';
 import { getSectionName, getSelectedSiteId } from 'calypso/state/ui/selectors';
-// import { useLaunchpadChecklist } from '../hooks/use-launchpad';
+import { useLaunchpadChecklist } from '../hooks/use-launchpad';
```
sharpirate, line 11: Can we rename `use-launchpad.tsx` to `use-launchpad-checklist.tsx` since we have another hook that's called `useLaunchpad` and the file name is the same - `use-launchpad.tsx` which can be confusing

jeyip (approved):

packages/help-center/src/components/help-center-launchpad.tsx
edanzer: Upon further discussion, we'll actually be creating a new useLaunchpad hook in data-stores that will retrieve all Launchpad data including checklists, and that can be used here as well as in Launchpad and other places throughout calypso. Once done, we'll be deleting this endpoint, so just going to skip making changes to it now.

Issue #72493, opened 2023-01-24
Stats: Refactor various highlight cards into one component


a8ck3n:
Currently have three different implementations of the highlight cards.

1. https://github.com/Automattic/wp-calypso/tree/trunk/packages/components/src/highlight-cards
2. https://github.com/Automattic/wp-calypso/tree/trunk/client/my-sites/stats/stats-email-top-row
3. https://github.com/Automattic/wp-calypso/tree/trunk/client/my-sites/stats/wordads

The general format is icon, title, and a display value. It might make sense to have the first two be props and the third be presented by the caller (ie: `<Card title='Clicks' icon='arrow'>{ /* child elements */ }</Card>`.

jsnmoon:
A couple of additional refactors I think we could do:

- [ ] Refactor the count comparison card to reuse the count card (or the base card).
- [ ] Make the `heading` and `icon` properties optional for the count and count comparison cards.
- [ ] Refactor `percentCalculator` out of the count comparison card (maybe?)
- [ ] Refactor the count card to include tooltip functionality, which can be disabled via component properties.
- [ ] Deduplicate highlight card styling spread across the Stats codebase.

Two outstanding tasks @a8ck3n noted in #75989:

- [ ] Apply a min-width to each count comparison card so that the [relative count doesn't overflow to the next line](https://github.com/Automattic/wp-calypso/pull/75989#issuecomment-1515964593).
- [ ] Refactor annual highlight cards [to encapsulate all render-related configuration to a single component](https://github.com/Automattic/wp-calypso/pull/75989#discussion_r1172336924).

PR #75989, opened 2023-04-19
Stats: Rename existing highlight cards

jsnmoon:
Related to #72493

## Proposed Changes

* Renames "HighlightCards" to "WeeklyHighlightCards".
* Renames "HighlightCard" to "CountComparisonCard".

These changes are in preparation for a larger refactor for `CountComparisonCard` and `WeeklyHighlightCards`. It also establishes a naming convention for other card collections like `AllTimeHighlightCards` and `SubscriberHighlightCards` that currently exist outside of the components package.

## Testing Instructions

* Open the live branch for this PR.
* Ensure that the traffic and insights pages of Calypso Stats look and work as expected.

## Pre-merge Checklist


a8ck3n:
This tests well for me. Not specific to this PR but I'm wondering if we should adjust the minimum width on these cards to prevent the following:

SCR-20230420-ovdb

My personal preference would be to prevent wrapping of the red/green status text.

client/my-sites/stats/annual-highlights-section/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { AnnualHighlightCards } from '@automattic/components';
+import AnnualHighlightCards from '@automattic/components/src/highlight-cards/annual-highlight-cards';
```
a8ck3n: I'm curious about this change to direct imports. I see it was changed back in a later commit. Was this just to keep the build system happy while the renaming was happening?

packages/components/src/highlight-cards/annual-highlight-cards.tsx
```diff
@@ -50,31 +50,31 @@ export default function AnnualHighlightCards( {
 			</div>
 
 			<div className="highlight-cards-list">
-				<HighlightCard
+				<CountComparisonCard
```
a8ck3n, line 53: Would it make sense to pull the `heading` and `icon` values up a level and provide them when we are setting up the `count` values?

Currently we transform the API data in the caller and provide the `counts` configuration object. If we instead generated an array of card data objects (heading, icon, count) in the caller, we could avoid having configuration spread across two files. We could then map the passed in array here so this component would no longer be concerned about the number of cards being displayed.

a8ck3n (approved):
One more comment. No blockers! Tests great!

jsnmoon:
> I'm wondering if we should adjust the minimum width on these cards to prevent the following (@a8ck3n)

Sounds like a good idea to me!

client/my-sites/stats/annual-highlights-section/index.tsx
jsnmoon: I wanted to verify if our build system was properly tree-shaking unused dependencies for each entry point; I was concerned that exporting additional components in `src/index.tsx` may bloat all bundles importing the components package.

Luckily, our build system does properly tree-shake unused dependencies. I added them back to `index.tsx` to improve the DevUX for consuming these components.

packages/components/src/highlight-cards/annual-highlight-cards.tsx
jsnmoon: Yes, I think that'd be a good next step for refactoring the highlight cards 🙂

client/my-sites/stats/annual-highlights-section/index.tsx
a8ck3n: Nice! Thanks for the clarification!

PR #75990, opened 2023-04-19
Removing Explat test code and launching onboarding-pm test variant

donlair:
## Proposed Changes

* Launch the test variant of as the Explat experiment `paid_media_signup_2023_03_legacy_free_presentation` new control. Context at pbxNRc-2rF-p2.
* Remove the Explat config.

## Testing Instructions
* Checkout this PR and start Calypso locally.
* Navigate to `/start/onboarding-pm` 
* Go through signup using and confirm that it works properly:
* On the select a custom domain, a free domain, and then the skip option;
* Confirm that the option to choose Free is not present if the user selected a custom domain or the skip option.
* Confirm that the Free option IS available if the user chose a free domain.
* Confirm that the 1 year and 2 year term lengths are available on the Plans step (Monthly option should not be presented).

### DIFF
--- a/client/signup/steps/domains/index.jsx
+++ b/client/signup/steps/domains/index.jsx
@@ -26,7 +26,6 @@ import {
 	getFixedDomainSearch,
 } from 'calypso/lib/domains';
 import { getSuggestionsVendor } from 'calypso/lib/domains/suggestions';
-import { loadExperimentAssignment } from 'calypso/lib/explat';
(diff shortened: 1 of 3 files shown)


gmovr (approved):
LGTM
- Tested with free/paid domain, and also "a domain I already own"
- Tested plans toggle (1y/2y), no monthly present and toggles work
- Tested plan selection, and went to checkout fine with a 2y plan.

PR #76015, opened 2023-04-20
Newsletter: Fix subscribers disclaimer font

### LINKED ISSUE #75686: Newsletter onboarding: fix link font-size
These paragraphs in the Add subcribers step have different font sizes. The `p` has `font-size: 14px` and all the links, `font-size: 13px`. It should be **Base/Regular (14px)**.

Image

escapemanuele:
## What
Thesparagraphs in the Add subcribers step have different font sizes. The p has font-size: 14px and all the links, font-size: 13px. It should be Base/Regular (14px).

image

## Testing
1. Live link
2. Reach the Subscribers step in the Newsletter flow
3. Inspect the font sizes

Fixes https://github.com/Automattic/wp-calypso/issues/75686

### DIFF
--- a/packages/subscriber/src/components/add-form/style.scss
+++ b/packages/subscriber/src/components/add-form/style.scss
@@ -190,17 +190,17 @@
 	}
 
 	.add-subscriber__form--disclaimer {
-		font-size: 0.875rem;
+		font-size: $font-body-small;
 		line-height: 1.25;
 		color: var(--studio-gray-50);
 		margin-bottom: 1.5rem;
 
 		a,
 		button {
+			font-size: $font-body-small;
 			color: var(--color-text);
 			text-decoration: underline;
 			height: unset;
-			font-size: 0.875rem;
 
 			&:hover {
 				color: var(--color-text-subtle);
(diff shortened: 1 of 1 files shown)


packages/subscriber/src/components/add-form/style.scss
```diff
@@ -200,6 +200,7 @@
 			color: var(--color-text);
 			text-decoration: underline;
 			height: unset;
+			font-size: 0.875rem;
```
simison: Why this value, and not a variable from the shared typography package?

https://github.com/Automattic/wp-calypso/blob/e7e2d47f075b46e1ddb875f5b4cd0576b8e1f800/packages/typography/styles/variables.scss#L46

The same in higher up at line 193.

If you just remove font-size+line-height lines, does it actually get defaulted to 14px?

simison:
Remember that this component also gets used inside Calypso, so it's important to double-check importing subscribers page.

Added Caribou to review mostly just as an FYI.

packages/subscriber/src/components/add-form/style.scss
escapemanuele: Thanks, changed it to use the variable in both places👌 

I tried removing the font-size and the line-height, but link becomes 13px and text 16px

simison (approved):
Looks good!

Checked also in-product form:

image

bogiii (approved):
Looks good to me as well! 🚀

PR #76024, opened 2023-04-20
Plans 2023: plans data-store foundation

chriskmnds:
## Proposed Changes

It connects the 2023 plans grid to the plans data-store and makes it a dependency for loading the grids i.e. the grids will not render until the plans' data are fetched and loaded in the store.

It does not introduce any selector or other functionality. We should be able to start tuning and using now. The setup will improve as more things from the plans' code are refactored and deprecated.

## Testing Instructions

* Go to `/plans/[SITE]` and ensure the grids are rendered
* Same for `/start/plans`

## Pre-merge Checklist


client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -1214,4 +1227,13 @@ const WrappedPlanFeatures2023Grid = ( props: PlanFeatures2023GridType ) => {
...
+export default withSelect( ( select: SelectFunction, ownProps: PlanFeatures2023GridType ) => {
+	const locale = getLocaleSlug();
+	locale && ( select( PLANS_STORE ) as PlansSelect ).getSupportedPlans( locale );
+	const datastorePlansLoaded = ( select( PLANS_STORE ) as PlansSelect ).hasFinishedResolution(
+		'getSupportedPlans',
+		[ locale ]
+	);
+
+	return { ...ownProps, datastorePlansLoaded };
+} )( WrappedPlanFeatures2023Grid );
```
chriskmnds, line 1242: This isn't necessary, since `PlanFeaturesMain` is connected, but _I think_ no harm in keeping if to ensure this isn't imported and used anywhere in a broken state. It's also the only properly typed usage.

chriskmnds:
I'm investigating the test failures as some mocking is needed and it seems to be non-trivial right now. poked some folks in Slack to see if it rings any bells: p1682072469919839-slack-C02A41GCS

southp:
Adding @aneeshd16 and @jdc91 here to join the fun, as I kept running out of time looking deeper into this and the test failures 😅

client/my-sites/plans-features-main/index.jsx
```diff
@@ -621,6 +623,7 @@ export class PlansFeaturesMain extends Component {
 			hidePlanTypeSelector,
 			is2023PricingGridVisible,
 			planTypeSelectorProps,
+			datastorePlansLoaded = true,
```
southp, line 626: My understanding is that it intentionally defaults to `true` here since the plans grid is not relying on the data store yet. Is that correct? Either way, it would be better if we leave an inline comment so our folks won't be confused when reading this :)

chriskmnds: > Is that correct?

This is only here to get the tests passing. It will never default to `true` in a running environment since the HOC feeds the property through. I couldn't get the `withSelect` HOC mocked, at least not without digging further, hence this. I've seen some fix at one place on WooCommerce I think, but a little beyond me right now.

chriskmnds: > the plans grid is not relying on the data store yet

Well, actually, as of this PR, the grid relies on the data store. It will not render anything until they are fetched in the resolver (hence `datastorePlansLoaded` prop)

southp (approved):
Per the decision made here: p1682588080427139/1682189056.570179-slack-CV2TX2PAN . Let's ship this and iterate from here :)

aneeshd16 (commented):
Hi! Just noting here that a parallel effort is going on to [migrate wp-data stores from registerStore to createReduxStore](https://github.com/Automattic/wp-calypso/issues/74399), which changes the way we access the store.

Do you think we should take up the migration first and then wire up the plans grid? I'm guessing that we'll be among the primary users of the plans data store.

chriskmnds:
We will not be reinvigorating the plans wp-data store as proposed in this PR. Some React-Query hooks have been added to the plans store for fetching the intro offers for the Woo promos and that can be the direction for migrating the rest of pricing into the data store if/when we come to it. 

If future experiments determine that a wp-data store is better, then this PR can serve as a input reference.

Closing now as irrelevant.

PR #76035, opened 2023-04-20
Add editor in reader behind a feature flag

creativecoder:
## Proposed Changes

Adds an instance of isolated-block-editor behind a feature flag, `reader/editor`.

This is meant to be a starting point that we can build on. See pe7F0s-Kk-p2

## Testing Instructions

* Open `/read` on calypso.localhost
* Click on the editor and create a post.
* Click publish to publish the post on your primary site.


package.json
```diff
@@ -304,7 +304,10 @@
 		"@wordpress/base-styles": "4.13.0",
 		"@wordpress/interface@^4.8.0": "patch:@wordpress/interface@npm%3A4.8.0#./.yarn/patches/@wordpress-interface-npm-4.8.0-8a39ad37cf.patch",
 		"@wordpress/data-controls": "2.22.0",
-		"@wordpress/element": "4.20.0"
+		"@wordpress/element": "4.20.0",
+		"@wordpress/block-editor": "10.4.0",
+		"@wordpress/block-library": "7.17.0",
```
tyxla: I wonder if we could pin to the versions we use broadly right now:

* `@wordpress/bock-editor@10.5.0`
* `@wordpress/block-library@7.19.0`

The other forced resolutions here seem to match what we're already using at this time.

This is, of course, temporary, since as @noahtallen mentioned in p4TIVU-auj-p2 we are in the middle of upgrading Calypso to the latest `@wordpress` packages and upgrading to React 18 in the meantime.

config/production.json
```diff
@@ -119,6 +119,7 @@
 		"purchases/new-payment-methods": true,
 		"push-notifications": true,
 		"reader": true,
+		"reader/editor": false,
```
tyxla: Note that a feature flag is always `false` by default, unless specifically set to `true`. That being said, a `false` flag can just be omitted and it will work the same way.

client/reader/post-editor/index.jsx
```diff
@@ -0,0 +1,78 @@
+import IsoloatedEditor, { ToolbarSlot } from '@automattic/isolated-block-editor';
+import { serialize } from '@wordpress/blocks';
+import { Button } from '@wordpress/components';
+import { select, useDispatch } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
```
tyxla: Is there a good reason to use `@wordpress/i18n` and not `i18n-calypso`?

client/data/posts/use-create-new-post.js
```diff
@@ -0,0 +1,28 @@
+import { useCallback } from 'react';
+import { useMutation } from 'react-query';
+import wpcom from 'calypso/lib/wp';
+
+export default function useCreateNewPost( mutationOptions = {} ) {
+	const mutation = useMutation(
+		async ( { siteId, post } ) =>
```
tyxla: The downside with this open `post` approach is that we don't really define a schema of what post data is expected here, and that is prone to errors and missing/unexpected data. Should we try to introduce some sort of schema, or at least document the expected data format better?

client/reader/post-editor/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return (
+		<div className="reader-post-editor">
+			<div className="reader-post-editor__editor">
+				<IsoloatedEditor settings={ editorSettings }>
+					<ToolbarSlot>
+						<Button
+							className="reader-post-editor__publish-button"
+							isPrimary="true"
```
tyxla: Did you mean:

```suggestion
							isPrimary
```

client/data/posts/use-create-new-post.js
```diff
@@ -0,0 +1,28 @@
+import { useCallback } from 'react';
+import { useMutation } from 'react-query';
```
tyxla: Nice to see React Query used for post creation!

Note that we might need to change the import statement since we're in the middle of upgrading the package to v4 in #74381.

client/package.json
```diff
@@ -40,6 +40,7 @@
 		"@automattic/happychat-connection": "workspace:^",
 		"@automattic/help-center": "workspace:^",
 		"@automattic/i18n-utils": "workspace:^",
+		"@automattic/isolated-block-editor": "2.22.0",
```
tyxla: Is there a need to pin it like that, or could we make it less specific with a `^`?

client/reader/post-editor/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+			hasFixedToolbar: true,
+		},
+	};
+
+	function publishPost() {
+		const blocks = select( 'core/block-editor' ).getBlocks();
+
+		if ( blocks && blocks.length > 0 ) {
+			createNewPost( primarySiteId, {
+				content: serialize( blocks ),
```
tyxla, line 54: Do we also need a post title?

client/reader/post-editor/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+	return (
+		<div className="reader-post-editor">
+			<div className="reader-post-editor__editor">
+				<IsoloatedEditor settings={ editorSettings }>
+					<ToolbarSlot>
+						<Button
+							className="reader-post-editor__publish-button"
+							isPrimary="true"
+							onClick={ publishPost }
+							disabled={ isLoading }
```
tyxla, line 68: Should it also be disabled when there are no blocks?

client/reader/following/main.jsx
```diff
@@ -7,6 +9,7 @@ function FollowingStream( { ...props } ) {
 	/* eslint-disable wpcalypso/jsx-classname-namespace */
 	return (
 		<Stream { ...props }>
+			{ config.isEnabled( 'reader/editor' ) && <ReaderPostEditor /> }
```
tyxla: Is this something we should async load? 

Otherwise, we'll bundle this for everyone, including production, even though we don't really render it.

package.json
creativecoder: Unfortunately, changing to those versions starts to trigger console errors--there are probably other package version changes that would need to be in a new version of isolated-block-editor to align with those versions of block-editor and block-library.

Using `wordpress/block-editor@10.5.0` and `@wordpress/block-library@7.19.0`, here's the error I see, when using slash to insert a block: 2ff1d-pb. I also tried changing just `wordpress/block-editor@10.5.0` (with `@wordpress/block-library@ 7.17.0`), but that results in a build error.

client/data/posts/use-create-new-post.js
creativecoder: Thanks for flagging the react-query update!

package.json
tyxla: Makes sense. I hope we'll get to address most of these issues as we upgrade (and maintain) Calypso and all related projects to the latest version of the `@wordpress` packages eventually.

client/reader/post-editor/index.jsx
creativecoder: I had the same idea, but found this surprisingly tricky because `select( 'core/block-editor' ).getBlocks()` seems to only work if run within the scope of IsolatedBlockEditor, rather than within the parent component.

I didn't see an immediate way to pass a function that the button component runs to determine if it's disabled, rather than passing it a predetermined value.

I'll plan to address this in a follow-up, though!

client/reader/following/main.jsx
creativecoder: Great point, now updated to async load.

client/reader/post-editor/index.jsx
creativecoder: It's not technically required, though we'll want to handle this in some way, eventually.

client/package.json
creativecoder: `^` doesn't work, since 0.23.0 updates to React 18 and so is incompatible. But I changed it to 0.22.x.

client/reader/post-editor/index.jsx
creativecoder: Yes, thanks!

client/data/posts/use-create-new-post.js
creativecoder: I've moved the file to .ts and imported the relevant Post type from `@wordpress/core-data` -- let me know if that seems like a good solution.

client/reader/post-editor/index.jsx
creativecoder: Nope, just habit. Now changed to use the Calypso package.

config/production.json
creativecoder: Good point--usually I err on the side of trying to be verbose, but I can see how the false flag values add clutter. Now removed.

creativecoder:
I've rebased this branch off of `try-react-18` and updated `isolated-block-editor` to the latest version. It's not working yet, but there's less version forcing needed than before: pe7F0s-Vt-p2

bogiii (approved):

bogiii:
I've reviewed and tested the part of the code that owns the @Automattic/caribou team; it works as expected without regression.

creativecoder:
I've rebased this branch off of trunk. I wouldn't say the editor works, as there are some UI/style issues, but it builds and you can interact with it without errors.

worldomonation (changes requested):
I am not 100% sure why @Automattic/kitkat was tagged for review, but my concern is that this PR completely breaks Unit Tests.

noahtallen:
> my concern is that this PR completely breaks Unit Tests.

I think this will get resolved after this can be rebased with #78711

PR #76039, opened 2023-04-20
Virtual themes: Open site assembler

### LINKED ISSUE #75332: Enable virtual theme to pass patterns to the Assembler.
### Details

Selecting a virtual theme should open up the theme in the Pattern Assembler so that the user can get all of the goodness that the PA offers - adjust the patterns, colours and fonts etc.

We've mooted a few ideas around changing the preview page in the design picker to accommodate this:
 * Removing or de-emphasizing the 'start from scratch' cta on the preview page
 * Offering the user the choice between 'Selecting', 'Opening in PA with pattern' and 'Opening in PA without pattern'
 * Opening the PA up with the pattern rather than the preview page in the design picker, so users can adjust before selecting
 * Allowing some virtual themes to skip the design picker - maybe it's currently less useful for a relatively full-featured virtual theme like a l-i-b, and the onus should be on working on the blocks not the patterns.

Current preview page:
image

I think the best, fastest next step is to 
 * Make the 'Continue' button always go to the PA, which will pre-populate with the selected pattern
 * ~~Remove the 'Start designing' CTA to the PA~~ @mmtr already removed this
 * Change the virtual theme prefix back to 'Design your own'

We should do a lego-style mini-cft like pekYwv-Ub-p2 before deploying

### Related

https://github.com/Automattic/wp-calypso/pull/75326

mmtr:
Fixes https://github.com/Automattic/wp-calypso/issues/75332

## Proposed Changes

Open the site assembler after selecting a virtual theme in the design picker so users can further customize the assembled design if the following conditions are satisfied:
- Viewport is large ([the assembler step is disabled otherwise](https://github.com/Automattic/wp-calypso/blob/25dfa286b5e056dac78e11e6b537e4122a0fb6de/packages/design-picker/src/components/pattern-assembler-cta/index.tsx#L26-L28)).
- The `pattern-assembler/dotcompatterns` feature flag is enabled ([patterns used by virtual themes are unavailable otherwise](https://github.com/Automattic/wp-calypso/blob/25dfa286b5e056dac78e11e6b537e4122a0fb6de/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/patterns-data.ts#L593-L595)).

https://user-images.githubusercontent.com/1233880/233394594-460e263c-40dd-49fa-9cba-637ee5ef72ee.mov

## Testing Instructions

- Select a virtual theme and continue with the design selection.
- Make sure you land in the site assembler.
- Make sure the pattern of the virtual theme is automatically pre-selected.
- Make sure you can add/remove/reorder patterns.
- Complete the site setup.
- Verify you get a site with BC3 and a homepage template with the expected patterns.
- Go back to `/setup/site-setup/designSetup?siteSlug=<SITE_DOMAIN>`.
- Resize the browser window to simulate a mobile viewport.
- Make sure you don't land in the site assembler after selecting a virtual theme.


client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -461,6 +463,27 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow, stepName } ) => {
...
+			isDesktop &&
+			isEnabled( 'pattern-assembler/dotcompatterns' );
+
+		if ( shouldGoToAssembler ) {
+			const assemblerDesign = {
+				..._selectedDesign,
+				design_type: 'assembler',
+			} as Design;
+
+			setSelectedDesign( assemblerDesign );
```
BogdanUngureanu, line 460: In the with theme assembler flow we are [also setting an intent](https://github.com/Automattic/wp-calypso/blob/1bc90b4d4fab8819184adba437ade50433839555/client/landing/stepper/declarative-flow/with-theme-assembler-flow.ts#L30). Do we need to do here too (I'm not familiar with intents 😅 )?

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -461,6 +463,27 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow, stepName } ) => {
...
+		const shouldGoToAssembler =
+			_selectedDesign?.is_virtual &&
+			_selectedDesign?.slug === BLANK_CANVAS_DESIGN.slug &&
+			isDesktop &&
+			isEnabled( 'pattern-assembler/dotcompatterns' );
+
+		if ( shouldGoToAssembler ) {
+			const assemblerDesign = {
+				..._selectedDesign,
+				design_type: 'assembler',
```
BogdanUngureanu: Maybe we should use the BLANK_CANVAS_DESIGN.design_type constant. :)

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
mmtr: I don't think we need it since we don't do that during the normal site setup flow:

https://github.com/Automattic/wp-calypso/blob/9ce211ef8474966426bbe9a95980932fbc4cfed3/client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx#L526-L548

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
mmtr: Technically, the virtual theme could use a different theme than Blank Canvas, so I'm bit reluctant to use data from a constant that seems to be specific to Blank Canvas.

But given that both virtual themes and the site assembler only work with Blank Canvas right now, I guess it's ok for now.

BogdanUngureanu (approved):
Works as expected! 🎉 🚢

okmttdhr (approved):
Worked as expected!

One thing I noticed is there's a weird space at the bottom when I selected one of the Link in Bio virtual theme. Is this expected? (This did not happen when I selected `Blank Canvas - single page`) CC: @Automattic/lego 

https://user-images.githubusercontent.com/5287479/233521851-2ae70a1e-a0f1-40c1-a7c3-90e11219c589.mov

mmtr:
> One thing I noticed is there's a weird space at the bottom when I selected one of the Link in Bio virtual theme. Is this expected?

Good catch @okmttdhr! Seems this a generic issue in the site assembler when the pattern is too big for the preview area? It also happens when opening the assembler from scratch:

https://user-images.githubusercontent.com/1233880/233655907-0486158f-64a0-4e0e-b362-913200e021ea.mov

miksansegundo:
> One thing I noticed is there's a weird space at the bottom when I selected one of the Link in Bio virtual theme. Is this expected?

> Seems this a generic issue in the site assembler when the pattern is too big for the preview area?

I working on fixing that [issue](https://github.com/Automattic/wp-calypso/issues/76129). It's related to patterns with `min-height: 100vh` (like LIB patterns).

Issue #75335, opened 2023-04-05
Fixing the with-theme flows


dsas:
### Details

We discovered some issues with the with-theme flow a [while ago](https://github.com/Automattic/dotcom-forge/issues/1454#issuecomment-1397197929), which prompted us to do a walkthrough paYJgx-2Vo-p2. 

In a nutshell, the problem is if the user selects a paid theme, even if they pay for it, we never activate it; instead, we activate a very old random theme from a hardcoded list of fallback themes. Bespin is addressing this for atomic flows but this still happens for premium themes when you choose the premium plan.

### Steps

- [x] Replace the hardcoded fallback theme list with the current default Core theme. (fbhepr%2Skers%2Sjcpbz%2S.pbasvt%2Sjcpbz%2Qbcgvbaf.cuc%3Se%3Qp62722n8%23867%2Q887-og)
- [x] In the /with-theme flow, ensure we actually activate the chosen theme as long as the site can support it.
- [ ] Install a fallback theme if the site doesn’t support the chosen theme.

### Related

https://github.com/Automattic/dotcom-forge/issues/2045

dsas:
@BogdanUngureanu the middle 'step' here can be ticked off right? It works, it's just not an ideal UX?

BogdanUngureanu:
@dsas yup, that's right! I've ticked that point :)

BogdanUngureanu:
>  Install a fallback theme if the site doesn’t support the chosen theme.

@dsas I don't think this is necessary anymore, right? After the change from D109689-code lands, the default theme (fallback) would be TT3.

dsas:
Sounds good 🎉

BogdanUngureanu:
Both of the tasks have been completed. Closing this issue. :)

PR #76040, opened 2023-04-20
Fix with-theme flow for premium, WooCommerce and DotORG themes

BogdanUngureanu:
If the theme is free, premium and WooCommerce, the user will now be pointed to the designSetup page.

If the theme is dot-org, the user will be sent to /marketplace/theme/install page.

However, this PR doesn't fix the installation issues for Marketplace themes.

Related to #75335

## Proposed Changes

* Redirect the users to designSetup for free, premium and WooCommerce themes and hide the back button.
* Redirect the users to the marketplace/install page for dot-org themes.
* From a technical PoV, the user is routed based on the `theme_type` URL parameter

## Testing Instructions

* Use the live link generated in this PR.
* Open an Incognito window and go to /themes (loggedout)
* Select a free, a premium and a WooCommerce theme and copy the links from the `Pick design` button. e.g: `/start/with-theme?ref=calypshowcase&theme=ctlg&style_variation=corn`
* To simplify things, go to your browser window where you are already logged-in and go to that URL.
* Create a new website and notice that you are now redirected to /designSetup page (with the theme preselected and the back button missing).
* For dot-org themes, you'll need to go directly to /start/with-theme/user?ref=calypshowcase&theme=astra&theme_type=dot-org and create a new site.
* After purchasing a business plan, you'll be redirected to the marketplace theme install page.
* However, if you haven't purchased a Business plan, the page will display an upgrade nudge saying that you need to purchase a Business plan in order to be able to install the theme. 

## Pre-merge Checklist


client/signup/config/flows.js
```diff
@@ -133,6 +133,16 @@ function getChecklistThemeDestination( { flowName, siteSlug, themeParameter } )
...
 }
 
+function getWithThemeDestination( { siteSlug, themeParameter, styleVariation, themeType } ) {
+	if ( 'dot-org' === themeType ) {
+		return `/marketplace/theme/${ themeParameter }/install/${ siteSlug }`;
+	}
+
+	const style = styleVariation ? `&style=${ styleVariation }` : '';
+
+	return `/setup/site-setup/designSetup?siteSlug=${ siteSlug }&theme=${ themeParameter }${ style }&hideBack=true`;
```
mmtr, line 143: What do you think about excluding free themes here? Users don't need to perform any additional step to activate them, so adding one extra step could affect the signup process.

BogdanUngureanu: Yeah, that's true. I wanted to leave it for two reasons:
- It would be consistent with the Premium and Woo Flows
- It will give another chance to users to select a style variation

Being consistent with Premium and Woo, means that with-theme will have two separate flows, instead of three:
 - free, premium, Woo
 - DotOrg (and in the future probably Marketplace)

I'm also fine with skipping the flow, as you said.

mmtr:
When testing with a Free theme (Storia), I get a blank screen caused by a `TypeError: e.designs is not iterable` JS error:

Screenshot 2023-04-21 at 16 06 20

BogdanUngureanu:
I was getting it locally too, strange, it didn't happen before. I've merged trunk and the error went away. Let's see if it breaks again on calypso.live. Nice find! :)

BogdanUngureanu:
Works as expected now, so the PR is back to RfR. :)

mmtr (approved):
Nice! The flow feels disconnected since the `designSetup` screen is clearly not part of the `with-theme` flow, but this is much better than a broken flow!

mmtr:
For WooCommerce themes, I think the next step should be to ensure that the theme is not added to the shopping cart as a one-off purchase (since they cannot be bought that way).

BogdanUngureanu:
> For WooCommerce themes, I think the next step should be to ensure that the theme is not added to the shopping cart as a one-off purchase (since it cannot be bought that way).

Yeah, I've noticed that too but thought it should be handled separately (probably when we'll deal with the one-off premium theme purchases), even though the WooCommerce themes shouldn't be purchased as separate themes in the first place.

PR #76070, opened 2023-04-21
Fix `planTypes` prop of the PlansFeaturesMain component and some accompanied refactoring.

southp:
## Proposed Changes

This PR is primarily for fixing up the `plansTypes` prop so the new pricing grid can accept a list of plan types to show like in the case of https://github.com/Automattic/wp-calypso/pull/75174/. Before this PR, by doing so the Free plan and the Enterprise plan would disappear when toggling to the monthly term. It is because the Free plan and the Enterprise plan don't really have a term, and they only have a placeholder-like annual term in their data definition. This PR fixes it by adding special cases for them in the function for converting plan types into actual plans.

Although that part is a smaller fix, what's happening underneath is actually a series of data processing flow of:

1. Given a list of "plan types".
2. Derive "plans" data from the given types and the term.
3. Derive "visible plans" data from the plans data, given related props like `hidePersonalPlan`, `hidePersonalPlan`, `flowName`, etc.

I feel the current code doesn't outline this processing pipeline clearly, so I've done some refactoring here aiming at clarifying it from the code level. A side effect is that the whole processing will now run only once per render; currently it is run multiple times per render since `getPlansForPlanFeatures` and `getVisiblePlansForPlanFeatures` are called at multiple places per rendering.

Some other misc tweaks:

* `getPlanFeatures()` is now `renderPlanFeatures()` to match its meaning better.
* `isEnabled( 'plans/personal-plan' )` check is removed since we've enabled it everywhere for ages already.

Lastly, I chose to not updating the `plansTypes` prop name as discussed in https://github.com/Automattic/wp-calypso/pull/75174#discussion_r1171167577 to limit the scope of this PR. That name will be up for iterations for sure :) 

## Testing Instructions

Confirm the plans page the same as in production for the following flows. Both the new pricing grid and the legacy one should work as it is.

* /start/ 
* /start/domain/ 
* /start/onboarding-pm/ 
* /setup/newsletter/
* /setup/link-in-bio/
* /plans while being logged-in

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plans-features-main/index.jsx
+++ b/client/my-sites/plans-features-main/index.jsx
@@ -384,7 +384,7 @@ export class PlansFeaturesMain extends Component {
 			return plan ? [ ...accum, plan ] : accum;
 		}, [] );
 
-		return this.props.showTreatmentPlansReorderTest ? plans.reverse() : plans;
+		return plans;
 	}
 
 	isPersonalCustomerTypePlanVisible() {
--- a/client/my-sites/plans-features-main/index.jsx
+++ b/client/my-sites/plans-features-main/index.jsx
@@ -228,7 +228,7 @@ export class PlansFeaturesMain extends Component {
 		);
 	}
 
-	renderPlanFeatures() {
+	renderPlanFeatures( plans, visiblePlans ) {
 		const {
 			basePlansPath,
 			currentPurchaseIsInAppPurchase,
@@ -257,8 +257,6 @@ export class PlansFeaturesMain extends Component {
 			isPlansInsideStepper,
 		} = this.props;
 
-		const plans = this.getPlansForPlanFeatures();
-		const visiblePlans = this.getVisiblePlansForPlanFeatures( plans );
 		const legacyText =
 			locale === 'en' ||
 			hasTranslation(
@@ -310,7 +308,7 @@ export class PlansFeaturesMain extends Component {
 					isLandingPage={ isLandingPage }
 					isLaunchPage={ isLaunchPage }
 					onUpgradeClick={ onUpgradeClick }
-					plans={ visiblePlans }
+					plans={ plans }
 					redirectTo={ redirectTo }
 					visiblePlans={ visiblePlans }
 					selectedFeature={ selectedFeature }
@@ -571,7 +569,7 @@ export class PlansFeaturesMain extends Component {
 			return this.show2023OnboardingPricingGrid( plans, visiblePlans );
 		}
 
-		return this.renderPlanFeatures( plans );
+		return this.renderPlanFeatures( plans, visiblePlans );
 	}
 
 	render() {
@@ -586,7 +584,6 @@ export class PlansFeaturesMain extends Component {
 		/*
 		 * We need to pass all the plans in order to show the correct features in the plan comparison table.
 		 * Pleas use the getVisiblePlansForPlanFeatures selector to filter out the plans that should not be visible.
-		 * Also, note that the old pricing gird only needs the visible plans. Thus, when it's the old pricing grid,
 		 * we pass `visiblePlans` to its `plans` prop.
 		 */
 		const plans = this.getPlansForPlanFeatures();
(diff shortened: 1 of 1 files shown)


client/my-sites/plans-features-main/index.jsx
```diff
@@ -441,6 +383,8 @@ export class PlansFeaturesMain extends Component {
 
 			return plan ? [ ...accum, plan ] : accum;
 		}, [] );
+
+		return this.props.showTreatmentPlansReorderTest ? plans.reverse() : plans;
```
aneeshd16: Since this test is no longer active, can we return `plans` here? Removing references to this test requires a more comprehensive cleanup, but since we are touching on this part, we might as well remove this reference.

client/my-sites/plans-features-main/index.jsx
```diff
@@ -602,16 +555,15 @@ export class PlansFeaturesMain extends Component {
...
 	}
 
-	renderPlansGrid() {
+	renderPlansGrid( plans, visiblePlans ) {
 		const { shouldShowPlansFeatureComparison, is2023PricingGridVisible } = this.props;
 
-		if ( is2023PricingGridVisible ) {
-			return this.show2023OnboardingPricingGrid();
+		// TODO: need to figure out if we can deprecate `shouldShowPlansFeatureComparison`
+		if ( is2023PricingGridVisible || shouldShowPlansFeatureComparison ) {
```
aneeshd16: This conditional means that the new pricing grid could possibly be shown in cases where `is2023PricingGridVisible` is false and `shouldShowPlansFeatureComparison` is true. This is an undocumented "feature", and I think we should solely rely on `is2023PricingGridVisible` to decide the visibility of the new grid since that's well documented.

If this case has actually happened (`is2023PricingGridVisible` = false and `shouldShowPlansFeatureComparison` = true), and the new pricing grid is being shown, then that should be considered a bug since it's unintentional.

I think we should remove this var from the conditional, and in the future, get rid of it altogether. 🙂

aneeshd16 (commented):
Tests out! I left some minor code comments below.

client/my-sites/plans-features-main/index.jsx
southp: Good call. I've removed it in https://github.com/Automattic/wp-calypso/pull/76070/commits/707f2ad5cc743f58f60c9f2a0287ed0cf30a55b1

client/my-sites/plans-features-main/index.jsx
southp: Great catch! It's actually a quite major logical flaw. 

After digging a bit more and investigating [how plans-features-main-pm uses this prorperty](https://github.com/Automattic/wp-calypso/blob/trunk/client/signup/steps/plans-pm/plans-features-main-pm.jsx#L45), which happens to be a copy of how things used to be, I now understand that that prop is used to mainly determine whether to render `PlansFeaturesComparison` or `PlansFeatures`; the former is for the desktop, and the latter is for the mobile.

To better clarify it, I've done some more rearragnement in https://github.com/Automattic/wp-calypso/pull/76070/commits/3cf40524a692a095997880d04cdb2cb0136ae29d to make this clearer by introducing a `renderLegacyPricingGrid()` function. 

I've also added a note that the legacy component should be loaded in async as a follow-up, which might have quite some implication on the page load speed.

aneeshd16 (approved):
LGTM!

client/my-sites/plans-features-main/index.jsx
```diff
@@ -121,120 +120,72 @@ export class PlansFeaturesMain extends Component {
...
+				customerType,
+				isJetpack,
+				availablePlans: visiblePlans,
+			} ),
+			siteId,
+			isReskinned,
+			isPlansInsideStepper,
+			intervalType,
+			hidePlansFeatureComparison,
+		};
```
aneeshd16, line 156: Our next step should probably be to convert this file to TS. I see that the new pricing grid does not use a lot of these props, and it would be great if we can trim this file further.

southp: Totally agree :)

PR #76079, opened 2023-04-21
Pending subscriptions tab visible for EN non-wp users only

JuanLucha:
## Proposed Changes

The pending subscriptions screen in the Subscription Manager will be visible only for external users with 'en' locale.
Screenshot 2023-04-21 at 11 02 46

## Testing Instructions

1. Build this PR locally.
2. Navigate to http://calypso.localhost:3000/subscriptions/settings?locale=pt-br (or specify some other non-EN locale). → When you click on the "Pending" tab, you should be directed to https://wordpress.com/email-subscriptions/?option=pending.
3. Navigate to http://calypso.localhost:3000/subscriptions/settings?locale=en (English locale) → When you click on the "Pending" link now, you should end up on the new "Pending" view in Calypso.

### DIFF
--- a/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
+++ b/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
@@ -60,7 +60,7 @@ const TabsSwitcher = () => {
 						{ translate( 'Comments' ) }
 					</NavItem>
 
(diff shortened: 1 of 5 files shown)


ivan-ottinger (approved):
Looks good and works as expected. 👍🏼👍🏼

The `Pending` string has been already translated, so it is alright to expose the whole "Pending" tab to all users.

PR #76084, opened 2023-04-21
Revert "Stats: Add new SubscribersHighlightSection component to Subscribers page"

a8ck3n:
Reverts Automattic/wp-calypso#75956 — build broken after merging.


jsnajdr (approved):
Approving so that you can move forward merging it.

PR #76085, opened 2023-04-21
E2E: Refactor Editor locator retrieval

WunderBart:
## Proposed Changes

Create async getters for the `editor` and `canvas` locators so we don't need to worry whether they're iframed or not. This change will:

- Allow us to drop the use of the initial params (Is the site Atomic or Simple? Is the theme block-based?).
- Allow us to quickly change themes without the need to determine what kind of theme it is (block-based vs. classic).
- Guard us against failures caused by iframe-related updates (e.g., theme switching to a block-based one).

## How to test

All the CI editor-related E2E jobs should pass.


packages/calypso-e2e/src/lib/components/editor-window.ts
```diff
@@ -0,0 +1,60 @@
...
+export class EditorWindow {
+	public page: Page;
+
+	/** */
+	constructor( page: Page ) {
+		this.page = page;
+	}
+
+	/** */
+	async getEditorFrame(): Promise< Page | Frame > {
```
worldomonation: I have a couple of questions here.

1. Was it difficult or unfeasible to return a single value in order to abstract the underlying implementation?
2. What's the reason for returning a `Page | Frame` instead of the more "modern" `Locator | FrameLocator`?

Let me elaborate for each question.

For 1, in the current implementation, I tried to present a unified interface in the form of returning a `Locator` to a top-level element of the page (un-iframed) or the first feasible element (iframed). You can see this in [these lines](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-e2e/src/lib/pages/editor-page.ts#L87-L90).

While the interfaces of Frame and Page are nearly identical, I would argue that it would be preferable to have a single return value so the consumers aren't having to cast the types.


Point 2 is the one that I comparatively have a stronger opinion on. 
I feel that we should use Locators as much as possible because Frames and many Page methods are being de-emphasized lately. I'm wondering if you chose to use Page/Frame because Locators haven't yet implemented a `waitForFunction` method yet.

worldomonation (commented):
If I'm following correctly, this PR proposes that we go back to calling the `getEditorFrame` (or equivalent method if switching to Locator) for each time we need to interact with the editor window.

I won't lie that I don't think this method is the best way because many issues can arise in a hypothetical scenario where the consumer forgets to call this method, the frame itself changes, or Playwright deprecates Frames. 

But, I am cognizant of the fact that the current method we use is also brittle and unsustainable. So maybe by trading a little bit of elegance for practicality our Editor interactions can be more reliable.

packages/calypso-e2e/src/lib/components/editor-window.ts
WunderBart: > While the interfaces of Frame and Page are nearly identical, I would argue that it would be preferable to have a single return value so the consumers aren't having to cast the types.

Fair point. We could return the `page.mainFrame()` (guaranteed to persist) to have a single type return value. 

> I feel that we should use Locators as much as possible because Frames and many Page methods are being de-emphasized lately. I'm wondering if you chose to use Page/Frame because Locators haven't yet implemented a waitForFunction method yet.

I don't have a strong opinion on this one, TBH, aside from the fact that it does give us more control and that it's difficult (at the current state of locators) to retrieve the frame parent from a locator in case it's needed. I can see that the `waitForFunction` doesn't seem to be that widely used, though (which is good).
The implementation in this PR allows for an easy transition from `Page/Frame` to `Locator`, so I'll test that on another branch, and we'll decide which one would be best to go with. How does that sound?

WunderBart:
Thanks for the feedback, @worldomonation - much appreciated!

> I won't lie that I don't think this method is the best way because many issues can arise in a hypothetical scenario where the consumer forgets to call this method, the frame itself changes, or Playwright deprecates Frames.

A couple of thoughts on this:

> consumer forgets to call this method

I might be missing something, but I don't think that a consumer forgetting to call this method would be a case because otherwise, they won't be able to interact with the editor. The parent has to be retrieved before interacting with the Editor/canvas.

> the frame itself changes

The frame itself can change (e.g., navigate to a different URL) but it won't change the reference to the object - it will still be available via the same reference (like the Page is when navigating). 

> Playwright deprecates Frames

This sounds unlikely, as Frames seem to be at the very base of the core structure, but one never really knows - Playwright does deprecate a lot of the API! 😅 I'd say there can be more Frame methods that can become deprecated, but ultimately the Frame itself will stay. 

---

The thing that worries me the most in the context of working with Frames, is that there's no build-in mechanism to wait for the frame object itself. `page.frame()` is synchronous, and if the frame is at the time unavailable, we'll get `null` in return. Also, we can't retrieve it via the `elementHandle` because it's been deprecated. 😄 That's why I am using locators to pierce through iframes in order to determine when they're ready to be retrieved as Frames, but it feels like a bit of a hacky way to do that. As [mentioned above](https://github.com/Automattic/wp-calypso/pull/76085#discussion_r1179349115), I'll try switching to Locators and see how it goes!

WunderBart:
@worldomonation, this one's ready for a proper review now! 🙏

packages/calypso-e2e/src/lib/components/editor-window.ts
worldomonation: > The implementation in this PR allows for an easy transition from Page/Frame to Locator, so I'll test that on another branch, and we'll decide which one would be best to go with. How does that sound?

That sounds fair to me. Let's keep it simple(r) in this PR by sticking to the Frame/Page dual-type or a single Frame return type.

Edit: it looks like I didn't read the rest of your (later) comments - you've already switched to using Locators.

worldomonation:
Looks like there's a failure in the `test/e2e/specs/blocks/blocks__shallowly_smoke_test_popular_blocks.ts` file that's causing the Unit Tests to fail:

```typescript
 test/e2e/specs/blocks/blocks__shallowly_smoke_test_popular_blocks.ts(64,39): error TS2554: Expected 1 arguments, but got 2.
```

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -148,73 +104,28 @@ export class EditorPage {
...
-		const timeout =
-			this.target === 'atomic' ? EXTENDED_EDITOR_WAIT_ATOMIC_TIMEOUT : EXTENDED_EDITOR_WAIT_TIMEOUT;
-
 		// In a typical loading scenario, this request is one of the last to fire.
 		// Lacking a perfect cross-site type (Simple/Atomic) way to check the loading state,
 		// it is a fairly good stand-in.
 		await Promise.all( [
 			this.page.waitForURL( /(post|page|post-new.php)/ ),
-			this.page.waitForResponse( /.*posts.*/, { timeout: timeout } ),
+			this.page.waitForResponse( /.*posts.*/, { timeout: 60 * 1000 } ),
```
worldomonation, line 114: **question:non-blocking** - is 60s going to be long enough for AT loading, where it tries to load the iframed version then redirects to the "raw" wp-admin based edtior?

If it is, would it make sense to export the `EditorComponent.EDITOR_TIMEOUT` value so that we have one defined value for the loading timeout?

packages/calypso-e2e/src/lib/pages/editor-page.ts
```diff
@@ -368,7 +294,7 @@ export class EditorPage {
 		openInlineInserter: OpenInlineInserter
 	): Promise< ElementHandle > {
 		// First, launch the inline inserter in the way expected by the script.
-		await openInlineInserter( this.editorWindow );
+		await openInlineInserter( await this.editor.canvas() );
```
worldomonation, line 301: **comment** - I see what you mean in your [response](https://github.com/Automattic/wp-calypso/pull/76085#issuecomment-1525938711) to my (admittedly) poorly-phrased comment:

> I might be missing something, but I don't think that a consumer forgetting to call this method would be a case because otherwise, they won't be able to interact with the editor. The parent has to be retrieved before interacting with the Editor/canvas.

Instead of placing the burden of retrieving the editor and canvas, you've built it into the individual methods in `EditorPage`. So as long as developers implementing new methods for EidtorPage calls the relevant methods to retrieve the editor/canvas, things will work just fine.

packages/calypso-e2e/src/lib/components/editor-component.ts
```diff
@@ -0,0 +1,99 @@
...
+	 * Returns the Editor parent locator. It will automatically resolve to the
+	 * proper locator, regardless if the Editor is Gutenframed or not.
+	 */
+	async parent(): Promise< Locator > {
+		if ( this.parentLocator ) {
+			return this.parentLocator;
+		}
+
+		try {
+			this.parentLocator = await Promise.any( [
```
worldomonation, line 34: **comment** - when did we start being able to use `Promise.any`!? When I last tried to use it, VSCodium rejected my attempts. This is great to see!

packages/calypso-e2e/src/lib/components/editor-component.ts
```diff
@@ -0,0 +1,99 @@
...
+				this.waitForUnframedCanvas(),
+			] );
+		} catch ( _error ) {
+			throw new Error( 'Timed out waiting for the Editor canvas' );
+		}
+
+		return this.canvasLocator;
+	}
+
+	/** */
```
worldomonation: **issue:blocking** - probably just an oversight, but all new methods in this PR needs some form of JSDocs.

worldomonation (approved):
Kudos on wrangling this quite nasty PR. I think your approach is much better than a tunnel-visioned me could come up with! 

Note: I've approved the PR in the interest of not delaying merge on this once the single blocking request is addressed.

packages/calypso-e2e/src/lib/pages/editor-page.ts
WunderBart: > is 60s going to be long enough for AT loading, where it tries to load the iframed version then redirects to the "raw" wp-admin based editor?

60s is the value [we previously used](https://github.com/Automattic/wp-calypso/pull/76085/files/48843472678df6810801eadad3a97b775146a060#diff-9a431df4f005592b6887e18be632044cbc38b331da80fc73f9a0c9fa7efcc5beL44-L45) specifically for the AT editor timeout, so it should be good at this point.

> If it is, would it make sense to export the EditorComponent.EDITOR_TIMEOUT value so that we have one defined value for the loading timeout?

Ultimately, I think we shouldn't need that timeout value to be used outside the `EditorComponent`. Waiting for the editor parent/canvas should be all we need to verify that they're ready to be interacted with (thinking optimistically 😛). For example, the above `waitForResponse(...)` should be replaceable with `await this.editor.parent()` since it means that the page/iframe is ready. I didn't want to do that in this PR since it requires a separate round of testing, though. If that's OK, I'd keep it as is and I'll address this bit in a follow-up PR.

packages/calypso-e2e/src/lib/pages/editor-page.ts
WunderBart: Indeed! Just to be clear: Each [component that we use to compose the EditorPage](https://github.com/Automattic/wp-calypso/blob/912231aadca2662b6527737f38860433075852e5/packages/calypso-e2e/src/lib/pages/editor-page.ts#L71-L87) needs to have the [EditorComponent](https://github.com/Automattic/wp-calypso/blob/912231aadca2662b6527737f38860433075852e5/packages/calypso-e2e/src/lib/components/editor-component.ts) instance passed to be able to use the parent/canvas getters. Regarding the public use of the EditorPage instance (inside tests), we still delegate the base element getters as we used to, with the following changes:
- sync `getEditorWindowLocator()` is now async `getEditorParent()`, and
- sync `getEditorCanvasLocator()` is now async `getEditorCanvas()`. 


It's very much like we did before, just the getters are now async. Hope it makes sense!

packages/calypso-e2e/src/lib/components/editor-component.ts
WunderBart: https://github.com/Automattic/wp-calypso/pull/76085/commits/912231aadca2662b6527737f38860433075852e5

PR #76092, opened 2023-04-21
Updates to writing prompt block in full post reader

davemart-in:
## Description

It was brought to my attention yesterday that the new writing prompts block was being shown on the reader full post view without any formatting. This in turn is affecting the native mobile apps. There is an internal discussion related to the mobile apps here: pctCYC-Mc-p2

### Before

before

### After

after

## Testing Instructions

1. Apply this PR locally or click the "Calypso live" link below
2. Navigate to `/read/blogs/167232230/posts/5558`


client/blocks/reader-full-post/_content.scss
```diff
@@ -442,3 +442,70 @@
...
+
+	.jetpack-blogging-prompt__label {
+		font-family: sans-serif;
+		font-size: $font-body-small;
+		margin-bottom: 16px;
+		padding-left: 22px;
+		position: relative;
+
+		&::before {
+			background: no-repeat url(https://s0.wp.com/wp-content/mu-plugins/jetpack-plugin/production/_inc/blocks/images/icon-cd4e9ecf53fadf95246d.svg);
```
creativecoder: I'm worried this url could change unexpectedly, given the hash as part of the filename. I think we should just add the icon directly to Calypso and load it from `calypso/assets/images/icons/light-bulb.svg`

client/blocks/reader-full-post/_content.scss
```diff
@@ -442,3 +442,70 @@
...
+			position: absolute;
+			width: 24px;
+			top: 1px;
+		}
+	}
+
+	.jetpack-blogging-prompt__text {
+		font-size: $font-title-medium;
+		margin-bottom: 24px;
+	}
```
creativecoder, line 525: Let's add styles for the quote variant, as well. Here's what we have in Jetpack.

```css
.jetpack-blogging-prompt.is-style-quote {
  border: unset;
  border-left: 4px solid #2f2f2f; /* $gray-800 */
  margin-bottom: 16px; /* $grid-unit-20 */
  padding: 0 24px;
}
```

creativecoder (commented):
This looks much better! I left a couple of suggestions, otherwise looks good to go.

client/blocks/reader-full-post/_content.scss
```diff
@@ -442,3 +442,70 @@
...
+	border: 1px solid var(--color-neutral-5);
+	border-radius: 2px;
+	box-sizing: border-box;
+	margin-bottom: 24px;
+	padding: 24px;
+
+	.jetpack-blogging-prompt__answers {
+		display: flex;
+
+		.image-wrapper {
```
thomashorta, line 471: For mobile it doesn't seem that there is a `div` with this class wrapping the `img` tag, so these styles are not being applied, causing the answers section to look bad on the mobile apps.

Here's an example of block that reaches the mobile clients (both iOS and Android):
```html
<div class="wp-block-jetpack-blogging-prompt jetpack-blogging-prompt">
    <div class="jetpack-blogging-prompt__label">
        Daily writing prompt
    </div>
    <div class="jetpack-blogging-prompt__text">
        Describe a random encounter with a stranger that stuck out positively to you.
    </div>
    <div class="jetpack-blogging-prompt__answers">
        <img class="jetpack-blogging-prompt__answers-gravatar" aria-hidden="true" src="redacted-url">
        <img class="jetpack-blogging-prompt__answers-gravatar" aria-hidden="true" src="redacted-url">
        <img class="jetpack-blogging-prompt__answers-gravatar" aria-hidden="true" src="redacted-url">
        <a class="jetpack-blogging-prompt__answers-link" href="https://wordpress.com/tag/dailyprompt-1912" target="_blank" rel="external noreferrer noopener">View all responses</a>
    </div>
</div>
```

I replaced the actual gravatar URLs with `redacted-url` just to not expose real data here, but they are valid URLs.

This is an example of how it looks on mobile.

client/blocks/reader-full-post/_content.scss
```diff
@@ -442,3 +442,70 @@
 	padding-left: 16px;
 	padding-right: 16px;
 }
+
+// Writing Prompts Block Styles
+.reader-full-post__story-content .jetpack-blogging-prompt {
+	border: 1px solid var(--color-neutral-5);
```
thomashorta: On the mobile app readers, this border color is not being properly loaded as we currently don't have that in the hardcoded styles we inject in the Reader WebView. I could find the code where we add those variables, so I believe we need to add this one as well for mobile.

It doesn't seem like there are actions to be taken on your side @davemart-in, but I am just commenting here so we keep track of this need.

[Here's the Android code](https://github.com/wordpress-mobile/WordPress-Android/blob/2b6c76488a53c9b4bf1baf4b543de46f7d954e46/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java#L541-L551) (cc @RenanLukas @develric)
[Here's the iOS code](https://github.com/wordpress-mobile/WordPress-iOS/blob/cc9a26e8ebc5560ac4e59cdcbe234c714e5b26f8/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift#L173-L192) (cc @dvdchr @wargcm)

client/blocks/reader-full-post/_content.scss
thomashorta: On the mobile app readers this is not looking great in Dark Mode, since this has a very dark gray color. I wonder if there's any way to tint the icon programmatically using CSS in this case. If there is, we (mobile team) will need to add some special treatment here inside the Reader renderers.

Here's an example of what that looks like on Android and iOS:
Android | iOS
--- | ---
 | 

**Note:** For supporting dark and light themes we already explicitly set some colors for the text and other things, so I believe it would make sense to add the special treatment on the mobile Reader renderers, using CSS ideally, though I have no idea how to tint icons like this.

thomashorta (commented):
I added a comment in the original reference (pctCYC-Mc-p2#comment-847) with some issues I found on the mobile app reader and I also added some comments about them here, for the ones I could identify the cause.

client/blocks/reader-full-post/_content.scss
thomashorta: The answer avatars are still not looking better on mobile, even after the latest changes here.

I wonder if something like this could work for both web and mobile clients, given the HTML:
```scss
.jetpack-blogging-prompt__answers {
    display: flex;

    .image-wrapper {
        height: 24px;
        width: 24px;
    }

    .image-wrapper,
    img {
        &:first-child {
            margin-left: -2px;
        }

        &:not(:first-child) {
            margin-left: -12px;
        }
    }

    .image-wrapper img,
    img {
        border: 2px solid #fff;
        border-radius: 50%;
        height: 24px;
        width: 24px;
    }
}
```

I have just some basic CSS knowledge so I don't know if I am oversimplifying things 😅 so feel free to suggest something else.

thomashorta: I just noticed that though the CSS generated by the code above causes the image to have the correct size, they still do not overlap as expected. Removing `display: flex` from `.jetpack-blogging-prompt__answers` made them work properly on Android, but not on iOS 🙄 so I am guessing some WebView / browser differences between the systems.

I am thinking it might not be worth trying to fix the answers section for the mobile clients right now and maybe just hide it for them at the moment, wdyt? cc @develric @wargcm 

For hiding it on both clients without app-side changes I was thinking we could add something like this to the [reader-mobile.scss file](https://github.com/Automattic/wp-calypso/blob/f47c6faf675d8afc4d7eb8cbadb9768f2158bbc9/client/assets/stylesheets/reader-mobile.scss), does that make sense?
```scss
.reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__answers {
    display: none;
}

// we might need this as well so we don't have a big gap below the text
.reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__text {
	margin-bottom: 0px
}
```

wargcm: I'm expecting _most_ mobile clients will only see posts with that section when viewing the prompt responses already. So I don't see a problem with hiding it if it takes a ton of work trying to get it to appear correctly.

client/blocks/reader-full-post/_content.scss
davemart-in: Updated in: https://github.com/Automattic/wp-calypso/pull/76092/commits/bb9ec63fe359868bf49ab430af89401df5f5b0fd

client/blocks/reader-full-post/_content.scss
davemart-in: Added in https://github.com/Automattic/wp-calypso/pull/76092/commits/a4f4356e9c27235b9cb47b0a0e0fe021d51c600e

client/blocks/reader-full-post/_content.scss
davemart-in: > ```
> <div class="wp-block-jetpack-blogging-prompt jetpack-blogging-prompt">
>     <div class="jetpack-blogging-prompt__label">
>         Daily writing prompt
>     </div>
>     <div class="jetpack-blogging-prompt__text">
>         Describe a random encounter with a stranger that stuck out positively to you.
>     </div>
>     <div class="jetpack-blogging-prompt__answers">
>         
>         
>         
>         <a class="jetpack-blogging-prompt__answers-link" href="https://wordpress.com/tag/dailyprompt-1912" target="_blank" rel="external noreferrer noopener">View all responses</a>
>     </div>
> </div>
> ```

@thomashorta I made some updates in https://github.com/Automattic/wp-calypso/pull/76092/commits/93fb8d4a294df38bab7e670b66047fd029f744f6 based on the code you provided above.  I pasted your markup into my local env and here's what I'm seeing now (Notice `.image-wrapper` is missing  in the markup):

CleanShot 2023-04-24 at 16 32 38@2x

client/blocks/reader-full-post/_content.scss
davemart-in: I went ahead and just used a hex value here to save you the hassle. Updated in: https://github.com/Automattic/wp-calypso/pull/76092/commits/3b86368cd9cf3ab0f8ca6dc12358aaa51615fedf

client/blocks/reader-full-post/_content.scss
davemart-in: @thomashorta is there any CSS class to denote that you are in dark mode?

This particular icon is loaded from an SVG file, which means that you won't be able to alter the stroke or fill dynamically. The SVG would need to be included in the DOM for it to be editable (which we can't do here, since we're styling code that is coming from the editor).

What I have done however is include a dark mode variation located at 'client/assets/images/icons/light-bulb-dark.svg' in https://github.com/Automattic/wp-calypso/pull/76092/commits/18e937f9890f57f765d26cd029f236e9d29648ff. You could just swap this out with CSS with something like:

```
.dark-mode .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before {
	background: no-repeat url(calypso/assets/images/icons/light-bulb-dark.svg);
}
```

Assuming there is a `.dark-mode` class or something similar that you could use.

client/blocks/reader-full-post/_content.scss
thomashorta: I think it would be better to use `--color-neutral-5` in this case. Can you revert that change?

We already have other `--color-neutral` usages currently and we do change the actual hex color being used depending on light vs dark mode.

For instance, in this case, we would use a very light gray on the white background (light mode) and a darker gray on the dark background (dark mode), so the border is always subtle. Having the color hardcoded to `#dcdcde` makes the border too contrasting, here's an example of how it is looking on Android right now:


I was already working on implementing the `--color-neutral-5` today, and this is what it looks like:

client/blocks/reader-full-post/_content.scss
thomashorta: Not sure why, but when targeting the latest `Calypso Live` URL the icon is not showing up anymore on mobile (both iOS and Android).

Opening the reader on the web it shows up correctly, though I noticed that when I inspect the CSS of the web page, this is the URL being used for the light-bulb: `url(/calypso/images/light-bulb-5a92e3ed7ded3d7f7cae.svg)`

And on mobile I see `url(calypso/assets/images/icons/light-bulb.svg)`.

Looking at the proxy it also seems like it's trying to reach `container-recursing-raman.calypso.live/calypso/calypso/assets/images/icons/light-bulb.svg`, but that returns a `404`. 🤔 

Do you know what could be going on?

thomashorta: > is there any CSS class to denote that you are in dark mode?

Not really, on Android we actually compute the colors on the app-side (Kotlin and Android APIs) and provide the variables manually for the current App Theme setting. 

On iOS I see we are using `@media (prefers-color-scheme: light)` and `@media (prefers-color-scheme: dark)` to provide the colors, so _maybe_ that works on Android as well... ~~I could not test with the current icons because of the error [I mentioned above](https://github.com/Automattic/wp-calypso/pull/76092#discussion_r1175820457).~~

I was able to test it by setting different colors to the border and it worked. So I believe for the icon we could have something like:
```css
@media (prefers-color-scheme: light) {
    .reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before {
        background: no-repeat url(calypso/assets/images/icons/light-bulb.svg);
        background-size: 24px 24px;
        content: "";
        display: inline-block;
        height: 24px;
        left: -5px;
        opacity: .65;
        position: absolute;
        width: 24px;
        top: 1px
    }
}

@media (prefers-color-scheme: dark) {
    .reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before {
        background: no-repeat url(calypso/assets/images/icons/light-bulb-dark.svg);
        background-size: 24px 24px;
        content: "";
        display: inline-block;
        height: 24px;
        left: -5px;
        opacity: .65;
        position: absolute;
        width: 24px;
        top: 1px
    }
}
```

Once the [issue above](https://github.com/Automattic/wp-calypso/pull/76092#discussion_r1175820457) with the icon not loading is solved, that is.

client/blocks/reader-full-post/_content.scss
thomashorta: Thanks for the changes!

On Android this is looking nice!
But on iOS the avatars do not overlap (though I don't think that it's looking bad, to be honest).

Android | iOS
--- | ---
 |

client/assets/images/icons/light-bulb-dark.svg
```diff
@@ -0,0 +1 @@
+<svg height="18" viewBox="0 0 14 18" width="14" xmlns="http://www.w3.org/2000/svg"><path d="m8.75005 16.0938h-4.68755c-.08594 0-.15625.0703-.15625.1562v.625c0 .3457.2793.625.625.625h3.75c.3457 0 .625-.2793.625-.625v-.625c0-.0859-.0703-.1562-.1562-.1562zm-2.3438-16.0938c-3.53711 0-6.40625 2.86914-6.40625 6.40625 0 2.37105 1.28906 4.44145 3.20313 5.54885v2.2637c0 .3457.27929.625.625.625h5.15622c.3457 0 .625-.2793.625-.625v-2.2637c1.9141-1.1074 3.2031-3.1778 3.2031-5.54885 0-3.53711-2.8691-6.40625-6.4062-6.40625zm2.498 10.7383-.7011.4062v2.293h-3.5938v-2.293l-.70115-.4062c-1.53711-.8887-2.50195-2.5254-2.50195-4.33205 0-2.76172 2.23828-5 5-5 2.7617 0 5 2.23828 5 5 0 1.80665-.9648 3.44335-2.502 4.33205z" fill="#fff" transform="translate(.59375 .25)"/></svg>
```
creativecoder: There look to be some extraneous properties here, I'd recommend
- Removing `height` and `width`
- Setting viewBox to the standard `0 0 24 24`
- Removing `fill` and `transform`

creativecoder: Realizing you may need the fill for the `dark` aspect, unless it's possible to use the original icon just change the fill for dark mode via css?

client/blocks/reader-full-post/_content.scss
creativecoder: Quote styling looks good with the update.

client/blocks/reader-full-post/_content.scss
davemart-in: I reverted it in https://github.com/Automattic/wp-calypso/pull/76092/commits/e4c77c6d576840ed74946a4182a50560af69fa30.

client/blocks/reader-full-post/_content.scss
davemart-in: in https://github.com/Automattic/wp-calypso/pull/76092/commits/0f7116bef0c58a61e97d378c92d7ba6ef20eca2d I moved the icons inline (therefore eliminating the need to load them externally). I also added the code to update the icon when dark mode is selected. Have a look and let me know if this is still posing any issues.

client/assets/images/icons/light-bulb-dark.svg
davemart-in: I kept height and width (both set to 20) in there because I think it helps ensure the icon doesn't get squished on some browsers. I updated viewbox to be `0 0 20 20`. I removed transform, but kept fill in place.

> Realizing you may need the fill for the dark aspect, unless it's possible to use the original icon just change the fill for dark mode via css?
The SVG would need to be loaded directly into the DOM for us to be able to dynamically change the fill. Since we're adding the icon after the DOM has loaded (using ::before), this option is not available.

davemart-in:
@thomashorta please have another look when you have a moment. I believe we should be close!

client/blocks/reader-full-post/_content.scss
thomashorta: > I see we are using @media (prefers-color-scheme: light) and @media (prefers-color-scheme: dark) to provide the colors, so maybe that works on Android as well.

Although it works on mobile, I just realized that it also affects web, which is not great since people can be in dark mode (for instance I use MacOS dark mode) but the reader page on web always has a white background. I did a quick test editing the CSS for a page in Chrome and using the `prefers-color-scheme` and it applies the `dark` variation for me.

So I guess this can't be used on the general CSS, maybe we will have to do something on the mobile clients anyway.

https://user-images.githubusercontent.com/5091503/234290590-d80618dd-6359-4ecf-a6e1-802cf96a5db2.mov

davemart-in:
@thomashorta I saw a comment come through from you via email, but I don't see it in the thread above for some reason.  I removed the dark color scheme preference code in https://github.com/Automattic/wp-calypso/pull/76092/commits/8de739142e3be4821d8710a004224b0761de9de0

client/blocks/reader-full-post/_content.scss
```diff
@@ -442,3 +442,92 @@
...
+			left: -5px;
+			opacity: 0.65;
+			position: absolute;
+			width: 24px;
+			top: 1px;
+		}
+	}
+
+	@media (prefers-color-scheme: dark) {
+		.reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before {
```
thomashorta: I believe this `@media` query needs to be nested inside `.jetpack-blogging-prompt__label` and this selector could be just `&::before`.

Currently, the compiled `css` is resulting in this matcher, because of the nesting:
```css
.reader-full-post__story-content .jetpack-blogging-prompt .reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before
```

---

Fixing the above makes things look good on mobile! 🎉  

**BUT** doing so introduces a different problem that I just realized today. Since this CSS is also being used for web, this `media` query will also apply on web browsers, which is not desirable, since even on a Dark theme, the Reader web page has a white background. Here's an example, injecting the resulting CSS on Chrome and switching between MacOS' Themes:

https://user-images.githubusercontent.com/5091503/234299989-c4295a0a-1665-4ef7-8c16-7ad9f0538a7f.mov

thomashorta (commented):
Thanks for the changes, I think we are very close!!! I just added a comment about an issue regarding the usage of `prefers-color-scheme` that I just realized.

client/blocks/reader-full-post/_content.scss
thomashorta: > I saw a comment come through from you via email, but I don't see it in the thread above for some reason. I removed the dark color scheme preference code in https://github.com/Automattic/wp-calypso/commit/8de739142e3be4821d8710a004224b0761de9de0

Sorry @davemart-in, GitHub is trolling me, I wrote that comment, and after I posted the page refreshed with your changes for today, so I deleted it in order to test it and make sure that was an actual problem, but it is.

I am not sure what would be the best way of achieving the correct icon color on mobile without making hardcoded workarounds in both Android and iOS WebView "renderers". 😞 

Maybe we should inject a `.darkmode` class somewhere so we can have better control of these things from the `css` when needed, as [you commented here](https://github.com/Automattic/wp-calypso/pull/76092#discussion_r1175764460). 🤔

thomashorta:
> I saw a comment come through from you via email, but I don't see it in the thread above for some reason. I removed the dark color scheme preference code in https://github.com/Automattic/wp-calypso/commit/8de739142e3be4821d8710a004224b0761de9de0

Sorry @davemart-in, GitHub is trolling me, I wrote that comment, and after I posted the page refreshed with your changes for today, so I deleted it in order to test it and make sure that was an actual problem, but[ it is](https://github.com/Automattic/wp-calypso/pull/76092#discussion_r1176565152).

I am not sure what would be the best way of achieving the correct icon color on mobile without making hardcoded workarounds in both Android and iOS WebView "renderers". 😞 

Maybe we should inject a `.darkmode` class somewhere so we can have better control of these things from the `css` when needed, as [you commented here](https://github.com/Automattic/wp-calypso/pull/76092#discussion_r1175764460). 🤔

davemart-in:
I'm out of ideas as well. Happy to proceed however you think is best. We could:

A) Just ship what I have in this PR and let you deal with the dark mode issue in each of the native implementations.
B) Add a CSS class to distinguish a `.darkmode`, which would then need to be added to both native apps.

Which do you prefer @thomashorta?

thomashorta:
> I'm out of ideas as well. Happy to proceed however you think is best. We could:
>
> A) Just ship what I have in this PR and let you deal with the dark mode issue in each of the native implementations.
> B) Add a CSS class to distinguish a .darkmode, which would then need to be added to both native apps.
> 
> Which do you prefer @thomashorta?

I think we could ship with what you have in this PR for now, with the `--color-neutral-10`, since that is already supported by the apps so deploying this change would make the block work basically 100% correctly in all clients. I do have to make a few minor changes in the mobile code and also figure out the icon color for dark mode on mobile.

Btw, regarding that icon color on mobile, I was wondering if it would make sense to add the specific `media` query for dark mode in the [reader-mobile.scss](https://github.com/Automattic/wp-calypso/blob/51da610b30610bc9d6203f175762e4901f94cef3/client/assets/stylesheets/reader-mobile.scss) file, since rules there would (AFAIK) only affect the mobile apps. 

If that's possible and makes sense, adding the lines below to that file would make the block work 99% correctly on all clients, without any mobile-side changes. That missing 1% is a minor bug I found on Android that needs to be fixed in the app.

```scss
@media (prefers-color-scheme: dark) {
    .reader-full-post__story-content .jetpack-blogging-prompt .jetpack-blogging-prompt__label::before {
        // stylelint-disable-next-line function-url-quotes
        background-image: url("data:image/svg+xml,%3Csvg fill='none' height='20' viewBox='0 0 20 20' width='20' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m12.3438 17.3438h-4.68755c-.08594 0-.15625.0703-.15625.1562v.625c0 .3457.2793.625.625.625h3.75c.3457 0 .625-.2793.625-.625v-.625c0-.0859-.0703-.1562-.1562-.1562zm-2.3438-16.0938c-3.53711 0-6.40625 2.86914-6.40625 6.40625 0 2.37105 1.28906 4.44145 3.20313 5.54885v2.2637c0 .3457.27929.625.625.625h5.15622c.3457 0 .625-.2793.625-.625v-2.2637c1.9141-1.1074 3.2032-3.1778 3.2032-5.54885 0-3.53711-2.8692-6.40625-6.4063-6.40625zm2.498 10.7383-.7011.4062v2.293h-3.59377v-2.293l-.70118-.4062c-1.53711-.8887-2.50195-2.52541-2.50195-4.33205 0-2.76172 2.23828-5 5-5 2.7617 0 5 2.23828 5 5 0 1.80664-.9648 3.44335-2.502 4.33205z' fill='%23fff'/%3E%3C/svg%3E");
    }
}
```

What do you think?

davemart-in:
@thomashorta, yep! That is correct. Updated in https://github.com/Automattic/wp-calypso/pull/76092/commits/1806f5cb39ba44a6a5bf0598c846989da8189bba

thomashorta (approved):
> That is correct. Updated in https://github.com/Automattic/wp-calypso/commit/1806f5cb39ba44a6a5bf0598c846989da8189bba

Awesome!!! Now it seems to be working correctly on Android, iOS, and Web! 🎉 

Thanks a lot for the changes and sorry again for the back and forth! 🙇 

Android | iOS
--- | ---
 |

davemart-in:
There were two seemingly unrelated E2E test failures, so I just rebased the branch to see if that would dismiss them.

PR #76093, opened 2023-04-21
Use tag stats to display post and site count totals

eoigal:
There is a WPCOM REST API endpoint available to get the stats for a tag.

For a request like;
```
https://public-api.wordpress.com/rest/v1.3/read/topics/creativity/stats?_envelope=1
```

We should get a JSON response like;

```
{"total_posts":71395,"posts_per_day":195.6027397260274,"total_sites":37076,"last_post_date_gmt":"2023-04-20T23:50:35+0000"}
```

This PR will call this endpoint and display the `total_posts` and the `total_sites` tallies on the tag page sidebar similar to the mockup below;

screenshot-2023-03-16-at-3 26 14-pm


davemart-in (approved):
LGTM. No JS Errors.

I bumped up the space between the two stats by just a bit.

PR #26978, opened 2018-08-31
Add notice about plan credits to /plans page to reduce users confusion about pricing

adamziel:
Related to p9jf6J-So-p2

This PR adds some messaging on `/plans` page that better explains the nature of discounted prices.

An example:

zrzut ekranu 2018-08-31 o 15 31 51

Some specifics are still being discussed so no description or test plan for now.

PR #22384, opened 2018-02-13
Signup: Set up A/B test for minimizing the Free plan on the plans page.

taggon:
This PR sets up an A/B test for minimizing the Free plan on the plans plage in sign up. See p8Eqe3-nL-p2 for more detail about the test.

## How to test

1. Set `minimizeFreePlan` A/B test to `minimized` by running the follow script in browser console:
```
localStorage.setItem( 'ABTests', '{"minimizeFreePlan_20180219": "minimized"}' );
```
2. Create a new website. Follow the signup flow and you will notice the free plan card disappeared. Instead, an action button for the free plan shows up at the far bottom of the page as follows:
wordpress_com

cc @anneforbush @fditrapani

PR #76100, opened 2023-04-21
Refactor displaying plan notices

ddc22:
## Introduction
This PR aims to refactor the plan notices section. Notices were first introduced to  the codebase using React portals #26978. Even though it's initially convenient to breakout out of react and reuse a chunk of logic to show a UI that lives out of your natural child tree it can pose some problems. Also by breaking out essentially a linear conditional decision chain into multiple functions there seems to be repetition of logic. The solution proposed here is to move the notices into a declarative component inside the render loop.

## The Problems and motivation for this change

1. When we start to access the DOM directly, this anti pattern tends to grow. As shown below we have tried to check if the mounting dom element exists by querying the dom directly before mounting this notice.

https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L874-L877

https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L918-L931

2. Moving out of DOM means state changes will now have to be manually verified.
As shown below we imperatively implement calls based on state which can become more convoluted with time. 
https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L903-L908

3. Duplication of condition logic due to not being in a single if/else chain. Several flags were hidden behind multiple functions that could have been consolidated.

#### hasDiscountNotice
https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L1001
https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L934

#### canUserPurchasePlan
https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L968
https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L896

## Use cases impacted
The below use cases were copied from #72952. The UI designs maybe a bit outdated but the idea remains the same. These are the main cases to watch out for. However since this is a code refactor nothing else should have been broken and should mirror production without any side effects.

**Pro-rated credits**
This should show when you already have an annual paid plan:

Screenshot 2023-02-03 at 13 51 44

**Upgrade disabled**
This should show if you're not the owner of the current paid plan site:

Screenshot 2023-02-03 at 13 12 42

**Marketing**
You'll need to sandbox the public API and follow the instructions in `2efed-pb`  to trigger it:

Screenshot 2023-02-03 at 13 09 34

Related to #26978 #22384 #75725

## Pre-merge Checklist


client/my-sites/plan-features-comparison/index.jsx
```diff
@@ -313,22 +309,6 @@ PlanFeaturesComparison.defaultProps = {
...
-			const annualDiscountPrice = getPlanDiscountedRawPrice( state, siteId, planName );
-			const annualRawPrice = getSitePlanRawPrice( state, siteId, planName );
-			if ( typeof annualDiscountPrice !== 'number' || typeof annualRawPrice !== 'number' ) {
-				return 0;
-			}
-
-			return annualRawPrice - annualDiscountPrice;
-		} )
-		.reduce( ( max, credits ) => Math.max( max, credits ), 0 );
-
```
ddc22, line 331: This is essentially a calculation of state which [belongs in a selector](https://redux.js.org/usage/deriving-data-selectors#calculating-derived-data-with-selectors).  And has been moved to   [calculate-plan-upgrade-credits.ts](https://github.com/Automattic/wp-calypso/pull/76100/files#diff-9541c9db66f4bb4bf6587c41379862986af7ceed675f394f868503b0393b0a31)

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Calculate available plan credits given a set of displayed plans
+ *
+ * @param {*} state app state
+ * @param {*} siteId the selected site id
+ * @param {string[]}  consideredPlans Plans that are considered for the given calcuation
+ * @returns {number} The amount of credits available for the given plans
+ */
+export function calculatePlanUpgradeCredits(
```
chriskmnds: Let's rename this to the more conventional selector form e.g. `getPlanUpgradeCredits`.

Although, thinking whether a selector is needed here or a more local hook instead, as it's only used for the `PlanNotices` component.

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,38 @@
...
+		return 0;
+	}
+
+	const creditsPerPlan = consideredPlans.map( ( planName ) => {
+		const availableForPurchase = isPlanAvailableForPurchase( state, siteId, planName );
+		if ( ! availableForPurchase ) {
+			return 0;
+		}
+		const annualDiscountPrice = getPlanDiscountedRawPrice( state, siteId, planName );
+		const annualRawPrice = getSitePlanRawPrice( state, siteId, planName );
```
chriskmnds: Can these be served by the `getPlanPrices` selector? If so, then can we go for that to keep unifying access to these pricing selectors?

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param {*} state app state
+ * @param {*} siteId the selected site id
+ * @param {string[]}  consideredPlans Plans that are considered for the given calcuation
+ * @returns {number} The amount of credits available for the given plans
+ */
+export function calculatePlanUpgradeCredits(
+	state: IAppState,
+	siteId: number | undefined,
+	consideredPlans: string[]
```
chriskmnds: I'm not sure if the adjective "considered" adds any value here. It feels unnecessary. I think plain `plans` would just work.

client/my-sites/plans-features-main/plan-notices.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { getCurrentUserCurrencyCode } from 'calypso/state/currency-code/selectors';
+import isSiteAutomatedTransfer from 'calypso/state/selectors/is-site-automated-transfer';
+import {
+	getSitePlanSlug,
+	isCurrentUserCurrentPlanOwner,
+} from 'calypso/state/sites/plans/selectors';
+import { calculatePlanUpgradeCredits } from 'calypso/state/sites/plans/selectors/calculate-plan-upgrade-credits';
+import { isCurrentPlanPaid, isJetpackSite } from 'calypso/state/sites/selectors';
+
+function usePlanUpgradeCreditsDisplay(
```
chriskmnds: Can we move hooks to a corresponding `/hooks` folder? I can see your intention here, but I prefer to see hooks living separately, exported, and tested fully (so let's also add a test for this as well).

client/my-sites/plans-features-main/plan-notices.tsx
```diff
@@ -0,0 +1,136 @@
...
+	const isUpgradeEligibleSite =
+		isSiteOnPaidPlan && isNotJetpackSiteOrIsAtomicSite && isHigherPlanAvailable();
+
+	return {
+		creditsValue,
+		isPlanUpgradeCreditEligible: isUpgradeEligibleSite && creditsValue > 0,
+	};
+}
+
+export default function PlanNotices( {
```
chriskmnds: NIT. I think singular form (`PlanNotice`) might reflect better what's served by the component (a single notice). 🤔

client/my-sites/plans-features-main/plan-notices.tsx
```diff
@@ -0,0 +1,136 @@
...
+): {
+	creditsValue: number;
+	isPlanUpgradeCreditEligible: boolean;
+} {
+	const isSiteOnPaidPlan = !! useSelector( ( state ) => isCurrentPlanPaid( state, siteId ) );
+	const currentSitePlanSlug = useSelector( ( state ) => getSitePlanSlug( state, siteId ) );
+	const creditsValue = useSelector( ( state ) =>
+		calculatePlanUpgradeCredits( state, siteId, visiblePlanNames )
+	);
+	const isNotJetpackSiteOrIsAtomicSite = !! useSelector(
```
chriskmnds: This name feels more indirect and hard to grasp than previously. If I understand, is this rewriting `isJetpackNotAtomic` from above?

client/my-sites/plans-features-main/plan-notices.tsx
```diff
@@ -0,0 +1,136 @@
...
+							amountInCurrency: formatCurrency( creditsValue, currencyCode ?? '' ),
+						},
+						components: {
+							b: <strong />,
+						},
+					}
+				) }
+			</Notice>
+		);
+	}
```
chriskmnds: Are these conditions meant to be ordered in some priority list, or are they replaceable? e.g. can we switch the last one with second-to-last and still have this working as expected?

client/my-sites/plans-features-main/plan-notices.tsx
```diff
@@ -0,0 +1,136 @@
...
+	);
+
+	if ( isNoticeDismissed ) {
+		return null;
+	} else if ( ! canUserPurchasePlan ) {
+		return (
+			<Notice
+				className="plan-features__notice"
+				showDismiss={ true }
+				onDismissClick={ () => setIsNoticeDismissed( true ) }
```
chriskmnds: Can we move this into a handler function `handleDismissNotice` and reuse it across the rest? Should we also wrap with `useCallback`?

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,38 @@
...
+	const creditsPerPlan = consideredPlans.map( ( planName ) => {
+		const availableForPurchase = isPlanAvailableForPurchase( state, siteId, planName );
+		if ( ! availableForPurchase ) {
+			return 0;
+		}
+		const annualDiscountPrice = getPlanDiscountedRawPrice( state, siteId, planName );
+		const annualRawPrice = getSitePlanRawPrice( state, siteId, planName );
+		if ( typeof annualDiscountPrice !== 'number' || typeof annualRawPrice !== 'number' ) {
+			return 0;
+		}
```
chriskmnds: I think this will need updating once refactored to `getPlanPrices`.

client/my-sites/plans-features-main/plan-notices.tsx
chriskmnds: Can we add some tests for this component? It would be nice to capture any priority rendering in the test if the current shape remains.

client/my-sites/plans-features-main/plan-notices.tsx
chriskmnds, line 1: Let's start moving components into a respective `/components` folder. It will help avoid a messy directory and start isolating parts better.

chriskmnds (changes requested):
Thanks, @jdc91 . Nice bit of refactoring! 💯 
I left some comments to hopefully clean up / improve things a little. Nothing critical, but will come back to test thoroughly once/if any feedback is accounted.

client/my-sites/plans-features-main/plan-notices.tsx
aneeshd16: Just chiming in here, I have already moved components to a /components folder in https://github.com/Automattic/wp-calypso/pull/75526, so +1 to @chriskmnds' suggestion.

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: It's actually a hard choice here. between `!isJetPackNotAtomic` (i.e. Not not double negation) or `isNotJetpackSiteOrIsAtomicSite`. I thought this to be much more readable than the other. It is a judgment call though I can change it if the other way is more readable 👍

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: I think they are in order. At least that is what I gathered from the previous implementation.

https://github.com/Automattic/wp-calypso/blob/af1a71803031f7d3e139201e3c3e50850625c81c/client/my-sites/plan-features-2023-grid/index.tsx#L903-L916

We are trying to populate 1 notice in a single location. So there is bound to be an order of precedence if not those conditions do not make sense. This is the order that I have explicitly signalled with the if else chain :)

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: Yes I have been meaning to do this but got bogged down with the refactor itself ! Will try to add some tests!

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
ddc22: I was thinking since it is a literal calculation of state to have it closer to redux state so that any future uses of this can reuse this, but looking back there is a possibility for code YAGNI. But the code has to live somewhere so this is more a matter of cohesion I guess. 

The other reason I didn't really move to a hook is because hook semantics like useState or useEffect are not required here and the only use would be to tap into state.

Given the direction of moving away from redux maybe its a good idea to move away from the state and probably rely on a hook and props. Let me know what you prefer, I am ok either way. 👍 

For now I will rename this to adhere with the naming convention.

chriskmnds: > Given the direction of moving away from redux maybe its a good idea to move away from the state and probably rely on a hook and props. Let me know what you prefer, I am ok either way. 👍

Technically speaking, this component (and selector) - the stuff in the entire PR - will probably not be subject to export to a package (like we want to do with the main grid component). So it seems ok to be utilizing Calypso state selectors here. The refactor you did here is pretty powerful in that regard, as it pulls out a big chunk of code that we don't need to "serve" the 2023 grid from a package 👍 

My suggestion for a hook is to avoid adding a selector that's only used for a specific use case (I think hooks can serve that purpose)

client/my-sites/plans-features-main/plan-notices.tsx
chriskmnds: `isJetPackNotAtomic` seems to be used quite a bit at places, so I'd keep that. it's also shorter and feels more "canonical" 🤔

client/my-sites/plans-features-main/plan-notices.tsx
chriskmnds: Thanks! I think the order was a little more clear/obvious previously (but might just be me having looked at the previous version). It was still "implied" and prone to error (someone adding another notice, not realizing the prioritization I guess).

I don't suppose there's a better (any?) design pattern we can apply for such cases? Otherwise, if we can guard the order via a couple of tests it would be great too I think.

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
chriskmnds: p.s. just FYI - the `getPlanPrices` selector will be on our immediate list of things to port into the plans data store (when we get to). We'll need to decide how to migrate existing uses (this included). It might be the case that we prefix with `deprecated__` and direct uses to go through the data store instead. but it's not an immediate concern for here I think (unless we port sooner than later)

ddc22: Are you talking about this selector? (I couldn't find any other selector)
I think that selector might need to be renamed to `getSitePlanPrices` because it gets the price information for the current site plan, not any plan that we have. 
 
https://github.com/Automattic/wp-calypso/blob/f8f12f8d5aedb387ef12f466fde5b616d20c251b/client/state/plans/selectors/get-plan-prices.ts#L14-L26

I am trying to retrieve the plan object for any given plan slug.

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: I don't think we need to use useCallback here. The child <Notice />  is a fairly simple component and there is nothing much to optimise there and we do not have a reason to expect it growing in complexity since it's simply a small notice.

Also since `setIsNoticeDismissed` is simply **setting state** do we need to add an indirection by adding it into another handler function? 

I think having it this way is more readable at a glance. WDYT?

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: I think having this as an explicit if else chain will be best. We can possibly move the contents of each if else chain into descriptive child components though. But the cyclomatic complexity will still remain which is I think the main source of complexity here.

I will add some tests here as well so that we are guarded from possibly regressions.

client/my-sites/plans-features-main/plan-notices.tsx
ddc22: I restructured all the component's inside plans-features-main/components @aneeshd16  since it is in a different tree +1

client/my-sites/plans-features-main/plan-notices.tsx
chriskmnds: > I don't think we need to use useCallback here. The child is a fairly simple component and there is nothing much to optimise there and we do not have a reason to expect it growing in complexity since it's simply a small notice.

👍 

> Also since setIsNoticeDismissed is simply setting state do we need to add an indirection by adding it into another handler function?

I thought about this too, but I'd still prefer a named function to reuse. The same definition is repeated three times if anything (currently three, could even grow).

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
chriskmnds: > I think that selector might need to be renamed to getSitePlanPrices because it gets the price information for the current site plan, not any plan that we have.

I don't think that's the case? It's a wrapper to the three/four main pricing selectors used for for plans (any plan). It will indeed compute a discounted price if it's also the site plan (when `siteId` is passed), but I think you can trust the raw price to be the correct price needed at places (as far as I recall, and unless something hasn't changed already).

ddc22: Sorry, I just read the comment on the function call and assumed that is what it did. 

Thanks [I tried to incorporate this](https://github.com/Automattic/wp-calypso/pull/76100/commits/9add02615948db9134f997671b443d97ed4a491c).

However for some reasons the values that I get do not tally with production credits values for some reasons. I did some investigation and could not figure it out right now ( Down the call stack we seem to be calling 2 different selectors [getPlanDiscountedRawPrice](https://github.com/Automattic/wp-calypso/blob/261706c3c01db16bb91c619728d2a69ec36953c8/client/state/sites/plans/selectors/get-plan-discounted-raw-price.ts), [getDiscountedRawPrice](https://github.com/Automattic/wp-calypso/blob/261706c3c01db16bb91c619728d2a69ec36953c8/client/state/plans/selectors/get-discounted-raw-price.ts) when comparing the 2 approaches)

So I thought of leaving this out for now to be investigated later.

chriskmnds: Interesting. I think it's best if we try and go through this selector as it returns values from both [getPlanDiscountedRawPrice](https://github.com/Automattic/wp-calypso/blob/261706c3c01db16bb91c619728d2a69ec36953c8/client/state/sites/plans/selectors/get-plan-discounted-raw-price.ts), [getDiscountedRawPrice](https://github.com/Automattic/wp-calypso/blob/261706c3c01db16bb91c619728d2a69ec36953c8/client/state/plans/selectors/get-discounted-raw-price.ts) to use in comparison. That was the idea pretty much, to unify the pricing.

Maybe @aneeshd16 can chime in? I believe I've seen some recent work on an update to one of the underlying selectors.

client/my-sites/plans-features-main/components/plan-notices.tsx
```diff
@@ -0,0 +1,101 @@
...
+							amountInCurrency: formatCurrency( creditsValue, currencyCode ?? '' ),
+						},
+						components: {
+							b: <strong />,
+						},
+					}
+				) }
+			</Notice>
+		);
+	}
```
aneeshd16: We probably don't need the `else` here since all of these are return statements. So this could be something like the following for a small readability win.

```js
if ( isNoticeDismissed ) {
  return null;
}

if ( ! canUserPurchasePlan ) {
  return <Notice...
}
```

client/my-sites/plans-features-main/index.jsx
```diff
@@ -607,7 +608,15 @@ export class PlansFeaturesMain extends Component {
...
-				<div className="plans-features-main__notice" />
+				<PlanNotice
+					visiblePlanNames={ visiblePlans }
+					siteId={ siteId }
+					isInSignup={ this.props.isInSignup }
+					discountInformation={ {
+						withDiscount: this.props.withDiscount,
+						discountEndDate: this.props.discountEndDate,
+					} }
+				/>
```
aneeshd16: Since the notice needs to be displayed only on `/plans`, and it never is displayed in signup flows, would it be possible to render the `<PlanNotice/>` component one level above, maybe here?
https://github.com/Automattic/wp-calypso/blob/1d90b031e834b9664a7e8b45d762ee421af3930d/client/my-sites/plans/main.jsx#L267

I realize that the `visiblePlans` prop won't be available there, but perhaps we can get it via some selector? This would truly separate the notice from the plans grid. 🤔

client/state/marketplace/purchase-flow/selectors.ts
```diff
@@ -1,7 +1,16 @@
...
 import { IAppState } from 'calypso/state/types';
 
 export function getPurchaseFlowState( state: IAppState ): IPurchaseFlowState {
-	return state.marketplace.purchaseFlow;
+	return (
+		state?.marketplace?.purchaseFlow ?? {
+			primaryDomain: null,
+			productSlugInstalled: null,
+			pluginInstallationStatus: MARKETPLACE_ASYNC_PROCESS_STATUS.UNKNOWN,
+		}
```
aneeshd16: `state.marketplace.purchaseFlow` would never be `undefined` and would always return the default state from the reducer, right? Maybe I am missing something here.

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
aneeshd16: Hi @jdc91, the commit you shared cannot be accessed. Can you please share the commit in a different branch so that we help debug?

client/my-sites/plans-features-main/index.jsx
ddc22: That is a really good suggestion Aneesh! 👍 
However, the logic for visible plans is inside plansFeaturesMain and might be coupled further. I can look into moving this a step up but I think its better to move it to a different PR to avoid scope creep.

client/my-sites/plans-features-main/components/plan-notices.tsx
ddc22: I actually wanted to explicitly highlight the fact that this is a linear chain of conditions. Do you think it's more beneficial to **not** have the `else` 🤔 ?

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
ddc22: I have moved the commit branched off from this PR at https://github.com/Automattic/wp-calypso/pull/76361

client/state/marketplace/purchase-flow/selectors.ts
ddc22: Technically you are correct but we are following a convention where we safeguard against the reducer being undefined since they are/maybe registered dynamically.

https://github.com/Automattic/wp-calypso/blob/6e6636e9633456a68032d28fafd6d734e08db379/client/state/types.ts#L15-L18
so we have to guard for undefined.

chriskmnds:
Thanks, @jdc91! The refactors and follow ups here are amazing 🙌  

Happy to review and get this moving along tomorrow, before your start / if nobody catches me to it :-)

p.s. is there another/previous related PR, or is this with all the fixes? 🤔

ddc22:
Thanks for the great feedback @chriskmnds and @aneeshd16 ! 🙇 
I have also added the tests for the [PlanNotice](https://github.com/Automattic/wp-calypso/pull/76100/commits/dd41fbc8d24a97ef894058c7c594d8fc513e68c6) component. 
After we ship this change, on top of this I will next be finalising the change for  [Plans 2023 : Plan purchase credits messaging tweak](https://github.com/Automattic/wp-calypso/pull/75725)

client/state/marketplace/purchase-flow/selectors.ts
aneeshd16: Oh okay! Maybe we can import the default state from the reducer and return that instead?
https://github.com/Automattic/wp-calypso/blob/11d41f2c0be1329084b7093e3780c2f6ad71ac76/client/state/marketplace/purchase-flow/reducer.ts#L12-L16

client/my-sites/plans-features-main/components/plan-notices.tsx
aneeshd16: Since we return early from each statement, it would still be a linear chain. TBH, there's not really much of a difference here, and I highlighted it here just in case you had not considered this approach. Either way works! 🙂

client/state/sites/plans/selectors/calculate-plan-upgrade-credits.ts
aneeshd16: Thanks! I'll have a look at this.

client/my-sites/plans-features-main/components/plan-notice.tsx
```diff
@@ -0,0 +1,101 @@
...
+	);
+	const activeDiscount = getDiscountByName( withDiscount, discountEndDate );
+	const { creditsValue, isPlanUpgradeCreditEligible } = usePlanUpgradeCreditsDisplay(
+		siteId,
+		visiblePlanNames
+	);
+
+	const handleDismissNotice = () => setIsNoticeDismissed( true );
+
+	if ( isNoticeDismissed ) {
```
chriskmnds, line 43: This also goes back to my comment. There's some indirection here i.e. `Notice` instances below interact with this value, but `MarketingMessage` doesn't, yet closing a notice will also block the marketing message from showing.

client/my-sites/plans-features-main/components/plan-notice.tsx
```diff
@@ -0,0 +1,101 @@
...
+						},
+						components: {
+							b: <strong />,
+						},
+					}
+				) }
+			</Notice>
+		);
+	}
+	return <MarketingMessage siteId={ siteId } />;
```
chriskmnds, line 100: Back to my comment. I may be doing something wrong, but `MarketingMessage` takes precedence for me over the notices. This is expected?

client/my-sites/plans-features-main/components/test/plan-notice.tsx
```diff
@@ -0,0 +1,171 @@
...
+				isInSignup={ false }
+				siteId={ 32234 }
+			/>
+		);
+		expect( screen.getByRole( 'status' ).textContent ).toBe(
+			'You have $100.00 of pro-rated credits available from your current plan. Apply those credits towards an upgrade before they expire!'
+		);
+	} );
+
+	test( 'A marketing message <PlanNotice /> when no other notices are available and marketing messages are available', () => {
```
chriskmnds, line 149: Obviously, we expect `MarketingMessage` to _not_ take precedence judging from this test (yet it always does for me in the app) 🤔

client/my-sites/plans-features-main/hooks/test/use-plan-upgrade-credits-display.tsx
```diff
@@ -0,0 +1,104 @@
...
+const plansList = [
+	TYPE_FREE,
+	TYPE_PERSONAL,
+	TYPE_PREMIUM,
+	TYPE_BUSINESS,
+	TYPE_ECOMMERCE,
+	PLAN_ENTERPRISE_GRID_WPCOM,
+];
+describe( 'usePlanUpgradeCreditsDisplay hook', () => {
+	beforeEach( () => {
```
chriskmnds, line 53: I think it's good practice to `clearAllMocks` before anything else here

client/my-sites/plans-features-main/hooks/use-plan-upgrade-credits-display.ts
```diff
@@ -0,0 +1,36 @@
...
+): {
+	creditsValue: number;
+	isPlanUpgradeCreditEligible: boolean;
+} {
+	const isSiteOnPaidPlan = !! useSelector( ( state ) => isCurrentPlanPaid( state, siteId ) );
+	const currentSitePlanSlug = useSelector( ( state ) => getSitePlanSlug( state, siteId ) );
+	const creditsValue = usePlanUpgradeCredits( siteId, visiblePlanNames );
+	const isJetpackNotAtomic = useSelector(
+		( state ) => isJetpackSite( state, siteId ) && ! isSiteAutomatedTransfer( state, siteId )
+	);
```
chriskmnds, line 20: NIT - This should really evolve out of components at some point, considering there are 7 or more definitions of the same in the codebase. Nothing to do here.

client/my-sites/plans-features-main/hooks/use-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,40 @@
+import { useSelector } from 'react-redux';
+import { getPlanDiscountedRawPrice } from 'calypso/state/sites/plans/selectors';
+import { getSitePlanRawPrice } from 'calypso/state/sites/plans/selectors/get-site-plan-raw-price';
+import isPlanAvailableForPurchase from 'calypso/state/sites/plans/selectors/is-plan-available-for-purchase';
+import type { PlanSlug } from '@automattic/calypso-products';
+
+/**
+ * Calculate available plan credits given a set of displayed plans
```
chriskmnds, line 8: NIT - I think it might help if we are a little more specific here on how these credits are derived. Initially, I thought this will do some sort of summation, but it picks the max credits resulting from any plan.

client/my-sites/plans-features-main/hooks/use-plan-upgrade-credits.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param {*} siteId the selected site id
+ * @param {string[]}  plans Plans that are considered for the given calculation
+ * @returns {number} The amount of credits available for the given plans
```
chriskmnds: Similarly - I think `The amount of credits available for the given plans` points me to a "summation". Anyway, minor.

client/my-sites/plans-features-main/index.jsx
```diff
@@ -607,7 +608,15 @@ export class PlansFeaturesMain extends Component {
 				<QuerySites siteId={ siteId } />
 				<QuerySitePlans siteId={ siteId } />
 				<HappychatConnection />
-				<div className="plans-features-main__notice" />
+				<PlanNotice
+					visiblePlanNames={ visiblePlans }
```
chriskmnds: Can we update it to just `visiblePlans`? 

I wish we could just settle on "plan" and stop referring to things with "plan name" or "plan type" when both refer to product store slugs.

chriskmnds (approved):
Thanks @jdc91 for your patience and for addressing the feedback. 🙌  

This tests out for me. I don't think I tested all the scenarios, but I tested "pro-rated credits", "not owner", and the "marketing message". This works like in `trunk`, so we should be able to proceed. (but pls test again and resolve any existing convos first) :-)

A couple of things I notice, unrelated to this PR (so also on trunk): 

- Marketing message always takes precedence for me e.g. on a site with pro-rated credits and the marketing message enabled, it shows the marketing message. Is this expected, or did I mess things up on my end?
- When I close the marketing message, it seems to go inactive for some period of time (is this correct?). The problem is, none of the other notices will take precedence at that point, so nothing is shown on reloads, even though notices exist.

I feel this entire setup may need some rethinking. The "marketing message" isn't really a notice right? i.e. I don't really see why it should block other notices from showing or the other way. Maybe something for a follow-up. 🤔

client/test-helpers/testing-library/index.js
```diff
@@ -31,3 +32,30 @@ export const renderWithProvider = (
 
 	return rtlRender( ui, { wrapper: Wrapper, ...renderOptions } );
 };
+
+export const renderHookWithProvider = (
```
chriskmnds, line 36: Interesting. 
I wish these helpers could be extracted and put into a finer package for reuse.

chriskmnds:
Also, I wanted to note, the marketing message can feel a little clunky, especially on slower connections or other delays (and not sure about that dark background):

https://user-images.githubusercontent.com/1705499/235131278-7b2d42b1-b644-43e6-b56f-8d7864df75a4.mov

Unrelated to the PR, obviously. :-)

client/state/marketplace/purchase-flow/selectors.ts
ddc22: Ah yes good call, for some reason this looked like a type in my head 😄

client/test-helpers/testing-library/index.js
ddc22: Definitely a good candidate to package out 🤔 .

client/my-sites/plans-features-main/index.jsx
ddc22: I wanted to be explicit here because, generally plan refers to the plan object. But I guess that can be communicated by the type.

client/my-sites/plans-features-main/components/plan-notice.tsx
```diff
@@ -0,0 +1,101 @@
...
+			>
+				{ translate(
+					'This plan was purchased by a different WordPress.com account. To manage this plan, log in to that account or contact the account owner.'
+				) }
+			</Notice>
+		);
+	} else if ( activeDiscount ) {
+		return (
+			<Notice
+				className="plan-features__notice-credits"
```
chriskmnds, line 61: @jdc91 just FYI - the classes passed here don't have significance it seems e.g. there's no styling depended on them. 

We will still want to pass something for updating some of the inner styles/margins added. This is something I'm working out on my end (for the reskin https://github.com/Automattic/wp-calypso/pull/75333). I am currently passing `plan-features-main__notice` for all of them.

ddc22: I thought this is a change for the `<Notice />`  component. Are we not changing the styling for the notice system wide?

client/my-sites/plans-features-main/components/test/plan-notice.tsx
ddc22: I actually checked this and the UI seem to honour the precedence 🤔 
notices

client/test-helpers/testing-library/index.js
ddc22: Do you think it's ok to picked this up separately?

client/my-sites/plans-features-main/components/test/plan-notice.tsx
chriskmnds: Cool. Thanks for confirming! was probably doing something wrong on my end then. Sorry 🤷

client/my-sites/plans-features-main/components/plan-notice.tsx
chriskmnds: Yes, we are (changing it system wide). It's just for a case of a bottom margin that I chose to only remove for these notices. sounds like something I may need investigate better though. I guess a Notice shouldn't really include a margin by default to begin with 🤔

ddc22:
> I feel this entire setup may need some rethinking. The "marketing message" isn't really a notice right? i.e. I don't really see why it should block other notices from showing or the other way. Maybe something for a follow-up.

I think the naming might also be clashing here. We essentially are trying to display something in **screen real estate**. Maybe it needs to be changed to signal this with something similar to `<PlansGridHeaderNotices />`
image

Not sure how what the UX limitations are to have multiple message here. In any case I would venture a guest that we need to limit it, because importance needs to be given to carrying the users eyes towards the plans grid. Maybe @lucasmendes-design could weigh in here too.

I have added a gif below of how this screen real estate is populated.

notices

client/test-helpers/testing-library/index.js
chriskmnds: Or not at all right now. Just a thought - didn't really expect a follow up 🙇 

This PR looks like in great shape AFAIC. Thanks for all the tests and follow ups.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7963342

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7963342).

PR #50187, opened 2021-02-17
Disable HotModuleReplacement

scinos:
#### Changes proposed in this Pull Request

Disable `HotModuleReplacement`

Having this enabled without a proper HMR setup is causing lots of troubles, mainly developers getting an error when they change a file or switch branches while `yarn start` is running:

image

Now, in order to see a change in the browser, the developer must do a page refresh.

Note that this will also the disable the notification about the state of the compilation:

image

I think we should at least try to get HMR working, and I plan to work on it soon. At least we should bring back the notification because it is working today and providing value. But for now, it looks like the friction caused by the error overweights the value of the notification.

#### Testing instructions

Run `yarn start`, change a file in Calypso (or switch branches) and verify it doesn't break anymore.

PR #76101, opened 2023-04-21
Add hot reloading

mreishus:
## Proposed Changes

* Try to add hot-reloading
  *  Noting that Sergio removed the `HotModuleReplacementPlugin()` in #50187 because the hot updates were failing - and he mentioned we should try to get it working later.
  * I tried various things and was eventually able to get something working with `pmmmwh/react-refresh-webpack-plugin`.

* PROS:
  *  It seems to work :)
* CONS:
  * I don't really know what I'm doing and just fiddled with stuff until it worked.
  * I really know very little about calypso's complicated setup.
  * Not tested very much.
  * Would not be surprised if redux, wordpress/data etc break it.
  * Uses pre 1.0 plugin https://github.com/pmmmwh/react-refresh-webpack-plugin to take advantage of react's new "fast refresh"
  * Had to `yarn add -D` packages (but hopefully they don't affect production build?)

## Testing Instructions

* `yarn && yarn start`
* Use calypso
* Make changes in your editor and see if they hot reload

https://user-images.githubusercontent.com/937354/233733014-568c6c4a-2c9c-4cc3-be8a-bad3ce1d612e.mp4

## Pre-merge Checklist


noahtallen:
> Had to yarn add -D packages (but hopefully they don't affect production build?)

That makes sense to me, and we can see in the matticbot comment that we aren't changing the production bundle

I also noticed in the docs that it wants us to update the babel config. Was that not needed?

> I really know very little about calypso's complicated setup.

What you did looks good to me! 😁 Calypso's setup is "easier than it looks" but feels very complicated when you're working with it, unfortunately :(

mreishus:
:thinking: I didn't even try adding the babel plugin yet, I was so excited to get something that worked I skipped over it :)

noahtallen:
I wonder what it even enables since this already works :P

noahtallen:
I also tested with a typescript component in packages (like button). Making small changes to it and its style file successfully update the page too!

mreishus:
I tried testing with and without the babel config. After adding the Babel config, I saw a small difference in the generated files:
```diff
41a42
> _c = CardHeading;
49a51,54
> 
> var _c;
> 
> __webpack_require__.$Refresh$.register(_c, "CardHeading");
```
I'm not sure what it actually changes but it might be a good thing? I'll push the patch I have:

mreishus:
Ah, found something that breaks:

* Load homepage
* Make this change
```diff
diff --git a/client/state/ui/reducer.js b/client/state/ui/reducer.js
index 3938cb9305..78947857b5 100644
--- a/client/state/ui/reducer.js
+++ b/client/state/ui/reducer.js
@@ -29,6 +29,7 @@ export const selectedSiteId = withSchemaValidation(
        ( state = null, action ) => {
                switch ( action.type ) {
                        case SELECTED_SITE_SET:
+                               console.log('SELECTED_SITE_SET', action.siteId);
                                return action.siteId || null;
                }
 
```
* See in browser: [webpack] hot update failed: Error: Different reducers on multiple calls to `addReducerToStore` for key: ui

mreishus:
That error is ours; I stopped it from happening during the hot reload process in 92a93cbb3174edca09ca8d777c2aeb6a1f9f15ed , but the old reducer is still in memory even after the hot reload process. I tried some things but I couldn't get hot reducer replacement to work. Maybe that's good enough and people just need to F5 after messing with reducers.

I see @sgomes originally worked on `addReducerToStore()`, so maybe he has some ideas :)

noahtallen:
> Maybe that's good enough and people just need to F5 after messing with reducers.

Yeah, definitely not opposed to that. The PR would be an improvement no matter what. It'd be cool if we could make it obvious when someone needs to hard reload (or even do it automatically), but that could even be for follow-up.

@jsnajdr might also have ideas about the reducer stuff

mreishus:
Reverted 92a93cb since it caused TS failures and caused the ETK to be changed

sgomes:
> I see @sgomes originally worked on `addReducerToStore()`, so maybe he has some ideas :)

Sorry, I don't have much to share on this one 😕 My understanding of the mechanics of HMR in Calypso at the time was superficial at best, and it's only gotten worse 😅

I can answer questions about the work I did on `addReducerToStore()`, though, if that helps!

client/webpack.config.js
```diff
@@ -176,14 +187,16 @@ const filePaths = {
...
-	entry: filterEntrypoints( {
-		'entry-main': [ path.join( __dirname, 'boot', 'app' ) ],
-		'entry-domains-landing': [ path.join( __dirname, 'landing', 'domains' ) ],
-		'entry-login': [ path.join( __dirname, 'landing', 'login' ) ],
-		'entry-stepper': [ path.join( __dirname, 'landing', 'stepper' ) ],
-		'entry-browsehappy': [ path.join( __dirname, 'landing', 'browsehappy' ) ],
-		'entry-subscriptions': [ path.join( __dirname, 'landing', 'subscriptions' ) ],
-	} ),
+	entry: addHotMiddlewareIfDev(
+		filterEntrypoints( {
```
tyxla: `addHotMiddlewareIfDev` could be part of `filterEntrypoints()` since it's already looping through all the enabled ones anyway.

client/webpack.config.js
```diff
@@ -113,6 +114,16 @@ function filterEntrypoints( entrypoints ) {
 	return allowed;
 }
 
+function addHotMiddlewareIfDev( entry ) {
+	if ( isDevelopment ) {
+		return Object.entries( entry ).reduce( ( acc, [ key, value ] ) => {
+			acc[ key ] = [ 'webpack-hot-middleware/client', ...value ];
+			return acc;
+		}, {} );
```
tyxla: Alternatively, we could wrap in a `Object.fromEntries()` which will do the job to reduce back to an object, and we can only do a `.map()` to change the value. Could be done as part of `filterEndpoints()`.

client/webpack.config.js
```diff
@@ -113,6 +114,16 @@ function filterEntrypoints( entrypoints ) {
...
+	}
+	return entry;
```
tyxla: Nit, but I generally recommend the other way around:

```suggestion
	if ( ! isDevelopment ) {
		return entry;
	}

	return Object.entries( entry ).reduce( ( acc, [ key, value ] ) => {
		acc[ key ] = [ 'webpack-hot-middleware/client', ...value ];
		return acc;
	}, {} );
```

That way the actual logic is one less tab nested, and that can arguably improve the readability for some folks.

tyxla (commented):
Thank you for working on this, it's amazing to see an opportunity to fix hot reloading! ❤️ 

The only thing I spotted initially was the webpack build monitor needs some love since it will remain indefinitely in the "Building" state after a change:

Screenshot 2023-04-25 at 14 59 09

Regardless, I've done some testing with changes in section chunks and async-loaded modules and both worked great! This is definitely exciting to see 😍

mreishus:
Rebased + Updated

client/webpack.config.js
mreishus: I first tried combining the loops like this:

```diff
diff --git a/client/webpack.config.js b/client/webpack.config.js                                                                        
index a9a964d377..7f814371e8 100644                                                                                                     
--- a/client/webpack.config.js                                                                                                          
+++ b/client/webpack.config.js                                                                                                          
@@ -107,21 +107,18 @@ function filterEntrypoints( entrypoints ) {                                                                       
        const allowed = {};
        Object.entries( entrypoints ).forEach( ( [ key, val ] ) => {
                if ( validEntrypoints.includes( key ) ) {
-                       allowed[ key ] = val;                                                                                           
+                       allowed[ key ] = addHotMiddlewareIfDev( val );                                                                  
                }
        } );

        return allowed;
 }

-function addHotMiddlewareIfDev( entry ) {                                                                                              
-       if ( isDevelopment ) {                                                                                                          
-               return Object.entries( entry ).reduce( ( acc, [ key, value ] ) => {                                                     
-                       acc[ key ] = [ 'webpack-hot-middleware/client', ...value ];                                                     
-                       return acc;                                                                                                     
-               }, {} );                                                                                                                
+function addHotMiddlewareIfDev( value ) {                                                                                              
+       if ( ! isDevelopment ) {                                                                                                        
+               return value;                                                                                                           
        }
-       return entry;                                                                                                                   
+       return [ 'webpack-hot-middleware/client', ...value ];                                                                           
 }

 /**
@@ -187,16 +184,14 @@ const filePaths = {                                                                                               
 const webpackConfig = {
        bail: ! isDevelopment,
        context: __dirname,
-       entry: addHotMiddlewareIfDev(                                                                                                   
-               filterEntrypoints( {                                                                                                    
-                       'entry-main': [ path.join( __dirname, 'boot', 'app' ) ],                                                        
-                       'entry-domains-landing': [ path.join( __dirname, 'landing', 'domains' ) ],                                      
-                       'entry-login': [ path.join( __dirname, 'landing', 'login' ) ],                                                  
-                       'entry-stepper': [ path.join( __dirname, 'landing', 'stepper' ) ],                                              
-                       'entry-browsehappy': [ path.join( __dirname, 'landing', 'browsehappy' ) ],                                      
-                       'entry-subscriptions': [ path.join( __dirname, 'landing', 'subscriptions' ) ],                                  
-               } )                                                                                                                     
-       ),                                                                                                                              
+       entry: filterEntrypoints( {                                                                                                     
+               'entry-main': [ path.join( __dirname, 'boot', 'app' ) ],                                                                
+               'entry-domains-landing': [ path.join( __dirname, 'landing', 'domains' ) ],                                              
+               'entry-login': [ path.join( __dirname, 'landing', 'login' ) ],                                                          
+               'entry-stepper': [ path.join( __dirname, 'landing', 'stepper' ) ],                                                      
+               'entry-browsehappy': [ path.join( __dirname, 'landing', 'browsehappy' ) ],                                              
+               'entry-subscriptions': [ path.join( __dirname, 'landing', 'subscriptions' ) ],                                          
+       } ),                                                                                                                            
        mode: isDevelopment ? 'development' : 'production',
        devtool: sourceMapType,
        output: {
```

Which works when there are entry points set.
However, I discovered the filterEntrypoints also has an early exit case in which it will need the loop anyway:

```diff
diff --git a/client/webpack.config.js b/client/webpack.config.js
index 7f814371e8..7550883665 100644
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -83,7 +83,12 @@ if ( shouldCreateSentryRelease ) {
 
 function filterEntrypoints( entrypoints ) {
        if ( ! process.env.ENTRY_LIMIT ) {
-               return entrypoints;
+               return Object.fromEntries(
+                       Object.entries( entrypoints ).map( ( [ key, value ] ) => [
+                               key,
+                               addHotMiddlewareIfDev( value ),
+                       ] )
+               );
        }
 
        const allowedEntrypoints = process.env.ENTRY_LIMIT.split( ',' );
```

Which seems like we really didn't get the benefit of combining the loops, and the logic for adding the middleware is now repeated in multiple spots in the function

client/webpack.config.js
mreishus: I changed it to use fromEntries here: 4dd9489c. However, I am wondering if this code should even be here, see discussion below.

client/webpack.config.js
mreishus: Fixed

mreishus:
> The only thing I spotted initially was the webpack build monitor needs some love since it will remain indefinitely in the "Building" state after a change:
> 
> Screenshot 2023-04-25 at 14 59 09

This one is definitely a toughie! For me, it works sometimes and not others. I used this to debug and analyze:

```diff
diff --git a/client/components/webpack-build-monitor/index.tsx b/client/components/webpack-build-monitor/index.tsx
index 241f7391d8..6b3848f752 100644
--- a/client/components/webpack-build-monitor/index.tsx
+++ b/client/components/webpack-build-monitor/index.tsx
@@ -27,6 +27,7 @@ const WebpackBuildMonitor: FunctionComponent = () => {
 	const [ buildState, setBuildState ] = useState( BuildState.INITIAL );
 
 	useEffect( () => connectToWebpackServer( setBuildState ), [] );
+	console.log({buildState});
 
 	const msg = getMessage( buildState );
 	if ( ! msg ) {
diff --git a/client/components/webpack-build-monitor/webpack-client.ts b/client/components/webpack-build-monitor/webpack-client.ts
index a228959f69..98363d7e55 100644
--- a/client/components/webpack-build-monitor/webpack-client.ts
+++ b/client/components/webpack-build-monitor/webpack-client.ts
@@ -81,14 +81,17 @@ function processUpdate( message: UpdateMessage, setBuildState: BuildStateSetter
 
 	// hot update already in progress, triggered by another message handler
 	if ( hot.status() !== 'idle' ) {
+		console.log( '[webpack] hot update already in progress. Status is: ', hot.status() );
 		return;
 	}
 
 	setBuildState( BuildState.UPDATING );
 
+	console.log( '==> Calling CHECK' );
 	hot
 		.check( true )
 		.then( ( updatedModules ) => {
+			console.log( '<== CHECK return then, set idle' );
 			setBuildState( BuildState.IDLE );
 			console.log( `[webpack] hot updated ${ updatedModules.length } modules:` );
 			for ( const updatedModuleId of updatedModules ) {
@@ -96,6 +99,8 @@ function processUpdate( message: UpdateMessage, setBuildState: BuildStateSetter
 			}
 		} )
 		.catch( ( error ) => {
+			console.log( '<== CHECK return catch' );
+			console.log( '[webpack] hot update failed [NEEDS RELOAD]' );
 			setBuildState( BuildState.NEEDS_RELOAD );
 			console.error( '[webpack] hot update failed:', error );
 		} );
@@ -128,6 +133,9 @@ export default function connectToWebpackServer( setBuildState: BuildStateSetter
 
 		const message = JSON.parse( m.data );
 
+		console.log( 'message action:', message.action );
+		console.log( 'hot status:', ( module as HotModule ).hot?.status() );
+
 		switch ( message.action ) {
 			case 'building':
 				setBuildState( BuildState.BUILDING );
diff --git a/client/my-sites/customer-home/main.jsx b/client/my-sites/customer-home/main.jsx
index e88a776681..ce47a98c29 100644
--- a/client/my-sites/customer-home/main.jsx
+++ b/client/my-sites/customer-home/main.jsx
@@ -110,7 +110,7 @@ const Home = ( {
 			<FormattedHeader
 				brandFont
 				headerText={ translate( 'My Home' ) }
-				subHeaderText={ translate( 'Your hub for posting, editing, and growing your site.' ) }
+				subHeaderText={ translate( 'Your hub2 for posting, editing, and growing your site.' ) }
 				align="left"
 				hasScreenOptions
 			/>
diff --git a/client/my-sites/themes/theme-showcase.jsx b/client/my-sites/themes/theme-showcase.jsx
index 7f3ea9524e..495f73e216 100644
--- a/client/my-sites/themes/theme-showcase.jsx
+++ b/client/my-sites/themes/theme-showcase.jsx
@@ -519,6 +519,7 @@ class ThemeShowcase extends Component {
 						</>
 					) }
 				</ThemesHeader>
+				<div><div>test again3</div></div>
 				<div className="themes__content" ref={ this.scrollRef }>
 					<QueryThemeFilters />
 					<div className="themes__controls">
```

* **When it works and the build badge updates:**
  * Both the  `'==> Calling CHECK'` and `'<== CHECK return then, set idle'` log
  * When the message comes back, we see:
```
message action: built
hot status: idle
```

IDLE status means everything is done and we can do the status updates we want.

* **When it's broke and the build badge gets stuck:**
   * Neither  `'==> Calling CHECK'` and `'<== CHECK return then, set idle'` log
   * When the message comes back, we see:
```
message action: built
hot status: check
```

CHECK status means we're in the middle of a check and cannot do updates. The processUpdate() function exits early and we're not free to call hot.check() - it would error. I would expect a new message to come through the handler when this check finishes but one never does.

After messing around with a lot, I think the most likely theory is a conflict between `client/components/webpack-build-monitor` and the `node_modules/webpack-hot-middleware/process-update.js`. They're essentially trying to do the same thing - call `module.hot.check()` and provide feedback on what happens - and depending on which one 'wins', the build badge gets updated or not.

Check the source code of the two files:
  * our [webpack-build-monitor](https://github.com/Automattic/wp-calypso/blob/4dd9489c4725253dac63ceed6b21a5193f92a76a/client/components/webpack-build-monitor/webpack-client.ts#L1)
  * the [process-update.js](https://github.com/webpack-contrib/webpack-hot-middleware/blob/master/process-update.js#L1) from webpack-hot-middleware

I think they're trying to do the same thing and in conflict. If this is right, then the way forward is to pick which one we want to use, and remove the other. Either delete the webpack-build-monitor, or remove adding `webpack-hot-middleware` to the entrypoints.

**Option 1 (rough):** (more cleanup needed but I don't want to show a huge diff)
```diff
diff --git a/client/layout/index.jsx b/client/layout/index.jsx
index 49e58a7ad0..bbae7a66aa 100644
--- a/client/layout/index.jsx
+++ b/client/layout/index.jsx
@@ -351,7 +351,6 @@ class Layout extends Component {
                                { 'development' === process.env.NODE_ENV && (
                                        <>
                                                <SympathyDevWarning />
-                                               <AsyncLoad require="calypso/components/webpack-build-monitor" placeholder={ null } />
                                        </>
                                ) }
                                { loadInlineHelp && (
```

**Option 2 (rough):** (more cleanup needed but I don't want to show a huge diff)
```diff
diff --git a/client/webpack.config.js b/client/webpack.config.js
index 5b28409076..30118fc460 100644
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -115,6 +115,7 @@ function filterEntrypoints( entrypoints ) {
 }
 
 function addHotMiddlewareIfDev( entrypoints ) {
+       return entrypoints;
        if ( ! isDevelopment ) {
                return entrypoints;
        }
```

I think the HMR process is working the same in both scenarios, but I'm not 100% certain. I don't have a strong preference either way.

tyxla:
I've been able to find a change that always breaks this webpack build monitor:

```
diff --git a/client/state/purchases/selectors/get-purchases.js b/client/state/purchases/selectors/get-purchases.js
index bda888338f..48b4e85768 100644
--- a/client/state/purchases/selectors/get-purchases.js
+++ b/client/state/purchases/selectors/get-purchases.js
@@ -10,6 +10,6 @@ import 'calypso/state/purchases/init';
  * @returns {Array} Purchases
  */
 export const getPurchases = createSelector(
-       ( state ) => createPurchasesArray( state.purchases.data ),
+       ( state ) => [ createPurchasesArray( state.purchases.data )[0] ],
        ( state ) => [ state.purchases.data ]
 );
```

Just pick a site with purchases (a plan for example) and go to `/purchases/subscriptions/SITE_SLUG`. This is basically a selector that a dynamic reducer pulls, which is also used in a bunch of other selectors. With this change, I always get the neverending "building" indicator.

I personally find the webpack build monitor very handy, and wouldn't want it removed. I'd instead look into fixing it - I'd guess that it is just not prepared to use the data in the same format that the new tooling we use for HMR works. Haven't debugged it myself, though, but I'd say this is a blocker.

mreishus:
Since we want to keep our build monitor, give it a shot with the `webpack-hot-middleware/client` removed. I believe this eliminates the conflict and allows the build monitor to receive all of the messages. I've rebased and updated the branch with 2916d0a05da1ecce750b34420a886a52a4825913.  In my testing, I haven't been able to have it get stuck in this configuration yet.

markbiek:
Wanted to note that this _does_ seem to fix the `window.AppBoot is not a function` error we often get when working in the Stepper framework.

mreishus:
* Rebase
* Connected the stepper entry point to the build monitor and have it accept hot reloads

https://user-images.githubusercontent.com/937354/236071979-b7bcf516-95fa-4453-8c7d-66b109948d3c.mp4

mreishus:
Ok, I don't think I really have any other ideas to implement here.

tyxla:
Thank you for working on this @mreishus, we appreciate your effort 🙌 

I personally would love to spend some more time testing it and definitely want to see it merged as it's such a huge boost to developer experience. I hope to have more time to commit to it next week - I appreciate your patience in the meantime.

jsnajdr:
I started testing this now. Last time I looked at hot reloading was several years ago, and `react-refresh` didn't exist yet.

client/webpack.config.js
```diff
@@ -226,6 +227,7 @@ const webpackConfig = {
 				cacheIdentifier,
 				cacheCompression: false,
 				exclude: /node_modules\//,
+				plugins: [ isDevelopment && require.resolve( 'react-refresh/babel' ) ].filter( Boolean ),
```
jsnajdr: nit: could be also
```js
plugins: isDevelopment ? [ require.resolve( 'react-refresh/babel' ) ] : []
```
which IMO is more elegant.

packages/calypso-build/webpack/transpile.js
```diff
@@ -41,6 +43,7 @@ module.exports.loader = ( {
 				cacheIdentifier,
 				cacheCompression,
 				presets,
+				plugins: plugins || [],
```
jsnajdr: nit: could you just pass the raw `plugins` value, just like with `presets`? The `babel-loader` should accept an `undefined` value and treat it as empty-ish.

client/landing/stepper/index.tsx
```diff
@@ -149,3 +152,7 @@ window.AppBoot = async () => {
 		document.getElementById( 'wpcom' )
 	);
 };
+
+if ( module.hot ) {
+	module.hot.accept();
+}
```
jsnajdr: This explicit `module.hot.accept()` shouldn't be needed. Or did you have to add it in order to make the Stepper landing page work?

The `react-refresh` webpack runtime adds code to the end of every module, and this code calls `module.hot.accept()` if the module exports only React components. Only for such modules, `react-refresh` can handle the hot update and replace the module at runtime without causing any errors.

For other modules, hot updates won't be accepted. Like when you modify a reducer or something like that. Then we can't guarantee that the new code will be really used. References to old code can still be in place.

This is how `react-refresh` works: it supports hot updates only for React components and nothing else.

If you want to support hot reloading for the Stepper's top-level UI, the best approach is to extract it to a separate file and then import it:
```js
import Stepper from './component';
render( <Stepper flow={ flow } />, document.getElementById( 'wpcom' ) );
```

client/boot/app.js
```diff
@@ -8,3 +8,7 @@ import 'calypso/assets/stylesheets/style.scss';
 window.AppBoot = () => {
 	bootApp( 'Calypso' );
 };
+
+if ( module.hot ) {
+	module.hot.accept();
+}
```
jsnajdr: This also shouldn't be necessary.

client/landing/stepper/index.tsx
```diff
@@ -140,6 +140,9 @@ window.AppBoot = async () => {
 							id="notices"
 						/>
 					</BrowserRouter>
+					{ 'development' === process.env.NODE_ENV && (
+						<AsyncLoad require="calypso/components/webpack-build-monitor" placeholder={ null } />
+					) }
```
jsnajdr, line 145: It would be nice if all entrypoints had this build monitor. The login page, the logged-out layout... I see you experimented with adding a module to every entrypoint in the `webpack.config.js`. That could work for `webpack-build-monitor` as well. It would have to mount into its own React root, but that's a positive change. It would work more like the dev helpers which are independent from the main React layout.

jsnajdr:
> @jsnajdr might also have ideas about the reducer stuff

This problem is no longer present, or is it? Reducer modules shouldn't accept hot updates, because they have non-React exports. A full reload should be forced in such a case. If you were previously getting errors about reducers not being equal, it seems that a reducer hot update _was_ accepted for some reason.

jsnajdr:
I've spent some time refreshing my knowledge about webpack hot reloading, learning about `react-refresh`, and testing this PR, and it works really well! I would remove the explicit calls to `module.hot.accept()`. Otherwise, this is mergeable.

We should follow up with adding the `webpack-build-monitor` to all entrypoints.

> After messing around with a lot, I think the most likely theory is a conflict between `client/components/webpack-build-monitor` and the `node_modules/webpack-hot-middleware/process-update.js`.

Yes, I can confirm that the `webpack-build-monitor` implements its own HMR client. It connects to the `/__webpack_hmr` event source and handles the incoming messages, including the hot updates. Running two HMR clients at once leads to conflicts, of course.

tyxla (approved):
I gave this a fair amount of testing and on top of Jarda's feedback I think it's good to go! 🚀 

There will surely be some improvements like getting `webpack-build-monitor` to work properly, but we can always follow-up. Also, having it shipped might unravel more bugs to work on as it gets tested by more developers. 

Thanks @mreishus for taking care of it 🙌

client/landing/stepper/index.tsx
mreishus: I think it is needed. I tested this branch by removing the explicit reloads:

```diff
diff --git a/client/boot/app.js b/client/boot/app.js
index 58f23397f7..dcf82c105e 100644
--- a/client/boot/app.js
+++ b/client/boot/app.js
@@ -8,7 +8,3 @@ import 'calypso/assets/stylesheets/style.scss';
 window.AppBoot = () => {
        bootApp( 'Calypso' );
 };
-
-if ( module.hot ) {
-       module.hot.accept();
-}
diff --git a/client/landing/stepper/index.tsx b/client/landing/stepper/index.tsx
index 0835a41c27..0df1207135 100644
--- a/client/landing/stepper/index.tsx
+++ b/client/landing/stepper/index.tsx
@@ -152,7 +152,3 @@ window.AppBoot = async () => {
                document.getElementById( 'wpcom' )
        );
 };
-
-if ( module.hot ) {
-       module.hot.accept();
-}
```

Then tried basic hot-reloading by making changes to the html provided by `client/my-sites/customer-home/main.jsx` and `client/landing/stepper/declarative-flow/internals/steps-repository/store-profiler/index.tsx`. (nothing in the data layer). I found that most of the time, the updates didn't work:

Screenshots from a session where it didn't work:




I'm not sure what's happening though, because I think it did work with them removed at least once. However, I did just restart calypso 5 or 6 times and can't seem to get any hot reloads working with these removed anymore.

packages/calypso-build/webpack/transpile.js
mreishus: Fixed in ef135252ebc

client/webpack.config.js
mreishus: Fixed in dfd3ee941e

mreishus:
* Rebased

mreishus:
I don't think it's worth merging with the `module.hot.accept();`s removed. I'm not able to do basic reloading with them gone.

client/landing/stepper/index.tsx
jsnajdr: It works for with the `accept` calls removed when I change only the JSX markup in `customer-home/main.jsx`. Then the dev server creates updates for the `home` and `runtime` assets:
```
webpack building...
webpack built be80d0323231fa6ba8ad in 4178ms
assets by status 167 MiB [cached] 2126 assets
assets by path . 1.12 MiB
  assets by info 21.3 KiB [immutable]
    asset home.3f0d1f8c4c2e3e97aa60.hot-update.js 20.4 KiB [emitted] [immutable] [hmr] (name: home)
    asset runtime.3f0d1f8c4c2e3e97aa60.hot-update.js 859 bytes [emitted] [immutable] [hmr] (name: runtime)
    asset runtime.3f0d1f8c4c2e3e97aa60.hot-update.json 38 bytes [emitted] [immutable] [hmr]
  asset home.js 1.02 MiB [emitted] (name: home)
  asset runtime.js 75.5 KiB [emitted] (name: runtime)
cached modules 51.6 MiB (javascript) 18 MiB (asset) 12.7 MiB (css/mini-extract) [cached] 15736 modules
runtime modules 56.2 KiB 24 modules
./my-sites/customer-home/main.jsx 9.74 KiB [built] [code generated]
webpack 5.68.0 compiled successfully in 4178 ms

Ready! All assets are re-compiled. Have fun!
```
and the browser accepts the update:
```
[webpack] hot updated 1 modules:
webpack-client.ts:97 ./my-sites/customer-home/main.jsx
```
because the `main.jsx` module has `react-refresh` code, added by the Babel and webpack plugins, that accepts the update if the React exports are compatible.

In your screenshots, however, I see there is also an update for the `sections` module, which is not accepted because its exports aren't React components:
```
[webpack] hot update failed: Error: Aborted because ./sections.is is not accepted
Update propagation: ./sections.js -> ./sections-middleware.js -> ./boot/common. js -> ./boot/app.js
```
Here webpack also asks the parent modules if they `accept` and that's why `module.hot.accept()` in the `./boot/app.js` module works.

I'm wondering where this `sections` update is coming from. How are the old and new modules different? What does webpack report on the server, in the terminal console, when building this update?

Anyway, I think you can merge this PR even with the `module.hot.accept()` calls, they are harmless. Once we figure out what's going on with the `sections` update we can fix that in a followup.

jsnajdr: Adding an unconditional `module.hot.accept()` call to the entrypoint module also has the effect that _every_ hot update will be accepted, unless a child module, i.e., the React Refresh runtime, explicitly rejected it. It does that explicit rejection only when a module goes from "React Refresh good" to "React Refresh bad". It doesn't have any opinion on modules that export something else than React components.

We don't want to accept any update, only the ones that can be safely applied.

mreishus:
Will look at deploying this tomorrow, and tracking down what's different about my setup. Got distracted by another project, sorry.

jsnajdr:
If it turns out that the `sections` module needs hot updates for a good reason, we can do the `module.hot.accept()` call only in that module, not in the entrypoint parent. The `sections` module is special, processed by its own webpack loader that adds generated code (for loading the section chunks), so it's quite plausible that it has some special behavior hot-reloading-wise, too.

mreishus:
I was able to get it working by removing the generic accepts and adding a new accept for sections. 46db022d7c9

mreishus:
Some things in stepper still don't work - but it's related to the complex flow system that uses dynamic requires. There will be probably need to be something in the future like https://github.com/Automattic/wp-calypso/commit/46db022d7c9b9a60841a0cc52dc3b9ab21beeea9 that implements a strategy for updating stepper.

See #77119 where I sketched out one idea that seems to work. But seems too hacky to include last minute.

client/landing/stepper/index.tsx
mreishus: This is a good idea. I tried putting together something quickly but I ended up stopping the app from booting. :)

tyxla:
@mreishus would you be interested to post about this on the calypso p2? It's a pretty great improvement that devs will love to find out about. 💯

client/sections-middleware.js
```diff
@@ -114,3 +114,11 @@ export const setupRoutes = () => {
 		}
 	}
 };
+
+if ( module.hot ) {
+	module.hot.accept( './sections', () => {
+		const updatedSections = require( './sections' ).default;
+		receiveSections( updatedSections );
+		setupRoutes();
+	} );
```
jsnajdr, line 123: I'm wondering in what situations does the `sections` hot update happen. When merely changing a React component, only the one module with that component should hot-update. But it seems that for you, it sometimes leads to a `sections` update, too? Did you investigate why did that happen? It would be nice to compare the incoming `sections.hot-update.js` module with the existing `sections.js`. What has changed?

mreishus: I'm not sure why it's happening. If I do these steps:
* `yarn && yarn start`
* Visit `http://calypso.localhost:3000/home/<mysite>`
* Make this change:
```diff
diff --git a/client/my-sites/customer-home/main.jsx b/client/my-sites/customer-home/main.jsx
index e88a776681..ce47a98c29 100644
--- a/client/my-sites/customer-home/main.jsx
+++ b/client/my-sites/customer-home/main.jsx
@@ -110,7 +110,7 @@ const Home = ( {
                        <FormattedHeader
                                brandFont
                                headerText={ translate( 'My Home' ) }
-                               subHeaderText={ translate( 'Your hub for posting, editing, and growing your site.' ) }
+                               subHeaderText={ translate( 'Your hub2 for posting, editing, and growing your site.' ) }
                                align="left"
                                hasScreenOptions
                        />
```

In my console I see:
```
[webpack] hot updated 2 modules: webpack-client.ts:94:13
./my-sites/customer-home/main.jsx webpack-client.ts:97:15
./sections.js
```

In my terminal I see:
```
webpack building...
<i> [webpack-dev-middleware] wait until bundle finished: /home/testsite5mmr.word press.com
<i> [webpack-dev-middleware] wait until bundle finished: /nostats.js?_ut=wpcom%3Auser_id&_ui=73763986
<i> [webpack-dev-middleware] wait until bundle finished: /version?1684511999671
webpack built cec808f1f61c2467932f in 10135ms
assets by status 168 MiB [cached] 2105 assets
assets by path . 2.29 MiB
  assets by info 288 KiB [immutable]
    asset boot_common_js.194cf1bb8654a5a5279b.hot-update.js 265 KiB [emitted] [immutable] [hmr]
    asset home.194cf1bb8654a5a5279b.hot-update.js 20.4 KiB [emitted] [immutable] [hmr] (name: home)
    asset runtime.194cf1bb8654a5a5279b.hot-update.js 2.94 KiB [emitted] [immutable] [hmr] (name: runtime)
    asset runtime.194cf1bb8654a5a5279b.hot-update.json 55 bytes [emitted] [immutable] [hmr]
  asset home.js 1.03 MiB [emitted] (name: home)
  asset boot_common_js.js 787 KiB [emitted]
  asset vendors-node_modules_socket_io-client_lib_index_js.js 146 KiB [emitted] (id hint: vendors)
  asset runtime.js 74.7 KiB [emitted] (name: runtime)
cached modules 52.2 MiB (javascript) 16.8 MiB (asset) 12.7 MiB (css/mini-extract) [cached] 15761 modules
runtime modules 55.4 KiB 24 modules
javascript modules 41.3 KiB
  cacheable modules 41 KiB
    modules by path ./ 28.6 KiB
      ./sections.js 18.9 KiB [code generated]
      ./my-sites/customer-home/main.jsx 9.74 KiB [built] [code generated]
    modules by path ../node_modules/ 12.4 KiB
      ../node_modules/socket.io-client/lib/manager.js 12.2 KiB [code generated]
      ../node_modules/engine.io-client/lib/index.js 140 bytes [code generated]
  ../packages/components/src/animation/features/ lazy ^.*$ namespace object 160 bytes [built]
  ./signup/steps/ lazy ^\.\/.*$ include: signup\/steps\/[0-9a-z/-]+\/index\.[j%7Ct]sx$ exclude: signup\/steps\/[0-9a-z/-]+\/test\/index\.[j%7Ct]sx$ chunkName: async-load-signup-steps-[request] namespace object 160 bytes [built]
webpack 5.68.0 compiled successfully in 10135 ms

Ready! All assets are re-compiled. Have fun!

```

And my new code block runs.
I'm not sure how to see what's different; I tried this form of logging:
```diff
diff --git a/client/sections-middleware.js b/client/sections-middleware.js
index f4b8a2f698..558bf1c9c6 100644
--- a/client/sections-middleware.js
+++ b/client/sections-middleware.js
@@ -11,6 +11,7 @@ import isSectionEnabled from './sections-filter';
 import { receiveSections, load } from './sections-helper';
 import { pathToRegExp } from './utils';
 
+console.log(sections);
 receiveSections( sections );
 
 function activateSection( section, context ) {
@@ -118,6 +119,7 @@ export const setupRoutes = () => {
 if ( module.hot ) {
        module.hot.accept( './sections', () => {
                const updatedSections = require( './sections' ).default;
+               console.log(updatedSections);
                receiveSections( updatedSections );
                setupRoutes();
        } );
```

Then using chrome devtools to assign each to a global variable, copy() it, paste them to files and diff them, and they were the exact same.

If I look at the network tab, the important requests seem to be to `http://calypso.localhost:3000/calypso/evergreen/home.5f3b16f21430644f17b8.hot-update.js` and `http://calypso.localhost:3000/calypso/evergreen/boot_common_js.5f3b16f21430644f17b8.hot-update.js`. I suppose the offender here is it sending me a new boot_common_js. But it's a huge string that's `evaled()` that I can't make much sense of - about 275K of text.  It does start out with a reference to sections though:

```
globalThis["webpackHotUpdatecalypso"]("boot_common_js",{

/***/ "./sections.js":
/***/ ((module, __webpack_exports__, __webpack_require__) => {
eval("__webpack_require__.r(__webpack_exports__);\n/* harmony export */ __webpack_requi ...
```

Edit: After changing devtool to false, I can see it is indeed the entire sections.js file it's sending me:
```
"use strict";
globalThis["webpackHotUpdatecalypso"]("boot_common_js",{

/***/ "./sections.js":
/***/ ((module, __webpack_exports__, __webpack_require__) => {

__webpack_require__.r(__webpack_exports__);
/* harmony export */ __webpack_require__.d(__webpack_exports__, {
/* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
/* harmony export */ });
/* provided dependency */ var __react_refresh_utils__ = __webpack_require__("../node_modules/@pmmmwh/react-refresh-webpack-plugin/lib/runtime/RefreshUtils.js");
__webpack_require__.$Refresh$.runtime = __webpack_require__("../node_modules/react-refresh/runtime.js");

/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ([{
  "name": "root",
  "paths": ["/"],
  "module": "calypso/root",
  "group": "root",
  "enableLoggedOut": true,
  "load": () => __webpack_require__.e(/* import() | root */ "root").then(__webpack_require__.bind(__webpack_require__, "./root.js"))
}, {
  "name": "customize",
  "paths": ["/customize"],
  "module": "calypso/my-sites/customize",
  "group": "sites",
  "load": () => Promise.all(/* import() | customize */[__webpack_require__.e("vendors-node_modules_wordpress_components_build-module_button_index_js-node_modules_wordpress-0aa56a"), __webpack_require__.e("vendors-node_modules_emotion_react_dist_emotion-react_browser_esm_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_ui_utils_space_js-node_modules_wordpre-0b15d0"), __webpack_require__.e("vendors-node_modules_lodash-es__createFlow_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_spinner_index_js"), __webpack_require__.e("vendors-packages_search_node_modules_fuse_js_dist_fuse_esm_js-node_modules_lodash-es_partition_js"), __webpack_require__.e("vendors-node_modules_wordpress_icons_build-module_library_close-small_js-node_modules_wordpre-067a0c"), __webpack_require__.e("vendors-node_modules_deep-freeze_index_js-node_modules_to-title-case_index_js-node_modules_lo-9cc1a5"), __webpack_require__.e("components_forms_form-text-input_style_scss"), __webpack_require__.e("state_media_init_js"), __webpack_require__.e("lib_query-manager_paginated_index_js-lib_query-manager_with-query-manager_js"), __webpack_require__.e("blocks_site_index_jsx-state_selectors_get-sites_js"), __webpack_require__.e("lib_titan_index_js"), __webpack_require__.e("components_count_index_jsx-components_forms_form-text-input_index_jsx-lib_analytics_page-view-be817b"), __webpack_require__.e("components_site-selector_index_jsx"), __webpack_require__.e("state_purchases_actions_js-state_purchases_selectors_fetching_js-state_purchases_selectors_ge-2a27a3"), __webpack_require__.e("blocks_site_placeholder_jsx-components_empty-component_index_jsx-components_search_index_jsx--227b29"), __webpack_require__.e("state_posts_init_js"), __webpack_require__.e("state_themes_init_js"), __webpack_require__.e("customize")]).then(__webpack_require__.bind(__webpack_require__, "./my-sites/customize/index.js"))
}, {
  "name": "sites-dashboard",
  "paths": ["/sites"],
  "module": "calypso/sites-dashboard",
  "group": "sites-dashboard",
  "load": () => Promise.all(/* import() | sites-dashboard */[__webpack_require__.e("vendors-node_modules_wordpress_components_build-module_button_index_js-node_modules_wordpress-0aa56a"), __webpack_require__.e("vendors-node_modules_emotion_react_dist_emotion-react_browser_esm_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_ui_utils_space_js-node_modules_wordpre-0b15d0"), __webpack_require__.e("vendors-node_modules_tanstack_query-core_build_lib_queryObserver_mjs-node_modules_tanstack_re-632692"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_spinner_index_js"), __webpack_require__.e("vendors-packages_search_node_modules_fuse_js_dist_fuse_esm_js-node_modules_lodash-es_partition_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_flex_context_js-node_modules_wordpress-7d345a"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_base-control_styles_base-control-styles_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_h-stack_component_js"), __webpack_require__.e("vendors-node_modules_react-intersection-observer_react-intersection-observer_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_dropdown-menu_index_js-node_modules_wo-0a91b5"), __webpack_require__.e("components_notice_style_scss"), __webpack_require__.e("components_forms_form-text-input_style_scss"), __webpack_require__.e("lib_titan_index_js"), __webpack_require__.e("components_count_index_jsx-components_forms_form-text-input_index_jsx-lib_analytics_page-view-be817b"), __webpack_require__.e("state_purchases_actions_js-state_purchases_selectors_fetching_js-state_purchases_selectors_ge-2a27a3"), __webpack_require__.e("packages_shopping-cart_src_shopping-cart-manager_ts"), __webpack_require__.e("state_plugins_init_js"), __webpack_require__.e("state_automated-transfer_actions_js"), __webpack_require__.e("packages_data-stores_src_site_index_ts-packages_data-stores_src_user_index_ts"), __webpack_require__.e("state_stats_init_js"), __webpack_require__.e("packages_data-stores_src_automated-transfer-eligibility_index_ts-packages_data-stores_src_onb-a29d3e"), __webpack_require__.e("components_site-preview-link_index_tsx-landing_stepper_hooks_use-site-copy_tsx-state_sites_la-f20c6e"), __webpack_require__.e("sites-dashboard")]).then(__webpack_require__.bind(__webpack_require__, "./sites-dashboard/index.ts"))
}, {
  "name": "account",
  "paths": ["/me/account"],
  "module": "calypso/me/account",
  "group": "me",
  "load": () => Promise.all(/* import() | account */[__webpack_require__.e("vendors-node_modules_wordpress_components_build-module_button_index_js-node_modules_wordpress-0aa56a"), __webpack_require__.e("vendors-node_modules_emotion_react_dist_emotion-react_browser_esm_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_ui_utils_space_js-node_modules_wordpre-0b15d0"), __webpack_require__.e("vendors-node_modules_lodash-es__createFlow_js"), __webpack_require__.e("vendors-node_modules_tanstack_query-core_build_lib_queryObserver_mjs-node_modules_tanstack_re-632692"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_spinner_index_js"), __webpack_require__.e("vendors-packages_search_node_modules_fuse_js_dist_fuse_esm_js-node_modules_lodash-es_partition_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_flex_context_js-node_modules_wordpress-7d345a"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_base-control_styles_base-control-styles_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_h-stack_component_js"), __webpack_require__.e("vendors-node_modules_path-browserify_index_js-node_modules_valid-url_index_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_input-control_input-base_js-node_modul-0e31f1"), __webpack_require__.e("vendors-node_modules_phone_dist_index_js"), __webpack_require__.e("vendors-node_modules_qrcode_react_lib_index_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_custom-select-control_index_js-node_mo-45131c"), __webpack_require__.e("vendors-node_modules_use-debounce_esm_useDebounce_js-node_modules_validator_index_js-node_mod-e10448"), __webpack_require__.e("vendors-node_modules_express-useragent_index_js"), __webpack_require__.e("vendors-node_modules_lottie-web_build_player_lottie_light_js"), __webpack_require__.e("vendors-node_modules_react-transition-group_esm_CSSTransition_js"), __webpack_require__.e("vendors-node_modules_wordpress_components_build-module_toggle-control_index_js-node_modules_w-f60480"), __webpack_require__.e("state_media_init_js"), __webpack_require__.e("lib_query-manager_paginated_index_js-lib_query-manager_with-query-manager_js"), __webpack_require__.e("blocks_site_index_jsx-state_selectors_get-sites_js"), __webpack_require__.e("lib_titan_index_js"), __webpack_require__.e("components_count_index_jsx-components_forms_form-text-input_index_jsx-lib_analytics_page-view-be817b"), __webpack_require__.e("components_site-selector_index_jsx"), __webpack_require__.e("packages_shopping-cart_src_shopping-cart-manager_ts"), __webpack_require__.e("lib_cart-values_cart-items_ts"), __webpack_require__.e("my-sites_checkout_calypso-shopping-cart-provider_tsx"), __webpack_require__.e("components_badge_index_jsx-components_inline-support-link_index_jsx-lib_analytics_track-compo-35a0b3"), __webpack_require__.e("lib_two-step-authorization_index_js-me_reauth-required_index_jsx"), __webpack_require__.e("blocks_edit-gravatar_index_jsx"), __webpack_require__.e("me_controller_js-my-sites_checkout_cart-manager-client_ts-components_badge_style_scss-compone-be08b4"), __webpack_require__.e("components_close-on-escape_index_jsx-components_forms_form-buttons-bar_index_jsx-components_f-28eff6"), __webpack_require__.e("account")]).then(__webpack_require__.bind(__webpack_require__, "./me/account/index.js"))
...(truncated)...
```
But I don't know why.

```

mreishus:
> @mreishus would you be interested to post about this on the calypso p2? It's a pretty great improvement that devs will love to find out about. 100

For sure. Scheduled a post to go off Monday (morning USA / afternoon Europe). Was worried it wouldn't be seen as much if I posted late on a Friday.

PR #76127, opened 2023-04-24
Staging Sites: Fix staging site check that breaks user account in some cases

### LINKED ISSUE #76099: User must make many attempts to actually log out, intermittently
### Quick summary

TL;DR: User frequently has to try several (or even many) times to log out and sees either blank pages or errors when trying, until after repetition, it finally works.

This is a weird, intermittent bug that a user is encountering. I cannot personally reproduce, nor have any HEs been able to. There is a javascript error I wanted to note in case it was relevant. I don't think [this PR](https://github.com/Automattic/wp-calypso/pull/73855) is relevant, but it's the only fairly recent thing I could find that referenced the javascript property that's sending an error to the user's browser.

See 6080495-zd-woothemes.

### Steps to reproduce

I added the console steps that begin and end this repro list, to see if the user could spot anything useful in the console that might point to a server-side issue, as this has occurred on multiple browsers and devices with cookies/etc. cleared.

1. Load the browser console.
2. Click the user avatar at the top right of the admin bar from any screen. This should go to /me.
3. Observe that a blank screen loads rather than the profile.
5. Click back to get to the prior page.
6. See step 3. Keep trying this until:
7. The message "We're sorry but an unexpected error has occurred" appears in the main content area to the right of the usual admin sidebar.
8. Look for any errors in the browser console (my debugging step for the user).

The user is logged out and sees an appropriate logout screen.

### What actually happened

The user reports:

> When I try to log out, the page My Profile appears, over and over again. Sometimes have to try three or four times to log out.

The user also reported that sometimes it takes many more tries (like a couple of dozen). And sometimes the logout works as expected, so it's intermittent.

Further, when they sent me info from the browser console, this one came up:

Screen Shot 2023-04-21 at 3 56 46 PM

It may not be related to this at all (I sort of doubt it is, in fact), but a javascript error seemed worth reporting in any case.

### Impact

One

### Available workarounds?

Simple

### Logs or notes

The user has cleared cache and cookies on multiple browsers and computers. They have also seen this behavior on their phone. I don't think we can consider this a browser-specific issue, hence my suggesting we look for console errors that _could_ point to some other (maybe server-side?) problem. I don't actually think the error sent back actually points to such an issue.

The user says he can live with the issue but wanted to bring it to our attention. This seems like it must be some kind of edge case, as I could find no other similar reports. So there may not be much to do here, but perhaps the js error is worth a look.

wojtekn:
Fixes https://github.com/Automattic/wp-calypso/issues/76099

## Proposed Changes

In this PR, I propose to use optional chaining to avoid breaking UI if not all site data is loaded.

The issue popped out on the domain upsell block on the customer profile page via staging site check from `client/sites-dashboard/utils.ts`. To reproduce without setting upsell on the data level:

1. Run Calypso on local environment
2. Open customer profile
3. Edit the `client/my-sites/customer-home/cards/features/domain-upsell/index.jsx` file and remove three other checks from the condition that includes a check for `isStagingSite`
4. Refresh customer profile

## Testing Instructions

1. Run Calypso on the local environment or use Calypso Live
2. Open customer profile
3. Confirm page loads without an error

Additionally, check if:

- staging site has a badge on the Sites page in both table and grid modes
- staging site has a badge in the left sidebar menu
- staging site doesn’t have a link to Staging sites on Sites page context menu

## Pre-merge Checklist

### DIFF
--- a/client/sites-dashboard/components/sites-ellipsis-menu.tsx
+++ b/client/sites-dashboard/components/sites-ellipsis-menu.tsx
@@ -263,7 +263,7 @@ const SiteDropdownMenu = styled( DropdownMenu )( {
 function useSubmenuItems( site: SiteExcerptData ) {
 	const { __ } = useI18n();
 	const siteSlug = site.slug;
-	const isStagingSite = site.is_wpcom_staging_site;
+	const isStagingSite = site?.is_wpcom_staging_site;
 	const isStagingSiteEnabled = isEnabled( 'yolo/staging-sites-i1' );
 	const hasStagingSitesFeature = useSafeSiteHasFeature( site.ID, FEATURE_SITE_STAGING_SITES );
 
--- a/client/sites-dashboard/components/sites-ellipsis-menu.tsx
+++ b/client/sites-dashboard/components/sites-ellipsis-menu.tsx
@@ -39,6 +39,7 @@ import {
 	isCustomDomain,
 	isNotAtomicJetpack,
 	isP2Site,
+	isStagingSite,
 } from '../utils';
 import type { SiteExcerptData } from 'calypso/data/sites/site-excerpt-types';
 
@@ -263,7 +264,7 @@ const SiteDropdownMenu = styled( DropdownMenu )( {
 function useSubmenuItems( site: SiteExcerptData ) {
 	const { __ } = useI18n();
 	const siteSlug = site.slug;
-	const isStagingSite = site?.is_wpcom_staging_site;
+	const isWpcomStagingSite = isStagingSite( site );
 	const isStagingSiteEnabled = isEnabled( 'yolo/staging-sites-i1' );
 	const hasStagingSitesFeature = useSafeSiteHasFeature( site.ID, FEATURE_SITE_STAGING_SITES );
 
@@ -283,7 +284,7 @@ function useSubmenuItems( site: SiteExcerptData ) {
 				sectionName: 'database_access',
 			},
 			{
-				condition: ! isStagingSite && isStagingSiteEnabled && hasStagingSitesFeature,
+				condition: ! isWpcomStagingSite && isStagingSiteEnabled && hasStagingSitesFeature,
 				label: __( 'Staging site' ),
 				href: `/hosting-config/${ siteSlug }#staging-site`,
 				sectionName: 'staging_site',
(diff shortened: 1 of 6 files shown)


danielbachhuber:
How did we end up with a Site object without the `is_wpcom_staging_site` attribute?

wojtekn:
@danielbachhuber we get it from `getSelectedSite`, which can return null.

kozer (approved):

danielbachhuber:
> we get it from `getSelectedSite`, which can return null.

@wojtekn Sounds good.

katinthehatsite:
Works as expected for me, no errors visible!

client/sites-dashboard/utils.ts
```diff
@@ -45,7 +45,7 @@ export const isP2Site = ( site: SiteExcerptNetworkData ) => {
 };
 
 export const isStagingSite = ( site: SiteExcerptNetworkData ) => {
```
sejas: @wojtekn , should we update the type of the `site` parameter?

```suggestion
export const isStagingSite = ( site: SiteExcerptNetworkData | undefined) => {
```

client/sites-dashboard/components/sites-ellipsis-menu.tsx
```diff
@@ -263,7 +263,7 @@ const SiteDropdownMenu = styled( DropdownMenu )( {
 function useSubmenuItems( site: SiteExcerptData ) {
 	const { __ } = useI18n();
 	const siteSlug = site.slug;
-	const isStagingSite = site.is_wpcom_staging_site;
+	const isStagingSite = site?.is_wpcom_staging_site;
```
sejas: Ideally, inside all the `sites-dashboard` folder we could use the `isStagingSite` function directly importing it from `client/sites-dashboard/utils.ts`

sejas (approved):
@wojtekn , the change looks good, and it's better to be safe.
 I've seen the site can be undefined in some cases during the deleting staging sites flow 👍 .

I added a couple of suggestions. No strong opinion though.

client/sites-dashboard/components/sites-ellipsis-menu.tsx
wojtekn: Thanks, I've modified the code to use this existing function.

danielbachhuber (approved):
I confirmed that the `Staging` label still appears as expected in a few places, and was able to successfully log out.

Landing this so we can get the fix in front of users.

PR #75594, opened 2023-04-11
Design Picker: Switch to `react-intersection-observer`

danielbachhuber:
See https://github.com/Automattic/wp-calypso/issues/74022
Previously https://github.com/Automattic/wp-calypso/pull/71064

## Proposed Changes

Switches to `react-intersection-observer` in the Design Picker because `calypso/lib/use-in-view` is buggy and it doesn't make sense to spend time reinventing the wheel.

## Testing Instructions

1. Create a new Premium site.
2. Hit `Continue` until you get to the Design Picker.
3. Click on a category and verify `calypso_signup_design_scrolled_to_end` is only fired once when you hit the end of the page (even if you scroll to the bottom of the page multiple times).

image

PR #71064, opened 2022-12-12
Track if user scrolls to end of Design Picker

bangank36:
#### Proposed Changes

* ~~Introduce a new hook to track scroll when it reaches the bottom of the page~~
* We decided to take advantage of [useInView](https://github.com/Automattic/wp-calypso/blob/30d5518166fbcc8f40ddc6167fd7ee70cd3c4c62/client/lib/use-in-view/index.ts#L3) to track instead of onscroll
* Trigger event `calypso_signup_design_scrolled_to_end` with 2 props `intent` and `site_category` when unified design picker page is scrolled to bottom

#### Testing Instructions

* Set up your browser so you can easily see Tracks events -- I would strongly recommend the [Tracks Vigilante Chrome extension](https://github.com/Automattic/tracks-chrome-extension)
* Open the Calypso live branch for this PR (or via your local Calypso environment)
* Navigate to the Design Picker step -- you can navigate directly there via `/setup/site-setup/designSetup?siteSlug=${ siteSlug }`
* Inspect the Tracks events logged for `calypso_signup_design_scrolled_to_end`. Also verify that the events have valid properties
* Confirm that the tracks event only fired once when category changed

Recording:

https://user-images.githubusercontent.com/10071857/207592582-0a99513f-f8e7-4686-a874-7b6daa81df5b.mp4

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
[...]

PR #76136, opened 2023-04-24
Design Picker: Switch to `react-intersection-observer` (v2)

danielbachhuber:
Previously https://github.com/Automattic/wp-calypso/pull/71064 (and https://github.com/Automattic/wp-calypso/pull/75594)

## Proposed Changes

Switches to `react-intersection-observer` in the Design Picker because `calypso/lib/use-in-view` is buggy and it doesn't make sense to spend time reinventing the wheel.

## Testing Instructions

1. Create a new Premium site.
2. Hit `Continue` until you get to the Design Picker.
3. Click on a category and verify `calypso_signup_design_scrolled_to_end` is fired when you hit the end of the page.

image


packages/design-picker/src/components/unified-design-picker.tsx
```diff
@@ -348,8 +348,20 @@ const UnifiedDesignPicker: React.FC< UnifiedDesignPickerProps > = ( {
...
-	// Track as if user has scrolled to bottom of the design picker
-	const ref = useInView< HTMLDivElement >( onViewAllDesigns, {}, [ categorization?.selection ] );
+	const [ viewedSelections, setViewedSelections ] = useState< string[] >( [] );
+
+	const { ref } = useInView( {
+		onChange: ( inView ) => {
+			if (
+				inView &&
+				categorization?.selection &&
+				! viewedSelections.includes( categorization?.selection )
```
okmttdhr: badge badge

This condition will likely prevent the event from firing if the user scrolls to the end a second time in the same category.

e.g., 

- Select `Blog`
- Scroll to the bottom 
	- The event fire 🔥 
- Scroll to the top and Select `Business`
- Scroll to the bottom 
	- The event fire 🔥 
- **Scroll to the top and Select `Blog` again**
- Scroll to the bottom 
	- **The event does not fire** 🧑‍🚒 

I thought this was not intended and should be changed to the original behavior, but WDYT? @Automattic/lego @autumnfjeld

danielbachhuber: > This condition will likely prevent the event from firing if the user scrolls to the end a second time in the same category.

This is correct. I thought this was the original intent...

taipeicoder: @danielbachhuber thank you for working on this 🙂
I went to the production environment to double-check, and confirmed that the event would trigger again after switching categories (which means that there is no need to keep track of the viewed categories).

autumnfjeld:
> Create a new Premium site.
@danielbachhuber  Why does the site have to be premium? 

The event calypso_signup_design_scrolled_to_end should fire the same way regardless of plan, ya?

I've never looked at data from this event before, but I guess it would make sense for it to behave as @okmttdhr  [describes](https://github.com/Automattic/wp-calypso/pull/76136#discussion_r1176078488), per category.    
Select Blog
Scroll to the bottom
The event fire 🔥
Scroll to the top and Select Business
Scroll to the bottom
The event fire 🔥
Scroll to the top and Select Blog again
Scroll to the bottom
The event does not fire 🧑‍🚒

Also we know for instance such an event is only useful for some categories where a scroll action is needed to see all designs.  For example there is only one podcast, so the event is not providing interesting data there.   In fact, why is it even firing on the podcast page?   Ohhhh, because of the Design your own CTA makes the page scrollable. Hmmm.  

It appears the event was already tracking "per category" (what @taipeicoder  is saying above).

image

At any rate, thanks for making it fire correctly! : )

danielbachhuber:
@autumnfjeld @taipeicoder I'm a little bit confused about this feedback. Should I:
1. Keep the existing behavior and only fire the event once per category.
2. Remove the category tracking and fire the event every time the user hits the bottom of the page.

Thanks!

taipeicoder:
> @autumnfjeld @taipeicoder I'm a little bit confused about this feedback. Should I:
> 
> 1. Keep the existing behavior and only fire the event once per category.
> 2. Remove the category tracking and fire the event every time the user hits the bottom of the page.
> 
> Thanks!

Hi @danielbachhuber, let's keep the existing production behavior (i.e.: not tracking viewed categories) for now. We can change it afterward if needed. Thank you!

danielbachhuber:
> let's keep the existing production behavior (i.e.: not tracking viewed categories) for now. We can change it afterward if needed.

@taipeicoder Sounds good, I've changed the behavior with 0265de98c6

PR #76156, opened 2023-04-24
Woo Express: Fix interval toggle overflow on small mobile screens

### LINKED ISSUE #75487: Woo Express Essential: Monthly/Yearly toggle overflow
CfT: pdtkmj-1mM-p2#comment-2527

The "Pay Monthly"/"Pay Annually" buttons overflow on small mobile screens (eg iPhone SE).

image

markbiek:
Fixed #75487

## Proposed Changes

On some older iPhones with small screens (like the SE), the plan interval toggle buttons overflowed the container.

image

This adds a new breakpoint for tiny mobile screens and shrinks the font slightly to prevent the overflow.

## Testing Instructions

* Using the calypso.live link, access the `/plans` page for a Woo Express site from either a small mobile device or use your browser's responsive display tools to view the site with a page with of 375px or less.
* Verify that the "Pay Annually" button isn't overflowing the container.
image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plans/ecommerce-trial/wooexpress-plans/index.tsx
+++ b/client/my-sites/plans/ecommerce-trial/wooexpress-plans/index.tsx
@@ -9,6 +9,8 @@ import {
 	getPlans,
 	isWooExpressPlan,
 } from '@automattic/calypso-products';
+import { useIsEnglishLocale } from '@automattic/i18n-utils';
+import { hasTranslation } from '@wordpress/i18n';
 import { useTranslate } from 'i18n-calypso';
 import page from 'page';
 import { useCallback, useMemo } from 'react';
@@ -46,6 +48,7 @@ export function WooExpressPlans( props: WooExpressPlansProps ) {
 		yearlyControlProps,
 	} = props;
 	const translate = useTranslate();
+	const isEnglishLocale = useIsEnglishLocale();
 
 	const mediumPlanAnnual = getPlans()[ PLAN_WOOEXPRESS_MEDIUM ];
 	const mediumPlanMonthly = getPlans()[ PLAN_WOOEXPRESS_MEDIUM_MONTHLY ];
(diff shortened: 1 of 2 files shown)


client/my-sites/plans/ecommerce-trial/wooexpress-plans/style.scss
```diff
@@ -16,6 +16,12 @@
 			max-width: 100%;
 			width: 100%;
 		}
+
+		@media (max-width: 375px) {
+			.segmented-control__text {
+				font-size: 0.75rem !important;
```
sixhours: Can we use `$font-body-extra-small` here instead of a hard-coded font value?

Bonus points if you can make the selector more specific to avoid `!important` IMO :D

daledupreez:
I also tested the new code with my language set to German, and while it is an improvement, we still hit trouble with long strings.

Rather than make the text smaller, might it make sense to move the "Save x%" outside the selector? We could still control what elements get displayed via CSS, but I think that's more likely to be robust. @FlorianBru, do you have any objections to us taking that approach?

free-trial-plans-narrow-german

markbiek:
I've pushed an update that tweaks the markup so we can target the `(Save nn%)` copy directly. I'm currently hiding the copy on extra-small mobile screens but we can make adjustments depending on design feedback.

**Extra-small mobile**
image

**Other sizes**
image

client/my-sites/plans/ecommerce-trial/wooexpress-plans/style.scss
```diff
@@ -16,6 +16,12 @@
 			max-width: 100%;
 			width: 100%;
 		}
+
+		@media (max-width: 375px) {
```
daledupreez: Is this the right breakpoint? I feels like it might need to be wider to handle German and/or other translations.

client/my-sites/plans/ecommerce-trial/wooexpress-plans/index.tsx
```diff
@@ -71,9 +71,12 @@ export function WooExpressPlans( props: WooExpressPlansProps ) {
...
 					<span>
-						{ translate( 'Pay Annually (Save %(percentageSavings)s%%)', {
-							args: { percentageSavings },
-						} ) }
+						{ translate( 'Pay Annually' ) }
+						<span className="wooexpress-plans__interval-savings">
+							{ translate( ' (Save %(percentageSavings)s%%)', {
+								args: { percentageSavings },
+							} ) }
+						</span>
```
daledupreez: To keep the full context in a single translation, I think it would be better to interpolate the wrapper span.

I also think we need to wrap this in a `hasTranslation()` check given that we likely have a number of non-English users that we don't want to break this for.
```suggestion
						{ isEnglishLocale || i18n.hasTranslation( 'Pay Annually {{span}}(Save %(percentageSavings)s%%){{/span}}' )
							translate( 'Pay Annually {{span}}(Save %(percentageSavings)s%%){{/span}}', {
								args: { percentageSavings },
								components: { span: <span className="wooexpress-plans__interval-savings" /> },
							}
							: translate( 'Pay Annually (Save %(percentageSavings)s%%)', {
								args: { percentageSavings },
							} )
						}
```

client/my-sites/plans/ecommerce-trial/wooexpress-plans/style.scss
markbiek: Hmm, that's a good point. The original issue was only happening on breakpoints much smaller than `$break-mobile` but that example was also English.

To accommodate languages like German, it's probably better to move this selector into the one above it that's targeting `$break-mobile` (which, I think, is 480px).

sixhours (approved):
Tested and this is working well for me!

PR #75682, opened 2023-04-13
Videopress landing remove plans grid

thedebian:
Fixes Automattic/greenhouse#1667

## Proposed Changes

This PR removes the choose a plan step and enforces the monthly premium plan in the cart (as it could easily be upgraded to a yearly plan on checkout).

## Testing Instructions

* Apply PR
* `yarn start`
* Go to `/setup/videopress`
* Follow the steps
* ✅ You should get redirected to the checkout page with a premium plan, no plan page should be displayed during the flow

PR #67539, opened 2022-09-08
VideoPress Onboarding v2

pgk:
#### Proposed Changes

* Creates a new onboarding flow for VideoPress.
* See also pxWta-1z5-p2
* This is a culmination of many different PRs that were merged to this branch.

Screen Shot 2022-10-12 at 5 30 42 PM

#### Testing Instructions

* Start the flow at http://calypso.localhost:3000/setup/intro?flow=videopress
* Walk through the onboarding, choosing a premium or better plan.
* You should end up with a Videomaker site.

Related to #

PR #76157, opened 2023-04-24
VideoPress Onboarding: Do not add cart item if missing site, cart key, or product

sirbrillig:
Originally added in https://github.com/Automattic/wp-calypso/pull/67539 and updated in https://github.com/Automattic/wp-calypso/pull/75682, the VideoPress onboarding code tries to add invalid products to the cart (specifically, cart items without a product slug). This happens if the newly created site does not exist, no cart key can be determined, or if the product itself cannot be found.

An example of this error can be found here: https://a8c.sentry.io/issues/4122116103/?project=6313676

Interestingly, TypeScript is set up to guard against all of these situations, but in both the above-mentioned PRs the code uses `as` and `any` to ignore these guards, allowing the bugs into the codebase and causing fatal errors.

## Proposed Changes

In this PR, we remove all unnecessary uses of `as` and `any`, instead bailing out of the function if `getNewSite()` fails, if the returned site data is missing its `site_slug` property, or if the `getPlanProduct()` fails to find a product.

**NOTE:** I chose to bail out silently as this was already done by the `try/catch` around `createVideoPressSite()` in the same function, but if you can think of a better UX, please suggest it. I'm not familiar with this area of the codebase so I'm not sure what the user should see when this happens.

## Testing Instructions

I'm not certain, but here's the test instructions for https://github.com/Automattic/wp-calypso/pull/75682

- Go to `/setup/videopress`
- Follow the steps
- You should get redirected to the checkout page with a premium plan, no plan page should be displayed during the flow.

### DIFF
--- a/client/landing/stepper/declarative-flow/videopress.ts
+++ b/client/landing/stepper/declarative-flow/videopress.ts
@@ -192,11 +192,11 @@ const videopress: Flow = {
 					.then( () => {
 						setProgress( 1.0 );
 						const redirectTo = encodeURIComponent(
-							`/setup/videopress/launchpad?siteSlug=${ newSite?.site_slug }&siteId=${ newSite?.blogid }`
+							`/setup/videopress/launchpad?siteSlug=${ newSite.site_slug }&siteId=${ newSite.blogid }`
 						);
 
 						window.location.replace(
-							`/checkout/${ newSite?.site_slug }?signup=1&redirect_to=${ redirectTo }`
+							`/checkout/${ newSite.site_slug }?signup=1&redirect_to=${ redirectTo }`
 						);
 					} );
 			} );
--- a/client/landing/stepper/declarative-flow/videopress.ts
+++ b/client/landing/stepper/declarative-flow/videopress.ts
@@ -18,6 +18,7 @@ import SiteOptions from './internals/steps-repository/site-options';
 import VideomakerSetup from './internals/steps-repository/videomaker-setup';
 import type { Flow, ProvidedDependencies } from './internals/types';
 import type { OnboardSelect, UserSelect } from '@automattic/data-stores';
+import type { MinimalRequestCartProduct } from '@automattic/shopping-cart';
 
 const videopress: Flow = {
 	name: VIDEOPRESS_FLOW,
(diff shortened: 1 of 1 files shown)


thedebian (approved):
It seems to work properly.
I'm wondering under which conditions getNewSite might fail in the first place ?
Also maybe we want to redirect the user to site options or something instead of just returning ?

sirbrillig:
Thanks for checking @thedebian! I'm not sure what could cause this to happen. My only guess is that perhaps creating the site is failing in a way that is not caught by the `catch`? In any case, I'll merge this for now to stop the errors from coming in but feel free to make additional changes here to improve the UX or add logging.

worldomonation:
FYI @sirbrillig I suspect this change has caused a spike in recent intermittent failure counts for the VideoPress spec. See below:

image

sirbrillig:
@worldomonation 🤔 It may be that it was the tests that were causing the fatal errors when they tried to add invalid items to the cart, and since this PR prevents the fatal errors, the tests are now failing? I have a feeling that there might be a race condition of some sort in this part of signup but I don't know anything about this area of the code so we should check with the team that maintains it, which I believe is @Automattic/greenhouse.

roundhill:
Strange one, we'll investigate 👀

worldomonation:
> since this PR prevents the fatal errors, the tests are now failing

That's definitely a possibility.
It appears folks from #cylon, @edanzer and @jeyip have run into this same issue when they ran the specs on CI and locally while working on a feature: p1682547346789469-slack-CHN6J22MP

thedebian:
I saw some stuff have been moved from the "processing" step to the submit of "chooseADomain" step, moving it back where it was, but calling "stepValidateSiteTitle" before (or another one, maybe adding more checks) should do the trick.
This way, people would get redirected to the options step if they land for a reason or another on the processing step without doing the previous steps (at the very least, having a site slug and such)

sirbrillig:
I'm not sure if it's the same cause as is being discussed here but I've seen the mentioned videopress e2e test fail in TeamCity and I can see from the screenshot that checkout appears to be stuck while loading, but it's hard to say what this means. I have a feeling it may be a slow shopping-cart endpoint (although why that would only appear for this particular e2e is beyond me) but it's impossible to tell without debug logs. However, https://github.com/Automattic/wp-calypso/pull/76377 may help us track this down by adding additional visible information to checkout's loading process when it is slow.

worldomonation:
> videopress e2e test fail in TeamCity and I can see from the screenshot that checkout appears to be stuck while loading, 

Sorry, a few days late in responding (had AFK on Friday) - the failure mode I was seeing last week is that the Cart screen was showing "no items in cart" for users that stepped through the VideoPress flow, as shown below:
image

However, incidence rate appears to have dropped after 04/25. See comparison below:
image
image

However, even at 3 occurrences over the last ~100 builds is higher than the incidence rate prior to 04/25.

----

Looks like I was mistaken in the first comment, thinking that #cylon was responsible for VideoPress. Looks like it's actually #greenhouse, so I'll be reaching out to them with details of [this comment](https://github.com/Automattic/wp-calypso/pull/76157#issuecomment-1523464675). Thanks!

Issue #76110, opened 2023-04-23
Reader: when x-post label runs to two lines, text appears too close to the avatars


bluefuton:
### Quick summary

In Reader, x-posts (cross posts) are shown like this:

<img width="619" alt="Screenshot 2023-04-24 at 9 43 41 AM" src="https://user-images.githubusercontent.com/17325/233868086-a2720d0d-7bb8-42b3-936b-413a23232029.png">

As shown above, if the label runs to two lines, the spacing between the avatars and the text is reduced too much, and the text is out of alignment with other x-posts.

The label can run to two lines when the author has x-posted to multiple blogs.

### Steps to reproduce

1. Start at https://wordpress.com/read
2. Find an x-post in your stream.
3. It could take a while for a x-post like this to happen naturally, so try lengthening the text in `reader__x-post-label` using your browser's dev tools.

### What you expected to happen

The spacing between the avatars and the text should be consistent.

### What actually happened

If there are two lines of text in the x-post label, the avatars and text are too close together.

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

### Logs or notes

Browser: Firefox v112 Mac

cc @davemart-in

bluefuton:
This was fixed in #76162. 🚀

PR #76162, opened 2023-04-24
Fix small issue with misalignment of x-post layout in Reader

davemart-in:
## Description

@bluefuton highlighted the following issue: https://github.com/Automattic/wp-calypso/issues/76110 where x-post formatting was inconsistent. I took a pass at fixing this, though I had to use http://calypso.localhost:3000/read/a8c to test it because I didn't have any x-posts in my normal following feed.

### Before

Screenshot 2023-04-26 at 9 50 52 AM

### After

Screenshot 2023-04-26 at 9 50 59 AM

## Testing instructions

- Apply this PR locally
- Head to http://calypso.localhost:3000/read/ or http://calypso.localhost:3000/read/a8c if you're an Automattician and look for a x-post which spans more than 2 lines.

### DIFF
--- a/client/blocks/reader-avatar/style.scss
+++ b/client/blocks/reader-avatar/style.scss
@@ -33,8 +33,6 @@
 		&.has-gravatar.has-site-icon {
 			max-height: 40px;
 			max-width: 40px;
-			min-height: 40px;
-			min-width: 40px;
 		}
 
 		&.has-site-and-author-icon {
--- a/client/blocks/reader-avatar/style.scss
+++ b/client/blocks/reader-avatar/style.scss
@@ -30,10 +30,12 @@
 			min-width: 32px;
 			min-height: 32px;
 		}
-		
+
 		&.has-gravatar.has-site-icon {
 			max-height: 40px;
 			max-width: 40px;
+			min-height: 40px;
+			min-width: 40px;
 		}
 
 		&.has-site-and-author-icon {
--- a/client/blocks/reader-avatar/style.scss
+++ b/client/blocks/reader-avatar/style.scss
@@ -30,6 +30,7 @@
 			min-width: 32px;
 			min-height: 32px;
 		}
+		
 		&.has-gravatar.has-site-icon {
 			max-height: 40px;
 			max-width: 40px;
--- a/client/blocks/reader-avatar/style.scss
+++ b/client/blocks/reader-avatar/style.scss
@@ -30,6 +30,12 @@
 			min-width: 32px;
 			min-height: 32px;
 		}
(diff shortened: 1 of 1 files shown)


client/blocks/reader-avatar/style.scss
```diff
@@ -30,6 +30,12 @@
 			min-width: 32px;
 			min-height: 32px;
 		}
+		&.has-gravatar.has-site-icon {
+			max-height: 40px;
+			max-width: 40px;
+			min-height: 40px;
+			min-width: 40px;
```
bluefuton, line 38:
```suggestion
			max-height: 40px;
			max-width: 40px;
```

Looks like these rules are duplicated :)

client/blocks/reader-avatar/style.scss
```diff
@@ -30,6 +30,10 @@
 			min-width: 32px;
 			min-height: 32px;
 		}
+		&.has-gravatar.has-site-icon {
```
bluefuton, line 34:
```suggestion
		
		&.has-gravatar.has-site-icon {
```

bluefuton (approved):
Tests out well for me 🚢 Thanks @davemart-in!

Screenshot 2023-04-26 at 9 38 57 AM

davemart-in:
@bluefuton I'm still seeing the issue after you removed the min-width/min-height. Here's what I'm seeing:

CleanShot 2023-04-26 at 09 14 57@2x

And here's what I see if I add it back in:

CleanShot 2023-04-26 at 09 17 13@2x

bluefuton:
My apologies @davemart-in - I thought I saw duplicate rules but my brain was playing tricks.

Is there any reason not to explicitly set the `width` and `height` to 40px if we're setting both `min-` and `max-` to the same value?

davemart-in:
> Is there any reason not to explicitly set the width and height to 40px if we're setting both min- and max- to the same value?

I don't know that we want to drop the min-width on `.reader-avatar.is-compact.has-gravatar` since this style is used in other places throughout the reader (where 32px should correctly be the min-width).

Setting a width of 40px doesn't seem to work:

CleanShot 2023-05-01 at 09 18 29@2x

Honestly, it's kind of odd. Not sure I've seen this before. I suspect this is somehow partially due to the parent element being set to `display: flex`. The min-width is somehow being interpreted as "we only need to reserve this amount of space".

davemart-in:
Going to go ahead and merge this, because it fixes the issue. Happy to refactor if another alternative approach makes more sense.

bluefuton:
Thanks for fixing this one @davemart-in! My A8C feed looks all nicely aligned now ❤️

PR #75073, opened 2023-03-29
Plans 2023: Fix calculation errors and cleanup price descriptions

chriskmnds:
Closes Automattic/martech#1648

## Proposed Changes

Fixes a calculation error in price descriptions and moves billing timeframe/savings properties to the respective component. These seem sensitive parts and we are moving around price computations. hopefully, nothing was missed.

The PR is based on https://github.com/Automattic/wp-calypso/pull/74960, which introduces the `usePlanPrices` hook that's used here.

### TODO

- [x] Some unit/UI tests while touching these parts

### Media

**Before**

Screenshot 2023-03-29 at 3 42 56 PM

**After**

Screenshot 2023-03-29 at 3 42 06 PM

## Testing Instructions

* Go to `/plans/[TERM]][SITE]` and confirm price descriptions are produced as expected.
* Switch to INR currency and confirm the price shows up correctly.
* Confirm prices, in general, are displayed correctly (this should also serve as additional checks for the changes introduced by https://github.com/Automattic/wp-calypso/pull/74960)

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76086, opened 2023-04-21
Plans 2023: data store slugs and more

chriskmnds:
Related to #76024 

## Proposed Changes

Continuation of #76024 to introduce the plans data store into the 2023 plan grids. 
- Introduces more slugs to cover the rest of the plans
- `billingTerm` included in the plan product structure (used in follow up PR)
- a selector to fetch plan product given a respective slug (used in follow up PR)

### Author Notes

The draft P2 post ( dvytD-p2  ) contains details about this work and motivation. For the time being, we move enough content into the plans data store to surface it for usage in the plans 2023 grids (continuing on how it's already set up to work).

## Testing Instructions

There are no UI integrations. One way to test:
- go to `/plans/[SITE]`
- open Redux dev tools and inspect the latest image for `automattic/onboard/plans`
- under `State`, confirm all the variant `planProducts` for different periods are visible
- confirm `billingTerm` is included in that structure

Screenshot 2023-04-26 at 12 09 42 PM

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #74797, opened 2023-03-22
Plans 2023: Fix 3Y plan price descriptions

chriskmnds:
Fixes https://github.com/Automattic/wp-calypso/issues/74638

## Proposed Changes

- Updates the 3Y plan price descriptions to show the correct text.
- Adds `currencyCode` to the data store product (and uses that in the respective component, in place of the Calypso selector)

### Media

**Before**

227237260-185a84d0-4630-4d6d-a32a-f666c368eca1

**After**

227233544-91efa92e-c0ce-422d-86e8-a641845836a4

### TODO

- [x] Ensure the selectors/resolvers are called at the right place to avoid any flicker with the store data coming in after the grid is rendered

## Testing Instructions

* Go to `/plans/3yearly/[SITE]` and ensure the price descriptions show the prices, as in the media. Do the same for `/plans/yearly` and `/plans/2yearly`.
* Go to `/start/plans` and ensure things render as before.
* In all cases above, there should be no flicker of price descriptions changing. The grids should load after both selectors have resolved the plans endpoint.

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76205, opened 2023-04-25
WooExpress: Fixes the timeframe description on the monthly view

### LINKED ISSUE #75299: WooExpress: Yearly discount information is not displayed below the pricing
The discount information(`SAVE x% BY PAYING ANNUALLY`) is not shown below the pricing described in the design.

### Calypso
Screen Shot 2023-04-04 at 17 13 00

lD0gTB4IqGsmSmsnjZNaP6-fi-2740-73590

### LINKED ISSUE #75483: Woo Express Essential: Monthly plan note differs from Performance
CfT: pdtkmj-1mM-p2#comment-2505

On the monthy plan toggle, Essential and Performance have different notes under them. The annual plan notes are the same.

Essential: "per month, billed monthy"
Performance: "Save 36% by paying annually".

image

valterlorran:
Closes #75483 #75299

## Proposed Changes

* Fixes the timeframe description for WooExpress plans following the design proposed here: peapX7-22h-p2. We were showing different messages for the Essential and the Performance plan(See image below). This problem was introduced on #75073.
* We can't use the `usePerMonthDescription` hook found [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx#L24-L89) because the way to fetch the yearly plan variant does not work for WooExpress plans.
* Fixes #75299 adding the promotion to the annual plan below the pricing when the user is viewing the monthly plan

#### Before
Screen Shot 2023-04-25 at 08 56 27

#### After
Screen Shot 2023-04-25 at 08 53 27

## Testing Instructions

* Navigate to `/setup/wooexpress` to create a new Woo Trial site.
* Navigate to `/plans/monthly/{YOUR-SITE-SLUG}
* You should `per month`, followed by `Save X% by paying annually` below the price


aneeshd16 (commented):
Hi! I am not able to reproduce the original issue in prod. Are there any additional steps required to reproduce this?

image

> We can't use the usePerMonthDescription hook found [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx#L24-L89) because the way to fetch the yearly plan variant does not work for WooExpress plans.

Can you please explain why it doesn't work? Those functions are generic, so if this doesn't work, it could be a sign of a bug in the plans config. It would be great to avoid duplication as much as possible. 🙂

valterlorran:
@aneeshd16 I'm trying to understand why you are not able to reproduce, but I think is something related to the pricing. I started a thread that you can follow here: p1682443098163709-slack-C0Q664T29

Regarding the `usePerMonthDescription` hook, one of the problems is that the `getPlanSlugForTermVariant` function doesn't work for WooExpress plans, because the plans related to WooExpress have the type defined as `TYPE_ECOMMERCE`, which wouldn't output a reliable result for the `filter` and the `find` here:
https://github.com/Automattic/wp-calypso/blob/fe2c6255756a966b402b02f29b18d0b22d227edb/packages/calypso-products/src/get-plan-term-variant.ts#L20-L21

Besides that, the designers also suggested a specific billing description for the monthly variation: `per month` instead of `per month, billed monthly`. I added to the plan properties, that we can access as `billingTimeFrame` prop.
https://github.com/Automattic/wp-calypso/blob/bcb48e1ec26e6bf5759866d8c667691db09f4bde/packages/calypso-products/src/plans-list.tsx#L2477

The designers also required a specific design for the promotion description( Save x% paying annually), which would add too much WooExpress-specific logic to the hook.

valterlorran:
@aneeshd16 We fixed the pricing problem here: p1682513229167499/1682443098.163709-slack-C0Q664T29 
However, you won't see the same thing as I'm seeing, because as I mentioned that function doesn't work properly for the WooExpress plan because it returns the eCommerce plan, but we need to compare it with the monthly price of the Essential plan with the yearly price of the Essential plan.
https://github.com/Automattic/wp-calypso/blob/83ee170393a4c709fb279d8a6e9f250903a44ca6/client/my-sites/plan-features-2023-grid/billing-timeframe.tsx#L53
It mistakenly falls in this condition for the Essential plan when we have the INR defined as our currency.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -88,12 +97,83 @@ function usePerMonthDescription( {
...
+	);
+};
+
+/**
+ * Retrieves the monthly and yearly plan slugs for a given WooExpress plan name.
+ *
+ * @param planName - The name of the WooExpress plan to retrieve slugs for.
+ * @returns An object containing the monthly and yearly plan slugs, or an empty object if the plan name is invalid.
+ */
+function getWooExpressPlanSlugs( planName: string ) {
```
aneeshd16: Added a note about this above.

client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -1079,3 +1079,11 @@ body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup_
 		color: #55347d;
 	}
 }
+
+.plan-features-2023-grid__discount-promotion {
```
aneeshd16: Would it be possible to move this CSS to a styled component? We're hoping to move the grid over to styled-components so that it looks consistent across usages.

aneeshd16 (commented):
> Regarding the usePerMonthDescription hook, one of the problems is that the getPlanSlugForTermVariant function doesn't work for WooExpress plans, because the plans related to WooExpress have the type defined as TYPE_ECOMMERCE, which wouldn't output a reliable result for the filter and the find here:

This could also output an incorrect result if the function is called for the WPCOM eCommerce plan. It's probably not breaking right now only because the Woo Express plans are defined after the eCommerce plans in [plans-list](https://github.com/Automattic/wp-calypso/blob/c098b309ec7d07989e51b38f403d15a4c458179a/packages/calypso-products/src/plans-list.tsx).

Ideally, each Woo Express plan should have its own `type`: `TYPE_WOOEXPRESS_SMALL`, `TYPE_WOOEXPRESS_MEDIUM` and `TYPE_WOOEXPRESS_PLUS`. So all other plan-related functions would also work as expected. We can also replace the `getWooExpressPlanSlugs` with `getPlanSlugForTermVariant` then.

Tagging @Automattic/martech-decepticons for additional thoughts.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
valterlorran: Do you think we could tackle that in a follow-up PR/issue? I'm not sure yet how adding those types can impact the Woo Express flows.

chriskmnds:
> > Regarding the usePerMonthDescription hook, one of the problems is that the getPlanSlugForTermVariant function doesn't work for WooExpress plans, because the plans related to WooExpress have the type defined as TYPE_ECOMMERCE, which wouldn't output a reliable result for the filter and the find here:
> 
> This could also output an incorrect result if the function is called for the WPCOM eCommerce plan. It's probably not breaking right now only because the Woo Express plans are defined after the eCommerce plans in [plans-list](https://github.com/Automattic/wp-calypso/blob/c098b309ec7d07989e51b38f403d15a4c458179a/packages/calypso-products/src/plans-list.tsx).
> 
> Ideally, each Woo Express plan should have its own `type`: `TYPE_WOOEXPRESS_SMALL`, `TYPE_WOOEXPRESS_MEDIUM` and `TYPE_WOOEXPRESS_PLUS`. So all other plan-related functions would also work as expected. We can also replace the `getWooExpressPlanSlugs` with `getPlanSlugForTermVariant` then.
> 
> Tagging @Automattic/martech-decepticons for additional thoughts.

Just a run-by comment from me, for now. Agree with this sentiment. Plans should have their slugs defined concretely, and `getPlanSlugForTermVariant` to be trusted universally - it sounds like the reasonable approach.

Happy to take a closer look, as there may be overlap or insight from work on migrating plans state to data stores.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
valterlorran: I started working on a draft PR here: https://github.com/Automattic/wp-calypso/pull/76535

markbiek (approved):
Working as expected!

valterlorran:
Hey @aneeshd16 & @chriskmnds.
Are you all ok with merging this as-is and addressing the types in the follow-up?  I'm working on it here: https://github.com/Automattic/wp-calypso/pull/76535
And I create a p2 post to get more inputs about the impacts of this change here: paYKcK-2U9-p2

aneeshd16:
> Hey @aneeshd16 & @chriskmnds.
Are you all ok with merging this as-is and addressing the types in the follow-up? I'm working on it here: https://github.com/Automattic/wp-calypso/pull/76535
And I create a p2 post to get more inputs about the impacts of this change here: paYKcK-2U9-p2

Thanks @valterlorran! I think the draft PR is definitely a step in the right direction! This would remove the need for a `getWooExpressPlanSlugs` function. 

We are currently in the process of decoupling the pricing grid from Calypso. One solution is a prop-based approach like https://github.com/Automattic/wp-calypso/pull/76534. So maybe we can have a prop in the pricing grid called `BillingTimeframeOverrides` and you would be able to pass the `WooExpressMonthlyPromotion` component as a prop to the pricing grid in https://github.com/Automattic/wp-calypso/blob/4910db94dde21aced9111956d87aae3389df408d/client/my-sites/plans/ecommerce-trial/wooexpress-plans/index.tsx#L142

@chriskmnds what do you think?

chriskmnds:
> @chriskmnds what do you think?

Sorry I missed this. Let me take another look tomorrow and write back if not late. But generally, the "optional props" based approach/plan that @aneeshd16 described is our current attempt at uncoupling the grid from Calypso state and dependencies.

Concerning the above, we should be considerate of prop drilling that may result, as things can go a couple of layers deep. There are ways to remedy - I've seen making use of a shared store, or defining a grid context where these configurations/props can live. Not something for this PR or to change the direction (of feeding the grid with more props).

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -88,12 +98,91 @@ function usePerMonthDescription( {
...
+) => {
+	const translate = useTranslate();
+	const { billingTimeframe, yearlyPlanSlug, monthlyPlanSlug } = props;
+
+	const annualPlan = getPlans()[ yearlyPlanSlug ];
+	const monthlyPlan = getPlans()[ monthlyPlanSlug ];
+	const planPrices = useSelector( ( state ) => ( {
+		annualPlanMonthlyPrice: getPlanRawPrice( state, annualPlan.getProductId(), true ) || 0,
+		monthlyPlanPrice: getPlanRawPrice( state, monthlyPlan.getProductId() ) || 0,
+	} ) );
```
chriskmnds: We could have used the [getPlanPrices](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/plans/selectors/get-plan-prices.ts) selector for these instead. `rawPrice` from there would correspond to `getPlanRawPrice`. It's another means to unify the pricing selectors, at least until we have a better data layer.

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -88,12 +98,91 @@ function usePerMonthDescription( {
...
+				args: {
+					percentageSavings,
+				},
+				comment: 'Translators: percentageSavings is the percentage saved by paying annually',
+		  } )
+		: null;
+	return (
+		<div>
+			<div>{ billingTimeframe }</div>
+			<DiscountPromotion>{ discountDescription }</DiscountPromotion>
```
chriskmnds: The `DiscountPromotion` styles look like the only thing that's new here. The rest are already computed for monthly plans [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx#L45-L63), right? 

Is there anything outside of the missing plan slugs that require repeating logic for computing the monthly-yearly discount? The whole lot should have been one or two conditions to existing logic as far as I can tell.

chriskmnds (changes requested):
Hi! Great work. I left some comments below to better orient myself with the changes. If my understanding is correct, and it's a matter of introducing the plan slugs to avoid all the logic repetition, then maybe we should go in that direction.

Thanks for seeking/waiting for our input here, despite having approval already. :-)

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
valterlorran: Hey @chriskmnds, we'll use the `usePerMonthDescription` to get the description once #76535 is merged. My intention is to have something similar to: 
```ts
const WooExpressMonthlyPromotion: FunctionComponent< WooExpressMonthlyPromotionProps > = (
	props
) => {
	const { billingTimeframe, discountDescription } = props;
	return (
		<div>
			<div>{ billingTimeframe }</div>
			<DiscountPromotion>{ discountDescription }</DiscountPromotion>
		</div>
	);
};
```

chriskmnds: Cool. I guess let's rebase this PR over https://github.com/Automattic/wp-calypso/pull/76535?

There's also a bit of refactoring happening in https://github.com/Automattic/wp-calypso/pull/74797 with this file, which reinvigorates a data store here. It's been approved, but mainly waiting for the WooExpress slugs to settle in, as they are needed in its [base branch](https://github.com/Automattic/wp-calypso/pull/76086).

valterlorran: Hey! I rebased as you suggested and updated the code accordingly. I checked the changes on the file, and I believe it won't affect my changes in this PR.

markbiek (approved):
LGTM!

PR #76216, opened 2023-04-25
LiB Domain: Async flow and fixed login return url

escapemanuele:
## What
The LiB Domain flow was working, but it had some issues.
This PR:
- Makes the flow load the steps asynchronously
- Encode the return url for the login. It was not working with two parameters ( `domain` and `provider` )
- Adds `domain` to the track event

## Testing

1. Live Link
2. Visit the flow: [Live Link]/setup/link-in-bio-domain?domain=[your test domain]&provider=[your test provider]
3. Try the flow in incognito and go through login
4. Check that the flow works. In particular, make sure that the `provider` parameter is still there after redirecting from login


alshakero (approved):
LGTM

PR #76218, opened 2023-04-25
add/Jetpack Zendesk Chat to partner dashboard

TheSteveK:
Related to [#](p7pQDF-8iG-p2)

## Proposed Changes

* This PR adds the Zendesk chat widget to cloud.jetpack.com/partner-portal
* The component is called from the Sidebar of the partner portal because that site has no footer, and adding one would have caused a lot of complications and delay
* Because this chat widget uses a different API key, a few changes had to be made to the existing chat widget code.
* A new property was created called keyType which expects either jpGeneral or jpAgency
* Based on which type is passed, the widget will fetch the corresponding API key
* New keys were added for the agency to the environment files

## Testing Instructions

* Use the Jetpack Cloud live link, or build this PR locally and head over to /partner-portal (note you have to be a Jetpack partner to access the portal). The process should be documented in the FG.
* Chat will behave and look as it does on the rest of Jetpack Cloud (you can also test /pricing to make sure it shows chat when appropriate)
* To verify different API keys are being used, you can use the Network tab and search for requests going to zendesk.com you should see a different API key for the agency dashboard and for general Jetpack Cloud

## Pre-merge Checklist

### DIFF
--- a/client/components/jetpack/jetpack-presales-chat-widget/index.tsx
+++ b/client/components/jetpack/jetpack-presales-chat-widget/index.tsx
@@ -13,6 +13,12 @@ type PresalesChatResponse = {
 	is_available: boolean;
 };
 
+export type KeyType = 'jpAgency' | 'jpGeneral';
+
+export interface ZendeskJetpackChatProps {
+	keyType: KeyType;
+}
+
 //the API is rate limited if we hit the limit we'll back off and retry
 async function fetchWithRetry(
 	url: string,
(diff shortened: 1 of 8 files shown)


client/components/jetpack/jetpack-presales-chat-widget/index.tsx
```diff
@@ -38,10 +44,16 @@ async function fetchWithRetry(
...
 }
 
-export const ZendeskJetpackChat: React.VFC = () => {
+export const ZendeskJetpackChat: React.VFC< { keyType: KeyType } > = ( { keyType } ) => {
 	const [ error, setError ] = useState( false );
 	const { data: isStaffed } = usePresalesAvailabilityQuery();
-	const zendeskChatKey = config( 'zendesk_presales_chat_key' ) as keyof ConfigData;
+	const zendeskChatKey = useMemo( () => {
+		return config(
+			keyType === 'jpAgency'
```
jkguidaven, line 52: Not a blocker, but maybe we can make `keyType` a string instead of union type here and pass the key name and avoid the need to map. This way, when we add a new key later, we do not need to modify this line of code.

But I understand why you did it here to prevent pulling config data, not necessarily a chat key.

jkguidaven (approved):
Tested both the partner portal and pricing page, and I can confirm the live chat appears. 

I have a minor comment but LGTM 👍

PR #76220, opened 2023-04-25
Plans: Show current plan price correctly

aneeshd16:
Fixes Automattic/martech#1695

## Proposed Changes

* If a currency-specific discount is active, and if a plan has been upgraded using pro-rated credits, the price shown on the /plans page is currently the discounted price. This is incorrect and gives the impression that the user was charged for more than the price shown on this page. This PR shows the current plan price without any discounts.
* Added a new hook, `usePlanPricesDisplay`, in the `client/my-sites/plan-features-2023-grid/hooks` directory. This returns the original and discounted prices to be displayed for a particular plan slug. This works according to the following rules:
  * If the passed plan slug is the current plan slug, it returns the "site plan raw price".
  * If the passed plan slug is not available for purchase,  it returns the raw price of the plan.
  * If the `withoutProRatedCredits` arg is false (false by default), the discounted price is either the "plan discounted raw price" (discounted price after applying pro-rated credits) or the "discounted raw price" (discounted price if there is a first-year currency specific discount).
  * If the `withoutProRatedCredits` arg is true (false by default), the discounted price is the "discounted raw price" (discounted price if there is a first-year currency-specific discount).
* The [`header-price`](https://github.com/Automattic/wp-calypso/blob/98c3ad52a21fa8d3fdea2e6711249d2f06621438/client/my-sites/plan-features-2023-grid/components/header-price.tsx) and [`billing-timeframe`](https://github.com/Automattic/wp-calypso/blob/98c3ad52a21fa8d3fdea2e6711249d2f06621438/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx) components should not be concerned with which prices to display and should be focused on only how to display the original and discounted prices.

Screenshots: (Current plan is the Business plan)

| Production  | This branch |
|--------|--------|
|image|image| 

## Testing Instructions

* Follow the test instructions in Automattic/martech#1695.
* The price should be displayed correctly.
* Also confirm that the prices for the plans lower than the current plan are displayed without any discount.

#### Testing for regressions:
Confirm that the prices are displayed correctly (same as production) in the following flows:
* `/start/plans`
* `/plans/<slug>` where `<slug>` is the site slug of a site on the free plan.
* `/setup/newsletter/plans`.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
+++ b/client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
@@ -11,35 +11,48 @@ import {
 import { formatCurrency } from '@automattic/format-currency';
 import { useIsEnglishLocale } from '@automattic/i18n-utils';
 import styled from '@emotion/styled';
-import i18n, { localize, TranslateResult, useTranslate } from 'i18n-calypso';
+import i18n, { TranslateResult, useTranslate } from 'i18n-calypso';
 import { FunctionComponent } from 'react';
 import { useSelector } from 'react-redux';
-import usePlanPrices from 'calypso/my-sites/plans/hooks/use-plan-prices';
 import { getCurrentUserCurrencyCode } from 'calypso/state/currency-code/selectors';
+import { usePlanPricesDisplay } from '../hooks/use-plan-prices-display';
 
 interface Props {
 	planName: string;
 	billingTimeframe: TranslateResult;
 	billingPeriod: number | null | undefined;
 	isMonthlyPlan: boolean;
+	currentSitePlanSlug?: string;
+	siteId?: number | null;
 }
 
 function usePerMonthDescription( {
 	isMonthlyPlan,
 	planName,
 	billingPeriod,
+	currentSitePlanSlug,
+	siteId,
 }: Omit< Props, 'billingTimeframe' > ) {
 	const translate = useTranslate();
 	const currencyCode = useSelector( getCurrentUserCurrencyCode );
 	const isEnglishLocale = useIsEnglishLocale();
-	const planPrices = usePlanPrices( {
+	const planPrices = usePlanPricesDisplay( {
 		planSlug: planName as PlanSlug,
 		returnMonthly: isMonthlyPlan,
+		currentSitePlanSlug,
+		siteId,
 	} );
-	const planYearlyVariantPricesPerMonth = usePlanPrices( {
+
+	// We want `planYearlyVariantPricesPerMonth` to be the raw price the user
+	// would pay if they choose an annual plan instead of the monthly one. So pro-rated
+	// (or site-plan specific) credits should not be taken into account.
+	const planYearlyVariantPricesPerMonth = usePlanPricesDisplay( {
 		planSlug:
 			getPlanSlugForTermVariant( planName as PlanSlug, TERM_ANNUALLY ) ?? ( '' as PlanSlug ),
 		returnMonthly: true,
+		currentSitePlanSlug,
+		siteId,
+		withoutProRatedCredits: true,
 	} );
 
 	if ( isWpComFreePlan( planName ) || isWpcomEnterpriseGridPlan( planName ) ) {
@@ -47,19 +60,16 @@ function usePerMonthDescription( {
 	}
 
 	if ( isMonthlyPlan ) {
-		// We want `yearlyVariantMaybeDiscountedPricePerMonth` to be the raw price the user
-		// would pay if they choose an annual plan instead of the monthly one. So pro-rated
-		// (or site-plan specific) credits should not be taken into account.
 		const yearlyVariantMaybeDiscountedPricePerMonth =
-			planYearlyVariantPricesPerMonth.discountedRawPrice ||
-			planYearlyVariantPricesPerMonth.rawPrice;
+			planYearlyVariantPricesPerMonth.discountedPrice ||
+			planYearlyVariantPricesPerMonth.originalPrice;
 
-		if ( yearlyVariantMaybeDiscountedPricePerMonth < planPrices.rawPrice ) {
+		if ( yearlyVariantMaybeDiscountedPricePerMonth < planPrices.originalPrice ) {
 			return translate( `Save %(discountRate)s%% by paying annually`, {
 				args: {
 					discountRate: Math.floor(
-						( 100 * ( planPrices.rawPrice - yearlyVariantMaybeDiscountedPricePerMonth ) ) /
-							planPrices.rawPrice
+						( 100 * ( planPrices.originalPrice - yearlyVariantMaybeDiscountedPricePerMonth ) ) /
+							planPrices.originalPrice
 					),
 				},
 			} );
@@ -67,14 +77,14 @@ function usePerMonthDescription( {
 	}
 
 	if ( ! isMonthlyPlan ) {
-		const discountedPrice = planPrices.planDiscountedRawPrice || planPrices.discountedRawPrice;
+		const discountedPrice = planPrices.discountedPrice;
 		const fullTermDiscountedPriceText =
 			currencyCode && discountedPrice
 				? formatCurrency( discountedPrice, currencyCode, { stripZeros: true } )
 				: null;
 		const rawPrice =
-			currencyCode && planPrices.rawPrice
-				? formatCurrency( planPrices.rawPrice, currencyCode, { stripZeros: true } )
+			currencyCode && planPrices.originalPrice
+				? formatCurrency( planPrices.originalPrice, currencyCode, { stripZeros: true } )
 				: null;
 
 		// TODO: Remove check once text is translated
@@ -185,4 +195,4 @@ const PlanFeatures2023GridBillingTimeframe: FunctionComponent< Props > = ( props
 	return <div>{ description }</div>;
 };
 
-export default localize( PlanFeatures2023GridBillingTimeframe );
+export default PlanFeatures2023GridBillingTimeframe;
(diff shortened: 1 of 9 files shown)


client/state/plans/selectors/get-plan-prices.ts
```diff
@@ -15,6 +19,8 @@ export interface PlanPrices {
 /**
  * A convenience function that returns the undiscounted and discounted prices
  * for a given plan and site.
+ * If the planSlug passed is the current plan, then this function returns
+ * { rawPrice: sitePlanRawPrice, discountedRawPrice: 0, planDiscountedRawPrice: 0 }
```
ddc22: Do you think this might be a bit confusing? I think it stems from the fact that we combine `getPlanPrices`, `getSitePlanPrices`. Because getPlanPrices essentially means I will get the price information for a given plan. And the first question that comes to my head, when I see this, is, why are the discounts not available when I pass in the current site's plan?

I think this is UI display logic that has been placed in the data layer, which should ideally live close to display logic. The fact that we do not show the discount information for the current purchased plan should be handled closer to the presentation layer (i.e. near the react component where this happens)

There is potential for duplication here, true! But the increase in complexity is not an ideal bargain I think 🤔

client/state/plans/selectors/get-plan-prices.ts
```diff
@@ -26,6 +32,19 @@ export function getPlanPrices(
...
+					planSlug === currentSitePlan?.productSlug &&
+					getSitePlanRawPrice( state, siteId, planSlug, { returnMonthly } ) ) ||
+				0,
+			discountedRawPrice: 0,
+			planDiscountedRawPrice: 0,
+		};
+	}
 
 	return {
 		rawPrice: ( productId && getPlanRawPrice( state, productId, returnMonthly ) ) || 0,
```
ddc22: I know this not related but is the `planSlug &&` condition check required below on line number 55?

https://github.com/Automattic/wp-calypso/blob/2a6fdd65640d80d4c9a71fa663b46c48c0bc8526/client/state/plans/selectors/get-plan-prices.ts#L53-L57

ddc22 (changes requested):
I had a query about the implementation.

client/state/plans/selectors/get-plan-prices.ts
aneeshd16: Thanks @jdc91, that makes sense to me and I agree with you. I think the root cause is that the billing timeframe and plan price are rendered by different components, with each component having it's own (duplicated) logic.

-----

image

-----

How do we establish a guarantee that both components will display the same price? One solution is to have a common parent component for both these, which calculates the price to be shown and pass it down to the billing-timeframe and plan-price components as props. This would stop the duplication of logic and also ensure that the prices shown are consistent between both components. WDYT?

ddc22: Ah yes! That is a valid concern and your solutions sounds just about right! 💯

client/state/plans/selectors/get-plan-prices.ts
```diff
@@ -26,6 +32,19 @@ export function getPlanPrices(
 ): PlanPrices {
 	const plan = getPlan( planSlug );
 	const productId = plan?.getProductId();
+	const currentSitePlan = getCurrentPlan( state, siteId );
+
+	if ( currentSitePlan?.productSlug === planSlug ) {
+		return {
+			rawPrice:
+				( siteId &&
```
jeyip: `Non-blocking`:

How do we feel about scooting this set of conditionals out of the object literal to make it easier to read?

```javascript
	const rawPrice =
		( siteId &&
			planSlug === currentSitePlan?.productSlug &&
			getSitePlanRawPrice( state, siteId, planSlug, { returnMonthly } ) ) ||
		0;

	if ( currentSitePlan?.productSlug === planSlug ) {
		return {
			rawPrice,
			discountedRawPrice: 0,
			planDiscountedRawPrice: 0,
		};
	}
```

jeyip:
## Testing
### Requirements

Tested with INR and MXN currency discounts

**Personal to Premium upgrade with pro-rated credits**
- [x] Premium Header price displays the non-discounted monthly price
- [x] Premium Billing Timeframe displays the non-discounted yearly price

**Premium to Business upgrade with pro-rated credits**
- [x] Business Header price displays the non-discounted monthly price
- [x] Business Billing Timeframe displays the non-discounted yearly price

**Personal to Premium upgrade with pro-rated credits to Business upgrade with pro-rated credits**
- [x] Business Header price displays the non-discounted monthly price
- [x] Business Billing Timeframe displays the non-discounted yearly price

### Browsers
- [x] Chrome

jeyip:
Question about behavior:

- I bought the currency discounted personal plan.
- I upgraded to the non-discounted premium plan with pro-rated credits
- I upgraded again to the non-discounted business plan with pro-rated credits

Just to clarify, what prices do we expect to display for the plans we're upgrading **_from_**?

In the situation above, the business plan we upgrade **_to_** displays non-discounted values correctly. The premium plan we upgraded **_from_**, however, displays currency discounted prices. I would've expected only the personal plan to display currency discounted prices.

Screenshot 2023-05-09 at 7 24 16 AM

client/my-sites/plan-features-2023-grid/components/header-price.tsx
```diff
@@ -3,15 +3,17 @@ import styled from '@emotion/styled';
...
+import { usePlanPricesDisplay } from '../hooks/use-plan-prices-display';
 import type { PlanProperties } from '../types';
 
 interface PlanFeatures2023GridHeaderPriceProps {
 	planProperties: PlanProperties;
 	is2023OnboardingPricingGrid: boolean;
 	isLargeCurrency: boolean;
 	isPlanUpgradeCreditEligible: boolean;
+	currentSitePlanSlug: string;
+	siteId: number;
```
jeyip: I'm seeing a type error for `siteId` which is at least partially responsible for the unit test CI failure

`Type 'number | undefined' is not assignable to type 'number | null'.
  Type 'undefined' is not assignable to type 'number | null'.ts(2322)`

aneeshd16:
Thanks for the review, @jeyip!

> In the situation above, the business plan we upgrade to displays non-discounted values correctly. The premium plan we upgraded from, however, displays currency-discounted prices. 

This is a great question! For non-purchasable plans, I think we should show the undiscounted price here. When a user buys a currency-discounted plan, they are given credits equivalent to the undiscounted value of the plan. So after they upgrade, the price of their old plan should reflect the actual value of the plan, which is the undiscounted price. What do you think?

> I would've expected only the personal plan to display currency-discounted prices.

I think both Personal and Premium should display the un-discounted prices since switching to those plans on the discounted prices is impossible after a site has already been created. It would end up looking like this:
image


Tagging @Automattic/martech-decepticons here for thoughts and suggestions.

client/my-sites/plan-features-2023-grid/hooks/use-plan-prices-display.ts
aneeshd16, line 1: I'll be adding tests once we agree with this approach (the new hook).

client/my-sites/plan-features-2023-grid/components/billing-timeframe.tsx
```diff
@@ -145,4 +155,4 @@ const PlanFeatures2023GridBillingTimeframe: FunctionComponent< Props > = ( props
 	return <div>{ perMonthDescription }</div>;
 };
 
-export default localize( PlanFeatures2023GridBillingTimeframe );
```
aneeshd16, line 188: `localize` wasn't required here.

jeyip:
> I think we should show the undiscounted price here. When a user buys a currency-discounted plan, they are given credits equivalent to the undiscounted value of the plan. So after they upgrade, the price of their old plan should reflect the actual value of the plan, which is the undiscounted price. What do you think?

> I think both Personal and Premium should display the un-discounted prices since switching to those plans on the discounted prices is impossible after a site has already been created. It would end up looking like this

I think this could make sense, but my only concern is that we run into the problem we were originally trying to address, where, for the first plan they purchase that's currency discounted ( in this example the Personal plan ), it "...gives the impression that the user was charged for more than the price shown on this page."

aneeshd16:
> I think this could make sense, but my only concern is that we run into the problem we were originally trying to address, where, for the first plan they purchase that's currency discounted ( in this example the Personal plan ), it "...gives the impression that the user was charged for more than the price shown on this page."

That's a good point! I've started a discussion here( pdgrnI-2cl-p2#comment-4003 ) on how to show the price for plans the user upgraded from. Do you think this PR is good to go for the current plan case?

jeyip:
> That's a good point! I've started a discussion here(pdgrnI-2cl-p2#comment-4003) on how to show the price for plans the user upgraded from. Do you think this PR is good to go for the current plan case?

Yes! In fact, I'm not sure what the team culture is around **follow-up PRs,** but I think what we have now with current plans is definitely an improvement over what already exists.

If the team is okay with iterative progress, I think it makes sense to move forward with what we have now and circle back once the p2 discussion is resolved to handle the previously purchased plans. 🙂

southp:
@aneeshd16 I've made a call in p2-pdgrnI-2cl#comment-4009 just now and that can be done as a follow-up. Along side with it, I'm seeing that the plans lower than the current plan have "credits applied" label, which is not accurate because it doesn't really applied: 

image

I know it's a separate issue, but could you also help fix it? 

Other than these two, I think this PR is good to go once rebased. Please feel free to ping me or @jeyip for another look once that's done :)

aneeshd16:
Thanks @southp for the review and help in taking this forward! I have updated the PR according to the option #1 described in pdgrnI-2cl-p2#comment-4003. The only difference is that instead of checking if the current plan is free, the code instead checks if the passed plan is available for purchase. IMO, this is a more robust check, and it's the same check used to show the "Downgrade" text on the action buttons.

southp (approved):
Thanks for addressing the feedback. I've confirmed:


* Logged-in as a paid plan owner using a currency having the first-year discount
    * The current plan and the lower-tierd plans show the original prices
    * The higher-tired plans show upgrade-credit-applied prices
    * The above applies to the monthly plans as well.
    * What the plans comparison grid shows agrees with the plans grid. 
* Logged-in as a paid plan owner using a currecnt having no first-year discount
    * Works as it is in production

Also, thanks for covering the new hook by a test suite. LGTM 🚢

PR #74313, opened 2023-03-10
Launchpad: Update launchpad "Choose a domain" flow

agrullon95:
Related to #74416

## Testing / Review Estimations

Testing: Medium
Review: Medium

## Proposed Changes

* Created a new domain upsell stepper flow to reuse the existing onboarding "Domain Selection" and "Plan selection" screens
* Modified the Launchpad "Choose a domain" task to navigate to this new domain upsell stepper flow instead of `/domains/add`

## Testing Instructions

* Checkout this branch
* Create a new launchpad-enabled site or use an existing site (write, build, or free flow) 
* Click the "Choose a domain" task
* Verify the "use your own domain link" adds the correct domain and plan to your cart on the checkout screen. Empty the cart without completing the purchase and return to the domain upsell flow through the launchpad screen. (You can use any valid domain like tinyurl.com)
* Complete the flow (select domain and any plan) and ensure the domain/plan has been purchased successfully.
* Verify the Launchpad "Choose a domain" task is marked as completed after upgrading to a paid plan
* Verify existing usage of the domains component isn't affected by these changes (`/setup/newsletter` and `/setup/link-in-bio use this component)
* Verify the "Choose a domain" task redirects to `/domains/manage` if the user is on a paid plan

**Domain selection screen**
CleanShot 2023-03-20 at 11 26 28@2x

**Plan selection screen**
image

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
[...]

PR #76233, opened 2023-04-25
Signup Stepper: Only add products to cart if a site slug exists

sirbrillig:
In https://github.com/Automattic/wp-calypso/pull/68722 and https://github.com/Automattic/wp-calypso/pull/74313 the domains and plans steps were added to the Stepper framework of signup. Those steps were configured to add products and plans to the shopping cart. However, they use `as string` to override TypeScript's protections when calling the `addPlanToCart()`, `addProductsToCart()`, and `addTempSiteToSourceOption()` helper functions. This defeats the purpose of TypeScript and could result in any number of hard-to-find bugs like those fixed in https://github.com/Automattic/wp-calypso/pull/76157. 

I followed the code in those helper functions to see what would happen if the data was _not_ a string (as could happen in production without these changes) and it would cause failed network calls to `/sites/undefined/themes/mine` and `/migrations/from-source/undefined`. Assuming no fatal errors, the cart items would then be added to the user's `no-site` cart. It's likely that these issues are not happening because the data in question actually _does_ always exist in practice, but we can confirm that without opting-out of TypeScript's guards by just using JavaScript `if` statements.

## Proposed Changes

In this diff we modify the `createSite()` function inside `SiteCreationStep` to remove the `as string` type casting. The result is that in order to safely call the helper methods, we must also add guards to prevent calling them if `site.siteSlug` (and in one case `siteData.ID`) does not exist.

We also modify `submit()` inside `useStepNavigation()` in the `domainUpsell` flow to prevent calling those same helpers when a site slug or flow name does not exist.

## Testing Instructions

The biggest risk here is that the type casting could be hiding incorrect types on the helper functions; that is, if `addPlanToCart()`, etc. are intended to actually work without a site slug, then this PR would break that functionality. So the first thing to verify is: are these flows intended to work without a `siteSlug`? I'm hoping that's not the case since those types have existed for longer than the code this PR touches.

I'm not familiar with this section of the codebase but we can probably use the testing instructions from the original diffs:

- Go to `http://calypso.localhost:3000/setup?flow=link-in-bio`
- Try the flow: free/paid domains, free/paid plans, `decide later`, filter domains, ...

And:

* Create a new launchpad-enabled site (http://calypso.localhost:3000/setup/free) 
* Click the "Choose a domain" task
* Verify the "use your own domain link" adds the correct domain and plan to your cart on the checkout screen. Empty the cart without completing the purchase and return to the domain upsell flow through the launchpad screen. (You can use any valid domain like tinyurl.com)
* Complete the flow (select domain and any plan) and ensure the domain/plan has been purchased successfully.
* Verify the Launchpad "Choose a domain" task is marked as completed after upgrading to a paid plan
* Verify existing usage of the domains component isn't affected by these changes (`/setup/newsletter` and `/setup/link-in-bio use this component)
* Verify the "Choose a domain" task redirects to `/domains/manage` if the user is on a paid plan

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
@@ -42,6 +42,13 @@ const DEFAULT_WOOEXPRESS_FLOW = 'pub/twentytwentytwo';
 const DEFAULT_NEWSLETTER_THEME = 'pub/lettre';
 const DEFAULT_START_WRITING_THEME = 'pub/livro';
 
+function hasSourceSlug( data: unknown ): data is { sourceSlug: string } {
+	if ( data && ( data as { sourceSlug: string } ).sourceSlug ) {
+		return true;
+	}
+	return false;
+}
+
 const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, data } ) {
 	const { submit } = navigation;
 	const { __ } = useI18n();
@@ -121,8 +128,9 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 			};
 		}
 
+		const sourceSlug = hasSourceSlug( data ) ? data.sourceSlug : undefined;
 		const site = await createSiteWithCart(
-			flow as string,
+			flow,
 			true,
 			isPaidDomainItem,
 			theme,
@@ -132,20 +140,20 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 			true,
 			username,
 			domainCartItem,
-			data?.sourceSlug as string
+			sourceSlug
 		);
 
-		if ( planCartItem ) {
-			await addPlanToCart( site?.siteSlug as string, flow as string, true, theme, planCartItem );
+		if ( planCartItem && site?.siteSlug ) {
+			await addPlanToCart( site.siteSlug, flow, true, theme, planCartItem );
 		}
 
-		if ( productCartItems?.length ) {
-			await addProductsToCart( site?.siteSlug as string, flow as string, productCartItems );
+		if ( productCartItems?.length && site?.siteSlug ) {
+			await addProductsToCart( site.siteSlug, flow, productCartItems );
 		}
 
-		if ( isMigrationFlow( flow ) ) {
+		if ( isMigrationFlow( flow ) && site?.siteSlug && siteData?.ID ) {
 			// Store temporary target blog id to source site option
-			addTempSiteToSourceOption( site?.siteId as number, siteData?.ID as number );
+			addTempSiteToSourceOption( site.siteId, siteData.ID );
 		}
 
 		return {
--- a/client/landing/stepper/declarative-flow/domain-upsell.ts
+++ b/client/landing/stepper/declarative-flow/domain-upsell.ts
@@ -70,16 +70,16 @@ const domainUpsell: Flow = {
 						const planCartItem = getPlanCartItem();
 						const domainCartItem = getDomainCartItem();
 						if ( planCartItem ) {
-							await addPlanToCart( siteSlug as string, flowName as string, true, '', planCartItem );
+							await addPlanToCart( siteSlug, flowName, true, '', planCartItem );
 						}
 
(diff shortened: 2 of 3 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
```diff
@@ -42,6 +42,13 @@ const DEFAULT_WOOEXPRESS_FLOW = 'pub/twentytwentytwo';
 const DEFAULT_NEWSLETTER_THEME = 'pub/lettre';
 const DEFAULT_START_WRITING_THEME = 'pub/livro';
 
+function hasSourceSlug( data: unknown ): data is { sourceSlug: string } {
```
agrullon95, line 45: Can we set `data: unknown` to `data: SiteDetails` here?

sirbrillig: 🤔 [SiteDetails](https://github.com/Automattic/wp-calypso/blob/f8584807c526b7e1abb68371df9e981d49fab431/packages/data-stores/src/site/types.ts#L105) doesn't have a `sourceSlug` property, so I don't think that's applicable here. Unless it's something missing from that type?

agrullon95: Youre right. I'm looking into that right now.  Doesnt seem to exist there.  Are we expecting SiteDetails to contain `sourceSlug`? Do you have an example of how the response looks with the sourceSlug property & value?

client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
```diff
@@ -132,20 +140,20 @@ const SiteCreationStep: Step = function SiteCreationStep( { navigation, flow, da
 			true,
 			username,
 			domainCartItem,
-			data?.sourceSlug as string
+			sourceSlug
```
agrullon95, line 143: Do we need to update the `sourceSlug` to consider undefined?

CleanShot 2023-04-25 at 13 57 31@2x

client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
sirbrillig: I don't know much about Stepper or where this data comes from, unfortunately. In this PR I'm just adjusting the code to operate the way it seems to be designed. 

In this case, `data` was being used as `data?.sourceSlug as string`, but that expression itself is very dangerous because the optional chaining shows that it may in fact resolve to `undefined` and not a string. Here I've just added a [TypeScript predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to narrow `Record<string, unknown>` (the type of `data` before this PR) to `undefined | { sourceSlug: string }` (which is how it's being used).

client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
sirbrillig: The question mark after `sourceSlug` counts as `undefined` so it is already supported.

agrullon95:
### Tests

- [x]  Confirm free launchpad site creation still functions as it previously did. (This occurs after the user clicks "Continue" on `/setup/free/freeSetup`)
- [x]  Confirm the Domain upsell flow is still adding the selected domain and plan to the cart.
- [x] Created new sensei site. I didn't test sensei specific features, but I was able to confirm the onboarding flow completed successfully and the "Sensei LMS" page was loading correctly.

client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx
agrullon95: I see. I'm not too familiar with `sourceSlug` either, but your changes seem to mimic the previous logic.

agrullon95 (approved):

escapemanuele:
I tested the following flows, and they all worked fine: link-in-bio, link-in-bio-domain, newsletter and site-setup ✌️

PR #76239, opened 2023-04-25
Added P2 label to the list of sites in Subscription Manager

### LINKED ISSUE #75078: `/subscriptions/sites`: Add `P2` tag to each P2 site on the list
When the site is a P2 site, it should have the following badge:

Markup on 2023-03-29 at 16:43:23

## Acceptance Criteria

- each P2 site has a badge / label as can be seen in the design: inDLaEQV8jJ21O4WXawIsJ-fi-712-15534

JuanLucha:
⚠️ Please do not merge until the D109848-code patch is merged and deployed. ⚠️ 

Fixes https://github.com/Automattic/wp-calypso/issues/75078

## Proposed Changes

This PR adds the P2 label to the list of sites in the new Subscription Manager, as seen in the Figma file: inDLaEQV8jJ21O4WXawIsJ-fi-712_18319
Screenshot 2023-05-23 at 19 25 24

## Testing Instructions

1. Apply this PR.
2. Go to this file `client/server/pages/index.js` and change the call back in `app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {` to it ends up like this:
```js
		return next();
	} );
```
3. Log in to your Calypso local and go to `http://calypso.localhost:3000/subscriptions/sites`.
4. If your user has P2 and non-P2 site subscriptions, you should see the P2 tag beside the site's name accordingly.

### DIFF
--- a/client/landing/subscriptions/components/site-list/site-row.tsx
+++ b/client/landing/subscriptions/components/site-list/site-row.tsx
@@ -81,7 +81,7 @@ export default function SiteRow( {
 					<span className="title-column">
 						<span className="name">
 							{ name }
-							{ organization_id && <span className="p2-label">P2</span> }
+							{ !! organization_id && <span className="p2-label">P2</span> }
 						</span>
 						<span className="url">{ hostname }</span>
 					</span>
--- a/client/landing/subscriptions/components/site-list/site-row.tsx
+++ b/client/landing/subscriptions/components/site-list/site-row.tsx
@@ -41,7 +41,7 @@ export default function SiteRow( {
 	URL: url,
 	date_subscribed,
 	delivery_methods,
-	organization_id,
+	is_wpforteams_site,
 }: SiteSubscription ) {
 	const hostname = useMemo( () => {
 		try {
@@ -81,7 +81,7 @@ export default function SiteRow( {
 					<span className="title-column">
 						<span className="name">
 							{ name }
-							{ !! organization_id && <span className="p2-label">P2</span> }
+							{ !! is_wpforteams_site && <span className="p2-label">P2</span> }
 						</span>
 						<span className="url">{ hostname }</span>
 					</span>
--- a/client/landing/subscriptions/components/site-list/site-row.tsx
+++ b/client/landing/subscriptions/components/site-list/site-row.tsx
@@ -41,6 +41,7 @@ export default function SiteRow( {
 	URL: url,
 	date_subscribed,
 	delivery_methods,
+	organization_id,
 }: SiteSubscription ) {
 	const hostname = useMemo( () => {
 		try {
@@ -78,7 +79,10 @@ export default function SiteRow( {
 				<span className="title-box" role="cell">
 					<SiteIcon iconUrl={ site_icon } size={ 48 } siteName={ name } />
 					<span className="title-column">
-						<span className="name">{ name }</span>
+						<span className="name">
+							{ name }
+							{ organization_id && <span className="p2-label">P2</span> }
+						</span>
 						<span className="url">{ hostname }</span>
 					</span>
 				</span>
--- a/client/landing/subscriptions/components/site-list/styles.scss
+++ b/client/landing/subscriptions/components/site-list/styles.scss
@@ -1,8 +1,7 @@
 @import "@automattic/color-studio/dist/color-variables";
 @import "@automattic/typography/styles/variables";
 @import "@wordpress/base-styles/breakpoints";
-@import "client/assets/stylesheets/style";
-@import "../../styles";
+@import "client/assets/stylesheets/p2-vars";
 
 .subscription-manager__site-list {
 
--- a/client/landing/subscriptions/components/site-list/styles.scss
+++ b/client/landing/subscriptions/components/site-list/styles.scss
@@ -83,6 +83,19 @@
 					}
 				}
 			}
+
+			.p2-label {
+				display: inline-block;
+				text-align: center;
+				width: 36px;
+				height: 20px;
+				background: #2a6aee;
+				color: $studio-green-0;
+				font-size: 12;
+				border-radius: 4px;
+				margin-left: 8px;
+				line-height: 20px;
+			}
 		}
 
 		.title-box,
--- a/client/landing/subscriptions/components/site-list/styles.scss
+++ b/client/landing/subscriptions/components/site-list/styles.scss
@@ -1,6 +1,8 @@
 @import "@automattic/color-studio/dist/color-variables";
 @import "@automattic/typography/styles/variables";
 @import "@wordpress/base-styles/breakpoints";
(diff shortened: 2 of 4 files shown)


ivan-ottinger:
I am reviewing this PR now.

client/landing/subscriptions/components/site-list/site-row.tsx
```diff
@@ -57,7 +58,10 @@ export default function SiteRow( {
 				<span className="title-box" role="cell">
 					{ siteIcon }
 					<span className="title-column">
-						<span className="name">{ name }</span>
+						<span className="name">
+							{ name }
+							{ organization_id && <span className="p2-label">P2</span> }
```
ivan-ottinger: When we look at the Figma design, there are actually two different tags:

1. blue for `P2` sites
2. green for `Paid` subscriptions

It is not clear to me whether the Paid subscriptions will only have `organization_id` of a certain value, or there will be some other property flagging Paid subscriptions.

With that said, so far it looks like that P2 sites have `organization_id === 1`. Because of that, we might want to check for that value specifically (because simple `organization_id &&` would evaluate as `true` for all `organization_id`s that are "non-zero" - which might include Paid sites as well)

Btw, I noticed that some of the P2s have `organization_id === 2`.

client/landing/subscriptions/components/site-list/styles.scss
```diff
@@ -74,6 +74,19 @@ $max-list-width: 1300px;
...
+				text-align: center;
+				width: 36px;
+				height: 20px;
+				background: #2a6aee;
+				color: $studio-green-0;
+				font-size: 12;
+				border-radius: 4px;
+				margin-left: 8px;
+				line-height: 20px;
+			}
```
ivan-ottinger: Since the same label would be used for Comments list as well, I wonder if it would make sense to move the styling to more general SCSS file that is shared across both lists. We could consider `client/landing/subscriptions/styles.scss` for instance.

client/landing/subscriptions/components/site-list/styles.scss
```diff
@@ -74,6 +74,19 @@ $max-list-width: 1300px;
...
 			}
+
+			.p2-label {
+				display: inline-block;
+				text-align: center;
+				width: 36px;
+				height: 20px;
+				background: #2a6aee;
+				color: $studio-green-0;
+				font-size: 12;
```
ivan-ottinger: The `font-size` value is missing `px`, but regardless, we can use variable instead. I think `$font-body-extra-small` should be the right one.

client/landing/subscriptions/components/site-list/styles.scss
```diff
@@ -74,6 +74,19 @@ $max-list-width: 1300px;
 					}
 				}
 			}
+
+			.p2-label {
+				display: inline-block;
+				text-align: center;
+				width: 36px;
+				height: 20px;
+				background: #2a6aee;
```
ivan-ottinger: I think there still isn't 100% agreement on what color should be used: p1682440802321329-slack-C02TCEHP3HA.

client/landing/subscriptions/components/site-list/site-row.tsx
ivan-ottinger: For `organization_id === 0`, the statement will render `"0"` string instead of `null`:

Markup on 2023-04-26 at 14:30:40

Depending on how the final logic will look like, we may want to use `? <span></span> : null` instead.

ivan-ottinger (commented):
The changes display nice blue tag for P2 sites. 🙂 👍🏼 

I have noticed a couple things we will need to address though (please see the comments below).

client/landing/subscriptions/components/site-list/site-row.tsx
JuanLucha: Thanks for pointing this out! In fact, making double not (`!! organization_id`) fixes the problem.

client/landing/subscriptions/components/site-list/styles.scss
JuanLucha: We reached a consensus. Applying the correct one now.

client/landing/subscriptions/components/site-list/styles.scss
JuanLucha: Good point! Using the variable now.

client/landing/subscriptions/components/site-list/styles.scss
JuanLucha: Great point! I'll move it right away.

client/landing/subscriptions/components/site-list/styles.scss
```diff
@@ -1,6 +1,8 @@
 @import "@automattic/color-studio/dist/color-variables";
 @import "@automattic/typography/styles/variables";
 @import "@wordpress/base-styles/breakpoints";
+@import "client/assets/stylesheets/style";
+@import "../../styles";
```
ivan-ottinger: I might be missing something, but this import shouldn't be necessary. 🤔

client/landing/subscriptions/components/site-list/styles.scss
```diff
@@ -1,6 +1,8 @@
 @import "@automattic/color-studio/dist/color-variables";
 @import "@automattic/typography/styles/variables";
 @import "@wordpress/base-styles/breakpoints";
+@import "client/assets/stylesheets/style";
```
ivan-ottinger: I would personally moved this import to the file where the CSS variable (for the P2 color) is used (`client/landing/subscriptions/styles.scss`) and considered importing just the P2 colors:

```scss
@import "client/assets/stylesheets/_p2-vars";
```

because `client/assets/stylesheets/style` includes a bunch of other imports we don't need for that one P2 color.

ivan-ottinger (commented):
Works as expected. 👍🏼👍🏼  Just two more things I noticed. 🙂

client/landing/subscriptions/components/site-list/site-row.tsx
ivan-ottinger: Perfect. 👍🏼👍🏼

ivan-ottinger (approved):
Nice! The proposed changes look good and work as expected 🙂👍🏼👍🏼  (with the D109848-code patch applied to my sandbox - which still needs a review).

- P2 label displays correctly on mobile as well
- I did not notice any regressions, errors, nor warnings.

I am adding the following warning to the PR description - to make sure this doesn't get merged before the patch is in accidentally.

⚠️ Please do not merge until the D109848-code patch is merged and deployed. ⚠️

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8011240

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8011240).

PR #76267, opened 2023-04-26
SHARING - Add post publish sharing modal

roo2:
In pe7F0s-gz-p2 we are adding a sharing modal after publishing a post. The purpose is to encourage bloggers to publish their post to social media and get views and engagement.

### Desktop

CleanShot 2023-05-03 at 14 56 37@2x

### Testing instructions

Sandbox a test site and 
Build and apply this version of the ETK to your sandbox with this command.
```
cd apps/editing-toolkit
yarn dev --sync
```
On your sandboxed blog, publish a new post
You should see the post publish sharing modal.

* clicking links should record tracks events
* The modal should not conflict with the other modals identified in https://github.com/Automattic/wp-calypso/pull/76615 or the jetpack [launchpad modal](https://github.com/Automattic/jetpack/blob/f60b8721717189e53129e856c75a9461a0d153fb/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/launchpad-save-modal.php)

#### Dismissing the modal.
The modal can be dismissed with the checkbox. The modal can be re-enabled from `wpsh` by running
```
update_blog_option( $site_id, 'sharing_modal_dismissed', false )
```


roo2:
@davemart-in I've added two sharing buttons and got the copy to clipboard working, there are a few known issues:

* The twitter logo is square. I looked into this and I think we might not be able to use the `SocialLogo` component.
* I need to check the facebook API key

I've changed line 46 to default the popup to open by default.
`const [ isOpen, setIsOpen ] = useState( true );`

roo2:
I didn't add the number of subscribers just because it would require another API call to get all of the site data every time the editor is loaded which we aren't making yet. What do you think @ollierozdarz?

eoigal (approved):
WFM. Nice job

davemart-in:
> I didn't add the number of subscribers just because it would require another API call to get all of the site data every time the editor is loaded which we aren't making yet.

I think that is fine.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/block-editor-nux.js
```diff
@@ -96,13 +97,16 @@ function WelcomeTour() {
...
-			<VideoPressCelebrationModal />
-		</>
-	),
+	render: () => {
+		return (
+			<>
+				<WelcomeTour />
+				<FirstPostPublishedModal />
+				<PostPublishedSharingModal />
+				<SellerCelebrationModal />
```
creativecoder: @roo2 FYI, `SellerCelebrationModal` is the model I had mentioned last week as something we should make sure we're not conflicting with. Since these are all grouped together, I'm assuming there's logic here to make sure there aren't multiple shown at the same time, is that right?

roo2: thanks, I forgot to investigate this actually! 😬 I will investigate today 👀

roo2:
added an opt-out function

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/post-published-sharing-modal/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+	const previousIsCurrentPostPublished = useRef( isCurrentPostPublished );
+	const shouldShowFirstPostPublishedModal = useSelect(
+		( select ) =>
+			(
+				select( 'automattic/wpcom-welcome-guide' ) as WpcomWelcomeGuideSelectors
+			 ).getShouldShowFirstPostPublishedModal(),
+		[]
+	);
+
+	const [ isOpen, setIsOpen ] = useState( false );
```
davemart-in: We'll need to make sure to remove this line, right?

davemart-in: Or is it fine because it's set to false?

davemart-in:
@roo2 Heads up: I'm currently working on making a few minor tweaks to the design:

- [x] Update copy at top of modal and link to subscribers section
- [x] Update link color consistency
- [x] Update "Share with"  to be "Get more traffic to your post by sharing"
- [x] Change title from "Congratulations!" to "Post published!" (my worry here is what if someone is publishing something sad or difficult in their life. Would "Congratulations" come off the wrong way?)

creativecoder:
Given the prominence of this new modal, can we do a string freeze on this PR when the copy is finalized and make sure it's translated before release to users?

I know waiting for translations is no fun, but showing non-English users a modal of gibberish every time they publish a post is a horrible UX, imho.

davemart-in:
@creativecoder, yep. I think that is a good call due to the prominence of this modal. All of my string changes should be in shortly.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/post-published-sharing-modal/index.tsx
roo2: yea false is correct

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7966477

roo2:
Refactored the seller celebration modal to work with this. To test the seller celebration modal:

* Create a new site and select the "sell" intent
* Sandbox the test site, reload proxy, refresh the page.
* In the post editor, add a "payments buttons" block,
* Save the page
* You should see the seller celebration modal
* The modal will only be shown once, to reset it, run in wpsh: `update_blog_option( $site_id, 'has_seen_seller_celebration_modal', false )`
Screen Shot 2023-05-04 at 12 47 16 pm

Note: I found that the seller celebration modal wasn't shown in the site editor despite site editor specific code existing, I tested in production and it seems to be an existing bug.

roo2:
Also refactored the video celebration modal to work with the sharing modal, to test:
* Sandbox a site with site_intent=videopress (with `update_blog_option($site_id, 'site_intent', 'videopress')`)
* Create a post and add a video block to it.
* Publish the post
* You should see the videopress celebration modal.
* On subsequent posts, you should see the sharing modal, reset the celebration modal with `update_blog_option( 218612585, 'has_seen_video_celebration_modal', false)`

Screen Shot 2023-05-04 at 2 52 51 pm

Note: the launchpad modal seems to already be interfering with the videopress celebration modal

roo2:
Also added code to avoid overlapping with the launchpad modal https://github.com/Automattic/jetpack/tree/f60b8721717189e53129e856c75a9461a0d153fb/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal

roo2:
I found that other congratulations modals have been disabled on atomic, apparently due to an issue with updating synced site options. I think we should leave this modal also disabled on atomic for now, and look into enabling it once we've released the first version.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/post-published-sharing-modal/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+import type { SelectFromMap } from '@automattic/data-stores';
+import './style.scss';
+
+type WpcomWelcomeGuideSelectors = SelectFromMap< typeof wpcomWelcomeGuideSelectors >;
+type CoreEditorPlaceholder = {
+	getCurrentPost: ( ...args: unknown[] ) => { link: string; title: string };
+	getCurrentPostType: ( ...args: unknown[] ) => string;
+	isCurrentPostPublished: ( ...args: unknown[] ) => boolean;
+};
+const FB_APP_ID = '249643311490';
```
Addison-Stavlo: Im not totally up to speed on the context here, but is this app ID something that should be set as an environment variable or hidden from the public eye in any way?

davemart-in: @Addison-Stavlo nice catch! I think you're right.

All of the [other instances](fbhepr%2Sfrnepu%3Scebwrpg%3Qpnylcfb%26cebwrpg%3Qwrgcnpx%26cebwrpg%3Qjcpbz%26shyy%3Q249643311490%26qrsf%3Q%26ersf%3Q%26cngu%3Q%26uvfg%3Q%26glcr%3Q%26keq%3Q%26aa%3Q3%26a%3Q25%26fgneg%3Q25-og) I see of this ID are in PHP files.

On https://designschool.org/ (in the top right corner) I just link to `https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Fdesignschool.org%2F` without any JS includes and it works. Wonder if we could do something similar here (but still show it in a popup vs. full screen).

Addison-Stavlo: > All of the [other instances](https://github.com/Automattic/wp-calypso/pull/fbhepr%2Sfrnepu%3Scebwrpg%3Qpnylcfb%26cebwrpg%3Qwrgcnpx%26cebwrpg%3Qjcpbz%26shyy%3Q249643311490%26qrsf%3Q%26ersf%3Q%26cngu%3Q%26uvfg%3Q%26glcr%3Q%26keq%3Q%26aa%3Q3%26a%3Q25%26fgneg%3Q25-og) I see

This link seems to be a 500 error for me. PHP or JSX, I wasn't sure if this was something that should be hidden from a public repo or not.

davemart-in: Sorry. My A8c github extension is rewriting the URL to prevent it from leaking. I'll post the link in our Slack channel.

> I wasn't sure if this was something that should be hidden from a public repo or not.

I'll let Eru weigh in when he's online.

Addison-Stavlo: I think I may have been confusing this with the App Secret.

> I'll let Eru weigh in when he's online.

👍

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/style.scss
```diff
@@ -0,0 +1,163 @@
+@import "@wordpress/base-styles/breakpoints";
+@import "@wordpress/base-styles/mixins";
+
+.components-modal__screen-overlay {
```
simison: I think this ends up affecting all the modals after the the CSS has been loaded, not just this specific modal?

Was there something that should be improved in the modal component directly in Gutenberg?

simison:
Some notes on that _"delivered to each of your subscribers"_ part:

- You could remove the line entirely to simplify everything going on in this modal; that said I also like the little celebration it adds.

- We have "paid newsletters" functionality coming up soon and it'll allow customers to limit who actually gets the post. Showing confirmation _"to each of your subscribers"_ can be confusing right after they've chosen to limit only "paid subscribers". To solve this, you could pull the copy technically from Jetpack functionality which will take above into account:
    image

You could simplify by leaving out the link; it's one more link to get distracted instead of sharing, and unclear what they would do at that page anyway?

Overall I'm conflicted by having the modal duplicate sidepanel and both visible at the same time, but I still think the modal is pretty nice in grabbing attention to a single thing at once vs adding more content to the sidebar. Only if the dynamics between the modal and sidepanel could be solved somehow? 

Screenshot 2023-05-08 at 10 23 52

davemart-in:
@simison Let's keep the strings frozen for the v1. Happy to continue iterating and improving things in future iterations.

simison:
Yep, should've clarified that not a blocker for merging; follow up later is totes fine.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/style.scss
roo2: Actually this isn't neccessary at all. Good catch, I added it to deal with the a8c "super admin bar" which puts the admin bar on top of everything, hiding the `X` button. But I realized this only applies for a12s, so I'll just remove this code.

roo2: Maybe it should be changed in the a8c super admin bar plugin, I'll take a look at that 👀

roo2: Oh interesting, it looks like it has just been addressed with the latest Gutenberg update, they now leave a gap at the top of the modal, which happens to fit the super admin bar perfectly 😄 

https://github.com/wordpress/gutenberg/pull/50039

simison: Great. :-) And modals can be dismissed with esc key anyway.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+		// The first post will show a different modal.
+		if (
+			! shouldShowFirstPostPublishedModal &&
+			! shouldShowSellerCelebrationModal &&
+			! shouldShowVideoCelebrationModal &&
+			launchpadScreenOption !== 'full' &&
+			! previousIsCurrentPostPublished.current &&
+			isCurrentPostPublished &&
+			postType === 'post'
+		) {
```
okmttdhr, line 69: badge

It might be bug-prone to manage this kind of condition every time we add a new modal. (A dev who will add a new modal needs to know this condition exists somehow.)🙄
Maybe we should add a shared context (or whatever), such as `hasPostEditorModal`, or create something like a "mediator" that controls the display of a modal? 
What do you think?

simison:
@roo2 not a blocker for merging, but ideally this code would be shipped in Jetpack-mu-wpcom plugin as ETK is feature-freeze (p58i-dox-p2)

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7966477).

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx
roo2: hmmm I'm not sure, I think the other modals have managed so far because they all show in diferent contexts. But if two modals where to show at the same time, we need to decide which has priority. But then the problem is that every modal needs to know about every other modal it may conflict with.

I guess it depends on how many more modals we plan on adding, if only one or two more, maybe this current approach is ok, but any more and it becomes increasingly complex.

It's also complicated because of the direction to move new features into jetpack. Modals implemented there won't have access to these flags :/ 

My preference is to go with this approach for now and then maybe next time we add a modal, we should migrate all modals into jetpack, and think about if we need to change the approach then.

roo2: I probably should have implemented this in jetpack anyway 😛 but then I would have to migrate everything or I wouldn't have access to the other modal states. My preference is to do that on the next modal, but I am biased 😉

davemart-in:
@roo2 looks like [translations are in](https://github.com/Automattic/wp-calypso/pull/76267#issuecomment-1543743661)!

> not a blocker for merging, but ideally this code would be shipped in Jetpack-mu-wpcom plugin as ETK is feature-freeze (p58i-dox-p2)

Thanks for the heads up @simison. I wish we had caught that earlier on in this project. I was unaware. Since this modal builds on the modals already in ETK, and since we're trying to get this rushed out (due to the Twitter API deprecation), and since we have a [separate PR with improvements to those modals](https://github.com/Automattic/wp-calypso/pull/76615) I'd vote that we go ahead and make an exception to the freeze in this one instance and get this deployed. 

Now that we know about the freeze, we'll obviously not add anything else to ETK moving forward.

cc @ebinnion

simison:
Yep I think it's fine as this isn't super significant amount or complex piece of code.

PR #76291, opened 2023-04-26
Staging sites: fix staging creation navigation bug

kozer:
- Related to https://github.com/Automattic/dotcom-forge/issues/2208

## Proposed Changes

When the user creates a new staging site, and move back and forth to the page while the staging creation is in progress, faces strange issues like:

- The button to create a site becomes visible again
- When the simple staging site has been created, an error message is displayed instead of the progress bar.

This pr fixes the above issues.

NOTE: This doesn't fix the case where the user reloads the page. Instead for that case, we have implemented a lock mechanism in D109053 that prevents the user to create a new site. It definitely not a perfect solution, but it's better UX-wise solution than before.

## Testing Instructions

1. Create a new site with a business plan and make it atomic.
2. Use this diff D109294-code
3. Navigate to the `/navigation` page.
4. Press the "Add Staging Site" button.
5. Imediatelly, press another link in the sidebar on the left, eg: "Pages"
6. Navigate back to `/hosting` using the sidebar.
7. Ensure that you see the progress bar.

| **Before** | **After** |
| :---:         |    :---:       |
|before1before2| after1|

## Pre-merge Checklist


sejas (commented):
@kozer , nice approach using `useIsMutating` 🙌 .

I think we still having an error, here two recordings:

I was able to reproduce the error message clicking the back and forward buttons:

https://user-images.githubusercontent.com/779993/234623524-1289105c-775b-4f37-8837-837672ad2a0e.mp4

Also when clicking different pages in the sidebar:

https://user-images.githubusercontent.com/779993/234624495-071ff478-a45e-4974-82a6-15320cb1fdb2.mp4

kozer:
Thanks @sejas for catching this! I did some updates that fix the issue you mentioned. Can you please check it again?

sejas (commented):
@kozer , Thanks for your changes.
I've found a different issue. Now we are not showing the error message in any case.

https://user-images.githubusercontent.com/779993/234832801-77dff91e-d51f-4de7-935f-a322aac9773c.mp4

Steps to reproduce:
- Create a staging site
- Invite a new user to the productions side
- Access the hosting configuration from the other user
- I expect to see an error saying `Unable to access the staging site. Please contact site owner.`
- It shows that it's creating the staging site.

I remember that some states could overlap we may need to re-think the whole logic/UX.

danielbachhuber:
> I remember that some states could overlap we may need to re-think the whole logic/UX.

Maybe we should abstract and write some tests around the state-processing logic? This would make it easier to grok what's going on.

kozer:
> I remember that some states could overlap we may need to re-think the whole logic/UX.

The logic is a bit complicated, I'll try to investigate it a bit more.

>Maybe we should abstract and write some tests around the state-processing logic? This would make it easier to grok what's going on.

This will help, for sure. I'll try to go around it for now and open a new ticket to revisit it, and write some tests about that component

kozer:
@sejas , @danielbachhuber , I decided to return `site_owner` as a property at the staging site list endpoint and use that to determine if the user has access, to fix the last bug @sejas mentioned.
For that reason, I have created diff D109294-code.

Initially, we discussed with @sejas moving all the logic regarding transfer status and site access to the `useStagingSite` hook. This included fetching each site again using `wp.site`, to get the site owner.
Even if fetching again the site is ok for our use case, it creates an "N+1 problem" in case we want to support multiple staging sites, and I wanted to be proactive as we already have an endpoint that fetches staging site information and is build with multiple staging sites support, to begin with.

I also decided to avoid moving the rest of the code to the `useStagingSite`, for the same reason, plus I didn't want to risk potential breakage of the current implementation
IMO, we need to revisit and refactor the component to properly support multiple staging sites ( as the endpoint has been built with this in mind ), and probably move staging status, etc to this endpoint as well.

sejas:
@kozer , I left a comment in D109294-code. I like the approach of computing that property in the backend.
Instead of `site_owner`, we should check if the user has permission.

sejas:
Do not forget to remove `useHasSiteAccess` from `client/my-sites/hosting/staging-site-card/test/index.js`

https://github.com/Automattic/wp-calypso/blob/8f4bbd610ab580652ff7345a0bc2e46cb3f62f86/client/my-sites/hosting/staging-site-card/test/index.js#L12

client/my-sites/hosting/staging-site-card/index.js
```diff
@@ -85,7 +84,8 @@ export const StagingSiteCard = ( { currentUserId, disabled, siteId, siteOwnerId,
 	const stagingSite = useMemo( () => {
 		return stagingSites && stagingSites.length ? stagingSites[ 0 ] : [];
 	}, [ stagingSites ] );
-	const hasSiteAccess = useHasSiteAccess( stagingSite.id );
+	const hasSiteAccess =
+		! stagingSite.id || Boolean( stagingSite.id && stagingSite?.user_has_permission );
```
sejas: I think we could remove `stagingSite.id` from the second part of the condition.
```suggestion
		! stagingSite.id || Boolean( stagingSite?.user_has_permission );
```

client/my-sites/hosting/staging-site-card/index.js
```diff
@@ -157,6 +152,38 @@ export const StagingSiteCard = ( { currentUserId, disabled, siteId, siteOwnerId,
...
+		if (
+			! wasCreating &&
+			! isStagingSiteTransferComplete &&
+			stagingSite.id &&
+			transferStatus !== transferStates.REVERTED &&
+			hasSiteAccess &&
+			! isReverting
+		) {
+			setWasCreating( true );
+		}
```
sejas, line 211: Here is a small edge case where we are displaying the success notice.

The way to reproduce is by clearing the calypso_sotre indexedDB and refreshing the page.

I think this error comes from receiving the staging site but in the redux store we don't have its status yet. So the transfer status is null until the request finishes.

I'm fine deploying this PR and creating an issue for that.

## Screencast

https://user-images.githubusercontent.com/779993/235206011-064bd478-c014-4d49-a28b-a5b918a2ffb4.mp4

## State:

Screenshot 2023-04-28 at 17 36 13

Not sure if we really need this useEffect hook.

sejas (approved):
@kozer , `useIsMutating` and `user_has_permission` from the backend are great additions.

I found an unexpected behaviour in the new useEffect that sets `wasCreating` to true.
We could create a follow up to improve or remove that useEffect.

client/my-sites/hosting/staging-site-card/index.js
kozer: @sejas we need, as `wasCreating` is used to set that the site is being created. I'm sure that we can simplify that, however, I think we need to revisit, and probably rewrite this component.
I don't mind filling a new issue for that. I find that issue, not a big deal, in comparison to what is happening right now when the user is moving back and forth to the pages.

danielbachhuber:
@kozer @sejas Is this ready to land?

kozer:
> @kozer @sejas Is this ready to land?

Yes, I'm about to land it

PR #76293, opened 2023-04-26
Tags page responsive tweaks

davemart-in:
## Description

@ollierozdarz brought up in our Slack channel that the responsive mobile version of the newly renovated tags page still had some issues.

### Before

- under some of the letter sections there’s a large gap (K for example)
- H1 doesn’t look like it’s pulling a mobile style
- Top tags styles were off
- The A-Z should be mobile specific (position fixed, vertical) to the mobile design of the page.

### After

https://user-images.githubusercontent.com/5634774/234612308-748f6564-3fa2-4bbb-889b-992be9f8624e.mp4

### DIFF
--- a/client/my-sites/sidebar/style.scss
+++ b/client/my-sites/sidebar/style.scss
@@ -743,7 +743,7 @@ $font-size: rem(14px);
 			min-height: initial;
 		}
 
-		.focus-content .layout__content:not('.has-no-sidebar') {
+		.focus-content:not(.has-no-sidebar) .layout__content {
 			padding: 47px 0 0;
 		}
 
--- a/client/my-sites/sidebar/style.scss
+++ b/client/my-sites/sidebar/style.scss
@@ -743,7 +743,7 @@ $font-size: rem(14px);
 			min-height: initial;
 		}
 
-		.focus-content .layout__content {
+		.focus-content .layout__content:not('.has-no-sidebar') {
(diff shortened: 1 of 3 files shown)


client/reader/tags/alphabetic-tags.tsx
```diff
@@ -87,6 +89,23 @@ const scrollToLetter = ( letter: string ) => {
 
 export default function AlphabeticTags( { alphabeticTags }: AlphabeticTagsProps ) {
 	const translate = useTranslate();
+	const [ isSticky, setIsSticky ] = useState( false );
+
+	useEffect( () => {
+		const handleScroll = () => {
+			const tableA = document.getElementById( 'alphabetic-tags-table-A' );
```
roo2: I think this needs to be done with a ref so that react is guaranteed to recognize and load it properly. I'll take a stab at changing it.

client/my-sites/sidebar/style.scss
```diff
@@ -743,7 +743,7 @@ $font-size: rem(14px);
 			min-height: initial;
 		}
 
-		.focus-content .layout__content {
+		.focus-content .layout__content:not('.has-no-sidebar') {
```
roo2: quotes caused errors for me

roo2 (approved):
I applied two small changes and LGTM apart from that, I'll just go ahead and merge this

PR #76295, opened 2023-04-26
Open Sites view in Subscription Manager for internal EN users

### LINKED ISSUE #76280: `/subscriptions/sites`: Allow logged-in users to access subscriptions management portal paths in Calypso
We must create the feature flag `subscription-management/internal-users/sites-view` to show the Sites view in Subscriptions Manager to internal users in English.

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/76280

## Proposed Changes

This PR opens the Sites view in the Subscription Manager to all internal users (users with a WordPress account) in English.

## Testing Instructions

2. With the WordPress user logged in in your browser, go to `http://calypso.localhost:3000/subscriptions/sites`. You should see the Sites view in Calypso.
3. Then check that clicking on `Comments` and `Pending` tabs takes you to `https://wordpress.com/email-subscriptions/?option=comments&locale=en` and `https://wordpress.com/email-subscriptions/?locale=en&option=pending` respectively.
4. Go directly to `http://calypso.localhost:3000/subscriptions/comments` and `http://calypso.localhost:3000/subscriptions/pending` and check that it also takes you to https://wordpress.com/email-subscriptions/?option=comments&locale=en` and `https://wordpress.com/email-subscriptions/?locale=en&option=pending` respectively.
5. Then, check all those links and clicks with an external EN user.

### DIFF
--- a/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
+++ b/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
@@ -3,6 +3,7 @@ import { SubscriptionManager } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
 import { useTranslate } from 'i18n-calypso';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
+import { useIsLoggedIn } from 'calypso/../packages/data-stores/src/reader/hooks';
 import Nav from 'calypso/components/section-nav';
 import NavItem from 'calypso/components/section-nav/item';
 import NavTabs from 'calypso/components/section-nav/tabs';
@@ -20,10 +21,11 @@ const TabsSwitcher = () => {
 	const { pathname } = useLocation();
 	const { data: counts } = SubscriptionManager.useSubscriptionsCountQuery();
 	const locale = useLocale();
+	const { isLoggedIn } = useIsLoggedIn();
 	const shouldEnableCommentsTab =
-		config.isEnabled( 'subscription-management-comments-view' ) && locale === 'en';
+		config.isEnabled( 'subscription-management-comments-view' ) && locale === 'en' && ! isLoggedIn;
 	const shouldEnablePendingTab =
-		config.isEnabled( 'subscription-management-pending-view' ) && locale === 'en';
+		config.isEnabled( 'subscription-management-pending-view' ) && locale === 'en' && ! isLoggedIn;
 
 	const getFullPath = ( subpath: string ) =>
 		`/subscriptions/${ subpath }${ locale !== 'en' ? '/' + locale : '' }`;
--- a/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
+++ b/client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
@@ -3,7 +3,6 @@ import { SubscriptionManager } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
 import { useTranslate } from 'i18n-calypso';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
-import { useIsLoggedIn } from 'calypso/../packages/data-stores/src/reader/hooks';
 import Nav from 'calypso/components/section-nav';
 import NavItem from 'calypso/components/section-nav/item';
 import NavTabs from 'calypso/components/section-nav/tabs';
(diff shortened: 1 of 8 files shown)


client/server/pages/index.js
```diff
@@ -830,12 +830,16 @@ function wpcomPages( app ) {
 	} );
 
 	app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {
-		if ( req.context.isLoggedIn ) {
+		const showSitesViewToInternalUsers =
+			config.isEnabled( 'subscription-management/internal-users/sites-view' ) &&
+			req.path === '/subscriptions/sites';
```
mashikag: We do not have to do the: 
```
req.path === '/subscriptions/sites'
```
As long as we ensure that the tabs within the subscriptions' portal point the logged-in user back to the re-skinned portal when the Comments, the Pending, or the Settings tab is clicked.

mashikag: In fact, if we do not do the check, then it will make it easier for us to test the other views before we have a feature flag for them. 😁

JuanLucha: If we don't do `req.path === '/subscriptions/sites'` the user can put `http://calypso.localhost:3000/subscriptions/comments` directly into their browser and they will go to Comments in Calypso. We need to show only if it's Sites.

ivan-ottinger: Yeah, I agree that we could leave all tabs accessible for logged-in user - since it is behind the feature flag anyways.

JuanLucha: True. Removing the check.

client/landing/subscriptions/components/tabs-switcher/tabs-switcher.tsx
```diff
@@ -3,6 +3,7 @@ import { SubscriptionManager } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
 import { useTranslate } from 'i18n-calypso';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
+import { useIsLoggedIn } from 'calypso/../packages/data-stores/src/reader/hooks';
```
mashikag: You should not import the hook that way. The hook should be exported [here](https://github.com/Automattic/wp-calypso/blob/1701398d37cca5972323860dd4dd504fdd4b3873/packages/data-stores/src/reader/index.ts#L1), and then you can use it via the already imported here `SubscriptionManager`.

JuanLucha: Oh, nice. Doing the change 👍

JuanLucha: Very true! Change done 👍

client/server/pages/index.js
```diff
@@ -830,12 +830,16 @@ function wpcomPages( app ) {
 	} );
 
 	app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {
-		if ( req.context.isLoggedIn ) {
+		const showSitesViewToInternalUsers = config.isEnabled(
+			'subscription-management/internal-users/sites-view'
+		);
+
+		if ( req.context.isLoggedIn && ! showSitesViewToInternalUsers ) {
```
mashikag: Just thinking about it again. 🤔 I am really confused why we do not simply remove this check. Without the need for a feature flag here at all. I think it would make sense to do so, right?

mashikag: But mind the feature flag is still needed for the tabs navigation.

client/server/pages/index.js
```diff
@@ -828,21 +828,6 @@ function wpcomPages( app ) {
 				res.send( renderJsx( 'support-user' ) );
 			} );
 	} );
-
-	app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {
-		if ( req.context.isLoggedIn ) {
```
mashikag, line 833: I think we should remove only the following condition. I am sorry, I was a bit too vague with my explanation on our call when I said we can remove the whole condition.
```ts
if ( req.context.isLoggedIn ) {
	// We want to show the old subscriptions management portal to the logged-in users, until new one in reader is developped for them
	return res.redirect( 'https://wordpress.com/email-subscriptions?option=settings' );
}
```

Now, that we want to allow logged-in users to all the subscriptions portal paths, we should extend the following condition:
```ts
if ( reqc.context.isLoggedIn || req.cookies.subkey ) {
	// If the user is logged-in, or otherwise has a subkey cookie , they are authorized to view the page
	return next();
}
```

Lastly, we still want to show the email-subscriptions landing page to logged-out users who have no subkey. Therefore, the last redirect should remain in place.

mashikag (approved):

PR #76132, opened 2023-04-24
Site Assembler - Large patterns like LIB are cut-off in the large preview

miksansegundo:
Closes #76129
Related #76039 #72321

## Proposed Changes

* Fix pattern thumbnail preview of patterns with `min-height: 100vh`
* Fix also the large preview for all device viewports

### Before

https://user-images.githubusercontent.com/1881481/233988424-4ad1fd28-90cf-4367-af3d-8f9f19084c25.mov 

### After 

https://user-images.githubusercontent.com/1881481/233988658-c4ab5b36-a25c-4d45-834d-ba436f004f97.mov

## Testing Instructions

- Go to /setup?siteSlug=<your_site>
- Click the `Continue` button until you land on the Design Picker step
- Scroll to the bottom and click on `Start Designing`
- Click `Sections` and select any pattern from the category `Link In Bio`
- Click to switch devices
- Resize the window
- Make sure the pattern height is always at least 100% of the viewport height
- You will have to scroll to see the pattern content when the device is mobile or tablet
- Make sure the pattern is never cut off and you see all the pattern content when scrolling

## Pre-merge Checklist

- [X] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [X] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [X] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [X] Have you checked for TypeScript, React or other console errors?
- [X] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [X] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [X] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Issue #76020, opened 2023-04-20
[Spike] Site Assembler: Use a single iframe for the large preview


arthur791004:
Currently, we display the patterns in the large preview via the iframe respectively. However, if the user selects so many patterns, the performance might not be good. Thankfully, our metrics show that the average number of the selected patterns is about 6, so it's fine with the current implementation.

Besides, there are still some pros and cons as below so it might still be better if we can use a single iframe to eliminate the workaround

**Pros for one iFrame**
* Fix the gap between patterns. See https://github.com/Automattic/wp-calypso/issues/73072
* The background color on the gap might be delayed if you’re changing a color variation with different background color
* The better viewport height

**Cons (or pros for keeping multiple iframes)**
* Action bar implementation “easier”  to handle in individual iFrame
* Drag and drop implementation “easier” to handle in individual iFrame
* Could have a CSS workaround to remove the gap

arthur791004:
### Status Update on May 3

**Demo**

https://user-images.githubusercontent.com/13596067/234808652-8d72b896-9895-4cc1-8191-e41a64a7e20a.mov

**Done**

* We're able to use a single iframe for the large preview.
* We're able to stick the action bar on the hovered pattern when it's scrolled out of the viewport
* We're able to remove the workaround for the gap and background between the header, sections, and footer
* We're able to remove the workaround for the viewport height issue

**Next**

* The performance on the Firefox looks bad, we might need to investigate how to fix it.
* We can investigate whether there are other simpler solutions

arthur791004:
### Status Update on May 8

**Done**

* We have resolved the performance issue on Firefox

**Next**

* Resolve minor issues and ship https://github.com/Automattic/wp-calypso/pull/76312

arthur791004:
Closing it as we have finished the spike and will ship it later 🙂

PR #76312, opened 2023-04-26
Pattern Assembler: Use single iframe for the large preview

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/76020, https://github.com/Automattic/wp-calypso/pull/72364, https://github.com/Automattic/wp-calypso/pull/76132

## Proposed Changes

As https://github.com/Automattic/wp-calypso/issues/76020 mentioned, there are pros and cons when using a single iframe. This PR is focusing on using a single iframe to see whether the results look good or not.

**The concept of the current approach**

The main idea is to add the dummy overlay onto the hovered pattern on the large preview outside the iframe. So, we can reuse the existing components, styles, etc.

However, it's not easy to detect whether the cursor is leaving the hovered pattern. The reason is we want to keep hovered status when your mouse is over the action bar of the pattern but we cannot detect the actual target when the target is outside of the iframe (we will get `window` instead of the actual one).

[Update on May 4]

Attach the `mouseleave` event on the container of the large preview so we're able to detect whether the cursor is leaving out and we won't need to forward the scroll. Moreover, this way also resolves the performance issue on Firefox!

~So we did a trick to detect the `mouseleave` event on the dummy overlay. But if we did that, we are not able to scroll inside the iframe. Thankfully, Gutenberg has [usePopoverScroll](https://github.com/WordPress/gutenberg/blob/0154f86ed19502bb205a83f0f545058ca824c4c2/packages/block-editor/src/components/block-popover/use-popover-scroll.js#L13) hook to forward the scroll event from their toolbar to the content. Hence, we can simply mimic this approach to forward the scroll event to the iframe.~

With this approach, we can also fix the issue that [the action bar is not sticky to the pattern](pekYwv-1j1-p2) pekYwv-1j1-p2 🎉

Note
* ~It's a bit slow on Firefox~

**Alternatives**

Here are some possible alternatives. Free free to let me know if you have any other ideas 🙂

* Re-implement the styles of the action bar and active box shadow inside the iframe again. We might need to hard-coded something since we cannot get CSS variables inside an iframe, e.g. the border radius of the outer frame, colors, etc. Or we can try to [copy styles](https://bobbyhadz.com/blog/javascript-copy-all-styles-from-one-element-to-another) from one element to another but I'm not familiar with this one.
* Un-iframe if we don't need to fetch the external styles. Currently, if we are passing the inline styles to the iframe, those styles will be wrapped with the `editor-styles-wrapper` class name. Hence, it might be possible to remove the iframe and render the patterns directly.

**Demo**

https://user-images.githubusercontent.com/13596067/234808652-8d72b896-9895-4cc1-8191-e41a64a7e20a.mov

**Sticky action bar**

image

**Gap and Colors**

image

**Patterns with 100vh height**

| Before | After | Site Editor |
| - | - | - |
| image | image | image |
| image | image | image |
| image | image | image |

## Testing Instructions

* Go to /setup/site-setup?siteSlug=<your_free_site>
  * Select any header, sections, and footer. Ensure the gap between them looks good
  * Select the pattern with 100vh height, e.g. Link in Bio patterns. Ensure the preview looks good
  * Select any color variation. Ensure the background color looks good. Also, when switching between different color variations, verify the background color of the gap won't be flashing.
  * Hover and scroll the patterns on the large preview. Ensure the action bar and active box shadow works as expected
  * Reorder or remove the patterns via the action bar on the large preview. Ensure the action bar and active box shadow works as expected

## Pre-merge Checklist


okmttdhr:
I'm just asking as an idea, but is it possible to implement it without iframes at all? (I'd like to know if there're any technical limitations or reasons why we have to use iframe, if any.)

arthur791004:
> I'm just asking as an idea, but is it possible to implement it without iframes at all? (I'd like to know if there're any technical limitations or reasons why we have to use iframe, if any.)

@okmttdhr The large preview needs to load the styles of the theme and pattern, especially for the external styles. If we remove the iframe, there might be some global pollution on the styles and the result might look weird. We can investigate in the future whether we can replace all external styles with inline styles, and then we might have a way to wrap the styles into a specific class name to avoid any pollution.

packages/block-renderer/src/components/block-renderer-container.tsx
```diff
@@ -14,17 +14,17 @@ import BlockRendererContext from './block-renderer-context';
...
 	inlineCss?: string;
 	viewportWidth?: number;
 	viewportHeight?: number;
-	maxHeight?: 'none' | number;
+	maxHeight?: number;
 	minHeight?: number;
-	isMinHeight100vh?: boolean;
-	minHeightFor100vh?: number;
+	/** Whether to disable the pointer-events */
+	disabled?: boolean;
```
miksansegundo: Maybe better use `disablePointerEvents` and remove the comment?

I get that the prop `disabled` has the implicit meaning of disabling pointer events but maybe better be explicit about the goal of the prop.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview/pattern-overlay.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+interface Props {
+	referenceElement?: HTMLElement;
+	children?: JSX.Element | null;
+	stickyContent?: JSX.Element | null;
+}
+
+const STICKY_MARGIN = 72;
+
+const useOverlayRect = ( referenceElement?: HTMLElement ) => {
```
miksansegundo, line 14: Can we add a comment for each `useLayoutEffect` to help us understand their goal quickly?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview/use-action-bar-props.ts
```diff
@@ -0,0 +1,61 @@
...
+	onMoveDownSection,
+	onDeleteHeader,
+	onDeleteFooter,
+}: UseActionBarArguments ) => {
+	const patternType = element?.dataset.type ?? '';
+	const position = Number( element?.dataset.position ?? 0 );
+	const actionBarProps = { patternType };
+	const [ , forceRecompute ] = useReducer( ( s ) => ( s + 1 ) % Number.MAX_SAFE_INTEGER, 0 );
+
+	useLayoutEffect( () => {
```
miksansegundo, line 30: Can we add a comment to explain the goal the code in this `useLayoutEffect`?

Maybe rename `forceRecompute` to be `forceUpdate` as in the React docs: [Is there something like forceUpdate?](https://legacy.reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate)

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-category-list.tsx
```diff
@@ -48,31 +46,46 @@ const ScreenCategoryList = ( {
...
+					target.closest( '.pattern-action-bar' ) || target.closest( '.pattern-assembler__sidebar' )
+				)
+			) {
+				handleFocusOutside();
+			}
+		};
+
+		// Clicking the large preview iframe won't trigger the click event on the current window.
+		// Instead, the current window will blur and that iframe will be focused.
+		const handleBlur = () => {
```
miksansegundo: This workaround doesn't seems to work on Safari because the Pattern List doesn't close.

https://user-images.githubusercontent.com/1881481/236124190-df7223dc-05ed-48dd-86c2-ed9ca2dc3752.mov

miksansegundo (approved):
LGTM 👍 Nice job and clean-up!

Tested all the scenarios and worked fine except for closing the patterns list on Safari when we click on the iframe. I can debug that and follow-up with a PR if you want help.

I appreciate the comments in the code 🙌 . We still can add some more comments in some areas.

I've not noticed bad performance on Firefox. I'll create an issue to investigate that and we can share screen recordings to compare. I see some warnings tho:

- Layout was forced before the page was fully loaded. If stylesheets are not yet loaded this may cause a flash of unstyled content. [srcdoc](about:srcdoc)
- Error while traversing the CSS: TypeError: e is undefined
- Empty string passed to getElementById()
- Cookie “tk_qs” does not have a proper “SameSite” attribute value.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-category-list.tsx
```diff
@@ -48,31 +46,46 @@ const ScreenCategoryList = ( {
...
-	};
-
 	const trackEventCategoryClick = ( name: string ) => {
 		recordTracksEvent( PATTERN_ASSEMBLER_EVENTS.SCREEN_CATEGORY_LIST_CATEGORY_CLICK, {
 			pattern_category: name,
 		} );
 	};
 
 	useEffect( () => {
-		wrapperRef?.current?.addEventListener( 'click', handleFocusOutside );
```
miksansegundo, line 71: This approach works on Safari on production.

taipeicoder (approved):
Tested in Chrome, Firefox, and Safari, and they all worked as intended 🙂
I noticed that in Safari, after removing a pattern, the new pattern which the cursor is hover is not focused. This isn't the case with Chrome and Firefox, where the new pattern has the hover state. But this seems to be an existing problem, so it can be addressed later.

autumnfjeld:
> I'm just asking as an idea, but is it possible to implement it without iframes at all? (I'd like to know if there're any technical limitations or reasons why we have to use iframe, if any.)

@okmttdhr  To add to @arthur791004 's [comment](https://github.com/Automattic/wp-calypso/pull/76312#issuecomment-1532387572).  In theory, we could take a whole different approach, like what is done in the Theme Showcase: the live demo site which is navigable/clickable.   Sometimes I wonder if this is actually a better approach because of the interactive nature of the live demo

image
image

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview/pattern-overlay.tsx
okmttdhr: It'd be best to extract/organize hooks and add some tests if possible 👍

okmttdhr (approved):
LGTM, and kudos to you 👍 

>https://github.com/Automattic/wp-calypso/pull/76312#issuecomment-1532387572
>https://github.com/Automattic/wp-calypso/pull/76312#issuecomment-1535576097

Thank you for these comments!

I've come to understand that it might be "realistic" to implement it without using an iframe. So that can be useful sometimes.
I think it's good to encapsulate in iframe for now since I don't see any particular problems.

It'd also be good to consider other solutions, as Autumn said, in the future.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-category-list.tsx
arthur791004: I tested it but sometimes it doesn't work. So, I made a change to bubble the events from the element inside the iframe to its parent window

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview/pattern-overlay.tsx
arthur791004: Comment added! As this hook is for specific usage and only used by the component, I think we can keep it here now. I'd prefer to extract it to a single file and add some tests at that time when it's re-used in other places.

miksansegundo:
I retested on Safari and the pattern panel now closes as expected 🥳

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview/pattern-overlay.tsx
okmttdhr: badge
Just commenting so someone can look at this later. 🙂 (totally non-blocking)

I think it's worth implementing tests even if the hook is not re-used since we're using a little bit of a tricky way to position the sticky content, and it seems to be easy to implement for protecting the function.

The test code can be something like the below;

```ts
describe('PatternOverlay', () => {
  it('updates position on window resize', () => {
    // assuming you have a way to force useOverlayRect to return a defined value
    render(<PatternOverlay />);
    fireEvent(window, new Event('resize'));
    // assert that the updatePosition function was called
  });

  it('updates position on scroll', () => {
    // assuming you have a way to force useOverlayRect to return a defined value
    render(<PatternOverlay />);
    fireEvent(window, new Event('scroll'));
    // assert that the updatePosition function was called
  });
});
```

arthur791004:
Second try: https://github.com/Automattic/wp-calypso/pull/76750

PR #76317, opened 2023-04-26
Checkout: Clear the cart when attempting to add renewal fails

sirbrillig:
## Proposed Changes

This PR modifies checkout's loading process so that it will clear the cart if a renewal was requested but there are no valid renewals prepared. This is because if a renewal was requested by URL and was invalid, we don't want the user to see what may have been in their cart previously to avoid the appearance that the URL succeeded.

This was extracted from https://github.com/Automattic/wp-calypso/pull/75309

## Testing Instructions

TBD


sirbrillig:
Nevermind. Even though these changes are separate, it is too hard to prove that they work without the rest of https://github.com/Automattic/wp-calypso/pull/75309. Closing in favor of that.

PR #76336, opened 2023-04-27
Plugin Selectors: Guard against nulls in preparation for using TypeScript plugin selectors

robertf4:
## Proposed Changes

As part of splitting https://github.com/Automattic/wp-calypso/pull/72363 into multiple files, this PR adds a lot of small changes which check for nulls. These changes will eventually be necessary when the TypeScript versions of the plugin selectors are used, but for now they just serve to reduce ambiguity and avoid a few possible null reference errors.

I also removed the renaming of `localSlug` to `locale` in `client/my-sites/plugins/utils.js` due to a linting error in `i18n-translate-identifier` which tries to prevent the renaming of translation variables. However this code still does not pass the linter, so I opened https://github.com/Automattic/wp-calypso/issues/76339.

## Testing Instructions

For this PR it is best to review the code visually to ensure that it is either functionally equivalent to what existed before or avoids a null reference error.

If you want to do functional testing, just run a flow that loads each component and look for errors in the console.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plugins/utils.js
+++ b/client/my-sites/plugins/utils.js
@@ -43,7 +43,7 @@ const getConfirmationText = ( sites, selectedPlugins, actionText ) => {
 
 		Object.keys( plugin.sites ).forEach( ( siteId ) => {
 			const site = sites.find( ( s ) => s.ID === parseInt( siteId ) );
-			if ( site.canUpdateFiles ) {
+			if ( site && site.canUpdateFiles ) {
 				sitesList[ site.ID ] = true;
 				siteName = site.title;
 			}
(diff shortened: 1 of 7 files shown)


client/my-sites/plugins/utils.js
```diff
@@ -15,16 +15,16 @@ export function getVisibleSites( sites ) {
 
 export function useLocalizedPlugins() {
 	const isLoggedIn = useSelector( isUserLoggedIn );
-	const { localeSlug: locale } = useTranslate();
```
robertf4, line 18: This change is related to https://github.com/Automattic/wp-calypso/issues/76339.

Based on the spirit of the rule it seems that it does not want variables coming out of `useTranslate` to be renamed so that tools can always find them. However the rule does not account for this variable, so I opened an issue.

tyxla (approved):
LGTM 👍 

There seem to be a few warnings that may need to be fixed before shipping

client/my-sites/plugins/utils.js
tyxla: Feel free to temporarily explicitly ignore this one if it obstructs merging in the meantime.

PR #76349, opened 2023-04-27
Enable the Unified Importer on atomic

zaerl:
This PR enables the Unified Importer on atomic.

## Proposed Changes

* Enable the config flag
* The unified importer is enabled only if the `importer/unified` flag is on and the site is atomic
* Updated the minimum required version to 12.1, the version we are waiting to be deployed

## Testing Instructions

* `/import/URL` in an atomic site: all importers.
* `/import/URL` in a Jetpack self-hosted site: "Want to import into your site?" page.
* The Tools > Import link of the left menu in a Jetpack self-hosted site should have `/import/URL` endpoint.
* `/import/URL` in a simple site: all importers.

### DIFF
--- a/client/lib/importer/importer-config.tsx
+++ b/client/lib/importer/importer-config.tsx
@@ -259,15 +259,14 @@ function getConfig(
 		weight: 0,
 	};
 
+	// TODO: remove the atomic once Jetpack 12.1 is deployed.
+	const hasUnifiedImporter = config.isEnabled( 'importer/unified' ) && !! args.isAtomic;
+
 	// For Jetpack sites, we don't support migration as destination,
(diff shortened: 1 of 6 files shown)


ouikhuan (approved):
LGTM!

vishnugopal:
@zaerl We may want a PR that initially enables the feature only for Atomic sites (in case Jetpack isn't launched by our deadline).

vishnugopal (approved):
Approving again, tested with an Atomic site and got this: 
CleanShot 2023-05-02 at 14 51 05@2x

Tested with a Jetpack site and got this:
CleanShot 2023-05-02 at 14 51 31@2x

which seems ok!

PR #76077, opened 2023-04-21
chore: upgrade `react-query` (v3) to `@tanstack/react-query` (v4)

flootr:
**Second attempt**. We had to revert the first attempt (#74381) due to some regressions found.

## Proposed Changes

* Upgrade the `react-query@3` package to `@tanstack/react-query@4`
* Include any adjustments necessary because of breaking changes

## Testing Instructions

[Overview of changes between v3 and v4.](https://tanstack.com/query/latest/docs/react/guides/migrating-to-react-query-4)

**Please note**, this PR touches almost all queries and mutations for which we use the `react-query` library (now `@tanstack/react-query`) as well as a few other areas in the Calypso codebase (except for a few `/apps` directories which are not imported or don't import related code from Calypso or other packages.

* Verify Calypso builds well and does not contain any new type errors or warnings
* Depending on the query or mutation, verify data is fetched as expected (compare with prod) and no errors are thrown during runtime
* Verify the local persistence plugin still works as expected

**Additional steps** to verify regressions found during the first attempt of merging the update don't occur anymore:

- Verify the HelpCenter is working correctly & articles are shown
- Verify the signup flows works correctly (a failing pre-release e2e test showed there was an error when selecting a category the new site will be about at `/setup/site-setup/vertical`)

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76357, opened 2023-04-27
Stored cards: Do not return isLoading when logged-out

sirbrillig:
## Proposed Changes

The `useStoredPaymentMethods()` hook fetches stored cards using `react-query`, but sometimes we don't want to fetch them. In that case, the consumer sets the `isLoggedOut` prop, which disables the fetching.

Recently there was an upgrade to `react-query` in https://github.com/Automattic/wp-calypso/pull/76077 and it appears that one of the side effects was that when you disable a query, it leaves `isLoading` true instead of setting it to false. The result is that when `useStoredPaymentMethods()` is disabled, it seems to load forever. This prevents checkout from loading if the user is logged-out.

## Testing Instructions

- Visit the following URL in an incognito window: `/checkout/jetpack/jetpack_security_t1_yearly?source=jetpack-plans&checkoutBackUrl=https%3A%2F%2Fcloud.jetpack.com%2Fpricing`
- Verify that checkout loads.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
@@ -166,7 +166,7 @@ export function useStoredPaymentMethods( {
 
 	return {
 		paymentMethods,
-		isLoading,
+		isLoading: isLoggedOut ? false : isLoading,
 		isDeleting: mutation.isLoading,
 		error: errorMessage,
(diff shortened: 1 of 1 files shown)


michaeldcain (approved):
The works for me when visiting checkout as a logged-out user: http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1

sirbrillig:
https://github.com/Automattic/wp-calypso/pull/76372 adds a test to guard against this bug in the future.

PR #76358, opened 2023-04-27
LinkInBio Flow: Plan step to use new grid

chriskmnds:
Related to Automattic/martech#1533

## Proposed Changes

Updates the Link In Bio flow to use the new 2023 grid with expected features and copy. See: https://github.com/Automattic/martech/issues/1533#issuecomment-1495798000 

This is based on https://github.com/Automattic/wp-calypso/pull/75174 and will need a rebase. The plan type selector may appear broken, but it will be fixed in the base branch once updated.

### Media

Screenshot 2023-04-27 at 2 09 13 PM

## Testing Instructions

1. Go to `/setup/link-in-bio/plans`
2. Verify the new grid matches the media and mocks. Confirm functionality checks out
3. Go to `/start/plans` and `/plans/[SITE]` and ensure no regressions

## Pre-merge Checklist


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7952664

Thank you @chriskmnds for including a screenshot in the description! This is really helpful for our translators.

southp:
Thanks for getting ahead on this 👏🏼 
Once rebased, let's also use this chance to get rid of this TODO block: https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/plans-features-main/index.jsx#L448 :D

chriskmnds:
We are ready for a review here. @Automattic/martech-decepticons

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7952664).

client/landing/stepper/declarative-flow/link-in-bio.ts
```diff
@@ -46,6 +47,13 @@ const linkInBio: Flow = {
 		];
 	},
 
+	useSideEffect() {
+		const { setHidePlansFeatureComparison } = useDispatch( ONBOARD_STORE );
+		useEffect( () => {
+			setHidePlansFeatureComparison( true );
+		} );
+	},
```
aneeshd16, line 55: Why do we need to set this in a data store? If I understand correctly, it is a static property of the flow (like the flow name) and should be passed as a prop to the PlansWrapper component. Is there any use case where we might want to enable the comparison grid based on user action?

I see that it's the same case with the Newsletter flow as well, so consider this a non-blocking comment which we can tackle separately.

aneeshd16 (approved):
Left a code-related comment, but everything tests out. LGTM!

client/landing/stepper/declarative-flow/link-in-bio.ts
chriskmnds: I am not entirely sure, but I wouldn't object to going through the store (and haven't while reviewing). I think the onboard store is meant to serve things in a way that leads to less prop drilling. Otherwise, maybe most everything from that store can be passed around as props (probably not a great idea).

> Is there any use case where we might want to enable the comparison grid based on user action?

None of these are "user action" based. They are side effects to the root rendering.

aneeshd16: > None of these are "user action" based. They are side effects to the root rendering.

This action sets a value which is then fetched by the plans-wrapper. I was referring to these values, which are picked from the store:
https://github.com/Automattic/wp-calypso/blob/e7903abc2f7c3a2dd3102885df6602b49bd7db36/client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx#L63-L69

The values of `hideFreePlan` and `domainCartItem` depend on user action, but `hidePlansFeatureComparison` is a read-only value, which is a property of the flow itself. Typically something like this would be behind a `flowName` check, which IMO is flawed, but we don't have any better solutions yet. (pdDR7T-vO-p2)

> I think the onboard store is meant to serve things in a way that leads to less prop drilling.

It does, but I'm concerned if it would lead to the "dependency-soup" issue we have in the classic signup framework. Although, since Stepper flows are typed, it reduces that issue somewhat. 🤔

chriskmnds: > This action sets a value which is then fetched by the plans-wrapper. I was referring to these values, which are picked from the store:

Gotcha. It's a good point. It's almost like it becomes an implicit dependency on the flow right? (that it must set this value to get the grid showing). I don't know if it's a problem per se, but it does feel less of a "shared state" (at least across flows) than the rest. I think central dependency management wasn't a recipe for putting together Stepper, and I wouldn't disagree with that tbh. 

> It does, but I'm concerned if it would lead to the "dependency-soup" issue we have in the classic signup framework. Although, since Stepper flows are typed, it reduces that issue somewhat. 🤔

There may be a bit of that risk, but with everything typed and within React, definitely nowhere near classic signup. The architecture seems solid without a ton of over-engineering, so it remains to be seen how well everything is served (hopefully soon, classic signup? `/start`).

simison:
Convo came up that these grids with 3 plans should be narrower (768px) as per figma. Currently:

image

PR #76359, opened 2023-04-27
Starter designs: Remove virtual theme params

mmtr:
Counterpart of D109046-code

## Proposed Changes

The `starter-designs` endpoint no longer needs a param to generate virtual designs – they are always included now.

## Testing Instructions

- Make sure virtual designs are still visible.


BogdanUngureanu (approved):
LGTM and works as expected! 🚢

PR #76360, opened 2023-04-27
Update posts and recommended posts on reader search page

eoigal:
This PR plans to update the default listing of posts/sites on the search page to recommended posts.

This is part of the project - pe7F0s-Le-p2

**Design**
preview

At the moment, search already falls back to showing related/recommended posts if no search query is entered.
I have updated this to use the recommended sites API endpoint instead and pulled out the most recently liked post from that site.

This PR uses a new algorithm that will lookup a list of popular sites based on views and based on those results, return a list of recommended sites posts in 1 column and include a sidebar with the recommended sites list.

### Testing

* Needs patch; D109829-code
* Go to Reader search and confirm it loads with recommended posts/sites based on popular sites - http://calypso.localhost:3000/read/search

### DIFF
--- a/client/reader/search-stream/post-results.jsx
+++ b/client/reader/search-stream/post-results.jsx
@@ -34,7 +34,6 @@ class PostResults extends Component {
 			! query || query === ''
 				? ( postKey ) => ( { ...postKey, isRecommendation: true } )
 				: defaultTransform;
-
 		return (
 			<Stream
 				{ ...this.props }
--- a/client/reader/search-stream/style.scss
+++ b/client/reader/search-stream/style.scss
@@ -109,7 +109,6 @@
 	box-sizing: border-box;
 	border-bottom: 1px solid var(--color-neutral-10);
 	display: flex;
-	//flex-basis: calc(50% - 15px);
 	margin: 0 0 0 15px;
 	padding: 20px 0;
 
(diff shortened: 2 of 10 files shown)


Addison-Stavlo (commented):
Im realizing I may have jumped the gun on testing this as you noted still possibly making some changes in slack. But I did get a chance to test this briefly. It works well and I like the new format visually and it reads especially better on mobile than it did before. 

I like how the posts & sites become differentiated with tabs once we get down to smaller screen widths.  A couple minor things to point out in the mobile view:

1. The sites list in tab view (title and items) could probably use a small amount of padding on the left as the items feel uncomfortably aligned with the edge of the screen:

Screenshot 2023-05-05 at 9 50 57 AM

2. Im noticing an issue where switching back to "Posts" tab seems to have a default scroll position that is not the at the top of the list:



3. Im curious if there are any thoughts around making those tabs live outside the scrollbox so they are always visible at the top on mobile (similar with the sites sidebar on desktop). So that a user that has scrolled through posts can still switch gears and access sites easily without scrolling back to the top?

davemart-in (approved):
Looks great for a first pass. I think we can continue to fiddle with things to speed up the load time. Once you fix that last little padding thing I'd say ship it!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7969998

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7969998).

PR #76377, opened 2023-04-27
Checkout: Render loading info and support link if loading is slow

sirbrillig:
## Proposed Changes

This PR modifies the checkout loading page so that if it takes longer than a certain amount of time, the user will see additional text appear. This text will include a description of the current loading step and a link to contact support.

Screenshot 2023-04-27 at 5 48 37 PM

Screenshot 2023-04-28 at 2 31 01 PM

## Testing Instructions

It's easiest to see this by breaking checkout loading. One way to do that is to apply the following diff and then visiting `/checkout` in calypso on this branch.

```diff
--- a/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
+++ b/client/my-sites/checkout/composite-checkout/hooks/use-stored-payment-methods.tsx
@@ -166,7 +166,7 @@ export function useStoredPaymentMethods( {

        return {
                paymentMethods,
-               isLoading: isLoggedOut ? false : isLoading,
+               isLoading: true,
                isDeleting: mutation.isLoading,
                error: errorMessage,
                deletePaymentMethod: deletePaymentMethodAndSimilar,
```

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/checkout-loading-placeholder.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-loading-placeholder.tsx
@@ -9,7 +9,7 @@ export function CheckoutLoadingPlaceholder( {
 }: {
 	checkoutLoadingConditions: Array< { name: string; isLoading: boolean } >;
 } ) {
-	const showLoadingInfoThresholdMs = 1000;
+	const showLoadingInfoThresholdMs = 5000;
 	const [ shouldShowLoadingInfo, setShowLoadingInfo ] = useState( false );
 	useEffect( () => {
 		const timer = setTimeout( () => {
--- a/client/my-sites/checkout/composite-checkout/components/checkout-loading-placeholder.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/checkout-loading-placeholder.tsx
@@ -21,12 +21,12 @@ export function CheckoutLoadingPlaceholder( {
 	}, [] );
 
 	return (
-		<div>
+		<>
 			{ shouldShowLoadingInfo && (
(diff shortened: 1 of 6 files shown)


client/my-sites/checkout/composite-checkout/components/checkout-loading-placeholder.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useRef } from 'react';
+import { CALYPSO_CONTACT } from 'calypso/lib/url/support';
+
+export function CheckoutLoadingPlaceholder( {
+	checkoutLoadingConditions,
+}: {
+	checkoutLoadingConditions: Array< { name: string; isLoading: boolean } >;
+} ) {
+	const initialRenderTime = useRef( Date.now() );
+	const showLoadingInfoThresholdMs = 1000;
```
michaeldcain: I know they're not measuring the exact same point, but I think we should bump this to just above the average load time from our RUM logs - `4000` looks like it would clear for most people. Is that too delayed to be helpful for tests?

sirbrillig: Yeah, totally. I also need to update the code so that it starts a timer itself rather than currently relying on its parent to just happen to re-render it (in my experiments it currently takes about 5 seconds to show up because this code just says "if you're rendered and it's been more than 1 second, then show the notice").

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7960137

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7960137).

PR #76385, opened 2023-04-28
Add E2E tests for theme switching

okmttdhr:
Related to #

## Proposed Changes

This is to test permutations for the Live Preview spike.

Just by running the E2E test, it automatically performs a theme switch and saves a screenshot of the screen we want to confirm instead of manually performing a theme switch every time.

## Testing Instructions

- `yarn workspace wp-e2e-tests build`
- `NODE_ENV=test yarn workspace wp-e2e-tests test -- test/e2e/specs/onboarding/theme-switch.ts`

## Pre-merge Checklist

### DIFF
--- a/test/e2e/specs/onboarding/theme-switch.ts
+++ b/test/e2e/specs/onboarding/theme-switch.ts
@@ -12,13 +12,14 @@ jest.setTimeout( TIMEOUT_1_HOUR );
 
 describe( DataHelper.createSuiteTitle( 'Theme Switch' ), () => {
 	// twentyeleven
-	const FROM = 'hemingway-rewritten';
+	// const FROM = 'hemingway-rewritten';
 	// const FROM = 'appleton';
-	// const FROM = 'pendant';
+	const FROM = 'pendant';
 
 	// const TO = 'hemingway-rewritten';
 	// const TO = 'pendant';
-	const TO = 'appleton';
+	const TO = 'business';
+	// const TO = 'appleton';
 
 	let page: Page;
 	let selectedDomain: string;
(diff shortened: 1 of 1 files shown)


worldomonation:
@okmttdhr Should I give this PR a pass, given that it's still in Draft state?

okmttdhr:
@worldomonation I'm sorry for the confusion. This implementation is only for testing theme switch locally (rather than manual testing by a human), and this PR is for sharing for @Automattic/lego.
I might eventually implement E2E tests for theme switching, but I'll create another PR in that case.

worldomonation:
> @worldomonation I'm sorry for the confusion. This implementation is only for testing theme switch locally (rather than manual testing by a human), and this PR is for sharing for @Automattic/lego. I might eventually implement E2E tests for theme switching, but I'll create another PR in that case.

Okay, just to clarify, this is an experimental PR and so I or KitKat doesn't need to review this, correct?

----

Side note, there are a few things in this PR that may cause issues with other specs because it modifies the state of `defaultUser`. I can note them in the relevant sections.

okmttdhr:
@worldomonation 
>Okay, just to clarify, this is an experimental PR and so I or KitKat doesn't need to review this, correct?

There is no need for you or KitKat to review this!

>Side note, there are a few things in this PR that may cause issues with other specs because it modifies the state of defaultUser. I can note them in the relevant sections.

Oh, thank you for your pointing that out.
Should I, perhaps, use another user or reset the state of `defaultUser`?

worldomonation:
> Oh, thank you for your pointing that out. Should I, perhaps, use another user or reset the state of `defaultUser`?

I don't recall whether the `defaultUser` used a block-based or non-block-based theme, but it would be best to stick with the same type of themes for the time being.

The concern I had was that, if `defaultUser` is used in Editor-related testing, the tests would be written to expect only one type of the theme. In the past we've had issues where a test account's theme was accidentally switched to a type it did not expect and led to many simultaneous failures.

With that said, I did take stock and it seems there are no crossover between `defaultUser` and Editor-related tests, so I think we're in the clear. That means **no need for you to clear/reset the state** of the user for now.

If you're going to be doing a lot more experimental work, we can set you up with a specific `legoUser` of some sort.

okmttdhr:
I understand! Thank you for sharing!

> The concern I had was that, if defaultUser is used in Editor-related testing, the tests would be written to expect only one type of the theme

For this time, I create a new site for every test for theme switching, so I don't think it will affect the existing test (unless accidents happen).

>If you're going to be doing a lot more experimental work, we can set you up with a specific legoUser of some sort.

But I'd love to have `logeUser` if we can!

okmttdhr:
Closing as this E2E no longer makes sense due to the UI changes.

PR #76390, opened 2023-04-28
Site Logs: The correct log entry stays expanded after logs are refreshed

p-jackson:
**Note to translators**: the **Expand row** text is the accessible label given to this button
CleanShot 2023-05-01 at 11 12 02@2x

## Proposed Changes

The site logs table currently uses a simple index as the React key. This worked fine when the table was one giant component. But now that each row is its own component with its own state (which we changed when introducing the expand button) it means we need a proper key.

That's because it creates this issue:
- expand the first row
- do something that causes other logs to be recorded
- wait for the site logs to auto refresh
- see that the _new_ first row is expanded and the previously expanded log entry has been collapsed

The timestamp would do a good job as a unique ID, except that for the PHP logs it only goes down to the nearest second, and so there are usually multiple log entries with an identical timestamp. So the new key takes the order of logs with identical timestamps into account too.

Here's what the fixed table looks like.

https://user-images.githubusercontent.com/1500769/235061023-cfc710b4-748b-4b6d-95a9-6fafaa684c5e.mov

## Testing Instructions

- expand the first row
- do something that causes other logs to be recorded
- wait for the site logs to auto refresh
- see that the new rows are collapsed and the previously expanded log entries are still expanded

### DIFF
--- a/client/my-sites/site-logs/components/site-logs-table/index.tsx
+++ b/client/my-sites/site-logs/components/site-logs-table/index.tsx
@@ -23,11 +23,15 @@ export const SiteLogsTable = memo( function SiteLogsTable( {
 	const columns = useSiteColumns( logs );
 	const siteGmtOffset = useCurrentSiteGmtOffset();
 
-	if ( isLoading && ! logs?.length ) {
+	const logsWithKeys = useMemo( () => {
+		return generateRowKeys( logs );
+	}, [ logs ] );
+
+	if ( isLoading && ! logsWithKeys.length ) {
 		return <Skeleton />;
 	}
 
-	if ( ! isLoading && ! logs?.length ) {
+	if ( ! isLoading && ! logsWithKeys.length ) {
 		return <>{ __( 'No log entries within this time range.' ) }</>;
 	}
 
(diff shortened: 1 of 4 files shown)


mpkelly (approved):
Looks good. I reproduced the issue and confirmed it's fixed.

I removed a `React` import that WebStorm was complaining about. It shouldn't be needed and tests are still are passing. 

image

vykes-mac (approved):

p-jackson:
Thanks @mpkelly

> I removed a `React` import that WebStorm was complaining about.

I think there must be something wrong with the TS config of the unit test files. You're right that it shouldn't be necessary, so I think WebStorm is correct. But I _do_ get an error about it in VS Code 🤷 But other unit test files give me similar errors so it hasn't been anything other than an annoyance.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7958960

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7958960).

PR #75947, opened 2023-04-19
AT E2E: use separate comments for like and unlike operations.

worldomonation:
Fixes https://github.com/Automattic/wp-calypso/issues/75669.

## Proposed Changes

This PR proposes a rethink of how we perform the Likes: Comments spec.

Until this point, the spec has used the same comment to test both like and unlike operations.
This was carried over from the Selenium-based framework where every interaction was prohibitively slow.

This is a problem for Playwright as the Like/Unlike button has a fade in/out animation which is unreliable. AT sites add further complication on top in the form of iframes.

With the ability to set up the test environment programmatically, it may make sense to use two separate comments, one for like and one for unlike, set the state prior to the test, then execute the click actions and the validation afterwards.

## Testing Instructions

Ensure the following build configurations are passing:
  - [x] Gutenberg E2E AT prod (desktop)
  - [x] Gutenberg E2E AT edge (desktop)
  - [x] Gutenberg E2E Simple prod (desktop)
  - [x] Gutenberg E2E Simple edge (desktop)

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Issue #75952, opened 2023-04-19
AT: comments do not immediately become available, sometimes returning a HTTP 404.


worldomonation:
### Quick summary

If a comment is left on am AT site and a script or user wants to interact with the comment immediately (eg. like the comment), the API can return a HTTP 404.

### Steps to reproduce

1. open two WordPress.com console windows.
2. submit a new comment on an existing post. (eg. /sites/${ siteID }/posts/${ postID }/replies/new)
3. immediately attempt to like the comment produced in Step 2 using an API call. (eg. /sites/${ siteID }/comments/${ commentID }/likes/new)

### What you expected to happen

The API responds with a success.

### What actually happened

In E2E suites:

```
ERROR: Likes: Comment > beforeAll

 Error: Failed to like 634 on site 204039232
    at RestAPIClient.commentAction (packages/calypso-e2e/src/rest-api-client.ts:799:10)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at test/e2e/specs/published-content/likes__comment.ts:73:3
```

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Platform (Simple and/or Atomic)

Atomic

### Logs or notes

The workaround is to simply wait a few seconds (like a real user would) in order to let the comment "settle" into place.

PR #76414, opened 2023-04-28
E2E: Try to stabilize Atomic likes (again)

WunderBart:
Related: https://github.com/Automattic/wp-calypso/pull/75947

## Proposed Changes

When adding a comment on an Atomic site, it takes a bit until that comment is settled. Sometimes, when we send the like request immediately after creating a comment, we get an `unknown_comment` error response. To address that, let's make a couple `like` request retries before failing the test.

## Testing Instructions

```
TEST_ON_ATOMIC=true yarn workspace wp-e2e-tests test likes__comment.ts
```

### DIFF
--- a/packages/calypso-e2e/src/lib/components/comments-component.ts
+++ b/packages/calypso-e2e/src/lib/components/comments-component.ts
@@ -105,9 +105,8 @@ export class CommentsComponent {
 			// See the like() method for info on the following method call.
 			await waitForWPWidgetsIfNecessary( this.page );
 
-			const likeButtonFrame = this.page
-				.frameLocator( `iframe[name^="like-comment-frame"]:below(:text("${ comment }"))` )
-				.first();
+			const commentContent = this.page.locator( '.comment-content', { hasText: comment } );
+			const likeButtonFrame = commentContent.frameLocator( "iframe[name^='like-comment-frame']" );
 
 			unlikeButton = likeButtonFrame.getByRole( 'link', { name: 'Liked by you' } );
 			unlikedStatus = likeButtonFrame.getByRole( 'link', { name: 'Like' } );
@@ -119,7 +118,6 @@ export class CommentsComponent {
 		}
 
 		await this.page.getByText( comment ).scrollIntoViewIfNeeded();
-		await unlikeButton.waitFor();
 		await unlikeButton.click();
 		await unlikedStatus.waitFor();
 	}
--- a/test/e2e/specs/published-content/likes__comment.ts
+++ b/test/e2e/specs/published-content/likes__comment.ts
@@ -67,7 +67,7 @@ describe( 'Likes: Comment', function () {
 		// The comment takes some time to settle. If we request the like
 		// immediately we might be getting the `unknown_comment` error. Let's do
 		// a few retries to make sure the like is getting through.
-		let likeRetryCount = 5;
+		let likeRetryCount = 10;
 		const likeRetryTimer = setInterval( async () => {
 			try {
 				await restAPIClient.commentAction(
--- a/test/e2e/specs/published-content/likes__comment.ts
+++ b/test/e2e/specs/published-content/likes__comment.ts
@@ -38,7 +38,7 @@ describe( 'Likes: Comment', function () {
 	let commentToBeUnliked: NewCommentResponse;
 	let restAPIClient: RestAPIClient;
 
-	beforeAll( async function () {
+	it( 'Setup the test', async function () {
 		page = await browser.newPage();
 
 		testAccount = new TestAccount( accountName );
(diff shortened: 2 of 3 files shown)


packages/calypso-e2e/src/lib/components/comments-component.ts
```diff
@@ -105,9 +105,8 @@ export class CommentsComponent {
 			// See the like() method for info on the following method call.
 			await waitForWPWidgetsIfNecessary( this.page );
 
-			const likeButtonFrame = this.page
-				.frameLocator( `iframe[name^="like-comment-frame"]:below(:text("${ comment }"))` )
-				.first();
+			const commentContent = this.page.locator( '.comment-content', { hasText: comment } );
+			const likeButtonFrame = commentContent.frameLocator( "iframe[name^='like-comment-frame']" );
```
WunderBart, line 109: Just using some modern syntax here, not a crucial change.

test/e2e/specs/published-content/likes__comment.ts
```diff
@@ -38,7 +38,7 @@ describe( 'Likes: Comment', function () {
 	let commentToBeUnliked: NewCommentResponse;
 	let restAPIClient: RestAPIClient;
 
-	beforeAll( async function () {
+	it( 'Setup the test', async function () {
```
WunderBart, line 41: Changed it to `it` because the title gives a bit more info immediately when looking at the error in CI.

WunderBart:
@worldomonation I think there's something wrong with the comment `like` request. The [request retries](https://github.com/Automattic/wp-calypso/pull/76414/files#diff-610fae8bc8ff71446ff0d48fee7ceed70ca894e44b4d30e959d45d88d741c019R67-R84) ensure that the response is what we expect it to be - a liked comment with the given ID. Even with that out of the way, that very comment doesn't always render liked❗ Maybe we should look a bit further into this? Maybe it's some caching issue? 

/cc @fullofcaffeine 

---

Edit: I was right 😄 

This is what the like response looks like after a few retries:

First try:
```js
{
  error: 'unknown_comment',
  message: 'Unknown comment'
}
```

A few retries later. Notice how the comment is seemingly liked (`i_like: true`), but the `like_count` is still `0`.

```js
{
  success: true,
  i_like: true,
  like_count: 0,
}
```

A couple more retries later, finally, a correct response. The comment is liked by us, and the count is now correctly at 1. 😌

```js
{
  success: true,
  i_like: true,
  like_count: 1,
}
```

This is quite a doozy to me. You can clearly see the payload inconsistency now. Possibly worth reporting to the AT team?

---

Another edit:

So the interesting part right now is that the response can get stuck at `like_count: 0` even after 30 retries (with 1s delay between). This is something that we might not be able to handle and which will keep making this spec flaky. Definitely, something to tackle with the AT team at this point?

packages/calypso-e2e/src/lib/components/comments-component.ts
worldomonation: I had an experimental branch where I made a similar change, so I can vouch it works.

test/e2e/specs/published-content/likes__comment.ts
```diff
@@ -64,22 +64,26 @@ describe( 'Likes: Comment', function () {
...
-		// unless time is given for the comment to "settle" in place.
-		// It could be argued that adding this arbitrary delay is
-		// "more representative" of users.
-		// @see: https://github.com/Automattic/wp-calypso/issues/75952
-		if ( envVariables.TEST_ON_ATOMIC ) {
-			await page.waitForTimeout( 5 * 1000 );
+		// The comment takes some time to settle. If we request the like
+		// immediately we might be getting the `unknown_comment` error. Let's do
+		// a few retries to make sure the like is getting through.
+		const likeRetryCount = 10;
```
worldomonation, line 70: **comment:non-blocking** - I have a WIP branch locally to try to stabilize this test as well, and I tried to extract the retry mechanism to `ElementHelper` and ran into difficulties generalizing the parameter definition and forgot to ask @dpasque whether generics would help with the scenario.

What I had was the following:
```typescript
// In the test itself
		if ( envVariables.TEST_ON_ATOMIC ) {
			ElementHelper.retryRequest(
				restAPIClient.getComment(
					testAccount.credentials.testSites?.primary.id as number,
					commentToBeLiked.ID
				)
			);
		}
```

```typescript
// in ElementHelper
export async function retryRequest< T, P >(
	func: () => Promise< void >,
	parameters: [ P ],
	{ timeout = 5000 }: { timeout?: number } = {}
): Promise< void > {
	for ( let retries = 3; retries > 0; retries -= 1 ) {
		try {
			return await func( ...parameters );
		} catch ( err ) {
			// Throw the error if final retry failed.
			if ( retries === 1 ) {
				throw err;
			} else {
				await new Promise( ( resolve ) => setTimeout( resolve, timeout ) );
			}
		}
	}
	return;
}
```

worldomonation (approved):

worldomonation:
> So the interesting part right now is that the response can get stuck at `like_count: 0` even after 30 retries (with 1s delay between). This is something that we might not be able to handle and which will keep making this spec flaky. Definitely, something to tackle with the AT team at this point?

This is definitely strange.

Previously I've filed https://github.com/Automattic/wp-calypso/issues/75952 for the first situation you've covered, where the comment is posted but the API returns a 404 when attempting to access the comment. 

I did waffle a bit on whether to expand the scope of https://github.com/Automattic/wp-calypso/issues/75952 to cover the later two situations you've mentioned, but have elected to file a new issue to cover the situation for `like_count` and `i_like` being out of sync.

Filed https://github.com/Automattic/wp-calypso/issues/76444.

worldomonation:
I've left a message in the AT channel, but not sure if it's the right place. p1682922721277819-slack-C7YPW6K40

test/e2e/specs/published-content/likes__comment.ts
WunderBart: If we only need the retry logic in one place, I don't think we don't need to worry about turning it into a helper. Is there any other place that it'd currently be useful?

WunderBart: I'd even say that this should generally not be encouraged, so I'd actually avoid creating a helper for it. We need this here because of a faulty app logic which, ultimately, should be addressed somewhere else.

PR #76415, opened 2023-04-28
Vdimitrakis/selfserve 410/campaign details page design

therocket-gr:
This is a PR for the Blaze new design for the campaign details page .

Related to SELFSERVE-410 ticket

## Proposed Changes

please see the screenshot below

Screenshot 2023-04-28 at 14 55 54

There are some TODOs in the code...

1. we are waiting for the API endpoint that will get a specific campaign id (at the moment the endpoint will fetch all the objects of all campaigns data ) 
2. when the above endpoint is complete we need to implement a hook to fetch the data using the campaign id coming from the url param
3. at the moment the mockup page is using the mockData.json to get the campaign object - please delete this file (and the import) when steps 1 and 2 are complete
4. we need to add a clicks_organic value in the data coming from the DSP api - this is a new param that we didnt have in the past
5. there is an TODO with a hardcoded value (that reads "Delivered 192% more than estimated").. i am not sure how to calculate this 

## Testing Instructions

*

## Pre-merge Checklist


jjolmo:
I've updated the PR and removed the WIP state

-Updated the rest of the missing elements (promote button, post insights, some final design changes)
- Created responsive version

client/my-sites/promote-post-i2/components/campaign-item-page/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+		comment: 'Validation error when filling out domain checkout contact details form',
+	}
+);
+
+export default function CampaignItemPage( props: Props ) {
+	const { campaignId } = props;
+	const translate = useTranslate();
+	const siteId = useSelector( getSelectedSiteId ) || 0;
+	const campaignQuery = useCampaignsQueryNew( siteId || 0, campaignId );
+	const { isModalOpen, selectedSiteId, selectedPostId, keyValue } = usePromoteParams();
```
sbarbosa: Please use this function instead, to open the campaign widget:
```
const onClickPromote = useOpenPromoteWidget( {
		keyValue: 'post-0',
		entrypoint: 'promoted_posts-header',
} );
```

That hook already takes into consideration if you are opening the widget from Jetpack.

client/my-sites/promote-post-i2/components/campaign-item-page/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+	if ( campaignsIsLoading ) {
+		return null;
+	}
+
+	return (
+		<Main wideLayout className="promote-post-i2">
+			<DocumentHead title={ translate( 'Advertising - Campaign details' ) } />
+			<CampaignItemDetails siteId={ +siteId } campaign={ campaign } />
+
+			{ selectedSiteId && selectedPostId && keyValue && (
```
sbarbosa: You won't need to include the widget here, if you call the `onClickPromote` function from my last comment, the widget will be loaded without the need to include it in your page.

client/my-sites/promote-post/index.js
```diff
@@ -42,4 +43,14 @@ export default () => {
 		makeLayout,
 		clientRender
 	);
+
+	page(
```
sbarbosa, line 47: We will also need this in the i2 controller. The Blaze app uses that controller, and in the future, we are planning to remove this one.

sbarbosa (commented):
I added a couple of comments on the code, some of them we need to change because of the Blaze app compatibility.

Apart from that, here is some minor things I found while testing the page:
* Clicking the Advertising link in the header redirects you to your primary blog post page. We need to change this to consider your current blog, and I think it could be better if it redirects you to the campaigns list page (like a back button).

* In the header, the label next to the status should be using gray color

* The hover in the Traffic breakdown section doesn't have a round border, and we are missing a padding in the count element. Right now it get stuck to the border and doesn't look very nice
Screenshot 2023-06-07 at 13 57 23

* The underline in the header doesn't fix the full width of the page. We may need to change this after we merge both the widescreen Pr and this one. So feel free to ignore this.

* In the support articles section, the label should be bold, and the "Powered by Blaze" have a margin-bottom separating it from the other links (according with the design) 

* The (i) next to the 300x250 in the Ad preview section doesn't show a tooltip. 
Screenshot 2023-06-07 at 15 13 55

client/my-sites/promote-post-i2/components/ad-preview/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export default function AdPreview( { htmlCode, isLoading }: Props ) {
+	const shadowHost = useRef< HTMLDivElement >( null );
+
+	useEffect( () => {
+		if ( ! isLoading && htmlCode && shadowHost.current ) {
+			const shadowRoot = shadowHost.current.attachShadow( { mode: 'open' } );
+			shadowRoot.innerHTML = `
+          <div class="viewer viewer--receipt}">
```
timur987: I am not an expert in class naming but I suppose that `viewer--receipt` should be either `viewer__receipt` or `viewer-receipt`. I tend to rely on [BEM](https://en.bem.info/methodology/naming-convention/) methodology and Calypso [guidelines](https://github.com/Automattic/wp-calypso/blob/HEAD/docs/coding-guidelines/css.md).

What do you think?

client/my-sites/promote-post-i2/components/campaign-item-details/style.scss
```diff
@@ -0,0 +1,459 @@
+@import "../../../../../node_modules/@wordpress/base-styles/breakpoints";
```
timur987: It seems that the import statement is dependent to relative path. Using of this might be a bit more safe (in the case of moving the component to a different folder:
```js
```

client/my-sites/promote-post-i2/components/campaign-item-page/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				{ noCampaignListMessage }
+			</Notice>
+		);
+	}
+
+	if ( ! campaign || campaignsIsLoading ) {
+		return <CampaignItemDetails isLoading={ true } />;
+	}
+
+	if ( campaignsIsLoading ) {
```
timur987: Do we need this "if" condition? I suppose that we will handle the case of loading by the previous "if" condition.

campaigns-is-loading-null

client/my-sites/promote-post-i2/components/campaign-item-page/style.scss
```diff
@@ -0,0 +1,183 @@
+@import "@automattic/color-studio/dist/color-variables";
+@import "@automattic/components/src/styles/typography";
+
+@mixin promote-post-i2-font {
```
timur987: Sorry for nitpicking, @therocket-gr. I noticed that this file contains copy/paste from `client/my-sites/promote-post-i2/style.scss` file. We can probably add something like
```css
@import "../../style.scss";
```

and leave styles that are different from the original [`style.scss`](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/promote-post-i2/style.scss). For sure, we can postpone this task and will do it later. Just wanted to highlight my intention to remove repetition in CSS.

timur987 (approved):
Thanks, @therocket-gr!

The changes LGTM and despite I have some suggestions (like the one related to CSS), I do not see any blockers to merge it and fix the rest of suggestions in following PRs (assuming you will apply suggestions from @sbarbosa, they seem to be important).

Just to be mentioned (we can consider it in the following PR either), I've also found several differences between the implementation and Figma:
Screenshot 2023-06-08 at 12 50 59

client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
```diff
@@ -0,0 +1,628 @@
...
+					fill="#1E1E1E"
+				/>
+			</svg>
+		</span>
+	);
+
+	const databars = [
+		{
+			label: translate( 'Ad' ),
+			value: clicks_total || 0,
```
timur987, line 177: We use these bars to show impressions:
image

But there we show clicks. Should be use `impressions_total || 0` instead?
(Might be wrong because the intention might be in showing redirects from clicks versus organic traffic to the target site)

client/my-sites/promote-post-i2/components/campaign-item-page/style.scss
jjolmo: I'm not sure. Remember that we will delete the whole /promote-post/ folder, so we should not extend any file from there. It should be replicated completely. Once we go live we just remove the folder /promote-posts/

client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
jjolmo: Well, not sure about that because the current implementation makes "sense" in a way (unless I failed understanding it)
You are comparing organic views (the ones that come from any source excluding from the ad) VS visits that are coming from the ad
The ad views are actually clicks. So 1 click in the ad = 1 visit to the post

Remember that 1 impression is just an ad being displayed to the user. It is not the action of that user clicking into the ad and seeing the page itself, so we shouldn't use impressions_total there

jjolmo:
> Just to be mentioned (we can consider it in the following PR either), I've also found several differences between the implementation and Figma:

Related to this: I have this label explaining the delivery, but we will only show it when there's overdelivery. It won't be displayed otherwise

For the "operating systems" I think you're right, but we should display "Devices" since there's a different targeting that is actually called "operating systems" (We are not using it)
But I think this is covered in "Audience".

It's a bit confusing, we might need to check the labels

client/my-sites/promote-post-i2/components/campaign-item-page/index.tsx
sbarbosa: I got confused in this point, I think it may be needed because this is a new page (not like the tabs that are all in the same page). 

I will test this locally to be sure that is needed 👍

client/my-sites/promote-post/index.js
jjolmo: I saw that the /promote-post-i2/ page definitions seemed to me that we weren't using them. Not 100% sure, but otherwise the page wouldn't be opening.

But yes, you're right, we need to keep a copy of the i1 into i2 folder

client/my-sites/promote-post-i2/components/campaign-item-page/style.scss
timur987: The file I suggest is `promote-post-i2/style.scss`, so it's quick unlikely that we will delete it.

client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
timur987: Got it, I was wrong.

client/my-sites/promote-post-i2/components/campaign-item-page/index.tsx
sbarbosa: This is necessary; in fact, we have a bug in the campaigns page because the widget is not rendered there.
I pushed a commit to this branch fixing that, now we have a common wrapper component that the pages use, and that component adds the body class and the widget.

client/my-sites/promote-post-i2/components/campaign-item-page/style.scss
sbarbosa: I made the change to import it from the root style like we have in other components. 👍

sbarbosa:
Ok, I addressed all the comments that we made in this PR. 

@timur987, @jjolmo, I had to change the location of the `delivery_percent` parameter because the backend is returning it in the root, not inside the campaign_stats object. 

I tested by hardcoding a delivery above 100% and it showed the message correctly (I had to append a % symbol next to it that was missing).

**I will detail what I changed:**
* Created a new MainWrapper component that host the logic for rendering the widget and adding the body class, this fixes an issue we have if you click the promote button from the campaigns page, and also unify all the code inside one place.
* Fixed all the style comments
* Fixed the render of the page for widescreens
* Fixed the header redirection links (now all redirections use the utility function for compatibility with Jetpack)
* Fixed the location of the `delivery_percent` attribute

The only thing that is missing is the tooltip next to the ad image size. We can take care of that in another PR when we get feedback from Bart.

I think that now is ready to be retested.

sbarbosa (approved):

client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
```diff
@@ -0,0 +1,636 @@
...
+	const countriesList = audience_list ? audience_list[ 'countries' ] : '';
+	const topicsList = audience_list ? audience_list[ 'topics' ] : '';
+	const OSsList = audience_list ? audience_list[ 'OSs' ] : '';
+
+	// Formatted labels
+	const ctrFormatted = clickthrough_rate ? `${ clickthrough_rate.toFixed( 2 ) }%` : '-';
+	const durationFormatted = getCampaignDurationFormatted( start_date, end_date );
+	const totalBudgetFormatted = `$${ formatCents( total_budget || 0 ) }`;
+	const totalBudgetLeftFormatted = `$${ formatCents( total_budget_left || 0 ) } ${ __( 'left' ) }`;
+	const overallSpendingFormatted = `$${ formatCents( overall_spending || 0 ) }`;
```
timur987: I checked the field `overall_spending` in A8C DSP and think that it's not used there anymore.

In opposition, I've found that in API response we have a field `total_budget_used`. I assume this field should be used instead.

([source](https://github.tumblr.net/Tumblr/a8c-dsp/blob/trunk/dsp/src/api/routes/v1/api/campaigns.route.ts#L498))
image

Here's how a response from the API look like for me (there's no field `overall_spending` there):
image

timur987 (approved):
Thanks for the feature @therocket-gr and for the final tweaks @sbarbosa!

Overall it looks great now, but I have a question about `overall_spending` field I left below.
Rather than that, it looks fabulous!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8041163

Thank you @therocket-gr for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8041163).

PR #72674, opened 2023-01-26
Users management: Replace the way of querying invites by relying on useQuery

bogiii:
#### Proposed Changes

Replaced the way of querying invites by relying on `useQuery`.
`useQuery` by default triggers a query on the window focus event, which matches the way of getting data for the Teams content.

#### Testing Instructions

* Go to `/people/team/{SITE_SLUG}`
* Invite your testing account
* Check if there is a newly created pending invite
* In another tab with the test user, accept the invite
* Go back to the initial tab
* Check if a test member is part of the `Team` block
* Check if a test member disappears from the `Pending Invites` block

More info about the issue we fixed in this PR: https://github.com/Automattic/wp-calypso/issues/72636

#### Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Closes https://github.com/Automattic/wp-calypso/issues/72636

PR #76416, opened 2023-04-28
Invites: Fix undefined query data warning

tyxla:
## Proposed Changes

This PR fixes the following warning when visiting `/people/team/{SITE_SLUG}`:

Screenshot 2023-04-28 at 16 25 42

Seems to have been revealed with the v4 upgrade in #76077.

According to [the docs](https://tanstack.com/query/v4/docs/react/reference/useQuery): the `queryFn`:

> Must return a promise that will either resolve data or throw an error. The data cannot be undefined.

We [don't really utilize the data](https://github.com/Automattic/wp-calypso/blob/391f81bea18051ef155398fe2de4aba51455ebe4/client/my-sites/people/team-invites/index.tsx#L26) from this query, since the query is only used to trigger a Redux fetch of the invites. Therefore, to fix the error, we're returning a `null` to both satisfy the non-undefined requirement and still keep returning nothing from the query.

See #72674 which introduced this query.

## Testing Instructions

* Head to `/people/team/{SITE_SLUG}`
* Verify you no longer see the error message.

### DIFF
--- a/client/data/invites/use-get-invites-query.ts
+++ b/client/data/invites/use-get-invites-query.ts
@@ -13,6 +13,7 @@ const useGetInvitesQuery = ( siteId: number ) => {
 		if ( siteId && ! requestingInProgress ) {
 			dispatch( requestSiteInvites( siteId ) );
 		}
+		return null;
 	} );
 };
 


flootr (approved):
Thanks! 👍🏻

PR #76419, opened 2023-04-28
localizeUrl: option to preserve trailing slash variation

sgomes:
The motivation behind this PR is to fix the `Log In` link in logged-out theme pages.

image

As of right now, clicking it results in the URL location changing, but nothing else happening. The bug seems to be caused by the fact that the login URL has a trailing slash, which the routing doesn't particularly like.

In order to fix this, a new option needed to be added to `localizeUrl`, so that the chosen trailing slash variation (yes/no) for a link can be preserved.

This PR adds the new option and updates `wpcom-template-parts` to use it, where appropriate.

CC @josephscott and @mreishus, who also looked at the bug. 

## Proposed Changes

* Add new `preserveTrailingSlashVariation` option to `localizeUrl` and `useLocalizeUrl`.
* Modify `wpcom-template-parts` to use the new option where appropriate
* Add unit tests for the new option

## Testing Instructions

* Ensure the old and new `i18n-utils` tests work
* Modify `packages/wpcom-template-parts/src/universal-header-navigation/index.tsx` (L263) to localise the URL `'/log-in'` instead of `'//wordpress.com/log-in'` (it needs to match the origin for the bug to manifest)
* Start a dev build
* Open a theme page while logged out (e.g. `/theme/verso`)
* Click the `Log In` link at the top
* Ensure that you get redirected to the login page

## Pre-merge Checklist

### DIFF
--- a/packages/i18n-utils/src/localize-url.tsx
+++ b/packages/i18n-utils/src/localize-url.tsx
@@ -170,14 +170,12 @@ export const urlLocalizationMapping: UrlLocalizationMapping = {
 };
 
 function hasTrailingSlash( urlString: string ) {
-	let url;
 	try {
-		url = new URL( String( urlString ), INVALID_URL );
+		const url = new URL( String( urlString ), INVALID_URL );
+		return url.pathname.endsWith( '/' );
 	} catch ( e ) {
 		return false;
 	}
-
-	return url.pathname.endsWith( '/' );
 }
 
 export function localizeUrl(
--- a/packages/i18n-utils/src/localize-url.tsx
+++ b/packages/i18n-utils/src/localize-url.tsx
@@ -169,10 +169,22 @@ export const urlLocalizationMapping: UrlLocalizationMapping = {
 	},
 };
 
+function hasTrailingSlash( urlString: string ) {
+	let url;
+	try {
+		url = new URL( String( urlString ), INVALID_URL );
+	} catch ( e ) {
+		return false;
+	}
+
+	return url.pathname.endsWith( '/' );
+}
+
 export function localizeUrl(
 	fullUrl: string,
 	locale: Locale = getDefaultLocale(),
-	isLoggedIn = true
+	isLoggedIn = true,
+	preserveTrailingSlashVariation = false
 ): string {
 	let url;
 	try {
@@ -191,6 +203,10 @@ export function localizeUrl(
 
 	if ( ! url.pathname.endsWith( '.php' ) ) {
 		// Essentially a trailingslashit.
+		// We need to do this because the matching list is standardised to use
+		// trailing slashes everywhere.
+		// However, if the `preserveTrailingSlashVariation` option is enabled, we
+		// remove the trailing slash at the end again, when appropriate.
 		url.pathname = ( url.pathname + '/' ).replace( /\/+$/, '/' );
 	}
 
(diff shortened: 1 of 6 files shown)


packages/i18n-utils/src/localize-url.tsx
```diff
@@ -169,10 +169,22 @@ export const urlLocalizationMapping: UrlLocalizationMapping = {
...
 
+function hasTrailingSlash( urlString: string ) {
+	let url;
+	try {
+		url = new URL( String( urlString ), INVALID_URL );
+	} catch ( e ) {
+		return false;
+	}
+
+	return url.pathname.endsWith( '/' );
```
tyxla: Nit: I might be missing something, but we should be able to just use a constant and return immediately, no?

```suggestion
	try {
		const url = new URL( String( urlString ), INVALID_URL );
		return url.pathname.endsWith( '/' );
	} catch ( e ) {
		return false;
	}
```

packages/wpcom-template-parts/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@automattic/wpcom-template-parts",
-	"version": "1.0.0",
+	"version": "1.0.1",
```
tyxla: Is this really necessary? Has this package ever been released? Looks like it's intended to be private.

packages/i18n-utils/src/test/localize-url.js
```diff
@@ -91,12 +91,35 @@ describe( '#localizeUrl', () => {
 	} );
 
 	test( 'trailing slash variations', () => {
-		expect( localizeUrl( 'https://automattic.com/cookies/', 'de' ) ).toEqual(
+		const loggedIn = false;
```
tyxla: Should this be renamed to reflect the actual arg name?

tyxla (approved):
Tests well, code looks good too! 👍 

Perhaps @Automattic/i18n folks will have something to add, but this looks good to go IMHO. 🚀 

I've added a few minor optional suggestions, nothing blocking.

yuliyan (approved):
Tested - works as expected 👍

I just wonder if this behavior of the `page` client router is how it's supposed to work or a bug?

sgomes:
> I just wonder if this behavior of the `page` client router is how it's supposed to work or a bug?

That's a good question 🤔 Any idea, @tyxla?

packages/wpcom-template-parts/package.json
sgomes: Good point, reverted 👍

packages/i18n-utils/src/test/localize-url.js
sgomes: Yup, I can definitely do that 👍

packages/i18n-utils/src/localize-url.tsx
sgomes: Nope, apparently I was the one missing something 😄 Done!

tyxla:
> > I just wonder if this behavior of the `page` client router is how it's supposed to work or a bug?
> 
> That's a good question 🤔 Any idea, @tyxla?

Its a bit nasty TBH. From what I understand, page.js recognizes the link as an internal one and tries to treat it as part of the app, but there is no existing route configured to handle that request, so nothing happens when the URL changes. So it's not really a page.js bug, it's more like an unintentional page.js usage. We don't want `page.js`working for those links. If it wasn't for the trailing slash fix, one way around it could be to manually add a click handler to each of those links that executes before the `page.js` one and stops further propagation.

sgomes:
Thank you for the reviews, @tyxla and @yuliyan! 👍

PR #76431, opened 2023-04-28
Adds Blaze dashboard standalone app

sbarbosa:
Related to SELFSERVE-412

## Proposed Changes

We are making the Advertising dashboard compatible to be used from a Standalone app, that will be consumed inside Jetpack (just like the stats page).

This PR adds a first version of the standalone app (based on the work done in the stats page), and also applies changes to make the advertising page compatible with Jetpack.

### Technical details of the changes:
**DSP Fetches to the API**
I changed to use the library `calypso/lib/wp` instead of directly calling the `request` method of the `wpcom-proxy-request`. It is the same on the inside because the wp library uses the `wpcom-proxy-request` handler to make the calls.

**Dynamic Advertising section URL prefix**
We need a dynamic URL prefix for the Advertising section because we won't use that prefix inside Jetpack (and on Woo in the future). I added a utility method that generates the final URL by getting the URL prefix from a configuration property.
I only spread the changes inside our Advertising module; there are links outside this module that still uses the `/advertising` hardcode prefix, but those aren't a problem because they won't line inside Jetpack or Woo.

**Refactor of the controller.js in promote-post-i2 section**
Because of how Webpack works in the Blaze Dashboard app, we need to be careful with the imports in the promote post section. 
I moved the `redirectToPrimarySite` function from the `controller.js` to the `index.js` because one of the components used in that function forced Webpack to bundle components that we don't need, and some of them conflict with the app breaking styles.
For now, I only detected that importing `calypso/controller` will make our final bundle bigger, forcing us to import the Layout component that will conflict with our custom Jetpack Layout. Let's try having all the logic that needs the `calypso/controller` living inside the index.js instead of the controller.

**Blaze Dashboard new App**
I am adding a new application inside the `/apps` folder. This app is based on the `odyssey-stats` app (an app used inside Jetpack to render the stats page from Calypso). 
have this in mind while reviewing the app, it has some changes from the other, and is still not finished.

## Testing Instructions
* Open the Live preview
* Login as A8C user
* Check out Tools > Advertising
* Verify that the page loads the products/campaigns correctly
* Click on the Promote button on one of the posts
* Verify that the DSP create campaign widget loads correctly
* Click on Cancel and then confirm the modal
* Verify that you get redirected to the post page.

## Pre-merge Checklist


apps/blaze-dashboard/README.md
```diff
@@ -0,0 +1,60 @@
...
+└── src/
+    ├── components/       ← blaze dashboard app only components. For now there is only a layout component.
+    ├── page-middleware/  ← page.js integration with React and everything
+    ├── app.js            ← entry point
+    └── routes.js         ← page.js routes
+```
+
+## Routing
+
+It utilizes the [hashbang (#!) in page.js](https://github.com/visionmedia/page.js), however it doesn't work out of the box, because we are using hardcoded paths in Calypso, so some tricks are done in Jetpack to intecept the anchor clicks and covert them to hashbangs.
```
timur987: Probably, `intecept` should be replaced with `intercept`, and `covert` with `convert`.

apps/blaze-dashboard/bin/build-languages.js
```diff
@@ -0,0 +1,227 @@
...
+const languages = require( '@automattic/languages' );
+const parse = require( 'gettext-parser' ).po.parse;
+const _ = require( 'lodash' );
+const mkdirp = require( 'mkdirp' );
+const fetch = require( 'node-fetch' );
+
+const LANGUAGES_BASE_URL = 'https://widgets.wp.com/languages/calypso';
+const LANGUAGES_REVISIONS_FILENAME = 'lang-revisions.json';
+const TEMP_PATH = './dist/temp-languages';
+const CALYPSO_STRINGS = './dist/odyssey-strings.pot';
```
timur987: Just a sanity check, do we really want to use Odyssey's strings .pot file?

apps/blaze-dashboard/package.json
```diff
@@ -0,0 +1,69 @@
...
+		"url": "https://github.com/Automattic/wp-calypso/issues"
+	},
+	"homepage": "https://github.com/Automattic/wp-calypso",
+	"scripts": {
+		"clean": "npx rimraf dist",
+		"build": "NODE_ENV=production yarn dev",
+		"build:stats": "calypso-build",
+		"dev": "yarn run calypso-apps-builder --localPath dist --remotePath /home/wpcom/public_html/widgets.wp.com/blaze-dashboard",
+		"show-stats": "NODE_ENV=production EMIT_STATS=true yarn build",
+		"translate": "rm -rf dist/strings && mkdirp dist && wp-babel-makepot '../../{client,packages,apps}/**/*.{js,jsx,ts,tsx}' --ignore '**/node_modules/**,**/test/**,**/*.d.ts' --base '../../' --dir './dist/strings' --output './dist/odyssey-strings.pot' && node bin/build-languages.js"
```
timur987: And here as well, do we want to use Odyssey's .pot file?

therocket-gr (approved):
tested and everything works
code changes LGTM

apps/blaze-dashboard/src/load-config.js
```diff
@@ -0,0 +1,20 @@
...
+
+// The option enables loading of the whole translation file, and could be optimized by setting it to `true`, which needs the translation chunks in place.
+// @see https://github.com/Automattic/wp-calypso/blob/trunk/docs/translation-chunks.md
+productionConfig.features[ 'use-translation-chunks' ] = false;
+
+// Override the redesign feature
+productionConfig.features[ 'promote-post/redesign-i2' ] = true;
+productionConfig.dsp_widget_js_src = 'https://widgets.wp.com/promote-v2/widget.js';
+
+// Note: configData is hydrated in https://github.com/Automattic/jetpack/blob/d4d0f987cbf63a864b03b542b7813aabe87e0ed3/projects/packages/stats-admin/src/class-dashboard.php#L214
```
timur987: Probably we need to put the path related to another source of the configData (currently, it's for admin-stats).

client/my-sites/promote-post-i2/controller.js
```diff
@@ -0,0 +1,30 @@
...
+
+	const postId = item?.split( '-' )[ 1 ];
+
+	context.primary = (
+		<BlazePressWidget
+			isVisible={ true }
+			siteId={ siteId }
+			postId={ postId }
+			keyValue={ item }
+			// source={ source }
```
timur987: Do we need this comment?

timur987 (approved):
Thanks, @sbarbosa!

I checked changes (they look great) and how they work in Calypso Live and locally.

I am not confident how it should work via jetpack (it seems testing instructions do not cover this topic, but at least I'm sure that the current behavior works as expected).

apps/blaze-dashboard/bin/build-languages.js
sbarbosa: You are right, I will rename the file here. Thanks!

apps/blaze-dashboard/src/load-config.js
sbarbosa: I was waiting to merge the PR related to Jetpack, and then go and update all the references in the comments here.
I don't know if we can point to a draft PR here (before it is merged).

I can add TODO in those comments and remove the links until the code is merged.
What do you think @timur987 ?

timur987: > I can add TODO in those comments and remove the links until the code is merged.

It sounds like a plan. That's definitely a good idea.

client/my-sites/promote-post-i2/controller.js
sbarbosa: Thanks for pointing this out!.
For consistency, I will use the same logic that we use in the Post item component. In there, the source is extracted from the URL query parameter `source`

apps/blaze-dashboard/src/load-config.js
```diff
@@ -0,0 +1,21 @@
...
+productionConfig.features.is_running_in_jetpack_site =
+	window.configData.features.is_running_in_jetpack_site ?? true;
+
+// The option enables loading of the whole translation file, and could be optimized by setting it to `true`, which needs the translation chunks in place.
+// @see https://github.com/Automattic/wp-calypso/blob/trunk/docs/translation-chunks.md
+productionConfig.features[ 'use-translation-chunks' ] = false;
+
+// Override the redesign feature
+productionConfig.features[ 'promote-post/redesign-i2' ] = true;
+productionConfig.dsp_widget_js_src = 'https://widgets.wp.com/promote-v2/widget.js';
```
benzhu56, line 14: Are we forcing to use v2 of campaign dashboard and v2 of blaze widget here? I don't know much about standalone apps in Calypso. Is the current version a standalone app? We are also trying to test the v2 widget redesign with Jetpack, but I'm concerned that with forcing to use `/promote-v2/widget.js` here, we'll expose v2 to the users before releasing it.

sbarbosa: We are forcing both of them here because right now, both are using the same feature flag, and by the time we merge this PR there was an error that if you used the `promote/widget.js` entry point, with the `isV2` flag in true it showed you an mix version of the widget (new redesign HTML but with old CSS). 

This app is not being used anywhere yet, we plan to release the first version when we finish all the redesign tasks and enable it for users in Jetpack after all the testing is done. 

I just tested it and the bug I mentioned before appears to be fixed, so we can make a PR removing this line, which will force the app to take the config value from production instead (that is the v1 entry point).

PR #76434, opened 2023-04-28
Plugins: Use typescript plugin selectors in plugins

robertf4:
## Proposed Changes

This PR updates the main Plugins code to use the new TypeScript plugins selectors which were developed beginning in https://github.com/Automattic/wp-calypso/pull/72363.

## Testing Instructions

Check that the following works:
Plugin details page (/plugins/:plugin_slug)
* Activating/deactivating
* Enabling/disabling autoupdate
* Updating

Site details page (/plugins/manage/:site_slug)
* Activating/deactivating via toggle
* Enabling/disabling autoupdate via toggle
* Updating all plugins via primary update button
* Updating one plugin via update link
* Activating/deactivating via "edit all"
* Enabling/disabling autoupdate via "edit all"
* Updating via "edit all"
* Removing via "edit all"
* Removing via ellipsis

Plugins manage page
Site details page (/plugins/manag)
* Activating/deactivating via toggle
* Enabling/disabling autoupdate via toggle
* Updating all plugins via primary update button
* Updating one plugin via update link
* Activating/deactivating via "edit all"
* Enabling/disabling autoupdate via "edit all"
* Updating via "edit all"
* Removing via "edit all"
* Removing via ellipsis

## Pre-merge Checklist


client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,14 @@ import { getSite } from 'calypso/state/sites/selectors';
 
 import 'calypso/state/plugins/init';
 
+// TODO: remove this function when client/state/plugins/installed/test/selectors-ts.ts
+//       replaces client/state/plugins/installed/test/selectors.js
+const isSiteDataOnPluginObject = ( plugin ) => {
```
robertf4: This is here to make these actions work while both the TypeScript and JavaScript versions of the selectors coexist, and it will be removed when the JavaScript selectors are removed. The TypeScript versions place these properties on the site object instead of the plugin object, so by testing for them we can determine which version of the data is being used and handle it appropriately.

robertf4: I changed my mind on this and decided to make it a feature of the `actions.js` file. Sometimes data returned from the server is dispatched directly into one of these actions. Data returned from the server always has the site-specific properties on the plugin object. Because of this I believe it is safest to have the actions able to handle both data structures.

tyxla: Do you think we can address this in a separate PR? So we could keep this substantially smaller and easier to review?

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,19 @@ import { getSite } from 'calypso/state/sites/selectors';
...
+ * Determines if site specific data is on the plugin object.
+ * If it is not, then it can be assumed it is on the sites property.
+ *
+ * @param {Object} plugin - plugin object
+ * @returns {boolean} True if any of the site specific properties are on the plugin, false otherwise.
+ */
+const isSiteDataOnPluginObject = ( plugin ) => {
+	return [ 'active', 'autoupdate', 'update', 'version' ]
+		.map( ( prop ) => plugin.hasOwnProperty( prop ) )
+		.includes( true );
```
tyxla: I wonder if we can just use `Array.prototype.some()`

```suggestion
		.some( ( prop ) => plugin.hasOwnProperty( prop ) );
```

client/my-sites/plugins/plugin-details-header/index.jsx
```diff
@@ -17,6 +17,16 @@ const PluginDetailsHeader = ( { plugin, isPlaceholder, isJetpackCloud } ) => {
...
+	// This gets the version of the plugin on the last site or
+	// the wporg version if that doesn't exist. This is preserving
+	// some pre-existing behavior during a refactor and doesn't
+	// necessarily have to remain this way in the future.
+	let version = plugin?.version;
+	if ( plugin.sites ) {
+		const sites = Object.values( plugin.sites );
+		version = sites.findLast( ( site ) => site.version ).version;
+	}
+
```
tyxla, line 29: I wonder if we should extract this to a selector. It's rarely a good idea to do such calculations at render time.

client/my-sites/plugins/plugin-management-v2/remove-plugin/index.tsx
```diff
@@ -12,12 +10,10 @@ interface Props {
 }
 
 export default function RemovePlugin( { site, plugin }: Props ) {
-	const pluginOnSite = useSelector( ( state ) => getPluginOnSite( state, site.ID, plugin.slug ) );
-
 	return (
 		<PluginRemoveButton
 			key={ `remove-plugin-${ site.ID }` }
-			plugin={ pluginOnSite }
+			plugin={ plugin }
```
tyxla, line 16: Nice 👍

client/my-sites/plugins/plugin-management-v2/test/plugin-row-formatter.test.tsx
```diff
@@ -116,9 +117,15 @@ describe( '<PluginRowFormatter>', () => {
...
+		plugin.sites = sitesOriginal;
 	} );
 
-	test( 'should render correctly and show activate and toggle checked value', async () => {
+	// These skipped test nevers actually worked, it just tests that the HTML toggle actually toggled.
+	// To verify comment out the entirety of the toggle callbacks in the relevant components and observe
+	// that the test still pass.
+	// They were passing because their toggle code didn't execute. I changed some test data which caused
+	// the toggle code to actually execute, and now it needs to be rewritten to use nock to simulate the server calls.
+	test.skip( 'should render correctly and show activate and toggle checked value', async () => {
```
tyxla: Fair enough, thanks for clarifying. Let's add a TODO to the comment, so we can follow-up and fix them eventually.

client/my-sites/plugins/plugin-management-v2/test/plugin-row-formatter.test.tsx
```diff
@@ -131,7 +138,7 @@ describe( '<PluginRowFormatter>', () => {
 		expect( toggleButton ).toHaveProperty( 'checked', true );
 	} );
 
-	test( 'should render correctly and show auto-update and toggle checked value', async () => {
+	test.skip( 'should render correctly and show auto-update and toggle checked value', async () => {
```
tyxla, line 142: Same as above.

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -30,48 +29,85 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
...
+	selectedPlugins: { [ pluginSlug: string ]: boolean };
+}
 
+type Action =
+	| 'updating'
+	| 'activating'
+	| 'deactivating'
+	| 'enablingAutoupdates'
+	| 'disablingAutoupdates'
+	| 'removing';
```
tyxla, line 79: I wonder if these types can be moved somewhere else at a reusable place?

client/state/plugins/installed/test/fixtures/plugins.js
```diff
@@ -59,21 +78,22 @@ export const jetpack = {
...
 	description:
 		'Bring the power of the WordPress.com cloud to your self-hosted WordPress. Jetpack enables you to connect your blog to a WordPress.com account to use the powerful features normally only available to WordPress.com users.',
 	author: 'Automattic',
 	author_url: 'http://jetpack.com',
 	network: false,
-	autoupdate: true,
 	log: [ 'Array' ],
+	active: true,
+	version: '4.2.2',
+	autoupdate: true,
```
tyxla: Just curious: why are these changes necessary?

tyxla (commented):
Thanks for the hard work getting this ready for review! Really nice work 👏 

There seems to be a failing check - some types are likely not exactly in place.

I've also asked a few questions, I think there might be a bit more opportunity to move to separate PRs. That could make the PR even easier to review and test.

client/state/plugins/installed/actions.js
robertf4: I created https://github.com/Automattic/wp-calypso/pull/76684 to change only the actions. We can get that one reviewed and merged first and then rebase this one to remove these changes from this PR. But in the meantime if I take these changes off this PR then this PR won't be able to be tested.

client/state/plugins/installed/actions.js
robertf4: Yes that works, I made this change in the new actions PR: https://github.com/Automattic/wp-calypso/pull/76684

client/state/plugins/installed/test/fixtures/plugins.js
robertf4: The plugin objects which are returned from the server always come from a site-specific endpoint, so they have these site-specific properties on them. However in the store the objects are modified so that site-specific properties are found on a site object in order to accommodate cases where plugins on multiple sites are being displayed.

So I need both versions of these objects for the unit tests. Mocked server responses use the version with the site-specific properties on the object, while calls to actions that would be made from component code use the version with the site-specific properties on site objects.

The rearranging of the props was an artifact of my development process which reduced the readability of the diff, so I undid those changes in https://github.com/Automattic/wp-calypso/pull/76684

client/my-sites/plugins/plugin-details-header/index.jsx
robertf4: I don't want to put this in a selector because this was basically undefined behavior, and making it a reusable selector would make it official. We need to answer the question of what version a plugin should show here if it has multiple versions on multiple sites. Maybe it could be a range, or it could be the most recent version of the plugin which is installed, or it could be the most recent version of a plugin which is available. No matter the case this logic should not be used anymore. I have kept this logic here because I don't want these changes to alter what is shown in the UI in any way.

@vitozev might be able to clarify if this is something that Jetpack Genesis could take a deeper look at.

client/my-sites/plugins/plugins-list/index.tsx
robertf4: I considered it, but these are used only inside of this component for the purposes of its internal implementation.

client/my-sites/plugins/plugin-details-header/index.jsx
vitozev: We can definitely look into this. If you have more context (e.g., where precisely in the UI we show this, etc.), I would appreciate it if you add this to our [maintenance board](1202619025189113-asboard), and we'll triage it.

robertf4:
I've set the base branch for this to https://github.com/Automattic/wp-calypso/pull/76684 until it gets merged. @tyxla feel free to review this in the meantime. I will simply rebase again if https://github.com/Automattic/wp-calypso/pull/76684 is changed further.

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -532,14 +575,18 @@ export class PluginsList extends Component {
 		const { plugins } = this.props;
 		return plugins.reduce( ( sites, plugin ) => {
 			Object.keys( plugin.sites ).map( ( pluginSiteId ) => {
-				if ( ! sites.find( ( site ) => site.ID === pluginSiteId ) ) {
-					const pluginSite = this.props.allSites.find( ( s ) => s.ID === parseInt( pluginSiteId ) );
-					sites.push( pluginSite );
+				if ( ! sites.find( ( site ) => site && site.ID === Number( pluginSiteId ) ) ) {
+					const pluginSite = this.props.allSites.find(
+						( s ) => s && s.ID === parseInt( pluginSiteId )
```
tyxla: Any good reason to prefer `parseInt()` to `Number()` for casting to a number (L578 vs L580)? Seems like we can try to be consistent about the one we prefer. In any case, from my understanding they'll work the same way when working with a string that contains a number.

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -178,17 +215,17 @@ export class PluginsList extends Component {
...
 		},
-		updates( plugin ) {
+		updates: ( plugin: Plugin ) => {
 			if ( this.isSelected( plugin ) ) {
 				return Object.keys( plugin.sites ).some( ( siteId ) => {
 					const sitePlugin = getSitePlugin( plugin, siteId, this.props.pluginsOnSites );
-					const site = this.props.allSites.find( ( s ) => s.ID === parseInt( siteId ) );
-					return sitePlugin?.update?.new_version && site.canUpdateFiles;
+					const site = this.props.allSites.find( ( s ) => s && s.ID === parseInt( siteId ) );
+					return site && sitePlugin?.update?.new_version && site.canUpdateFiles;
```
tyxla: Nit: We could just use optional chaining to make it shorter:

```suggestion
					return sitePlugin?.update?.new_version && site?.canUpdateFiles;
```

client/state/plugins/installed/test/fixtures/plugins.js
tyxla: I might be misunderstanding, but I still see those even though the branch is based on the one from https://github.com/Automattic/wp-calypso/pull/76684. Not a big deal of course.

Noting that the order of object properties has historically been non-guaranteed, so a more reliable way to write the test assertions is to use [`expect.objectContaining`](https://jestjs.io/docs/expect#expectobjectcontainingobject).

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -532,14 +575,18 @@ export class PluginsList extends Component {
 		const { plugins } = this.props;
 		return plugins.reduce( ( sites, plugin ) => {
 			Object.keys( plugin.sites ).map( ( pluginSiteId ) => {
-				if ( ! sites.find( ( site ) => site.ID === pluginSiteId ) ) {
-					const pluginSite = this.props.allSites.find( ( s ) => s.ID === parseInt( pluginSiteId ) );
-					sites.push( pluginSite );
+				if ( ! sites.find( ( site ) => site && site.ID === Number( pluginSiteId ) ) ) {
```
tyxla: I'd prefer using `some` to `find` if we don't need to utilize the found value.

```suggestion
				if ( ! sites.some( ( site ) => site && site.ID === Number( pluginSiteId ) ) ) {
```

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -230,20 +267,24 @@ export class PluginsList extends Component {
...
+		const flattenArrays = ( full: any[], partial: any[] ) => [ ...full, ...partial ];
 		this.removePluginStatuses();
 		const pluginAndSiteObjects = selectedPlugins
-			.filter( ( plugin ) => ! isDeactivatingOrRemovingAndJetpackSelected( plugin ) ) // ignore sites that are deactivating, activating or removing jetpack
-			.map( ( p ) => {
+			.filter( ( plugin: Plugin ) => ! isDeactivatingOrRemovingAndJetpackSelected( plugin.slug ) ) // ignore sites that are deactivating, activating or removing jetpack
+			.map( ( p: Plugin ) => {
 				return Object.keys( p.sites ).map( ( siteId ) => {
-					const site = this.props.allSites.find( ( s ) => s.ID === parseInt( siteId ) );
+					const site = this.props.allSites.find( ( s ) => s && s.ID === parseInt( siteId ) );
```
tyxla: We could as well use optional chaining:

```suggestion
					const site = this.props.allSites.find( ( s ) => s?.ID === parseInt( siteId ) );
```

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -178,17 +215,17 @@ export class PluginsList extends Component {
...
 			return false;
 		},
-		updates( plugin ) {
+		updates: ( plugin: Plugin ) => {
 			if ( this.isSelected( plugin ) ) {
 				return Object.keys( plugin.sites ).some( ( siteId ) => {
 					const sitePlugin = getSitePlugin( plugin, siteId, this.props.pluginsOnSites );
-					const site = this.props.allSites.find( ( s ) => s.ID === parseInt( siteId ) );
-					return sitePlugin?.update?.new_version && site.canUpdateFiles;
+					const site = this.props.allSites.find( ( s ) => s && s.ID === parseInt( siteId ) );
```
tyxla:
```suggestion
					const site = this.props.allSites.find( ( s ) => s?.ID === parseInt( siteId ) );
```

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -1,8 +1,7 @@
 import { recordTracksEvent } from '@automattic/calypso-analytics';
 import { WPCOM_FEATURES_MANAGE_PLUGINS } from '@automattic/calypso-products';
-import { localize } from 'i18n-calypso';
+import { localize, translate } from 'i18n-calypso';
```
tyxla: I wonder if we can just import the type if the only reason to have the import is to do a `typeof`? `i18n-calypso` comes with its own types, so we can just use them.

```
import type { translate } from 'i18n-calypso';
```

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -30,48 +29,86 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
...
+	selectedSite?: SiteDetails;
+	selectedSiteSlug?: string;
+	siteIsAtomic?: boolean;
+	siteIsJetpack?: boolean;
+	searchTerm: string;
+	recordGoogleEvent: typeof recordGoogleEvent;
+	allSites: ReturnType< typeof getSites >;
+	pluginsOnSites: ReturnType< typeof getPluginsOnSites >;
+	removePluginStatuses: typeof removePluginStatuses;
+	translate: typeof translate;
```
tyxla: As mentioned above, we should be able to use the `translate` type from `i18n-calypso` directly.

client/my-sites/plugins/plugins-list/index.tsx
```diff
@@ -30,48 +29,86 @@ import { isJetpackSite } from 'calypso/state/sites/selectors';
...
+	inProgressStatuses: PluginStatus[];
+	isPlaceholder: boolean;
+	plugins: Plugin[];
+	selectedSite?: SiteDetails;
+	selectedSiteSlug?: string;
+	siteIsAtomic?: boolean;
+	siteIsJetpack?: boolean;
+	searchTerm: string;
+	recordGoogleEvent: typeof recordGoogleEvent;
+	allSites: ReturnType< typeof getSites >;
```
tyxla: Given that `selectedSite` is `SiteDetails`, should this just be `SiteDetails[]`?

client/my-sites/plugins/plugin-details-header/index.jsx
robertf4: I created 1202619025189113-as-1204688670518465

client/my-sites/plugins/plugins-list/index.tsx
robertf4: I couldn't get `translate` to work as it seems to be declared only as a value. It did have the type specified in `I18N` though, so I used that.

client/my-sites/plugins/plugins-list/index.tsx
robertf4: Yes, that works.

client/my-sites/plugins/plugins-list/index.tsx
robertf4: In this case it was just pre-existing code that I failed to recognize. I switched it to `Number` for consistency.

robertf4:
@tyxla I think this is ready for a final review.

tyxla:
Any concerns with regards to the #76788 revert? Should we write some base e2e tests before we go ahead?

robertf4:
> Any concerns with regards to the https://github.com/Automattic/wp-calypso/pull/76788 revert? 

@tyxla I'm going to go over all this code again after https://github.com/Automattic/wp-calypso/pull/77575 merges and every selector has a return type. I'm about to head to a meetup so I probably won't get to this until next week, so hold off on further review of this one. 

> Should we write some base e2e tests before we go ahead?

It's certainly something that we should consider, but I'm not sure what the scope of it would be. I'm not all that familiar with writing e2e tests for Calypso in general. @vitozev might have further input as to whether that should be handled by Jetpack Genesis or not.

PR #76445, opened 2023-05-01
Odyssey Widget: use container query with media query as fallback

### LINKED ISSUE #76380: Odyssey Widget: Single column for when container is narrow
This apply for both mobile layout and also desktop layout, when the container is narrower than a certain number, we show single columned widget.

Image

kangzj:
Related to #76380

## Proposed Changes

- Replaces Media Query with [Container Query](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) , which has ~86% support, which is not-bad support rate for browsers.
- Fallback to Media Query
- Uses `<=599px` as a narrow breakpoint

A note here, named container doesn't seem to work on Firefox - we only used the unnamed container solution.

## Testing Instructions

* Adjust the width of the screen on Chrome, Firefox and Safari
* Ensure the widget switches to mobile layout when the container is narrower than `600px`
* Ensure it works well on Android Chrome (40% of market share)

image

## Pre-merge Checklist


apps/odyssey-stats/src/widget/hightlights.scss
```diff
@@ -19,7 +19,8 @@ $offset-padding-hovering: 8px;
 		display: none;
 		margin-top: 16px;
 
-		@media only screen and (max-width: $break-dashboard-mobile) {
+		/* stylelint-disable-next-line scss/at-rule-no-unknown */
+		@container (max-width: #{$break-dashboard-mobile}) {
```
jsnmoon: We may want to consider using `$break-small` (600px) from `@wordpress/base-styles/breakpoints` instead of defining our own breakpoint. Doing so will allow us to avoid adding stylelint exceptions.

jsnmoon (approved):
This works well on desktop Firefox, desktop Chrome, and mobile Chrome.

It didn't work in Samsung Internet Browser (v19) [as expected](https://caniuse.com/css-container-query-units), but that's a little concerning given it was released only a few months ago in October 2022. Is it possible to add window media query fallback?

At any rate, I don't think these concerns are worth blocking this PR.

grzegorz-cp:
It works fine in Chrome on Samsung S22.

apps/odyssey-stats/src/app.scss
```diff
@@ -63,7 +63,7 @@
 	}
 
 	// Black segmented controls.
-	.is-section-stats .segmented-control .segmented-control__item {
+	.is-section-stats ul.segmented-control li.segmented-control__item {
```
kangzj, line 66: Fix green Segment Control on summary page by the way:

image

apps/odyssey-stats/src/widget/hightlights.scss
kangzj: The linting error is from `@container` which is unknown to SCSS transpiler. I did change to use `@wordpress/base-styles/breakpoints`, but the error deprecating is still needed to pass linting.

kangzj:
@jsnmoon @grzegorz-cp Thanks for your review. I updated the code with media query fallback and could use another look please. Thank you so much for testing multiple browsers. 👍

apps/odyssey-stats/src/widget/index.scss
```diff
@@ -100,3 +95,12 @@ $segmentedControlItemBorderRadius: 4px;
 		color: var(--studio-gray-100);
 	}
 }
+
+/* stylelint-disable-next-line scss/at-rule-no-unknown */
+@container (max-width: #{$break-small}) {
```
dognose24: The Stats Widget container width could be under `600px` in the four columns dashboard layout even though the window width is `2560px`, which is a little weird. IMHO, I prefer to set `480px` for the container query that still accommodates enough information for two columns of module cards. cc @jsnmoon 

截圖 2023-05-02 下午3 12 46

dognose24 (approved):
Thanks for the nice improvement, @kangzj! Works as described! I have some thoughts on the container query breakpoint, which differs from the breakpoint for window width. 🙂

apps/odyssey-stats/src/widget/index.scss
kangzj: Agreed 👍

PR #76454, opened 2023-05-01
Plans: Move query to domain-suggestions store

aneeshd16:
Related to p1682942089709679-slack-CV2TX2PAN

## Proposed Changes

* This PR moves the `useGetWordPressSubdomain` query hook to the `domain-suggestions` data store to decouple it from the plans grid. Also added a generic `useGetDomainSuggestions` hook which can be used to fetch domain suggestions. 

## Testing Instructions

Same as #75526

## Pre-merge Checklist


client/my-sites/plan-features-2023-grid/components/features.tsx
```diff
@@ -1,8 +1,8 @@
 import { getPlanClass, FEATURE_CUSTOM_DOMAIN, isFreePlan } from '@automattic/calypso-products';
+import { useGetWordPressSubdomain } from '@automattic/data-stores/src/domain-suggestions';
```
chriskmnds: Thinking whether it would be better to have `import { DomainSuggestions } from '@automattic/data-stores';` instead - and reach the query via `DomainSuggestions.useGetWordPressSubdomain`. 

Although, could that imply having the store registered as a side effect? (I think it won't be the case here, as we still need to explicitly call the `register` function, but it might become the case once the store is refactored like the others) If not a concern, then I'd go for `import { DomainSuggestions } from '@automattic/data-stores';`

packages/data-stores/src/domain-suggestions/queries.ts
```diff
@@ -0,0 +1,67 @@
...
+						suggestion.currency_code && {
+							cost: getFormattedPrice( suggestion.raw_price, suggestion.currency_code ),
+						} ),
+				};
+			} );
+
+			return processedSuggestions;
+		},
+		enabled: !! search,
+		staleTime: 1000 * 60 * 5, // 5 minutes
```
chriskmnds: Maybe we can pull this out into a constant. 

Also a little concerning maybe that there's no consistency or design across most queries in data stores concerning this expiration time (some set to Infinity, others to some value). could this be configured better?

packages/data-stores/src/domain-suggestions/queries.ts
```diff
@@ -0,0 +1,67 @@
+import { useQuery } from '@tanstack/react-query';
+import { stringify } from 'qs';
+import wpcomProxyRequest from 'wpcom-proxy-request';
+import { getFormattedPrice, normalizeDomainSuggestionQuery } from './utils';
```
chriskmnds, line 4: Nice reuse

chriskmnds (approved):
Looks good. Tested and modal works as expected. Thanks for the follow-up @aneeshd16! 💯 

Not sure if expected, or something due to a rebase needed, but I no longer see the "_Free for the first year with annual paid plans_"  notif with this PR:

Screenshot 2023-05-04 at 2 47 00 PM

Happy to take another look and help confirm if any changes are needed.

aneeshd16:
Thanks @chriskmnds for the review!

> Not sure if expected, or something due to a rebase needed, but I no longer see the "Free for the first year with annual paid plans" notif with this PR:

I am not able to reproduce this. Can you reproduce this consistently? The PR does not affect the domains step, so maybe this is a temporary issue? 🤔

chriskmnds:
> I am not able to reproduce this. Can you reproduce this consistently? The PR does not affect the domains step, so maybe this is a temporary issue? 🤔

Sounds like some temporary issue I stumbled on. No longer see it now, locally or in the live image.

packages/data-stores/src/domain-suggestions/queries.ts
aneeshd16: > Maybe we can pull this out into a constant.

Done via 8e0788fce420e5ffc8a721c749804295c2ab1c00.

> Also a little concerning maybe that there's no consistency or design across most queries in data stores concerning this expiration time (some set to Infinity, others to some value). could this be configured better?

I think it depends on the query too. In some cases, the server data will never change, so it is safe to set the staleTime to Infinity. In our case, there is a small chance that the free subdomain could be picked up by some other user, so we set a finite expiration.

client/my-sites/plan-features-2023-grid/components/features.tsx
aneeshd16: Done via 5061f39b6bc2ab0314dedef4732688dbf8fa692e

chriskmnds:
Great work all around 🥇 Tests out for me

PR #76459, opened 2023-05-01
Limited Global Styles: Open launchbar popup by default

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2268

## Proposed Changes

Improves the Limited Global Styles notice displayed in the launchbar:
- The dialog is open by default when custom styles are previewed, so it's less likely to miss.
- Wording has been adapted to make the UX more intuitive.

State | Before | After
--- | --- | ---
Preview ON | Screenshot 2023-05-01 at 15 20 00 | Screenshot 2023-05-01 at 15 17 55
Preview ON (mobile) | Screenshot 2023-05-01 at 15 19 40 | Screenshot 2023-05-01 at 15 18 44
Preview OFF | Screenshot 2023-05-01 at 15 20 27 | Screenshot 2023-05-01 at 15 18 14
Preview OFF (mobile) | Screenshot 2023-05-01 at 15 19 32 | Screenshot 2023-05-01 at 15 18 32

## Testing Instructions

- Visit the frontend.
- Make sure the dialog is open by default.
- Make sure you can show the custom styles using the secondary button in the dialog.
- Make sure the wording is clear.


BogdanUngureanu:
Great job, Miguel, this looks really good. I've noticed some small issues though I think we should fix them before shipping it:

- On mobile resolutions, clicking on the modal text will redirect the users to the plans page - which is an unexpected behavior.
- The modal remains open between refreshes even though I clicked it on the panel to collapse it. This might be annoying to users.
- When the styles are hidden, the modal is always collapsed (inconsistent with the other state where is always open).
-  When clicking on the upgrade and on the Turn on/off preview buttons we aren't triggering the events we used to trigger.
- I think we should add a new event that tracks the impressions for the modal. This would be inline with the impression events we trigger for the other GS nudges.

mmtr:
> On mobile resolutions, clicking on the modal text will redirect the users to the plans page - which is an unexpected behavior.

I cannot replicate this – Am I missing any testing step?

https://user-images.githubusercontent.com/1233880/235696356-5dd25571-916f-42f2-ac9d-23a64663735f.mov

> The modal remains open between refreshes even though I clicked it on the panel to collapse it.

Solved in https://github.com/Automattic/wp-calypso/pull/76459/commits/58452f9676d97c8238515c556d34ab4bbaac331d

> When the styles are hidden, the modal is always collapsed (inconsistent with the other state where is always open).

Yes, this is intentional. The reason why the dialog is open by default is to provide awareness to users and ensure it's less likely to miss. If users turned off the preview, they are already aware of the notice, so we no longer need to display the dialog open by default.

> When clicking on the upgrade and on the Turn on/off preview buttons we aren't triggering the events we used to trigger

Good catch! Fixed in https://github.com/Automattic/wp-calypso/pull/76459/commits/a2bcb8eedd3b5d5a0224444cfa91e71429bad874

> I think we should add a new event that tracks the impressions for the modal

Added a new `action` property in https://github.com/Automattic/wp-calypso/pull/76459/commits/3d384095f4ac0d3587df8d0130244f016fb15c39 and https://github.com/Automattic/wp-calypso/pull/76459/commits/61c79814c3b32477ea4ba9cf4d9a63d4da14952b to the `wpcom_global_styles_gating_notice` event so we can determine whether the dialog has been displayed.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
```diff
@@ -23,13 +24,25 @@ document.addEventListener( 'DOMContentLoaded', () => {
...
+	}
 
 	popoverToggle?.addEventListener( 'click', ( event ) => {
 		event.preventDefault();
-		recordEvent( 'wpcom_global_styles_gating_notice' );
+		recordEvent( 'wpcom_global_styles_gating_notice', {
+			action: popover?.classList.contains( 'hidden' ) ? 'show' : 'hide',
+		} );
 		popover?.classList.toggle( 'hidden' );
+		localStorage.setItem( 'showLimitedGlobalStylesNotice', '0' );
```
BogdanUngureanu: Maybe the value should be toggled instead of always being persisted as '0'? If the user leaves it open, then it should remain opened? 🤔

BogdanUngureanu:
> I cannot replicate this – Am I missing any testing step?

Looks like I'm not able to reproduce it anymore. 

Looks much better, thanks for addressing the issues. There's one more thing (not blocking). When a user clicks on the Turn on/off preview button we log an event. However, we can't differentiate between the Turn on and the Turn off actions. What do you think if we add a new prop as we did for the modal collapse? :) 

I've also added a small comment about the modal collapse state.

BogdanUngureanu:
Non-blocking: On mobile view, "Upgrade" label might not be explicit enough, especially on closed modal states. But again, no alternative comes to mind. It's an improvement anyway. 

Non-blocking: Are we sure that having the modal open on mobile viewports is clear enough on how to close it? As a user, when visiting the front-end for the first time, it might be not intuitive enough that the modal is actually collapsible by clicking on the item in the launchbar.

mmtr:
> we can't differentiate between the Turn on and the Turn off actions. What do you think if we add a new prop as we did for the modal collapse? :)

It's a good idea, but slightly more complicated than tracking when the dialog is open because the JS code doesn't know whether the preview is turned on or not. Given that the tracking behavior of the preview button has not changed in this PR, I think we can leave that for a separate PR if we're interested in tracking such info.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
mmtr: https://github.com/Automattic/wp-calypso/pull/76459/commits/88331cd28674f165f5402fe409622f2d38de87e1

mmtr:
> Are we sure that having the modal open on mobile viewports is clear enough on how to close it?

Added a close button in https://github.com/Automattic/wp-calypso/pull/76459/commits/fc231a2e4d4054910c4a6538a560a7590959c98a so users have a more natural way of closing it.

mmtr:
As discussed in p1683108883912269-slack-C04DZ8M0GHW, by showing custom styles by default we also unexpectedly showing them in the post editor, so we're going to leave that change for a separate PR where we will also include a warning notice in the post editor.

davewhitley:
Current specs (pending design feedback) can be found here: https://www.figma.com/file/PgUzt3AFZ6v6rWrmXO1uDl/Limited-Global-Styles-(upsells)?type=design&node-id=783-7960&t=P0uDQIXKwpcVMTIa-4

BogdanUngureanu (approved):
LGTM! 🚢

davewhitley:
Could we get a final screenshot of the design now? After that I think we could go ahead and ship unless you want to wait until a decision is made on the 3rd additional button, "Reset custom styles".

mmtr:
@davewhitley here you are!
Screenshot 2023-05-04 at 17 44 50

davewhitley:
LGTM

Issue #70070, opened 2022-11-16
[Bug]: Activating Tsubaki opens a WooCommerce onboarding flow I can't escape


dsas:
### Quick summary

Installing the Tsubaki theme dumps me on a 'Add an address to accept payments' screen which I wasn't expecting. There is a back button at the top of this screen, it doesn't work. The back button in the browser is also broken.

There are three things going on here:
 1. I was surprised to be on some kind of setting screen without warning, other themes didn't do this.
 2. I couldn't navigate out of this screen.
 3. I didn't know whether the theme was active, there was no confirmation that it was without going to visit the front-end of the site myself.

### Steps to reproduce

 1. Have a site on a paid plan (I was testing with a simple site with a Pro plan)
 2. Go to Appearance
 3. Search for the Tsubaki theme
 4. Activate it
 5. End up on https://wordpress.com/setup/plugin-bundle/storeAddress?siteSlug=whatever.wordpress.com
 6. Press back at the top of the page
 7. Press back in the web browser

### What you expected to happen

I expected to go back to the previous page (theme info) so I could navigate to change to a different theme

Also it's unclear at this point whether the theme has actually changed or not

### What actually happened

Clicking the back button just put me on the same screen again. To be more specific pressing back and using the browser back both sent me to https://wordpress.com/setup/plugin-bundle/getCurrentThemeSoftwareSets?siteSlug=whatever.wordpress.com which instantly redirected me to https://wordpress.com/setup/plugin-bundle/storeAddress?siteSlug=whatever.wordpress.com

Additionally when I held down the back button to choose where to navigate to, all of these screens just said 'WordPress.com', which is technically correct but also useless.

### Browser

Mozilla Firefox

### Context

Exploratory testing

### Platform (Simple, Atomic, or both?)

Simple

### Reproducibility

Consistent

### Severity

Some (< 50%)

cuemarie:
📌 SCRUBBING : RESULT - *Replicated / Could Not Replicate / Uncertain*
- Tested on Simple – Replicated

📌 FINDINGS/SCREENSHOTS/VIDEO
- Replicated as described!  I had to navigate directly to WordPress.com get out of the setup for WooCommerce.  When I return to the site, it's still Simple, but Tsubaki is now active

#### Screenshots/Recordings

https://user-images.githubusercontent.com/27249804/235012340-40697fae-9e90-4dd3-bfbf-6062ad60096d.mp4

📌 ACTIONS
- Assigned to Serenity

[...]

PR #76474, opened 2023-05-01
Add back button if coming from theme activation

andres-blanco:
Related to #https://github.com/Automattic/wp-calypso/issues/70070
The idea behind this PR is to give the user a way out of the plugin-bundle flow after activating a theme. 
This might not be the best approach but at least resolves a strange problem for the user while we sort out the best way to fix it. 

## Testing Instructions

* Apply this PR.
* Set up a site with a Business plan.
* Activate any WooCommerce theme (the ones with the badge image)
* Select either of the options presented.
* Verify you land in the plugin-bundle flow but you are provided with a `Back` button:
image
* Click on the button and verify you land on the themes page for the site and the theme you selected is active.
* Go through the steps again and verify that removing the `comingFromThemeActivation=true` flag from the URL after you are redirected to the plugin-bundle flow removes the `Back` button.


sixhours (approved):
This works for me, and I think it's an improvement over users having nowhere to go when they unexpectedly land in this flow. 👍

PR #76487, opened 2023-05-02
Learn: mobile links and subscribe on content footer

escapemanuele:
## What
Update to the mobile design of the links and subscribe on the bottom of the content footer.

### Before
image

### After
image

## Testing
1. Checkout branch, `cd apps/happy-blocks`and `yarn dev --sync`
2. Sandbox `learncft.wordpress.com` and visit the site
3. Switch to mobile and check the result

### DIFF
--- a/apps/happy-blocks/block-library/support-content-footer/style.scss
+++ b/apps/happy-blocks/block-library/support-content-footer/style.scss
@@ -92,68 +92,6 @@ $system-font-family: -apple-system, BlinkMacSystemFont, "Noto Sans", "Segoe UI",
 		font-family: var(--wp--preset--font-family--system-font);
 	}
 
-	@media (max-width: $breakpoint-mobile) {
-		padding: 12px 24px !important;
-
-		h4 {
-			font-size: 1.25rem;
-			line-height: 26px;
-		}
-
-		p {
-			font-size: 0.875rem;
-			margin-top: 4px;
-			margin-bottom: 8px;
-			line-height: 20px;
-		}
-
-		.support-content-cta {
-			margin-bottom: 0;
-
-			.support-content-cta-left {
-				padding: 0 48px 16px 0;
-			}
-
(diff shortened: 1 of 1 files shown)


apps/happy-blocks/block-library/support-content-footer/style.scss
```diff
@@ -135,6 +135,22 @@ $system-font-family: -apple-system, BlinkMacSystemFont, "Noto Sans", "Segoe UI",
...
 			flex-direction: column;
+			padding-bottom: 36px;
+
+			.support-content-subscribe {
+				margin-top: 32px;
+			}
+
+			.support-content-subscribe-email {
+				padding: 10px 16px 10px 16px;
+				font-size: 0.875rem;
```
renancarvalho: According to the design, the font size is 1rem or 16px. Also, this is already being [set here,](https://github.com/Automattic/wp-calypso/blob/2fa774f46af739b806acdc8ab1ec2f4663b7418f/apps/happy-blocks/block-library/support-content-footer/style.scss#L206) I guess we can get rid of the font-size here.
The padding is being overwritten too. I haven't checked, maybe you can add your styles in the link above instead of having it again?

renancarvalho (commented):
Nice changes! I left a comment. :)

I think you might tackle it on another task or in this one, but the desktop spacing has also changed.

I think this is a design mistake, but there, there is no underline on the 'WordPress privacy' link text. 
image

apps/happy-blocks/block-library/support-content-footer/style.scss
escapemanuele: Checking here, it is 14px on mobile and 16px on desktop, or am I missing something?
image

I have to put the styles here, as this is for mobile and in the link you provided are the styles for desktop!

renancarvalho (approved):
Looks pretty! 🚢

PR #76442, opened 2023-04-30
Plans 2023: Partial list of TODOs for NPM migration

chriskmnds:
Related to #

## Proposed Changes

Code TODOs. Not for merging.

## Testing Instructions

*

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76491, opened 2023-05-02
Plans 2023: npm migration - useIsLargeCurrency updates

chriskmnds:
Related to https://github.com/Automattic/wp-calypso/pull/76442

## Proposed Changes

Tackles one of the TODOs in https://github.com/Automattic/wp-calypso/pull/76442. Moves `useIsLargeCurrency` under the grid folder and passes `siteId` to it.

## Testing Instructions

- Go to /plans/[free | paid] and /start/plans and ensure the grids render fine with correct prices
- Switch to INR locale and ensure plan prices do not overlap the column borders

## Pre-merge Checklist


client/my-sites/plan-features-2023-grid/components/plan-comparison-grid.tsx
```diff
@@ -448,12 +450,14 @@ const PlanComparisonGridHeader: React.FC< PlanComparisonGridHeaderProps > = ( {
...
+	siteId,
 } ) => {
 	const allVisible = visiblePlansProperties.length === displayedPlansProperties.length;
 
-	const isLargeCurrency = useIsLargeCurrency(
-		displayedPlansProperties.map( ( properties ) => properties.planName as PlanSlug )
-	);
+	const isLargeCurrency = useIsLargeCurrency( {
+		planSlugs: displayedPlansProperties.map( ( properties ) => properties.planName as PlanSlug ),
+		siteId: siteId,
```
ddc22: NIT
```suggestion
		siteId,
```

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -891,9 +893,12 @@ export class PlanFeatures2023Grid extends Component<
...
-		);
+		const isLargeCurrency = useIsLargeCurrency( {
+			planSlugs: ( props.planProperties || [] ).map(
+				( properties ) => properties.planName as PlanSlug
+			),
+			siteId: props.siteId,
+		} );
 		return <Component { ...props } isLargeCurrency={ isLargeCurrency } />;
 	};
 };
```
ddc22: NIT: Should this HOC live inside the file of `useIsLargeCurrency` ? 🤔

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -891,9 +893,12 @@ export class PlanFeatures2023Grid extends Component<
 
 const withIsLargeCurrency = ( Component: LocalizedComponent< typeof PlanFeatures2023Grid > ) => {
 	return function ( props: PlanFeatures2023GridType ) {
-		const isLargeCurrency = useIsLargeCurrency(
-			( props.planProperties || [] ).map( ( properties ) => properties.planName as PlanSlug )
-		);
+		const isLargeCurrency = useIsLargeCurrency( {
+			planSlugs: ( props.planProperties || [] ).map(
+				( properties ) => properties.planName as PlanSlug
```
ddc22: We already provide the plan name slugs array in the following prop.

https://github.com/Automattic/wp-calypso/blob/d4f1a05b13a27b47a33cad9b21cada117a294d39/client/my-sites/plan-features-2023-grid/index.tsx#L117-L117

Also we can probably change the type above to `Array<PlanSlug>` as well

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -891,9 +893,12 @@ export class PlanFeatures2023Grid extends Component<
 
 const withIsLargeCurrency = ( Component: LocalizedComponent< typeof PlanFeatures2023Grid > ) => {
```
ddc22, line 872: Given the feedback below. Should we maybe guard for a more generic property type here?

```typescript
type WithIsLargeCurrencyProps = {
        plans :  Array<PlanSlug>,
        siteId : number,
        [key: string]: any,
     }>
}

const withIsLargeCurrency = ( Component: LocalizedComponent< WithIsLargeCurrencyParam > ) => {

```

ddc22 (changes requested):
The feature works as expected. I left some comments about the implementation.

client/my-sites/plan-features-2023-grid/index.tsx
chriskmnds: Thanks for reviewing, @jdc91. I don't think the HOC should live in the same file as the hook. 

p.s. the HOC here is temporary until`PlanFeatures2023Grid` is refactored to a function component, at which point it will be removed. I wouldn't move it to its own file either.

client/my-sites/plan-features-2023-grid/index.tsx
chriskmnds: > We already provide the plan name slugs array in the following prop.

Good call. We do have these in `props.plans`. Updated.

> Also we can probably change the type above to Array<PlanSlug> as well

We can do that. 

p.s. these are being mapped to the data store type in https://github.com/Automattic/wp-calypso/pull/76086, which may end up having a different form (from the looks of it). It's still not fleshed out fully, but no plan types will be imported from `calypso-products` soon.

client/my-sites/plan-features-2023-grid/index.tsx
chriskmnds: I am not sure if this is necessary. I wouldn't bother, given the HOC is also temporary (and the context of this PR is at a different level). As far as usage goes, the injected property is added to the grid type, so it looks good to me for what it is. 

p.s. not sure if the suggested code above is correct either. If you give me the full correct typings, then happy to add them :-)

chriskmnds (commented):
Thanks for reviewing, @jdc91. Happy to see no breakage here.

client/my-sites/plan-features-2023-grid/index.tsx
ddc22: Interesting! Unrelated to this PR but I would like to dig into your thoughts behind 

>  I don't think the HOC should live in the same file as the hook.

If this HOC wasn't temporary would you think of it the same way?

ddc22 (approved):
LGTM 🚢

client/my-sites/plan-features-2023-grid/index.tsx
chriskmnds: Hey @jdc91. Asking the cool questions 😄  

> If this HOC wasn't temporary would you think of it the same way?

I believe so. So, I would rather not export a HOC (say, `withFoo`) from a directory of hooks `../hooks/use-foo.ts`, I guess in the same way that I'd not export a hook from a directory of selectors `../state/selectors/getFoo.ts`. 🤔 However, I can see that being part of a general design if done consistently (but is it ever?). I would not go out of my way to object, but I'd maybe raise it/suggest it. 

There are also times when I think it's ok for things to live next to each other, especially if smaller bits interdepended or part of some closed/isolated scope. An example I recall is this [context file](https://github.com/Automattic/wp-calypso/blob/trunk/packages/tour-kit/src/components/tour-kit-context.tsx) exporting both a provider and a hook to access it. I'd throw the HOC in there too. One could argue if even those could be broken out and exported from an index file (a bit too much maybe).

NITs? What do you think? :-)

ddc22: Right the way I mostly saw it is that the HOC was mainly targeted as a Class Component's way to access the given hook 😁. So in the interest of cohesion, having it closer to the hook as much as possible looked fitting. But I do see your point that it looks a bit unorthodox 😬 

> NITs?

True! But personally love to have these NIT conversations because I learn a lot from them 😀

chriskmnds: I guess even unorthodox ways can feel ok if done consistently and don't harm discoverability

PR #76497, opened 2023-05-02
Plans: Add more context when a plan and/or feature is selected

### LINKED ISSUE #75728: Newsletter: when picking paid plan from launchpad, plans page lacks direction
### Quick summary

The link from the global styles "you need an upgrade" step in the launchpad goes to the plans page where it's not clear which plan you need and why.

This worked when first launched - the step passes `plan=value_bundle&feature=advanced-design-customization` to the /plans page but has stopped working (along with all other uses of these GET vars) due to changes on the plans page.

### Steps to reproduce

1. Create a newsletter https://wordpress.com/setup/newsletter
2. Pick a custom color
3. In the launchpad note "need a plan" thingy

   image

4. After choosing the task, you see only 3 plans:

    image
URL is `https://wordpress.com/plans/SITE_SLUG?plan=value_bundle&feature=advanced-design-customization`

5. Click "pay monthly" button and more plans appear
    
    Screenshot 2023-04-13 at 23 28 26

Clearer next steps from picking the task. 

### What actually happened

I didn't see all plans, I didn't see which plan I needed for me. No features or anything was highlighted or explained.

### Impact

All

### Available workarounds?

_No response_

### Logs or notes

@Addison-Stavlo is this behaviour from the Launchpad, or something specific to the plans grid? Do you know whom to ping about the plans grid?

Some discussion: p1681461948232039-slack-C048CUFRGFQ

----

### Next steps
 - [ ] Talk to Marketing (or whoever owns the /plans page) to align on what needs to happen with the get vars and highlighting
 - [ ] Consider the language used in the /plans page and with global styles nudges and whether they could align better

mmtr:
Fixes https://github.com/Automattic/wp-calypso/issues/75728

## Proposed Changes

Before redesigning the plans page, we used to provide more context when a single plan and/or feature is "selected" with the `plan` and `feature` query params. You can check this by going to https://wordpress.com/plans?plan=value_bundle&feature=advanced-design-customization&flags=-onboarding%2F2023-pricing-grid.

Looks like we missed to replicate that in the new plans page, so this PR brings back these improvements:
- Show a "Suggested" label on the plan that has been selected.
- Show an additional header when some plans are hidden, so users can know there are more plans available.
- Highlight the selected feature.

**Main grid**
Desktop | Mobile 
--- | ---
Screenshot 2023-05-11 at 13 32 31 | Screenshot 2023-05-11 at 13 32 18

**Comparison grid**
Desktop | Mobile
--- | ---
Screenshot 2023-05-11 at 13 36 11 | Unchanged, it's difficult to highlight a feature here.

## Testing Instructions

- Go to `/plans/<SITE_DOMAIN>?plan=value_bundle&feature=style-customization`
- Make sure the Premium plan is labeled as "Suggested" instead of "Popular".
- Make sure the "Style customization" feature is highlighted.
- Make sure there is a header indicating that an upgrade is required to access the selected feature.
- Make sure there is a link to browse all plans.


simison:
Was this design from my suggestion in https://github.com/Automattic/wp-calypso/issues/75728#issuecomment-1508207116 ? Anyway, if you didn't yet, run it also by .com designer. ;-)

This is towards right direction. 👍  

This can also be a follow up PR, but it would be good to continue showing the "limited grid" even when switching monthly/yearly toggle. Now it feels buggy:

https://user-images.githubusercontent.com/87168/235698070-56459f4c-44e5-42c7-b2a7-b07e372ba1b7.mov

**Some design considerations —**

I would consider removing the secondary "best for devs" while "suggested" badge is visible. It's like we're recommending two things at the same time and "suggested" badge becomes less obvious  — which might be valid decision here, but feels like we dilute the primary feature-based recommendation.

image

Further down the plans page, when opening plan comparison, I expected to see "suggested" badge here:

Screenshot 2023-05-02 at 17 31 18

Even further down the grid, we could highlight again the feature they wanted:

image

mmtr:
> Was this design from my suggestion in https://github.com/Automattic/wp-calypso/issues/75728#issuecomment-1508207116 ? Anyway, if you didn't yet, run it also by .com designer. ;-)

Yup! And also based on how the previous "limited grid" looked like before the redesign. Good call about the .com designer review – I asked @lucasmendes-design to make a design pass on this (p1683122439624429-slack-C048CUFRGFQ).

> continue showing the "limited grid" even when switching monthly/yearly toggle

Nice catch! FWIW it was also misbehaving like that before the redesign. Anyway, it's relatively trivial to solve, so I fixed it in https://github.com/Automattic/wp-calypso/pull/76497/commits/00e9340d87d49c4bc9fd935cbb46a6534970dcfc

> consider removing the secondary "best for devs" while "suggested" badge is visible.

Another nice catch. This was already the case in the old layout where other badges were removed (e.g. the "popular" badge was not visible in the Premium plan if the Personal plan was selected: https://wordpress.com/plans?plan=personal-bundle&flags=-onboarding%2F2023-pricing-grid). Done in https://github.com/Automattic/wp-calypso/pull/76497/commits/af16be467bf3fa1cd8e29441633fa4a48c422334.

> Further down the plans page, when opening plan comparison, I expected to see "suggested" badge
> Even further down the grid, we could highlight again the feature they wanted

It makes totally sense! Handled in https://github.com/Automattic/wp-calypso/pull/76497/commits/a425da4a4712615ad0118d8c76b5709a82131e22.

client/my-sites/plan-features-2023-grid/components/plan-comparison-grid.tsx
```diff
@@ -293,6 +308,8 @@ type PlanComparisonGridProps = {
 	canUserPurchasePlan: boolean;
 	selectedSiteSlug: string | null;
 	onUpgradeClick: ( properties: PlanProperties ) => void;
+	selectedPlan?: string;
+	selectedFeature?: string;
```
ddc22, line 318: NIT: Can we add a few lines of comments, explaining this feature and the prop.

ddc22 (commented):
Thanks for picking up this change. Clearly this slipped through the cracks on the redesign. Code looks great. The feature works as explained as well. Minor suggestion added below. Differing to @lucasmendes-design for any design feedback.

BogdanUngureanu:
Tests well for me. It would be nice to also open the feature list by default on mobile. Right now the feature list is collapsed, and when you click on it, it will show the feature highlighted. 

Screenshot 2023-05-08 at 15 54 33

Screenshot 2023-05-08 at 15 56 51

mmtr:
> open the feature list by default on mobile

Good idea! Done in https://github.com/Automattic/wp-calypso/pull/76497/commits/3c5c650469c799f36fdd0a6ce83b91fd11fb6fc4

client/my-sites/plan-features-2023-grid/components/plan-comparison-grid.tsx
mmtr: Sure thing. Although these props are actually inherited and passed down from multiple components above, so I added the comments to `PlanFeatures2023GridProps` which is the component at the top level. See https://github.com/Automattic/wp-calypso/pull/76497/commits/37d204b4f3905fd2db0fed7a546517d943dcc61f.

BogdanUngureanu (approved):
Tests well for me!

lucasmendes-design:
Instead of a yellow background, we can simplify it and have it with bold text only. This is how we highlight features in the new plans grid. Changing the background is a bit strange, IMO. This selected feature will work in a similar way of "Free domain for one year". 

In addition, we can place the feature at the very top of the list. 

Desktop:
image

Mobile:
Reorder and put the feature at the very top of the list and the text is bold. 

cc: @mmtr

lucasmendes-design:
Do we "select" more than one feature? Is that common?

mmtr:
> Do we "select" more than one feature? Is that common?

Nope, only one. They are selected with the `?feature=` query param, and it only supports a single feature.

lucasmendes-design:
@mmtr thanks. So, this solution will work. I was afraid that we could select 10 features at once 😆

simison:
@lucasmendes-design what I'd like to be able to do in future is render plans grid inside product, which is specific to Newsletters, blogs, etc, based on either onboarding flow they came through, or some other link or upsell inside the product.

lucasmendes-design:
> what I'd like to be able to do in future is render plans grid inside product, which is specific to Newsletters, blogs, etc, based on either onboarding flow they came through, or some other link or upsell inside the product.

The initial idea was to integrate the plans grid inside the product, so we created a flexible UI system to render the plans grid everywhere. ([P2](https://pricingandpackaging.wordpress.com/2022/09/08/pricing-page-ideas-for-different-touch-points/#comment-2854)). To tailor the information based on onboarding flow and upsell I think we need to align with devs but I don't think this is a big issue. cc: @simison

mmtr:
Thanks @lucasmendes-design for your suggestion on moving the selected feature to the top and have it with bold text only. I did that in the last commits, in addition to remove the bold text from any other feature, and this is how it currently looks:

Desktop | Mobile
--- | ---
Screenshot 2023-05-11 at 13 32 31 | Screenshot 2023-05-11 at 13 32 18

Additionally, I added a neutral background to the feature in the comparison grid displayed below so it's highlighted there as well, but I'm happy to remove/improve if you think this is not a good idea:

Screenshot 2023-05-11 at 13 36 11

BogdanUngureanu:
Hey @mmtr, the changes look nice on the desktop (although I would've liked a more prominent highlight of the `Style customization`). However, I've noticed that the header texts look a bit off.

Screenshot 2023-05-11 at 18 59 22

mmtr:
> However, I've noticed that the header texts look a bit off.

Can you expand on that? With header texts you mean the "Upgrade your plan to access..." text? What does look off exactly?

We're using the same header we were showing in the old layout:
Screenshot 2023-05-12 at 11 08 54

BogdanUngureanu:
Hey Miguel, I was talking about the `Upgrade your plan to access this feature and more" and the `View all plans` link. 

On desktop resolutions, there's a bit of margin between them and the texts are centered. On mobile, the texts are on the left (including the view all plans link) and there's no small spacing between them.

It's not a huge problem, it just looked a bit off to me (it might be a subjective thing).

mmtr:
Oh I see, thanks for clarifying. I'd say that's unrelated to this PR since the header was introduced a long ago. Before the Plans page redesign, we were already displaying it aligned to the left and with less separation between the text and the link. We can improve it in a follow-up if needed.

BogdanUngureanu:
Sounds fair! Let's 🐑  it.

lucasmendes-design:
> Additionally, I added a neutral background to the feature in the comparison grid displayed below so it's highlighted there as well, but I'm happy to remove/improve if you think this is not a good idea:

I think we can use a subtle color for the comparison grid. cc: @mmtr 

**BG:** #FAFAFA 
**Font color:** #3C434A 

image

BogdanUngureanu:
Had a look again and it LGTM!

PR #76505, opened 2023-05-02
Cleaned up line clamping for recommended posts in Reader

davemart-in:
## Description

@ebinnion highlighted that the line clamping we were using for the recommended site cards was cutting content off at a specific height, which is not ideal. I took this opportunity to clean up quite a bit of code and replaced it with modern line clamping CSS.

### Before

before

### After

after

## Testing Instructions

- Apply this PR
- Head to http://calypso.localhost:3000/following/manage
- Just use dev tools to add more than 3 lines of text vs. trying to refresh until you find a card with enough text

### DIFF
--- a/client/blocks/reader-list-item/style.scss
+++ b/client/blocks/reader-list-item/style.scss
@@ -85,23 +85,11 @@
 .reader-list-item__by-text,
 .reader-list-item__site-excerpt,
 .reader-list-item__site-url-timestamp {
-	display: block;
-	line-height: 18px;
-	max-height: 16px * 2.6;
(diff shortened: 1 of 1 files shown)


creativecoder (approved):
Works well for me, my tests match the screenshots in the PR description. Tested on Firefox and Safari.

TIL you can use line-clamp with the webkit prefix and it works across current browsers--nice!

davemart-in:
> TIL you can use line-clamp with the webkit prefix and it works across current browsers--nice!

Yep! Thankfully it's one that most browsers have adopted.

Issue #75928, opened 2023-04-18
Stats: Subscribers - add an overview below the chart


grzegorz-cp:
This section looks like it's connected to the chart, especially in the context of the Traffic page, but it isn't.
The section below the Visitors chart on the Traffic page functions as tabs, but the elements on the Subscribers page are not interactive.

Create a new component showing the number of subscribers for today, 30, 60 and 90 days ago. Don't place it in the chart section but directly on the subscribers page between the chart and the grid.
Consider the impact of retrieving data for 90 days and check if you can make four calls to get data for a specific day and if it's a better solution.

This section uses the exact CSS as the top `All time stats`. Don't copy the styles. If the `All times stats` component doesn't do it already, abstract the boxes into a separate component and use it for both this and `All times stats` components instead of copying CSS (in case this component is in the stats directory and the other one is in Storybook components).

# Design
- Widescreen
![Image](https://user-images.githubusercontent.com/112354940/232926608-8f993bf1-ed6c-40b7-994d-feb13c988c20.png)

- Mobile
![Image](https://user-images.githubusercontent.com/112354940/232926639-86fb95d8-cfd1-4cef-ab92-7424395d958a.png)

# Development
- [x] add a new component below the chart https://github.com/Automattic/wp-calypso/pull/76436
- [x] add data to component from mock API https://github.com/Automattic/wp-calypso/pull/76507
- [ ] share CSS between this and the top `All times stats` components [(moved here)](https://github.com/Automattic/wp-calypso/issues/80107) 
- [ ] apply shortened numbers to all screen sizes unless decided differently by the designers [(moved here)](https://github.com/Automattic/wp-calypso/issues/80109)
- [x] make sure that the component scrolls correctly for small screens

PR #76507, opened 2023-05-02
Stats: Adds data to subscribers overview section

annacmc:
Related to #75928 

This PR includes improvements to the Subscribers Overview section. The changes made are outlined as follows:

- adds data from the mock API endpoint into the subscriber's overview cards
-  moved from making a single API request for all 90 days of stats and filtering out the required data to making individual queries for each required day (0, 30, 60, and 90 days ago). This approach, while increasing the number of requests, significantly reduces the amount of unnecessary data queried.
- migrated the section and react-query hook from JavaScript to TypeScript

## Testing Instructions

* Open the live branch for this PR.
* Navigate to `/stats/day/:siteSlug`.
* Add `?flags=stats/subscribers-section` to the URL to enable the subscriber stats page.
* Scroll down to locate the overview cards (immediately under the chart)
* Check that they load data correctly, still behave acceptably on resize etc. 
 
image

## Pre-merge Checklist


client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,90 @@
...
-	return overviewCardStats;
+interface SubscribersData {
+	period: string;
+	subscribers: number;
+	subscribers_change: number;
 }
 
-function SubscribersOverviewCards() {
-	const overviewCardStats = SubscribersOverviewCardStats();
+interface SubscribersDataResult {
```
grzegorz-cp, line 21: Those two definitions are also defined in `subscribers-section`. Let's avoid duplication. We can either export it from there and import it here, or a better option would be to keep it with the hook because more places will use it.
You could either convert the hook to TS and export it from there (this would be best) or create a file with types and place it in the hook directory.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,90 @@
...
+	siteId: number;
+}
+
+function extractCountsAtIndexes( subscribersData: SubscribersData[], indexes: number[] ): number[] {
+	return indexes.map( ( index ) => subscribersData[ index ]?.subscribers || 0 );
+}
+
+// all comparisons are being compared to todays's count to show growth up to today
+function SubscribersOverviewCardStats( subscribersData: number[] ) {
+	const daysToDisplay = [ 0, 30, 60, 90 ];
```
grzegorz-cp: If you plan to use indexes, let's avoid magic numbers and put them into a constant. This way, you can reuse the constant, and everything is in one place and has a name.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,90 @@
...
+	const overviewCardStats: {
+		heading: string;
+		count: number;
+		previousCount?: number;
+	}[] = [];
+
+	daysToDisplay.forEach( ( day, index ) => {
```
grzegorz-cp: This logic works, and it would be great if we had a significant number of modules.
With four components, I prefer the previous approach with an object with all 4 of them written down instead of a loop.
This is a personal preference, so you can go with this function, but I think the other one is more readable.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,90 @@
...
+		const count = subscribersData[ index ] || 0;
+
+		const cardStat = {
+			heading: ( index === 0
+				? translate( 'Today' )
+				: translate( '%d days ago', { args: day } ) ) as string,
+			count: count,
```
grzegorz-cp: `previousCount` is missing at the moment.

grzegorz-cp (commented):
Nice progress!
It's great that you added TS and that you wanted to optimise the way the cards array is created.
I left a few comments, and most of them are personal preferences, so if you have strong feeling about it, you don't have to change the code, but please add more comments.

I'm not sure about the approach towards getting data. Asking for 90 days to show four values is one thing, but you also rely on the data coming in the correct order and not having any gaps.
I wonder if making four queries would be more effective (if possible). This way, we gain more flexibility if, for some reason, we want to add more lines to one chart, and we are not locked to a specific shape of the whole dataset.

client/my-sites/stats/stats-subscribers-overview/index.tsx
annacmc: correct. After chatting with @jsnmoon yesterday afternoon I decided to pull out the previous count change indicators and move my doubts to a P2 which I will be posting today and will hopefully explain everything for now 😄

client/my-sites/stats/stats-subscribers-overview/index.tsx
annacmc: I agree with you here in regards to readability. I'm working on the four component approach today and think I will simplify things back to things without the loop.

client/my-sites/stats/stats-subscribers-overview/index.tsx
annacmc: Good idea, thanks 👍

client/my-sites/stats/stats-subscribers-overview/index.tsx
jsnmoon: Anna is correct — our current design specifications lead to an undesirable/misleading interface for a site with growing subscriber counts. Definitely merits further discussion in a P2 🙂

client/my-sites/stats/hooks/use-subscribers-query.tsx
```diff
@@ -1,25 +1,40 @@
...
+	// todo: add type
+): Promise< any > {
+	const query: { unit: string; quantity: number; http_envelope: number; date?: string } = {
+		unit: period,
+		quantity,
+		http_envelope: 1,
+	};
+
+	if ( date ) {
+		query.date = date;
```
grzegorz-cp: This changes the logic of the hook. Previously the date was filled with the current date if no date was passed. Was this change intentional?

I'm also concerned that previously the hook consumers only had to pass a Date object (and the internal hook logic adjusted it to the format it needed - `periodStart.toISOString().slice( 0, 10 )`), but now all of the consumers have to do the conversion. This introduces more points to track where the conversion happens and potentially more places where this could introduce issues. 

If the change is intentional, other places using the hook have to have the date updated and converted to a proper string.

client/my-sites/stats/hooks/use-subscribers-query.tsx
```diff
@@ -1,25 +1,40 @@
...
-			http_envelope: 1,
-			date: periodStart.toISOString().slice( 0, 10 ), // returns UTC time
-		}
+		query
 	);
 }
 
-function selectSubscribers( payload ) {
+export function selectSubscribers( payload: {
+	date: string;
```
grzegorz-cp, line 34: Do you think it would be beneficial to move this to an interface declaration at the top of the file?
This could be combined with my other comment about centralising the type definitions.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,128 @@
...
+		return translate( '%d days ago', { args: daysAgo } );
+	};
 
-function SubscribersOverviewCardStats() {
 	const overviewCardStats = [
+		// Card for Today
 		{
-			heading: translate( 'Today' ),
-			count: 1410,
+			heading: getHeading( 0, indexFirstCard ),
```
grzegorz-cp: I personally prefer your previous approach where it was a plain `translate` specific to each of the objects. It was more readable. 
The `index` variable is introduced only to distinguish between `today` and the other headers. The previous solution avoids this and makes it easier to find the translated string.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,128 @@
...
+	date.setDate( date.getDate() - daysToSubtract );
+	return date.toISOString().split( 'T' )[ 0 ];
+}
+
+// calculate the stats to display in the cards
+function SubscribersOverviewCardStats( subscribersData: SubscribersData[][] ) {
+	const getCount = ( index: number ) => {
+		const innerArray = subscribersData[ index ];
+		if ( Array.isArray( innerArray ) && innerArray.length > 0 ) {
+			return innerArray[ 0 ].subscribers;
```
grzegorz-cp: Looks like this function might be the same as `return subscribersData[ index ]?.[0]?.subscribers || 0`.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,128 @@
...
 	return overviewCardStats;
 }
 
-function SubscribersOverviewCards() {
-	const overviewCardStats = SubscribersOverviewCardStats();
+const SubscribersOverview: React.FC< SubscribersOverviewProps > = ( { siteId } ) => {
+	const period = 'day';
+	const quantity = 1;
+	const dates: string[] = [ dateToday, date30DaysAgo, date60DaysAgo, date90DaysAgo ];
+	const subscribersQueries = useQueries( {
```
grzegorz-cp, line 77: This data fetching should be moved either to the original hook (that can export both data fetching hooks) or to a new hook file.
`querySubscribers` and `selectSubscribers` should not be manipulated by components consuming data (this component). Abstracting it to a different file creates a separation between the data handling mechanism and the consumers, which shouldn't be concerned about the API.

grzegorz-cp (dismissed):
Good to see the progress!

Almost all of my comments can be addressed in future pull requests apart from the changes to the date inside `querySubscribers`. I believe it should still accept a Date object and convert it to the needed format inside (doing so also prevents you from having to update the chart).

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
 import { CountComparisonCard } from '@automattic/components';
+import { UseQueryResult, useQueries } from '@tanstack/react-query';
 import { translate } from 'i18n-calypso';
 import React from 'react';
+import {
+	querySubscribers,
+	selectSubscribers,
+} from 'calypso/my-sites/stats/hooks/use-subscribers-query';
+
+const indexFirstCard = 0; // 0 for today
```
kangzj, line 10: 1/3 I can see when the days are used, they are used as an array - I'm just wondering, would it be better to use an array here to avoid so many variables.

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
...
+	unit: string;
+	date: string;
+}
+
+interface SubscribersOverviewProps {
+	siteId: number;
+}
+
+// calculate the dates to query for
+const dateToday = calculateQueryDate( indexFirstCard );
```
kangzj, line 32: 2/3 And then do a `map` here to get dates as an array too

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
...
+// calculate the date to query for based on the number of days to subtract
+function calculateQueryDate( daysToSubtract: number ) {
+	const today = new Date();
+	const date = new Date( today );
+	date.setDate( date.getDate() - daysToSubtract );
+	return date.toISOString().split( 'T' )[ 0 ];
+}
+
+// calculate the stats to display in the cards
+function SubscribersOverviewCardStats( subscribersData: SubscribersData[][] ) {
```
kangzj, line 46: I would prefer a different name for `subscribersData` as it's misleading - only when `subscribersData` is of type `SubscribersData`, we'd use the name IMO. `subscribersData` here is actually a two dimensional array of SubscribersData. I would use something like `subscribersDataArrays` but I'm not very sure to be honest...

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
...
+
 	return overviewCardStats;
 }
 
-function SubscribersOverviewCards() {
-	const overviewCardStats = SubscribersOverviewCardStats();
+const SubscribersOverview: React.FC< SubscribersOverviewProps > = ( { siteId } ) => {
+	const period = 'day';
+	const quantity = 1;
+	const dates: string[] = [ dateToday, date30DaysAgo, date60DaysAgo, date90DaysAgo ];
```
kangzj, line 76: 3/3 then it's still an array

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
...
-					count={ overviewCardStat.count }
-					previousCount={ overviewCardStat.previousCount }
-					showValueTooltip
-					icon={ false }
-				/>
-			) ) }
-		</div>
-	);
-}
+	const isLoading = subscribersQueries.some( ( result ) => result.isLoading );
```
kangzj, line 86: What do you say if we deal the errors and loading status separately, because we still want to show the working cards even when some of them errored out or are still loading.

client/my-sites/stats/hooks/use-subscribers-query.tsx
```diff
@@ -1,25 +1,42 @@
...
-function querySubscribers( siteId, period, quantity, date ) {
-	const periodStart = date ? date : new Date();
+export function querySubscribers(
+	siteId: number,
+	period: string,
+	quantity: number,
+	date?: string
+): Promise< any > {
+	// Create a date object from the passed date string or create a new date object if no date is provided.
+	const periodStart = date ? new Date( date ) : new Date();
```
kangzj, line 11: Just a note here for the usage of `today`, no need to change anything. We'll consider this all together later.

https://github.com/Automattic/wp-calypso/issues/76912

client/my-sites/stats/stats-subscribers-overview/index.tsx
```diff
@@ -1,57 +1,113 @@
...
+}
+
+interface SubscribersDataResult {
+	data: SubscribersData[];
+	unit: string;
+	date: string;
+}
+
+interface SubscribersOverviewProps {
+	siteId: number;
```
kangzj:
```suggestion
	siteId: number | null;
```

client/my-sites/stats/stats-subscribers-chart-section/index.tsx
```diff
@@ -77,7 +77,7 @@ export default function SubscribersChartSection( {
 		siteId,
 		period,
 		quantity,
-		queryDate
+		queryDate.toISOString()
```
grzegorz-cp: Let's avoid changing the parameter when executing a function.
Can you keep this as a `Date` object? This way, we avoid converting a date to a string, a date and a string again.

grzegorz-cp: If a string is needed, convert it inside `useSubscribersQuery` instead of the function execution here.

grzegorz-cp (approved):
👍 Let's ship and iterate on the feedback.

PR #76525, opened 2023-05-03
Theme Showcase: Fix live demo loading for 3rd-party themes

taipeicoder:
## Proposed Changes

This PR fixes the issue where the live demo URL for 3rd-party themes would not load correctly in the preview iframe. This is because the preview iframe expects its embedded page to send a `partially-loaded` message. Since the demo URL is externally managed, it doesn't send the `partially-loaded` message, and so the component times out by opening the demo URL in a separate browser tab. 

In order to avoid the weird user experience, this PR proposed to open 3rd-party theme demo sites directly in a new tab, instead of opening the preview modal first.

See the attached screenshots to see how to access a theme's live demo:
| Theme card | Theme detail page |
| --- | --- |
| Screenshot 2023-05-03 at 11 31 10 AM | Screenshot 2023-05-03 at 11 31 21 AM |

## Testing Instructions

* Head to the Theme Showcase.
* Select any third-party theme, such as Scentina. Third-party themes are labeled with the "Paid" badge. 
* Click on the Live Demo for the theme.
* Ensure that the preview modal opens the demo URL in a new browser tab, instead of opening the preview modal first.
* Also test in the logged-out Theme Showcase.
* Also test with a first-party theme and ensure that it works as before.

## Pre-merge Checklist

- [x]  Have you checked for TypeScript, React or other console errors?

### DIFF
--- a/client/my-sites/themes/theme-preview.jsx
+++ b/client/my-sites/themes/theme-preview.jsx
@@ -21,14 +21,13 @@ import {
 	isThemeActive,
 	isInstallingTheme,
 	isActivatingTheme,
+	isExternallyManagedTheme as getIsExternallyManagedTheme,
 } from 'calypso/state/themes/selectors';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 import { connectOptions } from './theme-options';
 
 const DEFAULT_VARIATION_SLUG = 'default';
 const isDefaultVariationSlug = ( slug ) => ! slug || slug === DEFAULT_VARIATION_SLUG;
-const isUrlWpcomApi = ( url ) =>
-	url.indexOf( 'public-api.wordpress.com/wpcom/v2/block-previews/site' ) >= 0;
 
 class ThemePreview extends Component {
 	static displayName = 'ThemePreview';
@@ -217,7 +216,8 @@ class ThemePreview extends Component {
 	};
 
 	render() {
-		const { themeId, siteId, demoUrl, children, isWPForTeamsSite } = this.props;
+		const { themeId, siteId, demoUrl, children, isExternallyManagedTheme, isWPForTeamsSite } =
+			this.props;
 		const { showActionIndicator, showUnlockStyleUpgradeModal } = this.state;
 
 		if ( ! themeId || isWPForTeamsSite ) {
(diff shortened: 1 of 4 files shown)


client/my-sites/themes/theme-preview.jsx
```diff
@@ -27,6 +27,8 @@ import { connectOptions } from './theme-options';
 
 const DEFAULT_VARIATION_SLUG = 'default';
 const isDefaultVariationSlug = ( slug ) => ! slug || slug === DEFAULT_VARIATION_SLUG;
+const isUrlWpcomApi = ( url ) =>
+	url.indexOf( 'public-api.wordpress.com/wpcom/v2/block-previews/site' ) >= 0;
```
arthur791004: I guess we can simply check the origin is `public-api.wordpress.com`. Or we can use `isExternallyManagedTheme` to check whether it's 3rd party theme. WDYT?

taipeicoder: That's much better, thanks for the suggestion!

arthur791004:
Do you ever encounter this issue when opening the live demo of the Scentina or Nokul?

image

image

miksansegundo:
I reproduced the error when opening the live demo of Scentina and Nokul. It happens on this PR and on production too. The difference is that on this PR the demo site doesn't open on a new tab.

Screenshot 2566-05-03 at 16 16 48

Screenshot 2566-05-03 at 16 17 39

taipeicoder:
Seems that the demo sites for Scentina and Nokul are set with the CSP:

```
themeskingdom.com tk-2020.local 2020.themeskingdom.com tkdvl.com
```

Screenshot 2023-05-03 at 5 35 55 PM

Which prevents us from embedding them in an iframe 😞. After some investigation, I couldn't find a good way to detect which sites can be embedded or which sites cannot. Perhaps we can follow what WooCommerce https://woocommerce.com/document/scentina/ and open the demo in a new tab for third-party themes?

cc: @ianstewart

taipeicoder:
@arthur791004 @miksansegundo since there are issues with 3rd-party theme demo sites not being embeddable, as discussed in p1683107174351239/1683056613.727579-slack-C048CUFRGFQ, the short term solution is to just open those demo sites directly in a new tab. I've updated the PR description accordingly.

arthur791004 (approved):
Tested and looks good to me 👍

PR #76527, opened 2023-05-03
Unlocks Unified Importer for Jetpack users

zaerl:
This PR unlocks the Unified Importer for Jetpack users.

## Proposed Changes

* Unlock the Unified Importer for Jetpack users.
* The version compare now is done using the `version_compare` function to have a more granular control[^1].

## Testing Instructions

We have four scenarios to be tested here for the `http://calypso.localhost:3000/import/SITE_URL` page:

1. A **simple site**: a list of importers that call the simple importer.
2. An **atomic site**: a list of importers that call the Unified Importer.
3. A **Jetpack site** with minimum version **12.1:** a list of importers that call the Unified Importer.
4. A **Jetpack site** with version **12.0 or older**: a list of greyed-out importers and a warning that cite: "Your Jetpack plugin is out of date. To make sure you can import reliably, update Jetpack." (see screenshot).

For case three, a brand new JN site is ok.  For case four, if you already don't have one, you can create a JN site using a specific old enough version.

image

[^1]: https://www.php.net/manual/en/function.version-compare.php

### DIFF
--- a/client/blocks/jetpack-plugin-update-warning/index.tsx
+++ b/client/blocks/jetpack-plugin-update-warning/index.tsx
@@ -1,9 +1,10 @@
 import { useTranslate } from 'i18n-calypso';
 import { FC, useCallback, useMemo, useState } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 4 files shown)


ouikhuan (approved):
LGTM 🚢

Issue #76333, opened 2023-04-27
Assembler: April 2023 copy updates


autumnfjeld:
# Description

Following , [Colors & Fonts Walkthrough Tasks – Team Lego] pbxlJb-3EX-p2   a thorough copy review of the Assembler was done and is now ready for dev.

# Dev Tasks
Consult the Figma and make the "final" updates.

Figma: 7l53h5fxAUNjVRBQ82YFwn-fi-2424%3A60507

Look in the Final columns
> the “final” columns are now all done. If they are empty that means no change.
![image](https://user-images.githubusercontent.com/5665959/234762445-d75be9aa-bfd4-4939-abb0-566775527639.png)

Check with @autumnfjeld or @davewhitley if you have questions.

arthur791004:
Closing it via https://github.com/Automattic/wp-calypso/pull/76528. And we will continue to improve the copy. See https://github.com/Automattic/wp-calypso/issues/76614

PR #76528, opened 2023-05-03
Pattern Assembler: May 2023 Update copy

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/76333

## Proposed Changes

* Update the final copy as requested. See 7l53h5fxAUNjVRBQ82YFwn-fi-2424_60507

**CTA**

image

**Main**

image

**Screens**

| Header | Homepage | Footer | Colors | Fonts |
| - | - | - | - | - |
| image | image | image | image | image |

**Large Preview**

| Default | Select Colors or Fonts without any pattern |
| - | - |
| image | image |

**Continue**

| Default | Colors | Fonts | Both | Upgrade later |
| - | - | - | - | - |
| image | image | image | image | image |

**Upsell Modal**

| Colors | Fonts | Both |
| - | - | - |
| image | image | image |

## Testing Instructions

* Go to /setup/site-setup?siteSlug=<your_free_site>
* On the Pattern Assembler screen, verify the copy on each screen.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
@@ -1,7 +1,6 @@
 import { PatternRenderer } from '@automattic/block-renderer';
 import { DeviceSwitcher } from '@automattic/components';
 import { useStyle } from '@automattic/global-styles';
-import { useHasEnTranslation } from '@automattic/i18n-utils';
 import { __experimentalUseNavigator as useNavigator } from '@wordpress/components';
 import { Icon, layout } from '@wordpress/icons';
 import classnames from 'classnames';
@@ -44,7 +43,6 @@ const PatternLargePreview = ( {
 	recordTracksEvent,
 }: Props ) => {
 	const translate = useTranslate();
-	const hasEnTranslation = useHasEnTranslation();
 	const navigator = useNavigator();
 	const hasSelectedPattern = header || sections.length || footer;
 	const shouldShowSelectPatternHint =
@@ -68,33 +66,6 @@ const PatternLargePreview = ( {
 		event.preventDefault();
 		recordTracksEvent( PATTERN_ASSEMBLER_EVENTS.LARGE_PREVIEW_ADD_HEADER_BUTTON_CLICK );
 		goToSelectHeaderPattern();
-	};
-
-	const getDescription = () => {
-		if ( ! shouldShowSelectPatternHint ) {
-			return translate( "It's time to get creative. Add your first pattern to get started." );
-		}
-
-		const options = {
-			components: {
-				link: (
-					// eslint-disable-next-line jsx-a11y/anchor-is-valid
-					<a href="#" target="_blank" rel="noopener noreferrer" onClick={ handleAddHeaderClick } />
-				),
-			},
-		};
-
-		return hasEnTranslation(
-			'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}'
-		)
-			? translate(
-					'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}',
-					options
-			  )
-			: translate(
-					'You can view your color and font selections after you select a pattern, get started by {{link}}adding a header pattern{{/link}}',
-					options
-			  );
 	};
 
 	const renderPattern = ( type: string, pattern: Pattern, position = -1 ) => {
@@ -218,8 +189,27 @@ const PatternLargePreview = ( {
 			) : (
 				<div className="pattern-large-preview__placeholder">
 					<Icon className="pattern-large-preview__placeholder-icon" icon={ layout } size={ 72 } />
-					<h2>{ translate( 'Welcome to your blank canvas' ) }</h2>
-					<span>{ getDescription() }</span>
+					<h2>{ translate( 'Ready to start designing? Add some elements from the sidebar.' ) }</h2>
+					{ shouldShowSelectPatternHint && (
+						<span>
+							{ translate(
+								'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}',
+								{
+									components: {
+										link: (
+											// eslint-disable-next-line jsx-a11y/anchor-is-valid
+											<a
+												href="#"
+												target="_blank"
+												rel="noopener noreferrer"
+												onClick={ handleAddHeaderClick }
+											/>
+										),
+									},
+								}
+							) }
+						</span>
+					) }
 				</div>
 			) }
 		</DeviceSwitcher>
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
@@ -68,6 +68,37 @@ const PatternLargePreview = ( {
 		goToSelectHeaderPattern();
 	};
 
+	const getTitle = () => {
+		if ( ! shouldShowSelectPatternHint ) {
+			return translate( 'Welcome to your blank canvas.' );
+		}
+
+		return translate( 'Ready to start designing?' );
+	};
+
+	const getDescription = () => {
+		if ( ! shouldShowSelectPatternHint ) {
+			return translate( "It's time to get creative. Add your first pattern to get started." );
+		}
+
+		return translate(
+			'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}',
+			{
+				components: {
+					link: (
+						// eslint-disable-next-line jsx-a11y/anchor-is-valid
+						<a
+							href="#"
+							target="_blank"
+							rel="noopener noreferrer"
+							onClick={ handleAddHeaderClick }
+						/>
+					),
+				},
+			}
+		);
+	};
+
 	const renderPattern = ( type: string, pattern: Pattern, position = -1 ) => {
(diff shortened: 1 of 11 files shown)


miksansegundo (approved):
LGTM 👍

I verified that the copy is updated.

Maybe for this or next iterations we can use the following suggestions to improve the copy. cc: @lucasmendes-design @autumnfjeld  

### Avoid orphan words for the English locale and inconsistency between buttons
#### Case 1

|BEFORE|AFTER|
|-----|-----|
|Screenshot 2566-05-03 at 19 30 37|Screenshot 2566-05-03 at 17 46 09|

While this copy is shorter to avoid the orphan, it also avoids the redundancy of "continue".

#### Case 2
|BEFORE|AFTER|
|-----|-----|
|Screenshot 2566-05-03 at 17 04 27|Screenshot 2566-05-03 at 17 05 23|

In this case the copy is shorter and also helpful because reminds users that in the assembler they can only create the layout of a homepage, they cannot edit the content. Plus, it's also inviting users to continue because there are much more things they can do in the editor.

#### Case 3
|BEFORE|AFTER|
|-----|-----|
|Screenshot 2566-05-03 at 17 19 33|Screenshot 2566-05-03 at 17 18 48|

In this case the copy avoids adding the new term "palette" because we don't use that term in the editor or the theme showcase AFAIK when we users can choose style variations. See screenshot.

Screenshot 2566-05-03 at 18 37 06

#### Case 4
|BEFORE|AFTER|
|-----|-----|
|Screenshot 2566-05-03 at 17 06 13|Screenshot 2566-05-03 at 17 08 48|

In this case the copy is shorter thanks to removing "useful" (IMO it's not adding much value because not only the footer links on a website are useful) and replacing "information" with "info" (because I found that WordPress has a block called "Contact info". See screenshot)

Screenshot 2566-05-03 at 18 56 28

### Avoid adding new terms and styling inconsistency between screens
#### Case 1
|BEFORE|AFTER|
|----|----|
|Screenshot 2566-05-03 at 17 56 22|Screenshot 2566-05-03 at 17 32 28|

In this case, adding a subheader creates visual consistency with the placeholder on the styles screens.

I'm sure there are good reasons to use the term "elements". I'd rather use "layout and style options" because these are precise references to the labels users can see on the sidebar and suggests that they are optional.

#### Case 2

|BEFORE|AFTER|
|----|----|
|Screenshot 2566-05-03 at 18 01 39|Screenshot 2566-05-03 at 17 17 45|

On the placeholder for the Colors and Fonts screens, the sidebar contains colors or fonts elements:
 
Screenshot 2566-05-03 at 18 19 06

If a user adds these "elements" from the sidebar the preview won't show anything because the user must add patterns. Therefore, on the styles screens that sentence is not adding much value and by removing it we create a visual consistency with the placeholder on the other screens.

lucasmendes-design:
Hi Miguel, all the proposed copies are better than the previous ones. Use case 2 makes a lot of sense to me since our users don't know that they will add content after this step. 

**Footer and Header:** It's beneficial to mention that these components will always be visible. I think this is more valuable than"usually links and contact info". 

**Colors and fonts:** We can block users from selecting the color. 
 
Solution 1: we don't have an inactive state but it's something to consider. 
image

Solution 2: show on the SA page that users must select a pattern first. 
image

cc: @davewhitley @autumnfjeld

davewhitley:
I don't really have a problem with "palettes" because it's a common word that people understand. Much more so than "variation" 😄  . I like your suggestions though anyway.

The biggest feedbacks I have are:
- We shouldn't use "pattern" here so much, because people don't know what that is, and this isn't a good time for them to learn WordPress terminology as well as more common site terms like "header". I'm not saying we _shouldn't_ use it, but just use it sparingly. So in in Case 2(B)(the second one), we can avoid "pattern" by saying something like "You can view your color and font selections after you select a layout option. Get started by adding a header."
- "**Ready to add your content and much more?**" — this sounds a bit awkward and "content" is not a great word to use. Maybe: "Ready to keep editing?"
-

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7966546

Thank you @arthur791004 for including a screenshot in the description! This is really helpful for our translators.

autumnfjeld:
> We shouldn't use "pattern" here so much, because people don't know what that is, and this isn't a good time for them to learn WordPress terminology as well as more common site terms like "header". 

I like to think of the Assembler as an education tool for some basic WP concepts, patterns definitely being one of them.  When the user gets to the editor, I think they should clearly understand what patterns are and clearly link this Pattern term in the editor's editors inserter with what they have seen in the PA.    

image

autumnfjeld:
Some of the changes result in awkward visual effects, like @miksansegundo  mentioned above with header/subheader and consistency and I don't think this would have been clear to the people making suggestions in Figma as they weren't seeing the true visual impact of the copy changes.  So let's fix some of these things.


Four requests @arthur791004 
1. Undo the changes of the copy in the Blank Canvas.  Let's go back to what it is in production.  
image


2. Update the GS modal title to match Figma:
image

3. Delete the sentence "Add some elements form the sidebar." Should look like this:
image

4. Remove word "usually" since the footer always appears at the bottom (as far as I know?!?! 😄 ). Should be like this:
image



---------

@miksansegundo Thank you for the detailed feedback about copy!!! Very helpful. 


Some thoughts:

* I think info is too informal, even if it was found somewhere else.  
* I previously mentioned the "orphaned" words to @davewhitley , I think it was Dave, and I will roughly repeat his answer: we shouldn't make copy decisions based on english orphans because we can't design just for English.....all the other languages may or may not have the same issue.  Even the superfluous "u" 😆  in British English could knock a word to the next line. e.g. Colour in narrow text boxes.  
* I much prefer the previous "Welcome to your blank canvas" and agree with your suggestions Miguel,. We can keep iterating in the near future.   I'm not sure exactly what is wrong "Welcome to your blank canvas."  It has a lovely ring. 
* There was much discussion on the message in the blank canvas area if user goes to C&F, we don't want to block selection, because if user goes back to patterns they can see the C&F in the pattern list, which is kinda cool (I think).
   * @lucasmendes-design We previously discussed making the "Welcome message" or anything in the empty canvas into a pattern of some sort, so it would change color/font even if the user didn't pick their own pattern.  But either Matias or Pablo didn't like the idea, so I abandoned the idea.....sadly, as I thought it would be a way to do fun messaging in the empty canvas.
* I've moved the ideas we should take into next copy round in this issue: [Assembler Copy Updates June or July 2023#76614](https://github.com/Automattic/wp-calypso/issues/76614)

arthur791004:
Updated!

> Undo the changes of the copy in the Blank Canvas. Let's go back to what it is in production.

image

> Update the GS modal title to match Figma:

image

> Delete the sentence "Add some elements form the sidebar." Should look like this:

image

> Remove word "usually" since the footer always appears at the bottom (as far as I know?!?! 😄 ). Should be like this:
image

autumnfjeld:
Fantastico 🚢

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -70,31 +68,35 @@ const PatternLargePreview = ( {
...
-			? translate(
-					'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}',
-					options
-			  )
-			: translate(
-					'You can view your color and font selections after you select a pattern, get started by {{link}}adding a header pattern{{/link}}',
-					options
-			  );
+		return translate(
+			'You can view your color and font selections after you select a pattern. Get started by {{link}}adding a header pattern{{/link}}',
```
miksansegundo: Let's add a period at the end of this sentence: `Get started by {{link}}adding a header pattern{{/link}}.`

arthur791004: Oops, nice catch!

lucasmendes-design:
> We previously discussed making the "Welcome message" or anything in the empty canvas into a pattern of some sort, so it would change color/font even if the user didn't pick their own pattern. But either Matias or Pablo didn't like the idea, so I abandoned it.

I liked this idea, @autumnfjeld! Not sure but I believe all sites have a header, this is the default pattern that we could add when users land on the SA page. Without this, I think [one of the solutions above](https://github.com/Automattic/wp-calypso/pull/76528#issuecomment-1533111051) could work to communicate this.

autumnfjeld:
> > We previously discussed making the "Welcome message" or anything in the empty canvas into a pattern of some sort, so it would change color/font even if the user didn't pick their own pattern. But either Matias or Pablo didn't like the idea, so I abandoned it.
> 
> I liked this idea, @autumnfjeld! Not sure but I believe all sites have a header, this is the default pattern that we could add when users land on the SA page. Without this, I think [one of the solutions above](https://github.com/Automattic/wp-calypso/pull/76528#issuecomment-1533111051) could work to communicate this.


I think it we created a "landing pattern" it would need to be subtle, probably something the user would not actually use.     We might call it our "Welcome pattern".  As soon as the user picks their own pattern, the welcome pattern is replaced by a "real" pattern.     The  Welcome Pattern would have similar intro copy that exists now in production and should be designed to reflect font & color updates if the user "skips ahead" to C&F before they select a "real" pattern.    But again, this idea wasn't well received , so I don't know about pursuing it.  But I do really like the idea. ;) 

For now this messaging does the job:

image

lucasmendes-design:
Just a tiny improvement. I think this CTA is clearer than the blue one. 

image

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7966546

Thank you @arthur791004 for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7966546).

PR #76533, opened 2023-05-03
GitHub Integration: Only show in development and wpcalypso

danielbachhuber:
## Proposed Changes

Adds a `quake/github-integration-i1` feature flag so the GitHub integration only appears in development and wpcalypso environments.

It's confusing to have it appear in staging because it makes folks think the feature is available.

## Testing Instructions

1. Verify the "Deploy from GitHub" card still appears in your local environment.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -152,7 +152,7 @@
 		"publicize-preview": true,
 		"purchases/new-payment-methods": true,
 		"push-notifications": true,
-		"quake/github-integration-i1": true,
+		"github-integration-i1": true,
 		"reader": true,
 		"reader/comment-polling": false,
 		"reader/full-errors": true,
--- a/config/development.json
+++ b/config/development.json
@@ -152,6 +152,7 @@
 		"publicize-preview": true,
 		"purchases/new-payment-methods": true,
 		"push-notifications": true,
+		"quake/github-integration-i1": true,
 		"reader": true,
 		"reader/comment-polling": false,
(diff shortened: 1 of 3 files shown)


config/development.json
```diff
@@ -152,6 +152,7 @@
 		"publicize-preview": true,
 		"purchases/new-payment-methods": true,
 		"push-notifications": true,
+		"quake/github-integration-i1": true,
```
zaguiini: I'd name it just `github-integration`. Feels weird to see a feature scoped by iteration and team.

zaguiini (commented):
This only shows up for Automatticians, though. Is it a big deal?

danielbachhuber:
> This only shows up for Automatticians, though. Is it a big deal?

I've run into multiple instances where Automatticians thought the feature was live, presumably because they saw it in the UI and there's no indication it isn't a released feature.

config/development.json
danielbachhuber: Sure, I changed it with d85f3d9284

zaguiini:
We'll need to revisit this once GH is good to go. It should work with staging sites, too.

p-jackson:
> I've run into multiple instances where Automatticians thought the feature was live, presumably because they saw it in the UI and there's no indication it isn't a released feature.

I've edited this post so a12s looking for the feature can find it: p7DVsv-gCH-p2

PR #76537, opened 2023-05-03
Iterated on visual changes for blog-onboarding

paulopmt1:
Related to 2202-gh-Automattic/dotcom-forge

## Proposed Changes

* Removed Upgrade Plan label from writing flow:

image

* Removed unnecessary border on filter button:

image

## Testing Instructions

* Ensure the filter button border works as expected for the choose-domain option in the writing flow:

image

* Ensure the filter button border works as expected for domain upsell (from the left panel):

image

* Open the domain upsell without the extra parameter (domainAndPlanPackage=true) and make sure the border still appears in the filter button:

image

### DIFF
--- a/client/components/domains/register-domain-step/style.scss
+++ b/client/components/domains/register-domain-step/style.scss
@@ -378,3 +378,11 @@ body.is-section-signup.is-white-signup {
 		}
 	}
 }
+
+.search-component .search-filters__dropdown-filters button {
+	background-color: transparent;
+	border-color: transparent;
+	border-radius: 0;
+	border-right: none;
+	border-left: none;
+}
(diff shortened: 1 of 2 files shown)


client/components/domains/register-domain-step/style.scss
```diff
@@ -378,3 +378,11 @@ body.is-section-signup.is-white-signup {
 		}
 	}
 }
+
+.search-component .search-filters__dropdown-filters button {
```
DustyReagan, line 382:
```suggestion
.start-writing .search-component .search-filters__dropdown-filters button {
```
Should we be more specific with this rule, just in case?

DustyReagan (approved):
Looks great! Left a comment up to you. ☺️

client/components/domains/register-domain-step/style.scss
paulopmt1: I intentionally made it broad, from the domain upsell designs (pebzTe-Fx-p2#comment-1131), we should also have that border hidden

PR #75834, opened 2023-04-17
update/pricing page empty cart detect jetpack site

TheSteveK:
Related to # https://github.com/orgs/Automattic/projects/587/views/1?pane=issue&itemId=22640934

## Proposed Changes

* When a jetpack user goes directly to the checkout page (eg: https://wordpress.com/checkout/mygroovysite.com) the go back button redirects to the WordPress.com cart, but they should really be pointed to the Jetpack Pricing Page
* This PR updates the logic the use valid checkout back URL to check if the site in the URL is a jetpack site.  If so, it returns the go back URL as cloud.jepack.com/pricing/site 

## Testing Instructions

* Use the calpso.live site (or run this PR locally) add /checkout/[site] to the URL, where site is a valid jetpack site
* Note that you must not have anything in your cart for that site.
* You'll see the empty cart page, click the go back button
* Confirm you're redirected to the Jetpack pricing page

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76540, opened 2023-05-03
Do not treat Atomic sites as Jetpack in useValidCheckoutBackUrl

sirbrillig:
## Proposed Changes

`useValidCheckoutBackUrl()` is a function used to override the logic in the `leaveCheckout()` function when you click the "Go Back" button in checkout's empty cart view. It is also used for the "X" close button in the masterbar in all of checkout. Its name suggests that it is used for all checkout close URLs but in fact it only is used to override the normal behavior for Jetpack sites (and more recently, Akismet).

It looks like during its implementation, Atomic sites were not ignored in its logic. Therefore, it treats Atomic sites as Jetpack and redirects them to the Jetpack pricing page. This bug appears to have been introduced in https://github.com/Automattic/wp-calypso/pull/75834

In this diff we ignore Atomic sites for that logic, treating them as simple sites for the purpose of the "Go Back" and "X" buttons.

Screenshot 2023-05-03 at 10 03 35 AM

## Testing Instructions

- Visit `/checkout` for an Atomic site with an empty cart.
- Verify that when you click the "Go Back" button or the "X" button you are redirected to the `/plans` page.


gavande1 (approved):
LGTM 🚀

WBerredo (approved):
This fixed the issue for atomic sites, LGTM!

Tested with free, premium and Business plans (atomic and not)

Issue #76451, opened 2023-05-01
Site Assembler - Update categories in assembler and editor


miksansegundo:
### Keep most useful/cool/popular at the top and the rest ordered alphabetically

1. **All**
2. **Basic**
3. **Blog**
4. **Call to Action**
5. **Image Gallery**

The rest will be ordered alphabetically.

### Categories that will be hidden/removed

We’ll hide them in simple sites but could still appear in Atomic sites because core patterns and plugins can still use them.

- **Posts**
We’ll use “Blog” instead

- **Sidebar**
We’ll create “Columns” instead

- **Wireframe**
We’ll create “Basic” instead

- **Gallery/Image/Photography**
We’ll create “Image Gallery” instead

- **CV/Portfolio**
We’ll remove it and keep “Portfolio”

- **Editorial**
We’ll remove it and use “Text” and “Image Gallery” instead

### Categories that will be created/renamed

- **View All**
Will contain all patterns in the same list

- **Basic**
Will contain “Wireframe” patterns

- **Image Gallery**
Will contain “Gallery/Image/Photography” patterns

- **Text**
Will contain patterns with mostly text.

### Reassign patterns to most relevant categories 

Reorganize existing patterns to make sure they’re in the best and most relevant categories.

alaczek:
> Reorganize existing patterns to make sure they’re in the best and most relevant categories.
To be clear - this means going to the source site and re-categorizing them? Who is going to do this? :)

miksansegundo:
> Reorganize existing patterns to make sure they’re in the best and most relevant categories.
To be clear - this means going to the source site and re-categorizing them? Who is going to do this? :)

@alaczek I will do it and ask you for reviewing it when ready and help if needed.

miksansegundo:
These tasks are done with some minor adjustments on the category names for simplicity and have more alignment with other labels used in WordPress.

 - `All` rather than `View All`
 - `Blog Posts` rather than `Blog`
 - `Gallery` rather than `Image Gallery`

Here is the result:

<img width="338" alt="Screenshot 2566-05-16 at 19 19 10" src="https://github.com/Automattic/wp-calypso/assets/1881481/13e5f5c4-ab4f-47f0-bb17-6f26269c2693">



I'll follow up with a PR for Gutenberg to give us the ability to customize the categories order.

Issue #76443, opened 2023-05-01
Site Assembler - Show the category title and description in the patterns panel


miksansegundo:
## What

- Create a description for all categories so we can standardize the UX in the editor and the assembler.
- Show the category description on the top of the pattern list as in the Editor. 

## Why

The category description can help users to understand what each category means.

## How

Create a description for all categories. See pekYwv-1kJ-p2#comment-948

Show the title and description in the assembler as in the editor.

![Image](https://user-images.githubusercontent.com/1881481/235411018-526ea537-0872-4500-b7bb-2c74a6cfcf9b.png)

autumnfjeld:
We have copy updates from Francisco, so we can start dev work on this.   p1683570747538799/1683552689.470639-slack-CRWCHQGUB

miksansegundo:
Closed by https://github.com/Automattic/wp-calypso/pull/77087

PR #76575, opened 2023-05-04
Category updates - Move core query patterns from the category Posts to Blog in the editor

miksansegundo:
Related to #76451

## Proposed Changes

* Recategorize core query patterns from `Posts` to `Blog` in the editor (from the ETK plugin)

|BEFORE|AFTER|
|---|---|
|Screenshot 2566-05-09 at 12 44 25|Screenshot 2566-05-09 at 12 40 28|

Note: these core query patterns are not listed on the Assembler because we fetch patterns from the Dotcom pattern library only. We should copy these patterns in the Dotcom pattern library and then avoid the duplication on the editor.

### Install the ETK plugin from this branch

**Test on Simple sites**

* Sandbox your site
* Run `install-plugin.sh etk chore/recategorize-core-query-patterns`
* Don't forget to revert the ETK after testing with `install-plugin.sh etk --revert`

**Test on Atomic sites**

* Download the ETK plugin zip from Teamcity
* Upload it to an Atomic site and activate it

### Verify the changes
On Simple sites or Atomic sites when the ETK plugin is activated:

- Verify that you don't see the `Posts` category on the editor. 
- Verify that you see the seven core query patterns in the `Blog` category.

## Pre-merge Checklist

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
@@ -48,12 +48,24 @@ class Block_Patterns_From_API {
 
 		$this->utils = empty( $utils ) ? new \A8C\FSE\Block_Patterns_Utils() : $utils;
 
+		$blog_category = array(
+			'blog' => __( 'Blog', 'full-site-editing' ),
+		);
+
 		// Add categories to this array using the core pattern name as the key for core patterns we wish to "recategorize".
 		$this->core_to_wpcom_categories_dictionary = array(
-			'core/quote' => array(
+			'core/quote'                                => array(
 				'quotes' => __( 'Quotes', 'full-site-editing' ),
 				'text'   => __( 'Text', 'full-site-editing' ),
 			),
+			// Move core query patterns from 'Posts' to 'Blog'.
+			'core/query-standard-posts'                 => $blog_category,
+			'core/query-medium-posts'                   => $blog_category,
+			'core/query-small-posts'                    => $blog_category,
+			'core/query-grid-posts'                     => $blog_category,
+			'core/query-large-title-posts'              => $blog_category,
+			'core/query-offset-posts'                   => $blog_category,
+			'core/social-links-shared-background-color' => $blog_category,
 		);
 	}
 
--- a/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
@@ -48,24 +48,12 @@ class Block_Patterns_From_API {
 
 		$this->utils = empty( $utils ) ? new \A8C\FSE\Block_Patterns_Utils() : $utils;
 
-		$blog_category = array(
-			'blog' => __( 'Blog', 'full-site-editing' ),
-		);
-
 		// Add categories to this array using the core pattern name as the key for core patterns we wish to "recategorize".
 		$this->core_to_wpcom_categories_dictionary = array(
-			'core/quote'                                => array(
+			'core/quote' => array(
 				'quotes' => __( 'Quotes', 'full-site-editing' ),
 				'text'   => __( 'Text', 'full-site-editing' ),
 			),
-			// Move core query patterns from 'Posts' to 'Blog'.
-			'core/query-standard-posts'                 => $blog_category,
-			'core/query-medium-posts'                   => $blog_category,
-			'core/query-small-posts'                    => $blog_category,
-			'core/query-grid-posts'                     => $blog_category,
-			'core/query-large-title-posts'              => $blog_category,
-			'core/query-offset-posts'                   => $blog_category,
-			'core/social-links-shared-background-color' => $blog_category,
 		);
 	}
 
@@ -151,6 +139,8 @@ class Block_Patterns_From_API {
 		$this->update_core_patterns_with_wpcom_categories();
 
 		$this->update_pattern_post_types();
+
+		$this->update_query_patterns_with_blog_category();
 
 		return $results;
 	}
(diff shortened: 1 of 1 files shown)


apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
```diff
@@ -48,12 +48,24 @@ public function __construct( Block_Patterns_Utils $utils = null ) {
...
 		$this->utils = empty( $utils ) ? new \A8C\FSE\Block_Patterns_Utils() : $utils;
 
+		$blog_category = array(
+			'blog' => __( 'Blog', 'full-site-editing' ),
+		);
+
 		// Add categories to this array using the core pattern name as the key for core patterns we wish to "recategorize".
 		$this->core_to_wpcom_categories_dictionary = array(
-			'core/quote' => array(
+			'core/quote'                                => array(
```
miksansegundo: This formatting is applied automatically by PHPCBF.

taipeicoder:
I'm still seeing the Posts category, with only one pattern in it:

Screenshot 2023-05-05 at 1 59 43 PM

miksansegundo:
Thanks for testing @taipeicoder. I can't reproduce it. Can you share more details to help me reproduce it? What kind of site and what theme are you using?

miksansegundo:
I found the reason. Patterns from themes like TT3 can use the `posts` or `query` categories. I'll find a solution for this case. Here is the Meta Post pattern that you see: fbhepr%2Skers%2Sjcpbz%2Qcho%2Qgurzrf%2Sgjraglgjraglguerr%2Scnggreaf%2Scbfg%2Qzrgn.cuc%3Se%3Q89r4s9p5%235-og

miksansegundo:
I found a solution that works for all patterns.

arthur791004:
Would it be better to keep the original categories or simply add the prefix to those categories? I guess people might also search/look for the `posts` patterns 🤔

miksansegundo:
@arthur791004 can you add some more details to help me understand your comment?
 
> Would it be better to keep the original categories or simply add the prefix to those categories? I guess people might also search/look for the posts patterns 🤔

> Would it be better to keep the original categories 

What original categories do you mean? 

> simply add the prefix to those categories

What prefix do you mean?

miksansegundo:
> I guess people might also search/look for the posts patterns 🤔

Are you suggesting keeping two categories, `Posts` and `Blog`, with the same patterns?

The idea is that on Dotcom, we won't use the `Posts` category because themes are categorized as `Blog` and we want to align everywhere.

@taipeicoder, maybe you can help me understand why themes are not categorized using the category `Posts` but `Blog` instead. Based on that decision, we decided to create this issue.

We plan to add a search bar in the assembler soon that will search for keyword matches in the pattern names and categories. If we want `Blog` patterns to be found when searching for `post` or `posts`, we can add these keywords in the pattern name.

FYI. The Dotorg pattern directory uses the label `Posts` but internally the slug is `query`. See it in https://wordpress.org/patterns/categories/query/

Screenshot 2566-05-08 at 19 06 30

arthur791004:
> What original categories do you mean?

Yes, I mean keep two categories, Posts and Blog, with the same patterns.

> What prefix do you mean?

For example, we can just rename them below
* `Posts` to `Blog Posts`
* `Query` to `Blog Query`

> If we want Blog patterns to be found when searching for post or posts, we can add these keywords in the pattern name.

Not just for searching. People might want to show the `Posts` on their home page. So, when people browse the list, they try to look for the `Posts` but they cannot find that category since it's renamed to `Blog`. Does it make sense?

apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
```diff
@@ -278,4 +280,24 @@ private function update_pattern_post_types() {
 			}
 		}
 	}
+
+	/**
+	 * Ensure that all query or posts patterns use the category blog instead.
+	 */
+	private function update_query_patterns_with_blog_category() {
```
okmttdhr, line 291: We might be able to write a simple test here.
https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/test/class-block-patterns-from-api-test.php

miksansegundo:
@arthur791004 Thanks for the details. I think the name `Blog Posts` will work to solve that confusion. I'll update the PR.

miksansegundo:
I updated the PR to use `Blog Posts` and add a description, we will review the description later as part of #76443.

The position of `Blog Posts` is not at the top of the list because the editor expects the slug `posts` rather than `blog`. We'll address the order later in a follow-up PR.

arthur791004 (approved):
LGTM, thanks 👍

okmttdhr (approved):
LGTM with a non-blocking comment 🙌

apps/editing-toolkit/editing-toolkit-plugin/block-patterns/class-block-patterns-from-api.php
miksansegundo: Thanks for thinking about testing 🙌 
What kind of test would be useful for the changes in this PR?

okmttdhr: badge

What I was thinking was a test ensuring the structure changes of query patterns from `Posts` to `Blog`. We might be able to [mock the pattern object](https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/editing-toolkit-plugin/block-patterns/test/class-block-patterns-from-api-test.php#L34-L48) with query patterns `Posts` and ensure the changes to `Blog`.
(we can test it through `register_patterns` since `update_query_patterns_with_blog_category` is a private method)

It's gonna be a small test, so I don't mean we need to be hardcore for this.
But it might help prevent future changes break it 💫 , if someone else modifies the code in a way that patterns are incorrectly categorized, the test will fail.

miksansegundo: Got it. I think for the categories case, we should be more flexible than in other cases because pattern categorization can break easily as Designers can add patterns and categories easily from the admin.

Let's reconsider adding tests later when we know better which areas should not change or things will break and tests will be more helpful.

Issue #76495, opened 2023-05-02
Task: Happy blocks WordPress header logo inconsistency


renancarvalho:
### Details

The header has some inconsistency when resizing the browser, See video attached.
The spacing to the left changes as the user resizes.

https://user-images.githubusercontent.com/2653810/235684603-3590e4dd-4883-48c0-9fcc-76afbc65611c.mp4

escapemanuele:
This issue has been fixed!

PR #76591, opened 2023-05-04
Happy Blocks: Remove logo padding-left

renancarvalho:
## Proposed Changes

The WordPress header logo is moving as the browser resizes on different viewports. This PR fixes it by removing the padding-left being set on the logo in different viewports and leaving only left and right 24px in the `nav` wrapper as it should be according to the [designs.](IceKRdz5qlPYwuqzu6pBgF-fi-2272_29779)

## Testing Instructions

- Check the video in the issue #76495 to visualize the problem.
- Pull and run this branch
- Navigate to `cd apps/happy-blocks` and run `yarn dev --sync`
- Sandbox `learncft.wordpress.com` and `wordpress.com/support` and visit the site
- Check if the designs are matching and the WP logo is not moving as you resize the browser window.
- Check [learn](https://learncft.wordpress.com/) and mainly [support](https://wordpress.com/support/) since it is live
- Check mobile and desktop.

Fixes to #76495


escapemanuele (approved):
Looks like it is working great! 😄

PR #76593, opened 2023-05-04
Fix the infinite loading state on Jetpack Backups list

macbre:
After the migration to `@tanstack/react-query` (see #76077) we were incorrectly still using the `isLoading` flag (which was never changing to false) instead of `isInitialLoading`.

Before this change: https://github.com/Automattic/wp-calypso/blob/b17472abcf76eb36ae7960887d65f2d5ddede251/client/my-sites/backup/status/hooks.js#L43

_Under the hood `useRewindableActivityLogQuery()` uses `useQuery` from `@tanstack/react-query`._

## Testing Instructions

1. Start the local Calypso env and see whether your site's backups list loads fine.
2. Now, using [the "Switch to User" tool switch to the `userref` selecting `calypso.local` from the dropdown](https://mc.a8c.com/tools/support-user/). Navigate to Jetpack / Backups where you should see:

Screenshot 2023-05-04 at 15 20 18

### DIFF
--- a/client/my-sites/backup/status/hooks.js
+++ b/client/my-sites/backup/status/hooks.js
@@ -40,13 +40,13 @@ const useBackupDeltas = ( siteId, { before, after, number = 1000 } = {}, enabled
 
 	const isValidRequest = filter.before && filter.after;
 
-	const { data, isLoading } = useRewindableActivityLogQuery( siteId, filter, {
+	const { data, isInitialLoading } = useRewindableActivityLogQuery( siteId, filter, {
 		enabled: isValidRequest && enabled,
 		refetchOnWindowFocus: false,
 	} );
 
 	return {
-		isLoading,
+		isLoading: isInitialLoading,
 		deltas: getDeltaActivitiesByType( data ?? [] ),
 	};
 };
--- a/client/my-sites/backup/status/hooks.js
+++ b/client/my-sites/backup/status/hooks.js
@@ -46,7 +46,7 @@ const useBackupDeltas = ( siteId, { before, after, number = 1000 } = {}, enabled
 	} );
 
 	return {
-		isLoading: isInitialLoading,
+		isInitialLoading,
 		deltas: getDeltaActivitiesByType( data ?? [] ),
 	};
 };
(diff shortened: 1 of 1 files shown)


client/my-sites/backup/status/hooks.js
```diff
@@ -40,13 +40,13 @@ const useBackupDeltas = ( siteId, { before, after, number = 1000 } = {}, enabled
...
 
-	const { data, isLoading } = useRewindableActivityLogQuery( siteId, filter, {
+	const { data, isInitialLoading } = useRewindableActivityLogQuery( siteId, filter, {
 		enabled: isValidRequest && enabled,
 		refetchOnWindowFocus: false,
 	} );
 
 	return {
-		isLoading,
+		isLoading: isInitialLoading,
```
tyxla: I think it can be helpful to rename the property returned by the hook as well, so it reflects the actual behavior and matches the underlying React Query API that we utilize. Otherwise, it might be confusing because the hook consumer could expect it to use `isLoading` under the hood and that wouldn't be true.

flootr: I second what Marin said, otherwise I think this looks good.

macbre: Makes sense - improved in a1d3d16cb8.

Thanks!

jwebbdev (approved):
Looks good, tested with another user as well and it properly loads up the `Backup Scheduled` card.  Looks like other comments were addressed so I'll 👍 - just might have to rerun some tests it looks like.

tyxla (approved):
LGTM from my perspective 👍 🚀

client/my-sites/backup/status/hooks.js
```diff
@@ -142,7 +142,9 @@ export const useDailyBackupStatus = ( siteId, selectedDate ) => {
 
 	return {
 		isLoading:
-			lastBackupBeforeDate.isLoading || lastAttemptOnDate.isLoading || backupDeltas.isLoading,
+			lastBackupBeforeDate.isLoading ||
+			lastAttemptOnDate.isLoading ||
```
rustamshaikh, line 146:
```suggestion
			lastBackupBeforeDate.isInitialLoading ||
			lastAttemptOnDate.isInitialLoading ||
```

I think these are also affected and need change since in the end, they are calling `useRewindableActivityLogQuery` in `useLatestBackupAttempt` hook. In fact, there are some places in the codebase that uses `useRewindableActivityLogQuery` (and `isLoading`) that need fixing :(

macbre: I'd rather keep this fix as small as possible. We can tackle these ones in a separate PR maybe and perform more broader testing then. Sounds good?

rustamshaikh: Yep

rustamshaikh (approved):
LGTM, the site's backup page loads fine.

tyxla:
Thanks again @macbre! Please ping us if you encounter any other issues that might have been a cause of the React Query v4 migration.

PR #76467, opened 2023-05-01
Launchpad: Update getEnhancedTasks

sharpirate:
Fixes #76478 

### Time Estimate
Review: Short
Test: Mid

### Summary
Update getEnhancedTasks to read task statuses from the endpoint

### Testing
1. Because the `launchpad/checklist` API isn't working at the moment, sandbox a site and apply this patch to the sandbox  https://github.com/Automattic/jetpack/pull/30397
1. Check out this branch and run `yarn` and `yarn start` if needed.
2. Start any new launchpad enabled site
   - Newsletter http://calypso.localhost:3000/setup/newsletter/intro
   - Link-in-Bio http://calypso.localhost:3000/setup/link-in-bio/intro
   - VideoPress http://calypso.localhost:3000/setup/videopress/intro
   - Build http://calypso.localhost:3000/start ( select `Other` as the site goal )
   - Write http://calypso.localhost:3000/start ( select `Write & Publish` as the site goal )
3. Navigate to the Launchpad by completing onboarding
4. Smoke test all flows and make sure everything works as expected. Pay attention to the domain upsell task specifically.

PR #76602, opened 2023-05-04
Launchpad: Add actions to paid newsletters tasks

lezama:
Add Actions to these tasks:
Screenshot 2023-05-04 at 17 49 15

Based on top of https://github.com/Automattic/wp-calypso/pull/76467

# How to test
* Follow the instructions in https://github.com/Automattic/wp-calypso/pull/76467
* Visit the newsletter flow: http://calypso.localhost:3000/setup/newsletter/
* Select paid newsletters
* Check the Paid newsletter tasks in the launchpad step

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
@@ -391,7 +391,7 @@ export function getEnhancedTasks(
 					taskData = {
 						actionDispatch: () => {
 							recordTaskClickTracksEvent( flow, task.completed, task.id );
-							window.open( `/earn/payments/${ siteSlug }#launchpad`, '_blank' );
+							window.location.assign( `/earn/payments/${ siteSlug }#launchpad` );
 						},
 					};
 					break;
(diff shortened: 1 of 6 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -387,6 +387,25 @@ export function getEnhancedTasks(
 						completed: isEmailVerified,
 					};
 					break;
+				case 'set_up_payments':
+					taskData = {
+						actionDispatch: () => {
+							recordTaskClickTracksEvent( flow, task.completed, task.id );
+							window.open( `/earn/payments/${ siteSlug }#launchpad`, '_blank' );
```
simison: Why `_blank`and not in the same tab?

lezama: there's no clear way to get back to the Launchpad after going to the Earn page. Someone suggested new tab 🤔 , I think it makes sense for now.

Maybe a better pattern could be to make these links instead?

simison: That's valid concern, but I don't think this solves it. :-) Folks get lost with multiple tabs all of a sudden open, browser back button stops working, and it's tricky especially on mobile.

I think we need to look at the "My home Launchpad" work and think if "My home" navigation item in the sidebar could be changed somehow to solve this, because I agree it's a real problem.

lezama: @simison so no new tab for now?

simison: Let's find a better way to solve the problem and remove `_blank`

I wrote about this in p2 (p9Jlb4-7c4-p2/#comment-7904) but maybe we can think of something intermediate.

lezama:
Rebased and merged here https://github.com/Automattic/wp-calypso/pull/76793/files

PR #75044, opened 2023-03-28
VAT Info: Use string interpolation for tax names

sirbrillig:
## Proposed Changes

This is a follow-up to https://github.com/Automattic/wp-calypso/pull/74538 which localized the "VAT" term in checkout. Because the name of VAT changes based on country, it would be better to use a localized term. In https://github.com/Automattic/wp-calypso/pull/73686 we modified the labels for the VAT form use use the more generic term "Business Tax", but this is still a little unclear.

In this PR we switch each of these labels to replace "Business Tax" with the selected country's term for VAT (eg: "GST", "CT", etc.) on the user-level VAT form.

**NOTE:** This can only localize once a country is selected... which means that the button to reach the form will always read "Add VAT details" before the user reaches the form. To resolve that case, we use the [useGeoLocationQuery() hook](https://github.com/Automattic/wp-calypso/blob/136443bd1dfb5792563475a4a4981e0ffded1750/client/data/geo/use-geolocation-query.js#L3) to provide a fallback country.

This is part of https://github.com/Automattic/payments-shilling/issues/1475 and https://github.com/Automattic/payments-shilling/issues/1474

## Screenshots

Screenshot 2023-03-28 at 1 10 40 PM

Screenshot 2023-03-28 at 2 12 32 PM

Screenshot 2023-03-28 at 1 11 02 PM

Screenshot 2023-03-28 at 1 11 09 PM

Screenshot 2023-03-28 at 2 12 24 PM

## Testing Instructions

- Visit `/me/purchases/vat-details`.
- Select a country that uses the term "VAT" like the United Kingdom.
- Verify that all the form labels use the word "VAT".
- Select a country that uses a different term like Australia, which uses "GST".
- Verify that all the form labels use the word "GST" (except for the title, which won't change until the details are saved).
- Visit `/me/purchases/billing` and set your geolocation (eg: using a VPN) to somewhere that is in a country that does not use "VAT" (eg: Australia).
- Verify that the button at the bottom reads "Add GST details".

PR #76609, opened 2023-05-04
Payments: Replace fallback with "tax (VAT/GST/CT)" for "Add VAT details"

### LINKED ISSUE #75445: Matching unique Country Tax Name to profile address or currency
### What

Can the location or currency that a user is using in their profile be linked to the name of the tax id in the region, country or countries which use that currency?

### Why

It’s a small thing, but it could be a good win for making a user feel personally spoken to.

6175013-zen is an example of a user saying that they were confused by the VAT/GST reference. They are in Australia and it is called the ABN for Australian customers - they higlighted this when asking what the GST tax was.

### How

This can be automatically matched to location/currency and updated to the users Billing History, or wherever the country tax name is shown. 

For instance, the ID number that we'll ask the user for in their VAT details may not always be called **VAT** ID, so for example, when they change their country ID, can the related VAT name change?

See annotated screenshot below:
Markup on 2023-04-07 at 18:04:54

I considered that this might be too small or unnecessary to be worked on, but in case it is a small change, I decided to surface it.

sirbrillig:
## Proposed Changes

In https://github.com/Automattic/wp-calypso/pull/75044 we modified each mention of "VAT" in calypso checkout and payment management to use the local term for taxes in the user's selected (tax) country (eg: "CT" for Japan). However, there are cases where we do not know what the country should be. In that case, https://github.com/Automattic/wp-calypso/pull/75044 relies on the user's geolocation to determine which string to display, but this is imperfect since, for example, it's possible for a user that pays taxes in Japan to be using an IP address in the US.

In this PR we change the fallback text to be "tax (VAT/GST/CT)". For example, when you see the button to reach the VAT info page or view the title of the VAT info page when no country has been set, these will now read "Add tax (VAT/GST/CT) details" instead of "Add VAT details" (if you are in the US).

Since this is the fallback text, **this will have no effect if the user has a selected tax location or if they are geolocated in a country that has a tax acronym**.

Before (geolocated in the US since there is no acronym for taxes there):

Screenshot 2023-05-04 at 5 17 29 PM

Screenshot 2023-05-02 at 1 36 34 PM

After (geolocated in the US):

Screenshot 2023-05-04 at 5 09 14 PM

Screenshot 2023-05-04 at 5 09 27 PM

Fixes https://github.com/Automattic/wp-calypso/issues/75445

## Testing Instructions

- Use an account set to English and geolocated in a country without a tax acronym (eg: the US). Also make sure the user does not have VAT details saved.
- Visit `/me/purchases/billing` and verify the button at the bottom of the page reads "Add tax (VAT/GST/CT) details".
- Click the button.
- Verify that the title of the page reads "Add tax (VAT/GST/CT) details".

### DIFF
--- a/client/me/purchases/billing-history/main.tsx
+++ b/client/me/purchases/billing-history/main.tsx
@@ -43,7 +43,7 @@ function BillingHistory() {
 
 	const genericTaxName =
 		/* translators: This is a generic name for taxes to use when we do not know the user's country. */
-		i18n.hasTranslation( 'tax (VAT/GST/CT)' );
+		translate( 'tax (VAT/GST/CT)' );
 	const fallbackTaxName =
 		getLocaleSlug()?.startsWith( 'en' ) || i18n.hasTranslation( 'tax (VAT/GST/CT)' )
 			? genericTaxName
--- a/client/me/purchases/billing-history/main.tsx
+++ b/client/me/purchases/billing-history/main.tsx
@@ -1,6 +1,6 @@
 import config from '@automattic/calypso-config';
 import { CompactCard, Card } from '@automattic/components';
-import { useTranslate } from 'i18n-calypso';
+import i18n, { getLocaleSlug, useTranslate } from 'i18n-calypso';
 import DocumentHead from 'calypso/components/data/document-head';
 import QueryBillingTransactions from 'calypso/components/data/query-billing-transactions';
 import FormattedHeader from 'calypso/components/formatted-header';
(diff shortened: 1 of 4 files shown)


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7976612

JessBoctor:
I updated the [WordPress.com support doc](https://wordpress.com/support/vat-gst-other-taxes/) to match this change: 

**Before**
Screenshot 2023-05-09 at 09 36 08

**After**
Screenshot 2023-05-09 at 09 55 15

Notified documentation team in p1683644344486989-slack-docs-docs.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7976612).

PR #76610, opened 2023-05-04
Hide plan upsell when having monthly product in cart

spraveenitpro:
Related to [#](https://github.com/Automattic/payments-shilling/issues/1552)

## Proposed Changes

* Remove the upsell that appears on the right sidebar for monthly carts

## Testing Instructions

1. Add a monthly plan to the cart from the plans page http://calypso.localhost:3000/plans/monthly/{sitename}
2. On the checkout page we should see the upsell like this 

3. Apply the patch
4. Refresh the cart and now the upsell component should not be shown any more like this: 
5. For yearly plans, nothing should change.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -311,6 +311,7 @@ export default function WPCheckout( {
 	}
 
 	const isDIFMInCart = hasDIFMProduct( responseCart );
+	const showUpsellMonthly = responseCart?.products?.some( isMonthlyProduct );
 
 	return (
 		<CheckoutStepGroup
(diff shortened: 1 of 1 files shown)


sirbrillig:
> Show plan upsell when having monthly product in cart #76610

> Remove the upsell that appears on the right sidebar for monthly carts

I may be misunderstanding, but in the issue and the PR description, it seems that the goal is to hide the upsell when a monthly product is in the cart. But the code and the screenshots show the opposite: it shows the upsell _only_ when a monthly product is in the cart. Am I just confused?

spraveenitpro:
> I may be misunderstanding, but in the issue and the PR description, it seems that the goal is to hide the upsell when a monthly product is in the cart.

You are right! I got confused :) , I have now updated the testing instructions and the code 🙏🏽

sirbrillig:
Much more clear! Thanks! You should update the PR title as well.

client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
```diff
@@ -311,6 +311,7 @@ export default function WPCheckout( {
 	}
 
 	const isDIFMInCart = hasDIFMProduct( responseCart );
+	const showUpsellMonthly = responseCart?.products?.some( isMonthlyProduct );
```
sirbrillig: While it does the right thing I think this variable name might be a little misleading. Maybe something like `hasMonthlyProduct`? Or `isMonthlyProductInCart`?

sirbrillig (commented):
This looks good! One minor comment but otherwise I think this is good to go.

client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
spraveenitpro: I agree with this, `hasMonthlyProduct ` looks better.

sirbrillig (approved):
Looks good! Thanks!

PR #76615, opened 2023-05-05
refactor celebration modals to expose their logic

roo2:
Broken out from https://github.com/Automattic/wp-calypso/pull/76267.

In https://github.com/Automattic/wp-calypso/pull/76267, I'm adding a new sharing modal which will appear after publishing a post. However it is important that that modal does not overwrite these existing modals.

This PR prepares the way by exposing the logic of whether the videopress or seller modal should be shown into `use-*` hooks. In https://github.com/Automattic/wp-calypso/pull/76267, I can then add logic to only display the new modal if neither of these existing modals are going to be shown.

### Testing instructions 
Check out this branch, build the ETK plugin and upload it to your sandbox with:
```
cd apps/editing-toolkit
yarn dev --sync
```

Note: When I tested both modals saving in the Site Editor, I found that they were not shown. In both cases, the isEditorSaving flag was not updated on save. This appears to be an existing bug, and I'll leave this out of this PR (although, it's possible I'm not using the site editor in the way that is expected).

#### Video celebration modal:
* **Sandbox** a site with `site_intent` set to `videopress`
  * in `wpsh` run `update_blog_option( $site_id, 'site_intent', 'videopress' )`
  * in `wpsh` run `update_blog_option( $site_id, 'has_seen_video_celebration_modal', false )`
* Create a post and add a video block to it.
* Publish the post
* You should see the videopress celebration modal.

Screen Shot 2023-05-04 at 2 52 51 pm

Note: the [launchpad modal](https://github.com/Automattic/jetpack/blob/f60b8721717189e53129e856c75a9461a0d153fb/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/launchpad-save-modal.php) seems to already be conflicting with the videopress celebration modal to an extent, but they appear behind one another

#### Seller celebration modal:
* **Sandbox** a site with `site_intent` set to `sell`
  * in `wpsh` run `update_blog_option( $site_id, 'site_intent', 'sell' )`
  * in `wpsh` run `update_blog_option( $site_id, 'has_seen_video_celebration_modal', false )`
* In the post editor, add a "payments buttons" block,
* Save the page
* You should see the seller celebration modal
Screen Shot 2023-05-04 at 12 47 16 pm

#### First post published modal:
* **Sandbox** a site with `site_intent` set to `sell`
  * in `wpsh` run `update_blog_option( $site_id, 'site_intent', 'write' )`
  * in `wpsh` run `update_blog_option( $site_id, 'has_never_published_post', true )`
* In the post editor, publish a post.
* You should see the first post published modal

Screen Shot 2023-05-05 at 12 54 09 pm


mreishus:
I tried out the seller modal and it is working fine for me. I did notice that there are now 2 calls to the "has seen" hook:

https://github.com/Automattic/wp-calypso/blob/6a7c86dd4e6e64f9688c0c848ae9420ad6425418/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/seller-celebration-modal/index.jsx#L46

https://github.com/Automattic/wp-calypso/blob/6a7c86dd4e6e64f9688c0c848ae9420ad6425418/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/seller-celebration-modal/use-should-show-seller-celebration-modal.js#L14

I think they make two separate instances of the data and side effects, causing the fetch to happen twice on an editor pageload:



I'm not really sure of a fix. Maybe `SellerCelebrationModalInner` could also pull out "hasSeen" and pass it to "shouldShow" ?

```
const { updateHasSeenSCM, hasSeenSCM } = useHasSeenSellerCelebrationModal();
const shouldShowSellerCelebrationModal = useShouldShowSellerCelebrationModal( hasSeenSCM );
```
Just an idea though; haven't tried it and I'm not sure if it makes sense. I'm not the best with hooks.

roo2:
thanks @mreishus. hmm I was having a look at other approaches, I think the react way to do it is to create a context https://stackoverflow.com/questions/59570594/how-do-i-fetch-data-in-a-react-custom-hook-only-once

But that seems a bit complicated, I think your solution of passing the `hasSeenSCM` variable is a lot simpler 👍

roo2:
hmm actually on second thought, since I will need to call the functions again in https://github.com/Automattic/wp-calypso/pull/76267 anyway. I think I should look at using a context to share these variables and avoid duplicate calls.

mreishus:
One possible other idea is using react-query to fetch the data. It assigns a [query key](https://tanstack.com/query/v4/docs/react/guides/query-keys) to each query, so if there are multiple instances, only the first should fetch the data and the second should grab the cached data. 
I think we'd need to add a [QueryClientProvider](https://github.com/Automattic/wp-calypso/blob/e7903abc2f7c3a2dd3102885df6602b49bd7db36/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/help-center.js#L86) to [the root of the nux plugin](https://github.com/Automattic/wp-calypso/blob/e7903abc2f7c3a2dd3102885df6602b49bd7db36/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/block-editor-nux.js#L100), and it might end up making the nux plugin larger.

roo2:
ahh yep, ok I think that is how we would normally do it in calypso. But it looks like we haven't used that library yet in the ETK. I would be more keen to leave it as the PR currently is with the Providers because it doesn't need to pull in a new library, (and it's less work for me to change it now 😉 )

creativecoder:
Do any of these modals work on Atomic sites (and thus need to be tested)?

I started trying to test the VideoPress modal on my Atomic development site, but it doesn't look like it works. There are API requests going a `site-intent` endpoint that doesn't seem to exist on Atomic sites 😞

roo2:
yea no they don't. I see comments in the code about it here but I don't quite understand it: https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/class-wp-rest-wpcom-block-editor-first-post-published-modal-controller.php#L76

okmttdhr (approved):
LGTM for `First post published modal` 👍 

Having a test will give us more confidence that the functionality is not broken by the refactor.

roo2:
I think some tests would be good but it's hard to test end to end and I want to limit the scope of this and do just what's necessary for https://github.com/Automattic/wp-calypso/pull/76267, especially considering the feature freeze for the ETK

Addison-Stavlo (commented):
Sell and post publish modals appear to work as per instructions here.

I am having issues with the videopress modal however. 🤔

i created a new site in the videopress flow, went to the add video step, added a video, saved, and got the general launchpad modal instead of the videopress modal. I exited back to the launchpad. I then went back into the video step a second time, updated a paragraph block, and saved - now the videopress modal appeared.

roo2:
hmm I only tested by manually setting site_intent from wpsh, I'll need to find how to actually access the videopress flow to test it 😆.

roo2:
@Addison-Stavlo was it in the site editor at first? I found an existing bug where some of the modals don't seem to be fired through the site editor

roo2:
When I testsed, I got both modals show up one behind the other, which appears to be the existing behavior
Screen Shot 2023-05-15 at 9 58 09 am

Addison-Stavlo:
> @Addison-Stavlo was it in the site editor at first?

No, it was the post editor both times. 🤷‍♀️

PR #76627, opened 2023-05-05
Theme Showcase: Use mShots for first-party themes

taipeicoder:
## Proposed Changes

This PR proposes to update the Theme Showcase to use mShots for first-party themes. This allows the Theme Showcase to align with the Onboarding Design Picker in two ways:

1. Use mShots to display theme thumbnails for style variations. Currently, the Theme Showcase uses a pre-defined image.
2. Change the theme thumbnails when a style variation is selected. Currently, the Theme Showcase opens the Theme Detail page when a style variation is selected.

https://github.com/Automattic/wp-calypso/assets/797888/a882bc73-2c3f-46bd-8bce-5a6f9caf5f22

## Testing Instructions

* Head to the Theme Showcase.
* Find a theme with style variations, such as Storia.
* Ensure that the theme thumbnails change according to the selected style variation, as described above.
* Ensure that clicking the theme thumbnail lands users to the Theme Detail page, with the style variation preselected.
* Ensure that theme options, such as Live Demo, also have the selected style variation applied.
* Also test in the logged-out Theme Showcase.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/themes/themes-selection.jsx
+++ b/client/my-sites/themes/themes-selection.jsx
@@ -25,7 +25,7 @@ import {
 	isInstallingTheme,
 	prependThemeFilterKeys,
 } from 'calypso/state/themes/selectors';
-import { trackClick } from './helpers';
+import { addStyleVariation, trackClick } from './helpers';
 import SearchThemesTracks from './search-themes-tracks';
 import './themes-selection.scss';
 
@@ -136,18 +136,8 @@ class ThemesSelection extends Component {
 				'calypso_themeshowcase_theme_style_variation_more_click',
 				tracksProps
 			);
-		}
 
-		const url = this.props.getThemeDetailsUrl( themeId );
-		if ( url ) {
-			const [ urlBase, urlQuery ] = url.split( '?' );
-			const params = new URLSearchParams( urlQuery );
-			if ( variation ) {
-				params.set( 'style_variation', variation.slug );
-			}
-
-			const paramsString = params.toString().length ? `?${ params.toString() }` : '';
-			pageRouter( `${ urlBase }${ paramsString }` );
+			pageRouter( this.props.getThemeDetailsUrl( themeId ) );
 		}
 	};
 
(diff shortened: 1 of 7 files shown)


client/my-sites/themes/themes-selection.jsx
```diff
@@ -212,13 +202,21 @@ class ThemesSelection extends Component {
...
+					styleVariation
+				);
 				return action( t, context );
 			};
 		};
 
-		if ( options && options.preview ) {
-			options.preview.action = wrappedPreviewAction( options.preview.action );
+		if ( options ) {
+			options = addStyleVariation( options, styleVariation );
```
arthur791004: Since we wrapped all `getUrl` functions with the selected style variation, how about passing the style variation to the 3rd argument to the `getUrl` function to let those functions decide how to handle the style variation? Does it make sense?

For example, the checkout URL might be incorrect as it should append the style variation to the `redirect_url` instead of appending it to the end of the URL directly

arthur791004:
The height of mShot looks weird but it might be an existing issue and not a blocker for this PR

| Default | With style variation |
| - | - |
| image | image |

client/my-sites/themes/themes-selection.jsx
taipeicoder: Good catch! I've updated the code so that each option's `getUrl` is responsible for handling the style variation.

okmttdhr:
I'm commenting just to refer to https://github.com/Automattic/wp-calypso/pull/77052, as it's related to https://github.com/Automattic/wp-calypso/pull/76627#issuecomment-1545414055 (CMIIW).

okmttdhr (approved):
Sometimes it takes 5~10 sec to show the image for the first several times (maybe we have no cache yet), but it got faster after a while. (It seems weird the display is slow even when selecting the **default** style variation back from selecting another style variation.)
I assume it would not happen in production?

But the performance problems should be solved in another PR even if any.
So, LGTM 💫🙌

taipeicoder:
> Sometimes it takes 5~10 sec to show the image for the first several times (maybe we have no cache yet), but it got faster after a while. (It seems weird the display is slow even when selecting the **default** style variation back from selecting another style variation.) I assume it would not happen in production?
> 
> But the performance problems should be solved in another PR even if any. So, LGTM 💫🙌

Yes, the caching should improve once it's in production since we get more traffic there 🙂

PR #76655, opened 2023-05-08
Hosting Onboarding: Add i2 flag flag

p-jackson:
Fixes Automattic/dotcom-forge#2358

## Proposed Changes

* A small tidy of flags in 674a4fdd3b14ffbe28c96948f2dd9b65521ddeef
* Adds a new `hosting-onboarding-i2` flag

Any features of the Hosting Onboarding i2 project that need to be released simultaneously can use this flag. The flag is initially only enabled in development builds.

## Testing Instructions

Smoke test Calypso. There should be no observable changes.


p-jackson:
I'm merging this now so other project work can easily take advantage of these new flags.

klimeryk:
Thanks for the clean-up of the feature flags, @p-jackson 🙇 Just noticed there's one left that needs fixing: https://github.com/Automattic/wp-calypso/blob/e7903abc2f7c3a2dd3102885df6602b49bd7db36/config/stage.json#L171. Sorry to be a PITA about this, but once one flag goes out of order, it's gonna be a mess soon 😅 (I need to actually add a flag that alphabetically goes at the end, which is why I noticed these flags being out of order)

PR #76657, opened 2023-05-08
Hosting Onboarding: New site creation flow for users interested in "hosting"

p-jackson:
Fixes https://github.com/Automattic/dotcom-forge/issues/2362

## Proposed Changes

* Creates a new flow in the "stepper" framework for creating sites for users interested in "hosting" features
* This flow only shows plans which are compatible with WoA sites
* Ensure the plans grid is displayed using the 2023 style

Some implementation notes:
- The new flow uses it's own flow ID i.e. it uses `new-hosted-site` rather than re-using `hosting` which is already used to create accounts in the old signup framework. I kept them distinct otherwise the tracks events would get all mixed up.
- This gets the flow working, but we'll need to bring some of the style customisations we did for the `/start/hosting` flow (e.g. emphasising the refund period) over to this framework's "plans" step. This probably won't be too much work because it uses the same plan grid components under the hood.
- This PR doesn't include the new site setup step, that's coming in https://github.com/Automattic/dotcom-forge/issues/2360

Here's a video p1683599503739649-slack-C04H4NY6STW

## Testing Instructions

* Head to `/setup/new-hosted-site`
* It might not be styled perfectly, but you should be able to create a new site
* Ensure none of our changes bleed through to other flows e.g. we hide the free plan for our flow, ensure it's not hidden on other flows.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
+++ b/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
@@ -10,6 +10,7 @@ import {
 import { ONBOARD_STORE } from '../stores';
 import { recordSubmitStep } from './internals/analytics/record-submit-step';
 import type { Flow, ProvidedDependencies } from './internals/types';
+import './internals/hosting-site-creation-flow.scss';
 
 const hosting: Flow = {
 	name: HOSTING_SITE_CREATION_FLOW,
--- a/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
+++ b/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
@@ -1,0 +1,64 @@
+import { HOSTING_SITE_CREATION_FLOW } from '@automattic/onboarding';
+import { addQueryArgs } from '@wordpress/url';
+import {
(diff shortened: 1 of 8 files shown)


zaguiini (approved):
It would be cool if we could replicate the plan picker wrapper style so that it doesn't get too wide:

| This PR | `/start/hosting` picker |
| ------- | --------------- |
| image | image |

Important: only the picker gets narrower. The plan comparison table below is still almost full-width.

---

After picking a plan, I'm seeing a weird horizontal scrollbar at the top:

image

I do not see it on `/start/hosting`. It doesn't look related to your PR but we could use a follow-up if we see it in other stepper flows.

---

Approving, it works as expected 👍

zaguiini (changes requested):
Plan comparison is haunting us again 👻 

image

zaguiini:
For now, I'm hiding the plan comparison table. This behavior will be fixed on https://github.com/Automattic/wp-calypso/pull/76722.

When it's fixed, we can restore it. This PR doesn't seem like the right place to do it, so let's address it in a follow-up.

zaguiini (approved):

PR #76661, opened 2023-05-08
Preview limited Global Styles by default

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2267

## Proposed Changes

- Custom styles are previewed by default, so there is less confusion around the site-editor and the front-end not matching up.
- Both the site editor and the post editor display now a global notice when Global Styles are limited. Previous sidebar notices in the site editor have been removed. This notice has the following actions:
  - Upgrade now (primary): Visible in both editors – It opens the Plans page in a new tab.
  - Remove custom styles (link): Visible in the site editor only – It resets the styles to defaults (it’s easy and safe to do so in the editor). Once we have a support page explaining how to remove custom styles, we can show it in the post editor too.
- Add a new notice to the view canvas.
- Updates the E2E test to use the view canvas mode since that's the only place where we have a fragile CSS selector.

**Frontend**

Before | After
--- | ---
Screenshot 2023-05-18 at 11 03 58 | Screenshot 2023-05-18 at 11 06 33

**Site editor (view canvas)**

Before | After
--- | ---
Screenshot 2023-05-18 at 11 03 15 | Screenshot 2023-05-18 at 11 05 03

**Site editor (edit canvas)**

Before | After
---  | ---
Screenshot 2023-05-18 at 11 03 27 | Screenshot 2023-05-18 at 11 09 01

**Post editor**

Before | After
--- | ---
Screenshot 2023-05-18 at 11 03 47 | Screenshot 2023-05-18 at 11 06 20

## Testing Instructions

- Open the site editor.
- Make sure there is a global notice at the top of the editor.
- Make sure the "Upgrade now" action opens the Plans page.
- Make sure the "Remove custom styles" action resets the styles to defaults.
- Make sure the notice goes aways when there aren't custom styles.
- Open the post editor.
- Make sure the custom styles are visible.
- Make sure there is a global notice at the top of the editor.
- Make sure the "Upgrade now" action opens the Plans page.
- Make sure the "Preview your site" opens the frontend without custom styles previewed.
- Visit the frontend.
- Make sure custom styles are previewed by default.
- Go to `/setup/newsletter`.
- Pick a "premium" color.
- As soon as the site is created, sandbox it.
- Get to the Launchpad's last step.
- Ensure the preview is using the "premium" color.
- Follow E2E tests setup instructions from https://github.com/Automattic/wp-calypso/tree/trunk/test/e2e
- Sandbox the primary site of the `simpleSiteFreePlanUser` user.
- Run the E2E test:
  - Desktop: `yarn workspace wp-e2e-tests test -- specs/fse/fse__limited-global-styles.ts`
  - Mobile: `yarn workspace wp-e2e-tests test:mobile -- specs/fse/fse__limited-global-styles.ts`


BogdanUngureanu (approved):
LGTM! Should we wait shipping this until we clarify the nudges in the Post Editor? 🤔

mmtr:
@BogdanUngureanu I think we should yeah. I started adding the notices locally but it's not quite ready yet.

dsas:
I'm unsure about 'Remove custom styles' it looks like a link, just like it does in other places. In those other places though it is a link, here it performs a destructive (but undoable) action

dsas:
I know we want to encourage people more to Upgrading than resetting. But Upgrade now looks like a button and acts like a link, and Remove custom styles looks like a link and acts like a button.

Maybe they should both be buttons but with Upgrade now being a primary button and Remove custom styles being less emphasised.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php
```diff
@@ -423,10 +419,10 @@ function wpcom_display_global_styles_launch_bar( $bar_controls ) {
 
 	if ( wpcom_is_previewing_global_styles() ) {
 		$preview_text     = __( 'Turn off preview', 'full-site-editing' );
```
dsas: A separate issue perhaps but I think this is unclear in context:

image

Is it clear that this preview toggle is only for customized styles?

I think it should be 'Show customized styles' and 'Hide customized styles', or maybe "View with customized styles" and "View without customized styles"

Edit: Moved discussion to P2: pekYwv-1qM-p2#comment-1164

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.js
```diff
@@ -1,98 +1,124 @@
...
+				noDefaultClasses: true,
+			} );
+		}
 
-		for ( const entityTitle of preSavePanelTitles ) {
-			if ( entityTitle.textContent !== globalStylesConfig.label ) {
-				continue;
+		createWarningNotice(
+			__(
+				'Your site includes customized styles that are only visible to visitors after upgrading to the Premium plan or higher.',
```
BogdanUngureanu: What if we add a link on the `customized styles` that:
- In Site Editor will open the Global Styles sidebar (there's a method that can programmatically open the Ui without refresh).
- In the Post Editor will be a link to Site Editor, that would open the GS sidebar automatically (optional).

I think this would be helpful for users that come from onboarding and are unfamiliar with Global Styles (since they only interacted with Styles variations, fonts or colors).

I don't think we should point to the docs because:
 - We already do that on the front-end
 - They are already in edit mode; there's no point in redirecting them to docs and then coming back again to the editor to try it out (which might or might not happen).

mmtr: Unfortunately, we cannot add links without using the experimental `__unstableHTML` property which is highly discouraged: https://github.com/WordPress/gutenberg/blob/6606da447be0ba0885573add567dbb97b42c8ed2/packages/notices/src/store/selectors.js#L29-L33

BogdanUngureanu: I guess we can improve it in a future iteration. :)

dsas:
> Make sure the "Preview your site" opens the frontend without custom styles previewed.

This is different to the existing Preview menu: 
image

I think this difference in behaviour is ok, but I think ideally the new button might need to be clearer that it's a preview without styles.

Might be good to steal the new tab icon as well 🤔 or otherwise warn people it's going to open in a new tab. It's good for a11y too.

BogdanUngureanu:
> I think this difference in behaviour is ok, but I think ideally the new button might need to be clearer that it's a preview without styles.

Even more, what if we move that button directly in the preview section? Is it possible to add another item to that menu? This way, both modes (preview with styles and preview without styles) are in the same place.

mmtr:
> ideally the new button might need to be clearer that it's a preview without styles.

I totally agree! But I has been unable to come up with a good copy that is clear and not too wordy 😅 . "Preview your site without custom styles" seems too long.

> Might be good to steal the new tab icon as well

Yup, I tried that, but we cannot add icons / svg / custom React components to the actions passed to `createWarningNotice`.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/global-style-sidebar-notice.tsx
BogdanUngureanu, line 1: What if we leave the notice for mobile viewports? On mobile, the users are not getting an immediate notice that informs them that they have to upgrade. Maybe some of them might be confused about why they are receiving that notice. Plus, on these resolutions, the upgrade nudges are not competing since they aren't displayed at the same time. 

Before (please ignore the broken UI 😅 )

https://github.com/Automattic/wp-calypso/assets/995955/a43703d1-de8f-488b-8c95-a3fceee858e0

And after:

https://github.com/Automattic/wp-calypso/assets/995955/7e869a6e-2f75-4987-9a84-bd7d9d901706

mmtr: I wondered about that too, but ultimately decided that having one single notice could be enough:
- A user could save styles without closing the styles sidebar, so the notice could be missed. However, I think this is extremely unlikely, since a user would probably want to see how the site looks like before saving.
- A user could spend a very long time adjusting the styles before closing the styles sidebar and realizing they need to upgrade. However, I think this is also a non-realistic scenario since users would probably close the styles sidebar quite often to check the site rather than changing a lot styles blindly. We also show a modal the first time the user opens this sidebar.

Having that in consideration, I think showing a notice in the sidebar is only useful to cover very edge cases and not worth it. I like to have only one single notice to maintain.

dsas:
> unable to come up with a good copy that is clear and not too wordy

Yeah 🤔 We don't say what you're previewing in the normal GB menu, so we probably don't have to here. 'Preview without custom styles' is still a bit wordy but might be ok?

mmtr:
> Maybe they should both be buttons but with Upgrade now being a primary button and Remove custom styles being less emphasised.

Done in https://github.com/Automattic/wp-calypso/pull/76661/commits/8911cbc5539335ca91986f99b9dadce50c2c8bc7

Screenshot 2023-05-15 at 12 42 33

> 'Preview without custom styles' is still a bit wordy but might be ok?

Let's try that, we can always change it if needed. Done in https://github.com/Automattic/wp-calypso/pull/76661/commits/18acc7dc0071a3d27f726160fed6d1362a275a9c

Screenshot 2023-05-15 at 12 42 51

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.js
```diff
@@ -1,98 +1,124 @@
...
-
-	return (
-		<Notice
-			status="warning"
-			isDismissible={ false }
-			className="wpcom-global-styles-notice notice-margin"
-		>
-			{ createInterpolateElement(
-				__(
-					'Your changes include customized styles that will only be visible once you <a>upgrade to a Premium plan</a>.',
```
BogdanUngureanu, line 46: Something just crossed my mind. I think this notice is still needed in Site Editor, in browse mode, when the user saves a Style variation from the "Styles" panel. 

When the user clicks on the "Save" button, we display a save notice where we also mention that they are about to save a Global Styles change and they have to upgrade.

BogdanUngureanu: In that mode, from what I've observed, we don't display the Site Editor warning notice.

mmtr: Oh nice catch, I totally missed that. I'll add this notice back but only when on the "view" canvas mode.

dsas:
Looks like this now needs a very quick rebase

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
```diff
@@ -0,0 +1,185 @@
...
+				'full-site-editing'
+			),
+			{
+				id: NOTICE_ID,
+				isDismissible: false,
+				actions: actions,
+			}
+		);
+
+		trackEvent( 'calypso_global_styles_gating_notice_show' );
```
BogdanUngureanu: Looks like this event is triggered in browse mode, from the `Styles` mode. Is this intended? :)

mmtr: Nice catch! Fixed in https://github.com/Automattic/wp-calypso/pull/76661/commits/f199e38dc3733a233b49ce861bca04bfff52d627

BogdanUngureanu: Also, in PostEditor, the context is set to `site-editor` instead of `post-editor`

mmtr: Another excellent finding. Fixed in https://github.com/Automattic/wp-calypso/pull/76661/commits/4ed46614aeef43214ed45dfb6a0e0db5038452e6

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
```diff
@@ -0,0 +1,185 @@
...
+	const { createWarningNotice, removeNotice } = useDispatch( 'core/notices' );
+	const { editEntityRecord } = useDispatch( 'core' );
+
+	const upgradePlan = useCallback( () => {
+		window.open( wpcomGlobalStyles.upgradeUrl, '_blank' ).focus();
+		trackEvent( 'calypso_global_styles_gating_notice_upgrade_click', isSiteEditor );
+	}, [ isSiteEditor ] );
+
+	const previewSite = useCallback( () => {
+		window.open( wpcomGlobalStyles.previewUrl, '_blank' ).focus();
```
BogdanUngureanu: When in Post Editor mode, I think the preview URL should be the post itself, not the main page. Otherwise, the user doesn't have a way to see how their page/post would look like without styles.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
```diff
@@ -0,0 +1,186 @@
...
+		trackEvent( 'calypso_global_styles_gating_notice_preview_click', isSiteEditor );
```
BogdanUngureanu: Maybe we should add the post id as an extra prop? I **think** when you add a different extra prop, the event would be considered unique. This way, a user clicking from two posts will be considered two different clicks. 

The same thing should be for `calypso_global_styles_gating_notice_show` too; we want to register the event twice if the user saw the upgrade notice from two different posts/pages.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		const footer = document.querySelector( '.edit-site-sidebar__footer' );
+		if ( ! footer ) {
+			return;
+		}
+
+		const noticeContainer = document.createElement( 'div' );
+		footer.prepend( noticeContainer );
+		render( <GlobalStylesWarningNotice />, noticeContainer );
```
BogdanUngureanu, line 68: Nitpick: this triggers a warning in the console:

```
react_devtools_backend_compact.js:2367 Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot
```

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
```diff
@@ -0,0 +1,186 @@
...
+
+	if ( ! globalStylesInUse ) {
+		return null;
+	}
+
+	return (
+		<Notice status="warning" isDismissible={ false } className="wpcom-global-styles-notice">
+			{ createInterpolateElement(
+				__(
+					'Your site includes customized styles that are only visible to visitors after <a>upgrading to the Premium plan or higher</a>.',
```
BogdanUngureanu, line 34: Idea: what if we add a link on the "customized styles" text that would open the GS panel? We should also add an event to it too.

BogdanUngureanu:
Gutenberg edge now offers support for customizing Pages in Site Editor. Should we remove the `Remove styles` button from Pages displayed in the Site Editor? The user might have the expectation that the change would only affect the page, not Site Wide. 🤔

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
mmtr: I think "uniques" refers to distinct users, not events. See PCYsg-5hU-p2:

> Uniques – the number of distinct users associated with an event has been recorded in a timeframe

See also `/wp-content/lib/nosara/tracks-trends.php#281`

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
mmtr: I think that's worth trying! Maybe in a follow-up?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
mmtr: I agree that's a better approach, but it's not that trivial to do so because we first need to save the post and then preview it. We basically need to port/duplicate this logic: https://github.com/WordPress/gutenberg/blob/83a33e7c24ee1db8fe426052ac73137f51289234/packages/editor/src/components/post-preview-button/index.js#L149

While doable, I'm reluctant about making this PR more complex, so can we maybe keep the site preview for now and add the post preview in a immediate follow-up? Alternatively, I can also remove the preview button.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
mmtr: Unfortunately, `createRoot` is not exposed in the `@wordpress/element` version used by the ETK plugin.

mmtr:
> The user might have the expectation that the change would only affect the page, not Site Wide. 🤔

Wouldn't that be also the case when using the "Reset to defaults" button from the Styles sidebar (which is also available when editing pages in the site editor)?

Since they have a option to edit styles globally across the whole site, I think it's ok to have an action in the notice that affects the global styles.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
BogdanUngureanu: I think it makes sense to remove the Preview button for now and make a follow-up PR for it. :)

BogdanUngureanu:
> Wouldn't that be also the case when using the "Reset to defaults" button from the Styles sidebar (which is also available when editing pages in the site editor)?
> 
> Since they have a option to edit styles globally across the whole site, I think it's ok to have an action in the notice that affects the global styles.

I think it's debatable. We are adding this option for users that aren't familiar with Global Styles. Otherwise, they'll just use the "Reset" from the Global Styles panel, so some confusion might be possible since they aren't familiar with how GS work. 

Also, the notice is part of the editor UI whereas the reset is part of the Global Styles panel. Some users might consider that since the notice is in the editor, it might only affect that specific page. And, who knows, maybe that GS panel shouldn't be displayed for pages, in core. It's a pretty new addition (the pages) and most likely will be refined in the future.

dsas:
> We are adding this option for users that aren't familiar with Global Styles. Otherwise, they'll just use the "Reset" from the Global Styles panel, so some confusion might be possible since they aren't familiar with how GS work.

That's helpful - the venn diagram of users familiar with GS gating and users that know how to reset GS is probably a circle. It's not the sole reason - it's also much more convenient for everyone else. 

> And, who knows, maybe that GS panel shouldn't be displayed for pages, in core. It's a pretty new addition (the pages) and most likely will be refined in the future.

I can't see hiding GS panel on the [GS roadmap](https://github.com/WordPress/gutenberg/issues/41232), [GB phase 2 roadmap](https://github.com/WordPress/gutenberg/issues/33094) or the [plan for reintroducing content editing to the site editor](https://github.com/WordPress/gutenberg/issues/44461). In the absence of it being mentioned I wouldn't place any weight on the GS panel maybe disappearing, I think that would also make a pretty weird and confusing UI. It wouldn't be possible to work on some GS for a combination of blocks that only appears on one page at present, but you know you're going to add more pages in future.

If it does happen, we'll react to it then. Lets not block something on hypothetical unknown futures especially if we can quickly and easily react to it if it materialises.

I think it's ok to have an action in the notice that affects the whole site. If it turns out to be a problem we can remove it. We could also consider rewording the button to be clear that it applies globally, but I'm not sure that is necessary myself.

mmtr:
> Some users might consider that since the notice is in the editor, it might only affect that specific page

Is not that also true when editing the homepage? How would you differentiate users who think that are editing only one page from those who are editing a site?

Or putting it in a different way, check these two videos:

https://github.com/Automattic/wp-calypso/assets/1233880/8e11f5b1-61ea-4773-b6a3-1e7ec79b5229


https://github.com/Automattic/wp-calypso/assets/1233880/48429037-c9da-4037-91c9-5079337b1101


Do you think we should hide the action in the second scenario despite editing exactly the same content?

BogdanUngureanu:
Just did another testing session and everything looks nice! Let's ship it! 🐑

dsas (approved):
Non-blocking query: When removing styles I get a snack notice saying 'Site updated', should/could that message be something more relevant like 'Site customized styles updated'?

davewhitley:
Great work! This is great to see almost shipped. The only reservation I have at this point is that perhaps it should have a close button that temporarily dismisses the notice. Now that it's big and in the forefront, this could be a problem. People might just want it out of the way because they are working on something not styles related and accidentally click "Remove all styles" just to dismiss the notice.

Not sure if this is still relevant but we do have a doc on removing custom styles https://wordpress.com/support/using-styles/#reset-all-styles

mmtr:
> Non-blocking query: When removing styles I get a snack notice saying 'Site updated', should/could that message be something more relevant like 'Site customized styles updated'?

We don't have control over that notice, it's generated by Core: https://github.com/WordPress/gutenberg/blob/a0a1fd51e542b5c97f166f7be1679c2858dbb96e/packages/editor/src/components/entities-saved-states/index.js#L211

> The only reservation I have at this point is that perhaps it should have a close button that temporarily dismisses the notice.

I think that's a good idea! I'll add it before shipping this.

> we do have a doc on removing custom styles

Nice! I'll add a link to that in a follow-up PR.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7995550

Thank you @mmtr for including a screenshot in the description! This is really helpful for our translators.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notices.js
mmtr: @BogdanUngureanu I was about to implement this on a new PR, but I'm having second thoughts now. The notice would end up with two competing links (one for upgrading, and one for opening the Styles sidebar) since we don't have any good way of giving more visual priority to one of them. This is one of the reasons why we removed the "reset styles" link we had originally (see https://github.com/Automattic/dotcom-forge/issues/1221).

BogdanUngureanu: @mmtr let's see the CTR on that link tomorrow. If the notice is shown a lot but the CTR is low, I think having a link would make a lot of sense because we'll be moving the users to the panel for further customization.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7995550).

PR #76664, opened 2023-05-08
Social Connections: Support connecting multiple accounts

gmjuhasz:
This change updates the connections page in Calypso, so it is possible to connect multiple accounts.

## Proposed Changes

* Refactored the `createOrUpdateConnection ` function, as now we create a connection each time, because we support multiple pages per site.
* Changed the logic to determine conflicting connections. Before this change two connections were conflicting, if they had the same keyring ID, but had different connection ID-s. From now on, these won't be conflicts, as we support multiple connections / site. The only conflict now is they want to connect the same site again - This should not be possible from the UI for most sites.

## Testing Instructions

* Check out this change and start Calypso with `yarn start`
* Go to `http://calypso.localhost:3000/marketing/connections/gmjuhasztest.jurassic.tube` for your site.
* Sandbox D109448-code patch, and point the wp api to your sandbox in your hosts file.
* Add this blog sticker to your site : `add_blog_sticker ('jetpack-social-allow-multiple-connections', null, null,  <blog-id>); `
Now test the patch:
- Try to add another connection for a site. You might have to change the API settings, to allow access to the second page too. For Facebook click settings in the popup, and tick the new site, then continue.
- You should see the new site, and your previous one as connected. There should be no warning of conflicts.
CleanShot 2023-05-08 at 13 18 30 png

* After connecting, it should appear in a list
* Try disconnecting them 1 by 1.
* Try using the disconnect all button

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/marketing/connections/account-dialog.jsx
+++ b/client/my-sites/marketing/connections/account-dialog.jsx
@@ -89,9 +89,11 @@ class AccountDialog extends Component {
 	}
 
 	areAccountsConflicting( account, otherAccount ) {
+		// Accounts are conflicting, if they have the same ID and keyringConnectionId, which
+		// means they are the same account. Otherwise it's a different site for the same account.
 		return (
 			account.keyringConnectionId === otherAccount.keyringConnectionId &&
-			account.ID !== otherAccount.ID
+			account.ID === otherAccount.ID
 		);
 	}
 
--- a/client/my-sites/marketing/connections/account-dialog.jsx
+++ b/client/my-sites/marketing/connections/account-dialog.jsx
@@ -1,3 +1,4 @@
+import { isEnabled } from '@automattic/calypso-config';
 import { Dialog } from '@automattic/components';
 import classNames from 'classnames';
 import { localize } from 'i18n-calypso';
@@ -91,10 +92,11 @@ class AccountDialog extends Component {
 	areAccountsConflicting( account, otherAccount ) {
 		// Accounts are conflicting, if they have the same ID and keyringConnectionId, which
 		// means they are the same account. Otherwise it's a different site for the same account.
-		return (
-			account.keyringConnectionId === otherAccount.keyringConnectionId &&
-			account.ID === otherAccount.ID
-		);
+		const isIdConflicting = isEnabled( 'jetpack-social/multiple-connections' )
+			? account.ID === otherAccount.ID
+			: account.ID !== otherAccount.ID;
+
+		return account.keyringConnectionId === otherAccount.keyringConnectionId && isIdConflicting;
 	}
 
 	isSelectedAccountConflicting() {
--- a/client/my-sites/marketing/connections/account-dialog.jsx
+++ b/client/my-sites/marketing/connections/account-dialog.jsx
@@ -90,13 +90,15 @@ class AccountDialog extends Component {
 	}
 
 	areAccountsConflicting( account, otherAccount ) {
-		// Accounts are conflicting, if they have the same ID and keyringConnectionId, which
-		// means they are the same account. Otherwise it's a different site for the same account.
-		const isIdConflicting = isEnabled( 'jetpack-social/multiple-connections' )
-			? account.ID === otherAccount.ID
-			: account.ID !== otherAccount.ID;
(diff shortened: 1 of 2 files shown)


client/my-sites/marketing/connections/service.jsx
```diff
@@ -162,8 +162,8 @@ export class SharingService extends Component {
 		if ( service ) {
 			if ( keyringConnectionId ) {
 				// Since we have a Keyring connection to work with, we can immediately
-				// create or update the connection
-				this.createOrUpdateConnection( keyringConnectionId, externalUserId );
```
spsiddarthan: We should do this only if the feature flag is set, no?

client/my-sites/marketing/connections/service.jsx
```diff
@@ -218,24 +218,13 @@ export class SharingService extends Component {
...
 
 	/**
-	 * Create or update the connection
+	 * Create the connection for the site
 	 *
 	 * @param {number} keyringConnectionId Keyring conneciton ID.
-	 * @param {number} externalUserId      Optional. User ID for the service. Default: 0.
+	 * @param {number} externalUserId      User ID for the service.
 	 */
-	createOrUpdateConnection = ( keyringConnectionId, externalUserId = 0 ) => {
```
spsiddarthan, line 226: May be leave the `createOrUpdateConnection` as well till we get rid of the feature flag?

spsiddarthan (changes requested):
This looks good, @gmjuhasz, and thanks for fixing the conflicting connections part as well, I had a separate task for that, great work taking care of that. 

We just need to add the feature flag checks and update this PR.

client/my-sites/marketing/connections/service.jsx
gmjuhasz: Yeah, I'll change this when the Feature flag is ready!

client/my-sites/marketing/connections/service.jsx
gmjuhasz: Sure!

client/my-sites/marketing/connections/account-dialog.jsx
```diff
@@ -89,9 +89,11 @@ class AccountDialog extends Component {
 	}
 
 	areAccountsConflicting( account, otherAccount ) {
+		// Accounts are conflicting, if they have the same ID and keyringConnectionId, which
+		// means they are the same account. Otherwise it's a different site for the same account.
 		return (
```
spsiddarthan, line 98: This should be behind the feature flag as well, so we don't change the behaviour for existing accounts.

client/my-sites/marketing/connections/account-dialog.jsx
```diff
@@ -89,10 +90,13 @@ class AccountDialog extends Component {
...
 	areAccountsConflicting( account, otherAccount ) {
-		return (
-			account.keyringConnectionId === otherAccount.keyringConnectionId &&
-			account.ID !== otherAccount.ID
-		);
+		// Accounts are conflicting, if they have the same ID and keyringConnectionId, which
+		// means they are the same account. Otherwise it's a different site for the same account.
+		const isIdConflicting = isEnabled( 'jetpack-social/multiple-connections' )
+			? account.ID === otherAccount.ID
+			: account.ID !== otherAccount.ID;
```
pablinos: I'm not sure this would ever happen. If these IDs are the same then would that mean we've sent back a duplicate connection?

I think now we want to make sure that (if we can) we check the external ID for the connection. We don't want someone connecting the same Facebook page twice, but maybe we should handle that on the server and make sure that the site doesn't have two connections to the same external entity.

gmjuhasz: Yeah it would mean that it's duplicated. For facebook, you cannot pick the same site twice, but if there is a site that let's you do that, I think we should make sure to show it as a conflict, as it was already connected. Do you think we should just not check for the ID in case of multiple connections?

gmjuhasz: Updated to just return `false` for now as we agreed on the Slack huddle.

spsiddarthan (approved):

PR #76667, opened 2023-05-08
Use more granular feature checks in Hosting Configuration page

daledupreez:
## Proposed Changes

This PR tackles a number of changes stemming from some assumptions that were made when building the Hosting Configuration page in Calypso. The core issue was that all Atomic features were assumed to be controlled via the SFTP feature, but the recent introduction of free eCommerce trials invalidates that assumption, as those sites should have access to Atomic features but not SSH or SFTP access. As such, this PR tackles a few things:
 * We split out hosting features into `basic` and `advanced` buckets, with the following features falling into the basic Atomic bucket behind the `WPCOM_FEATURES_ATOMIC` feature:
    - Transferring the site to Atomic (which has come up in some cases where initial transfers failed)
    - Restoring plugins and themes from the plan
    - Clearing the site cache
      * **Note:** We currently don't include PHP version and web server configuration in this bucket, as it's not clear that free trials should get access to that functionality - it would be better if we have a dedicated feature for this, but we don't currently have that.
 * The logic for showing the activation and upsell banners has been updated:
    - If the site doesn't have the Atomic feature, we show the upsell banner
    - If the site is Atomic, is not transferring, and doesn't have advanced features (as identified by the SFTP feature), we show the upsell banner
       * This is a departure from the old logic, as we still want an upsell for sites that don't have full hosting access (which is currently limited to Woo Express trials)
    - Otherwise we show the activation banner (which may be null/empty)
 * We keep the existing card sequence in the main part of the page and the sidebar, but we re-order them such that any cards for unavailable features are pushed to a `FeatureExample` that appears below the features that are available.
   - This makes it easier for users to access the features they do have access to, and if there are unavailable features, the upsell banner updates in #76783 should make it clearer what those features are.

This PR depends on #76783 to ensure we show consistent upsell banner content for all site types.

### Unit tests

Review the test cases covered by the unit tests, and verify that the unit tests succeed:
```
yarn test-client client/my-sites/hosting/test/index.js
```

### Manual testing

* Run this branch locally or via Calypso.live using the link in [this comment](https://github.com/Automattic/wp-calypso/pull/76667#issuecomment-1538213633).
* For each of the test cases below, we want to navigate to _Settings_ -> _Hosting Configuration_ (i.e. `/hosting-config/:siteSlug`, and confirm that the expectations are met.

#### Test cases

  - A site on WPCOM Simple that doesn't have Atomic features, i.e. the site is free, or has a Personal or Premium plan
    * You should see the upsell banner and the _Support_ card, and all other feature cards should be obscured
Screenshot 2023-05-17 at 11 53 29

  - A site on WPCOM Simple that can go Atomic, e.g. a Simple site with the WordPress.com Business plan
    * You should see the activation banner that allows you to enable hosting features, and the _Support_ card should be visible. All other cards should be obscured.
Screenshot 2023-05-17 at 12 48 07

  - An Atomic site that already has access to all Atomic features - this includes the Business, eCommerce, and paid Woo Express plans
    * You should see all cards, and no upsell or activation banner
Screenshot 2023-05-17 at 11 53 12 

  - An Atomic site with access to limited features, which specifically requires a site with the Woo Express trial plan added -- you can create such a site by working through the flow from https://woocommerce.com/start.
    * You should see the upsell banner, the _Restore plugins and themes_ card, the _Cache_ card, the _Site backup card, and the _Support_ card. All other cards in the main section should be obscured.
Screenshot 2023-05-17 at 11 52 56

  - Use the Atomic revert tool to revert your Woo Express trial site to WordPress.com Simple.
  - Once the revert completes, navigate to _Settings_ -> _Hosting Configuration_
  - Verify that you are shown the activation banner which allows you to request activation of the site's Atomic features
Screenshot 2023-05-17 at 12 14 38

  - Click on the "Activate" CTA in the banner, and proceed with the transfer
  - Verify that you see the "Please wait" message in the activation banner while the site transfers to Atomic
Screenshot 2023-05-17 at 12 14 57

## Pre-merge Checklist

- [N/A] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [N/A] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

### DIFF
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -5,8 +5,7 @@ import {
 	FEATURE_SITE_STAGING_SITES,
 	WPCOM_FEATURES_ATOMIC,
 } from '@automattic/calypso-products';
-import { englishLocales } from '@automattic/i18n-utils';
-import i18n, { localize } from 'i18n-calypso';
+import { localize } from 'i18n-calypso';
 import { Component, Fragment, useMemo } from 'react';
 import wrapWithClickOutside from 'react-click-outside';
 import { connect } from 'react-redux';
@@ -208,7 +207,6 @@ class Hosting extends Component {
 			isSiteAtomic,
 			isTransferring,
 			isWpcomStagingSite,
-			locale,
 			requestSiteById,
 			siteId,
 			siteSlug,
@@ -272,18 +270,17 @@ class Hosting extends Component {
 					/>
 				);
 
-				// Don't imply additional access if site doesn't have the SFTP feature.
-				const noticeText =
-					isBasicHostingDisabled &&
-					( englishLocales.includes( locale ) ||
-						i18n.hasTranslation( 'Please activate your hosting access.' ) )
-						? translate( 'Please activate your hosting access.' )
-						: translate( 'Please activate the hosting access to begin using these features.' );
-
 				return (
 					<>
 						{ failureNotice }
-						<Notice status="is-info" showDismiss={ false } text={ noticeText } icon="globe">
+						<Notice
+							status="is-info"
+							showDismiss={ false }
+							text={ translate(
+								'Please activate the hosting access to begin using these features.'
+							) }
+							icon="globe"
+						>
 							<TrackComponentView eventName="calypso_hosting_configuration_activate_impression" />
 							<NoticeAction
 								onClick={ clickActivate }
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -6,7 +6,7 @@ import {
 	WPCOM_FEATURES_ATOMIC,
 } from '@automattic/calypso-products';
 import { localize } from 'i18n-calypso';
-import { Component, Fragment, useMemo } from 'react';
+import { Component, Fragment } from 'react';
 import wrapWithClickOutside from 'react-click-outside';
 import { connect } from 'react-redux';
 import DocumentHead from 'calypso/components/data/document-head';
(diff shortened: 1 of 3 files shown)


danielbachhuber (commented):
> We need to test the following cases:

Can we add some JavaScript tests for these scenarios?

client/my-sites/hosting/main.js
```diff
@@ -143,27 +147,28 @@ class Hosting extends Component {
...
+					/>
+				);
+
+				// Don't imply additional access if site doesn't have the SFTP feature.
+				const noticeText =
+					isBasicHostingDisabled &&
+					( englishLocales.includes( locale ) ||
+						i18n.hasTranslation( 'Please activate your hosting access.' ) )
+						? translate( 'Please activate your hosting access.' )
+						: translate( 'Please activate the hosting access to begin using these features.' );
```
p-jackson: IMO the extra case—taking a WooExpress site Atomic—doesn't warrant its own copy. 1) it doesn't seem that likely 2) "these features" doesn't feel like it implies the SFTP to me at least, it's more implies the "advanced" features mentioned above in the sub title.

woo express that is simple

p-jackson (commented):
> Can we add some JavaScript tests for these scenarios?

This'd be great! Might be a heavy lift because it doesn't look like there are any RTL tests for this page yet, so there'll be a bunch of up-front mocking to do. But there are so many permutations of this page, it'd be great to get coverage of them without having to manually set up all the situations.

> We need to test the following cases

I've tested these cases and things are working as expected.

This case where the WooExpress trial site is on Atomic doesn't look great IMO. The main issue for me is the gradient. Would it look better to just have a solid greyed out look rather than something that fades to transparent?

CleanShot 2023-05-10 at 11 40 11@2x

daledupreez:
Thanks for the feedback, @danielbachhuber and @p-jackson!

I realised that a big part of my broader confusion was due to us using the older upsell content/style when the site has an eCommerce trial plan. When we implemented an alternate upsell path for those plans, we used the older nudge, as you were still experimenting with the now-default upsell. I've split out #76783 to use the new style, and will keep iterating on this PR so we do the following:
* Show the activation banner to transfer to Atomic when the site has the base Atomic feature, but is not Atomic, or is transferring to Atomic
* Refactor the code so we show all available cards at the top of the screen, and keep all unavailable feature cards behind the `FeatureExample` block
  - The core thinking here is that if the user has basic Atomic features, the only things they can access are the plugin restore and clear cache features
  - While the web server settings are technically a general platform feature, I don't know that we want folks with plain Atomic features to manipulate their PHP version and other web server settings

daledupreez:
> > We need to test the following cases:
>
> Can we add some JavaScript tests for these scenarios?

Absolutely! I just wanted to explore the approach and get some feedback before writing tests.

danielbachhuber:
> I've split out https://github.com/Automattic/wp-calypso/pull/76783 to use the new style, and will keep iterating on this PR so we do the following:

Cool, sounds good.

> While the web server settings are technically a general platform feature, I don't know that we want folks with plain Atomic features to manipulate their PHP version and other web server settings

We may need to keep the PHP version switcher available, so they can easily roll back when a new PHP version is available, etc.

client/my-sites/hosting/main.js
daledupreez: Agreed.

daledupreez: And the one case where it would occur is for stuck transfers, which we have started tripping over. So this is more likely to be something seen by HEs, but they still need a path for re-triggering a transfer.

client/my-sites/hosting/main.js
```diff
@@ -53,6 +54,125 @@ import WebServerSettingsCard from './web-server-settings-card';
...
+
+const MainCards = ( {
+	hasStagingSitesFeature,
+	isAdvancedHostingDisabled,
+	isBasicHostingDisabled,
+	isGithubIntegrationEnabled,
+	isWpcomStagingSite,
+	siteId,
+} ) => {
+	const mainCards = useMemo( () => {
```
daledupreez: Is this unnecessary? I am thinking that I'm mostly depending on the incoming props, which means React should (hopefully) identify that the component's props are unchanged.

p-jackson: I'm pretty sure React re-renders the component whenever the parent re-renders; regardless of whether the props have changed (unless you use `React.memo`). But the `useMemo` still isn't necessary IMO. The filtering logic doesn't look expensive, so unless there's a measurable performance impact I think the logic is clearer without pre-emptively memoising things.

sixhours:
I went through the testing instructions and the upsell card does *not* show on a fresh Woo Express Trial site until after I've reverted it and restored it to Atomic.

Upon purchasing the Trial and navigating to `/hosting-config/siteSlug`:

Screen Shot 2023-06-01 at 9 56 44 AM

After reverting to Simple, then activating hosting features to transfer the site back to Atomic:

Screen Shot 2023-06-01 at 10 04 24 AM

daledupreez:
Thanks for the review, @sixhours!

I dug into that specific behaviour for the newly-created Woo Express trial site, and I found that this is a situation that is almost certainly transient, and is linked to the site not being fully synched yet.

I believe the relevant code is here:
https://github.com/Automattic/wp-calypso/blob/b60e83c34d848e3b4cbc977cc062c0e4069216fc/client/blocks/upsell-nudge/index.jsx#L79-L93

This is the component that we use for the upsell nudge, albeit via the `HostingUpsellNudge` wrapper component.

I _think_ we can move forward with the patch, as that behavour seems to be a timing issue that will resolve after a few minutes.

sixhours (approved):
In that case, 🚢 it! :D

PR #76684, opened 2023-05-08
Plugins: Make plugin actions handle site data

robertf4:
## Proposed Changes

When the TypeScript selectors in https://github.com/Automattic/wp-calypso/pull/76434 are used, the plugin actions will need to handle objects both with and without site data on them. This PR makes that change.

## Testing Instructions

This PR should create no functional changes. Just regression test the plugin actions that correspond to these actions.

## Pre-merge Checklist

### DIFF
--- a/client/state/plugins/installed/actions.js
+++ b/client/state/plugins/installed/actions.js
@@ -52,16 +52,22 @@ import { getSite } from 'calypso/state/sites/selectors';
 import 'calypso/state/plugins/init';
 
 /**
- * Determines if site specific data is on the plugin object.
- * If it is not, then it can be assumed it is on the sites property.
+ * Determines the truthiness of a site specific property regardless of whether it is on the plugin object
+ * or on one of the plugin's site objects.
  *
- * @param {Object} plugin - plugin object
- * @returns {boolean} True if any of the site specific properties are on the plugin, false otherwise.
+ * @param {string} prop - The site property to check. One of 'active', 'autoupdate', 'update', or 'version'.
+ * @param {Object} plugin - The plugin object
+ * @param {number} siteId - The ID of the site if it exists
+ * @returns {boolean} True if the site specific property is on the plugin, false otherwise.
  */
-const isSiteDataOnPluginObject = ( plugin ) => {
-	return [ 'active', 'autoupdate', 'update', 'version' ].some( ( prop ) =>
-		plugin.hasOwnProperty( prop )
-	);
+const pluginHasTruthySiteProp = ( prop, plugin, siteId ) => {
+	if ( ! [ 'active', 'autoupdate', 'update', 'version' ].includes( prop ) ) {
+		throw new Error( `${ prop } is not a site property.` );
+	}
+
+	return plugin.hasOwnProperty( prop )
+		? plugin?.[ prop ]
+		: siteId && plugin.sites?.[ siteId ]?.[ prop ];
 };
 
 /**
@@ -144,11 +150,7 @@ export function activatePlugin( siteId, plugin ) {
 			pluginId,
 		};
 
-		if (
-			isSiteDataOnPluginObject( plugin )
-				? plugin.active
-				: siteId && plugin.sites?.[ siteId ]?.active
-		) {
+		if ( pluginHasTruthySiteProp( 'active', plugin, siteId ) ) {
 			return dispatch( { ...defaultAction, type: PLUGIN_ACTIVATE_REQUEST_SUCCESS, data: plugin } );
 		}
 
@@ -215,11 +217,7 @@ export function deactivatePlugin( siteId, plugin ) {
 			pluginId,
 		};
 
-		if (
-			isSiteDataOnPluginObject( plugin )
-				? ! plugin.active
-				: siteId && ! plugin.sites?.[ siteId ]?.active
-		) {
+		if ( ! pluginHasTruthySiteProp( 'active', plugin, siteId ) ) {
 			return dispatch( {
 				...defaultAction,
 				type: PLUGIN_DEACTIVATE_REQUEST_SUCCESS,
@@ -282,9 +280,7 @@ export function togglePluginActivation( siteId, plugin ) {
 			return;
 		}
 
-		if (
-			isSiteDataOnPluginObject ? ! plugin.active : siteId && ! plugin.sites?.[ siteId ]?.active
-		) {
+		if ( ! pluginHasTruthySiteProp( 'active', plugin, siteId ) ) {
 			dispatch( activatePlugin( siteId, plugin ) );
 		} else {
 			dispatch( deactivatePlugin( siteId, plugin ) );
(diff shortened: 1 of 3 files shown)


client/state/plugins/installed/actions.js
```diff
@@ -131,7 +144,11 @@ export function activatePlugin( siteId, plugin ) {
 			pluginId,
 		};
 
-		if ( plugin.active ) {
+		if (
+			isSiteDataOnPluginObject( plugin )
+				? plugin.active
+				: siteId && plugin.sites?.[ siteId ]?.active
```
tyxla: This looks a bit like a code smell - why are we checking if the plugin has `version` or `update` when we're going to be working with the `active` prop? 

Perhaps instead we could pass the prop we're looking for to the function and return true only if that prop is there, and not if some unrelated props are there?

Also, this seems a bit repetitive. Perhaps we can bake in the check inside the function?

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,25 @@ import { getSite } from 'calypso/state/sites/selectors';
...
+ * Determines the truthiness of a site specific property regardless of whether it is on the plugin object
+ * or on one of the plugin's site objects.
+ *
+ * @param {string} prop - The site property to check. One of 'active', 'autoupdate', 'update', or 'version'.
+ * @param {Object} plugin - The plugin object
+ * @param {number} siteId - The ID of the site if it exists
+ * @returns {boolean} True if the site specific property is on the plugin, false otherwise.
+ */
+const pluginHasTruthySiteProp = ( prop, plugin, siteId ) => {
+	if ( ! [ 'active', 'autoupdate', 'update', 'version' ].includes( prop ) ) {
```
robertf4, line 64: Ideally, if this was TypeScript then I would have just made `prop` have a type of `'active' | 'autoupdate' | 'update' | 'version'`. I did this instead to ensure someone doesn't accidentally misuse this function for a prop that isn't site-specific.

client/state/plugins/installed/actions.js
robertf4: I changed the approach. Let me know what you think.

client/state/plugins/installed/actions.js
```diff
@@ -278,7 +297,10 @@ export function updatePlugin( siteId, plugin ) {
 			pluginId,
 		};
 
-		if ( ! plugin?.update || plugin?.update?.recentlyUpdated ) {
+		if (
+			! pluginHasTruthySiteProp( 'update', plugin, siteId ) ||
+			( siteId && plugin?.sites?.[ siteId ]?.update?.recentlyUpdated )
```
robertf4, line 302: This is the only check which breaks the pattern since it is looking for `plugin.update.recentlyUpdated`. I figured it was better to just have this exception than to complicate `pluginHasTruthySiteProp` by making it handle this case as well.

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,25 @@ import { getSite } from 'calypso/state/sites/selectors';
...
+ * @param {number} siteId - The ID of the site if it exists
+ * @returns {boolean} True if the site specific property is on the plugin, false otherwise.
+ */
+const pluginHasTruthySiteProp = ( prop, plugin, siteId ) => {
+	if ( ! [ 'active', 'autoupdate', 'update', 'version' ].includes( prop ) ) {
+		throw new Error( `${ prop } is not a site property.` );
+	}
+
+	return plugin.hasOwnProperty( prop )
+		? plugin?.[ prop ]
```
tyxla: Seems like we don't need the optional chaining since `hasOwnProperty()` is `true`:

```suggestion
		? plugin[ prop ]
```

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,25 @@ import { getSite } from 'calypso/state/sites/selectors';
 
 import 'calypso/state/plugins/init';
 
+/**
+ * Determines the truthiness of a site specific property regardless of whether it is on the plugin object
+ * or on one of the plugin's site objects.
+ *
+ * @param {string} prop - The site property to check. One of 'active', 'autoupdate', 'update', or 'version'.
+ * @param {Object} plugin - The plugin object
+ * @param {number} siteId - The ID of the site if it exists
```
tyxla: If it exists? This adds some confusion, since the ID is expected to always be a number. Is the idea that we might have the site ID but the plugin data for the site may be missing?

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,25 @@ import { getSite } from 'calypso/state/sites/selectors';
...
+ * @returns {boolean} True if the site specific property is on the plugin, false otherwise.
+ */
+const pluginHasTruthySiteProp = ( prop, plugin, siteId ) => {
+	if ( ! [ 'active', 'autoupdate', 'update', 'version' ].includes( prop ) ) {
+		throw new Error( `${ prop } is not a site property.` );
+	}
+
+	return plugin.hasOwnProperty( prop )
+		? plugin?.[ prop ]
+		: siteId && plugin.sites?.[ siteId ]?.[ prop ];
```
tyxla: If the site ID is there and the prop doesn't exist, this will return `undefined`, while the function promises to always return a boolean.

tyxla: Also, any chance `plugin.sites` will be undefined? That particular optional chaining might be unnecessary.

client/state/plugins/installed/actions.js
tyxla: Thank you, I think it looks great when abstracted 👍

client/state/plugins/installed/actions.js
robertf4: Just bad wording on my part. I removed it.

client/state/plugins/installed/actions.js
robertf4: The documentation was incorrect in this case.

robertf4: Yes, it is undefined if the data object has come directly from the server. The `sites` prop is calculated on the client by the plugin selector.

client/state/plugins/installed/actions.js
```diff
@@ -51,6 +51,25 @@ import { getSite } from 'calypso/state/sites/selectors';
...
+/**
+ * Determines the truthiness of a site specific property regardless of whether it is on the plugin object
+ * or on one of the plugin's site objects.
+ *
+ * @param {string} prop - The site property to check. One of 'active', 'autoupdate', 'update', or 'version'.
+ * @param {Object} plugin - The plugin object
+ * @param {number} siteId - The ID of the site
+ * @returns {any} The value of the prop being requested
+ */
+const pluginHasTruthySiteProp = ( prop, plugin, siteId ) => {
```
tyxla, line 63: I appreciate addressing the documentation issues, but I think the function name still suffers from the same issue as before. The function name hints to expect a boolean, but it actually returns the prop (if there). Also, the function description still doesn't properly address that problem IMHO - it doesn't make it clear as to what the function returns. Furthermore, if I understand correctly, we seem to be using the function as a boolean (below in this PR), but it will return the prop value which may be of any type, so we rely on implicit type casting to boolean - is there a good reason for that? I think that area of the PR still needs some work.

robertf4: In this case I intended the function to return a boolean, so I have updated it to actually do that.

tyxla (approved):
LGTM, thanks 👍

PR #76685, opened 2023-05-08
Checkout: Convert CheckoutThankYou to TypeScript

sirbrillig:
## Proposed Changes

This PR converts the `CheckoutThankYou` component and several of its dependencies to TypeScript with as few functional changes as possible.

This required one large refactor because the component had a very hard-to-read and highly dynamic method called `getComponentAndPrimaryPurchaseAndDomain()` that TypeScript wouldn't even allow to exist. The method should really be replaced but to simplify this PR, I retained the method's original functionality after removing its component class return value. The wrapper `PurchaseDetailsWrapper` now should handle the actual rendering it used to be used for.

Screenshot 2023-05-10 at 4 03 50 PM

## Testing Instructions

I added some tests but there are a lot of complicated logic patterns in this component that are hard to understand. I have a feeling that much of the code this component contains can be refactored or removed, but this step should simplify doing so.

For example, as far as I can tell we cannot ever render the [generic domain registration details component](https://github.com/Automattic/wp-calypso/blob/4903ab03c1ed33f0b759ae1312b99e3e5547f2e9/client/my-sites/checkout/checkout-thank-you/index.tsx#L947) because it is only rendered [if there is a receipt ID](https://github.com/Automattic/wp-calypso/blob/4903ab03c1ed33f0b759ae1312b99e3e5547f2e9/client/my-sites/checkout/checkout-thank-you/index.tsx#L733) and [a different component is rendered](https://github.com/Automattic/wp-calypso/blob/4903ab03c1ed33f0b759ae1312b99e3e5547f2e9/client/my-sites/checkout/checkout-thank-you/index.tsx#L589) when [a domain registration was purchased](https://github.com/Automattic/wp-calypso/blob/4903ab03c1ed33f0b759ae1312b99e3e5547f2e9/client/my-sites/checkout/checkout-thank-you/index.tsx#L484).

Some checkout URLs to test follow. Each one should look the same before and after this change:

- `/checkout/thank-you/no-site`
- `/thank-you/example.com/unknown-receipt` (where `example.com` should be changed to a real site hostname.)
- The thank-you page for a domain-only purchase (begin the purchase process at `/domains`).
- The thank-you page for a plan (begin the purchase process at `/plans`).

### DIFF
--- a/client/my-sites/checkout/checkout-thank-you/index.tsx
+++ b/client/my-sites/checkout/checkout-thank-you/index.tsx
@@ -26,9 +26,7 @@ import {
 } from '@automattic/calypso-products';
 import { Card } from '@automattic/components';
 import { localize } from 'i18n-calypso';
-import { find, get } from 'lodash';
 import page from 'page';
-import PropTypes from 'prop-types';
 import { Component } from 'react';
 import { connect } from 'react-redux';
 import PlanThankYouCard from 'calypso/blocks/plan-thank-you-card';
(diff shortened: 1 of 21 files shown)


client/my-sites/checkout/checkout-thank-you/index.tsx
```diff
@@ -721,19 +730,9 @@ export class CheckoutThankYou extends Component {
 						/>
 					) }
 
-					{ ! isSimplified && ComponentClass && (
+					{ ! isSimplified && ! this.isGenericReceipt() && (
```
sirbrillig: This refactor may be wrong. I can't recall how I came up with it but this prevents the generic domain registration component from being rendered, amongst other things, I suspect.

sirbrillig: Upon review, I actually think this logic is correct. However, to make the refactor easier to read, I've restored the original behavior by using a string in place of `ComponentClass`.

spraveenitpro:
**Functional Testing:**

<details>
<summary>/checkout/thank-you/no-site page before patch</summary>


</details>

<details>
<summary>/checkout/thank-you/no-site page after patch</summary>


</details>

<details>
<summary>/thank-you/example.com/unknown-receipt page before patch</summary>

 
</details>

<details>
<summary>/thank-you/example.com/unknown-receipt page after patch</summary>

 
</details>

<details>
<summary>The thank-you page for a domain-only purchase before patch</summary>



</details>

<details>
<summary>The thank-you page for a domain-only purchase after patch</summary>


</details>

<details>
<summary>The thank-you page for a plan purchase before patch</summary>




</details>

<details>
<summary>The thank-you page for a plan purchase after patch</summary>


</details>

**Static Analysis:**

I looked through the code and agree that they are lots of dependencies for this specific component, I was able to see that you have typed them as expected. Reminded of concepts from pbOQVh-29D-p2 . LGTM

PR #76697, opened 2023-05-09
Sites page: Add new empty state CTAs

zaguiini:
Related to https://github.com/Automattic/dotcom-forge/issues/2365

## Proposed Changes

This adds a new empty state component to the sites page, replacing the old one (after fetching sites). Screenshot:

page

## Testing Instructions

Make sure that `hosting-onboarding-i2` flag is activated and run this branch on an account with zero sites

1. Check that both CTAs appear and they link to `/setup/new-hosted-site` and `/setup/import-focused`, respectively;
2. Check that hovering/focusing the CTAs activate the sparkling animation, and moving the mouse out of the link removes the sparkles after a while;
3. Check that turning off the flag changes the first CTA to `/start`.


client/sites-dashboard/components/no-sites-message.tsx
```diff
@@ -160,6 +163,33 @@ export const NoSitesMessage = ( { status, statusSiteCount }: SitesContainerProps
 		);
 	}
 
+	if ( isEnabled( 'hosting-onboarding-i2' ) ) {
```
p-jackson: Do we need some other condition here? Like the user just came from the `/start/hosting` flow? Or they have the "is dev" flag set?

zaguiini: I see no problem in this behavior being the default for every user. This CTA provides more options than the existing one. What I can do, though, would be to switch the CTA target for users with the flag disabled.

javierarce:
This is very nice! 

Some comments:

- I'd prefer it if we showed the stars at once. Your proposal is nice, but it's probably too much for this particular use. I added a simple prototype [here](https://www.figma.com/proto/Z71gE7aAlUwy39G8QUVPtr/Hosting-Onboarding?page-id=1%3A63&type=design&node-id=130-7986&scaling=min-zoom&starting-point-node-id=130%3A7986).
- Could we match the position of the stars with the ones in the Figma file?
- For the mobile version, I'd prefer to show the text under the icon to avoid having problems with long strings. 
- Could we rename "Import a site" and say "Migrate a site" instead?

image

zaguiini:
Thanks @javierarce!

About the mobile version: it feels kinda weird to have it vertically because it might culminate in vertical scrolling -- that's why I made it wider. But I have no problems proceeding with your version if you're aware of this trade-off.

Also, did we confirm the top-right corner CTA too? I suggested somewhere to remove the line separator and display the popover with the three options with the thumbnail on click instead of having a primary action on CTA click.

vykes-mac:
This looks nice! I'm with @javierarce comments on the placement of the text. Even though might result in more scrolling the UI feels more balance and a small scroll is not such of a hindrance on touch devices.

zaguiini:
Ok, the CTAs are vertical again on small screens.

What's left:

- top-right CTA
- fixing a bug where the loading skeleton flashes briefly

@javierarce I wonder if we should hide the page header ("Sites"). I thought of keeping it so the user knows that this is a dashboard instead of solely a step in the process.

javierarce:
> This looks nice! I'm with @javierarce comments on the placement of the text. Even though might result in more scrolling the UI feels more balance and a small scroll is not such of a hindrance on touch devices.

We could also reduce the size of the buttons on mobile a little bit to accommodate the options for more screens.

> @javierarce I wonder if we should hide the page header ("Sites"). I thought of keeping it so the user knows that this is a dashboard instead of solely a step in the process.

Hiding the header was also on my design proposal. In my opinion, showing it at this step of the flow is unnecessary and doesn't help the user much. The current solution does show the header, but only because it also includes the button (something we obviously don't need)

image

I updated [the Migrate icon](Z71gE7aAlUwy39G8QUVPtr-fi-104_8977) so the meaning is more clear

image

client/sites-dashboard/components/sites-dashboard-ctas.tsx
```diff
@@ -0,0 +1,38 @@
...
+	CREATE_SITE_ICON_SPARKLES,
+	MigrateSiteIcon,
+	MIGRATE_SITE_ICON_SPARKLES,
+} from './sites-dashboard-cta-icons';
+import { SparklingCTA } from './sparkling-cta';
+
+export const CreateSiteCTA = () => {
+	const { __ } = useI18n();
+
+	const newSiteURL = isEnabled( 'hosting-onboarding-i2' ) ? '/setup/new-hosted-site' : '/start';
```
p-jackson: I think I have the same concern about _only_ using the feature flag to decide which site creation flow to use. When we launch this feature, `hosting-onboarding-i2` turns `true`, and all users will be seeing our `new-hosted-site` flow.

That means any regular user, who deletes all their sites to start again, if they click this button they'll only be offered the Business and Commerce plan and be taken straight to WoA. These users may just want to create a free blog.

I think we want this button to go to `/setup/new-hosted-site` only if (1) the feature is enabled, and (2) they've just come from the `/hosting` LOHP. Otherwise we have no hint that they're interested in only WoA sites.

zaguiini:
Putting this PR on hold since there are some design unknowns: p1683728703732539/1683636035.909399-slack-C9EJ7KSGH

mpkelly:
>Putting this PR on hold since there are some design unknowns

I gave it a test anyway. One thing I noticed though that we might want to address in the final PR is that when you click the WordPress logo from `/sites` you will see this:

image

It feels weird to only have an option to create a site and not import one and see a different illustration. So whatever final solution we use I think we need to apply it to this page too.

zaguiini:
Oh, thanks @mpkelly. This looks like a different issue, though... Why is that clicking on W leads to `/stats/day`?

mpkelly:
>Oh, thanks @mpkelly. This looks like a different issue, though... Why is that clicking on W leads to /stats/day?

It should probably not link anywhere when you're already on the `/sites` page. Is that what you're thinking? If so, I can change that in a separate PR. 

But we use `/stats/day` because there are no sites and that's one of the pages that still exists.

zaguiini:
>It should probably not link anywhere when you're already on the /sites page. Is that what you're thinking

Yeah, it should have no effect if `/sites` is their default page. If it is not, it should send them to wordpress.com

>If so, I can change that in a separate PR.

Yes please 👍

javierarce:
Here's a new proposal for this screen based on this new design (CZ07QdsC7bmWbLiSO5K9T5-fi-1389_59681) of the blog onboarding:

image

@saygunnyc could you help us review it so we can move on with this task? I'd like to avoid reusing the same label for the buttons and the titles, I'm having difficulty thinking of a better alternative except maybe this:

image

saygunnyc:
Noting we don't have to use all elements (title, description, icon), esp. if they become redundant. It's possible to remove descriptions if that helps. cc @michaelpick for help regarding copies.

michaelpick:
Hey, @saygunnyc I really like the approach of the earlier, simple, visual iterations of this upstream! The copy is tighter with the later ones, though, as I think we had some tangle between 'import' and 'migrate' which felt like a bit of a mismatch here (import for both makes more sense).

If we're rolling with design parity with blog, newsletter etc. flows, I'd say the biggest content improvement would be cutting the current duplication of headline and CTA. Rather than trying to come up with synonyms for headings, it would make a lot more sense to remove the current headings and/or replace them with the supporting copy descriptions.

- Page H1: 'Let's add your first site'
  - Feels in keeping with other surfaces 👍 
 
- Description: Build a new site from scratch 
  - Adds additional information and brings value 👍 
  
- Description: Bring an existing site to WordPress.com
  - Adds additional information and brings value 👍
 
- CTA: Create a site 👍 
  - Simple and clear
- CTA: Import a site 👍 
  - Unidirectional language. Feels more intent-aligned and clearer than 'migrate', which is omnidirectional.

I'd lose the current h2s, which are duplicative whichever way we try to reword:
- Create a site / New site 👎 
- Migrate a site / Existing site 👎 

Either by bumping the descriptions to h2s, incorporating earlier design exploration of clear visual interaction (those sparkles are really nice!), or by placing the descriptions under the buttons, for instance. I'm sure there are other solutions!

zaguiini:
@javierarce this is how it looks now:

| Desktop | Mobile |
| -------- | ------- |
| image | image |

javierarce:
> @javierarce this is how it looks now:

Thanks, @zaguiini! Based on @michaelpick & @saygunnyc feedback (thanks for your feedback!) and to avoid being redundant, we could adjust the screen and replace the secondary titles with the descriptions like this:

image

It's not the most attractive screen right now, but most users will only see it once, and we can revisit it in the future.

[Figma file](https://www.figma.com/file/Z71gE7aAlUwy39G8QUVPtr/Hosting-Onboarding?type=design&node-id=1-63)

saygunnyc:
> and we can revisit it in the future

Agreed!

saygunnyc:
> Hey, @saygunnyc, I really like the approach of the earlier, simple, visual iterations of this upstream! 

I also like it being visual, however, currently, illustrations are tricky. 😅  As @javierarce said, we could revisit and give these a bit more of a lift, once if we lean towards utilizing illustrations again. Noting we'll be re-using this on other flows as needed as well.

mpkelly:
> > It should probably not link anywhere when you're already on the /sites page. Is that what you're thinking
> 
> Yeah, it should have no effect if `/sites` is their default page. If it is not, it should send them to wordpress.com
> 
> > If so, I can change that in a separate PR.
> 
> Yes please 👍

Deployed https://github.com/Automattic/wp-calypso/pull/77240

zaguiini:
Ok, cool. This is how it looks now:

image

I'm merging this PR as it is, and we can revisit it in the future as needed.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8010512

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8010512).

PR #76699, opened 2023-05-09
Backup: Add render switch to validate multisites and backup feature on backup page

Initsogar:
If a customer is trying to access the backup page directly in Calypso using a multisite with backup capabilities, they are getting a message saying "WordPress multi-sites are not supported".

This is because when accessing directly the page, we are not querying site features before. So the idea of this PR is to add a new switch component to validate if a site is part of a Jetpack multisite that can be used as part of the [showUnavailableForMultisites()](https://github.com/Automattic/wp-calypso/blob/0650cc21ec42c79e8c1536a4966d60976d0cf5a4/client/my-sites/backup/controller.js#L97) middleware.
## Proposed Changes

* Add new switch component to validate if a site is part of a Jetpack multisite
* Use new switch component on `showUnavailableForMultisites` backup controller middleware.

## Testing Instructions

First, we should grab a site that is part of a multisite and also has (or not) the backup feature. You can test this PR in your local environment by returning `true` on [`isJetpackSiteMultiSite` selector](https://github.com/Automattic/wp-calypso/blob/11e5d35381f271cf1c88d7af1f8d45cb820c1dc5/client/state/sites/selectors/is-jetpack-site-multi-site.js#L12) to simulate the multisite condition. 

Then navigate directly to http://jetpack.cloud.localhost:3000/backup/{site_slug}.

If your site has no backup feature, you should see the following message:

| Jetpack Cloud | Calypso Blue |
|---|---|
| image | image |

If your site has the backup feature, you should see the backup section as usual:

image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/backup/controller.js
+++ b/client/my-sites/backup/controller.js
@@ -105,11 +105,7 @@ export function showUnavailableForMultisites( context, next ) {
 	);
 
 	context.primary = (
-		<MultisiteNoBackupPlanSwitch
-			trueComponent={ message }
-			falseComponent={ context.primary }
-			loadingComponent={ <p>Loading..</p> }
-		/>
+		<MultisiteNoBackupPlanSwitch trueComponent={ message } falseComponent={ context.primary } />
 	);
 
 	next();
--- a/client/my-sites/backup/controller.js
+++ b/client/my-sites/backup/controller.js
@@ -14,8 +14,8 @@ import getRewindState from 'calypso/state/selectors/get-rewind-state';
 import getSelectedSiteId from 'calypso/state/ui/selectors/get-selected-site-id';
 import BackupUpsell from './backup-upsell';
 import BackupCloneFlow from './clone-flow';
-import IsMultiSiteSwitch from './is-multisite-switch';
 import BackupsPage from './main';
+import MultisiteNoBackupPlanSwitch from './multisite-no-backup-plan-switch';
 import BackupRewindFlow, { RewindFlowPurpose } from './rewind-flow';
 import WPCOMBackupUpsell from './wpcom-backup-upsell';
 import WpcomBackupUpsellPlaceholder from './wpcom-backup-upsell-placeholder';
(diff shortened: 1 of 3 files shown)


client/my-sites/backup/controller.js
```diff
@@ -97,21 +96,21 @@ export function showUnavailableForVaultPressSites( context, next ) {
...
-			<BackupUpsell reason="multisite_not_supported" />
-		) : (
-			<WPCOMBackupUpsell reason="multisite_not_supported" />
-		);
-	}
+	context.primary = (
+		<IsMultiSiteSwitch
+			trueComponent={ message }
+			falseComponent={ context.primary }
+			loadingComponent={ <p>Loading..</p> }
```
Initsogar: Let's use a loading placeholder instead.

client/my-sites/backup/is-multisite-switch.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { isJetpackSiteMultiSite } from 'calypso/state/sites/selectors';
+import getSelectedSiteId from 'calypso/state/ui/selectors/get-selected-site-id';
+
+type Props = {
+	trueComponent: ReactNode;
+	falseComponent: ReactNode;
+	loadingComponent?: ReactNode;
+};
+
+const IsMultiSiteSwitch: FunctionComponent< Props > = ( {
```
jwebbdev: It's a bit wordy, but I wonder if something like `IsMultiSiteNoPlanSwitch` or similar might be a less misleading name.

mavegaf (approved):
Tested and works as expected 👍

client/my-sites/backup/multisite-no-backup-plan-switch.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const MultisiteNoBackupPlanSwitch: FunctionComponent< Props > = ( {
+	trueComponent,
+	falseComponent,
+	loadingComponent,
+} ) => {
+	const siteId = useSelector( getSelectedSiteId );
+
+	const isMultiSite =
+		useSelector( ( state ) => siteId && isJetpackSiteMultiSite( state, siteId ) ) || false;
```
macbre, line 26:
```suggestion
		useSelector( ( state ) => siteId && isJetpackSiteMultiSite( state, siteId ) );
```

A left-over from debugging?

macbre:
@Initsogar - thanks for your help here!

client/my-sites/backup/multisite-no-backup-plan-switch.tsx
Initsogar: @macbre The `isJetpackSiteMultiSite` selector can also be `null`, but for this use case it would be better to always return a boolean, so in case it is null I'm returning false, so we could use it for the `renderCondition` prop in `RenderSwitch`. Maybe there's a better way to handle it in TS? Thoughts?

macbre: Oh, then this code does make sense.

PR #75542, opened 2023-04-11
Hosting onboarding: enable money back guarantee copy below CTA at checkout

zaguiini:
Related to https://github.com/Automattic/dotcom-forge/issues/2121.

## Proposed Changes

The checkout step for folks from the `/start/hosting` flow should display the money-back guarantee copy below the CTA.

It should consider the frequency (monthly, yearly) of the plan that was chosen.

image

## Testing Instructions

1. Checkout this branch;
2. Start the flow at `/start/user` and chose a paid plan;
3. Check that the copy does not show up below the CTA;
4. Close the checkout (empty cart) and start the hosting flow at `/start/hosting`;
5. Chose a paid plan and proceed to checkout;
6. Verify that the copy DOES show up below the CTA.

PR #76717, opened 2023-05-09
Checkout: display 14-day refund copy below CTA for biennially and triennially products

zaguiini:
Related to https://github.com/Automattic/wp-calypso/pull/75542. See p1683625556906589-slack-C04H4NY6STW.

## Proposed Changes

If the user is buying a product with a multi-year tenure, we should display only the 14-day refund copy.

| Before | After |
| ------- | ----- |
| image | image |

## Testing Instructions

1. Go to `/start/hosting`;
2. Choose a monthly plan;
3. Proceed to checkout;
4. Verify that the "7 day money back guarantee" copy shows up below the checkout CTA;
5. Change to a yearly plan;
6. Verify that the "14 day money back guarantee" copy shows up below the checkout CTA;
7. Change to biyearly, then triyearly plan;
8. Verify that you still see the "7 day money back guarantee" copy below the checkout CTA.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -1,4 +1,10 @@
-import { isYearly, isJetpackPurchasableItem, isMonthlyProduct } from '@automattic/calypso-products';
+import {
+	isYearly,
+	isJetpackPurchasableItem,
+	isMonthlyProduct,
+	isBiennially,
+	isTriennially,
+} from '@automattic/calypso-products';
 import { Gridicon } from '@automattic/components';
 import {
 	MainContentWrapper,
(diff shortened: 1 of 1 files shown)


michaeldcain (approved):
This works as described and looks good to me. Thanks for fixing it!

niranjan-uma-shankar:
Thanks @zaguiini, but I think there are still a couple of scenarios to address.

Firstly, a somewhat minor issue (since a lower volume of users will run into this), if the cart has only a one-time purchase product (non-subscription product), then we show both the 7-day and 14-day refund copies. Example:

wordpress com_checkout_rhbfgherjvferh462378467382 wordpress com

Secondly, again somewhat minor by nature of its rarity, is that when the cart has a 2-yearly plan and a monthly product, say a 2-year Business plan and a monthly Yoast plugin, then the refund text would say - 
`14 day money back guarantee on yearly subscriptions`
`7 day money back guarantee on monthly subscriptions`

The only contention I have with this is that there is no yearly subscription in the cart, so it's not clear (at least to me) if the 14-day refund applies to the bi-yearly product.

Finally, **a more critical issue** that I think a majority of our users would run into, is that we claim a 14-day moneyback guarantee if the cart has a domain product. Domains refunds are eligible only within 96 hours. While we do mention the 96-hour policy in the checkout fine print, a big bold 14-day moneyback guarantee is more likely to catch someone's attention and we might end up fielding dissatisfied customers in support. The effect is even more pronounced when there's _only_ a domain product in cart and the copy still suggests a 14-day moneyback offer - there is no confusing the aspect that this applies to the only item in cart but that doesn't represent our policy. Since so many of our paid users buy domains in the signup flow, we should maybe address this with higher priority, or hide the copy temporarily till we are able to arrive at a good solution. 

wordpress com_checkout_ev23ghcee23r23rgv23ghrc2jh3g wordpress com_signup=1

PR #76737, opened 2023-05-09
Earn Page: Fix toggle under warning 

lezama:
> There is something weird there. Clicking behind the info checks the checkbox in the other panel? Or is there a duplicated checkbox that is lying there by mistake?

https://github.com/Automattic/wp-calypso/assets/104869/d1f19000-e934-49ef-9a95-1cee51b173f3

### DIFF
--- a/client/my-sites/earn/memberships/add-edit-plan-modal.jsx
+++ b/client/my-sites/earn/memberships/add-edit-plan-modal.jsx
@@ -1,7 +1,6 @@
 import { Dialog, FormInputValidation } from '@automattic/components';
 import formatCurrency from '@automattic/format-currency';
 import { ToggleControl } from '@wordpress/components';
-import classnames from 'classnames';
 import { useTranslate } from 'i18n-calypso';
 import { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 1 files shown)


millerf:
The change is size when switching tab is a bit unsettling...

https://github.com/Automattic/wp-calypso/assets/790558/093da761-a5de-47bb-8daf-690a4f0ec65a

client/my-sites/earn/memberships/add-edit-plan-modal.jsx
```diff
@@ -414,19 +413,8 @@ const RecurringPaymentsPlanAddEditModal = ( {
...
-				>
-					{ renderGeneralTab() }
-				</div>
-				<div
-					className={ classnames( 'memberships__dialog-section', {
-						'is-visible': currentDialogTab === TAB_EMAIL,
-					} ) }
-				>
-					{ renderEmailTab() }
+				<div className="memberships__dialog-section is-visible">
```
simison, line 416: The CSS hiding seems hacky and root of some of the problems, I think it's best to remove `is-visible` and untangle this:

https://github.com/Automattic/wp-calypso/blob/afed935719c5c68a2bf5f756386379dedc2d967d/client/my-sites/earn/memberships/style.scss#L316-L318

https://github.com/Automattic/wp-calypso/blob/afed935719c5c68a2bf5f756386379dedc2d967d/client/assets/stylesheets/shared/functions/_z-index.scss#L230

They are there to ensure section is always as high as taller section of the two, as noted here:

https://github.com/Automattic/wp-calypso/blob/afed935719c5c68a2bf5f756386379dedc2d967d/client/my-sites/earn/memberships/style.scss#L312-L313

This causes two problems however:

- Click through actions (can be fixed with `pointer-events: none;`)
- Elements like tooltips, notices, modals etc have their own z-index which collides with z-indexes used here. Could be fixed by adjusting z-indices but it's fragile.

millerf: > I think it's best to remove is-visible and untangle this

Yes.

> The CSS hiding seems hacky 

Unfortunately, that is pretty much how Earn was made, apparently. We have a lot of work to bring every up to a certain standard...

edanzer: Hey @lezama @simison @millerf - I took a look at this. There seems to be a bit of unusual css around these tabs. I found what I think could be a very simple solution. This solution doesn't wrestle with any of the underlying complexities here, so it's not a 'thorough' solution. It just bumps the z-index from 1 to 2.

But best I can tell, this works, is simple, and should not affect anything else:
https://github.com/Automattic/wp-calypso/pull/76840/files

edanzer:
Got approvals on the alternative PR, so closing this in favor of https://github.com/Automattic/wp-calypso/pull/76840

PR #74883, opened 2023-03-24
Marketplace Thank You: Query data for Themes

epeicher:
Resolves #74878 

## Proposed Changes

* Query data read by selectors. This includes:
  * Products list
  * Purchases
  * Active Theme
* Instead of hiding an invalid theme. Display the theme with the action buttons disabled. This minimizes the layout shifts when loading the products and purchases for valid themes.

## Testing Instructions

* Navigate to the Marketplace Thank You page using this URL: `/marketplace/thank-you/:siteId?themes=:themeId`
* Using a `:siteId` that is Business or above and a `:themeId` that is free or you have purchased for that `:siteId`
* Activate the Theme
* Open the Developer Tools > Application and delete the IndexedDB
* Refresh the page
* Make sure you can see the buttons `Customize this design` meaning the Theme is considered `Active`
* Change the URL and add a theme that you have not purchased, e.g. `/marketplace/thank-you/:siteId?themes=lincoln`
* Make sure the CTA buttons for the Theme are disabled

### Regression testing
- To test the small update in `QueryProductsList` component. 
- Navigate to Themes and do some queries
- Make sure you do not see unexpected errors in the console

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76741, opened 2023-05-09
Jetpack Pro Dashboard: Add a hardcoded parameter to fetch prices in USD

heydemoura:
## Proposed Changes

* Fix agency prices showing up as their store setting currency instead of USD only.
* The absolute amount of the prices on the pro dashboard pricing page would vary according to your wordpress.com account currency, it should only show amounts from USD prices.

## Testing Instructions
- Change the currency of your wordpress.com account to anything other than USD
- Login to the Jetpack Pro Dashboard - If you don't have an agency account, request one from Avalon.
- On the Pro Dashboard, go to Licensing > Prices. You should only see prices in USD.

| Before  | After |
| ------------- | ------------- |
| image | image  |

## Pre-merge Checklist

### DIFF
--- a/client/components/data/query-products-list/index.jsx
+++ b/client/components/data/query-products-list/index.jsx
@@ -16,13 +16,13 @@ const request =
 		dispatch( requestProductsList( props ) );
 	};
 
-export function useQueryProductsList( { type = 'all', persist } = {} ) {
+export function useQueryProductsList( { type = 'all', currency, persist } = {} ) {
 	const dispatch = useDispatch();
 
 	// Only runs on mount.
 	useEffect( () => {
-		dispatch( request( { type, persist } ) );
-	}, [ dispatch, type, persist ] );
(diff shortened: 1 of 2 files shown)


rcoll:
This seems to fix the problem as intended, but tagging @Automattic/jetpack-genesis for a review because I'm unsure if `useQueryProductsList` could be negatively impacted in any other use cases.

vitozev (approved):
LGTM! As we don't provide a default value for the `currency`, it will be always `undefined` by default, and also should be emitted when [we stringify the object here](https://github.com/Automattic/wp-calypso/blob/trunk/packages/wpcom.js/src/lib/util/send-request.js#L64-L65). I think this shouldn't cause any regression.

I've manually tested [one flow](https://github.com/Automattic/wp-calypso/pull/74883) that involves direct call of the `useQueryProductsList()` and it works as expected.

The changes in the Jetpack Cloud are also working as expected.

Issue #76743, opened 2023-05-10
Reader: Public tag pages sticky menu positioning


ebinnion:
### Quick summary

Occasionally, when scrolling down and back up quickly, I am able to get the sticky menu with characters to stick incorrectly to the top of the page.

<img width="1119" alt="Screenshot 2023-05-09 at 7 40 33 PM" src="https://github.com/Automattic/wp-calypso/assets/1126811/5d9a8430-2036-4f4f-8cbe-99331a404233">

https://github.com/Automattic/wp-calypso/assets/1126811/ec16d0b3-b53e-4b66-a025-5fc151793b2d

### Steps to reproduce

1. Start at https://wordpress.com/tags, logged in or out.
2. Scroll down page enough to get the sticky menu to show then back up. Quickly.
3. Observe that the sticky menu occasionally sticky at the very top.

### What you expected to happen

The sticky menu to stay just above the "A" section when I scrolled to the top of the page.

### What actually happened

The sticky menu got stuck to the top of the viewport.

### Impact

All

### Available workarounds?

Yes, easy to implement

### Logs or notes

This was tested on Mac OS with Microsoft Edge browser, logged in and out.

karenroldan:
📌 REPRODUCTION RESULTS

- I was able to reproduce the issue. 

📌 FINDINGS/SCREENSHOTS/VIDEO

- I tested with Microsoft Edge, Firefox, Safari, and Chrome.

![reader](https://github.com/Automattic/wp-calypso/assets/67279475/e71c3a2d-71ed-489a-80fa-c35a9f08f43b)


📌 ACTIONS

- Assigned to @Automattic/loop

roo2:
updated https://github.com/Automattic/wp-calypso/pull/76747 with a fix. The issue was happening because of scrolling very quickly over the sticky panel's starting location, it is easiest to reproduce with a very short browser window

jordesign:
No longer relevant with the current Reader design - so closing this.

PR #76747, opened 2023-05-10
Extend stickypanel to support mobile and absolute position

roo2:
Fixes #https://github.com/Automattic/wp-calypso/issues/76743 by modifying the `<StickyPanel>` component to always use the `scroll` event rather than relying on the IntersectionObserver. see also slack discusion p1683679396865339-slack-C03NLNTPZ2T

The issue with using the IntersectionObserver for sticky panels seems to be fundamental [Stackoverflow link](https://stackoverflow.com/questions/61951380/intersection-observer-fails-sometimes-when-i-scroll-fast ). Basically, if you scroll very quickly over the starting location of the panel, the IntersectionObserver won't report any events, and the panel will stay shown.

The IntersectionObserver was only added for performance reasons, so we can simply fall back to the window.scroll implementation.

I also extended the sticky panel to use the width of the original element for the spacer, this is still needed to fix alignment for absolute positioned elements (the mobile view of the /tags page)

I also enabled the sticky panel component to work with mobile devices if minWidth=0 is passed in.

### Testing instructions

* go to the /tags page on mobile and desktop, the panel should stick as before and the bug in https://github.com/Automattic/wp-calypso/issues/76743 should no longer be reproducable.
* Go to the media gallery page, the stickypanel should still work on desktop (not mobile) /media/$site_slug

### DIFF
--- a/client/components/sticky-panel/index.jsx
+++ b/client/components/sticky-panel/index.jsx
@@ -63,12 +63,18 @@ function renderStickyPanel( props, state ) {
 			<div className="sticky-panel__content" style={ getBlockStyle( state ) }>
 				{ props.children }
 			</div>
-			<div className="sticky-panel__spacer" style={ { height: state.spacerHeight } } />
+			<div
+				className="sticky-panel__spacer"
+				style={ { height: state.spacerHeight, width: state.blockWidth } }
+			/>
 		</div>
 	);
 }
 
 function isWindowTooSmall( minLimit ) {
+	if ( minLimit === 0 ) {
+		return false;
+	}
 	return ( minLimit !== false && minLimit >= window.innerWidth ) || isMobile();
 }
 
--- a/client/components/sticky-panel/index.jsx
+++ b/client/components/sticky-panel/index.jsx
@@ -1,6 +1,6 @@
 import { isMobile } from '@automattic/viewport';
 import classNames from 'classnames';
-import { throttle, defer } from 'lodash';
+import { throttle } from 'lodash';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
 import ReactDom from 'react-dom';
@@ -9,8 +9,6 @@ import afterLayoutFlush from 'calypso/lib/after-layout-flush';
 import './style.scss';
 
 const RESIZE_RATE_IN_MS = 200;
-
-const hasIntersectionObserver = typeof window !== 'undefined' && 'IntersectionObserver' in window;
 
 const commonPropTypes = {
 	minLimit: PropTypes.oneOfType( [ PropTypes.bool, PropTypes.number ] ),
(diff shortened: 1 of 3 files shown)


client/components/sticky-panel/index.jsx
```diff
@@ -63,12 +63,18 @@ function renderStickyPanel( props, state ) {
...
+			<div
+				className="sticky-panel__spacer"
+				style={ { height: state.spacerHeight, width: state.blockWidth } }
+			/>
 		</div>
 	);
 }
 
 function isWindowTooSmall( minLimit ) {
+	if ( minLimit === 0 ) {
```
roo2, line 76: if minLimit has explicitly been set to 0, we can ignore the isMobile check

client/reader/tags/alphabetic-tags.tsx
```diff
@@ -89,24 +89,8 @@ const scrollToLetter = ( letter: string ) => {
 
 export default function AlphabeticTags( { alphabeticTags }: AlphabeticTagsProps ) {
 	const translate = useTranslate();
-	const [ isSticky, setIsSticky ] = useState( false );
 	const tagsTableRef = createRef< HTMLDivElement >();
 
-	useEffect( () => {
```
roo2, line 95: no longer neccessary now that the panel is fixed 🎉

Addison-Stavlo (commented):
Im testing this build on the calypso live link and am still seeing the issue noted in https://github.com/Automattic/wp-calypso/issues/76743

sticky-wierd

roo2:
oh wow, let's take another look, I just realized this may have been a completely seperate issue to the one I fixed and it just happened to not reproduce for me after my fix was applied 😂

Addison-Stavlo (commented):
The fix for the stickiness is working well in testing.

Im wondering about the performance that was originally saved here. 🤔 

If we wanted to keep the more performant option with intersectionOberserver we could probably keep that component but add an extra safety: a function on scroll that is debounced by a half a second or something.  That way we keep the perf, the interaction is smooth with normal interactions with the interactionObserver, and we have an extra function that barely fires whenever scroll stops just as a safety to ensure the component is where it should be.

Addison-Stavlo (commented):
The fix for the stickiness is working well in testing.

Im wondering about the performance that was originally saved here. 🤔 

If we wanted to keep the more performant option with intersectionOberserver we could probably keep that component but add an extra safety: a function on scroll that is debounced by a half a second or something.  That way we keep the perf, the interaction is smooth with normal interactions with the interactionObserver, and we have an extra function that barely fires whenever scroll stops just as a safety to ensure the component is where it should be.

roo2:
ok, that would be even better 👍

roo2:
updated to use scroll as a backup throttled to 500ms as suggested

Addison-Stavlo (approved):
This code looks good and its definitely an improvement!

I don't want to hold up the improvement any longer, so approving at this point. However, I think if we use `debounce` instead of `throttle` we could reduce the timer (and this visible delay) substantially:  

scroll-bar-thing

Since throttle still triggers on the interval of the timer, reducing the delay to a really low value would kind of negate the perf reasons of using it in the first place. But debounce will hold off entirely until the scroll event itsn't being triggered, so we could probably get away with a much lower delay there.

roo2:
ok, I think this sticky panel will be perfect now 👌😄

PR #76754, opened 2023-05-10
Add-ons: Create Storage Add On Card Item

sharpirate:
Fixes [1722-gh-Automattic/martech
](1722-gh-Automattic/martech)
### Proposed Changes
Create the card item for the storage add on

### Testing Instructions
1. Checkout this branch
2. Run `yarn` & `yarn start`
3. Go to `http://calypso.localhost:3000/add-ons/YOUR_SITE_SLUG`
4. You should see 2 new cards items - a 50 GB and a 100 GB storage add ons:

image

5. Click on the **Buy add-on** button. You should get redirected to the cart with the correct storage amount applied:

image


a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8005670

Thank you @Sharpirate for including a screenshot in the description! This is really helpful for our translators.

gmovr:
Thanks!

I've tested this manually for now and noticed one thing (when trying with the filtering from https://github.com/Automattic/wp-calypso/pull/76878 after rebasing on top of the latest commit in trunk)

If I already own 50 GB, there's no way to buy 100 GB (it's stuck on the Manage add-on button). Ideally, if a user already has 50 GB, we should select 100 by default, or at least check if they can buy 100, and then if they choose that, go ahead and change the button to the purchase link). 

This is a screenshot from a site with 50 GB of added space (Free plan). With the 53 GB total, they should be able to upgrade to 100 GB. 

Screenshot 2023-05-24 at 14 19 53

gmovr:
We also need to update the icon :)

@saygunnyc -- Do you have an icon we can use for the space add-on?

saygunnyc:
I have designed the add-ons on this Figma – dyXNnR0oHu8H47eUES8ktJ-fi-91%3A4553 (including an icon 😄) and shared them on this P2 – p9Jlb4-7p0-p2.

Add-ons


Is there a reason for grouping the 50 GB and 100 GB add-ons in a single add-on? Having them separate would be better for visibility on both options. It is hierarchically weird to have a title called 100 GB on the add-on and a dropdown within with 50 GB option.

gmovr:
> I have designed the add-ons on this Figma – [dyXNnR0oHu8H47eUES8ktJ-fi-91%3A4553](https://href.li/?https://www.figma.com/file/dyXNnR0oHu8H47eUES8ktJ/?node-id=91%3A4553) (including an icon 😄) and shared them on this P2 – [p9Jlb4-7p0-p2](https://href.li/?https://wp.me/p9Jlb4-7p0).
> 
> Add-ons
> Is there a reason for grouping the 50 GB and 100 GB add-ons in a single add-on? Having them separate would be better for visibility on both options. It is hierarchically weird to have a title called 100 GB on the add-on and a dropdown within with 50 GB option.

Thanks! I missed that. I always get lost in Figma, so I didn't see that view :D

client/my-sites/add-ons/hooks/use-add-ons.ts
```diff
@@ -42,13 +49,39 @@ const useAddOns = (): ( AddOnMeta | null )[] => {
...
+	if ( ! isStorageAddonAvailable() ) {
+		addOnsActive.pop();
+	}
 
 	return useSelector( ( state ): ( AddOnMeta | null )[] => {
 		return addOnsActive.map( ( addOn ) => {
 			const product = getProductBySlug( state, addOn.productSlug );
-			const name = getProductName( state, addOn.productSlug );
-			const description = getProductDescription( state, addOn.productSlug );
+			const name = addOn.name ? addOn.name : getProductName( state, addOn.productSlug );
```
sharpirate, line 84: I've added a name override because we want two different names ("50 GB Storage" and "100 GB Storage") for the same product slug (PRODUCT_1GB_SPACE). Since we're differentiating the two cards based on quantity and not based on different product slugs, we can't use the `getProductName` in this case

sharpirate:
@gmovr @saygunnyc Hey, I wanted to double-check on some minor design considerations. There's a slight difference in the formatting of the pricing between the old add-ons and the new storage add-ons in the current design:

Old: **$2 monthly, billed yearly**
New: **$50/month, billed yearly**
image

I would imagine want to stick to one formatting for the sake of consistency, I was wondering which one we should use?

Another small thing is, in the design we only have the **$** sign and not the currency code **US$**:

Design:
image

Production:
image

Wanted to make sure we want to keep using the currency code as we do now on production?

gmovr (approved):
Works well! I tested to toggle off the flag in `development.json` (to mimic how it is in prod), and the cards are not visible. And the previous add-ons are not affected (checkout links work). 

I notice some things that are not blockers:
- Warning in the console about encountering two children with the same key. I see that you've fixed this in https://github.com/Automattic/wp-calypso/pull/77387
- If I purchase one add-on, e.g. 50 GB, both 50 GB and 100 GB will shown as if they were purchased, with the "manage add-on" button. I think this will be fixed with the filtering in https://github.com/Automattic/wp-calypso/pull/77387 as well. 

Lookas good to deploy as it's behind a feature flag, and it's not breaking anything. We can do any design follow-ups (and other follow-ups) in separate PRs.

saygunnyc:
Good eye, @Sharpirate !

Let's keep the existing formatting. $X monthly, billed Xly.

> Wanted to make sure we want to keep using the currency code as we do now on production?''

We don't use US$ on the plans grid; I'd follow that. I believe we only add the additional indicator –M, CA on non-US $ instances.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8005670).

saygunnyc:
After reviewing the CFT pdtkmj-1tb-p2 and this comment, pdtkmj-1tb-p2#comment-2659, I agree it's not visible on the front-end that storage options are not stackable, and it becomes confusing for the user. They might easily think they purchased 150 GB storage, but they upgraded their 50 GB to 100 GB.

As it will always be an upgrade on the existing storage option after the first purchase, I modified to be a better fit for upgrading functionality, re-introducing the dropdown.

**A-First purchase – Multiple options**
This is what the user sees before any storage-related transaction if they have multiple storage options to choose from.

Add-ons_First


**B-After purchase**
This is what the user sees after purchasing a storage option, where more options are available.

Add-ons_AfterPurchase

**C-Changing dropdown**
This is what happens once the user selects a higher storage option. The total price changes above, and the CTA changes to Upgrade.

Add-ons_AfterPurchase_Upgrade

**A.1-First purchase – Single option**
If there's a single option (depending on users' eligibility), then this is how the add-on will look like.

Firsttime_Singleoption


**D-After purchase – Single option or After the best option is purchased**
No upgrade, downgrade, only manage

After_Toptierpurchased



This should remove the risk of confusion. Please let me know what you think.
And here is a link to the Figma file: dyXNnR0oHu8H47eUES8ktJ-fi-111%3A5863

PR #76722, opened 2023-05-09
Stepper: hide plans in the plans step according to the current flow

zaguiini:
Related to https://github.com/Automattic/wp-calypso/pull/76657.

## Proposed Changes

Let's use the `hideXPlan` prop to hide the plans when using the stepper. Using it fixes the [comparison table](https://github.com/Automattic/wp-calypso/pull/76657#pullrequestreview-1418592479), just like it did in the `/start/hosting` flow.

Screenshot from the Link-in-Bio flow:

image

It's actually possible to display the plan comparison table now:

image

## Testing Instructions

1. Checkout this branch;
2. Go to `/setup/link-in-bio`;
3. Proceed to the plans step;
4. Check that the plan comparison is hidden;
5. Comment [this line](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/link-in-bio.ts#L53);
6. Check that the plan comparison shows up and correctly renders the features according to the displayed plans.

PR #76557, opened 2023-05-03
Import focused flow: Skip preview step if it comes from move to wpcom plugin

bogiii:
Closes #76217

## Proposed Changes

* Skip the preview step if it comes from the "Move to WordPress.com" plugin

## Testing Instructions

* Create a `jurassic.ninja` website with the `Move to WordPress.com` plugin installed
* Press the `Get started` button
* Establish Jetpack connection
* Check if you ended up with Import type chooser screen

Before this change, it was a `Preview` screen and then an `Import type chooser`

Screenshot 2023-05-03 at 22 39 50

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #76763, opened 2023-05-10
Fix: planType variable breaks features in 2023 Plans Comparison Grid

claudiucelfilip:
Related to #76557 

## Proposed Changes

This is an alternative fix to #76722 by addressing adjustin how `<PlansFeaturesMain>` handles filtered plans.
The plans are filtered using the `planTypes` based on flows [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx#L51). 
In the case of the 2023 Plans Grid, the features are compounded from plan to plan, meaning that if a a lower-tie plan is missing, the higher plans don't inherit its features (resulting into [this issue](https://github.com/Automattic/wp-calypso/pull/76657#pullrequestreview-1418592479)).

The proposed fix ensures the `plans` variable always gets the default `planTypes` whereas the `visiblePlans` respect the `planTypes` property. These changes are only applied to the 2023 Plans Grid.

## Testing Instructions 

You can follow the steps from #76722 :
1. Checkout this branch;
2. Go to `/setup/link-in-bio`;
3. Proceed to the plans step;
4. Check that the plan comparison is hidden;
5. Comment [this line](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/link-in-bio.ts#L53);
6. Check that the plan comparison shows up and correctly renders the features according to the displayed plans.

7. Also double-check for regressions on `/start/plans` and `/plans/{SITE_ID}`

### DIFF
--- a/client/my-sites/plans-features-main/index.jsx
+++ b/client/my-sites/plans-features-main/index.jsx
@@ -622,9 +622,21 @@ export class PlansFeaturesMain extends Component {
 		 * we pass `visiblePlans` to its `plans` prop.
 		 */
 		const term = this.getPlanBillingPeriod( intervalType, getPlan( selectedPlan )?.term );
-		const planTypes = this.props.planTypes || this.getDefaultPlanTypes();
-		const plans = this.getPlansFromTypes( planTypes, GROUP_WPCOM, term );
-		const visiblePlans = this.getVisiblePlansForPlanFeatures( plans );
+		const defaultPlanTypes = this.getDefaultPlanTypes();
+		const planTypes = this.props.planTypes || defaultPlanTypes;
+		let plans = this.getPlansFromTypes( planTypes, GROUP_WPCOM, term );
+		const filteredPlans = plans;
+
+		/*
+		 * We need to keep all the plans in the plans variable,
+		 * The filtered planTypes should be reflected in visible plans only.
+		 */
+		if ( is2023PricingGridVisible ) {
+			plans = this.getPlansFromTypes( defaultPlanTypes, GROUP_WPCOM, term );
+		}
+
+		const visiblePlans = this.getVisiblePlansForPlanFeatures( filteredPlans );
+
 		const kindOfPlanTypeSelector = this.getKindOfPlanTypeSelector( this.props );
 
 		// If advertising plans for a certain feature, ensure user has pressed "View all plans" before they can see others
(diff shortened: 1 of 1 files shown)


zaguiini:
Does that mean we can get rid of the `hideXPlan` props?

claudiucelfilip:
> Does that mean we can get rid of the hideXPlan props?

I think so, yes. I'd keep the `hideBusinessPlan` check [from here](https://github.com/Automattic/wp-calypso/pull/76722/files#diff-2545c62593dd002397248377b1c0feb643bd2acc7694054ce094997b1c029aceR493-R495) - it looks like that was missing there, although it might not be used yet.

zaguiini:
>I'd keep the hideBusinessPlan check [from here](https://github.com/Automattic/wp-calypso/pull/76722/files#diff-2545c62593dd002397248377b1c0feb643bd2acc7694054ce094997b1c029aceR493-R495)

Ok, but this is not yet included in this PR 😅

claudiucelfilip:
> > I'd keep the hideBusinessPlan check [from here](https://github.com/Automattic/wp-calypso/pull/76722/files#diff-2545c62593dd002397248377b1c0feb643bd2acc7694054ce094997b1c029aceR493-R495)
> 
> Ok, but this is not yet included in this PR 😅

Good point! I pushed it now.

zaguiini (approved):
This is working great, much better than my solution. WDYT of dropping the `hideXPlan` props as a follow-up? It seems redundant now that it's possible to specify an array of visible plans.

I think we could also remove the `hidePlansFeatureComparison` that are clear workarounds to this limitation.

claudiucelfilip:
> WDYT of dropping the `hideXPlan` props as a follow-up? It seems redundant now that it's possible to specify an array of visible plans.

Yes, but we might want to be careful: this `<PlansFeaturesMain>` seems like it's used in [multiple places](https://github.com/search?q=repo%3AAutomattic%2Fwp-calypso%20%3CPlansFeaturesMain&type=code) and I'm not certain the all of them use the new grid. 
Also, the `hideFreePlan` may have some [distinct uses](https://github.com/search?q=repo%3AAutomattic%2Fwp-calypso+hideFreePlan&type=code) in relation to the domain selection step.

> I think we could also remove the `hidePlansFeatureComparison` that are clear workarounds to this limitation.

Maybe in a follow-up PR, to keep things separate?

simison:
In the [other PR ](https://github.com/Automattic/wp-calypso/pull/76722) the two plans were taking all the space of previously 4 plans — how does that look with this PR? They should be narrower.

[Similar note here](https://github.com/Automattic/wp-calypso/pull/76358#issuecomment-1545459637) for 3 plans.

PR #76771, opened 2023-05-10
Stepper: restore plan comparison on `/setup/new-hosted-site`

zaguiini:
## Proposed Changes

The plan comparison was hidden in specific onboarding flows because of a bug where the comparison table was misleading. More explanation [here](https://github.com/Automattic/wp-calypso/pull/76763).

Since the bug was fixed, it makes sense to restore the comparison table.

## Testing Instructions

1. Open `/setup/new-hosted-site`;
2. Check that the "Compare Plans" button is back below the plans grid;
3. Click the button and check that the plan comparison table is displaying accurate results relative to the visible plans.

### DIFF
--- a/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
+++ b/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
@@ -1,13 +1,10 @@
 import { HOSTING_SITE_CREATION_FLOW } from '@automattic/onboarding';
-import { useDispatch } from '@wordpress/data';
 import { addQueryArgs } from '@wordpress/url';
-import { useEffect } from 'react';
 import {
 	setSignupCompleteSlug,
 	persistSignupDestination,
 	setSignupCompleteFlowName,
 } from 'calypso/signup/storageUtils';
-import { ONBOARD_STORE } from '../stores';
 import { recordSubmitStep } from './internals/analytics/record-submit-step';
 import type { Flow, ProvidedDependencies } from './internals/types';
 import './internals/hosting-site-creation-flow.scss';
@@ -62,19 +59,6 @@ const hosting: Flow = {
 		};
 
 		return { submit };
-	},
-	useSideEffect() {
-		const { setHideFreePlan, setHidePlansFeatureComparison } = useDispatch( ONBOARD_STORE );
-
-		useEffect(
-			() => {
-				setHideFreePlan( true );
-				setHidePlansFeatureComparison( true );
-			},
-			// We only need to hide the free plan once, when the flow is mounted.
-			// eslint-disable-next-line react-hooks/exhaustive-deps
-			[]
-		);
 	},
 };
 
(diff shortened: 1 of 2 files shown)


client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
```diff
@@ -63,19 +60,6 @@ const hosting: Flow = {
 
 		return { submit };
 	},
-	useSideEffect() {
-		const { setHideFreePlan, setHidePlansFeatureComparison } = useDispatch( ONBOARD_STORE );
-
-		useEffect(
-			() => {
-				setHideFreePlan( true );
```
escapemanuele, line 71: Hi @zaguiin 👋 
Just checking, do we want to have the free plan back on this flow?

zaguiini: No, we don't. Did it come back on your side? I don't see it when browsing `/setup/new-hosted-site` 🤔

zaguiini: Oh, you're right @escapemanuele! Great catch :) Updated the PR -- it should now hide the free plan on `/setup/new-hosted-site`. I was seeing the subtext with the free plan CTA. Not anymore after the latest commit!

escapemanuele: That's great; glad you fixed that! When you start to delete code it is difficult to stop 😄

zaguiini: Time for approval then @escapemanuele ;)

escapemanuele:
Hey @zaguiini, sorry for the delay.

I wanted to understand better a particular thing. I created the `setHidePlansFeatureComparison` in [this PR](https://github.com/Automattic/wp-calypso/pull/75174) to remove the feature comparison in the Newsletter flow.
You can see [this comment](https://github.com/Automattic/wp-calypso/pull/75174#issuecomment-1503400548) from @lucasmendes-design, it was intended and not because it was not working.

He said: `Currently, I don't think it makes sense to have the comparison plans grid for the tailored flows since we don't have enough contextual features to present and compare. We can iterate on that for the next version.`

Have we considered this? If so, I'm happy to proceed 😄

zaguiini:
@escapemanuele oh OK, for these kinds of flows, it makes sense to keep the comparison CTA hidden then. I'll make it visible only for `new-hosted-site`.

zaguiini:
@escapemanuele now it should be working as intended -- LIB and Newsletter should be left unchanged; `/setup/new-hosted-site` should display the plan comparison and hide the free site plan.

client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
escapemanuele: Sorry @zaguiini, does this still hide the free plan? I see `setHideFreePlan( true )` being removed!

zaguiini: It does 👍 `planTypes` takes precedence over this prop: https://github.com/Automattic/wp-calypso/pull/76771/files#diff-957ebc95d02ddd8f1ba8103e61c3b31f701ca8a93e8c8b43728aa2b5eadf1d37R95

escapemanuele: Thanks for clarifying this 👍

escapemanuele (approved):
Thanks for addressing my comments 👍  Good to go!

PR #76776, opened 2023-05-10
Reader post sharing / Press this - replace created post content with an embed of the original

Addison-Stavlo:
Resolves item 3 of pe7F0s-P6-p2

## Proposed Changes

* This replaces the content of the post that is started from sharing a post from the reader/press-this with an embed of the original post. 
After:
Screenshot 2023-05-10 at 11 43 51 AM

## Notes on the Embed block

(EDIT- these issues have all been resolved) Working on this seems to be highlighting some issues with the core embed block for WP sites. Fixing these is out of scope of this PR, but seems like they may be a blocker for making these changes. Here are 3 examples of using the embed block in the editor and how it is rendered on the front end:

### In the editor:
Screenshot 2023-05-10 at 12 12 25 PM

* The first and third embed blocks are cut-off
* The second (middle) embed block only renders a link.

### On the front-end:
Screenshot 2023-05-10 at 12 09 50 PM

* The first and third embed blocks are still cut-off, however the third is less-so than before. Buttons for comments and sharing are still cut off on the first.
* The second (middle) that was only a link in the editor ends up rendering as expected on the front-end.

## Testing Instructions

* Build and sync this branch of `wpcom-block-editor` to your sandbox.  ( PCYsg-l4k-p2 for info on this step for wpcom-block-editor )
* sandbox the public-api, widgets, and site you will share the post from.
* Build this calypso branch (or similarly use the calypso.live link supplied below).
* Go to the reader.
* Find a post including a featured image.
* Use the sharing (curved arrow) icon to open the sharing popover.
* select the site you would like to share this from.
* Screenshot 2023-05-09 at 1 40 49 PM
* when the editor loads, verify the only post content is an embed of the original post.

## Pre-merge Checklist


roo2:
can confirm that the embed block looks pretty poor in this context.
For the post I reblogged, when I publish the page, it basically just renders as a link

shared: https://traveltheworldwithlouise.com/2023/05/10/brae-park-fort-langley/
how my post looks: pbNmQQ-1Bf-p2
In the editor:
Screen Shot 2023-05-11 at 2 12 45 pm

I don't think we should merge unless we can make some improvements to the embed block

apps/wpcom-block-editor/src/default/features/press-this.js
```diff
@@ -3,40 +3,14 @@ import { dispatch } from '@wordpress/data';
...
-			blocks.push(
-				createBlock( 'core/quote', {
-					value: `<p>${ text }</p>`,
-					citation: link,
-				} )
-			);
-		}
-
-		dispatch( 'core/editor' ).resetEditorBlocks( blocks );
+		dispatch( 'core/editor' ).resetEditorBlocks( [ createBlock( 'core/embed', { url: url } ) ] );
```
roo2: I think we can improve the appearance by passing some more attributes to the core embed block https://developer.wordpress.org/block-editor/reference-guides/core-blocks/#embed

Addison-Stavlo:
> can confirm that the embed block looks pretty poor in this context.
For the post I reblogged, when I publish the page, it basically just renders as a link

Yeah, that seemes similar to the middle example in the PR description above.  In the editor it just renders a link, but on the published front-end seems perfectly fine:

Screenshot 2023-05-12 at 11 53 50 AM

It seems to get turned into the "link" variation instead of the "wp-embed" like the others, even if I force it in the attributes. 🤔 I wonder what is in common with posts that fail to embed like this. 

> I don't think we should merge unless we can make some improvements to the embed block

Agreed. 

- [x] Regarding the first issue of the embeds showing up but having the site info and buttons at the button cut off, this seems to have been due to the marketing banner on free sites and should be resolved with D110647-code
- [x] For the issue where the editor preview ends up being only a link while the embed shows up on the front-end, will still need to investigate.

roo2:
I tested again, for me it was still showing just a normal link, I looked into it and added a new embed block to the post, the one I added manually had `type=wp-embed` but the one added by the Press This button was `type=link`.

Addison-Stavlo:
>  I looked into it and added a new embed block to the post, the one I added manually had type=wp-embed but the one added by the Press This button was type=link. 

With the last commit in [add caption and wp-embed type](https://github.com/Automattic/wp-calypso/pull/76776/commits/538a43d2b5517c774dcab72735aecc392f6f7dbe), we started adding them as the `wp-embed` type specifically, but this doesnt seem to help. In all of my testing these sites (seems to be simple sites on custom domains) cause the block to revert from `wp-embed` to `link`. A 3PC issue maybe? not sure yet. - [issue here for refernce](https://github.com/Automattic/wp-calypso/issues/76892)

Addison-Stavlo:
Embed block bugs should be resolved across simple/atomic, this is ready for testing again.

davemart-in (approved):
Looking great! Here's what I saw:

https://github.com/Automattic/wp-calypso/assets/5634774/67c3dadf-08ee-4599-bd76-fa452ab9cfd0

Not sure if we need that last line of text after the embed, but it's not a blocker for launch. 

Nice work pushing through on this one!!

Addison-Stavlo:
> Not sure if we need that last line of text after the embed, but it's not a blocker for launch.

Ah yes, the caption! Il clean that up and remove it here in a minute and we can include that in this PR. I originally added that playing around to make the broken embeds feel less broken. But those are now resolved 😁

Addison-Stavlo:
> Not sure if we need that last line of text after the embed, but it's not a blocker for launch.

Updated and removed that caption.

apps/wpcom-block-editor/src/default/features/press-this.js
```diff
@@ -3,30 +3,15 @@ import { dispatch } from '@wordpress/data';
...
-		if ( text ) {
-			blocks.push(
-				createBlock( 'core/quote', {
-					value: `<p>${ text }</p>`,
-					citation: link,
-				} )
-			);
-		}
-
-		dispatch( 'core/editor' ).resetEditorBlocks( blocks );
```
roo2, line 30: for the few minutes between the calypso change being deployed and the wpcom-block-editor change being deployed, I think embed's won't embed anything, because it relies on `text` being set to embed the link and the calypso part will no longer pass `text` through

roo2 (approved):
LGTM. I think to deploy this you should split the wpcom-block-editor part out and deploy it first

roo2:
Also, this is what happens if the URL cannot be embedded. LGTM 👍  what do you think? I think it happened because the link is `http://` only, not `https://` ?
Screen Shot 2023-05-26 at 11 49 52 am

roo2:
Actually I think that is a bug testing on atomic sites! I actually didn't realise this feature supports atomic but it does, try reblogging onto an atomic site. It may be another existing issue unrelated to this PR.
Screen Shot 2023-05-26 at 11 49 52 am

roo2:
actually, yes before this diff, the embeds work on Atomic because they won't use embed blocks.
Screen Shot 2023-05-26 at 12 10 58 pm

roo2 (changes requested):
We'd better fix the atomic issue before merging

apps/wpcom-block-editor/src/default/features/press-this.js
Addison-Stavlo: great point!

Addison-Stavlo:
@roo2  what exactly is the issue with Atomic?  We just fixed embedding simple site posts onto atomic yestereday, and embedding atomic onto simple has been working. Was it the insecure URL??

Addison-Stavlo:
Screenshot 2023-05-26 at 7 41 38 AM

 Reblogging onto an atomic site with this branch i see this ^^

Addison-Stavlo:
@roo2 - I have been testing this a bit. I can ONLY get what you are showing if the post I am trying to embed is from my WoA dev blog on the dev server pools.  Ive tested a handful of atomic sites in different directions.  I have also tested the link you are trying to embed both on an atomic site from my a8c account as well as a non-a8c account and it worked in both cases (also even with forcing https to http).  I also followed that site and tried sharing from the flow in this branch, and in all tests I am able to embed that post on my sites both simple and atomic:

Screenshot 2023-05-26 at 8 40 19 AM

Perhaps this is an environment testing issue? Im tempted to ship and revert or address this separately if there are issues.

I imagine if your atomic dev site is set up to sandbox the public-api, and you have this build patched without having pulled other more recent changes that fixed embedding simple sites on atomic, that you would see this issue.

Addison-Stavlo:
Others were unable to repro as well.  p1685109927964359/1685101737.324109-slack-C03NLNTPZ2T

Im going to split this and ship it out.

apps/wpcom-block-editor/src/default/features/press-this.js
Addison-Stavlo: Note - i was able to merge this all together and deploy the wpcom-block-editor build before the calypso build got to staging.

Addison-Stavlo: Actually - I forgot about the atomic cache with wpcom-block-editor.  Atomic sites wont pick up the wpcom-block-editor change until tomorrow. For that reason [i removed the calypso side param removals for now ](https://github.com/Automattic/wp-calypso/pull/77451) so atomic can continue to function until they pick up on that update. Come next week we can just revert that removal removal... remove the removals removal... 😆

roo2:
> I imagine if your atomic dev site is set up to sandbox the public-api, and you have this build patched without having pulled other more recent changes that fixed embedding simple sites on atomic, that you would see this issue.

Oh it could have been something like that, I was testing on an atomic site I'd been using for testing things a long time ago, I might have done something like that. 👍 

Thanks for looking into it

PR #76777, opened 2023-05-10
Support chat using Zendesk Messaging

klimeryk:
Well, this one is a doozy.

We're switching from Happy Chat to Zendesk Messaging for live chat support. This PR is big, but it's also the smallest I could get given the scope of the project. Any cleanup of Happy Chat code, old contact forms, etc. will be done as a follow-up. I want to keep this PR as easy to revert as possible if needed.

## Proposed Changes

* Disable Happy Chat feature flag and any availability checks that trigger connections to the websocket.
* Load the Zendesk Messaging script instead.
* Authenticate using already established endpoint (just with `type=zendesk`).
* Check availability using the `/help/messaging/is-available` endpoint.
* Pass the initial chat data using Zendesk user fields.
* Use the support history to figure out if the user has ongoing conversations or not (thankfully, a Messaging conversation is a "ticket"!).
* Additionally, make sure the Zendesk launcher is visible if any unread messages are reported by Zendesk's APIs directly. This is a good safeguard and makes it more dynamic.
* If the browser is blocking requests to Zendesk, show a notice and link to relevant support doc.

## Testing Instructions

The live chat support flow should feel the same up to the point where you click the chat CTA (`Chat with us`/`Still chat with us`). At that point, the Help Center should get hidden, while the Zendesk launch and widget should become visible.

Screenshot 2023-05-16 at 19 05 25

Trying to show the Help Center should automatically minimize the Zendesk widget, to avoid clutter and confusing UI. The Zendesk launcher remains visible, so the user can come back to the conversation at any time.

If you have Enhanced Tracking Protection (or similar strict privacy settings and/or ad-blockers), you will see this notice when choosing the Live Chat support option:
Screenshot 2023-05-16 at 19 07 32

Other support options should not show this notice, of course.

Verify that the Zendesk widget follows you on different paths in Calypso. Where Help Center is, the widget should also be there.

If you have an active, ongoing conversation, it should be detected and on refreshes the launcher should appear. If new messages arrive, it will be reflected with a count/notification on the launcher as well.

Screenshot 2023-05-16 at 17 41 35

Verify these behaviors in different environments that Help Center has to work in:
 - Simple sites
 - Atomic sites
 - Editor

## Pre-merge Checklist

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * WP_REST_Help_Center_User_Fields file.
+ *
+ * @package A8C\FSE
+ */
+
+namespace A8C\FSE;
+
+use Automattic\Jetpack\Connection\Client;
+
+/**
+ * Class WP_REST_Help_Center_User_Fields.
+ */
+class WP_REST_Help_Center_User_Fields extends \WP_REST_Controller {
+	/**
+	 * WP_REST_Help_Center_User_Fields constructor.
+	 */
+	public function __construct() {
+		$this->namespace = 'help-center';
+		$this->rest_base = '/zendesk/user-fields';
+	}
+
+	/**
+	 * Register available routes.
+	 */
+	public function register_rest_route() {
+		register_rest_route(
+			$this->namespace,
+			$this->rest_base,
+			array(
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'callback'            => array( $this, 'update_user_fields' ),
+				'permission_callback' => array( $this, 'permission_callback' ),
+				'args'                => array(
+					'fields' => array(
+						'type'     => 'object',
+						'required' => true,
+					),
+				),
+			)
+		);
+	}
+
+	/**
+	 * Callback to update user fields in Zendesk
+	 *
+	 * @param \WP_REST_Request $request    The request sent to the API.
+	 *
+	 * @return WP_REST_Response
+	 */
+	public function update_user_fields( \WP_REST_Request $request ) {
+		$body = Client::wpcom_json_api_request_as_user(
+			'help/zendesk/update-user-fields',
+			'2',
+			array(
+				'method' => 'POST',
+			),
+			$request['fields']
+		);
+
+		if ( is_wp_error( $body ) ) {
+			return $body;
+		}
+		$response = json_decode( wp_remote_retrieve_body( $body ) );
+		return rest_ensure_response( $response );
+	}
+
+	/**
+	 * Callback to determine whether the user has permission to access.
+	 */
+	public function permission_callback() {
+		return is_user_logged_in();
+	}
+}
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
@@ -2,6 +2,7 @@ window.configData = {
 	env_id: 'development',
 	i18n_default_locale_slug: 'en',
 	google_analytics_key: 'UA-10673494-15',
+	zendesk_support_chat_key: 'cec07bc9-4da6-4dd2-afa7-c7e01ae73584',
 	client_slug: 'browser',
 	twemoji_cdn_url: 'https://s0.wp.com/wp-content/mu-plugins/wpcom-smileys/twemoji/2/',
 	site_filter: [],
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
@@ -39,7 +39,7 @@ window.configData = {
 	is_running_in_jetpack_site: false,
 	gutenboarding_url: '/new',
 	features: {
-		happychat: true,
+		happychat: false,
 	},
 	signup_url: '/',
 	discover_blog_id: 53424024,
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
@@ -1,8 +1,8 @@
 window.configData = {
-	env_id: 'development',
+	env_id: 'production',
 	i18n_default_locale_slug: 'en',
 	google_analytics_key: 'UA-10673494-15',
-	zendesk_support_chat_key: '715f17a8-4a28-4a7f-8447-0ef8f06c70d7',
+	zendesk_support_chat_key: 'cec07bc9-4da6-4dd2-afa7-c7e01ae73584',
 	client_slug: 'browser',
 	twemoji_cdn_url: 'https://s0.wp.com/wp-content/mu-plugins/wpcom-smileys/twemoji/2/',
 	site_filter: [],
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
@@ -2,7 +2,7 @@ window.configData = {
 	env_id: 'development',
 	i18n_default_locale_slug: 'en',
 	google_analytics_key: 'UA-10673494-15',
-	zendesk_support_chat_key: 'cec07bc9-4da6-4dd2-afa7-c7e01ae73584',
+	zendesk_support_chat_key: '715f17a8-4a28-4a7f-8447-0ef8f06c70d7',
 	client_slug: 'browser',
 	twemoji_cdn_url: 'https://s0.wp.com/wp-content/mu-plugins/wpcom-smileys/twemoji/2/',
 	site_filter: [],
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -255,10 +255,6 @@ export const HelpCenterContactForm = () => {
 		Boolean( supportSite?.is_wpcom_atomic )
 	);
 
-	if ( ! thirdPartyCookiesAllowed ) {
-		return <ThirdPartyCookiesNotice />;
-	}
-
 	const enableGPTResponse = config.isEnabled( 'help/gpt-response' );
 	const showingSibylResults = params.get( 'show-results' ) === 'true';
 	const showingGPTResponse = params.get( 'show-gpt' ) === 'true';
@@ -311,7 +307,6 @@ export const HelpCenterContactForm = () => {
 						section: sectionName,
 					} );
 
-					//navigate( '/inline-chat' );
 					setShowHelpCenter( false );
 					setShowMessagingLauncher( true );
 					setShowMessagingWidget( true );
@@ -500,6 +495,10 @@ export const HelpCenterContactForm = () => {
 
 		return isSubmitting ? formTitles.buttonSubmittingLabel : formTitles.buttonLabel;
 	};
+
+	if ( ! thirdPartyCookiesAllowed ) {
+		return <ThirdPartyCookiesNotice />;
+	}
 
 	// TODO: A/B test
 	if ( enableGPTResponse && showingGPTResponse ) {
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -309,6 +309,7 @@ export const HelpCenterContactForm = () => {
 					setShowHelpCenter( false );
 					setShowMessagingLauncher( true );
 					setShowMessagingWidget( true );
+					resetStore();
 					break;
 				}
 				break;
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -310,7 +310,12 @@ export const HelpCenterContactForm = () => {
 						section: sectionName,
 					} );
 
-					submitZendeskUserFields( { route: sectionName } )
+					submitZendeskUserFields( {
+						messaging_source: sectionName,
+						messaging_initial_message: message,
+						messaging_plan: '', // Will be filled out by backend
+						messaging_url: supportSite?.URL,
+					} )
 						.then( () => {
 							setShowHelpCenter( false );
 							setShowMessagingLauncher( true );
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -510,7 +510,7 @@ export const HelpCenterContactForm = () => {
 		return isSubmitting ? formTitles.buttonSubmittingLabel : formTitles.buttonLabel;
 	};
 
-	if ( mode === 'CHAT' && ! thirdPartyCookiesAllowed ) {
+	if ( mode === 'CHAT' && thirdPartyCookiesAllowed === false ) {
 		return <ThirdPartyCookiesNotice />;
 	}
 
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -9,6 +9,7 @@ import { FormInputValidation, Popover } from '@automattic/components';
 import {
 	useSubmitTicketMutation,
 	useSubmitForumsMutation,
+	useUpdateZendeskUserFieldsMutation,
 	useSiteAnalysis,
 	useUserSites,
 	AnalysisReport,
@@ -161,6 +162,8 @@ export const HelpCenterContactForm = () => {
 	const locale = useLocale();
 	const { isLoading: submittingTicket, mutateAsync: submitTicket } = useSubmitTicketMutation();
 	const { isLoading: submittingTopic, mutateAsync: submitTopic } = useSubmitForumsMutation();
+	const { isLoading: submittingZendeskUserFields, mutateAsync: submitZendeskUserFields } =
+		useUpdateZendeskUserFieldsMutation();
 	const userId = useSelector( getCurrentUserId );
 	const { data: userSites } = useUserSites( userId );
 	const userWithNoSites = userSites?.sites.length === 0;
@@ -219,7 +222,7 @@ export const HelpCenterContactForm = () => {
 	);
 
 	const ownershipStatusLoading = ownershipResult?.result === 'LOADING';
-	const isSubmitting = submittingTicket || submittingTopic;
+	const isSubmitting = submittingTicket || submittingTopic || submittingZendeskUserFields;
 
 	// if the user picked a site from the picker, we don't need to analyze the ownership
 	if ( currentSite && sitePickerChoice === 'CURRENT_SITE' ) {
@@ -307,10 +310,16 @@ export const HelpCenterContactForm = () => {
 						section: sectionName,
 					} );
 
-					setShowHelpCenter( false );
-					setShowMessagingLauncher( true );
-					setShowMessagingWidget( true );
-					resetStore();
+					submitZendeskUserFields( { route: sectionName } )
+						.then( () => {
+							setShowHelpCenter( false );
+							setShowMessagingLauncher( true );
+							setShowMessagingWidget( true );
+							resetStore();
+						} )
+						.catch( () => {
+							setHasSubmittingError( true );
+						} );
 					break;
 				}
 				break;
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -42,6 +42,7 @@ import { BackButton } from './back-button';
 import { HelpCenterGPT } from './help-center-gpt';
 import { HelpCenterOwnershipNotice } from './help-center-notice';
 import { SibylArticles } from './help-center-sibyl-articles';
+import ThirdPartyCookiesNotice from './help-center-third-party-cookies-notice';
 import type { HelpCenterSelect } from '@automattic/data-stores';
 import './help-center-contact-form.scss';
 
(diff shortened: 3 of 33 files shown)


apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
```diff
@@ -1,7 +1,8 @@
 window.configData = {
-	env_id: 'development',
+	env_id: 'production',
```
klimeryk, line 2: This part is confusing, because it seems this file is used both in dev and production? But with the new checks, I think it should default to `production` so that the end bundle has the correct env. And if needed, one can change it to `development` in their local version.

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
```diff
@@ -0,0 +1,75 @@
...
+	 * Register available routes.
+	 */
+	public function register_rest_route() {
+		register_rest_route(
+			$this->namespace,
+			$this->rest_base,
+			array(
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'callback'            => array( $this, 'update_user_fields' ),
+				'permission_callback' => array( $this, 'permission_callback' ),
```
alshakero, line 34:
```suggestion
				'permission_callback' => array( $this, 'is_user_logged_in' ),
```

packages/help-center/src/components/help-center.tsx
```diff
@@ -2,51 +2,136 @@
...
-		setUnreadCount( unreadCount );
-	}, [ unreadCount, setUnreadCount ] );
+		const jwt = messagingAuth?.user.jwt;
+		if ( typeof window.zE !== 'function' || ! jwt ) {
+			return;
+		}
+
+		window.zE( 'messenger', 'loginUser', function ( callback ) {
+			callback( jwt );
+		} );
```
alshakero, line 85: Not sure how much of a good idea, but it may be nice to move this to `onSuccess` of `useMessagingAuth` query for fewer effects and  more cohesive code.

packages/help-center/src/components/help-center.tsx
```diff
@@ -2,51 +2,136 @@
...
+			showMessagingLauncher: helpCenterSelect.isMessagingLauncherShown(),
+			showMessagingWidget: helpCenterSelect.isMessagingWidgetShown(),
+		};
+	}, [] );
+
+	useEffect( () => {
+		if ( typeof window.zE !== 'function' ) {
+			return;
+		}
+		if ( showMessagingLauncher ) {
```
alshakero, line 100: I would make `showMessagingLauncher` a string instead of a boolean and just pass it around.

packages/help-center/src/hooks/use-messaging-auth.ts
```diff
@@ -0,0 +1,40 @@
...
+				apiNamespace: 'wpcom/v2',
+				apiVersion: '2',
+				method: 'POST',
+		  } ) ) as MessagingAuth )
+		: ( ( await apiFetch( {
+				path: addQueryArgs( '/help-center/authenticate/chat', params ),
+				method: 'POST',
+				global: true,
+		  } as APIFetchOptions ) ) as MessagingAuth );
+}
```
alshakero: The `async` is creating a promise around another promise. 

```suggestion
function requestMessagingAuth() {
	const currentEnvironment = config( 'env_id' );
	const params = { type: 'zendesk', test_mode: String( currentEnvironment === 'development' ) };
	const wpcomParams = new URLSearchParams( params );
	return canAccessWpcomApis()
		? wpcomRequest<MessagingAuth>( {
				path: '/help/authenticate/chat',
				query: wpcomParams.toString(),
				apiNamespace: 'wpcom/v2',
				apiVersion: '2',
				method: 'POST',
		  } ) 
		: apiFetch<MessagingAuth>( {
				path: addQueryArgs( '/help-center/authenticate/chat', params ),
				method: 'POST',
				global: true,
		  } as APIFetchOptions ) ;
}
```

packages/help-center/src/hooks/use-messaging-availability.ts
```diff
@@ -0,0 +1,47 @@
...
+import wpcomRequest, { canAccessWpcomApis } from 'wpcom-proxy-request';
+import type { MessagingAvailability } from '../types';
+
+interface APIFetchOptions {
+	global: boolean;
+	path: string;
+}
+
+async function requestMessagingAvailability() {
+	const currentEnvironment = config( 'env_id' );
```
alshakero, line 14: After we trained HEs for months about proxying status, I think this will create confusion again. Can the same logic apply here? `proxied => dev, no => prod`?

packages/help-center/src/hooks/use-messaging-availability.ts
```diff
@@ -0,0 +1,47 @@
...
+import { addQueryArgs } from '@wordpress/url';
+import wpcomRequest, { canAccessWpcomApis } from 'wpcom-proxy-request';
+import type { MessagingAvailability } from '../types';
+
+interface APIFetchOptions {
+	global: boolean;
+	path: string;
+}
+
+async function requestMessagingAvailability() {
```
alshakero: No need for async.

packages/help-center/src/hooks/use-messaging-availability.ts
```diff
@@ -0,0 +1,47 @@
...
+	const currentEnvironment = config( 'env_id' );
+	const params = {
+		group: 'wpcom_messaging',
+		environment: currentEnvironment === 'development' ? 'development' : 'production',
+	};
+	const wpcomParams = new URLSearchParams( params );
+	return canAccessWpcomApis()
+		? ( ( await wpcomRequest( {
```
alshakero:
```suggestion
		? ( ( wpcomRequest<MessagingAvailability>( {
```

packages/help-center/src/hooks/use-messaging-availability.ts
```diff
@@ -0,0 +1,47 @@
...
+				query: wpcomParams.toString(),
+				apiNamespace: 'wpcom/v2',
+				apiVersion: '2',
+				method: 'GET',
+		  } ) ) as MessagingAvailability )
+		: ( ( await apiFetch( {
+				path: addQueryArgs( '/help-center/support-availability/messaging', params ),
+				method: 'GET',
+				global: true,
+		  } as APIFetchOptions ) ) as MessagingAvailability );
```
alshakero:
```suggestion
		  } ) ) as MessagingAvailability )
		: ( ( apiFetch<MessagingAvailability>( {
				path: addQueryArgs( '/help-center/support-availability/messaging', params ),
				method: 'GET',
				global: true,
		  } as APIFetchOptions ) ))
```

packages/help-center/src/hooks/use-zendesk-config.ts
```diff
@@ -0,0 +1,18 @@
+import { useQuery } from '@tanstack/react-query';
+
+async function requestZendeskConfig() {
+	return await window.fetch( 'https://wpcom.zendesk.com/embeddable/config', {
+		method: 'GET',
+	} );
+}
```
alshakero:
```suggestion
function requestZendeskConfig() {
	return window.fetch( 'https://wpcom.zendesk.com/embeddable/config', {
		method: 'GET',
	} );
}
```

packages/help-center/src/hooks/use-zendesk-config.ts
```diff
@@ -0,0 +1,18 @@
+import { useQuery } from '@tanstack/react-query';
+
+async function requestZendeskConfig() {
+	return await window.fetch( 'https://wpcom.zendesk.com/embeddable/config', {
+		method: 'GET',
```
alshakero: Default method is `GET`.
```suggestion
		
```

packages/help-center/src/components/help-center.tsx
```diff
@@ -2,51 +2,136 @@
...
+		} else {
+			window.zE( 'messenger', 'hide' );
+		}
+	}, [ showMessagingLauncher ] );
+
+	useEffect( () => {
+		if ( typeof window.zE !== 'function' ) {
+			return;
+		}
+		if ( showMessagingWidget ) {
```
alshakero, line 111: Same for `showMessagingWidget`, can we make it a string?

packages/help-center/src/components/help-center.tsx
```diff
@@ -2,51 +2,136 @@
...
+				}
+			} );
+		}
 
-	const { unreadCount, closeChat } = useHCWindowCommunicator( isMinimized || ! show );
+		loadScript(
+			'https://static.zdassets.com/ekr/snippet.js?key=' + encodeURIComponent( zendeskKey ),
+			setUpMessagingEventHandlers,
+			{ id: ZENDESK_SCRIPT_ID }
+		);
```
alshakero: This would be nice to be wrapped in a factory function in the Help Center to avoid repetition. Something like

`ZDInstance({ onUnread, onClose, onOpen })`

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -487,6 +510,10 @@ export const HelpCenterContactForm = () => {
 		return isSubmitting ? formTitles.buttonSubmittingLabel : formTitles.buttonLabel;
 	};
 
+	if ( mode === 'CHAT' && ! thirdPartyCookiesAllowed ) {
```
alshakero: To avoid showing the notice during loading:

```suggestion
	if ( mode === 'CHAT' && thirdPartyCookiesAllowed === false ) {
```

packages/help-center/src/hooks/use-zendesk-config.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export default function useZendeskConfig( enabled: boolean ) {
+	return useQuery< Response >( [ 'getZendeskConfig' ], requestZendeskConfig, {
+		staleTime: Infinity,
+		retry: false,
+		meta: {
+			persist: false,
+		},
+		enabled,
```
alshakero, line 17: Almost a blocker: you can add a `select` here to process the resolved data to add another field for 3rd party cookies, then you can do:

```js
const { thirdPartyCookieAllowed } = useZendeskConfig( Boolean( chatStatus?.is_user_eligible ) );
```

...anywhere you want without all the store boilerplate, the effects, and you get free caching.

packages/help-center/src/components/help-center-third-party-cookies-notice.tsx
```diff
@@ -0,0 +1,46 @@
...
+	const { __ } = useI18n();
+	const sectionName = useSelector( getSectionName );
+
+	useEffect( () =>
+		recordTracksEvent( 'calypso_helpcenter_third_party_cookies_notice_open', {
+			force_site_id: true,
+			location: 'help-center',
+			section: sectionName,
+		} )
+	);
```
alshakero: Blocker: this will run many times, needs an empty dep array.

packages/help-center/src/components/help-center.tsx
```diff
@@ -2,51 +2,136 @@
...
+	const email = useSelector( getCurrentUserEmail );
+	const { data: supportHistory } = useSupportHistory( 'ticket', email );
 	useEffect( () => {
-		if ( data?.status === 'assigned' ) {
-			setShowHelpCenter( true );
+		if ( supportHistory?.some( ( ticket ) => ticket.channel === 'native_messaging' ) ) {
+			setShowMessagingLauncher( true );
 		}
-	}, [ data, setShowHelpCenter ] );
+	}, [ setShowMessagingLauncher, supportHistory ] );
```
alshakero: Same as the other effect, we can add a `select` in `useSupportHistory` that returns a boolean.

alshakero (changes requested):
Reviewed, but didn't get to test yet. The majority are non-blockers, only one is actually a blocker, the rest all are optional.

arcangelini:
@klimeryk are there some steps to testing this with ZD chat in staging? I have tried just opening things as I would with HelpCenter normally, but I am getting an error:

```
FoundError: No Zendesk user ID found.
    at onmessage (index.js:474:64)
```

packages/help-center/src/hooks/use-messaging-availability.ts
arcangelini:
```suggestion
	return canAccessWpcomApis()
		? await wpcomRequest< MessagingAvailability >( {
				path: '/help/messaging/is-available',
				query: wpcomParams.toString(),
				apiNamespace: 'wpcom/v2',
				apiVersion: '2',
				method: 'GET',
		  } )
		: await apiFetch< MessagingAvailability >( {
				path: addQueryArgs( '/help-center/support-availability/messaging', params ),
				method: 'GET',
				global: true,
		  } as APIFetchOptions );
```

The above suggestions in one edit.

packages/help-center/src/hooks/use-messaging-availability.ts
klimeryk: From my experience, it's more useful for HEs to be able to easily test if chat is available for production users. So I'd prefer to take this opportunity to clear this up. Especially since I imagine a lot of them will be testing this flow after we launch it and I don't want additional confusion of staging versus production 🙇

packages/help-center/src/components/help-center.tsx
klimeryk: I'm not sure I understand - could you clarify what are the benefits of it being a string and how that would make it easier to pass it around?

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
arcangelini: Actually I think it would be:

```suggestion
				'permission_callback' => 'is_user_logged_in',
```

And this would alleviate from having to add the `permission_callback` below.

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
```diff
@@ -0,0 +1,75 @@
...
+		return rest_ensure_response( $response );
+	}
+
+	/**
+	 * Callback to determine whether the user has permission to access.
+	 */
+	public function permission_callback() {
+		return is_user_logged_in();
+	}
+}
```
arcangelini, line 75: This would be removed if you switch the callback.

I think this is how all the other endpoints currently all, so no worries if you don't change this. I think we are going to clean all these up in the near future to make them all the same.

arcangelini (commented):
Wasn't able to test the changes, but read through. Everything seems to be good from my perspective.

As you highlighted before I do think this should prompt a general spring cleaning of Help Center to make it easier to understand and update since many decisions were made with happychat in mind.

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
klimeryk: I like the shorter syntax (and you can see the check inline), but I'd prefer to keep it as-is:
 - It's more consistent with other endpoints.
 - It's easier to extend the permission check - the diff would be smaller.

packages/help-center/src/hooks/use-messaging-availability.ts
arcangelini: This was specifically referring to Omar's suggestions. Haha just a rewrite that was easier to read.

packages/help-center/src/components/help-center.tsx
klimeryk: Hmmm, I like the idea, but for now I'd prefer to keep it as-is. I'm worried about race conditions here and want to see how the code will perform in real scenarios, at WordPress.com's scale. The part I'm worried about is the `typeof window.zE !== 'function'` check. The Zendesk script needs to be loaded first, only then we can authenticate. In my testing it worked well, but I might need to adjust the code to be dependent on the `loadScript`'s callback to fire first.

packages/help-center/src/hooks/use-messaging-auth.ts
klimeryk: Thanks for pointing that (and similar instances) out! I was copying code around and looks like it sneaked in. Fixed in https://github.com/Automattic/wp-calypso/pull/76777/commits/bcbcc946e4440b6c606700fea168825b586e6007

packages/help-center/src/hooks/use-messaging-availability.ts
klimeryk: Fixed in https://github.com/Automattic/wp-calypso/pull/76777/commits/bcbcc946e4440b6c606700fea168825b586e6007

packages/help-center/src/hooks/use-zendesk-config.ts
klimeryk: Fixed in https://github.com/Automattic/wp-calypso/pull/76777/commits/bcbcc946e4440b6c606700fea168825b586e6007

packages/help-center/src/components/help-center-contact-form.tsx
klimeryk: Good call! https://github.com/Automattic/wp-calypso/pull/76777/commits/26b1c8e046893c2346e12b90c1075c37677a3a44

packages/help-center/src/components/help-center-third-party-cookies-notice.tsx
klimeryk: Very nice catch! Fixed in https://github.com/Automattic/wp-calypso/pull/76777/commits/852ace6f1fc649f5c48b6cc730a1b64d6ec2cf2a. Technically, the effect depends on `sectionName`, but I've kept the dependencies empty (`[]`) so that it only fires on mount. Otherwise, it'd fire every time when the notice is on screen and the user is changing sections.

packages/help-center/src/hooks/use-messaging-availability.ts
klimeryk: Fixed as per https://github.com/Automattic/wp-calypso/pull/76777#discussion_r1196209571 🙇

apps/editing-toolkit/editing-toolkit-plugin/help-center/class-wp-rest-help-center-user-fields.php
klimeryk: As mentioned above https://github.com/Automattic/wp-calypso/pull/76777#discussion_r1196186208, I'll keep this as-is for now. Like you said, it's more consistent with other endpoints and if we're gonna clean it up, we can do it then.

packages/help-center/src/components/help-center.tsx
klimeryk: Yup, absolutely! I'll definitely clean up this code once this is launched and any major bugs are fixed. I want to keep it in an "exploded" format for now, so that it's easier to move things around and quickly refactor things 🙇‍♂️

packages/help-center/src/components/help-center.tsx
alshakero: This `if` is basically converting `true` to `'show'` and `false` to `'hide'`. I prefer 

```js
window.zE( 'messenger', showMessagingLauncher )
```

It's inspired by this https://kentcdodds.com/blog/stop-using-isloading-booleans

klimeryk:
> @klimeryk are there some steps to testing this with ZD chat in staging? I have tried just opening things as I would with HelpCenter normally, but I am getting an error:

Thanks for testing! I'll reach out on Slack to get more info on which user you've used to test 🙇

packages/help-center/src/hooks/use-zendesk-config.ts
klimeryk: I like the idea! Definitely happy with less boilerplate. Unfortunately, `select` does not seem to work my use-case. From my testing, it seems it's only called when the promise resolves. If there's a network error (which is exactly what I'm interested in here), it's never called.
But this gave me an idea to simplify this code and indeed just use `useZendeskConfig` directly.

klimeryk: Fixed in https://github.com/Automattic/wp-calypso/pull/76777/commits/f4b09067cd81bba7d2c79efae4a40c6189c537d2.
While testing I notice that despite having `retry: false`, react-query _really_ seems to hate the request failing and tries to retry it on focus, mount, etc. So I had to explicitly disable those too.

packages/help-center/src/components/help-center.tsx
klimeryk: I appreciate the suggestion, but in this case I prefer to keep it as-is. `useSupportHistory` returns an array of previous support interactions of given type - plain and simple. Throwing in an additional field there would make its return value less clear. And technically, it can also request a `chat` support history (though, with the advent of Zendesk Messaging for chat that needs cleaning up cause Messaging "chats" are actually tickets too), where this property would not apply.

arcangelini:
Tested in Simple and Atomic, everything seems to be A-OK. One suggestion that I mentioned in Slack was that we redirect the user who HAS an active chat session before we give them the chat contact form. This way we prevent any confusions and that extra irrelevant API call to send the update-fields.

Not a blocker, but definitely something we should add soon after.

Markup 2023-05-17 at 13 17 09

Also, since now the chat history is shared in the Zendesk ticket history the warning of a previous sumbmission may neeed to be re-worded or updated? Right now it think the user submitted a ticket when actually its an open chat. Again, this is just a thought and not blocking in any way.

Markup 2023-05-17 at 13 42 52

klimeryk:
Thanks so much for testing and sharing your feedback!

>  One suggestion that I mentioned in Slack was that we redirect the user who HAS an active chat session before we give them the chat contact form. This way we prevent any confusions and that extra irrelevant API call to send the update-fields.

Yup, definitely, I'll rework it so that if the user selects Live chat option, they'll immediately go to Zendesk Messaging. I'll do that in a follow-up PR, though, as we need to launch now 🙇‍♂️ I'll be on the lookout for any reports about the confusion around the notice and ask for feedback how/if we should it improve it as well - thanks for pointing it out!

PR #76803, opened 2023-05-11
Reader: adds view count, when available, to posts in feed

creativecoder:
## Proposed Changes

Adds a view count for each post in the reader feed, when available.

Screenshot 2023-05-10 at 22 16 30

## Testing Instructions

- Apply D110502-code and sandbox public-api site
- Load a feed in the reader (e.g. `/read`)
- You should see a view count at the bottom of each post card, when available.


eoigal (approved):
WFM

PR #76812, opened 2023-05-11
Reader - Add suggested follows modal

eoigal:
This PR will add a modal that will show a list of recommended follows based on the site the user has just followed.

follow-1

We will only show this modal on Reader site page. We may also add it to Reader search page.

Project Ref: pe7F0s-O2-p2

### PR highlights

The PR requires the patch at code-D111404. The patch passes back extra site information for each related post, saving having to make an extra `QueryReaderSite` call for each related post, allowing the dialog to render more quickly.

This PR adds a new scope of `SCOPE_SUGGESTED_FOLLOWS` to allow us to pass extra params to the related posts endpoint.

The PR requests the relatedPosts when the full site page loads, **not** when the dialog loads. This is done this way to avoid waiting on the dialog to load while the related posts request is processing.

Another thing of note, the Follow button style is left unchanged for now. The designs have a button style but this would require a change to the followbutton component which is something I think we can come back to.

### Testing
* Apply patch code-D111404 on sandbox
* Go to a reader site page like http://calypso.localhost:3000/read/feeds/35154032
* Click the `Follow` button and you should see a dialog like the image below - confirm it works and looks as expected

Screenshot 2023-05-19 at 16 15 16


creativecoder (commented):
Picking through a list of related posts to construct a list of sites in the client seems somewhat inefficient. What would it take to generate a list of suggested sites on the server and just use that data directly?

eoigal:
> Picking through a list of related posts to construct a list of sites in the client seems somewhat inefficient. What would it take to generate a list of suggested sites on the server and just use that data directly?

I could replicate the related posts endpoint or pass a parameter to change the response to list sites instead but I don't think it would give us much in terms of efficiency. What overhead do you see on the client side could be removed by doing this?

davemart-in (commented):
Tested. This works as expected. Here's what I saw:

CleanShot 2023-05-22 at 10 52 59

I've got a few 1:1s right now, but then I'd like to jump in and just possibly touch up a few of the styles if that is okay.

davemart-in:
Here's a list of the updates I've made:

- reader-suggested-follow-item_link opens in a new tab now
- Added external link icon (so you know at a glance that it will open in a new tab)
- Changed description text to be, “While you're at it, you might check out these sites.”
- Updated CSS around width of modal
- Updated font styles for header and description
- Increased padding around content in modal
- Updated close button positioning
- Added border-radius to modal

Here's what it looks like now:

CleanShot 2023-05-22 at 14 27 58

Only thing I couldn't figure out how to do easily was update the backdrop color. Ollie mocked it up with a darker background, which I think looks better TBH. Here is a preview of both:

CleanShot 2023-05-22 at 14 29 30

CleanShot 2023-05-22 at 14 30 19

eoigal:
Overrode the modal backdrop when showing the suggested follows modal, looks now like this;

Screenshot 2023-05-22 at 20 29 36

PR #76814, opened 2023-05-11
Happy blocks: Adjust search box on inner pages for support

renancarvalho:
## Proposed changes

This PR makes the search box visible on `/support` inner pages. We are still not sure what is the best UX/Design to provide this to the user, so this is a temporary but valuable solution IMO.

## How to test

- Pull and run this branch
- Navigate to `cd apps/happy-blocks` and `run yarn dev --sync`
- Sandbox `wordpress.com/support` 
- Follow the [steps here](https://github.com/Automattic/wpsupport3/pull/438)


escapemanuele (approved):
Tested and this works great as a temp solution 👍

PR #76816, opened 2023-05-11
E2E: wait for previews to load before clicking on a site assembler component.

### LINKED ISSUE #76806: Flaky E2E: Select "Header"
### Spec file

specs/onboarding/setup__site-assembler.ts

### TeamCity ID

9994327

### Logs

```
Error: expect(received).toBe(expected) // Object.is equality

Expected: 1
Received: 0
at Object.<anonymous> (/home/teamcity-2/buildAgent/work/c4a9d5b38c1dacde/test/e2e/specs/onboarding/setup__site-assembler.ts:63:52)
```

worldomonation:
Fixes #76806

## Proposed Changes

This PR adds a wait for `networkidle` state before clicking on a site assembler component. 

## Testing Instructions

  - [ ] Gutenberg E2E Simple production (desktop)
  - [ ] Gutenberg E2E Simple production (mobile)
  - [ ] Calypso E2E (desktop)
  - [ ] Calypso E2E (mobile)
  - [ ] Pre-Release E2E

## Pre-merge Checklist

### DIFF
--- a/packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
@@ -32,6 +32,13 @@ export class SiteAssemblerFlow {
 	 */
 	async selectLayoutComponentType( type: LayoutType ): Promise< void > {
 		await this.page.getByRole( 'button', { name: type } ).click();
+
+		// Each header element contains an iframe that loads the preview in the card.
+		// These previews are lazy-loaded and thus only load if the card is in the
+		// viewport.
+		// By waiting for the network activity to cease for the lazy-load,
+		// mis-clicks and clicks being swallowed can be prevented.
+		await this.page.waitForLoadState( 'networkidle', { timeout: 15 * 1000 } );
 	}
 
 	/**
--- a/packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
@@ -32,13 +32,6 @@ export class SiteAssemblerFlow {
 	 */
 	async selectLayoutComponentType( type: LayoutType ): Promise< void > {
 		await this.page.getByRole( 'button', { name: type } ).click();
-
-		// Each header element contains an iframe that loads the preview in the card.
-		// These previews are lazy-loaded and thus only load if the card is in the
-		// viewport.
-		// By waiting for the network activity to cease for the lazy-load,
-		// mis-clicks and clicks being swallowed can be prevented.
-		await this.page.waitForLoadState( 'networkidle', { timeout: 15 * 1000 } );
 	}
 
 	/**
@@ -47,6 +40,28 @@ export class SiteAssemblerFlow {
 	 * @param {number} index Index of the item to choose. Defaults to 0.
 	 */
 	async selectLayoutComponent( index = 0 ): Promise< void > {
-		await this.page.locator( '.pattern-list-renderer__pattern-list-item' ).nth( index ).click();
+		const target = this.page
+			.getByRole( 'listbox', { name: 'Block patterns' } )
+			.getByRole( 'option' )
+			.nth( index )
+			.locator( 'iframe' );
+
+		// The iframe has an attribute of "height" that is present only if the preview has loaded.
+		// By checking for the presence of the "height" attribute on the last visible
+		// iframe, we can be sure that all visible component card has loaded.
+		// If the last iframe does not yet contain the "height" attribute, then wait
+		// for a predetermined timeout, then re-evaluate.
+		while (
+			! ( await target.evaluate( ( element ) => {
+				if ( element.style.height ) {
+					return true;
+				}
+				return false;
+			} ) )
+		) {
+			await this.page.waitForTimeout( 1000 );
+		}
+
+		await target.click();
 	}
 }


packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
```diff
@@ -32,6 +32,13 @@ export class SiteAssemblerFlow {
 	 */
 	async selectLayoutComponentType( type: LayoutType ): Promise< void > {
 		await this.page.getByRole( 'button', { name: type } ).click();
+
+		// Each header element contains an iframe that loads the preview in the card.
+		// These previews are lazy-loaded and thus only load if the card is in the
+		// viewport.
+		// By waiting for the network activity to cease for the lazy-load,
+		// mis-clicks and clicks being swallowed can be prevented.
+		await this.page.waitForLoadState( 'networkidle', { timeout: 15 * 1000 } );
```
worldomonation: Another way to go about this same thing that depends less on `networkidle` but on an implementation detail is the following:

```typescript
		// Each header element contains an iframe that loads the preview.
		// These elements are lazy-loaded and thus only load if visible in the
		// viewport.
		// Grab the last one that's visible first.
		const lastVisibleElement = this.page
			.getByRole( 'listbox', { name: 'Block patterns' } )
			.locator( 'iframe' )
			.last();

		// The iframe has an attribute of "height" that is present only if the
		// preview has loaded.
		// By checking for the presence of the "height" attribute on the last visible
		// iframe, we can be sure that all visible component card has loaded.
		// If the last iframe does not yet contain the "height" attribute, then wait
		// for a predetermined timeout, then re-evaluate.
		while (
			! ( await lastVisibleElement.evaluate( ( element ) => {
				if ( element.style.height ) {
					return true;
				}
				return false;
			} ) )
		) {
			await this.page.waitForTimeout( 1000 );
		}
```

I'm inclined to stick with `networkidle` because it seems to be the more legible and thus maintainable option of the two. But if I've missed something obvious, please feel free to point it out!

dpasque: I'm definitely good with trying out the `networkidle` check here!

But... I think it makes me a little nervous conceptually? Mostly because there's not a document load happening here -- we're just clicking a button that loads a bunch of elements asynchronously. `waitForLoadState`, from their docs, is predominantly for waiting for document loads. So if that event has already fired for the current URL/document, I think it will just return immediately without doing any extra waiting for network requests.

That said, I realize we don't have a lot of better options here. 😕 The only other idea I can think of is maybe modifying the selector for when we actually click on a pattern/layout component so that it automatically waits until the specific item is loaded in?

Looking at the DOM, it looks like when the item has loaded, we add a new div that looks like...
`<div class="block-renderer">...</div>`

So what if we target that inner div with the click?

like...

```typescript
await this.page.locator( '.pattern-list-renderer__pattern-list-item' ).nth( index ).locator('.block-renderer').click();
```

Would that wait for that inner div to appear before clicking?

dpasque (approved):
I'm good with trying network idle to start! 👍 I threw out one other idea though, feel free to take it or leave it!

dpasque (approved):
I'm good with trying network idle to start! 👍 I threw out one other idea though, feel free to take it or leave it!

packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
worldomonation: Good point you raise. I have some data to back up why waiting for `networkidle` should suffice, as long as our servers don't hiccup during the E2E.

> I think it will just return immediately without doing any extra waiting for network requests.

Keep in mind that `waitForLoadState('networkidle')` will wait for all network activity to cease for at least 500ms, which includes not just XHR requests but also any other media and JS that's being loaded.

To test, I added a `throw new Error` after the last action in `Select Header`, and ran the two different approaches:
1. `waitForLoadState('networkidle')`
2. using the convoluted `iframe height` attribute for the last visible component preview.

In both cases, the total runtime is ~15 seconds.

**`networkidle`**
```
Test Suites: 1 failed, 1 total
Tests:       1 failed, 2 skipped, 3 passed, 6 total
Snapshots:   0 total
Time:        14.485 s
```

**iframe wait method**
```
Test Suites: 1 failed, 1 total
Tests:       1 failed, 2 skipped, 3 passed, 6 total
Snapshots:   0 total
Time:        15.495 s
```

Dissecting further, we see that in both cases, there is a ~4 second wait (give or take some time based on network performance).

**`networkidle`**
```
  pw:api   navigated to "about:srcdoc" +701ms
  pw:api   navigated to "about:srcdoc" +199ms
  pw:api   navigated to "about:srcdoc" +226ms
  pw:api   "networkidle" event fired +4s
  pw:api <= page.waitForLoadState succeeded +2ms
  pw:api => locator.click started +0ms
```

**iframe wait method**
```
  pw:api <= locator.evaluate succeeded +193ms
  console.log
    waiting for 1000ms

      at SiteAssemblerFlow.selectLayoutComponentType (../../packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts:66:12)

  pw:api => page.waitForTimeout started +1ms
  pw:api   navigated to "about:srcdoc" +202ms
  pw:api <= page.waitForTimeout succeeded +956ms
  pw:api => locator.evaluate started +0ms
  pw:api waiting for getByRole('listbox', { name: 'Block patterns' }).locator('iframe').last() +5ms
  pw:api   locator resolved to visible <iframe tabindex="-1" loading="lazy" aria-hidden="true" …></iframe> +3ms
  pw:api <= locator.evaluate succeeded +11ms
  console.log
    waiting for 1000ms

      at SiteAssemblerFlow.selectLayoutComponentType (../../packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts:66:12)

  pw:api => page.waitForTimeout started +2ms
  pw:api <= page.waitForTimeout succeeded +1s
  pw:api => locator.evaluate started +1ms
  pw:api waiting for getByRole('listbox', { name: 'Block patterns' }).locator('iframe').last() +4ms
  pw:api   locator resolved to visible <iframe tabindex="-1" loading="lazy" aria-hidden="true" …></iframe> +3ms
  pw:api <= locator.evaluate succeeded +9ms
  console.log
    waiting for 1000ms

      at SiteAssemblerFlow.selectLayoutComponentType (../../packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts:66:12)

  pw:api => page.waitForTimeout started +2ms
  pw:api <= page.waitForTimeout succeeded +1s
  pw:api => locator.evaluate started +1ms
  pw:api waiting for getByRole('listbox', { name: 'Block patterns' }).locator('iframe').last() +5ms
  pw:api   locator resolved to visible <iframe tabindex="-1" loading="lazy" aria-hidden="true" …></iframe> +3ms
  pw:api <= locator.evaluate succeeded +9ms
  console.log
    waiting for 1000ms

      at SiteAssemblerFlow.selectLayoutComponentType (../../packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts:66:12)

  pw:api => page.waitForTimeout started +1ms
  pw:api <= page.waitForTimeout succeeded +1s
  pw:api => locator.evaluate started +0ms
```

Thus we can be certain (at least at this time) that the either of these options should work for us.

-----

> Looking at the DOM, it looks like when the item has loaded, we add a new div that looks like...
<div class="block-renderer">...</div>

Unfortunately, the `div.block-render` element is always shown, even if the preview is not loaded. See the following screenshot of the trace file, where the `div.block-render` element whose preview has not loaded is still present.

image

I ran a separate experiment with the changes in this PR removed - once for the unchanged version in trunk, and once more with your snippet.

**unchanged trunk**
```
  pw:api => locator.click started +1ms
  pw:api   not waiting, "load" event already fired +1ms
  pw:api waiting for getByRole('button', { name: 'Header' }) +3ms
  pw:api   locator resolved to <button id="/header" data-wp-c16t="true" aria-label="Hea…>…</button> +2s
  pw:api attempting click action +0ms
  pw:api   waiting for element to be visible, enabled and stable +0ms
  pw:api   element is visible, enabled and stable +18ms
  pw:api   scrolling into view if needed +0ms
  pw:api   done scrolling +1ms
  pw:api   performing click action +6ms
  pw:api   click action done +43ms
  pw:api   waiting for scheduled navigations to finish +0ms
  pw:api   navigations have finished +17ms
  pw:api <= locator.click succeeded +9ms
  pw:api => locator.click started +0ms
  pw:api waiting for locator('.pattern-list-renderer__pattern-list-item').nth(1) +38ms
  pw:api   navigated to "about:srcdoc" +19ms
  pw:api   locator resolved to <button tabindex="0" role="option" id="id-ptqtob-2" aria…>…</button> +157ms
  pw:api attempting click action +41ms
  pw:api   waiting for element to be visible, enabled and stable +0ms
  pw:api   navigated to "about:srcdoc" +104ms
  pw:api   element is visible, enabled and stable +125ms
  pw:api   scrolling into view if needed +0ms
  pw:api   done scrolling +34ms
  pw:api   navigated to "about:srcdoc" +71ms
  pw:api   performing click action +163ms
  pw:api   click action done +190ms
```

**Dan's snippet**
```
pw:api => locator.click started +1ms
  pw:api waiting for locator('.pattern-list-renderer__pattern-list-item').nth(1).locator('.block-renderer') +206ms
  pw:api   locator resolved to <div class="block-renderer">…</div> +72ms
  pw:api attempting click action +28ms
  pw:api   waiting for element to be visible, enabled and stable +0ms
  pw:api   navigated to "about:srcdoc" +96ms
  pw:api   element is visible, enabled and stable +6ms
  pw:api   scrolling into view if needed +0ms
  pw:api   done scrolling +104ms
  pw:api   performing click action +200ms
  pw:api   navigated to "about:srcdoc" +37ms
  pw:api   click action done +1ms
  pw:api   waiting for scheduled navigations to finish +0ms
  pw:api   navigated to "about:srcdoc" +4ms
  pw:api   navigations have finished +108ms
  pw:api <= locator.click succeeded +176ms
```

The results look about the same between the two. In my mind, this makes sense because the `div.block-rederer` element is already present upon `load` even firing. Playwright then proceeds to click on the element, but in rare instances when the preview loads the result is that elements shift and the click fails.

worldomonation: With all the above said though, I took a look at the documentation and apparently `networkidle` is [now discouraged](https://playwright.dev/docs/api/class-page#page-wait-for-load-state-option-state).

In its place, the team suggests using web assertions, but those exist only for folks using Playwright Test Runner, which isn't us. I have the slightest hunch that (like all things Microsoft) they ostensibly claim that Playwright will work with many test runners, but in practice if you want the latest and greatest features you'd have to use their first-party runner.

-----

Side note, the Core Playwright E2E specs use Playwright Test Runner, and having used web assertions there, it is indeed quite powerful. Instead of having to create helper methods or manually craft while loops to wait for an element to appear, you can simply do something like this:

```typescript
await expect( async () => 
    await page.getByRole('button', {name: "Click me!"} ).getAttribute( 'height' ).toBeTruthy() 
).toPass()
```

The snippet above will keep polling for the statement in the expect clause to be true until timeout is reached. There are so many places we can use something like this, including this PR.

dpasque: Oh wow, thank you for the detailed investigation! 😄  Given all that, I'm totally fine with going forward with `networkidle` for now! 👍 

Re: Playwright Test Runner... I agree on all accounts. It's kind of frustrating that it's definitely the "first class" approach, but I'd also be excited to migrate and try it out!

worldomonation: Well, looks like your concern with `networkidle` has shown itself when I triggered the Pre-Release Tests on the branch. Back to the drawing board...

```
page.waitForLoadState: Timeout 15000ms exceeded.
at SiteAssemblerFlow.selectLayoutComponentType (/home/teamcity-3/buildAgent/work/c4a9d5b38c1dacde/packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts:42:19)
    at Object.<anonymous> (/home/teamcity-3/buildAgent/work/c4a9d5b38c1dacde/test/e2e/specs/onboarding/setup__site-assembler.ts:60:4)
```

image

packages/calypso-e2e/src/lib/flows/site-assembler-flow.ts
```diff
@@ -40,6 +40,28 @@ export class SiteAssemblerFlow {
...
+				if ( element.style.height ) {
+					return true;
+				}
+				return false;
+			} ) )
+		) {
+			await this.page.waitForTimeout( 1000 );
+		}
+
+		await target.click();
```
WunderBart, line 65: What if we pierce through the iframe, forcing it to have the inner target ready instead? For example, I think something like this might work:

```js
async selectLayoutComponent( index = 0 ): Promise< void > {
	await this.page
		.getByRole( 'listbox', { name: 'Block patterns' } )
		.getByRole( 'option' )
		.nth( index )
		.frameLocator( 'iframe' )
		.locator( '.block-editor-iframe__body' )
		.click();
}

```

worldomonation: > What if we pierce through the iframe, forcing it to have the inner target ready instead? For example, I think something like this might work:
> 
> ```js
> async selectLayoutComponent( index = 0 ): Promise< void > {
> 	await this.page
> 		.getByRole( 'listbox', { name: 'Block patterns' } )
> 		.getByRole( 'option' )
> 		.nth( index )
> 		.frameLocator( 'iframe' )
> 		.locator( '.block-editor-iframe__body' )
> 		.click();
> }
> ```

Blargh, you come in with a good suggestion right after I hit merge! 

I'm going to quickly try your idea in another PR.

PR #76823, opened 2023-05-11
Happy blocks: Update WPCom nav bar for happy blocks

renancarvalho:
## Proposed Changes

The main idea of this PR is to **not** show the WPCom to logged-in users. For those, Calypso bar is the only one visible.
Besides that, it also does:
- Visually matches the logged out [WPCom](https://wordpress.com/) nav bar.
- For logged-in users, only load header.scss which contains the `happy blocks` header styles.
- For logged-out users, it loads both, header.scss and view.scss (nav bar styles).
- CSS clean up.

Before: 
image

After:
image

## Testing Instructions

- Pull and run this branch
- Navigate to `cd apps/happy-blocks` and `run yarn dev --sync`
- Sandbox wordpress.com/support and visit the site

For logged-in users: 
 - The header should not be present.
 - Check on the network tab if view.scss for the header is loaded, **It should not**. The only one visible is the footer one. 

For logged-out users:
 - The header should be present.
 - Check on the network tab if view.scss for the header is loaded, **It should**. 

*** No changes on mobile


apps/happy-blocks/block-library/universal-header/header.scss
```diff
@@ -0,0 +1,210 @@
+@import "@automattic/calypso-color-schemes";
+@import "~@automattic/color-studio/dist/color-variables";
+
+$breakpoint-mobile: 782px; //Mobile size.
```
renancarvalho, line 4: This was extracted from view.scss, looks like a lot of changes but it is not.

escapemanuele (approved):
Cool! Let's ship this 👍

PR #76836, opened 2023-05-11
BlazePress search campaign / post endpoint

jjolmo:
Related to #

## Proposed Changes

* Added new blaze/posts paginated endpoint
* Applied infinite scrolling with server pagination
* Wired the search bar and filters in both campaigns and posts list
* revamped the main page and reduced the code to include the new GET endpoint
* Added missing styles and Skeleton loader for entries

## Testing Instructions
- Add this diff in your wpcom sandbox D109592
- checkout this branch (https://github.tumblr.net/Tumblr/a8c-dsp/pull/91) in the DSP server 
- Go to /advertising page in calypso
- You should see all the possible posts to be blazed from the current selected blog. If the amount is big enough (20 items) it will add more entries as you scroll. You can change this number at the wpcom diff
- You should be able to search by post title and order by different options in the search bar
- Go to campaigns tab. You should see all the user campaigns
- You should be able to search campaigns and filter them by status

## Pre-merge Checklist

image

You won't see images in the campaign list since there's no image serving in your locale system when running the DSP server

image

There's some nice algorithm added at the wpcom side to get the first canonical image from a post. It works quite good!


client/data/promote-post/use-promote-post-campaigns-query-paged.ts
```diff
@@ -0,0 +1,140 @@
...
+	OSs: 'OSs',
+};
+
+export type AudienceList = {
+	[ key in keyof typeof AudienceListKeys ]: string;
+};
+
+export type Campaign = {
+	content_config: {
+		clickUrl: string;
```
sbarbosa, line 25: It would be nice if we could unify the format for these fields. The rest of the fields use the snake case format, expect for the two that use camel case.

client/data/promote-post/use-promote-post-posts-query-paged.ts
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if ( searchOptions.order ) {
+		searchQueryParams += `&order_by=${ searchOptions.order.orderBy }`;
+		searchQueryParams += `&order=${ searchOptions.order.order }`;
+	}
+
+	return searchQueryParams;
+};
+
+async function queryPosts( siteId: number, queryparams: string ) {
```
sbarbosa, line 28: Nice!, now we have the possibility to separate this request from the rest of the WP requests and have a specific endpoint in our package. 

In Jetpack, all requests but the one that the DSP does are redirected to the stats packages (by default), but with this, I can just redirect it to the blaze package without affecting other calls.

client/my-sites/promote-post-i2/components/campaigns-table/index.tsx
```diff
@@ -24,9 +46,27 @@ export default function CampaignsTable( props: Props ) {
 					</tr>
 				</thead>
 				<tbody>
-					{ campaigns.map( ( campaign ) => {
-						return <CampaignItem key={ `item-${ campaign.campaign_id }` } campaign={ campaign } />;
-					} ) }
+					{ isLoading && ! isFetchingPageResults ? (
+						<>
+							<CampaignItemLoading />
```
sbarbosa, line 51: Not in this PR, but we need to iterate on this idea. It could be nice to have the same number of posts from the last call (with a maximum) as loading when performing another filter.

Right now, if you have less than 4 posts, the experience is a little strange, you see that the page gains a vertical scroll and then comes back only with a single post. 

But its better to start a discussion with Bart to see how he wants to do it.

client/my-sites/promote-post-i2/components/posts-list/index.tsx
```diff
@@ -1,55 +1,112 @@
...
+						<EmptyContent
+							className="promote-post-i2__empty-content"
+							title={ translate( 'You have no posts or pages.' ) }
+							line={ translate(
+								"Start by creating a post or a page and start promoting it once it's ready"
+							) }
+							illustration={ null }
+						/>
+					) : (
+						<>{ translate( 'No posts match your search' ) }</>
```
sbarbosa, line 94: This message doesn't take the correct margin, and it appears stuck to the border. We can treat these small design details in a future PR if you want. Let me know, and I can create a ticket for this.

Screenshot 2023-05-11 at 18 45 55

sbarbosa (approved):
LGTM 👍

Tested it and worked correctly.
I found a couple of small design details that can be treated on a future PR (this feature is still under a FF).

client/data/promote-post/use-promote-post-campaigns-query-paged.ts
jjolmo: Count with it! This information will be used in the new campaign individual page. I'll normalize the fields there!

client/data/promote-post/use-promote-post-posts-query-paged.ts
jjolmo: Would be great if I could get this endpoint tested in jetpack directly. I've also improved the wpcom endpoint to give more compatibility .
If there's something you might need to explore that could ease the jetpack integration, please tell me.

client/my-sites/promote-post-i2/components/campaigns-table/index.tsx
jjolmo: Sounds nice, just to give some ideas:

Since it's a loader, doing calls to any specific service would fail the purpose of adding this loader, but we might be able to get an approximate number of posts from the calypso state, and if we don't have it, just default to a number of rows.

Anyways, faking the data with skeletons usually creates this unusual situation.

client/my-sites/promote-post-i2/components/posts-list/index.tsx
jjolmo: whoops. I applied margins at the last step. I'll fix it

client/data/promote-post/use-promote-post-posts-query-paged.ts
jeherve: > Would be great if I could get this endpoint tested in jetpack directly

You should be able to test the endpoint with this PR, as long as you have Jetpack sites connected to your WordPress.com account at the moment.

Another way would be to add a code snippet to your Jetpack site to make queries to the endpoint from there directly, and log the results to see what you get. Something like this:

```php
/**
 * Test Blaze's Posts endpoint
 *
 * @see D109592-code
 *
 * @return void
 */
function jeherve_test_blaze_endpoints() {
	$site_id  = \Automattic\Jetpack\Connection\Manager::get_site_id();
	$url      = sprintf(
		'/sites/%1$d/blaze/posts?title=%2$s',
		$site_id,
		'A new post about things' // Change for any post or any parameter you want to test on your site.
	);
	$response = \Automattic\Jetpack\Connection\Client::wpcom_json_api_request_as_blog(
		$url,
		'2',
		array( 'method' => 'GET' ),
		null,
		'wpcom'
	);
	$result   = json_decode( wp_remote_retrieve_body( $response ), true );

	error_log( print_r( $result, true ) );
}
add_action( 'init', 'jeherve_test_blaze_endpoints' );
```

jjolmo:
Let's merge this!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8008425

Thank you @jjolmo for including a screenshot in the description! This is really helpful for our translators.

client/my-sites/promote-post-i2/components/empty-promotion-list/index.tsx
```diff
@@ -1,51 +1,40 @@
...
-		body: __( "Start by creating a post or a page and start promoting it once it's ready." ),
-	},
-};
+	let title;
+	let subtitle;
+
+	if ( type === 'campaigns' ) {
+		title = translate( 'You have no campaigns' );
+		subtitle = translate(
+			'You have not created any campaigns Click {{learnMoreLink}}Promote{{/learnMoreLink}} to get started.',
```
emilyaudela, line 20: There is a missing period after the first sentence.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8008425).

PR #76853, opened 2023-05-11
Move product logos to components package

ddc22:
Related to #76442

## Proposed Changes

* Adds logos added for the new Plans2023 grid to Automattic packages and adds a demo of said logos.

image

Caveat: There seem to be several other logos scattered throughout the codebase, however they are not scoped into this change. The main motivation for this change is to decouple the plan comparison grid so that it can be moved into a package as laid out in #76442

## Testing Instructions
- There shouldn't' be any visible changes from production, except for the new Logo section.
-  Go through the plans step `/start/plans` of the sign up flow and make sure everything works in the plans step and the all logos are displayed properly.
- Go to the `/plans` page and make sure of the same

*

## Pre-merge Checklist


packages/components/src/index.ts
```diff
@@ -25,7 +25,6 @@ export { SpinnerExample } from './spinner/example';
 export { GMClosureNotice } from './gm-closure-notice/gm-closure-notice';
 export { default as WordPressLogo } from './wordpress-logo';
```
chriskmnds: NIT - Should we move`WordPressLogo` along into the logos folder? also for `WooCommerceWooLogo` below

packages/components/src/logos/bloomberg-logo/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+		</g>
+	</svg>
+);
+export { BloombergLogo };
```
chriskmnds: NIT - why not a default export, and name it from the index file, as it was done with the previous ones?

My preference would also be for default exports foremost, generally.

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -20,33 +20,29 @@ import {
...
-import DisneyLogo from 'calypso/assets/images/onboarding/disney-logo.svg';
-import FacebookLogo from 'calypso/assets/images/onboarding/facebook-logo.svg';
-import SalesforceLogo from 'calypso/assets/images/onboarding/salesforce-logo.svg';
-import SlackLogo from 'calypso/assets/images/onboarding/slack-logo.svg';
-import TimeLogo from 'calypso/assets/images/onboarding/time-logo.svg';
-import vipLogo from 'calypso/assets/images/onboarding/vip-logo.svg';
-import wooLogo from 'calypso/assets/images/onboarding/woo-logo.svg';
 import QueryActivePromotions from 'calypso/components/data/query-active-promotions';
 import FoldableCard from 'calypso/components/foldable-card';
-import JetpackLogo from 'calypso/components/jetpack-logo';
```
chriskmnds, line 49: Comparison grid probably needs updating [here](https://github.com/Automattic/wp-calypso/blob/update/move-logos-to-components/client/my-sites/plan-features-2023-grid/components/plan-comparison-grid.tsx#L23).

packages/components/src/logos/facebook-logo/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+		xmlns="http://www.w3.org/2000/svg"
+		{ ...props }
+	>
+		<title>WordPress VIP client logo for Facebook</title>
+		<mask id="mask0_233_14595" maskUnits="userSpaceOnUse" x="0" y="10" width="77" height="9">
+			<path d="M76.9837 10H0.5V18.1053H76.9837V10Z" fill="white" />
+		</mask>
+		<g mask="url(#mask0_233_14595)">
+			<path
+				fill-rule="evenodd"
```
chriskmnds: This is throwing a warning in the console. I think this should be `fillRule`. also for a few other cases below

packages/components/src/logos/facebook-logo/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+				clip-rule="evenodd"
```
chriskmnds: Same. `clip-rule` -> `clipRule`. a few more case below

packages/components/src/logos/vip-logo/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+		width="38"
+		height="17"
+		viewBox="0 0 38 17"
+		fill="none"
+		xmlns="http://www.w3.org/2000/svg"
+		{ ...props }
+	>
+		<title>WPVIP logo</title>
+
+		<g clip-path="url(#clip0_233_14550)">
```
chriskmnds: Similar. `clip-path` -> `clipPath`

chriskmnds (commented):
Looks good. Thanks for jumping in on this and doing a great refactor. I left some comments, but nothing critical it seems. It tests out and the docs show these as outlined.

southp:
What's the advantages of embedding SVGs directly into scripts, instead of being individual files? The downside is that a non-dev can no longer easily access these logos just using a image viewer anymore. Since we are just wrapping them into components here, it looks like it doesn't do much outside of what an SVG tag already can, so I'm curious ot learn :)

ddc22:
@southp I guess the upside is basically performance. SVGs are simple DOM elements that just have to be rendered immediately. An image will have to be requested externally and rendered there after. I can acknowledge though that this is not a huge gain in performance for the average case.

> The downside is that a non-dev can no longer easily access these logos just using a image viewer anymore

I was not aware of a use case like this in this area do you think we should cater to this requirement even with devdocs being readily available? 🤔

southp:
Thanks for the clarification. Whether it's a performance gain or not would largely depend on the context. When the bandwidth of downloading one single document isn't an issue, embedding them will likely be a performance gain. However, when the bandwidth is more restricted, browsers can still render the document by using standard placeholders for these images while downloading them. If there is a case that we use the same image for multiple times in one page, SVGs will be downloaded once and reused, but I'm not sure if react nowadays would handle this case :) 

Saying all that, I'm not against that we give this a go. Let's keep it this way if there is no other concern :) 

> I was not aware of a use case like this in this area do you think we should cater to this requirement even with devdocs being readily available? 🤔

Good question. Again, let's keep it this way and see how it goes –– we can point them to the devdocs when needed and ask them how they feel :)

packages/components/src/logos/bloomberg-logo/index.tsx
ddc22: I agree to conform with the existing pattern. But my preference has always been for named exports 😬. This is primarily because we retain naming consistency of the export across modules. And when we want to refer to an export with an alias it's always explicit. Is it just a preference on your side for default exports?

ddc22:
@southp good point 🙌 . Definitely depends on the context for performance. Let's keep a lookout for any DX or performance bottlenecks, and I am happy to iterate on this 👍

packages/components/src/logos/bloomberg-logo/index.tsx
chriskmnds: > This is primarily because we retain naming consistency of the export across modules.

I think that's a hard rule to follow and would only really make sense if we actually forbit default exports. Otherwise, what is the real value of it - can it be trusted in search, for instance? _Maybe_ in a small code base you can depend on such, but I doubt that can be the case in larger projects. 

I personally seldom (never?) search the code base by module name. It's not something that I can trust in a realistic scenario. I'd typically either "find references" via the editor (99% of the time) or search by file name. So whether a module is renamed across files doesn't come into my experience much. I think!

> Is it just a preference on your side for default exports?

I don't think I've come up with this myself, probably something that I've grown into. I believe we also have de facto patterns at places (state selectors I think?) to have a single default export. My preference would be a default export for the "main" functionality you are exporting from a file, and named exports for anything else that cannot live in a separate file. Something along these lines. :-)

chriskmnds: and if we all switch to something else, more restrictive, I'll adapt :D 

Consistency is key. like we talked about last time. even unorthodox patterns can be ok if done consistently

ddc22:
@chriskmnds can you give me approval if you feel like this is good to go 😄

chriskmnds:
Tests out as outlined in signup/admin. Thanks for working on this. 😎 

I am not entirely sure why, but the `ProductLogoExample` doesn't show up on `/devdocs/design` in the live image + my local instance is now also broken when visiting that (but not sure if related). I searched in Slack really quick and found what looks like relevant convo (p1671091105143639-slack-C7YPUHBB2). It looks like we may need to add a static `displayName` property to the example for it to show e.g. 

`ProductLogoExample.displayName = 'ProductLogo';`

and also bump the version of `@automattic/components` (along with a changelog)? Can we confirm though with someone from the Calypso components team if we need to do this?

(but weird, definitely had this working last time in local calypso 😕 )

> What's the advantages of embedding SVGs directly into scripts, instead of being individual files?

Good point @southp! I think it's becoming common to do it via the React component. I can see the raw SVG form though being more accessible to someone/designer who wants to load in editor and redo/replace if need be (vs having to go through the React wrapper)? Or maybe reuse it more easily outside of React. Other than that, they can inspect the file in `/devdocs` (assuming one exists)

chriskmnds:
> @chriskmnds can you give me approval if you feel like this is good to go 😄

caught you to it https://github.com/Automattic/wp-calypso/pull/76853#issuecomment-1562817927 😆

ddc22:
This was fixed by  #77401, @chriskmnds can you have another look?

chriskmnds:
> This was fixed by #77401, @chriskmnds can you have another look?

Still don't see it? Shouldn't these show up here? https://container-distracted-cartwright.calypso.live/devdocs/design

chriskmnds (approved):
Great job! LGTM

(still curious about whether we need to bump the `@automattic/components` package version and a changelog entry 🤔 )

ddc22:
@chriskmnds we don't seem to bump the change log entry that often. Happy to bump this if a requirement arises.

image

PR #76854, opened 2023-05-11
Hosting onboarding: add site options step to `/setup/new-hosted-site`

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2360.

## Proposed Changes

This PR adds the data center picker, along with a site title suggestion, to the new `/hosting` onboarding flow. See the issue for details.

image

## Testing Instructions

1. Patch your sandbox with this diff: D110494-code;
2. Checkout this branch;
3. Browse `/setup/new-hosted-site`;
4. Verify that the site title gets auto-populated with a suggestion;
5. Verify that the recommended data picker is selected, but you can pick a custom one;
6. Buy the plan;
7. Verify that the site title is the same as the one suggested in the first step and that the domain slug is related to the suggestion;
8. Verify that the picked data center was communicated correctly*.

*You can use this snippet to log the picked data center right after checkout. Add this to `0-sandbox` **before pressing the Pay button**:

```php
function sandbox_activate_product( $blog_id, $user_id, $product_id, $txn_id, $verb, $meta, $volume, $extra ) {
	l( $extra );
}

add_action( 'activate_product', 'sandbox_activate_product', 9, 8 );
```

And you should see `{ extra: { geo_affinity: 'chosen_data_center' } }` on your sandbox logs.


javierarce:
Could we hide the dropdown when the optimal data center is selected? We don't need to show it by default and it would make this screen look nicer.

image

mpkelly (approved):
It works as advertised. Code looks fine although the conditional logic is a bit complex in places. 

I agree that removing illustrations feels like a separate PR unrelated to this one.

client/landing/stepper/declarative-flow/internals/steps-repository/plans/index.tsx
```diff
@@ -27,13 +35,13 @@ const plans: Step = function Plans( { navigation, flow } ) {
...
 			return __( 'Back' );
 		}
 	};
 
 	const shouldHideBackButton = () => {
-		if ( flow === DOMAIN_UPSELL_FLOW ) {
+		if ( flow === DOMAIN_UPSELL_FLOW || isHostingSiteCreationFlow( flow ) ) {
 			return false;
 		}
 		return true;
```
p-jackson: Would it be clearer expressed like this?

```suggestion
		return flow !== DOMAIN_UPSELL_FLOW && ! isHostingSiteCreationFlow( flow );
```

client/landing/stepper/declarative-flow/internals/steps-repository/plans/index.tsx
```diff
@@ -9,8 +15,10 @@ const plans: Step = function Plans( { navigation, flow } ) {
 	const { submit } = navigation;
 	const { __ } = useI18n();
 
-	const handleSubmit = () => {
-		const providedDependencies: ProvidedDependencies = {};
+	const handleSubmit = ( plan: MinimalRequestCartProduct | null ) => {
+		const providedDependencies: ProvidedDependencies = {
+			...plan,
+		};
```
p-jackson: Do the plan properties really want to be spread into the dependencies object? I know it's unlikely, but imagine if someone added a `goToCheckout` prop to the plan object. We'd end up overriding it with our own `goToCheckout flag.

It feels like the values submitted by a step should be distinct from the plans's own properties. i.e.

```suggestion
		const providedDependencies: ProvidedDependencies = { plan };
```

client/landing/stepper/declarative-flow/internals/steps-repository/site-options/style.scss
```diff
@@ -149,3 +149,50 @@
...
+
+	.data-center-picker__options {
+		font-size: $font-body-small;
+
+		.form-select {
+			width: fit-content;
+			border-radius: 4px;
+			font-size: $font-body-small;
+		}
+	}
```
p-jackson: It feels like some of these styles belong in the `client/blocks/data-center-picker` code, rather than in the step repository. I understand that some of the theming we want to do in the stepper framework, but wouldn't some of the rules (e.g. `.form-select` and the wrapping rule) go in the component itself?

In fact there seems to be something wrong with the wrapping rule. It's not allowing the description to fit correctly on small screens.

wrap

client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
```diff
@@ -60,6 +84,20 @@ const hosting: Flow = {
...
 		return { submit };
 	},
+	useSideEffect( currentStepSlug ) {
+		const { resetOnboardStore } = useDispatch( ONBOARD_STORE );
+
+		useEffect(
+			() => {
+				if ( currentStepSlug === 'options' ) {
+					resetOnboardStore();
+				}
```
p-jackson: This does something when the user navigates directly to `/setup/new-hosted-site/options`, but not when the user goes to `/setup/new-hosted-site` (which is the usual URL we use to start a flow). That's because `currentStepSlug === undefined`.

In my experience, deciding when is the best time to reset the flow is one of the trickiest parts of the onboarding process. No easy answers 🤷 Reseting the flow for each fresh load of the `options` step seems a good place to start though.

The `free` flow does it [for every page load](https://github.com/Automattic/wp-calypso/blob/51fb46a0166b32b40ef986bc5d898b13775c3233/client/landing/stepper/declarative-flow/free.ts#L38) but that seems too aggressive for me. If the user refreshes on the plans step we should retain their geo affinity etc.

p-jackson (commented):
### Plans back button not working

back

This button doesn't work. That's because `plans` step's `handleGoBack()` function is special cased to only do something for the `DOMAIN_UPSELL_FLOW`. @alshakero now that all navigation is modelled as "submitting" a step, should there be some generic understanding for what it means to go back? e.g. `submit?.( { back: true } );`
Without it, every flow is going to have to customise the plan step with it's own bespoke way of going backwards. I don't _think_ it'll break anything if we introduce it now, since we won't be introducing a new code path for flows where the back button is hidden, and basically all flows hide the back button.

### Using the site options step

Just a thought that the site options step seems to be getting pretty darn big. There's so many different modes and permutations, and it makes me wonder whether it's cleaner and easier to test if we just copy-paste the existing site options step i.e. have a new step that only includes the two options we care about.
The down side is we'd need to come up with our own slug.

p-jackson (commented):
### Plans back button not working

back

This button doesn't work. That's because `plans` step's `handleGoBack()` function is special cased to only do something for the `DOMAIN_UPSELL_FLOW`. @alshakero now that all navigation is modelled as "submitting" a step, should there be some generic understanding for what it means to go back? e.g. `submit?.( { back: true } );`
Without it, every flow is going to have to customise the plan step with it's own bespoke way of going backwards. I don't _think_ it'll break anything if we introduce it now, since we won't be introducing a new code path for flows where the back button is hidden, and basically all flows hide the back button.

### Using the site options step

Just a thought that the site options step seems to be getting pretty darn big. There's so many different modes and permutations, and it makes me wonder whether it's cleaner and easier to test if we just copy-paste the existing site options step i.e. have a new step that only includes the two options we care about.
The down side is we'd need to come up with our own slug.

alshakero:
Thanks for the ping!

> This button doesn't work. That's because plans step's handleGoBack() function is special cased to only do something for the DOMAIN_UPSELL_FLOW. @alshakero now that all navigation is modelled as "submitting" a step, should there be some generic understanding for what it means to go back? e.g. `submit?.( { back: true } )`;

Yes, that sounds like a good idea. Although I feel it's like I'd be tricking myself around not using `goBack`. Maybe I'm mistaken and `goBack` is a legitimate need, and we shouldn't depreciate it? `goBack` isn't too bad because it doesn't make the step less reusable and doesn't leak information about the flow into the step.

We can just remove [this](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/types.ts#L20-L22) if you agree. WDYT?

zaguiini:
Thanks for all the feedback, @p-jackson! Please let's give this PR another round of review.

>We can just remove [this](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/types.ts#L20-L22) if you agree. WDYT?

Sounds good @alshakero. To not mix concerns, it'd make more sense to do this on a separate PR.

p-jackson:
> Maybe I'm mistaken and `goBack` is a legitimate need, and we shouldn't depreciate it? `goBack` isn't too bad because it doesn't make the step less reusable and doesn't leak information about the flow into the step.

@alshakero Fair enough, insisting on `submit( { back: true } )` over `goBack()` is probably just forcing people to jump through hoops when they're obviously the same thing 😄 

FWIW, the thing I like about only having `submit()` is that it forces everyone to think of their `useStepNavigation` hook as a finite-state machine where going back isn't anything special. So it was a good proposal.

javierarce:
Nice work! I noticed a few areas for improvement:

- We should remove the vertical scrolling on the page.
- In other onboarding flows, we use a different pattern for required steps. For instance, [in the newsletter setup](https://wordpress.com/setup/newsletter/newsletterSetup), the submit button remains active, but we display an error message if the user fails to provide a newsletter name. We also follow this pattern in the importing stage of other onboardings. To maintain consistency, let's enable the continue button and show an error message if the user attempts to proceed without a title.
- It could be related to using Calypso live for testing, but the continue button's background color is pink instead of blue.
- The labels' color is too dark. In other onboardings, we use color: `var(--studio-gray-60);`.
- The non-selected state of the [SelectCard component](https://www.figma.com/file/eqlepL7MHifK8Bgr1OWqLr/Dotcom-Design-Library?type=design&node-id=1858-1994&t=Slxg49xw3fWg53FN-4) should have a `1px` border-width instead of `2px`. The description's color should be `color: var(--studio-gray-60);`.
- [The icon](https://www.figma.com/file/eqlepL7MHifK8Bgr1OWqLr/Dotcom-Design-Library?type=design&node-id=47-2902&t=Slxg49xw3fWg53FN-4) in the form-setting-explanation should be `20px`, not `24px`. Additionally, let's increase the paragraph's top margin from `6px` to `8px`.

alshakero:
> forces everyone to think of their useStepNavigation hook as a finite-state machine

Genuinely smiled wide. That's precisely my intention.

zaguiini:
>It could be related to using Calypso live for testing, but the continue button's background color is pink instead of blue.

That's because of the theme used in Calypso. We're using the primary color for the button.

>[The icon](https://www.figma.com/file/eqlepL7MHifK8Bgr1OWqLr/Dotcom-Design-Library?type=design&node-id=47-2902&t=Slxg49xw3fWg53FN-4) in the form-setting-explanation should be 20px, not 24px. Additionally, let's increase the paragraph's top margin from 6px to 8px.

That'd go against the current components, creating inconsistencies. I'll go with the existing patterns for now.

p-jackson (approved):
Changes are working well. I like the refactor to the options step that allows multiple flows to re-use the nice `options` step name but without having to merge all of the logic into one giant uber step.

I've tested the `videopress` and `site-setup` steps to ensure they're still working as before. No regressions that I could find. There was a change to the behaviour of the `/setup/videopress/options` step in that the "Continue" button is now always enabled rather than waiting for users to enter a valid site title. But this is inline with @javierarce's suggestion above to make the various flows consistent, so I think it's a good change.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7994852

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7994852).

PR #76878, opened 2023-05-12
Add-ons: Filter Storage Add On Options

sharpirate:
This PR is based on https://github.com/Automattic/wp-calypso/pull/76754
Fixes 1721-gh-Automattic/martech

### Proposed Changes
Implement filtering for the available storage options based on the current max storage:

1. We want to offer 50 and 100 GB for now (this will likely change in the future)
2. A site should not have more than 200 GB of storage. If a user already has 200 GB of storage (e.g. provided by a plan), they should not have the option to add more storage.
3. Potentially, a user may be eligible for 50 GB, not 100 GB, depending on their current allowed storage.

### Testing
1. Checkout this branch
2. Run `yarn` & `yarn start`
3. Go to `http://calypso.localhost:3000/add-ons/YOUR_SITE_SLUG`
4. You should see the new card item for the storage add on:

image

The values in the dropdown should be dynamic depending on the current storage usage for a site.


client/my-sites/add-ons/hooks/use-add-ons.ts
```diff
@@ -19,10 +22,55 @@ export interface AddOnMeta {
...
 	name: string | React.ReactChild | null;
+	quantity?: number;
 	description: string | React.ReactChild | null;
 	displayCost: string | React.ReactChild | null;
+	variations?: Partial< AddOnMeta >[];
 }
 
+const AVAILABLE_PRODUCTS = [ 50, 100 ];
+
+const buildStorageAddOn = (
```
sharpirate: This method filters available options based on the current storage limit for a site. The implementation is using mocked data for now, I'll be updating the method to use real data from `useMediaStorageQuery `

gmovr:
Sorry for the review bomb ... I tried to change the base branch to make it easier to review, but I realized they are not branched off the same trunk commit. 🤦

gmovr:
We also need to update the icon :) 

@saygunnyc  -- Do you have an icon we can use for the space add-on?

sharpirate:
Closing this in favor of https://github.com/Automattic/wp-calypso/pull/77387

PR #76891, opened 2023-05-12
Reader: Update related posts to pass variable size param

eoigal:
This PR is some ground work for the suggested follows project - https://github.com/Automattic/wp-calypso/pull/76812

We intend to use the related post endpoint to help us figure out suggested follows. At the moment, calypso only requests 2 related posts. This PR will make that variable so we can request more.

There should be no change to behaviour with this PR.

### Test
Go to a Reader post and confirm related posts still shows 2 related posts;

http://calypso.localhost:3000/read/feeds/94606731/posts/4701442242

Screenshot 2023-05-12 at 17 03 22

Project Ref: pe7F0s-O2-p2


davemart-in (approved):
Tested https://container-boring-einstein.calypso.live/read/feeds/94606731/posts/4701442242 

Works as expected. No JS errors. Ship it!

PR #75730, opened 2023-04-13
Add reader tag sidebar

eoigal:
As part of the public tags v3 project, we are adding a new sidebar to tag pages;

screenshot-2023-03-16-at-3 26 14-pm

This PR will add this sidebar.

This PR makes use of an existing wpcom endpoint `WPCOM_REST_API_V2_Endpoint_Read_Tags_Cards`.

This endpoint returns 4 related tags and 2 related sites based on a tag.

We can update this endpoint to support returning a variable amount of results so we can show more results in sidebar.

The PR does not include the Post or Author counts. There isn't a count total available from this endpoint so not rendering this until we have this in place. I don't think this is a blocker to getting this sidebar done.

PR #76899, opened 2023-05-13
Reader tag pages: update column styles

creativecoder:
Related to https://github.com/Automattic/wp-calypso/pull/75730

## Proposed Changes

Updates the column styles on the newly designed reader tag pages to better match the design.

Additional markup changes to support the design include:

- Adding a `streamHeader` prop to `ReaderStream`, so a custom header can be added at the top of the main column in the two-column view, rather than above both columns.
- Modifies `ReaderStream` so the `.stream__two-column` class is only present when two columns are displayed side-by-side, for styles that only apply to the two column layout.
- Adds the `.tag-stream__main` class to the `main` element on the tag page to more easily customize the tag page (it appears this was present before, but removed at some point).

image

## Testing Instructions

* Visit `/tag/{tag-slug}` and compare with designs (pe7F0s-IS-p2)
* Check both large (two column) and small screens (tabbed layout)
* Spot check other Reader pages: Following, Search, etc to check for regressions

### DIFF
--- a/client/reader/stream/style.scss
+++ b/client/reader/stream/style.scss
@@ -584,16 +584,12 @@
 	align-items: flex-start;
 	display: flex;
 	flex-wrap: nowrap;
+	gap: 80px;
 	justify-content: flex-start;
+}
 
-	.reader__content {
-		padding-right: 56px;
-		position: relative;
-	}
-
-	stream__right-column {
-		padding-left: 24px;
-	}
+.reader__content {
+	position: relative;
 }
 
 .stream__right-column {
--- a/client/reader/stream/style.scss
+++ b/client/reader/stream/style.scss
@@ -1,7 +1,8 @@
 // Reader-specific tweaks to components
 
 // Overrides default 720px width
-.is-reader-page .following.main {
+.is-reader-page .following.main,
+.is-reader-page .tag-stream__main.main {
 	max-width: 600px;
 
 	&.reader-two-column {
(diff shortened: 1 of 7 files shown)


client/reader/stream/style.scss
```diff
@@ -655,98 +652,93 @@
 		line-height: 18px;
 	}
 
-	.stream__right-column {
```
creativecoder, line 658: At this point and below for this file, there are no css property changes.

Several blocks are being moved from `.stream__two-column .stream__right-column` to `.stream__right-column` so the styles apply to the right column on both two column and tabbed views.

Addison-Stavlo (commented):
This is looking pretty good! 

In testing the first thing I noticed was the position of the 'back' button in mobile view gets uncomfortably close to the title. However, this seems to be pre-existing and can probably be addressed separately: 

Screenshot 2023-05-15 at 8 45 44 AM

<hr/> 

> Spot check other Reader pages: Following, Search, etc to check for regressions 


WRT this, i found a few differences between running this branch and production. Most seem pretty minor, but noting them blow:

1. The images seem to extend outside the column slightly on this branch, both for dual-column and tab views of the search page. 

reader-search-columns

2. Items in the right column are not concatenated as early / at as short of a length (this seems like an improvement!). Its especially noticeable on my long titled test sites:

less-concat

3. There seems to be a different font size/colors for subheading links on the likes (and conversations) pages:


fonts-diff

creativecoder:
@Addison-Stavlo Thanks for the detailed testing/review!

> the 'back' button in mobile view gets uncomfortably close to the title.

I went ahead and removed the repositioning styles for various breakpoints--it seems to look fine with the default positioning.

> The images seem to extend outside the column slightly on this branch, both for dual-column and tab views of the search page.

You're right--there was padding and margin inside the container that doesn't seem necessary; the design looks to be flush with the column. So I've removed it.

> Items in the right column are not concatenated as early / at as short of a length

Yes, that was intentional!

> There seems to be a different font size/colors for subheading links on the likes (and conversations) pages

Thanks for catching that! I missed one place that needed the `reader__content` class now that I moved it one level above the `InfiniteList` component. Should now be resolved.

---

I'd appreciate another look when you have a moment.

PR #76919, opened 2023-05-15
Add/post publish modal v2

roo2:
re-adds the post publish modal but sets it to disabled so that translation can be started on translate.wordpress.org

see pxLjZ-6mM-p2 and p1683059619368299-slack-C03NLNTPZ2T


apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/block-editor-nux.js
```diff
@@ -104,7 +105,9 @@ registerPlugin( 'wpcom-block-editor-nux', {
 			<HasSeenVideoCelebrationModalProvider>
 				<ShouldShowFirstPostPublishedModalProvider>
 					<WelcomeTour />
-					<PostPublishedModal />
+					<FirstPostPublishedModal />
+					{ /* todo: enable once translations are complete */ }
+					{ /* <PostPublishedSharingModal /> */ }
```
simison, line 110: You could always check for locale and release for English. :-)

Smaller release gives you some heads ups on what kind of feedback you get, too.

PR #68417, opened 2022-09-28
Signup: Allow theme slug as a default dependency in plans step

daledupreez:
#### Proposed Changes

* This PR adds support for the plans step to specify `themeSlugWithRepo` as a default dependency, which is accomplished by adding some logic to the `isPlanFulfilled()` function to add `defaultDependencies?.themeSlugWithRepo` to the submitted step data if the value is present
* This PR also specifies `themeSlugWithRepo` to be `pub/twentytwentytwo` for the `ecommerce` and `ecommerce-monthly` flows, where the plans step is not actually shown to users, so the changes from https://github.com/Automattic/wp-calypso/pull/68411 don't apply, because the `onSelectPlan()` callback is never triggered

#### Testing Instructions

* Open the live testing branch and navigate to `/start/ecommerce` to start the `ecommerce` signup flow (this is normally triggered from the [Ecommerce landing page](https://wordpress.com/ecommerce/) )
* Work through the flow and verify that your site is created with the Twenty Twenty Two theme
* Repeat the test above starting from `/start/ecommerce-monthly`, and verify that you also get a site created with the Twenty Twenty Two theme.
* Navigate to `/start` and work through the main signup flow - verify that there are no changes to any of those steps. (There shouldn't be, as the only steps that have added `themeSlugWithRepo` to `defaultDependencies` are the two I modified!)

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Related to #67562

PR #68411, opened 2022-09-28
Initial theme for ecommerce sites: twentytwentytwo (Redux)

mreishus:
#### Proposed Changes

* New ecommerce sites will now have their initial theme set separately compared to the rest of sites.
* That Initial theme for ecommerce sites will be twentytwentytwo.

#### Reimplementation

* I did not deploy #68164 because it failed pre-release e2e tests:
  * `CALYPSO_BASE_URL=http://calypso.localhost:3000/ yarn jest specs/onboarding/ftme__write.ts`
  * `CALYPSO_BASE_URL=http://calypso.localhost:3000/ yarn jest specs/plans/plans__signup-free.ts`

It turns out the root cause of the failure is `isCommerce` throwing an exception when it is passed `null`. This meant that clicking on "free" to make a free site only threw an exception and did not let the user continue.

Scope limited fix:
```diff
diff --git a/client/signup/steps/plans/index.jsx b/client/signup/steps/plans/index.jsx
index 3b9b0af44c..a7dae5af70 100644
--- a/client/signup/steps/plans/index.jsx
+++ b/client/signup/steps/plans/index.jsx
@@ -129,7 +129,7 @@ export class PlansStep extends Component {
                        const signupVals = { cartItem };
 
                        // Buying an eCommerce plan defaults to the pub/twentytwentytwo theme (All remaining flows)
-                       if ( isEcommerce( cartItem ) ) {
+                       if ( cartItem && isEcommerce( cartItem ) ) {
                                signupVals.themeSlugWithRepo = 'pub/twentytwentytwo';
                        }
 ```

### Testing Instructions

* Use `/start` to create a new ecommerce site. It should have the 2022 theme.
* Use `/start` to create other sites, not using ecommerce. They should continue to have zoologist.
* Use `/start` to create free sites. This should continue to work.
* Anything else you can think of.

Related to #67562
Related to #68164

PR #76921, opened 2023-05-15
Add themeSlugWithRepo dep to plans-site-selected

jsnajdr:
I got a Sentry notification that a JS error is being reported in the wild:
```
This step (plans-site-selected) provides an unspecified dependency [themeSlugWithRepo]. Make sure to specify it in /signup/config/steps-pure.js, using the providesDependencies property.
```
I was notified because I migrated the Signup code from Flux to Redux 4 years ago, and I am not familiar what's been done in Signup since then, but I think I was able to figure it out anyway:

In #68411 and #68417 @daledupreez and @mreishus merged a change that's adding a `themeSlugWithRepo` dep to several plans-selection steps, because the plans selection UI can specify a theme after selecting an eCommerce plan. But the `plans-site-selected` step didn't get the new dependency. And it seems it needs it.

### DIFF
--- a/client/signup/config/steps-pure.js
+++ b/client/signup/config/steps-pure.js
@@ -101,7 +101,7 @@ export function generateSteps( {
 			stepName: 'plans-site-selected',
 			apiRequestFunction: addPlanToCart,
 			dependencies: [ 'siteSlug' ],
-			providesDependencies: [ 'cartItem' ],
+			providesDependencies: [ 'cartItem', 'themeSlugWithRepo' ],
 			props: {
 				hideFreePlan: true,
 				hideEnterprisePlan: true,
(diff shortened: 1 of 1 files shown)


daledupreez (approved):
This looks like the right fix to me, based on the fact that both steps resolve to the underlying `plans` step per [this code](https://github.com/Automattic/wp-calypso/blob/d6a146f21f49ea67e5b7e0d54990b3c5f6d0bc3c/client/signup/config/step-components.js#L40-L41).

I've also looked through the code that sets this up ([here](https://github.com/Automattic/wp-calypso/blob/d6a146f21f49ea67e5b7e0d54990b3c5f6d0bc3c/client/lib/signup/step-actions/index.js#L1302-L1312), which seems to set `themeSlugWithRepo` for the step submission only when it has an incoming value. I _think_ it's safe to do this, as we're also doing this elsewhere.

PR #76923, opened 2023-05-15
Stats: Omit base URL on children for clicks/refers

### LINKED ISSUE #76922: Stats: Omit base domain for children click/referrer rows
Showing the base domain for child rows of Clicks/Referrers modules makes them hard to read:

Image

jsnmoon:
Fixes #76922.

## Proposed Changes

Before|After
-|-
image|image

The child label will no longer begin with the parent's base URL.

## Testing Instructions

* Open the live branch of this PR and navigate to the Stats Dashboard.
* Ensure that the child rows in the referrers module and clicks modules no longer begin with the parents' base URL.

## Pre-merge Checklist

### DIFF
--- a/client/state/stats/lists/utils.js
+++ b/client/state/stats/lists/utils.js
@@ -746,7 +746,7 @@ export const normalizers = {
 		const dataPath = query.summarize ? [ 'summary', 'clicks' ] : [ 'days', startOf, 'clicks' ];
 		const statsData = get( data, dataPath, [] );
 
-		return statsData.map( ( item ) => {
+		const output = statsData.map( ( item ) => {
 			const hasChildren = item.children && item.children.length > 0;
 			const newRecord = {
 				label: item.name,
@@ -760,7 +760,9 @@ export const normalizers = {
 			if ( item.children ) {
 				newRecord.children = item.children.map( ( child ) => {
 					return {
-						label: child.name,
+						// Remove the parent name from the child name.
+						// If the child name is the same as the parent name, use a slash instead.
+						label: child.name?.replace( item.name, '' ) || '/',
 						value: child.views,
 						children: null,
 						link: child.url,
@@ -771,6 +773,7 @@ export const normalizers = {
 
 			return newRecord;
 		} );
+		return output;
 	},
 
 	/*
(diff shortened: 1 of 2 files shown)


grzegorz-cp (approved):
👍

client/state/stats/lists/utils.js
```diff
@@ -793,7 +796,13 @@ export const normalizers = {
 		const parseItem = ( item ) => {
 			let children;
 			if ( item.children && item.children.length > 0 ) {
-				children = item.children.map( parseItem );
+				children = item.children.map( ( child ) => {
+					const parsed = parseItem( child );
+					// Remove the parent name from the child name.
+					// If the child name is the same as the parent name, use a slash instead.
+					parsed.label = child.name?.replace( item.name, '' ) || '/';
```
dognose24, line 803: In this case, the `item.name` should be a domain URL (e.g., whereareyoufrom.wordpress.com), so it works well. Could we use the regular expression to find the match at the start position of the string to ensure we do not accidentally remove the line of the rest of the URL?

```suggestion
					const regex = new RegExp( `^${ item.name }` );
					parsed.label = child.name?.replace( regex, '' ) || '/';
```

client/state/stats/lists/utils.js
```diff
@@ -760,7 +760,9 @@ export const normalizers = {
 			if ( item.children ) {
 				newRecord.children = item.children.map( ( child ) => {
 					return {
-						label: child.name,
+						// Remove the parent name from the child name.
+						// If the child name is the same as the parent name, use a slash instead.
+						label: child.name?.replace( item.name, '' ) || '/',
```
dognose24, line 765: Could we apply the exact regular expression replacement the same as suggestions for another change?

dognose24 (approved):
Works well! Left comments for the string check, which will not be the blocker. 🙂

client/state/stats/lists/utils.js
kangzj: I'm actually fine with Jason's approach as replace would only replace the first occurrence and all children are prefixed with that. However theoretically yours is safer, I'll go ahead and merge and let's see if we need to do that later.

kangzj (approved):

client/state/stats/lists/utils.js
dognose24: Sounds reasonable to me as well! 👍🏼

PR #76944, opened 2023-05-15
Subscription Management: Fix handling of invalid urls.

### LINKED ISSUE #76898: Subscription Management: JavaScript error when using link to login
### Quick summary

See screenshot below. When I start with an incognito window, go to `/subscriptions`, enter my a8c email, then paste the link in to this incognito window, I end up with JavaScript errors and a blank page.

Screenshot 2023-05-12 at 6 57 52 PM

### Steps to reproduce

1. Start with incognito window and logged out.
2. Enter in a8c email.
3. Go to email and get link.
4. Paste into incognito window.
5. Verify JS error.

No error.

### What actually happened

Error.

### Impact

Some. I can reproduce on my a8c account but not on my personal WPCOM account.

### Available workarounds?

No and the platform is unusable

Simple, Atomic, Self-hosted

### Logs or notes

Tested on Mac OS and Microsoft Edge.

yansern:
Fixes https://github.com/Automattic/wp-calypso/issues/76898

## Proposed Changes

Site subscription with `false` in the url property causes WSOD. The `false` value may come from purged Jetpack-connected site (even though it doesn't seem to apply to other purged Jetpack-connected sites I've seen).

The fix is done on client-side instead of server-side because the server-side API endpoint is also shared with Reader, and I'd like to avoid diving into Reader's code to see how they are handling falsy URLs (and if they are handling it).

## Testing Instructions

**Setup**
1. On your sandbox, modify `class[...]read-following-mine-v1-2-endpoint.php` at line 423, make it return `false`.
2. Sandbox `public-api.wordpress.com`.

**Test**
1. Test as external user by setting `subkey` on the `calypso.localhost:3000` cookie.
2. Go to `/subscriptions`.
3. It should render the list just fine instead of WSOD'ing.

Screenshot of how it looks like when rendered:

image

Error Reference:

image

image

## Pre-merge Checklist

### DIFF
--- a/client/landing/subscriptions/components/site-list/site-row.tsx
+++ b/client/landing/subscriptions/components/site-list/site-row.tsx
@@ -32,7 +32,7 @@ export default function SiteRow( {
 	date_subscribed,
 	delivery_methods,
 }: SiteSubscription ) {
-	const hostname = useMemo( () => new URL( url ).hostname, [ url ] );
+	const hostname = useMemo( () => ( url ? new URL( url ).hostname : '' ), [ url ] );
 	const siteIcon = useMemo( () => {
 		if ( site_icon ) {
 			return <img className="icon" src={ site_icon } alt={ name } />;
@@ -88,7 +88,12 @@ export default function SiteRow( {
 
 	return (
 		<li className="row" role="row">
-			<a href={ url } rel="noreferrer noopener" className="title-box" target="_blank">
+			<a
+				{ ...( url && { href: url } ) }
+				rel="noreferrer noopener"
+				className="title-box"
+				target="_blank"
+			>
 				<span className="title-box" role="cell">
 					{ siteIcon }
 					<span className="title-column">
--- a/client/landing/subscriptions/components/site-list/site-row.tsx
+++ b/client/landing/subscriptions/components/site-list/site-row.tsx
@@ -32,7 +32,13 @@ export default function SiteRow( {
 	date_subscribed,
 	delivery_methods,
 }: SiteSubscription ) {
-	const hostname = useMemo( () => ( url ? new URL( url ).hostname : '' ), [ url ] );
+	const hostname = useMemo( () => {
+		try {
+			return new URL( url ).hostname;
+		} catch ( e ) {
+			return '';
+		}
+	}, [ url ] );
 	const siteIcon = useMemo( () => {
 		if ( site_icon ) {
 			return <img className="icon" src={ site_icon } alt={ name } />;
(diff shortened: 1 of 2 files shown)


client/landing/subscriptions/components/site-list/site-row.tsx
```diff
@@ -32,7 +32,7 @@ export default function SiteRow( {
 	date_subscribed,
 	delivery_methods,
 }: SiteSubscription ) {
-	const hostname = useMemo( () => new URL( url ).hostname, [ url ] );
+	const hostname = useMemo( () => ( url ? new URL( url ).hostname : '' ), [ url ] );
```
mashikag: If we are not planning to change the API, then we should also correct our types, i.e. in `SiteSubscription` type we should say that it may have value of `boolean`, or even to be more precise `false`.

JuanLucha (approved):

client/landing/subscriptions/components/site-list/site-row.tsx
yansern: Done. https://github.com/Automattic/wp-calypso/pull/76944/commits/83a4cc9a04553fbf18502a258a10b8b5206096b5

yansern:
Relooking at this issue with fresher eyes, I'm going to use `try..catch` to handle more edge cases.

image

obenland:
> The fix is done on client-side instead of server-side because the server-side API endpoint is also shared with Reader, and I'd like to avoid diving into Reader's code to see how they are handling falsy URLs (and if they are handling it).

`is_valid_blog_for_feed_subscription()` is supposed to filter out spam sites and deleted sites—wouldn't purged Jetpack-connected sites fall within that bucket?

obenland:
@yansern Not sure if you got a notification for this? ^

yansern:
> `is_valid_blog_for_feed_subscription()` is supposed to filter out spam sites and deleted sites—wouldn't purged Jetpack-connected sites fall within that bucket?

Thanks!  Looking at the API endpoint, generating the subscription site list for both external users and logged-in users go through the same `build_subscriptions` function which calls the `is_valid_blog_for_feed_subscription()` function.

If the function is already called, it means it hasn't catered for other edge cases.

What exact other edge cases are there? I'm not sure. I will set up a logstash to monitor sites that produce these false URLs and determine what these other edge cases are.

> @yansern Not sure if you got a notification for this? ^

Thanks for the persistent reminder! You leave no stone unturned!

yansern:
Issue created: https://github.com/Automattic/wp-calypso/issues/77487

PR #76946, opened 2023-05-15
General Settings: Add accessible name to timezone drop down

### LINKED ISSUE #75438: General/Settings Site timezone drop down needs an accessible name
### Quick summary

The drop-down selector for 'Site timezone' on General/Settings has no accessible name and can be missed by screen readers.

This is also a violation of WCAG 2A and must be fixed before we can achieve A-level compliance.

### Steps to reproduce

1. Start on Settings/General
2. Inspect the Site timezone drop-down selector `<select class="form-select">`
3. Confirm the drop-down does not have an a11y label or any accessible name for screen readers or other assistive technology.

Alternatively, use a developer tool like axe DevTools to inspect the Settings/General screen for a11y rule violations.

The drop-down selector for Site timezone should have an accessible name that screen readers can access.

### What actually happened

The Site timezone drop-down selector `<select class="form-select">` should have an accessible name and the options in the drop-down should also be labeled in a way that allows screen reader users to select a locale for their timezone.

### Impact

Some (< 50%)

### Available workarounds?

Simple

### Logs or notes

_No response_

danielbachhuber:
Fixes https://github.com/Automattic/wp-calypso/issues/75438

## Proposed Changes

Fixes the "Select element must have an accessible name" warning by associating the `<label>` with the `<select>` element by assigning an `id=` to the latter:

image

## Testing Instructions

1. Navigate to 'General Settings'.
2. Verify the `id=` attribute appears as expected.
3. Run the axe DevTools extension and verify the "Select element must have an accessible name" error no longer appears.

### DIFF
--- a/client/components/timezone/README.md
+++ b/client/components/timezone/README.md
@@ -22,12 +22,11 @@ export default class extends React.Component {
 
 ## Props
 
-`selectedZone` - **optional** String value to define the selected timezone.
-
-`includeManualOffsets` - **optional** Boolean value to include/exclude the manual offsets from the
+* `selectedZone` - **optional** String value to define the selected timezone.
+* `id` - **optional** String value to add an `id=` attribute to the `<select>` element.
+* `includeManualOffsets` - **optional** Boolean value to include/exclude the manual offsets from the
 list of timezones, the default value is `true`
-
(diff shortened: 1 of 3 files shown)


wojtekn (approved):
Works as expected, axe devtools doesn't show the error anymore.

PR #76949, opened 2023-05-15
Hosting onboarding: Add shimmer loading animation to sites with AT transfer in progress on Sites page

vykes-mac:
Closes https://github.com/Automattic/dotcom-forge/issues/2376

## Proposed Changes
This PR adds a loading banner to an atomic transfer site to show that the transfer is in progress.
* Add loader notice that shows when site is transferring and when the transfer is completed.
* Add hook to check atomic transfer status. 

## Testing Instructions

1. Go to  `/sites` page.
2. get siteId of an atomic site
3. using  react dev tools select the top most component then right click the `store` and click `set as global variable`
4.  In the console run `$reactTemp0.dispatch({type: 'AUTOMATED_TRANSFER_STATUS_SET', siteId:atomicId, status: 'active' });` `$reactTemp` is what ever global variable the store is set to.
6.  Verify that the loading notice is shown when `active`.
7.  Verify `site activated notice` is shown when transfer completes (`it will fetch complete from server for exisiting AT site`)

* 
image
image

## Pre-merge Checklist


client/sites-dashboard/components/sites-dashboard.tsx
```diff
@@ -330,6 +330,8 @@ function useShowSiteCreationNotice( allSites: SiteExcerptData[], newSiteSlug: st
 			return;
 		}
 
+		site.is_creating = true;
```
zaguiini: We can't modify the `site` object directly because this object is being used by React and could affect the diffing process (i.e. knowing when to re-render). We should pass a new copy to the child by spreading the properties and appending it to the bottom as a new property.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
```diff
@@ -36,6 +43,15 @@ export const SiteItemThumbnail = ( {
 }: SiteItemThumbnailProps ) => {
 	const { __ } = useI18n();
 	const classes = classNames( props.className, disallowSelection );
+	const newSiteId = site.is_creating ? site.ID : null;
+	const transferStatus = useCheckSiteTransferStatus( newSiteId );
+
+	const isTransferCompleted = transferStatus === transferStates.COMPLETE;
+	const isTransferring = transferStatus !== null && ! isTransferCompleted;
```
zaguiini: I'd say to move this piece of logic to the back-end, abstracting it and using the same flag on both the hosting configuration page and the sites dashboard.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
```diff
@@ -36,6 +43,15 @@ export const SiteItemThumbnail = ( {
...
+
+	if ( isTransferCompleted && newSiteId ) {
```
zaguiini: I don't think we should use the new site ID from the query parameter to decide whether or not to show the processing flag. There is no dependency between the two 🤔

What I'm trying to say is that correlation is not cause. What if I have an existing site that is transitioning? It won't show up as transitioning on the Sites dashboard, despite being in a transition state.

zaguiini (commented):
@javierarce could you help us here? I feel a better option would be to use the "Plan expired" banner inside the site to indicate processing

client/sites-dashboard/components/sites-dashboard.tsx
vykes-mac: > @javierarce could you help us here? I feel a better option would be to use the "Plan expired" banner inside the site to indicate processing

image

Think showing banner like this over the thumbnail like the expired plan make sense, just to change the wording.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
vykes-mac: by backend you mean extracting to it's own custom hook that's re-useable? If you php I don't think there's any need for that, there's an endpoint that already gives the transfer status and redux actions setup to fetch the status.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
vykes-mac: What you're saying is that the status check should be done on all sites of the dashboard and not just the site that was just created with the `new-site` queryParam?

vykes-mac: siteId is used to fetch the transfer status from the back-end, but everything here is fixed to coming soon sites and only newly created sites from the new hosting flow.

client/sites-dashboard/components/sites-dashboard.tsx
javierarce: Could we show the loading ellipsis and a title instead? This would look nicer than using the banner and we would be using the same loading style as in the onboardings. 

image

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
zaguiini: >What you're saying is that the status check should be done on all sites of the dashboard and not just the site that was just created with the new-site queryParam?

Exactly.

client/sites-dashboard/components/sites-site-item-thumbnail.tsx
zaguiini: >If you php I don't think there's any need for that, there's an endpoint that already gives the transfer status and redux actions setup to fetch the status.

OK 👍 

>by backend you mean extracting to it's own custom hook that's re-useable?

I meant to create an abstraction. I wasn't sure about the implementation, that's why I proposed moving to the BE. But it could well be in the FE as well.

client/sites-dashboard/components/sites-dashboard.tsx
zaguiini: Yes @vykes-mac

@javierarce I was thinking of maintaining consistency. I don't like the idea of modifying the thumbnail 🤔 Adding the fixed banner on top of the tile image makes more sense to me, but you have the final saying 😁 

Like this:

image

But instead of error, it would be an info banner with a spinner. On a white background.

vykes-mac: the current implementation I have here, makes it easy to change the title and add a loader as @javierarce mentioned but personally I prefer the banner similar to what's show on the hosting config page when we activate those features

vykes-mac: Something like this



https://github.com/Automattic/wp-calypso/assets/47489215/b30df07d-6cba-4e6c-bc07-8e2884857ebb

zaguiini: That looks great @vykes-mac

Is there any way to dim the card slightly so that it shows up as disabled?

Making it gray could also work

zaguiini: Well, thinking better, maybe it's not a good idea since the user can already see the site through `/home/%s`, so let's not dim it.

zaguiini:
Didn't work for me 😢 

My steps:

1. Create a free site;
2. Search that site on the sites page;
3. Buy a Business plan;
4. Kickoff Atomic transfer by activating the hosting features;
5. Going back to the sites page.

I do see the domain change to `.wpcomstaging.com` but I don't see the transfer banner, even though a transfer is in progress.

client/sites-dashboard/components/sites-dashboard.tsx
javierarce: Ok, for now, let's use @vykes-mac's proposed solution to keep consistency with the expired notice.

vykes-mac:
> Didn't work for me 😢
> 
> My steps:
> 
> 1. Create a free site;
> 2. Search that site on the sites page;
> 3. Buy a Business plan;
> 4. Kickoff Atomic transfer by activating the hosting features;
> 5. Going back to the sites page.
> 
> I do see the domain change to `.wpcomstaging.com` but I don't see the transfer banner, even though a transfer is in progress.

🤔  I wonder if the transfer had completed before going to the `/sites` page. Let me check some more.

vykes-mac:
@zaguiini can you try again, fixed the issue. I had a check that was checking that the site was atomic before showing the loader but the site wouldn't have been atomic yet so it was bypassing the call to fetch the status.

client/sites-dashboard/hooks/use-check-site-transfer-status.tsx
```diff
@@ -0,0 +1,53 @@
...
+	const dispatch = useDispatch();
+
+	const transferStatus = useSelector( ( state ) => getAutomatedTransferStatus( state, siteId ) );
+	const isFetchingTransferStatus = useSelector( ( state ) =>
+		isFetchingAutomatedTransferStatus( state, siteId )
+	);
+
+	const isTransferCompleted = transferStatus === transferStates.COMPLETE;
+	const isTransferring =
+		transferStatus !== null && transferStatus !== transferStates.NONE && ! isTransferCompleted;
```
zaguiini, line 28: We should check for `.FAILURE` and `.ERROR` here too.

zaguiini (approved):
This looks good to me. We should (I'm not 100% sure) be able to skip the checking if `site_options.is_automated_transfer` returns `true`. But then I'm not sure if reversions set this flag to false.

zaguiini (approved):
This looks good to me. We should (I'm not 100% sure) be able to skip the checking if `site_options.is_automated_transfer` returns `true`. But then I'm not sure if reversions set this flag to false.

client/sites-dashboard/hooks/use-check-site-transfer-status.tsx
vykes-mac: yeah we should check for failures. This currently doesn't work on the list view, should we do the list view also like the plan expired? A column shows the expire notice.

zaguiini: That's something that we need to figure out with @javierarce but I'm fine with the list view being the same as the plan expired, except with a spinner included.

vykes-mac:
@javierarce can get input here, this is what I have for the table view but I'm not sure I'm feeling it.
image

javierarce:
> @javierarce can get input here, this is what I have for the table view but I'm not sure I'm feeling it.

Could we do something different for this view? It would make more sense to show the indication in the Status column, replacing the `Coming soon` and the `Launch checklist` elements

vykes-mac:
> Could we do something different for this view? It would make more sense to show the indication in the Status column, replacing the Coming soon and the Launch checklist elements

That make sense. Are you okay with the indication that's there or would you change it to something else?

zaguiini:
For me, it'd make more sense to display the progress in the "Status" column, instead of the site's current status. The Atomic transfer is not related to the Plan, but rather to the site status. It looks like a natural progression to me: AT -> Coming soon -> Public/Private.

I'd also drop the dark background and the exclamation at the end, @vykes-mac

javierarce:
I agree with @zaguiini, I think we could do something like this:

https://github.com/Automattic/wp-calypso/assets/4933/cec8e80f-d55a-470f-ad12-47c1f70d25ef

And then my suggestion would be to extend the same idea to the grid layout. This would look cleaner and nicer than the notice:


https://github.com/Automattic/wp-calypso/assets/4933/f25ca69f-d4ac-473c-9b70-e99d022a37f7

vykes-mac:
Updated based on @javierarce suggestions

vykes-mac:
@zaguiini give this another go.

zaguiini (approved):
Looks good to me!

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8010540

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8010540).

PR #76953, opened 2023-05-15
Marketing > Traffic: Remove gray background from 'Related Posts'

### LINKED ISSUE #75599: Color contrast issues on /marketing/traffic examples
### Quick summary

The example links on /marketing/traffic are too light blue for the grey background that they're on. This happens in a few different places, so a default color somewhere may need to be updated to a darker color or this particular instance may need a CSS override.

The text links are in the parent element:
`<div class="related-posts__preview-items">`

### Steps to reproduce

1. On /marketing/traffic
2. Inspect the link texts in the post example and manually run the foreground and background colors through a color contrast checker or use an accessibility checker like the axe DevTools chrome browser extension.

All text on the page should have a color contrast of at least 4.5:1 at a normal font size.

### What actually happened

Three links have just a shade too light text color. We can darken the color by a single shade or make the text bigger.

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

danielbachhuber:
Fixes https://github.com/Automattic/wp-calypso/issues/75599

## Proposed Changes

Replaces the gray background on 'Related posts' with a gray border to address accessibility issues with contrast.

### Before

image

### After

image

## Testing Instructions

1. Navigate to the 'Traffic' tab under 'Marketing'.
2. Run axe DevTools.
3. Verify the 'Related posts' elements no longer appear under the 'Elements must have sufficient color contrast' warning (but the page description will still be flagged).


kozer:
Nice work @danielbachhuber !
Your test works as expected. However, under the same element, I get the following:

highlight_error_traffic_marketing

It would be nice if we can fix that as well!

katinthehatsite (commented):
It works as described in the PR. One thing that I wanted to note is that when I enable the option `Show post category or tags`, the text appears in a light grey color which might be hard to see. I am not an accessibility expert though:

Screenshot 2023-05-16 at 10 32 43 AM

Here is the screenshot from the axe Dev Tools:

Screenshot 2023-05-16 at 10 37 41 AM


Also, just a general note on the testing instructions: it would be good to have a more precise path to `Navigate to the 'Traffic' tab under 'Marketing'` so something like **Tools > Marketing > Traffic** or https://wordpress.com/marketing/traffic. It took me a bit of time to find Traffic 😅

javierarce:
I agree with @kozer and @katinthehatsite. If we set the opacity of `.related-posts__preview-post-context` to 1 and changed the color to `var(--color-text-subtle)` we would fix the problem with that element.

javierarce:
By the way, if you use Chrome, you can easily conduct a basic accessibility check by simply highlighting an element. 

Screenshot 2023-05-16 at 11 06 32

javierarce (approved):
LGTM

danielbachhuber:
Good suggestions, thank you @kozer @katinthehatsite @javierarce . I tweaked the category coloring with f4b6f5e9c6eac0b0590322f318a53604b7ccf1da and updated the screenshots.

PR #76958, opened 2023-05-16
Enable sharing modal for english speakers

roo2:
Enables the sharing modal only for English speakers. see https://github.com/Automattic/wp-calypso/pull/76267

### Testing instructions
Set language to 'en' or 'en-gb'
Publish a new post 
You should see the sharing modal.


Addison-Stavlo (approved):
Confirmed this shows up for EN and EN (UK), while not showing up for others like ES.

If I change the locale for the site under "Settings -> General": I continue to get the modal.

Screenshot 2023-05-16 at 7 47 46 AM

If I change the locale under account settings, the modal no longer present:

Screenshot 2023-05-16 at 7 52 11 AM

✅  The above makes sense to me. I was confused which made sense at first but this is an interface item so the account setting for interfaces makes the most sense here.

roo2:
ahh yes good point about the interface language vs site language. I didn't consider that, but it looks like I used the right one :)

PR #76959, opened 2023-05-16
Reader: fix search feed polling request errors

creativecoder:
## Proposed Changes

* Limit polling on the Search page of Reader to 10 recommended sites, to prevent an error response and multiple retries when requesting more than 10.

## Testing Instructions

* Visit `/read/search`, open the Network tab, and wait until the client makes a polling request to `v1.2/read/recommendations/sites`.
* Without this change (production/staging), you get an error response with multiple retries: `{ error: "Bad Request", message: "Can only return 10 site results at a time" }`
* With this change, you should see a successful request.

### DIFF
--- a/client/state/data-layer/wpcom/read/streams/index.js
+++ b/client/state/data-layer/wpcom/read/streams/index.js
@@ -164,6 +164,9 @@ const streamApis = {
 				algorithm: 'read:recommendations:sites/es/2',
 				posts_per_site: 1,
 			} ),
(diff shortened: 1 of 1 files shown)


roo2 (approved):
LGTM!

creativecoder:
I wonder what purpose polling serves on the search page? Since I don't see the need to check for updated content when it's not tied to a feed/site, maybe we can remove it from this page all-together.

But I'll go ahead and merge this now to fix the immediate problem.

PR #76960, opened 2023-05-16
E2E: refactor Support specs into one flow.

### LINKED ISSUE #76916: Flaky E2E: Enter empty search keyword and expect no results to be shown
### Spec file

pecs/support/support__popover-invalid.ts

### TeamCity ID

10011482

### Logs

image

```
page.waitForSelector: Timeout 10000ms exceeded.
waiting for locator(':has-text("Sorry, there were no matches.")') to be visible
at SupportComponent.noResultsShown (/home/teamcity-2/buildAgent/work/c4a9d5b38c1dacde/packages/calypso-e2e/src/lib/components/support-component.ts:130:19)
    at Object.<anonymous> (/home/teamcity-2/buildAgent/work/c4a9d5b38c1dacde/test/e2e/specs/support/support__popover-invalid.ts:49:27)
```

worldomonation:
Fixes https://github.com/Automattic/wp-calypso/issues/76916.

## Proposed Changes

This PR is a semi-major refactor of how we think about Support related E2E specs. These specs were a near carbon copy of the ones that existed back in the Selenium framework days (prior to 2021). 

Now that we have the ability to execute component testing with `react-testing-library, most of what's covered in the specs (even in revised form) should be tested at the component level. 

However, to address the issue of permanent failure in https://github.com/Automattic/wp-calypso/issues/76916, I've decided to restructure the specs as a temporary bridge until they are moved into the component testing level.

Key changes:
- merge all scenarios (invalid, valid) into one.
- discontinue testing the support card present in `/home`.
- discontinue parametrization of the overall suite, instead using the environment variables to run on the targeted environment.

## Testing Instructions

  - [ ] Calypso E2E (desktop)
  - [ ] Calypso E2E (mobile)
  - [ ] Pre-Release E2E

## Pre-merge Checklist


packages/calypso-e2e/src/lib/components/support-component.ts
```diff
@@ -38,16 +24,7 @@ export class SupportComponent {
...
-
-	/**
-	 *
-	 */
-	async waitForQueryComplete(): Promise< void > {
-		await Promise.all( [
-			this.page.waitForSelector( selectors.resultsPlaceholderHelpCenter, { state: 'hidden' } ),
-			this.page.waitForSelector( selectors.resultsPlaceholder, { state: 'hidden' } ),
-		] );
+		this.root = this.page.getByLabel( 'Help Center' );
```
worldomonation: This is a technique/style I learned from working on the Gutenberg Playwright migration, specifically [this comment](https://github.com/WordPress/gutenberg/pull/45784#discussion_r1189538014).

The idea is that, when using role-based selectors (as recommended by Playwright now), we trade legibility for some ambiguity in the locator structure. For instance, the following locator is easy to read and understand, but there is potential to match on multiple elements on a page.

```
page.getByRole('button', { name: 'home' });
```

By always selecting from a parent/root locator it is possible to understand the overall structure hierarchy.

Note: if I was following the Gutenberg code style, this would probably be named something like `this.helpCenter`instead of `this.root`.

test/e2e/specs/support/support__popover.ts
```diff
@@ -0,0 +1,76 @@
...
+			await supportComponent.clearSearch();
+		} );
+
+		it( 'Search for "theme" returns articles and show me where links ', async function () {
+			const keyword = 'theme';
+			await supportComponent.search( keyword );
+
+			const [ articles, links ] = await supportComponent.getResultCount();
+			expect( articles ).toBeTruthy();
+			expect( links ).toBeTruthy();
```
worldomonation: This change in assertion is also from my experience with Gutenberg Playwright migration. In working on the PR, one thing that was repeated over the course of the PR is to [not structure tests around implementation details](https://github.com/WordPress/gutenberg/pull/45784#discussion_r1189525405). 

In the case of Support testing, my thoughts are that the following constitute implementation detail:
- how the results are presented;
- the order in which the results are presented.

Additionally, we cannot control:
- names of the returned resources;
- how many results are returned.

Throw on top of this the fact that the reworked Support popover does not have a true "empty results" state anymore and it becomes apparent the existing component is no longer fit for purpose.

By removing the assertion for `default state is shown` or `empty results are shown` and instead testing that the query returned _some sort of results_ for the expected categories, implementation detail testing is removed as much as possible.

test/e2e/specs/support/support__home.ts
worldomonation, line 1: This spec has been removed because of a couple of reasons:
- the support card present in `/home` serves up an older version of the inline help.
- usage for the support card in `/home` is lower than the popover help.

For the second point, I [compared]() the Tracks events for the two different Supports that are in production.
- `calypso_inlinehelp_show` (new Support)
- `calypso_inlinehelp_article_open` (legacy Support)

For the legacy Support, the `calypso_inlinehelp_article_open` stands in for the number of times a user interacts with the card and opens a link. It is not the perfect approximation, but there is no Tracks event exclusively for the user interacting with the card.

For a 7-day trailing period as of today, the popover has ~66k interactions while the card has just under half at ~31k interactions. So while not trivial, the card variant sees significantly fewer interactions.

worldomonation:
Looks like the current solution failed in a loop at 69 (nice) iterations.

image

packages/calypso-e2e/src/lib/components/support-component.ts
```diff
@@ -60,11 +37,7 @@ export class SupportComponent {
...
 		}
 
-		// This Promise.all wrapper contains many calls due to a certain level of uncertainty
-		// when the support popover is launched.
 		await Promise.all( [
-			// Waits for all placeholder CSS elements to be removed from the DOM.
-			this.waitForQueryComplete(),
 			this.page.waitForSelector( selectors.supportPopoverButton() ),
 			this.page.click( selectors.supportPopoverButton() ),
 		] );
```
WunderBart: I'm not sure I understand this chain - why do we need to wait for the same selector we're about to click here? I think it's what internally happens with `click()` - before it clicks, it waits for the given selector to be clickable, so the `click` itself should be sufficient.

packages/calypso-e2e/src/lib/components/support-component.ts
```diff
@@ -156,16 +100,14 @@ export class SupportComponent {
...
 	 */
 	async search( text: string ): Promise< void > {
-		// Wait for the response to the request and ensure the status is HTTP 200.
+		await this.helpCenter.getByPlaceholder( 'Search for help' ).fill( text );
+
 		await Promise.all( [
-			this.page.waitForResponse( selectors.successfulSearchResponse ),
-			this.page.fill( selectors.searchInput, text ),
-			this.waitForQueryComplete(),
+			this.page.waitForResponse( /wpcom\?query/ ),
```
WunderBart: I think adding a comment with a full example URL that we're matching here would be useful.

packages/calypso-e2e/src/lib/components/support-component.ts
```diff
@@ -85,69 +58,40 @@ export class SupportComponent {
...
+		return [
+			await this.page
+				.getByLabel( 'Search Results' )
+				.getByRole( 'list', { name: 'Recommended resources' } )
+				.count(),
+			await this.page
+				.getByLabel( 'Search Results' )
+				.getByRole( 'list', { name: 'Show me where to' } )
+				.count(),
+		];
```
WunderBart: It would be easier to follow if we returned the named results here, e.g.:

```suggestion
		return {
			articleCount: await this.page
				.getByLabel( 'Search Results' )
				.getByRole( 'list', { name: 'Recommended resources' } )
				.count(),
			linkCount: await this.page
				.getByLabel( 'Search Results' )
				.getByRole( 'list', { name: 'Show me where to' } )
				.count(),
		};
```

test/e2e/specs/support/support__popover.ts
WunderBart: In this case, I'd go for the `isGreaterThan( 0 )` matcher instead of `toBeTruthy()`. The reason for that is that `articles` or `links` could be an array as well, and `expect( [] ).toBeTruthy()` would return true. Being a little bit more explicit here could potentially guard us against some false negatives in the future.

test/e2e/specs/support/support__showme-where.ts
```diff
@@ -20,17 +21,20 @@ describe( 'Support: Show me where', function () {
...
-		await supportComponent.clickResult( 'where', 1 );
+		[ newPage ] = await Promise.all( [
+			page.waitForEvent( 'popup' ),
+			supportComponent.clickResult( 'Create a new site' ),
+		] );
 	} );
 
 	it( 'Signup flow is started', async function () {
-		await page.waitForURL( /start/ );
+		await newPage.waitForURL( /start/ );
```
WunderBart, line 39: Same as above - would be nice to have an example URL in the comment here 🙏

WunderBart (approved):
Just a few non-blocking comments. Thanks! 🚀

test/e2e/specs/support/support__showme-where.ts
worldomonation: Example added in 23ed24e76e93e4ec1b0598cbedc0e6ad2f89f4dd.

test/e2e/specs/support/support__popover.ts
worldomonation: > The reason for that is that articles or links could be an array as well, and expect( [] ).toBeTruthy() would return true.

Great point and something I have yet to come to grips with despite a couple of years of working in JS/TS ecosystem. 

Assertions changed in 23ed24e76e93e4ec1b0598cbedc0e6ad2f89f4dd, but may change again if the loop stress test fails.

packages/calypso-e2e/src/lib/components/support-component.ts
worldomonation: Sounds good, will do. 

Comment and example added in 23ed24e76e93e4ec1b0598cbedc0e6ad2f89f4dd.

packages/calypso-e2e/src/lib/components/support-component.ts
worldomonation: Thanks for the suggestion, makes total sense. 

Incorporated into fix in 23ed24e76e93e4ec1b0598cbedc0e6ad2f89f4dd.

packages/calypso-e2e/src/lib/components/support-component.ts
worldomonation: I think this was a mistake on my part. Can't think of reasons to explain this, all I can maybe come up with is that I mistyped a selector and forgot about it.

Edit: it looks like this was in place in the original file. Even more of a head scratcher.

packages/calypso-e2e/src/lib/components/support-component.ts
```diff
@@ -38,116 +14,79 @@ export class SupportComponent {
...
-
-	/**
-	 *
-	 */
-	async waitForQueryComplete(): Promise< void > {
-		await Promise.all( [
-			this.page.waitForSelector( selectors.resultsPlaceholderHelpCenter, { state: 'hidden' } ),
-			this.page.waitForSelector( selectors.resultsPlaceholder, { state: 'hidden' } ),
-		] );
+		this.helpCenter = this.page.locator( '.components-card__body > .inline-help__search' );
```
worldomonation, line 17: Changed: locator now points to the results field (in red) not the entire Inline Help modal.

This is because there is no selector to select the entire modal that is visible.

Screenshot 2023-05-17 at 3 48 52 PM

worldomonation:
The new solution wasn't perfect (failed at 94 iterations) but it was much better than the previous ones I had come up with, plus I didn't want to block https://github.com/Automattic/wp-calypso/pull/76734.

Issue #76332, opened 2023-04-27
Integrate the site picker screen into the migration flow


ouikhuan:
<img width="1023" alt="Screenshot 2023-05-09 at 12 05 55" src="https://github.com/Automattic/wp-calypso/assets/1241413/b233609a-9a68-445e-90d4-868f9d8374e3">

PR #76972, opened 2023-05-16
Update SiteDashboard components in order to be customizable

bogiii:
Related to #76332

## Proposed Changes

* Update SiteGridItem component supporting additional customization, such as:
  * show/hide launch nag
  * show/hide badge section
  * show/hide thumbnail link
  * show/hide site renew link
  * created site selection action "Select this site" button
  * provide select site callback method
* Update content controls component supporting customization, such as:
  * provide custom onQueryParamChange callback handler
  * show/hide mode switcher if there is no callback method
* Sites grid
  * introduce `siteSelectorMode` boolean property predefining set of properties
  * provide `onSiteSelectBtnClick` callback handler

Markup on 2023-05-16 at 11:44:04

## Testing Instructions

* Go to `/sites`
* Check does everything look the same as before this chunk of changes

* The usage of this customization could be tested in the next PR related to [the Migration Plugin i2](https://github.com/orgs/Automattic/projects/331/views/25) project.

## Pre-merge Checklist

### DIFF
--- a/client/sites-dashboard/components/sites-grid-item.tsx
+++ b/client/sites-dashboard/components/sites-grid-item.tsx
@@ -1,9 +1,10 @@
 import { PLAN_ECOMMERCE_TRIAL_MONTHLY } from '@automattic/calypso-products';
+import { Button } from '@automattic/components';
 import { useSiteLaunchStatusLabel, getSiteLaunchStatus } from '@automattic/sites';
 import { css } from '@emotion/css';
 import styled from '@emotion/styled';
 import { useI18n } from '@wordpress/react-i18n';
-import { AnchorHTMLAttributes, memo } from 'react';
+import { memo } from 'react';
 import { useInView } from 'react-intersection-observer';
 import { SiteExcerptData } from 'calypso/data/sites/site-excerpt-types';
 import { displaySiteUrl, getDashboardUrl, isStagingSite } from '../utils';
@@ -41,6 +42,16 @@ const badges = css( {
 	marginInlineStart: 'auto',
 } );
 
+const selectAction = css( {
+	display: 'flex',
+	gap: '8px',
+	alignItems: 'center',
+	marginInlineStart: 'auto',
+	button: {
+		whiteSpace: 'nowrap',
+	},
+} );
+
 export const siteThumbnail = css( {
 	aspectRatio: '16 / 11',
 	width: '100%',
@@ -72,10 +83,23 @@ const ellipsis = css( {
 
 interface SitesGridItemProps {
 	site: SiteExcerptData;
+	showLaunchNag?: boolean;
+	showBadgeSection?: boolean;
+	showThumbnailLink?: boolean;
+	showSiteRenewLink?: boolean;
+	onSiteSelectBtnClick?: ( site: SiteExcerptData ) => void;
 }
 
-export const SitesGridItem = memo( ( { site }: SitesGridItemProps ) => {
+export const SitesGridItem = memo( ( props: SitesGridItemProps ) => {
 	const { __ } = useI18n();
+	const {
+		site,
+		showLaunchNag = true,
+		showBadgeSection = true,
+		showThumbnailLink = true,
+		showSiteRenewLink = true,
+		onSiteSelectBtnClick,
+	} = props;
 
 	const isP2Site = site.options?.is_wpforteams_site;
 	const isWpcomStagingSite = isStagingSite( site );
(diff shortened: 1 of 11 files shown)


client/sites-dashboard/components/sites-grid-item.tsx
```diff
@@ -1,9 +1,10 @@
 import { PLAN_ECOMMERCE_TRIAL_MONTHLY } from '@automattic/calypso-products';
```
danielbachhuber, line 1: Can we make sure to update `client/sites-dashboard/components/sites-grid-item/README.md` and `client/sites-dashboard/components/sites-grid-item/docs/example.jsx` too?

client/sites-dashboard/components/sites-grid.tsx
```diff
@@ -27,16 +27,31 @@ interface SitesGridProps {
...
 
-export const SitesGrid = ( { sites, isLoading, className }: SitesGridProps ) => {
+export const SitesGrid = ( props: SitesGridProps ) => {
+	const { sites, isLoading, className, siteSelectorMode = false, onSiteSelectBtnClick } = props;
+	const additionalProps = siteSelectorMode
+		? {
+				showLaunchNag: false,
+				showBadgeSection: false,
+				showThumbnailLink: false,
+				showSiteRenewLink: false,
```
danielbachhuber, line 41: If we're disabling all of this UI, does it make sense to reuse the `<SitesGridItem>` component? Will it be a problem if `<SitesGridItem>` introduces new UI in the future? What benefit do you receive from using `<SitesGridItem>` vs. simply copying or exporting some features from it?

bogiii: I was thinking the same before I decided to edit the existing one.

There is a big chance to get additional requirements from design to include some stuff that exists now, for example, a renewal link or something similar... To avoid future copy/paste, I would go with this approach.

Anyway, we didn't introduce any more complexity except toggling existing html blocks.

client/sites-dashboard/components/sites-grid-item.tsx
bogiii: Sure, I'll do that!

client/sites-dashboard/components/sites-grid.tsx
danielbachhuber: @bogiii Sounds good. Let's add some tests (maybe snapshot?) covering the various permutations too, then. Tests will help ensure we don't accidentally break the design in the future.

zaguiini: Ok @bogiii your reasoning makes sense. I would suggest the same as Daniel: to invert the approach and go with generic components that can be more customizable.

I was thinking of creating a `<SitesGridItem />` component that accepts props, for example, `siteTitleRightAdornment`, instead of disabling specific sections of the grid item. But either approach works... let's not worry too much about it.

client/sites-dashboard/components/sites-grid.tsx
```diff
@@ -27,16 +27,31 @@ interface SitesGridProps {
 	className?: string;
 	isLoading: boolean;
 	sites: SiteExcerptData[];
+	siteSelectorMode?: boolean;
+	onSiteSelectBtnClick?: ( site: SiteExcerptData ) => void;
```
zaguiini, line 31: I'd say `onSiteSelect` is a better name. Why highlight that this event was fired through a click of a button?

bogiii: It was also the thing that I thought about how to call it best.

I mentioned the button in the callback name because there are more site links, but only the "Select this site" button will trigger provided callback method.

zaguiini: Why not make the whole area clickable? It's very possible that the user might click the thumbnail instead of the button for example.

bogiii: I just followed the design:

Screenshot 2023-05-16 at 20 41 37

If I rethink, I like having an explicit button for picking a site since the grid item component has options to provide more CTAs over the Thumbnail, such as plan renewal or similar.

client/sites-dashboard/components/sites-grid-item/docs/example.jsx
```diff
@@ -34,7 +34,17 @@ const SitesGridItemExample = () => {
...
+						site={ site }
+						key={ site.ID }
+						showLaunchNag={ false }
+						showBadgeSection={ false }
+						showThumbnailLink={ false }
+						showSiteRenewLink={ false }
+						onSiteSelectBtnClick={ () => {
+							// console.log( _site );
+						} }
+					/>
```
danielbachhuber: Can we include an example with all of these enabled too?

client/sites-dashboard/components/test/sites-grid-item.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+describe( '<SitesGridItem>', () => {
+	test( 'Default render', () => {
+		const tree = renderer
+			.create(
+				<SitesGridItem site={ makeTestSite( { title: 'The example site' } ) as SiteExcerptData } />
+			)
+			.toJSON();
+		expect( tree ).toMatchSnapshot();
+	} );
```
danielbachhuber, line 31: We should include at least one more example of non-default renders too.

danielbachhuber (approved):
Grid view appears as expected:

image

All of my other feedback is non-blocking.

ouikhuan (approved):
Looks great to me! Works as described 👍

PR #76978, opened 2023-05-16
Subscription Management: Fix menu keep shown in Post Subscriptions list after unfollow

### LINKED ISSUE #76123: `/subscriptions/comments`: When you unsubscribe from a post/comment, the settings popup stays open for the next row on the list
When you unsubscribe from a post/comment, the settings popup stays open for the next row on the list:

Markup on 2023-04-24 at 11:47:32

This can confuse the user thinking they haven't unsubscribed yet. As a result, they may accidentally unsubscribe from another site they hadn't planned to unsubscribe from in the first place.

I suspect the issue may be related to the virtualized list.

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/76123

## Proposed Changes

**The problem**

In the new Subscription Manager, when the user unfollows a Post, the popup they need to use is shown even when the row is not there anymore. If the user then clicks on "unfollow" in that menu, the actual registry occupying that place in the row is removed. This could lead the user to remove by mistake a subscription to the wrong post.

**Why this is happening**

The comment list is using a virtualized list component to display such a huge amount of rows. Virtualized lists sometimes avoid removing elements from the DOM so they can be reused later, like in the case of removing an element of the list. As the dom element is not removed, and the open menu is nested in the rows themselves, the menu was kept open.

**How was this problem fixed**

The virtualized list was generating its own `key` for the rows in the list. I entangled that `key` with the `post.id`, so the row cannot be reused in case of being removed. The key would change according to its content, forcing a redraw of the row.

## Testing Instructions

1. Apply this PR.
2. In the file `client/server/pages/index.js`, add this in line number 883:
Screenshot 2023-05-16 at 14 40 25
3. Go to `http://calypso.localhost:3000/subscriptions/comments`.
4. Click on the ellipsis menu and click "unfollow". The row and the menu popup should disappear.

### DIFF
--- a/client/landing/subscriptions/components/comment-list/comment-list.tsx
+++ b/client/landing/subscriptions/components/comment-list/comment-list.tsx
@@ -31,7 +31,12 @@ const CommentList = ( { posts }: CommentListProps ) => {
 			{ posts ? (
 				<VirtualizedList items={ posts }>
 					{ ( { item, key, style, registerChild } ) => (
-						<CommentRow key={ key } style={ style } forwardedRef={ registerChild } { ...item } />
+						<CommentRow
+							key={ `${ item.id }-${ key }` }
+							style={ style }
+							forwardedRef={ registerChild }
+							{ ...item }
+						/>
 					) }
 				</VirtualizedList>
 			) : null }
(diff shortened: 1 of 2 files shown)


yansern (approved):
This tests well. Thanks!

ivan-ottinger:
Thank you for this fix, @JuanLucha! 🙇🏼‍♂️ That bug was annoying.

PR #76982, opened 2023-05-16
Fix/remove paid feature for earn

millerf:
## Proposed Changes

* Now that Earn features are on free plan,  Most of the gating/ nudges need to be removed

## Testing plan

* Run this diff with `yarn start`
* Go to any simple WPCOM site dashboard and make sure no nudge to upgrade a plan appears, but there are nudges for Stripe


client/my-sites/earn/home.tsx
```diff
@@ -21,39 +11,32 @@ import EmptyContent from 'calypso/components/empty-content';
...
 import getSiteBySlug from 'calypso/state/sites/selectors/get-site-by-slug';
 import { getSelectedSiteSlug, getSelectedSiteId } from 'calypso/state/ui/selectors';
 import { isRequestingWordAdsApprovalForSite } from 'calypso/state/wordads/approve/selectors';
 import type { Image } from 'calypso/components/promo-section/promo-card/index';
 import type { AppState, SiteSlug } from 'calypso/types';
-
+// eslint-disable-next-line import/order
 import './style.scss';
+// eslint-disable-next-line import/order
+import { isJetpackSite } from 'calypso/state/sites/selectors';
```
romarioraffington: I think we can easily fix this eslint error by moving the import to the correct line?

millerf: Fixed in 5772198b9ab4421249cf065c2e599af95f7d0858. Thanks, I overlooked it...

edanzer (approved):
Testing as expected here. And code looks good - at least based on my understanding of what we're doing. Mostly removing code here. So I partly confirmed the nudges are gone, and partly just smoke tested to confirm no other regressions from deleting still-used code somewhere. 

Note: There are some failing e2e tests related to domain upsells. This appears to be a broader issue affecting many calypso PRs, not anything specific to this one.

n3f (approved):
Didn't have the time to do a deep dive on the actual code, but testing it appears to work as expected.

PR #76986, opened 2023-05-16
Help Center: move site picker

arcangelini:
## Proposed Changes

This is *part 1* of refactoring the Help Center contact form. I have broken these changes out to make my original PR smaller. Original PR #72699

This moves the site picker / dropdown into its own component allowing the main contact form component to be a little less cluttered. All the logic should be exactly the same just passed over to another file. The component is responsible for the dropdown or the text field. All details are passed in as props.

There was also an error with the `SitePicker` component due to a missing key so I added that as well.

## Testing Instructions

1. Pull branch and `yarn start`
2. Log in and open Help Center.
3. If you have a site available on your user then it should auto-populate the drop down with that site.
4. If you don't then you should see the text field.
5. Check against WordPress.com for regression.

**If you need a test account for a user with no sites I have one you can use. Please ping me.**

### DIFF
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -660,8 +660,7 @@ export const HelpCenterContactForm = () => {
 				setSitePickerChoice={ setSitePickerChoice }
 				currentSite={ currentSite }
 				siteId={ sitePickerChoice === 'CURRENT_SITE' ? currentSite?.ID : 0 }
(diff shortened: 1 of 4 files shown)


escapemanuele (approved):
Tested this and it all seems to be working. Also thanks for reducing the scope of the PR 👍

escapemanuele:
Also tested this using a user with no sites and it is showing the text input as advertised 🥇

PR #76990, opened 2023-05-16
Quarentined e2e tests because of wpcom etk bug

lupus2k:
## Proposed Changes

As ETK works on wpcom and tests runs for it, if ETK has an error, all tests are failing.
We quarentined 2 tests that are failing and we will bring it back after the ETK bugs were fixed.


lupus2k:
We muted those tests instead of quarentined. Thanks @youbrokesomething

PR #76997, opened 2023-05-16
Payments: Support Error.prototype.cause in all locations that record errors

sirbrillig:
## Proposed Changes

The `CheckoutErrorBoundary` component is a React error boundary with several features including allowing its parent to provide a logging function for the error message. In https://github.com/Automattic/wp-calypso/pull/75293 we modified the component to record both the original error message inside the `Error.prototype.cause` property and the displayed message in the `Error.prototype.message` property. These are then unraveled using a `convertErrorToString()` function.

However, that PR only updated one of the locations that records error messages from `CheckoutErrorBoundary`, meaning that the other locations no longer recorded their original errors.

This PR modifies all the other places that use `CheckoutErrorBoundary` to use `convertErrorToString()` as well.

Props to @pottedmeat for noticing this issue.

## Testing Instructions

TypeScript should mostly cover this.

### DIFF
--- a/client/me/purchases/controller.jsx
+++ b/client/me/purchases/controller.jsx
@@ -21,6 +21,7 @@ import {
 	billingHistory,
 } from 'calypso/me/purchases/paths';
 import PurchasesNavigation from 'calypso/me/purchases/purchases-navigation';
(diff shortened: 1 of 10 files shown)


pottedmeat (approved):
Code looks good and no compile errors.

PR #77000, opened 2023-05-17
Fix errors on scroll of unmounted sticky component

roo2:
reported here: p1684284352464299-slack-C04U5A26MJB

I added a backup scroll event to the sticky panel but didn't detatch it on unmount.

There is also an issue with using `ReactDom.findDOMNode( this )`. If the component is unmounted, even checking this will throw a JS error.

Even after I detatched and cancelled the onscroll event, `ReactDom.findDOMNode( this )` was still being called once and throwing an error.

I changed the file to use a reference instead which can safely be checked if the component is unmounted see https://stackoverflow.com/questions/39767482/is-there-a-way-to-check-if-the-react-component-is-unmounted

Related to https://github.com/Automattic/wp-calypso/pull/76747

## Testing Instructions

Go to the /tags page
scroll down until the panel "sticks"
click on a link (if possible, tab to a link, and hit enter while currently scrolling)
on the next page, you should not see an error on scrolling
I was only able to test the "StickyPanelWithScrollEvent" component by hacking the code as follows. It is a fallback for old browsers.

```
export default StickyPanelWithScrollEvent;

// export default hasIntersectionObserver
// 	? StickyPanelWithIntersectionObserver
// 	: StickyPanelWithScrollEvent;
```

### DIFF
--- a/client/components/sticky-panel/index.jsx
+++ b/client/components/sticky-panel/index.jsx
@@ -2,8 +2,7 @@ import { isMobile } from '@automattic/viewport';
 import classNames from 'classnames';
 import { throttle, debounce, defer } from 'lodash';
 import PropTypes from 'prop-types';
-import { Component } from 'react';
-import ReactDom from 'react-dom';
+import { Component, createRef } from 'react';
 import afterLayoutFlush from 'calypso/lib/after-layout-flush';
 
 import './style.scss';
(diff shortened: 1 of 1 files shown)


roo2:
one sec, need to update tests...

roo2:
updated code so signature of getDimensions doesn't change

PR #77002, opened 2023-05-17
Checkout: Fix fetching of receipt data for thank-you page

sirbrillig:
We store purchase receipt data in the calypso Redux data store. It gets injected by two different sources: the transactions endpoint response (`/me/transactions`) and the receipt endpoint response (`/me/billing-history/receipt/:receiptId`, but note: the receipt endpoint has two "modes" depending on if `?format=display` is set; in this case we're talking about when that param is _not_ set which only happens on the checkout thank-you page). 

However, the format of those two receipt objects is slightly different. Most notably, in the transactions endpoint, the `purchases` property is a key-value store where the key is the site ID and the value is an array of purchases, but in the receipt endpoint, `purchases` is just an array of purchases.

Because of this difference, the code that processes the receipt data to save it in the Redux store must handle both formats, but it does not do so correctly. Currently it handles only the transactions endpoint response, which means that if the checkout thank-you page fetches its own receipt data, it will fail to find any purchases, leading to a blank page for the user.

In most instances of the checkout flow, this failure is hidden because the transactions endpoint response will have filled in the data just after checkout and before we reach the thank-you page, but that is not the case for redirect transactions like PayPal since their receipt data isn't available before the user leaves checkout.

## Proposed Changes

In this PR, we modify the code that processes receipt data so that it can handle the formats of both the transaction endpoint and the receipts endpoint.

Fixes https://github.com/Automattic/payments-shilling/issues/1673

## Screenshots

Before:

Screenshot 2023-05-16 at 4 44 17 PM

After:

Screenshot 2023-05-16 at 4 57 15 PM

## Testing Instructions

Automated tests are included, but it's worth testing both versions of the experience to make sure they work as expected with real data.

### Testing transactions endpoint data

- Visit `/domains` and follow the domain-only purchase flow (choose a `.live` domain).
- When you reach checkout, select a credit card (new or existing) as the payment method.
- Complete the purchase.
- If you see an upsell page, click "Skip for now".
- Verify that you see the receipt page.

### Testing receipt endpoint data

- Visit `/domains` and follow the domain-only purchase flow (choose a `.live` domain).
- When you reach checkout, select PayPal as the payment method.
- Complete the flow (you can use a test PayPal account as per these instructions: PCYsg-IA-p2).
- If you see an upsell page, click "Skip for now".
- Verify that you see the receipt page.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
@@ -84,7 +84,12 @@ export default async function existingCardProcessor(
 
 	return submitWpcomTransaction( formattedTransactionData, dataForProcessor )
 		.then( async ( stripeResponse ) => {
-			if ( stripeResponse?.message?.payment_intent_client_secret ) {
+			if (
+				stripeResponse &&
+				'message' in stripeResponse &&
+				typeof stripeResponse.message !== 'string' &&
+				stripeResponse.message?.payment_intent_client_secret
+			) {
 				debug( 'transaction requires authentication' );
 				// 3DS authentication required
 				reduxDispatch( recordTracksEvent( 'calypso_checkout_modal_authorization', {} ) );
@@ -100,10 +105,12 @@ export default async function existingCardProcessor(
 			return stripeResponse;
 		} )
 		.then( ( stripeResponse ) => {
-			if (
-				stripeResponse?.redirect_url &&
-				! stripeResponse?.message?.payment_intent_client_secret
-			) {
+			const hasPaymentIntent =
+				stripeResponse &&
+				'message' in stripeResponse &&
+				typeof stripeResponse.message !== 'string' &&
+				stripeResponse?.message?.payment_intent_client_secret;
+			if ( stripeResponse?.redirect_url && ! hasPaymentIntent ) {
 				debug( 'transaction requires redirect' );
 				return makeRedirectResponse( stripeResponse.redirect_url );
 			}
--- a/client/my-sites/checkout/composite-checkout/lib/multi-partner-card-processor.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/multi-partner-card-processor.ts
@@ -131,14 +131,19 @@ async function stripeCardProcessor(
 	debug( 'sending stripe transaction', formattedTransactionData );
 	return submitWpcomTransaction( formattedTransactionData, transactionOptions )
 		.then( async ( stripeResponse ) => {
-			if ( stripeResponse?.message?.payment_intent_client_secret ) {
+			if (
+				stripeResponse &&
+				'message' in stripeResponse &&
+				typeof stripeResponse.message !== 'string' &&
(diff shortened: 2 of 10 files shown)


spraveenitpro:
**Functional Testing:**

Some warnings are documented in the screenshot below.

<details>
<summary>Testing receipt endpoint data -  Without Patch</summary>


</details>

<details>
<summary>Testing receipt endpoint data -  With Patch </summary>



As you can see we got a warning `Warning: Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. You can only call Hooks`
</details>

<details>
<summary> Testing transactions endpoint data - Without Patch</summary>

 
</details>

<details>
<summary> Testing transactions endpoint data - With Patch</summary>



As you can see we got a warning `Warning: Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. You can only call Hooks`
</details>

client/my-sites/checkout/composite-checkout/lib/existing-card-processor.ts
spraveenitpro, line 1: This is a great pattern and improves readability🙏🏽

client/my-sites/checkout/composite-checkout/lib/multi-partner-card-processor.ts
spraveenitpro, line 1: I like how everything is explicit which makes it easier to review.

client/my-sites/checkout/composite-checkout/lib/normalize-transaction-response.ts
```diff
@@ -1,24 +1,26 @@
...
 	response: unknown
-): WPCOMTransactionEndpointResponse {
+): WPCOMTransactionEndpointResponse | FailedResponse {
 	if ( ! response ) {
 		return emptyResponse;
 	}
 	const transactionResponse = response as WPCOMTransactionEndpointResponse;
 	if (
-		transactionResponse.message ||
+		( 'message' in transactionResponse && transactionResponse.message ) ||
```
spraveenitpro, line 19: Great approach here! From my understanding, this can limit runtime errors from accessing the message property if it does not exist.

spraveenitpro (approved):

client/my-sites/checkout/composite-checkout/lib/normalize-transaction-response.ts
sirbrillig: Yup! And more importantly it "convinces" TypeScript that `transactionResponse.message` is ok to access (although we have to do both because `'message' in transactionResponse` will return true even if `transactionResponse.message` is undefined).

sirbrillig:
> As you can see we got a warning Warning: Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. You can only call Hooks

Thanks. I'll investigate this. Since it has to do with the thank-you page it should be easier to track down now that https://github.com/Automattic/wp-calypso/pull/76685 is merged. But since it's happening with or without this diff, it's not a blocker.

Issue #76719, opened 2023-05-09
Task: Explore consolidating the Pattern Picker and Design Carousel


daledupreez:
### Details

As per this Slack thread - p1683640282530249/1683589244.945149-slack-C02FMH4G8 - we should explore whether we can consolidate the Pattern Picker in `packages/pattern-picker` with the Design Carousel in `packages/design-carousel`.

### Checklist

- [ ] Work out if we can consolidate more of the code and CSS for these two components.

daledupreez:
This was fixed in https://github.com/Automattic/wp-calypso/pull/77182.

PR #77003, opened 2023-05-17
Update designCarousel to include link-in-bio related changes.

agrullon95:
Related to #76719

## Proposed Changes

* DesignCarousel and Pattern-picker packages have similar functionality, so it is best to try to consolidate them for easier maintainability. This PR updates the designCarousel package to work with link-in-bio flow. 

## Testing Instructions

* Checkout this branch
* Execute `yarn start`
* Navigate to http://calypso.localhost:3000/setup/ecommerce/designCarousel to ensure it is working as intended and looks correct.
* Navigate to http://calypso.localhost:3000/setup/link-in-bio/patterns to ensure it is working as intended and looks correct.
* Navigate to http://calypso.localhost:3000/setup/link-in-bio-tld/patterns to ensure it is working as intended and looks correct.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/design-carousel/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/design-carousel/style.scss
@@ -1,5 +1,123 @@
+@import "../style";
+
 .designCarousel.step-container .step-container__header .formatted-header {
 	.formatted-header__subtitle {
 		text-align: center;
+	}
+}
+
+.design-carousel__select {
+	// Hide Gridicon Heart icon by default
+	svg {
+		display: none;
+	}
+}
+
+.link-in-bio.patterns,
+.link-in-bio-tld.patterns {
+	height: 100vh;
+
+	@supports ( height: 100svh ) {
+		height: 100svh;
+	}
+
+	padding: 25px 0 0;
+
+	@include break-medium {
+		padding: 40px 0 0;
+	}
+
+	.step-container {
+		width: 100%;
+		height: 100%;
+
+		&.designCarousel {
+
+			@media (max-width: 600px) {
+				padding: 60px 0 0;
+			}
+
+		}
+
+		@include break-small {
+			height: auto;
+			padding-top: 2px;
+		}
+
+		@include onboarding-break-mobile-landscape {
+			padding: 25px 0 0;
+			height: auto;
+		}
+
+		@media screen and (orientation: landscape) and ( max-height: 720px ) {
+			padding: 5px;
+		}
+
+		.step-container__header {
+			margin-top: 0;
+			margin-bottom: 20px;
+
+			@media (min-width: 600px) {
+				margin-top: 24px;
+			}
+
+			@include break-medium {
+				margin-bottom: 32px;
+				margin-top: 54px;
+			}
+
+			@include onboarding-break-mobile-landscape {
+				margin: 10px;
+			}
+		}
+
+		.step-container__content {
+			height: 100%;
+			min-height: 100%;
+			max-height: 100%;
+
+			@include break-small {
+				height: auto;
+			}
+		}
+
+		.step-container__jetpack-powered {
+			margin-top: 20px;
+		}
+
+		.design-carousel__cta {
+			margin-top: 10px;
+		}
+
+
+		.design-carousel__select {
+			width: 28vh;
+
+			svg {
+				display: none;
+			}
+
+			use {
+				fill: #fff;
+			}
+
+			&:not(:disabled):hover {
+				svg {
+					display: inline;
+				}
+
+				span {
+					display: none;
+				}
+			}
+		}
+
+		.design-carousel__item-mobile-only {
+			margin: 20px 0;
+		}
+
+		.design-carousel__item .mshots-image__container {
+			height: 100%;
+		}
 	}
 }
--- a/client/landing/stepper/declarative-flow/link-in-bio.ts
+++ b/client/landing/stepper/declarative-flow/link-in-bio.ts
@@ -31,7 +31,6 @@ const linkInBio: Flow = {
 			},
 			{ slug: 'domains', asyncComponent: () => import( './internals/steps-repository/domains' ) },
 			{ slug: 'plans', asyncComponent: () => import( './internals/steps-repository/plans' ) },
-
 			{
 				slug: 'patterns',
 				asyncComponent: () => import( './internals/steps-repository/design-carousel' ),
--- a/client/landing/stepper/declarative-flow/link-in-bio.ts
+++ b/client/landing/stepper/declarative-flow/link-in-bio.ts
@@ -31,7 +31,11 @@ const linkInBio: Flow = {
 			},
 			{ slug: 'domains', asyncComponent: () => import( './internals/steps-repository/domains' ) },
 			{ slug: 'plans', asyncComponent: () => import( './internals/steps-repository/plans' ) },
-			{ slug: 'patterns', asyncComponent: () => import( './internals/steps-repository/patterns' ) },
+
+			{
+				slug: 'patterns',
+				asyncComponent: () => import( './internals/steps-repository/design-carousel' ),
+			},
 			{
 				slug: 'siteCreationStep',
 				asyncComponent: () => import( './internals/steps-repository/site-creation-step' ),
--- a/packages/design-carousel/package.json
+++ b/packages/design-carousel/package.json
@@ -29,9 +29,9 @@
 		"watch": "tsc --build ./tsconfig.json --watch"
 	},
 	"dependencies": {
+		"@automattic/design-picker": "workspace:^",
 		"@automattic/i18n-utils": "workspace:^",
 		"@automattic/onboarding": "workspace:^",
-		"@automattic/design-picker": "workspace:^",
 		"classnames": "^2.3.1",
 		"react-query": "^3.32.1",
 		"swiper": "^4.5.1",
--- a/packages/design-carousel/package.json
+++ b/packages/design-carousel/package.json
@@ -31,6 +31,7 @@
 	"dependencies": {
 		"@automattic/i18n-utils": "workspace:^",
 		"@automattic/onboarding": "workspace:^",
+		"@automattic/design-picker": "workspace:^",
 		"classnames": "^2.3.1",
 		"react-query": "^3.32.1",
 		"swiper": "^4.5.1",
--- a/packages/design-carousel/src/index.tsx
+++ b/packages/design-carousel/src/index.tsx
@@ -1,9 +1,12 @@
-import { useStarterDesignsQuery } from '@automattic/data-stores';
+import { Gridicon } from '@automattic/components';
+import { StarterDesigns, useStarterDesignsQuery } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
+import { MShotsOptions } from '@automattic/onboarding/src';
 import { Button } from '@wordpress/components';
+import { useMediaQuery } from '@wordpress/compose';
 import { Icon, chevronLeft, chevronRight } from '@wordpress/icons';
 import { useI18n } from '@wordpress/react-i18n';
-import React, { useEffect, useRef } from 'react';
+import { useEffect, useRef } from 'react';
 import Swiper from 'swiper';
 import { Item } from './item';
 import 'swiper/dist/css/swiper.css';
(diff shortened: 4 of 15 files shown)


agrullon95:
The Link-in-bio mobile viewport still needs to be tweaked to match production:


CleanShot 2023-05-17 at 00 19 30@2x


CleanShot 2023-05-17 at 00 19 03@2x

client/landing/stepper/declarative-flow/link-in-bio.ts
```diff
@@ -31,7 +31,11 @@ const linkInBio: Flow = {
 			},
 			{ slug: 'domains', asyncComponent: () => import( './internals/steps-repository/domains' ) },
 			{ slug: 'plans', asyncComponent: () => import( './internals/steps-repository/plans' ) },
-			{ slug: 'patterns', asyncComponent: () => import( './internals/steps-repository/patterns' ) },
+
```
escapemanuele: We don't need the empty line here 👍

packages/design-carousel/package.json
```diff
@@ -31,6 +31,7 @@
 	"dependencies": {
 		"@automattic/i18n-utils": "workspace:^",
 		"@automattic/onboarding": "workspace:^",
+		"@automattic/design-picker": "workspace:^",
```
escapemanuele: I think you may need to run `yarn` in `packages/design-carousel` and commit the `yarn.lock` and `package.json` files 👍

escapemanuele:
Hey @agrullon95, great PR! I tested and it is working, I left a couple of comments to address and then maybe the tests will pass 🚀

packages/design-carousel/package.json
agrullon95: Thanks. I also added it to the tsconfig.json file. Now its working.

client/landing/stepper/declarative-flow/internals/steps-repository/design-carousel/style.scss
```diff
@@ -1,5 +1,123 @@
...
+
+.design-carousel__select {
+	// Hide Gridicon Heart icon by default
+	svg {
+		display: none;
+	}
+}
+
+.link-in-bio.patterns,
+.link-in-bio-tld.patterns {
```
simison, line 17: Should these be just "mobile patterns" instead of scoping them to a single flow? Makes it less cumbersome to re-use for other flows in future.

agrullon95: Yes thats a great idea. I can make these changes in a follow up PR. That should make it easier to add flows without worrying about necessary CSS changes.

simison: Yes, totes fine for a follow-up. Thank you!

Remember to consider re-usability across all code, CSS, JS, etc.

packages/design-carousel/src/index.tsx
```diff
@@ -1,9 +1,12 @@
-import { useStarterDesignsQuery } from '@automattic/data-stores';
+import { Gridicon } from '@automattic/components';
+import { StarterDesigns, useStarterDesignsQuery } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
+import { MShotsOptions } from '@automattic/onboarding';
```
daledupreez: Should this not be a type import?
```suggestion
import type { MShotsOptions } from '@automattic/onboarding';
```

packages/design-carousel/src/index.tsx
```diff
@@ -101,6 +192,8 @@ export default function DesignCarousel( { onPick, selectedDesignSlugs }: DesignC
 					} }
 				>
 					<span>{ __( 'Continue' ) }</span>
+					<Gridicon icon="heart" size={ 18 } />
+					{ /* Heart icon when hovering over continue missing here */ }
```
daledupreez: What is this comment for? I am a bit unclear on the intent.

packages/design-picker/src/utils/designs.ts
```diff
@@ -33,7 +33,7 @@ export const getDesignPreviewUrl = (
 			? options.viewport_height || DEFAULT_VIEWPORT_HEIGHT
 			: undefined,
 		source_site: 'patternboilerplates.wordpress.com',
-		use_screenshot_overrides: options.use_screenshot_overrides,
+		use_screenshot_overrides: options.use_screenshot_overrides || true,
```
daledupreez: This feels like it should either be `?? true` or just `true`, as the `||` will fall through to `true` if the incoming value is falsey.

daledupreez (commented):
I have some comments up from inspection, which likely need a bit of discussion.

packages/design-carousel/src/index.tsx
agrullon95: This was a leftover comment for me. I removed it.

packages/design-picker/src/utils/designs.ts
agrullon95: @daledupreez  Good catch.

What do you think about changing it to:

```
...( options.use_screenshot_overrides && {
			use_screenshot_overrides: options.use_screenshot_overrides,
		} ),
```

This way its added only when necessary and I can update the design-carousel item component:

```
<MShotsImage
	url={ getDesignPreviewUrl( design, {
	     use_screenshot_overrides: true,
	} ) }
	options={ options }
	alt={ design.title }
        aria-labelledby=""
/>
```

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+	};
 
 	const { data: allDesigns } = useStarterDesignsQuery( {
 		_locale: locale,
 	} );
 
-	let selectedDesigns = allDesigns?.designs;
+	const shouldDisplayMobileViewOnly = ( flow: string | undefined ) => {
+		switch ( flow ) {
+			case 'link-in-bio':
```
alshakero: Let's use [`isLinkInBioFlow`](https://github.com/Automattic/wp-calypso/blob/trunk/packages/onboarding/src/utils/flows.ts#LL28C14-L28C29) instead. And since this function doesn't depend on the state, we can hoist it up and outside the component; to avoid recreating it at every render.

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				return true;
+			default:
+				return false;
+		}
+	};
```
alshakero:
```suggestion
	const shouldDisplayMobileViewOnly = ( flow: string | undefined ) => {
		switch ( true ) {
			case isLinkInBioFlow(flow):
				return true;
			default:
				return false;
		}
	};
```

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
-			selectedDesignSlugs.includes( design.slug )
-		);
+	const getLinkInBioDesigns = ( allDesigns: StarterDesigns | undefined ) => {
+		const designs = allDesigns
+			? allDesigns?.designs.filter(
+					( design ) =>
+						design.is_virtual &&
+						design.categories.some( ( category ) => category.slug === 'link-in-bio' )
+			  )
+			: [];
```
alshakero:
```suggestion
		const designs = allDesigns?.designs.filter(
					( design ) =>
						design.is_virtual &&
						design.categories.some( ( category ) => category.slug === 'link-in-bio' )
			  )
			?? [];
```

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+		locale: string | undefined,
+		selectedDesignSlugs: string[] | undefined,
+		flow: string | undefined
+	) => {
+		if ( ! allDesigns || ! locale ) {
+			return null;
+		}
+
+		switch ( flow ) {
+			case 'link-in-bio':
```
alshakero: Same as above, this may break for other variations of LiB flow (eg `link-in-bio-tld`).

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				return getLinkInBioDesigns( allDesigns );
+			case 'ecommerce':
```
alshakero: Let's use [constants](https://github.com/Automattic/wp-calypso/blob/2693b31ceeb00c7cf352c69255311641280637ed/packages/onboarding/src/utils/flows.ts#L13).

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				return false;
+		}
+	};
 
-	if ( selectedDesigns && selectedDesignSlugs ) {
-		// If we have a restricted set of designs, filter out all unwanted designs
-		const filteredDesigns = selectedDesigns.filter( ( design ) =>
-			selectedDesignSlugs.includes( design.slug )
-		);
+	const getLinkInBioDesigns = ( allDesigns: StarterDesigns | undefined ) => {
```
alshakero: Should be outside the component.

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
-		selectedDesigns = selectedDesignSlugs
-			.map( ( selectedDesignSlug ) =>
-				filteredDesigns.find( ( design ) => design.slug === selectedDesignSlug )
-			)
-			.filter( ( selectedDesign ) => !! selectedDesign ) as Design[];
-	}
+		return designs;
+	};
+
+	const getEcommerceDesigns = (
```
alshakero: Should be outside the component. But would need `selectedDesign` argument.

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				)
+				.filter( ( selectedDesign ) => !! selectedDesign ) as Design[];
+		}
+
+		return selectedDesigns;
+	};
+
+	const getCarouselOptions = ( flow: string | undefined ) => {
+		switch ( flow ) {
+			case 'link-in-bio':
```
alshakero: Should use `isLinkInBioFlow`.

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				.map( ( selectedDesignSlug ) =>
+					filteredDesigns.find( ( design ) => design.slug === selectedDesignSlug )
+				)
+				.filter( ( selectedDesign ) => !! selectedDesign ) as Design[];
+		}
+
+		return selectedDesigns;
+	};
+
+	const getCarouselOptions = ( flow: string | undefined ) => {
```
alshakero: Can be outside the component.

packages/design-carousel/src/index.tsx
```diff
@@ -1,45 +1,121 @@
...
+				return desktopOptions;
+		}
+	};
+
+	const getFlowDesigns = (
+		allDesigns: StarterDesigns | undefined,
+		locale: string | undefined,
+		selectedDesignSlugs: string[] | undefined,
+		flow: string | undefined
+	) => {
```
alshakero: Should be outside the component.

alshakero (changes requested):
Left mostly optional feedback. Really good work. Should be tested with LiB flow variations because I think it's currently broken in two of them.

packages/design-carousel/src/index.tsx
agrullon95: I moved this function outside. Im not sure if I need the selectedDesign argument since its derived from allDesigns.

packages/design-carousel/src/index.tsx
agrullon95: Updated to use constants for ecommerce and link-in-bio.

packages/design-carousel/src/index.tsx
agrullon95: Good catch. I updated these cases to use the `isLinkInBioFlow` function.

agrullon95:
> Left mostly optional feedback. Really good work. Should be tested with LiB flow variations because I think it's currently broken in two of them.

@alshakero Great feedback. I updated the PR to include the changes you suggested above.
Including using constants for the flows, using the `isLinkInBioFlow` function to cover the different link-in-bio flows, and moved the functions outside of the components.

packages/design-picker/src/utils/designs.ts
daledupreez: I think that should work!

packages/design-carousel/src/index.tsx
```diff
@@ -12,19 +15,33 @@ import type { Design } from '@automattic/design-picker/src/types';
 type DesignCarouselProps = {
 	onPick: ( design: Design ) => void;
 	selectedDesignSlugs?: string[];
+	flow?: string | null | undefined;
```
daledupreez, line 18: I have a general comment here: can we not move _all_ the logic from functions below into props that are specified by callers? This should simplify the component and avoid reaching back out into external logic. As things stand, there is a _lot_ of flow-specific logic here, and it feels like pushing some of the decision-making up to the calling component will be easier to maintain in future.
Essentially, I am thinking we could update the props as per the suggestion here.
```suggestion
	flow?: string | null | undefined;
	designFilter?: ( design: Design ) => boolean; // My syntax could be wrong here; this would also mean `selectedDesignSlugs` could probably be dropped
	displayOnlyMobileView?: boolean;
	carouselDesktopOptions?: MShotsOptions; // should be defaulted if not present
	carouselMobileOptions?: MShotsOptions; // should be defaulted if not present
```

Maybe we can tackle that in a follow-up PR, but I think it would be a much cleaner interface.

agrullon95: @daledupreez - That makes sense and is definitely something I can change in a follow-up PR. If everything is working with the current PR on your end, I can merge this and start working on those changes.

alshakero (approved):
Such a neat PR now. Absolutely beautiful. 

Tested LiB and LiB TLD, and ecommerce.

daledupreez (approved):
Let's get this shipped!

I tested the eCommerce flow, and that looks good to me.

PR #77013, opened 2023-05-17
Subscriptions Management: Add "Cancel subscription" button handling

### LINKED ISSUE #76981: Subscription Management: Integrate "Cancel subscription" API on individual site subscription page.
## Dev notes
Use `SubscriptionManager.useSiteUnsubscribeMutation();` the returned mutate function requires the following payload:
```json

```

### API:
`POST` https://public-api.wordpress.com/rest/v1.1/read/following/mine/delete?http_envelope=1
### Payload:
```json
{"source": "calypso", "url": "http://yourdomain55.wordpress.com"}
```

ivan-ottinger:
Resolves https://github.com/Automattic/wp-calypso/issues/76981.

## Proposed Changes

* Attach the `Cancel subscription` button to the `useSiteUnsubscribeMutation` (handles both external and logged-in users).
* Introduce the `showSettings` state that sets to `false` when the unsubscribe mutation is successful. This is planned to be set to `true` on successful resubscribes (that will then make the settings visible again).
* Fix `Notice` component styling to match with the original Figma design (inDLaEQV8jJ21O4WXawIsJ-fi-1195_17698).

ℹ️ Please note the following:
- The PR does not implement the Notice's close `X` button as it didn't make sense to me to have it there. Closing the Notice would result in a page with no action available to the user (apart from navigating to the page with all Subscriptions).
- The subsequent PR will implement the `Resubscribe` button / link.

### Testing as an external user
1. Check out the PR and build the app.
2. Authenticate as an external user (`subkey` "dance" in the incognito browser window).
3. Make sure your external user is subscribed to some sites already.
4. Navigate to `http://calypso.localhost:3000/subscriptions/site/[blog_id]` where `[blog_id]` needs to be replaced with a specific blog ID your user is subscribed to.
5. When the page loads, you should be able to click on the "Cancel subscription" button on the bottom of the page.
6. Once the button is clicked, the user should be unsubscribed from the site (this can be checked at https://wordpress.com/subscriptions/sites), the settings should be hidden and the success notice should display:

Markup on 2023-05-17 at 13:12:26

### Testing as an logged-in user
1. Check out the PR.
2. Navigate to `client/server/pages/index.js` and temporarily add `return next();` to the line `833`:

Markup on 2023-04-28 at 17:50:44

3. Navigate to `client/landing/subscriptions/components/site-subscription-page/site-subscription-page.tsx` and replace the siteAddress with the address of the site you are trying to unsubscribe from:

Markup on 2023-05-17 at 13:11:00

4. Build the app.
5. Log in to a test WordPress.com account that has some existing Site subscriptions.
6. Navigate to `http://calypso.localhost:3000/subscriptions/site/[blog_id]` where `[blog_id]` needs to be replaced with a specific blog ID your user is subscribed to.
7. When the page loads, you should be able to click on the "Cancel subscription" button on the bottom of the page.
8. Once the button is clicked, the user should be unsubscribed from the site (this can be checked at https://wordpress.com/subscriptions/sites), the settings should be hidden and the success notice should display:

Markup on 2023-05-17 at 13:12:26

### Testing mutation error state
1. Follow the steps as above, but instead of navigating to `http://calypso.localhost:3000/subscriptions/site/[blog_id]` with correct `[blog_id]` used, replace it with non-existent one, e.g. `http://calypso.localhost:3000/subscriptions/site/hello-folks`.
2. Clicking the "Cancel subscription" button should result in an error message:

Markup on 2023-05-17 at 13:13:35

## Pre-merge Checklist

### DIFF
--- a/client/landing/subscriptions/components/site-subscription-page/site-subscription-page.tsx
+++ b/client/landing/subscriptions/components/site-subscription-page/site-subscription-page.tsx
@@ -2,9 +2,10 @@ import { Gridicon } from '@automattic/components';
 import { Reader, SubscriptionManager } from '@automattic/data-stores';
 import { Button } from '@wordpress/components';
 import { useTranslate } from 'i18n-calypso';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 import FormattedHeader from 'calypso/components/formatted-header';
+import { Notice, NoticeType } from 'calypso/landing/subscriptions/components/notice';
 import { SiteIcon } from 'calypso/landing/subscriptions/components/site-icon';
 import PoweredByWPFooter from 'calypso/layout/powered-by-wp-footer';
 import SiteSubscriptionSettings from './site-subscription-settings';
@@ -41,8 +42,18 @@ const SiteSubscriptionPage = () => {
 	} = data;
 	const [ showSettings, setShowSettings ] = useState( true );
 
-	const { mutate: unsubscribe, isLoading: unsubscribing } =
-		SubscriptionManager.useSiteUnsubscribeMutation();
+	const {
+		mutate: unsubscribe,
+		isLoading: unsubscribing,
+		isSuccess: unsubscribed,
+		error: unsubscribeError,
+	} = SubscriptionManager.useSiteUnsubscribeMutation();
+
+	useEffect( () => {
+		if ( unsubscribed ) {
+			setShowSettings( false );
+		}
+	}, [ unsubscribed ] );
 
 	if ( ! blogId || isError ) {
 		return <div>Something went wrong.</div>;
@@ -52,11 +63,6 @@ const SiteSubscriptionPage = () => {
 		// Full page Wordpress logo loader
 		return <div>Loading...</div>;
 	}
-
-	const handleUnsubscribe: () => void = () => {
-		setShowSettings( false );
-		unsubscribe( { blog_id: blogId, url: data.siteUrl } );
-	};
 
 	const subHeaderText =
 		subscribers > 1
@@ -86,6 +92,27 @@ const SiteSubscriptionPage = () => {
 						<FormattedHeader brandFont headerText={ siteName } subHeaderText={ subHeaderText } />
 					</header>
 
+					{ unsubscribed && (
+						<Notice type={ NoticeType.Success }>
+							{ translate(
+								'You have successfully unsubscribed and will no longer receive emails from %s.',
+								{
+									args: [ data.siteName ],
+									comment: 'Name of the site that the user has unsubscribed from.',
+								}
+							) }
+						</Notice>
+					) }
+
+					{ unsubscribeError && (
+						<Notice type={ NoticeType.Error }>
+							{ translate( 'There was an error when trying to unsubscribe from %s.', {
+								args: [ data.siteName ],
+								comment: 'Name of the site that the user tried to unsubscribe from.',
+							} ) }
+						</Notice>
+					) }
+
 					{ showSettings && (
 						<>
 							<SiteSubscriptionSettings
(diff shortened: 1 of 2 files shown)


client/landing/subscriptions/components/site-subscription-page/site-subscription-page.tsx
```diff
@@ -76,19 +92,49 @@ const SiteSubscriptionPage = () => {
...
+									args: [ data.siteName ],
+									comment: 'Name of the site that the user has unsubscribed from.',
+								}
+							) }
+						</Notice>
+					) }
+
+					{ unsubscribeError && (
+						<Notice type={ NoticeType.Error }>
+							{ translate( 'There was an error when trying to unsubscribe from %s.', {
```
ivan-ottinger, line 109: Hey @saygunnyc 👋🏼 Do we have a preferred copy that should display when unsubscribing fails for some reason?

The text you can see below is just temporary I have used so far:

Markup on 2023-05-17 at 12:01:28

Thank you!

client/landing/subscriptions/components/site-subscription-page/site-subscription-page.tsx
```diff
@@ -38,6 +40,20 @@ const SiteSubscriptionPage = () => {
 		deliveryFrequency,
 		subscribers,
 	} = data;
+	const [ showSettings, setShowSettings ] = useState( true );
```
mashikag, line 43: Do we anticipate the settings not to be displayed in some other cases, then unsubscribed? 🤔

ivan-ottinger: The reason I kept the `showSettings` state there (instead of relying on `unsubscribed` - as with the Success notice) was to have an easy way to display the settings again when the upcoming "Resubscribe" button is clicked and the user is subscribed again. 🤔 

Do you think that it should be better to simply rely on `unsubscribed`? I just felt that I don't have full control over its value, but I might be overthinking thia.

ivan-ottinger: To clarify a bit more: Let's imagine that the mutation is successful and hence `unsubscribed` is set to `true`. Then the user clicks the "Resubscribe" link (not implemented yet), but the `unsubscribed` hasn't changed. As a result, the settings would be still hidden.

mashikag: > Do you think that it should be better to simply rely on unsubscribed?

No. 😁 Your answer 100% satisfactory to me. It is just something that I did not really consider when asking the question. I thought about the `unsubscribed` const in terms of the subscription state, rather than the result of the call to unsubscribe API. 

Thanks for the explanation! 🙌

mashikag:
While it makes sense to not be able to remove the unsubscribe successful message, what about the error that the unsubscribe action was unsuccessful? I think we can display the X on those, what do you think? Otherwise, all looks good to me.

mashikag (approved):

mashikag (approved):

mashikag:
Also, I noticed that when we try to unsubscribe for a site that a user is not subscribed to with a logged-in user, we will receive a success message because of the response:
Screen Shot 2023-05-17 at 14 52 30

while if we do the same with the external user we will get an error:
Screen Shot 2023-05-17 at 15 08 12

However, I think this is still fine, because in theory we should not display the page if subscription does not exist.

ivan-ottinger:
> Also, I noticed that when we try to unsubscribe for a site that a user is not subscribed to with a logged-in user, we will receive a success message because of the response:

> However, I think this is still fine, because in theory we should not display the page if subscription does not exist.

That's a great catch! Thanks, Maciej!

> While it makes sense to not be able to remove the unsubscribe successful message, what about the error that the unsubscribe action was unsuccessful? I think we can display the X on those, what do you think? Otherwise, all looks good to me.

Good point! I have created a dedicated task for that: https://github.com/Automattic/wp-calypso/issues/77020.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/7991852

Thank you @ivan-ottinger for including a screenshot in the description! This is really helpful for our translators.

ivan-ottinger:
Just noting that the following fix has been addressed in https://github.com/Automattic/wp-calypso/pull/77033.

> While it makes sense to not be able to remove the unsubscribe successful message, what about the error that the unsubscribe action was unsuccessful? I think we can display the X on those, what do you think? Otherwise, all looks good to me.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/7991852).

PR #77021, opened 2023-05-17
Staging Sites: fix card buttons and long URL on mobile display

katinthehatsite:
* Closes https://github.com/Automattic/dotcom-forge/issues/2210

## Proposed Changes

This PR addresses inconsistencies in the "Manage staging site" and "Delete staging site" on mobile devices. Some issues it addresses are:

* if a staging site has a very long URL, it gets cut off on mobile devices on a staging card https://wordpress.com/hosting-config/{productionsite}
* on some mobile devices and in some browsers, the buttons "Manage staging site" and "Delete staging site" are not getting wrapped and overflow 

## Testing Instructions

**To test long URL being cut off**

This is mostly an edge case as it happens only when the URL is very long:

* Create a production Atomic site with a very long domain (40 characters is a good size, for example, `kathasasitewithaverylongdomainfortesting.wpcomstaging.com`)
* Go to `https://wordpress.com/hosting-config/{yourproductionsite}`
* Create a staging site that should also have a long URL
* Check the view of the long domain on mobile (using Chrome Dev Tools)
* Confirm that the URL does not overflow and goes on the next line:

Screenshot 2023-05-17 at 1 23 22 PM

This is what it looked like **before**:

Screenshot 2023-05-17 at 1 21 56 PM

**Issue 2 - Updated**:

For now, we decided to keep the original design on the desk

* Create a production Atomic site with a very long domain (40 characters is a good size, for example, `kathasasitewithaverylongdomainfortesting.wpcomstaging.com`)
* Go to `https://wordpress.com/hosting-config/{yourproductionsite}`
* Create a staging site
* Check the buttons "Manage staging site" and "Delete staging site"  maintain their original design on the desktop
* Switch to mobile view (using Chrome Dev Tools) is fine
* Check that the buttons "Manage staging site" and "Delete staging site" are stacked on the top of each other and fill the screen like so:

Screenshot 2023-05-17 at 4 37 37 PM

**Previous notes on possible design solution**:

Another part of the issue,  it was reported that on some screens the buttons "Manage staging site" and "Delete staging site" either overflow or the distance between the buttons and the sides of the screen is consistent on different mobile devices and browsers. You can see an example of the issue happening on this screencast:

https://github.com/Automattic/wp-calypso/assets/25575134/a15a619a-383c-4a0b-88fa-bb7396745b9b

This is something that is expected at the moment because we are using `display: 'flex'` for the container that has the buttons so if they fit on the screen, they will be placed there like so: `expected margin on the left + button + gap between buttons + second button + whatever space is left`. The key part of this is `whatever space is left` as this depends on the screen size of the used device: on some devices, there is more space and on some less which creates inconsistency of how much space is on each side of the button.

One option that we can do is to stack the buttons one on top of the other either by using `flex-wrap: wrap` on the parent container and `flex-grow` on the children. This change will affect how the buttons are displayed on the **bigger screens** though:

Screenshot 2023-05-17 at 3 18 56 PM

Screenshot 2023-05-17 at 3 18 28 PM

Screenshot 2023-05-17 at 3 18 50 PM

Another option would be to use a media query to also stack the buttons on top of each other but keep the size that they have now. This would preserve the original design but would be a less responsive solution:

Screenshot 2023-05-17 at 3 24 27 PM

The bigger screen remains unaffected in this case and is what it is now. 

@javierarce @danielbachhuber - what do you think would be a better approach here? 

## Pre-merge Checklist


katinthehatsite:
Quick comment: I have not implemented anything for the Issue 2 yet so the Calypso live link only works for testing the first issue.

katinthehatsite:
Update: 

I talked with @danielbachhuber  and we decided to keep the original design on the desktop but make the buttons fill in the whole screen on mobile. I updated the testing instructions in the PR for issue 2.

javierarce:
> I talked with @danielbachhuber and we decided to keep the original design on the desktop but make the buttons fill in the whole screen on mobile. I updated the testing instructions in the PR for issue 2.

Great, that was going to be my suggestion… but we should do the same for the other buttons on the page. I'll open an issue for that.

Should we also handle the issue with long titles here? I think we could add an ellipsis for this case.

Screenshot 2023-05-17 at 18 11 28

danielbachhuber:
> but we should do the same for the other buttons on the page. I'll open an issue for that.

Cool, sounds good.

> Should we also handle the issue with long titles here? I think we could add an ellipsis for this case.

I just spent a half hour trying to get it to work but I wasn't able to 🙃 Let's handle this at some point in the future.

danielbachhuber (approved):

danielbachhuber:
> I just spent a half hour trying to get it to work but I wasn't able to

It works with a fixed width on the element:

```diff
diff --git a/client/my-sites/hosting/staging-site-card/index.js b/client/my-sites/hosting/staging-site-card/index.js
index 34d9021b9e..8190cac03c 100644
--- a/client/my-sites/hosting/staging-site-card/index.js
+++ b/client/my-sites/hosting/staging-site-card/index.js
@@ -72,6 +72,11 @@ const SiteNameContainer = styled.div( {
 const SiteName = styled.a( {
 	fontWeight: 500,
 	marginInlineEnd: '8px',
+	textOverflow: 'ellipsis',
+	overflow: 'hidden',
+	whiteSpace: 'nowrap',
+	display: 'inline-block',
+	maxWidth: '220px',
 	'&:hover': {
 		textDecoration: 'underline',
 	},
```

I don't want to set a fixed width, though.

katinthehatsite:
> I don't want to set a fixed width, though.

@danielbachhuber I changed `SiteNameContainer` from `display: flex` to `display: block` and it sort of seemed to work but I don't know if it is a good approach. Let me know what you think!

Alternatively, I am wondering if we need the staging label at all there since it is clear for the customers from the other information that is already there that the site is staging...

danielbachhuber:
> I changed `SiteNameContainer` from `display: flex` to `display: block` and it sort of seemed to work but I don't know if it is a good approach. Let me know what you think!

@katinthehatsite It changes the design slightly, but I don't think it solves the ellipsis thing.

> Alternatively, I am wondering if we need the staging label at all there since it is clear for the customers from the other information that is already there that the site is staging...

I will defer to @javierarce on this one. Worth noting we have a slightly different variation on the design for the Sites page:

image

I'm not sure if we should try to create a unified component at some point...

PR #76987, opened 2023-05-16
Site Editor Domain Upsell: Add callout using ETK V2

lupus2k:
Fixes https://github.com/Automattic/wp-calypso/issues/76321
Fixes https://github.com/Automattic/wp-calypso/issues/73445

This is a new version of [this PR](https://github.com/Automattic/wp-calypso/pull/76855), because it was reverted (p1684250921623969-slack-C02FMH4G8)

## Proposed Changes
Add Domain Upsell callout to Gutenberg using [editing toolkit](https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/README.md)

#### Conditions to show the callout
- User has a verified email
- Site has no custom domain
- Site is not a P2
- Not show if has `siteIntent` (like start-writing, link-in-bio, etc)

#### Track events
- View: `calypso_block_editor_domain_upsell_callout_view`
- Click: `calypso_block_editor_domain_upsell_callout_click`
- Dismiss: `calypso_block_editor_domain_upsell_callout_dismiss`

## Testing Instructions

- You need to sandbox your testing site domain
- Go to `/apps/editing-toolkit`
- Run `yarn dev --sync`
- On Calypso, go to `/posts` (and `/pages`) and edit or create new.
- Or, change the view to classic and use `/wp-admin`
- You should see the callout if conditions matches.
- Check CTA
- Check dismiss (you can re-enable it by running `wp.data.dispatch( 'core/preferences' ).set( 'core/edit-post', 'domain_upsell_callout_dismiss', false );` in the console and reloading the page.
- Check track events

To test on Atomic Sites, follow the instructions here: PCYsg-ly5-p2 (**ETK and Atomic Sites**)

## Screenshots
image
image
image
image
image

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
[...]

PR #77040, opened 2023-05-17
Check & fix e2e tests after ETK merge

lupus2k:
This is a follow-up of https://github.com/Automattic/wp-calypso/pull/76987

We want to check and fix (if needed) the e2e tests after merging ETK's new version.


dpasque (approved):
Approving this for now -- if CI runs green, we should be good to merge!

PR #77046, opened 2023-05-18
React 18

### LINKED ISSUE #75167: Tooling: Add Storybook for `client/*`
# Overview

Currently, we have some Storybooks under the `packages` directory (e.g., [tour-kit](https://github.com/Automattic/wp-calypso/blob/1afe45179cfa7ac98bcdb30dd4d739cb4589f019/packages/tour-kit/.storybook/main.js), [components](https://github.com/Automattic/wp-calypso/blob/1afe45179cfa7ac98bcdb30dd4d739cb4589f019/packages/components/.storybook/main.js#)), but no Storybook for under the `client` directory.

**I would suggest that we can have Storybook under the `client` directory.** This way, we can add a "story" for every piece of UI (for components/pages/etc).

# Context

Without Storybook, we need to run a local server when developing UI-related things, which is slow to debug and not easy to test all the conditions.

I would be happy to be able to develop based on Storybook at a larger granularity (e.g., per page) without limiting Storybook to so-called "components" only. https://storybook.js.org/docs/react/get-started/why-storybook#build-uis-in-isolation

There are several advantages to using Storybook for UI development.

- Rapid development: Since components are isolated, developers can quickly prototype and iterate on their designs without waiting for a full build and refresh cycle, which can be time-consuming with a local server. 
- Organizing UI: Complex conditions can be organized and developed by component, which is helpful in finding, understanding, testing, and debugging UI patterns
- Storybook makes component testing easy with its play function, addon, etc. https://storybook.js.org/docs/react/writing-tests/introduction

# Tasks

_Some tasks may be better isolated as another task._

- [ ] CI
	- [ ] Run build on PR
	- [ ] Deploy to GitHub pages (or somewhere)
		- it'd be nice to have Storybook deployed for each PR with GitHub pages or something.
- [ ] Tips & Documentation
	- [ ] Update 
		- [ ] https://github.com/Automattic/wp-calypso/blob/trunk/docs/components.md
		- [ ] https://github.com/Automattic/wp-calypso/blob/trunk/docs/component-libraries.md
	- [ ] General guidelines to use Storybook (how to run it? how to write a story?)
	- [ ] Direction on how best to use Storybook.
	- [ ] Retire Live Docs in devdocs https://wpcalypso.wordpress.com/devdocs/design
- [ ] Add some stories
	- There are several stories that I would like to add as examples. The settings in preview.js might change accordingly.
	- [x] A component that uses redux store https://github.com/Automattic/wp-calypso/pull/75942
	- [x] A component from another package https://github.com/Automattic/wp-calypso/pull/75943
	- [ ] A component that has images
- [ ] Migrate to [Storybook 7.0](https://storybook.js.org/blog/storybook-7-0/)
- [ ] Addons
	- [ ] [addon-essentials](https://storybook.js.org/docs/react/essentials/introduction)
	- [ ] [Figma integration](https://storybook.js.org/docs/react/sharing/design-integrations)
	- [ ] [Accessibility addon](https://storybook.js.org/addons/@storybook/addon-a11y).

# Related Resources

- Heads-up: The “Universal” Storybook Project for Calypso – codep2 p7H4VZ-4gp-p2
- Which Button should I use when working with the Calypso repo? pdKhl6-1hu-p2
	- https://github.com/Automattic/wp-calypso/issues/71143
- Slack conversation p1680657358609729-slack-C03DY3RJ5
- Test Diversification: UI Component Tests – Software Testing p6fDka-Ol-p2
	- https://github.com/WordPress/gutenberg/blob/trunk/test/storybook-playwright/README.md

noahtallen:
This PR updates Calypso to React 18, and updates WordPress packages to the first version which supported React 18 (which is 4-5 months old at this point.)

**Note: this does not yet enable the new rendering mode -- that will likely be a follow-up due to the size of this PR!**

### Prerequisite or split-out work:
- https://github.com/Automattic/wp-calypso/pull/77121
- #78292 
 
### Notable breaking changes which impact us (excluding concurrent mode)
- Various React types have been deprecated and changed
  - ReactChild, ReactFragment, and ReactText have been deprecated
  - `useCallback` now requires types for the callback arguments, which means most of those need to be updated
  - `children` isn't automatically a prop, so we need to add it, which means everywhere we use children needs to be updated

### To do:
- Smoke test everything!

### Done: (🤞 )
- [x] SSR does not appear to be working in prod (calypso.live) due to `useSyncExternalStore` being called without a 3rd argument in `use-subscription`. (This breaks SSR.)
- [x] Fix an issue where TextEncoder wasn't defined in ReactDOMServer, causing notifications panel to not open.
- [x] ~Importing `parseWithAttributeSchema` from `@wordpress/blocks` crashes the application -- something related to the unlock/lock private APIs. This unfortunately prevents the editor routes from loading.~ (Will potentially need to fix this in a future WP package update)
- [x] Import react dom server from browser bundles
- [x] Fix issue with SSR and layout effects
- [x] Fix babel issue related to `declare context` use (breaking docker build)
- [x] Fix unit tests
- [x] Fix all type issues!
- [x] Fix this type issue related to `createElement`: `Argument of type 'ForwardRefExoticComponent<ButtonProps & RefAttributes<any>>' is not assignable to parameter of type 'string | FunctionComponent<BaseButtonProps & _ButtonProps & Omit<Omit<DetailedHTMLProps<ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>, "ref">, "disabled" | ... 1 more ... | keyof BaseButtonProps> & RefAttributes<...>> | ComponentClass<...>'.` ([example](https://github.com/Automattic/wp-calypso/blob/b29d7fb87c1ff882f6adaa1a50598ad917f669d5/packages/subscriber/src/components/add-form/index.tsx#L430-L433)) -- doesn't seem to happen any more
- [x] Fix type issue related to `ExternalLink` and `useInterpolateElement` (punted)
- [x] Add `stream` browserify to calypso-build and client 
- [x] Fix peer dependency violation warnings (doesn't block working on other issues)
- [x] Updated types related to children/ReactNode -- `@types/react` has changed to be more strict about this.
- [x] Updated some types related to event handlers and useCallback.
- [x] For some components, use `ComponentProps< typeof Component >` now that prop types aren't directly available for wp components.
- [x] Add the `stream` package -- apparently this was needed for react-dom? Webpack wouldn't compile without it.
- [x] Updated WordPress components usage to use variant rather than `isX`. (E.g. `variant="tertiary"` rather than `isTertiary`)
- [x] Updated package versions across the repo (React, WordPress, and testing library)
- [x] Tried to stub out global styles so that we can work on other things, until we find a solution for this.
- [x] Added some `@types/` packages required by `@wordpress/components` (See https://github.com/WordPress/gutenberg/pull/50231 for more info, but we'll need to do this in Calypso until we get a fix in core.)
- [x] Fix type issue with internal translations related to deprecated ReactChild
- [x] Fix type issue with certain store selectors and actions
- [x] Fix type issues with search results in the help center
- [x] Update stripe packages for compatibility with React 18
- [x] Fix `@automattic/global-styles` -- it relies extensively on private file imports (e.g. `@wordpress/edit-site/build-module/some/deep/path`), which have been drastically changed internally.


apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nav-sidebar/src/components/create-page/index.tsx
```diff
@@ -8,13 +8,6 @@ import { get } from 'lodash';
 
 import './style.scss';
 
-const Button = ( {
```
noahtallen: This doesn't appear to do anything anymore

packages/help-center/src/components/help-center-contact-form.tsx
```diff
@@ -268,23 +268,18 @@ export const HelpCenterContactForm = () => {
...
 				title: preventWidows( decodeEntities( result.title ) ),
 			} );
 
 			if ( result.blog_id ) {
-				params.set( 'blogId', result.blog_id );
+				params.set( 'blogId', result.blog_id.toString() );
 			}
 
-			navigate( `/post/?${ params }`, searchResult );
+			navigate( `/post/?${ params }` );
```
noahtallen, line 227: As far as I can tell, search result doesn't do anything. I'm double checking with the original author to be sure, though.

packages/help-center/src/components/help-center-header.tsx
```diff
@@ -85,8 +85,8 @@ const HelpCenterHeader = ( { isMinimized = false, onMinimize, onMaximize, onDism
 								element={ __( 'Contact Options', __i18n_text_domain__ ) }
 							/>
 							<Route path="/inline-chat" element={ __( 'Live Chat', __i18n_text_domain__ ) } />
-							<Route path="/contact-form" element={ SupportModeTitle } />
-							<Route path="/post" element={ ArticleTitle } />
+							<Route path="/contact-form" element={ <SupportModeTitle /> } />
```
noahtallen, line 88: This component expects to receive a rendered element now

packages/i18n-calypso/types/index.d.ts
```diff
@@ -56,7 +60,7 @@ export type TranslateOptionsPluralText = TranslateOptionsPlural & { textOnly: tr
 
 // Translate hooks, like component interpolation or highlighting untranslated strings,
 // force us to declare the return type as a generic React node, not as just string.
-export type TranslateResult = React.ReactChild;
```
noahtallen: The ReactChild types are deprecated, and they also don't seem to work any more. Since the intention originally was to avoid using ReactNode because of its falsey type options, I changed to use ExistingReactNode which filters the types on ReactNode.

tyxla:
Great work @noahtallen!  🚀 Thank you!

I think we should first address the peer dep situation in #77121 which is quite complex on its own 😅 I'll devote some time to helping there as well with whatever I can.

noahtallen:
@tyxla I made a bunch of changes to the yarnrc file, including reworking how we ignore certain warnings. It should both be much more clear what our discard statements are doing, and should also dynamically cover more scenarios that are the same issue. I also documented the packages we're waiting for `@wordpress/` to update. I also moved a few of the discard statements into the `packageExtensions` area, because I think that's a more ideal way to solve errors where that works.

I also ignored the warnings related to the date picker package we decided not to update just yet.

Finally, I also synced our react-virtualized fork (https://github.com/Automattic/react-virtualized/pull/3) and published the update to npm, ultimately incorporating it here. 

This means all the violations/warnings happening on `yarn install` are resolved now!

noahtallen:
I added a bunch of little commits trying to get some insight into why yarn install is failing in CI and not locally. Here's what's happening. ([TeamCity link](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10134423?buildTab=log&focusLine=349&logView=linear&linesState=295))

```sh
RUN  cat packages/data-stores/package.json && yarn explain peer-requirements p031cd && yarn install && yarn explain peer-requirements p031cd && cat packages/data-stores/package.json

### packages/data-stores/package.json
{
  "name": "@automattic/data-stores",
  ...
   "peerDependencies": {
    ...
     "react": "^17.0.2", ### WTF???
   }
}

### First explain peer requirements (it's fine for some reason?)
YN0000: @automattic/domain-picker@workspace:packages/domain-picker [abd57] provides react-dom@npm:18.2.0 [35689] with version 18.2.0, which satisfies the following requirements:
YN0000: @automattic/data-stores@workspace:packages/data-stores [abd57] → ^18.2.0 ✓

### Yarn install output:
┌ Resolution step
│ @automattic/domain-picker@workspace:packages/domain-picker [abd57] provides react-dom (p031cd) with version 18.2.0, which doesnt satisfy what @automattic/data-stores requests
...

### Second explainer for same ID, now broken!!
@automattic/domain-picker@workspace:packages/domain-picker [abd57] provides react-dom@npm:18.2.0 [441bf] with version 18.2.0, which doesnt satisfy the following requirements:
@automattic/data-stores@workspace:packages/data-stores [abd57] → ^17.0.2 ✘

### packages/data-stores/package.json
{
  "name": "@automattic/data-stores",
  ...
   "peerDependencies": {
    ...
     "react": "^17.0.2", ### incorrect!!!
   }
}
```

Some of these symptoms are very similar to p9oQ9f-1I5-p2. This is so bizarre

tyxla:
> This means all the violations/warnings happening on yarn install are resolved now!

Glad to hear - nice work!

> I added a bunch of little commits trying to get some insight into why yarn install is failing in CI and not locally. Here's what's happening. 

For what it's worth I've tried on both my machines to repro it and could not. I wonder if we have some sort of a cache on the TeamCity level that we need to bust 🤔

noahtallen:
Yeah, if I check the contents of a package.json file outside of the docker build, it is correct and matches the branch. But inside the build, after running `COPY . /calypso/`, the contents of the file are different. Suggesting that Docker is copying files from something other than the current branch. (The layer cache is not used, and both docker build and my "outside the build file check" run from the same directory. The `--no-cache=true` flag also makes no difference, which makes sense as the layer cache is not being used here for the COPY step anyways.)

noahtallen:
Ok, the issue is _not_ happening any more. I rebased the PR and squashed a ton of commits, and the weird CI issue with peer dependency warnings is resolved.

creativecoder:
I'm testing this with my [isolated-block-editor in Reader PR](https://github.com/Automattic/wp-calypso/pull/76035/files) (rebasing the PR from this branch), and I'm seeing this error in the JS console. Is this still expected?

```
Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot [backend.js:6:2315](moz-extension://50f4159d-9276-c945-bd28-93ecf5f31579/build/backend.js)
    r moz-extension://50f4159d-9276-c945-bd28-93ecf5f31579/build/backend.js:6
    React 3
    render app:///./controller/web-util.js:12
    middleware app:///../node_modules/page/page.js:1170
    nextEnter app:///../node_modules/page/page.js:713
    makeLayoutMiddleware app:///./controller/shared.js:59
    middleware app:///../node_modules/page/page.js:1170
    nextEnter app:///../node_modules/page/page.js:713
    following app:///./reader/controller.js:195
    middleware app:///../node_modules/page/page.js:1170
    nextEnter app:///../node_modules/page/page.js:713
    sidebar app:///./reader/controller.js:152
    middleware app:///../node_modules/page/page.js:1170
    nextEnter app:///../node_modules/page/page.js:713
    updateLastRoute app:///./reader/controller.js:117
    middleware app:///../node_modules/page/page.js:1170
    nextEnter app:///../node_modules/page/page.js:713
    redirectLoggedOutToSignup app:///./controller/index.web.js:194
    ...
```

tyxla:
> I'm testing this with my [isolated-block-editor in Reader PR](https://github.com/Automattic/wp-calypso/pull/76035/files) (rebasing the PR from this branch), and I'm seeing this error in the JS console. Is this still expected?

It's expected - React 18 adds a new concurrent rendering mode and a new API to support it, so it will display that warning if you try to use the old one. The old one is still supported, though, and we're going to be using it for the time being in Calypso. Since we're not aiming to change to the new API in this PR, the answer is yes: it's still expected.

noahtallen:
See p1686704845701869-slack-C7YPUHBB2 for a status update

packages/i18n-calypso/types/index.d.ts
jsnajdr: I think you want to avoid `ReactNode` here, because that's a very wide type that includes all things you can pass as `children`. That includes `undefined`, arrays, iterables etc.

Then there's the much narrower type for things that can be returned from a `render` function. That's `ReactElement | null`. You can't even return strings or numbers from a `render` function -- although it works perfectly at runtime, it's a type error and you need to wrap them in fragments.

`TranslateResult` wants to be somewhere in between. What if it was `ReactElement | string`? That should cover all cases. We don't want `TranslateResult` to be `null`-able, because we sometimes call `.toString()` on it etc. and I think it would cause a lot of unwanted type errors.

noahtallen: Sounds like a great idea!

noahtallen:
Following-up on the thoughts mentioned in p4TIVU-aCA-p2, it'd be worth taking the time to split out some code. Even though I've spent a lot of effort on the PR, I think to make this manageable, we might need to delete some of that progress if I can't easily separate it out.

The two big ideas I have are:

- Move the components work to another PR. This will be tough because I did a fair amount of work on type issues related to it. But we should pin components to a much lower version to avoid the type upgrade which isn't strictly needed for this PR.
- Try to handle some of the React type updates in another PR. For example, we can probably migrate away from every instance of `ReactText` and `ReactChild` without needing the React update. 

Otherwise, a lot of this is probably intertwined enough it might need to stay in this PR

packages/i18n-calypso/types/index.d.ts
noahtallen: One thing that isn't clear to me is what we actually expect the type to be. 

Here are the React 18 type definitions (updated since React 17!):

```ts
type ReactNode = ReactElement | string | number | ReactFragment | ReactPortal | boolean | null | undefined

type ReactFragment = Iterable<ReactNode>
interface ReactPortal extends ReactElement {
  key: Key | null;
  children: ReactNode;
}
```

`ReactNode` is supposed to represent anything that React can render. (Though I suppose that means anything you can put in JSX, not anything you can return from render!) In this case, I filtered out the null, undefined, and boolean types to create the (admittedly oddly named) `ExistingReactNode` type, which leaves us with `ReactElement | ReactFragment | string | number` (considering that ReactPortal is just a more specific type of `ReactElement.` 

But since the result of a translate operation is a variable you would put into jsx (e.g. `<p>{translate( 'hello world ')}</p>`), it seems like any truthy `ReactNode` result would work fine.

Since the current state of the PR is accepting `ReactElement | string | number | ReactFragment | ReactPortal`, I don't know if further filtering the type to just `ReactElement | string` would resolve any of the type issues.

Most of the issues I've been running into are that the input args often don't strictly exist! Take this example:

```tsx
{ translate( 'Establishing a connection to {{strong}}%(targetSite)s{{/strong}}.', {
  args: {
    targetSite: targetSite ? targetSite : siteSlug,
  },
  components: {
    strong: <strong />,
  },
} ) }
```

The type issue here is that `siteSlug` is nullable. (Because `useSelector( getSelectedSiteId )` can return null.) Though I suspect that while under this code path, a site is probably selected, the types aren't strong enough to see that.

Similar issues account for many of the errors I've been working through.

noahtallen: I guess the core question is whether there are any limitations on what substitutions can handle (beyond non-nullable etc). If it can handle inputs that are ReactPortal or ReactFragment, it seems fine to accept those

The core reason we're getting errors now is that the React 17 typed `ReactFragment` as `{} | Iterable<ReactNode>`, which is a super loose type. That's been fixed now, which probably actually makes our types work more correctly now

noahtallen:
I split out #78292 to include type fixes for react types. I'll probably have to revert some stuff in this PR, but that's ok.

tyxla:
> I split out #78292 to include type fixes for react types. I'll probably have to revert some stuff in this PR, but that's ok.

Nice to see #78292 land - I hope it won't cause too much trouble getting this branch up to date. 

Do you see #78350 as a blocker too, or is it one that we can address separately from this one?

noahtallen:
> Do you see https://github.com/Automattic/wp-calypso/pull/78350 as a blocker too, or is it one that we can address separately from this one?

Not a blocker at all! I see it as completely separate.

> I hope it won't cause too much trouble getting this branch up to date.

for now, I began working on getting the branch up to date by branching out again to avoid overwriting anything in the current branch: https://github.com/Automattic/wp-calypso/pull/78451 (with the ultimately goal to bring everything back into this PR)

noahtallen:
@tyxla Alright, I finally got everything rebased. 

Beyond integrating the type work on trunk, here are the other things added now:
- I pinned every WordPress package, not just components. This was necessary for downgrading components to work in the first place. I think we should continue pinning them in the future too.
- I added a bunch more tsconfig references. I'm not sure why these errors have been popping up in this PR and not trunk. In the future, we should add something to enforce explicit dependencies and tsconfig references!
- Added `stream` package fallback in webpack, required to get various apps to build.
- I reverted the global styles changes. These were related to WordPress updating, and won't be an issue until after this PR, when we update WordPress packages again (we'll have to re-open that PR then: #77319)
- Deduped and resolved all peer dependency issues (again)

And worked on a **ton of type issues**. Looks like all these are passing in this PR now too!

I can run calypso locally now, but it looks like the docker build is failing in this branch. The cause is that I needed to add `declare context` in a legacy class component to get types working, but babel isn't happy about transforming this. I tried adding `allowDeclareFields`, and that worked locally, but it didn't seem to work in CI. So that's the next step.

After that, I think the unit tests would be good to work on, along with e2e tests once the docker build is working. 

Finally, it's fixing any major issues we find while testing! (Such as the notifications panel not loading.)

tyxla:
> I can run calypso locally now, but it looks like the docker build is failing in this branch. The cause is that I needed to add declare context in a legacy class component to get types working, but babel isn't happy about transforming this. I tried adding allowDeclareFields, and that worked locally, but it didn't seem to work in CI. So that's the next step.

~Resolved this in https://github.com/Automattic/wp-calypso/pull/77046/commits/d678031f83cd54d49e858ece142e3d070a8b98ff. Basically it was all about fixing the babel config presets syntax when providing options for the preset.~ Scratch that. It resolved for me locally, but not for CI yet again.  Will have to dig further.

tyxla:
In https://github.com/Automattic/wp-calypso/pull/77046/commits/28030999c9b220f8c02fbdba4d08b7cafe3be815 I've walked around the issue by using type assertions for now so we don't block the already huge PR. 

The docker image now builds properly.

tyxla:
I've pushed a couple of commits to fix all failing package and client unit tests too - they're now green ✅ 

The only remaining failures should be e2e tests. I won't have a chance to work on them today, so feel free to commandeer @noahtallen 🤝

noahtallen:
> I've walked around the issue by using type assertions for now so we don't block the already huge PR.

Fair enough! Since we only had it in one place, not a huge problem IMO.

> I've pushed a couple of commits to fix all failing package and client unit tests too - they're now green ✅

Very happy to hear! I'll start looking into the e2e issues next 👍

noahtallen:
I made some progress fixing issues:
- Fixed a warning about `useLayoutEffect` in SSR. i thought this might fix the e2e failure, but it didn't.
- Fixed the notifications panel not loading by changing how we import `ReactDOMServer`.
- Fixed unit test failures caused by the above by adding `TextEncoder` to our tests.

The main e2e failure now is that SSR doesn't work. Unfortunately, SSR is working _locally_, but it doesn't seem to work in calypso.live. (Making it hard to troubleshoot!) The pages load, but they are client-side rendered so it's a bit slower. No clear indication of any issues in the browser console or e2e logs. (The SSR tests check for a `data-calypso-ssr` attribute only added by the SSR pipeline -- I've verified this isn't getting added and that pages are being client-side rendered only in calypso.live.)

Probably the biggest issue troubleshooting this is that we don't capture the calypso _server_ logs anywhere for calypso.live, so we can't see the specific error. (We only persist add staging and production logs to logstash.)

So we might only be able to reproduce by running Calypso more close to production locally, which is always tricky to set up! But I think this is the main issue to resolve before we can mark this PR ready for review.

noahtallen:
Oddly enough, I can replicate it locally now:

```sh
01:01:23.380Z ERROR calypso: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering. (reqId=fca10cf8-1459-4b9b-8f86-9412305e0551, url=/themes/, env=development, userAgent="Firefox 114", path=/themes/)
    Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
        at Object.useSyncExternalStore (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5218:11)
        at Object.useSyncExternalStore (/Users/noahallen/source/wp-calypso/build/node_modules/react/cjs/react.development.js:1676:21)
        at useSubscription (/Users/noahallen/source/wp-calypso/build/node_modules/use-subscription/cjs/use-subscription.development.js:37:15)
        at useRtl (/Users/noahallen/source/wp-calypso/build/webpack:/wp-calypso/packages/i18n-calypso/src/rtl.js:22:24)
        at Popover (/Users/noahallen/source/wp-calypso/build/webpack:/wp-calypso/packages/components/src/popover/index.jsx:376:22)
        at renderWithHooks (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5662:16)
        at renderIndeterminateComponent (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5735:15)
        at renderElement (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5950:7)
        at renderNodeDestructiveImpl (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6108:11)
        at renderNodeDestructive (/Users/noahallen/source/wp-calypso/build/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6080:14)
```

noahtallen:
The [react docs](https://react.dev/reference/react/useSyncExternalStore#reference) state the third argument to `useSyncExternalStore` is:

> optional `getServerSnapshot`: A function that returns the initial snapshot of the data in the store. It will be used only during server rendering and during hydration of server-rendered content on the client. The server snapshot must be the same between the client and the server, and is usually serialized and passed from the server to the client. If you omit this argument, rendering the component on the server will throw an error.

Looking at the source code for `use-subscription`, which is where the error originates (via `useRtl`):

https://github.com/facebook/react/blob/8ec962d825fc948ffda5ab863e639cd4158935ba/packages/use-subscription/src/useSubscription.js#L29, it doesn't include that. So I guess we can't use that package at all in SSR code in the future!

But I also found that with our update to `use-subscription` v1.8, it basically just wraps `useSyncExternalStore`, and I'm guessing we don't want that behavior until we migrate to concurrent rendering. So I think we can just downgrade it to 1.6 to get the "legacy" behavior.

noahtallen:
Success! All checks are green, and I've verified SSR is working along with the RTL subscription hook.

Next up, we just need to do a lot of manual testing!

tyxla:
I'm doing some thorough testing this week and will be fixing things as I find them. For now, after quite some thorough testing of logged-in features, I've only managed to find issues that also exist in `trunk` like #78679 and #78678 😅

tyxla:
I've also done some signups through various flows, including purchases. Couldn't find any issues so far! 🟢 

I think we should be able to land this, but I'd recommend doing it early in the day of the one who lands it, so we'll have the opportunity to monitor and revert if something suddenly goes bad.

tyxla (approved):
Amazing work @noahtallen, let's 🚢 it when you're ready!

arthur791004 (approved):
Tested the post-signup onboarding flow, especially for the Design Picker and Assembler screen, and both are still working well. Nice job 🎉🎉🎉

noahtallen:
I did further testing of the Calypso apps like ETK, notifications, and wpcom-block-editor, and these are also working well.

noahtallen:
Next up, updating WordPress packages to latest: #78711

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8353914).

PR #77039, opened 2023-05-17
Plan Comparison Grid: Update breakpoints in Onboarding

jeyip:
Related to https://github.com/Automattic/martech/issues/1530

## Proposed Changes

* Previously, the same breakpoints were being applied to both the plan features grid **and** the plan comparison grid
* These changes add new, plan comparison grid specific breakpoints
* Updates the number of plans that we render at each breakpoint

## Testing Instructions

* Check out this branch and run `yarn start`
* Navigate to calypso.localhost:3000/start
* Choose some free domain and land on the plans page
* Verify that the number of plans shown at each breakpoint is correct
* Verify that the width of each plan column and the gutter space between columns is correct

**ONBOARDING FLOW**
- From 1600px to 1281px, show all plans. 
- From 1280px to 1024px, show 4 plans. 
- From 1023px to 880px, show 3 plans
- From 880px to mobile, we show 2 plans. 

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #77047, opened 2023-05-18
Plan Comparison Grid: Update breakpoints in /plans

jeyip:
Related to https://github.com/Automattic/martech/issues/1530

## Proposed Changes

* Previously, the same breakpoints were being applied to both the plan features grid **and** the plan comparison grid
* These changes add new, plan comparison grid specific breakpoints
* Updates the number of plans that we render at each breakpoint

## GIFS
BEFORE

AFTER

## Testing Instructions

* Check out this branch and run `yarn start`
* Navigate to calypso.localhost:3000
* Navigate to Upgrades > Plans
* Verify that the number of plans shown at each breakpoint is correct
* Verify that the width of each plan column and the gutter space between columns is correct

### DIFF
--- a/client/my-sites/plan-features-2023-grid/index.tsx
+++ b/client/my-sites/plan-features-2023-grid/index.tsx
@@ -297,32 +297,34 @@ export class PlanFeatures2023Grid extends Component<
 			selectedFeature,
 		} = this.props;
 		return (
-			<div className="plans-wrapper">
-				<QueryActivePromotions />
-				<div className="plan-features">
-					<div className="plan-features-2023-grid__content">
-						<div>
-							<div className="plan-features-2023-grid__desktop-view">
-								{ this.renderTable( planProperties ) }
-							</div>
-							<div className="plan-features-2023-grid__tablet-view">
-								{ this.renderTabletView() }
-							</div>
-							<div className="plan-features-2023-grid__mobile-view">
-								{ this.renderMobileView() }
+			<>
+				<div className="plans-wrapper">
+					<QueryActivePromotions />
+					<div className="plan-features">
+						<div className="plan-features-2023-grid__content">
+							<div>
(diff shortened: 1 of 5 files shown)


southp:
The gaps between the description texts and the CTA buttons are gone:

image

It's not the case on the parent branch :)

client/my-sites/plan-features-2023-grid/index.tsx
```diff
@@ -311,32 +311,34 @@ export class PlanFeatures2023Grid extends Component<
...
-						<div>
-							<div className="plan-features-2023-grid__desktop-view">
-								{ this.renderTable( planProperties ) }
-							</div>
-							<div className="plan-features-2023-grid__tablet-view">
-								{ this.renderTabletView() }
-							</div>
-							<div className="plan-features-2023-grid__mobile-view">
-								{ this.renderMobileView() }
+			<>
```
southp: This looks like the DOM structure update we reverted in https://github.com/Automattic/wp-calypso/pull/77039/commits/661333abced19f27c62b1513e861d4dbb9b23e5e. If we still agree it's not necessary, let's remove this update :)

southp (changes requested):

client/my-sites/plan-features-2023-grid/index.tsx
jeyip: Ah I thought I'd done this, but it's my mistake. Totally agree with the DOM structure update. I'll update this PR now

jeyip: Fixed! [28f2f99](https://github.com/Automattic/wp-calypso/pull/77047/commits/28f2f9985bde96be603d8f3095cd49782de0029e)

southp (approved):
Thanks for addressing all the feedback. It works as described and the code also LGTM :)

PR #77053, opened 2023-05-18
fixes to sharing modal and enable for English users

roo2:
I had some issues on deploying the sharing modal last time. p1684314937793019-slack-C025Q5RK2

* SVG icons where not loaded from the production CDN 
  * The svg `<use>` tag does not support loading from external domains 
  * https://github.com/w3c/svgwg/issues/707
* Modal was visible on P2s and private blogs

This PR fixes those two issues and re-enables the sharing modal.
See https://github.com/Automattic/wp-calypso/pull/76267
Screen Shot 2023-05-18 at 2 22 32 pm

The SVG issue required inlining a copy of the social logos svg into the ETK

### Testing instructions.
* sandbox a test site
* Apply to your sandbox with `cd apps/editing-toolkit && yarn dev --sync`
* Publish a new post
* verify that no request is made to an svg file named `social-logos-d55401f99bb02ebd6cf4.svg` or similar.
* sandbox a p2 or private site.
* publish a post and verify the modal is not shown
* see instructions from https://github.com/Automattic/wp-calypso/pull/76267


apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx
```diff
@@ -30,6 +32,9 @@ const SharingModal: React.FC = () => {
 	const { launchpadScreenOption } = window?.launchpadOptions || {};
 	const { isDismissed, updateIsDismissed } = useSharingModalDismissed( isDismissedDefault );
 	const { __ } = useI18n();
+	const localeSlug = useLocale();
+	const isEnglish = localeSlug?.startsWith( 'en' );
+	const isPrivateBlog = window?.wpcomGutenberg?.blogPublic === '-1';
```
eoigal, line 37: I don't see how this will block on P2. If the P2 is public, I _think_ it will still show.. is that ok?

davemart-in: I suspect that is fine. I suspect the primary issue is to not show it on private blogs and P2s since you'd never really want to promote private posts.

eoigal: Yeah, if p2 is public the share modal pops up. Think that's ok but just wanted to double check;

Screenshot 2023-05-18 at 17 10 47

eoigal (approved):
WFM. Nice job

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx
Addison-Stavlo: > I suspect the primary issue is to not show it on private blogs and P2s since you'd never really want to promote private posts.
> > Yeah, if p2 is public the share modal pops up. Think that's ok

I agree here as well. 👍

Addison-Stavlo (approved):
Looks good and tests well on my end as well!

apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx
roo2: yep, I didn't even think about P2s being public but I agree that the thing we care about is private blogs. A public P2 would probably want to share and if not they can dismiss it.

PR #77072, opened 2023-05-18
Stepper Framework: Un-deprecate the `goBack()` navigation function

p-jackson:
## Proposed Changes

Originally discussed in https://github.com/Automattic/wp-calypso/pull/76854#issuecomment-1549382116, this PR un-deprecates the `goBack()` navigation function.

`goBack()` is really just a special version of `submit()` i.e. it can be achieved by calling `submit( { goBack: true } )`. But as was brought up in the above PR, it'd be nice for all the flows to use a consistent flag name for step submissions that represent going backwards. And if they all have a consistent flag name, isn't that the same thing as having an explicit `goBack()` function?

Another reason to favour a `goBack()` function is that it's more type safe. The `ProvidedDependencies` parameter passed to `submit()` isn't strongly typed, so you have to check the value of the `goBack` flag without knowing whether it'll be there. Whereas you know the `goBack()` function is always being called by someone who wanted to go back.
It doesn't solve it for all the other cases where `ProvidedDependencies` has to be accessed, but it's something.

CC @Automattic/vertex and @edanzer who did the work to deprecate this method in the first place.

I'm open to discussion. There's definitely something appealing about the idea of only having one function for moving the state-machine to another state.

## Testing Instructions

I've tested the `/setup/new-hosted-site` flow but I'm not sure how to test the domain upsell flow. I think it's liked to from one of the launch pads?

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/domain-upsell.ts
+++ b/client/landing/stepper/declarative-flow/domain-upsell.ts
@@ -46,6 +46,12 @@ const domainUpsell: Flow = {
 			window.location.assign( location );
 		};
 
+		function goBack() {
+			if ( currentStep === 'plans' ) {
+				navigate( 'domains' );
+			}
+		}
+
 		async function submit( providedDependencies: ProvidedDependencies = {} ) {
 			switch ( currentStep ) {
 				case 'domains':
@@ -62,9 +68,6 @@ const domainUpsell: Flow = {
 					navigate( 'plans' );
 
 				case 'plans':
-					if ( providedDependencies?.goBack ) {
-						return navigate( 'domains' );
-					}
 					if ( providedDependencies?.goToCheckout ) {
 						const planCartItem = getPlanCartItem();
 						const domainCartItem = getDomainCartItem();
@@ -85,7 +88,7 @@ const domainUpsell: Flow = {
 			}
 		}
 
-		return { submit, exitFlow };
+		return { submit, exitFlow, goBack };
 	},
 };
 
--- a/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
+++ b/client/landing/stepper/declarative-flow/hosting-site-creation-flow.ts
@@ -42,6 +42,12 @@ const hosting: Flow = {
 
 		const flowName = this.name;
 
+		const goBack = () => {
+			if ( _currentStepSlug === 'plans' ) {
+				navigate( 'options' );
+			}
+		};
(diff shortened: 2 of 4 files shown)


alshakero (approved):
Looks great. Thanks for the elaborate explanation. Yes, I'm really into finite states design, but this isn't far from it. Unless someone starts passing parameters to `goBack` which would really throw a wrench.

p-jackson:
> Unless someone starts passing parameters to `goBack` which would really throw a wrench.

`submit()` is still as powerful as it used to be, so hopefully if there are any special versions of "back" that require parameters then we'll remember to fall back to using `submit()` for that special case.

p-jackson:
I discovered how to test the `domain-upsell` step. It's part of the `/setup/free` flow.

CleanShot 2023-05-19 at 10 32 21@2x

The back button works as expected (it takes the user back to the launchpad)

edanzer:
Hey - Looks like this was just merged! But wanted to post a note anyways. You won't get any strong objections from me. 

The idea behind deprecating goBack() at the time was that effortlessly navigating back and forth between stepper screens meant that each screen would need a lot of conditional logic to be able to handle users at different stages/steps/states in the flow or onboarding journey. I think there was some thought that the same sort of thing had made the prior onboarding framework a bit messy and there was a desire to encourage a way of thinking about stepper that avoided that. 

Even at the time, though, I understood the counterpoints, and could have gone either way.

p-jackson:
Thanks @edanzer, I hear ya! Hopefully this will still achieve what want by keeping all the branchy navigation code out of the components and keeping each flow's "go back" logic separate 🤞

client/landing/stepper/declarative-flow/domain-upsell.ts
```diff
@@ -46,6 +46,12 @@ const domainUpsell: Flow = {
 			window.location.assign( location );
 		};
 
+		function goBack() {
```
simison, line 49: @alshakero noted _"Unless someone starts passing parameters to goBack which would really throw a wrench."_:

Could this be added as a comment to guide future developers?

(Edit: meant to live the comment for `client/landing/stepper/declarative-flow/internals/types.ts` file but you get the idea :-) )

alshakero: Done in https://github.com/Automattic/wp-calypso/pull/77168

PR #77095, opened 2023-05-18
E2E: Fix failing specs for the Gutenberg v15.8.0 release

WunderBart:
#### Fix Gutenberg atomic E2E tests edge (mobile):
- specs/editor/editor__schedule.ts: Editor: Schedule: Schedule: future
- specs/editor/editor__post-advanced-flow.ts: Editor: Advanced Post Flow: Revert post to draft
- specs/editor/editor__privacy-private.ts: Editor: Privacy (Private): Create a Private page
- specs/editor/editor__privacy-password.ts: Editor: Privacy (Password): Create a Password page
- specs/editor/editor__privacy-public.ts: Editor: Privacy (Public): Create a Public page
- specs/editor/editor__revisions.ts: Editor: Revisions
- specs/blocks/blocks__core-jetpack-extended.ts: Blocks: Jetpack Extended Core Blocks: Validating blocks in published post.: Instagram
- specs/editor/editor__page-basic-flow.ts: Editor: Basic Post Flow
- specs/editor/editor__post-basic-flow.ts: Editor: Basic Post Flow: Preview

### DIFF
--- a/packages/calypso-e2e/src/lib/blocks/block-flows/instagram.ts
+++ b/packages/calypso-e2e/src/lib/blocks/block-flows/instagram.ts
@@ -37,15 +37,15 @@ export class InstagramBlockFlow implements BlockFlow {
 	 * @param {EditorContext} context The current context for the editor at the point of test execution
 	 */
 	async configure( context: EditorContext ): Promise< void > {
-		const editorParent = await context.editorPage.getEditorParent();
-		const embedUrlLocator = editorParent.locator( selectors.embedUrlInput );
+		const editorCanvas = await context.editorPage.getEditorCanvas();
+		const embedUrlLocator = editorCanvas.locator( selectors.embedUrlInput );
 		await embedUrlLocator.fill( this.configurationData.embedUrl );
 
-		const embedButtonLocator = editorParent.locator( selectors.embedButton );
+		const embedButtonLocator = editorCanvas.locator( selectors.embedButton );
 		await embedButtonLocator.click();
 
 		// We should make sure the actual Iframe loads, because it takes a second.
-		const instagramIframeLocator = editorParent
+		const instagramIframeLocator = editorCanvas
 			.frameLocator( selectors.editorInstagramIframe )
 			.locator( 'body' );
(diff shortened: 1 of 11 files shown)


packages/calypso-e2e/src/lib/blocks/block-flows/instagram.ts
```diff
@@ -37,15 +37,17 @@ export class InstagramBlockFlow implements BlockFlow {
 	 * @param {EditorContext} context The current context for the editor at the point of test execution
 	 */
 	async configure( context: EditorContext ): Promise< void > {
-		const editorParent = await context.editorPage.getEditorParent();
-		const embedUrlLocator = editorParent.locator( selectors.embedUrlInput );
+		const editorCanvas = await context.editorPage.getEditorCanvas();
```
fullofcaffeine, line 40: It looks like we want to use `getEditorCanvas` for anything that renders inside the editor (because of the canvas iframe), right? This seems to be the case for simple and AT as well.

fullofcaffeine:
I've run each test locally for simple,AT x mobile,desktop. They all passed. I'll merge this and re-run E2Es builds for simple and AT. Thanks for wrangling this, @WunderBart!

packages/calypso-e2e/src/lib/blocks/block-flows/instagram.ts
WunderBart: It will probably be the main cause of failing tests until we update the parent selectors in the tests so that the frame context is correct (editor parent, aka Gutenframe vs. editor canvas). For elements that are located within the canvas, we should always be using the canvas frame as the root, and for the elements outside - the editor parent frame. It’s been the same with the upstream GB tests ever since the canvas iframe was introduced. Currently, in Calypso, not all selectors are scoped like that because part of the framework/tests was written before the iframe was introduced. Once we update those parts (like those that are failing now), it should stop failing regardless if the canvas iframe is there or not, thanks to https://github.com/Automattic/wp-calypso/pull/76085.

PR #77100, opened 2023-05-18
Limited Global Styles: Add link to reset styles support doc

mmtr:
## Proposed Changes

Adds a link to the limited Global Styles notice in the post editor and in the launchbar that opens the support doc explaining how to reset styles.

Launchbar | Post editor
--- | ---
Screenshot 2023-05-18 at 16 35 58 | Screenshot 2023-05-18 at 16 16 24

## Testing Instructions

- Apply these ETK changes to your sandbox
- Sandbox a site with limited Global Styles
- Visit the frontend
- Make sure the launchbar notice contains a "Reset custom styles" link
- Click on it
- Make sure it opens the expected support doc
- Visit the open editor
- Make sure the post editor global notice contains a "Reset custom styles" link
- Click on it
- Make sure it opens the expected support doc

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
@@ -27,6 +27,7 @@ document.addEventListener( 'DOMContentLoaded', () => {
 	const upgradeButton = container.querySelector( '.launch-bar-global-styles-upgrade' );
 	const previewButton = container.querySelector( '.launch-bar-global-styles-preview' );
 	const closeButton = container.querySelector( '.launch-bar-global-styles-close' );
+	const resetButton = container.querySelector( '.launch-bar-global-styles-reset' );
 
 	const limitedGlobalStylesNoticeAction =
 		localStorage.getItem( 'limitedGlobalStylesNoticeAction' ) ?? 'show';
@@ -66,5 +67,11 @@ document.addEventListener( 'DOMContentLoaded', () => {
 			action: checkbox.checked ? 'show' : 'hide',
 		} );
 		window.location = previewButton.href;
+	} );
+
+	resetButton?.addEventListener( 'click', ( event ) => {
+		event.preventDefault();
+		recordEvent( 'wpcom_global_styles_gating_notice_reset' );
+		window.open( resetButton.href, '_blank' ).focus();
 	} );
 } );
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
@@ -71,7 +71,7 @@ document.addEventListener( 'DOMContentLoaded', () => {
 
 	resetButton?.addEventListener( 'click', ( event ) => {
 		event.preventDefault();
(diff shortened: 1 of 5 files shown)


BogdanUngureanu:
Can we use the external URL icon we use for external links (e.g. the learn more, the front-end preview, etc)?

BogdanUngureanu:
I wonder, wouldn't it be better to rename the link to something neutral like "Learn more" (since we don't actually reset the styles, we are just pointing them to the docs)?

BogdanUngureanu:
Also, as an alternative, what if we add a link to the site-editor that would open the GS panel? This might be useful for Post Editor. The link would be called something like "Manage styles". This way, both modification, and resetting are covered by the same link.

mmtr:
> rename the link to something neutral like "Learn more" (since we don't actually reset the styles, we are just pointing them to the docs)?

We're linking them though to a specific section in the Styles docs about removing the styles. "Learn more" would be a better fit if we weren't linking to that specific section.

The idea behind the [?] icon is to somehow set the expectation that this link will provide some help about how to perform such action. 

> Also, as an alternative, what if we add a link to the site-editor that would open the GS panel?

We discarded that option because the "reset styles" button is pretty hidden in the site editor:

> I opted to not link to the editor, where I’m sure they’ll get lost and have no idea how to reset styles.
>
> pekYwv-1qM-p2

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
```diff
@@ -67,4 +68,10 @@ document.addEventListener( 'DOMContentLoaded', () => {
 		} );
 		window.location = previewButton.href;
 	} );
+
+	resetButton?.addEventListener( 'click', ( event ) => {
+		event.preventDefault();
+		recordEvent( 'wpcom_global_styles_gating_notice_reset' );
```
BogdanUngureanu: Shouldn't the event be `calypso_global_styles_gating_notice_reset_support_click`? 🤔

BogdanUngureanu (changes requested):
The link looks good to me on the front-end and in the post editor. However, I think the event we trigger on the front-end should be renamed.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/wpcom-global-styles-view.js
mmtr: I wanted to follow the same naming format we're using for the other events in the frontend: `wpcom_global_styles_gating_notice_upgrade` and `wpcom_global_styles_gating_notice_preview`. We don't use the `_click` suffix in those, and there is only one word after the `wpcom_global_styles_gating_notice_` prefix.

BogdanUngureanu: Oh I see. It should probably be prefixed with `wpcom_`. The thing that confused me was the event name implies that it's triggered when the user actually resets the styles, like the one from the site-editor: `calypso_global_styles_gating_notice_reset_click`. 

But we don't do that, we just point them to support docs. What if we rename it to `wpcom_global_styles_gating_notice_reset_support`? It would be similar to the event from the Post Editor :)

mmtr: Fair enough!

BogdanUngureanu (approved):
Nice work! Let's 🐑  it!

PR #77103, opened 2023-05-18
Added P2 label to the list of comments in Subscription Manager

JuanLucha:
## Proposed Changes

This PR adds the P2 label to the list of comments in the new Subscription Manager, as seen in the Figma file: inDLaEQV8jJ21O4WXawIsJ-fi-712_18319

## Testing Instructions

1. Apply this PR.
2. Go to this file `client/server/pages/index.js` and change the call back in `app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {` to it ends up like this:
```js
		return next();
	} );
```
3. Log in to your Calypso local and go to `http://calypso.localhost:3000/subscriptions/comments`.
4. If your user has P2 and non-P2 site subscriptions, you should see the P2 tag beside the site's name accordingly.

### DIFF
--- a/client/landing/subscriptions/components/comment-list/comment-row.tsx
+++ b/client/landing/subscriptions/components/comment-list/comment-row.tsx
@@ -24,6 +24,7 @@ const CommentRow = ( {
 	date_subscribed,
 	forwardedRef,
 	style,
+	is_wpforteams_site,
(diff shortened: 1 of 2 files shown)


yansern (approved):
Thanks!

PR #77105, opened 2023-05-18
Fixed pending tab not loading for logged in users in Subscription Manager

### LINKED ISSUE #77079: Subscription Manager: Subkey not found error when user is logged in.
### Quick summary

When a logged-in user goes to the `pending` tab in Subscription Manager, they will find this in the console sometimes:

Screenshot 2023-05-18 at 12 04 43

We need to find out how to reproduce this and fix it.

It could be related to the check for the `subkey` cookie. That was supposed to be done only when this feature was available for external users. Now that this feature is open for all users, the check should change accordingly.

### Steps to reproduce

Note: there is no guarantee this process will reproduce the problem. Part of the problem is I only sometimes found this problem.

1. log in to WordPress.com.
2. Go to `https://wordpress.com/subscriptions/comments`.
3. Change to the `pending` tab. You should see the loading screen for some time.
4. Before the loading is finished, change back to the `comments` tab. You should see the problem in the console.

No error in the console.

### What actually happened

This error appears in the console:
Screenshot 2023-05-18 at 12 04 43

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/77079

## Proposed Changes

**The problem**

When a logged-in user goes to the `pending` tab in Subscription Manager, they will find this in the console sometimes:

Screenshot 2023-05-18 at 12 04 43

**The solution**

Turns out the problem was the queries to retrieve the pending site and comment subscriptions ignoring the logged status of the user:
Screenshot 2023-05-18 at 18 40 26

As you can see, the variable `isLoggedIn` wasn't being passed to the `callApi` method. Additionally, the nameSpace was required too, because being a logged-in user, the method called to make the request inside `callApi` is `wpcomRequest`. Internally, if the namespace is not passed, the URL of the request is formed like this:

fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Qcebkl%2Swdhrel%2Qcebivqre.wf%3Se%3Qq5r14q9q%23144-og

Our endpoint doesn't live there, but in `wpcom/v2` namespace, that's why we have to pass this parameter too:

Screenshot 2023-05-18 at 18 47 12

## Testing Instructions

Apply this PR.
As an external user, go to `https://wordpress.com/subscriptions/pending` with the browser console open.
Check that it works and that there is no error message in the console.
Change to other tabs and check the same.
As an internal user, go to `https://wordpress.com/subscriptions/pending` with the browser console open.
Check that it works and that there is no error message in the console.
Change to other tabs and check the same.

### DIFF
--- a/packages/data-stores/src/reader/queries/use-pending-post-subscriptions-query.ts
+++ b/packages/data-stores/src/reader/queries/use-pending-post-subscriptions-query.ts
@@ -13,30 +13,33 @@ type SubscriptionManagerPendingPostSubscriptionsQueryProps = {
 	sort?: ( a?: PendingPostSubscription, b?: PendingPostSubscription ) => number;
 };
 
-const callPendingBlogSubscriptionsEndpoint =
-	async (): Promise< PendingPostSubscriptionsResult > => {
-		const pendingPosts = [];
-		const perPage = 1000; // TODO: This is a temporary workaround to get all pending subscriptions. We should remove this once we decide how to handle pagination.
+const callPendingBlogSubscriptionsEndpoint = async (
+	isLoggedIn: boolean
+): Promise< PendingPostSubscriptionsResult > => {
+	const pendingPosts = [];
+	const perPage = 1000; // TODO: This is a temporary workaround to get all pending subscriptions. We should remove this once we decide how to handle pagination.
 
-		const incoming = await callApi< SubscriptionManagerPendingPostSubscriptions >( {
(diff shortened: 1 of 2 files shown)


yansern:
This might explain the discrepancy of having pending count 1 but having no pending sites list. In our dev environments, we may have subkey set while logged in.

yansern (approved):
Kudos for debugging this.

PR #77112, opened 2023-05-18
Hosting configuration: poll Atomic transfer status when loading the page

### LINKED ISSUE #72371: [Bug]: UI gets stuck after activating hosting features
### Quick summary

When I activate hosting features, the page gets stuck on the “Please wait while we activate the hosting features.“ message. Hosting features seem to be activated, but the page doesn't reload.

### Steps to reproduce

1. Create a site with a Business plan
2. Navigate to Settings -> Hosting Configuration
3. Activate hosting features and confirm domain change

The page reloads after the process is done.

### What actually happened

"Please wait while we activate the hosting features." is displayed all the time.

### Browser

Google Chrome/Chromium

### Context

_No response_

Atomic

### Other notes

_No response_

### Reproducibility

Intermittent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

After some time, I can refresh page manually.

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2151. Closes https://github.com/Automattic/wp-calypso/issues/72371. 

## Proposed Changes

This PR tries to fix the problem where the current WPCOM tab doesn't get notified whenever the transfer to Atomic is completed.

It fetches the status on page load and also re-fetches the site whenever the transfer is completed.

The goal is to create an abstraction (i.e. to re-use the transfer status polling logic both in /hosting-configuration and /sites. In the future, as we split HC into multiple pages, this hook and abstraction will be much appreciated since AT transfer is a crucial part of our hosting features activation and management. It also counts as maintenance work.

## Testing Instructions

In both cases, you can follow the progress by opening the dev tools and filtering by `automated-transfers` in the Network tab.

### Without leaving the current tab

1. Pick a Business site;
3. Go to `/hosting-config`;
4. Click "Activate" the Business features;
5. Verify that after a while the site is transferred to Atomic and the page automatically reflects that.

### Refreshing the tab while the transfer is in progress

1. Pick a Business site;
2. Go to `/hosting-config`;
3. Click "Activate" the Business features;
4. Refresh the page;
5. Verify that after a while the site is transferred to Atomic and the page automatically reflects that.

### DIFF
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -72,6 +72,15 @@ class Hosting extends Component {
 		this.props.fetchAutomatedTransferStatus( this.props.siteId );
 	}
 
+	componentDidUpdate( prevProps ) {
+		const { COMPLETE } = transferStates;
+		const { transferState, requestSiteById, siteId } = this.props;
+
+		if ( prevProps.transferState !== COMPLETE && transferState === COMPLETE ) {
+			requestSiteById( siteId );
+		}
+	}
+
 	render() {
 		const {
 			teams,
@@ -81,7 +90,6 @@ class Hosting extends Component {
 			isECommerceTrial,
 			isWpcomStagingSite,
 			isTransferring,
-			requestSiteById,
 			siteId,
 			siteSlug,
 			translate,
@@ -112,10 +120,6 @@ class Hosting extends Component {
 				( isTransferring && COMPLETE !== transferState ) ||
 				( isDisabled && COMPLETE === transferState )
 			) {
-				if ( COMPLETE === transferState ) {
-					requestSiteById( siteId );
-				}
-
 				let activationText = translate( 'Please wait while we activate the hosting features.' );
 				if ( this.state.clickOutside ) {
 					activationText = translate( "Don't leave quite yet! Just a bit longer." );
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -113,7 +113,11 @@ class Hosting extends Component {
 		};
 
 		const getAtomicActivationNotice = () => {
-			const { COMPLETE, FAILURE } = transferStates;
+			const { COMPLETE, FAILURE, RELOCATING_REVERT } = transferStates;
+
+			if ( transferState === RELOCATING_REVERT ) {
+				return null;
+			}
 
 			// Transfer in progress
 			if (
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -69,12 +69,7 @@ class Hosting extends Component {
 	}
 
 	componentDidMount() {
-		const { COMPLETE } = transferStates;
-		// Check if a reverted site still has the COMPLETE status
-		if ( this.props.transferState === COMPLETE ) {
-			// Try to refresh the transfer state
-			this.props.fetchAutomatedTransferStatus( this.props.siteId );
-		}
+		this.props.fetchAutomatedTransferStatus( this.props.siteId );
 	}
 
 	render() {
--- a/client/my-sites/hosting/main.js
+++ b/client/my-sites/hosting/main.js
@@ -5,8 +5,7 @@ import {
 	FEATURE_SITE_STAGING_SITES,
 } from '@automattic/calypso-products';
 import { localize } from 'i18n-calypso';
-import { Component, Fragment } from 'react';
-import wrapWithClickOutside from 'react-click-outside';
+import { Fragment, useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import DocumentHead from 'calypso/components/data/document-head';
 import QueryKeyringConnections from 'calypso/components/data/query-keyring-connections';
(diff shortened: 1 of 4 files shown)


client/state/automated-transfer/selectors/is-automated-transfer-active.js
```diff
@@ -10,8 +10,15 @@ import 'calypso/state/automated-transfer/init';
  * @param {string} status name of current state in automated transfer
  * @returns {?boolean} is transfer currently active? null if unknown
  */
-export const isActive = ( status ) => ( status ? status === transferStates.START : null );
```
zaguiini, line 13: We have several transitioning states (e.g. `provisioned`, `relocating_switcheroo`) that indicate that the transfer is in progress. I'm flipping this condition so that it won't break if we add more steps in the future.

client/my-sites/hosting/main.js
```diff
@@ -69,11 +69,15 @@ class Hosting extends Component {
 	}
 
 	componentDidMount() {
+		this.props.fetchAutomatedTransferStatus( this.props.siteId );
```
zaguiini: The problem with calling this method is that it duplicates polling when transferring by clicking "Activate" and not reloading the page.

In this scenario, [`initiateThemeTransfer`](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/themes/actions/theme-transfer.js#L56) and [`receiveStatus`](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/data-layer/wpcom/sites/automated-transfer/status/index.js#LL25C1-L25C1) are polling the back-end (which duplicates the number of necessary requests).

But that's not all: _both_ are dispatching tracks events whenever the transfer gets completed. This is not a bad thing per se, but in our situation, it's not desired. We should find a way of not dispatching duplicated Tracks events only in this scenario.

zaguiini (commented):
If I were to be completely honest, I'm not a big fan of this implementation. The end goal is to abstract all this [read and poll transfer status] functionality and reuse it in https://github.com/Automattic/wp-calypso/pull/76949#issuecomment-1553376685.

But let's take one step at a time, I guess.

client/state/automated-transfer/selectors/is-automated-transfer-active.js
vykes-mac: > If I were to be completely honest, I'm not a big fan of this implementation. The end goal is to abstract all this [read and poll transfer status] functionality and reuse it in https://github.com/Automattic/wp-calypso/pull/76949#issuecomment-1553376685.
> 
> But let's take one step at a time, I guess.

I have thought about this and it seem much effort to achieve at the moment, It seem we would have to convert the hosting page to functional component to get some reusable api that doesn't involve creating/modifying the set of redux actions. It would be easier with hooks.

client/my-sites/hosting/main.js
vykes-mac: > In this scenario, [initiateThemeTransfer](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/themes/actions/theme-transfer.js#L56) and [receiveStatus](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/data-layer/wpcom/sites/automated-transfer/status/index.js#LL25C1-L25C1) are polling the back-end (which duplicates the number of necessary requests).

Yup! both those actions should be using the same polling mechanism to DRY out the code.

mpkelly: I commented on Slack too. I think the problem comes from us using [initiateThemeTransfer](https://github.com/Automattic/wp-calypso/blob/trunk/client/state/themes/actions/theme-transfer.js#L56) to check for an Atomic transfer. IMO, It would be better to have an `initiateAtomicTransfer` and a new endpoint that could be polled for Atomic site transfers.

vykes-mac: This will attempt to fetch even for simple sites, I guess that might not be an issue but it's something I noticed while working on the `/sites` transfer status loader.

vykes-mac:
when I revert an atomic site it's stuck here. Even after refreshing many times. The fact that the status is now set to `reverted` is throwing it off.

image
 

image

zaguiini:
This PR is blocked by https://github.com/Automattic/wp-calypso/pull/77186, which will remove the [polling limitation](https://github.com/Automattic/wp-calypso/pull/77112#discussion_r1198162805), bringing us one step closer to our goal of abstracting and reusing this logic on the Sites page.

client/state/automated-transfer/selectors/is-automated-transfer-active.js
p-jackson: Is this change to `isActive` still relevant now that we're using a hook to get the transfer state? And if it is, should we import the list of end states from `use-atomic-transfer-query.ts`

client/my-sites/hosting/main.js
```diff
@@ -52,189 +51,176 @@ import WebServerSettingsCard from './web-server-settings-card';
...
+		hasStagingSitesFeature,
+	} = props;
+
+	useEffect( () => {
+		function onClickOutside( event ) {
 			event.preventDefault();
 			event.stopImmediatePropagation();
-			this.setState( { clickOutside: true } );
+			setHasClickedOutside( true );
+			window.removeEventListener( 'click', onClickOutside );
```
p-jackson, line 81: This `removeEventListener` doesn't look necessary because setting `hasClickedOutside=true` will already cause the `useEffect` to re-run (`hasClickedOutside` is in the deps array). And since the `useEffect` re-runs it means the cleanup function will get called.

client/my-sites/hosting/main.js
```diff
@@ -264,4 +250,4 @@ export default connect(
 		fetchAutomatedTransferStatus,
 		requestSiteById: requestSite,
 	}
-)( localize( wrapWithClickOutside( Hosting ) ) );
+)( localize( Hosting ) );
```
p-jackson, line 248: Wow, `wrapWithClickOutside` is some truly old school React 😅. It would be possible to keep using this 3rd party library—just in the spirit of keeping the refactor smaller. To do so we could implement the required `handleClickOutside` component method using the seldom used `React.useImperativeHandle()` hook!

I think it's good as it is though if it tests fine. May as well take the opportunity of a big class-to-function refactor to make the change.

client/state/atomic-transfer/hooks/use-atomic-transfer-query.ts
```diff
@@ -0,0 +1,61 @@
...
+	transferStates.FAILURE,
+	transferStates.ERROR,
+	transferStates.REVERTED,
+] as const;
+
+export const useAtomicTransferQuery = (
+	siteSlug: SiteSlug,
+	{ refetchInterval }: UseAtomicTransferQueryOptions
+) => {
+	const { data } = useQuery( [ 'sites', siteSlug, 'atomic', 'transfers', 'latest' ], {
```
p-jackson, line 39: Maybe a nit pick, but what about a cache key like `['sites', siteSlug, 'atomic-transfers', 'latest']`

I think it makes more sense when deciding how the cache should be invalidated.
Invalidating `['sites',siteSlug,'atomic-transfers']` would invalidate all transfers, whether it's the latest or not
Using the key as written, what would invalidating `['sites', siteSlug, 'atomic']` mean? Invalidate transfer state, GitHub state, SSH keys, all at the same time?

client/state/atomic-transfer/hooks/use-atomic-transfer-query.ts
```diff
@@ -0,0 +1,61 @@
...
+interface UseAtomicTransferQueryOptions {
+	refetchInterval?: number;
+}
+
+const endStates = [
+	transferStates.NONE,
+	transferStates.COMPLETE,
+	transferStates.FAILURE,
+	transferStates.ERROR,
+	transferStates.REVERTED,
```
p-jackson, line 32: Should this list of end states also include `RELOCATING_REVERT`? I don't know how to produce this state, but [the comment doc](https://github.com/Automattic/wp-calypso/blob/a76fb98cf130dd021d70ce9378fa5f7b75e9b5e4/client/state/automated-transfer/constants.ts#L19) makes it seems like it's the same sort of thing as `NONE` and `REVERTED`

client/state/automated-transfer/selectors/is-automated-transfer-active.js
p-jackson: Further to whether this is still relevant, it seems this selector's change in behaviour is what's causing the unit test failure

zaguiini:
Thanks for the comments, @p-jackson! This is still in the works, but with a lower priority. I'll try to give it some attention tomorrow.

zaguiini:
@p-jackson I think that the work done in https://github.com/Automattic/wp-calypso/pull/77433 is enough to include the component in this PR. There is going to be more information added, but the core functionality is there (in `WithAtomicTransfer`).

Issue #4336, opened 2016-03-26
Editor: insert/edit link keyboard shortcut doesn't work


danhauk:
When writing a post in the editor the insert/edit link keyboard shortcut does nothing. Other shortcuts (bold, italic, etc.) work as expected. Clicking the link button in the MCE panel works as expected.

#### Steps to reproduce
1. Start a new post or edit an existing post.
2. Press Cmd/Ctrl + k to insert a link

#### What I expected
I expected the link modal window to appear so I can add a link to the post.

#### What happened instead
Nothing happens.

#### Browser / OS version
OS X 10.11.3
Chrome 49.0.2623.108 & Safari 9.0.3

lancewillett:
I'm able to repeat this with Safari 9.1 on OS X 10.11.4.

Also reported by a Linux desktop app user in https://github.com/Automattic/wp-desktop/issues/161

lancewillett:
Hmm, this was supposedly fixed with https://github.com/Automattic/wp-calypso/pull/1747 — regression perhaps?

DigitalDavo:
Confirmed on version 1.3.2-alpha version of Wordpress.com desktop app on windows too. This is almost the most important feature for using Calypso as a blogging tool.

lancewillett:
@DigitalDavo You can add links in other ways: highlight a word, paste (as long as URL is in clipboard). Highlight word and use toolbar icon to add link.

DigitalDavo:
Thanks @lancewillett , I never knew I could just paste over the top. I was so used to CTRL-K since Windows Live Writer days that I was happy that it worked in previous versions of calypso desktop.

aduth:
Fixed by #6097, props @zpnk

Issue #15, opened 2015-11-17
Security: Add Self XSS Warning to Dev Console


hoverduck:
Add a warning in the Javascript console similar to this example from www.facebook.com
![image](https://cloud.githubusercontent.com/assets/7233112/11216543/b98d7088-8d19-11e5-98f9-0aafc65e32ee.png)

erickstm87:
Hey I'd like to take this one. It would be my first post. Not too familiar with xss attacks though. Is there a specific portion of the site where this warning would be needed?

chrispwns:
Is this still open? This seems like a relatively easy fix and I'd love to have a chance at it. Assuming it's open would something like the attached image work for you? I went ahead and threw that together really quick just to see how console formatting worked in JS because it interested me.
![self-xss-warning](https://cloud.githubusercontent.com/assets/10478216/20857420/95ae2220-b8f1-11e6-9ea6-3a52b4b5af18.png)

PR #77121, opened 2023-05-19
Update packages to versions which include React 18 as peer dependencies

noahtallen:
This is a step of #77046 which can be merged earlier. It updates dependencies such that everything includes React 18 as a peer dependency.

This doesn't include _every_ package -- some are more tightly coupled with React 18, which are included in the original PR. The packages I update here should work both before and after that one, making it easier to split out the work.

How to find these packages? In #77046, "simply" run `yarn explain peer-requirements ...` for various peer dependency violation warnings to find packages which need to be updated. 

## Packages being updated:
- react-popper
- react-redux **(major)**
- react-query **(major)** (one or two places didn't get updated to the new version yet)
- html-to-react
- qrcode.react **(major)**
- ~react-day-picker **(major)**~
- react-element-to-jsx-string **(major)**
- react-live **(major)**
- redux-dynamic-middlewares
- react-modal
- all `@storybook` packages **(major)**
- all `@stripe` packages **(major)**
- use-subscription
- reakit

## Potential issues:
After looking through the changelogs, these could impact us:
- react-redux API changes. Unsure if we use anything they removed.
- react-query in the one or two packages which haven't been updated.
- qrcode.react changed the way it exports things, so I updated these.
- storybook: This update was very large, but I got it mostly working. However, there may be small issues I didn't catch. (Also note that on first boot of the main project, it takes a while to compile.)

## Probable issues:
- react-day-picker now uses **date-fns** and changed some other things. (https://react-day-picker.js.org/guides/upgrading) Our internal use of some of its utilities will have to change. EDIT: We removed this for now.


noahtallen:
Currently stuck on:

```sh
#15 79.55 [webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve 'react/jsx-runtime' in '/calypso/node_modules/react-live/dist'

[19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_932_932&logView=flowAware)
  #15 79.55 Did you mean 'jsx-runtime.js'?

[19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_933_933&logView=flowAware)
  #15 79.55 BREAKING CHANGE: The request 'react/jsx-runtime' failed to resolve only because it was resolved as fully specified

[19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_934_934&logView=flowAware)
  #15 79.55 (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').

[19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_935_935&logView=flowAware)
  #15 79.55 The extension in the request is mandatory for it to be fully specified.

[19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_936_936&logView=flowAware)
  #15 79.55 Add the extension to the request.
```

tyxla (commented):
> Currently stuck on:
> 
> ```shell
> #15 79.55 [webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve 'react/jsx-runtime' in '/calypso/node_modules/react-live/dist'
> 
> [19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_932_932&logView=flowAware)
>   #15 79.55 Did you mean 'jsx-runtime.js'?
> 
> [19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_933_933&logView=flowAware)
>   #15 79.55 BREAKING CHANGE: The request 'react/jsx-runtime' failed to resolve only because it was resolved as fully specified
> 
> [19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_934_934&logView=flowAware)
>   #15 79.55 (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
> 
> [19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_935_935&logView=flowAware)
>   #15 79.55 The extension in the request is mandatory for it to be fully specified.
> 
> [19:01:31 ](https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/10045651?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=10045651_936_936&logView=flowAware)
>   #15 79.55 Add the extension to the request.
> ```

Should be resolved by https://github.com/Automattic/wp-calypso/pull/77121/commits/f60bb2914f7541e7efcc137c70e958e766152a0e

jsnajdr:
Let's also merge #77151 first, and then when rebasing this PR, we can remove all the `apps/inline-help` changes.

tyxla:
Okay, I've done some updates:

- Resolving the `react/jsx-runtime` webpack module resolution errors.
- Fixing unit tests
- Updating storybook to use the new version CLI
- Fixing storybook config

This advanced how far in the pipeline we succeeded, but the PR is now blocked by some `react-redux` type inconsistencies. 

It seems like in the newest major `react-redux` version, types are quite messed up. For example `DefaultRootState` no longer exists, and in order to use the typed versions of `useSelector`, `useDispatch` and `useStore`, we'll likely need to do some type definition module magic. That's where I'm pausing, but it may be a good place for you to continue.

noahtallen:
Alright, I pushed this a fair bit further:

1. Converted state/index.js to typescript
2. Added useSelector and useDispatch wrapper functions with type info
3. Changed every import of those two functions (in `.ts` or `.tsx` files under client) to import from `state/calypso` instead of from react-redux.
4. Removed uses of the removed default store type
5. Added manual types for the state in a few `connect` calls which were missing it.

I also noticed we have two different store types. `AppState` is defined globally and is just `any`. One is `IAppState` and has a few reducers typed, so I used that as the default.

Some remaining issues from `yarn tsc --noEmit --project client/tsconfig.json`:
- Somewhere along the way, `yarn build-packages` started failing, only after running `yarn clean` (possibly related to other internal type issues below)
- tsc is complaining about missing type definitions in JS files imported by state/index.ts 
- There is an issue with `payment_method` types in purchases, not sure why it's coming up now. (Potentially related to async thunk stuff below?)
- Issues with `store.getState()` in one or two places.
- `setStore` type issues
- The dispatch/thunk types don't account for async.

Re: async thunk. I didn't quite get to solving this, but the core issue is that the `ThunkDispatch` doesn't always work. In particular, there are action creators with return types typed in JSDoc which are not handled by the `ThunkDispatch` type. This means I had to add `(action: object) => void` as an acceptable signature for `dispatch`. Unfortunately, I couldn't get that working with a union or extending with an interface, so I ended up using an intersection type. I think that causes tsc to be unable to correctly infer that an async action is being dispatched, so we need to look into the type def for `CalypsoDispatch` to fix that.

client/package.json
```diff
@@ -116,6 +116,7 @@
 		"d3-scale": "^3.2.1",
 		"d3-selection": "^1.4.2",
 		"d3-shape": "^1.3.7",
+		"date-fns": "^2.30.0",
```
tyxla: ~`date-fns` seems to be unused, maybe we can remove it?~

Scratch that, seems like it's a peer dependency of `react-day-picker`.

tyxla: The `react-day-picker` upgrade seems like a bigger change, since it's using `date-fns` and it breaks the `DateRange` component because of lacking locale data. I've reverted the upgrade in https://github.com/Automattic/wp-calypso/pull/77121/commits/54b6d950d05ab0f190e99a4a4a6cca2da6c84fa3 which seems to fix the tests and the component, while still appearing to be working with React 18.

tyxla:
Thanks for the thorough update, Noah! I'll continue in the same spirit.

I made some more progress today, resolved all TS errors and all checks appear to be great, with the exception of a few unit tests.

I did the following:

- Fixed some code style problems (minor markdown lint errors)
- Tuned help center tsconfig to not check JS in order to fix some TS errors without converting code to TS.
- Reorganized some TS modules in help center components - added a new module and deleted one.
- Added more help center tsconfig references so it would properly find all referenced packages
- Introduced a basic `useStore()` react-redux wrapper, to complement the `useSelector` and `useDispatch` ones
- Converted memberships reducer to TS, extracted reducer type
- Expanded the `createReduxStore()` return type to support dynamically registered reducers
- Used `any` instead of `AnyAction` to support simple action creator functions in JS files
- Documented the `any` type and added a TODO to fix it
- Added explicit type casting of token to string type
- Reverted the `react-day-picker` upgrade to fix some tests and deduped packages
- Fixed some of the unit test failures

At this point, I'm stepping away from this one for today - feel free to take over.

As a next step, we should fix the remaining unit test failures, and given everything is green, we should do some very thorough testing of all affected areas. 

On a side note, I find it quite unusual that the `help-center` package imports from the Calypso client. It seems like it should be the other way around, and we're breaking the idea of having the package independent from the monorepo.

client/package.json
noahtallen: Sounds good 👍

noahtallen:
The current unit test failures are due to `react-testing-library` using the new render mode by default. While we can probably fix this by adding `legacyRoot: true` to every single call to render (not a huge fan of that), I'm going to see if we can fix whatever issue is making our code incompatible with the new render mode. The main error is `Cannot read properties of undefined (reading '0') at useSyncExternalStore`

noahtallen:
What I got done today:
- In 001f881, I refactored a test suite to use better RTL practices. It was originally mocking useState, which was causing issues with useSyncExternalStore. (Note that RTL is not using the legacy renderer, but the new one.) This took more time than I was hoping since I wanted the tests to also follow RTL best practices.
- In 59c1995, I added a hopeful fix for issues with the emotion `css` type on components. One new PR on trunk wasn't working in this branch.
- In e834e7c, I switched the unit tests to wait until everything has completed -- currently they exit when one of the tasks has failed, which means we can't see the results of the unit tests.
- In e71d09f, fixed an issue in the media library unit test due to the maximum update depth being exceeded. The root cause was that we were passing our own little store implementation to `connect`, which didn't have any provider wrapped around it. I switched to use `renderWithProvider` instead, and this cleared things up. (After fixing a different issue related to redux-thunk needing our mocked action to receive a function returning an object instead of undefined.)
- Assorted fixes for new commits on trunk which are adding/changing state connections.

Remaining issues I've noted:
- An issue with higher order components and emotion types. Unfortunately, emotion types aren't showing up in my IDE either. Adding the tsconfig JSX setting was needed to get the `css` prop to work, but is messing up types in a couple places.
- The `manage-purchase/test/change-payment-method.tsx` suite is failing to execute for some reason

tyxla (commented):
I've just fixed the remaining unit test failures and merged the latest trunk, and we're all green now 🎉 

For the next week, I'm planning to do some thorough testing and a deep dive into the major package update breaking changes.

noahtallen (commented):
I did a fair amount of testing today, including:

- Login
- Site creation
- Plan purchasing
- Various admin actions like deleting and editing posts
- Devdocs
- qrcode in a handful of places (This is one of the major upgrades)

I also tested the new builds for these apps:
- notifications
- editing toolkit plugin
- wpcom block editor

Didn't detect any issues so far!

noahtallen:
Also, storybook was broken (though we weren't getting the failure), so I made some more updates in 2ad3045 to the config, and it looks like Storybook 7 is successfully working.

In 4a0ddb2 and c816abc, I refactored away from `import url from 'url'` to using 'new URL'. There was a webpack failure which could be related to using a node-style URL, and there were a handful of places we were importing it directly.

I also fixed more merge conflicts with trunk.

tyxla:
> Didn't detect any issues so far!

This is excellent news, @noahtallen! I won't get to this today due to other commitments but I will try to devote some time tomorrow. Hopefully, everything will be good and we can proceed to ship it this week!

tyxla:
Did some research for any breaking changes and have the following conclusions:

- "react-popper": "^2.2.5" => "^2.3.0",
	- diff: https://github.com/floating-ui/react-popper/compare/v2.2.5...v2.3.0
	- No major API changes, basically React 18 support
- "react-redux": "^7.2.6", => "^8.0.5",
	- diff: https://github.com/reduxjs/react-redux/compare/v7.2.6...v8.0.5
	- a few major API changes, nothing utilized in Calypso
- "redux": "^4.1.2" => "^4.2.1",
	- diff: https://github.com/reduxjs/redux/compare/v4.1.2...v4.2.1
	- `createStore` is now deprecated, in favor of @reduxjs/toolkit (see #4336)
		- no need to change it immediately, could be its own separate migration
- "redux-thunk": "^2.3.0" => "^2.4.2",
	- diff: https://github.com/reduxjs/redux-thunk/compare/v2.3.0...v2.4.2
	- no major API changes
	- optionally: we could use `import 'redux-thunk/extend-redux'` to expand the dispatch type to support thunks.
- "react-query": "^3.32.1", => "@tanstack/react-query": "^4.29.1",
	- doesn't seem to be used directly
	- can most likely be removed as a dependency
- "@stripe/react-stripe-js": "^1.4.1" => 2.1.0
	- diff: https://github.com/stripe/react-stripe-js/compare/v1.4.1...v2.1.0
	- a few API changes, nothing utilizd
- "html-to-react": "^1.4.5" => "^1.5.1",
	- diff: https://github.com/mikenikles/html-to-react/compare/a10462807630c61396d5a0d583f80b2a22a9695c...268e5d50944b0c804277deb160da5c78172d1c96
	- nothing affecting
- "qrcode.react": "^1.0.1" => "^3.1.0"
	- diff: https://github.com/zpao/qrcode.react/compare/v1.0.1...v3.1.0
	- default export deprecated
		- library claims that they deprecated it in https://github.com/zpao/qrcode.react/releases/tag/v3.0.0
		- but it seems it's still supported: https://github.com/zpao/qrcode.react/commit/59c4857aba23a6d9d7e35297a9423601c342a6e5
		- we're using the default export: https://github.com/Automattic/wp-calypso/blob/6aa43c308bc6277e683aec021413a11569ce6a73/packages/components/src/app-promo-card/index.tsx#L4
		- worth considering updating, although it seems to still be working (at the bottom of https://wordpress.com/stats/day/SITENAME)
- "react-element-to-jsx-string": "^14.3.2", => "^15.0.0",
	- diff: https://github.com/algolia/react-element-to-jsx-string/compare/75fcbf8...master
	- fixes only, no API changes
- "react-live": "^2.3.0", => "^4.1.1",
	- diff: https://github.com/FormidableLabs/react-live/compare/v2.3.0...react-live%404.1.1
	- claims to have large internal changes
	- tested the devdocs playground, works well.
- "react-modal": "^3.14.3", => "^3.16.1",
	- diff: https://github.com/reactjs/react-modal/compare/fd193f2a66b3ad54b4cecfe1c2d3bb5d1475ae6b...master
	- no big changes
- "reakit": "^1.2.3" => "^1.3.11",
	- doesn't seem to be used directly at all
	- we could remove the dependency
- "redux-dynamic-middlewares": "^2.0.0" => "^2.2.0",
	- no major changes
- "use-subscription": "^1.5.1" =>  "^1.8.0"
	- updated to use `useSyncExternalStore` internally
	- we should be able to use `use-sync-external-store` eventually

TL;DR: we're good 😉

tyxla (approved):
On top of checking the library upgrading for breaking changes, I've also done some thorough testing of various sections (login, signup, site creation - a few free and paid flows) and a bunch of logged-in sections (me, various site settings, stats, reader, plans, checkout) and everything appears to be working great 🎉 

I think we should be good to merge this one, and p2 about it to get all dotcom teams to test and keep an eye on the features they are developing. 

Great work @noahtallen 🙌 ❤️

noahtallen:
Very happy to hear! Your research matches mine as well. A few additions:

- I refactored several qrcode instances to the named import; must have missed that one!
- Reakit is a peer dependency of something else right now, but we’ll at least be able to remove reakit-utils in #77046 
- Otherwise very minor changes beyond the redux upgrade!

PR #77131, opened 2023-05-19
Hosting Onboarding: hosting flow only contains signup step, which includes an "I am a dev" checkbox

p-jackson:
Related to https://github.com/Automattic/dotcom-forge/issues/2364

## Proposed Changes

- Removes all steps from `/start/hosting` except signup
- Adds "I am a dev" checkbox to the passwordless signup (only for this flow)
- "Continue as user" version of signup works, even when it's the last step in the flow.

CleanShot 2023-05-22 at 15 31 35@2x

We should use the state of the "is dev" checkbox for social signup too. But it might be better to leave that for a follow up PR.

## Testing Instructions

* Test locally (where `hosting-onboarding-i2` flag is enabled)
   * Test `/start/hosting` when logged out
   * Test `/start/hosting` when logged in (should see the "Continue as user" version of signup)
   * Test the "Have an account? Log in" link
      * I'm not 100% sure what behaviour we want here, but it takes me to the flow destination (i.e. `/sites`) even when that's not my user's preference, so I think that's correct.
   * Confirm the flow only has one step
   * Confirm the user lands on the `/sites` page
   * Confirm the `/start/onboarding` signup behaviour hasn't changed
* Test with `calypso.live` (where `hosting-onboarding-i2` flag is disabled)
   * Test `/start/hosting` should not have changed compared to production
   * Confirm the user lands on the `/sites` page with the `?new-site=` query param
* Apply D111469-code to sandbox and test new users get the correct "is dev" flag

## Pre-merge Checklist

### DIFF
--- a/client/blocks/signup-form/index.jsx
+++ b/client/blocks/signup-form/index.jsx
@@ -1,6 +1,7 @@
 import config from '@automattic/calypso-config';
 import { Button, FormInputValidation } from '@automattic/components';
 import { localizeUrl } from '@automattic/i18n-utils';
+import { CheckboxControl } from '@wordpress/components';
 import classNames from 'classnames';
 import debugModule from 'debug';
 import { localize } from 'i18n-calypso';
@@ -40,6 +41,7 @@ import TextControl from 'calypso/components/text-control';
 import wooDnaConfig from 'calypso/jetpack-connect/woo-dna-config';
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import formState from 'calypso/lib/form-state';
+import { preventWidows } from 'calypso/lib/formatting';
 import { getLocaleSlug } from 'calypso/lib/i18n-utils';
 import {
 	isCrowdsignalOAuth2Client,
@@ -95,6 +97,7 @@ class SignupForm extends Component {
 		handleLogin: PropTypes.func,
 		handleSocialResponse: PropTypes.func,
 		isPasswordless: PropTypes.bool,
+		showIsDevAccountCheckbox: PropTypes.bool,
 		isSocialSignupEnabled: PropTypes.bool,
 		locale: PropTypes.string,
 		positionInFlow: PropTypes.number,
@@ -119,6 +122,7 @@ class SignupForm extends Component {
 		displayUsernameInput: true,
 		flowName: '',
 		isPasswordless: false,
+		showIsDevAccountCheckbox: false,
 		isSocialSignupEnabled: false,
 		horizontal: false,
 		shouldDisplayUserExistsError: false,
@@ -150,6 +154,7 @@ class SignupForm extends Component {
 
 		this.state = {
 			submitting: false,
+			isDevAccount: false,
 			isFieldDirty: {
 				email: false,
 				username: false,
@@ -924,6 +929,25 @@ class SignupForm extends Component {
 		return false;
 	}
 
+	isDevAccountCheckbox() {
+		if ( ! this.props.showIsDevAccountCheckbox ) {
+			return null;
+		}
+
+		const { translate } = this.props;
+		return (
+			<CheckboxControl
+				label={ preventWidows(
+					translate(
+						"I'm a developer. Boost my WordPress.com experience and give me early access to developer features"
+					)
+				) }
+				checked={ this.state.isDevAccount }
+				onChange={ ( isDevAccount ) => this.setState( { isDevAccount } ) }
+			/>
+		);
+	}
+
 	emailDisableExplanation() {
 		if ( this.props.disableEmailInput && this.props.disableEmailExplanation ) {
 			return (
@@ -1152,6 +1176,7 @@ class SignupForm extends Component {
 					} ) }
 				>
 					{ this.getNotice() }
+					{ this.isDevAccountCheckbox() }
 					<PasswordlessSignupForm
 						step={ this.props.step }
 						stepName={ this.props.stepName }
@@ -1162,6 +1187,7 @@ class SignupForm extends Component {
 						disabled={ this.props.disabled }
 						disableSubmitButton={ this.props.disableSubmitButton }
 						queryArgs={ this.props.queryArgs }
+						isDevAccount={ this.state.isDevAccount }
 					/>
 
 					{ showSeparator && (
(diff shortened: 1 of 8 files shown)


mpkelly (approved):
Successfully followed the test plan. The code looks fine.

javierarce:
I mentioned this on Slack, but I'd like to propose this alternative with a checkbox and a shorter copy: "I am a developer. Boost my WordPress.com use and give me early access to developer features."

image

zaguiini:
@javierarce it feels a little ambiguous right now because of the social login. Will people logging through Google also be registered as developers?

javierarce:
> @javierarce it feels a little ambiguous right now because of the social login. Will people logging through Google also be registered as developers?

I was just assuming we had decided only to mark developers if they were using the left form. If that hasn't changed, I think we could use this alternative

image

cc: @saygunnyc

client/signup/config/flows.js
```diff
@@ -188,7 +188,11 @@ function getDIFMSiteContentCollectionDestination( { siteSlug } ) {
 }
 
 function getSitesDestination( { siteSlug } ) {
-	return addQueryArgs( { 'new-site': siteSlug }, '/sites' );
+	if ( siteSlug ) {
+		return addQueryArgs( { 'new-site': siteSlug }, '/sites' );
+	}
+
+	return '/sites';
```
zaguiini: Is this change necessary? I thought that if the value is undefined, the query parameter is omitted 🤔 

Oh, there are two `addQueryArg` functions in the codebase. One behaves as I described; the other does not.

vykes-mac:
> I mentioned this on Slack, but I'd like to propose this alternative with a checkbox and a shorter copy: "I am a developer. Boost my WordPress.com use and give me early access to developer features."

I second @javierarce comment here. checkbox works better in the use case than a toggle

vykes-mac:
> I'm not 100% sure what behaviour we want here, but it takes me to the flow destination (i.e. /sites) even when that's not my user's preference, so I think that's correct.

This seems like the sensible path, it's still coming from hosting. I suppose it's similar to creating the account from hosting without checking the "I am a developer" checkbox.

client/signup/config/flows.js
p-jackson: Quite right

https://github.com/Automattic/wp-calypso/blob/93a6d449e4dee3bbea8dc1fbbc0f3953927f8ea8/client/lib/url/add-query-args.ts#L28

It's not needed.

The reason I assumed it _wouldn't_ strip `null` or `undefined` values was because how else would you add a query param that was value-less. I guess that's just not possible with this `addQueryArgs` function 🤷

p-jackson:
Thanks for the feedback, currently implementing the checkbox design and moving above the email and social options.

zaguiini:
Have we considered checking the box by default? We're targeting developers in this LOHP, so it makes sense to have this on as a sensible default.

A few visual comments...

It looks like the checkbox and the text are not center-aligned:

image

What is it with the outline? What's the reason behind the border?

image

javierarce:
> What is it with the outline? What's the reason behind the border?

That comes from [this component](https://www.figma.com/file/eqlepL7MHifK8Bgr1OWqLr/Dotcom-Design-Library?type=design&node-id=1584-1929&t=902Au5m47saOfyuw-4) we are using.

> Have we considered checking the box by default? We're targeting developers in this LOHP, so it makes sense to have this on as a sensible default.

In general, it makes sense to default to the most probable option… but I think adding some friction could reduce false positives from non-developers that forget to uncheck the box.

The problem with that checkbox is that we are giving incentives to every type of user… who wouldn't want a better experience and something for free?

zaguiini:
>non-developers that forget to uncheck the box.

Will it be that bad if they don't? They can always disable it by themselves on the settings page. I think that non-ticking it by default would be worse than a few false positives 🤔

p-jackson:
> Have we considered checking the box by default?

I'm lightly in favour of having it unchecked by default. I think the checkbox copy is already so enticing that we're more likely to get false positives than we are to get false negatives.

I'm going to merge this PR with it defaulting to unchecked, but please don't take this as the last word! I just want to get the code changes merged and switching the default is a very quick change to make.

> That comes from this component we are using.

Thanks for sharing the figma @javierarce! Afaict we haven't implemented that checkbox component anywhere else yet so I'm doing it for the first time here. I have some feedback on this double border when the control is checked:

CleanShot 2023-05-24 at 14 58 00@2x

It seems like it'll be easily confused with the focus rectangle. When it has focus it'll have yet _another_ blue border around it. Although maybe for this style of control the focus rectangle should be around the entire rounded rectangle?

p-jackson:
I'm merging this as is (even though checkbox styling could still be improved a bit, and "is dev" doesn't work for social logins yet) just so we can get the functionality on `trunk`. I'll tasks for the remaining tasks.

Since it's behind a feature flag it shouldn't matter that I'm merging early.

However heads up that there is a slight change to production behaviour. In order to get the checkbox to line up nicely with the signup form elements I had to rejig how the layout is done, and that resulted in me being able to make the error messages line up with the signup form elements nicely too. Hopefully a picture  will explain it:

CleanShot 2023-05-24 at 15 13 03@2x

Before this PR the error notice stuck out a bit wider than the signup form. So it looks better now IMO, but just wanted to be explicit that this PR does change production behaviour too.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8007545

javierarce:
> Thanks for sharing the figma @javierarce! Afaict we haven't implemented that checkbox component anywhere else yet so I'm doing it for the first time here. I have some feedback on this double border when the control is checked:

If it's helpful, we recently implemented something pretty similar [here](https://github.com/Automattic/wp-calypso/pull/76854) (with a radio button instead of a checkbox, but the behaviour and style should be similar).

javierarce:
Oh, actually that component exists in https://wordpress.com/setup/site-setup/goals

p-jackson:
So it does! Should have remembered that since I see that screen so often 🙈

https://github.com/Automattic/dotcom-forge/issues/2530

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8007545).

PR #77151, opened 2023-05-19
Remove the experimental apps/inline-help app

jsnajdr:
Removes the experimental `apps/inline-help` app which is a proof of concept I worked on in January 2022 and is abandoned. The production Help Center is being developed in the ETK.

I see it often wastes people's time on various migrations. Currently it is @noahtallen with #77121, before that it was @flootr with React Query upgrade, and then, looking at `git log apps/inline-help`, there were ESLint fixes, Enzyme upgrades, Calypso Apps updates...

**How to test:**
Verify that I didn't break any build or deployment process by removing the app.


tyxla (approved):
Thanks @jsnajdr 👍 

It's worth mentioning that we should consider removing the "Inline Help" build configuration from TeamCity before or after this PR is merged.

jsnajdr:
Removing the TeamCity configuration in c0e5d057d0. I knew that something like this will need to be done, but wasn't sure what exactly.

jsnajdr:
The "Inline Help (WPCom Plugins)" TeamCity job is still running and failing even after I removed it from the `.kt` file. I guess I also need to remove it somewhere in the TeamCity UI?

tyxla:
> The "Inline Help (WPCom Plugins)" TeamCity job is still running and failing even after I removed it from the `.kt` file. I guess I also need to remove it somewhere in the TeamCity UI?

I think it will get removed as soon as we land it in `trunk`.

klimeryk:
Awesome, thanks for removing this - I was actually looking into removing it as part of sunsetting Happy Chat 🙇 (pdm0RZ-BU-p2)

jsnajdr:
> I think it will get removed as soon as we land it in `trunk`.

Yes, that's exactly what happened :+1:

tyxla:
Nice, thanks for confirming 🙌

noahtallen:
Should we also remove `apps/happychat`?

tyxla:
Yep. It's part of that same happychat sunsetting project FWIW - pdm0RZ-BU-p2 - as @klimeryk mentioned above.

noahtallen:
Cool!

klimeryk:
`apps/happychat` (and _much_ more!) being removed in https://github.com/Automattic/wp-calypso/pull/78433 - would love extra eyes and reviews 🙏

tyxla:
Awesome, just reviewed #78433. Great cleanup!

Issue #76933, opened 2023-05-15
Site Assembler: Get rid of the headstart process


arthur791004:
## Description

Referring to pbxlJb-3Ph-p2, we have to get rid of the headstart process so that we're able to use Site Assembler to customize the home template, and header/footer template parts on the AT sites.

## Acceptance Criteria

When a user finishes the PA step on the AT site, 
- [x] The home page should be the same as the preview on the Pattern Assembler screen.
- [x] The header should be the same as the selected header on the Pattern Assembler screen.
- [x] The footer should be the same as the selected footer on the Pattern Assembler screen.
- [x] The active theme should be BC3
- [x] The “Homepage displays” setting should be “default”

## Dev Tasks

- [x] https://github.com/Automattic/wp-calypso/pull/77152
- [x] D111641-code
- [x] https://github.com/Automattic/wp-calypso/pull/77431

## Related

- https://github.com/Automattic/dotcom-forge/issues/2398

arthur791004:
Closing it via https://github.com/Automattic/wp-calypso/pull/77152 and D111641-code

PR #77152, opened 2023-05-19
Site Assembler: Switch to use the site-assembler endpoint

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/76933

## Proposed Changes

* Replace the `applyThemeWithPatterns` action with the new one called `assembleSite` to get rid of the headstart process. And it will generate the template/template-parts based on the selected patterns on the PA screen
* Support the `isNewSite` query parameter to decide whether to reset the content of your site
* Use the `activateOrInstallThenActivate` to active or install the theme on your site to support the AT site

https://github.com/Automattic/wp-calypso/assets/13596067/a54b47d2-8357-47fd-8cbe-cbd5402bbf46

## Testing Instructions

Apply D111641-code to your sandbox before testing

**Logged-in Theme Showcase**
  * Head to `/themes/<your_site>`. Your site can be either a simple site or AT site.
  * Scroll down to select the BCPA CTA, and ensure you land on the PA screen
    image
  * On the PA screen
    * Select a header
    * Select patterns on your homepage
    * Select a footer
    * Select colors
    * Select fonts
    * Continue
  * When you land on the site editor, ensure your homepage is built with the selected patterns and styles
  * Head to your site and ensure it looks good!
  * Head to `/pages/<your_site>`, and ensure your pages are still there
  * Head to `/posts/<your_site>`, and ensure your posts are still there

**Logged-out Theme Showcase**
  * Head to `/themes`
  * Scroll down to select the BCPA CTA, and ensure you land on the PA screen as above
  * Follow the guidance to create a new site
  * When your site is created, ensure you land on the PA screen
  * On the PA screen, follow the above instructions to select patterns and styles
  * When you land on the site editor, ensure your homepage is built with the selected patterns and styles
  * Head to your site and ensure it looks good!
  * Head to `/pages/<your_site>`, and ensure you don't have any existing page
  * Head to `/posts/<your_site>`, and ensure your have 4 initial posts

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -1,7 +1,7 @@
 import { isEnabled } from '@automattic/calypso-config';
 import { useSyncGlobalStylesUserConfig } from '@automattic/global-styles';
 import { useLocale } from '@automattic/i18n-utils';
-import { StepContainer, WITH_THEME_ASSEMBLER_FLOW } from '@automattic/onboarding';
+import { StepContainer, isSiteAssemblerFlow, isSiteSetupFlow } from '@automattic/onboarding';
 import {
 	__experimentalNavigatorProvider as NavigatorProvider,
 	__experimentalNavigatorScreen as NavigatorScreen,
@@ -81,7 +81,7 @@ const PatternAssembler = ( {
 	const siteId = useSiteIdParam();
 	const siteSlugOrId = siteSlug ? siteSlug : siteId;
 	const locale = useLocale();
-	const isNewSite = !! useQuery().get( 'isNewSite' );
+	const isNewSite = !! useQuery().get( 'isNewSite' ) || isSiteSetupFlow( flow );
 	const isSiteJetpack = useSelector( ( state ) => isJetpackSite( state, site?.ID ) );
 
 	// The categories api triggers the ETK plugin before the PTK api request
@@ -638,7 +638,7 @@ const PatternAssembler = ( {
 		<StepContainer
 			className="pattern-assembler__sidebar-revamp"
 			stepName="pattern-assembler"
-			hideBack={ navigatorPath !== NAVIGATOR_PATHS.MAIN || flow === WITH_THEME_ASSEMBLER_FLOW }
+			hideBack={ navigatorPath !== NAVIGATOR_PATHS.MAIN || isSiteAssemblerFlow( flow ) }
 			goBack={ onBack }
 			goNext={ goNext }
 			isHorizontalLayout={ false }
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -1,3 +1,4 @@
+import { isEnabled } from '@automattic/calypso-config';
 import { useSyncGlobalStylesUserConfig } from '@automattic/global-styles';
 import { useLocale } from '@automattic/i18n-utils';
 import { StepContainer, WITH_THEME_ASSEMBLER_FLOW } from '@automattic/onboarding';
@@ -9,11 +10,13 @@ import {
 import { useDispatch, useSelect } from '@wordpress/data';
 import classnames from 'classnames';
 import { useState, useRef, useMemo } from 'react';
-import { useDispatch as useReduxDispatch } from 'react-redux';
+import { useDispatch as useReduxDispatch, useSelector } from 'react-redux';
 import PremiumGlobalStylesUpgradeModal from 'calypso/components/premium-global-styles-upgrade-modal';
-import { ActiveTheme } from 'calypso/data/themes/use-active-theme-query';
 import { createRecordTracksEvent } from 'calypso/lib/analytics/tracks';
-import { setActiveTheme } from 'calypso/state/themes/actions';
+import { isJetpackSite } from 'calypso/state/sites/selectors';
+import { activateOrInstallThenActivate, setActiveTheme } from 'calypso/state/themes/actions';
+import { getThemeIdFromStylesheet } from 'calypso/state/themes/utils';
+import { useQuery } from '../../../../hooks/use-query';
 import { useSite } from '../../../../hooks/use-site';
 import { useSiteIdParam } from '../../../../hooks/use-site-id-param';
 import { useSiteSlugParam } from '../../../../hooks/use-site-slug-param';
@@ -46,6 +49,7 @@ import type { StepProps } from '../../types';
 import type { OnboardSelect } from '@automattic/data-stores';
 import type { DesignRecipe, Design } from '@automattic/design-picker/src/types';
 import type { GlobalStylesObject } from '@automattic/global-styles';
+import type { ActiveTheme } from 'calypso/data/themes/use-active-theme-query';
 import './style.scss';
 
 const PatternAssembler = ( {
@@ -61,7 +65,7 @@ const PatternAssembler = ( {
 	const [ activePosition, setActivePosition ] = useState( -1 );
 	const [ isPatternPanelListOpen, setIsPatternPanelListOpen ] = useState( false );
 	const { goBack, goNext, submit } = navigation;
-	const { applyThemeWithPatterns } = useDispatch( SITE_STORE );
+	const { applyThemeWithPatterns, assembleSite } = useDispatch( SITE_STORE );
 	const reduxDispatch = useReduxDispatch();
 	const { setPendingAction } = useDispatch( ONBOARD_STORE );
 	const selectedDesign = useSelect(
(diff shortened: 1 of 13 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -367,16 +373,38 @@ const PatternAssembler = ( {
...
+		if ( ! siteSlugOrId || ! site?.ID || ! themeId ) {
 			return;
 		}
 
-		setPendingAction( () =>
-			applyThemeWithPatterns( siteSlugOrId, design, syncedGlobalStylesUserConfig ).then(
-				( theme: ActiveTheme ) => reduxDispatch( setActiveTheme( site?.ID || -1, theme ) )
-			)
-		);
+		if ( isEnabled( 'pattern-assembler/logged-in-showcase' ) ) {
```
miksansegundo, line 383: I think we have to create a new feature flag, something like `pattern-assembler/without-headstart`, rather than using `pattern-assembler/logged-in-showcase`.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -367,16 +373,38 @@ const PatternAssembler = ( {
...
+							headerHtml: header?.html,
+							footerHtml: footer?.html,
+							globalStyles: syncedGlobalStylesUserConfig,
+							shouldResetContent: isNewSite,
+						} )
+					)
+			);
+		} else {
+			setPendingAction( () =>
+				applyThemeWithPatterns( siteSlugOrId, design, syncedGlobalStylesUserConfig ).then(
```
miksansegundo, line 403: Just noting that `applyThemeWithPatterns` is [used by virtual themes](https://github.com/search?q=repo%3AAutomattic%2Fwp-calypso%20applyThemeWithPatterns&type=code) in general and tailored flows (`link-in-bio` and `free`) without passing the param `sourceSiteId` so the default value (blankcanvas3demo.wordpress.com) is used from this [constant](https://github.com/Automattic/wp-calypso/blob/b63c4cc849f4ea7fe78a7c26877a1ef2a9198cb7/packages/data-stores/src/site/constants.ts#L3).

This means that all the posts from `blankcanvas3demo` site are copied in the user's site because `trimContent` is [hardcoded](https://github.com/Automattic/wp-calypso/blob/b63c4cc849f4ea7fe78a7c26877a1ef2a9198cb7/packages/data-stores/src/site/actions.ts#L562) to be `false` because otherwise only three posts will be copied. 

For sites assembled with the theme `blankcanvas3` we copy posts from `blankcanvas3demo` but for other themes maybe we should use their demo sites. The new endpoint `/site-assembler` handles that and we should use it for virtual themes too. cc: @mmtr @taipeicoder

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
arthur791004: I considered that before but I decided to bundle this endpoint with BCPA in the logged-in theme showcase since we can have full tests via CfT to ensure everything works well and ship them together. Or do you prefer to switch to this endpoint first?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -367,16 +373,38 @@ const PatternAssembler = ( {
...
+							activateOrInstallThenActivate( themeId, site?.ID, 'assembler', false, false )
+						)
+					)
+					.then( () =>
+						assembleSite( siteSlugOrId, isSiteJetpack ? themeId : stylesheet, {
+							homeHtml: sections.map( ( pattern ) => pattern.html ).join( '' ),
+							headerHtml: header?.html,
+							footerHtml: footer?.html,
+							globalStyles: syncedGlobalStylesUserConfig,
+							shouldResetContent: isNewSite,
```
okmttdhr, line 397: badge Beginner question again 💡 
Why do we need to set `shouldResetContent: true` when it's a new site?
Does it mean that there is already content even for a new site?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -367,16 +373,38 @@ const PatternAssembler = ( {
...
+		if ( isEnabled( 'pattern-assembler/logged-in-showcase' ) ) {
+			setPendingAction( () =>
+				Promise.resolve()
+					.then( () =>
+						reduxDispatch(
+							activateOrInstallThenActivate( themeId, site?.ID, 'assembler', false, false )
+						)
+					)
+					.then( () =>
+						assembleSite( siteSlugOrId, isSiteJetpack ? themeId : stylesheet, {
```
okmttdhr, line 392: badge
What will happen if `/sites/<>/site-assembler` respond with 500? Could the user see a message or something on UI?

arthur791004: The [processing-step](https://github.com/Automattic/wp-calypso/blob/a46d327d187c79654b6e8abc18cc8be387cd5657/client/landing/stepper/declarative-flow/internals/steps-repository/processing-step/index.tsx#L76) will handle this kind of error, and we can [decide whether to go when submitting the result](https://github.com/Automattic/wp-calypso/blob/a46d327d187c79654b6e8abc18cc8be387cd5657/client/landing/stepper/declarative-flow/site-setup-flow.ts#L284)

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
arthur791004: Yes. There are initial pages/posts when the site is created by default but those content might not be what we want in WordPress.com

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
okmttdhr: That's good to know! Thank you for even telling me where the code is!

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
miksansegundo: I see. Testing all together makes sense. I thought you missed to update the flag name. Thanks for the details.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -77,6 +81,8 @@ const PatternAssembler = ( {
 	const siteId = useSiteIdParam();
 	const siteSlugOrId = siteSlug ? siteSlug : siteId;
 	const locale = useLocale();
+	const isNewSite = !! useQuery().get( 'isNewSite' ) || isSiteSetupFlow( flow );
```
miksansegundo, line 84: This way to check if a site is new uses the entry points to the assembler (`with-theme-assembler` and `site-setup`). We can also use this way to identify new sites and add the site option `wpcom_site_setup=assembler`.

Another way is to use the property `time_posted` from the entry in `wp_signup_log`. We use it in the Blog Report Card to show how old a site is. We can assume that a site is new when it has been created in the past 24h.

What are you thoughts on using `time_posted`? It seems easier to maintain because the flows and assembler entry points will change in the future.

arthur791004: We can get the creation time called `registered` from the `get_blog_details` function on the server site. If we can use the creation time to determine whether a site is newly created within the specific time range (e.g. 24 hours), then I can use that data 🙂

Let's create an issue and do it in the follow-up PR. WDTY?

arthur791004: Oh, or we can simply use `site.options.created_at` to determine whether it's a new site or not

arthur791004: But there might be some edge cases. For example, if the user creates a new site, and then goes through the `with-theme-assembler` flow from the Theme Showcase again within the specific time range, should we treat it as a new site? On the contrary, if the user creates a new site, hangs on the onboarding flow and land on the PA after 24 hours, how could we do 🤔

miksansegundo: Let's create an issue to discuss cons/pros and decide what solution is better and easier to maintain. I think there is not a perfect solution.

miksansegundo (approved):
It's looking good and working fine on my tests. Let's do more testing to make sure we identify the edge cases.

Screenshot 2566-05-26 at 12 42 46

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
miksansegundo: What are your thoughts on this @taipeicoder?

arthur791004: Here is the [isNewSite](https://github.com/Automattic/wp-calypso/blob/97a7f20365454dbb30c4b5fdd98e360a9ab93cb1/client/state/sites/selectors/is-new-site.js#L13) selector in Calypso now.

taipeicoder (approved):
Tested:
- Theme Showcase - Simple site ✅
- Theme Showcase - Atomic site ✅
- Onboarding Design Picker ✅
- LoTS ✅

🎉

okmttdhr (commented):
Thank you and LGTM ✅ 👍

Issue #77093, opened 2023-05-18
Subscription Management: Update site subscriptions count when the user unsubscribes or resubscribes


ivan-ottinger:
### Details

The number of site subscriptions should decrement or increment depending on if the user unsubscribed or resubscribed:

![Markup on 2023-05-18 at 15:24:34](https://github.com/Automattic/wp-calypso/assets/25105483/8d19243f-5ce9-4e20-b4a8-97e5039b5cae)

Details:

- the text should state "subscribers" instead of "followers" (contrary to the screenshot)
- the number should be formatted with commas (as can be seen in the screenshot)
- let's use optimistic updates

Design: inDLaEQV8jJ21O4WXawIsJ-fi-712_30652

PR #77160, opened 2023-05-19
Subscriptions Management: Add optimistic updates to all settings / toggles on the Individual Site Subscription page

### LINKED ISSUE #77174: Subscription Management: Add optimistic updates to all settings / toggles on the Individual Site Subscription page.
### Details

The goal of this task is to add optimistic updates to all settings / toggles on the Individual Site Subscription page:

Markup on 2023-05-19 at 16:46:28

### Checklist

_No response_

### Related

_No response_

ivan-ottinger:
Resolves https://github.com/Automattic/wp-calypso/issues/77174.

## Proposed Changes

Add optimistic update to:

* `NotifyMeOfNewPosts` toggle
* `EmailMeNewPosts` toggle
* `Delivery Frequency` selector
* `EmailMeNewComments` toggle

Markup on 2023-05-22 at 16:39:28

**Please note**

ℹ️ This PR does not address optimistic updates for subscribe / unsubscribe actions. These require a bit different approach - since once you unsubscribe, information like Site title, Site icon and subscriptions count become unavailable - while we still need it for display. This will be addressed this in a separate task (https://github.com/Automattic/wp-calypso/issues/77093).

ℹ️ The `EmailMeNewComments` toggle always returns to the original state because there's a bug in the API which needs to be addressed separately (https://github.com/Automattic/wp-calypso/issues/77173).

### Testing as an external user
1. Check out the PR and build the app.
2. Authenticate as an external user (`subkey` "dance" in the incognito browser window).
3. Make sure your external user is subscribed to some sites already.
4. Navigate to `http://calypso.localhost:3000/subscriptions/site/[blog_id]` where `[blog_id]` needs to be replaced with a specific blog ID your user is subscribed to.
5. When the page loads, you should be able to change the "Email delivery frequency" setting (this can be checked at https://wordpress.com/subscriptions/sites).

### Testing as an logged-in user
1. Check out the PR.
2. Navigate to `client/server/pages/index.js` and temporarily add `return next();` to the line `833`:

Markup on 2023-04-28 at 17:50:44

3. Build the app.
4. Log in to a test WordPress.com account that has some existing Site subscriptions.
5. Navigate to `http://calypso.localhost:3000/subscriptions/site/[blog_id]` where `[blog_id]` needs to be replaced with a specific blog ID your user is subscribed to.
7. When the page loads, you should be able to switch all the toggles (this can be checked at https://wordpress.com/subscriptions/sites).

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/reader/hooks/use-cache-key.ts
+++ b/packages/data-stores/src/reader/hooks/use-cache-key.ts
@@ -3,10 +3,16 @@ import useIsLoggedIn from './use-is-logged-in';
 // change version to invalidate all cache keys
 const version = 'v1';
 
-const useCacheKey = ( key: string[] ) => {
+const useCacheKey = ( key: string[], blog_id?: string | number ) => {
 	const { id, isLoggedIn } = useIsLoggedIn();
 
-	return [ ...key, version, isLoggedIn ? 'logged-in' : 'not-logged-in', id ? id : '' ];
+	return [
+		...key,
+		version,
+		isLoggedIn ? 'logged-in' : 'not-logged-in',
+		id ? id : '',
+		blog_id ?? '',
+	];
 };
 
 export default useCacheKey;
--- a/packages/data-stores/src/reader/hooks/use-cache-key.ts
+++ b/packages/data-stores/src/reader/hooks/use-cache-key.ts
@@ -3,16 +3,10 @@ import useIsLoggedIn from './use-is-logged-in';
 // change version to invalidate all cache keys
 const version = 'v1';
 
-const useCacheKey = ( key: string[], blog_id?: string | number ) => {
+const useCacheKey = ( key: string[] ) => {
 	const { id, isLoggedIn } = useIsLoggedIn();
 
-	return [
-		...key,
-		version,
-		isLoggedIn ? 'logged-in' : 'not-logged-in',
-		id ? id : '',
-		blog_id ?? '',
-	];
+	return [ ...key, version, isLoggedIn ? 'logged-in' : 'not-logged-in', id ? id : '' ];
 };
 
 export default useCacheKey;
--- a/client/landing/subscriptions/components/site-subscription-page/site-subscription-settings.tsx
+++ b/client/landing/subscriptions/components/site-subscription-page/site-subscription-settings.tsx
@@ -22,7 +22,7 @@ const SiteSubscriptionSettings = ( {
 	const { mutate: updateNotifyMeOfNewPosts, isLoading: updatingNotifyMeOfNewPosts } =
 		SubscriptionManager.useSiteNotifyMeOfNewPostsMutation( blogId );
 	const { mutate: updateEmailMeNewPosts, isLoading: updatingEmailMeNewPosts } =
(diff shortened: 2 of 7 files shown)


packages/data-stores/src/reader/hooks/use-cache-key.ts
```diff
@@ -3,10 +3,16 @@ import useIsLoggedIn from './use-is-logged-in';
 // change version to invalidate all cache keys
 const version = 'v1';
 
-const useCacheKey = ( key: string[] ) => {
+const useCacheKey = ( key: string[], blog_id?: string | number ) => {
 	const { id, isLoggedIn } = useIsLoggedIn();
 
-	return [ ...key, version, isLoggedIn ? 'logged-in' : 'not-logged-in', id ? id : '' ];
+	return [
+		...key,
```
yansern: Why`...key` wouldn't work but `blog_id ?? ''` would?

ivan-ottinger: Could you please share a little bit more what you mean?

Do you mean why I am setting the `blog_id` as an extra parameter to the `useCacheKey`?

The reason is that I wanted to avoid checking for `blog_id` value everywhere `useCacheKey` is used - since `blog_id` can be string, but also a number. So each `useCacheKey` would need to have something like:
 
```js
const siteSubscriptionDetailsCacheKey = useCacheKey(
		[ 'read', 'site-subscription-details', ...( blog_id !== undefined ? [ String( blog_id ) ] : [] ) ]
	);
```

Of course, this could be written in a different way.

Edit: Or, maybe we could make sure that `blog_id` is always string. 🤔

ivan-ottinger: I have rewritten the logic in https://github.com/Automattic/wp-calypso/pull/77160/commits/cc01485c9638903bd2252907dc85ae6175a60537.

yansern: > Could you please share a little bit more what you mean?

Yes so because `...key` would essentially spread out the array items, the end result would be the same as adding blog_id to the array separately.

But anyway, I see that you have rewrote the logic, so all is good. :)

JuanLucha:
This is not a blocker to merge this PR because this could be handled when we integrate this screen with the rest of the application, but following your test instructions for external users, I tried to change the delivery frequency of a site I'm not subscribed to, and I got this error:
Screenshot 2023-05-22 at 17 19 15

yansern (approved):
Thanks!

yansern:
> This is not a blocker to merge this PR because this could be handled when we integrate this screen with the rest of the application, but following your test instructions for external users, I tried to change the delivery frequency of a site I'm not subscribed to, and I got this error: Screenshot 2023-05-22 at 17 19 15

Edge case. They shouldn't be able to land there.

ivan-ottinger:
> This is not a blocker to merge this PR because this could be handled when we integrate this screen with the rest of the application, but following your test instructions for external users, I tried to change the delivery frequency of a site I'm not subscribed to, and I got this error:

Thank you for catching and reporting that case, @JuanLucha!

---

Thank you for merging, @yansern!

PR #77024, opened 2023-05-17
Support chat using Zendesk Messaging - attempt 2

klimeryk:
Reverts Automattic/wp-calypso#77016

Second attempt at https://github.com/Automattic/wp-calypso/pull/76777

We're switching from Happy Chat to Zendesk Messaging for live chat support. This PR is big, but it's also the smallest I could get given the scope of the project. Any cleanup of Happy Chat code, old contact forms, etc. will be done as a follow-up. I want to keep this PR as easy to revert as possible if needed.

## Proposed Changes

* Disable Happy Chat feature flag and any availability checks that trigger connections to the websocket.
* Load the Zendesk Messaging script instead.
* Authenticate using already established endpoint (just with `type=zendesk`).
* Check availability using the `/help/messaging/is-available` endpoint.
* Pass the initial chat data using Zendesk user fields.
* Use the support history to figure out if the user has ongoing conversations or not (thankfully, a Messaging conversation is a "ticket"!).
* Additionally, make sure the Zendesk launcher is visible if any unread messages are reported by Zendesk's APIs directly. This is a good safeguard and makes it more dynamic.
* After a refresh, if there's an active Messaging conversation, the Zendesk launcher should automatically appear.
* If the browser is blocking requests to Zendesk, show a notice and link to relevant support doc.
* **New**: support activity is fetched from data stored on the WPCOM user. No external APIs (like Zendesk's) are contacted. That will allow us to provide the best UX, without running into throttling issues.

## Testing Instructions

The live chat support flow should feel the same up to the point where you click the chat CTA (`Chat with us`/`Still chat with us`). At that point, the Help Center should get hidden, while the Zendesk launch and widget should become visible.

Screenshot 2023-05-16 at 19 05 25

Trying to show the Help Center should automatically minimize the Zendesk widget, to avoid clutter and confusing UI. The Zendesk launcher remains visible, so the user can come back to the conversation at any time.

If you have Enhanced Tracking Protection (or similar strict privacy settings and/or ad-blockers), you will see this notice when choosing the Live Chat support option:
Screenshot 2023-05-16 at 19 07 32

Other support options should not show this notice, of course.

Verify that the Zendesk widget follows you on different paths in Calypso. Where Help Center is, the widget should also be there.

If you have an active, ongoing conversation, it should be detected and on refreshes the launcher should appear. If new messages arrive, it will be reflected with a count/notification on the launcher as well.

Screenshot 2023-05-16 at 17 41 35

[...]

PR #77164, opened 2023-05-19
Remove signup/inline-help and related code

klimeryk:
Part of cleaning up Happy Chat: pdm0RZ-BU-p2

Related to #77024

## Proposed Changes

* Remove the unused `signup/inline-help` feature flag and related code.

## Testing Instructions

This does not seem to be enabled, so just general smoke testing and relying on e2e test should hopefully be enough.


alshakero (commented):
We also have `happychat-connection` package that can be removed.

klimeryk:
Right - that's definitely on my radar: pdm0RZ-BU-p2. Just wanted to split it into smaller pieces that are less likely to break things. And if they do, it will be easier to figure out which part 🙇

PR #77165, opened 2023-05-19
Messaging: Improve handling of Zendesk key for auth

klimeryk:
## Proposed Changes

* Switches to sandbox key for development env.
* Ensures we refetch the JWT once the key changes, so that we don't get 403 responses from Zendesk.

## Testing Instructions
Good to start by cleaning your state, so that the previous JWT keys are removed: just add `?flags=force-sympathy` to the URL.

- Open Help Center -> Still need help.
- In the Network tab, you should have a request to `https://public-api.wordpress.com/wpcom/v2/help/authenticate/chat?_envelope=1&type=zendesk&test_mode=true`.
- There should also be a successful POST request to `https://woothemes1654197491support.zendesk.com/sc/sdk/v2/apps/633dcb21690c2c00f4ca2eb8/login`.
- Refresh Calypso, redo the steps. However, now there should be _no_ request to `/wpcom/v2/help/authenticate/chat` (the JWT returned there can and should be cached for a week). But there should still be a successful POST to Zendesk's `login`.

Now, stop Calypso, switch the key back to production one (`cec07bc9-4da6-4dd2-afa7-c7e01ae73584`), restart Calypso. You'll also need to hardcode `test_mode: 'false'` here: https://github.com/Automattic/wp-calypso/blob/bdff6b74ae544e736e33e620f12225a3ffa453e0/packages/help-center/src/hooks/use-messaging-auth.ts#L15.

- Redo the steps. Now there should be again a request to `/wpcom/v2/help/authenticate/chat` (to fetch the correct JWT for production). And there should still be a successful POST to Zendesk's `login`, but on the `wpcomsupport.zendesk.com` host.

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/help-center/src/config.js
@@ -1,8 +1,12 @@
+const isDevelopment = process.env.NODE_ENV === 'development' ? 'development' : 'production';
+
 window.configData = {
-	env_id: 'production',
+	env_id: isDevelopment ? 'development' : 'production',
 	i18n_default_locale_slug: 'en',
 	google_analytics_key: 'UA-10673494-15',
-	zendesk_support_chat_key: 'cec07bc9-4da6-4dd2-afa7-c7e01ae73584',
+	zendesk_support_chat_key: isDevelopment
+		? '715f17a8-4a28-4a7f-8447-0ef8f06c70d7'
+		: 'cec07bc9-4da6-4dd2-afa7-c7e01ae73584',
 	client_slug: 'browser',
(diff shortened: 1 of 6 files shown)


alshakero (commented):
This would mean that in Calypso, people will get dev, in the Editor, they will get prod.

Is that OK?

klimeryk:
> This would mean that in Calypso, people will get dev, in the Editor, they will get prod.

It's an improvement ;) I was mostly interested in fixing the general development case, with ETK being an outlier (in terms of testing support flows). But thanks for the nudge - I've added logic for ETK in https://github.com/Automattic/wp-calypso/pull/77165/commits/f2129f9550bb9dbd8e9838a554ae4f6415338e62 that should help 🙏

PR #77185, opened 2023-05-19
Test deleting reference + file to show warning on install-plugin

lupus2k:
I created this PR for testing purposes. Our goal is to provide a warning to users who remove a file and its corresponding `required_once` statement in the same deployment.

Diff: D111576-code

Don't worry about failed tests, the only needed is `Editing ToolKit (WPCom Plugins)`


lupus2k:
Not needed anymore, diff was merged.

PR #77005, opened 2023-05-17
Integrate logged in tags page with reader sidebar

roo2:
project thread pe7F0s-Qa-p2

Takes the existing /tags page and updates it to include the logged in sidebar as described in task one of the project thread.
Screen Shot 2023-05-17 at 5 17 34 pm

Screen Shot 2023-05-17 at 5 18 04 pm

### Testing instructions

Logged out /tags page should remain unchanged

Test logged in tags page in mobile, tablet and desktop. I've attempted to align the left padding with the individual tag pages.

PR #77192, opened 2023-05-22
Fix bug resizing window with absolute positioned sticky panel

roo2:
Fixes issue raised here https://github.com/Automattic/wp-calypso/pull/77005#issuecomment-1551369419

Basically if you are in a desktop sized window and have the panel in sticky mode, then resize the window to mobile size, the width of the sticky panel was not adjusted and it took up the full width of the screen.

The sticky panel relied on taking the width of the panel from "un-sticky" mode and then not adjusting it again until sticky mode is toggled.

This diff forces the width to be checked on scroll or resize, it also sets the mobile width in css to make sure it is always available.

https://github.com/Automattic/wp-calypso/assets/22446385/5112ee5a-4464-43e2-ad98-82738b173057

### Testing instructions
Go to /tags in a desktop browser
scroll down until "sticky mode" is engaged
resize the window to mobile size.
The panel should be sized correctly.

### DIFF
--- a/client/components/sticky-panel/index.jsx
+++ b/client/components/sticky-panel/index.jsx
@@ -153,7 +153,8 @@ class StickyPanelWithIntersectionObserver extends Component {
 			return this.setState( { isSticky: false } );
 		}
 
-		if ( isSticky !== this.state.isSticky ) {
+		const dimensionUpdates = getDimensionUpdates( this.state._ref.current, this.state );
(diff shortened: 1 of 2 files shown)


davemart-in (approved):
Looks to be fixed! Nice work! No JS errors. Here's what I'm seeing:

https://github.com/Automattic/wp-calypso/assets/5634774/175ccf31-bc04-4edb-b1f7-2b9c1c2ed9f0

PR #77198, opened 2023-05-22
useStoredPaymentMethods: improve types for query key and the wp.req.get call

jsnajdr:
Looking at the issue with disabled queries and `isLoading` state, I notice the `useStoredPaymentMethods` hook doesn't work very well with the `type` and `expired` default values. They should be defaulted in the `useStoredPaymentMethods` hook, not only in the `fetchPaymentMethods` function. Otherwise we can get different query key values for the same state: the `all` type can be represented as both `"all"` and `undefined` in the key, and `false` expired value as both `false` and `undefined`.

Also I'm improving the `wp.req.get` call by putting the query params to the `query` parameter of `wp.req.get`. If will take care of properly serializing the values and constructing the query string.

@sirbrillig does this look good?


sirbrillig (approved):
Thanks! This makes a lot of sense. I didn't consider the effects on the key of undefined values.

I tested various places that use stored payment methods and they appear to work as expected.

PR #77210, opened 2023-05-22
i18n: Add error handler for the require chunk translations handler

yuliyan:
Related to p1684544878901539-slack-C04U5A26MJB

Prior to adding the error handler, if a translation chunk request fails, it would cause the entire Webpack dynamic import to fail which will likely result in crashing the entire app. Adding the error handler for the translation chunk requests would prevent the dynamic import failures and only cause some missing translation, but the rest of the app should remain working as normal.

## Proposed Changes

* Add error handling for the require chunk translations handler.

## Testing Instructions

* Code review.
* Boot Calypso locally with `BUILD_TRANSLATION_CHUNKS=true ENABLE_FEATURES=wpcom-user-bootstrap,use-translation-chunks yarn start`.
* Change UI to a Mag-16 language.
* Delete the language files for the selected language. E.g., for Spanish: `find public/languages/es-* -not -name "*manifest*" -delete`
* Confirm translation chunks fetch requests fail, but the page still loads correctly.

## Pre-merge Checklist

### DIFF
--- a/client/lib/i18n-utils/switch-locale.js
+++ b/client/lib/i18n-utils/switch-locale.js
@@ -218,6 +218,7 @@ function getIsTranslationChunkPreloaded( chunkId, localeSlug ) {
 	if ( typeof window !== 'undefined' ) {
 		return (
 			window.i18nLanguageManifest?.locale?.[ '' ]?.localeSlug === localeSlug &&
+			window.i18nTranslationChunks &&
 			chunkId in window.i18nTranslationChunks
 		);
 	}
@@ -286,12 +287,15 @@ function addRequireChunkTranslationsHandler( localeSlug = i18n.getLocaleSlug(), 
 			return;
 		}
 
-		const translationChunkPromise = getTranslationChunkFile( chunkId, localeSlug ).then(
-			( translations ) => {
+		const translationChunkPromise = getTranslationChunkFile( chunkId, localeSlug )
+			.then( ( translations ) => {
 				addTranslations( translations, userTranslations );
 				loadedTranslationChunks[ chunkId ] = true;
-			}
-		);
+			} )
+			.catch( ( error ) => {
+				debug( `Encountered an error loading translation chunk ${ chunkId }.` );
+				debug( error );
+			} );
 
 		promises.push( translationChunkPromise );
 	};
--- a/client/lib/i18n-utils/switch-locale.js
+++ b/client/lib/i18n-utils/switch-locale.js
@@ -4,6 +4,7 @@ import { isDefaultLocale, getLanguage } from '@automattic/i18n-utils';
 import debugFactory from 'debug';
 import i18n from 'i18n-calypso';
 import { forEach, throttle } from 'lodash';
+import { captureException } from 'calypso/lib/sentry';
 
 const debug = debugFactory( 'calypso:i18n' );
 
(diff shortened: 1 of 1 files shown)


jsnajdr:
While testing this, I found one bug that might be causing some of the Sentry errors. Load Calypso with Translation Chunks and the Quick Language Switcher:
```
ENABLE_FEATURES=use-translation-chunks,quick-language-switcher yarn start
```
Now switch language using the masterbar switcher, say from `de` to `es`. Then look at `window.__requireChunkCallback__`. It contains *two* callbacks, one from the previous language, one from the new one. When switching language without reloading, as the quick switcher does, nobody calls `removeRequireChunkTranslationsHandler` with the old callback.

Both handlers load chunks, each for a different language, and they try to add it to common global `i18n` object with `addTranslations()`.

client/lib/i18n-utils/switch-locale.js
```diff
@@ -286,12 +287,15 @@ function addRequireChunkTranslationsHandler( localeSlug = i18n.getLocaleSlug(),
...
+		const translationChunkPromise = getTranslationChunkFile( chunkId, localeSlug )
+			.then( ( translations ) => {
 				addTranslations( translations, userTranslations );
 				loadedTranslationChunks[ chunkId ] = true;
-			}
-		);
+			} )
+			.catch( ( error ) => {
+				debug( `Encountered an error loading translation chunk ${ chunkId }.` );
+				debug( error );
```
jsnajdr, line 302: It would be great if we could report to Sentry that we failed to load a translation chunk. Including info on the chunk name and locale. That would be an improvement over the current `TypeError: Failed to fetch`. And would help us figure out what is causing these failures.

Calypso client already integrates the `@sentry/browser` library. Maybe @fullofcaffeine or @noahtallen know how to report custom errors there.

jsnajdr (approved):
Thanks, this works as expected. Before this PR, a translation chunk load failure would fail the entire dynamic `import( 'something' )` call, because the handler calls `promises.push` with a promise that rejects. And webpack then does `Promise.all( promises )`.

Adding the `.catch` handler swallows the error, the dynamic chunk is still loaded, but without the translations -- they are optional.

yuliyan:
> While testing this, I found one bug that might be causing some of the Sentry errors. Load Calypso with Translation Chunks and the Quick Language Switcher:
> 
> ```
> ENABLE_FEATURES=use-translation-chunks,quick-language-switcher yarn start
> ```
> 
> Now switch language using the masterbar switcher, say from `de` to `es`. Then look at `window.__requireChunkCallback__`. It contains _two_ callbacks, one from the previous language, one from the new one. When switching language without reloading, as the quick switcher does, nobody calls `removeRequireChunkTranslationsHandler` with the old callback.
> 
> Both handlers load chunks, each for a different language, and they try to add it to common global `i18n` object with `addTranslations()`.

Thank you for reporting this, @jsnajdr! I was able to reproduce it and will address it in a separate PR.

client/lib/i18n-utils/switch-locale.js
noahtallen: Yes! You can use `import { captureException } from 'calypso/lib/sentry';`, and then call `captureException(error)`. (You could create a `new Error()` to include extra information.)

It's important to import this function from our internal lib instead of from the sentry package to avoid loading extra code when it's not needed. Our lib makes sure sentry is only async loaded for a small percent of requests, but lets you use these stub functions anywhere, even if they are sometimes no-op when sentry isn't loaded.

yuliyan: Thank you for the suggestion! I've updated the PR to report the error to Sentry.

However, I was unable to test it locally (I believe Sentry is not enabled in local environment). @noahtallen, do you happen to know how to test if the error reporting works as expected?

noahtallen: I don't think it would be blocking to test the error flow, but I think you could test it locally by commenting out these lines of code:

https://github.com/Automattic/wp-calypso/blob/d6c65e414a1041b53f2efc726d4bf664e0484718/client/lib/sentry/index.ts#L135-L148

And then somehow triggering the exception. I think it will show up in Sentry at that point

yuliyan: Thanks - I managed to to test it on calypso.live and it looks like errors are now properly reported to Sentry. I'll proceed with merging the PR.

PR #77225, opened 2023-05-22
[GB][E2E] Rotate user password for the AT E2E edge site

fullofcaffeine:
## Proposed Changes

After I copied it to login to smoke test the new GB version in the AT GB edge testing site, it was left in the clipboard, and after a failed copy, I accidentally pasted the password at my browser's search bar and pressed enter, so it was sent as the query param for a Google search. The risk is virtually inexistent as the connection is secure, and there was no contextual information sent along, but better safe than sorry, so I'm rotating it just in case.

## Testing Instructions

1. Checkout `trunk`
2. Follow the instructions in this fieldguide: PCYsg-vnR-p2 to decrypt the config file
3. Move it to `decrypted-secrets.trunk.json`
4. Checkout this branch and decrypt the config file again
7. Diff the trunk and the new file: `diff decrypted-secrets.trunk.json decrypted-secrets.json` and make sure only the expected changes are there.
8. (Optional) Try running some E2E test to make sure the JSON is valid and is properly loaded


fullofcaffeine:
I already tested this locally and made sure the diff looked good, so I'll go ahead and merge this once the tests pass. Once I merge, I'll update the password in the AT edge test site to match.

PR #77265, opened 2023-05-23
removing Coupon code for Akismet checkout

mikestottuk:
Removes the Coupon Code area from Akismet checkout since there are no coupons for that product.

To test:-

- Spin up Calypso locally
- Visit http://calypso.localhost:3000/checkout/akismet/ak_plus_yearly_1?checkoutBackUrl=https://akismet.com/pricing/

The coupon code should not show in this branch

- Visit something like http://calypso.localhost:3000/checkout/{site}#step2

The coupon code should show (since won't be Akismet checkout)

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout-order-review.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout-order-review.tsx
@@ -179,13 +179,15 @@ export default function WPCheckoutOrderReview( {
 				/>
 			</WPOrderReviewSection>
 
-			<CouponFieldArea
-				isCouponFieldVisible={ isCouponFieldVisible }
-				setCouponFieldVisible={ setCouponFieldVisible }
-				isPurchaseFree={ isPurchaseFree }
-				couponStatus={ couponStatus }
-				couponFieldStateProps={ couponFieldStateProps }
-			/>
+			{ ! isAkismetCheckout() && (
+				<CouponFieldArea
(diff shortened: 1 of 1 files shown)


client/my-sites/checkout/composite-checkout/components/wp-checkout-order-review.tsx
```diff
@@ -227,6 +228,10 @@ function CouponFieldArea( {
 		);
 	}
 
+	if ( isAkismetCheckout() ) {
```
IanRamosC: I'd recommend moving this check to the render of `WPCheckoutOrderReview`. That way, we avoid calling this function/component entirely

IanRamosC (commented):
The code tests well, but I left a comment about a possible improvement. Also, do we want to A/B test this?

mikestottuk:
> Also, do we want to A/B test this?

We'll just do a before / after test and if it looks positive for Akismet build in A/B to test for WP.com

client/my-sites/checkout/composite-checkout/components/wp-checkout-order-review.tsx
mikestottuk: If it's moved there, it removes the whole top area where you can interact with what's in the order.

Screenshot 2023-05-31 at 13 14 29

IanRamosC (approved):
LGTM!

sirbrillig (approved):
This should be fine. 👍

PR #77300, opened 2023-05-23
DISCOVER - Add suggested follows modal (with new endpoint)

eoigal:
This PR will add a modal that will show a list of suggested follows based on the site the user has just followed.

follow-1

We will only show this modal on Reader site page. We may also add it to Reader search page.

Project Ref: pe7F0s-O2-p2

### PR highlights

The PR requires the patch at code-D111621. The patch adds a new related sites endpoint. 

The endpoint will look up a top viewed post for this blog and use it with `Reader_Post_Related` to get a list of related post IDs and site IDs. In the case where no top view post is found in stats, it will fallback to the `post_id` passed in the request. This `post_id` is the latest post on this site.

The endpoint will then filter out an undesirable sites and return a list of related sites.

This PR uses `useQuery` to request the list of related sites.

Another thing of note, the Follow button style is left unchanged for now. The designs have a button style but this would require a change to the followbutton component which is something I think we can come back to.

### Testing
* Apply patch code-D111621 on sandbox
* Go to a reader site page like http://calypso.localhost:3000/read/feeds/35154032
* Click the `Follow` button and you should see a dialog like the image below - confirm it works and looks as expected

Screenshot 2023-05-19 at 16 15 16


davemart-in (approved):
Everything checks out fine. There might be a slight second more delay before the modal pops up, but I think it's fine. Here's what I'm seeing:

https://github.com/Automattic/wp-calypso/assets/5634774/a97bc1db-b4db-4329-abf0-9f1b6c0ada72

eoigal:
The modal was a little laggy so have sped it up by not waiting for posts to load before making API call to get suggested follows. To allow for this, we no longer wait for the posts to load on reader. This was necessary to get the latest post ID. Now we just look that up on the endpoint if necessary.

I've also updated code-D111621 to be a little quicker too.

The modal now also includes a placeholder while the suggested follows call is being processed. You should only see this if you click follow quickly after the site page loads.

davemart-in (approved):
Very zippy this time! LGTM. 🚢 it.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8009923

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8009923).

Issue #77048, opened 2023-05-18
Rewrite @automattic/global-styles to align with Calypso's React 18 upgrade


taipeicoder:
Related issue: #77046

## Description

There is an ongoing effort to upgrade Calypso to React 18. WordPress packages will also be updated to the first version that supports React 18. As a result, the package `@automatic/global-styles` will be heavily affected since it has a lot of hard dependencies with WordPress packages, especially `@wordpress/edit-site`.

Please refer to this discussion for more context: p1684368118561169-slack-CRWCHQGUB

## Acceptance Criteria
- [x] Rewrite `@automattic/global-styles` using the WordPress packages updated in #77046.
- [x] (Optional) Avoid importing from the `build_module` path. Please make a separate issue if it's going to be a follow-up task.

arthur791004:
Here are the functions we need to rewrite ([commit](https://github.com/Automattic/wp-calypso/pull/77046/commits/19f4a493055cb229eec5b3975ee81ed1531912ab)):

* GlobalStylesContext
* mergeBaseAndUserConfigs
* useSettings
* useStyle
* useGlobalStylesOutput

arthur791004:
Closing it via https://github.com/Automattic/wp-calypso/pull/77319. Will draft a p2 to describe how it works!

PR #77319, opened 2023-05-24
React 18: Fix @automattic/global-styles package

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/77048, https://github.com/Automattic/wp-calypso/pull/77046 -- this change fixes compatibility with WordPress packages after the react 18 update

## Proposed Changes

* Create a `gutenberg-bridge` folder inside the `@automattic/global-styles` package and
  * Unlock the private APIs we need and re-export them for Calypso to use.
  * Re-implement `mergeBaseAndUserConfigs` function as Gutenberg doesn't export it
  * Create `withExperimentalBlockEditorProvider` to wrap the component with `ExperimentalBlockEditorProvider` and eliminate the [compatStyles](https://github.com/WordPress/gutenberg/blob/f77958cfc13c02b0c0e6b9b697b43bbcad4ba40b/packages/block-editor/src/components/iframe/index.js#L127). The reason is the `compatStyles` are not required on Calypso and it will cause warnings since some of `compatStyles` might not have an `id` so that we cannot use it as key when rendering.
* Use the fixed version of `@wordpress/private-apis` to ensure we have the consistent `__private` symbol and avoid we're not able to unlock the private APIs due to the unequal private symbol. Or is dedupe enough?

There is a warning from Gutenberg below, but it looks like GB won't encounter this one 🤔
* `Warning: validateDOMNesting(...): <svg> cannot appear as a child of <head>.`

## Testing Instructions

**Theme Showcase**

* Go to `/themes/<your_site>`
* Ensure the style variation discs look good
* Pick a theme with style variations
* Ensure the style variations still look good on the theme details page

**Design Picker**

* Continue until you land on the Design Picker screen
* Ensure the style variation discs look good
* Pick a theme with style variations
* Ensure the style variations still look good on the preview screen

**Pattern Assembler**

* Continue until you land on the Design Picker screen
* Scroll down to the bottom
* Pick the BCPA CTA (Start designing)
* Select header/homepage/footer/colors/fonts
* Ensure everything looks good as before

## Pre-merge Checklist


packages/global-styles/src/components/global-styles-provider/index.tsx
```diff
@@ -17,57 +16,62 @@ const cleanEmptyObject = ( object: any ) => {
...
+				const updatedConfig = callback( currentConfig );
+				return {
+					styles: cleanEmptyObject( updatedConfig.styles ) || {},
+					settings: cleanEmptyObject( updatedConfig.settings ) || {},
+				};
+			} );
+		},
+		[ setUserConfig ]
+	);
+	return [ !! true, userConfig, setConfig ];
```
noahtallen: What's the purpose of `!! true` here?

packages/global-styles/src/gutenberg-bridge/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * See https://github.com/Automattic/wp-calypso/issues/77048
+ */
+import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
+import { createHigherOrderComponent } from '@wordpress/compose';
+import { __dangerousOptInToUnstableAPIsOnlyForCoreModules } from '@wordpress/private-apis';
+import mergeWith from 'lodash/mergeWith';
+import type { GlobalStylesObject } from '../types';
+
+const { unlock } = __dangerousOptInToUnstableAPIsOnlyForCoreModules(
+	'I know using unstable features means my plugin or theme will inevitably break on the next WordPress release.',
```
noahtallen, line 14: I hadn't realized they added this message, but I'm definitely a fan since it forces us to be explicit about it 😅

packages/global-styles/src/gutenberg-bridge/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+	const mergeTreesCustomizer = ( _: unknown, srcValue: unknown ) => {
+		// We only pass as arrays the presets,
+		// in which case we want the new array of values
+		// to override the old array (no merging).
+		if ( Array.isArray( srcValue ) ) {
+			return srcValue;
+		}
+	};
+
+	return mergeWith( {}, base, user, mergeTreesCustomizer );
```
noahtallen: We would like to avoid lodash if possible, but this could be a use case which is too difficult to replicate natively.

noahtallen (approved):
Overall, this looks good to me. I really like the gutenberg-bridge approach, which should ideally make it easier to change things in the future. I tested all the features you mentioned in the PR, and they are all testing great for me! Do we use any of these in the editor via ETK?

One thing I'm worried about is that the unlock API package version is from a newer release than the other Gutenberg packages. However, the unlock package doesn't have dependencies on other packages in Gutenberg, so I think it will be ok.

I think that most of the CI failures are from the underlying React PR and not from this one.

packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts
```diff
@@ -1,21 +1,20 @@
...
-	// 	setUserConfig( () =>
-	// 		globalStyles
-	// 			.filter( Boolean )
-	// 			.reduce( ( prev, current ) => mergeBaseAndUserConfigs( prev, current ), {} )
-	// 	);
-	// }, [ globalStyles, enabled ] );
-	// return user;
+	const { user, setUserConfig } = useContext( GlobalStylesContext );
+	useEffect( () => {
+		if ( ! enabled ) {
```
noahtallen: Note: this file changed on trunk, so it might need to be refreshed. (I resolved the merge conflict in the base PR to maintain compatibility with this branch)

packages/global-styles/src/gutenberg-bridge/index.tsx
arthur791004: Yes, we have to take time to see how to support `mergeWith` natively but I think it's not the goal of this PR. So, I'd prefer to do it later 🙂

packages/global-styles/src/components/global-styles-provider/index.tsx
arthur791004: To indicate the config is ready. We don't want to fetch the current user config from the server at the current stage, so it's always ready at the begining

packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts
arthur791004: Rebased. Thanks for noting that!

packages/global-styles/src/gutenberg-bridge/index.tsx
tyxla: I'd also love to not see new Lodash usage introduced. 

We've been using `deepmerge` as a replacement for `merge` and `mergeWith` in Gutenberg and it mostly works well. In this case, it appears to be a simple scenario where we want to disable merging arrays and just have them overriden. This can be achieved with something like:

```
import deepmerge from 'deepmerge';
import { isPlainObject } from 'is-plain-object';

deepmerge( obj1, obj2, {
	isMergeableObject: isPlainObject,
} );
```

and if you want to avoid using `is-plain-object` for some reason, you could do something like (untested):

```
import deepmerge from 'deepmerge';

deepmerge( obj1, obj2, {
	isMergeableObject: ( value ) => ! Array.isArray( value ),
} );

File: packages/global-styles/src/components/global-styles-provider/index.tsx [hunk 5]
```diff
@@ -1,7 +1,6 @@
-// import { GlobalStylesContext } from '@wordpress/edit-site/build-module/components/global-styles/context';
-// import { mergeBaseAndUserConfigs } from '@wordpress/edit-site/build-module/components/global-styles/global-styles-provider';
 import { isEmpty, mapValues } from 'lodash';
```
tyxla: I bet those Lodash usages should be pretty straightforward to remove. Let me know if I can aid with suggestions.

packages/subscriber/src/components/add-form/index.tsx
```diff
@@ -377,14 +379,16 @@ export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
...
-				target: '__blank',
-				rel: 'noreferrer',
-				href: localizeUrl(
-					'https://wordpress.com/support/launch-a-newsletter/import-subscribers-to-a-newsletter/'
-				),
-			} ),
+			Button: (
+				<Button
+					variant="link"
+					target="__blank"
```
tyxla: I assume you meant:

```suggestion
					target="_blank"
```

packages/subscriber/src/components/add-form/index.tsx
```diff
@@ -356,13 +356,15 @@ export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
...
-								variant: 'link',
-								target: '__blank',
-								href: localizeUrl(
-									'https://wordpress.com/support/launch-a-newsletter/import-subscribers-to-a-newsletter/'
-								),
-							} ),
+							Button: (
+								<Button
+									variant="link"
+									target="__blank"
```
tyxla:
```suggestion
									target="_blank"
```

packages/help-center/src/components/help-center-notice.tsx
```diff
@@ -39,10 +39,16 @@ function getResponses( siteName?: string ) {
 					),
 					{
 						hosted_on_our_services: (
-							<ExternalLink href={ localizeUrl( 'https://wordpress.com/support/com-vs-org/' ) } />
+							<ExternalLink
+								href={ localizeUrl( 'https://wordpress.com/support/com-vs-org/' ) }
+								children={ null }
```
tyxla, line 44: Why is that necessary? `children` will be `undefined` by default.

packages/help-center/src/components/help-center-notice.tsx
```diff
@@ -39,10 +39,16 @@ function getResponses( siteName?: string ) {
...
+							/>
 						),
 						wordpress_org_community_forums: (
-							<ExternalLink href={ localizeUrl( 'https://wordpress.org/support/forums/' ) } />
+							<ExternalLink
+								href={ localizeUrl( 'https://wordpress.org/support/forums/' ) }
+								children={ null }
```
tyxla, line 50: Why is that necessary? `children` will be `undefined` by default.

packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts
```diff
@@ -1,21 +1,20 @@
...
+	useEffect( () => {
+		setUserConfig( () =>
+			globalStyles
+				.filter( Boolean )
+				.reduce(
+					( prev, current ) => mergeBaseAndUserConfigs( prev, current ),
+					{} as GlobalStylesObject
+				)
+		);
+	}, [ globalStyles ] );
```
tyxla, line 16: Is there a chance `globalStyles` will be a new object on every re-render and `setUserConfig` will be called every time? Should we memoize the config merging and call `setUserConfig` only if the memoized version changes?

packages/global-styles/src/gutenberg-bridge/index.tsx
tyxla: Yup, I love it too 😉 Great tactics from a devex perspective. 👍

packages/global-styles/src/components/font-pairing-variations/preview.tsx
```diff
@@ -25,130 +22,132 @@ interface Props {
...
+										fontFamily: textFontFamily,
+										fontStyle: textFontStyle,
+										color: '#000000',
+										fontSize: '16px',
+										fontWeight: textFontWeight,
+										lineHeight: '1em',
+										textAlign: 'center',
+									} }
+								>
+									{ translate( 'Default' ) }
```
tyxla: Note that such simple strings often needs a context for the translators to know how to translate it in various languages.

packages/global-styles/src/components/global-styles-provider/index.tsx
tyxla: I think Noah was suggesting that we could remove the `!!` boolean casting, since it's already a boolean.

packages/help-center/src/components/help-center-notice.tsx
noahtallen: The problem is that some components require children via TS, which is hard to support in some cases (like when interpolating)

packages/subscriber/src/components/add-form/index.tsx
arthur791004: Fixed. But I found there are lots of `__blank` in our codebase 😂

packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts
arthur791004: We handle this case outside when using this hook.

> Should we memoize the config merging and call setUserConfig only if the memoized version changes?

How could we detect the memoized version changes? If the `globalStyles` is a new object on every re-render, then the memoized version will be changed as well.

packages/global-styles/src/gutenberg-bridge/index.tsx
arthur791004: Updated! TIL: `deepmerge` 👍

packages/global-styles/src/components/global-styles-provider/index.tsx
arthur791004: Done!

packages/subscriber/src/components/add-form/index.tsx
tyxla: Bummer. It's a good opportunity to fix them 😅  Fixed them all in #77484

packages/help-center/src/components/help-center-notice.tsx
tyxla: Makes sense, thanks.

Intuitively, I would suggest resolving this by altering the type to make the `children` prop optional, but a double-check at the `ExternalLink` code makes me think twice about it. Why do we have an external link without any text? Sounds like it could be an accessibility issue.

arthur791004: Actually, it has as it used by `createInterpolateElement` function to replace the custom-defined tag with the specified component, and there is the text between the tag that is the text for the `ExternalLink`. For example,

```js
// Input
createInterpolateElement(
  '{{a}}Learn more{{/a}}',
  { a: <ExternalLink /> }
);

// Output
<ExternalLink>Learn more</ExternalLink>
```

tyxla: Cool! Then (sorry for going in circles 😅 ), I'd ask the following question: what do you feel is the best in this case - passing `children={ null }`, ignoring the TS rule, or fixing the `ExternalLink` type to support an optional `children` prop`? If you picked one of those, why did you prefer that choice?

packages/global-styles/src/hooks/use-sync-global-styles-user-config.ts
tyxla: I was mostly thinking out loud and leaving this for your consideration, but if it's handled already, no need to worry about it 👍

packages/help-center/src/components/help-center-notice.tsx
arthur791004: I think the `ExternalLink` still needs the `children` so we should not make it optional. So, I'd prefer to pass `children={ null }` or even use `ts-expect-error` to bypass the rule and see whether we can fix this issue via `createInterpolateElement` as it should let TS know it will pass the children to that component. However, I'm not whether it's possible or not 😓

tyxla: Sounds good, thanks for confirming, let's keep the `children={ null }` then 👍 No need to overcomplicate things.

noahtallen: I also asked about this in Slack a couple weeks ago (it's a broader issue uncovered in the react 18 update): p1684448743933079-slack-CK1QYS6P6

noahtallen:
I'll go ahead and merge this into the main React 18 PR so that we have all the fixes in one place.

arthur791004:
Thank you!

noahtallen:
I've been doing a lot more work on the React 18 PR, and in an effort to split out much as possible from that PR, I've pinned the WordPress packages to a lower version (versions from around December 2022). The benefit is that we won't need to bundle this fix into the React 18 PR, but we'll still need to land it in the PR _after_ that when we update the WordPress packages to the current versions. So I restored the branch in anticipation of that

noahtallen:
Reopened to merge onto the wordpress package updates: #78715

PR #77325, opened 2023-05-24
Happy Blocks: Fix footer image position

renancarvalho:
Related to https://github.com/Automattic/wpsupport3/issues/452

## Proposed Changes

Fixes Happy blocks footer image and missing paddings.

Padding issue:
image

Misaligned image issue:
image

## Testing Instructions

- Pull and run this branch
- Checkout branch and `yarn dev --sync` from `apps/happy-blocks`
- Check [support](https://wordpress.com/support/) and [learn](https://learncft.wordpress.com/) if:
	-  The paddings around the tablet viewport is 24px like the other elements
	- The image on the desktop is aligned to the right on the footer.


apps/happy-blocks/block-library/support-content-footer/style.scss
```diff
@@ -11,8 +11,7 @@ $system-font-family: -apple-system, BlinkMacSystemFont, "Noto Sans", "Segoe UI",
 	max-width: none !important;
 	align-items: normal !important;
 	box-sizing: border-box;
-	padding-bottom: 20px;
-	padding-top: 20px;
+	padding: 20px 24px;
```
escapemanuele, line 14: Could you try without this change? I see the padding correctly even without changing this, I think it would be better so this can inherit the horizontal padding from other styles!

escapemanuele (commented):
Thank you for this, the image is indeed a lot better now 👍 

I left a small comment, nothing too important!

apps/happy-blocks/block-library/support-content-footer/style.scss
renancarvalho: This padding issue is only visible on the tablet viewport. It gets like this 
image
Could you please verify if you see it like this before this PR changes?

escapemanuele: image

I see this in production, padding coming from [here](https://github.com/Automattic/wpsupport3/blob/feat/add_breadcrumbs_learn/styles/templates/_style.scss#L56). What are you using as width/device for testing?

renancarvalho: Sorry for the lack of information, support is fine, this issue is visible on learn since they have different HTML structure.

On support, the footer content is inside `<main>` which sets the padding. Learn, on the other hand, has it outside, this is why this issue is surfacing on `/learn`

escapemanuele: Oooh ok then!

escapemanuele (approved):
Good to go!

apps/happy-blocks/block-library/support-content-footer/style.scss
renancarvalho: Also, posts pages on /support have it outside `<main>`. These cases will also be fixed. 👍

PR #77327, opened 2023-05-24
Jetpack Connect: Remove dependency on Happy Chat

klimeryk:
Part of sunsetting Happy Chat: pdm0RZ-BU-p2

I'm not sure what was the original intent here - JPOP does not do chat, so at best this would have connected to WPCOM HEs. Plus, it seems some of these flows are when user is logged out - which did not work with Happy Chat (which requires WPCOM user). All in all, I think it's best to go back to the fallback contact URL and then we can see if any changes are needed (like showing Help Center).

I've checked the Tracks events that would have been fired if these flows were working as expected and it seems they've never been fired:
- calypso_jpc_authorize_chat_initiated
- calypso_jpc_authorize_chat_initiated
- calypso_jpc_noqueryarguments_chat_initiated
- calypso_jpc_sso_chat_initiated
- calypso_jetpack_disconnect_chat_initiated

I'd appreciate double-checking that, though, as I might have missed something.

## Proposed Changes

* Remove dependency on Happy Chat.
* Use the default (or overridden) contact `url` from `HelpButton`.

## Testing Instructions
Make sure the contact link shows up on these pages and that it does not load Happy Chat, just directs to a contact page.

* http://calypso.localhost:3000/jetpack/connect
* http://calypso.localhost:3000/jetpack/connect/authorize

Screenshot 2023-05-24 at 12 19 14

Screenshot 2023-05-24 at 12 19 34


sergeymitr (approved):
Looks good, tests well 👍

PR #77329, opened 2023-05-24
Use Help Center in cancel purchase form instead of Happy Chat

klimeryk:
Part of sunsetting Happy Chat: pdm0RZ-BU-p2

We've switched to Zendesk Messaging. And with that, Happy Chat is being removed from WordPress.com. Help Center should be used instead.

## Proposed Changes

* Open chat from Help Center instead of directly invoking Happy Chat.

## Testing Instructions

TBD


klimeryk:
Abandoning in favor of https://github.com/Automattic/wp-calypso/pull/78173

PR #77357, opened 2023-05-24
Add plural support to translate function passed to the BlazePress widget

dlind1:
Related to 622-gh-Automattic/i18n-issues

## Proposed Changes

* Add support for the third parameter in the translate function passed to the BlazePress widget, to support translations with plural forms.

## Testing Instructions

* Follow the instructions from 135-gh-Tumblr/a8c-dsp

## Pre-merge Checklist

### DIFF
--- a/client/components/blazepress-widget/index.tsx
+++ b/client/components/blazepress-widget/index.tsx
@@ -26,7 +26,11 @@ export type BlazePressPromotionProps = {
 	source?: string;
 };
 
-type BlazePressTranslatable = ( original: string, extra?: TranslateOptionsText ) => string;
+type BlazePressTranslatable = (
+	original: string,
+	plural_or_extra?: string | TranslateOptionsText,
+	extra?: TranslateOptionsPluralText
+) => string;
 
 export function goToOriginalEndpoint() {
 	const { pathname } = getUrlParts( window.location.href );
@@ -94,7 +98,7 @@ const BlazePressWidget = ( props: BlazePressPromotionProps ) => {
 					(
 						original: string,
 						plural_or_options?: string | TranslateOptionsText,
-						options?: TranslateOptionsText
+						options?: TranslateOptionsPluralText
 					): string => {
 						if ( plural_or_options && options ) {
 							// This is a special case where we re-use the translate in another application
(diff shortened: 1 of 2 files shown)


jjolmo (approved):
LGTM

client/lib/promote-post/index.ts
```diff
@@ -25,7 +25,7 @@ declare global {
 				urn: string;
 				onLoaded?: () => void;
 				onClose?: () => void;
-				translateFn?: ( value: string, options?: any ) => string;
+				translateFn?: ( value: string, plural_or_options?: any, options?: any ) => string;
```
yuliyan: Could be replaced with `typeof translate` and `import { translate } from 'i18n-calypso/types';`.

client/components/blazepress-widget/index.tsx
```diff
@@ -91,12 +95,20 @@ const BlazePressWidget = ( props: BlazePressPromotionProps ) => {
 					props.postId,
 					onClose,
 					source,
-					( original: string, options?: TranslateOptionsText ): string => {
-						if ( options ) {
+					(
```
yuliyan: This wrapper around `translate` seems to be redundant as it doesn't seem to do anything but pass the same arguments to the original translate function.

I think we can just pass the `translate` function from the `useTranslate` hook.

client/components/blazepress-widget/index.tsx
```diff
@@ -26,7 +26,11 @@ export type BlazePressPromotionProps = {
 	source?: string;
 };
 
-type BlazePressTranslatable = ( original: string, extra?: TranslateOptionsText ) => string;
+type BlazePressTranslatable = (
```
yuliyan: This type def could be omitted if we use the translate function from the `useTranslate` hook as is, instead of creating a wrapper around.

yuliyan (changes requested):

client/lib/promote-post/index.ts
dlind1: Done!

client/components/blazepress-widget/index.tsx
dlind1: Thanks, done!

yuliyan (approved):
Thank you for addressing the feedback, LGTM!

PR #77387, opened 2023-05-25
Add-ons: Storage Add On Filtering

sharpirate:
This PR is based on https://github.com/Automattic/wp-calypso/pull/76754
Fixes 1721-gh-Automattic/martech

### Proposed Changes
Implement filtering for the available storage options based on the current max storage:

1. We want to offer 50 and 100 GB for now (this will likely change in the future)
2. A site should not have more than 200 GB of storage. If a user already has 200 GB of storage (e.g. provided by a plan), they should not have the option to add more storage.
3. Potentially, a user may be eligible for 50 GB, not 100 GB, depending on their current allowed storage.
4. User should be able to purchase each space upgrade only once

### Testing
1. Checkout this branch
2. Run `yarn` & `yarn start`
3. Create a new test site
4. Go to `http://calypso.localhost:3000/add-ons/YOUR_SITE_SLUG`
5. You should see the 2 new storage upgrades: 50 GB and 100 GB:

image

6. Sandbox store and `public-api.wordpress.com`
7. Use a test card to buy the 50 GB upgrade
8. Once the purchase is complete, click on the Add-Ons link in the side menu so you can go back to `http://calypso.localhost:3000/add-ons/YOUR_SITE_SLUG`.

Note: You probably won't see the updated data just yet because there seems to be an issue with`getPastBillingTransactions` not being able to read the data consistently. I've found a workaround for now that allows for testing:
9. Visit `http://calypso.localhost:3000/purchases/billing-history/calffff.wordpress.com` and verify you can see the storage purchase in the billing history tab
10. Manually change the URL back to `http://calypso.localhost:3000/add-ons/YOUR_SITE_SLUG`
11. You should see the 50 GB storage showing as "Manage" and the 100 GB storage should be available for purchase

https://github.com/Automattic/wp-calypso/assets/20927667/96a32b94-bda3-4dfb-adb8-2cf38139e549

12. You should then be able to purchase the 100 GB storage upgrade
13. Repeat the workaround from before and verify you can manage both add ons

image

14. Do some more general testing:
• Verify the add ons display correctly according to the 200 GB storage limit 
• Verify that you can only buy the add ons in an ascending order - you should be able to buy the 50 GB and then the 100 GB but if you start out with the 100 GB the 50 GB shouldn't be visible

### DIFF
--- a/client/my-sites/add-ons/hooks/use-add-ons.ts
+++ b/client/my-sites/add-ons/hooks/use-add-ons.ts
@@ -14,7 +14,7 @@ import {
 	getProductName,
 } from 'calypso/state/products-list/selectors';
 import getBillingTransactionFilters from 'calypso/state/selectors/get-billing-transaction-filters';
-import getPastBillingTransactions from 'calypso/state/selectors/get-past-billing-transactions';
+import { useBillingTransactions } from 'calypso/state/sites/hooks/use-billing-history';
 import { STORAGE_LIMIT } from '../constants';
 import customDesignIcon from '../icons/custom-design';
 import spaceUpgradeIcon from '../icons/space-upgrade';
@@ -85,11 +85,12 @@ const useAddOns = ( siteId?: number ): ( AddOnMeta | null )[] => {
 	// if upgrade is bought - show as manage
 	// if upgrade is not bought - only show it if available storage and if it's larger than previously bought upgrade
 	const { data: mediaStorage } = useMediaStorageQuery( siteId );
+	const { billingTransactions } = useBillingTransactions( 'past' );
 
 	return useSelector( ( state ): ( AddOnMeta | null )[] => {
-		const transactions = getPastBillingTransactions( state );
 		const filter = getBillingTransactionFilters( state, 'past' );
-		const filteredTransactions = transactions && filterTransactions( transactions, filter, siteId );
+		const filteredTransactions =
+			billingTransactions && filterTransactions( billingTransactions, filter, siteId );
 
 		const spaceUpgradesPurchased: BillingTransactionItem[] = [];
 
--- a/client/my-sites/add-ons/hooks/use-add-ons.ts
+++ b/client/my-sites/add-ons/hooks/use-add-ons.ts
@@ -7,6 +7,7 @@ import { useTranslate } from 'i18n-calypso';
 import useMediaStorageQuery from 'calypso/data/media-storage/use-media-storage-query';
 import { filterTransactions } from 'calypso/me/purchases/billing-history/filter-transactions';
 import { useSelector } from 'calypso/state';
+import { BillingTransactionItem } from 'calypso/state/billing-transactions/types';
 import {
 	getProductBySlug,
 	getProductDescription,
@@ -90,7 +91,7 @@ const useAddOns = ( siteId?: number ): ( AddOnMeta | null )[] => {
 		const filter = getBillingTransactionFilters( state, 'past' );
 		const filteredTransactions = transactions && filterTransactions( transactions, filter, siteId );
 
-		const spaceUpgradesPurchased = [];
+		const spaceUpgradesPurchased: BillingTransactionItem[] = [];
 
 		if ( filteredTransactions?.length ) {
 			for ( const transaction of filteredTransactions ) {
@@ -116,7 +117,7 @@ const useAddOns = ( siteId?: number ): ( AddOnMeta | null )[] => {
 				// if storage add on is already purchased
 				if (
 					spaceUpgradesPurchased.findIndex(
-						( item ) => parseInt( item.licensed_quantity ) === addOn.quantity
+						( item ) => Number( item.licensed_quantity ) === addOn.quantity
 					) >= 0
 				) {
 					return {
@@ -131,7 +132,7 @@ const useAddOns = ( siteId?: number ): ( AddOnMeta | null )[] => {
 				const availableStorageUpgrade = STORAGE_LIMIT - currentMaxStorage;
 
 				// if the current storage add on option is greater than the available upgrade, remove it
-				if ( ( addOn.quantity as number ) > availableStorageUpgrade ) {
+				if ( ( addOn.quantity ?? 0 ) > availableStorageUpgrade ) {
 					return null;
 				}
 			}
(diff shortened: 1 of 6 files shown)


client/my-sites/add-ons/hooks/use-add-ons.ts
```diff
@@ -73,10 +75,28 @@ const useAddOns = (): ( AddOnMeta | null )[] => {
...
+			// if storage add ons are not enabled in the config, remove them
+			if ( ! isStorageAddonEnabled() ) {
+				return false;
+			}
+
+			const currentMaxStorage = mediaStorage?.max_storage_bytes / Math.pow( 1024, 3 );
+			const availableStorageUpgrade = STORAGE_LIMIT - currentMaxStorage;
+
+			// if the current storage add on option is greater than the available upgrade, remove it
+			if ( ( addOn.quantity as number ) > availableStorageUpgrade ) {
```
sirbrillig: Really minor and not a blocker but this can be done without needing to resort to `as` (in this case it's quite safe but I always like to encourage folks to avoid `as` when possible... and it's not always possible so this is just a suggestion).

```suggestion
			if ( ( addOn.quantity ?? 0 ) > availableStorageUpgrade ) {
```

gmovr:
Thanks for all the work on this!

I'm leaving some comments:
- I tested with Premium and Free plans, and the core logic works well.
    - I'm sent to the correct checkout URL for each quantity.
    - When I purchase an add-on (after re-fetching the data*) the purchase status works, and I see the correct purchase status of 50 GB vs 100 GB. E.g. I can buy 50 GB, while still seeing 100 GB as purchasable. Upgrading to 100 GB will then change the buttons for both 50 and 100 GB to "manage", which is correct.
- On a Business plan, I'm not seeing the space add-ons (which is the correct behavior for now). 


*Stale data
- The reason why the purchase state is not correct, and you need to go to /billing-history/ is because at the add-ons screen the `state.billingTransactions` is not updated (and the selectors you've used rely on this data). 
- This data comes from the `QueryBillingTransactions` component. I'm not sure if this is correct, as I'm not too familiar with the code, but I tested this by mounting the `QueryBillingTransactions` component in main.tsx in the add-ons route, and the data is then in sync without going to the billing transactions view first. 

Screenshot 2023-06-06 at 11 58 40

@sirbrillig -- I see you've worked on that code. Will using the billing history here be the best way? We're trying to maintain the licensed quantity (current quantity owned by user) for rendering the purchase vs manage button. 

The previous add-ons code only checks the product slug, while for a tiered-volume product, the slug will be the same, while the quantity/licensed quantity will differ.

gmovr (approved):
As this is behind a feature flag, you can deploy it (rebase to ensure mobile e2e passes), as I don't want to block it.

When I go to prod, there are no changes to the add-ons screen. You can follow up with the needed filtering and any optimization here (can we memoize any of these checks? or do other optimizations?)

Once a new PR is opened for fixing the filtering of licensed quantity, the final PR will be to open the feature flag in prod. 👍

sirbrillig:
> I see you've worked on that code. Will using the billing history here be the best way? We're trying to maintain the licensed quantity (current quantity owned by user) for rendering the purchase vs manage button.

I don't fully grok the question, but if you want to take an action in calypso based on the most up-to-date data for a user or site, the best thing to do is to query that data as close to the place that uses it as possible. 

Right now we use query components and Redux state for a lot of data but over time we should be removing that and replacing it with `react-query` hooks ([example where we replaced the stored cards state](https://github.com/Automattic/wp-calypso/pull/74676)) to make sure the data is always up-to-date.

If you believe the data returned by the billing history endpoint will do what you need, you might consider creating a `useBillingHistory` hook (or something like that) that fetches that data. We could then later replace all uses of the Redux state with that hook.

Does that help?

sharpirate:
I've implemented a hook `useBillingTransactions` for fetching the billing history with react-query which seems to be working well as per my testing. Would love some feedback @gmovr @sirbrillig

client/state/sites/hooks/use-billing-history.ts
```diff
@@ -0,0 +1,32 @@
...
+} from 'calypso/state/billing-transactions/types';
+
+export const billingTransactionsQueryKey = 'use-stored-payment-methods';
+
+interface BillingHistory {
+	billing_history: BillingTransaction[];
+}
+
+const fetchBillingTransactions = ( transactionType?: BillingTransactionsType ) =>
+	wp.req.get( `/me/billing-history` + ( transactionType ? `/${ transactionType }` : '' ), {
```
sirbrillig: If you pass `'upcoming'` to this hook, then the endpoint will return an `upcoming_charges` array [as shown here](https://github.com/Automattic/wp-calypso/blob/87cd675fd6f7e45c1730ceb489fee3bcda392bc8/client/state/billing-transactions/actions.ts#L34), but that isn't covered by the types here.

Honestly, the two types of transactions (past and future) are different enough that I think they might deserve different hooks rather than trying to clarify which data will be returned with which option. I think that 90% of the time callers just need the `'past'` data, so I suggest that you make this hook only return that data (removing the option to pass `transactionType`) and forget about `'upcoming'`, leaving that for another hook in the future.

gmovr:
Screenshot 2023-06-15 at 14 15 25

I tested again, now the manage/purchase works 👍  -- but after buying 100 GB, I'm unable to see it (it only shows manage on 50 GB). I think it's because we check if 100 GB is purchasable based on current storage (which it won't be).

I'm also seeing 50 GB purchasable, if I buy 100 GB first, and not upgrading from 50.

sharpirate:
@sirbrillig I've updated the hook to only fetch `past` billing transactions

@gmovr 
> I'm also seeing 50 GB purchasable, if I buy 100 GB first, and not upgrading from 50.

I've updated the PR to filter out space upgrades that are smaller than the smallest currently purchased so that way a user can only purchased upgrades in an ascending order

> but after buying 100 GB, I'm unable to see it (it only shows manage on 50 GB)

I wasn't able to repro this issue, it seems to be working fine for me. It's weird because I'm first checking if the space upgrade is purchased and only then am I checking the current storage limit... Can you provide a bit more details on how to repro this issue?

sirbrillig:
>  I've updated the hook to only fetch past billing transactions

Thanks! This looks good.

gmovr (approved):
Tested again, and this works well. The code looks good too. 

I see some minor bugs. I'll make separate GH issues for them (e.g. cancelling an add-on, and changing button state after the data is fetched, which can take some time). 

They are all pretty minor and not blocking, as this is behind a feature flag anyways. 🚢

PR #77392, opened 2023-05-25
Newsletter onboarding: skip intro step if ref newsletter-lp

### LINKED ISSUE #77245: Newsletter onboarding: tighten experience from the landing page into flow
Few independent tasks regarding landing page (https://wordpress.com/newsletter/), in order of priority:

- [ ] Skip intro when coming from the landing page (`?ref=newsletter-lp`) but show it otherwise
- [ ] Update Intro & loading page visuals to better match with landing page https://github.com/Automattic/wp-calypso/issues/77309
- [ ] Skip "goal" selection (that we haven't implemented yet) if we know from landing page that they want to migrate a newsletter, or that they want a paid newsletter.

Landing page PT pdZzpt-rd-p2

lezama:
Resolves: https://github.com/Automattic/wp-calypso/issues/77245
> Skip intro when coming from the landing page (?ref=newsletter-lp) but show it otherwise

### Proposed Changes
* Add and adjust conditional logic for Newsletter flows to handle both logged in state and ?ref=newsletter-lp.

### Testing Instructions

**Test Logged In User**
While logged into your WordPress.com account:
1) Go to http://calypso.localhost:3000/setup/newsletter - you should see the intro step
2) Go to http://calypso.localhost:3000/setup/newsletter/intro - you should see the intro step
3) Go to http://calypso.localhost:3000/setup/newsletter?ref=newsletter-lp - you should see the setupNewsletter step

**Test Non-Logged In User**
While NOT logged into your WordPress.com account:
1) Go to http://calypso.localhost:3000/setup/newsletter. Confirm you see the intro step. Click 'Launch you newsletter' button, and confirm you go to account creation. Create account or login, and confirm you go to Setup. 
2) Go to http://calypso.localhost:3000/setup/newsletter/intro. Confirm you see the intro step. Click 'Launch you newsletter' button, and confirm you go to account creation. Create account or login, and confirm you go to Setup. 
3) Go to http://calypso.localhost:3000/setup/newsletter?ref=newsletter-lp. Confirm you go directly to account creation. Create account or login, and confirm you go to Setup screen.

### DIFF
--- a/client/landing/stepper/declarative-flow/newsletter.ts
+++ b/client/landing/stepper/declarative-flow/newsletter.ts
@@ -110,16 +110,15 @@ const newsletter: Flow = {
 
 		function submit( providedDependencies: ProvidedDependencies = {} ) {
 			recordSubmitStep( providedDependencies, '', flowName, _currentStep );
+			const logInUrl = getStartUrl();
 
 			switch ( _currentStep ) {
 				case 'intro':
 					clearSignupDestinationCookie();
-					// This redirect for non-logged in users nearly duplicates one above,
-					// but is needed to avoid a short flash of the newsletterSetup screen.
-					if ( ! userIsLoggedIn ) {
-						return window.location.assign( getStartUrl() );
+					if ( userIsLoggedIn ) {
+						return navigate( 'newsletterSetup' );
 					}
-					return navigate( 'newsletterSetup' );
+					return window.location.assign( logInUrl );
 
 				case 'newsletterSetup':
 					clearSignupDestinationCookie();
--- a/client/landing/stepper/declarative-flow/newsletter.ts
+++ b/client/landing/stepper/declarative-flow/newsletter.ts
@@ -141,6 +141,8 @@ const newsletter: Flow = {
 						persistSignupDestination( destination );
 						setSignupCompleteSlug( providedDependencies?.siteSlug );
 						setSignupCompleteFlowName( flowName );
+
+						// Return to subscribers after checkout
 						const returnUrl = encodeURIComponent(
 							`/setup/${ flowName }/subscribers?siteSlug=${ providedDependencies?.siteSlug }`
 						);
@@ -151,7 +153,7 @@ const newsletter: Flow = {
 							) }?redirect_to=${ returnUrl }&signup=1`
 						);
 					}
-					// If the user chooses a free plan, we need to redirect to the subscribers and not checkout.
+					// If the user chooses a free plan, we need to redirect to the subscribers directly and not checkout.
 					return window.location.assign(
 						`/setup/${ flowName }/subscribers?siteSlug=${ providedDependencies?.siteSlug }`
 					);
--- a/client/landing/stepper/declarative-flow/newsletter.ts
+++ b/client/landing/stepper/declarative-flow/newsletter.ts
@@ -75,6 +75,8 @@ const newsletter: Flow = {
 		const { setStepProgress } = useDispatch( ONBOARD_STORE );
 		const query = useQuery();
 		const isComingFromMarketingPage = query.get( 'ref' ) === 'newsletter-lp';
+		const isLoadingIntroScreen =
+			! isComingFromMarketingPage && ( 'intro' === _currentStep || undefined === _currentStep );
 
 		const flowProgress = useFlowProgress( {
 			stepName: _currentStep,
@@ -82,11 +84,17 @@ const newsletter: Flow = {
 		} );
 		setStepProgress( flowProgress );
 		const locale = useLocale();
+
 		const getStartUrl = () => {
 			return locale && locale !== 'en'
 				? `/start/account/user/${ locale }?variationName=${ flowName }&pageTitle=Newsletter&redirect_to=/setup/${ flowName }/newsletterSetup`
 				: `/start/account/user?variationName=${ flowName }&pageTitle=Newsletter&redirect_to=/setup/${ flowName }/newsletterSetup`;
 		};
+
+		// Unless showing intro step, send non-logged-in users to account screen.
+		if ( ! isLoadingIntroScreen && ! userIsLoggedIn ) {
+			window.location.assign( getStartUrl() );
+		}
 
 		// trigger guides on step movement, we don't care about failures or response
 		wpcom.req.post(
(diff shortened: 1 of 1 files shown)


client/landing/stepper/declarative-flow/newsletter.ts
```diff
@@ -93,14 +106,14 @@ const newsletter: Flow = {
...
 			const logInUrl = getStartUrl();
 
+			if ( ! userIsLoggedIn ) {
+				clearSignupDestinationCookie();
+				return window.location.assign( logInUrl );
+			}
+
 			switch ( _currentStep ) {
 				case 'intro':
-					clearSignupDestinationCookie();
```
edanzer, line 98: In the old version, clearSignupDestinationCookie() was invoked whether a user was logged in or not. In the new version, it is only invoked if the user is NOT logged in. Not sure if there's any concern about that difference or not?

edanzer:
@lezama - Just tested this. 

As a logged in user, this seems to work as expected. 
   - GOOD. http://calypso.localhost:3000/setup/newsletter -> go to intro step
   - GOOD. http://calypso.localhost:3000/setup/newsletter?ref=newsletter-lp - go to setupNewsletter step
   
As a logged out user, I think something may not be quite right for the lp case. 
   - GOOD. http://calypso.localhost:3000/setup/newsletter -> go to intro step. If I then click the button to continue, I get the account page. If I setup an account, I go to setupNewsletter screen. This works as before. 
   - QUESTIONS. http://calypso.localhost:3000/setup/newsletter?ref=newsletter-lp
      - So this goes to setupNewsletter step screen. If I fill out the form and click continue, I go to the account page. If I create an account, I then get sent back to the setupNewsletter screen. This still ultimately works, and I'm not sure if that's intended flow, but seems like we'd ideally want to avoid going back to the same screen?
      - If so, seems like we'd have two options. First, we send a user directly to the account screen first, then only to the setup screen. Second, we still start with setup, then account, and go directly to domains.

lezama:
> QUESTIONS. http://calypso.localhost:3000/setup/newsletter?ref=newsletter-lp
So this goes to setupNewsletter step screen. If I fill out the form and click continue, I go to the account page. If I create an account, I then get sent back to the setupNewsletter screen. This still ultimately works, and I'm not sure if that's intended flow, but seems like we'd ideally want to avoid going back to the same screen?

it should go to the account page first? will take a look!

client/landing/stepper/declarative-flow/newsletter.ts
```diff
@@ -24,8 +25,16 @@ const newsletter: Flow = {
...
+		const isComingFromMarketingPage = query.get( 'ref' ) === 'newsletter-lp';
+
 		return [
-			{ slug: 'intro', asyncComponent: () => import( './internals/steps-repository/intro' ) },
+			// Load intro step component only when not coming from the marketing page
+			...( ! isComingFromMarketingPage
+				? [
+						{ slug: 'intro', asyncComponent: () => import( './internals/steps-repository/intro' ) },
+				  ]
+				: [] ),
```
simison, line 37: I found the optional spread be simpler to look at, but it's also "one liner trickery". Happy to revert back to previous if you think that looked better. :-)

client/landing/stepper/declarative-flow/newsletter.ts
```diff
@@ -129,6 +140,8 @@ const newsletter: Flow = {
 						persistSignupDestination( destination );
 						setSignupCompleteSlug( providedDependencies?.siteSlug );
 						setSignupCompleteFlowName( flowName );
+
+						// Return to subscribers after checkout
```
simison, line 151: Unrelated to this PR but I liked the clarification.

client/landing/stepper/declarative-flow/newsletter.ts
```diff
@@ -157,9 +170,8 @@ const newsletter: Flow = {
 			switch ( _currentStep ) {
 				case 'launchpad':
 					return window.location.assign( `/view/${ siteSlug }` );
-
 				default:
-					return navigate( 'intro' );
+					return navigate( isComingFromMarketingPage ? 'newsletterSetup' : 'intro' );
```
simison, line 181: I liked having this conditional in the same line where it's used, instead of declaring on top of the file, and using down in the file.

simison:
Made some changes and elaborated on them with inline comments.

There's some bugginess with logged-out state, that I couldn't figure out yet.

> it should go to the account page first? 

Yes

client/landing/stepper/declarative-flow/newsletter.ts
lezama: works for me

client/landing/stepper/declarative-flow/newsletter.ts
simison: I think `clearSignupDestinationCookie` could be moved to `useSideEffect` — it's a side-effect (we clear cookies), it runs always and early.

edanzer: I went ahead and made this change.

edanzer:
This PR has been updated to handle mixed logic (which step we're loading, whether logged in, whether ?ref=newsletter-lp is set). I tested all the mixed scenarios reasonably well. 

Worth noting that we could simplify logic if we ALWAYS showed the account/login page first for non-logged in users. There's extra logic now to show the 'intro' page (and only that page) to non-logged in users. 

If we made the above change logic would be: 
```
if ( not logged in ) { 
  // always go to account page
  // after login/account creation, go to 'intro' or 'setup' depending on ?ref=newsletter-lp
}
```

client/landing/stepper/declarative-flow/newsletter.ts
```diff
@@ -24,8 +25,16 @@ const newsletter: Flow = {
 		return translate( 'Newsletter' );
 	},
 	useSteps() {
+		const query = useQuery();
+		const isComingFromMarketingPage = query.get( 'ref' ) === 'newsletter-lp';
+
```
edanzer, line 30: Worth noting we duplicate the two lines above on line 77-78 below, because we need the same info in both useSteps and useStepsNavigation hooks. If the duplication bothers us, we could remove the lines here, always return the same set of steps, but add a conditional redirect in useStepsNavigation.

(pseudocode, but something like ):
```
if ( useIsLoggedIn && isComingFromMarketingPage && stepIsIntroOrDefault ) {
  navigate( 'newsletterSetup' )
}
```

simison: I was thinking that too but it's fine — I don't think anyone will miss it when doing changes.

The best would be if signup page would be part of stepper.

simison (approved):
Code looks neat and tested all scenarios:

- Logged out without ref: intro → signup → login → setup →
- Logged in without ref: intro → setup → 
- Logged out with ref: signup → login → setup →
- Logged in with ref: setup →

PR #77408, opened 2023-05-25
Subscription Management: Fix infinite fetching of comment subscriptions.

JuanLucha:
## Proposed Changes

This PR fixes the infinite calls to the comment list endpoint. It's far from perfect, but it's a temporal fix, only valid while we fix the backend to return the correct number of comment subscriptions.

It still makes more calls than needed, but that's due to the asynchronous nature of the state hook used to stop the fetching.

## Testing Instructions

1. Apply this patch.
2. Add `return next();` right after the line `app.get( [ '/subscriptions', '/subscriptions/*' ], function ( req, res, next ) {` in the file `client/server/pages/index.js`.
3. Start the application.
4. Go to `http://calypso.localhost:3000/subscriptions/comments` with the network tab from the developer tools open.
5. Check that the calls to `/wpcom/v2/post-comment-subscriptions` stops after some of them.

### DIFF
--- a/packages/data-stores/src/reader/queries/use-post-subscriptions-query.ts
+++ b/packages/data-stores/src/reader/queries/use-post-subscriptions-query.ts
@@ -1,5 +1,5 @@
 import { useInfiniteQuery } from '@tanstack/react-query';
-import { useCallback, useEffect, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 import { callApi } from '../helpers';
 import { useCacheKey, useIsLoggedIn, useIsQueryEnabled } from '../hooks';
 import type { PostSubscription } from '../types';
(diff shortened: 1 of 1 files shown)


yansern:
The PR is not working for me, it is still making infinite requests.
Also, I don't think it should stop fetching just because the comments array is empty, e.g. page 1 could contain 500 irrelevant items, but page 2 contains say 100 items that are relevant.

I think we should stop querying the moment all the pages have been queried, e.g. 5000 items of 500 pagination size = 5 pages, so when page number reaches number 5, it should stop querying.

yansern (approved):
This works. Thanks! Although I think `useRef` is not needed here if the logic is calculated directly in `getNextPageParam`.

yansern:
Tested again with the updated changes. Works as expected.

PR #77412, opened 2023-05-25
Sites page: use consistent URLs in the CTAs

zaguiini:

## Proposed Changes

Standardizes and centralizes the generation of the Sites page CTAs. This is necessary because we're sharing query parameters between states and at least two variations of the CTA's target so we need to extract this logic and reuse it.

## Testing Instructions

The "Create site"/"Add new site" button should point to...

Developer account in the hosting flow: `/start/new-hosted-site?hosting-flow=true`
Developer account not in the hosting flow: `/start/new-hosted-site`
Not a developer account: `/start`

Check that the "Import existing site" CTA points to the exact same place in both of the experiences: `/start/import?source=sites-dashboard`.

The `ref=calypso-nosites` property should be added conditionally, checking if the user has a site or not.

### DIFF
--- a/client/sites-dashboard/components/sites-dashboard-ctas.tsx
+++ b/client/sites-dashboard/components/sites-dashboard-ctas.tsx
@@ -1,7 +1,9 @@
 import { useI18n } from '@wordpress/react-i18n';
+import { addQueryArgs } from '@wordpress/url';
 import { useSelector } from 'react-redux';
 import getCurrentQueryArguments from 'calypso/state/selectors/get-current-query-arguments';
 import { useSitesDashboardCreateSiteUrl } from '../hooks/use-sites-dashboard-create-site-url';
+import { TRACK_SOURCE_NAME } from '../utils';
 import { EmptyStateCTA } from './empty-state-cta';
 
 export const CreateSiteCTA = () => {
@@ -31,7 +33,10 @@ export const MigrateSiteCTA = () => {
 		<EmptyStateCTA
 			description={ __( 'Bring a site to WordPress.com' ) }
 			label={ __( 'Migrate a site' ) }
-			target="/start/import"
+			target={ addQueryArgs( '/start/import', {
+				source: TRACK_SOURCE_NAME,
+				ref: 'calypso-nosites',
+			} ) }
 		/>
 	);
 };
--- a/client/sites-dashboard/components/sites-dashboard-ctas.tsx
+++ b/client/sites-dashboard/components/sites-dashboard-ctas.tsx
@@ -1,9 +1,8 @@
 import { useI18n } from '@wordpress/react-i18n';
-import { addQueryArgs } from '@wordpress/url';
 import { useSelector } from 'react-redux';
 import getCurrentQueryArguments from 'calypso/state/selectors/get-current-query-arguments';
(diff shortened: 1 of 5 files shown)


p-jackson:
@zaguiini I agree that we should have a `useSitesDashboardCreateSiteUrl` hook to contain the logic, but I disagree that the `is_dev_account` flag should come into it.

As this PR is currently written:
- Developers can't make simple sites. Checking the "I'm a developer" box said I'd get early access to advanced features, it didn't say anything about losing the ability to make free WordPress.com sites (which is what we're famous for)
- Non-developers who were interested in hosting features (they just came from the `/hosting` LP) are getting offered a simple site

How I imagined this working is that `useSitesDashboardCreateSiteUrl` would have logic that looked up where the user has just come from. If they've just come from the `/hosting` LP then they go through the `new-hosted-site` flow. Otherwise we use the regular `/start` flow. That way devs and non-devs alike who are on the overall hosting flow (or journey, whatever broad term refers to the LOLP-to-new-site progression) are seeing the experience we're creating for this project.

From my point of view, maybe there's some future change we want to make where the `is_dev_account` flag plays a larger roll in what plans we recommend to you. Maybe the regular `/start/plans` step should have a toggle to switch between Simple and Atomic compatible plans, since that's something devs can understand. But when we do it it should work for all flows. e.g. The "Add new site" button in the Calypso sidebar shouldn't show different plans to what you see if you start the site creation process from `/sites`

p-jackson:
Not sure if what I've said is actually clear so here's some untested code for what I mean https://github.com/Automattic/wp-calypso/pull/77420/files

Seems to me that the original project was about the hosting flow, and we happened to sneak in a little checkbox about being a developer, but that's not really a critical part of the flow IMO

mpkelly:
I'm with @p-jackson here. I just checked ten random dev accounts, and almost all of them have Simple sites, so I think we should reconsider this change. 

And as mentioned on the call earlier, I wonder how many people will see this UI anyway, as you have to break out of the onboarding flow, which tries to get new users to create a site from the start.

zaguiini:
Thanks @Automattic/quake, created a P2 post to discuss this further: pet6gk-u0-p2

zaguiini:
Alright @Automattic/quake I think this is good to go now. This was trickier to achieve than I was expecting, but here we are. I'm even starting to doubt if this is even valuable, but this feeling goes away as soon as I review the testing instructions and how difficult would be to maintain separate implementations.

vykes-mac:
> Developer account in the hosting flow: /start/new-hosted-site?hosting-flow=true
> Developer account not in the hosting flow: /start/new-hosted-site

Did you mean  `/setup/new-hosted-site` instead?

For dev accounts testing shows` /setup/***` instead of `/start`

zaguiini:
Yes, sorry!Sent from my iPhoneOn 31 May 2023, at 15:42, vykes-mac ***@***.***> wrote:﻿

Developer account in the hosting flow: /start/new-hosted-site?hosting-flow=true
Developer account not in the hosting flow: /start/new-hosted-site

Did you mean  /setup/new-hosted-site instead?

vykes-mac (approved):
Looks good to me

PR #77414, opened 2023-05-26
Import hosted site flow: allow going back to /sites

zaguiini:

## Proposed Changes

This PR adds the capability to go back after clicking "Import a site" from the `/sites` page.

One thing I don't like is that a full-page refresh happens. How to replace the component instead?

## Testing Instructions

1. Enable the `hosting-onboarding-i2` flag;
2. In an account with no sites, click the "Import a site" button;
3. Verify that the "Skip to dashboard" button on the top-right corner is gone;
4. Verify that the "Back" button in the top-left corner shows up;
5. Click it and you should be redirected to `/sites`.

### DIFF
--- a/client/landing/stepper/declarative-flow/import-flow.ts
+++ b/client/landing/stepper/declarative-flow/import-flow.ts
@@ -220,6 +220,16 @@ const importFlow: Flow = {
 
 		const goBack = () => {
 			switch ( _currentStep ) {
+				case 'import': {
+					const source = urlQueryParams.get( 'source' );
+
+					if ( source === 'sites-dashboard' ) {
+						return window.location.assign( '/sites' );
+					}
+
+					return;
+				}
+
 				case 'importList':
 					// eslint-disable-next-line no-case-declarations
 					const backToStep = urlQueryParams.get( 'backToStep' );
--- a/client/landing/stepper/declarative-flow/import-flow.ts
+++ b/client/landing/stepper/declarative-flow/import-flow.ts
@@ -113,6 +113,8 @@ const importFlow: Flow = {
 		};
 
 		const submit = ( providedDependencies: ProvidedDependencies = {} ) => {
+			const source = urlQueryParams.get( 'source' );
+
 			switch ( _currentStep ) {
 				case 'importReady': {
 					const depUrl = ( providedDependencies?.url as string ) || '';
@@ -126,7 +128,7 @@ const importFlow: Flow = {
 						return exitFlow( providedDependencies?.url as string );
 					}
 
-					return navigate( providedDependencies?.url as string );
+					return navigate( addQueryArgs( providedDependencies?.url as string, { source } ) );
 				}
 				case 'importReadyPreview': {
 					return navigate( providedDependencies?.url as string );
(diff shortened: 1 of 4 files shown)


client/landing/stepper/declarative-flow/import-flow.ts
```diff
@@ -219,13 +220,23 @@ const importFlow: Flow = {
...
+					return;
+				}
+
 				case 'importList':
 					// eslint-disable-next-line no-case-declarations
 					const backToStep = urlQueryParams.get( 'backToStep' );
 
 					if ( backToStep ) {
-						const path = `${ backToStep }?siteSlug=${ siteSlugParam }`;
+						const path = addQueryArgs( backToStep, { siteSlug: siteSlugParam, source } );
```
p-jackson, line 241: By default, the `navigate()` function retains all query parameters as you move between steps. However as soon as you modify the query parameters you're expected to manage all of them yourself.

https://github.com/Automattic/wp-calypso/blob/57df03537af291e9775a93cb293cddfdc03bd55c/client/landing/stepper/declarative-flow/internals/index.tsx#L92

So we need to manually pass the `?source` param manually.

p-jackson (approved):
Tested and navigation works as expected 👍 
I noticed that if you advanced a little further into the flow—past the first step—then the `source` param ends up getting dropped. So I've pushed a fix to ensure the query param is preserved as you navigate backwards and forwards.

mpkelly:
It works well in most cases, but I found a way to get the old "Skip to dashboard" UI to reappear. Here's a short video:

https://github.com/Automattic/wp-calypso/assets/6851384/a8819b3c-2569-4c65-a6f7-afca7eef2d43

I will try and fix this now.

mpkelly:
It looks like there is another problem but not related to this patch. If you do as I did in the video above (choose the WordPress.org importer), but don't hit "Back", then the importing UI gets stuck and just shows the spinner. The console logs contain this error:

```GET https://public-api.wordpress.com/rest/v1.1/sites/?http_envelope=1&locale=en-gb 404```

Still investigating this one but should probably be a separate issue.

p-jackson:
I've fixed the issue where clicking back while the import is in progress drops the `?source` query param (which is what caused the "Skip to dashboard" UI to appear.

https://github.com/Automattic/wp-calypso/assets/1500769/babfdadf-4db1-46cc-ac8e-dd20a9c0fada

The issue was there are a number of places in `submit()` which overwrite all query parameters, and so `?source` gets lost. The underlying issue appears to be that the `getFinalImporterUrl()` function calculates the navigation URL [inside the context of the step](https://github.com/Automattic/wp-calypso/blob/6e9d2dab95d123e50bfd22c4abf6d118d048b4db/client/landing/stepper/declarative-flow/internals/steps-repository/import-ready/index.tsx#L33) and then passes the URL to `submit()`. But I think the approach is supposed to be that the logic for determining the navigation URLs is supposed to be done in the `useStepNavigation()` hook rather than by the step itself. That'd make it easier to manage the query params without having to add code to each step for dealing with the `?source` param.

I've only fixed the bug for the example in [the video above](https://github.com/Automattic/wp-calypso/pull/77414#issuecomment-1563947078), but there are other navigations where query params are being dropped, so it could be something that needs a refactor CC @Automattic/caribou

zaguiini:
Thanks, @p-jackson. I'm afraid that we'll need to put this PR on hold until we figure out a way to use this flow without a pre-selected site (p1685139634815589-slack-C04H4NY6STW).

vishnugopal:
What's the status of this PR? Are we circling back to this? 🤔

zaguiini:
We'll need to redo most of it since we have added a new flow @vishnugopal. I'll take care of this today.

zaguiini:
Looks like this PR is not necessary. Clicking "Back" on the `/setup/import-hosted-site` flow does get the user back to `/sites`. The same thing happens if browsing `/start/import` from `/sites`. I'm closing this PR.

PR #77415, opened 2023-05-26
Sites: use consistent import site URLs

zaguiini:

**This PR is blocked by https://github.com/Automattic/wp-calypso/pull/77414**.

## Proposed Changes

After https://github.com/Automattic/wp-calypso/pull/76697, we are using a new target for the import site flow: `/setup/import-focused`.

It makes sense to provide a consistent experience for the users, so now the CTA, both for users without a site (empty state) and users with a site to go to the same place, especially because they are using the more developer-focused flow.

The `/start/import` flow also doesn't look great. It starts asking for a domain for your new site, but we first need to check whether we can import the old site. So it makes sense to ask them for their existing site first -- the `/setup/import-focused` flow does exactly that.

## Testing Instructions

Open `/sites` with two different accounts: one with sites, and one without.

Check that the CTA for "Import a site" (user without sites) and "Import an existing site" (user with sites) points to `/setup/import-focused`.


zaguiini:
This was achieved in https://github.com/Automattic/wp-calypso/pull/77412.

PR #77434, opened 2023-05-26
Reader: Fix follow sources

eoigal:
I found a bug in Reader Follow source when logging stats/track events.

The `source` is assigned to a prop `source` but this is never set, the prop `follow_source` is.

The trouble with fixing this is that it has potential to break stats already in place. Since the prop wasn't being set, the source falls back to the path location.

This PR removes the use of `followSource` prop (that eventually gets set to `follow_source`) in most cases where we want to continue using the path location.

There are a few exceptions where I think there is value in including the `followSource` prop.

### Exceptions

The Reader Follows Manage page has multiple places where the follow button is rendered.

Screenshot 2023-05-26 at 11 58 23
Screenshot 2023-05-26 at 11 59 51

1. Followed Site - source continue to be `following_manage`
2. Following Site Recommendations - source will now be `reader-following-manage-recommendation`
3. Following Site Search - source will now be `reader-following-manage-search-result`

The Reader Suggested Follows dialog/modal will now register the source `reader-suggested-follows-dialog` rather than the path location of where the modal is rendered.

The Reader Search page also renders the follow button in a few places. We'll continue to register the source in stats as `search` except in one place; 

Screenshot 2023-05-26 at 12 03 02

* Reader site search results (sidebar) - source will now be `search-results-sites`

### Testing
I tested this by adding;
`console.log( 'compare follows', getLocation( window.location.pathname ), source );` to [recordFollow](https://github.com/Automattic/wp-calypso/blob/1c80ab46a1ca80e382ad477ea5ca27f45675db50/client/reader/stats.js#L217)
and
`console.log( 'compare unfollows', getLocation( window.location.pathname ), source );` to [recordUnfollow](https://github.com/Automattic/wp-calypso/blob/1c80ab46a1ca80e382ad477ea5ca27f45675db50/client/reader/stats.js#L215)

Confirm expected stat source is logged to stats already in use at `reader_follows` on mc stats.

### DIFF
--- a/client/blocks/reader-full-post/index.jsx
+++ b/client/blocks/reader-full-post/index.jsx
@@ -38,7 +38,6 @@ import ReaderCommentIcon from 'calypso/reader/components/icons/comment-icon';
 import ReaderMain from 'calypso/reader/components/reader-main';
 import { isDiscoverPost, isDiscoverSitePick } from 'calypso/reader/discover/helper';
 import DiscoverSiteAttribution from 'calypso/reader/discover/site-attribution';
-import { READER_FULL_POST } from 'calypso/reader/follow-sources';
 import { canBeMarkedAsSeen, getSiteName, isEligibleForUnseen } from 'calypso/reader/get-helpers';
 import readerContentWidth from 'calypso/reader/lib/content-width';
 import LikeButton from 'calypso/reader/like-button';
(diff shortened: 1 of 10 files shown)


Addison-Stavlo (approved):
This tests well for me and does as advertised.

> The trouble with fixing this is that it has potential to break stats already in place. Since the prop wasn't being set, the source falls back to the path location.  

I guess the other option is leaving stats consistent and tracking these other items with a new tracking event or something. Im unsure what makes the most sense.

PR #77437, opened 2023-05-26
Simply Pricing Page for Jetpack Cloud

mikestottuk:
This PR hides the navigation menu and login links from the pricing page via CSS rules. It also moves the Jetpack logo down a bit on mobile devices

Screenshot 2023-05-26 at 13 35 55

Mobile device (old)
Screenshot 2023-05-26 at 13 39 20

Mobile device (new)

Screenshot 2023-05-26 at 13 40 28

### DIFF
--- a/client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
+++ b/client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
@@ -1344,7 +1344,7 @@
 	}
 	//logo is really close to the top of the page on mobile
 	.jetpack-logo {
-		margin-top: 25px;
+		margin-top: 1.5rem;
 	}
 }
 
--- a/client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
+++ b/client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
@@ -21,6 +21,13 @@
 		color: #101517;
 		font-size: 1.125rem; /* stylelint-disable-line scales/font-sizes */
 		transition: background 0.1s;
+		.jetpack-logo {
+			display: inline-block;
+			text-align: center;
+		}
+		.force-hide {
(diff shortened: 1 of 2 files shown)


client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
```diff
@@ -1334,5 +1342,9 @@
 			z-index: 9999999;
 		}
 	}
+	//logo is really close to the top of the page on mobile
+	.jetpack-logo {
+		margin-top: 25px;
```
IanRamosC: I'd recommend using `rem` instead of `px` for consistency with other spacing values we have on the codebase. `1.5rem` equals to `24px` which is the closer we can get to 25px. WDYT?

IanRamosC (commented):
The change looks ok if we only want to hide it for now, and enable the navigation again in the future. I couldn't tell from the P2 comment if this is a permanent change or not. If it's permanent, we should remove the code entirely, not only hiding it via CSS.

mikestottuk:
> The change looks ok if we only want to hide it for now, and enable the navigation again in the future. I couldn't tell from the P2 comment if this is a permanent change or not. If it's permanent, we should remove the code entirely, not only hiding it via CSS.

Yes, sorry I should have been clearer in the PR - it's a test - if it doesn't increase conversions or makes them worse we'd just undo via the CSS changes which I figured would be easier

client/jetpack-cloud/sections/pricing/jpcom-masterbar/style.scss
mikestottuk: Sure, I can do that. In that style.scss there's a bunch of other `px` so either one fits but can patch since most of the other margin-tops are in rem though

IanRamosC (approved):
Code looks good and tests well! :shipit:

PR #77441, opened 2023-05-26
Plans: Fix "Suggested" label position

mmtr:
## Proposed Changes

Fixes the issue reported by @mikemayhem3030 in p1685088113454759-slack-C029GN3KD and introduced by https://github.com/Automattic/wp-calypso/pull/76497 which was causing the "Suggested" label of the filtered plans page to be wrongly positioned.

Before | After
--- | ---
Screenshot 2023-05-26 at 15 10 22 | Screenshot 2023-05-26 at 15 21 37
Screenshot 2023-05-26 at 15 25 30 | Screenshot 2023-05-26 at 15 21 43
Screenshot 2023-05-26 at 15 25 36 | Screenshot 2023-05-26 at 15 21 49
Screenshot 2023-05-26 at 15 25 42 | Screenshot 2023-05-26 at 15 21 56

## Testing Instructions

- Switch to a free site and go to `/plans`.
- Add the `?plan=` query param to test all possible plans: `personal-bundle`, `value_bundle`, `business-bundle`, `ecommerce-bundle`.
- Make sure the "Suggested" label is correctly positioned.

### DIFF
--- a/client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
+++ b/client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
@@ -1,4 +1,9 @@
-import { isBusinessPlan, isPersonalPlan, isPremiumPlan } from '@automattic/calypso-products';
+import {
+	isBusinessPlan,
+	isPersonalPlan,
+	isPremiumPlan,
+	planLevelsMatch,
+} from '@automattic/calypso-products';
 import { isNewsletterFlow, isLinkInBioFlow } from '@automattic/onboarding';
 import type { PlanProperties } from '../types';
 
@@ -14,19 +19,31 @@ interface Props {
 	visiblePlans: PlanProperties[];
 	flowName: string;
 	currentSitePlanSlug?: string;
+	selectedPlan?: string;
 }
 
-const useHighlightIndices = ( { visiblePlans, flowName, currentSitePlanSlug }: Props ) => {
+const useHighlightIndices = ( {
+	visiblePlans,
+	flowName,
+	currentSitePlanSlug,
+	selectedPlan,
+}: Props ) => {
 	return visiblePlans.reduce< number[] >( ( acc, { planName }, index ) => {
 		let isHighlight = false;
 
+		const isCurrentPlan = currentSitePlanSlug === planName;
+		const isSuggestedPlan = !! ( selectedPlan && planLevelsMatch( planName, selectedPlan ) );
+
 		if ( flowName && isNewsletterFlow( flowName ) ) {
-			isHighlight = isPersonalPlan( planName ) || currentSitePlanSlug === planName;
+			isHighlight = isPersonalPlan( planName ) || isCurrentPlan;
 		} else if ( flowName && isLinkInBioFlow( flowName ) ) {
-			isHighlight = isPremiumPlan( planName ) || currentSitePlanSlug === planName;
+			isHighlight = isPremiumPlan( planName ) || isCurrentPlan;
 		} else {
 			isHighlight =
-				isBusinessPlan( planName ) || isPremiumPlan( planName ) || currentSitePlanSlug === planName;
+				( isBusinessPlan( planName ) && ! selectedPlan ) ||
+				( isPremiumPlan( planName ) && ! selectedPlan ) ||
+				isCurrentPlan ||
+				isSuggestedPlan;
 		}
 
 		if ( isHighlight ) {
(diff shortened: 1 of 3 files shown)


client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
```diff
@@ -14,19 +19,31 @@ interface Props {
...
+	visiblePlans,
+	flowName,
+	currentSitePlanSlug,
+	selectedPlan,
+}: Props ) => {
 	return visiblePlans.reduce< number[] >( ( acc, { planName }, index ) => {
 		let isHighlight = false;
 
+		const isCurrentPlan = currentSitePlanSlug === planName;
+		const isSuggestedPlan = !! ( selectedPlan && planLevelsMatch( planName, selectedPlan ) );
```
chriskmnds, line 35: I notice that for the highlight labels, we also check whether the plan is available for purchase ([here](https://github.com/Automattic/wp-calypso/blob/9c49dc51529c46f5da45fa8596b0f16f72b72085/client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts#L25)). Do we need the same condition here, or otherwise update the other and remove? I think it's best if we keep the logic is identical across these two hooks (and sorry for that! 😬).

One way it can lead to "breakage" is if you are on a higher-tier plan and highlight a lower tier via the URL param - the borders around the badge don't get updated correctly. It's probably not the intended use, but still:

Screenshot 2023-05-26 at 4 47 36 PM

chriskmnds (commented):
Tests out! Except for one probably minor point, I commented below. :-)

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
chriskmnds: p.s. If you need to port the logic over, I fiddled with it a bit and something like this might help:

Somewhere in [highlight-adjacency-matrix.ts](https://github.com/Automattic/wp-calypso/blob/9c49dc51529c46f5da45fa8596b0f16f72b72085/client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts):

```
const isAvailableForPurchase = ( planName: string ) =>
	createSelector(
		( state ) => {
			const selectedSiteId = getSelectedSiteId( state );
			return selectedSiteId ? isPlanAvailableForPurchase( state, selectedSiteId, planName ) : false;
		},
		[ getSelectedSiteId ]
	);
```

then [inside the loop](https://github.com/Automattic/wp-calypso/blob/9c49dc51529c46f5da45fa8596b0f16f72b72085/client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts#L31):

```
const isSuggestedPlan = 
    selectedPlan && planLevelsMatch( planName, selectedPlan ) && isAvailableForPurchase;
```

chriskmnds: not sure if the suggested code makes sense actually :D

mmtr: I tried it but it doesn't work. Seems that `isAvailableForPurchase` always returns true, so I guess it's cached/memoized when it shouldn't. Ideally we should use the `isPlanAvailableForPurchase` selector within the `reduce` callback (or even better: reuse the `useHighlightLabel` hook) but it's not possible because of the hooks restrictions.

chriskmnds: Yeah, that's the problem here. We either need to violate the "rules-of-hooks" or work it out differently. 

This line was wrong, as it's not even calling the selector 😬 :

```
const isSuggestedPlan = 
    selectedPlan && planLevelsMatch( planName, selectedPlan ) && isAvailableForPurchase;
```


This works for me:

```
const useIsAvailableForPurchase = ( planName: string ) =>
	useSelector( ( state ) => {
		const selectedSiteId = getSelectedSiteId( state );
		return selectedSiteId ? isPlanAvailableForPurchase( state, selectedSiteId, planName ) : false;
	} );
```

and then:

```
		const isSuggestedPlan =
			!! ( selectedPlan && planLevelsMatch( planName, selectedPlan ) ) &&
			useIsAvailableForPurchase( planName );
```

but it violates the rule `React Hook "useIsAvailableForPurchase" cannot be called inside a callback.`

chriskmnds (approved):
Let's ship and worry about https://github.com/Automattic/wp-calypso/pull/77441/files#r1206800405 in a follow-up. 

I'll create an issue for it and put it on our agenda

PR #77477, opened 2023-05-29
Enable sharing modal on atomic

roo2:
Enables the post publish sharing modal on Atomic sites.

see https://github.com/Automattic/wp-calypso/pull/76267

Also should be released after D112153-code has enabled transferring the 'sharing_modal_dismissed' option

### Testing instructions
see: PCYsg-ly5-p2#atomic-testing
With this version of the ETK installed on your atomic site, create a test post and publish it
You should see the modal
Select the "Don't show again" checkbox.
Publish another post and you should not see the modal.


creativecoder (commented):
I'd strongly recommend prefixing this option with `wpcom_`. Otherwise it could easily conflict if another plugin author happens to pick the same option name for their modal.

roo2:
since it's already released for simple sites, I'll add a check for the current value but also add the prefix before I add it to the atomic site whitelist

creativecoder (approved):
Works as advertised, following the testing instructions on my local Atomic dev site.

creativecoder:
I'm not sure if the Blaze Dashboard failure is a problem? It doesn't look related to the change.

image

Issue #77369, opened 2023-05-25
Writing tests for the the new migration flow


ouikhuan:
Context: https://github.com/Automattic/wp-calypso/pull/77478#issuecomment-1567951576

Before we add more e2e flows to our codebase, we should first write unit and component tests to verify the methods and the associated React components work as intended.

PR #77478, opened 2023-05-29
Turn on new migration plugin / import flow feature flag on wpcalypso

ouikhuan:
Related to #77472 

## Proposed Changes

* This PR adds the new migration plugin flow feature flag to the `wpcalypso` env to prepare for the upcoming CfT.
* Based on the flow changes on the feature flag, we also needs to update our existing e2e tests.

## Testing Instructions

TeamCity test run against the live link of this PR

## Pre-merge Checklist

### DIFF
--- a/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
@@ -14,8 +14,7 @@ const selectors = {
 	// The "content only" "continue" button of '/start/from/importing/wordpress'
 	wpContentOnlyContinueButton:
 		'.content-chooser .import-layout__column:nth-child(2) > div > div:last-child button:text("Continue")',
-	wpPlansContentOnlyOptionButton:
-		'.import__import-everything--redesign > .import__footer-button-container:last-child button:text("Use the content-only import option")',
+	wpPreMigrationContentOnlyOptionButton: 'role=button[name="Use the content-only import option"]',
 
 	// ImporterDrag page
 	importerDrag: ( text: string ) => `div.importer-wrapper__${ text }`,
@@ -117,10 +116,10 @@ export class StartImportFlow {
 	}
 
 	/**
-	 * Validates that we've landed on the WordPress plans page.
+	 * Validates that we've landed on the WordPress pre-migration page.
 	 */
-	async validateWordPressPlansPage(): Promise< void > {
-		await this.page.waitForSelector( selectors.wpPlansContentOnlyOptionButton );
+	async validateWordPressPreMigrationPage(): Promise< void > {
+		await this.page.waitForSelector( selectors.wpPreMigrationContentOnlyOptionButton );
 	}
 
 	/**
@@ -138,10 +137,10 @@ export class StartImportFlow {
 	}
 
 	/**
-	 * Continue 'content only' WordPress migration on plans page.
+	 * Continue 'content only' WordPress migration on pre-migration page.
 	 */
-	async contentOnlyWordPressPlansPage(): Promise< void > {
-		await this.page.click( selectors.wpPlansContentOnlyOptionButton );
+	async contentOnlyWordPressPreMigrationPage(): Promise< void > {
+		await this.page.click( selectors.wpPreMigrationContentOnlyOptionButton );
 	}
 
 	/**
--- a/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
@@ -14,6 +14,8 @@ const selectors = {
 	// The "content only" "continue" button of '/start/from/importing/wordpress'
 	wpContentOnlyContinueButton:
 		'.content-chooser .import-layout__column:nth-child(2) > div > div:last-child button:text("Continue")',
+	wpPlansContentOnlyOptionButton:
+		'.import__import-everything--redesign > .import__footer-button-container:last-child button:text("Use the content-only import option")',
 
 	// ImporterDrag page
 	importerDrag: ( text: string ) => `div.importer-wrapper__${ text }`,
@@ -115,6 +117,13 @@ export class StartImportFlow {
 	}
 
 	/**
+	 * Validates that we've landed on the WordPress plans page.
+	 */
+	async validateWordPressPlansPage(): Promise< void > {
+		await this.page.waitForSelector( selectors.wpPlansContentOnlyOptionButton );
+	}
+
+	/**
 	 * Validates that we've landed on the importer drag page.
 	 */
 	async validateImporterDragPage( importer: string ): Promise< void > {
@@ -126,6 +135,13 @@ export class StartImportFlow {
 	 */
 	async contentOnlyWordPressPage(): Promise< void > {
 		await this.page.click( selectors.wpContentOnlyContinueButton );
+	}
+
+	/**
+	 * Continue 'content only' WordPress migration on plans page.
+	 */
+	async contentOnlyWordPressPlansPage(): Promise< void > {
+		await this.page.click( selectors.wpPlansContentOnlyOptionButton );
 	}
 
 	/**
--- a/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
+++ b/packages/calypso-e2e/src/lib/flows/start-import-flow.ts
@@ -116,13 +116,6 @@ export class StartImportFlow {
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


bogiii (approved):
Looks good to me! 👍

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
```diff
@@ -14,6 +14,8 @@ const selectors = {
 	// The "content only" "continue" button of '/start/from/importing/wordpress'
 	wpContentOnlyContinueButton:
 		'.content-chooser .import-layout__column:nth-child(2) > div > div:last-child button:text("Continue")',
+	wpPlansContentOnlyOptionButton:
+		'.import__import-everything--redesign > .import__footer-button-container:last-child button:text("Use the content-only import option")',
```
worldomonation: **comment:blocking** - in general, we try to avoid using CSS selectors like this, preferring to use [role-based selectors](https://playwright.dev/docs/release-notes#highlights-2). You can see this in action in most of the recent PRs authored by myself or others involved in the E2E maintenance effort.

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
```diff
@@ -114,6 +116,13 @@ export class StartImportFlow {
 		await this.page.waitForSelector( selectors.wpContentOnlyContinueButton );
 	}
 
+	/**
+	 * Validates that we've landed on the WordPress plans page.
+	 */
+	async validateWordPressPlansPage(): Promise< void > {
+		await this.page.waitForSelector( selectors.wpPlansContentOnlyOptionButton );
```
worldomonation: **question:non-blocking**- without knowing the screen that this is referring to, I can't say for certain but the description sounds like it's checking that the user has been redirected to the`/plans` endpoint.

If so, can I suggest usage of the `PlansPage.validateActivePlan` or `PlansPage.validateActiveTab` in order to validate that the expected action took place _without_ introducing a new single-use method?

Alternatively, checking for `page.waitForURL` in the test is always a possibility.

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
```diff
@@ -128,6 +137,13 @@ export class StartImportFlow {
 		await this.page.click( selectors.wpContentOnlyContinueButton );
 	}
 
+	/**
+	 * Continue 'content only' WordPress migration on plans page.
+	 */
+	async contentOnlyWordPressPlansPage(): Promise< void > {
+		await this.page.click( selectors.wpPlansContentOnlyOptionButton );
```
worldomonation: **question:non-blocking** - could you give a screenshot or link to the endpoint for this?

worldomonation (changes requested):
Thanks for your PR.

I left a few questions, some of them blocking. If you aren't sure what any of those mean, do let one of us know!

Just a heads up that we're actively trying to discourage usage of E2E tests as the only form of solution. It sounds like this test can easily be implemented as a component test, doubly so because this E2E spec never actually imports any site. Has the team considered using the [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) to write component tests instead?

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
ouikhuan: Thanks, I've changed it to the role-based selector now.

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
ouikhuan: Actually, it's a pre-migration page with plans information in it. I've renamed the function a little, and I hope it's clearer now.

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
ouikhuan: Please see the screenshot below, it's a pre-migration page with plans information in it. Users can click the content-only option for free import here. 
Screen Shot 2023-05-29 at 11 22 58 PM

ouikhuan:
> Just a heads up that we're actively trying to discourage usage of E2E tests as the only form of solution. It sounds like this test can easily be implemented as a component test, doubly so because this E2E spec never actually imports any site. Has the team considered using the [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) to write component tests instead?

Thanks for the suggestion! Since our importer flow has changed due to the recent change, we need to update this existing e2e flow as well so it won't break the testing for now. We're planning to [add more tests](https://github.com/Automattic/wp-calypso/issues/77369) to the new flow soon and we can implement some component tests as well.

worldomonation:
> We're planning to https://github.com/Automattic/wp-calypso/issues/77369 to the new flow soon and we can implement some component tests as well.

Just to clarify - **KitKat is discouraging the addition of new E2E specs** when such E2Es are either:
- the only method of testing a feature, or;
- the predominanet method of testing a feature.

In other words, we are trying to [push tests down the testing pyramid](p6fDka-L9-p2), meaning it is highly recommended to first write unit and component tests to verify the methods and the associated React components work as intended. 

The purpose of an E2E test is not to serve as the primary method of testing a feature, but to be a sanity check to ensure **all components** - backend, frontend, API, third-party services, literally everything - is working together as intended.

For additional reference, I wrote a [series on E2E tests](p6fDka-Ld-p2) and in it is a checklist that everyone wanting to write E2E tests should ask themselves. A new E2E test should only ever be written once all questions and tradeoffs are considered.

packages/calypso-e2e/src/lib/flows/start-import-flow.ts
```diff
@@ -114,6 +115,13 @@ export class StartImportFlow {
 		await this.page.waitForSelector( selectors.wpContentOnlyContinueButton );
 	}
 
+	/**
+	 * Validates that we've landed on the WordPress pre-migration page.
+	 */
+	async validateWordPressPreMigrationPage(): Promise< void > {
+		await this.page.waitForSelector( selectors.wpPreMigrationContentOnlyOptionButton );
+	}
```
worldomonation: **issues:blocking** - we are now recommending [developers use role-based locators](https://playwright.dev/docs/locators) unless there is no alternative. 

This is how you'd construct a locator:

```
await this.page.locator( selectors.wpPreMigrationContentOnlyOptionButton ).waitFor();
```

**suggestion:blocking** - since the `validateWordPressPreMigrationPage` and `contentOnlyWordPressPreMigrationPage` methods only differ in that one waits for a selector and one performs a click, is it really necessary to create two new methods for this? 

Since Playwright performs [automatic actionability checks prior to clicking](https://playwright.dev/docs/api/class-locator#locator-click), there is no need to do `waitForSelector` and a `click` in separate methods.

----

Taken together, the overall suggestion would be to:
1. combine the methods into one, and rename the method.
2. use a locator, not the deprecated `page.waitForSelector` method.

```suggestion
	async clickPremigrationOptionButton(): Promise< void > {
		await this.page.locator( selectors.wpPreMigrationContentOnlyOptionButton ).click();
	}
```

test/e2e/specs/onboarding/setup__importer.ts
```diff
@@ -10,6 +10,7 @@ declare const browser: Browser;
 describe( DataHelper.createSuiteTitle( 'Site Import' ), () => {
 	const credentials = SecretsManager.secrets.testAccounts.defaultUser;
 
+	let redesignImporterFeature = false;
```
worldomonation, line 13: **suggestion:blocking** - in the interest of making sure temporary workarounds are removed later, could we please get a Tampermonkey-obfuscated link to the project thread here, with comments to remove this variable, the change in Line 23 and the if/else added to Line 48-54 once this redesigned importer is made default?

The same thing goes for the other file as well. Thanks!

worldomonation (changes requested):
Some blocking changes required, plus the previous comment on the actual need for more E2Es to consider. Thanks!

ouikhuan:
> Some blocking changes required, plus the previous comment on the actual need for more E2Es to consider. Thanks!

Thanks for the detailed review and the suggestions @worldomonation ! I've updated the code based on your feedback, and also updated the tasks to include unit tests and component tests before adding more e2e tests.

worldomonation (approved):
Thanks for addressing the feedback thoroughly, I appreciate it!

worldomonation:
@ouikhuan This change has led to a huge spike in the flaky failure rate for the spec and quite disruptive to the workflow.

I'm going to strongly suggest that this PR be reverted for now to work out why the failure has spiked.
image

PR #77484, opened 2023-05-29
Fix: Replace target="__blank" with target="_blank"

tyxla:
## Proposed Changes

Fixing all `target="__blank"` (note the double underscore) to be `target="_blank"` (with a single underscore).

Came up in https://github.com/Automattic/wp-calypso/pull/77319#discussion_r1206599911

## Testing Instructions

Not needed.


arthur791004 (approved):
👍

PR #77499, opened 2023-05-29
Help Center: Refactor Zendesk Messaging hooks

klimeryk:
This is a much needed follow-up to #77024.
This re-organizes the code for Zendesk Messaging so that it can be interacted with using clearly defined hooks.

## Proposed Changes

* Split out the code that was concentrated in Help Center into easier to re-use hooks.
* Introduce a `usePresalesChat` that uses those to simplify loading presales chat.

## Testing Instructions

This a doozy. There's two main flows to test: Help Center and pre-sales chat.

Log in to the `woothemes1654197491` Zendesk instance - that's the one `development` connects to.

### Important behaviours to look out for

We need to make sure that we call the `https://public-api.wordpress.com/wpcom/v2/help/authenticate/chat?_envelope=1&type=zendesk&test_mode=true` as late as possible. Otherwise we'll run into throttling issues with Zendesk. The JWT returned by it is heavily cached for one week, but it's expensive to get it the first time nevertheless.

`https://public-api.wordpress.com/wpcom/v2/help/messaging/is-available?_envelope=1&group=wpcom_presales&environment=development` endpoint is less crucial - we cache the API calls there (they are not user-specific), so we don't have to be as careful with calls.

### Help Center
The happy path:
 - User is eligible for chat (they are on a paid plan that has live chat support - like Premium and above).
 - Verify they have Live chat option in Help Center.
 - Verify they can start a chat and that `update-user-fields` endpoint is triggered with the data to pass to the new chat.
 - If there's an active chat, going to Help Center -> Live chat should immediately open the Zendesk Widget.
 - If there's an active chat, on refresh the Zendesk Messaging launcher should pop up on its own.

#### WordPress.com
- Go to checkout (try upgrading your plan or buying a domain name).
- If you're a free user, verify that the Zendesk script is getting loaded, you're getting authenticated, and finally the Zendesk Launcher is being shown (light blue color) - if there's agent availability (check `is-available` endpoint's return).
- If you're a paid user, the Zendesk script is probably loaded already, but you might still need to get authenticated. And the launcher should still come up automatically. The launcher icon will be the deeper color than the pre-sales one.
- Once you leave checkout, the launcher should disappear - unless you have the widget open.

#### JP Presales

@TheSteveK @CodeyGuyDylan could you help me test the Jetpack-related flows? 🙏


packages/data-stores/src/help-center/actions.ts
```diff
@@ -128,17 +128,24 @@ export const setUserDeclaredSite = ( site: SiteDetails | undefined ) =>
...
 	} as const );
 
+export const resetStore = () =>
+	( {
+		type: 'HELP_CENTER_RESET_STORE',
+	} as const );
+
 export const startHelpCenterChat = function* ( site: HelpCenterSite, message: string ) {
-	yield setInitialRoute( '/inline-chat' );
+	yield setInitialRoute( '/contact-form?mode=CHAT' );
```
klimeryk, line 137: I know this is not the same as what `inline-chat` offered, but `inline-chat` has been removed, and this action is only used in the dead `client/me/help/help-contact` form (that I will be removing very soon).

klimeryk:
@worldomonation, could you help me figure out why the E2E tests are failing? 🙏 
```
locator.waitFor: Timeout 10000ms exceeded.
waiting for locator('.components-card__body > .inline-help__search') to be visible
at SupportComponent.openPopover (/home/teamcity-1/buildAgent/work/c4a9d5b38c1dacde/packages/calypso-e2e/src/lib/components/support-component.ts:35:25)
    at Object.<anonymous> (/home/teamcity-1/buildAgent/work/c4a9d5b38c1dacde/test/e2e/specs/support/support__popover.ts:31:3)
```

This looks related to my changes, but I've also seen it a bit flaky - which could _also_ be related to my changes! 😬 FWIW, I've not changed the UI of the Help Center, so it should work as previously. So it means that Help Center is not loading in that test case for some reason - I'd love to understand why and when (especially if it's signalling a regression/broken flow!).

Happy to go on a call or help in any other way to figure this out 🙇

worldomonation:
> This looks related to my changes, but I've also seen it a bit flaky - which could _also_ be related to my changes! 😬 FWIW, I've not changed the UI of the Help Center, so it should work as previously. So it means that Help Center is not loading in that test case for some reason - I'd love to understand why and when (especially if it's signalling a regression/broken flow!).

Hi! I took a look at the artifacts produced by the test run and noticed the popover isn't appearing on the screen, but the Inline Help button is active (note the highlighting).

image

I was able to reproduce it in the calypso.live branch manually by clicking immediately on the Inline Help button on the navbar as the page loaded. My guess is that the interaction by Playwright occurs while the rest of the `/home` content is being loaded and this is too fast for the Support popover or Calypso to handle.

As for a solution, I can think of a couple of ways:
- if you made changes to the loading process of the Support popover, removing those changes would likely solve the problem.
- I can investigate a better way to wait for the loading process of the `/home` dashboard so the click isn't swallowed up.

If you'd like the second option, let me know as I will mark the existing failure as a known issue and mute the test step.

P.S. I'm also thinking the Support specs should really become a React component test instead, but I don't know enough React to implement this.

TheSteveK:
> @TheSteveK @CodeyGuyDylan could you help me test the Jetpack-related flows?

It seems to work OK on Jetpack cloud, though I don't have access to the ZD sandbox.  It looks like it's still connecting to production as it brings up my chat history when I start a new chat.

Also, the Jetpack and Akismet themes have chat with authentication implemented there too, will there need to be changes to those?

klimeryk:
> It seems to work OK on Jetpack cloud, though I don't have access to the ZD sandbox. It looks like it's still connecting to production as it brings up my chat history when I start a new chat.

Thanks for testing, @TheSteveK! It's connecting to product on all environments except `development` - so if you used the `calypso.live` link, it'd indeed connect to production.

> Also, the Jetpack and Akismet themes have chat with authentication implemented there too, will there need to be changes to those?

Sorry, I'm not sure I understand - where is "there too" in this context? You mean in their respective WordPress themes directly? That should continue to work as-is, I'm only interested here in the React implementation in Calypso. Ultimately, they both do the same thing at the end - call the `loginUser` callback that Zendesk provides. It's just the road to get there that needed refactoring in Calypso.

TheSteveK:
> Sorry, I'm not sure I understand - where is "there too" in this context? You mean in their respective WordPress themes directly? That should continue to work as-is, I'm only interested here in the React implementation in Calypso. Ultimately, they both do the same thing at the end - call the `loginUser` callback that Zendesk provides. It's just the road to get there that needed refactoring in Calypso.

Yes, so where the widget is embedded in the theme, it calls the authentication API at /rest-api-plugins/endpoints/help-authentication.php  the widget does check that the user is logged in before attempting authentication (bad stuff happened otherwise). I just wasn't sure if the help-authentication api needed to be changed to not run into rate limiting issues

klimeryk:
> > Sorry, I'm not sure I understand - where is "there too" in this context? You mean in their respective WordPress themes directly? That should continue to work as-is, I'm only interested here in the React implementation in Calypso. Ultimately, they both do the same thing at the end - call the `loginUser` callback that Zendesk provides. It's just the road to get there that needed refactoring in Calypso.
> 
> Yes, so where the widget is embedded in the theme, it calls the authentication API at /rest-api-plugins/endpoints/help-authentication.php the widget does check that the user is logged in before attempting authentication (bad stuff happened otherwise). I just wasn't sure if the help-authentication api needed to be changed to not run into rate limiting issues

If it's like this right now, it's fine. We should definitely add necessary precautions on the backend side of things to avoid abuse. My main concern with this patch is to make sure that the Jetpack/Akismet parts that I modified make sense - authenticating only logged in users, etc.

klimeryk:
@worldomonation, thanks for having a look!

> I was able to reproduce it in the calypso.live branch manually by clicking immediately on the Inline Help button on the navbar as the page loaded. My guess is that the interaction by Playwright occurs while the rest of the /home content is being loaded and this is too fast for the Support popover or Calypso to handle.

Thanks for details. I tried different users (paid, free), with network throttling (3G):
 - In Firefox, it looks like the Help Center loads early and the `?` button appears, you can click it, Help Center appears - but then the rest of Calypso loads and the state is reset and Help Center is not shown anymore. But the `?` button is also not active anymore - so it does not seem the same case as in e2e tests.
 - In Chromium, I was only able to partially reproduce it - there does indeed seem to be an initial delay between clicking the `?` button and Help Center showing up, but it does show up eventually (and that's only when I force slow 3G throttling).


https://github.com/Automattic/wp-calypso/assets/3392497/530509a6-99d2-469a-a8f4-d699d19e0590

Could you confirm that it does not indeed show up for you at all? Are there any errors in the console? Any requests hanging/waiting?

FWIW, the behaviour I recorded above is the same for production and this test branch. I've not changed how Help Center is loaded - the only thing that could be related is indirectly influencing it by adding dependencies, shifting code around bundles (but nothing significant), etc.

klimeryk:
> P.S. I'm also thinking the Support specs should really become a React component test instead, but I don't know enough React to implement this.

I can try to help with this if it will help make this more stable/reliable 🙇 Also cc @Automattic/vertex who wrote the original code and have lots of experience with it (and React! 😬)

CodeyGuyDylan:
I tested out the flows that I have worked on and everything seemed to work correctly 😄

worldomonation:
>> P.S. I'm also thinking the Support specs should really become a React component test instead, but I don't know enough React to implement this.

> I can try to help with this if it will help make this more stable/reliable

These specs are indeed better off as React component tests, especially [these test steps](https://github.com/Automattic/wp-calypso/blob/trunk/test/e2e/specs/support/support__popover.ts#L39-L63) that are essentially validations around whether a search term produces the expected results. I also think this would be a great opportunity for development teams to write component tests!

I can spin up the tasks as long as I know what team to assign them to - would it be vertex?
Edit: [task](https://github.com/Automattic/wp-calypso/issues/77724).

> Could you confirm that it does not indeed show up for you at all? Are there any errors in the console? Any requests hanging/waiting?

I can't reproduce it now so it might have been a fluke. 
In any case, I will take a look. It would be necessary for any Support spec that remains after the migration to component tests to be able to toggle the popover reliably.

client/components/jetpack/portal-nav/index.tsx
```diff
@@ -53,6 +53,12 @@ export default function PortalNav( { className = '' }: Props ) {
 	const isPartnerPortalRoute = useSelector( isPartnerPortal );
 	const selectedSiteId = useSelector( getSelectedSiteId );
 	const manageSiteLink = useManageSiteLink();
+
+	usePresalesChat(
+		'jpAgency',
```
yoavf, line 58: Nitpick, but even though there's a strict enum-like type for this, it feels  _a bit_ like a magic string when just reading the code. 

Maybe a comment specifying `getGroupName()` as where it's defined?

Or, have `usePresalesChat` accept an object and then you can do something like
```js
usePresalesChat( {
	groupKey : 'jpAgency'
	enabled: hasJetpackPartnerAccess && isSectionNameEnabled( 'jetpack-cloud-partner-portal' )
})
```

Same goes for the same call in `client/my-sites/plans/jetpack-plans/product-store/index.tsx`

yoavf (approved):
The code looks good to me. Definitely an improvement over the previous version.
It would be great though, if this could be reviewed by an actual Calypso expert :)

klimeryk:
@worldomonation, after (another) rebase after the weekend, looks like the tests magically resolved themselves. I'll merge and deploy, since other blockers have been removed and this patch is a blocker for further work 🙏 But I'll definitely follow up on improvements to make sure the tests are reliable/stable.

client/components/jetpack/portal-nav/index.tsx
klimeryk: Fully agree, but it feels like a losing battle against what seems to be the convention. A compromise could be introducing these strings as named constants somewhere, so that we can take advantage of the enum behaviour without "magic" strings around. I'll keep that in mind for follow-up patches 🙇‍♂️

PR #77516, opened 2023-05-30
Fix scroll on tags page. 

roo2:
Fixes issues reported here p7DVsv-hf4-p2#comment-46057

Used a debounced scroll function so that focus doesn't prematurely cancel scroll.

I also changed to use the "center" option of the underlying scrollIntoView function to avoid the top of the element being cut off https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView. I couldn't find another way to do this with our `scrollIntoViewport` library (or `element.scrollIntoView`) because neither support taking an offset from the element. 

## Testing Instructions

go to /tags page and click on the letters to scroll down, scroll should complete every time.
When logged in, the section you scroll to should be centered vertically and not cut off


davemart-in (approved):
Tested logged out and logged in. Works as expected. Nicely done @roo2. 🚢 it!

PR #76788, opened 2023-05-10
Marketplace: Use TypeScript plugin selectors in Marketplace

robertf4:
## Proposed Changes

As a part of the work to improve the performance of the plugins selectors which began in https://github.com/Automattic/wp-calypso/pull/72363, the plugins selectors were changed to be in TypeScript. This PR switches the Marketplace code to use the TypeScript versions of the selectors in preparation for the removal of the JavaScript ones.

## Testing Instructions

Verify that you can still purchase and install plugins through the marketplace, and that the purchase flow is unaltered.

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #77517, opened 2023-05-30
Marketplace: Use TypeScript plugin selectors in Marketplace

robertf4:
## Proposed Changes

As a part of the work to improve the performance of the plugins selectors which began in https://github.com/Automattic/wp-calypso/pull/72363, the plugins selectors were changed to be in TypeScript. This PR switches the Marketplace code to use the TypeScript versions of the selectors in preparation for the removal of the JavaScript ones.

This is a second attempt at what https://github.com/Automattic/wp-calypso/pull/76788 was doing. What has changed is the return type of `getStatusForPlugin` was specified in https://github.com/Automattic/wp-calypso/pull/77575, which now properly causes a compile error when `pluginInstallStatus.error` is accessed. This access was changed to `pluginInstallStatus?.error`.

## Testing Instructions

Plugin installation:
1. Switch to an site with a plan that supports plugins.
2. Visit `/plugins/:site`.
3. Go through the flow to purchase a paid plan, and verify that it is installed and activated at the end of the flow.
4. Return to `/plugins/:site` and go through the flow to install a free plugin. There is a section called "The free essentials" which contains only free plugins. Verify that it is installed and activated at the end of the flow.

Marketplace test:
1. Visit `/marketplace/test/:site` and verify that the display is identical to what is seen in production, and that there are no console errors.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/marketplace/pages/marketplace-product-install/index.tsx
+++ b/client/my-sites/marketplace/pages/marketplace-product-install/index.tsx
@@ -459,7 +459,7 @@ const MarketplaceProductInstall = ( {
 		// Catch the rest of the error cases.
 		if (
 			pluginUploadError ||
-			pluginInstallStatus.error ||
+			pluginInstallStatus?.error ||
 			( atomicFlow && automatedTransferStatus === transferStates.FAILURE )
 		) {
 			return (
--- a/client/my-sites/marketplace/pages/marketplace-product-install/index.tsx
+++ b/client/my-sites/marketplace/pages/marketplace-product-install/index.tsx
@@ -26,8 +26,11 @@ import { getAutomatedTransferStatus } from 'calypso/state/automated-transfer/sel
 import { getPurchaseFlowState } from 'calypso/state/marketplace/purchase-flow/selectors';
 import { MARKETPLACE_ASYNC_PROCESS_STATUS } from 'calypso/state/marketplace/types';
 import { installPlugin, activatePlugin } from 'calypso/state/plugins/installed/actions';
-import { getPluginOnSite, getStatusForPlugin } from 'calypso/state/plugins/installed/selectors';
-import { isPluginActive } from 'calypso/state/plugins/installed/selectors-ts';
+import {
+	getPluginOnSite,
+	getStatusForPlugin,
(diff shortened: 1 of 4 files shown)


tyxla (approved):
Tests well and the code looks good 👍 

Perhaps we should get a second thumbs up just in case from someone from @Automattic/wpcom-marketplace?

robertf4:
I added @Automattic/bespin which I believe is the review account for marketplace.

epeicher (commented):
I have tested it, and it works as described. I have tested installing the plan, installing a free plugin and a paid plugin and it works as expected. Also I have not found any unexpected errors in the console.

Regarding the `Marketplace test` test step, I have found an inconsistency between production and this branch; the 3rd party plugins list does not return the `active: true` field. Please find below screenshots of the results:
|This branch | Production|
|-------|------|
|CleanShot 2023-06-14 at 10 44 23@2x|CleanShot 2023-06-14 at 10 44 47@2x|

I think that is related to the `getFilteredAndSortedPlugins` selector change in `marketplace-test` page. I don't think it affects production code, so I would not block this PR from progressing, but I wanted to point it out as it is indicated in the Test steps.

robertf4:
> I think that is related to the getFilteredAndSortedPlugins selector change in marketplace-test page. I don't think it affects production code, so I would not block this PR from progressing, but I wanted to point it out as it is indicated in the Test steps.

@epeicher I didn't notice that, so I'm glad you did. In this case it is expected. `getFilteredAndSortedPlugins` changed the data structure of the plugin objects so that the `active` property is no longer found on the plugin object and is only found on the site objects. This is because the plugin is only active or inactive on a site, and the plugin can be installed on multiple sites at the same time. Previously these site specific properties were just being set to whatever the value was for the last site in the list, and now they are only found on the site objects to remove that ambiguity.

epeicher:
> This is because the plugin is only active or inactive on a site, and the plugin can be installed on multiple sites at the same time. Previously these site specific properties were just being set to whatever the value was for the last site in the list, and now they are only found on the site objects to remove that ambiguity.

That makes total sense, thanks for fixing that @robertf4

PR #77519, opened 2023-05-30
New hosted site flow: only hide other plans when onboarding through /hosting

zaguiini:
Related to pet6gk-u0-p2.

## Proposed Changes

It doesn't make sense to hide all plans whenever the `/setup/new-hosted-site` is used because developers might want to create personal or premium sites for example.

Instead, let's hide all plans except Business and Commerce only when the user is coming immediately after signing up.

I thought about splitting this PR into smaller ones, but I can't see how it can be done in a non-complicated way. Let me know if you have any ideas.

## Testing Instructions

All this with the `hosting-onboarding-i2` feature flag turned on and as a developer

### Coming from /hosting

1. Browse `/start/hosting`;
2. Go through the "Continue as <your username>" step;
3. Verify that `?hosting-flow=true` was appended to `/sites`;
4. Click "Create a new site" and check that the flow has this order: options, plans (Biz and Woo), checkout.

### Not coming from /hosting

1. Browse `/sites`;
2. Click "Create a new site" and verify that `?hosting-flow=true` is not in the URL;
3. Check that the flow has this order: plans (all except Enterprise), options, checkout;
4. If picking Free, you should go to `/sites` after the progressbar is done;
5. If picking Personal or Premium, you should go to checkout and then `/sites`;
6. If picking Business or Commerce, you should see the site options, checkout, and then `/sites`.

### Not a developer

If you're not a developer, "Create new site" on `/sites` should lead you to `/start`.


mpkelly:
Tested and working. I will resolve this merge issue and then do a code review.

mpkelly (approved):
Code looks fine. Rebased and re-tested and everything still works. Will merge this.

p-jackson:
> It doesn't make sense to hide all plans whenever the `/setup/new-hosted-site` is used because developers might want to create personal or premium sites for example.

This wasn't what I was expecting from the P2 discussion pet6gk-u0-p2#comment-407. But I can see how it could have been saying this.

What I was imagining was that devs would continue to go down the regular old `/start` flow when they're not coming from hosting. That's because it's not just about offering the other plans, but also devs should have the opportunity to use the goals step, and the pattern builder, etc., just like every other use.

I don't want to block the PR from merging if that's what you were expecting. Just wanted to remind you of the other stuff in that default `/start` flow that we're also missing.

mpkelly:
> What I was imagining was that devs would continue to go down the regular old `/start` flow when they're not coming from hosting. That's because it's not just about offering the other plans, but also devs should have the opportunity to use the goals step, and the pattern builder, etc., just like every other use.

@p-jackson that's what is happening in this PR. Now the only time you see only Business & eCommerce is when you come directly from `/hosting`.

p-jackson:
Ok I just misunderstood the diff 😅 Manually testing things cleared that up

The logic for where the user goes also depends on `is_dev_account`, and if they're not a dev then they go down the `/start` flow. So we're not always skipping those `/start` steps I was worried about.

I think there's some logic to the idea that devs don't need to go through goal setting step since they're more likely to know what they're doing. On the other hand, `is_dev_account` to me doesn't necessarily mean "WordPress" expert, so things like the pattern assembler could still be good.

And I still feel like `is_dev_account=false` users should be able to use `/setup/new-hosted-site`, especially if they've just been browsing our LOHP and want to check out these cool features.

But [I was definitely wrong](https://github.com/Automattic/wp-calypso/pull/77519#issuecomment-1568006779) about how it worked. Don't let me block merging. I don't want to create an endless back and forth about how it should work when it's something we've already discussed on P2 😄

PR #77537, opened 2023-05-30
Reader: Tidy up Search sites sidebar styles

eoigal:
This PR intends to fix some style issues with the Reader search site results in the sidebar. One of the main issues is that the newest styles in Search for Popular Sites is different to the older Site results sidebar styling. This will try to make them more alike;

### Popular Sites
Screenshot 2023-05-30 at 14 30 13

### Search Sites Sidebar - Before
Screenshot 2023-05-30 at 14 24 14

1 - Updates square to circular site icons as used elsewhere on Reader
2 - Updated site title styling to match Popular Sites styling
3 - Updated site description to match Popular Sites styling

### Search Sites Sidebar - After
Screenshot 2023-05-30 at 11 48 29

### DIFF
--- a/client/blocks/reader-avatar/index.jsx
+++ b/client/blocks/reader-avatar/index.jsx
@@ -21,6 +21,7 @@ const ReaderAvatar = ( {
 	preferBlavatar = false,
 	showPlaceholder = false,
 	onClick,
+	iconSize = null,
 } ) => {
 	let fakeSite;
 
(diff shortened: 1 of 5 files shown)


davemart-in:
While we're in there, can we line clamp the title to just 2 lines? For instance, this looks a tad excessive:

CleanShot 2023-06-01 at 09 40 55@2x

You can just add text via the DOM to simulate, vs. trying to find a site with 4 lines.

Other than that, this is looking solid.

Issue #75327, opened 2023-04-05
Site Assembler - Sections Pattern Inserter, Categories, and Large Preview UX


miksansegundo:
Related [Prototype](https://www.figma.com/proto/7l53h5fxAUNjVRBQ82YFwn/Site-Assembler?page-id=2100%3A60662&node-id=2100-60914), https://github.com/Automattic/wp-calypso/issues/75521#issuecomment-1511988853

## What

- Users feel frustrated when the sidebars cover the larger view on the right
- Users must do too many clicks to add section patterns

|BEFORE|AFTER|
|----|----|
|![Image](https://user-images.githubusercontent.com/1881481/230035617-604ddf7b-3778-4c96-9d0c-c8714ce69334.png)|![Image](https://user-images.githubusercontent.com/1881481/230033449-b1f49609-dc55-4e63-a159-5780d45f8eaa.png)|

## Why

For a better UX, less clicks and easier to visualize the added patterns in the large preview

## How

- When user clicks to start the process of adding section patterns
  - The "Sections" screen won't show if there aren't patterns added. User goes directly to the "Add patterns" screen
  - The "Add patterns" screen has the first category open by default (currently this would be Text patterns).
  - The large preview will not be partially hidden by side nav because the patterns panel won't float over it

miksansegundo:
@autumnfjeld @davewhitley What are your thoughts? Let's move it to ready for development when ready.

davewhitley:
I would like to try it out and see if it's a better experience. We'd need to zoom out the large preview though I think for it to work.

lucasmendes-design:
Hi @miksansegundo, what is the latest on this? Is there anything I can help?

miksansegundo:
We could try adding a quick CSS transition. We decided to not use it on the current version because it didn't perform well.

What if we open the first category automatically when we are on this screen "Add patterns"?

autumnfjeld:
I think this issue has evolved a bit and could encompass the following criteria:

* Do not block the large preview with any sidebars.  Large preview should fit the available area to the right of any open sidebars. 
* When user clicks to start the process of adding section patterns then the categories list AND the first category should open showing the list of patterns (currently this would be Text patterns).
  * The empty list will NOT show
  * The large preview will not be hidden by side navs

If this makes sense to you @miksansegundo & @lucasmendes-design let's update the issue description to make that clear and start working on this. :) 


For reference this is what I mean by "empty list".

[...]

PR #77547, opened 2023-05-30
Site Assembler - Sections Pattern Inserter, Categories, and Large Preview UX

miksansegundo:
Related to #75327

## Proposed Changes

- The "Homepage" screen won't show if there aren't patterns added. The user goes directly to the "Add patterns" screen
- The "Add patterns" screen has the first category open by default
- Allow toggling the category to hide the patterns panel
- The large preview is fully visible and the patterns panel doesn't float over it

## Testing Instructions

- Click on the Calypso Live link in the first comment
- Create a new site with any plan
- Continue until the Design Picker
- Scroll down and click `Start designing`
- Click `Homepage` and verify
  - The  `Homepage` screen only shows when there are patterns selected
  - The `Back` and `Save` buttons go to the previous screen
- Verify that the large preview is fully visible
- Verify you can switch devices and scroll the large preview
- Verify you can click on a category to toggle the patterns panel
- Test patterns from the "Link in bio" category to test that they use the 100% viewport height 

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -10,6 +10,7 @@ import {
 } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { useDispatch, useSelect } from '@wordpress/data';
+import classnames from 'classnames';
 import { useState, useRef, useMemo } from 'react';
 import PremiumGlobalStylesUpgradeModal from 'calypso/components/premium-global-styles-upgrade-modal';
 import { createRecordTracksEvent } from 'calypso/lib/analytics/tracks';
@@ -66,6 +67,7 @@ const PatternAssembler = ( {
 	const wrapperRef = useRef< HTMLDivElement | null >( null );
 	const [ activePosition, setActivePosition ] = useState( -1 );
 	const [ surveyDismissed, setSurveyDismissed ] = useState( false );
+	const [ isPatternPanelListOpen, setIsPatternPanelListOpen ] = useState( false );
 	const { goBack, goNext, submit } = navigation;
 	const { applyThemeWithPatterns, assembleSite } = useDispatch( SITE_STORE );
 	const reduxDispatch = useReduxDispatch();
@@ -546,7 +548,13 @@ const PatternAssembler = ( {
 	}
 
 	const stepContent = (
-		<div className="pattern-assembler__wrapper" ref={ wrapperRef } tabIndex={ -1 }>
+		<div
+			className={ classnames( 'pattern-assembler__wrapper', {
+				'pattern-assembler__pattern-panel-list--is-open': isPatternPanelListOpen,
+			} ) }
+			ref={ wrapperRef }
+			tabIndex={ -1 }
+		>
 			<Notices noticeList={ noticeList } noticeOperations={ noticeOperations } />
 			<div className="pattern-assembler__sidebar">
 				<NavigatorScreen path={ NAVIGATOR_PATHS.MAIN }>
@@ -601,6 +609,7 @@ const PatternAssembler = ( {
 						selectedPattern={ sectionPosition !== null ? sections[ sectionPosition ] : null }
 						onSelect={ onSelect }
 						recordTracksEvent={ recordTracksEvent }
+						onTogglePatternPanelList={ setIsPatternPanelListOpen }
 					/>
 				</NavigatorScreen>
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -556,6 +556,7 @@ const PatternAssembler = ( {
 						recordTracksEvent={ recordTracksEvent }
 						surveyDismissed={ surveyDismissed }
 						setSurveyDismissed={ setSurveyDismissed }
+						patterns={ sections }
 					/>
 				</NavigatorScreen>
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -564,7 +564,7 @@ const PatternAssembler = ( {
 						recordTracksEvent={ recordTracksEvent }
 						surveyDismissed={ surveyDismissed }
 						setSurveyDismissed={ setSurveyDismissed }
-						hasSections={ sections.length }
+						hasSections={ !! sections.length }
 					/>
 				</NavigatorScreen>
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -564,7 +564,7 @@ const PatternAssembler = ( {
 						recordTracksEvent={ recordTracksEvent }
 						surveyDismissed={ surveyDismissed }
 						setSurveyDismissed={ setSurveyDismissed }
-						patterns={ sections }
+						hasSections={ sections.length }
 					/>
 				</NavigatorScreen>
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -10,7 +10,6 @@ import {
 } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { useDispatch, useSelect } from '@wordpress/data';
-import classnames from 'classnames';
 import { useState, useRef, useMemo } from 'react';
 import PremiumGlobalStylesUpgradeModal from 'calypso/components/premium-global-styles-upgrade-modal';
 import { createRecordTracksEvent } from 'calypso/lib/analytics/tracks';
@@ -66,7 +65,6 @@ const PatternAssembler = ( {
 	const [ sectionPosition, setSectionPosition ] = useState< number | null >( null );
 	const wrapperRef = useRef< HTMLDivElement | null >( null );
 	const [ activePosition, setActivePosition ] = useState( -1 );
-	const [ isPatternPanelListOpen, setIsPatternPanelListOpen ] = useState( false );
 	const [ surveyDismissed, setSurveyDismissed ] = useState( false );
 	const { goBack, goNext, submit } = navigation;
 	const { applyThemeWithPatterns, assembleSite } = useDispatch( SITE_STORE );
@@ -548,13 +546,7 @@ const PatternAssembler = ( {
 	}
 
 	const stepContent = (
-		<div
-			className={ classnames( 'pattern-assembler__wrapper', {
-				'pattern-assembler__pattern-panel-list--is-open': isPatternPanelListOpen,
-			} ) }
-			ref={ wrapperRef }
-			tabIndex={ -1 }
-		>
+		<div className="pattern-assembler__wrapper" ref={ wrapperRef } tabIndex={ -1 }>
 			<Notices noticeList={ noticeList } noticeOperations={ noticeOperations } />
 			<div className="pattern-assembler__sidebar">
 				<NavigatorScreen path={ NAVIGATOR_PATHS.MAIN }>
@@ -607,8 +599,6 @@ const PatternAssembler = ( {
 						replacePatternMode={ sectionPosition !== null }
 						selectedPattern={ sectionPosition !== null ? sections[ sectionPosition ] : null }
 						onSelect={ onSelect }
-						wrapperRef={ wrapperRef }
-						onTogglePatternPanelList={ setIsPatternPanelListOpen }
 						recordTracksEvent={ recordTracksEvent }
 					/>
 				</NavigatorScreen>
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -9,6 +9,15 @@
 	margin-inline-start: 10px;
 	box-sizing: border-box;
 	z-index: 1;
+	transition: margin-inline-start 0.1s ease-out;
+
+	&.device-switcher__container--frame-bordered .device-switcher__frame {
+		box-shadow: none;
+	}
+
+	.pattern-assembler__pattern-panel-list--is-open & {
+		margin-inline-start: 365px;
+	}
 }
 
 .pattern-large-preview__patterns {
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -16,20 +16,20 @@
 		position: relative;
 	}
 
-	&.device-switcher__container--is-computer .pattern-large-preview__patterns {
-		--pattern-large-preview-device-computer-scale: 0.52;
-		--pattern-large-preview-device-computer-width: 1024;
-		position: absolute;
-		left: 0;
-		top: 0;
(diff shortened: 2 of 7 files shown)


lucasmendes-design:
Great work. 

**Some notes:** 
- I'm worried about how it will look on small screens.  
- I can't deselect the category anymore. It's always selected and expanded. The video below is the comparison between this and in-production. Not sure if it was intentional but I think we could collapse the categories. It helps for small screens. WDYT? 

https://github.com/Automattic/wp-calypso/assets/98343583/ca717419-3d78-4196-9cba-94c02477252c

miksansegundo:
> I'm worried about how it will look on small screens.

Me too. I think the preview on the desktop viewport (when we scale it down), could look too small.

> I can't deselect the category anymore. It's always selected and expanded. 

This is intentional. The reasoning is that the `Add patterns` screen goal is adding patterns and previewing them (fully) in the large preview. When you are done adding patterns, you click save/back and the large preview will get expanded again for you to continue working on your homepage on a bigger canvas.

I get your point, a user on a small screen could want to make the large preview larger. As a user, what would motivate you to click on the same category to toggle? I think that I would click if I see the patterns panel floating over. 

One idea, is to make the assembler responsive to smaller screens with screen breakpoints to adjust paddings.

autumnfjeld:
Just an FYI: This is what happens in the editor when the browser window is narrow:

https://github.com/Automattic/wp-calypso/assets/5665959/4f2c4413-521e-479d-a7d0-5c49e93c41df




What was your browser viewport width in your demo @miksansegundo ? Full laptop monitor width?

autumnfjeld:
@miksansegundo @taipeicoder 
My two cents are, that if we made any design/dev effort to change the viewing/responsive/etc for "small screens" that this is its own project and  we don't move forward on it until all the work in pekYwv-1Df-p2   is done and we have a mini CfT or internal CfT on this work. Then we see if anyone comments on the small screen. Would want to know a dev effort estimate before deciding to do this work.

Because the user can close the pattern selection panel, meaning after they click save, they will see their homepage in full view.   So maybe there should be some informative copy about that? Not sure. But my point is the user has the ability to see their homepage in a "proper" width once they are out of the pattern selection state. 

@lucasmendes-design  expressed similar ideas:
> I can't deselect the category anymore. It's always selected and expanded. The video below is the comparison between this and in-production. Not sure if it was intentional but I think we could collapse the categories. It helps for small screens. WDYT?

autumnfjeld:
It might be helpful for the user to have a way to close the pattern panel.   Close the patten panel, so only the category list shows. Allowing more flexibility in viewing options while user is selecting patterns.

image

miksansegundo:
> Allowing more flexibility in viewing options while user is selecting patterns.

@autumnfjeld what do you mean by viewing options? If the pattern panel is closed, users cannot add patterns.

I think when adding patterns they should be visible. I think users will only find it helpful to close the panel to make the large preview larger when it's too small but I don't see how it gives more flexibility to add patterns. 

@lucasmendes-design happy to explore where to put a button to close the panel if you also see the need for that feature.

miksansegundo:
> What was your browser viewport width in your demo @miksansegundo ? Full laptop monitor width?

It was just a portion of my 39.5-inch panoramic screen 😅

Here is a full screenshot of a 16-inch laptop:
Screenshot 2566-05-31 at 12 04 26

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -541,6 +541,7 @@ const PatternAssembler = ( {
 						onSelect={ onMainItemSelect }
 						onContinueClick={ onContinueClick }
 						recordTracksEvent={ recordTracksEvent }
+						patterns={ sections }
```
arthur791004: [nit] Maybe `hasSections` is enough. WDYT?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
```diff
@@ -9,6 +9,15 @@
...
 	box-sizing: border-box;
 	z-index: 1;
+	transition: margin-inline-start 0.1s ease-out;
+
+	&.device-switcher__container--frame-bordered .device-switcher__frame {
+		box-shadow: none;
+	}
+
+	.pattern-assembler__pattern-panel-list--is-open & {
+		margin-inline-start: 365px;
```
arthur791004, line 20: I guess we have to increase the margin to balance the whitespace on both sides of the large preview and keep the box-shadow. WDYT?

arthur791004:
I'd also prefer to toggle the pattern panel list 😆 We can allow people to unselect the category or have a `X` button on that panel to close it.

For the small screen, I think the pattern panel list should be on the top of the sidebar instead of growing on the right side. Here is the example in the site editor, and you can see the pattern list is on the top of the category list and there is a back button to allow users to go back to category list on the small screen

image

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/screen-category-list.tsx
```diff
@@ -125,12 +108,8 @@ const ScreenCategoryList = ( {
 							aria-describedby={ description }
 							aria-current={ isOpen }
 							onClick={ () => {
-								if ( isOpen ) {
-									setSelectedCategory( null );
-									onTogglePatternPanelList?.( false );
```
arthur791004, line 130: I'd prefer to keep this feature to toggle the panel list. Let's wait for the final decision 🙂

miksansegundo: Okay. Let's keep it.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
miksansegundo: I agree about the margins. I think the shadow doesn't work well.
I did the changes following the design in Figma. Let's wait for Lucas' design review before updating that.

miksansegundo:
A new version is ready for testing.

https://github.com/Automattic/wp-calypso/assets/1881481/a82ececf-7878-4120-9fe2-691299206a8d


Thank y'all for the feedback!

lucasmendes-design:
Hi @miksansegundo, I tested and it looks good to me 😃 Thanks!

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
arthur791004: Sure 🙂

arthur791004:
> Scale to 1280px the desktop viewport

I think this behavior is a bit weird. For example, the viewport width is only fixed for the desktop device. If you switch to a tablet or mobile, the content will become responsive.

If we want to have a “fixed” viewport width for the device content, I assume we have to make the `DeviceSwithcer` component support this behavior so the viewport width will be fixed for all devices.

https://github.com/Automattic/wp-calypso/assets/13596067/52f2dbc3-199d-4ba4-b2c5-b7fc49c30078

miksansegundo:
> If we want to have a “fixed” viewport width for the device content, I assume we have to make the DeviceSwithcer component support this behavior so the viewport width will be fixed for all devices.

👍 I agree with moving the feature to `DeviceSwithcer`, not sure if we necessarily have to fix the viewport width for all devices but we can test it.

autumnfjeld:
> @autumnfjeld what do you mean by viewing options? If the pattern panel is closed, users cannot add patterns.

I meant all the different "states" the "views" the user has within the  PA. Example
* When selecting patterns, pattern panel open, the view is narrow
* When done selecting patterns, pattern panel is closed, the view is wide

Agree, when the pattern panel is closed, user can't add patterns. My point was more that the user can switch between these views, that the user does have options for a wider view, even if it is "effort" to switch between these views.

autumnfjeld:
> 👍 I agree with moving the feature to DeviceSwithcer, not sure if we necessarily have to fix the viewport width for all devices but we can test it.

@miksansegundo @lucasmendes-design   Feels like this should be separate work, if it is needed at all.  Peopel will use the width of their screens in most cases, so that shoud be the design target.   Or maybe I'm missing something of why this would be worth the dev work?

autumnfjeld:
@miksansegundo  Just tested the live link. Flow is a great improvement!!!   Nice job!  Might share your demo video in a comment of the P2, because this is such a news worthy improvement. 😄

miksansegundo:
I think the desktop viewport width should be `1024 px minimum` and `ideally 1280 px`.

A Macboox 16" screen has `1728 x 1000 pixels` and the large preview gets only `963 px` **without scaling it,** which feels "okay" but it's still too small for a desktop.

Screenshot 2566-06-06 at 12 23 25


A Macboox 13" screen has `1280 x 800 pixels` and the large preview gets only `515 px` **without scaling it,** which feels like a tablet viewport. 

Screenshot 2566-06-06 at 12 21 36

We can address the scaling on another PR but I think they should be shipped together because `963 px` or less is not enough for the desktop preview.

cc: @autumnfjeld @lucasmendes-design

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -135,10 +137,61 @@ const PatternLargePreview = ( {
 		);
 	};
 
+	const deviceComputerViewportScale = () =>
+		frameRef.current!.clientWidth / DEVICE_COMPUTER_VIEWPORT_WIDTH;
+
+	const setLargePreviewStyleForViewport = () => {
+		if ( NAVIGATOR_PATHS.SECTION_PATTERNS === navigator.location.path ) {
```
okmttdhr: badge

I think we don't need this condition as it's already in `useEffect`!

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -135,10 +137,61 @@ const PatternLargePreview = ( {
...
+	// Scale down the computer device viewport
+	useEffect( () => {
+		if ( NAVIGATOR_PATHS.SECTION_PATTERNS === navigator.location.path ) {
+			setTimeout(
+				() => {
+					updateViewportHeight();
+					setLargePreviewStyleForViewport();
+				},
+				// Wait for .pattern-large-preview transition
+				205
```
okmttdhr: badge

We might be able to create a variable for `205` to be consistent.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -135,10 +137,61 @@ const PatternLargePreview = ( {
...
+			);
+		}
+	}, [ navigator.location ] );
```
okmttdhr: badge

If we just want to execute a function in a certain condition, wouldn't it be ok to remove useEffect? (I'm sorry if it doesn't make sense here.)
https://react.dev/learn/you-might-not-need-an-effect

```suggestion
		if ( NAVIGATOR_PATHS.SECTION_PATTERNS === navigator.location.path ) {
			setTimeout(
				() => {
					updateViewportHeight();
					setLargePreviewStyleForViewport();
				},
				// Wait for .pattern-large-preview transition
				205
			);
		}
```

arthur791004:
@miksansegundo I assume we can leave the viewport issue as it is (without scaling implemented in this PR) as the user encounters the same issue now when they are on the small screen. Let's make the `DeviceSwitcher` component support the fixed viewport width in the follow-up PR and move this forward. WDYT?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
miksansegundo: I noticed that the value `navigator.location` was not updated when is out of the effect. I'll check it again.

miksansegundo:
@arthur791004 Okay, let's do that. I'll use this PR as the base for my next. Note that there are still some changes in this PR to fix issues. See my new comments on the code.

@okmttdhr Thanks for the review. I'll address the feedback in the next PR because I removed from this PR those changes.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -169,12 +169,13 @@ const PatternLargePreview = ( {
 		};
 	}, [ activePosition, header, sections, footer ] );
 
+	// Update viewport height on window resize
 	useEffect( () => {
 		const handleResize = () => updateViewportHeight();
-		window.addEventListener( 'resize', handleResize );
+		window.addEventListener( 'resize', handleResize, true );
```
miksansegundo, line 175: `true` makes it a passive event for better browser optimization.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -169,12 +169,13 @@ const PatternLargePreview = ( {
...
 
 		return () => window.removeEventListener( 'resize', handleResize );
-	} );
+	}, [] );
```
miksansegundo, line 178: Only needs to run on mount

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.tsx
```diff
@@ -193,9 +194,10 @@ const PatternLargePreview = ( {
 			frameRef={ frameRef }
 			onDeviceChange={ ( device ) => {
 				recordTracksEvent( PATTERN_ASSEMBLER_EVENTS.PREVIEW_DEVICE_CLICK, { device } );
+				setDevice( device );
 				// Wait for the animation to end in 200ms
+				// The animation is triggered by the setDevice above
 				window.setTimeout( () => {
-					setDevice( device );
```
miksansegundo, line 198: Moving `setDevice( device );` out of the timeout because the animation is triggered by it.

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-list-panel.tsx
```diff
@@ -32,7 +32,7 @@ const PatternListPanel = ( {
 	return (
 		<div key="pattern-list-panel" className="pattern-list-panel__wrapper">
 			<div className="pattern-list-panel__title">{ category?.label }</div>
-			<p className="pattern-list-panel__description">{ category?.description }</p>
+			<div className="pattern-list-panel__description">{ category?.description }</div>
```
miksansegundo, line 35: No need for a `<p>` so better use `<div>`

arthur791004 (approved):
Tested and looks good to me 👍

[nit] Some changes might not be required as it just made the order different 😅

okmttdhr (approved):
LGTM✅

Worked well on Storybook also.

https://github.com/Automattic/wp-calypso/assets/5287479/bcbd6ca4-8928-4236-91f7-312cdf632951

autumnfjeld:
@miksansegundo 
> We can address the scaling on another PR but I think they should be shipped together because 963 px or less is not enough for the desktop preview.

If the other PR is coming soon, go ahead and ship this one now. Because the improvements here are so good. Let's get these to our users with this PR. :)

autumnfjeld:
Question: would it be worthwhile for us to add a track event somewhere in onboarding , maybe when the PA screen loads that tracks viewport widith?  We could something like 
`calypso_signup_viewportwidth = ` and use settings we use for responsive breakpoints. I'd need to find the code, is categorised as S M L ?

Or we could make our own designations:

`calypso_signup_assembler_viewportwidth = LT680 | 681to768| 769to1060 | 1061to1200 | 1201to1480 | GT1481`. I am just making up numbers. You get the idea :)

arthur791004:
> Question: would it be worthwhile for us to add a track event somewhere in onboarding , maybe when the PA screen loads that tracks viewport widith?

I think it should be global property for all events as it's very useful 🙂

PR #77551, opened 2023-05-30
Add `canUseTheme` selector

mmtr:
Related to https://github.com/Automattic/dotcom-forge/issues/2438

## Proposed Changes

In order to reuse the same component for rendering the theme type badges in the design picker and in the theme showcase (https://github.com/Automattic/dotcom-forge/issues/2438) we need a single way to infer whether the site has needed plan to activate a certain theme type.

This PR introduces a new `canUseTheme` selector to centralize that logic rather than duplicating it across the codebase. The new selector will be used a by a follow-up PR which will refactor how the theme type badges are rendered.

## Testing Instructions

N/A. The new selector is still unused, but you can manually test it in this follow-up PR: https://github.com/Automattic/wp-calypso/pull/77555

### DIFF
--- a/client/state/themes/selectors/can-use-theme.js
+++ b/client/state/themes/selectors/can-use-theme.js
@@ -17,12 +17,12 @@ import { getThemeType } from 'calypso/state/themes/selectors';
 import 'calypso/state/themes/init';
 
 /**
- * Get the theme type.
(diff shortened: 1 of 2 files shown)


BogdanUngureanu (approved):
LGTM!

PR #77558, opened 2023-05-30
Checkout: Do not save VAT info at checkout when logged-out

sirbrillig:
## Proposed Changes

When using checkout, it's possible to enter VAT details. The VAT validation endpoint works with logged-out users, so this is fine. However, when validation completes, checkout attempts to _save_ the VAT details, which will never work for a logged-out user.

This PR changes the code to avoid trying to save the VAT details during checkout if the user is logged-out.

Screenshot of the error before this PR:

Screenshot 2023-05-30 at 2 23 48 PM

Screenshot after this PR:

Screenshot 2023-05-30 at 2 29 44 PM

## Testing Instructions

An automated test is included but you can test this manually as well:

- Visit checkout logged-out (eg: using an incognito window, visit `/checkout/jetpack/jetpack_security_t1_yearly`).
- In the contact step, select a country that allows VAT (eg: "United Kingdom") and a valid postal code (eg: `M1 1AE`).
- Check the box to add VAT details.
- Enter valid VAT details (eg: if you are sandboxed you can use the number `553557881` and any other info).
- Click "Continue".
- Verify that the step completes successfully.

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/wp-checkout.tsx
@@ -402,6 +402,7 @@ export default function WPCheckout( {
 							// When the contact details change, update the VAT details on the server.
 							try {
 								if (
+									! isLoggedOutCart &&
 									vatDetailsInForm.id &&
 									! areVatDetailsSame( vatDetailsInForm, vatDetailsFromServer )
 								) {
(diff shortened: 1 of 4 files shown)


jboland88 (commented):
This tested well for me and fixes the issue for entering VAT info when logged out - thanks for adding the new tests!

PR #77563, opened 2023-05-30
Jetpack Cloud: Make skip-link on pricing page scroll to main content

### LINKED ISSUE #77562: Jetpack Pricing Page: skip to main feature is broken
## What's the issue
In the Jetpack pricing page, the _Skip to main content_ feature does not work.

## How to reproduce the issue
- Visit https://cloud.jetpack.com/pricing
- Press the `Tab` key until you see the _Skip to main content_ link appear
- Press the link
- Press the `Tab` key multiple times: notice the focus keeps moving across the navigation instead of the main content of the page

## Expected behavior
Clicking _Skip to main content_ should move the focus to the main content of the page. This is broken because several elements have an `id` with the `content` value.

mattgawarecki:
This PR seeks to fix the "Skip to main content" accessibility skip-link on the pricing page, which currently doesn't work because the ID it looks for (`#content`) is not unique on the page.

Resolves #77562.

## Proposed Changes

* Rename the ID for the pricing page's main content anchor, so that browsers can scroll to it.

## Testing Instructions

1. Verify the issue: follow the instructions in #77562 to see the bug in action.
2. Visit the same pricing page using the code in this pull request (either via Calypso Live or in a local testing environment).
3. Follow the same instructions as in step 1. The page should correctly scroll to the main content of the page, and pressing the Tab key should start its navigation sequence at the top of the main content.

### DIFF
--- a/client/jetpack-cloud/sections/pricing/jpcom-masterbar/index.tsx
+++ b/client/jetpack-cloud/sections/pricing/jpcom-masterbar/index.tsx
@@ -24,7 +24,7 @@ import useUserMenu from './use-user-menu';
 
 import './style.scss';
 
-export const MAIN_CONTENT_ID = 'content';
+export const MAIN_CONTENT_ID = 'pricing-content';
 
 const JETPACK_COM_BASE_URL = 'https://jetpack.com';
 


yashwin (approved):
LGTM. Works as expected.

PR #77572, opened 2023-05-30
Newsletters: Add skip button to subscribers form

### LINKED ISSUE #77247: Newsletter onboarding: add separate "skip" to subscriber step
Instead of combined skip/continue CTA, add a separate skip below the main CTA.

Image

Mindful that the same form is inside product (Users->Subscribers->Add) and we don't want the skip there.

FYI @crisbusquets

edanzer:
Resolves https://github.com/Automattic/wp-calypso/issues/77247

## Proposed Changes

* Adds an optional skip link to the AddSubcribersForm component, via a prop and conditional mark up. 
* Display the new skip link on the Subscribers screen during the Newsletter onboarding flow.
* Disable the Add and Continue button until subscribers have been added (now that skip link is separate).
* These changes also revealed a small issue where AddSubcribersForm component was invoking a method before initialization when given a specific set of prop values.  

## Testing Instructions

1) Check out this branch and run yarn and yarn start if needed. 
2) Start a newsletter site at http://calypso.localhost:3000/setup/newsletter/intro and proceed to Subscribers screen.
3) Confirm the main button is disabled. Add a valid email to one field, and confirm the button is now enabled. 
4) Confirm the Skip for now link shows as expected and, when clicks, takes you to Launchpad.
5) Test to confirm no regressive breakage in other places where the AddSubscriberForm is used. You can test this by going to http://calypso.localhost:3000/people/add-subscribers/YOURDOMAIN. You should see the Add Subscribers form on this page, but the Skip for now link should not show.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
@@ -32,7 +32,7 @@ const Subscribers: Step = function ( { navigation } ) {
 
 	const submitButtonText = isImportValid
 		? translate( 'Add and continue' )
-		: translate( 'Skip for now' );
+		: translate( 'Add subscribers' );
 
 	return (
 		<StepContainer
@@ -51,14 +51,17 @@ const Subscribers: Step = function ( { navigation } ) {
 							flowName="onboarding_subscribers"
 							submitBtnName={ submitButtonText }
 							onImportFinished={ handleSubmit }
+							onSkipBtnClick={ handleSubmit }
 							onChangeIsImportValid={ ( isValid ) => setIsImportValid( isValid ) }
-							allowEmptyFormSubmit={ true }
+							allowEmptyFormSubmit={ false }
 							manualListEmailInviting={ ! isUserEligibleForSubscriberImporter }
 							showCsvUpload={ isEnabled( 'subscriber-csv-upload' ) }
 							recordTracksEvent={ recordTracksEvent }
 							titleText={ translate( 'Ready to add your first subscribers?' ) }
 							subtitleText={ subtitleText }
 							showSubtitle={ true }
+							showSkipLink={ true }
+							submitBtnAlwaysEnable={ false }
 						/>
 					) }
 				</div>
--- a/packages/subscriber/src/components/add-form/index.tsx
+++ b/packages/subscriber/src/components/add-form/index.tsx
@@ -69,6 +69,7 @@ export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
 		onChangeIsImportValid,
 		titleText,
 		subtitleText,
+		showSkipLink,
 	} = props;
 
 	const {
@@ -510,7 +511,7 @@ export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
 					>
 						{ submitBtnName }
 					</NextButton>
-					{ props.showSkipLink && (
+					{ showSkipLink && (
 						<div className="add-subscriber__form-skip-link-wrapper">
 							<button className="add-subscriber__form-skip-link" onClick={ props.onSkipBtnClick }>
 								{ translate( 'Skip for now' ) }
--- a/packages/subscriber/src/components/add-form/index.tsx
+++ b/packages/subscriber/src/components/add-form/index.tsx
@@ -43,6 +43,7 @@ interface Props {
 	onChangeIsImportValid?: ( isValid: boolean ) => void;
 	titleText?: string;
 	subtitleText?: string;
+	showSkipLink?: boolean;
 }
 
 export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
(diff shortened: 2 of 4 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
```diff
@@ -51,14 +51,17 @@ const Subscribers: Step = function ( { navigation } ) {
 							flowName="onboarding_subscribers"
 							submitBtnName={ submitButtonText }
 							onImportFinished={ handleSubmit }
+							onSkipBtnClick={ handleSubmit }
 							onChangeIsImportValid={ ( isValid ) => setIsImportValid( isValid ) }
-							allowEmptyFormSubmit={ true }
+							allowEmptyFormSubmit={ false }
```
simison, line 56: Is this feature in the form component needed at all anymore?

Alternatively you could allow empty submits, if folks submit empty form you could just continue as if they pressed skip. No strong opinion way or the other.

simison:
Added Caribou team for optional review, but ok to ignore if folks are busy and we can review just within our team. 👋

packages/subscriber/src/components/add-form/style.scss
```diff
@@ -55,6 +55,17 @@
 	.components-notice.is-info {
 		background: var(--studio-gray-0);
 	}
+
+	.add-subscriber__form-skip-link-wrapper {
+		display: flex;
+		justify-content: flex-start;
+		padding: 16px 0;
```
bogiii: Please use rem unit instead of pixel. `16px` => `1rem`

packages/subscriber/src/components/add-form/index.tsx
```diff
@@ -503,6 +510,13 @@ export const AddSubscriberForm: FunctionComponent< Props > = ( props ) => {
 					>
 						{ submitBtnName }
 					</NextButton>
+					{ props.showSkipLink && (
+						<div className="add-subscriber__form-skip-link-wrapper">
+							<button className="add-subscriber__form-skip-link" onClick={ props.onSkipBtnClick }>
```
bogiii: To be consistent, please take props constants at the beginning of the components and not access them directly from props. Everything is okay with this approach; I just want to follow the same style.

See: https://github.com/Automattic/wp-calypso/pull/77572/files#diff-313c3e20cc7156984e3eca7448261037ca4d5e5199f089e22c9f4ba2dba0ec05R55-R72

bogiii (approved):
I tested the Users management side, and there is no regression. All looks good!

I left some comments for the minor changes.

simison (approved):
Tested 
- skipping, 
- adding subscriber, 
- coming back from launchpad and skipping, 
- and coming back from launchpad and adding subscriber.

Code is a bit more than I expected; I expected just duplicated CTA with different copy/styling, but I see the complications come from using the form in couple different places. All good!

packages/subscriber/src/components/add-form/index.tsx
edanzer: Sure @bogiii - happy to update that!

client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
edanzer: The reason I'm setting this to false is that it is needed to disable the button based on the logic of the component. 

We could leave the submit button enabled and I can skip changing this prop. 

Or we could alter the component and be sure other places that use the component/prop work as affected. 

As it was, I was already making more changes than expected and was just trying to minimize changes to only what was needed for the PR and to work with the component as written. 

But I could could go different ways on this depend on what we'd like to see. I don't have strong feelings.

packages/subscriber/src/components/add-form/index.tsx
edanzer: Updated. And good catch. It's odd that I didn't do that initially. I think I'm used to seeing props destructured as part of the component definition for functional components, rather than within the component.

packages/subscriber/src/components/add-form/style.scss
edanzer: Updated. And thanks for taking the time to review @bogiii !

client/landing/stepper/declarative-flow/internals/steps-repository/subscribers/index.tsx
edanzer: My sense on this is I'd probably not update the component itself for this PR (ie, removing that prop and adjusting other uses of the component if needed). But we could just not use the prop here, and allow users to click either the 'Add and Continue' button or the skip link. But I'd probably keep this as is - if users can click the Add button to continue, it seems to negate the need for the 'skip' link in the first place. I think that's basically the logic behind how this was originally implemented.

simison: Yep, fine as-is! thanks for elaborating

PR #77575, opened 2023-05-30
Plugins: Specify selector types when TypeScript cannot infer

robertf4:
## Proposed Changes

This PR addresses a few issues noted in this recent incident report: p58i-eyE-p2

* The types of the selectors are specified when TypeScript cannot infer them. This would have avoided the fatal error which led to the incident report.
* An error in `getPluginOnSite` where the `sites` property could be accessed before checking for its existence is resolved.

## Testing Instructions

Use the TypeScript tools in your favorite IDE to ensure that all selectors have return types which are either correctly specified or inferred. (`requestPluginsError` is an exception. It is used only as a truthy/falsey but I suspect it is actually an object and I do not know its type.)

Verify that the unit tests pass and Calypso has no type errors. There is no functional testing to do as the altered functions are not in use yet.

## Pre-merge Checklist

### DIFF
--- a/client/state/plugins/installed/selectors-ts.ts
+++ b/client/state/plugins/installed/selectors-ts.ts
@@ -11,9 +11,11 @@ import type {
 	InstalledPluginData,
 	Plugin,
 	PluginFilter,
+	PluginSite,
 	PluginSites,
 	PluginStatus,
 } from './types';
+import type { SiteDetails } from '@automattic/data-stores';
 import type { AppState } from 'calypso/types';
 
 import 'calypso/state/plugins/init';
@@ -34,7 +36,7 @@ export function isEqualSlugOrId( pluginSlug: string, plugin: Plugin ) {
 	return plugin.slug === pluginSlug || plugin?.id?.split( '/' ).shift() === pluginSlug;
 }
 
-export function isRequesting( state: AppState, siteId: number ) {
+export function isRequesting( state: AppState, siteId: number ): boolean {
 	if ( typeof state.plugins.installed.isRequesting[ siteId ] === 'undefined' ) {
 		return false;
 	}
@@ -46,7 +48,7 @@ export function isRequestingForSites( state: AppState, sites: number[] ) {
 	return sites.some( ( siteId ) => isRequesting( state, siteId ) );
 }
 
-export function isRequestingForAllSites( state: AppState ) {
+export function isRequestingForAllSites( state: AppState ): boolean {
 	return state.plugins.installed.isRequestingAll;
 }
 
@@ -59,7 +61,7 @@ const getSiteIdsThatHavePlugins = createSelector(
 		return Object.keys( state.plugins.installed.plugins ).map( ( siteId ) => Number( siteId ) );
 	},
 	( state: AppState ) => [ state.plugins.installed.plugins ]
-);
+) as ( state: AppState ) => number[];
 
 /**
  * The server returns plugins store at state.plugins.installed.plugins are indexed by site, which means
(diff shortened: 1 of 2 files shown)


client/state/plugins/installed/selectors-ts.ts
```diff
@@ -280,7 +282,7 @@ export const getPluginOnSite = createSelector(
 		};
 	},
 	( state: AppState ) => [ getAllPluginsIndexedByPluginSlug( state ) ]
-);
+) as ( state: AppState, siteId: number, pluginSlug: string ) => Plugin & PluginSite;
```
robertf4, line 284: I left this return type as `Plugin & PluginSite` instead of declaring that a different type as I don't think this pattern is ideal, so I think it's best if this type isn't named. It should be clear that these functions are returning a combination of two types.

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -351,9 +353,13 @@ export const getSiteObjectsWithoutPlugin = createSelector(
...
 	( state: AppState, siteIds: number[], pluginSlug: string ) => [ pluginSlug, ...siteIds ].join()
-);
+) as ( state: AppState, siteIds: number[], pluginSlug: string ) => SiteDetails[];
 
-export function getStatusForPlugin( state: AppState, siteId: number, pluginId: string ) {
+export function getStatusForPlugin(
+	state: AppState,
+	siteId: number,
+	pluginId: string
+): { [ key: string ]: string } | undefined {
```
robertf4: Although this used to return `false` and this change caused the error, I still think it is best to return `undefined` here. The function name suggests it returns the plugin status, not a boolean which indicates if the status is undefined.

A search of this code shows that this selector is only used in two components and both are TypeScript, so the explicit typing here will prevent the previous error from occurring again.

robertf4:
I need to check something before this gets reviewed so I moved it back to In Progress.

tyxla:
> I need to check something before this gets reviewed so I moved it back to In Progress.

Sounds good, feel free to re-request review when ready.

robertf4:
@tyxla it's good for review now!

client/state/plugins/installed/selectors-ts.ts
```diff
@@ -353,7 +354,11 @@ export const getSiteObjectsWithoutPlugin = createSelector(
 	( state: AppState, siteIds: number[], pluginSlug: string ) => [ pluginSlug, ...siteIds ].join()
 );
 
-export function getStatusForPlugin( state: AppState, siteId: number, pluginId: string ) {
+export function getStatusForPlugin(
+	state: AppState,
+	siteId: number,
+	pluginId: string
+): { [ key: string ]: string } | undefined {
```
tyxla: Should this function return type be `PluginStatus | undefined` instead?

client/state/plugins/installed/types.ts
```diff
@@ -36,13 +36,15 @@ export type Plugin = {
 	wporg?: boolean;
 };
 
+export type PluginSite = {
+	active: boolean;
+	autoupdate: boolean;
+	update?: PluginUpdate;
+	version: string;
+};
+
```
tyxla, line 41: I find the overlap with the `InstalledPluginData` type a bit awkward. Can we do anything to dedupe it?

client/state/plugins/installed/selectors-ts.ts
robertf4: Yes, it can be, I just didn't have an `error` prop on the `PluginStatus` so I missed that it is the same object.

client/state/plugins/installed/types.ts
robertf4: I changed it to use this to construct `InstalledPluginData`.

tyxla (approved):
Looks good to me, thanks 👍 🚀

PR #77495, opened 2023-05-29
Importer flow: Mark find_available_url to true if from migration flow

ouikhuan:
Related to #77473 

## Proposed Changes

* Make sure the site creation step on the migration plugin flow always finds available URL.

## Testing Instructions

* Start with `http://calypso.localhost:3000/setup/newsletter/intro`, go through steps and setup a newsletter
* Prepare a JN site with Move to WordPress.com plugin and click `Get started` CTA
* Click `create a new one` button on sitePicker step
* Make sure a new site is created

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #77588, opened 2023-05-31
Hosting onboarding: allow importing a site without specifying a target WPCOM site

mpkelly:
Closes https://github.com/Automattic/dotcom-forge/issues/2585.

## Proposed Changes

Let's add a new `/setup/import-hosted-site` so that developers can import sites without a target site already established in WPCOM.

[We're not adding the data center picker and plugins/themes description right now](https://github.com/Automattic/wp-calypso/pull/77588#issuecomment-1573842072).

## Testing Instructions

1. Enable `hosting-onboarding-i2`;
2. Make yourself a developer by toggling "I am a developer" in `/me`;
3. Browse Sites > Import existing site: it should land you on `/setup/import-hosted-site`;
4. Pass a Jurassic Ninja site.

The import flow should work as normal. You should be asked to upgrade near the end. Upgrade and pay for Business with free credits.

The site on WPCOM should have its URL derived from the source site's title;

You should see a CTA to go to the site's home.

### DIFF
--- a/client/landing/stepper/declarative-flow/import-hosted-site.ts
+++ b/client/landing/stepper/declarative-flow/import-hosted-site.ts
@@ -238,6 +238,18 @@ const importHostedSiteFlow: Flow = {
 
 		const goBack = () => {
 			switch ( _currentStep ) {
+				case 'importList':
+					// eslint-disable-next-line no-case-declarations
+					const backToStep = urlQueryParams.get( 'backToStep' );
+
+					if ( backToStep ) {
+						const path = `${ backToStep }?siteSlug=${ siteSlugParam }`;
+
+						return navigate( path );
+					}
+
+					return navigate( 'import' );
+
 				case 'importReady':
 				case 'importReadyNot':
 				case 'importReadyWpcom':
--- a/client/landing/stepper/declarative-flow/import-hosted-site.ts
+++ b/client/landing/stepper/declarative-flow/import-hosted-site.ts
@@ -127,18 +127,6 @@ const importHostedSiteFlow: Flow = {
 
 		const goBack = () => {
 			switch ( _currentStep ) {
-				case 'importList':
-					// eslint-disable-next-line no-case-declarations
-					const backToStep = urlQueryParams.get( 'backToStep' );
-
-					if ( backToStep ) {
-						const path = `${ backToStep }?siteSlug=${ siteSlugParam }`;
-
-						return navigate( path );
-					}
-
-					return navigate( 'import' );
-
 				case 'importReady':
 				case 'importReadyNot':
 				case 'importReadyWpcom':
--- a/client/landing/stepper/declarative-flow/import-hosted-site.ts
+++ b/client/landing/stepper/declarative-flow/import-hosted-site.ts
@@ -132,6 +132,9 @@ const importHostedSiteFlow: Flow = {
 
 		const goBack = () => {
 			switch ( _currentStep ) {
+				case 'import':
+					return window.location.assign( '/sites' );
+
 				case 'importReady':
 				case 'importReadyNot':
 				case 'importReadyWpcom':
--- a/client/landing/stepper/declarative-flow/import-hosted-site.ts
+++ b/client/landing/stepper/declarative-flow/import-hosted-site.ts
@@ -1,38 +1,24 @@
 import { isEnabled } from '@automattic/calypso-config';
-import { Design, isBlankCanvasDesign } from '@automattic/design-picker';
+import { isBlankCanvasDesign } from '@automattic/design-picker';
 import { IMPORT_HOSTED_SITE_FLOW } from '@automattic/onboarding';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { ImporterMainPlatform } from 'calypso/blocks/import/types';
-import useAddTempSiteToSourceOptionMutation from 'calypso/data/site-migration/use-add-temp-site-mutation';
-import { useSourceMigrationStatusQuery } from 'calypso/data/site-migration/use-source-migration-status-query';
-import { useSiteExcerptsQuery } from 'calypso/data/sites/use-site-excerpts-query';
-import { SITE_PICKER_FILTER_CONFIG } from 'calypso/landing/stepper/constants';
 import MigrationError from 'calypso/landing/stepper/declarative-flow/internals/steps-repository/migration-error';
 import { ProcessingResult } from 'calypso/landing/stepper/declarative-flow/internals/steps-repository/processing-step/constants';
 import { useQuery } from 'calypso/landing/stepper/hooks/use-query';
 import { useSiteSlugParam } from 'calypso/landing/stepper/hooks/use-site-slug-param';
 import { ONBOARD_STORE } from 'calypso/landing/stepper/stores';
 import { useSiteSetupFlowProgress } from '../hooks/use-site-setup-flow-progress';
-import DesignSetup from './internals/steps-repository/design-setup';
 import ImportList from './internals/steps-repository/import-list';
 import ImportReady from './internals/steps-repository/import-ready';
 import ImportReadyNot from './internals/steps-repository/import-ready-not';
 import ImportReadyPreview from './internals/steps-repository/import-ready-preview';
 import ImportReadyWpcom from './internals/steps-repository/import-ready-wpcom';
 import ImportWithSiteAddressStep from './internals/steps-repository/import-with-site-address';
-import ImporterBlogger from './internals/steps-repository/importer-blogger';
-import ImporterMedium from './internals/steps-repository/importer-medium';
-import ImporterSquarespace from './internals/steps-repository/importer-squarespace';
-import ImporterWix from './internals/steps-repository/importer-wix';
 import ImporterWordpress from './internals/steps-repository/importer-wordpress';
-import MigrationHandler from './internals/steps-repository/migration-handler';
-import PatternAssembler from './internals/steps-repository/pattern-assembler';
 import ProcessingStep from './internals/steps-repository/processing-step';
-import SiteCreationStep from './internals/steps-repository/site-creation-step';
-import SitePickerStep from './internals/steps-repository/site-picker';
 import { Flow, ProvidedDependencies } from './internals/types';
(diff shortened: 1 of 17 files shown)


client/landing/stepper/hooks/use-get-site-suggestions-query.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const useGetSiteSuggestionsQuery = ( {
+	enabled,
+	onSuccess,
+}: {
+	enabled: boolean;
+	onSuccess?: ( response: SuggestionsResponse ) => void;
+} ) =>
+	useQuery( {
+		queryKey: [ 'site-suggestions' ],
```
p-jackson: I see that the new `useGetSiteSuggestionsQuery` hook is just a refactor so it has the same behaviour as before.

I'm wondering how each instance of the hook could be made independent? Now that it's shared, if `useGetSiteSuggestionsQuery` was used in two different components, would we want them to share the same suggestion? I feel we'd want them each to have unique suggestions.

One way might to give each instance it's own unique query key? i.e.

```ts
import { useInstanceId } from '@wordpress/components';

// ...

const instanceId = useInstanceId();

useQuery( {
	queryKey: [ 'site-suggestions', instanceId ],
	// ...
} );
```

Maybe we could also return the result of `useQuery` from the custom hook. Just for completeness.

client/landing/stepper/declarative-flow/internals/steps-repository/import-ready-preview/index.tsx
```diff
@@ -42,14 +45,35 @@ const ImportReadyPreview: Step = function ImportStep( props ) {
...
+
+			if ( ! blogName ) {
+				const response = await getSiteSuggestions();
+
+				if ( response.success ) {
+					blogName = response.suggestions[ 0 ].title;
+				}
+			}
+
+			await createSite( {
```
p-jackson: Interesting, I didn't think about how you could create the site here. Usually we'd create the site by navigating to the `siteCreationStep` step. By navigating to that step it ensures [the store is aware of the pending action](https://github.com/Automattic/wp-calypso/blob/304627d2290055a366905054b93c4815c1184894/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx#L176) (for the progress bar) and that all the correct props are passed to the call to create the site ([e.g. flow name, site visibility, etc.](https://github.com/Automattic/wp-calypso/blob/304627d2290055a366905054b93c4815c1184894/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx#L137-L149))

We don't _have_ to reuse the existing `siteCreationStep`. It's used by a number of different flows and has logic for Woo and all sorts of stuff now. It'll probably work our of the box for us though since the `new-hosted-site-options` step is already saves things in the correct place in the store for them to be picked up by the `siteCreationStep` step.

client/landing/stepper/declarative-flow/import-hosted-site.ts
```diff
@@ -0,0 +1,291 @@
...
+			return navigate( 'processing' );
+		};
+
+		const handleMigrationRedirects = ( providedDependencies: ProvidedDependencies = {} ) => {
+			const userHasSite = sites && sites.length > 0;
+
+			if ( providedDependencies?.status === 'inactive' ) {
+				// This means they haven't kick off the migration before, so we send them to select/create a new site
+				if ( ! providedDependencies?.targetBlogId ) {
+					return userHasSite ? navigate( 'sitePicker' ) : navigate( 'siteCreationStep' );
```
p-jackson: It looks like once upon the `import-focused` flow used the `siteCreationStep` step too, since this logic has been copied over into our new flow. This is how you'd navigate to the step: the logic in `submit` would determine that a site needs to be created and we'd call `navigate( 'siteCreationStep' )`.

p-jackson (commented):
I know it's a work in progress, I just thought I'd take a look at how it's going.

If you want to take a look at a shorter flow which needs to to multiple async operations one after another (like in our case create a site before starting the existing import process) then you can look at [the `'wooexpress'` flow](https://github.com/Automattic/wp-calypso/blob/304627d2290055a366905054b93c4815c1184894/client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts#L20).

client/landing/stepper/declarative-flow/import-hosted-site.ts
mpkelly: Thanks for the feedback, @p-jackson. I wonder if adding an explicit step for site creation using the options component is a better solution. That solves the problem of adding the data centre picker too. I based my solution on @zaguiini's POC, but maybe I have taken it too literally. 

There is a sporadic error when navigating from `importReadyPreview` to `importerWordpress`, which I think is related to us creating the site on the fly. You temporarily see "Uh oh. Page not found", which I believe is due to the site not being ready. So that could be another thing we'd fix by adding an explicit step.

zaguiini:
@mpkelly let's go to `/sites` only if this is the first site created by the user. You can understand more about this decision [here](https://github.com/Automattic/wp-calypso/pull/77678). There's also the implementation and you could check if you're able to replicate that in this new flow.

zaguiini:
I tried putting the data center picker and the transferred themes and plugins within the `importerWordpress` step. This is how it looks:

image

It's obviously not good, and I doubt it has any value. Imported sites have been transferred to Atomic since ever and there is not much point in showing the themes and plugins. Let's revisit it in the future and go with what we have right now.

zaguiini:
Merged, but we'll need a follow-up to address this scenario: https://github.com/Automattic/dotcom-forge/issues/2640

ouikhuan:
Sometimes site creation will fail because there might be domain items metadata when you try to create a site. See this line here: https://github.com/Automattic/wp-calypso/blob/trunk/packages/onboarding/src/cart/index.tsx#L110

When there's a domainItem here the `find_available_url` will be marked as false, and you'll get a Blogname already exists error. We faced the same issue previously and to be honest, we're not sure if it's a bug from other flows or if it's on purpose. It's also difficult to reproduce it. Eventually, we find if you try to setup a newsletter, you'll have domainItem stored. To see how to reproduce this and the fix, you can see the PR here: https://github.com/Automattic/wp-calypso/pull/77495


https://github.com/Automattic/wp-calypso/assets/4074459/92ae3869-670f-43a9-81c3-a38a0f5f43af



Also, when you click the go back button on the page, you'll miss the `source` param so it thinks you are in the normal import flow. If you try to type in the site address again, it'll be blocked as it does for normal flow.


https://github.com/Automattic/wp-calypso/assets/4074459/6f21b37a-92d7-4eab-b90f-90bf77e7de99

I realized there's [another PR ](https://github.com/Automattic/wp-calypso/pull/77414)handling go back issue, if that PR will be ready soon then please ignore the feedback above about the goBack button, thanks!

client/landing/stepper/declarative-flow/internals/steps-repository/import-with-site-address/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+					goToStep={ ( step, section ) =>
+						navigation.goToStep?.( generateStepPath( step, section ) )
+					}
+				/>
+			) }
+		</ImportWrapper>
+	);
+};
+
+export default ImportWithSiteAddressStep;
```
bogiii, line 37: Since here we introduced a new Step whose purpose is entering the source site address; I'm wondering what the difference is between this step and the step we already have here: https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/import/index.tsx

It seems like exactly the same steps. Isn't it?

client/landing/stepper/declarative-flow/internals/steps-repository/import/style.scss
```diff
@@ -11,3 +11,16 @@
 		margin: auto;
 	}
 }
+
+@media ( max-width: 1000px ) {
```
bogiii, line 15: We should use the defined breakpoints mixins, such as `@include break-xlarage {}`.

There are constants:
`break-xlarge` => 1080px
`break-large` => 960px

Why `1000px`, is this something from design?

mpkelly: It can be changed to `break-xlarge`. I will add a PR for that.

client/landing/stepper/declarative-flow/internals/steps-repository/import-with-site-address/index.tsx
mpkelly: The extra component was added to pass in `disableImportListStep` as we're not supporting the `ImportList` step in this iteration. 

I couldn't see a way to override this. Ideally this step would have taken `props`:

```ts
	useSteps() {
		return [
			{ slug: 'import', component: ImportWithSiteAddressStep }, //pass props here with component name
			
			```

Comment (bogiii) [reply client/landing/stepper/declarative-flow/internals/steps-repository/import-with-site-address/index.tsx:37]: I think that we should avoid making another import/capture step doing the same stuff.

In the regular `import` step [where we set up `CaptureStep`](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/import/index.tsx#L67) you can provide additional props based on the flow that consumes the step at the moment.

```
const { navigation, flow } = props;

<CaptureStep
	disableImportListStep={ IMPORT_HOSTED_SITE_FLOW === flow } // <------- solution
	goToStep={ ( step, section ) =>
		navigation.goToStep?.( generateStepPath( step, section ) )
	}
/>
```

@mpkelly what do you think?

mpkelly: Ok, let's go with that.

client/landing/stepper/declarative-flow/internals/steps-repository/import/style.scss
mpkelly: There is a PR here. I didn't need these `@media` rules and was able to remove some more CSS. There is a `global.css` which does the styling for these flows, and I just need to register a class there. 

https://github.com/Automattic/wp-calypso/pull/77899

PR #77214, opened 2023-05-22
ESLint: Enable `@tanstack/eslint-plugin-query`

tyxla:
## Proposed Changes

Enable the `@tanstack/eslint-plugin-query` ESLint plugin with its recommended rules but make them non-blocking warnings for now.

## Testing Instructions

Run ESLint locally and verify you're seeing **warnings** related to these rules:

- `@tanstack/query/prefer-query-object-syntax`
- `@tanstack/query/exhaustive-deps`

PR #77598, opened 2023-05-31
media: fix `@tanstack/query/prefer-query-object-syntax`

flootr:
Related to #77214 

## Proposed Changes

PR fixes `@tanstack/query/prefer-query-object-syntax` violations in `client/data/media`.

It's worth noting that the object syntax will be the only available option in a future major version (see [docs](https://tanstack.com/query/v4/docs/react/eslint/prefer-query-object-syntax)).

## Testing Instructions

* Verify the following actions related to media files are still working:
  * update meta information like title, caption, alt text, etc
  * edit an image (flip, rotate, etc)
  * delete a media file
* Verify checks are green

### DIFF
--- a/client/data/media/use-delete-media-mutation.js
+++ b/client/data/media/use-delete-media-mutation.js
@@ -2,8 +2,9 @@ import { useMutation } from '@tanstack/react-query';
 import wp from 'calypso/lib/wp';
 
 export const useDeleteMediaMutation = ( queryOptions = {} ) => {
-	return useMutation(
-		( { siteId, mediaId } ) => wp.req.post( `/sites/${ siteId }/media/${ mediaId }/delete` ),
-		queryOptions
-	);
+	return useMutation( {
+		mutationFn: ( { siteId, mediaId } ) =>
+			wp.req.post( `/sites/${ siteId }/media/${ mediaId }/delete` ),
+		...queryOptions,
+	} );
 };
--- a/client/data/media/use-delete-media-mutation.js
+++ b/client/data/media/use-delete-media-mutation.js
@@ -3,8 +3,8 @@ import wp from 'calypso/lib/wp';
 
 export const useDeleteMediaMutation = ( queryOptions = {} ) => {
 	return useMutation( {
+		...queryOptions,
 		mutationFn: ( { siteId, mediaId } ) =>
 			wp.req.post( `/sites/${ siteId }/media/${ mediaId }/delete` ),
-		...queryOptions,
 	} );
 };
--- a/client/data/media/use-edit-media-mutation.js
+++ b/client/data/media/use-edit-media-mutation.js
@@ -33,14 +33,14 @@ const startEditMediaItem = ( siteId, item ) => ( dispatch, getState ) => {
 export function useEditMediaMutation( queryOptions ) {
 	const dispatch = useDispatch();
 
-	const mutation = useMutation(
-		( { siteId, mediaId, payload } ) =>
+	const mutation = useMutation( {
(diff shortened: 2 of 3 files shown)


client/data/media/use-delete-media-mutation.js
```diff
@@ -2,8 +2,9 @@ import { useMutation } from '@tanstack/react-query';
...
 
 export const useDeleteMediaMutation = ( queryOptions = {} ) => {
-	return useMutation(
-		( { siteId, mediaId } ) => wp.req.post( `/sites/${ siteId }/media/${ mediaId }/delete` ),
-		queryOptions
-	);
+	return useMutation( {
+		mutationFn: ( { siteId, mediaId } ) =>
+			wp.req.post( `/sites/${ siteId }/media/${ mediaId }/delete` ),
+		...queryOptions,
```
tyxla: The subtle downside with this approach is that if `queryOptions` has `mutationFn` (although it shouldn't), it will override the one we specify above. Perhaps we should reorder them and have `mutationFn` declared below `...queryOptions`? WDYT?

client/data/media/use-edit-media-mutation.js
```diff
@@ -33,14 +33,14 @@ const startEditMediaItem = ( siteId, item ) => ( dispatch, getState ) => {
...
-		( { siteId, mediaId, payload } ) =>
+	const mutation = useMutation( {
+		mutationFn: ( { siteId, mediaId, payload } ) =>
 			wp.req.post( {
 				path: `/sites/${ siteId }/media/${ mediaId }/edit`,
 				formData: Object.entries( payload ),
 			} ),
-		queryOptions
-	);
+		...queryOptions,
```
tyxla: Same as above with regards to potentially overriding `mutationFn`.

client/data/media/use-update-media-mutation.js
```diff
@@ -3,11 +3,11 @@ import { useCallback } from 'react';
...
 
 export const useUpdateMediaMutation = ( queryOptions = {} ) => {
-	const mutation = useMutation(
-		( { siteId, mediaId, updates } ) =>
+	const mutation = useMutation( {
+		mutationFn: ( { siteId, mediaId, updates } ) =>
 			wp.req.post( `/sites/${ siteId }/media/${ mediaId }`, updates ),
-		queryOptions
-	);
+		...queryOptions,
```
tyxla: Same as above with regards to potentially overriding `mutationFn`.

tyxla (approved):
LGTM 👍 Thanks 🚀 

Left a minor question, but it's optional and non-blocking IMHO.

client/data/media/use-delete-media-mutation.js
flootr: Agreed, I had the exact same thing in mind when refactoring those. Overriding the `mutationFn` would kinda defeat the purpose of having a custom mutation hook in the first place (so it's not likely). One possible solution I can see (other than switching the arguments around) is to destructure only what we expect from the `queryOptions` objects. Another thing which comes to mind is typing `queryOptions`.

tyxla: I agree - typing does the job in #77602, so it's one option, but if you don't want to overcomplicate things, we can just reorder the lines to put `mutationFn` at the end. Otherwise, I agree that folks shouldn't be overriding `mutationFn`, but there's nothing to stop them from doing that 🙀 😉

flootr: Added https://github.com/Automattic/wp-calypso/pull/77598/commits/28ec108780490afa0ab080c9d028aec96d37174c which takes care of preventing accidental `mutationFn` overrides.

PR #77620, opened 2023-05-31
Reader: Add Suggested Follows to Search Post Results Stream

eoigal:
Follow up to https://github.com/Automattic/wp-calypso/pull/77300

This adds the suggested follows modal to the Reader search page post results.

When a user clicks on the follow button, they should get a modal showing a list of related/suggested sites to follow;

### Example

https://github.com/Automattic/wp-calypso/assets/5560595/711ccb2a-34cd-4b08-affd-faf3f4ab61dc

### Highlights

This PR will pass the post ID that the follow click was for. The thinking here is that this is the content that encouraged the user to follow this site, we should use the content from this post to find related sites... rather than lookup the most viewed post on that site as is happening now.

I works with change to the related sites endpoint to pass the `post_id` arg when known - code-D112470.

Another things is the post stream is also shown in the `Following` page - we don't want this popup there (yet?) so I have added a prop to only show this on the search page for now.

Lastly... I had to add an exception to the click redirect that normally redirects the users to the Reader full post page when it detects a click event. Since the modal relies on a click event to close the modal dialog, we need an exception to avoid this redirect.

### Testing
* Apply PR and code-D112470
* Go to Reader full post page like - http://calypso.localhost:3000/read/search
* Search something and click the follow button under the list of search posts results
* You should see modal like in example above

### DIFF
--- a/client/blocks/reader-post-actions/index.jsx
+++ b/client/blocks/reader-post-actions/index.jsx
@@ -1,11 +1,13 @@
 import classnames from 'classnames';
 import { localize } from 'i18n-calypso';
 import PropTypes from 'prop-types';
+import { useState } from 'react';
 import CommentButton from 'calypso/blocks/comment-button';
 import { shouldShowComments } from 'calypso/blocks/comments/helper';
 import PostEditButton from 'calypso/blocks/post-edit-button';
 import ShareButton from 'calypso/blocks/reader-share';
 import { shouldShowShare } from 'calypso/blocks/reader-share/helper';
+import ReaderSuggestedFollowsDialog from 'calypso/blocks/reader-suggested-follows/dialog';
 import ReaderViews from 'calypso/blocks/reader-views';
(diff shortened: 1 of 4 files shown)


davemart-in:
@eoigal I tried testing this by applying the dotcom patch and testing with calypso live. I'll send you a video of what I saw (basically, it didn't work, not sure why, but I did see some errors in the console).

davemart-in (approved):
Okay, my mistake. I clicked the wrong Follow link when testing this previously. Retested and it works as expected! Nicely done.

creativecoder:
@eoigal I came across this while tracking down some things related to the follow button, and noticed that we have a separate instance of the `ReaderSuggestedFollowsDialog` component for every individual post in the stream. Is that necessary?

What do you think of refactoring to render the modal component outside the list, and then pass down a callback function to the `ReaderPostActions`, `RelatedPostCard`, etc components in the list to open it for the specific site/post being followed?

PR #77629, opened 2023-05-31
ETK - update site intent hook to use existing data on window object

Addison-Stavlo:
Related to #

## Proposed Changes
Replaces the api request used here with data that is already [added to the window via jetpack](https://github.com/Automattic/jetpack/blob/e135711f9a130946dae1bca6c9c0967350331067/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/launchpad-save-modal.php#LL31C8-L31C34).

Use of this previous endpoint was causing many 404 errors on atomic sites, and triggering other rate limitation and breaking features as a result. p1685550995632009-slack-C03N25JPCE4

We could update to using a different endpoint that works for both simple and atomic (such as https://github.com/Automattic/wp-calypso/pull/77627). However, it looks like the permissions check on the read end for that is not what we want and may need to be updated to suit our needs. 

Reading the window for this data should work for what we want, at least until we can get the above endpoint updated.
*

## Testing Instructions

SIMPLE
* Patch this ETK build on your sandbox.
* smoke test features using this hook, such as the domain upsell callout, nav sidebar, etc.  Similarly you can log the result and verify it is the same as before.  Verify this data is present on the first render.

ATOMIC
* patch this ETK build to an atomic site (find the editing-toolkit build under the tests and checks below, click on that, in teamcity click the 'artificats' tab, and download the zip for the plugin. upload this to an atomic site)
* Load the editor and verify there are no longer 7 404s requests for 'site-intent'
* smoke test features using this hook noted above.

## Pre-merge Checklist

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
@@ -1,8 +1,9 @@
 // FIXME: We can use `useSiteIntent` from `@automattic/data-stores` and remove this.
 // https://github.com/Automattic/wp-calypso/pull/73565#discussion_r1113839120
 const useSiteIntent = () => {
-	// We can skip the request altogether since this information is already added to the window. We
-	// could update this to use the launchpad endpoint in jetpack-mu-wpcom, but that may require
+	// We can skip the request altogether since this information is already added to the window in
+	// https://github.com/Automattic/jetpack/blob/e135711f9a130946dae1bca6c9c0967350331067/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/launchpad-save-modal.php#LL31C8-L31C34
+	// We could update this to use the launchpad endpoint in jetpack-mu-wpcom, but that may require
 	// permissions changes as it requires 'manage_options' to read
 	// https://github.com/Automattic/jetpack/blob/e135711f9a130946dae1bca6c9c0967350331067/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php#L121.
 	return {
--- a/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
+++ b/apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
@@ -1,27 +1,13 @@
-import apiFetch from '@wordpress/api-fetch';
-import { useState, useEffect, useCallback } from '@wordpress/element';
-
 // FIXME: We can use `useSiteIntent` from `@automattic/data-stores` and remove this.
 // https://github.com/Automattic/wp-calypso/pull/73565#discussion_r1113839120
 const useSiteIntent = () => {
-	const [ siteIntent, setSiteIntent ] = useState( undefined );
-	const [ siteIntentFetched, setSiteIntentFetched ] = useState( false );
-
-	const fetchSiteIntent = useCallback( () => {
-		apiFetch( { path: '/wpcom/v2/site-intent' } )
-			.then( ( result ) => {
-				setSiteIntent( result.site_intent );
-				setSiteIntentFetched( true );
-			} )
-			.catch( () => {
-				setSiteIntent( undefined );
-				setSiteIntentFetched( true );
-			} );
-	}, [] );
-
-	useEffect( () => {
-		fetchSiteIntent();
-	}, [ fetchSiteIntent ] );
-	return { siteIntent, siteIntentFetched };
+	// We can skip the request altogether since this information is already added to the window. We
+	// could update this to use the launchpad endpoint in jetpack-mu-wpcom, but that may require
+	// permissions changes as it requires 'manage_options' to read
+	// https://github.com/Automattic/jetpack/blob/e135711f9a130946dae1bca6c9c0967350331067/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php#L121.
+	return {
+		siteIntent: window.Jetpack_LaunchpadSaveModal?.siteIntentOption,
+		siteIntentFetched: true,
+	};
 };
 export default useSiteIntent;


lupus2k (approved):
Tested with `start-writing` flow, `design-first` flow & domain upsell. All works great.

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
```diff
@@ -1,27 +1,14 @@
...
-import { useState, useEffect, useCallback } from '@wordpress/element';
-
 // FIXME: We can use `useSiteIntent` from `@automattic/data-stores` and remove this.
 // https://github.com/Automattic/wp-calypso/pull/73565#discussion_r1113839120
 const useSiteIntent = () => {
-	const [ siteIntent, setSiteIntent ] = useState( undefined );
-	const [ siteIntentFetched, setSiteIntentFetched ] = useState( false );
-
-	const fetchSiteIntent = useCallback( () => {
-		apiFetch( { path: '/wpcom/v2/site-intent' } )
```
okmttdhr, line 11: When did the endpoint `/wpcom/v2/site-intent` become unavailable?
I was wondering because this hook itself seems to have been in use since March 2022.

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
```diff
@@ -1,27 +1,14 @@
-import apiFetch from '@wordpress/api-fetch';
-import { useState, useEffect, useCallback } from '@wordpress/element';
-
 // FIXME: We can use `useSiteIntent` from `@automattic/data-stores` and remove this.
 // https://github.com/Automattic/wp-calypso/pull/73565#discussion_r1113839120
 const useSiteIntent = () => {
```
okmttdhr, line 3: Should we deprecate this hook 👇  as well, or not?
https://github.com/Automattic/wp-calypso/blob/trunk/packages/data-stores/src/queries/use-site-intent.ts#L11

okmttdhr (approved):
Thank you for your fixing! LGTM ✅

I left the two questions!

arthur791004 (approved):

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
arthur791004: It's unavailable on the Jetpack site at the beginning 😓

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
arthur791004: I think we have to implement it on Jetpack repo or [switch to another endpoint](https://github.com/Automattic/wp-calypso/pull/77627). But maybe we can simply return the intent from the options of the site settings.

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
okmttdhr: 😇

eoigal (approved):
lgtm

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
Addison-Stavlo: Its available from simple sites and requests to the public api. However, in the editor atomic sites make these requests directly to the atomic sites themselves where this endpoint doesn't exist on its own.

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
Addison-Stavlo: > Should we deprecate this hook 

Im not entirely sure if that would work in this situation or not (in the editor), but maybe is more robust with wpcom-proxy-request under the hood where this hook was simply api-fetch. Id imagine that hook works fine for atomic in calypso?

> I think we have to implement it on Jetpack repo or https://github.com/Automattic/wp-calypso/pull/77627. But maybe we can simply return the intent from the options of the site settings.

That other endpoint you link lives in jetpack (actually jetpack-mu-wpcom to be more specific) which fits pretty well for most cases, but I held off as it seems like the GET permissions need to be updated for the use we wanted here in the editor.  Reading intent from options where they exist is an option as well.

Addison-Stavlo:
Thank you all for the reviews and testing!

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
arthur791004: Yes, I know. But people might feel confused about why we need to get the site intent from the launchpad endpoint as they're different features 😓

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
arthur791004: Yes, that's true 👍

apps/editing-toolkit/editing-toolkit-plugin/dotcom-fse/lib/site-intent/use-site-intent.js
Addison-Stavlo: > But people might feel confused about why we need to get the site intent from the launchpad endpoint as they're different features

True!  Im wondering if we should put an inline script in jetpack-mu-wpcom that injects some necessary static info on the window on editor load. Like what we are using here, but just setup in a place where it is intended to be general info that should always be there in the editor (like site_intent). We wouldn't even need to make requests for info like that, since we wouldn't expect it to change (especially not within an editor session).

creativecoder: > Im wondering if we should put an inline script in jetpack-mu-wpcom...

If site_intent is used across multiple places, I think that makes sense! There's no need for each individual feature to re-request this data, and adding it as a JS global server-side looks like the most efficient/most performant way to get the data.

PR #77638, opened 2023-06-01
Remove flaky site creation tests and fix jest callbacks

noahtallen:
## Proposed Changes
I noticed several flaky unit tests over the past week or two. While they were each different tests, the stack trace always pointed to the file in this PR. As far as I can tell, these tests didn't use the correct Jest syntax for callbacks -- as a result, Jest didn't know that the test was finished. In fact, I think it went on to run other tests, then the callback was executed (which failed), and then the other test failed as a result.

After fixing the callback issue via a promise wrapper, I found two tests failing consistently. After troubleshooting for a bit, I noticed the functionality they covered was actually removed two weeks ago, and two very similar tests were removed at that time: https://github.com/Automattic/wp-calypso/pull/76656

But due to the test issue, that wasn't caught. So this PR makes jest wait for the tests to finish and removes those two tests.

## Testing Instructions
Unit tests should pass.

### DIFF
--- a/client/lib/signup/test/step-actions.js
+++ b/client/lib/signup/test/step-actions.js
@@ -2,8 +2,8 @@
  * @jest-environment jsdom
  */
 
+import nock from 'nock';
 import flows from 'calypso/signup/config/flows';
-import { useNock } from 'calypso/test-helpers/use-nock';
 import { createSiteWithCart, isDomainFulfilled, isPlanFulfilled } from '../step-actions';
 
 jest.mock( 'calypso/signup/config/steps', () => require( './mocks/signup/config/steps' ) );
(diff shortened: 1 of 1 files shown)


client/lib/signup/test/step-actions.js
```diff
@@ -12,31 +12,41 @@ jest.mock( 'calypso/signup/config/flows-pure', () =>
...
-				return {
-					error: 'error',
-					message: 'something goes wrong!',
-					requestBody,
-				};
-			} );
-	} );
-
-	test( 'should find available url if siteUrl is empty and enable auto generated blog name', () => {
+	nock( 'https://public-api.wordpress.com' )
```
noahtallen, line 30: useNock was deprecated.

client/lib/signup/test/step-actions.js
```diff
@@ -12,31 +12,41 @@ jest.mock( 'calypso/signup/config/flows-pure', () =>
...
+			resolve();
+		}, ...args );
+	} );
+}
+
 describe( 'createSiteWithCart()', () => {
 	// createSiteWithCart() function is not designed to be easy for test at the moment.
 	// Thus we intentionally mock the failing case here so that the parts we want to test
 	// would be easier to write.
-	useNock( ( nock ) => {
```
noahtallen, line 19: useNock was deprecated.

kozer (approved):

PR #77643, opened 2023-06-01
Enable updating already set plan and site name in the launchpad start writing flow

lsl:
Fixes https://github.com/Automattic/dotcom-forge/issues/2588

## Proposed Changes

* Enables updating site name and changing plans after already completing those steps in the launchpad "start writing" flow.

Before

Screenshot 2023-06-01 at 13-55-06 Almost ready to launch — WordPress com

After (plan now enabled regardless of completing the domain step)

Screenshot 2023-06-01 at 13-55-25 Almost ready to launch — WordPress com

Also now able to click these links a second time after completing the step:

https://github.com/Automattic/wp-calypso/assets/811776/98ab9703-65f3-4276-8302-71712d99faaf

## Testing Instructions

* With an account with no sites
* http://calypso.localhost:3000/setup/start-writing
* Publish your first post
* Complete launchpad steps in any order
* Update answers in any order
* Launching should still work
* No flows should be broken
* No other flows should be broken (ie test /start launchpad flow as well)

https://github.com/Automattic/wp-calypso/assets/811776/158067c8-adfe-4f02-93aa-6a60b2030856

Notes:
* The plan and domain screens are missing page variations depending on whether a domain or plan is already selected. I'm not sure whether that is a blocker for merging this PR as is or not. We can probably do those in another PR or two.
* We can enable the "Write first post" link but there are two issues: 1, it creates a new post, 2 returning on subsequent post publishing the url fails to update in the launchpad step and launching then fails with a browser error (mal site warning). If we can fix that url we can probably just enable users to write more than one first post or edit the one they created.
Screenshot(22)

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
@@ -110,7 +110,7 @@ export function getEnhancedTasks(
 								} )
 							);
 						},
-						disabled: task.completed && ! isStartWritingFlow( flow ),
+						disabled: task.completed && ! isBlogOnboardingFlow( flow ),
 					};
 					break;
 				case 'setup_newsletter':
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
@@ -170,7 +170,8 @@ export function getEnhancedTasks(
 							window.location.assign( plansUrl );
 						},
 						badge_text: ! task.completed ? null : translatedPlanName,
-						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isBlogOnboardingFlow( flow ),
+						disabled:
+							( task.completed || ! domainUpsellCompleted ) && ! isBlogOnboardingFlow( flow ),
 					};
 					break;
 				case 'subscribers_added':
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
@@ -170,7 +170,7 @@ export function getEnhancedTasks(
 							window.location.assign( plansUrl );
 						},
 						badge_text: ! task.completed ? null : translatedPlanName,
-						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isStartWritingFlow( flow ),
+						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isBlogOnboardingFlow( flow ),
 					};
 					break;
 				case 'subscribers_added':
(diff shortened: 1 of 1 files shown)


candy02058912:
> We can enable the "Write first post" link but there are two issues: 1, it creates a new post, 2 returning on subsequent post publishing the url fails to update in the launchpad step and launching then fails with a browser error (mal site warning). If we can fix that url we can probably just enable users to write more than one first post or edit the one they created.

I feel like we shouldn't let the user go back to the step "Write your first post" since we would like them to move forward to launching their site. They can make changes to the post after they've launched the site.

If we'd really want to support "Write your first post" to be revisited, then maybe only allowing "editing the original post"?
IMO I think the launchpad should get the user closer to launching with every step completed and not let the users get stuck too long in the launchpad. cc: @saygunnyc @DustyReagan

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -110,7 +110,7 @@ export function getEnhancedTasks(
 								} )
 							);
 						},
-						disabled: task.completed,
+						disabled: task.completed && ! isStartWritingFlow( flow ),
```
candy02058912:
```suggestion
						disabled: task.completed && ! isBlogOnboardingFlow( flow ),
```

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -170,7 +170,7 @@ export function getEnhancedTasks(
 							window.location.assign( plansUrl );
 						},
 						badge_text: ! task.completed ? null : translatedPlanName,
-						disabled: task.completed || ! domainUpsellCompleted,
+						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isStartWritingFlow( flow ),
```
candy02058912:
```suggestion
						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isBlogOnboardingFlow( flow ),
```

candy02058912 (commented):
Makes sense to support this for both `start-writing` and `design-first` flow as they're both part of the blog onboarding flow family.

lsl:
Good idea, committed.

lupus2k (approved):

lupus2k (approved):
It works as expected 🎉

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
```diff
@@ -170,7 +170,7 @@ export function getEnhancedTasks(
 							window.location.assign( plansUrl );
 						},
 						badge_text: ! task.completed ? null : translatedPlanName,
-						disabled: task.completed || ! domainUpsellCompleted,
+						disabled: ( task.completed || ! domainUpsellCompleted ) && ! isBlogOnboardingFlow( flow ),
```
paulopmt1: Minor: My linter suggested it be on a new line:

image

image

paulopmt1 (approved):
Nice work @lsl, it's working as expected, and I didn't find regressions in other flows.

client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts
lsl: This might be due to the github commit, will fix.

lsl: oops nevermind looks like @lupus2k got to it!

PR #77029, opened 2023-05-17
Plans: Add parameter to show only specific plans for the Jetpack app plan selection flow

staskus:
Related to #76968

### Details

pcdRpT-2Bo-p2
Scope: pc8eDl-Vv-p2
Discussion: p1684158877461119-slack-CFFF01Q4V

Create a WP.com Plans selection for the Jetpack app limiting it to only showing WP.com Personal and Premium plans.

## Proposed Changes

Check `jetpackAppPlans` parameter to set `planTypes` props of `PlansFeaturesMain` component to limit plan selection to Personal and Premium Plans:
- /plans/{siteURL}?**jetpackAppPlans=true**
- /plans/yearly/{siteURL}?domainAndPlanPackage=true&**jetpackAppPlans=true**

Follow a similar pattern to #66022 which added `domainAndPlanPackage` parameter

### Normal plan selection flow
1. Select simple/atomic site
2. Visit /plans/{siteURL}?**jetpackAppPlans=true** page
3. Confirm that only personal and premium plans are visible
4. Switch between annual and monthly plans
5. Confirm that only personal and premium plans are visible
6. Confirm that after selecting the plan, the checkout view is opened

### Domain and plan package plan selection flow
1. Select simple/atomic site
2. Visit /plans/yearly/{siteURL}?**domainAndPlanPackage=true**&**jetpackAppPlans=true** page
3. Confirm that only personal and premium plans are visible
4. Confirm that header with steps and back button is not visible
5. Confirm that after selecting the plan, the checkout view is opened

### Regression normal plan selection flow
1. Select simple/atomic site
2. Visit /plans page
3. Confirm that all the available plans are offered

### Regression domain and plan package plan selection flow
1. Select simple/atomic site
2. Visit /plans/yearly/{siteURL}?**domainAndPlanPackage=true** page
3. Confirm that all the available plans are offered
4. Confirm that header with steps and back button appears on top

### Before. Plans for domain and plan package

https://github.com/Automattic/wp-calypso/assets/4062343/108e0f26-593e-44e8-8961-53dcd6b53ebc

### After. Plans for domain and plan package and `jetpackAppPlans=true`

https://github.com/Automattic/wp-calypso/assets/4062343/08acb533-ddb8-4b5a-a926-1406eb068bee

### After. Plans with `jetpackAppPlans=true`

https://github.com/Automattic/wp-calypso/assets/4062343/ccc040dc-6dc3-4d2a-93ff-056e261f1635

## Pre-merge Checklist

[...]

PR #67965, opened 2022-09-19
Plugin: new onboarding flow adding a plugin

taggon:
#### Proposed Changes

* This PR creates a new onboarding flow starting from the CTA button to logging in or creating an account and adding the plugin to the checkout page. Project thread: pau2Xa-4j7-p2

#### Testing Instructions

* Choose a plugin - free or paid - from the `/plugins` page as a logged-out user.
* Click on the `Start with this plugin` button to start the new onboarding flow.
  Yoast_SEO_Plugin_—_WordPress_com_및_Plugin__new_onboarding_flow_adding_a_plugin_by_taggon_·_Pull_Request__67965_·_Automattic_wp-calypso
* On the plans step there should be only two plans: `Business` and `eCommerce`.
  Create_a_site_—_WordPress_com
* Once the checkout is successfully processed you are taken to the plugin installation flow.
* Make sure the new website has both the plan and the plugin.

#### Pre-merge Checklist

- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

PR #77667, opened 2023-06-01
Plans 2023: (foundation) Sync plans grid views across signup/tailored flows and admin

### LINKED ISSUE #77187: The pricing grid should show consistent content both in the sign-up flows and in the plan admin page.
### What

The pricing grid component is showing different content in several tailored flows. However, it always shows the default on in the plans admin page, i.e. /plans. The orignal reporting thread is: p1684403954874079-slack-CV2TX2PAN

e.g. in the Newsletter flow:
image

after signing up, here is what they see in /plans:
image

### Why

By making the pricing grid showing consistent content and the number of plans, it's a major step forward reducing user's confusion on . This will even become more important if we attempt to remove the plans step in the tailored flows.

### How

_No response_

### LINKED ISSUE #78116: Plans 2023: Convert PlansFeaturesMain to TypeScript
PlansFeaturesMain has been gradually becoming home for a lot of the complexity that we pull out of the plans grids component, especially since our focus on simplifying the latter and aiming for a package.

Unfortunately, this is still untyped, prone to hassle as we refactor or implement additional functionality. It's almost like a cloud that obscures the view from either side.

We should refactor and convert to TypeScript ASAP - although after (or on top of) https://github.com/Automattic/wp-calypso/pull/77667, as that refactors a lot of the code in that file.

chriskmnds:
Fixes https://github.com/Automattic/wp-calypso/issues/77187
Fixes https://github.com/Automattic/wp-calypso/issues/78116

## Proposed Changes

- Refactors and types the `PlansFeaturesMain` component.
- Introduces plans with intent:

Attempts to sync the plans rendered during a signup flow to the plans rendered after in admin when vising `/plans` with the created site (where relevant).

The main change is the ability to define an "intent" for rendering the grid when using the wrapper `PlansFeaturesMain` instead of passing an explicit list of `planTypes`. The intent is in turn converted centrally in a hook to a respective list of plans for `PlanFeatures2023Grid` to render. 

If intent is missing, it will either be derived from the current site or default to `default` (corresponding to the default list of plans we see at `/start/plans`).

**This is a foundation PR. No change except for Newsletter applies.** 

### TODO

- [x] Derive intent from current/selected site
- [ ] (in a follow-up) Find a different name for the wrapper - maybe `PricingView`?
- [ ] Figure out the case of plans rendered under plugins
- [x] Fix jetpack-app plans

## Testing Instructions

- Use field-guide page for plan configurations (`/plans-2023/` page)
- [ ] Check several tailored flows that they work as before in Stepper (`/setup`) and Classic Signup (`/start`) - end to end (through checkout, etc.):
    - [ ] `/setup/start-writing` & `/setup/design-first` -> same plans (Free, Personal, Premium, Business)
    - [ ] `/setup/link-in-bio` -> link-in-bio features in plans (Free, Personal, Premium)
    - [ ] `/setup/newsletter` -> newsletter features in plans (Free, Personal, Premium) 
        - [ ] if you grab the site and go to `/plans`, it should render Newsletter plans
    - [ ] `/setup/new-hosted-site` 
        - [ ] with `?hosting-flow=true` arg -> only Business & Commerce
        - [ ] without -> all/default plans except Enterprise
- [ ] Check `/plans` and `/start/plans` pages with `?plans=[plan_slug e.g. business-bundle]` and ensure selected plan is marked with "Suggested" label
- [ ] Check `/plans` page with site from `/start/p2` and `/setup/wooexpress` -> should render P2 & WooExpress specific plans
- [ ] Check plan notices that they render properly (e.g. use paid plan on `/plans` to show discounts). No plan notices in Signup
- [ ] Check plan type selector (toggle) that it works properly
- [ ] Check domain upsell modal shows/works correctly -> after selecting a paid domain in `/start`, pick a Free plan in the plan step
- [ ] Check `/plans` page for **jetpack-app** and **plugins**

## Pre-merge Checklist


client/my-sites/plan-features-2023-grid/grid-context.tsx
```diff
@@ -0,0 +1,21 @@
+import { createContext, useContext } from '@wordpress/element';
+import type { Intent } from 'calypso/my-sites/plans-features-main/hooks/use-plan-types-with-intent';
+
+interface PlansGridContext {
+	intent: Intent;
```
chriskmnds: Ideally, the actual grid package would not need to know anything about intent. So eventually deprecating this will take things a step further (it will mean feeding the grid with the respective plan features and the highlight labels for plans). 

Context can stay though and can be adapted to cover all the props passed into the grid when consumed/imported.

client/my-sites/plan-features-2023-grid/grid-context.tsx
```diff
@@ -0,0 +1,21 @@
+import { createContext, useContext } from '@wordpress/element';
+import type { Intent } from 'calypso/my-sites/plans-features-main/hooks/use-plan-types-with-intent';
```
chriskmnds: Created this explicit dependency for the same reason mentioned [below](https://github.com/Automattic/wp-calypso/pull/77667/files#r1213036575). If we get to extracting a package before, we can export the Intent type from here and use in the `use-plan-types-with-intent` hook instead.

chriskmnds:
@simison you can test newsletter flow/site and leave any comments what works/doesn't, and we decide if to be tackled here or in follow-ups. The comparison grid for once looks a little empty in `/plans/[newsletter-site]`

On my end:

`/plans/[newsletter-site]`:

Screenshot 2023-06-01 at 4 12 37 PM

`setup/newsletter`:

Screenshot 2023-06-01 at 4 13 05 PM

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
```diff
@@ -0,0 +1,130 @@
...
+
+	return null;
+};
+
+const usePlanTypesWithIntent = ( props: Props ): PlanTypesWithIntent | null => {
+	// TODO: refactor: sitePlanSlug === currentSitePlanSlug
+	// TODO: refactor: pass siteId
+	const { selectedPlan, sitePlanSlug, hideEnterprisePlan, currentSitePlanSlug } = props;
+	const intentFromSiteMeta = useIntentFromSiteMeta();
+	const intent = props.intent || intentFromSiteMeta || 'default';
```
chriskmnds: maybe move to consuming end

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+interface PlanTypesWithIntent {
+	intent: Intent;
+	planTypes: string[];
+}
+
+const useIntentFromSiteMeta = (): Intent | null => {
+	const selectedSiteId = useSelector( getSelectedSiteId ) ?? undefined;
+	const intent = useSiteIntent( selectedSiteId );
```
chriskmnds: Need to follow up on conclusions from p1685550995632009-slack-C03N25JPCE4  as there seems to have been some issues with endpoint used in a similar hook in ETK. [Here](https://github.com/Automattic/wp-calypso/pull/77629#discussion_r1212467275) there's chat about deprecating the data-stores' one (this) as well.

simison:
@chriskmnds 

> You can test newsletter flow/site and leave any comments what works/doesn't, and we decide if to be tackled here or in follow-ups. The comparison grid for once looks a little empty in /plans/[newsletter-site]

Looking great! What do you mean by comparison grid looking empty? I have 3 points to address.

Not necessarily in this PR, but we should have an "escape hatch" to see all the plans still. They might be aware of other plans.

If they get linked to this page from upsells telling to upgrade "business" (there's usually URL argument then), and we shouldn't then hide those plans in that scenario. The features in the plans grid should still be newsletter focused:

Test the flow from Settings:

Screenshot 2023-06-07 at 18 45 43

Screenshot 2023-06-07 at 18 45 55


@lucasmendes-design we should solve a problem where none of the features in the comparison grid are available until higher plans:

Screenshot 2023-06-07 at 18 39 51

- Show all plans always for comparison? (defeats purpose of focused plans)
- Don't show some sections at all?

Screenshot of the entire page like I tested it, click to see bigger:

lucasmendes-design:
Hi @simison, they can see all plans for sure. It's missing one feature in the comparison grid. See the video below from /pricing page: 

https://github.com/Automattic/wp-calypso/assets/98343583/89d49093-3236-415a-8c02-f8e92fd9fef0

> Not necessarily in this PR, but we should have an "escape hatch" to see all the plans still. They might be aware of other plans.

However, it's not possible to fit all 5 plans inside calypso because we don't have space so users can select which plan they want to compare. 

On bigger screens, it's possible but in general it will be very difficult to fit. 

> we should solve a problem where none of the features in the comparison grid are available until higher plans:

The commerce features were added by Niranja two weeks ago I think. It's something new. I would recommend hiding them if Business or Commerce aren't being compared.

chriskmnds:
> The commerce features were added by Niranja two weeks ago I think. It's something new. I would recommend hiding them if Business or Commerce aren't being compared.

cc @niranjan-uma-shankar

simison:
> However, it's not possible to fit all 5 plans inside calypso because we don't have space so users can select which plan they want to compare.

@lucasmendes-design  I meant a way to toggle from 3 plans view into the full view:

Screenshot 2023-06-08 at 12 51 10

lucasmendes-design:
Ah, sorry about that. I miss understood. We also have that on the blog flow. But we didn't implement that in Calypso because, at that time, we showed all plans in this context. :) 

Here is an example. 

image

lucasmendes-design:
I think we can add that button so users can see all plans. (View all plans - Hide plans).

chriskmnds:
> If they get linked to this page from upsells telling to upgrade "business" (there's usually URL argument then), and we shouldn't then hide those plans in that scenario. The features in the plans grid should still be newsletter focused:
> 
> Test the flow from Settings:
> 
> Screenshot 2023-06-07 at 18 45 43 

This is a big one. 

For this (and escape hatch) we need updated features for Business & Commerce plans for existing tailored flows (link-in-bio/newsletter), but also for every new flow going forward. 

I would propose a middle ground for this PR and handle the rest in a follow-up. Middle-ground being to work like it does now: show the default (all plans) grid in case of query string plans. 

I'm creating a couple of follow-up issues. I don't mind this PR being put "on hold" until any of the rest are developed - although it will make life a little harder with nested PRs:

- https://github.com/Automattic/wp-calypso/issues/78113
- https://github.com/Automattic/wp-calypso/issues/78114

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+interface PlanTypesWithIntent {
+	intent: Intent;
+	planTypes: string[];
+}
+
+const usePlanTypesWithIntent = ( props: Props ): PlanTypesWithIntent | null => {
+	// TODO: refactor: sitePlanSlug === currentSitePlanSlug
+	// TODO: refactor: pass siteId
```
chriskmnds: here next

client/my-sites/plans/ecommerce-trial/wooexpress-plans/index.tsx
```diff
@@ -119,6 +119,7 @@ export function WooExpressPlans( props: WooExpressPlansProps ) {
 	);
 
 	const plansTableProps = {
+		// TODO clk
```
chriskmnds, line 122: here next

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -90,19 +79,18 @@ const PlansWrapper: React.FC< Props > = ( props ) => {
...
 	const headerText = __( 'Choose a plan' );
 	const isInSignup = props?.flowName === DOMAIN_UPSELL_FLOW ? false : true;
-
-	const hideFreePlan = planTypes ? ! planTypes.includes( TYPE_FREE ) : reduxHideFreePlan;
-
-	const translate = useTranslate();
+	const intent = getIntent( props?.flowName, props.hostingFlow );
+	const hideFreePlan = intent
+		? reduxHideFreePlan && 'blog-onboarding' === intent
+		: reduxHideFreePlan;
```
aneeshd16: We are setting the value of `hideFreePlan` based on the value of the intent. I understand why we made this change here, but it seems like this breaks the rule that "plan types are derived from intent only in the `usePlanTypesWithIntent` hook". It also means that if any Stepper flow sets `reduxHideFreePlan` to `true`, it won't always be respected. For example, the LIB flow sets the value to true [here](https://github.com/Automattic/wp-calypso/blob/dfcaffcac16c408f3be4fd9f38ef76f80ff1ddf3/client/landing/stepper/declarative-flow/link-in-bio-domain.ts#L121), but it will have no effect.

I think we should set `hideFreePlan` to the value of `reduxHideFreePlan`, and that should always override the planTypes that we derive from the intent in `usePlanTypesWithIntent`. Would that make sense?

client/my-sites/plan-features-2023-grid/grid-context.tsx
aneeshd16, line 1: 🙌

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
```diff
@@ -27,19 +26,32 @@ const useHighlightLabel = ( { planName, flowName, currentSitePlanSlug, selectedP
...
-	if ( flowName && isNewsletterFlow( flowName ) ) {
+	if ( 'newsletter' === intent ) {
 		if ( isPersonalPlan( planName ) ) {
 			return translate( 'Best for Newsletter' );
 		}
-	} else if ( flowName && isLinkInBioFlow( flowName ) ) {
+
+		if ( isCurrentPlan ) {
+			return translate( 'Your plan' );
+		}
```
aneeshd16: I was wondering if we could take this outside the if ladder, to L30. If it's the current plan, we should always return `Your plan`, right? 🤔

client/my-sites/plans-features-main/hooks/use-intent-from-site-meta.ts
```diff
@@ -0,0 +1,35 @@
+import { useSiteIntent } from '@automattic/data-stores';
+import { useSelector } from 'react-redux';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import type { Intent } from 'calypso/my-sites/plans-features-main/hooks/use-plan-types-with-intent';
+
+interface IntentFromSiteMeta {
+	processing: boolean;
```
aneeshd16, line 7: Nit: I was wondering if we can use the default `isFetching` from react-query instead of a custom `processing`.

client/my-sites/plans-features-main/hooks/use-intent-from-site-meta.ts
```diff
@@ -0,0 +1,35 @@
...
+	intent: Intent | null | undefined;
+}
```
aneeshd16, line 9: IMO, since this involves a network call, we should also expose `isError` so that the component can handle it gracefully.

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
```diff
@@ -0,0 +1,105 @@
...
+
+interface Props {
+	intent: Intent;
+	selectedPlan: string;
+	sitePlanSlug: string;
+	hideEnterprisePlan: boolean;
+}
+
+interface PlanTypesWithIntent {
+	intent: Intent;
```
aneeshd16: I was confused by this hook since it accepts `intent` and returns `intent`.

Essentially it is casting unknown values of `intent` to `default`. I think this logic should be a part of `useIntentFromSiteMeta` and then `usePlanTypesWithIntent` can return just the `planTypes`.

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
```diff
@@ -0,0 +1,105 @@
...
+			return {
+				intent,
+				planTypes: [ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM, TYPE_BUSINESS ],
+			};
+		case 'plugins': // is this used?
+			return {
+				intent,
+				planTypes: [
+					...( currentSitePlanType ? [ currentSitePlanType ] : [] ),
+					TYPE_BUSINESS, // this can match the currentSitePlanType. needs investigation.
```
aneeshd16: I was wondering if you have investigated this already?

aneeshd16 (commented):
I did one pass through the code but haven't done any testing yet. Left some comments. :) The grid context is nicely done! 🙌 

The term `intent` and `SiteIntent` have already been used in a different context in Calypso:
https://github.com/Automattic/wp-calypso/blob/97a488006ca85c98dd406ce5fbe2519e48acf7c3/packages/data-stores/src/onboard/constants.ts#L13-L22

Introducing another meaning (and another type) for this term does increase confusion. Also, by re-purposing the server state (`useSiteIntent`), we run the risk of overwriting the value of the intent previously set:
https://github.com/Automattic/wp-calypso/blob/384b89485abc509fd781795bc3f0fa95a405da21/packages/data-stores/src/site/actions.ts#L320

This `intent` is already used in events and funnels (30d3d-pb), so any changes to the existing `intent` schema would need to be done carefully.

I don't have a good solution in mind apart from renaming `intent` in this PR to something else. 😬 Maybe `flowName` was good enough?

I was wondering if you had something better in mind?

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
chriskmnds: I think that's correct. `your plan` should take precedence 🤔  Should move this out. 

In any case, this hook will be removed next.

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: > For example, the LIB flow sets the value to true [here](https://github.com/Automattic/wp-calypso/blob/dfcaffcac16c408f3be4fd9f38ef76f80ff1ddf3/client/landing/stepper/declarative-flow/link-in-bio-domain.ts#L121), but it will have no effect.

Yes, good catch. `hideFreePlan` should pass through as `reduxHideFreePlan`, although that's also a property I want to deprecate next (all of the `hideXYZPlan` should go as they are largely unnecessary to be propagating throughout). I will revisit and see how to incorporate the value from `reduxHideFreePlan` into the intent hook.

client/my-sites/plans-features-main/hooks/use-intent-from-site-meta.ts
chriskmnds: I wanted "processing" (and the whole layer in this hook) to be an abstraction, for now. No errors or react-query specific things propagating further. Along with an `undefined` intent, it's enough to continue processing and fail gracefully (or pick up once defined). 

It could be something to engineer more down the line.

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
chriskmnds: Definitely. No need for this to pass back the `intent` value. It's always passed in as either `default` or something else. Unless I overlooked something, there's no casting per se i.e. `return { intent: 'default', ... }` below may as well be `return { intent, ... }`. I think it just lingered this way from the hook doing more originally than it does now (it would derive the site intent at some point).

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
chriskmnds: Nope. Needs to be tested in other PR, if this is even used. Do we render plans in `/plugins`?

chriskmnds:
Thanks @aneeshd16 . Pretty valuable comments. Regarding:

> This intent is already used in events and funnels (30d3d-pb), so any changes to the existing intent schema would need to be done carefully.
> 
> I don't have a good solution in mind apart from renaming intent in this PR to something else. 😬 Maybe flowName was good enough?
> 
> I was wondering if you had something better in mind?

I am actually fairly okay with introducing a plans `intent` here. I would have a hard time actually using something else for this. Even renaming to `PlansIntent` feels unnecessary right now, but that's the closest alternative. Definitely not `flowName`. 

I also don't mind that "intent" is used at other places and in different contexts. Why be concerned about it - it's a perfectly valid term for whatever context it is used AFAIC. Technically, there's no overlap - one could define a "foo" intent for plans and that's perfectly fine too, for now. Also a `SiteIntent` might also expand into a different intent for plans, given whatever conditioning's added into the respective abstraction. 

> but haven't done any testing yet

No need to test anything in this PR for functionality. Use https://github.com/Automattic/wp-calypso/pull/78428 for all testing as it carries more changes.

client/my-sites/plans-features-main/hooks/use-intent-from-site-meta.ts
chriskmnds: Same with https://github.com/Automattic/wp-calypso/pull/77667#discussion_r1239997834 When it comes to wanting to be explicit about `isError`, if we ever want to, then we'll add it. It feels like YAGNI right now. If it fails, it fails gracefully with a `null` intent, at least that's what's "intended" (gotta love "intent" word).

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: > It also means that if any Stepper flow sets reduxHideFreePlan to true, it won't always be respected. For example, the LIB flow sets the value to true [here](https://github.com/Automattic/wp-calypso/blob/dfcaffcac16c408f3be4fd9f38ef76f80ff1ddf3/client/landing/stepper/declarative-flow/link-in-bio-domain.ts#L121), but it will have no effect

This isn't the case right now actually. it looks like whatever's passed by the flows into data store gets overridden [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx#L103). I know I was trying to mimic same effect with the changes, and that's just what's happening here. It's a little weird. Technically only flow plan lists with no free plan get the `hideFreePlan` property derived from store.

It does look like a bug, but also a major change of logic/intention if we do start removing free plan on paid domain selection. This should be raised up separately if we have concerns about it. Not the sort of change I'd want to incorporate here though.

chriskmnds: In any case, I worked out the change in https://github.com/Automattic/wp-calypso/pull/78428/commits/8ce104d4f3d9bd4b029eb632a89d4ea1040ef7af to respect store flag unconditionally (and how I'd approach getting rid of props drilling along the way), and reverted in https://github.com/Automattic/wp-calypso/pull/78428/commits/7309fd36e3b7971923f235e5121bd5f178902693 as that deviates from production.

If you have a spare cycle, then maybe confirm with me if you are seeing the same.

packages/data-stores/src/onboard/index.ts
```diff
@@ -40,7 +40,6 @@ export function register(): typeof STORE_KEY {
 			'goals',
 			'hasUsedDomainsStep',
 			'hasUsedPlansStep',
-			'hideFreePlan',
```
chriskmnds, line 45: also revert, or keep with intention (confirm)

southp:
Here are the scenarios that I'm checking. It's mostly an expansion to the test plan in the description :) 
I'll update this comment along the way.

#### The classic signup flows
The plans & features should be in-sync for the following flows:

- [x] The main onboarding flow, /start
- [x] /start/launch-site
- [ ] The Free plan + paid domain modal should work as expected.
    - [x] Continue with the Personal plan
    - ❌Continue with the Free plan -> this seems broken. After clicking, the modal disappears and leaves a few secs of time for the user to click around. In production, the progress page appears immediately after click.
- [x] A plan-specific flow, /start/premium
- [x] /start/p2
- [x] Spot-checking if any of the above flows work through checkout.

Some more ad hoc flows worth checking:
- [x] The domain-only flow, /start/domain
    - [x] Just buy a domain
    - [x] Create a new site (only 4 plans should be shown)

#### The tailored flows
The plans & features should be in-sync for the following flows:

- [x] /setup/newsletter 
- [ ] ❌ /setup/link-in-bio -> the default plans grid was shown in /plans. Fwiw, I tested as a new sign-up.
- [ ] ❌/setup/start-writing -> the default plans grid was shown in /plans
- [ ] ❌/setup/design-first -> the default plans grid was shown in /plans
- [ ] 🤔/setup/videopress -> It's hard to determine. It no longer has the plans step, but clicking "view plans" also lead to the checkout directly. I believe it's a bug unrelated to this PR.
- [x] /setup/free
- [ ] ❌/start/hosting (this flow starts from the classic framework) -> the standard grid was shown in /plans
- [x] /setup/wooexpress
- [ ] ❌/setup/new-hosted-site
    - [ ] ❌?hosting-flow=true (only Business & Commerce) -> the standard grid was shown in /plans
    - [ ] ❌?hosting-flow=false (all plans are shown except for the Enterprise plan) -> the standard grid was shown in /plans
- [x] Spot-checking if any of the above flows work through checkout.

#### Misc cases

- [ ] The plans grid should work as expected on an Atomic site.
- [x] Check plan notices that they render properly (e.g. use paid plan on /plans to show discounts). 
- [x] No plan notices in Signup
- [x] The term toggle should work properly
- [ ] /plans page for jetpack-app
- [ ] /plans page for plugins
- [x] Upgrading from the plans page should work as expected.
- [x] /plans page for an account having both a newsletter site and a "standard" site.

jeyip:
Lots of overlap with James's tests, so I only included new issues I saw or confirmed existing failures

### Classic Signup Flows
- [x] /start/domain
  - ~After choosing a custom domain and clicking on "Start site," only the Personal, Premium, Business, and Commerce plans are shown in onboarding. In /plans, all plans are shown.~ See https://github.com/Automattic/wp-calypso/pull/77667#issuecomment-1610906033

### Tailored Flows
- [x] /setup/wooexpress
  - As far as I can tell, the wooexpress flow has no plans page in onboarding, so there's nothing to sync
- [ ] ❌ /setup/new-hosted-site?hosting-flow=true
  - I saw the standard grid instead of just the business and ecommerce plans
- [ ] ❌ /setup/new-hosted-site?hosting-flow=false
  - I saw the standard grid instead of all plans except enterprise

### Misc
- [ ] ❌  Check /plans and /start/plans pages with ?plans=[plan_slug e.g. business-bundle] and ensure selected plan is marked with "Suggested" label
  - This might be user error, but I couldn't get this to work when navigating to `/plans/{SITE_SLUG}?plans=business-bundle`
- [x] Check plan notices that they render properly (e.g. use paid plan on /plans to show discounts). No plan notices in Signup
- [x] Check plan type selector (toggle) that it works properly
  - I believe you're describing the comparison grid toggle? If so, then it works as expected
- [ ] ❌  Check domain upsell modal shows/works correctly -> after selecting a paid domain in /start, pick a Free plan in the plan step
  - Continue with the Free plan doesn't do anything when clicked
 
 @chriskmnds would you mind describing how to test this more thoroughly? `Check /plans page for jetpack-app and plugins`

southp:
@jeyip 

> After choosing a custom domain and clicking on "Start site," only the Personal, Premium, Business, and Commerce plans are shown in onboarding. In /plans, all plans are shown.

This is actually the expected behavior. In general we want to discourage the combination of a paid domain + the Free plan, and we have implemented various ways for historical reasons. In `/start/domain`, its way is to simply hide the Free plan when a custom domain is selected.

jeyip:
> This is actually the expected behavior. In general we want to discourage the combination of a paid domain + the Free plan, and we have implemented various ways for historical reasons. In /start/domain, its way is to simply hide the Free plan when a custom domain is selected.

Thanks for the heads up! I went ahead and updated my testing notes to reflect this

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
```diff
@@ -27,23 +26,24 @@ const useHighlightLabel = ( { planName, flowName, currentSitePlanSlug, selectedP
 	);
 	const isSuggestedPlan =
 		selectedPlan && planLevelsMatch( planName, selectedPlan ) && isAvailableForPurchase;
+	const { intent } = usePlansGridContext();
 
-	if ( flowName && isNewsletterFlow( flowName ) ) {
+	if ( isCurrentPlan ) {
```
jeyip, line 31: `Non-blocking / Nit`:

This hook could use a test in `client/my-sites/plan-features-2023-grid/hooks/test`

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -51,23 +44,19 @@ interface Props {
 	hostingFlow: boolean;
 }
 
-function getPlanTypes( flowName: string | null, hideFreePlan: boolean, hostingFlow: boolean ) {
+function getPlansIntent( flowName: string | null, hostingFlow: boolean ): PlansIntent | null {
```
jeyip, line 47: I understand the reasoning behind defining an intent for the plans grid, but, out of curiosity, why do we need to map `flowName` to a unique slug? Ex. NEWSLETTER_FLOW -> 'plans-newsletter'.

What if we just used the `flowName` to dictate which plan is shown instead?

jeyip: After digging through recent conversations, Christos explains his reasoning in https://github.com/Automattic/wp-calypso/pull/77667#issuecomment-1604516032

jeyip: > Why be concerned about it - it's a perfectly valid term for whatever context it is used AFAIC. Technically, there's no overlap - one could define a "foo" intent for plans and that's perfectly fine too, for now. Also a SiteIntent might also expand into a different intent for plans, given whatever conditioning's added into the respective abstraction.

If I understand this correctly, I'm not sure I agree. Given how large our codebase is, it's not always immediately obvious why certain decisions were made, and for devs without context, having _both_ a `siteIntent` _and_ a `plansIntent` implies a need for both.

From what I can tell, `siteIntent` alone would serve our current needs ( definitely correct me if I'm wrong ). Although this might change in the future, there _is_ a cost, and it's one extra layer of indirection when trying to understand our plans grid.

With that being said, I don't feel like what I bring up should stop the PR from proceeding, so I'll resolve the conversation for now

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
```diff
@@ -51,23 +44,19 @@ interface Props {
...
 	switch ( flowName ) {
 		case START_WRITING_FLOW:
 		case DESIGN_FIRST_FLOW:
-			return hideFreePlan
-				? [ TYPE_PERSONAL, TYPE_PREMIUM, TYPE_BUSINESS ]
-				: [ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM, TYPE_BUSINESS ];
+			return 'plans-blog-onboarding';
 		case NEWSLETTER_FLOW:
-			return [ TYPE_FREE, TYPE_PERSONAL, TYPE_PREMIUM ];
+			return 'plans-newsletter';
```
jeyip, line 53: Could we create shareable constants for the plan intent slugs? It'd be used here and in hooks like `client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts`, `client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts`, etc.

EX.
```javascript
export const PLANS_NEWSLETTER = "plans-newsletter";
export const PLANS_LINK_IN_BIO = "plans-link-in-bio";
// etc...
```

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
```diff
@@ -4,7 +4,7 @@ import {
 	isPremiumPlan,
 	planLevelsMatch,
 } from '@automattic/calypso-products';
-import { isNewsletterFlow, isLinkInBioFlow } from '@automattic/onboarding';
+import { usePlansGridContext } from '../grid-context';
```
jeyip, line 7: `Non-blocking / Nit`:

This hook could use a test in `client/my-sites/plan-features-2023-grid/hooks/test`

packages/calypso-products/src/main.ts
```diff
@@ -718,7 +718,7 @@ export const getPopularPlanSpec = ( {
 	isJetpack,
 	availablePlans,
 }: {
-	flowName: string;
+	flowName?: string | null;
```
jeyip, line 721: The typing here says that flowName may be `undefined` (because it's an optional argument), `string`, or `null` Just wanna confirm that this is the intention?

client/signup/config/step-components.js
```diff
@@ -28,7 +28,6 @@ const stepNameToModuleName = {
 	plans: 'plans',
 	'plans-new': 'plans',
 	'plans-business': 'plans',
-	'plans-ecommerce': 'plans',
```
jeyip, line 31: `Non-blocking / Nit`:

Could you explain more about which flow is being affected by the removal of `plans-ecommerce` here? I'm assuming one of the `start` flows?

client/my-sites/plans/main.jsx
```diff
@@ -272,10 +270,10 @@ class Plans extends Component {
 				redirectTo={ this.props.redirectTo }
 				withDiscount={ this.props.withDiscount }
 				discountEndDate={ this.props.discountEndDate }
-				site={ selectedSite }
+				siteId={ selectedSite?.ID }
 				plansWithScroll={ false }
 				hidePlansFeatureComparison={ this.props.isDomainAndPlanPackageFlow }
-				planTypes={ planTypes }
+				intent={ isJetpackAppPlans ? 'plans-jetpack-app' : null }
```
jeyip: `Non-blocking / Nit`:

What made us move to the inline ternary? I find it easier to read as it was structured originally with planTypes

```
const intent = this.props.jetpackAppPlans ? 'plans-jetpack-app' : null ;

<PlansFeaturesMain
  {...rest}
  intent={intent}
/>
```

packages/data-stores/src/onboard/index.ts
jeyip: `Non-blocking / Nit`:

Would you mind elaborating on what you mean here?

jeyip:
As far as I can tell, I don't have any glaring concerns or blocking requests for changes to the PR. A lot was already covered by Aneesh in previous conversations

With that being said, there's a large volume of changes to very public-facing portions of the site, so I'm planning to give these changes a second pass tomorrow morning. For the time being, once issues found in the testing notes https://github.com/Automattic/wp-calypso/pull/77667#issuecomment-1609003996 + https://github.com/Automattic/wp-calypso/pull/77667#issuecomment-1610264290  are **_either_** discussed or resolved, then +1 from me

southp:
I'm doing another round of test based on the new understanding of the expected behavior. ~This comment will again be updated along the way.~

@chriskmnds I've done the round of testing and found only one broken case; will shift into code review mode soon :) 

#### The classic signup flows
The plans & features should be in-sync for the following flows:

- [x] The main onboarding flow, /start
- [x] /start/launch-site
- [ ] The Free plan + paid domain modal should work as expected.
    - [x] Continue with the Personal plan
    - [ ] ❌Continue with the Free plan -> Re-confirmed that it is broken.
- [x] A plan-specific flow, /start/premium
- [x] /start/p2
- [x] Spot-checking if any of the above flows work through checkout.

Some more ad hoc flows worth checking:
- [x] The domain-only flow, /start/domain
    - [x] Just buy a domain
    - [x] Create a new site (only 4 plans should be shown)

#### The tailored flows
The plans & features should be in-sync for the following flows:

- [x] /setup/newsletter 
- [x] /setup/link-in-bio
- [x] /setup/start-writing
- [x] /setup/design-first
- [x] /setup/videopress 
- [x] /setup/free
- [x] /start/hosting (this flow starts from the classic framework)
    - [x] Create a site
    - [x] Migrate a site (this will lead to /start/import)
- [x] /setup/wooexpress
- [x] /setup/new-hosted-site (@jeyip FYI this flow works fine for me somehow)
    - [x] ?hosting-flow=true (only Business & Commerce)
    - [x] ?hosting-flow=false (all plans are shown except for the Enterprise plan)
- [x] Spot-checking if any of the above flows work through checkout.

#### Misc cases

- [x] The plans grid should work as expected on an Atomic site.
- [x] Check plan notices that they render properly (e.g. use paid plan on /plans to show discounts). 
- [x] No plan notices in Signup
- [x] The term toggle should work properly
- [x] /plans page for jetpack-app (@jeyip see https://github.com/Automattic/wp-calypso/pull/77029 on how to test it :) )
- [x] /plans page for plugins (@jeyip I believe this refers to the plugins flow. You can access it through `/plugins` whilel logged-out. By getting any of the plugin, it will then lead you to `/start/business` flow. We once attempted to make it install the selected plugin as well in https://github.com/Automattic/wp-calypso/pull/67965, but it is on-hold almost permanently now.)
- [x] Upgrading from the plans page should work as expected.
- [x] /plans page for an account having both a newsletter site and a "standard" site.

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: thanks.. maybe down the line. too early to over-engineer here

client/my-sites/plan-features-2023-grid/hooks/use-highlight-adjacency-matrix.ts
chriskmnds: maybe down the line. lots of things changing here.

client/signup/config/step-components.js
chriskmnds: `/start` flow. 

There's no use for this step anywhere. so removed, instead of updating with intent.

packages/calypso-products/src/main.ts
chriskmnds: Yeah, intentional. A lot of these types are created to match underlying selectors that may return either or

client/my-sites/plan-features-2023-grid/hooks/use-highlight-label.ts
chriskmnds: this will get removed next. 
p.s. we will likely see more changes coming through that won't contain tests I'm afraid.

client/my-sites/plans/main.jsx
chriskmnds: Probably something to clean up, as the comment above also remained in the code.

In general, I don't mind _not_ creating additional variables when there's a single use for them. In this case, I see little difference between

`intent={ isJetpackAppPlans ? 'plans-jetpack-app' : null  }`

and 

```
const intent = isJetpackAppPlans ? 'plans-jetpack-app' : null;
...
intent={ intent }
```

or things like 

```
const hideFreePlan = this.props.isDomainAndPlanPackageFlow;
...
hideFreePlan={ hideFreePlan }
```

I totally get it this is fine to do and helps some folks with reading "React code" I guess. However, I've seen things get off hand when everything is tried to be reassigned before assigned to props. verbose variables, thinking ahead (that something may become more "complex" to compute down the line), etc. It can feel noisy.

packages/data-stores/src/onboard/index.ts
chriskmnds: this lingered from some experimental code that got reverted. however, it's also one of the changes I want to explore next

For some reason, we persist `hideFreePlan` in the store, presumably to deal with redirections or something. However, this being persisted means the Onboard store will have a value for this when other flows/places attempt to query. One change I'd like to explore next is to remove all the props for hiding individual plans (most of them can already be skipped with "intent" here) e.g. `hideFreePlan` can be derived from store directly when a flow renders the plans-grid calypso wrapper (i.e. `PlansFeaturesMain` can just query the store for it). One less prop to pass around as it's a value already instantiated centrally in the store, and one less pattern (`hideFooPlan`) is a good idea. 

^ probably not if it's persisted though as it may get ugly

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
southp: No, `/plugins` is defaulted to the Business plan. From my test, currently it leads to `/start/business`, so we don't need to worry about it.

client/my-sites/plans-features-main/hooks/use-visible-plans-for-plan-features.ts
```diff
@@ -0,0 +1,90 @@
...
+
+	if ( hideBusinessPlan ) {
+		plans = plans.filter( ( planSlug ) => ! isBusinessPlan( planSlug ) );
+	}
+
+	if ( hideEcommercePlan ) {
+		plans = plans.filter( ( planSlug ) => ! isEcommercePlan( planSlug ) );
+	}
+
+	// TODO clk - should probably remove this (why the retriction to these plans?)
```
southp, line 75: Yes, let's remove this in a follow-up. First of all it doesn't make sense anymore now that we use it for plans outside of dotcom plans. Second of all, it's now Typescript so it can be more efficiently by better typing if we really want.

client/signup/config/steps-pure.js
```diff
@@ -249,32 +244,21 @@ export function generateSteps( {
...
 			fulfilledStepCallback: isPlanFulfilled,
 		},
+
 		'plans-new': {
 			stepName: 'plans',
 			providesDependencies: [ 'cartItem' ],
 			fulfilledStepCallback: isPlanFulfilled,
 		},
 
-		'plans-ecommerce': {
```
southp, line 258: Good catch. This step isn't used anywhere now.

client/my-sites/plans-features-main/hooks/use-visible-plans-for-plan-features.ts
chriskmnds: Perfect. already removing it in other PR

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: I think of this intent (the plans' intent) as a specific and clear concern for the plans grid. There's no use of it outside of plans. The type itself is `PlansIntent`, and that's what other places in the app can/will see if need be. I see little reason to rename internal variables to "plansIntent" either, in the same way that I see no reason to rename `usePlansWithIntent` to `usePlansWithPlansIntent` (it's verbose and doesn't sound very confident about the code that runs things for me).

Question whether the word "intent" is owned by Site meta and cannot be used in any other context, then I wholly disagree. The code and types should be clear and separated enough to avoid confusion (and I think that's the case here). If you confuse `PlansIntent` type with `SiteIntent`, then might be worth giving it a bit more time. 

We can revisit this. It just works for me as a mental model right now.

> From what I can tell, siteIntent alone would serve our current needs ( definitely correct me if I'm wrong ).

I don't think that's the case. The plans' intent can be defined prior to site existence, hence values like "default" (and potentially more defaults that are not "wpcom" - so "default" may evolve to "default-wpcom", etc.).

chriskmnds: > I see little reason to rename internal variables to "plansIntent" either

This definitely the next thing I'd consider. But I'd definitely hate to start seeing `usePlansWithPlansIntent` stuff around. Maybe just the context variable.

client/my-sites/plans/main.jsx
chriskmnds: doing it like you mentioned 👍

client/my-sites/plans-features-main/hooks/use-plan-types-with-intent.ts
chriskmnds: Sound. will remove the comments.

@southp do we need to revisit anything at all related to plans rendered under /plugins routes? I wasn't able to follow the routes and get the plans view, but haven't tested enough

packages/data-stores/src/onboard/index.ts
chriskmnds: I will leave this change to see if it leads to any breakage that we haven't spotted. We've been testing a lot of things with it, so hopefully this can safely not persist. It may be false hope though as I don't think there's actual use of it right now (last I recall, it gets overridden in the plans wrapper).

southp: Agree. Let's leave it as it is for now. I believe [I've covered most of the cases](https://github.com/Automattic/wp-calypso/pull/77667/#issuecomment-1612356342) and doesn't see any obvious breakage :)

southp:
I've confirmed that:

* The paid domain + Free plan modal now works as expected.
* The rest of the onboarding flow works as expected. 
* The plans page for the Jetpack app still works as expected with the change of https://github.com/Automattic/wp-calypso/pull/77667/commits/9639ff9bf63eaf37c885d5d663deb08661612b62. 
* The `/plugins` flow still works as expected with the change of https://github.com/Automattic/wp-calypso/pull/77667/commits/c56e57e97928badc8b6b664ec174bd3122ade59f

Once the typing error is fixed, I believe this PR will be ready to ship :)

packages/data-stores/src/onboard/index.ts
chriskmnds: perfect

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: to be clear - we'll engineer a lot more around this going forward. also include a README in the end to cement it, depending on what stays. "intent" variables will likely end up being exported from plans data-store package

client/landing/stepper/declarative-flow/internals/steps-repository/plans/plans-wrapper.tsx
chriskmnds: > From what I can tell, siteIntent alone would serve our current needs ( definitely correct me if I'm wrong ).

Technically speaking, I think you would not be wrong, in principle. 

Considering the case where a plans grid is rendered within a space that does not involve all the plans, but is also not part of a flow ( e.g. plans under `/plugins`, or the jetpack app, or some marketting/logged-out page ) - we could either: 
- give it a "default" intent that defines all the plans, and conditionally pass properties to hide the ones we don't need. like we've been doing so far on some occasions in `/start`
- or define an intent, like "plans-plugins" or "plans-foo-upsell" and declare the plans to use

it's all variable. we'll be narrowing in on things as we go. eventually if `SiteIntent` becomes the only explicit intent that's needed, then we'll map one into the other.

jeyip:
## Testing
### Requirements

#### The classic signup flows
The plans & features should be in-sync for the following flows:

- [x] The main onboarding flow, /start
- [x] /start/launch-site
- [x] The Free plan + paid domain modal should work as expected.
    - [x] Continue with the Personal plan
    - [x] Continue with the Free plan
- [x] A plan-specific flow, /start/premium ( Didn't see an onboarding plans page )
- [x] /start/p2 ( Didn't see an onboarding plans page )
- [x] Spot-checking if any of the above flows work through checkout.

Some more ad hoc flows worth checking:
- [x] The domain-only flow, /start/domain
    - [x] Just buy a domain
    - [x] Create a new site (only 4 plans should be shown)

#### The tailored flows

/plans and onboarding plans page in sync
- [x] /setup/newsletter

Not in sync but should continue to behave as expected
- [x] /setup/link-in-bio
- [x] /setup/start-writing
- [x] /setup/design-first
- [x] /setup/videopress 
- [x] /setup/free
- [x] /start/hosting
    - [x] Create a site
    - [x] Migrate a site (this will lead to /start/import)
- [x] /setup/wooexpress
- [x] /setup/new-hosted-site (@southp this flow isn't syncing for me, but I'll defer to your testing)
    - [x] ?hosting-flow=true (only Business & Commerce)
    - [x] ?hosting-flow=false (all plans are shown except for the Enterprise plan)
- [x] Spot-checking if any of the above flows work through checkout.

### Browsers
- [x] Chrome

jeyip (approved):

jeyip:
Per our discussions about focusing on testing and less on code review here p1688022806475799/1687988246.215679-slack-CV2TX2PAN, I've gone ahead and given this an approval 🚢

chriskmnds:
> Per our discussions about focusing on testing and less on code review here

but please do post any other concerns you have about the code

southp (approved):
Just noticed that I should have left an explicit approval. I'm soo looking forward to use TS in this project and see the syncing live  :)

chriskmnds:
> Just noticed that I should have left an explicit approval. I'm soo looking forward to use TS in this project and see the syncing live :)

Sound. We'll really start to hone in on improvements from this point onward.

chriskmnds:
Thanks for the amazing reviews and all that testing done! 🙇

PR #77674, opened 2023-06-01
Sites: put created site at the top of the list

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2616.

## Proposed Changes

It's difficult to find the created site if the user has multiple entries. Especially because the sites list gets ordered.

This PR makes sure that the `?new-site` is always on top, making it impossible for the user to miss t.

## Testing Instructions

1. Browse `/setup/new-hosted-site`;
2. Pick the free plan;
3. Verify that you have been redirected to `/sites` and that your newly created site shows up as the first site on the list.

### DIFF
--- a/client/sites-dashboard/components/sites-dashboard.tsx
+++ b/client/sites-dashboard/components/sites-dashboard.tsx
@@ -212,7 +212,7 @@ export function SitesDashboard( {
 				<SitesDashboardSitesList
 					sites={ allSites }
 					filtering={ { search } }
-					sorting={ sitesSorting }
+					sorting={ { ...sitesSorting, newSiteID } }
 					grouping={ { status, showHidden: true } }
 				>
 					{ ( { sites, statuses } ) => {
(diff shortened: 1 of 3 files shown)


wojtekn:
@zaguiini, thanks for working on fixing the issue I had in mind! :-) 

I'm wondering if there is a better way to fix it. The "Magic" sorting mode sorts sites favoring those which the user interacted recently with. Instead of relying on the URL parameter, could we solve it on the interaction logging level (or sorting algorithm level) to favor each newly created site?

Also, besides that, would it make sense to redirect the user to the site's home and not to the `/sites` page?

zaguiini:
>I'm wondering if there is a better way to fix it. The "Magic" sorting mode sorts sites favoring those which the user interacted recently with. Instead of relying on the URL parameter, could we solve it on the interaction logging level (or sorting algorithm level) to favor each newly created site?

Yeah, that makes sense. I was thinking of favoring only when it was first created, then defaulting to default sorting after a refresh. But then the problem will happen again 🤔 

Because of that,

>Also, besides that, would it make sense to redirect the user to the site's home and not to the /sites page?

It would indeed make more sense to redirect the user to the site immediately. I guess that for the first site, it makes sense to redirect to `/sites` so that they are aware of our dashboard. But for subsequent site purchases maybe it's a better idea indeed to send them to the site immediately.

zaguiini:
But on the other hand, we can modify magic sorting to always put fresh sites on top. I think that's a better idea. Let me try and to that instead.

PR #77678, opened 2023-06-01
New hosted site: Go to site home if the user has more than one site

zaguiini:
Supersedes https://github.com/Automattic/wp-calypso/pull/77674. Closes https://github.com/Automattic/dotcom-forge/issues/2616.

## Proposed Changes

In the onboarding flow, it makes sense to redirect the user to `/sites` after they purchase the site because we want them to be aware of our dashboard.

But this behavior might not be wanted in subsequent purchases because most of the time, after a site is created, it makes sense to modify it immediately. So why send them to `/sites`? They are already aware of the dashboard, and know how it works, so let's send them to `/home/%s`.

See more in p1685625431206379-slack-C0347E545HR.

## Testing Instructions

- Login to an account with zero sites and create a site through `/setup/new-hosted-site`. You should be redirected to `/sites`;
- Create a new site with the same account and flow. You should be redirected to `/home/%s`.

### DIFF
--- a/client/landing/stepper/declarative-flow/new-hosted-site-flow.ts
+++ b/client/landing/stepper/declarative-flow/new-hosted-site-flow.ts
@@ -9,6 +9,7 @@ import {
 	persistSignupDestination,
 	setSignupCompleteFlowName,
 } from 'calypso/signup/storageUtils';
+import getSites from 'calypso/state/selectors/get-sites';
 import { useQuery } from '../hooks/use-query';
 import { ONBOARD_STORE } from '../stores';
 import { isInHostingFlow } from '../utils/is-in-hosting-flow';
@@ -55,6 +56,7 @@ const hosting: Flow = {
 		];
 	},
 	useStepNavigation( _currentStepSlug, navigate ) {
+		const siteCount = useSelector( getSites ).length;
 		const hostingFlow = useSelector( isInHostingFlow );
 		const { setSiteTitle, setPlanCartItem, setSiteGeoAffinity } = useDispatch( ONBOARD_STORE );
 		const { siteGeoAffinity, planCartItem } = useSelect(
@@ -73,9 +75,13 @@ const hosting: Flow = {
 			}
 
 			if ( _currentStepSlug === 'processing' ) {
-				const destination = addQueryArgs( '/sites', {
-					'new-site': providedDependencies.siteId,
-				} );
+				const destination =
+					siteCount === 0
+						? addQueryArgs( '/sites', {
+								'new-site': providedDependencies.siteId as number,
+						  } )
+						: `/home/${ providedDependencies.siteSlug }`;
+
 				persistSignupDestination( destination );
 				setSignupCompleteSlug( providedDependencies?.siteSlug );
 				setSignupCompleteFlowName( flowName );


vykes-mac:
I think we should still redirect to `/sites`  especially when having multiple sites to further emphasize it as the central location to manage your sites.  I don't think we should redirect there just for awareness purposes but as the main point of interacting with multiple sites. 

This would require some re-working of sorting algorithms but I don't think we should redirect to `/sites` only for the first site.

I also think it's better to make the experience consistent rather one thing here and another thing even though I followed the same action.

zaguiini:
Thanks for the feedback, @vykes-mac. I used to think just like you but some arguments and deep thoughts made me reconsider that approach. Here are a couple of them:

#### Redirect to `/sites` closes the onboarding loop

The original idea didn't include making `/sites` the destination for every post-checkout step. Hosting onboarding i2 is more focused on going full circle: registering for an account, creating/importing a site, and verifying that there is a way to manage multiple sites as an advanced hosting feature.

#### Sites page should be a birds-eye view, or a hub, to quickly access the sites and their functionalities

We worked on allowing the users to pick `/sites` the home page in the past with the idea of making it the starting point for their story, not the end. Furthermore, the Sites page should be a hub to quickly access and diagnose site issues, exactly what we're going to implement in Sites Requiring Attention i1. Right now, it's not it.

#### The user will want to immediately manage the created site

For the first purchase, and as we're trying to consolidate the idea of the page and sell it to the user, it makes sense to redirect them to `/sites` because then that page would be actionable, and it would click immediately that "this is the go-to place when wanting to navigate between sites or take quick actions".

However, for subsequent ones, the user will already be aware of the page, what it does, and how to get to it. Let's be more efficient and place them directly at the created site's home.

#### No other flow redirects to `/sites`

The idea of going back to `/sites` might work for `/setup/(new|import)-hosted-site` but there are a ton of onboarding flows in Calypso.

If we want to make it consistent, we'd need to change at least 100 entries to redirect to `/sites` after checkout... What I'm trying to say is that redirecting to `/sites` exclusively on the `/setup/new-hosted-site` flow doesn't make it more or less consistent. That'd be a huge effort 😅 

I do agree that's inconsistent, but every other flow sends the user to the home page of the site directly. We'd need to figure that out first.

#### There are some sorting issues

And finally, sorting issues. [I have tried](https://github.com/Automattic/wp-calypso/pull/77674) putting the `?new-site` at the top of the list, but @wojtekn raised good points about the sorting algorithms, how magic should behave, and, to top it: refreshing the page doesn't stick the new site to the top. Even if it did, it'd be very weird to have a site starting with "not A" before sites starting with "A" in the case of alphabetic sorting, for example.

Still talking about alphabetic sorting, the chances of the new site being below the fold or on the new page are considerable. What should we do if the new site is not even showing up on the page?

Yolo is going to improve magic sorting (p1685626274840809/1685625431.206379-slack-C0347E545HR), and then maybe it will get a little bit better. But right now this isn't a reality.

### If we're still not convinced...

I have been thinking about adding a guided tour to `/sites` after their first purchase. We would demonstrate the features of the page and, at the end, say that for future purchases, they'll be sent to the site's home directly. This would strike a good balance between "not ever going to /sites" and "going every time even when it doesn't make sense".

I thought about adding a toggle at the end of the guided tour but that would bring the same inconsistency problems described above.

client/landing/stepper/declarative-flow/new-hosted-site-flow.ts
```diff
@@ -55,6 +56,7 @@ const hosting: Flow = {
 		];
 	},
 	useStepNavigation( _currentStepSlug, navigate ) {
+		const siteCount = useSelector( getSites ).length;
```
p-jackson, line 59: I feel the site count from the user object is going to be more reliable because the full site list is large and takes a long time to load. Feels more likely we'll have the user object loaded.

```suggestion
		const siteCount = useSelector( state => getCurrentUser( state )?.site_count );
```

zaguiini: Yeah, I was using that first, but I was afraid of the number not updating if the user doesn't refresh the page. I'll give it another shot and check whether it works as we think it does 🧑‍🔬

p-jackson:
You make some good points @zaguiini. In fact, so many good points in favour of `/home`, and only one in favour of `/sites` (the idea of "going full circle") that you make me think we should just be consistent and _always_ go `/home`.

The reason this won't be super nice for for the hosting flow is that `/home` is not very "hosting" friendly. Lots of support articles and checklists on there that are more aimed consumers; not people who are trying to preview our hosting features. As a stop gap we could take Atomic sites to `/hosting-config`, that way at least we're taking them to the main Calypso dashboard.

The other reason why going `/home` for sites that have just gone Atomic, is that personally I find the main Calypso dashboard a bit buggy while a transfer is in progress. Better not to encourage them to poke around until the transfer is complete. Dropping them on `/sites` is another way to delay that process.

Sorry if I keep harping on about this, but at the start of the project I imagined we would sidestep this issue by making the `/setup/new-hosted-site` flow a one off. Once you finished or dropped out of the hosting flow you just go back to using the regular `/start` site creation flow everywhere. It'd continue to create simple sites that land on `/home`. I think doing it this way avoids the rabbit holes of figuring out what the experience is supposed to be like for devs/users-interested-in-hosting generally; all we've got to figure out is this one flow.

client/landing/stepper/declarative-flow/new-hosted-site-flow.ts
p-jackson: Oh that's an interesting point. Worth checking.

zaguiini:
>The reason this won't be super nice for for the hosting flow is that /home is not very "hosting" friendly. Lots of support articles and checklists on there that are more aimed consumers; not people who are trying to preview our hosting features.

We could create a new view for dev accounts that bought sites from `/setup/(new|import)-hosted-site`. There's time for that.

>As a stop gap we could take Atomic sites to /hosting-config, that way at least we're taking them to the main Calypso dashboard.

I don't think that's any better than sending them to `/sites` as most of the Atomic features are there, and we're showing a non-blocking spinner on that page, meaning they can poke around.

>The other reason why going /home for sites that have just gone Atomic, is that personally I find the main Calypso dashboard a bit buggy while a transfer is in progress. Better not to encourage them to poke around until the transfer is complete. Dropping them on /sites is another way to delay that process.

What if we don't redirect Business sites immediately to `/home` but, instead, show the progress bar just like it happens when you buy a Commerce plan? This specific plan only redirects to `/home` after the site has been transferred.

>Sorry if I keep harping on about this, but at the start of the project I imagined we would sidestep this issue by making the /setup/new-hosted-site flow a one off. Once you finished or dropped out of the hosting flow you just go back to using the regular /start site creation flow everywhere.

It doesn't need to be a one-off. I see no problem in making `/setup/new-hosted-site` the go-to onboarding flow for subsequent purchases in `/sites` CTAs. It will transfer Business and Commerce sites to Atomic by default (that's actually a good thing), and the user can still buy Free, Personal, or Premium plans. 

Personally, for developers, I think this onboarding flow is better, even if we disconnect it entirely from the /hosting onboarding flow as we're going to skip shenanigans (like asking for intent, domain name, etc) and go straight to what matters: site creation and plan. The developer can figure out the rest.

zaguiini:
So what I propose is:

- `/setup/(new|import)-hosted-site` as the site creation flow for developer accounts, both on `/sites` and in the "Add new site" buttons in the Calypso sidebar (this is already implemented);
- Wait for Atomic transfer to complete before redirecting to `/home/%s`;
- Redirect to `/home/%s` immediately if not a Business/Commerce site;
- New home view for sites bought through `/setup/(new|import)-hosted-site`.

In the future, as `/sites` get better, more actionable, and informative, we can revisit this and find a way to always redirect to `/sites` after checkout, also handling _how_ to display the created site there. One potential idea would be to detach it from the list/grid directly and show it on top for the remainder of the session instead of removing it on refresh.

zaguiini:
Closing in favor of https://github.com/Automattic/wp-calypso/pull/77723.

PR #77680, opened 2023-06-01
New hosted site: display site options for non-Atomic sites

zaguiini:
## Proposed Changes

@wojtekn found it weird that they're getting a `<username><unique-id>.wordpress.com` site and that doesn't make sense indeed.

I think we should offer the site options step for 

## Testing Instructions

1. Open `/setup/new-hosted-site` and select Free, Personal or Premium plan;
2. Check that you're redirected to the options step and there's a suggestion name;
3. Go to checkout (or home in case of free) and see that the site slug is related to the site title.

In the case of a non-Atomic plan, you should not see the data center picker. Picking Business or Commerce should display the data center picker.

Check that `/setup/new-hosted-site?hosting-flow` is unaffected and the first step is `siteOptions` with the data center picker available.


vykes-mac (approved):
This makes sense and work as described.

wojtekn:
Thanks for fixing it @zaguiini . I've tested it and it works great now.

PR #77694, opened 2023-06-01
Fix: Remove unwanted exception

andres-blanco:
When the `AssertConditionState.FAILURE` state is returned by the `useAssertionHooks` method and exception was being thrown. This is unwanted behaviour since is not actionable and it's triggering errors in sentry.

This PR removes the throw and just returns an empty component.

## Testing Instructions

* Apply this patch or use the calypso.live link below.
* Navigate to a `/setup` stepper component without including the `siteSlug` or `siteId` in the URL (eg: `/setup/site-setup/goals`).
* Verify no exception is thrown.
* Verify that the URL you checked is working ok with the correct URL parameters.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/index.tsx
@@ -152,7 +152,7 @@ export const FlowRenderer: React.FC< { flow: Flow } > = ( { flow } ) => {
 				return <StepperLoader />;
 			/* eslint-enable wpcalypso/jsx-classname-namespace */
 			case AssertConditionState.FAILURE:
-				throw new Error( assertCondition.message ?? 'An error has occurred.' );
+				return <></>;
 		}
 
 		const StepComponent = 'asyncComponent' in step ? lazy( step.asyncComponent ) : step.component;


client/landing/stepper/declarative-flow/internals/index.tsx
```diff
@@ -152,7 +152,7 @@ export const FlowRenderer: React.FC< { flow: Flow } > = ( { flow } ) => {
 				return <StepperLoader />;
 			/* eslint-enable wpcalypso/jsx-classname-namespace */
 			case AssertConditionState.FAILURE:
-				throw new Error( assertCondition.message ?? 'An error has occurred.' );
+				return <></>;
```
noahtallen, line 155: If you do want to observe the error messages, you could also use log2logstash or something like that. In all the cases we were talking about, the page is redirected anyways, so returning an empty component seems fine. Though, are there other cases where the page won't get redirected which we should try to cover?

noahtallen (approved):

client/landing/stepper/declarative-flow/internals/index.tsx
andres-blanco: I inspected all appearances of `AssertConditionState.FAILURE` and they all redirect. 

I think it's not worth logging here given that this state was introduced to provide a way for steps to check for preconditions prior to being rendered, so getting an empty component is not really an unwanted behaviour. Having said that, the naming looks definitely wrong, but I will merge this to prevent the Sentry logs from appearing.

noahtallen: Thanks!

PR #69616, opened 2022-10-31
Third-Party Premium Themes: Adding CTA on themes page

valterlorran:
#### Proposed Changes

* Add the appropriate CTAs for the third-party premium themes on the theme(`/theme/{themeId}/{siteSlug}`) page.
* Fixed the CSS for the CTA.

**Site with a Business plan**
203050568-885e17ae-764c-487b-a89b-84cf99efd7d7

**Site without a Business plan**
Screen Shot 2022-11-21 at 09 00 46

Screen Shot 2022-11-22 at 10 58 13

**User not authenticated**
Screen Shot 2022-11-21 at 08 57 15

#### Testing Instructions

Follow the instructions in this link D92217-code to sandbox the public-api.

1. Testing when the theme is not purchased but the user is eligible for externally managed themes.
1.1 Access `/theme/makoney/{SITE-SLUG}`, where the site has the business plan.
1.2 You should see the CTA inviting users to subscribe to the theme.
1.3 Clicking on the CTA should add the external-managed theme to the cart and redirect to the checkout page.
1. Testing when the theme is not purchased and the user is not eligible for externally managed themes.
2.1 Access `/theme/makoney/{SITE-SLUG}` with a free site.
2.2 You should see the CTA inviting users to upgrade.
2.3 Clicking on the CTA should add the external-managed theme, add the monthly Business plan and the theme to the cart, and redirect to the checkout page.

Related to #69468 #69480

Depends on:
 * #69832
 * #69852
 * #69831
 * #69853
 * #70214 
 * #70227

Closes #71083

PR #77696, opened 2023-06-01
Theme SSR: Use state to control the disabling of button and banner

### LINKED ISSUE #76663: Theme Details: refreshing the page while logged-out results to CTAs not working
### Quick summary
In Theme Details logged out page, refreshing the page results to CTAs being disabeld.

### Steps to reproduce

1. Log out or use incognito
2. Load `https://wordpress.com/theme/miniature`
3. Refresh the page

CTA buttons "Pick this Design" and "Subscribe banner" to be available

### What actually happened

CTA buttons "Pick this Design" and "Subscribe banner" were disabled

CleanShot 2023-05-08 at 14 04 15@2x

WBerredo:
Fixes #76663

## Proposed Changes

Use a state variable to control the disabling of the button and banner.
This variable is used to guarantee the initial state will always be the same (disabled) and then it should verify the isLoading status.
This was implemented because SSR needs to have the same initial markup as mentioned on https://github.com/facebook/react/issues/20152#issuecomment-721066999.

PS: After solving this issue, we are still having issues on SSR as we have a random text being assigned on footer, this is not tackled on this PR

CleanShot 2023-06-01 at 15 01 53@2x

## Testing Instructions

**Prod**
* On an incognito tab, go to a theme details page. Ex: `/theme/miniature`
* Verify on the first load the banner and the `Pick this design` button are enabled
* Reload the page and check how both are DISABLED

**This version**
* On an incognito tab, go to a theme details page. Ex: `/theme/miniature`
* Verify on the first load the banner and the `Pick this design` button are enabled
* Reload the page and check how both are STILL ENABLED 

| Before  | After |
| ------------- | ------------- |
|CleanShot 2023-06-01 at 14 49 52@2x|CleanShot 2023-06-01 at 14 52 41|

## Pre-merge Checklist


noahtallen:
> PS: After solving this issue, we are still having issues on SSR as we have a random text being assigned on footer, this is not tackled on this PR

By the way, we do have a markup cache applied for SSR'd pages. As a result, it's likely that random text in the footer could get saved in that cache

noahtallen (approved):
Works well for me 👍

WBerredo:
> By the way, we do have a markup cache applied for SSR'd pages. As a result, it's likely that random text in the footer could get saved in that cache

Do you have any docs or any example PR on how to do it?

noahtallen:
I don't know if there are docs, but the problem is that we cannot avoid having a cache due to performance constraints on the server. Serving a page from the cache often takes <10ms, but rendering the entire page can take 400-500ms or more due to extra network requests back to WordPress.com servers. 

Here's where the cache is set:
https://github.com/Automattic/wp-calypso/blob/778ea6db1488ee3f53b640c7edca55d6009a43b2/client/server/render/index.js#L107

Here's where we get the markup from cache:
https://github.com/Automattic/wp-calypso/blob/778ea6db1488ee3f53b640c7edca55d6009a43b2/client/server/isomorphic-routing/index.js#L17-L21

Here's where we skip rendering when the cache exists:
https://github.com/Automattic/wp-calypso/blob/71777122e390109cb5d6c8d48179cd2ed52f902a/client/controller/shared.js#L24

Whether the cache is used depends on whether the server instance still has the markup in memory, which depends on other factors (for example, there's no cache immediately after a new deploy.)

We can't change that functionality, but I was mostly curious what you're trying to solve related to this:

> we are still having issues on SSR as we have a random text being assigned on footer

WBerredo:
>  I was mostly curious what you're trying to solve related to this:

> we are still having issues on SSR as we have a random text being assigned on footer

We are receiving the following warning about the client and server versions being different, and as they differ the client version is picked. As it should always happen (as the footer text is randomly selected) I'm assuming we will never use the SSR'd version, am I right about it?

noahtallen:
> We are receiving the following warning about the client and server versions being different, and as they differ the client version is picked. As it should always happen (as the footer text is randomly selected) I'm assuming we will never use the SSR'd version, am I right about it?

Right, this random data is generated once on the server and once on the client: https://github.com/Automattic/wp-calypso/blob/45d208d3198ea576b9c6356a0b9baaa7b5da1472/packages/wpcom-template-parts/src/hooks/use-automattic-branding-noun.ts#L19

You would need to find a way to prevent the client from generating that data again I think. I imagine this would happen even without the cache too :)

WBerredo:
It will be tracked as a separate issue here: #77799

Thanks for the input @noahtallen.

jsnajdr:
The random footer issue can be solved by generating an empty markup on the initial render and then updating it in an effect. SSR does only the initial render, so it will generate empty non-random markup. The client will hydrate it correctly and then will update it with the real random markup.
```js
function randomRoger( translate ) {
  const variants = [
    translate( 'An Automattic foo' ),
    translate( 'An Automattic bar' ),
  ];
  return variants[ Math.floor( Math.random() * variants.length ) ];
}

function Footer() {
  const [ roger, setRoger ] = useState( null );
  useEffect( () => {
    setRoger( randomRoger( translate ) );
  }, [ translate ] );
  if ( ! roger ) {
    return null;
  }
  return <div className="lp-logo">{ roger }</div>;
}
```

client/my-sites/theme/main.jsx
```diff
@@ -152,7 +152,13 @@ class ThemeSheet extends Component {
...
 	};
 
+	/**
+	 * ssrLoadingTrigger is a flag to guarantee the isLoading method always starts with true
+	 * It allows the initial state to always be the same to make SSR work properly
+	 * It starts with true but is right way updated to false on componentDidMount
+	 */
 	state = {
+		ssrLoadingTrigger: true,
 		showUnlockStyleUpgradeModal: false,
```
jsnajdr, line 165: Unrelated to this PR, but the `isAtomicTransferCompleted` would also deserve initialization to `false` here.

client/my-sites/theme/main.jsx
```diff
@@ -188,8 +198,12 @@ class ThemeSheet extends Component {
 
 	isLoading = () => {
 		const { isLoading, isThemeActivationSyncStarted } = this.props;
-		const { isAtomicTransferCompleted } = this.state;
-		return isLoading || ( isThemeActivationSyncStarted && ! isAtomicTransferCompleted );
+		const { isAtomicTransferCompleted, ssrLoadingTrigger } = this.state;
+		return (
+			isLoading ||
+			( isThemeActivationSyncStarted && ! isAtomicTransferCompleted ) ||
+			ssrLoadingTrigger
```
jsnajdr: It would be better to solve the root cause of the issue which is: the `isLoading` is `true` when the theme is externally managed (like Miniature is) and when the products list (returned by `getProductsList`) is empty. The products list is empty when it hasn't been yet loaded from the REST endpoint.

On the server, the products list is not in the server Redux state and the `/theme` SSR endpoint is not loading the list. It's empty and the SSR code uses `isLoading = true`.

In the browser, the products list can be stored in the local IndexedDB cache, and the products list initial state is initialized with that data. The products list is never empty, and `isLoading` is `false`.

There is a REST endpoint fetch happening, but it only updates the existing data, `isLoading` is never `true` during the load. One would have to call the `isProductsListFetching` selector to detect that load. And on initial render that value is `false` anyway because the fetch hasn't started yet, it's triggered in an effect.

If the products list is not in the local IndexedDB cache, then in the browser you also get `isLoading = true`, which is in sync with the server and there is no problem. The hydration mismatch happens only when there is a cached products list in IndexedDB.

jsnajdr: The code that computes `isLoading` from the "is loading cart" and "is products list empty" values was added in #69616 by @valterlorran. I'm wondering why are we interested in these values, because it seems that the `ThemeSheet` component is not looking at any products list data or cart data -- why does it care about them being loaded?

client/my-sites/theme/main.jsx
WBerredo: Considering we don't initialize it, it starts as `undefined`, which will return the proper value on our checks (`! isAtomicTransferCompleted`). But I agree we could be more explicit about it.

client/my-sites/theme/main.jsx
WBerredo: On the logged-in version, there are changes in the CTA and the banner, and if they are clicked at this previous state it points to the wrong places (tested now). So I imagine this is the reason why it's disabled at this point.

CleanShot 2023-06-07 at 11 03 48

WBerredo: Another solution would be having the `disabled` property set via state with a fixed initial state. It would be a similar approach to what was suggested for the footer issue.

WBerredo:
The implementation for the random footer issue was implemented here: #77916

client/my-sites/theme/main.jsx
jsnajdr: > On the logged-in version, there are changes in the CTA and the banner

Thanks, I think I understand what is happening now. On your screen capture, the label changes from "Upgrade to subscribe" to "Subscribe to activate". That change happens probably when the `/sites/:site/features` request (initiated from top-level layout component) finishes fetching, because only after that we know that the site is eligible for external themes. The `disabled` state doesn't watch for this, however. It's entirely possible that the wrong "Upgrade to subscribe" button is rendered as active.

The products list is not needed for rendering the Theme page, it's needed by the `onButtonClick` hander, which executes the `addExternalManagedThemeToCart` function. That function assumes that the products list is loaded. It looks at the products list to figure the right products (plan and theme) to add to the cart.

It would be better if this `addExternalManagedThemeToCart` function itself did the products list check. It's an async thunk, it only needs a few lines like this to be added:
```js
const hasProductsList = ( state ) => Object.keys( getProductsList( state ) ).length > 0;
if ( ! hasProductsList( state ) ) {
  await dispatch( requestProductsList( { type: 'all' } ) );
}
// should have valid products list now. Important to get fresh state from `getState`.
const products = getProductsList( getState() );
```
Having this, you can completely remove `QueryProductsList` and `getProductsList` from `theme/main.jsx`. After clicking the button, the handler itself will wait for products to load.

The `isLoadingCart` selector is there to disable the button after it was clicked. The `isLoadingCart` state is set to `true` by the `addExternalManagedThemeToCart` action. If you start loading products in `addExternalManagedThemeToCart`, it would be good to set `isLoadingCart` to `true` before that async load, so that the button is disabled immediately after clicking.

> Another solution would be having the `disabled` property set via state with a fixed initial state.

Yes, this also works, although it doesn't address the core issue. It's putting another hack on top of hacky code. You can do:
```js
const [ buttonDisabled, setButtonDisabled ] = useState( true );
const areProductsLoaded = ! isExternallyManagedTheme || Object.values( getProductsList( state ) ).length > 0;
useEffect( () => {
  if ( areProductsLoaded ) {
    setButtonDisabled( false );
  }
}, [ areProductsLoaded ] );
```
Here `buttonDisabled` always starts as `true`, and even if products are already loaded at initial render, it's reset to `false` only in an effect, a little bit later. Both server and client initial render have reliably always a disabled button.

The `disabled` prop for the button still needs to also take `isCartLoading` into account, so that it gets disabled also after clicking.

WBerredo: I've used the second alternative adapted to a class component.

---

I did some tests with the first alternative, but after removing the ProductList check we would still need to add additional checks to keep the button disabled while its text is changing, and I was afraid to fall into the same case where we would have different initial states depending if its cached or not.

jsnajdr: > after removing the ProductList check we would still need to add additional checks to keep the button disabled while its text is changing

The text is changing as a reaction to the `/sites/:id/features` response arriving and being stored in the Redux state. Not as a reaction to the Products list. And nobody is waiting for the features.

It would be nice to implement the fixes that I suggested at some point. It would untangle some of the spaghetti that accumulated in the Themes code over time. It's not a blocker for this PR, but worth considering.

client/my-sites/theme/main.jsx
jsnajdr: Yes, everything still works even if it's not initialized, but it's a good custom to declare it. If this component was converted to TypeScript, the type checker would start complaining about this.

client/my-sites/theme/main.jsx
```diff
@@ -173,6 +178,11 @@ class ThemeSheet extends Component {
 		if ( this.props.themeId !== prevProps.themeId ) {
 			this.scrollToTop();
 		}
+
+		if ( this.state.disabledButton !== this.isLoading() ) {
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState( { disabledButton: this.isLoading() } );
+		}
```
jsnajdr, line 193: This "props to state sync" code should also be in `componentDidMount`. If `this.isLoading()` is `false` on initial mount, and then the component never rerenders again, then `state.disabledButton` would stay `true` and we don't want that. The "never rerenders again" case is theoretical, the component will almost certainly be rerendered, but it's good to make components resilient.

jsnajdr (approved):
Good to go after adding a `state.disabledButton` sync also to `componentDidMount`.

client/my-sites/theme/main.jsx
WBerredo: Added at https://github.com/Automattic/wp-calypso/pull/77696/commits/bce2f0d71f8be5defe3f63328f2b2ea0e5c18418

client/my-sites/theme/main.jsx
WBerredo: Added at https://github.com/Automattic/wp-calypso/pull/77696/commits/bce2f0d71f8be5defe3f63328f2b2ea0e5c18418

client/my-sites/theme/main.jsx
WBerredo: I created a task to handle this in the future https://github.com/Automattic/wp-calypso/issues/78120

PR #77702, opened 2023-06-01
URL Search component: Fix browser hard refreshing on search param update

bogiii:
## Proposed Changes

After [Update packages to versions which include React 18 as peer dependencies](https://github.com/Automattic/wp-calypso/commit/8d1be7918a1c79edfba76eea679a234f0b2bb11c) changes, URL Search component doesn't work as before; after each keyboard typing, the browser hard-refresh happens instead of only updating the query param.

`page.replace` method expects pathName.
Before this fix, it was full URL, ex: `https://calypso.localhost:3000/people/subscribers/{SLUG}`

It is fixed by providing only pathName + query params.

Screen Capture on 2023-06-01 at 22-30-32

## Testing Instructions

* Go to `/people/team/{SLUG}`
* Search by any term
* Check if there is no browser refreshing

## Pre-merge Checklist

### DIFF
--- a/client/lib/url-search/index.js
+++ b/client/lib/url-search/index.js
@@ -6,12 +6,12 @@ const debug = debugFactory( 'calypso:url-search' );
 
 /**
  * Function for constructing the url to page to. Here are some examples:
- * 1. { uri: 'google.com', search:'hello' } --> 'google.com?s=hello'
+ * 1. { uri: 'google.com', search:'hello' } --> 'https://google.com/?s=hello'
  * 2. {
  *     uri: 'wordpress.com/read/search?q=reader+is+awesome',
  *     search: 'reader is super awesome'
  *     queryKey: 'q',
- *    } --> 'wordpress.com/read/search?q=reader+is+super+awesome'
+ *    } --> result.searchParams.get( 'q' ) === 'reader is super awesome'
  *
  * @param {Object} options the options object
  * @param {string} options.uri the base uri to modify and add a query to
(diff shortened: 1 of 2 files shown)


noahtallen:
Sorry for breaking this! By the way, here's where I updated the test:
https://github.com/Automattic/wp-calypso/pull/77121/commits/18d144d1d88df7bbb9ab1086c068f8bb89e20a32#diff-573bd74fdcbaf65a8d85b422e2f64a9c445d0c72e8f9257582a5dce0daccc4e0

Here's where I updated the logic:
https://github.com/Automattic/wp-calypso/pull/77121/commits/c816abc2c0af278ff7e2d4fdf7db6d810618d0de#diff-cebfad6600ff9542a9322fc3e7789ed559d28ff7783bb3c40301b14165363378

I think the core change was that the function now includes the protocol, but previously it still included the domain name. (E.g. it wasn’t returning just the pathname)

zaerl (approved):
Work as expected. 🚢

bogiii:
> I think the core change was that the function now includes the protocol, but previously it still included the domain name. (E.g. it wasn’t returning just the pathname)

Thanks for providing more details to me.
Obviously, when the URL has a protocol part, the `page` lib does a page refresh.

I'm not quite sure what should be the right solution here:
- Reverting back by removing the protocol part, or
- Leaving this solution + tests update, since the protocol and origin will remain the same after page `replace` and `redirect` methods invokes

@noahtallen, what do you think?

noahtallen:
Since I broke it, I'll went ahead and removed the protocol. I like the idea of keeping the previous behavior as much as possible. While the function is only used in one place, it seems weird that even though the domain is provided as an argument, it wouldn't be fully included in the result.

noahtallen:
Oh, interesting, it doesn't work with only the protocol removed.

PR #62543, opened 2022-04-05
[E2E] Allow "unspecified" env vars and bring its interface closer to `getTestAccountByFeature`'s

fullofcaffeine:
## Changes proposed in this Pull Request

As I was writing a P2 about cross-target testing (still a WIP) and to announce the `getTestAccountByFeature` [abstraction](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-e2e/src/lib/utils/get-test-account-by-feature.ts), I noticed a couple of things were less than ideal in the current implementation:

### There was no easy way to communicate the absence of a certain artifact when using `envVariables` as an input to `getTestAccountByFeature`:

So far, the absence of an artifact such as the `GUTENBERG_EDGE` env var meant the actual attribute in the [envVariables](https://github.com/Automattic/wp-calypso/blob/trunk/packages/calypso-e2e/src/env-variables.ts) object was set to `false`, implicitly meaning the use of a "stable" release. This is fine to an extent, but there are a couple of problems with it:

1. Its implicit nature can be confusing. It's not 100% clear what version is being used if `GUTENBERG_EDGE` is `false`. You can assume `stable`, but it's not really obvious.
2. It's impossible to write a test against a site that might not have Gutenberg installed. This might be a pretty rare event, but I can see that happening with other plugins or artifacts we might want to test against. A tangible example might be CoBlocks. What if we want to declare a criterion for a site that doesn't have CoBlocks installed (not the case now, but could happen in the future?). Right now (before the changes here), it's not possible to when `getTestAccountByFeature` is connected to the `envVariable` object via the `envToFeatureKeys` ([example](https://github.com/Automattic/wp-calypso/blob/trunk/test/e2e/specs/blocks/blocks__shallowly_smoke_test_popular_blocks.ts#L30-L31)) because the `envVariables` will forcedly assign `false` to the `COBLOCKS_EDGE` attribute if the env var is not set, which right now means the `stable` plugin is installed and not that it's absent.

### The new `unspecified` state for feature artifacts

If a feature-like artifact - like a plugin - is to be incorporated and configurable via the criteria that are selectable by `getTestAccountByFeature`,  then you should decide what states it should have. In the case of plugins like CoBlocks and Gutenberg - as of this changeset - we have three:

* `stable`
* `edge`
* `unspecified` (=== `undefined`)

[...]

PR #77713, opened 2023-06-02
[E2E][AT][GB] Add new test criteria for AT GB edge nightly E2E testing

fullofcaffeine:
## Proposed Changes

Follow-up to: https://github.com/Automattic/wp-calypso/pull/77639 and https://github.com/Automattic/wp-calypso/pull/77711

* Add a new test criteria that will be selected when a `GUTENBERG_NIGHTLY` env var is set, which will cause the `gutenberg` version to be set to `nightly` and the right `nightly` account/site to be selected for the test to run against.

## Testing Instructions

* Unit tests should pass.
* E2E tests should pass.
* Test with the GB AT nighly build (Desktop suffices), make sure no tests fail because of unexpected JS errors or account not found errors. Result is [here](https://teamcity.a8c.com/buildConfiguration/calypso_WPComTests_gutenberg_atomic_nightly_desktop/10183210?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true). As long as there are no "account not found" errors, it should be considered 🟢


packages/calypso-e2e/src/lib/utils/get-test-account-by-feature.ts
```diff
@@ -4,10 +4,10 @@ import type { JetpackTarget, SupportedEnvVariables } from '../../types/env-varia
 
 export type TestAccountEnvVariables = Pick<
 	SupportedEnvVariables,
-	'GUTENBERG_EDGE' | 'COBLOCKS_EDGE' | 'TEST_ON_ATOMIC' | 'JETPACK_TARGET'
+	'GUTENBERG_EDGE' | 'GUTENBERG_NIGHTLY' | 'COBLOCKS_EDGE' | 'TEST_ON_ATOMIC' | 'JETPACK_TARGET'
 >;
 
-type Env = 'edge' | 'stable';
+type Env = 'edge' | 'stable' | 'nightly';
```
dpasque, line 10: **Suggestion (non-blocking / take-it-or-leave-it 😄 ):** Out of curiosity, what do you think of collapsing these two ENV vars `GUTENBERG_EDGE` and `GUTENBERG_NIGHTLY` into just a single ENV var like `GUTENBERG_TARGET` that has these three values? 

Totally optional of course, and I realize this may be super annoying to actually implement 😆 (because you'd have to tweak the old Gutenberg TC builds), but I think it might read a bit nicer and scale a bit cleaner. We took this approach with Jetpack and it's worked well there so far! I just sometimes find combining multiple booleans into one enum-like type to sometimes be a bit harder to follow, because there's often implicit hierarchical rules involved. 😄

dpasque (approved):
Exciting! 😄 LGTM! I had one suggestion, but it might be kind of annoying to implement, so feel free to take it if you like it! Otherwise 🚀

packages/calypso-e2e/src/lib/utils/get-test-account-by-feature.ts
fullofcaffeine: @dpasque Jinx! I actually did that in this (somewhat old) PR: https://github.com/Automattic/wp-calypso/pull/62543. I will add a note to revisit it soon and extract that part into a smaller PR. I think there are other improvements there that should be revisited, but the PR is too big and complex, so I need to split it - and refresh my memory about it - as it's been a while!

fullofcaffeine: There's also this https://github.com/Automattic/wp-calypso/pull/62235, which is an experiment to add a wildcard syntax sugar, which could reduce the number of current repeat rules (for example, in this change, we needed to add expand the combination of nightlies to the coblocks rules and the siteEditor rules in addition to the general rule. In theory, wildcards would allow a single nightlies rule, and we could "ignore" the other attribute's values.

fullofcaffeine: ~~Actually! I think https://github.com/Automattic/wp-calypso/pull/62543 implements a simpler alternative to wildcards 🤔 -- anyway, I'll try to revive it soon, giving priority to your suggestion of simplifying the way we pass the gutenberg version.~~ No, it doesn't. With the current implementation of `getAccountByFeature`, "ignoring" a set value basically means having it set to all its possible values. The best way to do that without repeating the rule (like we do now), would be an wildcard syntax sugar.

PR #77720, opened 2023-06-02
Fix - Plan upgrade credit information on the checkout page

ddc22:
Related to :  Automattic/martech#1665
Additional Context : pdgrnI-2nz-p2
## Proposed Changes
* Increase clarity about the plan upgrade credit in the checkout page by including an additional line of explanation as shown below.

| Existing      | Proposed |
| ----------- | ----------- |
| Yearly plan : image         |   image  |
| Monthly Plan :  image  |     image     |

## Testing Instructions

* Purchase a plan under Ecommerce in a site
* Go to the `/plans` page
* Click on `Upgrade` on any plan
* The above changes should be visible at checkout.
* Make sure the above changes work in different terms like monthly and annual.

## Pre-merge Checklist

### DIFF
--- a/packages/wpcom-checkout/src/checkout-line-items.tsx
+++ b/packages/wpcom-checkout/src/checkout-line-items.tsx
@@ -29,7 +29,7 @@ import formatCurrency from '@automattic/format-currency';
 import styled from '@emotion/styled';
 import { useViewportMatch } from '@wordpress/compose';
 import { useTranslate } from 'i18n-calypso';
-import { useState, PropsWithChildren } from 'react';
+import { useState, PropsWithChildren, ReactChild } from 'react';
 import { getLabel, getSublabel } from './checkout-labels';
 import { getItemIntroductoryOfferDisplay } from './introductory-offer';
 import { isWpComProductRenewal } from './is-wpcom-product-renewal';
@@ -776,19 +776,17 @@ function isCouponApplied( { coupon_savings_integer = 0 }: ResponseCartProduct ) 
 	return coupon_savings_integer > 0;
 }
 
-function FirstTermDiscountCallout( { product }: { product: ResponseCartProduct } ) {
+function PlanUpgradeCreditInformation( { product }: { product: ResponseCartProduct } ) {
 	const translate = useTranslate();
-	const planSlug = product.product_slug;
 	const origCost = product.item_original_subtotal_integer;
 	const finalCost = product.item_subtotal_integer;
+	const discount = origCost - finalCost;
+	const planSlug = product.product_slug;
 	const isRenewal = product.is_renewal;
 
-	// Do not display discount reason if there is an introductory offer.
-	if ( product.introductory_offer_terms?.enabled ) {
-		return null;
-	}
-
 	if (
+		// Do not display discount reason if there is an introductory offer.
+		product.introductory_offer_terms?.enabled ||
 		// Do not display discount reason for non-wpcom, non-jetpack products.
 		( ! isWpComPlan( planSlug ) && ! isJetpackProductSlug( planSlug ) ) ||
 		// Do not display discount reason if there is no discount.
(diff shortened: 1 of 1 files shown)


packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -798,22 +796,42 @@ function FirstTermDiscountCallout( { product }: { product: ResponseCartProduct }
...
+					isSmallestUnit: true,
+					stripZeros: true,
+				} ),
+			},
+		} );
 	}
 
 	if ( isBiennially( product ) || isTriennially( product ) ) {
-		return <DiscountCallout>{ translate( 'Discount for first term' ) }</DiscountCallout>;
+		oneTimeExplanation = translate( 'Discount: %(discount)s Applied in first term only', {
```
sirbrillig: Could these assignments be changed back to early return statements? They are generally easier to reason about since the order is more clear (the first one that matches wins instead of the other way around) and because without an early return you have to follow all the code in the function to see what will happen even if a condition does match.

packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -798,22 +796,42 @@ function FirstTermDiscountCallout( { product }: { product: ResponseCartProduct }
 		// Do not display discount reason if a coupon is applied.
 		isCouponApplied( product )
 	) {
-		return null;
```
sirbrillig, line 801: There may be a reason I don't understand but what's the benefit of changing this `null` to an empty fragment? React will render nothing when a component returns `null`.

sirbrillig (commented):
Could you add a "before" screenshot to complement the "after" screenshots in the PR description? Otherwise it's a little hard to see what has changed.

packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -798,22 +796,42 @@ function FirstTermDiscountCallout( { product }: { product: ResponseCartProduct }
...
+					isSmallestUnit: true,
+					stripZeros: true,
+				} ),
+			},
+		} );
 	}
 
 	if ( isYearly( product ) ) {
-		return <DiscountCallout>{ translate( 'Discount for first year' ) }</DiscountCallout>;
+		oneTimeExplanation = translate( 'Discount: %(discount)s Applied in first year only', {
```
sirbrillig: I'm not certain, but I think that "applied" should not be capitalized since it's part of an ongoing sentence. We would not write, "A discount was Applied".

Also it might (?) be a good idea to have a translator's comment above this translation to make it clear what "%(discount)" is (a formatted number).

packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -798,22 +796,42 @@ function FirstTermDiscountCallout( { product }: { product: ResponseCartProduct }
...
 		isCouponApplied( product )
 	) {
-		return null;
+		return <></>;
 	}
-
+	let oneTimeExplanation: ReactChild = '';
 	if ( isMonthlyProduct( product ) ) {
-		return <DiscountCallout>{ translate( 'Discount for first month' ) }</DiscountCallout>;
+		oneTimeExplanation = translate( 'Discount: %(discount)s Applied in first month only', {
```
sirbrillig: Should we use `hasTranslation` on these so that they can be merged before the new strings are translated?

packages/wpcom-checkout/src/checkout-line-items.tsx
ddc22: The types were screaming for some reason when I returned null. Maybe it was a false alarm 🤔

packages/wpcom-checkout/src/checkout-line-items.tsx
ddc22: I was actually planning to string freeze this change after a confidence check from a reviewer 👍

ddc22:
@sirbrillig I did a minor change to the implementation and also have included the proper screenshots.  👍

packages/wpcom-checkout/src/checkout-line-items.tsx
sirbrillig: Sounds good!

sirbrillig:
Thank you @jdc91! Mostly I think this is great. I pushed a couple of changes which just refine what you were already doing. Let me know if you like them and I'll do a final review (just in case you want to add anything else or have questions about what I did).

ddc22:
@sirbrillig You read my mind, that is what I wanted to do all along 😁. I did that halfway and thought maybe this is just a string and should probably be a function call. But this approach looks much better! 👍 
Please do proceed to test it 🙌

packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -776,19 +776,17 @@ function isCouponApplied( { coupon_savings_integer = 0 }: ResponseCartProduct )
...
 	const finalCost = product.item_subtotal_integer;
+	const upgradeCredit = origCost - finalCost;
+	const planSlug = product.product_slug;
 	const isRenewal = product.is_renewal;
 
-	// Do not display discount reason if there is an introductory offer.
-	if ( product.introductory_offer_terms?.enabled ) {
-		return null;
-	}
-
```
ddc22, line 790: Nice change 👍

sirbrillig:
I took some additional screenshots of how this affects both checkout and the mini-cart in the masterbar.

| Before | After |
|--------|--------|
| before-77720 | after-77720 |
| mini-cart-before-77720 | mini-cart-after-77720 | 

This does not affect the "Discount for first X" text used by introductory offers:

jetpack-after-77720

sirbrillig (approved):
This looks fine from my point of view. It does increase the vertical height of line items but I don't know if it matters a lot. It may also change again with the checkout redesign (p9Jlb4-6FB-p2) but that does not block this change and this definitely provides a better explanation of the pro-rated discount (aka: "Upgrade credit") than before.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8028148

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8028148).

Issue #77321, opened 2023-05-24
Virtual Themes: Add PA 'opt-out' mechanism


dsas:
Roadmap issue: https://github.com/Automattic/dotcom-forge/issues/2528

Add a tag/meta/etc to patterns so pattern authors can opt that virtual theme out of going via the Pattern Assembler.

Some Virtual Themes are currently not much use in Pattern Assembler and therefore we should skip it to avoid the user experience feeling broken. Themes that fall into this category are those that aren't likely to need headers and footers, or have a strong opinionated design with hard-coded background colours or images such that setting colours and fonts doesn't actually change the homepage.

Given all Virtual Themes skip the PA in flows where it's not available (e.g. one of the free flows) then this _should_ just work, however we might need to be careful to set expectations over which theme the user will have installed afterwards.

dsas:
Added to the current sprint as a low priority on the assumption this is a well-aligned, quick change that we can tackle if we have time. If that's not the case then lets bump into a subsequent sprint.

PR #77752, opened 2023-06-02
Design Picker: Skip flagged virtual themes from going to SA

BogdanUngureanu:
When the design_type property is set on the design and equals "assembler", then we should point the user to the Site Assembler.

This is needed so that we can exclude some Virtual themes from being sent to the Site Assembler. 

Related to https://github.com/Automattic/wp-calypso/issues/77321

## Proposed Changes

* Send the user to Site Assembler only when the `design_type` is set to `assembler`

## Testing Instructions

* Sandbox your site and apply D112585-code
* Use the live link and go to `setup/site-setup/designSetup?siteSlug=your-site`
* Click on the Link in Bio category
* Click on the `Link in Bio: Anime` virtual theme
* Make sure that you're sent to the preview page of the theme (not the Site Assembler)
* Click on another virtual theme and notice that you're sent to the SIte Assembler (only the Anime theme is flagged in the back-end for now - all LiB will be flagged this way once this PR lands).

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
+++ b/packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
@@ -9,6 +9,7 @@ import type {
 	SoftwareSet,
 	StyleVariation,
 	PreviewData,
+	DesignType,
 } from '@automattic/design-picker/src/types';
 
 interface StarterDesignsQueryParams {
@@ -37,6 +38,7 @@ interface StarterDesign {
 	software_sets?: SoftwareSet[];
 	is_virtual: boolean;
 	preview_data: PreviewData | null;
+	design_type?: DesignType;
 }
 
 export function useStarterDesignsQuery(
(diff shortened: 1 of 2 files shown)


packages/data-stores/src/starter-designs-queries/use-starter-designs-query.ts
```diff
@@ -101,7 +104,7 @@ function apiStarterDesignsToDesign( design: StarterDesign ): Design {
 		is_bundled_with_woo_commerce,
 		price,
 		software_sets,
-		design_type: is_premium ? 'premium' : 'standard',
+		design_type: design_type ?? ( is_premium ? 'premium' : 'standard' ),
```
dsas, line 107: I felt wary about changing this because I thought it was adding a new value to design_type here which might throw off some existing stats. Virtual themes that go via the site assembler are already tracked as `design_type: standard` on the grid though and `design_type: assembler` on the assembler page.

PR #71937, opened 2023-01-11
Checkout: hide discount percentage in variant picker if purchase is jetpack

TheSteveK:
#### Proposed Changes

* per 1199980337313591-as-1203595785054343/f and p1671636858945399-slack-C03RCHNVC0H jetpack products should NOT show the price before discount on the variation dropdowns.  This PR hides those sections if the product or plan is Jetpack (WP.com may still wish to display that)
* The code checks the product variant against the isJetpackPlan and isJetpackProduct functions and if true the "original" price and discount percent are not displayed.

Before:

before

After:

after

#### Testing Instructions

* using a jetpack site, add a product with a discount (like videopress) to your cart.  Head over to the Calyspo live link and append /checkout/:site to the URL
* Click the variation dropdown, confirm there's no discount percent pricing showing.  Then, select the other variation.
* Click the drop down again to verify there's still no discount showing.
* Verify wp.com still shows the extra info by adding a wordpress.com plan to a wp.com site's cart.  Repeat the same steps above, but make sure there is a discount displaying

#### Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

Related to #

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - 0-as-1203595785054343

PR #77755, opened 2023-06-02
Update variant dropdown for multiple-year plans to display discounts

IanRamosC:
Related to p1685720048306269-slack-C02LK1W8T4Z

## Proposed Changes

* Add back the discount percentage for Jetpack plans to the variant dropdown (for 2-year and 3-year plans). The idea behind this is to let the users know that the 2-year plans have additional discounts compared to the regular discount from an Intro Offer

## Testing Instructions

* Visit the Calypso live link
* Go to `/checkout/jetpack/jetpack_security_t1_yearly#step2`
* Confirm that the two-year plan displays the discount

Before | After
--- | ---
image | image

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
@@ -134,6 +134,7 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
 	);
 	const productBillingTermInMonths = variant.productBillingTermInMonths;
 	const isIntroductoryOffer = introCount > 0;
+	const shouldRenderJetpackDiscount = isJetpack && introCount > 1;
 	const translate = useTranslate();
 	const billingTermInYears = () => {
 		if ( productBillingTermInMonths > 12 ) {
@@ -244,10 +245,12 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
 				) }
 			</Label>
 			<PriceTextContainer>
-				{ discountPercentage > 0 && ! isJetpack && ! isMobile && ! isIntroductoryOffer && (
-					<DiscountPercentage percent={ discountPercentage } />
-				) }
-				{ discountPercentage > 0 && ! isJetpack && ! isIntroductoryOffer && (
+				{ discountPercentage > 0 &&
+					! isMobile &&
+					( ! isIntroductoryOffer || shouldRenderJetpackDiscount ) && (
+						<DiscountPercentage percent={ discountPercentage } />
+					) }
+				{ discountPercentage > 0 && ( ! isIntroductoryOffer || shouldRenderJetpackDiscount ) && (
 					<DoNotPayThis>{ formattedCompareToPriceForVariantTerm }</DoNotPayThis>
 				) }
 				<Price aria-hidden={ isIntroductoryOffer }>{ formattedCurrentPrice }</Price>
--- a/client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
+++ b/client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
@@ -134,7 +134,6 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
 	);
 	const productBillingTermInMonths = variant.productBillingTermInMonths;
 	const isIntroductoryOffer = introCount > 0;
-	const shouldRenderJetpackDiscount = isJetpack && introCount > 1;
 	const translate = useTranslate();
 	const billingTermInYears = () => {
 		if ( productBillingTermInMonths > 12 ) {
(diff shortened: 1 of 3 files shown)


client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
```diff
@@ -244,10 +245,12 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
 				) }
 			</Label>
 			<PriceTextContainer>
-				{ discountPercentage > 0 && ! isJetpack && ! isMobile && ! isIntroductoryOffer && (
-					<DiscountPercentage percent={ discountPercentage } />
-				) }
-				{ discountPercentage > 0 && ! isJetpack && ! isIntroductoryOffer && (
+				{ discountPercentage > 0 &&
+					! isMobile &&
+					( ! isIntroductoryOffer || shouldRenderJetpackDiscount ) && (
```
CodeyGuyDylan: Is there any way we can clear up this logic a bit? I think I understand what we are checking but I think it would be nice to save this to some variable with a comment above explaining it clearly

CodeyGuyDylan (changes requested):
The only reason I am requesting changes is to potentially clean up some hard-to-read logic, but I do have a couple of thoughts:

If we are showing the % off in the dropdown, should we also show it here?
Screenshot 2023-06-02 at 10 11 01 AM

I'm sure this has been thought of, but this all seems very busy to me:
image
If there was any way to shorten it to be more concise, I think that would be great. As a user, I like to do things quickly and I don't think I'd have a good time decoding all this 😅 

That being said, I can't think of a way to shorten it myself, and I do think all the information is necessary, so there may not be much we can do about that

client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
```diff
@@ -244,10 +245,12 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
 				) }
 			</Label>
 			<PriceTextContainer>
-				{ discountPercentage > 0 && ! isJetpack && ! isMobile && ! isIntroductoryOffer && (
-					<DiscountPercentage percent={ discountPercentage } />
-				) }
-				{ discountPercentage > 0 && ! isJetpack && ! isIntroductoryOffer && (
```
jboland88, line 250: Curious question - do we know why Jetpack products were excluded here previously?

sirbrillig:
I have three small requests: 

1. Could you add a "Before" screenshot to complement the "After" screenshot you have in the description? Otherwise it's hard to tell what exactly is changed 
2. Have you checked how this will look at mobile width (unless it doesn't affect that width)? A screenshot of that might be good too.
3. The description of this PR, `Add back the discount for Jetpack plans in the variant drop for plans longer than one-year plans` doesn't mention the most important thing to be recorded: why? I realize there's an Asana link but that requires having a very specific login, could disappear over the years, and (I checked) doesn't explain why this task is being done either.

CodeyGuyDylan:
> Have you checked how this will look at mobile width (unless it doesn't affect that width)? A screenshot of that might be good too.

I can answer this one 😄 , in this PR, the percentage discount is hidden on mobile widths

IanRamosC:
> If we are showing the % off in the dropdown, should we also show it here?

I think we are just avoiding being redundant, because we already mention that in the card title:
image

>That being said, I can't think of a way to shorten it myself, and I do think all the information is necessary, so there may not be much we can do about that

TBH, I don't know either. It's using the existing components, so it'd probably require design input.

client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
IanRamosC: I'll try to. Can't promise anything. :D 
Due to the nature of each condition, it's a bit hard to simplify the logic here.

client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
IanRamosC: Yep, it was a design request because we were avoiding displaying the discount and old price in other places. Context: p1671636858945399-slack-C03RCHNVC0H

IanRamosC:
I appreciate your input/feedback, @sirbrillig. I added more context to the PR description with your suggestions. I'm also replying to each item individually below.

> I have three small requests:
> 
> 1. Could you add a "Before" screenshot to complement the "After" screenshot you have in the description? Otherwise it's hard to tell what exactly is changed

👍🏾 

> 3. The description of this PR, `Add back the discount for Jetpack plans in the variant drop for plans longer than one-year plans` doesn't mention the most important thing to be recorded: why? I realize there's an Asana link but that requires having a very specific login, could disappear over the years, and (I checked) doesn't explain why this task is being done either.

The idea is to display the additional discount that users are getting (on top of the first-year discount) if they choose the two-year plan. I started this thread p1685720048306269-slack-C02LK1W8T4Z to discuss it a bit further, in case we think we have too much information in that dropdown.

client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
sirbrillig: A related link: https://github.com/Automattic/wp-calypso/pull/71937

sirbrillig:
Thank you for the extra info! What to display in these options has been debated for a long time (see pbOQVh-24K-p2 for some history) and it's definitely a struggle between not visually overwhelming the user with prices and percentages and explanations while at the same time accurately explaining what they will pay. I like to keep track of these changes and their effects since it's easy to lose track of things we've done before.

elliottprogrammer (approved):
@CodeyGuyDylan's feedback regarding cleaning up the logic looks like it has been addressed and everything looks good here to me.
LGTM! 👍

IanRamosC:
> Thank you for the extra info! What to display in these options has been debated for a long time (see pbOQVh-24K-p2 for some history) and it's definitely a struggle between not visually overwhelming the user with prices and percentages and explanations while at the same time accurately explaining what they will pay. I like to keep track of these changes and their effects since it's easy to lose track of things we've done before.

Thanks for letting us know. I'll keep that in mind whenever I review a PR that touches the dropdown. Also, I appreciate your efforts to keep that clean.

PR #74392, opened 2023-03-13
Add page for expired eCommerce trials

daledupreez:
Related to:
* Automattic/wc-calypso-bridge#1005
* #74387

## Proposed Changes

* This change implements a new page for expired eCommerce trials
* For now, this doesn't include the header image, but the PR does include all the recommended copy and styles beyond that, including mobile styles

## Testing Instructions

* Run this branch locally or via the Calypso.live branch
* Ensure you can inspect Tracks events (e.g. by using Tracks Vigilante)
* For an eCommerce trial site without any purchases other than the free eCommerce trial, navigate to `/plans/my-plan/trial-expired/:siteSlug`
* Verify that you see the trial expired page, and you don't see a link to manage your previous purchases
* Verify that we logged a `calypso_page_view` event for the page load
* Compare the layout in desktop and mobile screens to the designs in lD0gTB4IqGsmSmsnjZNaP6-fi-2232%3A21618, and the copy matches the content from p1678485569139339/1678450383.291469-slack-C03Q87XT1QF
* Verify that toggling the interval works
* Verify that clicking on the CTAs adds the appropriate product to your cart, and triggers a `calypso_wooexpress_expired_trial_upgrade_cta_clicked` Tracks event with the `location` event prop set to `footer` or `main-price-card`
* Now make a purchase on the site, such as a third party premium theme subscription
* Refresh the page, and verify that you see the "Manage your previous purchases" link
* Verify that clicking on the link takes you to manage your purchases for the current site

### Screenshots

Screenshot 2023-03-15 at 18 23 21
Screenshot 2023-03-15 at 18 23 56
Screenshot 2023-03-15 at 18 24 19
Screenshot 2023-03-15 at 18 24 33
Screenshot 2023-03-15 at 18 24 54
Screenshot 2023-03-15 at 18 25 06

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
[...]

PR #77758, opened 2023-06-02
Fix conflicting interval toggle styles in plans pages

ddc22:
Closes Automattic/martech#1522
## Proposed Changes

- Moves styled component related styles for the segment control component into sass
- Cleans up any other plans related segment control sass dead code 

## Additional Context

TLDR; accessing the same element (.segment-control) via multiple approaches (SCSS, Emotion) can lead to a nasty, intermittently visible bug, which is hard to diagnose and solve. It happens when both approaches match the same specificity.

---

This bug used to appear intermittently. It appeared because of two conflicting css transpilations fighting for the same specificity. For example, if one is consuming any component, if that said component has implemented its styles using a `.scss` file and then if we use `Styled Component child selectors` to modify the said component's styles, this can lead to said bug. For example ...

https://github.com/Automattic/wp-calypso/blob/7d5a48bbeb604774bb6c2d0b4fa6df7c840268a2/client/my-sites/plans-features-main/components/plan-type-selector.tsx#L337-L396

And in general styled components and sass approaches when mixed to apply styles to the same elements. Can lead to this bug. The bug that we faced with this issue was intermittent, it appeared and disappeared at times. It was incredibly hard to diagnose and solve. The reason for this is `SASS` and `Styled Component` both have a compile step which renders **_internal css_** onto the page. If by chance they both end up targeting the same specificity of styles for a given element, then the order of precedence for the `<style />`  tag is what will decide which rule wins. And the transpilation order of the <style /> tag cannot be guaranteed.

### So how do we solve this?

There are three approaches 
1 - Move all the `child selectors` shown above to sass. ( This PR )
2 - Move the Segment Control sass styles to `styled components` ( Ignored since would lead to refactors effecting many other components )
3 - Leave everything as it is and get a higher specificity in the Style Component selector (Will fix the issue short term but has the potential to reintroduce the bug)

I think we should be careful when mixing these approaches. Generally there wont be any issues but problems can arise if we are targeting the same element with both approaches and specificity becomes equal. And those bugs are build order dependent and therefore extremely hard to investigate. I think generally its a good practice to avoid targeting the same element with these two different approaches.

## Testing Instructions

- Go to each of the following and confirm the interval toggle matches the one from Figma (xkhhUlDowF13dRoXJqlRDM-fi-1_95):
    - `/setup/link-in-bio/plans` to test a Stepper flow with plans
    - `/start/plans` to test Signup flow with plans
    - `/plans[free]` to test Admin with plans

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/plans/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/plans/style.scss
@@ -125,48 +125,6 @@
 			text-decoration: underline;
 		}
 
-		.plans-features-main {
-			ul.segmented-control {
-				background-color: #f2f2f2;
-				border-radius: 6px;  /* stylelint-disable-line scales/radii */
-				border-color: var(--studio-gray-5);
-
-				li.segmented-control__item {
-					border: 6px;
-					padding: 2px;
-					&:first-child {
-						margin-right: 5px;
-					}
-					&.is-selected {
-						a.segmented-control__link {
-							padding-top: 6px;
-							padding-bottom: 6px;
-							border: 0.5px solid #0000000a;
-							background-color: var(--studio-white);
-							color: var(--color-text);
-							border-radius: 5px;  /* stylelint-disable-line scales/radii */
-							box-shadow: 0 3px 8px #0000001f, 0 3px 1px #0000000a;
-						}
-					}
-					a.segmented-control__link {
-						border: solid 1px #f2f2f2;
-						border-radius: 5px;  /* stylelint-disable-line scales/radii */
-					}
-				}
-				li.segmented-control__item:not(.is-selected) {
-					a.segmented-control__link {
-						border: solid 1px #f2f2f2;
-						border-radius: 5px;  /* stylelint-disable-line scales/radii */
-						&:hover {
-							background-color: unset !important;
-							border: 1px solid var(--studio-gray-10);
-							border-radius: 5px;  /* stylelint-disable-line scales/radii */
-						}
-					}
-				}
-			}
-		}
-
 		.plans__loading {
 			width: 100%;
 			display: flex;
--- a/client/my-sites/plan-features-2023-grid/style.scss
+++ b/client/my-sites/plan-features-2023-grid/style.scss
@@ -728,58 +728,6 @@ body.is-section-signup.is-white-signup,
 	}
 }
 
-.plans-features-main.is-pricing-grid-2023-plans-features-main  ul.segmented-control.price-toggle,
-body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup__step .segmented-control.price-toggle,
-#primary .is-pricing-grid-2023-plans-features-main .segmented-control.price-toggle {
-	background-color: #f2f2f2;
-	width: 100%;
-	margin: 0 auto;
-
-	@include plans-2023-break-small {
-		max-width: fit-content;
-		width: auto;
-	}
-
-	.segmented-control__link {
-		padding: 6px 11px;
-	}
-
-	li.segmented-control__item:not(.is-selected) {
-		border: 6px;
-		padding: 2px;
-		font-weight: 500;
-
-		& a.segmented-control__link {
-			border: 1px solid #f2f2f2;
-			border-radius: 5px; /* stylelint-disable-line scales/radii */
-
-			&:hover {
-				border-color: rgba(0, 0, 0, 0.04);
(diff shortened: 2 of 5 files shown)


ddc22:
Closed in favour of #77776

chriskmnds:
@jdc91 https://github.com/Automattic/wp-calypso/pull/77776 is what I had on my end. I approached this from the point of migrating everything to the `IntervalTypeToggleWrapper` component (and whatever remains, if any, should be things specific that can't be done from inside the component - outside of a media query, there shouldn't be any?).

You can take whatever may be useful from there and reuse here, as this looks more complete. (please don't close your PR). I think we should be explicit on why there needs to be both a styled component (with some styles) and additional SASS since we are "fixing" this once and for all.

Thanks for working on this! I have a lot on my plate with open PRs and can move focus elsewhere 🙇

aneeshd16 (changes requested):
> There are three approaches
1 - Move all the child selectors shown above to sass. ( This PR )

This approach sounds good to me! Great start!

On `/setup/newsletter/plans`, I see a regression:

| Production | This PR |
|--------|--------|
|image|image| 

The size of the toggle, colours, margins and hover colour is different. I'm not sure but this could be because of the styles in this file: https://github.com/Automattic/wp-calypso/blob/55be137de83f012093a92fe4f7878d1990e03b02/client/landing/stepper/declarative-flow/internals/steps-repository/plans/style.scss#L129

There are also some styles in https://github.com/Automattic/wp-calypso/blob/4910db94dde21aced9111956d87aae3389df408d/client/my-sites/plans-features-main/components/plan-interval-selector/style.scss which I feel can be removed/consolidated so that all styles for the toggle come from a single place.

ddc22:
Thanks for pointing out the stepper styles 👍  I have fixed them as well.
image

---

> There are also some styles in https://github.com/Automattic/wp-calypso/blob/4910db94dde21aced9111956d87aae3389df408d/client/my-sites/plans-features-main/components/plan-interval-selector/style.scss which I feel can be removed/consolidated so that all styles for the toggle come from a single place.

You are correct the change at #74392, should probably use this component 🤔 . I am planning to pick it up in a subsequent PR. @daledupreez is there a reason to introduce a separate plan selector component for eCommerce trials  in #74392 ? Or can we possibly use the existing component?

aneeshd16 (approved):

daledupreez:
> You are correct the change at https://github.com/Automattic/wp-calypso/pull/74392, should probably use this component 🤔 . I am planning to pick it up in a subsequent PR. @daledupreez is there a reason to introduce a separate plan selector component for eCommerce trials in https://github.com/Automattic/wp-calypso/pull/74392 ? Or can we possibly use the existing component?

The second component includes some discount text inline:
Screenshot 2023-06-07 at 12 48 22

That was something the Woo Express team specifically wanted, as we default the display to monthly for these plans given the price tags, and we want to provide a clear signal that longer-term plans offer notable savings over the monthly options.

So we _could_ consider refactoring them into a single component, but I can't recall how much divergence there was between the two toggles beyond that aspect.

ddc22:
I have created a task to track the other task - Automattic/martech#1821 @aneeshd16

chriskmnds:
I would have requested changes here I'm afraid. :/ 

I don't think we've had _three_ options to reason with here. I think we only had one (and not exactly mentioned). The goal should have been to make this interval toggle independent of environment, with a default state that is known and is what we expect (cause there's only one that is expected of this component at any point of use). To put it differently, [these](https://github.com/Automattic/wp-calypso/pull/77758/files#diff-bad007f09f6802c0aa05c2c5a3215beb51a3d277e86941257d6987142d4954deR2-R4) context-dependent rules should have been removed, and the styles/overrides done within the component's own context (not the environment's). 

Right now we don't even know what the default state is of the respective component. It makes our job for NPM migration not any easier than it was before. Once again, I feel we are patching at the surface, not fixing at the root (I think we've had a similar conversation recently when attempting to fix the grid responsiveness). A grid package will inevitably export an interval selector, just like the original grid package did ([here](https://github.com/Automattic/wp-calypso/tree/trunk/packages/plans-grid/src/plans-interval-toggle)).

It's not hard goal to think in terms of packages, and the sooner we embody that thinking into our workflows, the sooner we'll be doing a much better job at refactors. When we care about self-contained and isolated code (we will always do as long as I'm around), then might help to think in terms of "what if this was living in a package", or "what would this look like in a Storybook demo" - and we'll be attacking the ugly mess that surrounds the current plans grid (that was largely the result of folks _not_ thinking this way) way more effectively.

PR #77771, opened 2023-06-02
Import onboarding flow: Mark the return type of the `getUrlData` selector as nullable

zaguiini:
Follow up for https://github.com/Automattic/wp-calypso/pull/77770.

## Proposed Changes

We're unsafely accessing the `urlData` through the `getUrlData` selector. There is a chance that the data doesn't exist, so we need to type it as `UrlData | null`.

This PR does that and fixes the type errors by either null coalescing where this data is optional, or short-circuiting if the data is mandatory.

## Testing Instructions

The import flows should work normally, as should every other onboarding flow that uses `siteCreationStep`. There are no wrong usages left, we're just making sure that this won't happen again.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/importer/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/importer/index.tsx
@@ -89,7 +89,7 @@ export function withImporterWrapper( Importer: ImporterCompType ) {
 		useEffect( checkFromSiteData, [ fromSiteData?.url ] );
 		useEffect( () => onComponentUnmount, [] );
 
-		if ( ! importer ) {
+		if ( ! importer || ! fromSiteData ) {
 			stepNavigator.goToImportCapturePage?.();
 			return null;
 		}
@@ -161,7 +161,7 @@ export function withImporterWrapper( Importer: ImporterCompType ) {
 		/**
 	 	↓ Renders
 		 */
-		function renderStepContent() {
+		const renderStepContent = () => {
 			if ( isLoading() ) {
 				return <LoadingEllipsis />;
 			} else if ( ! siteSlug || ! site || ! siteId ) {
@@ -188,7 +188,7 @@ export function withImporterWrapper( Importer: ImporterCompType ) {
 					showConfirmDialog={ ! isMigrateFromWp }
 				/>
 			);
-		}
+		};
 
 		return (
 			<>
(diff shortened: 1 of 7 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/importer/index.tsx
```diff
@@ -83,7 +83,7 @@ export function withImporterWrapper( Importer: ImporterCompType ) {
 			}
 		}, [ siteId, site ] );
 
-		if ( ! importer ) {
+		if ( ! importer || ! fromSiteData ) {
```
ouikhuan: We can't have this here because this is where we might have users come directly. For example, we have our progress screen and plans page here where users might come directly if they dropped off the flow earlier. `http://calypso.localhost:3000/setup/import-focused/importerWordpress?siteSlug=${target_site}&from=${source_site}&option=everything`. We'll [fetch the site data ](https://github.com/Automattic/wp-calypso/blob/3b914914177a418dacef236015ae0597e2f79d99/client/landing/stepper/declarative-flow/internals/steps-repository/importer/index.tsx#L79)if there's no data fetched yet here. If we add the check here, it'll always be redirected to the site address URL input page before it tries to fetch the site data.

ouikhuan (changes requested):

client/landing/stepper/declarative-flow/internals/steps-repository/importer/index.tsx
zaguiini: Right. What's the appropriate place for checking, then? Right before usage? What should we do if there is no `fromSiteData`? Is this data optional?

ouikhuan: We can remove the check here and check it before usage if needed. We do have some other pages that don't require the `fromSiteData`. For example, `https://wordpress.com/setup/import-focused/importerWordpress?siteSlug=${target_site}&from=&option=content` which allows you to import the content to your destination site by uploading a xml file. We also allow users to upload xml files from other platform like Blogger, Medium etc.

zaguiini:
Alright @ouikhuan please give it another round of review 😄 You can merge and deploy the PR if you approve the changes 👍

ouikhuan (approved):
LGTM!

ouikhuan (approved):
LGTM!

PR #77776, opened 2023-06-03
Plans 2023: Fix interval toggle

chriskmnds:
Related to https://github.com/Automattic/martech/issues/1522

## Proposed Changes

Updates the interval toggle to have consistent styles across usages in Stepper/Signup/Admin.

### TODO

- [ ] remove sensei flow styles.
- [ ] migrate the few remaining SASS styles to the styled component
- [ ] some cleanup

## Testing Instructions

- Go to each of the following and confirm the interval toggle matches the one from Figma (xkhhUlDowF13dRoXJqlRDM-fi-1_95):
    - `/setup/link-in-bio/plans`
    - `/start/plans`
    - `/plans[free]`

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/plan-features-2023-grid/style.scss
+++ b/client/my-sites/plan-features-2023-grid/style.scss
@@ -717,8 +717,43 @@ body.is-section-signup.is-white-signup,
 }
 
 .plans-features-main.is-pricing-grid-2023-plans-features-main  ul.segmented-control.price-toggle,
-body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup__step .segmented-control.price-toggle,
-#primary .is-pricing-grid-2023-plans-features-main .segmented-control.price-toggle {
+body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup__step ul.segmented-control.price-toggle,
+#primary .is-pricing-grid-2023-plans-features-main ul.segmented-control.price-toggle {
+	background-color: #f2f2f2;
+	border-radius: 6px;  /* stylelint-disable-line scales/radii */
+
+	@include plans-2023-break-small {
+		max-width: fit-content;
+		width: auto;
+	}
+
+	li.segmented-control__item {
+		&.is-selected {
+			a.segmented-control__link {
+				padding-top: 6px;
+				padding-bottom: 6px;
+				border: 0.5px solid #0000000a;
+				background-color: var(--studio-white);
+				color: var(--color-text);
+				border-radius: 5px;  /* stylelint-disable-line scales/radii */
+				box-shadow: 0 3px 8px #0000001f, 0 3px 1px #0000000a;
+			}
+		}
+		a.segmented-control__link {
+			border: solid 1px #f2f2f2;
+			border-radius: 5px;  /* stylelint-disable-line scales/radii */
+
+			&:focus {
+				outline: none;
+			}
+		}
+	}
+}
+
+// .plans-features-main.is-pricing-grid-2023-plans-features-main  ul.segmented-control.price-toggle,
+// body.is-section-signup.is-white-signup .is-onboarding-2023-pricing-grid .signup__step .segmented-control.price-toggle,
+// #primary .is-pricing-grid-2023-plans-features-main .segmented-control.price-toggle
+.foo {
 	background-color: #f2f2f2;
 	width: 100%;
 	margin: 0 auto;
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/plans/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/plans/style.scss
@@ -125,48 +125,6 @@
 			text-decoration: underline;
 		}
 
-		.plans-features-main {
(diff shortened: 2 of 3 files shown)


client/my-sites/plan-features-2023-grid/style.scss
```diff
@@ -717,8 +717,43 @@ body.is-section-signup.is-white-signup,
...
+		}
+		a.segmented-control__link {
+			border: solid 1px #f2f2f2;
+			border-radius: 5px;  /* stylelint-disable-line scales/radii */
+
+			&:focus {
+				outline: none;
+			}
+		}
+	}
```
chriskmnds, line 750: These should move to the styled component and there shouldn't be any left in the SASS files. cc @jdc91

ddc22: I was thinking maybe we should move everything to SASS because the bugs occur when style components and sass try to compete for specificity and sometimes order can conflict.

The **segment control component** is based on sass and whatever the approach followed by the component library should be followed by the client (The consumer of the component).

https://github.com/Automattic/wp-calypso/blob/19f3e85f4094222aead702a39ecb87d3d8ccc338/packages/plans-grid/src/segmented-control/index.jsx#L5-L9

Now we could move the styles in the library component(segment control) to styled component, but that would explode the complexity of the refactor because other components already have used this component extensively.

We could keep going with a hybrid approach and try to fix this bug, but in the future there is a high likelihood that these types of bugs could occur in production but still could look good in DEV environments.

This is why I have moved all the styles that are attacking the same specificity in multiple places and encapsulated them into a [sass module in one location](https://github.com/Automattic/wp-calypso/pull/77758/files#diff-bad007f09f6802c0aa05c2c5a3215beb51a3d277e86941257d6987142d4954de).

https://github.com/Automattic/wp-calypso/pull/77758

Feel free to extract learnings from the above PR 👍

chriskmnds: but isn't this a problem in general with having both styled components _and_ CSS styles?

> The segment control component is based on sass and whatever the approach followed by the component library should be followed by the client (The consumer of the component).

Not sure I follow this rule. A styled component can't import other components without an accompanying CSS file? (cause this is what you are saying) 🤔 

> that would explode the complexity of the refactor because other components already have used this component extensively

Not sure why we need to care about this or why it would explode the refactor. There is only one section left to migrate into the styled component in this PR (but it needs more specificity to override the parent's).

To understand the problem - we own and render the `IntervalToggleWrapper`, which defines its own styles. Why does it matter if other components are using the imported `SegmentedControl`?

ddc22: TLDR; accessing the same element (.segment-control) via multiple approaches (SCSS, Emotion) can lead to a nasty, intermittently visible bug, which is hard to diagnose and solve. It happens when both approaches match the same specificity.

---

> A styled component can't import other components without an accompanying CSS file?

Nope sorry I was not clear above. What I meant was if one is consuming any component, if that said component has implemented its styles using a `.scss` file. We should not use `Styled Component child selectors` to modify the child component styles. For example ...

https://github.com/Automattic/wp-calypso/blob/7d5a48bbeb604774bb6c2d0b4fa6df7c840268a2/client/my-sites/plans-features-main/components/plan-type-selector.tsx#L337-L396

And in general styled components and sass approaches should not be mixed to apply styles to the same elements. Because this can lead to bugs. 

If you can remember the initial bug was intermittent, it appeared and disappeared at times. It was incredibly hard to diagnose and solve.

The reason for this is `SASS` and `Styled Component` both have a compile step which renders **_internal css_** onto the page. If by chance both end up with the same specificity of styles for a given element, then the order of precedence for the `<style />`  tag wins. And the  transpilation order of the <style /> tag cannot be guaranteed.

So how do we solve this?

There are three approaches 
1 - Move all the `child selectors` shown above to sass. (Which is what I have done)
2 - Move the Segment Control sass styles to `styled components`
3 - Leave everything as it is and get a higher specificity in the Style Component selector (I think this is what this PR leads to eventually but do let me know if I am wrong)

The only problem with approach 3 is that if someone else does a change where for a given element the specificity between styled components and sass becomes equal. This again resurfaces the hard to reproduce intermittent bug I explained above.

> Not sure why we need to care about this or why it would explode the refactor. 

I was specifically talking about approach 2. The only additional advantage with approaches 2 ( and 1 ) is that we reduce the chance of this intermittent bug happening by bringing everything into one realm and possibly keeping the developer in one realm. And the exploding complexity was if we went with approach 2 :) Which can be ignored completely.

> but isn't this a problem in general with having both styled components and CSS styles?

Yes I guess so. This is why I think we should be careful when mixing these approaches. Generally there wont be any issues but problems can arise if we are targeting the same element with both approaches and specificity becomes equal. And those bugs are build order dependent and therefore extremely hard to investigate. I think generally its a good practice to avoid targeting the same element with these two approaches. But I am looking forward to your take on it as well !

---

I was actually thinking about `CodeP2` ing this, because it was an extremely tough bug and we lost a lot of dev hours to this in the early days of this change. WDYT?

PR #77779, opened 2023-06-05
Revert "ESLint: Enable `@tanstack/eslint-plugin-query` (#77214)"

roo2:
see p1685923446140549-slack-C7YPUHBB2

It appears that the update to the linter works on local machines but causes linting of the newspack-blocks project to fail in CI which was responsible for replacing the translation text domains. See example diff generated D112800-code


tyxla:
For posterity: @flootr opened a take 2 in https://github.com/Automattic/wp-calypso/pull/77787, and we're working to address the root problem.

PR #77780, opened 2023-06-05
Add /setup/blog binary selection screen for blog onboarding

lsl:
Fixes https://github.com/Automattic/dotcom-forge/issues/2141

## Proposed Changes

* Adds binary selection page for blog onboarding 

## Testing Instructions

* http://calypso.localhost:3000/setup/blog
* ~Logged out should force a login or user creation before loading the page, both flows should return back to this screen~
* Clicking on one of the buttons should move you to the editor, design picker, or login screen before hand as required.

Screenshot

Screenshot 2023-06-05 at 18-01-00 Create your blog — WordPress com

- [ ] Mobile/small screen breakpoints need work
- [x] Wordpress logo missing?
- [x] Figma icons needed

### DIFF
--- a/client/landing/stepper/declarative-flow/blog.ts
+++ b/client/landing/stepper/declarative-flow/blog.ts
@@ -1,5 +1,13 @@
+import { useLocale } from '@automattic/i18n-utils';
 import { translate } from 'i18n-calypso';
-import { Flow } from 'calypso/landing/stepper/declarative-flow/internals/types';
+import { redirect } from 'calypso/landing/stepper/declarative-flow/internals/steps-repository/import/util';
+import {
+	AssertConditionResult,
+	AssertConditionState,
+	Flow,
+} from 'calypso/landing/stepper/declarative-flow/internals/types';
+import { useSelector } from 'calypso/state';
+import { getCurrentUserSiteCount, isUserLoggedIn } from 'calypso/state/current-user/selectors';
 
 const Blog: Flow = {
 	name: 'blog',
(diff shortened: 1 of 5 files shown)


lsl:
Could I get a quick approach check? Not really sure if this is the expected way we want to achieve this or not. There was some kind of intents gathering steps in the existing code but I couldn't figure out how / or if it was relevant.

candy02058912:
> Could I get a quick approach check? Not really sure if this is the expected way we want to achieve this or not.

I initially thought of creating a simple page that linked to `/setup/start-writing` or `/setup/design-first` and add a check to see if the user is logged in. So this is a fresh new take on this. Would like to know how the other @Automattic/explorers think.

> There was some kind of intents gathering steps in the existing code but I couldn't figure out how / or if it was relevant.

The intents were used to keep the state of the editor/launchpad which I don't think we would need that for this step.

client/landing/stepper/declarative-flow/internals/steps-repository/blogger-intent/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+			<StepContainer
+				stepName="intent"
+				hideBack={ true }
+				hideSkip={ true }
+				hideNext={ true }
+				showJetpackPowered={ true }
+				stepContent={
+					<div className="blogger-intent__container">
+						<h1 className="blogger-intent__heading">
+							{ translate( "Let's start your blog, %(username)s", {
```
paulopmt1: Should we add the exclamation point?

image

client/landing/stepper/declarative-flow/blog.ts
```diff
@@ -0,0 +1,75 @@
...
+			redirect( `/` );
+			result = {
+				state: AssertConditionState.CHECKING,
+				message: `${ flowName } requires no preexisting sites`,
+			};
+		}
+
+		return result;
+	},
+};
```
paulopmt1, line 73: Not sure we need to check/redirect user here since we already do it inside of each flow.

paulopmt1:
Thanks for working on it @lsl!

> I initially thought of creating a simple page

I thought the same @candy02058912, but maybe this strategy can help us add more steps if we want in the future with a little more complexity. I would remove the extra checks we already do inside the flows, though.

lsl:
Main thing was just needing that username and at the time me not being sure how the flows fit together / if they should be merged (I don't think they should be).

I'll strip out the username / redirects and simplify the step before deploying later today.

lsl:
Testing this, I don't know if I like how this says "Start writing" on the button and then drops you on another screen that says "lets get started" asking you to sign up right after clicking "start writing".

Leaning towards reverting that last commit and shipping so I can work on the styling. Applying [26cb9d6](https://github.com/Automattic/wp-calypso/pull/77780/commits/26cb9d62f561c22c8f9f6a5f23aa8b855110e582) again later shouldn't be too much trouble.

DustyReagan (approved):
Looks good to iterate on! 👍

client/landing/stepper/declarative-flow/blog.ts
lsl: responded below

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8028454

Thank you @lsl for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8028454).

PR #77782, opened 2023-06-05
Enable search page for logged out users

roo2:
see pe7F0s-Ti-p2

* Makes the `/read/search` page public
* Updates its style and header to match the `/tag/$tag` pages
* Uses trending tags rather than subscribed tags to generate recommendations.
* By applying D112875-code, the default search results will be available to logged out users

### Blank state
Screen Shot 2023-06-06 at 2 49 18 pm

### After search
Screen Shot 2023-06-06 at 2 49 35 pm

### Testing instructions
The public version of the page will not be publicly available until we update nginx config for wordpress.com
Therefore it is most important to test that the logged in search has not been affected.
Ensure design, search and suggestions still work as before in the logged in`/read/search` page.

### DIFF
--- a/client/layout/logged-out.jsx
+++ b/client/layout/logged-out.jsx
@@ -76,6 +76,8 @@ const LayoutLoggedOut = ( {
 
 	const isReaderTagPage =
 		sectionName === 'reader' && window.location.pathname.startsWith( '/tag/' );
+	const isReaderSearchPage =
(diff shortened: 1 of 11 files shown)


eoigal (approved):
Nice job, wfm

ebinnion:
I'm not sure that this makes wordpress.com/read/search public. I just get redirected to https://en.search.wordpress.com/?q=undefined. Am I testing incorrectly?

PR #77795, opened 2023-06-05
Analytics: Add prop to GA4 purchase event if purchasing yearly or higher WPcom plan

gmovr:
Related to Automattic/Martech#1754

## Proposed Changes

* When implementing GA4 via Gtag, we are not able to use custom conversions based on the `items` array, only the top-level cart. This is a limitation in Gtag (ref: https://support.google.com/analytics/answer/10085872?hl=en)
* There are two ways around this: we can either implement the whole conversion tracking in GTM, or we can add this prop to the top level of the cart.
* @robertsreberski  -- what do you think of this? It seems a bit hacky, e.g. if we want to add other conditions, and it's WPcom specific, and does not make sense for Akismet or Jetpack -- but it's important for a conversion test we're doing.

Screenshot 2023-06-05 at 13 28 37

## Testing Instructions
* Ensure ad-tracking is enabled, either via a flag `?flags=ad-tracking` or by changing the values in your local `development.json` file.
* Store sandbox
* Either via the Google Analytics 4 debug view, or by setting a breakpoint before `recordOrderInWPcomGA4` returns, confirm that:
    * After purchasing a plan, the new prop `contains_yearly_or_higher_wpcom_plan` is visible, and it will return `false` for a monthly plan, and `true` for a yearly or higher (e.g. 2 or 3 year) plan. 

## Pre-merge Checklist

### DIFF
--- a/client/lib/analytics/utils/cart-to-ga-purchase.ts
+++ b/client/lib/analytics/utils/cart-to-ga-purchase.ts
@@ -62,6 +62,8 @@ export function cartToGaPurchase(
 		currency: 'USD', // we track all prices in USD
 		value,
 		items: items.map( ( product ) => productToGaItem( product, cart.currency ) ),
-		contains_yearly_or_higher_wpcom_plan: containsYearlyOrHigherWPcomPlan,
+		...( 'wpcom' === cartInfoType
+			? { contains_yearly_or_higher_wpcom_plan: containsYearlyOrHigherWPcomPlan }
+			: {} ),
 	};
 }
--- a/client/lib/analytics/utils/cart-to-ga-purchase.ts
+++ b/client/lib/analytics/utils/cart-to-ga-purchase.ts
@@ -9,6 +9,7 @@ export type GaPurchase = {
 	transaction_id: string;
 	coupon: string;
 	items: GaItem[];
+	contains_yearly_or_higher_wpcom_plan?: boolean;
 };
 
 const parseCartInfo = ( cartInfo: WpcomJetpackCartInfo ) => ( {
@@ -45,6 +46,15 @@ export function cartToGaPurchase(
 ): GaPurchase {
 	const cartInfoType = getCartInfoType( cartInfo );
 
+	// When using gtag.js, we can't access the `items` array to make custom events in GA4.
+	// To get around this limitation, we need to set this property on the top level purcahse object.
+	const containsYearlyOrHigherWPcomPlan = cartInfo.wpcomProducts?.some( ( product ) => {
+		if ( ! product.months_per_bill_period ) {
+			return false;
+		}
+		return product.months_per_bill_period >= 12;
+	} );
+
 	const { value, items } = parseCartInfo( cartInfo )[ cartInfoType ];
 	return {
 		transaction_id: orderId,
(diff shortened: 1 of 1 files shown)


gmovr:
Rebase for failing e2e

travisw (approved):
LGTM 🚀 

I see the value as `false` with a monthly plan purchase, and `true` when I purchased 1 year and three year. Just noting also that according to the GA4 debugger there are two purchase events being fired, unless I'm reading it wrong.
I used [this extension](https://chrome.google.com/webstore/detail/analytics-debugger/ilnpmccnfdjdjjikgkefkcegefikecdc/related).

### 1 month

Screenshot 2023-06-09 at 12 45 40 PM

### 1 year

Screenshot 2023-06-09 at 12 46 56 PM

### 3 year

Screenshot 2023-06-09 at 12 48 08 PM

client/lib/analytics/utils/cart-to-ga-purchase.ts
```diff
@@ -45,12 +46,22 @@ export function cartToGaPurchase(
...
+	} );
+
 	const { value, items } = parseCartInfo( cartInfo )[ cartInfoType ];
 	return {
 		transaction_id: orderId,
 		coupon: cart.coupon,
 		currency: 'USD', // we track all prices in USD
 		value,
 		items: items.map( ( product ) => productToGaItem( product, cart.currency ) ),
+		contains_yearly_or_higher_wpcom_plan: containsYearlyOrHigherWPcomPlan,
```
robertsreberski: I would suggest excluding `contains_yearly_or_higher_wpcom_plan` for carts other than wpcom:
```suggestion
		...( 'wpcom' === cartInfoType ? { contains_yearly_or_higher_wpcom_plan: containsYearlyOrHigherWPcomPlan } : {} ),
```

robertsreberski (changes requested):
> @robertsreberski -- what do you think of this? It seems a bit hacky, e.g. if we want to add other conditions, and it's WPcom specific, and does not make sense for Akismet or Jetpack -- but it's important for a conversion test we're doing.

This test makes sense to have at least some quick insight into purchases. The only more future-proof improvement I could think of would be to introduce `int` instead of `boolean` in `contains_yearly_or_higher_wpcom_plan`. If we could do e.g. `yearly_value_wpcom_plan` that would be `0` (for monthly) and then `1`, `2` (for yearly different terms) - we could have access to more specific custom conversions directly from the GA4 panel.

Maybe it's not needed in your case, but I don't think it's harder to implement at this stage or create redundant logic - so I thought I'll mention that.

---

One think I would consider changing is to not add a new property to conversion object for `cartInfoType` = Jetpack or Akismet . Even though it doesn't break anything - it is a bit redundant and makes implementation confusing imo.

client/lib/analytics/utils/cart-to-ga-purchase.ts
gmovr: Thanks! I've opted for just combining yearly and higher (incl 2y and 3y plans) for now, as the "and" filtering in Google isn't great, so this will be cleaner in the Google analytics setup. I've added the changes you suggested though for non-wpcom carts, which makes sense, as the prop is already optional in the type definition. :)

robertsreberski (approved):
LGTM, let's ship and test :shipit:

PR #77801, opened 2023-06-05
Growth: Minimal stepper onboarding flow

chriskmnds:
Closes Automattic/growth-foundations#70

## Proposed Changes

Adds a minimal `/setup/onboarding-pm-dev` flow that should be identical to `/start` -> with "log in", "domains", "plans", "checkout", steps.

### Not covered under this PR

- Update the term toggle to 1y/2y
-  Processing step: the progress bar is not filled-up properly
-  Processing step: the progress bar should attach to the top as https://github.com/Automattic/wp-calypso/pull/77801#discussion_r1226681785 said
-  In the main onboarding flow, there is a Business upsell shown after picking the Premium plan. We should replicate that here.
-  Confirm that is_domain_mapping and is_domain_transfer are properly set for calypso_signup_completion event.

## Testing Instructions

* Go to `/setup/onboarding-pm-dev` and confirm the flow end to end. It should behave identically to `/start`

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
@@ -244,7 +244,7 @@
 // Domain upsell flow
 
 .domain-upsell.domains,
-.onboarding-pm-dev.domains {
+.onboarding-media.domains {
 
 	$domain-stepper-component-width: 1040px;
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
@@ -244,7 +244,7 @@
 // Domain upsell flow
 
 .domain-upsell.domains,
-.onboarding-pm.domains {
+.onboarding-pm-dev.domains {
 
 	$domain-stepper-component-width: 1040px;
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
@@ -244,7 +244,7 @@
 // Domain upsell flow
 
 .domain-upsell.domains,
-.onboarding.domains {
+.onboarding-pm.domains {
 
 	$domain-stepper-component-width: 1040px;
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
@@ -243,7 +243,8 @@
 
 // Domain upsell flow
 
-.domain-upsell.domains {
+.domain-upsell.domains,
+.onboarding.domains {
 
 	$domain-stepper-component-width: 1040px;
 
--- a/client/landing/stepper/declarative-flow/onboarding-pm.ts
+++ b/client/landing/stepper/declarative-flow/onboarding-pm.ts
@@ -1,6 +1,6 @@
 import { UserSelect } from '@automattic/data-stores';
 import { useLocale } from '@automattic/i18n-utils';
-import { ONBOARDING_PM_LOW, useFlowProgress } from '@automattic/onboarding';
+import { ONBOARDING_PM_FLOW, useFlowProgress } from '@automattic/onboarding';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { addQueryArgs } from '@wordpress/url';
 import wpcom from 'calypso/lib/wp';
@@ -16,7 +16,7 @@ import { AssertConditionState, ProvidedDependencies } from './internals/types';
 import type { Flow } from './internals/types';
 
 const onboarding: Flow = {
-	name: ONBOARDING_PM_LOW,
+	name: ONBOARDING_PM_FLOW,
 	useSteps() {
 		return [
 			{
@@ -38,7 +38,7 @@ const onboarding: Flow = {
 		];
 	},
 	useStepNavigation( currentStep, navigate ) {
-		const flowName = ONBOARDING_PM_LOW;
+		const flowName = ONBOARDING_PM_FLOW;
 		const { setStepProgress } = useDispatch( ONBOARD_STORE );
 		const flowProgress = useFlowProgress( { stepName: currentStep, flowName } );
 		const { domain, provider } = useDomainParams();
@@ -125,7 +125,7 @@ const onboarding: Flow = {
 		};
 	},
 	useAssertConditions() {
-		const flowName = ONBOARDING_PM_LOW;
+		const flowName = ONBOARDING_PM_FLOW;
 		const userIsLoggedIn = useSelect(
 			( select ) => ( select( USER_STORE ) as UserSelect ).isCurrentUserLoggedIn(),
 			[]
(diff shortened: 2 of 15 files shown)


client/landing/stepper/declarative-flow/onboarding-pm.ts
```diff
@@ -0,0 +1,150 @@
...
+					return;
+				case 'processing': {
+					// clearSignupDestinationCookie();
+					if ( providedDependencies?.goToCheckout ) {
+						const destination = `/setup/site-setup/goals?siteSlug=${ providedDependencies.siteSlug }&notice=purchase-success`;
+						const returnUrl = encodeURIComponent( destination );
+
+						persistSignupDestination( destination ); // not sure if this is needed
+						setSignupCompleteSlug( providedDependencies?.siteSlug );
+						setSignupCompleteFlowName( flowName );
```
chriskmnds: move out of condition block

client/landing/stepper/declarative-flow/onboarding-pm.ts
```diff
@@ -0,0 +1,150 @@
...
+				slug: 'plans',
+				asyncComponent: () => import( './internals/steps-repository/plans' ),
+			},
+			{
+				slug: 'siteCreationStep',
+				asyncComponent: () => import( './internals/steps-repository/site-creation-step' ),
+			},
+			{
+				slug: 'processing',
+				asyncComponent: () => import( './internals/steps-repository/processing-step' ),
```
chriskmnds, line 36: Processing step may need tuning / review actions that consume the progress bar.

southp:
Summary of the things left that I'm aware of at the moment.

- [ ] Update the term toggle to 1y/2y
- [ ] Processing step: the progress bar is not filled-up properly
- [ ] Processing step: the progress bar should attach to the top as [this comment](https://github.com/Automattic/wp-calypso/pull/77801#discussion_r1226681785) said
- [ ] In the main onboarding flow, there is a Business upsell shown after picking the Premium plan. We should replicate that here.
- [ ] Confirm that `is_domain_mapping` and `is_domain_transfer` are properly set for `calypso_signup_completion` event.

ddc22:
@chriskmnds @southp I have tracked the issues you raised here under https://github.com/Automattic/martech/issues/1882. I will do some testing of this flow to see if the bare minimum functionality works here, so that we can get it merged.

client/landing/stepper/declarative-flow/internals/steps-repository/domains/style.scss
```diff
@@ -243,7 +243,8 @@
 
 // Domain upsell flow
 
-.domain-upsell.domains {
+.domain-upsell.domains,
+.onboarding-pm-dev.domains {
```
chriskmnds:
```suggestion
.onboarding-pm.domains {
```

sounds like

chriskmnds: sorry, the other way. sounds like [this](https://github.com/Automattic/wp-calypso/pull/77801/files#diff-5bc8c2fe5f3c46f9140fb3c093ad37430873fc9825b0bad5f7d1c263bd5b3021R140) should be updated instead

packages/onboarding/src/utils/flows.ts
```diff
@@ -29,6 +29,12 @@ export const UPDATE_DESIGN_FLOW = 'update-design';
 export const DOMAIN_UPSELL_FLOW = 'domain-upsell';
 export const BULK_DOMAIN_TRANSFER = 'bulk-domain-transfer';
 
+// TODO:
+// Once it is finalized, update it to just `onboarding-pm`
+export const ONBOARDING_PM_FLOW = 'onboarding-pm-dev';
```
chriskmnds: I don't fully understand why we rename to `onboarding-pm-dev`. It's still `dev` until we start pointing users to it right? 

won't this create overhead when updating back to `onboarding-pm` - like CSS classes derived from flow's name (that already some are slipping through the cracks)

chriskmnds: don't mean this as any form of blocker btw. I think I can see the value of having this intermediary "dev" flow now - maybe to run queries and analyses over (and forget about once the other ships), without having to keep track of dates and things like that to separate the two later :-)

ddc22: I think The main rational for dev prefix is the twin flow code magic we are doing to maintain flow name consistency.

Some pre context: We already do have in production a `/start/onboarding-pm` flow

https://github.com/Automattic/wp-calypso/blob/397a28df4a1c05737fbeadfc85798ac66338de5e/client/signup/config/flows-pure.js#L142-L150

## The story of the twin flow 
(I think you know and implemented this 😆  so feel free to ignore, I will document it here for posterity)

So we are doing some code magic gymnastic to make the flow look consistent from the analytics side. The main reason we have to do this is because the `user step` is not yet ported into the stepper framework. 

So we have to redirect to the user step in the Onboarding flow in order to get the user logged in.

## Why not just create a flow with  a random name that has only the user step and get it over with?

We need to maintain the consistency of the flow name. So the flow name on stepper and flow name on the signup framework needs to be identical.

So this is where the problem arises where a `/start/onboarding-pm` flow already exists. We cannot change that flow to cater to the needs of the twin flow. And we need to maintain flow name consistency for analytics testing. This is what I understood as the rational for this approach.

Some context can also be found here: p1687350686732589-slack-C059R8S4ALT

@southp Did I get this right?

ddc22 (approved):
The addition of a minimal flow LGTM !

The issues mentioned at https://github.com/Automattic/wp-calypso/pull/77801#issuecomment-1595655578 has been moved to Automattic/martech#1882.

🚢

chriskmnds:
> The addition of a minimal flow LGTM !
> 
> The issues mentioned at [#77801 (comment)](https://github.com/Automattic/wp-calypso/pull/77801#issuecomment-1595655578) has been moved to [Automattic/martech#1882](https://github.com/Automattic/martech/issues/1882).
> 
> 🚢

Nothing to address [here](https://github.com/Automattic/wp-calypso/pull/77801#discussion_r1246196679) ?

packages/onboarding/src/utils/flows.ts
southp: @jdc91 Yeah, you got it right :) 

However, to rethink about that we can also just come up with a name that's not `onboarding-pm` to avoid a follow-up renaming; just that somehow this name "onboarding-pm" sticks to me the most somehow. It'd also be safer to deploy, since there will be a time gap between deploying the new paid media flow and updating the link in the paid media ads. If we just replace the current `/start/onboarding-pm` by the twin flow, there will be a period of time that the ads links to the "broken" twin flow. 

I previously thought a way of addressing this is to use an explicit redirect destination like this: https://github.com/Automattic/wp-calypso/pull/77958/files#diff-004d43d24ca95a3f3a3b10e34e557ebc0348147c8c806c7c311b7909385e0a3aR47 . That way, even if the users start from `/start/onboarding-pm`, the new paid media flow will still work. All these tricky things won't be necessary though if we come up with a better name than `onboarding-pm`. 

@chriskmnds @jdc91 The more I think about it, the more I think we should just rename it. Does that make sense? If so, what will be the name? e.g. `/setup/start-pm`, `/setup/pm`, or `/setup/onboarding-media`, etc.?

chriskmnds: `/setup/onboarding-media` sounds nice :-)

southp: Cool. I've updated it as https://github.com/Automattic/wp-calypso/pull/77801/commits/c557edc2d520f6253509899b16d0e8b8a6c1fe1d, https://github.com/Automattic/wp-calypso/pull/77801/commits/f8604020bb071aed61c506e302583937e71bdfb6 and https://github.com/Automattic/wp-calypso/pull/77801/commits/1d7b08fe4aae16cedfdef6e2c515574e859ff067. I'll also be deploying this soon since they should be major enough for another round of review. Saying all that, we can of course still follow up with any concern that we might have; it's not public facing anyway :)

PR #77803, opened 2023-06-05
Help Center AI quick answers: log the answer source in tracks

yoavf:
While testing Docsbot, answers generated by docsbot will contain a "source" key. If the key isn't set, then the answer comes from Sitebot.

This logs the source in thumbs up/down events, and in the `calypso_helpcenter_show_gpt_response` tracks event.

## Testing Instructions

* Apply D112433-code (and activate a variation as per the instructions there)
* Open the Help Center, and click on "Still need help"
* Choose any contact method
* Enter your message and click on "Continue"
* After the answer is loaded, click thumbs up/down
* Verify in the dev console that tracks requests for the events `calypso_helpcenter_gpt_response_thumbs_up`, `calypso_helpcenter_gpt_response_thumbs_down` and `calypso_helpcenter_show_gpt_response` contain an `answer_source` property

## Pre-merge Checklist

### DIFF
--- a/packages/data-stores/src/queries/use-jetpack-search-ai.ts
+++ b/packages/data-stores/src/queries/use-jetpack-search-ai.ts
@@ -24,6 +24,7 @@ export type JetpackSearchAIResult = {
 	step: string;
 	urls: AIResponseURL[];
 	terms: string[];
+	source: null | string;
 };
 
 export function useJetpackSearchAIQuery( config: JetpackSearchAIConfig ) {
--- a/packages/data-stores/src/queries/use-jetpack-search-ai.ts
+++ b/packages/data-stores/src/queries/use-jetpack-search-ai.ts
@@ -24,7 +24,7 @@ export type JetpackSearchAIResult = {
 	step: string;
 	urls: AIResponseURL[];
 	terms: string[];
-	source: null | string;
+	source: string;
 };
 
 export function useJetpackSearchAIQuery( config: JetpackSearchAIConfig ) {
@@ -48,5 +48,8 @@ export function useJetpackSearchAIQuery( config: JetpackSearchAIConfig ) {
 		keepPreviousData: false,
 		enabled: config.enabled && !! config.query,
 		retry: false,
+		select: ( data ) => {
+			return { ...data, source: data.source ?? 'sitebot' };
+		},
 	} );
 }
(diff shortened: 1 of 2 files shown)


packages/data-stores/src/queries/use-jetpack-search-ai.ts
```diff
@@ -24,6 +24,7 @@ export type JetpackSearchAIResult = {
 	step: string;
 	urls: AIResponseURL[];
 	terms: string[];
+	source: null | string;
```
klimeryk: To avoid duplicating the `data?.source | 'sitebot'` checks below, you could add [a `select` callback to the `useQuery` call](https://tanstack.com/query/v4/docs/react/reference/useQuery) below and make sure the `source` field is set correctly. Then you could even adjust the type here to be always `string`.

Example from a different part of Calypso code:
https://github.com/Automattic/wp-calypso/blob/61d721e66a191447821a02c29684cf65772d67ee/packages/data-stores/src/support-queries/use-support-activity.ts#L26

Props to @alshakero for pointing out this feature of `useQuery`.

Something like this:
```
select: ( data ) => {
	return { ...data, source: data.source ?? 'sitebot', };
},
```

packages/help-center/src/components/help-center-gpt.tsx
```diff
@@ -101,17 +102,19 @@ export function HelpCenterGPT() {
...
 	} );
 
-	const doThumbsUp = () => {
+	const doThumbsUp = ( source: string ) => {
 		setFeedbackGiven( true );
 		recordTracksEvent( 'calypso_helpcenter_gpt_response_thumbs_up', {
 			location: 'help-center',
+			answer_source: source,
 		} );
 	};
```
klimeryk: If you want to be able to use it as previously as a callback:
```suggestion
	const doThumbsUp = ( source: string ) => {
		return () => {
			setFeedbackGiven( true );
			recordTracksEvent( 'calypso_helpcenter_gpt_response_thumbs_up', {
				location: 'help-center',
				answer_source: source,
			} );
		};
	};
```
Then it could be used as `<Button onClick={ doThumbsUp( data?.source ) }>&#128077;</Button>`

klimeryk (approved):
Tested both variations: showing the result and upvoting. All worked as expected. Left two code suggestions, but not blockers, just up to you if you want to apply them.

packages/help-center/src/components/help-center-gpt.tsx
yoavf: This makes so much sense, thanks!

packages/data-stores/src/queries/use-jetpack-search-ai.ts
yoavf: TIL about select. Awesome, this change really simplifies the code!

PR #77824, opened 2023-06-06
Editor Domain Upsell: Check for site-intent on php

lupus2k:
Fixes https://github.com/Automattic/dotcom-forge/issues/2617

## Proposed Changes
As some signup flows keep the `site-intent` alive after finishing the process, we checked for the launchpad status if the `site-intent` is still alive, and we need to make it in PHP.

#### Conditions to show the callout
- User has a verified email
- Site has no custom domain
- Site is not a P2
- Not show if it has `site-intent` and not launched (like start-writing, link-in-bio, etc)
- Should show if it has `site-intent` and is launched (like write flow)

## Testing Instructions

- You need to sandbox your testing site domain
- Go to `/apps/editing-toolkit`
- Run `yarn dev --sync`
- On Calypso, go to `/posts` (and `/pages`) and edit or create new.
- Or, change the view to classic and use `/wp-admin`
- You should see the callout if conditions match.
- Test it with a site created using the `write` flow as this flow keeps the `site-intent` alive after launching the site.

To test on Atomic Sites, follow the instructions here: PCYsg-ly5-p2 (**ETK and Atomic Sites**)

## Pre-merge Checklist

- [ ] ~~Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)~~
- [ ] ~~For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?~~

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
@@ -83,7 +83,22 @@ class WPCOM_Domain_Upsell_Callout {
 	private function should_not_show_callout() {
 		$blog_id = defined( 'IS_WPCOM' ) && IS_WPCOM ? get_current_blog_id() : \Jetpack_Options::get_option( 'id' );
 
-		return $this->blog_has_custom_domain( $blog_id ) || $this->user_has_email_unverified() || $this->blog_is_p2( $blog_id );
+		return $this->has_unlaunched_launchpad() || $this->blog_has_custom_domain( $blog_id ) || $this->user_has_email_unverified() || $this->blog_is_p2( $blog_id );
+	}
+
+	/**
+	 * Check for site-intent and not launched site on launchpad.
+	 */
+	private function has_unlaunched_launchpad() {
+		$site_intent      = defined( 'IS_WPCOM' ) && IS_WPCOM ? get_option( 'site_intent' ) : \Jetpack_Options::get_option( 'site_intent' );
+		$task_statuses    = get_option( 'launchpad_checklist_tasks_statuses', array() );
+		$site_is_launched = ! isset( $task_statuses['site_launched'] ) || ! $task_statuses['site_launched'];
+
+		if ( ! $site_intent || ! $site_is_launched ) {
+			return false;
+		}
+
+		return true;
 	}
 
 	/**
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
@@ -92,9 +92,9 @@ class WPCOM_Domain_Upsell_Callout {
 	private function has_unlaunched_launchpad() {
 		$site_intent      = get_option( 'site_intent' );
 		$task_statuses    = get_option( 'launchpad_checklist_tasks_statuses', array() );
-		$site_is_launched = ! isset( $task_statuses['site_launched'] ) || ! $task_statuses['site_launched'];
+		$site_is_launched = isset( $task_statuses['site_launched'] ) && $task_statuses['site_launched'];
 
-		if ( ! $site_intent || ! $site_is_launched ) {
+		if ( ! $site_intent || ( $site_intent && $site_is_launched ) ) {
 			return false;
 		}
 
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
@@ -90,7 +90,7 @@ class WPCOM_Domain_Upsell_Callout {
 	 * Check for site-intent and not launched site on launchpad.
 	 */
 	private function has_unlaunched_launchpad() {
-		$site_intent      = defined( 'IS_WPCOM' ) && IS_WPCOM ? get_option( 'site_intent' ) : \Jetpack_Options::get_option( 'site_intent' );
+		$site_intent      = get_option( 'site_intent' );
 		$task_statuses    = get_option( 'launchpad_checklist_tasks_statuses', array() );
 		$site_is_launched = ! isset( $task_statuses['site_launched'] ) || ! $task_statuses['site_launched'];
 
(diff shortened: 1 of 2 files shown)


apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
```diff
@@ -83,7 +83,22 @@ public function enqueue_script_and_style() {
...
 
-		return $this->blog_has_custom_domain( $blog_id ) || $this->user_has_email_unverified() || $this->blog_is_p2( $blog_id );
+		return $this->has_unlaunched_launchpad() || $this->blog_has_custom_domain( $blog_id ) || $this->user_has_email_unverified() || $this->blog_is_p2( $blog_id );
+	}
+
+	/**
+	 * Check for site-intent and not launched site on launchpad.
+	 */
+	private function has_unlaunched_launchpad() {
+		$site_intent      = defined( 'IS_WPCOM' ) && IS_WPCOM ? get_option( 'site_intent' ) : \Jetpack_Options::get_option( 'site_intent' );
```
lsl: What's the is_wpcom check here for? (I notice its not needed for the next line)

lupus2k: I think I copied it from the $blog_id line, it should not be there. I will remove it

lupus2k: Done!

lsl: Looks like it's just wpcom blog id that gets stored as a jetpack option for atomic sites. 

Going by this line that popped up in grep https://github.com/Automattic/wp-calypso/blob/df14f3f4e8333002d640409e532a1623fe854036/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/index.php#L30

lupus2k: Interesting, Do you think we have to do the same on the domain upsell file?

lsl: If it's needing a wpcom blog/site id and running in an atomic context then yes.

lsl: link?

apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
```diff
@@ -83,7 +83,22 @@ public function enqueue_script_and_style() {
...
+		return $this->has_unlaunched_launchpad() || $this->blog_has_custom_domain( $blog_id ) || $this->user_has_email_unverified() || $this->blog_is_p2( $blog_id );
+	}
+
+	/**
+	 * Check for site-intent and not launched site on launchpad.
+	 */
+	private function has_unlaunched_launchpad() {
+		$site_intent      = get_option( 'site_intent' );
+		$task_statuses    = get_option( 'launchpad_checklist_tasks_statuses', array() );
+		$site_is_launched = ! isset( $task_statuses['site_launched'] ) || ! $task_statuses['site_launched'];
```
DustyReagan: Would it make sense to check `get_option( 'launch-status' ) === 'unlaunched';` here instead? 🤔

DustyReagan (approved):
I left a comment for consideration, but this seems to be working well to me. I tested on simple and on Atomic by manually updating the `option` values using `/_cli` P9HQHe-je-p2

apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
DustyReagan: I think the `option` table gets sync'd/transferred when a site goes Atomic. So, `get_option` would work for `launchpad_checklist_tasks_statuses` and `site-intent`. But if I remember correctly `blog_id` on Atomic is always 1. So we have to call `\Jetpack_Options::get_option( 'site_intent' )` if we want the `blog_id` as wpcom knows it. We don't need `blog_id` for the `get_option` function so it's fine as is.... I think! 😅

lupus2k: I remember we have some similar problem with Gifting.

apps/editing-toolkit/editing-toolkit-plugin/wpcom-domain-upsell-callout/class-wpcom-domain-upsell-callout.php
lupus2k: You're right, but we have to change the if because we want to show the upsell to `unlaunched` sites but not while within the `launchpad`, like this:

```
$site_is_unlaunched = get_option( 'launch-status' ) === 'unlaunched';
if ( ! $site_intent || ( $site_intent && ! $site_is_unlaunched ) ) {
```

DustyReagan: Yeah, might could even do `$site_is_launched = get_option( 'launch-status' ) === 'launched';` to get rid of the `!` and make it more clear. I was thinking 'launch-status' might be a more definitive "source of truth" on whether the site is launched than 'launchpad_checklist_tasks_statuses['site_launched']'. But maybe there's a good reason to use `launchpad_checklist_tasks_statuses` instead?

lupus2k: I was thinking, what happens if the user finished the process and launched the site in Launchpad but then passed it to `coming soon`?

DustyReagan: Oh, I see! I'm not sure if that sets 'launch-status' to 'unlaunched' or not. 🤔 But, that makes sense why you might want to use `launchpad_checklist_tasks_statuses` instead. Cool with me! ☺️ 🚀

PR #77841, opened 2023-06-06
Extend the Woo GTM tracking in Calypso to also fire in the woo express stepper flow

dkd903:
As we don't get enough data in the Woo GTM container for trial upgrades (users going from a free woo express trial to a paid plan) yet, we should add the Woo GTM container and track users earlier in the flow. E.g. in the stepper flow here: https://wordpress.com/setup/wooexpress/

Related to # 1749-gh-Automattic/martech

## Proposed Changes

* Introduce a new method `recordGTMDatalayerEvent` and load the Woo GTM container in the same.
* Call the method in the stepper flow step changes routine

## Testing Instructions

* Install the GADebugger Chrome Extension and enable
* Switch to the branch on this PR and start Calypso
* Visit /setup/wooexpress/ and cycle through the various steps
* Open Chrome Console and notice that you can see Tag manager events for the various steps

## Pre-merge Checklist


client/lib/analytics/ad-tracking/woo/record-gtm-datalayer-event.ts
```diff
@@ -0,0 +1,38 @@
...
+	if ( stepName ) {
+		loadGTMContainer( TRACKING_IDS.wooGoogleTagManagerId )
+			.then( () => initGTMContainer() )
+			.then( () => {
+				debug(
+					`recordGTMDatalayerEvent: Initialized GTM container ${ TRACKING_IDS.wooGoogleTagManagerId }`
+				);
+
+				// We ensure that we can track with GTM
+				if ( ! mayWeTrackByTracker( 'googleTagManager' ) ) {
```
gmovr: Can we check consent higher up in this function, i.e. before loading and initializing GTM? I think it should work.

client/lib/analytics/ad-tracking/woo/record-gtm-datalayer-event.ts
```diff
@@ -0,0 +1,38 @@
...
+import { debug, TRACKING_IDS } from '../constants';
+import { initGTMContainer, loadGTMContainer } from '../gtm-container';
+
+/**
+ * Sends a step event to Google Tag Manager for the woo express stepper flow.
+ *
+ * @param {string} stepName
+ * @returns {void}
+ */
+export default async function recordGTMDatalayerEvent( stepName = '' ) {
```
gmovr:
```ts
export default function recordGTMDatalayerEvent( stepName: string ): void {
```

but if you want the function to be `async` you can't return `void`, and also you need to `await` for it to resolve/reject where you use the function (ex in the Stepper flow). 

e.g.

```ts
export default async function recordGTMDatalayerEvent( stepName: string ): Promise< void > { //...


// and where you use it
await recordGTMDatalayerEvent( stepName);
```

I don't think we need it to be async in this case.

client/lib/analytics/ad-tracking/woo/record-gtm-datalayer-event.ts
```diff
@@ -0,0 +1,38 @@
...
+	if ( stepName ) {
```
gmovr: as long as we set stepName as required in the type definitions, then there's no need to check it here IMO

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
```diff
@@ -110,6 +111,7 @@ const wooexpress: Flow = {
 
 		function submit( providedDependencies: ProvidedDependencies = {}, ...params: string[] ) {
 			recordSubmitStep( providedDependencies, intent, flowName, currentStep );
+			recordGTMDatalayerEvent( currentStep );
```
gmovr: I'm not too familiar with Stepper. But I see that there's a `useSideEffect` hook (optional) that we can use, ref the stepper readme file. It says it's for tracking functions and other side-effects.

@Automattic/martech-decepticons who's been working with Stepper. Should we use this, or can we keep the tracking logic in the submit function?

gmovr (changes requested):

gmovr:
Thanks for working on this! I left some comments 👍  cc: @markbiek and @daledupreez  who've been working on this flow previously.

gmovr:
Could you add some testing instructions as well?

client/lib/analytics/ad-tracking/woo/record-gtm-datalayer-event.ts
dkd903: Makes sense; better to return early if tracking is not permitted. Updated the PR.

client/lib/analytics/ad-tracking/woo/record-gtm-datalayer-event.ts
dkd903: @gmovr  - Apologies, that was indeed a typo :-) I copied the other method and was building this off of that other piece of code. Just before creating the PR, I might have done an undo and the async got pushed up. Thanks for spotting it.

jeyip:
Following the testing instructions, but when I visit `/setup/woo-express` I'm redirected to my home 🤔 Is there any other environment configuration I need to do?



_Edit:_ I think I figured it out 👍 After some digging, I'm guessing that you intended to write `/setup/wooexpress` and not `/setup/woo-express` I'm, however, not seeing any onboarding steps to walk through. I'm also not seeing anything in the GA debugger 😞 

Judging from [this initial scaffolding PR's testing instructions](https://github.com/Automattic/wp-calypso/pull/71611), having non-interactable steps appears to be expected behavior, but then I wonder what exactly we're intending to track in this PR? Would you mind elaborating a bit more @dkd903?

client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts
jeyip: I believe `useSideEffect` is called whenever a step in the flow is rendered, whereas `submit` would be called whenever the submit callback is invoked, often _to mark the completion_ of a step cc @alshakero to verify

To answer your question, it depends on whether or not you'd like to track a step being seen or a step being completed by the end user

Ex. To elaborate further on `submit`, the trial WooExpress flow has several steps ( [link](https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts#L24-L29) ) If we look at any of the code rendering these views, we'll see the `submit` callback tied to some logic in the file

Below is the `submit` function in the `siteCreationStep`

https://github.com/Automattic/wp-calypso/blob/4144f48528e8f5d3c787376d2a05aabff3702d2c/client/landing/stepper/declarative-flow/internals/steps-repository/site-creation-step/index.tsx#L56

It's probably invoked whenever the creator of the step has decided that the `siteCreationStep` has finished

dkd903:
> 

Thank you @jeyip You can find some additional context on this thread: p1684695879052389-slack-C1A7F9WG5

jeyip:
In the slack thread I see that @kevinbates says "And just to clarify, we don't need an event when the free trial flow is started, only when it's completed. :slightly_smiling_face:"

In that case, I would actually move the event tracking to here above the `exitFlow` method

https://github.com/Automattic/wp-calypso/blob/8252bfb81eeeca07da01374be668d94a07355b8e/client/landing/stepper/declarative-flow/trial-wooexpress-flow.ts#L135-L137

dkd903:
@jeyip  - I pushed a new commit to address your comment.

jeyip:
> I'm also not seeing anything in the GA debugger 😞

Hey @dkd903, I mentioned this earlier [here](https://github.com/Automattic/wp-calypso/pull/77841/#issuecomment-1587862400), but I'm still unable to see any events fired off in the Google Analytics Debugger.

@gmovr + @dkd903 It sounds like you were both successfully seeing events triggered though? I'm assuming there's still something left in my dev environment that needs to be set up. If either of you wouldn't mind triggering the event, recording a gif, and posting it here to confirm functionality, then consider it a +1 from my end.

dkd903:
> > I'm also not seeing anything in the GA debugger 😞
> 
> Hey @dkd903, I mentioned this earlier [here](https://github.com/Automattic/wp-calypso/pull/77841/#issuecomment-1587862400), but I'm still unable to see any events fired off in the Google Analytics Debugger.
> 
> @gmovr + @dkd903 It sounds like you were both successfully seeing events triggered though? I'm assuming there's still something left in my dev environment that needs to be set up. If either of you wouldn't mind triggering the event, recording a gif, and posting it here to confirm functionality, then consider it a +1 from my end.

Thanks @jeyip - I am AFK today, will respond to you by tomorrow, first thing.

gmovr:
Thanks for the review. For testing ad-tracking related code, you can enable `ad-tracking` and `cookie-banner` in the development.json config file. Another option is to add `?flags=ad-tracking,cookie-banner` to the URL, however, this is less reliable, as the flag may disappear upon redirects.

Taking this for a spin now to test the functionality, after the latest changes.

gmovr:
I tested via the GTM debugger and the browser network inspector (e.g. looking at `window.dataLayer` for the `user interaction` data layer event. Works well. 

We may have some follow-ups if firing this on this step is insufficient, but that will be fine to add later.

@dkd903 -- I restarted the test run, but if it still fails, can you rebase it? It doesn't look related, maybe just a timeout -- but better to be sure. 

cc: @kevinbates

gmovr:
Yeah, failure in that one build task is related to some language files in stats. Rebase should fix this 👍

gmovr (approved):
Works well, thanks for handling the follow-ups! Let's rebase it to make sure tests pass, and it's good to go 🚢

dkd903:
Closing this PR due to a bad merge. Opened a new PR: https://github.com/Automattic/wp-calypso/pull/78348

PR #77845, opened 2023-06-06
reader search popular sites - add follow button

Addison-Stavlo:
Related to # pe7F0s-Rz-p2

## Proposed Changes

* This attempts to add the follow button to the popular sites list in reader search (when no search input is present), by using the same component that renders sites results when a search input is present.

I will circle back to add more images and thoughts here, but right now the search results started coming back empty so its a bit hard to test.

## Testing Instructions

*

## Pre-merge Checklist

### DIFF
--- a/client/blocks/reader-subscription-list-item/connected.jsx
+++ b/client/blocks/reader-subscription-list-item/connected.jsx
@@ -18,6 +18,7 @@ class ConnectedSubscriptionListItem extends Component {
 		onComponentMountWithNewRailcar: PropTypes.func,
 		showNotificationSettings: PropTypes.bool,
 		showLastUpdatedDate: PropTypes.bool,
+		showFollowedOnDate: PropTypes.bool,
 		isEmailBlocked: PropTypes.bool,
 		isFollowing: PropTypes.bool,
 		followSource: PropTypes.string,
@@ -29,6 +30,7 @@ class ConnectedSubscriptionListItem extends Component {
 		onComponentMountWithNewRailcar: noop,
 		showNotificationSettings: true,
 		showLastUpdatedDate: true,
+		showFollowedOnDate: true,
 	};
 
 	componentDidMount() {
@@ -56,6 +58,7 @@ class ConnectedSubscriptionListItem extends Component {
 			siteId,
 			showNotificationSettings,
 			showLastUpdatedDate,
+			showFollowedOnDate,
 			isFollowing,
 			followSource,
 			railcar,
@@ -70,6 +73,7 @@ class ConnectedSubscriptionListItem extends Component {
 				url={ url }
 				showNotificationSettings={ showNotificationSettings }
 				showLastUpdatedDate={ showLastUpdatedDate }
+				showFollowedOnDate={ showFollowedOnDate }
 				isFollowing={ isFollowing }
 				followSource={ followSource }
 				railcar={ railcar }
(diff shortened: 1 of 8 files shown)


davemart-in:
@Addison-Stavlo the follow button itself LGTM. Here's what I'm seeing:

https://github.com/Automattic/wp-calypso/assets/5634774/8005db42-27b1-4b22-b5a4-cd2d337c7722

That said, after you follow one of these, some extra data is being injected into the component. A cog shows up and the alignment of the additional text looks off:

https://github.com/Automattic/wp-calypso/assets/5634774/c9ec2a93-cf5c-45a3-8cd8-adfadc5bdc27

Can we just prevent this additional stuff from being added to the DOM? So, like this post-follow:

CleanShot 2023-06-06 at 12 01 22

Addison-Stavlo:
> after you follow one of these, some extra data is being injected into the component. A cog shows up and the alignment of the additional text looks off:

@davemart-in   Yes, i wanted to ask about that!  This seems to be the current behavior in production of the component that is here when you search for something:

Screenshot 2023-06-06 at 12 49 45 PM

regarding the extra text that shows up (followed on date), I was wondering if we should remove this from both places?  (both with and without search inputs?) they seem to clutter everything in general.

The cog seems fine to me, but id be happy to hide that as well if necessary. If so, do you think we should hide it in both cases? It seems like they should have the same behavior.

Id also ask the same questions of which of these makes sense to keep when we are talking about the single-column view of this on reduced screen widths.  Here is what it looks like on this state of this PR (which seems to be the same as the search result parallel that currently exists):
Screenshot 2023-06-06 at 1 19 38 PM

Since i refactored to use the component from the sites search results here, i left some of that above behavior the same for parity and wanted to get your thoughts on what we should do here regarding them. It seems like w/ and w/o search those sites list items should have visual/behavioral parity, but if thats not the case and we want them to be different we can go that way as well.

davemart-in:
> regarding the extra text that shows up (followed on date), I was wondering if we should remove this from both places?

I think I'd prefer to keep this additional context on the "manage followed sites" screen. I think it can be helpful to know when you followed a site. We could perhaps tighten up the styles on that page though.

> The cog seems fine to me, but id be happy to hide that as well if necessary. If so, do you think we should hide it in both cases? It seems like they should have the same behavior.

I'd vote to keep it on the manage screen, but hide it here. These are different contexts IMO. The manage screen is meant to be a full on management experience. This sidebar is just meant to be a prompt to highlight additional sites that you might want to follow.

> Id also ask the same questions of which of these makes sense to keep when we are talking about the single-column view of this on reduced screen widths.

I'd hide the cog and extra info here as well. Again, this is meant for a different context than the manage screen.

> It seems like w/ and w/o search those sites list items should have visual/behavioral parity, but if thats not the case and we want them to be different we can go that way as well.

In this case, I think it's okay to show different things based on the context of what we're trying to achieve.

Addison-Stavlo:
Updating here after chatting with Dave yesterday.  What i was referring to wasnt the following/manage areas, but the same site sidebar on the search page once a search result is present. The decision is to remove everything but the follow button for all of the site items components on this search page (and add it to the places it isn't), while ensuring their use in following/manage and other areas is unaffected.

davemart-in (approved):
Great job! Works as expected. Here's what I'm seeing:

CleanShot 2023-06-07 at 10 26 06

Manage screen still has all of the right context:

CleanShot 2023-06-07 at 10 27 17@2x

PR #77848, opened 2023-06-06
Sites: track empty state view

zaguiini:
## Proposed Changes

We should track the page views for the empty sites page so that we can understand how many people are starting the onboarding flow after signup.

## Testing Instructions

1. Open the dev tools, change the log level to verbose;
2. Type `localStorage.setItem('debug', 'calypso:analytics')` in the console;
3. Browse `/sites` with an account with no sites;
4. Check that the page view event was dispatched with `site_count: 0`.


mpkelly (approved):
Now fires in both cases - when there are sites and when there are none. 

Empty:
image

Has sites (existing functionality):

image

PR #77849, opened 2023-06-06
Signup: store `is_dev_account` property in Tracks

zaguiini:
## Proposed Changes

**PLEASE ADD THE PROPERTY TO TRACKS BEFORE MERGING THE PR**

Adds `is_dev_account` as a property to the `calypso_signup_actions_submit_step` event.

**PLEASE ADD THE PROPERTY TO TRACKS BEFORE MERGING THE PR**

## Testing Instructions

2. Browse `/start/hosting` with the `hosting-onboarding-i2` feature flag turned on;
3. Create an account in passwordless mode and check that `calypso_signup_actions_submit_step` was dispatched with `flow: user-hosting` and `is_dev_account: bool`;
4. Create an account by logging in with a social account (Google or Apple) and check that `calypso_signup_actions_submit_step` was dispatched with `flow: user-hosting` and `is_dev_account: bool`;

### DIFF
--- a/client/blocks/signup-form/index.jsx
+++ b/client/blocks/signup-form/index.jsx
@@ -941,7 +941,12 @@ class SignupForm extends Component {
 			<SelectCardCheckbox
 				className="signup-form__is-dev-account-checkbox signup-form__span-columns"
 				checked={ this.state.isDevAccount }
-				onChange={ ( isDevAccount ) => this.setState( { isDevAccount } ) }
+				onChange={ ( isDevAccount ) => {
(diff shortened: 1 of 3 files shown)


mpkelly:
Edit. Ok, so the event props need to be updated before testing. TIL.

mpkelly (approved):
It works and the event fires each time the checkbox is toggled. One thing to note is that this event will only fire if the user interacts with the checkbox. Maybe that's fine but there is no implicit recording of "not at developer", even though we will sometimes record `is_dev_account: false` when they toggle on then off again.

PR #77851, opened 2023-06-06
Hosting flows: add `is_in_hosting_flow` property to `calypso_signup_step_start`

zaguiini:
## Proposed Changes

**PLEASE ADD THE PROPERTY TO TRACKS BEFORE MERGING THE PR**

Adds `is_in_hosting_flow` as a property to the `calypso_signup_step_start` event.

**PLEASE ADD THE PROPERTY TO TRACKS BEFORE MERGING THE PR**

## Testing Instructions

2. Browse `/start/new-hosted-site?hosting-flow=true`;
3. Check that `calypso_signup_step_start` was dispatched with `is_in_hosting_flow: true`;
4. Remove the query parameter, refresh the page and check that the same event gets dispatched but with `is_in_hosting_flow` set to `false`.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/index.tsx
@@ -13,11 +13,9 @@ import {
 	getSignupCompleteFlowNameAndClear,
 	getSignupCompleteStepNameAndClear,
 } from 'calypso/signup/storageUtils';
-import { useSelector } from 'calypso/state';
 import { useSite } from '../../hooks/use-site';
 import useSyncRoute from '../../hooks/use-sync-route';
 import { ONBOARD_STORE } from '../../stores';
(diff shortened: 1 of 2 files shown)


mpkelly:
This is not working for me whether `?hosting-flow=true` is present or not. I am running the local Calypso branch, and the code in the file that has been changed does not run when viewing `/start/new-hosted-site?hosting-flow=true`. The Tracks event fires, but it doesn't contain the new property.

mpkelly:
Ok, so it looks like this was in the wrong place. I moved it (see https://github.com/Automattic/wp-calypso/pull/77851/commits/561f83298835ce42b381de51232f68c1c1cb79ed), and it now works as expected. Going to merge this in. @zaguiini, please review my changes. 

Screenshot 2566-06-08 at 16 25 39


image

mpkelly (approved):

Issue #77684, opened 2023-06-01
Subscribers Page: Heading with count


ivan-ottinger:
Create "Total" heading with the number of subscribers. If the API providing the number of subscribers isn't ready yet, feel free to use placeholder number:

![Markup on 2023-06-06 at 17:39:19](https://github.com/Automattic/wp-calypso/assets/25105483/8bd3646a-1a98-4ab0-97c1-b28bdff441f6)

Design: h8tMpJlCqNFemEerU9owHI-fi-1880_61261

ivan-ottinger:
🔴 We are waiting for the design finalization.

ivan-ottinger:
The design is still in progress, but we can already introduce the section based on the WIP design: h8tMpJlCqNFemEerU9owHI-fi-1880_61261

ivan-ottinger:
This task has been resolved by https://github.com/Automattic/wp-calypso/pull/77852.

PR #77852, opened 2023-06-06
Subscribers Page: add the subscriber list

### LINKED ISSUE #77679: Subscribers Page: Create a list of subscribers with mock data
The goal of this task is to create a list of subscribers. If the API / query returning the real data is not ready yet, feel free to use mock data.

The following columns should be covered:

- Name
- Since
- "Action" (ellipsis `...` menu) - doesn't have to be clickable yet

Details on the _Name_ column:

- for WPCOM user, the column should be displayed as can be seen in the design
- for external user, the email address should be formatted in a similar way as the WPCOM subscriber's name and there should be no gray `@username` line (as the username does not exist for external users)

If that makes sense, the pagination task can be worked on simultaneously: https://github.com/Automattic/wp-calypso/issues/77731.

Design: h8tMpJlCqNFemEerU9owHI-fi-1764_57090

JuanLucha:
Closes https://github.com/Automattic/wp-calypso/issues/77679 and https://github.com/Automattic/wp-calypso/issues/77684

## Proposed Changes

This PR introduces the subscriber list and the total counter on the Subscribers Page:

Screenshot 2023-06-05 at 23 01 36 2

Note: we are using mocks at this stage.

## Testing Instructions

2. Go to this URL: `http://calypso.localhost:3000/subscribers/[the-domain-of-your-site]`.
3. Check you can see a list like the image above.

### DIFF
--- a/client/my-sites/subscribers/main.tsx
+++ b/client/my-sites/subscribers/main.tsx
@@ -1,19 +1,63 @@
 import config from '@automattic/calypso-config';
+import { useLocale } from '@automattic/i18n-utils';
 import { translate } from 'i18n-calypso';
+import { useSelector } from 'react-redux';
+import { Item } from 'calypso/components/breadcrumb';
 import DocumentHead from 'calypso/components/data/document-head';
+import FixedNavigationHeader from 'calypso/components/fixed-navigation-header';
 import Main from 'calypso/components/main';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
+import { useSubscribersQuery } from './queries';
+import { SubscriberList } from './subscriber-list/subscriber-list';
+import './styles.scss';
 
 export const Subscribers = () => {
 	const isSubscribersPageEnabled = config.isEnabled( 'subscribers-page' );
+	const locale = useLocale();
+	const selectedSiteId = useSelector( getSelectedSiteId );
+	const initialState = { data: { total: 0, subscribers: [] } };
+	const result = useSubscribersQuery( selectedSiteId );
+	const {
+		data: { total, subscribers = [] },
+	} = result && result.data ? result : initialState;
+
+	const navigationItems: Item[] = [
+		{
+			label: 'Subscribers',
+			href: `/subscribers`,
+			helpBubble: (
+				<span>
+					{ translate(
+						'Add subscribers to your site and send them a free or paid {{link}}newsletter{{/link}}.',
+						{
+							components: {
+								link: (
+									<a
+										href="https://wordpress.com/support/launch-a-newsletter/#about-your-subscribers"
+										target="blank"
+									/>
+								),
+							},
+						}
+					) }
+				</span>
+			),
+		},
+	];
 
 	if ( ! isSubscribersPageEnabled ) {
 		return null;
 	}
 
 	return (
-		<Main>
+		<Main wideLayout className="subscribers">
 			<DocumentHead title={ translate( 'Subscribers' ) } />
-			Subscribers
+			<FixedNavigationHeader navigationItems={ navigationItems }></FixedNavigationHeader>
+			<div className="subscribers__header-count">
+				<span className="subscribers__title">{ translate( 'Total' ) }</span>{ ' ' }
+				<span className="subscribers__subscriber-count">{ total }</span>
+			</div>
+			<SubscriberList subscribers={ subscribers } locale={ locale } />
 		</Main>
 	);
 };
--- a/client/my-sites/subscribers/subscriber-list/styles.scss
+++ b/client/my-sites/subscribers/subscriber-list/styles.scss
@@ -1,0 +1,125 @@
+@import "@automattic/color-studio/dist/color-variables";
+@import "@wordpress/base-styles/breakpoints";
+
+.subscriber-list {
+	margin: 0;
+
+	.row {
+		border-block-end: 1px solid $studio-gray-5;
+		display: flex;
+		align-items: center;
+		flex-direction: row;
+		padding-top: 20px;
+		padding-bottom: 20px;
+
+		* {
+			flex: 1;
+		}
+
+		.hidden {
+			display: none;
+		}
+
+		&.header {
+			padding-bottom: $font-code;
+			padding-top: 0;
+
+			@media (max-width: $break-small) {
+				display: none;
+			}
+		}
+
+		.subscriber-list__checkbox-column,
+		.subscriber-list__profile-column,
+		.subscriber-list__subscription-type-column,
+		.subscriber-list__rate-column,
+		.subscriber-list__since-column,
+		.subscriber-list__menu-column {
+			font-weight: 400;
+			font-size: $font-body-small;
+			font-size: rem(14px);
+			line-height: 20px;
+			letter-spacing: -0.15px;
+			color: $studio-gray-60;
+		}
+
+		.subscriber-list__profile-column {
+			display: flex;
+			align-items: center;
+			flex: 3;
+			min-width: 0;
+
+			.subscriber-profile {
+				display: flex;
+				min-width: 0;
+
+				.subscriber-profile__user-image {
+					height: 40px;
+					width: 40px;
+					border-radius: 50%;
+					margin-right: 12px;
+					flex: 0;
+				}
+
+				.subscriber-profile__user-details {
+					display: flex;
+					flex-direction: column;
+					justify-content: center;
+					max-width: 100%;
+					overflow: hidden;
+					min-width: 0;
+
+					.subscriber-profile__name {
+						overflow: hidden;
+						text-overflow: ellipsis;
+						white-space: nowrap;
+						font-weight: bold;
+						line-height: 17px;
+						min-width: 0;
+						font-size: rem(15px);
+					}
+
+					.subscriber-profile__email {
+						overflow: hidden;
+						text-overflow: ellipsis;
+						white-space: nowrap;
+						font-size: rem(12px);
+						color: $studio-gray-60;
+						min-width: 0;
+					}
+				}
+			}
+		}
+
+		.subscriber-list__checkbox-column {
+			.form-checkbox {
+				height: 20px;
+				width: 20px;
+			}
+		}
+
+		.subscriber-list__menu-column {
+			flex-basis: 36px;
+			flex-grow: initial;
+
+			.gridicon {
+				fill: $studio-gray-50;
+			}
+		}
+
+		&:last-child {
+			border-bottom: none;
+		}
+	}
+}
+
+@media ( max-width: $break-medium ) {
+	.subscriber-list {
+		padding-left: 16px;
+		padding-right: 16px;
+
+		.subscriber-list__since-column {
+			display: none;
+		}
+	}
+}
--- a/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
+++ b/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
@@ -1,0 +1,61 @@
+import { useTranslate } from 'i18n-calypso';
+import FormCheckbox from 'calypso/components/forms/form-checkbox';
+import { Subscriber } from '../types';
+import { SubscriberRow } from './subscriber-row';
+import './styles.scss';
+
+type SubscriberListProps = {
+	subscribers: Subscriber[];
+	locale: string;
+};
+
+export const SubscriberList = ( { subscribers, locale }: SubscriberListProps ) => {
+	const translate = useTranslate();
+	return (
+		<ul className="subscriber-list" role="table">
+			<li className="row header" role="row">
+				<span className="subscriber-list__checkbox-column hidden" role="columnheader">
+					<FormCheckbox />
+				</span>
+				<span className="subscriber-list__profile-column" role="columnheader">
+					{ translate( 'Name' ) }
+				</span>
+				<span className="subscriber-list__subscription-type-column hidden" role="columnheader">
+					{ translate( 'Subscription type' ) }
+				</span>
+				<span className="subscriber-list__rate-column hidden" role="columnheader">
+					{ translate( 'Open Rate' ) }
+				</span>
+				<span className="subscriber-list__since-column" role="columnheader">
+					{ translate( 'Since' ) }
+				</span>
+				<span className="subscriber-list__menu-column" role="columnheader"></span>
+			</li>
+			{ subscribers.map(
+				( {
+					user_id,
+					subscription_id,
+					display_name,
+					email_address,
+					subscriptions,
+					openRate,
+					date_subscribed,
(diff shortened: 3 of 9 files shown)


ivan-ottinger:
Great progress, Juan! ☺️ 

A couple of initial things I noticed:

- the sorting arrows don’t need to be in for the Milestone I. But if you already have them, that’s good!
- the “Subscribers ℹ️” header section / stripe is not part of the Milestone I - as this is still in design; we can simply display the list of subscribers with the total number; in other words, for Milestone I we are aiming for the green boxes (as discussed in pdDOJh-1TE-p2:

Markup on 2023-06-07 at 09:32:21

- “Open rate” column can be hidden - as it is not part of Milestone I.
- “Status” columns is being reworked (as can be seen in the design WIP: h8tMpJlCqNFemEerU9owHI-fi-1880_61261 - screens with yellow background); I would personally skip it altogether at this point as it will likely change a lot

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
+import moment from 'moment';
+import { Item } from 'calypso/components/breadcrumb';
 import DocumentHead from 'calypso/components/data/document-head';
-import Main from 'calypso/components/main';
+import FixedNavigationHeader from 'calypso/components/fixed-navigation-header';
+import FormCheckbox from 'calypso/components/forms/form-checkbox';
+import './styles.scss';
+import { SubscriberProfile } from './subscriber-profile';
+
+const subscribers = [
```
phcp: Maybe moving this mock to a json file would be cleaner since there are many lines on the array.

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
 			<DocumentHead title={ translate( 'Subscribers' ) } />
-			Subscribers
-		</Main>
+			<div className="main is-wide-layout">
+				<FixedNavigationHeader navigationItems={ navigationItems }></FixedNavigationHeader>
+				<div className="header-count">
+					<span className="title">{ translate( 'Total' ) }</span>{ ' ' }
+					<span className="subscriber-count">{ subscribers && subscribers.length }</span>
+				</div>
+				<table className="subscriber-list">
```
phcp: This table could be moved to `SubscribersList` component, it would make this file more readable and easier to introduce the next features.

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
+							( {
+								subscription_id,
+								display_name,
+								email_address,
+								subscription,
+								openRate,
+								date_subscribed,
+								avatar,
+							} ) => (
+								<tr key={ subscription_id }>
```
phcp: Same for the rows, moving to a `SubscriberRow` or something similar would make it more readable.

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
+									<td className="checkbox-column">
+										<FormCheckbox />
+									</td>
```
phcp: On the phase 1 we aren't including the checkbox, so maybe it would worth it to remove for now. h8tMpJlCqNFemEerU9owHI-fi-1880_61261

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
+											displayName={ display_name }
+											email={ email_address }
+										/>
+									</td>
+									<td className="subscription-type-column">{ subscription }</td>
+									<td className="rate-column">{ openRate }</td>
+									<td className="since-column">
+										{ moment( date_subscribed ).locale( locale ).format( 'LL' ) }
+									</td>
+									<td className="ellipsis-menu">•••</td>
```
phcp: You can possibly replace it for a `<Gridicon icon="ellipsis" ... />

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
 		return null;
 	}
 
 	return (
-		<Main>
+		<>
 			<DocumentHead title={ translate( 'Subscribers' ) } />
-			Subscribers
-		</Main>
+			<div className="main is-wide-layout">
```
phcp: If there is no issue with the Main component, it could be replaced by `<Main wideLayout>`.

client/my-sites/subscribers/main.tsx
```diff
@@ -1,19 +1,306 @@
...
+				<FixedNavigationHeader navigationItems={ navigationItems }></FixedNavigationHeader>
+				<div className="header-count">
+					<span className="title">{ translate( 'Total' ) }</span>{ ' ' }
+					<span className="subscriber-count">{ subscribers && subscribers.length }</span>
+				</div>
+				<table className="subscriber-list">
+					<tbody>
+						<tr className="table-header">
```
phcp: [EsLint Calypso rules](https://github.com/Automattic/wp-calypso/blob/trunk/packages/eslint-plugin-wpcalypso/docs/rules/jsx-classname-namespace.md) recommends to use BEM naming conventions on classNames. You can check the rules here: https://en.bem.info/methodology/naming-convention/. In this case all the classNames nested to subscriber-list would be updated to include the namespace prefix. Example: `table-header` => `subscriber-list__ table-header`

client/my-sites/subscribers/subscriber-profile.tsx
```diff
@@ -0,0 +1,19 @@
...
+	avatar,
+	displayName,
+	email,
+}: {
+	avatar: string;
+	displayName: string;
+	email: string;
+} ) {
+	return (
+		<div className="user-wrapper">
```
phcp: Same here for the className conventions.

phcp (commented):
Good job Juan! Added some comments related to componentization and lint stuff. Let me know your thoughts about it.

client/my-sites/subscribers/main.tsx
JuanLucha: Hi, Paulo!
The good news is I'm removing the mock as the endpoint and the query is already in the trunk. I'll push this change today.

client/my-sites/subscribers/main.tsx
JuanLucha: Very nice! 🚀

client/my-sites/subscribers/main.tsx
JuanLucha: Great!

client/my-sites/subscribers/main.tsx
JuanLucha: Indeed. I'll push the menu with the delete button in my next commit.

client/my-sites/subscribers/main.tsx
JuanLucha: I'll try that 👌

client/my-sites/subscribers/main.tsx
JuanLucha: I'll change them

client/my-sites/subscribers/subscriber-profile.tsx
JuanLucha: I'll change them

client/my-sites/subscribers/main.tsx
phcp: Maybe in this case it could be in a different PR/task as you already have the mock implemented. (But of course if you already applied the query you can ignore my message 😄)

client/my-sites/subscribers/main.tsx
phcp: This node `<div className="main is-wide-layout">` isn't required as you added the Main component on the top.

client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+				<span className="subscriber-list__subscription-type-column" role="columnheader">
+					{ translate( 'Subscription type' ) }
+				</span>
+				<span className="subscriber-list__rate-column" role="columnheader">
+					{ translate( 'Open Rate' ) }
+				</span>
+				<span className="subscriber-list__since-column" role="columnheader">
+					{ translate( 'Since' ) }
+				</span>
+				<span className="subscriber-list__menu-column" role="columnheader"></span>
```
phcp: In this case, I think using two different selectors for rows and list is making it a bit confusing on Sass (sorry for the bad suggestion 😬), maybe using just the `subscriber-list__` as a namespace would be cleaner and also reusable on both components (`SubscriberList` and `SubscriberRow`). For the specificities of each column you could use additional selectors for now, like `checkbox-column`. Another thing is that you can include a `hidden` selector just for the hidden columns, it adds some clarity on what is rendered without even look to the css.

```
		<ul className="subscriber-list" role="table">
			<li className="subscriber-list__row header" role="row">
				<span className="subscriber-list__column checkbox-column hidden" role="columnheader">
					<FormCheckbox />
				</span>
				<span className="subscriber-list__column profile-column" role="columnheader">
					{ translate( 'Name' ) }
				</span>
				<span className="subscriber-list__column hidden" role="columnheader">
					{ translate( 'Subscription type' ) }
				</span>
				...
```

client/my-sites/subscribers/subscriber-list/subscriber-row.tsx
```diff
@@ -0,0 +1,42 @@
...
+			<span className="subscriber-row__rate-column" role="cell">
+				{ openRate }
+			</span>
+			<span className="subscriber-row__since-column" role="cell">
+				{ moment( date_subscribed ).locale( locale ).format( 'LL' ) }
+			</span>
+			<span className="subscriber-row__menu-column" role="cell">
+				<Gridicon icon="ellipsis" size={ 24 } />
+			</span>
+		</li>
```
phcp: Similarly, you could consider the row component as part of the subscriber-list tree, so it could look like this:
```
		<li className="subscriber-list__row" role="row">
			<div className="subscriber-list__column checkbox-column hidden" role="cell">
				<FormCheckbox />
			</div>
			<span className="subscriber-list__column profile-column" role="cell">
				<SubscriberProfile avatar={ avatar } displayName={ display_name } email={ email_address } />
			</span>
			<span className="subscriber-list__column hidden" role="cell">
				{ subscriptions && subscriptions.map( ( subscription ) => <div>{ subscription }</div> ) }
			</span>
			...
```

phcp (commented):
Good job Juan! Thank you for working on the feedback. I just noticed that the font-size for the Profile column isn't matching with the Figma design h8tMpJlCqNFemEerU9owHI-fi-1682_52557, and CI is failing because of a lint issue. I think that after fixing these things you are good to go. There are also a few things related to the selectors that could be improved (I left it as a comment), but I don't see it as a blocker, we can address them on follow-up PRs as we will be touching a lot on these components for the next weeks.

client/my-sites/subscribers/types/index.ts
```diff
@@ -24,5 +24,7 @@ export type Subscriber = {
 	email_address: string;
 	avatar: string;
 	display_name: string;
+	subscriptions: SubscriptionPlan[];
```
ivan-ottinger, line 27: I might have missed something, but why is `subscriptions: SubscriptionPlan[];` necessary here?

client/my-sites/subscribers/subscriber-list/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+
+		.subscriber-list__checkbox-column,
+		.subscriber-list__profile-column,
+		.subscriber-list__subscription-type-column,
+		.subscriber-list__rate-column,
+		.subscriber-list__since-column,
+		.subscriber-list__menu-column {
+			font-weight: 400;
+			font-size: $font-body-small;
+			font-size: rem(14px);
```
ivan-ottinger, line 40: The `font-size` is set two times here. I believe you wanted to keep `font-size: $font-body-small;` only.

client/my-sites/subscribers/subscriber-list/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+					min-width: 0;
+
+					.subscriber-profile__name {
+						overflow: hidden;
+						text-overflow: ellipsis;
+						white-space: nowrap;
+						font-weight: bold;
+						line-height: 17px;
+						min-width: 0;
+						font-size: rem(15px);
```
ivan-ottinger, line 79: Here we could use font-size variable instead (from `@import "@automattic/typography/styles/variables";`)

client/my-sites/subscribers/subscriber-list/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+					}
+
+					.subscriber-profile__email {
+						overflow: hidden;
+						text-overflow: ellipsis;
+						white-space: nowrap;
+						font-size: rem(12px);
```
ivan-ottinger, line 86: The same as in the previous comment - we can use font-size variable instead.

client/my-sites/subscribers/subscriber-list/subscriber-row.tsx
```diff
@@ -0,0 +1,42 @@
...
+			</span>
+			<span className="subscriber-list__subscription-type-column hidden" role="cell">
+				{ subscriptions && subscriptions.map( ( subscription ) => <div>{ subscription }</div> ) }
+			</span>
+			<span className="subscriber-list__rate-column hidden" role="cell">
+				{ openRate }
+			</span>
+			<span className="subscriber-list__since-column" role="cell">
+				{ moment( date_subscribed ).locale( locale ).format( 'LL' ) }
+			</span>
```
ivan-ottinger, line 36: I think we could consider using the `<TimeSince />` component here. This will allow us to get rid of passing the locale down the component tree and make the code consistent with what we already have in the _Subscriptions Management_ portal (e.g. in `client/landing/subscriptions/components/site-subscriptions-manager/site-row.tsx`)

ivan-ottinger (commented):
The list is displaying nicely! 👍🏼👍🏼 I would have just a couple more suggestions (please see below and in the code). 🙂

The email address (if only that is displayed) and the ellipsis menu on the right should be vertically aligned:

Markup on 2023-06-09 at 10:20:32

When the tooltip is clicked on, there's this border displayed that shouldn't be there:
Markup on 2023-06-09 at 10:18:17

I would say that these two things can be addressed in a separate PR.

ivan-ottinger (approved):
We have agreed to merge this PR already and then create a separate PR addressing the latest suggestions I shared above.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8034059

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8034059

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8034059).

PR #77853, opened 2023-06-06
New hosted site: Track site title and geo affinity changes

zaguiini:
## Proposed Changes

Adds `has_site_title` and `has_geo_affinity` as properties to the `calypso_signup_site_options_submit` event.

This is useful to understand whether the users are going with the defaults defined by us.

## Testing Instructions

2. Browse `/start/new-hosted-site`;
3. Pick a Business plan;
4. Go through the site options step, changing the title and data center;
5. Check that `calypso_signup_site_options_submit` was dispatched with `has_site_title` and `has_geo_affinity` according to your activities on the page.

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/site-options/new-hosted-site-options.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/site-options/new-hosted-site-options.tsx
@@ -72,8 +72,7 @@ export const NewHostedSiteOptions = ( { navigation }: Pick< StepProps, 'navigati
 		}
 
 		recordTracksEvent( 'calypso_signup_site_options_submit', {
-			has_site_title: true,
-			has_changed_title: hasChangedInput.current.title,
+			has_site_title: hasChangedInput.current.title,
 			has_changed_geo_affinity: hasChangedInput.current.geoAffinity,
 		} );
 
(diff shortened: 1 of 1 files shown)


p-jackson:
The `calypso_signup_site_options_submit` event already has a `has_site_title` prop. On the surface that seems like no problem to add our new `has_changed_title` props, because they seem like two different things semantically.

The problem is that `has_site_title` is already documented as meaning "the **_user_** has set a blog name"

https://github.com/Automattic/tracks-events-registration/blob/d1c88fa83d9b081e7441cfdaed9c671e39ff739d/events/calypso_signup_site_options_submit.event.json#L31-L32

Back when the prop was added the user was never shown an auto-generated site name. Leaving it blank was how you got the default name. Our new flow changes things up a bit.

Either:
- We find out if anyone is actively monitoring `has_site_title`, and if not then we can that prop description to "was site title empty"
- We don't add the new `has_changed_title` prop and instead rework the logic that sets the `has_site_title` prop so that it it's actually based on **_user_** sets the blog name

mpkelly (approved):
It's working as expected. 

image

Edit. Just saw the comment from @p-jackson. How do we check the first option?

zaguiini:
>How do we check the first option?

I already did by looking at the funnels that this event is included. We're not using the property in any funnel, so it's fine to re-use it. I'll update the description and use the existing property.

PR #77871, opened 2023-06-07
Adjust the mobile style of the sharing modal

### LINKED ISSUE #77786: Sharing modal looks off visually in mobile viewport
### Quick summary

_No response_

### Steps to reproduce

1. Have a site, which has already published a post so that you don't get other onboarding modals
2. Publish a post, see "sharing modal"
3. Test mobile screen width:

Screenshot 2023-06-05 at 09 26 34

Scroll up/down:

Screenshot 2023-06-05 at 09 26 45

Desktop version is fine:

Screenshot 2023-06-05 at 09 28 36

Visually better mobile version:

- Consider removing the image, or change it to different more mobile optimized version. It's silly to need to scroll on mobile.
- Consider removing most of the core Modal component overrides; the core modal is and will get iterated, so more overrides you do, the harder it is to manage those and things will break.
- Not sure what happened to header toolbar compared to core modals where it's visible behind the scrim
- Core version has fixed header (see screenshot below) but the sharing modal doesn't have that:

    https://github.com/Automattic/wp-calypso/blob/b8062d6f337ba951807da5cbe44e8196e3ecb340/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/index.tsx#L169

    Screenshot 2023-06-05 at 09 42 54

### What actually happened

Visually things look off.

### Impact

All

### Available workarounds?

Simple, Atomic

### Logs or notes

- Firefox browser

roo2:
Fixes https://github.com/Automattic/wp-calypso/issues/77786

Adjusts height of modal components so that the modal will not require scrolling on mobile devices 
* I targeted iPhone SE which is 375w*667h
I Adjusted font styles to match the mobile design of the videopress modal.
I left the desktop version as it is.

**Before:**
Screen Shot 2023-06-07 at 11 13 29 am

**After:**
Screen Shot 2023-06-07 at 11 01 32 am

**Videopress modal for reference.**
Screen Shot 2023-06-07 at 11 01 41 am

### Testing instructions
* Sandbox a test site.
* Sync changes to your sandbox with `cd apps/editing-toolkit && yarn dev --sync`
* Publish a new post
* Using the mobile emulator in chrome dev tools, select iphone SE (and test other resolutions)

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/style.scss
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-block-editor-nux/src/sharing-modal/style.scss
@@ -2,9 +2,6 @@
 @import "@wordpress/base-styles/mixins";
 
 .wpcom-block-editor-post-published-sharing-modal {
-	@media only screen and (max-width: 600px) {
-		border-radius: 0;
-	}
 	.components-modal__content {
 		margin-top: 0;
 		padding-bottom: 0;
@@ -17,6 +14,7 @@
 				p a {
 					color: var(--color-text);
 					text-decoration: underline;
+					white-space: nowrap;
 
 					&:hover {
 						color: var(--wp-components-color-accent, var(--wp-admin-theme-color, #007cba));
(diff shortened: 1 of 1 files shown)


simison:
Better! The Illustration doesn't work supper well in how it takes the empty space (too high vs wide), but it's already better so I'd ship this and check if you can get a better illustration for mobile specifically.

roo2:
I think there was quite a bit of back and forth on the illustration and now we've lost our designer Ollie, I would personally think we're better off just sticking with that current illustration despite it not being the best fit for mobile.

PR #77857, opened 2023-06-06
Launchpad: Added complementary styles to Customer home launchpad

nelsonec87:
Related to https://github.com/Automattic/wp-calypso/issues/77285

## Proposed Changes

* Added title, progress and styles to the Customer Home Launchpad component.

## Testing Instructions

* Change this line on `client/my-sites/customer-home/locations/primary/index.jsx:82` from:

`[ TASK_SITE_SETUP_CHECKLIST ]: SiteSetupList,` to `[ TASK_SITE_SETUP_CHECKLIST ]: LaunchpadKeepBuilding,`

* Open the customer home page and compare the designs with the result

Figma:
image

Screenshot:
image

PR #77872, opened 2023-06-07
Launchpad: Add taskFilter to Launchpad component

nelsonec87:
Depends on https://github.com/Automattic/wp-calypso/pull/77857

## Proposed Changes

* Added `taskFilter` prop to `<Launchpad />` component, so we can change the tasks that are shown (like it's done on Stepper)
* Added unit tests to test the filter

## Testing Instructions

* Visual inspection
* `yarn run test-packages launchpad`

### DIFF
--- a/packages/launchpad/src/launchpad.tsx
+++ b/packages/launchpad/src/launchpad.tsx
@@ -1,6 +1,6 @@
 import { useLaunchpad } from '@automattic/data-stores';
 import Checklist from './checklist';
-import { Task } from './types';
+import type { Task } from './types';
 
 export interface LaunchpadProps {
 	siteSlug: string;
--- a/packages/launchpad/src/launchpad.tsx
+++ b/packages/launchpad/src/launchpad.tsx
@@ -1,16 +1,25 @@
 import { useLaunchpad } from '@automattic/data-stores';
 import Checklist from './checklist';
+import { Task } from './types';
 
 export interface LaunchpadProps {
 	siteSlug: string;
 	checklistSlug?: string | 0 | null | undefined;
 	makeLastTaskPrimaryAction?: boolean;
+	taskFilter?: ( tasks: Task[] ) => Task[];
 }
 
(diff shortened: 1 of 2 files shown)


packages/launchpad/src/launchpad.tsx
```diff
@@ -1,15 +1,19 @@
 import { useLaunchpad } from '@automattic/data-stores';
 import Checklist from './checklist';
+import { Task } from './types';
```
daledupreez: This looks like it should be a type import.
```suggestion
import type { Task } from './types';
```

packages/launchpad/src/test/launchpad.tsx
```diff
@@ -0,0 +1,46 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import Launchpad from '../launchpad';
+import '@testing-library/jest-dom';
+import { Task } from '../types';
```
daledupreez: This also looks like it should be a type import.
```suggestion
import type { Task } from '../types';
```

packages/launchpad/src/test/launchpad.tsx
```diff
@@ -0,0 +1,46 @@
...
+	describe( 'when taskFilter is provided', () => {
+		it( 'then render only tasks returned from the callback', () => {
+			const filter = ( tasks: Task[] ) => {
+				// return only the first task
+				return [ tasks[ 0 ] ];
+			};
+
+			render( <Launchpad siteSlug="any site" taskFilter={ filter } /> );
+			const checklistItems = screen.queryAllByRole( 'listitem' );
+			expect( checklistItems.length ).toBe( 1 );
```
daledupreez, line 43: Might it be worth verifying that the single task is the one we expect it to be? I think that would be a better way to verify that we are actually running our filter.

daledupreez (approved):
This looks good to me, but I do have two nits RE type imports, as well as a question about whether we can make one of the tests more specific.

daledupreez (approved):
The updates also look to me!

PR #77877, opened 2023-06-07
Hosting import flow: reset onboarding store when starting flow to avoid managed_subdomain_not_available API error.

mpkelly:
Related to https://github.com/Automattic/wp-calypso/pull/77588#issuecomment-1576189253

## Proposed Changes

* Reset the onboarding store when starting the new import flow. This avoids using a stale domain from a previous flow which then produces a `400` API response. 

## Testing Instructions

* On trunk, go to `/setup/newsletter` or `/setup/link-in-bio` and choose the free domain and Free plan. 
* Wait until you're sent to Launchpad, then change the URL to `/setup/import-hosted-site` and use a JN site (I used https://dark-potter.jurassic.ninja/) to start the import flow. Continue through the flow and you should see an error:

Screenshot 2566-06-07 at 11 54 12

Screenshot 2566-06-07 at 11 54 19

The problem is that the payload contains `blog_name` with value `test9042.home.blog`, the free domain from the newsletter flow. 

* Checkout this branch or use Calypso Live
* Repeat the process above and confirm that you confirm the flow (at least until it tells your to upgrade to Business)

This PR fixes the bug, but I wonder why `siteTitle` from a previously completed flow is still in the `dependencyStore` anyway? Can you see a better way to fix it?

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/import-hosted-site.ts
+++ b/client/landing/stepper/declarative-flow/import-hosted-site.ts
@@ -1,6 +1,7 @@
 import { isEnabled } from '@automattic/calypso-config';
 import { IMPORT_HOSTED_SITE_FLOW } from '@automattic/onboarding';
 import { useDispatch } from '@wordpress/data';
+import { useEffect } from 'react';
 import { ImporterMainPlatform } from 'calypso/blocks/import/types';
 import MigrationError from 'calypso/landing/stepper/declarative-flow/internals/steps-repository/migration-error';
 import { ProcessingResult } from 'calypso/landing/stepper/declarative-flow/internals/steps-repository/processing-step/constants';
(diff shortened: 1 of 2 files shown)


ouikhuan:
@mpkelly , I'm still seeing the error after applying this fix, but I got a blog_name_exists error instead. The domain I picked is a wordpress.com sub-domain by the way, but I think the issue is the same.

https://github.com/Automattic/wp-calypso/assets/4074459/28f03236-5af5-4e70-a000-c1313f367332

Also, since I already have this earlier, it is still loaded when I go straight to the `import-hosted-site` flow. So I think the `removeSiteSlugDependency` might not be getting called at all.

mpkelly:
Thanks, @ouikhuan. I also still see the issue when I don't manually enter the URL. So I think `removeSiteSlugDependency` is only called when loading the entire UI. 

I still think the fix is to remove the stale `blog_name` from the previous flow, so I will look at doing that once the Newsletter flow completes. 

Edit. This bug happens when completing the Link in Bio flow also. It doesn't happen when using `/start`, though. So it looks like some logic is missing from the tailored flows.

mpkelly:
I compared `/start/free`, `/setup/link-in-bio`, and `/setup/newsletter`. The only thing I could see was that the free flow had a reset call `resetOnboardStore`. I added this same reset to the new import flow, and it seems to fix the problem, although it feels like a framework step rather than something to do at the flow level. Let me know your thoughts during the review.

ouikhuan (approved):
This looks good to me, tested and work as expected! Looking into the codebase I also see free flow is doing the same, some other files are using it in a hook, too. So I think we're fine here, nice fix! 👍

PR #77897, opened 2023-06-07
Hosting Configuration: Only show data center picker when eligible

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/2672

## Proposed Changes

Only shows the data center picker when the site is eligible for transfer and has no blocking holds:

image

image

Previously, the data center picker would show if there was a blocking hold:

image

## Testing Instructions

1. Create a new Business site and navigate to 'Hosting Configuration'.
2. Click 'Activate'.
3. Verify the data center picker appears and works as expected.
4. Take the site Atomic.
5. After the site is Atomic, visit `/hosting-config/activate/<site-slug>`.
6. Verify the data center picker doesn't appear.

### DIFF
--- a/client/blocks/eligibility-warnings/index.tsx
+++ b/client/blocks/eligibility-warnings/index.tsx
@@ -212,7 +212,7 @@ export const EligibilityWarnings = ( {
 				</CompactCard>
 			) }
 
-			{ showDataCenterPicker && (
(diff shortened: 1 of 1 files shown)


kozer (approved):
@danielbachhuber , works as expected!

PR #77900, opened 2023-06-07
Hosting import flow: allow picking an existing site

zaguiini:
Closes https://github.com/Automattic/dotcom-forge/issues/2640.

## Proposed Changes

The user should be able to pick an existing site as a target for the migration.

## Testing Instructions

In an account with zero sites, check that we're creating one in the background after confirming that we want to import an existing site.

In an account with existing sites, check that the site picker is shown after agreeing to import an existing site. Check that it's possible to use an existing site, as well as create a new one by clicking the "create a new site" links.


mpkelly (approved):
This is working for me for the following:

1. No sites - automatically create a site ✅ 
2. Has sites - choose an existing site ✅ 
3. Has sites - decide to skip and create a new site ✅ 

The only part I didn't like is when you're on the "Upgrade to Business" screen and hit back, you go to `/setup/import-hosted-site/import?siteSlug=pantropicalaphid94.wordpress.com`, and it has a `siteSlug`. If you want to pick a different site, you won't be able to. I'm not sure what's best to do here, but we probably want to remove this param in case the user picked the wrong site. At the same time, it's probably not a big issue, so I will approve this.

zaguiini:
>If you want to pick a different site, you won't be able to. 

Yeah, it should go back to a cleared state (i.e. with no params). I'll address this in a follow-up PR.

mpkelly:
> > If you want to pick a different site, you won't be able to.
> 
> Yeah, it should go back to a cleared state (i.e. with no params). I'll address this in a follow-up PR.

@zaguiini Seeing as though you didn't merge yet, I will address this in a commit now.

mpkelly:
@zaguiini, I added https://github.com/Automattic/wp-calypso/pull/77900/commits/d0de9f0913fbc8921001c77408a42dfb4f2379e8 to remove the `siteSlug` on going back but keep existing params. This allows the user to change sites if they've made a mistake.

PR #66050, opened 2022-07-28
Update blocked plugins that break on Atomic and violate ToS

Nic-Sevic:
Update blocked plugins to include additional plugins that break on Atomic, violate ToS, and do not need outreach.
Based on p9F6qB-78n-p2

#### Proposed Changes

adds the following plugins to the incompatibility list:
vamtam-offline-jetpack
plugin-detective
wp-hide-security-enhancer
redis-cache
wp-phpmyadmin-extension
antihacker
deactivate-xml-rpc-service
disable-xml-rpc
disable-xml-rpc-api
disable-xml-rpc-fully
disable-xml-rpc-unset-x-pingback
manage-xml-rpc
sg-security
simple-xml-rpc-disabler
stopbadbots
wee-remove-xmlrpc-methods
wp-hide-security-enhancer

#### Pre-merge Checklist

- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have you sent any new strings for translation (PCYsg-1vr-p2) ASAP?

Related to #

PR #77906, opened 2023-06-07
Syncing incompatible plugins

RafaelFunchal:
Syncing incompatible plugins with WPcomSH/class-jetpack-plugin-compatibility.php


Nic-Sevic:
@danielbachhuber Would you be able to help us resolve this and get the change merged in? We don't have access to teamCity to try to resolve the build failure. Also, I added this to the Serenity - Forge Web Engineering group as Forge Build Group is now deprecated. If this is not the right place, please let me know 🙏 

This is related to a plugin we removed from the incompatible list in wpcomsh here - robo-gallery: https://github.com/Automattic/wpcomsh/pull/1384
At the time the calypso part was missed so that's being added now
It's the same goal as this previous PR: https://github.com/Automattic/wp-calypso/pull/66050

It looks like @RafaelFunchal also added some other plugins which have been added to the wpcomsh list but not the calypso list since (@hideandgeek404 pinging you for awareness that a partner calypso pr is needed on these as well)

ultimate-reset
jetpack-backup
performance-lab
wp-optimizer
wp-security-hardening
flo-launch
blogmatic-poster
blogmatic
db-access-adminer
pexlechris-adminer
trafficzion
wpstagecoach
civicrm

The changes these create are in the plugin marketplace. Robo Gallery should no longer showa blocked status and the additions should now show it, like so:
2023-06-08 at 11 39

danielbachhuber:
> Would you be able to help us resolve this and get the change merged in?

@Nic-Sevic Yep, happy to help.



> Also, I added this to the Serenity - Forge Web Engineering group as Forge Build Group is now deprecated. If this is not the right place, please let me know

Serenity isn't quite the right board (they're another team in the Forge Web Engineering Group). I've added it to our Team YOLO board.

> This is related to a plugin we removed from the incompatible list in wpcomsh here - robo-gallery: [Automattic/wpcomsh#1384](https://github.com/Automattic/wpcomsh/pull/1384) At the time the calypso part was missed so that's being added now It's the same goal as this previous PR: #66050
> 
> It looks like @RafaelFunchal also added some other plugins which have been added to the wpcomsh list but not the calypso list since

Sounds good, all of this is fine by me.

PR #68837, opened 2022-10-10
Agency Dashboard: Add Request to Check Sites Connection Health

wilmersondasilva:
Currently, the connection health status is populated in the site profiles index during an audit, which means that there might be a significant delay between the time site went down and the index update. More context can be found pedEI5-2s-p2 here.

In WordPress.com, we use the `/test-connection` endpoint that makes an API request to the remote site to verify the connection's health. 

The API was changed so we can use this endpoint also on Calypso green.

**Backend task**: 1202619025189113-as-1203022769710885

#### Proposed Changes

- In the Calypso dashboard, the connection status of each site will be fetched.
- If the connection status of the site is not healthy, an error message will appear to the user.

#### Testing Instructions

- Make sure to have a partner account.
- Create an ephemeral site
- Connect Jetpack
- Ensure the site is now visible in the [Agency Dashboard](https://calypso.live/?image=registry.a8c.com/calypso/app:build-45797&env=jetpack)
- Delete the ephemeral site
- Visit the site to ensure it is gone
- Refresh the [Agency Dashboard](https://calypso.live/?image=registry.a8c.com/calypso/app:build-45797&env=jetpack) page
- Verify that there is now an indication about the unhealthy site connection

#### Screenshots

image

#### Pre-merge Checklist

- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?

Related to 1202619025189113-as-1203022769710885

PR #77910, opened 2023-06-07
Jetpack Cloud: Show more accurate value in site counter

mattgawarecki:
This pull request aims to improve the accuracy of the count displayed in the "All My Sites" component in the Jetpack Cloud sidebar, by retrieving a different/more appropriate value.

Resolves `1202619025189113-as-1203147992256213`, `1202619025189113-as-1203225745976434`.
Issue referenced in #68837.

## Proposed Changes

* Add a new selector `getCurrentUserJetpackVisibleSiteCount` to retrieve the value of `jetpack_visible_site_count` on the `user` state tree.
* Use the aforementioned new selector in Jetpack Cloud environments where the `realtime-site-count` feature flag is disabled or not set (i.e., Horizon).
* Add documentation inside the `connect` function to explain how site count is determined in various circumstances.

## Known issues

Only active sites are retrieved and filtered when `realtime-site-count` is enabled, but `jetpack_visible_site_count` counts both active and inactive sites. This means that in cases where someone with post editing capabilities logs in, the value in their site counter will vary in environments where the feature flag is in different states.

_**Example:** I have post edit capabilities on 5 Jetpack sites. One of these sites is inactive, meaning it hasn't communicated with Jetpack's servers in the last 30 days. In environments with `realtime-site-count` enabled, the site counter will show 4; however, environments with `realtime-site-count` disabled will show 5._

### Testing the fix

1. Visit cloud.jetpack.com in both production/staging, and in your testing environment (e.g., this PR's live branch). Complete the following steps in each environment. *(**NOTE:** If you use the live branch, `realtime-site-count` is disabled by default.)*
2. Note the number displayed in the site counter in your sidebar.
3. Add the query parameter `flags=realtime-site-count` to your browser's URL to enable real-time site counting, if necessary. Note again the displayed count.
4. Add the query parameter `flags=-realtime-site-count` (notice the leading minus sign) to your browser's URL to disable real-time site counting, if necessary. Note again the displayed count. It may be higher than when the flag was enabled.
5. Verify that the difference in the displayed count in your test environment is smaller or non-existent, compared to the difference in production/staging.

### Testing for regression issues

1. Visit wordpress.com in both production/staging, and in your testing environment.
2. Complete the steps in the previous test case, ensuring that the values are identical between both environments.

image

### DIFF
--- a/client/blocks/all-sites/index.jsx
+++ b/client/blocks/all-sites/index.jsx
@@ -6,7 +6,11 @@ import PropTypes from 'prop-types';
 import { Component } from 'react';
 import { connect } from 'react-redux';
 import Count from 'calypso/components/count';
-import { getCurrentUserVisibleSiteCount } from 'calypso/state/current-user/selectors';
+import isJetpackCloud from 'calypso/lib/jetpack/is-jetpack-cloud';
+import {
+	getCurrentUserJetpackVisibleSiteCount,
+	getCurrentUserVisibleSiteCount,
+} from 'calypso/state/current-user/selectors';
 import getSites from 'calypso/state/selectors/get-sites';
 
 import './style.scss';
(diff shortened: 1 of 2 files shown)


oikeme (approved):
Changes work as expected from the testing instructions and with the updates to the visible site count. LGTM

PR #77916, opened 2023-06-07
SSR Footer: Use an empty state for Automattic branding before set the final value

### LINKED ISSUE #77799: Marketplace: SSR Theme Details footer is random
The footer uses a random text to put next to `an Automattic` as `medley` or `experiment`.
This causes the client (random) version to always be different from the server (fixed) version, causing the client version to be picked. 

As we want the page to be SSR'd, we have to find a way to have the same version for the client and server, maybe caching the client version.

More context on #77696

WBerredo:
Related to #77696
Fixes #77799

## Proposed Changes

SSR requires the initial state to be the same for the client and server version, thus having a random text on the footer makes it always use the client version.

In order to align the initial state for both cases, the initial state for the Automattic branding will be only the "Automattic" part, and the text will be added later.

CleanShot 2023-06-07 at 15 47 11

## Testing Instructions

**Trunk**

* On a logged-out tab
* Go to a page that displays the footer. Ex: `/theme/miniature`
* Reload the page
* Open the `console` and you SHOULD see an error stating there is a mismatch between the client and server version and you'll see the client version on the screen

**This branch**

* On a logged-out tab
* Go to a page that displays the footer. Ex: `/theme/miniature`
* Scroll down to the bottom of the page
* Reload the page
* You should see only `Automattic` at first, and then the correct branding
* Open the `console` and you SHOULD NOT see any errors 

CleanShot 2023-06-07 at 15 47 11

## Pre-merge Checklist


packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
```diff
@@ -472,7 +473,14 @@ const UniversalNavbarFooter = ( {
 	const isEnglishLocale = useIsEnglishLocale();
 	const pathNameWithoutLocale =
 		currentRoute && removeLocaleFromPathLocaleInFront( currentRoute ).slice( 1 );
-	const automatticBranding = useAutomatticBrandingNoun();
+	const [ automatticBranding, setAutomatticBranding ] = useState( { article: '', noun: '' } );
```
jsnajdr, line 480: Here `article` and `noun` should be replaced with `prefix` and `suffix`. Only in English there is article before and a noun after. In Spanish, for example (https://wordpress.com/es/themes/miniature), there is only prefix:

Screenshot 2023-06-08 at 9 26 22

Also it seems that the translator didn't understand that Automattic is a company name. 🙁

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
```diff
@@ -472,7 +473,14 @@ const UniversalNavbarFooter = ( {
...
+	const currentAutomatticBranding = useAutomatticBrandingNoun();
+
+	useLayoutEffect( () => {
+		setAutomatticBranding( currentAutomatticBranding );
+		// Automattic Branding should be set only once after the page load
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [] );
```
jsnajdr: Here a new random slogan is generated on every render, but used only once in a mount effect. That's weird and the ESLint warning warns about that weirdness.

It would be better to extract the `useTranslate` function out of `useAutomatticBrandingNoun`, and `useTranslate` was also the only reason why it was a hook:
```js
useEffect( () => {
  setAutomatticBranding( getAutomatticBrandingNoun( translate ) );
}, [ translate ] );
```

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
WBerredo: > Here article and noun should be replaced with prefix and suffix. Only in English there is article before and a noun after.

I agree, but I would prefer to keep this PR focused on guaranteeing an initial state to eliminate the warning. As it is not a regression introduced in this PR I would handle it in a separate issue.

Prod version:

CleanShot 2023-06-08 at 12 32 25@2x

jsnajdr: > I would handle it in a separate issue.

Yes that's fine :+1:

> Prod version:

That's a different random selection, with the same error: the "automático" word shouldn't be there.

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
WBerredo: Updated at https://github.com/Automattic/wp-calypso/pull/77916/commits/1425f4aefe75704dba82e9839706210f7c355ff2

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
WBerredo: PR created here: #78018

jsnajdr (approved):
Looks good, thanks for addressing all feedback :+1:

gcsecsey (approved):
Looks good to me too, and works as described.

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
```diff
@@ -471,10 +473,15 @@ const UniversalNavbarFooter = ( {
...
 	const localizeUrl = useLocalizeUrl();
 	const locale = useLocale();
+	const translate = useTranslate();
 	const isEnglishLocale = useIsEnglishLocale();
 	const pathNameWithoutLocale =
 		currentRoute && removeLocaleFromPathLocaleInFront( currentRoute ).slice( 1 );
-	const automatticBranding = useAutomatticBrandingNoun();
+	const [ automatticBranding, setAutomatticBranding ] = useState( { article: '', noun: '' } );
+
+	useLayoutEffect( () => {
```
noahtallen, line 482: By the way `useLayoutEffect` has been causing warnings in the React 18 PR #77046. There's more info about it here: https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

But the core idea is that since effects don't run on the server in SSR, `useLayoutEffect` can be problematic because its purpose is to update content _before_ people see it. When SSR generates a layout without using the effect, that content wouldn't include what the layout effects expects to add.

Thankfully, this isn't a problem for us because the default content is just the Automattic logo, so we have a graceful fallback.

What I did to fix the warning in #77046 was to conditionally define `useLayoutEffect` in SSR mode like this:

```js
const useIsomorphicEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect;
```

And it seems to resolve the issue. Just wanted to drop a comment here as well.

noahtallen: Interestingly, it looks like this warning will be removed in a future React version: https://github.com/facebook/react/pull/26395

jsnajdr: This particular code would be better off if it used just `useEffect`. Without SSR, the benefit of `useLayoutEffect` would be that the version of the page with the incomplete footer would never be actually painted. The first paint would occur only after the layout effect sets the footer to the final random value.

But with SSR, the initial version is always painted, because it's a part of the initial HTML markup. `useLayoutEffect` doesn't make any difference compared to `useEffect`.

PR #77920, opened 2023-06-07
Checkout: Convert ManagedContactDetailsFormFields to TypeScript

sirbrillig:
## Proposed Changes

This converts the `ManagedContactDetailsFormFields` component to TypeScript and in the process refactors it to render explicit input components rather than using React's `createElement()`.

Before and after this change, the component should look something like this:

Screenshot 2023-06-12 at 7 10 36 PM

## Testing Instructions

- Visit the calypso home page for a site and click "Upgrades" > "Domains" in the sidebar.
- Follow the flow to add a domain to your shopping cart.
- You should end up in checkout.
- Click "Edit" on the contact details step if it is already completed.
- Verify that the step looks the same before and after this change.
- Change the viewport width to be sure that the form looks good at different widths.
- Change the country (eg: Canada and France have quite different fields) to verify that the form looks good with different sets of fields.

### DIFF
--- a/client/components/domains/contact-details-form-fields/managed-contact-details-form-fields.tsx
+++ b/client/components/domains/contact-details-form-fields/managed-contact-details-form-fields.tsx
@@ -38,7 +38,7 @@ import './style.scss';
 const debug = debugFactory( 'calypso:managed-contact-details-form-fields' );
 
 export interface ManagedContactDetailsFormFieldsProps {
-	eventFormName: string;
+	eventFormName?: string;
 	contactDetails: DomainContactDetailsData;
 	contactDetailsErrors: DomainContactDetailsErrors;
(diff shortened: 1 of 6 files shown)


spraveenitpro (commented):
I did some static analysis and it LGTM, Great work replacing the propTypes

spraveenitpro:
**Functional Testing:**

<details>
<summary>Before Patch</summary>


</details>


<details>
<summary>After patch:</summary>


</details>

*  Verify that the step looks the same before and after this change. ✅
*  Change the viewport width to be sure that the form looks good at different widths. ✅
*  Change the country (eg: Canada and France have quite different fields) to verify that the form looks good with different sets of fields. ✅

spraveenitpro (approved):

PR #77925, opened 2023-06-07
Adds the credits section to the new Advertising page

sbarbosa:
## Proposed Changes

Recently we added the possibility of using credits to promote a post. That functionality already lives inside the Advertising page. 
This PR aims to include that new section in the new design.

**Desktop view:**
Screenshot 2023-06-07 at 19 52 19

**Mobile view:**
Screenshot 2023-06-07 at 19 52 35

**Mobile navigation when you have all tabs enabled:**

## Testing Instructions

You will need to have credits for your account; ping me if you need some.

* Check out this branch or enable the `promote-post/redesign-i2` flag in the live preview URL
* Go to Tools->Advertising
* Check that the credit tab is displayed on desktop at the right of the navigation bar
* Click on that tab, and check that the credits page is shown correctly
* Change to a mobile viewport and check that the Credits tabs is shown next to the Campaigns tab
* Click it and verify that the page is shown correctly

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/promote-post-i2/components/credit-balance/index.tsx
+++ b/client/my-sites/promote-post-i2/components/credit-balance/index.tsx
@@ -6,8 +6,9 @@ interface Props {
 	balance?: number;
 }
 
-const CreditBalance = ( { balance }: Props ) => {
-	if ( balance === undefined || balance === 0 ) {
+const CreditBalance = ( { balance = 0 }: Props ) => {
+	// Hide the section if balance is invalid or is 0
+	if ( ! balance || balance === 0 ) {
 		return null;
 	}
 
--- a/client/my-sites/promote-post-i2/components/credit-balance/index.tsx
+++ b/client/my-sites/promote-post-i2/components/credit-balance/index.tsx
@@ -1,0 +1,37 @@
+import { memo } from '@wordpress/element';
+import { translate, numberFormat } from 'i18n-calypso';
+import InlineSupportLink from 'calypso/components/inline-support-link';
+
+interface Props {
+	balance?: number;
+}
+
+const CreditBalance = ( { balance }: Props ) => {
+	if ( balance === undefined || balance === 0 ) {
+		return null;
+	}
+
+	return (
+		<div className="promote-post-i2__aux-wrapper">
+			<div className="empty-promotion-list__container">
+				<h3 className="empty-promotion-list__title wp-brand-font">
+					${ numberFormat( balance, 2 ) }
+				</h3>
+				<p className="empty-promotion-list__body">
+					{ translate(
+						'Available credits that will be automatically applied toward your next campaigns. {{learnMoreLink}}Learn more.{{/learnMoreLink}}',
+						{
+							components: {
+								learnMoreLink: (
+									<InlineSupportLink supportContext="blaze_credits" showIcon={ false } />
+								),
+							},
+						}
+					) }
+				</p>
+			</div>
+		</div>
+	);
+};
+
+export default memo( CreditBalance );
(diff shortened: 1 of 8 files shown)


client/my-sites/promote-post-i2/main.tsx
```diff
@@ -126,12 +133,20 @@ export default function PromotedPosts( { tab }: Props ) {
...
-			itemCount: totalCampaignsUnfiltered || null,
+			itemCount: totalCampaignsUnfiltered,
+		},
+		{
+			id: 'credits',
+			name: translate( 'Credits' ),
+			className: 'pull-right',
+			itemCount: creditBalance,
+			isCountAmount: true,
+			enabled: creditBalance !== undefined && creditBalance !== 0,
```
timur987: Just thinking out loud. It's interesting whether it will be even more right to use `creditBalance > 0` instead of `creditBalance !== 0`. Assuming that dealing with floating point would work better with `>` operator. But the current solution looks workable either.

client/my-sites/promote-post-i2/main.tsx
```diff
@@ -235,24 +252,37 @@ export default function PromotedPosts( { tab }: Props ) {
...
+				<>
+					<PageViewTracker
+						path={ getAdvertisingDashboardPath( '/:site/credits' ) }
+						title="Advertising > Credits"
+					/>
+					<CreditBalance balance={ creditBalance } />
+				</>
+			) }
+
+			{ /* Render products tab */ }
```
timur987: Perhaps it should be `Render posts tab`.

client/my-sites/promote-post-i2/components/credit-balance/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { memo } from '@wordpress/element';
+import { translate, numberFormat } from 'i18n-calypso';
+import InlineSupportLink from 'calypso/components/inline-support-link';
+
+interface Props {
+	balance?: number;
+}
+
+const CreditBalance = ( { balance }: Props ) => {
+	if ( balance === undefined || balance === 0 ) {
```
timur987: Wonder could we swap the condition this way:
```js
if ( balance > 0 ) {
    return (
        // The balance markup
    );
}

return null;
```

timur987 (approved):
Thanks, @sbarbosa!

The code LGTM and works as advertised.
Left several minor non-blocking suggestions though.

client/my-sites/promote-post-i2/main.tsx
sbarbosa: You are right; we won't be having negative credits, so its safe to make the change to `>` instead of the equals. 
I need to maintain the `creditBalance !== undefined` to differentiate the case between an invalid amount (NaN) and a blank response from the API. 

If by any reason the backend start sending invalid balances, the interface will display the tab but without an amount, and that will give us a clue where the error could be.

sbarbosa: I change my mind regarding showing the page if we aren't able to parse the balance that the backend gives us. 

I will go for your solution and only show the tab/section if we have a valid amount.

client/my-sites/promote-post-i2/components/credit-balance/index.tsx
sbarbosa: I will change the condition to also exclude the case for invalid numbers (NaN).
Regarding the change of order, I think it is easier to follow the code if we take these conditions on top (like we do with loadings). 

From my point of view, if I scroll down to the bottom of the component, I would like to see the main render of that component. But it may be just my way of thinking.

sbarbosa:
@timur987 , I applied some changes to the code. I took your advice and used the `credits > 0` to enable the tab, and did a minor change on the credits section to hide it if you have an invalid amount.

Let me know what you think! Thanks!

timur987:
@sbarbosa, changes LGTM. Checked on both user with credits and without. It works like a charm!

PR #77940, opened 2023-06-08
Look up blaze credit balance using site id instead of user id

lsl:
Fixes https://github.com/Automattic/dotcom-forge/issues/2641

## Proposed Changes

* Use selected site id instead of current user id for fetching blaze credits.

## Testing Instructions

* http://calypso.localhost:3000/advertising/testingtesterson32.wordpress.com/campaigns?flags=-promote-post/redesign-i2 (modify your site url)

Should no longer see 404 errors with Invalid Site Specified coming from centralize.php (line link in issue) and /credits/balance endpoint should be responding with e.g. 0 credits.

### DIFF
--- a/client/data/promote-post/use-promote-post-credit-balance-query.ts
+++ b/client/data/promote-post/use-promote-post-credit-balance-query.ts
@@ -1,18 +1,17 @@
 import { useQuery } from '@tanstack/react-query';
 import { useSelector } from 'react-redux';
 import { requestDSP } from 'calypso/lib/promote-post';
-import { getCurrentUser } from 'calypso/state/current-user/selectors';
+import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 
 const useCreditBalanceQuery = ( queryOptions = {} ) => {
-	const currentUser = useSelector( getCurrentUser );
-	const wpcomUserId = currentUser?.ID;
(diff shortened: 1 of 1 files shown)


candy02058912 (commented):
My situation is a bit weird:
Tried `trunk` and `update/blaze-credits`, with a11n and non-a11n accounts, the credit balance are all showing in both cases.

lsl:
I'm not 100% sure this is the solution/problem, I added some logging and saw my user name in the logs where there should be a site id, I didn't have time to trace it but swapping the user id for site id did thread through to my logging correctly.

Is it loading a balance for someone else maybe? Your user ID might match a valid site ID. A good check would be to give your test user a balance and check it's showing the right amount.

candy02058912 (approved):
> Is it loading a balance for someone else maybe? Your user ID might match a valid site ID. A good check would be to give your test user a balance and check it's showing the right amount.

I think this may be it?

LGTM 👍  I tested loading credits into the non-a11n account and confirmed that this PR works.
nona11n
nona11n

lsl:
@candy02058912 what do you see when loading without the patch? (e.g. in prod)

candy02058912:
> @candy02058912 what do you see when loading without the patch? (e.g. in prod)

empty

PR #77955, opened 2023-06-08
Inline Help/Help Center: use the same useHelpSearchQuery hook

yoavf:
The default search results for the inline help are not currently loading: 

Screenshot 2023-06-08 at 13 18 01

Reported in p9Jlb4-7c4-p2#comment-8216

## Proposed Changes

* This updates calypso useHelpSearchQuery hook to use the new endpoint (see https://github.com/Automattic/wp-calypso/pull/76734/ )
* It then removes the custom hook in the Help Center, and uses the calypso hook instead.
* Because the `/help` screen search results need data that the new endpoint does not provide, this also brings back the old hook as `-v1` and uses it for this screen.

## Testing Instructions

* In calypso home, make sure you see the default search results in the help widget:

Screenshot 2023-06-08 at 13 19 32

* In `/help` search results should include both Documentation and Forum links

* The Help Center default results (and search results) should stay the same.

## Pre-merge Checklist


yoavf:
@alshakero : Does this make sense to you? I didn't want to just duplicate the hook in Calypso. 
I'll test this on atomic/JP of course

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,74 @@
...
+		onSuccess: async ( data ) => {
+			if ( ! data[ 0 ]?.content ) {
+				const newData = await Promise.all(
+					data.map( async ( result: SearchResult ) => {
+						const article: { [ content: string ]: string } = canAccessWpcomApis()
+							? await wpcomRequest( {
+									path: `help/article/${ result.blog_id }/${ result.post_id }`,
+									apiNamespace: 'wpcom/v2/',
+									apiVersion: '2',
+							  } )
```
alshakero: Pretty rare case, but I think the `async/await` is complicating things here:

```js
Promise.all(data.map(result =>
	wpcomRequest({
		path: `help/article/${result.blog_id}/${result.post_id}`,
		apiNamespace: 'wpcom/v2/',
		apiVersion: '2',
	}).then(article => ({ ...result, content: article.datum }))
))
```

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,74 @@
...
+							  } as APIFetchOptions );
+						return { ...result, content: article.content };
+					} )
+				);
+				queryClient.setQueryData( [ 'help', search ], newData );
+			}
+		},
+		refetchOnWindowFocus: false,
+		refetchOnMount: true,
+		enabled: true,
```
alshakero, line 77: Enabled by default
```suggestion
```

alshakero (changes requested):
Thanks for the ping! I think this goes the opposite direction. We should [default to packages, unless we can't](p4TIVU-arC-p2). 

I'd improve the Help Center hook and use it in Calypso. Packages are typed and accessible from both places without pulling any extraneous dependencies.

When you import a package, you import only that, when you import something from Calypso, you really don't know what you're pulling with it.

client/data/help/use-help-search-query.ts
flootr: the same is true for `refetchOnMount`

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,90 @@
...
+			apiFetch( {
+				global: true,
+				path: `/help-center/fetch-post?blog_id=${ blogId }&post_id=${ postId }`,
+			} as APIFetchOptions );
+	}
+
+	return useQuery< any >( {
+		queryKey: [ 'help-center-search', search, sectionName ],
+		queryFn: fetchArticlesAPI,
+		onSuccess: async ( data ) => {
```
flootr, line 64: Please take into consideration that `onSuccess` callbacks for queries (not mutations!) are [deprecated](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#callbacks-on-usequery-and-queryobserver-have-been-removed) and will be removed in the next major version of `@tanstack/query` so it may be worthwhile to explore different solutions to fetching additional content.

simison:
If the PR works, the feedback here can probably addressed in a follow-up PR? The important thing would be getting a bug fixed sooner and optimization can wait.

simison (approved):
Looks to work well:

image

yoavf:
@simison 
> Looks to work well:

Yes, but then the `/help` screen was broken because it expects a different format for the data:
Screenshot 2023-06-18 at 12 33 33

To solve this asap I brought back the old hook in `use-help-search-query-v1.ts`.

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,90 @@
...
 
-	if ( search ) {
-		params.append( 'query', search );
+	// set up article list fetching
+	let fetchArticlesAPI: () => Promise< unknown >;
+	const params = `?query=${ encodeURIComponent( search ) }&locale=${ encodeURIComponent(
+		locale
+	) }&section=${ encodeURIComponent( sectionName ) }`;
+	if ( canAccessWpcomApis() ) {
+		fetchArticlesAPI = () =>
```
alshakero: I would make `fetchArticlesAPI` a function that has these `ifs` and `elses` inside it. In which case, you can put it outside the hook to avoid recreating it with every render.

```js
function fetchArticlesAPI() {
	if ( canAccessWpcomApis() ) {
		return wpcomRequest( {
				path: `help/search/wpcom${ params }`,
				apiNamespace: 'wpcom/v2/',
				apiVersion: '2',
		} );	
	} else {		
		return apiFetch( {
				global: true,
				path: `/help-center/search${ params }`,
			} as APIFetchOptions );
	}
}
```

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,90 @@
...
+				apiNamespace: 'wpcom/v2/',
+				apiVersion: '2',
+			} );
+	} else {
+		fetchArticleAPI = ( blogId: number, postId: number ) =>
+			apiFetch( {
+				global: true,
+				path: `/help-center/fetch-post?blog_id=${ blogId }&post_id=${ postId }`,
+			} as APIFetchOptions );
+	}
```
alshakero: +1.

client/data/help/use-help-search-query.ts
```diff
@@ -1,26 +1,90 @@
...
+							result.blog_id,
+							result.post_id
+						);
+						return { ...result, content: article.content };
+					}
+					return result;
+				} )
+			);
+
+			queryClient.setQueryData( [ 'help', search ], newData );
```
alshakero: Does this query key exist? I can't find it. If you're trying to alter the results of the query itself (and not another query), [select](https://tanstack.com/query/v4/docs/react/reference/useQuery#:~:text=select%3A%20(data%3A%20TData)%20%3D%3E%20unknown) is better, IMO.

alshakero (changes requested):
Found one issue with the code. Will test in a following iteration.

client/data/help/use-help-search-query.ts
yoavf: Thanks!  implemented

client/data/help/use-help-search-query.ts
yoavf: @alshakero Not sure I understand (this isn't something I've changed here - it's already in [the existing hook](https://github.com/Automattic/wp-calypso/blob/trunk/packages/help-center/src/hooks/use-help-search-query.ts))

alshakero: What this line does, it sets the data of a query that has the same key. Eg

```js
useQuery( [ 'help', search ], ... );
```

And I searched Calypso and it doesn't exist, so I imagine it's a leftover. It's probably not doing anything here.

alshakero (approved):
Works and looks great! Thank you so much for picking this up.

As for the stale query key, it's up to you.

client/data/help/use-help-search-query.ts
yoavf: Thanks @alshakero  - cleaned that up!

PR #77960, opened 2023-06-08
Assembler: The new site from LoTS won't land on Assembler if vpw is 960px

arthur791004:
Related to `tracks/funnel/view.php?42784/2023-06-01/to/2023-06-07`

## Proposed Changes

Referring to the funnels, there is a 50% drop rate after people select a plan but it's unreasonable. After investigation, the `PatternAssemblerCta` component uses the `useViewportMatch` hook to determine the viewport width, and it will include the minimums. However, we use the `isDesktop` function inside `getChecklistThemeDestination` to determine whether to go to the assembler screen after creating the new site, but it doesn't include the minimums.

image

So, this PR proposes to support the `>=960px` breakpoint on Calypso to align the behavior. In the long term, I assume we have to include minimums as the `break-*` mixins from Gutenberg also include minimums.

Note that People might still drop out after selecting a plan, and the main reason might be people selecting a non-free plan.

image

## Testing Instructions

* Go to the Theme Showcase without logging in, and ensure your viewport width is 960px
* Scroll down and select BCPA CTA
* On the User screen, log in or register a new one
* On the domain screen, select a domain
* On the plan screen, select a plan
* After the new site is created, ensure you land on the Assembler screen

## Pre-merge Checklist

### DIFF
--- a/packages/viewport/src/index.ts
+++ b/packages/viewport/src/index.ts
@@ -47,6 +47,7 @@ const isServer = typeof window === 'undefined' || ! window.matchMedia;
 
 const noop = () => null;
 
+export type QueryOption = { min?: number; max?: number };
 export type QueryItem = false | MinimalMediaQueryList | MediaQueryList;
 export type ListenerCallback = ( matches: boolean ) => void;
 export type UnsubcribeCallback = () => void;
@@ -66,7 +67,7 @@ function addListenerFunctions(
 	};
 }
 
-function createMediaQueryList( args?: { min?: number; max?: number } ): QueryItem {
+function createMediaQueryList( args?: QueryOption ): QueryItem {
 	const { min, max } = args ?? {};
 	if ( min !== undefined && max !== undefined ) {
 		return isServer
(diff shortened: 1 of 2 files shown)


packages/viewport/src/index.ts
```diff
@@ -89,38 +90,41 @@ function createMediaQueryList( args?: { min?: number; max?: number } ): QueryIte
 	return false;
 }
 
-const mediaQueryLists: Record< string, QueryItem > = {
-	'<480px': createMediaQueryList( { max: 480 } ),
```
arthur791004, line 93: I think we should create the media query list on demand. Any thoughts?

packages/viewport/src/index.ts
```diff
@@ -89,38 +90,41 @@ function createMediaQueryList( args?: { min?: number; max?: number } ): QueryIte
...
+	'<800px': { max: 800 },
+	'<960px': { max: 960 },
+	'<1040px': { max: 1040 },
+	'<1280px': { max: 1280 },
+	'<1400px': { max: 1400 },
+	'>480px': { min: 480 },
+	'>660px': { min: 660 },
+	'>782px': { min: 782 },
+	'>800px': { min: 800 },
+	'>=960px': { min: 959 },
```
arthur791004, line 108: Support the new breakpoint temporarily to align the behavior of the `break-*` mixins from Gutenberg.

packages/viewport/src/index.ts
taipeicoder: What do you mean on demand?

packages/viewport/src/index.ts
```diff
@@ -89,38 +90,41 @@ function createMediaQueryList( args?: { min?: number; max?: number } ): QueryIte
...
+	if ( ! mediaQueryOptions.hasOwnProperty( breakpoint ) ) {
 		try {
 			// eslint-disable-next-line no-console
 			console.warn( 'Undefined breakpoint used in `mobile-first-breakpoint`', breakpoint );
 		} catch ( e ) {}
 		return undefined;
 	}
 
-	return mediaQueryLists[ breakpoint ];
+	return createMediaQueryList( mediaQueryOptions[ breakpoint ] );
```
taipeicoder, line 127: I think this is what you meant by creating the media query list on demand? Could you explain more why you think it should be moved here?

packages/viewport/src/index.ts
arthur791004: I mean creating them when we need to check the viewport width as they will check the viewport immediately. Hence, we got the result initially in the previous way, but it isn't required.

For example, people might open the page with the small viewport window and then increase the window width. If there is a sub-page (or a component not rendered initially) and need to use the viewport to determine something, the result might be wrong as the initial viewport is small.

taipeicoder: I see your point 🙂
It seems like a pretty significant change, so it's worth making a simple announcement in dotcom-devs to let people know about this!

taipeicoder (approved):
Works as intended 👍

packages/viewport/src/index.ts
arthur791004: Posted. See p1686282782913399-slack-C03DY3RJ5

miksansegundo (approved):
LGTM 👍  Tested using 960px as viewport width.

|BEFORE|AFTER|
|----|----|
| <video src="https://github.com/Automattic/wp-calypso/assets/1881481/4ca2dfd6-3d9b-48ec-8e33-c6dc96856c48"> | <video src="https://github.com/Automattic/wp-calypso/assets/1881481/44451c57-6a0e-49c7-aef9-6089879bb2c6"> |


@autumnfjeld This is related to the importance of tracking the window size.

okmttdhr (approved):
LGTM 🙌

PR #77962, opened 2023-06-08
Assembler - Device Switcher - Fixed device viewport 

miksansegundo:
Related to #77547

## Proposed Changes

* Added `isFixedViewport` prop to use a fixed viewport per device
  * computer `1080px`
  * tablet `782px`
  * phone `480px`
* Added `isShowFrameShadow` prop to hide the shadow
* Scaled the action bar and border on hovered patterns

https://github.com/Automattic/wp-calypso/assets/1881481/2236405f-c989-4269-b496-8cd5ec3d9eca

### Follow-up issues
- Fix the height of patterns with 100vh so when a category is toggled or the window resized they use 100% viewport height.  

## Testing Instructions

- Click on the Calypso Live link in the first comment
- Create a new site with any plan
- Continue until the Design Picker
- Scroll down and click `Start designing`
- Click `Homepage` and add patterns
- Verify that the large preview is scaled for each device type
- Verify that is scaled after resizing the window or toggling a category to hide the patterns panel
- Test patterns from the "Link in bio" category to test that they use the 100% viewport height 

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -78,10 +78,10 @@
 	&::after {
 		content: "";
 		position: absolute;
-		left: 2px;
-		right: 2px;
-		top: 2px;
-		bottom: 2px;
+		left: 1px;
+		right: 1px;
+		top: 1px;
+		bottom: 1px;
 		border-radius: 2px;
 	}
 
@@ -104,7 +104,7 @@
 
 		&::after {
 			// Scale up the border of a hovered pattern in fixed viewport
-			box-shadow: 0 0 0 calc(2px * (1 / var(--viewport-scale))) var(--color-primary-light);
+			border: calc(2px * (1 / var(--viewport-scale))) solid var(--color-primary-light);
 		}
 	}
 }
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -13,6 +13,18 @@
 
 	.pattern-assembler__pattern-panel-list--is-open & {
 		margin-inline-start: 365px;
+
+		&.device-switcher__container--is-computer .pattern-large-preview__patterns {
+			--pattern-large-preview-device-computer-scale: 0.52;
+			--pattern-large-preview-device-computer-width: 1024;
+			position: absolute;
+			left: 0;
+			top: 0;
+			transform: scale(var(--pattern-large-preview-device-computer-scale));
+			transform-origin: left top;
+			width: calc(var(--pattern-large-preview-device-computer-width) * 1px);
+			max-height: calc(100% / var(--pattern-large-preview-device-computer-scale));
+		}
 	}
 }
 
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -71,6 +71,10 @@
 		}
 	}
 
+	.block-renderer * {
+		transition: all 0.1s ease-out;
+	}
+
 	&::after {
 		content: "";
 		position: absolute;
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
@@ -13,6 +13,18 @@
 
 	.pattern-assembler__pattern-panel-list--is-open & {
 		margin-inline-start: 365px;
+
+		&.device-switcher__container--is-computer .pattern-large-preview__patterns {
+			--pattern-large-preview-device-computer-scale: 0.52;
+			--pattern-large-preview-device-computer-width: 1024;
+			position: absolute;
+			left: 0;
+			top: 0;
+			transform: scale(var(--pattern-large-preview-device-computer-scale));
+			transform-origin: left top;
+			width: calc(var(--pattern-large-preview-device-computer-width) * 1px);
+			max-height: calc(100% / var(--pattern-large-preview-device-computer-scale));
+		}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
```diff
@@ -99,7 +103,8 @@
 		}
 
 		&::after {
-			box-shadow: 0 0 0 2px var(--color-primary-light);
+			// Scale up the border of a hovered pattern in fixed viewport
+			box-shadow: 0 0 0 calc(2px * (1 / var(--viewport-scale))) var(--color-primary-light);
```
arthur791004: The scaled box-shadow might look weird. I assume we can leave it as it is. WDYT?

image

packages/components/src/device-switcher/fixed-viewport.tsx
```diff
@@ -0,0 +1,61 @@
...
+	device: string;
+	enabled?: boolean;
+}
+
+const FixedViewport = ( { children, frameRef, device, enabled }: Props ) => {
+	const viewportWidth = frameRef?.current?.parentElement?.clientWidth as number;
+	const viewportScale = useViewportScale( device, viewportWidth );
+
+	if ( ! enabled ) {
+		return <>{ children }</>;
```
arthur791004: Is the fragment required?

packages/components/src/device-switcher/device-switcher.tsx
```diff
@@ -11,29 +13,75 @@ interface Props {
...
 	children,
 	className = '',
 	defaultDevice = DEVICE_TYPES.COMPUTER,
 	isShowDeviceSwitcherToolbar,
 	isShowFrameBorder,
+	isShowFrameShadow = true,
+	isFixedViewport,
 	frameRef,
 	onDeviceChange,
+	onAnimationEnd,
```
arthur791004: This sounds weird. How about `onViewportChange`?

arthur791004:
Need to rebase onto trunk 😉

packages/components/src/device-switcher/fixed-viewport.tsx
miksansegundo: The fragment is required to avoid this error, any other ideas?

Screenshot 2566-06-12 at 15 36 10

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
miksansegundo: It looks like the scale is not working well. It shouldn't be so thick. Should we use a border instead?

packages/components/src/device-switcher/fixed-viewport.tsx
miksansegundo: I did a change so we don't need it. WDYT?

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/pattern-large-preview.scss
miksansegundo: I cannot reproduce the issue, have you found steps to reproduce it? The issue seems to be the scale not being updated.

I think it's fine to use shadow rather than border because it works fine when the scale is correct. 

Screenshot 2566-06-12 at 16 03 50

Screenshot 2566-06-12 at 16 01 45

Screenshot 2566-06-12 at 16 03 57

miksansegundo: I just reproduced it by making the window smaller, so small that seems to be an edge case but I think I can fix it.

Screenshot 2566-06-12 at 16 16 41

miksansegundo: Using a border fixes it! Is there any reason to use a shadow?

https://github.com/Automattic/wp-calypso/assets/1881481/14a6a51e-f4ef-48a1-923f-928bb72e59c6

arthur791004:
The text on the large preview looks a bit small. Maybe we don't need to fix the viewport if there is no selected pattern.

image

arthur791004:
The result looks good. However, people can enter the assembler screen with a width `960px` and above, but it's a bit small 🥲

image

We might avoid preselecting the first category, and open the pattern list on the top of the category list on the smaller screen. Just like the editor. WDYT?

miksansegundo:
> The text on the large preview looks a bit small. Maybe we don't need to fix the viewport if there is no selected pattern.

Done ✅ Good idea! 

> people can enter the assembler screen with a width 960px and above, but it's a bit small 🥲

Let's follow up with a responsive version for the 960px breakpoint and consider others like mobile later. cc: @lucasmendes-design  🎨

> We might avoid preselecting the first category, and open the pattern list on the top of the category list on the smaller screen.

+1

arthur791004 (approved):
Tested and LGTM 👍 

Maybe we have to disable the fixed viewport under a certain breakpoint (e.g.: 1280px) temporarily as I think the experience of the large preview is not good when the pattern list panel is opening. But it's just my 2 cents

miksansegundo:
Thanks @arthur791004, let's keep iterating! 🏎️ 🙂

@lucasmendes-design Here some more ideas for the responsive version in small screens:

- reduce paddings in layout
- don't open the first category initially 
- close the patterns panel after click 
- don't show the desktop preview and default to tablet

I can demo these scenarios and others in a walkthrough call to discuss more options.

@autumnfjeld there is some design work and then a few hours for dev to make these tweaks.

lucasmendes-design:
Agreed, we can think about mobile in further explorations. 

Some thoughts and suggestions: 
- For now, I suggest not opening the patterns when users click on the homepage option for all viewports. Currently, when users select the homepage, we open by default the category All. 
- +1 `close the patterns panel after click`. We just need to save where users are in the list. Let's suppose that they scrolled down and when they open the patterns category again it is at the beginning of the list, which I think it's not a great UX. They must go to the same position. 
- For very small screens, we may have the previous behavior. It means the side menu will open over the site preview. This could be an easy thing to do and can solve the problem initially. 
image

- We can try a different approach as well and have a more compact component for the side menu. We have some space to remove from the menu. Wix example below: 
image

image

Is it possible to mockup these things @miksansegundo?

miksansegundo:
Thanks for the feedback. I'm shipping this PR and then continue working on fixes for the breakpoints 1280px and 960px.

miksansegundo:
@lucasmendes-design I'll work on the improvements in https://github.com/Automattic/wp-calypso/issues/78239. Let's continue the discussion there 🙂

PR #77984, opened 2023-06-09
Refactor check for whether a style variation is default or not

taipeicoder:
## Proposed Changes

This PR consolidates how we check whether a style variation is the default one or not in a utility function in the package `@automattic/design-picker`.

## Testing Instructions

* Head to the onboarding design picker.
* Ensure that the style variation selection in the theme card, and theme preview works as before.
* Head to the theme showcase.
* Ensure that the style variation selection in the theme card, and theme detail page works as before.

## Pre-merge Checklist

### DIFF
--- a/packages/design-picker/src/utils/global-styles.ts
+++ b/packages/design-picker/src/utils/global-styles.ts
@@ -1,0 +1,5 @@
+import { DEFAULT_GLOBAL_STYLES_VARIATION_SLUG } from '../constants';
+
+export function isDefaultGlobalStylesVariationSlug( slug?: string ): boolean {
+	return ! slug || slug === DEFAULT_GLOBAL_STYLES_VARIATION_SLUG;
+}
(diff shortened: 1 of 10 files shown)


packages/design-picker/src/utils/global-styles.ts
```diff
@@ -0,0 +1,5 @@
+import { DEFAULT_GLOBAL_STYLES_VARIATION_SLUG } from '../constants';
+
+export function isDefaultGlobalStylesVariationSlug( slug?: string ): boolean {
```
arthur791004, line 3: Maybe we can move this into `@automattic/global-styles` package as it also has the [isDefaultGlobalStyleVariationSlug](https://github.com/Automattic/wp-calypso/blob/2bb04b6f040a6a2464fd1d17942b150be1b34b71/packages/global-styles/src/components/global-styles-variations/index.tsx#L31) function

arthur791004 (approved):
LGTM 👍

packages/design-picker/src/utils/global-styles.ts
taipeicoder: I thought about it, but `@automattic/global-styles` should be loaded async. I'm not sure if importing it from there would add a lot to the entire JS bundle. WDYT @arthur791004

arthur791004: I guess the tree shaking can handle this 🤔 Let's make a minor change to test it!

PR #78011, opened 2023-06-09
Subscribers Page: some fixes based on feedback

### LINKED ISSUE #78003: Subscribers Page: fixes based on feedback
- The email address (if only that is displayed) and the ellipsis menu on the right should be vertically aligned:

Image

- When the tooltip is clicked on, there's this border displayed that shouldn't be there:

Image

- Remove `subscriptions` from the Subscriber type in `client/my-sites/subscribers/types/index.ts`.
- Use font variables in `client/my-sites/subscribers/subscriber-list/styles.scss` instead of pixel values directly.
- Use the TimeSince component in the `since` column in `client/my-sites/subscribers/subscriber-list/subscriber-row.tsx`.

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/78003

## Proposed Changes

The fixes done in this PR are:

1. Remove subscriptions from the Subscriber type in `client/my-sites/subscribers/types/index.ts`.
2. Use font variables in `client/my-sites/subscribers/subscriber-list/styles.scss` instead of pixel values directly.
3. Use the TimeSince component in the since column in `client/my-sites/subscribers/subscriber-list/subscriber-row.tsx`.

## Testing Instructions

2. Go to this URL: http://calypso.localhost:3000/subscribers/[the-domain-of-your-site].
3. Check you can see a list like the design: h8tMpJlCqNFemEerU9owHI-fi-812_37857

### DIFF
--- a/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
+++ b/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
@@ -8,7 +8,7 @@ type SubscriberListProps = {
 	subscribers: Subscriber[];
 };
 
-export const SubscriberList = ( subscribers: SubscriberListProps ) => {
+export const SubscriberList = ( { subscribers }: SubscriberListProps ) => {
 	const translate = useTranslate();
 	return (
 		<ul className="subscriber-list" role="table">
--- a/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
+++ b/client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
@@ -6,10 +6,9 @@ import './styles.scss';
 
 type SubscriberListProps = {
 	subscribers: Subscriber[];
-	locale: string;
 };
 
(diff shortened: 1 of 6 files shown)


client/my-sites/subscribers/subscriber-list/subscriber-list.tsx
```diff
@@ -37,7 +36,6 @@ export const SubscriberList = ( { subscribers, locale }: SubscriberListProps ) =
 					subscription_id,
 					display_name,
 					email_address,
-					subscriptions,
 					openRate,
```
TimBroddin, line 39: Just a heads up: in my upcoming PR I've renamed openRate to open_rate because that's probably how it's going to come in from the API later on (they use snake case 😑).

TimBroddin (commented):
I got this:

CleanShot 2023-06-09 at 14 25 21@2x

It seems like there was a typo. I'm pushing the fix.

TimBroddin (approved):
All other changes are fine!

PR #78014, opened 2023-06-09
Hide non-SSO jetpack sites from reader-share site list

Addison-Stavlo:
Related to # p1686313489610819/1686239372.646259-slack-C029GN3KD

## Proposed Changes

* It was noticed that jetpack connected self hosted sites without SSO enabled do not work with reader share. the underlying problem is more broad, as they are redirected to the jetpack site's wp-admin to login when attempting to access the editor in general. We will discuss more how to better resolve this in p2, but for now we may want to hide non-SSO enabled jetpack sites from the reader share list.

## Testing Instructions

* apply this calypso branch
* goto the reader and try to share a post
* search for a non-SSO jetpack connected site in the sites list, verify they are not present. (you may need to create and jetpack connect a self hosted site without enabling SSO).
* search for other sites you would expect to see here and verify they are present.

## Pre-merge Checklist

### DIFF
--- a/client/blocks/reader-share/index.jsx
+++ b/client/blocks/reader-share/index.jsx
@@ -152,8 +152,10 @@ class ReaderShare extends Component {
 		}
 	};
 
+	// Hide jetpack sites without SSO enabled as they do not currently work with this feature.
+	// pe7F0s-X2-p2
 	sitesFilter = ( site ) => {
-		if ( site.jetpack && ! site.is_wpcom_atomic ) {
+		if ( site?.jetpack && ! site?.options?.active_modules?.includes( 'sso' ) ) {
 			return false;
 		}
 		return true;
--- a/client/blocks/reader-share/index.jsx
+++ b/client/blocks/reader-share/index.jsx
@@ -152,6 +152,13 @@ class ReaderShare extends Component {
 		}
 	};
 
+	sitesFilter = ( site ) => {
+		if ( site.jetpack && ! site.is_wpcom_atomic ) {
+			return false;
+		}
+		return true;
+	};
+
 	render() {
 		const { onCopyLinkClick } = this.props;
 		const buttonClasses = classnames( {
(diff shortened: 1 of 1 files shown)


creativecoder (changes requested):
How about using the `getEditorUrl` selector to construct the editor url? That should should fix the issue.

Addison-Stavlo:
> How about using the getEditorUrl selector to construct the editor url? That should should fix the issue.

You mean for the `pickSiteToShareTo` function used from here?  More below on why id go another route, but in simple testing that selector seems to still give the iframed URL in this case anyways.

I was thinking if we want to fix this we try to fix the underlying issue - whatever is handling the editor redirect for non-SSO sites in general to send to the editor url that is trying to be accessed instead of landing in base wp-admin after login. That way it resolves the same issues when trying to enter the editor from calypso page/post lists and other areas than just reader sharing. 

I was thinking we could dig into this as a follow up and circle back to un-hide these, but was also uncertain on the priority given the issue is not very widespread or broadly noted by users and easily resolved with a quick toggle in the jetpack connection. 🤔

Addison-Stavlo:
Quoting myself from above:
> I was thinking if we want to fix this we try to fix the underlying issue...  I was thinking we could dig into this as a follow up

@creativecoder I was thinking roughly something like https://github.com/Automattic/wp-calypso/pull/78068 - although given editor auth is a high-traffic area and there could be a lot going on there im unsure how much that may need to snowball or raise other considerations, but the approach seems safely limited to the individual case currently. If we can get something like that merged in the short-term then id say we can drop this PR, otherwise id say we merge this as is and follow-up on the other as time fits as I don't think this is seen as a high priority issue to spend much time on (at least my understanding from previous discussions with @davemart-in)

creativecoder:
@Addison-Stavlo Let's do both!

Ideally we won't need to redirect from Gutenframe, when we know it won't work in the first place, since that's just extra time to load the editor, which already tends to take a long time to load. But agreed that we should have that fallback in place, and I've approved the PR.

In order to help avoid getting to a place where we need a redirect, in the first place, I put together https://github.com/Automattic/wp-calypso/pull/78148, which will go directly to the wp-admin url with non-SSO Jetpack sites.

PR #78018, opened 2023-06-09
Footer i18n: Update footer branding to be well supported in multiple languages

WBerredo:
Related to #77916

## Proposed Changes

Replace `Automattic` text for a variable containing Automattic SVG  in order to avoid the company name being translated

CleanShot 2023-06-12 at 12 33 55@2x

## Testing Instructions

**This branch**
* On a logged-out tab
* Go to a page that displays the footer. Ex: /theme/miniature
* Change the language (Ex: `Español`) and confirm the name Automattic is not being translated

**Prod**
* On a logged-out tab
* Go to a page that displays the footer. Ex: /theme/miniature
* Change the language (Ex: `Español`) and confirm the name Automattic  is being translated to `automático`

## Pre-merge Checklist


packages/wpcom-template-parts/src/utils.ts
```diff
@@ -1,24 +1,30 @@
...
+		translate( 'An %(Automattic)s contraption', options ),
+		translate( 'An %(Automattic)s creation', options ),
+		translate( 'An %(Automattic)s experiment', options ),
+		translate( 'An %(Automattic)s invention', options ),
+		translate( 'An %(Automattic)s joint', options ),
+		translate( 'An %(Automattic)s medley', options ),
+		translate( 'An %(Automattic)s opus', options ),
+		translate( 'An %(Automattic)s production', options ),
+		translate( 'An %(Automattic)s ruckus', options ),
+		translate( 'An %(Automattic)s thingamajig', options ),
```
jsnajdr: This is a good idea, to turn "Automattic" into a variable. Having this, you could pass the SVG directly into the `translate` function:
```js
function AutomatticBrand() {
  return <>
    <span hidden>Automattic</span>
    <svg><path d="draw the letters"/></svg>
  </>;
}

function getAutomatticBranding( translate, brand ) {
  return translate( 'An %(Automattic) thing', { args: { Automattic: brand } } );
}

getAutomatticBranding( translate, <AutomatticBrand /> );
```
With interpolated translation string, you no longer need to split manually into prefix and suffix.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8035323

Thank you @WBerredo for including a screenshot in the description! This is really helpful for our translators.

packages/wpcom-template-parts/src/utils.ts
```diff
@@ -1,24 +1,30 @@
...
 	];
 
-	const branding = automatticRoger[ Math.floor( Math.random() * ( 10 - 0 + 1 ) + 0 ) ];
+	const branding = automatticRoger[ Math.floor( Math.random() * ( 10 - 0 + 1 ) + 0 ) ]?.toString();
```
jsnajdr: I'm wondering what these `±0` arithmetics are supposed to achieve. To select a random array element, seems to me that simple `roger[ Math.floor( Math.random() * roger.length ) ]` is sufficient.

Also, once you avoid the `.split` call and use interpolation to insert the A8c logo, the `.toString` call is no longer needed. In theory `branding` can be a HTML/JSX markup then.

WBerredo: > I'm wondering what these ±0 arithmetics are supposed to achieve. To select a random array element, seems to me that simple roger[ Math.floor( Math.random() * roger.length ) ] is sufficient.

Agreed, I'm also not sure why are they needed, but I suppose it must be needed as it is a non-trivial addition.

packages/wpcom-template-parts/src/utils.ts
WBerredo: I really liked this solution and tried to implement it, but apparently, when interpolating the translation with a markup it always generates a different signature.

CleanShot 2023-06-12 at 12 59 15@2x

WBerredo: This is a great idea, ty.

I've implemented it at https://github.com/Automattic/wp-calypso/pull/78018/commits/c9c94345dac6275b4d39bf5f1991fe01018c6a41

packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx
```diff
@@ -477,7 +464,9 @@ const UniversalNavbarFooter = ( {
 	const isEnglishLocale = useIsEnglishLocale();
 	const pathNameWithoutLocale =
 		currentRoute && removeLocaleFromPathLocaleInFront( currentRoute ).slice( 1 );
-	const [ automatticBranding, setAutomatticBranding ] = useState( { article: '', noun: '' } );
+	const [ automatticBranding, setAutomatticBranding ] = useState< React.ReactChild >(
```
jsnajdr: The `React.ReactChild` type (or rather `React.JSX.Element`) will be automatically inferred from the default value, doesn't need to be specified.

packages/wpcom-template-parts/src/utils.tsx
```diff
@@ -0,0 +1,42 @@
+import { translate as translateMethod } from 'i18n-calypso';
+
+export const getAutomatticBrandingNoun = ( translate: typeof translateMethod ) => {
+	const options = {
+		description:
+			'Branding to be shown on the Footer of the page, Automattic\'s variable will always contains the word "Automattic"',
```
jsnajdr: This `options.description` field, will it be correctly extracted by the `i18n` tool? I have some doubts about that. We might have to repeat it verbatim for every variant, however ugly that is.

epeicher (approved):
I have tested it, and it works as described. Using the Calypso link and changing the language to _Español_, I cannot see the word `Automattic` being translated while it is translated in wordpress.com. 

LGTM! 🚢 
CleanShot 2023-06-13 at 13 35 53@2x

jsnajdr (approved):
Looks good now :+1:

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8035323).

PR #78067, opened 2023-06-12
Subscription Management: Allow user to resubscribe to a site.

### LINKED ISSUE #78057: Subscription Management: Let user resubscribe if user accidentally unsubscribe from a site.
On the sites page of Subscription Management, let user resubscribe if the user accidentally unsubscribe from a site.

When a person clicks on Unsubscribe, the item should remain there.
   - You can use `filter: grayscale(100%)` to turn it gray.
   - We can hide the content on "New Posts", "New Comments" & "Email Frequency" column (or keep it there is fine too).
   - Under the "Since" column, it should display "Unsubscribed".

The dropdown menu where the "Unsubscribe" button is now shows the button "Subscribe". Other settings in the dropdown menu is either disabled or hidden. It would be nice if it remembers what was the settings the user chose before.

The lifetime of this item is temporary, it goes away on page refresh.

Bear in mind that when resubscribed, the backend returns a new subscription id.

image

TimBroddin:
Fixes #78057

## Proposed Changes

This PR allows a user to resubscribe, in case they accidentally remove a subscriber.

https://github.com/Automattic/wp-calypso/assets/528287/1c993db2-17f6-47b9-a1e3-5324394f22e5

It also adds support to pass a function as the child of SettingsPopover. This function is called with a `close` function, which the child can use to close the dialog. I did this because when unsubscribing, the dialog would stay open, which seems a bit weird.

```js
	<SettingsPopover className="site-settings-popover">
			{ ( close: () => void ) => ( ... )
		</SettingsPopover>
```

The `useSiteSubscribeMutation` and `useSiteUnsubscribeMutation` now both take an extra parameter called `doNotInvalidateSiteSubscriptions` to make these mutations skip the invalidation of the queries, making this magic possible.

## Testing Instructions

1. Apply this PR
2. Surf to http://calypso.localhost:3000/subscriptions/sites
3. Unsubscribe from a site
4. This site should be grayed out in the list
5. Click the three dots again
6. Click resubscribe
7. You should be subscribed again

## Pre-merge Checklist

### DIFF
--- a/client/landing/subscriptions/components/settings/site-settings/site-settings.tsx
+++ b/client/landing/subscriptions/components/settings/site-settings/site-settings.tsx
@@ -11,6 +11,8 @@ import NotifyMeOfNewPostsToggle from './notify-me-of-new-posts-toggle';
 import './styles.scss';
 import '../styles.scss';
 
+const ResubscribeIcon = UnsubscribeIcon;
+
 type SiteSettingsProps = {
 	notifyMeOfNewPosts: boolean;
 	onNotifyMeOfNewPostsChange: ( value: boolean ) => void;
@@ -79,28 +81,55 @@ const SiteSettings = ( {
 type SiteSettingsPopoverProps = SiteSettingsProps & {
 	onUnsubscribe: () => void;
 	unsubscribing: boolean;
+	onResubscribe: () => void;
+	resubscribing: boolean;
+	isDeleted: boolean;
 };
 
 export const SiteSettingsPopover = ( {
 	onUnsubscribe,
 	unsubscribing,
+	onResubscribe,
+	resubscribing,
+	isDeleted,
 	...props
 }: SiteSettingsPopoverProps ) => {
 	const translate = useTranslate();
 	return (
 		<SettingsPopover className="site-settings-popover">
-			<SiteSettings { ...props } />
+			{ ( close: () => void ) => (
+				<>
+					<SiteSettings { ...props } />
 
-			<hr className="subscriptions__separator" />
+					<hr className="subscriptions__separator" />
 
-			<Button
-				className={ classNames( 'unsubscribe-button', { 'is-loading': unsubscribing } ) }
-				disabled={ unsubscribing }
-				icon={ <UnsubscribeIcon className="settings-popover__item-icon" /> }
-				onClick={ onUnsubscribe }
-			>
-				{ translate( 'Unsubscribe' ) }
-			</Button>
+					{ isDeleted ? (
+						<Button
+							className={ classNames( 'resubscribe-button', { 'is-loading': resubscribing } ) }
+							disabled={ resubscribing }
+							icon={ <ResubscribeIcon className="settings-popover__item-icon" /> }
+							onClick={ () => {
+								onResubscribe();
+								close();
+							} }
+						>
+							{ translate( 'Resubscribe' ) }
+						</Button>
+					) : (
+						<Button
+							className={ classNames( 'unsubscribe-button', { 'is-loading': unsubscribing } ) }
+							disabled={ unsubscribing }
+							icon={ <UnsubscribeIcon className="settings-popover__item-icon" /> }
+							onClick={ () => {
+								onUnsubscribe();
+								close();
+							} }
+						>
+							{ translate( 'Unsubscribe' ) }
+						</Button>
+					) }
+				</>
+			) }
 		</SettingsPopover>
 	);
 };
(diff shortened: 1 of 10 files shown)


client/landing/subscriptions/components/site-subscriptions-manager/site-row.tsx
```diff
@@ -87,9 +89,11 @@ export default function SiteRow( {
 		SubscriptionManager.useSiteEmailMeNewCommentsMutation();
 	const { mutate: unsubscribe, isLoading: unsubscribing } =
 		SubscriptionManager.useSiteUnsubscribeMutation();
+	const { mutate: resubscribe, isLoading: resubscribing } =
+		SubscriptionManager.useSiteSubscribeMutation();
 
 	return (
-		<li className="row" role="row">
+		<li className={ classnames( 'row', { deleted: isDeleted } ) } role="row">
```
yansern: With the move to a notification popup, could we instead, reuse the `isDeleted` flag so that it doesn't render the list item?

packages/data-stores/src/reader/mutations/use-site-subscribe-mutation.ts
```diff
@@ -62,9 +64,29 @@ const useSiteSubscribeMutation = ( blog_id?: string ) => {
...
+				queryClient.setQueryData( siteSubscriptionsCacheKey, {
+					...previousSiteSubscriptions,
+					pages: previousSiteSubscriptions.pages.map( ( page ) => {
+						return {
+							...page,
+							total_subscriptions: page.total_subscriptions - 1,
+							subscriptions: page.subscriptions.map( ( siteSubscription ) => ( {
+								...siteSubscription,
+								isDeleted:
+									siteSubscription.blog_ID === params.blog_id ? false : siteSubscription.isDeleted,
```
yansern, line 83: When resubscribing, the timestamp is incorrect (due to optimistic update). Could we optimistically update the `since` timestamp to the current time when a person resubscribes around here?

client/landing/subscriptions/components/settings/site-settings/site-settings.tsx
```diff
@@ -79,28 +81,55 @@ const SiteSettings = ( {
...
+							disabled={ unsubscribing }
+							icon={ <UnsubscribeIcon className="settings-popover__item-icon" /> }
+							onClick={ () => {
+								onUnsubscribe();
+								close();
+							} }
+						>
+							{ translate( 'Unsubscribe' ) }
+						</Button>
+					) }
```
yansern: Let's move it to a notification popup, and with that, these lines are no longer needed.

yansern:
For the snackbox / notice, if we can't make it float on the corner in this PR, it is fine. Just render it under the menu item first. :)

TimBroddin:
Ready for review again @yansern

yansern (approved):
Thanks! I've tested it and I feel that:
- There is a delay before the notice shows.
  - I wonder if we can do this optimistically.
- When the notice shows, it goes away too quickly before I could hit Resubscribe.
  - I wonder if we can just make it show until the user remove the notice itself.
- When I hit resubscribe, I wish it could scroll to the item on top (as I clicked it when I was scrolled down). It would also be nice to have some kind of indicator that shows that this item was just added back into the list.


In order to unblock other work that is relying on the resubscribe feature, I'm going to approve this PR now, and leave the rest of the UX tweaking as separate issues.

TimBroddin:
> Thanks! I've tested it and I feel that:
> 
> * There is a delay before the notice shows.
>   
>   * I wonder if we can do this optimistically.

We could, but there's a nasty edge case: what if the user clicks the button before the subscription is actually deleted? The resubscribe would fail, since already subscribed, and the subscription would be deleted anyway.  

> * When the notice shows, it goes away too quickly before I could hit Resubscribe.
>   
>   * I wonder if we can just make it show until the user remove the notice itself.

Yes, I'll add that in since it's a 3 character change.

> * When the user resubscribe, I wish it could scroll to the item on top (as I clicked it when I was scrolled down). It would also be nice to have some kind of indicator that shows that this item was just added back into the list.

I'll make an issue for this.

> 
> In order to unblock other work that is relying on the resubscribe feature, I'm going to approve this PR now, and leave the rest of the UX tweaking as separate issues.



> Thanks! I've tested it and I feel that:
> 
> * There is a delay before the notice shows.
>   
>   * I wonder if we can do this optimistically.
> * When the notice shows, it goes away too quickly before I could hit Resubscribe.
>   
>   * I wonder if we can just make it show until the user remove the notice itself.
> * When the user resubscribe, I wish it could scroll to the item on top (as I clicked it when I was scrolled down). It would also be nice to have some kind of indicator that shows that this item was just added back into the list.
> 
> In order to unblock other work that is relying on the resubscribe feature, I'm going to approve this PR now, and leave the rest of the UX tweaking as separate issues.

TimBroddin:
I'll merge this in, when changing the notice not to go away, it doesn't disappear automatically when clicking the resubscribe button. I'll make a ticket for this.

PR #34902, opened 2019-07-26
Fix for Jetpack SSO 3rd party cookie blocking issue with gutenframe

glendaviesnz:
#### Changes proposed in this Pull Request

Some browsers can prevent the cookies to be stored if they are coming from a different domain than the one we're currently visiting. This was causing issues to AT sites with a custom domain when using Gutenframe.

This PR seeks to solve that by performing a redirection to the custom domain, so the authentication cookies can be stored and then we redirect back to Calypso.

#### Testing instructions

* Make sure you have an AT test site using a custom domain.
* Make sure that the "Prevent cross-site tracking" option is enabled in the Safari privacy settings:
Screen Shot 2019-08-02 at 12 10 35

* Make sure there are no cookies saved for the custom domain of your AT test site:
Screen Shot 2019-08-02 at 12 13 18
Screen Shot 2019-08-02 at 12 13 55

* Start Calypso running on master.
* In Safari, go to `calypso.localhost:3000/block-editor/post` and select your AT test site.
  * If you're asked to enter your credentials, go first to `wordpress.com/login` and enter your credentials there, so the `wordpress.com` auth cookies can be stored and read by `calypso.localhost`.
* The editor should hang on the loading screen.
* Switch to this branch and reload `calypso.localhost:3000/block-editor/post`.
* You should notice the URL redirect to your custom domain to authenticate, and then redirect back to Calypso. This redirect should only happen once per browser session.
 
Fixes #33142.

PR #78068, opened 2023-06-12
Editor auth redirect - use native editor for redirect_to url for non-SSO jetpack sites

Addison-Stavlo:
Related to # https://github.com/Automattic/wp-calypso/pull/78014 - 78014 hides non-SSO jetpack sites from the reader-sharing site list until we can find a working solution for them. This PR attempts to resolve the underlying issue and allow those to work as expected.

## Proposed Changes

* When a non-SSO Jetpack site fails authentication when loading the editor, ensure the `redirect_to` url uses that site's native wp-admin editor. In the state previous to this PR, the `redirect_to` url was filled with the original wpcom editor address which does not work as a redirect from that site causing the user to land in the root of wp-admin after login.

NOTE - I think this is the safe approach as we only make changes for non-SSO connected Jetpack sites after auth failure. I am uncertain if these changes would be beneficial to other circumstances or not and am limiting this to avoid any other potential regressions.

## Testing Instructions

* Run this calypso build.
* Enter calypso and select a self-hosted site that is connected via Jetpack without SSO (wpcom login) enabled.
* Go to posts and "add new".
* After being prompted to login on your self-hosted site, verify you are loaded into the editor. (previously this landed in root wp-admin).
* Go back to the calypso and visit the reader.
* Select a post to share in the reader and select the same site noted above as the site to share from.
* After logging into the self-hosted site, verify you land in the editor. An embed block should be inserted as a result of reader-sharing. (previously this was also landing users in root wp-admin).

## Pre-merge Checklist

### DIFF
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -127,8 +127,6 @@ export const authenticate = ( context, next ) => {
 		isDesktop || // The desktop app can store third-party cookies.
 		context.query.authWpAdmin; // Redirect back from the WP Admin login page to Calypso.
 
-	const isJetpackNonSSO = isJetpack && ! isSSOEnabled( state, siteId );
-
 	if ( isDesktop && isJetpack && ! isSSOEnabled( state, siteId ) ) {
 		isAuthenticated = false;
 	}
@@ -164,7 +162,7 @@ export const authenticate = ( context, next ) => {
 
 	// If non-SSO Jetpack lets ensure return URL uses the sites native editor, as the dotcom
 	// redirect does not happen.
-	if ( isJetpackNonSSO ) {
+	if ( isJetpack && ! isSSOEnabled( state, siteId ) ) {
 		const postType = determinePostType( context );
 		const postId = getPostID( context );
 
--- a/client/gutenberg/editor/controller.js
+++ b/client/gutenberg/editor/controller.js
@@ -127,6 +127,11 @@ export const authenticate = ( context, next ) => {
 		isDesktop || // The desktop app can store third-party cookies.
 		context.query.authWpAdmin; // Redirect back from the WP Admin login page to Calypso.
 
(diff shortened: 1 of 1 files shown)


Addison-Stavlo:
@glendaviesnz  👋  - I noticed you have some history in this function from a few years back so pinging you if you happen to remember and have some context or info to share. Regarding Jetpack + SSO + the editor in calypso.

I see that in https://github.com/Automattic/wp-calypso/pull/34902 the `authWpAdmin` arg was added to the `redirect_to` url when sending the user to login through wp-admin.  So it seems as if the flow for these sites would be calypso -> wp_admin login -> back to calypso editor.  Was this for when SSO was enabled or disabled?

Currently I am exploring to fix a bug when SSO is disabled. It seems that the `redirect_to` url is not respected on the self-hosted site if it is the calypso URL: after login it goes to root wp-admin instead of going back to calypso.  In testing, it seems like in this case the `redirect_to` url is only respected if it is part of the site we are logging into, and ignored otherwise.  I wasn't sure if this was expected behavior when SSO is disabled, or a regression from what you may have had working in the past?

This current PR changes the `redirect_to` url to use the self-hosted site's editor instead (only in the case of jetpack + ! SSO) which seems to fix our issues, but im curious if I could be missing more context/cases here.

glendaviesnz:
> @glendaviesnz 👋 - I noticed you have some history in this function from a few years back so pinging you if you happen to remember and have some context or info to share. Regarding Jetpack + SSO + the editor in calypso.

That PR seems like a lifetime ago 😄. Looking back on it, and from what I remember we only really looked at the SSO use case, so I don't think you are missing any context re the non-SSO use case on this PR.

creativecoder (approved):
Can confirm this fixes the Reader reblog issues on my personal site that doesn't have sso enabled.

In terms of the scope of this change, it seems okay to me, but I haven't tested any other scenarios we might need to consider.

PR #78072, opened 2023-06-12
Launchpad: Added launchpad as secondary card

nelsonec87:
## Proposed Changes

* Added Launchpad Keep building checklist to the customer home, on the secondary spot.
* This PR should not change anything until it's backend counterpart is enabled.
* It set up calypso to handle the Launchpad component as an option for the `secondary` spot in the customer home (left column).

## Testing Instructions
* Run this PR and see that nothing changed.
* Sandbox the API and apply [D113406](D113406-code) 
* Open customer home again:

Design:
image

HTML:
image

### DIFF
--- a/client/my-sites/customer-home/cards/launchpad/style.scss
+++ b/client/my-sites/customer-home/cards/launchpad/style.scss
@@ -4,6 +4,9 @@
 
 .launchpad-keep-building {
 	padding: 24px;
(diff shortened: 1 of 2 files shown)


andres-blanco (approved):
The Launchpad does appear on the homepage when the diff is applied:
image

PR #78074, opened 2023-06-12
Fixes webview integration with the Jetpack mobile app

sbarbosa:
Related to SELFSERVE-519

## Proposed Changes

This PR make sure that the webview loaded by the Jetpack mobile App doesn't contain any invalid section or links.

The proposed changes are:
* Hide the new header buttons and subtitles in mobile Jetpack
* Hide the Support button in the campaign details page
* Added a margin-top to the search component

Like in the previous version, I continue to hide the navigation bar.

## Testing Instructions

You need to test this in a browser with a specific user agent. In Chrome for example, in the device dimension dropdown, select Edit and then Add Custom.

Add this in the `User agent string` field: 
`Mozilla/5.0 (iPhone; CPU iPhone OS 12_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/16B91 wp-iphone/12.1`

* Load the live preview and navigate to Tools->Advertising
* In the URL, activate the feature flag `promote-post/redesign-i2` and refresh the page.
* Verify that the WebView version of the Posts page is displayed
**Note:** No header or navigation, and without the `View` link in the posts.

Screenshot 2023-06-12 at 16 06 48

* Exit the mobile view mode and refresh the page
* Go to the Campaigns page and then select a campaign from the list to see the detail
* Go to mobile mode again, in the URL, activate the feature flag `promote-post/redesign-i2` and refresh the page
* Verify that the Campaign details page is displaying the WebView version of the page
**Note:** No header (only the Back button), and without the `Contact Support` link in the page.

Screenshot 2023-06-12 at 16 32 41

Screenshot 2023-06-12 at 16 06 39

## Pre-merge Checklist


timur987 (commented):
Thanks, @sbarbosa!

Overall changes LGTM but I've found that the filter next to the search bar is positioned a bit higher than the search bar (it's relevant to both posts and campaigns):

Campaigns:
Screenshot 2023-06-13 at 20 24 45

Posts:
Screenshot 2023-06-13 at 20 29 19

For mobile view, "Posts" column name should be on the same level as Sort order dropdown:
Screenshot 2023-06-13 at 20 30 28

For mobile Campaigns, I can't simulate that I have campaigns (as I usually do sandboxing public-api.wordpress.com), so I can't see the list of Campaigns, but I've found that mobile view shows the rest of columns (all columns excluding the "Campaign" column, that should be hidden for mobile view):

Screenshot 2023-06-13 at 20 32 27

sbarbosa:
Thanks for the review @timur987 ! I will check what you mentioned.
One thing to notice is that you need to load the page on a movie device screen with a specific user agent. When you use that user agent, there are section of the page that are hidden (header and menu for example).

I will take a look at the thing you mentioned regarding the alignment issue, maybe some of the hidden components are getting in the way. WIll post new images after I fix this.

timur987:
> One thing to notice is that you need to load the page on a movie device screen with a specific user agent. When you use that user agent, there are section of the page that are hidden (header and menu for example).

Yep, I used Firefox with overridden User Agent string as you've mention in testing instructions. It seems better now (I mean, the "Post" and "Recently Updated" signs are inline:
Screenshot 2023-06-13 at 21 41 44

On desktop it still has improper alignment for the same elements:
Screenshot 2023-06-13 at 21 44 18

sbarbosa:
Ok @timur987, I think I got it now. I changed the approach to using padding instead of margin (it was simpler this way, the margin is already being overridden in a media query).

Now the alignment should be correct for mobile & desktop. I couldn't reproduce your last point (related to the columns being displayed). Maybe it got fixed with the merge with trunk (I think this PR did have your last PR with the responsive view for that page).

Screenshot 2023-06-13 at 16 07 43


Screenshot 2023-06-13 at 16 07 57

timur987 (approved):
Thanks, @sbarbosa!

The latest padding changes did the trick and it looks great for both

campaigns

image

and posts 

image

PR #77892, opened 2023-06-07
Pattern Assembler: Use the stylesheet from the activated result to support self-hosted site

arthur791004:
Related to #

## Proposed Changes

* The stylesheet on the self-hosted site connecting to Jetpack should have `-wpcom` suffix. Otherwise, you're not able to install the BC3. So, this PR is focusing on the following items to resolve the issue.
  * Query the theme first to fix the issue that appending `-wpcom` suffix to the theme id when installing doesn't work
  * Use the returned stylesheet or id to assemble your homepage. For example, the theme attribute on the template part
* Besides, the self-hosted site should not open the upsell modal when either Colors or Fonts is selected. Hence, check whether the current site is a self-hosted site inside the `wpcom_should_limit_global_styles` function to avoid this issue.

## Testing Instructions

* Run `cd apps/editing-toolkit && yarn dev --sync` to sync ETK plugin to your sandbox
* Go to the simple site, atomic site, or self-hosted site connecting to Jetpack (e.g. the site created via JN)
* Head to Theme Showcase
* Scroll down and select BCPA CTA
* Select header, footer, colors, and fonts
* Continue
* The upsell modal should not visible on your self-hosted site
* When you land on the site editor, you're able to see the header and footer. Before, you will see the error on the header/footer as the theme attribute on the template part is incorrect

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [x] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [x] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
[...]

PR #67882, opened 2022-09-16
Site Switcher: Use fuzzy search

zaguiini:
#### Proposed Changes

Closes https://github.com/Automattic/wp-calypso/issues/67316. As an effort to standardize the search mechanisms between the sites page and instances of the site selector, let's use the fuzzy search functionality. It's yielding good results and producing a better UX.

It's blocked by https://github.com/Automattic/wp-calypso/pull/67881, a PR that adds the `title` computed field to the site excerpt query, which is used as one of the keys for fuzzy searching: `name` is not a very good index since it might be empty. `title` is a computed value between the `name` and the `slug`. Let's include all of them when fuzzy searching, along with the site URL.

#### Testing Instructions

1. Open wordpress.com;
2. Misstype some site name or URL in the site switcher;
3. Check that the "No sites found" message is displayed;
4. Apply this branch or open the dev site exposed in this PR;
5. Type the same search param in the site switcher;
6. Check that the site now appears, despite the typo;
7. Try searching in other instances of the site selector, such as the `/marketing/tools` page.

| `trunk` | This PR |
| ------- | -------- |
| image | image |

PR #78078, opened 2023-06-12
Fix some TypeScript issues across various files

sirbrillig:
## Proposed Changes

There are a ton of TypeScript errors in trunk currently. I believe this is preventing our TS checker from finding new errors, resulting in more errors being added over time. 

This diff fixes some of the errors. I tried to only make changes that were unambiguously trivial rather than risk making changes to functionality.

## Errors I could not fix here

- A bunch of errors relating to the differing definitions of the `Campaign` type introduced by https://github.com/Automattic/wp-calypso/pull/76836.
- `createHigherOrderComponent` errors in `withSitesSorting` added by https://github.com/Automattic/wp-calypso/pull/67881.
- `createHigherOrderComponent` errors in `searchSites` added by https://github.com/Automattic/wp-calypso/pull/67882.
- A pretty serious-looking `useStepNavigation` type error on the first parameter in the stepper introduced in https://github.com/Automattic/wp-calypso/pull/77780.
- An error in the dispatched type introduced by https://github.com/Automattic/wp-calypso/pull/77892.
- An error with the properties passed to `Tooltip` added by https://github.com/Automattic/wp-calypso/pull/78043.

## Testing Instructions

I think the only way is to go over each of my changes statically and make sure they make sense.

Run `yarn typecheck` on the repo to see remaining errors. `trunk` currently has 29 errors. After this PR we're down to 9 errors.


client/my-sites/stats/hooks/use-subscribers-query.tsx
```diff
@@ -1,3 +1,4 @@
+import { isValueTruthy } from '@automattic/wpcom-checkout';
```
noahtallen, line 1: Interesting! Is this a workaround for type issues related to `filter(Boolean)`?

noahtallen (approved):
Overall looks good to me; but the e2e test failure is still happening after re-running it (edit: the failures are also happening in a different PR so probably unrelated.)

client/my-sites/earn/components/commission-fees.tsx
```diff
@@ -19,7 +19,7 @@ const CommissionFees = ( { className, commission, iconSize = 16 }: CommissionFee
 					'On your current plan, WordPress.com charges {{em}}%(commission)s{{/em}}.{{br/}} Additionally, Stripe charges are typically %(stripe)s. {{a}}Learn more{{/a}}',
 					{
 						args: {
-							commission: '' + parseFloat( commission ) * 100 + '%',
+							commission: '' + parseFloat( String( commission ) ) * 100 + '%',
```
tyxla, line 22: It's interesting that we take a number, parse it to a string, and then to a float again 🤔 Pre-existing, so no need to change anything, but it felt odd to me.

client/state/ui/selectors/get-selected-site-id.js
```diff
@@ -3,7 +3,6 @@ import 'calypso/state/ui/init';
 /**
  * Returns the currently selected site ID.
  *
- * @param  {Object}  state Global state tree
```
tyxla, line 6: I find it odd that we're removing these; we are actually returning an object here. Perhaps we should change it to return `AnyAction`?

client/state/user-settings/actions.js
```diff
@@ -50,7 +50,6 @@ export const fetchUserSettingsSuccess = ( settingValues ) => ( {
  * Post settings to WordPress.com API at /me/settings endpoint
  *
  * @param {Object} settingsOverride - default settings object
- * @returns {Object} Action object
```
tyxla, line 53: Same as above

tyxla (approved):
Nice work 👍  🚀 

Left a few comments, but nothing is mandatory / critical. 

Thanks 🙌

client/my-sites/stats/hooks/use-subscribers-query.tsx
sirbrillig: Yup. I added it ages ago for that purpose.

client/my-sites/earn/components/commission-fees.tsx
sirbrillig: Yeah, in theory if the types are right this `parseFloat` is not necessary, but the types are probably wrong since this is here 😩. Adding this double-casting should preserve the existing behavior so that if `commission` is a string, it will be properly converted back to a number.

client/state/ui/selectors/get-selected-site-id.js
sirbrillig: We could do that, but I think the inferred types are actually accurate enough for us for now.

Really, this file should be converted to TS and we can use `AppState` or `IAppState` for the `state` property, but IMO I don't think it's necessary in this PR.

In the below change, by removing the `{Object}`, the return type becomes:

```
{
    type: string;
    settingsOverride: Object;
}
```

There's still an `Object` in there from the `param` but it's more accurate and meets the `AnyAction` type. Eventually this file should also be converted to TS but I figured it was out-of-scope for this fix.

PR #62965, opened 2022-04-21
Sentry enhancements in Calypso

noahtallen:
### Changes proposed in this Pull Request
Some follow-up enhancements to Calypso

* Include the user ID when logging errors. This makes the impact of an issue more clear, as we can tell how many users were effected.
* Fix the global `window.COMMIT_SHA` value. Not sure when it got broken, but we just need to pass it in as an arg to the Docker build.
* Create releases and upload sourcemaps using the webpack plugin. (You can see sentry running in the build log here: https://teamcity.a8c.com/buildConfiguration/calypso_BuildDockerImage/7786712?showRootCauses=false&showLog=7786712_5738_220)

### The webpack plugin
The main problem to solve is that the plugin runs in dev mode, calypso.live, everything. We _don't_ want to create releases all the time. I think we should only create releases for prod. But not _just_ the environment being production, but trunk production builds in TeamCity. (This has been solved by passing data in to Docker/Webpack via TeamCity.)

The remaining problem is that enabling sourcemaps & sentry will make the build take longer. A normal build time right now can be about 5:15, but the example build with source maps was about 7:15. However, it's hard to measure exactly how much longer it is. I noticed many normal builds are above 7 minutes as well, so it's hard to say if this much of an increase would be normal.

The problem is that it would be fairly hard to solve that problem. Ideally, I would run a separate TeamCity step. But that means the sourcemaps were not generated as part of the same build which generated the files deployed to production. That could be problematic. But the only way to avoid that is to enable sourcemaps for prod builds... which defeats the purpose of generating them separately. 

However, the impact could be low because we only have sourcemaps on for trunk builds. So branch builds, like PRs, aren't impacted by sourcemaps being turned on.
  
### Testing instructions

* Load calypso.live for the branch
* Verify that `window.COMMIT_SHA` is the expected value. (it contains '(unknown)' on trunk)

PR #78081, opened 2023-06-13
(devex)(error-tracking) Generate JS source-maps for the notifications app

fullofcaffeine:
## Proposed Changes

Generate source-maps for the `notifications` _standalone_ app JS assets. When `NODE_ENV` is _not_ `production`, it will use `inline-cheap-source-map` as `devtool`. For production builds, `source-map` will be used, following the same pattern of other Calypso apps (such as ETK).

### Production build

1. Run `yarn build`
2. Verify `dist/build.min.js.map` is present (see screenshot below):
Screenshot from 2023-06-12 17-57-22

#### Verify that source-maps are applied in production by sandboxing `widgets.wp.com`:

1. Sandbox `widgets.wp.com`;
2. Run `yarn build`
3. `cd dist` and then run `rsync -azP  --exclude .git --exclude --delete user@sandbox:~/public_html/widgets.wp.com/notifications`
4. Open the notifications panel by clicking the bell icon on the upper-right corner of the screen
5. Verify source-maps have been applied:

Screenshot from 2023-06-12 18-27-10

This tests that source-maps have been applied to the standalone app.

#### Verify source-maps are applied in production in the `wp-calypso` web app:

Source-maps are already being generated for the Calypso web app. The notification is rendered within the Calypso client, which already has source-map generation setup (see https://github.com/Automattic/wp-calypso/pull/62965). It uses `hidden-source-maps`, so they are not loaded by the devtools and we can't easily verify, but we'll trust it's already working :) '

### Development mode

1. Run `yarn dev`
7. Open `dist/build.min.js`, scroll to the bottom and notice the inline source-map:
Screenshot from 2023-06-12 17-56-27

## Pre-merge Checklist

### DIFF
--- a/apps/notifications/webpack.config.js
+++ b/apps/notifications/webpack.config.js
@@ -10,6 +10,7 @@ const HtmlWebpackPlugin = require( 'html-webpack-plugin' );
 const { BundleAnalyzerPlugin } = require( 'webpack-bundle-analyzer' );
 
 const shouldEmitStats = process.env.EMIT_STATS && process.env.EMIT_STATS !== 'false';
+const isDevelopment = process.env.NODE_ENV !== 'production';
 
 /**
  * Return a webpack config object
@@ -108,6 +109,7 @@ function getWebpackConfig(
 				} ),
 			new ExtensiveLodashReplacementPlugin(),
 		].filter( Boolean ),
+		devtool: isDevelopment ? 'inline-cheap-source-map' : 'source-map',
 	};
 }
 


apps/notifications/webpack.config.js
```diff
@@ -102,6 +103,7 @@ function getWebpackConfig(
 			new ExtensiveLodashReplacementPlugin(),
 			BuildMetaPlugin( { outputPath } ),
 		].filter( Boolean ),
+		devtool: isDevelopment ? 'inline-cheap-source-map' : 'source-map',
```
fullofcaffeine, line 112: A couple of Calypso apps use `source-map` as the `devtool` on production, but the Calypso client uses `hidden-source-map`. I think from the perspective of devex, `source-map` is better, while from a perf. perspective, one could argue `hidden-source-maps` saves some bytes. I decided to leave it as `source-map`, but happy to hear other people's thoughts about it, if any!

noahtallen:
> cd dist and then run rsync -azP  --exclude .git --exclude --delete user@sandbox:~/public_html/widgets.wp.com/notifications

By the way, there are two built-in ways to do this:

1. `yarn build --sync`
2. `install-plugin.sh notifications generate/js-source-maps-for-notifications-app` (this would more closely match the production release asset!)

I used the install-plugin approach, but I'm noticing the map file isn't getting applied:

```sh
~/public_html (trunk) $ install-plugin.sh notifications generate/js-source-maps-for-notifications-app
Downloading...
Unzipping...
Checking for unsafe deploy...
Updating files with version control...

Build info:
  - TeamCity build number: 1.65272
  - Commit hash: 171fb9e
  - Commit link: https://github.com/Automattic/wp-calypso/commit/171fb9e0181e64542fb0cd7811458259e1b3f60c
  - Direct download: https://teamcity.a8c.com/repository/download/calypso_WPComPlugins_Notifications/notifications-release-build.tcbuildtag/notifications.zip?guest=1&branch=generate/js-source-maps-for-notifications-app

Successfully installed Notifications build 1.65272
~/public_html (trunk) $ git status
On branch trunk
Your branch is up to date with 'origin/trunk'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   widgets.wp.com/notifications/build.min.css
	modified:   widgets.wp.com/notifications/build.min.js
	modified:   widgets.wp.com/notifications/build.min.js.LICENSE.txt
	modified:   widgets.wp.com/notifications/build.min.rtl.css
	modified:   widgets.wp.com/notifications/build_meta.json
	modified:   widgets.wp.com/notifications/index.html
	modified:   widgets.wp.com/notifications/rtl.html
	modified:   wp-content/lib/calypso-apps/cache-busters.json
```

Even though the `.min.js.map` file is in the buildfile. It's possible we also need to update the install-script to allow .map files?

fullofcaffeine:
> Even though the .min.js.map file is in the buildfile. It's possible we also need to update the install-script to allow .map files?

~~Interesting! Thanks for letting me know, I'll look into that!~~ [See comment below](https://github.com/Automattic/wp-calypso/pull/78081#issuecomment-1588363218).

fullofcaffeine:
> yarn build --sync

Ahh, nice. I knew about `dev --sync` but forgot about `build --sync`, thanks for the nudge :)

fullofcaffeine:
> Interesting! Thanks for letting me know, I'll look into that!

Ah, I think I know what's happening. We do remove .map files from production and upload them to Sentry using [this build](https://teamcity.a8c.com/buildConfiguration/calypso_WPComPlugins_GutenbergUploadSourceMapsToSentry/10191998?buildTab=log), so that's expected. That means it might make sense to change the `devtool` to `hidden-source-map` OR to adjust the `install-plugin.sh` script to allow pushing source-maps. However - if my memory doesn't fail - we remove source-maps in the `install-plugin.sh` to reduce the size of the commit and reduce the risk of causing a `hunt suc dracones` situation (the svn commit size limit error message).

noahtallen:
Ah, forgot about that!

> However - if my memory doesn't fail - we remove source-maps in the install-plugin.sh to reduce the size of the commit

I don't think the size issue should be a problem for notifications (ETK could be a different story). So if we need to push it to work around the issue, that'd be fine. However, if we want to match the ETK+Gutenberg behavior and add notifications source maps to that build too, hidden source map sounds good

fullofcaffeine:
For uploading soruce-maps to Sentry for the `notifications` app, it _should_ work as is. The TC build for `notifications` should contain a complete build with all assets, including source-maps, no? 

I'd love to commit source-maps to the SVN repo (so that users and devs could see source-mapped JS assets in the devtools), but given the SVN size limit, I'll probably refrain from doing so atm.

fullofcaffeine:
> For uploading soruce-maps to Sentry for notifications, it should work as is. The TC build for notifications should contain a complete build with all assets, including source-maps, no?

See:
* https://github.com/Automattic/wp-calypso/blob/trunk/.teamcity/_self/lib/wpcom/WPComPluginBuild.kt#L114
* https://github.com/Automattic/wp-calypso/blob/trunk/apps/notifications/package.json#L24
* https://github.com/Automattic/wp-calypso/blob/trunk/.teamcity/_self/projects/WPComPlugins.kt#LL409C2-L409C2

noahtallen:
Yep! The artifact does have the map file:

image

noahtallen (approved):
What you mentioned sounds good to me 👍

apps/notifications/webpack.config.js
noahtallen: Another point is that you get warnings in the browser console when a file includes a link to a source map, but one isn't served from the same location. If it doesn't make a difference to Sentry, we could just use hidden source map for all the Calypso apps

fullofcaffeine:
There's a related discussion about capturing exceptions generated by the standalone `notifications` app here: p1686680165242869-slack-C7YPUHBB2. @noahtallen brought up a good point about errors generated from `notifications` not having access to any error-tracking handling because it runs from an iframe. This shouldn't block the merge of this PR, but we'll need at least two other PRs for error-tracking to start working on the `notifications` standalone app:
1. A PR that exposes error-tracking to the WPCOM simple frontend
1. A PR that adds logic in the notifications standalone app to capture errors and forward them through the iframe proxy to the parent frame, so that the Sentry instance in the WPAdmin context (exposed by the ETK) can pick it up. It will be recorded in the `gutenberg-wp-admin` Sentry project. cc @noahtallen 
1. The backend PHP for the error-tracking should only activate Sentry for logged-in users. There's no such guard right now, but since we use the `admin_` actions, Sentry is never really loaded in the frontend.

PR #78085, opened 2023-06-13
Signup: Add Domain Origin Prop to Tracks Event

sharpirate:
Fixes 71-gh-Automattic/growth-foundations

### Proposed Changes
Add `signup_domain_origin` prop to `calypso_signup_complete` tracks event to differentiate between all the places in the UI from which a user can select a domain.

### Testing
1. Checkout this branch
2. Run `yarn` and `yarn start`
3. Go to `http://calypso.localhost:3000/start`
4. Open [Tracks Vigilante](https://github.com/Automattic/tracks-chrome-extension) so you can monitor tracks events
5. Click on **Choose my domain later** in the sidebar. Continue through the flow until you see the `calypso_signup_complete` event fire in tracks vigilante. Confirm the even has the following property:
`signup_domain_origin: 'choose later'`
6. Click on **Use a domain I own** in the sidebar. Continue through the flow until you see the `calypso_signup_complete` event fire in tracks vigilante. Confirm the even has the following property:
`signup_domain_origin: 'use your domain'`
7. Select the free .wordpress.com domain. Continue through the flow until you see the `calypso_signup_complete` event fire in tracks vigilante. Confirm the even has the following property:
`signup_domain_origin: 'free'`
8. Select the one of the custom suggested domains. Continue through the flow until you see the `calypso_signup_complete` event fire in tracks vigilante. Confirm the even has the following property:
`signup_domain_origin: 'custom'`
9. If for some reason the domain origin can't be retrieved from `localStorage`, the value of the `signup_domain_origin` should be `not set`

### DIFF
--- a/client/lib/analytics/signup.js
+++ b/client/lib/analytics/signup.js
@@ -65,6 +65,8 @@ export function recordSignupComplete(
 		);
 	}
 
+	const signUpDomainOrigin = window.localStorage.getItem( 'SIGNUP_DOMAIN_ORIGIN' );
+
 	// Tracks
 	// Note that Tracks expects blog_id to differntiate sites, hence using
 	// blog_id instead of site_id here. We keep using "siteId" otherwise since
@@ -83,7 +85,10 @@ export function recordSignupComplete(
 		starting_point: startingPoint,
 		is_transfer: isTransfer,
 		is_mapping: isMapping,
+		signup_domain_origin: signUpDomainOrigin,
 	} );
+
+	window.localStorage.removeItem( 'SIGNUP_DOMAIN_ORIGIN' );
 
 	// Google Analytics
 	const flags = [
--- a/client/lib/analytics/signup.js
+++ b/client/lib/analytics/signup.js
@@ -9,6 +9,7 @@ import { gaRecordEvent } from 'calypso/lib/analytics/ga';
 import { identifyUser } from 'calypso/lib/analytics/identify-user';
 import { addToQueue } from 'calypso/lib/analytics/queue';
 import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
+import { SIGNUP_DOMAIN_ORIGIN_KEY } from './utils/signup_domain_origin';
 
 const signupDebug = debug( 'calypso:analytics:signup' );
 
@@ -65,7 +66,7 @@ export function recordSignupComplete(
 		);
 	}
 
-	const signUpDomainOrigin = window.localStorage.getItem( 'SIGNUP_DOMAIN_ORIGIN' );
+	const signUpDomainOrigin = window.localStorage.getItem( SIGNUP_DOMAIN_ORIGIN_KEY );
 
 	// Tracks
 	// Note that Tracks expects blog_id to differntiate sites, hence using
@@ -88,7 +89,7 @@ export function recordSignupComplete(
 		signup_domain_origin: signUpDomainOrigin,
 	} );
 
-	window.localStorage.removeItem( 'SIGNUP_DOMAIN_ORIGIN' );
+	window.localStorage.removeItem( SIGNUP_DOMAIN_ORIGIN_KEY );
 
 	// Google Analytics
 	const flags = [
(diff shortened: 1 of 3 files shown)


client/lib/analytics/utils/signup_domain_origin.js
```diff
@@ -0,0 +1,8 @@
+const SIGNUP_DOMAIN_ORIGIN = {
```
sharpirate: I just came up with these strings as POC, I'm open to changing them to fit our tracks props convention

client/signup/steps/domains/index.jsx
```diff
@@ -155,6 +156,11 @@ class DomainsStep extends Component {
 	};
 
 	handleAddDomain = ( suggestion, position ) => {
+		const domainOrigin = suggestion?.is_free
+			? SIGNUP_DOMAIN_ORIGIN.free
+			: SIGNUP_DOMAIN_ORIGIN.custom;
+		window.localStorage.setItem( 'SIGNUP_DOMAIN_ORIGIN', domainOrigin );
```
gmovr: If we would like to change this key in the future, it would be easier only to have it in one place. Maybe set the key in a constant since it's used to set, read, and delete the same value. This can probably live in the same file in lib/analytics where the `SIGNUP_DOMAIN_ORIGIN` is defined. :) 

Do we need to handle the (rare?) cases where local storage is not defined, e.g. in Safari; I vaguely remember something about this, but I tested, and it works fine. 🤔 

It also appears we already check it in client/boot, and then do in-memory if localStorage is undefined e.g. here: /wp-calypso/client/boot/polyfills.js -- so probably calling it directly is fine

client/lib/analytics/signup.js
```diff
@@ -65,6 +65,8 @@ export function recordSignupComplete(
 		);
 	}
 
+	const signUpDomainOrigin = window.localStorage.getItem( 'SIGNUP_DOMAIN_ORIGIN' );
```
gmovr: If this is not set, do we want to keep it `null`, or can we fall back to something else? E.g. via null coalescing `??` 

It's more of a definition (Tracks events support blank values). It works the way it is.

gmovr (commented):
Works well. I left a few minor comments.

client/lib/analytics/utils/signup_domain_origin.js
gmovr: These seem fine to me. They're clear, and leaves no ambiguity to the domain origin. 👍  We can change them if requested later.

gmovr (approved):
LGTM

client/lib/analytics/utils/signup_domain_origin.js
```diff
@@ -0,0 +1,21 @@
...
+	return window.localStorage.getItem( SIGNUP_DOMAIN_ORIGIN_KEY ) || SIGNUP_DOMAIN_ORIGIN.not_set;
+}
+
+export function setDomainOrigin( origin ) {
+	window.localStorage.setItem( SIGNUP_DOMAIN_ORIGIN_KEY, origin );
+}
+
+export function removeDomainOrigin() {
+	window.localStorage.removeItem( SIGNUP_DOMAIN_ORIGIN_KEY );
+}
```
ddc22, line 21: This file can grow exponentially with data being accessed across various areas of the codebase via getter setters in multiple locations.

The signup framework already has the dependency store to funnel data across the steps.  The data then can be extracted as shown below and then fed into the `recordSignupComplete` function.

https://github.com/Automattic/wp-calypso/blob/6f87983e61d1dd9b5f4dd8a057ffcb41ef426b72/client/signup/main.jsx#L481-L503

ddc22 (commented):
The changes introduced in this PR are a bit problematic. We seem to be using a hacky way to access event related data from 2 unrelated places via the local storage.

There already is a dependency store in the signup framework that funnels data elements like this across the steps.

sharpirate:
Hey @jdc91, thanks for the feedback! I would be happy to refactor the code in a follow-up! Can I reach out to you if I have any questions along the way? Thanks!

ddc22:
Sure @Sharpirate please do reach out at any time 🙂

PR #78087, opened 2023-06-13
Migrate to null coalescing for safer Sentry detail check

noahtallen:
Resolves this sentry alert: https://a8c.sentry.io/issues/4241112100/?project=6313676

It shouldn't be that much different, but I guess null coalescing is slightly safer here? The error is `exceptionEvent.detail is not an Object. (evaluating '"reason" in exceptionEvent.detail').

There shouldn't be anything to check really, other than that the code is doing the same thing.

### DIFF
--- a/client/lib/sentry/index.ts
+++ b/client/lib/sentry/index.ts
@@ -160,7 +160,7 @@ export async function initSentry( { beforeSend, userId }: SentryOptions ) {
 				// Sentry does this, presumably for browsers being browsers 🤷
 				// https://github.com/getsentry/sentry-javascript/blob/4793df5f515575703d9c83ebf4159ac145478410/packages/browser/src/loader.js#L205
 				// @ts-expect-error Fixing up browser weirdness
-				exceptionEvent.reason ?? exceptionEvent?.detail?.reason ?? exceptionEvent
+				exceptionEvent?.reason ?? exceptionEvent?.detail?.reason ?? exceptionEvent
 			);
 
 		window.addEventListener( 'error', errorHandler );
--- a/client/lib/sentry/index.ts
+++ b/client/lib/sentry/index.ts
@@ -159,15 +159,10 @@ export async function initSentry( { beforeSend, userId }: SentryOptions ) {
 			void rejectionQueue.push(
 				// Sentry does this, presumably for browsers being browsers 🤷
 				// https://github.com/getsentry/sentry-javascript/blob/4793df5f515575703d9c83ebf4159ac145478410/packages/browser/src/loader.js#L205
-				// eslint-disable-next-line no-nested-ternary
-				'reason' in exceptionEvent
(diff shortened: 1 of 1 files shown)


client/lib/sentry/index.ts
```diff
@@ -159,15 +159,10 @@ export async function initSentry( { beforeSend, userId }: SentryOptions ) {
...
 				// https://github.com/getsentry/sentry-javascript/blob/4793df5f515575703d9c83ebf4159ac145478410/packages/browser/src/loader.js#L205
-				// eslint-disable-next-line no-nested-ternary
-				'reason' in exceptionEvent
-					? exceptionEvent.reason
-					: // @ts-expect-error Fixing up browser weirdness
-					'detail' in exceptionEvent && 'reason' in exceptionEvent.detail
-					? // @ts-expect-error Fixing up browser weirdness
-					  exceptionEvent.detail.reason
-					: exceptionEvent
+				// @ts-expect-error Fixing up browser weirdness
```
noahtallen, line 162: Expect error is required because .detail doesn't normally exist.

client/lib/sentry/index.ts
```diff
@@ -159,15 +159,10 @@ export async function initSentry( { beforeSend, userId }: SentryOptions ) {
...
+				exceptionEvent.reason ?? exceptionEvent?.detail?.reason ?? exceptionEvent
```
noahtallen: Fallback goes in this order:
1. event.reason
2. event.detail.reason
3. event

client/lib/sentry/index.ts
```diff
@@ -159,15 +159,10 @@ export async function initSentry( { beforeSend, userId }: SentryOptions ) {
...
-				// eslint-disable-next-line no-nested-ternary
-				'reason' in exceptionEvent
-					? exceptionEvent.reason
-					: // @ts-expect-error Fixing up browser weirdness
-					'detail' in exceptionEvent && 'reason' in exceptionEvent.detail
-					? // @ts-expect-error Fixing up browser weirdness
-					  exceptionEvent.detail.reason
-					: exceptionEvent
+				// @ts-expect-error Fixing up browser weirdness
+				exceptionEvent?.reason ?? exceptionEvent?.detail?.reason ?? exceptionEvent
```
tyxla, line 163: The only way this is safer is when `exceptionEvent` is not an object - then the `in` operator would crash with a `TypeError: Cannot use 'in' operator to search for 'reason'`. It is safer, though, and given the rest of the checks are pretty safe, I don't see any other way for this to break.

tyxla (approved):
LGTM 👍

client/lib/sentry/index.ts
noahtallen: That makes sense, then! The crash must happen when `exceptionEvent.detail` is truthy, but not an object.

PR #77528, opened 2023-05-30
Help Center: Reduce the amount of AI search requests

yoavf:
## Proposed Changes

* Increase the delay to 1000 from 500 ms. This should help reduce the unnecessary requests that we make. 
* Switch to making just one request for AI search.  Since we're not showing the interim results (urls), we can simplify the code and reduce the number of outgoing requests.

## Testing Instructions
* Open the help center
* Click on "Still need help"
* Choose one of the options
* In the form, start typing your message.
* Notice how many requests go out to the jetpack-search/ai endpoint
* Make sure the response is shown as expected

## Pre-merge Checklist

- [x] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [x] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [x] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #78102, opened 2023-06-13
Help Center: Reduce further the amount of AI search requests

yoavf:
Follow up to #77528

## Proposed Changes

* Increase the delay to 3000ms from (1000). This should help reduce the unnecessary requests that we make, as fewer non-finished messages will be sent to the server
* Clean up the use of the response in the contact form since we don't really need them there.

## Testing Instructions

* Open the help center
* Click on "Still need help"
* Choose one of the options
* In the form, start typing your message.
* Notice how many requests go out to the jetpack-search/ai endpoint
* Make sure the response is shown as expected after clicking continue.

*

## Pre-merge Checklist

### DIFF
--- a/packages/help-center/src/components/help-center-contact-form.tsx
+++ b/packages/help-center/src/components/help-center-contact-form.tsx
@@ -184,8 +184,8 @@ export const HelpCenterContactForm = () => {
 		supportSite = currentSite as HelpCenterSite;
 	}
 
-	const [ debouncedMessage ] = useDebounce( message || '', 1000 );
-	const [ debouncedSubject ] = useDebounce( subject || '', 1000 );
+	const [ debouncedMessage ] = useDebounce( message || '', 3000 );
+	const [ debouncedSubject ] = useDebounce( subject || '', 3000 );
 
 	const enableGPTResponse =
 		config.isEnabled( 'help/gpt-response' ) && ! ( params.get( 'disable-gpt' ) === 'true' );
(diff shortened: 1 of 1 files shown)


klimeryk (approved):
Tested and LGTM. The longer debounce makes sense 👍 
I've noticed that for some reason the `useJetpackSearchAIQuery` hook does not cache the query as expected (at least by me):
 - Go to the contact form (I've used Live chat).
 - Write your message (`How do i paypal`).
 - Click Continue.
 - Wait for the suggestion to load.
 - Go back.
 - Click Continue again.
 - There's again a `search` endpoint query.
The subsequent calls are fast, because it's cached on the backend's side. But I'm not sure if it's expected we make those calls - I'd expect that `useQuery` would take care of that for us.

I've tried to quickly investigate, but couldn't figure out why it's happening. The `queryKey` looks good, don't see flags that would cause this behaviour.

It's not a blocker - this is happening on production as well. Just mentioning it in case it's causing additional calls - but for the purpose of minimizing OpenAI calls it should not matter, since these should be already cached.

yoavf:
> I've noticed that for some reason the `useJetpackSearchAIQuery` hook does not cache the query as expected (at least by me):

Right, that was intentional because there's caching on the backend (see [here](https://github.com/Automattic/wp-calypso/pull/75838#discussion_r1175304013)). I think we can still optimize it, but I need to look into the useQuery documentation to be sure.

Issue #77991, opened 2023-06-09
Theme Showcase: Two Blank Canvas when searching for the theme


arthur791004:
Referring to pdtkmj-1sJ-p2#comment-2641, there are two Blank Canvas on Theme Showcase when searching for the theme. This might be expected as https://wordpress.com/themes is pulling results from WPORG Themes Showcase. But we have to think about how to avoid this situation

arthur791004:
Closing it via https://github.com/Automattic/wp-calypso/pull/78111. We can come back to this issue if there are more and more duplicate and retired themes from the search result. See https://github.com/Automattic/wp-calypso/pull/78111#issuecomment-1590347698 for more long term solutions.

PR #78111, opened 2023-06-13
Theme Showcase: Hide the retired themes from the search result

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/77991

## Proposed Changes

* Avoid showing the retired themes from the search result if those themes are also available in wporg

| Before | After |
| - | - |
| image | image |

## Testing Instructions

* Go to Theme Showcase
* Search “blank canvas”
* Ensure you won't see 2 blank canvas theme as one of them is retired.

## Pre-merge Checklist

### DIFF
--- a/client/components/themes-list/index.jsx
+++ b/client/components/themes-list/index.jsx
@@ -17,7 +17,7 @@ import getSiteEditorUrl from 'calypso/state/selectors/get-site-editor-url';
 import isAtomicSite from 'calypso/state/selectors/is-site-automated-transfer';
 import siteHasFeature from 'calypso/state/selectors/site-has-feature';
 import { upsellCardDisplayed as upsellCardDisplayedAction } from 'calypso/state/themes/actions';
-import { DEFAULT_THEME_QUERY } from 'calypso/state/themes/constants';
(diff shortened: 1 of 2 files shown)


taipeicoder (approved):
I think this approach works in the short term. In the future, we can think of an approach where list of retired themes come from the endpoint. The reason for this is that the list of retired themes might depend on the user. And that list is defined in the endpoint (fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sjcpbz%2Qgurzrf.cuc%3Se%3Qs41838np%2315-og).

arthur791004:
Yes! In the long term, we have different choices for doing that
* Get the retired themes from the endpoint as you mentioned
* Get the wporg themes from our endpoint (it will forward to the wporg API, and format the response)
* Maybe we can introduce a new taxonomy term to retire the wpcom themes, and then we don't need to define the list on the wpcom 🤔

autumnfjeld:
FYI: @Automattic/t-rex

Copons:
Thanks for the heads-up!
We have already considered the new constant in https://github.com/Automattic/wp-calypso/pull/78174.
We have moved the matching themes logic into a separate helper, but that does not change the behaviour introduced here.

PR #78126, opened 2023-06-13
Remove `FseThemes` component

mmtr:
While investigating what's needed to show dotOrg themes in the theme showcase (https://github.com/Automattic/dotcom-forge/issues/1450) I noted that the `FseThemes` component is no longer used.

This PR removes that component to clean up a little bit the codebase.


client/my-sites/themes/themes-selection.jsx
```diff
@@ -51,7 +51,6 @@ class ThemesSelection extends Component {
 		isRequesting: PropTypes.bool,
 		isThemeActive: PropTypes.func,
 		placeholderCount: PropTypes.number,
-		customizedThemesList: PropTypes.array,
```
mmtr, line 54: This prop was only used by the `FseThemes` component, so we can remove it now.

rcrdortiz:
Great job on reducing Calypso's size by removing dead code, the code LGTM.

There are no testing instructions for this PR so I've just tested that the themes showcase continues to work as expected (display listing and search). https://container-nervous-leakey.calypso.live/themes/{yoursitehere}

If you want me to test anything else besides that let me know, otherwise 🚢

rcrdortiz (approved):

PR #77353, opened 2023-05-24
Checkout: Update colors, spacing, and typography for new design

sirbrillig:
## Proposed Changes

This PR modifies the appearance of checkout in relatively small ways (although the visual effect is considerable) to match an updated design. This should not really change any functionality; checkout's step behavior should remain the same before and after this change.

This is part of a larger project of redesigning checkout: p9Jlb4-6FB-p2 Specifically this is the first milestone: pbOQVh-3pj-p2

### Before

Screenshot 2023-06-20 at 6 49 42 PM

jetpack-before

### After

Screenshot 2023-06-20 at 6 26 35 PM

jetpack-after

## Testing Instructions

This is a hard thing to test completely because checkout has so many variations. This should not change any functionality, so the first thing is to make sure that checkout renders successfully. Beyond that, it's a good idea to test any variation you can think of. Here are some suggestions:

- Checkout (and any of the other variations) at wide ("desktop") widths.
- Checkout (and any of the other variations) at around 900px wide ("tablet") widths.
- Checkout (and any of the other variations) at around 500px wide ("mobile") widths.
- Checkout for a Jetpack product.
- Checkout for an Akismet product.
- Checkout for a Gift product.
- Checkout with and without having a coupon added.
- Checkout with and without a domain in the cart.
- Checkout with an empty cart.
- Checkout after completing a synchronous (ie: non-3DS credit card) payment before being redirected.
- Checkout with a third-party-developer (eg: Yoast) plugin in the cart.
- Checkout in an RTL locale.
- Checkout with a monthly, yearly, and biennial WPCom plan in the cart.

Some non-checkout things that should be tested as well because they may be affected by this:

- Adding a payment method to an existing subscription.
- Viewing the mini-cart in the masterbar.
- Viewing the one-click checkout upsell after checkout completes.

PR #78128, opened 2023-06-13
Reader: only use layout overrides in `is-section-reader`

michaeldcain:
The styles added in #73464 for the logged-out `/tags` page can apply to any page if the user has already visited the Reader (and loaded the styles). This can both break intentionally full-width layouts (like the one in #77353) and result in unexpected bugs when a page implicitly relies on the max-width (or lack thereof). 

This PR makes them specific to `.is-section-reader`.

**To test:**
- logged-out, visit calypso.localhost:3000/tags and verify that the sidebar-less `.layout_content` has `max-width: 1140px`
- navigate around and verify that the styling of the pages looks correct
- logged-in, visit calypso.localhost:3000/tags and verify that the `.layout_content` has `max-width: 600px`
- navigate around and verify that the styling of the pages looks correct

### DIFF
--- a/client/reader/tags/style.scss
+++ b/client/reader/tags/style.scss
@@ -35,7 +35,7 @@
 	}
 }
 
-.has-no-sidebar .layout__content {
+.is-section-reader.has-no-sidebar .layout__content {
 	padding-left: 24px;
 	padding-right: 24px;
 	max-width: 1140px;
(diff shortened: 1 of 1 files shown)


sirbrillig (approved):
Amazing. Thank you!

PR #78137, opened 2023-06-13
Reader: Reintroduce the Discover page

eoigal:
This PR is part of the project to reintroduce the discover page to Reader

Screenshot 2023-06-16 at 16 42 40

We will keep the page private until we are complete by keeping the `isDiscoverEnabled` flag disabled.

I'm also limiting the scope of this initial PR to try to do the following;

- [x] Add Discover to side menu
- [x] Update Discover stream to use `read/tags/cards` endpoint
- [x] Add discover sidebar that lists recommended sites returned in `read/tags/cards` request

In a follow up PR we can look to add the horizontal menu above the discover post stream as per designs.
Some tasks I can think of...
* Pass recommended/related tags to a new horizontal tab menu - might be able to use the `READER_TAGS_RECEIVE` action
* Create horizontal tab component that renders tags based off action like `READER_TAGS_RECEIVE`

### Test
* Check with `isDiscoverEnabled` ([here](https://github.com/Automattic/wp-calypso/blob/de0756112aa5264e0761f47c9120d2b3789c682d/client/reader/discover/helper.js#L22)) disabled that the Discover menu isn't visible and the Following post streams (and others) work as expected.
* Enable the `isDiscoverEnabled` flag and confirm the discover menu item appear
* Confirm the Discover page loads as expected with a recommended post stream on the left and a popular sites sidebar on the right.

### DIFF
--- a/client/data/reader/use-related-meta-by-tag.ts
+++ b/client/data/reader/use-related-meta-by-tag.ts
@@ -104,6 +104,8 @@ const selectFromCards = ( response: {
 	related_tags: selectRelatedTags( response ),
 } );
 
+//TODO: this is the endpoint I need to use to get the related sites and tags for a given tag
+
 export const useRelatedMetaByTag = ( tag: string ): UseQueryResult< RelatedMetaByTag | null > => {
 	const tag_recs_per_card = 10;
 	const site_recs_per_card = 5;
--- a/client/data/reader/use-related-meta-by-tag.ts
+++ b/client/data/reader/use-related-meta-by-tag.ts
@@ -104,8 +104,6 @@ const selectFromCards = ( response: {
 	related_tags: selectRelatedTags( response ),
 } );
 
-//TODO: this is the endpoint I need to use to get the related sites and tags for a given tag
-
 export const useRelatedMetaByTag = ( tag: string ): UseQueryResult< RelatedMetaByTag | null > => {
 	const tag_recs_per_card = 10;
 	const site_recs_per_card = 5;
--- a/client/reader/stream/index.jsx
+++ b/client/reader/stream/index.jsx
@@ -597,7 +597,7 @@ export default connect(
 		const stream = getStream( state, streamKey );
 		const selectedPost = getPostByKey( state, stream.selected );
 		const streamKeySuffix = streamKey?.substring( streamKey?.indexOf( ':' ) + 1 );
-		const recommendedSites = getReaderRecommendedSites( state, 'seed?' ) || [];
+		const recommendedSites = getReaderRecommendedSites( state, 'discover-recommendations' ) || [];
 
 		return {
 			blockedSites: getBlockedSites( state ),
--- a/client/reader/stream/index.jsx
+++ b/client/reader/stream/index.jsx
@@ -41,6 +41,7 @@ import {
 	getTransformedStreamItems,
 	shouldRequestRecs,
 } from 'calypso/state/reader/streams/selectors';
+import { getReaderTags } from 'calypso/state/reader/tags/selectors';
 import { viewStream } from 'calypso/state/reader-ui/actions';
 import { resetCardExpansions } from 'calypso/state/reader-ui/card-expansions/actions';
 import getPrimarySiteId from 'calypso/state/selectors/get-primary-site-id';
@@ -50,7 +51,6 @@ import PostLifecycle from './post-lifecycle';
 import PostPlaceholder from './post-placeholder';
 import ReaderListFollowedSites from './reader-list-followed-sites';
 import './style.scss';
-import { getReaderTags } from 'calypso/state/reader/tags/selectors';
 
 const WIDE_DISPLAY_CUTOFF = 900;
 const GUESSED_POST_HEIGHT = 600;
@@ -462,7 +462,8 @@ class ReaderStream extends Component {
 	};
 
 	render() {
-		const { translate, forcePlaceholders, lastPage, streamHeader, streamKey, tag, tags } = this.props;
+		const { translate, forcePlaceholders, lastPage, streamHeader, streamKey, tag, tags } =
+			this.props;
 		const wideDisplay = this.props.width > WIDE_DISPLAY_CUTOFF;
 		let { items, isRequesting } = this.props;
 		const hasNoPosts = items.length === 0 && ! isRequesting;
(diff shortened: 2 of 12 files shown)


client/state/data-layer/wpcom/read/streams/index.js
```diff
@@ -100,6 +101,16 @@ const streamApis = {
 		path: ( { streamKey } ) => `/read/feed/${ streamKeySuffix( streamKey ) }/posts`,
 		dateProperty: 'date',
 	},
+	discover: {
+		path: () => '/read/tags/cards',
+		dateProperty: 'date',
+		query: ( extras, { tags } ) =>
+			getQueryString( {
+				...extras,
+				tags: tags ? Object.values( tags )?.map( ( tag ) => tag.slug ) : [],
```
Addison-Stavlo, line 207: Noting a couple things for us to circle back to later.

Here, if the request to retrieve tags has not completed we end up with an empty feed on discover that never updates. You can test this by reloading on 'Follow' and instantly hitting the 'Discover' tab before the followed tags have loaded, a list will never resolve on the discover tab.  With something like useQuery we would make the hook dependent on tags which would be one way to resolve it, there is probably another way to resolve this in this structure but Im not exactly sure how atm.. so pointing this out for later.

I imagine in the future we might will fill in some default tags to the empty array to ensure there is some discover feed even hen the user has no tags followed? If we do that this bug will evolve to showing the discover feed for those base tags in that circumstance, and never updating to the appropriate feed for the followed tags once those are retrieved.

client/state/data-layer/wpcom/read/streams/index.js
```diff
@@ -301,6 +358,9 @@ export function handlePage( action, data ) {
 	} else {
 		actions.push( receivePosts( streamPosts ) );
 		actions.push( receivePage( { streamKey, query, streamItems, pageHandle, gap } ) );
+		if ( streamSites.length > 0 ) {
+			actions.push( receiveRecommendedSites( { seed: 'seed?', sites: streamSites } ) );
```
Addison-Stavlo: I noticed `recieveRecommendedSites` was filtering out some of the results of `streamSites` when i tried it earlier (ones that i wouldn't initially expect to be filtered out here). This could be something to circle back to and investigate to see if we need to create a new action/selector combo for something like `recieveDiscoverSites` or not.

Addison-Stavlo (commented):
Just adding a couple notes in to remind us to circle back to later. Great progress here so far!

client/state/data-layer/wpcom/read/streams/index.js
eoigal: This was due to using `feed_ID` instead of `feedId` in the site object mapping.

client/state/data-layer/wpcom/read/streams/index.js
eoigal: We have a list of default tags that we could possible use in cases where no tags are found. This would be a good/simple workaround until we improve the UX around onboarding users to tags.

client/data/reader/use-related-meta-by-tag.ts
```diff
@@ -104,6 +104,8 @@ const selectFromCards = ( response: {
 	related_tags: selectRelatedTags( response ),
 } );
 
+//TODO: this is the endpoint I need to use to get the related sites and tags for a given tag
+
```
Addison-Stavlo: Do we still need this TODO for anything?  I think its already being used by the tags sidebar for related tags/sites.

client/reader/stream/index.jsx
```diff
@@ -588,6 +597,7 @@ export default connect(
 		const stream = getStream( state, streamKey );
 		const selectedPost = getPostByKey( state, stream.selected );
 		const streamKeySuffix = streamKey?.substring( streamKey?.indexOf( ':' ) + 1 );
+		const recommendedSites = getReaderRecommendedSites( state, 'seed?' ) || [];
```
Addison-Stavlo: Probably good to change `seed?` to `discover-recommendations` or something? (and in the recieveRecommendedSites)

client/state/data-layer/wpcom/read/streams/index.js
Addison-Stavlo: That will be great to prevent the blank feed. I think we will still need to circle back to ensure the feed updates when the users followed tags are received though. We don't have to do that here.

client/state/data-layer/wpcom/read/streams/index.js
```diff
@@ -243,16 +257,33 @@ export function handlePage( action, data ) {
 		// and offsets must be used
 		const { after } = date_range;
 		pageHandle = { before: after };
+	} else if ( next_page_handle ) {
+		pageHandle = { page_handle: next_page_handle };
+	}
```
Addison-Stavlo: What did this fix for us? Looking at this, it seems like `next_page` handles things the same way above.. i wonder why there was a separate `next_page_handle` needed by whatever might be using it?

client/state/data-layer/wpcom/read/streams/index.js
```diff
@@ -267,6 +298,34 @@ export function handlePage( action, data ) {
...
+		} ) );
+
+		streamSites = cardRecommendedSites.map( ( site ) => {
+			return {
+				url: site.URL,
+				site_icon: site.icon?.ico,
+				site_description: site.description,
+				site_name: site.name,
+				feed_URL: site.feed_URL,
+				feed_ID: site.feed_ID, // Not sure if this is needed yet
```
Addison-Stavlo: it sounds like this is what we needed for recieveRecommendedSites? and do we also need the `feedId` version below as well?

Addison-Stavlo (commented):
Just a few Qs as pointed out above.

Other than that, just a couple console logs to remove and I think were off to a good start.

client/data/reader/use-related-meta-by-tag.ts
eoigal: Oops no, this can be reverted. It was only a reminder.

client/state/data-layer/wpcom/read/streams/index.js
eoigal: As I understand it, we need to this for the Stream paging requests to request next batch of posts. Without setting the `pageHandle` the paging will not load next batch of posts.

PR #78132, opened 2023-06-13
TESTING: Cause TypeScript error to see if the CI catches it

sirbrillig:
This PR is an experiment; do not merge it!

Here we cause a TypeScript error to see if the automated CI checks report it.

PR #78141, opened 2023-06-13
Add proper rejection handling for allsettled

noahtallen:
Unit tests on trunk are not failing as expected. I think the issue is due to using allSettled in the code which checks how tasks finish, so I added checks for the results of the tasks. This wasn't caught earlier because I only switched to allSettled recently in #77121 as a side task.

Testing: unit tests should probably fail in this branch.

### DIFF
--- a/bin/unit-test-suite.mjs
+++ b/bin/unit-test-suite.mjs
@@ -56,38 +56,38 @@ const testWorkspaces = {
 	testId: 'check_storybook',
 };
 
-try {
-	// Since this task is so much larger than the others, we give it a large amount
-	// of CPU and run it by itself. We let other tasks complete in parallel with
-	// less CPU since they'll still finish much more quickly.
-	const testClientTask = runTask( testClient );
+// Since this task is so much larger than the others, we give it a large amount
+// of CPU and run it by itself. We let other tasks complete in parallel with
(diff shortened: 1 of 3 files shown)


bin/unit-test-suite.mjs
```diff
@@ -5,6 +5,17 @@ import runTask from './teamcity-task-runner.mjs';
...
 const globPromise = util.promisify( glob );
 
+// Promise.allSettled with some extra code to throw an error.
+async function completeTasks( promises ) {
+	const results = await Promise.allSettled( promises );
+	const exitCodes = results
+		.filter( ( { status } ) => status === 'rejected' )
+		.map( ( { reason } ) => reason );
+	if ( exitCodes.length ) {
+		throw exitCodes[ 0 ];
```
noahtallen, line 15: Basically passes the exit code of one of the tasks up to the error handler.

sirbrillig (approved):
Hopefully we'll see the same thing here, but https://github.com/Automattic/wp-calypso/pull/78132 fails correctly with this change applied!

noahtallen:
Seems to work: 
image

tyxla:
Nice, thank you Noah 🙌

PR #78147, opened 2023-06-13
Launchpad: Set the correct checklist slug

nelsonec87:
## Proposed Changes

* Start using the right checklist in the Launchpad component (in the customer home)

## Testing Instructions

* Open the customer home and see if the right checklist is shown

image


sixhours (approved):
Looking good! :shipit:

PR #78148, opened 2023-06-13
Reader: use correct editor url for sharing posts

creativecoder:
Related to https://github.com/Automattic/wp-calypso/pull/78014

## Proposed Changes

* Update the `shouldLoadGutenframe` selector to account for Jetpack sites that don't support SSO and so can't use Gutenframe
* Determine the correct editor url for a site when sharing a post from Reader, so we don't have to take the extra time for Gutenframe to redirect.

Combined, this fixes an issue with Jetpack sites that don't support SSO, when sharing a post from Reader.

## Testing Instructions

- Set up an Atomic or Jetpack site and turn off SSO from the Jetpack settings
- Go to `/read` and click on the Share arrow below a post
- Select the non-SSO site
- Verify that you end up with a new post on the selected site that has an embed link to the post from Reader

### DIFF
--- a/client/blocks/reader-share/index.jsx
+++ b/client/blocks/reader-share/index.jsx
@@ -1,15 +1,16 @@
 import config from '@automattic/calypso-config';
 import { Button } from '@automattic/components';
+import { addQueryArgs } from '@wordpress/url';
 import classnames from 'classnames';
 import { localize, translate } from 'i18n-calypso';
 import { defer } from 'lodash';
-import page from 'page';
 import PropTypes from 'prop-types';
 import { createRef, Component } from 'react';
 import { connect } from 'react-redux';
 import PopoverMenuItem from 'calypso/components/popover-menu/item';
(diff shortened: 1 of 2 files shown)


Addison-Stavlo (commented):
Im having mixed results in testing this branch.

Testing in sharing from reader i tried it a few times from this branch:
* Test 1 - redirected to wp-admin login but the `redirect_to` was still calypso, thus after login we ended up at wp-admin root.
* Test 2 - redirected straight into the wp-admin editor as expected.
* Test 3 - same result as test 1.
* Test 4 - same result as test 2.
* test 5 - back to 1 again.

it seems pretty consistently circular  here, im wondering if it somehow has something to do with having authenticated on the self hosted site recently or something (although doesn't seem like it should for this selector)? 🤔 

note im going to ship https://github.com/Automattic/wp-calypso/pull/78068 shortly so it may become more difficult to test this way.

creativecoder:
Thanks for testing @Addison-Stavlo , I'll see if I can figure out what's going on.

PR #78150, opened 2023-06-13
Switch yarn frozen lockfile to immutable flag

noahtallen:
Yarn frozen lockfile has been deprecated -- this just tweaks everywhere we mentioned it to use `--immutable` instead, which does the same thing. (I think in the current version of yarn, frozen lockfile is just an alias for that until it's removed.)


worldomonation:
Hey sorry I didn't get around to review it, I had gone to bed by that time. Looks fine to me though, retroactively approved.

noahtallen:
Yeah, np. Figured this one wasn't worth the waiting time :)

Issue #77616, opened 2023-05-31
Launchpad : Add Tracks events


sixhours:
We need to ensure we log the following Tracks events with their event properties:

- [x] **View task list** - user was shown a specific task list. Handled in #78204

Tracks registration located in `client/my-sites/customer-home/cards/launchpad/keep-building.tsx`
`calypso_launchpad_tasklist_view`

* Task list slug/ID - `checklist_slug`
* Task list completion state - `is_complete`
* Included tasks - `tasks`
* Surrounding context (where the task list was shown) - `context`

- [x] **View task** - the task was shown to the user on the front end. Handled in #78204
* Task ID
* Task list slug/ID
* Surrounding context
* Task completion state

- [x] **Click task** - user clicked on a specific task.

Tracks registration located in `client/my-sites/customer-home/cards/launchpad/keep-building.tsx`
`calypso_launchpad_task_clicked` (This is an existing event in the Stepper)

* Task ID - `task_id`
* Task list slug/ID - `checklist_slug`
* Surrounding context - `context`
* Task completion state - `is_completed`

- [x] **Complete task** - the task was completed -- this may need to be handled by the back-end API; Handled in https://github.com/Automattic/jetpack/pull/31444
* Task ID

- [ ] **Complete task list** - a given task list was completed.
* Task list ID, 
* Number of visible tasks in the list
* Number of completed tasks in the list
* Flag to indicate whether all tasks were complete

P2 with more details: paYKcK-30X-p2

sixhours:
All the above events have been added with the exception of the task list completion event, which we decided was probably not necessary at this time. Closing this as done.

PR #78152, opened 2023-06-13
Launchpad: Add actions to tasks

### LINKED ISSUE #78146: Launchpad: Build in front-end functionality
When a user clicks on a task, they should be redirected to a screen to complete that task as per the "Link to / Action" column on the Tasks spreadsheet.

As an example, in the Stepper, this functionality is handled by `client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.tsx`

We need to implement similar front-end behavior for the `LaunchpadKeepBuilding` component.

nelsonec87:
## Proposed Changes

Fixes https://github.com/Automattic/wp-calypso/issues/78146

* This PR builds on top of https://github.com/Automattic/wp-calypso/pull/78147
* This PR adds actions to the Launchpad task on customer home

## Testing Instructions

* Open customer home
image

* Click on the items and confirm you are taken to the respective page
* You may also use react tools to check the `actionDispatch` property of the tasks.

### DIFF
--- a/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
+++ b/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
@@ -3,10 +3,18 @@ import { useLaunchpad } from '@automattic/data-stores';
 import { Launchpad, Task } from '@automattic/launchpad';
 import { useTranslate } from 'i18n-calypso';
 import { connect } from 'react-redux';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { getSiteSlug } from 'calypso/state/sites/selectors';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 
 import './style.scss';
+
+function recordTaskClickTracksEvent( is_completed: boolean, task_id: string ) {
+	recordTracksEvent( 'calypso_launchpad_task_clicked', {
+		task_id,
+		is_completed,
+	} );
+}
 
 interface LaunchpadKeepBuildingProps {
 	siteSlug: string | null;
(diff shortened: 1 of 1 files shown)


nelsonec87:
Rebased to fix conflicts after the base branch changed

client/my-sites/customer-home/cards/launchpad/keep-building.tsx
```diff
@@ -3,11 +3,19 @@ import { useLaunchpad } from '@automattic/data-stores';
...
 import { useTranslate } from 'i18n-calypso';
 import { connect } from 'react-redux';
+import { recordTracksEvent } from 'calypso/lib/analytics/tracks';
 import { getSiteSlug } from 'calypso/state/sites/selectors';
 import { getSelectedSiteId } from 'calypso/state/ui/selectors';
 
 import './style.scss';
 
+function recordTaskClickTracksEvent( is_completed: boolean, task_id: string ) {
+	recordTracksEvent( 'calypso_launchpad_task_clicked', {
```
sixhours, line 13: Yay Tracks events!

This should be `click` instead of `clicked`. The FG has more info about naming conventions for Tracks events.

We also need to include the task list name (`keep-building`) and the surrounding context (probably `customer-home`) in the event properties. The Tracks event task for reference: https://github.com/Automattic/wp-calypso/issues/77616

sixhours (approved):
Thanks for getting this out so fast! Looks great, I have a minor comment on the corresponding Tracks event that is not a blocker for shipping this. 👍

client/my-sites/customer-home/cards/launchpad/keep-building.tsx
nelsonec87: This is actually the same Tracks event from the launchpad on stepper :) 

Should we create a new one?

sixhours: Just bringing the Slack discussion here for posterity, we'll keep the same event name since it already exists and we don't want to introduce more confusion, but we should definitely add the context and task list id to the event properties.

nelsonec87: Related PR https://github.com/Automattic/wp-calypso/pull/78181

PR #78174, opened 2023-06-14
Theme showcase: Include WP.org themes

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2744
Fixes https://github.com/Automattic/dotcom-forge/issues/2745
Fixes https://github.com/Automattic/dotcom-forge/issues/2746
Fixes https://github.com/Automattic/dotcom-forge/issues/2748
Fixes https://github.com/Automattic/dotcom-forge/issues/2749

## Proposed Changes

- Includes the 100 most popular WP.org themes in the theme showcase when searching a term (gated behind the `themes/interlaced-dotorg-themes` feature flag).
- Results are interlaced with WP.com themes according to this logic:
  - If the search term has an exact match, that theme is the first result, regardless of its WP.com or WP.org origin.
  - WP.com themes take first place over WP.org themes.
  - Retired WP.org themes or duplicate WP.org themes (those that are also WP.com themes) are excluded.

**Examples**

<table><tr><td>
Screenshot 2023-06-14 at 13 17 39</td><td>
Screenshot 2023-06-14 at 13 18 00</td></tr><tr><td>
Screenshot 2023-06-14 at 13 18 14</td><td>
Screenshot 2023-06-14 at 13 18 28</td></tr></table>

## Testing Instructions

- Go to Appearance > Themes.
- Scroll down to the bottom and make sure that WP.org themes don't show up.
- Enter a search theme such that matches a WP.com theme (e.g. Disco)
- Make sure that the WP.com theme is the first result.
- Make sure that WP.com themes are next.
- Make sure that WP.org themes are included at the end.
- Enter a search theme such that matches a WP.org theme (e.g. Astra).
- Make sure that the WP.org theme is the first result.
- Make sure that WP.com themes are next.
- Make sure that WP.org themes are included at the end.
- Make sure there are no regressions for Atomic and Jetpack sites.
- Make sure there are no regressions while logged out.
- Make sure WP.org themes are not interlaced when the `themes/interlaced-dotorg-themes` feature flag is disabled (add the `?flags=-themes/interlaced-dotorg-themes` query param).

### DIFF
--- a/client/lib/wporg/index.js
+++ b/client/lib/wporg/index.js
@@ -130,6 +130,7 @@ export function fetchThemesList( options = {} ) {
 		'request[search]': search,
 		'request[page]': page,
 		'request[per_page]:': number,
+		'request[browse]': 'popular',
 	};
 
 	return getRequest( WPORG_THEMES_ENDPOINT, query );
--- a/client/my-sites/themes/helpers.js
+++ b/client/my-sites/themes/helpers.js
@@ -116,7 +116,11 @@ export function getSubjectsFromTermTable( filterToTermTable ) {
 }
 
 /**
- * Interlace WP.com themes with WP.org themes.
+ * Interlace WP.com themes with WP.org themes according to the logic below:
+ * - WP.org themes are only included if there is a search term.
+ * - If the search term has an exact match (either a WP.com or a WP.org theme), that theme is the first result.
+ * - WP.com themes are prioritized over WP.org themes.
+ * - Retired WP.org themes or duplicate WP.org themes (those that are also WP.com themes) are excluded.
  *
  * @param wpComThemes List of WP.com themes.
  * @param wpOrgThemes List of WP.org themes.
@@ -134,12 +138,15 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
 		);
 	};
 
+	const includeWpOrgThemes = !! searchTerm;
 	const wpComThemesSlugs = wpComThemes.map( ( theme ) => theme.id );
-	const validWpOrgThemes = wpOrgThemes.filter(
-		( theme ) =>
-			! wpComThemesSlugs.includes( theme?.id?.toLowerCase() ) && // Avoid duplicate themes. Some free themes are available in both wpcom and wporg.
-			! RETIRED_THEME_SLUGS_SET.has( theme?.id?.toLowerCase() ) // Avoid retired themes.
-	);
+	const validWpOrgThemes = includeWpOrgThemes
+		? wpOrgThemes.filter(
+				( theme ) =>
+					! wpComThemesSlugs.includes( theme?.id?.toLowerCase() ) && // Avoid duplicate themes. Some free themes are available in both wpcom and wporg.
+					! RETIRED_THEME_SLUGS_SET.has( theme?.id?.toLowerCase() ) // Avoid retired themes.
+		  )
+		: [];
 
 	const matchingTheme = wpComThemes.find( isMatchingTheme );
 	const restWpComThemes = matchingTheme
--- a/client/my-sites/themes/helpers.js
+++ b/client/my-sites/themes/helpers.js
@@ -125,6 +125,9 @@ export function getSubjectsFromTermTable( filterToTermTable ) {
  */
 export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPage ) {
 	const isMatchingTheme = ( theme ) => {
+		if ( ! searchTerm ) {
+			return false;
+		}
 		return (
 			theme.name?.toLowerCase() === searchTerm?.toLowerCase() ||
 			theme.id?.toLowerCase() === searchTerm?.toLowerCase()
--- a/client/my-sites/themes/helpers.js
+++ b/client/my-sites/themes/helpers.js
@@ -121,7 +121,7 @@ export function getSubjectsFromTermTable( filterToTermTable ) {
  * @param wpComThemes List of WP.com themes.
  * @param wpOrgThemes List of WP.org themes.
  * @param searchTerm Search term.
- * @param isLastPage Whether the list of WP.com has reached the last page.
+ * @param isLastPage Whether the list of WP.com themes has reached the last page.
  */
 export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPage ) {
 	const isMatchingTheme = ( theme ) => {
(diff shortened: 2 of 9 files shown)


client/my-sites/themes/themes-selection.jsx
```diff
@@ -218,20 +217,64 @@ class ThemesSelection extends Component {
...
+			: validWpOrgThemes;
+
+		return [
+			...( matchingTheme ? [ matchingTheme ] : [] ),
+			...( matchingWpOrgTheme ? [ matchingWpOrgTheme ] : [] ),
+			...restThemes,
+			// Only include WP.org themes when searching a term and after all previous themes are visible.
+			...( query.search && isLastPage ? restWpOrgThemes : [] ),
+		];
+	};
```
rcrdortiz: Consider moving the theme mix and merging logic into a separate repository or module. This would adhere better to the Single Responsibility Principle, as the theme-selection screen's primary role should be orchestrating the necessary components to display the screen, rather than implementing mixing and merging strategies.

By separating this logic, we can simplify the theme-selection screen and make our codebase easier to navigate. Plus, it opens up the opportunity for more focused unit tests, improving our test coverage of core theme logic.
We could then also add some nice unit tests and increase our coverage of core theme logic.

mmtr: Thanks for you feedback! Done in https://github.com/Automattic/wp-calypso/pull/78174/commits/09b6ad12c03030facdaabb763838cfa3a4ce1ac8 (and tests added in https://github.com/Automattic/wp-calypso/pull/78174/commits/1627d78e5f0e24d89959a663df3757f891d70d8a).

config/development.json
```diff
@@ -200,6 +200,7 @@
 		"themes/atomic-homepage-replace": true,
 		"themes/display-thank-you-page": true,
 		"themes/display-thank-you-page-for-woo": true,
+		"themes/interlaced-dotorg-themes": true,
```
Copons, line 203: How about adding the flag to `wpcalypso` too? It would be super helpful for testing!

client/my-sites/themes/search-themes-tracks.js
```diff
@@ -16,31 +16,6 @@ function useThemeSearchDetails( query ) {
...
-			( theme ) =>
-				theme?.name?.toLowerCase() === search.toLowerCase() ||
-				theme?.id?.toLowerCase() === search.toLowerCase()
-		);
-		return matchingWporgThemes.length;
-	}, [ search, themesCount, wporgThemes ] );
-
-	return { actualCount, themesCount, totalCount, wporgThemesCount };
-}
-
```
Copons, line 43: This will break the MC `themes/searches` list.

We probably don't need the `actual_count` to be a fix for the old incorrect `result_count`, but I think there's still value in tracking how many Dotcom vs Dotorg results we return on any given query.

I'm totally open to chat more about this though, as it's a change I did almost half a year ago and barely remember it. 😄 

(more info and links at paYJgx-2Pl-p2)

client/lib/wporg/index.js
```diff
@@ -130,6 +130,7 @@ export function fetchThemesList( options = {} ) {
 		'request[search]': search,
 		'request[page]': page,
 		'request[per_page]:': number,
+		'request[browse]': 'popular',
```
Copons, line 133: You mentioned earlier that we might not need this. Is it still the case?

client/my-sites/themes/helpers.js
```diff
@@ -112,3 +114,47 @@ export function getSubjectsFromTermTable( filterToTermTable ) {
 			return obj;
 		}, {} );
 }
+
+/**
+ * Interlace WP.com themes with WP.org themes.
```
Copons: This could use a longer description, as the logic is not immediately clear to me while reading the code — and I know exactly how the logic is supposed to work!

For example, it was unclear to me why Dotorg themes are only listed in searches. Turned out it was caused by the `includeWpOrgThemes` elsewhere.
Maybe just defining the parameter as optional (or, probably, giving it a default `[]` value) would have been a helpful clue. 🤔 

Either way, I think the logic here is complicated enough to warrant an explanation for our future selves.

client/lib/wporg/index.js
mmtr: It's needed if we want to emphasize quality themes (which I think we want according to the discussion in p1686746868986769/1686743304.564759-slack-C04DZ8M0GHW).

config/development.json
mmtr: https://github.com/Automattic/wp-calypso/pull/78174/commits/addbe548fbaa5a3f9a9401ea6dea308dad013ae8

client/my-sites/themes/search-themes-tracks.js
mmtr: https://github.com/Automattic/wp-calypso/pull/78174/commits/968cce9b097f95730a2c10f429e1f24c11f36c14

client/my-sites/themes/helpers.js
mmtr: https://github.com/Automattic/wp-calypso/pull/78174/commits/330128ca73a49c0b3939d3fc08ec8ea4cef153f2

client/my-sites/themes/test/helpers.js
```diff
@@ -0,0 +1,38 @@
+import { interlaceThemes } from 'calypso/my-sites/themes/helpers';
+
+describe( 'helpers', () => {
```
rcrdortiz, line 3: Thanks for refactoring the component and creating tests!

client/my-sites/themes/helpers.js
```diff
@@ -112,3 +114,54 @@ export function getSubjectsFromTermTable( filterToTermTable ) {
...
+	const restWpOrgThemes = matchingWpOrgTheme
+		? validWpOrgThemes.filter( ( theme ) => theme.id !== matchingWpOrgTheme.id )
+		: validWpOrgThemes;
+
+	return [
+		...( matchingTheme ? [ matchingTheme ] : [] ),
+		...( matchingWpOrgTheme ? [ matchingWpOrgTheme ] : [] ),
+		...restWpComThemes,
+		// Include WP.org themes after the last page of WP.com themes.
+		...( isEnabled( 'themes/interlaced-dotorg-themes' ) && isLastPage ? restWpOrgThemes : [] ),
```
rcrdortiz, line 165: Should we also make the search results for exact coincidence use this config value?

mmtr: Nope, that would break the existing behavior in production which displays WP.org results when there is an exact match.

Copons (approved):
LGTM!
Excellent work here @mmtr!

I presume https://github.com/Automattic/dotcom-forge/issues/2747 will be worked on a separate PR, so IMHO we can ship this and proceed with a UX review by Marvel + other interested parties. 🙂

PR #78179, opened 2023-06-14
Help Center: Remove Happy Chat references

klimeryk:

## Proposed Changes

* Removes references to Happy Chat from the `help-center` package.

## Testing Instructions

Happy Chat should already not be loaded/used by Help Center, so automated tests and some smoke testing should be enough.


alshakero (approved):
Tested and all works. If TS is happy, I'm happy. 

Note: if this isn't meant to break Happychat, I suggest going to `apps/happychat`, running `yarn dev --sync`, and checking that https://widgets.wp.com/calypso-happychat still works.

klimeryk:
Thanks for the review and tip on `apps/happychat`, @alshakero 🙇 I think it should be fine as-is - even if Happy Chat is loaded somewhere, there's no one available to chat. So in a way, it's already "broken". I'll merge this on Monday, along with a bigger patch I'm preparing that removes `apps/happychat`, `packages/happychat-connection`, etc. ❣️ Just wanted to split it into smaller patches when possible.

PR #78200, opened 2023-06-14
Show Subscription Type column in Subscribers Page

### LINKED ISSUE #78182: Subscribers Page: Show plans on subscription type column.
Show plans that users subscribe to on subscription type column.

If there are multiple plans, show multiple plans.

image

JuanLucha:
Closes https://github.com/Automattic/wp-calypso/issues/78182

## Proposed Changes

The changes in this PR show the Subscription Type column on the Subscribers page. Currently, it shows all the paid subscriptions of the user, or "Free tier" if there is no paid subscription.

Screenshot 2023-06-14 at 19 17 42

## Testing Instructions

You will need to perform this test with a website that has paid users.

2. Go to this URL: http://calypso.localhost:3000/subscribers/[the-domain-of-your-site]. You should see the subscription plans of your subscribers listed in the Subscription Type column.


client/my-sites/subscribers/hooks/use-add-subscriptions.ts
```diff
@@ -0,0 +1,65 @@
+import { getCurrencyObject } from '@automattic/format-currency';
+import { useTranslate } from 'i18n-calypso';
+import { useMemo } from 'react';
+import { Subscriber, SubscriptionPlan } from '../types';
+
+const useAddSubscriptions = ( rawSubscribers: Subscriber[] ): Subscriber[] => {
```
phcp: I really like the idea of using a hook to process the subscription plans. One of the cool things about React hooks is to reuse logic across different React components. But in this case, it wouldn't work on the Individual Subscription page as it depends on the list.

I suggest moving this to the row level. You could create something like `const subscriptionPlans = useSubscriptionPlans( subscriber )`, which would return the formatted subscriber plans. With that, you can still benefit from useMemo and useTranslate if needed, and it can be reused in other scenarios. Another advantage you can take, you won't have to add extra frontend properties to the Subscriber type.

TimBroddin:
The formatting is also a bit off compared to the design:

| | |
|-|-|
| image | image | 

You can maybe use `renewal_period` or `title` for this.

client/my-sites/subscribers/hooks/use-add-subscriptions.ts
JuanLucha: Great! Implementing it right away!

TimBroddin (approved):
I think Free between brackets looks a bit off. In the design Free is just rendered as is.

CleanShot 2023-06-16 at 11 54 20@2x

I'm approving to unblock 😊

JuanLucha:
> I think Free between brackets looks a bit off. In the design Free is just rendered as is.
> CleanShot 2023-06-16 at 11 54 20@2x
> 
> I'm approving to unblock 😊

Changes done 👍

PR #78204, opened 2023-06-14
Launchpad: Added task/tasklist viewed Tracks event

nelsonec87:
Related to https://github.com/Automattic/wp-calypso/issues/77616

## Proposed Changes

* Added task view Tracks event
* Added tasklist view Tracks event

## Testing Instructions
* Create a new site from /start with the "Promote myself or business" intent.
* Launch your site
* You should see the new Launchpad task list on Customer Home.
* Run on your console: `localStorage.debug='calypso:analytics'`
* Check data send on Devtools (if you can't see, make sure the "Verbose" level is enabled)
* It should be sent for the checklist and all listed tasks
 
image

### DIFF
--- a/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
+++ b/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
@@ -41,6 +41,13 @@ const LaunchpadKeepBuilding = ( { siteSlug }: LaunchpadKeepBuildingProps ): JSX.
 		const sortedTasks = [ ...completedTasks, ...incompleteTasks ];
 
 		return sortedTasks.map( ( task: Task ) => {
+			recordTracksEvent( 'calypso_launchpad_task_view', {
+				checklist_slug: checklistSlug,
+				task_id: task.id,
+				is_completed: task.completed,
+				context: 'customer-home',
+			} );
+
 			let actionDispatch;
 
 			switch ( task.id ) {
--- a/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
+++ b/client/my-sites/customer-home/cards/launchpad/keep-building.tsx
@@ -34,6 +34,15 @@ const LaunchpadKeepBuilding = ( { siteSlug }: LaunchpadKeepBuildingProps ): JSX.
 	const numberOfSteps = checklist?.length || 0;
 	const completedSteps = ( checklist?.filter( ( task: Task ) => task.completed ) || [] ).length;
 
+	recordTracksEvent( 'calypso_launchpad_tasklist_viewed', {
+		checklist_slug: checklistSlug,
+		tasks: `,${ checklist?.map( ( task: Task ) => task.id ).join( ',' ) },`,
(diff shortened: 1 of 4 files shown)


sixhours (changes requested):
Tested this, and I'm seeing the view events for the tasks with the expected data, but they are recorded twice on load, which could skew the data:

Screen Shot 2023-06-19 at 12 25 39 PM

The task list view is only recorded once with the expected data.

sixhours:
I'm not seeing the duplicate events! 🥳 But it looks like you might have removed the tracking for viewing the task list itself. 
Screen Shot 2023-06-21 at 9 02 31 AM

nelsonec87:
> I'm not seeing the duplicate events! 🥳 But it looks like you might have removed the tracking for viewing the task list itself. Screen Shot 2023-06-21 at 9 02 31 AM

It got removed after rebase :( 

It's fixed now!

client/my-sites/customer-home/cards/launchpad/keep-building.tsx
```diff
@@ -34,13 +34,29 @@ const LaunchpadKeepBuilding = ( { siteSlug }: LaunchpadKeepBuildingProps ): JSX.
 	const numberOfSteps = checklist?.length || 0;
 	const completedSteps = ( checklist?.filter( ( task: Task ) => task.completed ) || [] ).length;
 
+	recordTracksEvent( 'calypso_launchpad_tasklist_viewed', {
+		checklist_slug: checklistSlug,
+		tasks: `,${ checklist?.map( ( task: Task ) => task.id ).join( ',' ) },`,
```
sixhours, line 39: Is this comma at the beginning a typo?

Screen Shot 2023-06-21 at 11 59 31 AM

nelsonec87: No, it's to make it easier to filter later, so all ids have `,` at the beginning and at the end :)

markbiek: I also find it a little odd having the `,` at the beginning. Can you explain a little more about how it's being used?

nelsonec87: @markbiek It was a tip from Dale to make our life easier when creating the funnels. Example: if we want to see if the `site_title` task id is present, we just need to check if the `tasks` string contains `,site_title,`, and we don't have to worry about It being the first or last item on the list nor about the existence of a `my_site_title_other_id` task.

sixhours: Oh interesting! In that case, approved. :D

sixhours (approved):

client/my-sites/customer-home/cards/launchpad/keep-building.tsx
markbiek: Great trick, thanks for the details!

PR #78220, opened 2023-06-15
Use browser language for logged out users

roo2:
As part of opening up the reader to the public, we would like to be able to localize logged out pages for non-en users. project thread: pe7F0s-Ti-p2

For logged in pages, the user's localeSlug is used to set the lang in the calypso server [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/server/pages/index.js#L383)

For logged out pages, we can take the lang from the sent `Accept-Language` header. The header is split to take the first language in case multiple languages exist.

I found that the search endpoint /v1.2/read/search/ already uses the `Accept-Language` header so no change was necessary on the backend.

Note that this can affect the logged in interface language when testing on localhost because of an existing limitation with how the server detects whether the user is logged in or not. On localhost, the `wordpress_logged_in` cookie does not exist so the *setUpLogged**Out**Route* code path is invoked even if the user is logged **in**. Therefore on localhost, this change makes the users browser locale overwrite their interface language. 

Testing:
Set your browser language to a mag-16 language.
In an incognito window, open the `/read/search` page or a `/tag/$tag` page
The page should be rendered in your preferred language.
Test language selection is still based on interface language for logged in users (except on localhost)

### DIFF
--- a/client/server/pages/index.js
+++ b/client/server/pages/index.js
@@ -8,6 +8,7 @@ import {
 	isTranslatedIncompletely,
 	isDefaultLocale,
 	getLanguageSlugs,
+	getLanguage,
 } from '@automattic/i18n-utils';
 import bodyParser from 'body-parser';
 import cookieParser from 'cookie-parser';
@@ -297,7 +298,16 @@ function setUpLoggedOutRoute( req, res, next ) {
 		};
 	}
 	const acceptLang = req.get( 'Accept-Language' )?.split( ',' )[ 0 ];
-	req.context.lang = acceptLang || req.context.lang;
+
+	const language = getLanguage( acceptLang );
+
+	if ( language ) {
+		req.context.lang = language.langSlug;
+		req.context.store.dispatch( {
+			type: LOCALE_SET,
+			localeSlug: language.langSlug,
+		} );
+	}
 
 	Promise.all( setupRequests )
 		.then( () => {
--- a/client/server/pages/index.js
+++ b/client/server/pages/index.js
@@ -296,6 +296,8 @@ function setUpLoggedOutRoute( req, res, next ) {
 			subscriptionManagementSubkey: req.cookies.subkey,
 		};
 	}
+	const acceptLang = req.get( 'Accept-Language' )?.split( ',' )[ 0 ];
+	req.context.lang = acceptLang || req.context.lang;
 
 	Promise.all( setupRequests )
 		.then( () => {
--- a/client/state/data-layer/wpcom/read/streams/index.js
+++ b/client/state/data-layer/wpcom/read/streams/index.js
@@ -1,5 +1,4 @@
 import warn from '@wordpress/warning';
-import i18n from 'i18n-calypso';
 import { random, map, includes, get } from 'lodash';
 import { keyForPost } from 'calypso/reader/post-key';
 import XPostHelper from 'calypso/reader/xpost-helper';
@@ -212,14 +211,13 @@ export function requestPage( action ) {
 	const fetchCount = pageHandle ? PER_FETCH : INITIAL_FETCH;
 	// eslint-disable-next-line no-extra-boolean-cast
 	const number = !! gap ? PER_GAP : fetchCount;
-	const lang = i18n.getLocaleVariant() || i18n.getLocaleSlug();
 	return http( {
 		method: 'GET',
 		path: path( { ...action.payload } ),
 		apiVersion,
 		query: isPoll
-			? pollQuery( [], { lang, ...algorithm } )
-			: query( { ...pageHandle, ...algorithm, number, lang }, action.payload ),
+			? pollQuery( [], { ...algorithm } )
+			: query( { ...pageHandle, ...algorithm, number }, action.payload ),
 		onSuccess: action,
 		onFailure: action,
 	} );
--- a/client/state/data-layer/wpcom/read/streams/index.js
+++ b/client/state/data-layer/wpcom/read/streams/index.js
@@ -1,4 +1,5 @@
 import warn from '@wordpress/warning';
+import i18n from 'i18n-calypso';
 import { random, map, includes, get } from 'lodash';
 import { keyForPost } from 'calypso/reader/post-key';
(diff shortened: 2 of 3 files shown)


eoigal (commented):
This isn't working for me but perhaps I'm not testing properly. I'm using the calypso.live test domain to test it.

I've updated the language  in an incognito Firefox to spanish. I don't see the `lang` param being set when debugging the search endpoint. 

I do see the `Accept-language` header like so in a public-api request;

Screenshot 2023-06-15 at 09 48 01

This looks like the `acceptLang` should be set to `es-ES`... is that enough to switch languages for logged out users... I don't see how the lang property is set?

TimBroddin (approved):
I tested this and it works!

CleanShot 2023-06-15 at 11 34 43@2x

I don't think the failing E2E tests have anything to do with this PR, but best to make sure everything passes before merging.

TimBroddin:
> This isn't working for me but perhaps I'm not testing properly. I'm using the calypso.live test domain to test it.
> 
> I've updated the language in an incognito Firefox to spanish. I don't see the `lang` param being set when debugging the search endpoint.
> 
> I do see the `Accept-language` header like so in a public-api request;
> 
> Screenshot 2023-06-15 at 09 48 01
> This looks like the `acceptLang` should be set to `es-ES`... is that enough to switch languages for logged out users... I don't see how the lang property is set?

I think incognito mode might ignore the preferred language in Firefox. I tested it in Chrome, under a new profile, with the language changed here: `chrome://settings/?search=language`

client/server/pages/index.js
```diff
@@ -296,6 +296,8 @@ function setUpLoggedOutRoute( req, res, next ) {
 			subscriptionManagementSubkey: req.cookies.subkey,
 		};
 	}
+	const acceptLang = req.get( 'Accept-Language' )?.split( ',' )[ 0 ];
+	req.context.lang = acceptLang || req.context.lang;
```
noahtallen: This isn't an expert opinion (and this is a non-blocking comment, since this probably won't break anything), but I also noticed there slightly more handling done for the logged-in user here:

https://github.com/Automattic/wp-calypso/blob/2dfbefa26af382afb413a2a0ac7c7c2ab898d207/client/server/pages/index.js#L376-L389

I'm not sure if that's really needed for this, though.

Also, other pages like `/themes` do locale with a path (like https://wordpress.com/es/themes). I don't know why we did that for those pages, but I'm curious if there was a specific reason that makes the /es path useful in some cases (and if that'd also be useful for this feature!)

creativecoder: > Also, other pages like /themes do locale with a path (like https://wordpress.com/es/themes).

Taking a guess: maybe this has to do with how the themes page is server side rendered?

noahtallen: Good point, that definitely makes it easier to cache per route. (And maybe there are related SSR benefits?) Though it'd be pretty easy to cache the language based on this too. Maybe it wasn't a thing when SSR was first added

roo2: I noticed also that the /themes page isn't currently redirecting to the language version that's set in the browser. Yea hmm maybe it is useful for SEO too? I'm not sure

roo2:
I needed to process the language further, and some other things to support FF and Chrome.
* Split out the language slug from the variant e.g. for `es-ES`, the lang should be set to `es`.
* change to lower case. `es-MX` should be `es-mx`
* change underscores for hyphens. `zh_TW` should be `zh-tw`
* I had to add an exception for "Chinese" to map to "Chinese Simplified". 

I'll get on browserstack and make sure this handles headers sent by Edge as well

danielbachhuber:
@roo2 What sort of code review were you hoping for from me?

roo2:
@danielbachhuber, I'm not sure exactly, I added you as it was a suggestion because you recently touched the files I think 🙂

roo2:
I've tested edge and did more general testing for logged in and logged out users in the calypso.live environment

creativecoder (commented):
Testing with Firefox, this seems to work as described

- Change preferred webpage language in browser
- /read/search, /tags, and /tag/{slug} are translated (with noted exceptions)

---

 There some oddities, I'll mention them here in case they're relevant

- Tag slugs are not translated (do we need to update the API or API call to support that)?
- The html returned from the server is not translated. There's a flash of content as the text switches to a different language after page load. Is there some way we can support setting language server-side?
- /themes (another logged out page rendered by Calypso) is not translated. Should this support the language header, as well?

I also see that changing the browser language setting doesn't make changes to the production marketing pages (wordpress.com/pricing, for example). There must be a different way those are translated? Should these be using the same mechanism for setting the language?

roo2:
> The html returned from the server is not translated. There's a flash of content as the text switches to a different language after page load. Is there some way we can support setting language server-side?

Ahh I found this was a result of caching, this must be a reason why the lang was included in the URL in other cases, I think I will need to change the implementation

client/server/pages/index.js
```diff
@@ -296,6 +297,17 @@ function setUpLoggedOutRoute( req, res, next ) {
 			subscriptionManagementSubkey: req.cookies.subkey,
 		};
 	}
+	const acceptLang = req.get( 'Accept-Language' )?.split( ',' )[ 0 ];
```
yuliyan, line 300: I noticed that as a side-effect of this change, some other routes are now being localized based on the detected language from the `Accept-Language` header. For example, `/accept-invite` and `/pricing` from Jetpack Cloud are now localized, and some other routes like `/devdocs` get partially localized - e.g. the `lang` and `dir` attributes on the `html` tag are set based on the browser's language, but the rest of the page is rendered in English.

Is setting the locale on the server side necessary? If not, I suppose the locale can then just be set on the client side, similarly to how we do it on `/start/user` with the [LocaleSuggestions component](https://github.com/Automattic/wp-calypso/blob/trunk/client/components/locale-suggestions/index.jsx#L31).

roo2:
I think this needs a bit of a step back and try a new approach after some discussion and a p2 post pe7F0s-ZH-p2

roo2:
closing in favour of https://github.com/Automattic/wp-calypso/pull/78656 for now

Issue #78091, opened 2023-06-13
Theme Showcase: Activate option shown for retired themes


taipeicoder:
## Description

As discussed in p1686629515306039/1686174991.330239-slack-C048CUFRGFQ. From @retnonindya:

> This is a GIF screen recording. I’m searching for a retired theme, and I use maxwell as an example.
When I searched for maxwell, I got the result of the WPORG theme. When I checked the three-dots icon, I don’t see Activate.
However, when I went to the theme’s detail page, I noticed a flash of plan information (?) then the info about the theme is retired.
Then, the three dots icon gives the Activate option.

![Image](https://github.com/Automattic/wp-calypso/assets/797888/8364e0be-5587-46e1-9931-343d6d3a809a)

## Acceptance Criteria
- [ ]  The options Activate and Try & Customize should not be available for retired themes.

cuemarie:
📌 SCRUBBING : RESULT - *Replicated / Could Not Replicate / Uncertain*
- Tested on Simple – **Replicated**

📌 FINDINGS/SCREENSHOTS/VIDEO
- Replicated as described!  Tested on a free site, non-admin account, unproxied.
- Before viewing the Info page where the theme is listed as retired, the theme's sticker is `Community`, but after going back, it's tagged with `Premium`.  This means at first it prompts me to upgrade to Business, but later prompts me to upgrade to Premium.

#### Screenshots/Recordings

https://github.com/Automattic/wp-calypso/assets/27249804/76f597d3-9700-4aff-974c-24dd00fb926b

📌 ACTIONS
- Assigned to @Automattic/lego

taipeicoder:
Closing via #78231

PR #78231, opened 2023-06-15
Theme Showcase: Do not show wporg themes if they are also retired wpcom themes

taipeicoder:
Related to #78091

## Proposed Changes

The current implementation shows a wporg theme even if its wpcom counterpart is retired. For instance, the theme Maxwell exists in both wporg and wpcom, but it's retired in wpcom. This leads to a scenario where the Theme Showcase will show the wporg version of the theme, but when the user clicks on the theme, the detail page will say that the theme is retired. This behavior is confusing, and might lead to issues as described in #78091.

This PR updates the logic so that retired wpcom themes are always omitted from the Theme Showcase, even if their wporg counterparts exist. In other words, when users search for Maxwell, even though the theme exists in wporg, it will not be shown in the Theme Showcase listing because it's retired in wpcom.

## Testing Instructions

* Head to the Theme Showcase.
* Search for a retired theme that also exists in wporg, such as Maxwell or Adelle.
* Ensure that the theme is not shown in the theme listing.

## Pre-merge Checklist


arthur791004:
Tested but the Maxwell theme is `null` so the activate action is still there

image

client/state/themes/selectors/is-theme-retired.js
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Whether a WPCOM theme given by its ID is retired.
+ *
+ * @param  {Object} state   Global state tree
+ * @param  {string} themeId Theme ID
+ * @returns {boolean}        True if the theme is retired
+ */
+export function isThemeRetired( state, themeId ) {
+	return isRetired( getTheme( state, 'wpcom', themeId ) );
```
arthur791004: The theme might not be fetched. So, one option is to get theme from wporg as well and add the slug to `RETIRED_THEME_SLUGS_SET`

taipeicoder: Hmmm the `retired` property is only from wpcom, which is why we only query for the theme in wpcom. How can I reproduce the theme being `null` 🤔 ?

arthur791004: The theme was null when I went to the Theme Showcase and searched “Maxwell”. Or maybe you can head to the search result page directly 🤔

taipeicoder: @arthur791004 with D113775-code merged, the query will no longer return `null`. Could you test again? Thanks!

arthur791004: We have to pass `retired=true` to themes endpoint to get the result

taipeicoder: Oops, I forgot to commit 😅

arthur791004 (approved):
Tested again and looks good to me 👍

arthur791004:
BTW, D113775-code has a side effect on the search result...we have to hide the retired themes from wpcom on client side as well

image

taipeicoder:
> BTW, D113775-code has a side effect on the search result...we have to hide the retired themes from wpcom on client side as well
> 
> image

Was your search term "blank canvas". I could only see the one in wpcom 🤔

Screenshot 2023-06-16 at 7 00 44 PM

Were you using the calypso.live link?

arthur791004:
> Were you using the calypso.live link?

Yes, I was using the calypso.live link and searched for the “blank canvas”. 

**Here is the result on production**

image

**Here is the result from the endpoint**

image

taipeicoder:
@arthur791004 I've changed the PR proposal to an approach that I think makes more sense, and updated the description accordingly. I've moved the PR to Needs Review again 🙂

client/my-sites/themes/helpers.js
```diff
@@ -162,7 +163,12 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
 	const restWpComThemes = matchingTheme
 		? wpComThemes.filter( ( theme ) => theme.id !== matchingTheme.id )
 		: wpComThemes;
-	interlacedThemes.push( ...restWpComThemes );
+
+	// The themes endpoint returns retired themes when search term exists.
+	// See: https://github.com/Automattic/wp-calypso/pull/78231
+	interlacedThemes.push(
+		...( searchTerm ? restWpComThemes.filter( ( theme ) => ! theme.retired ) : restWpComThemes )
```
arthur791004, line 170: [nit] We can filter out the retired themes regardless of the search term. But the current way is also fine 👍

arthur791004 (approved):
Tested again and looks good 🎉

image

image

client/my-sites/themes/helpers.js
taipeicoder: I'll leave the logic in for now, since I like the idea of the small performance gain! 🙂

Issue #20475, opened 2017-12-04
Activity Log: help button in error banner should trigger HappyChat


eliorivero:
Reported by @beaulebens in https://github.com/Automattic/wp-calypso/issues/20108#issuecomment-347361355
> That button loads VaultPress documentation, but it should really just trigger a Happychat.

PR #78244, opened 2023-06-15
Remove/happychat jetpack rewind activity log

klimeryk:

## Proposed Changes

* Remove `HappychatButton` from Activity Log-related pages. This was introduced 6 years ago in https://github.com/Automattic/wp-calypso/issues/20475. It's not the best user experience to just drop them in chat as a band-aid for missing proper error messages and guidance in the UI. So let's remove it and think of ways of better supporting them in that context. Looking at the related Tracks events, those buttons were either not used at all or very rarely.

## Testing Instructions

- Go to `/activity-log/` on your Atomic and/or Jetpack site, make sure it's rendering correctly, without links to chat.

Screenshot 2023-06-14 at 14 26 18


yoavf (approved):
Great cleanup @klimeryk !

PR #71333, opened 2022-12-19
Limited Global Styles: Add E2E test

mmtr:
Part of https://github.com/Automattic/dotcom-forge/issues/1397

#### Proposed Changes

Adds a new E2E test to check that, when Global Styles are limited, the upgrade modal and the upgrade pre-save notice show up in the Site Editor.

Global Styles became a paid feature (available on Premium plans and above) for sites created after Dec 15, 2022. See p7DVsv-fY6-p2.

That means that Free and Personal sites created after that date would have Limited Global Styles. We basically allow them to try Global Styles, but any custom style is not loaded in the frontend.

As part of the Limited Global Styles flow, we show upgrade banners in the Site Editor. However, those are fragile because they rely on a CSS selector to manipulate the DOM and inject those banners, but Gutenberg can change the DOM at any point without further notice.

So with this test, we hope to catch those regressions when new Gutenberg versions are released to WPCOM so we can update the Limited Global Styles implementation with new CSS selectors.

#### Testing Instructions

- Follow setup instructions from https://github.com/Automattic/wp-calypso/tree/trunk/test/e2e
- Run the test:
  - Desktop: `yarn workspace wp-e2e-tests test -- specs/fse/fse__limited-global-styles.ts`
  - Mobile: `yarn workspace wp-e2e-tests test:mobile -- specs/fse/fse__limited-global-styles.ts`

PR #78246, opened 2023-06-15
Limited Global Styles: Skip E2E test for Atomic sites

mmtr:
## Proposed Changes

Prevents the Limited Global Styles E2E test from running on Atomic sites since they always full access to Global Styles.

See https://github.com/Automattic/wp-calypso/pull/71333#issuecomment-1592490057

## Testing Instructions

Unsure! I think this cannot be manually tested?

### DIFF
--- a/test/e2e/specs/fse/fse__limited-global-styles.ts
+++ b/test/e2e/specs/fse/fse__limited-global-styles.ts
@@ -2,8 +2,9 @@
  * @group gutenberg
  */
 
-import { TestAccount, FullSiteEditorPage } from '@automattic/calypso-e2e';
+import { envVariables, TestAccount, FullSiteEditorPage } from '@automattic/calypso-e2e';
 import { Browser, Page } from 'playwright';
+import { skipDescribeIf } from '../../jest-helpers';
 
 declare const browser: Browser;
 
@@ -12,8 +13,12 @@ declare const browser: Browser;
  * style variations (eg. Twenty Twenty-Three).
  *
  * @see https://github.com/Automattic/wp-calypso/issues/78107
+ *
+ * We skip Atomic sites because they are not affected by Limited Global Styles.
+ *
+ * @see https://github.com/Automattic/wp-calypso/pull/71333#issuecomment-1592490057
  */
-describe( 'Site Editor: Limited Global Styles', function () {
+skipDescribeIf( envVariables.TEST_ON_ATOMIC )( 'Site Editor: Limited Global Styles', function () {
 	let page: Page;
 	let fullSiteEditorPage: FullSiteEditorPage;
 	let testAccount: TestAccount;


test/e2e/specs/fse/fse__limited-global-styles.ts
```diff
@@ -12,8 +13,12 @@ declare const browser: Browser;
  * style variations (eg. Twenty Twenty-Three).
  *
  * @see https://github.com/Automattic/wp-calypso/issues/78107
+ *
+ * We skip Atomic sites because they are not affected by Limited Global Styles.
+ *
+ * @see https://github.com/Automattic/wp-calypso/pull/71333#issuecomment-1592490057
  */
-describe( 'Site Editor: Limited Global Styles', function () {
+skipDescribeIf( envVariables.TEST_ON_ATOMIC )( 'Site Editor: Limited Global Styles', function () {
```
rcrdortiz, line 21: If we add this test we can implicitly test that the code is working as intended, since otherwise the test would be failing. With this change, if the build completes we can guarantee that your changes are working.

Besides this, the code looks good and I don't think there should be any problems (since the build for this PR should have failed if there were).

```suggestion
skipDescribeIf( envVariables.TEST_ON_ATOMIC )( 'Site Editor: Limited Global Styles', function () {
    test( 'Site should not be atomic', () => {
		expect( envVariables.TEST_ON_ATOMIC ).toBeFalsy();
	} );

```

rcrdortiz (approved):
Nice job, clean and easy changes! The LGTM also tests well.

I've reviewed the e2e tests for this build and there are no errors related to these changes. We could add another check, it would be redundant but would explicitly fail if we are testing on atomic.

```
diff --git a/test/e2e/specs/fse/fse__limited-global-styles.ts b/test/e2e/specs/fse/fse__limited-global-styles.ts
index f821b747dd..32eb1b0e08 100644
--- a/test/e2e/specs/fse/fse__limited-global-styles.ts
+++ b/test/e2e/specs/fse/fse__limited-global-styles.ts
@@ -19,6 +19,10 @@ declare const browser: Browser;
  * @see https://github.com/Automattic/wp-calypso/pull/71333#issuecomment-1592490057
  */
 skipDescribeIf( envVariables.TEST_ON_ATOMIC )( 'Site Editor: Limited Global Styles', function () {
+       test( 'Site should not be atomic', () => {
+               expect( envVariables.TEST_ON_ATOMIC ).toBeFalsy();
+       } );
+
        let page: Page;
        let fullSiteEditorPage: FullSiteEditorPage;
        let testAccount: TestAccount;
```

worldomonation:
> We could add another check, it would be redundant but would explicitly fail if we are testing on atomic.

That is not recommended, as it will cause TeamCity to mark every single Gutenberg build for AT to be erroneously marked as a fail.

worldomonation:
The failure for `E2E Tests (mobile) (Web app) — TeamCity build failed ` is known and tracked in https://github.com/Automattic/wp-calypso/issues/78255.

worldomonation (approved):
Nicely done.

PR #77493, opened 2023-05-29
Plans 2023: Single pricing grid. Remove feature flag

chriskmnds:
Related to https://github.com/Automattic/wp-calypso/issues/76766
Closes https://github.com/Automattic/wp-calypso/issues/77651

## Proposed Changes

Cleans up all the plumbing needed previously to render the new grid along with the old in Signup/start and Stepper.

Technically removes all references to:
- `is2023OnboardingPricingGrid`
- `onboarding/2023-pricing-grid` flag
- `is2023PricingGridEnabled`
- `is2023PricingGridVisible`

## Testing Instructions

Same as in https://github.com/Automattic/wp-calypso/pull/76768

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #78262, opened 2023-06-15
Plans: Remove references to Happy Chat

klimeryk:

## Proposed Changes

* Removes FAQ component from Plans page - [it's confirmed dead](https://github.com/Automattic/wp-calypso/pull/78262#issuecomment-1593144469).

## Testing Instructions

- Do a code search and confirm that this is indeed dead code.
- Although E2E tests should take care of this, spot check a few flows: /start, /plans/<site slug>, /setup/newsletter/plans.


klimeryk:
Looks like the `showFAQ` flag was indeed hardcoded in https://github.com/Automattic/wp-calypso/pull/77493 from a previous value of `showFAQ: !! props.showFAQ && ! is2023PricingGridVisible,`. @aneeshd16, @chriskmnds, should the `WpcomFAQ` component then be simply removed? If it's not gonna be used, it's dead code, but still requires maintenance, as seen by this PR.

aneeshd16:
Thanks for the ping @klimeryk! I checked the usage of the component, and I think we can remove it safely. I'll put up a PR for this shortly.

client/my-sites/plans-features-main/index.jsx
```diff
@@ -26,7 +26,6 @@ import {
 	getPlanPath,
 	GROUP_WPCOM,
 	PLAN_PERSONAL,
-	TITAN_MAIL_MONTHLY_SLUG,
```
AtrumGeost, line 29: It looks like this was removed intentionally, I'm just curious why since doesn't seem related to Happy Chat orthe wpcom-faq implementations.

AtrumGeost (commented):
Thanks for working on this 🙏 It removes noise and makes everyone's work easier.

I searched through the modified files and the references are properly removed. I just left a question about something that dragged my curiosity.

client/my-sites/plans-features-main/index.jsx
klimeryk: It goes back to this line:
https://github.com/Automattic/wp-calypso/pull/78262/files#diff-2545c62593dd002397248377b1c0feb643bd2acc7694054ce094997b1c029aceL422

Which was removed because that whole section is dead. `titanMonthlyRenewalCost` is set on this line:
https://github.com/Automattic/wp-calypso/pull/78262/files#diff-2545c62593dd002397248377b1c0feb643bd2acc7694054ce094997b1c029aceL557

Which depends on the `TITAN_MAIL_MONTHLY_SLUG` constant. Since it was not longer referenced, I removed it. Hope that clears it up!

PR #78263, opened 2023-06-15
Theme showcase: De-emphasize WP.org classic themes

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2747

## Proposed Changes

De-emphasize the WP.org classic themes in the theme showcase by moving them to the bottom of the results.

Before | Astra
--- | ---
Screenshot 2023-06-15 at 15 38 42 | Screenshot 2023-06-15 at 15 38 13

Note: The code of the interlacing logic has been rewritten to improve its readability. 

## Testing Instructions

- Go to Appearance > Themes.
- Enter a search term that produces results that include WP.org block themes and WP.org classic themes (e.g. Hotel).
- Make sure WP.org block themes (e.g. Elevated Lite, Strategist FSE) are prioritized over WP.org classic themes (e.g. Entr, Avantex).

### DIFF
--- a/client/my-sites/themes/helpers.js
+++ b/client/my-sites/themes/helpers.js
@@ -166,25 +166,22 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
 
 	// 3. WP.org themes (only if the list of WP.com themes has reached the last page).
 	if ( isEnabled( 'themes/interlaced-dotorg-themes' ) && isLastPage ) {
+		const restWpOrgThemes = matchingTheme
+			? wpOrgThemes.filter( ( theme ) => theme.id !== matchingTheme.id )
+			: wpOrgThemes;
+
 		// 3.1. WP.org block themes.
-		const restWpOrgBlockThemes = validWpOrgThemes.filter(
-			( theme ) =>
-				theme.id !== matchingTheme?.id &&
-				theme.taxonomies?.theme_feature?.some(
-					( feature ) => feature?.slug === 'full-site-editing'
-				)
+		const restWpOrgBlockThemes = restWpOrgThemes.filter( ( theme ) =>
+			theme.taxonomies?.theme_feature?.some( ( feature ) => feature?.slug === 'full-site-editing' )
 		);
 		interlacedThemes.push( ...restWpOrgBlockThemes );
 
-		// 3.2 WP.org classic themes.
-		const restWpOrgClassicThemes = validWpOrgThemes.filter(
+		// 3.2. WP.org classic themes.
+		const restWpOrgClassicThemes = restWpOrgThemes.filter(
 			( theme ) =>
-				theme.id !== matchingTheme?.id &&
-				( ! theme.taxonomies ||
-					! theme.taxonomies.theme_feature ||
-					theme.taxonomies.theme_feature.every(
-						( feature ) => feature?.slug !== 'full-site-editing'
-					) )
+				! theme.taxonomies ||
+				! theme.taxonomies.theme_feature ||
+				theme.taxonomies.theme_feature.every( ( feature ) => feature?.slug !== 'full-site-editing' )
 		);
 		interlacedThemes.push( ...restWpOrgClassicThemes );
 	}
--- a/client/my-sites/themes/helpers.js
+++ b/client/my-sites/themes/helpers.js
@@ -166,23 +166,25 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
 
 	// 3. WP.org themes (only if the list of WP.com themes has reached the last page).
 	if ( isEnabled( 'themes/interlaced-dotorg-themes' ) && isLastPage ) {
-		const restWpOrgThemes = matchingTheme
-			? wpOrgThemes.filter( ( theme ) => theme.id !== matchingTheme.id )
-			: wpOrgThemes;
+		const restWpOrgBlockThemes = [];
+		const restWpOrgClassicThemes = [];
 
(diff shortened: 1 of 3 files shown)


client/my-sites/themes/helpers.js
```diff
@@ -148,20 +149,45 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
...
-		...restWpComThemes,
-		// Include WP.org themes after the last page of WP.com themes.
-		...( isEnabled( 'themes/interlaced-dotorg-themes' ) && isLastPage ? restWpOrgThemes : [] ),
-	];
+	interlacedThemes.push( ...restWpComThemes );
+
+	// 3. WP.org themes (only if the list of WP.com themes has reached the last page).
+	if ( isEnabled( 'themes/interlaced-dotorg-themes' ) && isLastPage ) {
+		// 3.1. WP.org block themes.
+		const restWpOrgBlockThemes = validWpOrgThemes.filter(
```
Copons: What about extracting the "rest of Dotorg themes" once before checking if they are block or classic, as we do for Dotcom ones?

E.g.

```js
const restWpOrgThemes = matchingTheme
	? wpOrgThemes.filter( ( theme ) => theme.id !== matchingTheme.id )
	: wpOrgThemes;

const restWpOrgBlockThemes = restWpOrgThemes.filter( ... );
const restWpOrgClassicThemes = restWpOrgThemes.filter( ... );
```

client/my-sites/themes/helpers.js
```diff
@@ -148,20 +149,45 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
...
+		// 3.2 WP.org classic themes.
+		const restWpOrgClassicThemes = validWpOrgThemes.filter(
+			( theme ) =>
+				theme.id !== matchingTheme?.id &&
+				( ! theme.taxonomies ||
+					! theme.taxonomies.theme_feature ||
+					theme.taxonomies.theme_feature.every(
+						( feature ) => feature?.slug !== 'full-site-editing'
+					) )
+		);
```
Copons: Wonder if we could do a single loop here with foreach (or reduce even). 🤔 

E.g.

```js
const restWpOrgBlockThemes = [];
const restWpOrgClassicThemes = [];

wpOrgThemes.forEach( theme, => {
	if ( theme.id === matchingTheme?.id ) {
		return;
	}
	if ( theme.taxonomies?.theme_feature?.some(
		( feature ) => feature?.slug === 'full-site-editing'
	) {
		restWpOrgBlockThemes.push( theme );
	} else {
		restWpOrgClassicThemes.push( theme );
	}
} );
```

client/my-sites/themes/helpers.js
```diff
@@ -148,20 +149,45 @@ export function interlaceThemes( wpComThemes, wpOrgThemes, searchTerm, isLastPag
...
+		interlacedThemes.push( ...restWpOrgBlockThemes );
+
+		// 3.2 WP.org classic themes.
+		const restWpOrgClassicThemes = validWpOrgThemes.filter(
+			( theme ) =>
+				theme.id !== matchingTheme?.id &&
+				( ! theme.taxonomies ||
+					! theme.taxonomies.theme_feature ||
+					theme.taxonomies.theme_feature.every(
```
Copons: I guess this can be simplified with the optional chaining like it was done for `restWpOrgBlockThemes`. 🙂

Copons (commented):
It is a bit hard to verify that block themes are actually promoted over classic ones, but the code logic LGTM, and at a glance it seems to be working correctly.

client/my-sites/themes/helpers.js
mmtr: That's how I had it at some point and ultimately decided to remove the `restWpOrgThemes` var, because that way we only need to iterate `wpOrgThemes` twice instead of three times. 

It's probably a micro-optimization that doesn't have too much impact, so happy to bring it back.

client/my-sites/themes/helpers.js
mmtr: Oh, that's neat!

mmtr: https://github.com/Automattic/wp-calypso/pull/78263/commits/c9fe8395d3156e73db5c413c22057ba89123565f

client/my-sites/themes/helpers.js
mmtr: Not needed anymore after using the single loop you suggested.

client/my-sites/themes/helpers.js
mmtr: Not needed anymore after using the single loop you suggested.

mmtr:
> verify that block themes are actually promoted over classic ones

One way I used while developing is to check the API response that fetches the WP.org themes, inspect the tags of themes, and verify that those with the `full-site-editing` tag are promoted. Not a super smooth process, but does the trick.

Copons:
> > verify that block themes are actually promoted over classic ones
> 
> One way I used while developing is to check the API response that fetches the WP.org themes, inspect the tags of themes, and verify that those with the `full-site-editing` tag are promoted. Not a super smooth process, but does the trick.

Yup, I've been doing something similar checking the "features" tags in the details screen of a sample of themes. 🙂 

Screenshot 2023-06-15 at 16 06 55

Copons (approved):
Thank you Miguel, LGTM! ✨

rcrdortiz:
Added the ready-to-merge tag, I'm not sure if this was purposely left in the needs review state so feel free to move it back if needed.

PR #78279, opened 2023-06-15
Hosting onboarding: reuse "I am a developer" copy from /me

zaguiini:
## Proposed Changes

During the CfT, it was related that using "Boost" my experience might be misleading because of Jetpack Boost. Let's reword it. It's better to use the same wording that exists in `/me`.

## Testing Instructions

Check that the wording is different on `/start/hosting` and that the copy is shared between this page and `/me`.

| `trunk` | This PR |
| ------- | -------- |
| image | image |

### DIFF
--- a/client/blocks/signup-form/index.jsx
+++ b/client/blocks/signup-form/index.jsx
@@ -51,6 +51,7 @@ import {
 import { login, lostPassword } from 'calypso/lib/paths';
 import { addQueryArgs } from 'calypso/lib/url';
 import wpcom from 'calypso/lib/wp';
+import { getIAmDeveloperCopy } from 'calypso/me/profile/get-i-am-a-developer-copy';
 import { isP2Flow } from 'calypso/signup/utils';
 import { recordTracksEventWithClientId } from 'calypso/state/analytics/actions';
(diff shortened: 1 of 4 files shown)


mpkelly:
Tested this on desktop and mobile, and it looks good. Nice work moving it into a single component to keep it consistent. 

For me, "make my experience more powerful" and "boost" mean the same thing. I would prefer something explicit that more accurately describes what dev mode means today, _like_ this:

Screenshot 2566-06-16 at 14 22 43

We can update this text later as we add more developer-specific features.

zaguiini:
Thanks @mpkelly the problem with changing the wording and making it more objective is that we will need to update it whenever a relevant feature comes out, which is a painful maintenance point. Giving a rough idea of what being a developer is like is a better option.

mpkelly:
>Thanks @mpkelly the problem with changing the wording and making it more objective is that we will need to update it whenever a relevant feature comes out

@zaguiini, Yeah, I get that, and it's not ideal. But I think we will attract too many "devs" if we aren't more explicit. It's merged now anyway so let's see how it does.

PR #78281, opened 2023-06-15
Site options: do not generate title and autofocus the input

zaguiini:
## Proposed Changes

During the CfT, we received negative feedback about autogenerating titles. It's an important feature of a hosted site because the domain is derived from the site name, and we provide paid domain suggestions based on it, so that's something that the user needs to input.

[Another option](https://github.com/Automattic/wp-calypso/compare/hosting/do-not-add-suggestion-new-hosted-site...hosting/do-not-enforce-suggestion?expand=1) would be to not add anything but still let them pick a suggestion:

image

@javierarce how do you imagine the right adornment? Is this version OK? For small screens, what icon do you recommend?

## Testing Instructions

Pick any plan in `/setup/new-hosted-site` and verify that the input is empty and focused on mount.

image


mpkelly:
> how do you imagine the right adornment?

I prefer this approach myself. Some people might not care what the name is and can save themselves the stress of thinking of something. 

Another thing we might consider is showing a preview of their domain name, so it's clear what the full implications of choosing a title are: 

Screenshot 2566-06-16 at 14 40 20

zaguiini:
Awesome. Waiting for @javierarce's input.

javierarce:
> Another thing we might consider is showing a preview of their domain name, so it's clear what the full implications of choosing a title are:

Can we do this? We'll need to check if the domain is available and, if not, generate an alternative.

For the action inside the button, we could do something like this (similar to the way the filter action in the domain search bar works):

image

The first icon is one that exists in Gutenberg (shuffle), the other ones come from material (one it's called "auto fix" and the other one "auto awesome"). 

cc: @Automattic/dotcom-design for some design feedback about the idea.

zaguiini:
>Another thing we might consider is showing a preview of their domain name, so it's clear what the full implications of choosing a title are

>Can we do this? We'll need to check if the domain is available and, if not, generate an alternative

Technically, we can, but I don't think we should. The user might think that’s the domain they’ll get for free when signing up for an annual plan. This is misleading because `.wpcomstaging.com` is actually a free domain, not the paid one that they're getting for free.

There’s no space to say “temporary domain” or something like that (and even if there was space, it wouldn't sound great). So in order to avoid confusion, we're not going to add the suggested domain.

zaguiini:
This is the final version:

image

crisbusquets:
> This is the final version:
> 
> image

Thanks, @zaguiini !

P.S.: This could be also useful for the Newsletter onboarding (cc @simison @lezama @edanzer ).

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8050494

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8050494).

PR #78283, opened 2023-06-15
Subscription Management: Pass email_id url param to unsubscribe API endpoint on individual site subscription page.

### LINKED ISSUE #78293: Subscription Management: Fix tracks event `wpcom_subscriber_unsubscribe` not recorded.
When we started redirecting users from unsubscribe link to the new individual subscriptions page, there is a regression where the `wpcom_subscriber_unsubscribe` is no longer being recorded. Context: p1686830372954469/1686788048.219179-slack-CA6GBMY2W

In order for this to work:
- the tracks `email_id` needs to be passed over from the unsubscribe link to the new individual subscriptions page
- the `email_id` needs to be passed back to the API call that does the unsubscription
- the unsubscribe API needs to record tracks event if the email_id param exists

Further discussion here:
p1686841144096129-slack-C02TCEHP3HA

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/78293

## Proposed Changes

This PR makes the unsubscribe action in the Individual Subscription page pass email_id to the unsubscribe endpoint if that parameter is on the URL.
Screenshot 2023-06-15 at 18 47 38

## Testing Instructions
Note: use these instructions for internal and external users.

For internal users the endpoint should have a structure like this:
Screenshot 2023-06-15 at 18 38 06

For external users the endpoint should have a structure like this:
Screenshot 2023-06-15 at 18 39 53

1. Apply this patch and start the application.
2. Go to `http://calypso.localhost:3000/subscriptions/site/{id of any page you are subscribed to}`.
3. Open the network tab in the developer tools.
4. Click on `Cancel subscription` and confirm the cancellation.
5. Check the URL of the unsubscribe endpoint. It shouldn't contain any reference to `email_id` on it.
6. Resubscribe using the resubscribe button on the body of the page.
7. Add the `email_id` parameter with any number, so the URL is like this: `http://calypso.localhost:3000/subscriptions/site/{id of any page you are subscribed to}?email_id={put any number you want here}`.
8. Click on `Cancel subscription` and confirm the cancellation.
9. Check the URL of the unsubscribe endpoint. It should contain a new query parameter `email_id` on it.


packages/data-stores/src/reader/helpers/index.ts
```diff
@@ -83,23 +83,48 @@ const applyCallbackToPages = < K extends string, T >(
 	};
 };
 
+function getQueryParamFromUrl( param: string ): string | null {
+	let urlObj;
+	try {
+		urlObj = new URL( window.location.href );
```
phcp: Instead of accessing the `window` inside the `data-store`, you could consider adding the `email_id` as an optional param on the unsubscribe mutation. What do you think?

JuanLucha: I like it! I'll implement it 👍

client/landing/subscriptions/components/site-subscription-page/site-subscription-details.tsx
```diff
@@ -3,6 +3,7 @@ import { useLocale } from '@automattic/i18n-utils';
 import { Button } from '@wordpress/components';
 import { useTranslate, numberFormat } from 'i18n-calypso';
 import { useEffect, useState, useMemo } from 'react';
+import { useLocation } from 'react-router';
```
yansern: Try using `getQueryArgs()`.
```suggestion
import { getQueryArgs } from 'calypso/lib/query-args';
```

client/landing/subscriptions/components/site-subscription-page/site-subscription-details.tsx
```diff
@@ -51,7 +53,9 @@ const SiteSubscriptionDetails = ( {
 				'You currently have paid subscriptions with this site. Paid subscriptions must be cancelled separately by clicking the Manage Subscriptions button or going to https://wordpress.com/me/purchases.\n\nPress OK to proceed with unsubscribing from the site.\nPress Cancel to go back.'
 			)
 		) {
-			unsubscribe( { blog_id: blogId, url } );
+			const query = new URLSearchParams( location.search );
+			const emailId = query.get( 'email_id' ) || undefined;
+			unsubscribe( { blog_id: blogId, url, emailId } );
```
yansern: The version with `getQueryArgs()`.
```suggestion
			const emailId = getQueryArgs()?.email_id;
			unsubscribe( { blog_id: blogId, url, emailId } );
```

client/landing/subscriptions/components/site-subscription-page/site-subscription-details.tsx
```diff
@@ -31,6 +32,7 @@ const SiteSubscriptionDetails = ( {
 	const translate = useTranslate();
 	const localeSlug = useLocale();
 	const [ notice, setNotice ] = useState< NoticeState | null >( null );
+	const location = useLocation();
```
yansern: Not needed if using `getQueryArgs()`.
```suggestion
```

yansern:
Reason to use `getQueryArgs()` is because `react-router` doesn't exist when rendered under Calypso and this might incur a bit of an overhead. (In the future, the individual sites subscription page will also be rendered inside Calypso.)

yansern (approved):
Moved email_id to post body instead. 
All works now.

packages/data-stores/src/reader/helpers/index.ts
```diff
@@ -88,20 +88,21 @@ const getSubscriptionMutationParams = (
...
-	url?: string
+	url?: string,
+	emailId?: string
 ) => {
 	if ( isLoggedIn ) {
 		return {
 			path: `/read/following/mine/${ action }`,
 			apiVersion: '1.1',
-			body: { source: 'calypso', url: url },
+			body: { source: 'calypso', url: url, ...( emailId ? { email_id: emailId } : {} ) },
```
yansern, line 98: Note to self: This is not used at the moment for the loggedIn endpoint. Still need to decide whether it is needed.

PR #78292, opened 2023-06-15
Migrate away from deprecated react types

noahtallen:
## Proposed Changes
React 18 includes fixes and changes for certain React types. Notably, `ReactFragment` is changing from `{} | Iterable<ReactNode>` to simply `Iterable<ReactNode>`. Additionally, the `ReactChild` and `ReactText` types have been deprecated.

The first change is big because it makes `ReactFragment` much narrower -- so types which used to be accepted now aren't.

The core place this impacts us is the translation function, which relied on `ReactFragment` before. With the `{}` type, that means the `translation` function currently accepts substitutions which might not exist. We don't want that behavior; we only want to accept substitutions which do exist.

Since this requires a lot of changes, I propose we separate it from #77046 and work on it here:

1. We can patch `@types/react` to simply remove `{}` on `ReactFragment`. This lets us get the biggest part of the upgrade in the repo more quickly. (We'll remove this after #77046 is merged.)
2. Ban `ReactChild` and `ReactText` across the repo to prevent new uses.
3. Migrate away from `ReactChild` and `ReactText`
4. Fix type issues associated with all these changes

## Testing Instructions

Typecheck should pass. Most changes should be to types.

### DIFF
--- a/.yarn/patches/@types-react-npm-17.0.39-b4ac1f7bfe.patch
+++ b/.yarn/patches/@types-react-npm-17.0.39-b4ac1f7bfe.patch
@@ -1,0 +1,13 @@
+diff --git a/index.d.ts b/index.d.ts
+index 2d617bcb6e6d378d99a89fdf37fd632eb9fa325b..094587a884e2dc28be3f2cfb77bb1089dd64e476 100755
+--- a/index.d.ts
++++ b/index.d.ts
+@@ -236,7 +236,7 @@ declare namespace React {
+      * @deprecated Use either `ReactNode[]` if you need an array or `Iterable<ReactNode>` if its passed to a host component.
+      */
+     interface ReactNodeArray extends ReadonlyArray<ReactNode> {}
+-    type ReactFragment = {} | Iterable<ReactNode>;
++    type ReactFragment = Iterable<ReactNode>;
+     type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
+ 
+     //
--- a/client/blocks/importer/components/importer-drag/config.tsx
+++ b/client/blocks/importer/components/importer-drag/config.tsx
@@ -66,7 +66,7 @@ export function getImportDragConfig( importer: Importer, supportLinkModal?: bool
 			),
 		},
 	};
-	const importerData = importerConfig()[ importer ];
+	const importerData = importerConfig( {} )[ importer ];
 
 	importerData.title = translate( 'Import content from %(importerName)s', {
 		...options,
--- a/client/components/jetpack/threat-item-header/index.tsx
+++ b/client/components/jetpack/threat-item-header/index.tsx
@@ -31,9 +31,11 @@ const getThreatMessage = ( threat: Threat ) => {
 
 	switch ( getThreatType( threat ) ) {
 		case 'core':
-			return translate( 'Vulnerable WordPress version: %s', {
-				args: [ version ?? 'unknown' ],
-			} );
+			return version
+				? translate( 'Vulnerable WordPress version: %s', {
+						args: [ version ],
+				  } )
+				: translate( 'Vulnerable WordPress version.' );
 
 		case 'core_file':
 			return translate( 'Infected core file: %s', {
--- a/client/components/jetpack/threat-item-header/index.tsx
+++ b/client/components/jetpack/threat-item-header/index.tsx
@@ -32,7 +32,7 @@ const getThreatMessage = ( threat: Threat ) => {
 	switch ( getThreatType( threat ) ) {
 		case 'core':
 			return translate( 'Vulnerable WordPress version: %s', {
-				args: [ version ],
+				args: [ version ?? 'unknown' ],
 			} );
 
 		case 'core_file':
@@ -72,7 +72,7 @@ const getThreatMessage = ( threat: Threat ) => {
 					count: Object.keys( threat.rows ).length,
 					args: {
 						threatCount: Object.keys( threat.rows ).length,
-						threatTable: threat.table,
+						threatTable: threat.table as string,
 					},
 				}
 			);
--- a/client/components/jetpack/threat-item/utils.ts
+++ b/client/components/jetpack/threat-item/utils.ts
@@ -31,7 +31,7 @@ export const getThreatMessage = ( threat: Threat ): string | TranslateResult => 
 	switch ( getThreatType( threat ) ) {
 		case 'core':
 			return translate( 'The installed version of WordPress (%s) has a known vulnerability.', {
-				args: [ version ],
+				args: [ version ?? 'unknown' ],
 			} );
 
 		case 'core_file':
@@ -71,7 +71,7 @@ export const getThreatMessage = ( threat: Threat ): string | TranslateResult => 
 					count: Object.keys( threat.rows ).length,
 					args: {
 						threatCount: Object.keys( threat.rows ).length,
-						threatTable: threat.table,
+						threatTable: threat.table as string,
 					},
 				}
 			);
--- a/client/components/jetpack/threat-item/utils.ts
+++ b/client/components/jetpack/threat-item/utils.ts
@@ -30,9 +30,11 @@ export const getThreatMessage = ( threat: Threat ): string | TranslateResult => 
 
 	switch ( getThreatType( threat ) ) {
 		case 'core':
-			return translate( 'The installed version of WordPress (%s) has a known vulnerability.', {
-				args: [ version ?? 'unknown' ],
-			} );
+			return version
+				? translate( 'The installed version of WordPress (%s) has a known vulnerability.', {
+						args: [ version ],
+				  } )
+				: translate( 'The installed version of WordPress has a known vulnerability.' );
 
 		case 'core_file':
 			return translate( 'Compromised WordPress core file: %s', {
--- a/client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
+++ b/client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
@@ -277,7 +277,7 @@ export default function UpsellStep( { upsell, site, purchase, ...props }: StepPr
 							'But we’d love to see you stick around to build on what you started. ' +
 							'How about a free month of your %(currentPlan)s plan subscription to continue building your site?',
 						{
-							args: { planName: getPlan( purchase.productSlug )?.getTitle() },
+							args: { planName: getPlan( purchase.productSlug )?.getTitle() ?? '' },
 						}
 					) }
 				</Upsell>
--- a/client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
+++ b/client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
@@ -21,7 +21,7 @@ import type { Purchase } from 'calypso/lib/purchases/types';
 import type { TranslateResult } from 'i18n-calypso';
 
 type UpsellProps = {
-	children: React.ReactChild;
+	children?: React.ReactNode;
 	image: string;
 	title: TranslateResult;
 	acceptButtonText: TranslateResult;
--- a/client/jetpack-cloud/sections/partner-portal/primary/onboarding-widget/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/primary/onboarding-widget/index.tsx
@@ -48,7 +48,7 @@ export default function OnboardingWidget( { isLicensesPage }: { isLicensesPage?:
 			description: translate(
 				'Manage features and monitor your clients’ sites by adding them to your Jetpack Pro Dashboard. To do so, connect the sites to Jetpack using your {{strong}}%(userEmail)s{{/strong}} user account.',
 				{
-					args: { userEmail: user?.email },
+					args: { userEmail: user?.email ?? '' },
 					components: {
 						strong: <strong />,
 					},
--- a/client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
@@ -24,6 +24,11 @@ export default function Prices() {
 		const userYearlyProduct = Object.values( userProducts ).find(
 			( p ) => p.product_id === product.product_id
 		);
+
+		if ( userYearlyProduct === undefined ) {
+			return null;
+		}
+
 		const userMonthlyProduct =
 			userYearlyProduct &&
 			Object.values( userProducts ).find(
@@ -57,9 +62,9 @@ export default function Prices() {
 					<div>
 						{ translate( '%(price)s/month', {
 							args: {
-								price:
-									userMonthlyProduct &&
-									formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions ),
+								price: userMonthlyProduct
+									? formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions )
+									: translate( 'unknown' ),
 							},
 						} ) }
 					</div>
--- a/client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
@@ -29,6 +29,9 @@ export default function Prices() {
 			return null;
 		}
 
+		const dailyAgencyPrice = ( product.amount * 12 ) / 365;
+		const dailyUserYearlyPrice = userYearlyProduct.cost / 365;
+
 		const userMonthlyProduct =
 			userYearlyProduct &&
 			Object.values( userProducts ).find(
@@ -36,10 +39,9 @@ export default function Prices() {
 					p.billing_product_slug === userYearlyProduct.billing_product_slug &&
 					p.product_term === 'month'
 			);
-
-		const dailyAgencyPrice = ( product.amount * 12 ) / 365;
-		const dailyUserMonthlyPrice = userMonthlyProduct ? ( userMonthlyProduct.cost * 12 ) / 365 : 0;
-		const dailyUserYearlyPrice = userYearlyProduct ? userYearlyProduct.cost / 365 : 0;
+		const dailyUserMonthlyPrice = userMonthlyProduct
+			? ( userMonthlyProduct.cost * 12 ) / 365
+			: undefined;
 
 		return (
 			<tr key={ product.product_id }>
@@ -48,26 +50,28 @@ export default function Prices() {
 					<LicenseBundleCardDescription product={ product } />
 				</td>
 				<td>
-					<div className="prices__mobile-description">
-						<div>{ translate( 'Jetpack.com Pricing' ) }</div>
-						<span className="prices__th-detail">{ translate( 'billed monthly' ) }</span>
-					</div>
-					<div>
-						{ translate( '%(price)s/day', {
-							args: {
-								price: formatCurrency( dailyUserMonthlyPrice, 'USD', currencyFormatOptions ),
-							},
-						} ) }
-					</div>
-					<div>
-						{ translate( '%(price)s/month', {
-							args: {
-								price: userMonthlyProduct
-									? formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions )
-									: translate( 'unknown' ),
-							},
-						} ) }
-					</div>
+					{ userMonthlyProduct && dailyUserMonthlyPrice && (
+						<>
+							<div className="prices__mobile-description">
+								<div>{ translate( 'Jetpack.com Pricing' ) }</div>
+								<span className="prices__th-detail">{ translate( 'billed monthly' ) }</span>
+							</div>
+							<div>
+								{ translate( '%(price)s/day', {
+									args: {
+										price: formatCurrency( dailyUserMonthlyPrice, 'USD', currencyFormatOptions ),
+									},
+								} ) }
+							</div>
+							<div>
+								{ translate( '%(price)s/month', {
+									args: {
+										price: formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions ),
+									},
+								} ) }
+							</div>
+						</>
+					) }
 				</td>
 				<td>
 					<div className="prices__mobile-description">
--- a/client/jetpack-cloud/sections/partner-portal/unassign-license-dialog/index.tsx
+++ b/client/jetpack-cloud/sections/partner-portal/unassign-license-dialog/index.tsx
@@ -78,7 +78,7 @@ export default function UnassignLicenseDialog( {
 					'Unassigning this license means that the site {{bold}}%(siteUrl)s{{/bold}} will no longer have access to {{bold}}%(product)s{{/bold}}. Once this action is completed, you will be able to assign the license to another site. You will continue to be billed.',
 					{
 						components: { bold: <strong /> },
-						args: { siteUrl, product },
+						args: { siteUrl: siteUrl ?? '', product },
 					}
 				) }
 				&nbsp;
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/blogger-intent/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/blogger-intent/index.tsx
@@ -25,6 +25,8 @@ const BlogIntent: Step = function BlogIntent() {
 		[]
 	);
 
+	const username = currentUser?.display_name ?? currentUser?.username;
+
 	return (
 		<>
 			<DocumentHead title={ translate( 'Create your blog' ) } />
@@ -37,9 +39,11 @@ const BlogIntent: Step = function BlogIntent() {
 				stepContent={
 					<div className="blogger-intent__container">
 						<h2 className="blogger-intent__heading">
-							{ translate( "Let's start your blog, %(username)s!", {
-								args: { username: currentUser?.display_name || currentUser?.username },
-							} ) }
+							{ username
+								? translate( "Let's start your blog, %(username)s!", {
+										args: { username },
+								  } )
+								: translate( "Let's start your blog!" ) }
 						</h2>
 						<div className="blogger-intent__content">
 							<div className="blogger-intent__row">
--- a/client/landing/subscriptions/components/sort-controls/sort-controls.tsx
+++ b/client/landing/subscriptions/components/sort-controls/sort-controls.tsx
@@ -22,8 +22,8 @@ const SortControls: < T extends string >( props: SortControlsProps< T > ) => Rea
 } ) => {
 	const translate = useTranslate();
 	const sortingLabel = useMemo(
-		() => options.find( ( option ) => option.value === value )?.label,
-		[ options, value ]
+		() => options.find( ( option ) => option.value === value )?.label ?? translate( 'unknown' ),
+		[ options, value, translate ]
 	);
 
 	return (
--- a/client/landing/subscriptions/components/sort-controls/sort-controls.tsx
+++ b/client/landing/subscriptions/components/sort-controls/sort-controls.tsx
@@ -22,9 +22,13 @@ const SortControls: < T extends string >( props: SortControlsProps< T > ) => Rea
 } ) => {
 	const translate = useTranslate();
 	const sortingLabel = useMemo(
-		() => options.find( ( option ) => option.value === value )?.label ?? translate( 'unknown' ),
-		[ options, value, translate ]
+		() => options.find( ( option ) => option.value === value )?.label,
+		[ options, value ]
 	);
+
+	if ( ! sortingLabel ) {
+		throw new Error( 'In SortControl, props.value must exist in props.options.' );
+	}
 
 	return (
 		<Dropdown
--- a/client/lib/domains/resolve-domain-status.tsx
+++ b/client/lib/domains/resolve-domain-status.tsx
@@ -20,12 +20,11 @@ import { transferStatus, type as domainTypes, gdprConsentStatus } from './consta
 import type { ResponseDomain } from './types';
 import type { Purchase } from 'calypso/lib/purchases/types';
 import type { CalypsoDispatch } from 'calypso/state/types';
-import type { I18N } from 'i18n-calypso';
-import type { ReactChild } from 'react';
+import type { I18N, TranslateResult } from 'i18n-calypso';
 
-export type ResolveDomainStatusReturn =
+type ResolveDomainStatusReturn =
 	| {
-			statusText: ReactChild | Array< ReactChild >;
+			statusText: TranslateResult;
 			statusClass:
 				| 'status-error'
 				| 'status-warning'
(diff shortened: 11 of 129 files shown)


packages/components/src/highlight-cards/annual-highlight-cards.tsx
```diff
@@ -29,7 +29,7 @@ export default function AnnualHighlightCards( {
 
 	const header = (
 		<h3 className="highlight-cards-heading">
-			{ Number.isFinite( year )
+			{ year != null && Number.isFinite( year )
```
noahtallen, line 32: isFinite doesn't restrict the type enough :(

noahtallen:
In cd26e5e, I went through every instance of `ReactChild` and refactored to what made sense for the scenario. I tried to go with the most restrictive type as much as possible, so using `string` was viable in a lot of places. I also used `TranslateResult` in several places which were dealing with more complex translations, `ReactElement` when components were being used, and finally `ReactNode` when we were dealing with children.

I also removed `ReactFragment` in one place in favor of `ReactNode`.

noahtallen:
In 85338b5, I went through all the `translate` calls that were breaking. This was relatively discouraging, but I followed this approach:

- If it was easy to add an extra guard without changing functionality, I did that.
- If I could change the substitution to a blank string without breaking the sentence, I did that.
- If there was a quick way to make the root type issue better, I fixed that.
- If I couldn't find a better solution, I casted to the right type (`as string`)

None of this should functionally change how these translations were working. Ideally, the types wouldn't be this loose in the first place.

I really wish this part of our codebase was much better. Despite having strong type tools at our disposal, all of these issues happen because the vast majority of our code can't guarantee that some value exists. Part of this is the fault of the redux store -- (for example, when you get the site title, the site might not exist, and you might get no value back). But this means basically everything is doing way more null checks than we really should have to be doing.

The other problem is when types aren't architected with guarantees in mind from the start. For example, a static map of information, with constant keys known ahead of time, and yet the function to get a value can return undefined. It should be using keyof, etc to get really solid type info around that. But this wasn't as big a problem as the first thing.

noahtallen:
Also, there is this eslint issue that might be a bug:

```sh
TypeError: Cannot read properties of undefined (reading 'name')
Occurred while linting client/my-sites/marketing/do-it-for-me/difm-lite-in-progress.tsx:69
Rule: "wpcalypso/i18n-no-variables"
```

I can't tell what it's seeing on that line, and neither does my IDE

tyxla:
> Also, there is this eslint issue that might be a bug:
> 
> ```shell
> TypeError: Cannot read properties of undefined (reading 'name')
> Occurred while linting client/my-sites/marketing/do-it-for-me/difm-lite-in-progress.tsx:69
> Rule: "wpcalypso/i18n-no-variables"
> ```
> 
> I can't tell what it's seeing on that line, and neither does my IDE

Fix coming up for this in #78305. I'll cherry-pick it here just to ensure that it fixes the issue, but we can land it separately if it does.

tyxla:
Seems like checks passed with #78305. I'll go ahead and merge it and rebase the branch again.

.yarn/patches/@types-react-npm-17.0.39-b4ac1f7bfe.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/index.d.ts b/index.d.ts
+index 2d617bcb6e6d378d99a89fdf37fd632eb9fa325b..094587a884e2dc28be3f2cfb77bb1089dd64e476 100755
+--- a/index.d.ts
++++ b/index.d.ts
+@@ -236,7 +236,7 @@ declare namespace React {
+      * @deprecated Use either `ReactNode[]` if you need an array or `Iterable<ReactNode>` if its passed to a host component.
+      */
+     interface ReactNodeArray extends ReadonlyArray<ReactNode> {}
+-    type ReactFragment = {} | Iterable<ReactNode>;
++    type ReactFragment = Iterable<ReactNode>;
```
tyxla, line 10: 👍 Sounds like a good compromise to me - it will save us from doing a bunch of changes in #77046.

client/blocks/importer/components/importer-drag/config.tsx
```diff
@@ -66,7 +66,7 @@ export function getImportDragConfig( importer: Importer, supportLinkModal?: bool
 			),
 		},
 	};
-	const importerData = importerConfig()[ importer ];
+	const importerData = importerConfig( {} )[ importer ];
```
tyxla, line 69: Could also be done in the definition of `importerConfig` (AKA `getConfig()`) but that's fine too.

client/components/jetpack/threat-item-header/index.tsx
```diff
@@ -32,7 +32,7 @@ const getThreatMessage = ( threat: Threat ) => {
 	switch ( getThreatType( threat ) ) {
 		case 'core':
 			return translate( 'Vulnerable WordPress version: %s', {
-				args: [ version ],
+				args: [ version ?? 'unknown' ],
```
tyxla: cc @Automattic/yamato-scan: if `version` can be `undefined`, this can be improved to support an extra string in that case, like just "Vulnerable WordPress version" without the version in it.

client/components/jetpack/threat-item/utils.ts
```diff
@@ -31,7 +31,7 @@ export const getThreatMessage = ( threat: Threat ): string | TranslateResult =>
 	switch ( getThreatType( threat ) ) {
 		case 'core':
 			return translate( 'The installed version of WordPress (%s) has a known vulnerability.', {
-				args: [ version ],
+				args: [ version ?? 'unknown' ],
```
tyxla: cc https://github.com/orgs/Automattic/teams/yamato-scan: if version can be undefined, this can be improved to support an extra string in that case, like just "The installed version of WordPress has a known vulnerability." without the version in it.

client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
```diff
@@ -277,7 +277,7 @@ export default function UpsellStep( { upsell, site, purchase, ...props }: StepPr
 							'But we’d love to see you stick around to build on what you started. ' +
 							'How about a free month of your %(currentPlan)s plan subscription to continue building your site?',
 						{
-							args: { planName: getPlan( purchase.productSlug )?.getTitle() },
+							args: { planName: getPlan( purchase.productSlug )?.getTitle() ?? '' },
```
tyxla, line 280: The string looks almost good if `planName` is an empty string. It will only have double spacing. cc @taggon

client/jetpack-cloud/sections/partner-portal/primary/onboarding-widget/index.tsx
```diff
@@ -48,7 +48,7 @@ export default function OnboardingWidget( { isLicensesPage }: { isLicensesPage?:
 			description: translate(
 				'Manage features and monitor your clients’ sites by adding them to your Jetpack Pro Dashboard. To do so, connect the sites to Jetpack using your {{strong}}%(userEmail)s{{/strong}} user account.',
 				{
-					args: { userEmail: user?.email },
+					args: { userEmail: user?.email ?? '' },
```
tyxla, line 51: If the email is nullish, we'll end up with double spacing in this string. Is this something we can improve, @vitozev?

client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
```diff
@@ -57,9 +62,9 @@ export default function Prices() {
 					<div>
 						{ translate( '%(price)s/month', {
 							args: {
-								price:
-									userMonthlyProduct &&
-									formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions ),
+								price: userMonthlyProduct
+									? formatCurrency( userMonthlyProduct.cost, 'USD', currencyFormatOptions )
+									: translate( 'unknown' ),
```
tyxla: Hmm, having an unknown price feels off. This should definitely be fixed; I'd instead expect us to not render the price at all if `userMonthlyProduct` is nullish. This seems like something we can fix in this PR actually. WDYT @noahtallen?

client/jetpack-cloud/sections/partner-portal/unassign-license-dialog/index.tsx
```diff
@@ -78,7 +78,7 @@ export default function UnassignLicenseDialog( {
 					'Unassigning this license means that the site {{bold}}%(siteUrl)s{{/bold}} will no longer have access to {{bold}}%(product)s{{/bold}}. Once this action is completed, you will be able to assign the license to another site. You will continue to be billed.',
 					{
 						components: { bold: <strong /> },
-						args: { siteUrl, product },
+						args: { siteUrl: siteUrl ?? '', product },
```
tyxla, line 81: cc @Automattic/jetpack-avalon - if `siteUrl` is nullish, the string above will end up with a double space and an empty `<strong />` block - is this something we can improve as a follow-up?

client/landing/stepper/declarative-flow/internals/steps-repository/blogger-intent/index.tsx
```diff
@@ -37,9 +39,11 @@ const BlogIntent: Step = function BlogIntent() {
...
 					<div className="blogger-intent__container">
 						<h2 className="blogger-intent__heading">
-							{ translate( "Let's start your blog, %(username)s!", {
-								args: { username: currentUser?.display_name || currentUser?.username },
-							} ) }
+							{ username
+								? translate( "Let's start your blog, %(username)s!", {
+										args: { username },
+								  } )
+								: translate( "Let's start your blog!" ) }
```
tyxla, line 46: A bummer that we need a new string, but I think that's the best way to get it right 👍

client/landing/subscriptions/components/sort-controls/sort-controls.tsx
```diff
@@ -22,8 +22,8 @@ const SortControls: < T extends string >( props: SortControlsProps< T > ) => Rea
 } ) => {
 	const translate = useTranslate();
 	const sortingLabel = useMemo(
-		() => options.find( ( option ) => option.value === value )?.label,
-		[ options, value ]
+		() => options.find( ( option ) => option.value === value )?.label ?? translate( 'unknown' ),
```
tyxla: It feels like `unknown` would be an odd choice here. Perhaps there's a better default sorting label we can use?

cc @Automattic/apex WDYT?

client/lib/domains/resolve-domain-status.tsx
```diff
@@ -347,7 +346,6 @@ export function resolveDomainStatus(
 						},
 						args: {
 							expiryDate: moment.utc( domain.expiry ).format( 'LL' ),
-							renewCta,
```
tyxla, line 350: I wonder what this was doing here in the first place 🤔

client/me/purchases/manage-purchase/index.tsx
```diff
@@ -1181,7 +1184,7 @@ class ManagePurchase extends Component<
 								<div className="manage-purchase__contact-partner">
 									{ translate( 'Please contact your site host %(partnerName)s for details', {
 										args: {
-											partnerName: getPartnerName( purchase ),
+											partnerName: getPartnerName( purchase ) ?? '',
```
tyxla, line 1187: I think that's a reasonable fallback. Perhaps it would be nice to fix the case of double spacing when partner name is nullish.

client/me/purchases/pre-cancellation-dialog/index.tsx
```diff
@@ -246,7 +246,7 @@ export const PreCancellationDialog = ( {
 						<FormattedHeader
 							brandFont
 							headerText={ translate( 'Are you sure you want to cancel your %(label)s plan?', {
-								args: { label: planLabel },
+								args: { label: planLabel ?? '' },
```
tyxla: Perhaps we can use the "current" string as a fallback? Then the string would become:

```
Are you sure you want to cancel your current plan?
```

If we decide to go that way, though, we may need to add context to the string to make it clearer for translators.

client/my-sites/checkout/checkout-thank-you/gift/gift-thank-you.tsx
```diff
@@ -94,7 +94,7 @@ export default function GiftThankYou( { site }: { site: number | string } ) {
 						width: '63px',
 					} }
 					thankYouTitle={ translate( 'All done! Thank you for supporting %(siteName)s.', {
-						args: { siteName },
+						args: { siteName: siteName ?? translate( 'this website' ) },
```
tyxla, line 97: 👍

client/my-sites/checkout/checkout-thank-you/jetpack-checkout-thank-you.tsx
```diff
@@ -75,8 +75,8 @@ const JetpackCheckoutThankYou: FunctionComponent< Props > = ( {
 					>
 						{ translate( '%(productName)s was added to your site %(siteName)s.', {
 							args: {
-								productName,
-								siteName,
+								productName: productName as string, // We know this exists via hasProductInfo
+								siteName: siteName ?? '',
```
tyxla, line 79: If `siteName` is nullish, the string will end up pretty awkwardly with a space and a dot. Anything we can address separately, @rcoll or @Automattic/jetpack-martech?

client/my-sites/checkout/checkout-thank-you/licensing-thank-you-manual-activation-license-key.tsx
```diff
@@ -38,7 +38,7 @@ const LicensingActivationInstructions: FC< JetpackLicenseKeyProps > = ( {
 			/>
 			<LicensingActivation
 				title={ translate( 'Activate your %(productName)s plan', {
-					args: { productName },
+					args: { productName: productName ?? '' },
```
tyxla, line 41: If `productName ` is nullish, the string will end up having just double space. Anything we can address separately, @rcanepa  or @Automattic/jetpack-martech?

client/my-sites/checkout/composite-checkout/components/item-variation-picker/variant-dropdown-price.tsx
```diff
@@ -135,20 +135,19 @@ export const ItemVariantDropDownPrice: FunctionComponent< {
...
-	};
 
 	const translatedIntroOfferDetails = () => {
 		const args = {
 			formattedCurrentPrice,
 			formattedPriceBeforeDiscounts,
-			billingTermInYears: billingTermInYears(),
 			introCount,
 		};
+		// Add billing term in years for multi-year plans. (Only to be used when productBillingTermInMonths > 12)
```
tyxla, line 145: This can be a bit error-prone, but let's hope that the comment can help.

I've confirmed that we've used it only when the billing term in months is actually > 12.

client/my-sites/email/email-non-domain-owner-message/index.tsx
```diff
@@ -96,7 +96,7 @@ export const EmailNonDomainOwnerMessage = ( props: EmailNonDomainOwnerMessagePro
 		},
 		args: {
 			ownerUserName,
-			selectedDomainName: domain?.name,
+			selectedDomainName: domain?.name ?? '',
```
tyxla, line 99: Does it make sense if we default this to `'this domain'`? cc @Automattic/nomado for opinion

packages/i18n-calypso/types/index.d.ts
```diff
@@ -25,7 +29,7 @@ export interface ComponentInterpolations {
 
 export interface TranslateOptions {
 	/**
-	 * Arguments you would pass into sprintf to be run against the text for string substitution.
+	 * Arguments you would pass into sprintf to be run against the text for string substitution. Each substitution must exist. Iff a substitution shouldn't exist in some cases, pass an empty string to make the substitution explicit.
```
tyxla:
```suggestion
	 * Arguments you would pass into `sprintf` to be run against the text for string substitution. Each substitution must exist. If a substitution shouldn't exist in some cases, pass an empty string to make the substitution explicit.
```

packages/wpcom-checkout/src/checkout-line-items.tsx
```diff
@@ -752,7 +752,7 @@ export function LineItemSublabelAndPrice( { product }: { product: ResponseCartPr
 	const isDomainMapping = productSlug === 'domain_map';
 
 	if ( ( isDomainRegistration || isDomainMapping ) && product.months_per_bill_period === 12 ) {
-		const premiumLabel = product.extra?.premium ? translate( 'Premium' ) : null;
+		const premiumLabel = product.extra?.premium ? translate( 'Premium' ) : '';
```
tyxla, line 755: A bummer that we'll end up with an empty leading space in the case of an empty premium label. Anything we can/should separately fix, @sirbrillig?

client/my-sites/marketing/do-it-for-me/difm-landing.tsx
```diff
@@ -312,7 +312,7 @@ export default function DIFMLanding( {
 				'{{sup}}*{{/sup}}One time fee, plus an additional purchase of the %(plan)s plan. A WordPress.com professional will create layouts for up to %(freePages)d pages of your site. It only takes 4 simple steps:',
 				{
 					args: {
-						plan: planTitle,
+						plan: planTitle ?? '',
```
tyxla, line 317: I think we can totally fix this above where we declare `planTitle`. After all, `getPlan()` will always return the plan definition properly for those plans, so I'd say it's better if we instruct TS that it can't be nullish.

client/my-sites/marketing/do-it-for-me/difm-lite-in-progress.tsx
```diff
@@ -63,18 +63,18 @@ function WebsiteContentSubmissionPending( { primaryDomain, siteId, siteSlug }: P
...
-		},
 	};
 
 	const lineText = contentSubmissionDueDate
 		? translate(
 				'Click the button below to provide the content we need to build your site by %(contentSubmissionDueDate)s.{{br}}{{/br}}' +
 					'{{SupportLink}}Contact support{{/SupportLink}} if you have any questions.',
-				lineTextTranslateOptions
+				{
+					...lineTextTranslateOptions,
```
tyxla, line 73: 👍 Nice improvement.

FWIW this will no longer choke our custom ESLint rules since #78305 landed.

client/my-sites/people/team-invite/index.tsx
```diff
@@ -56,7 +56,7 @@ function TeamInvite( props: Props ) {
 				<PageViewTracker path="/people/new/:site" title="People > Invite People" />
 				<HeaderCake onClick={ goBack }>
 					{ translate( 'Add team members to %(sitename)s', {
-						args: { sitename: site.name },
+						args: { sitename: site.name ?? '' },
```
tyxla: Perhaps we can change the string to default to "this site"?

client/my-sites/people/team-invite/index.tsx
```diff
@@ -76,7 +76,7 @@ function TeamInvite( props: Props ) {
 
 			<HeaderCake onClick={ goBack }>
 				{ translate( 'Add team members to %(sitename)s', {
-					args: { sitename: site.name },
+					args: { sitename: site.name as string },
```
tyxla: Same as above - changing the string to default to `"this site"` would make this less awkward if `site.name` is nullish for some reason.

client/my-sites/plans-comparison/plans-comparison-action.tsx
```diff
@@ -55,7 +55,7 @@ function getButtonText( props: Partial< Props >, translate: TranslateFunc ): Tra
 
 	return translate( 'Start with %(plan)s', {
 		args: {
-			plan: planTitle,
+			plan: planTitle ?? '',
```
tyxla: This entire component would be so broken if `plan` is nullish 😅 

Should we default to just `Start` if there's no plan title for some reason? Otherwise, "Start with" is kind of awkward.

client/signup/steps/difm-site-picker/site-delete-confirmation-dialog.tsx
```diff
@@ -105,8 +105,8 @@ function SiteDeleteConfirmationDialog( {
 										strong: <strong />,
 									},
 									args: {
-										siteTitle,
-										siteAddress: siteDomain,
+										siteTitle: siteTitle ?? '',
+										siteAddress: siteDomain ?? '',
```
tyxla, line 109: The string would end up awkward with some additional spacing if `siteTitle` and/or `siteAddress` are empty. Is this something we can/should address as a side task, cc @Automattic/martech-decepticons?

tyxla (commented):
Thanks for separating the deprecated React type changes, that really helps make the primary React upgrade PR smaller. 

I've done a thorough review and made some suggestions, also pinged a few teams for follow-up fixes that could improve some of the strings in the case of nullish args. 

There are a couple of things I think we can improve in this PR before shipping it - let me know what you think!

Fantastic work, @noahtallen! 🚀

client/jetpack-cloud/sections/partner-portal/primary/onboarding-widget/index.tsx
vitozev: IIRC, this is the WPCOM user email address, which technically should always be present, but I agree with you. There are some upcoming projects (soon) to improve the onboarding flows, and we might not have this one anymore, but I'll keep your feedback in mind. Thanks!

client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
```diff
@@ -351,7 +351,7 @@ function useAddRenewalItems( {
 					translate(
 						"I tried and failed to create products matching the identifier '%(productAlias)s'",
 						{
-							args: { productAlias },
+							args: { productAlias: productAlias as string },
```
sirbrillig: Hm... I'd be inclined to use a fallback here rather than type casting. I feel that `as string` hides what could happen here and might encourage what I consider risky code in the future if other developers copy this example ("Even though the type here says this can be undefined, I don't want to handle that case and I'm pretty sure this is always a string" is a common reaction for devs new to TS and can be a major cause of errors). How about this?

```suggestion
							args: { productAlias: productAlias ?? '' },
```

client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
```diff
@@ -440,7 +440,7 @@ function useAddProductFromSlug( {
 					translate(
 						"I tried and failed to create products matching the identifier '%(productAlias)s'",
 						{
-							args: { productAlias: productAliasFromUrl },
+							args: { productAlias: productAliasFromUrl as string },
```
sirbrillig:
```suggestion
							args: { productAlias: productAliasFromUrl ?? '' },
```

client/blocks/importer/components/importer-drag/config.tsx
noahtallen: Right -- I had to update the type signature, and this was the easiest approach to getting the consumer to work (the type signature has default values for properties of the input object)

noahtallen:
Thanks for the comments @sirbrillig and @tyxla! Ideally, we wouldn't have so many null issues in the first place, and have better guarantees through the type system. But that would take a lot of effort -- this PR has to fix nearly 300 _different_ issues!

Ultimately, my approach in the PR was to try to find some balance on each issue one by one. If it was easy to add a more thorough fix, I did so, but ultimately, this mostly doesn't change how these components work on trunk. In reality, a lot of this data exists at runtime, but many components/types aren't designed to have strong type guarantees. (Hence why I used `as string` in a lot of places.)

For example, everything about a certain component might be designed with the assumption that a site is selected, and there could even be code that uses something like `site.ID` further up the tree (and even a component which might short-circuit if site doesn't exist!). But since the `getSelectedSite` selector doesn't have any return type guarantees, every component underneath it in the tree might add `site | undefined` or similar to its props to just get the types to work.

The vast majority of changes in this PR (related to `translate`) fall into that category. At the end of the day, the proper fix is to get more strict type guarantees throughout these components. But IMO that's too much to handle in this PR (which only exists to unblock #77046, which is blocking a lot of other things!)

I'll go through each of the comments so far and implement improvements, since those are likely higher impact, but I just wanted to mention most of this was an active consideration as I was working on the PR! There is not a straightforward solution, and at the end of the day, we'll need to improve our education around this so that teams know how to better utilize strong type guarantees. And hopefully the changes in this PR help people notice that more frequently!

client/components/jetpack/threat-item-header/index.tsx
noahtallen: Added a second string, though I doubt version is actually ever missing here

client/components/marketing-survey/cancel-purchase-form/step-components/upsell-step.tsx
noahtallen: I'm not convinced the plan ever doesn't exist here 🤔

client/jetpack-cloud/sections/partner-portal/primary/onboarding-widget/index.tsx
noahtallen: I'll leave this one as is for now 👍

client/jetpack-cloud/sections/partner-portal/primary/prices/index.tsx
noahtallen: The reason this was more challenging is that it's a `td` cell where there are four cells in a row (description, monthly price, yearly price, and agency price). I adjusted this to keep the `td` (to avoid messing up the layout), but the old code also had a default price of `0` (which we should probably avoid advertising ever :P)

So this is somewhat better, but I'm still guessing that the monthly price does exist

client/jetpack-cloud/sections/partner-portal/unassign-license-dialog/index.tsx
noahtallen: I think in this case, the `siteUrl` should always exist (otherwise, what would we be unassigning the license from), so I'd recommend trying to make the types for that stronger

client/landing/stepper/declarative-flow/internals/steps-repository/blogger-intent/index.tsx
noahtallen: I bet username has to exist at this point anyways since the user is logged in :/

client/landing/subscriptions/components/sort-controls/sort-controls.tsx
noahtallen: I thought about this a bit more, and with `SortControls`, it should be impossible to have a value which doesn't exist in the options. When selecting an option, the value is passed to the parent component and then passed back down. I've changed this to throw an error to indicate that this scenario should never happen, and verified that the consumers don't break this behavior. (Both consumers use enums for sort terms from what I can tell.)

noahtallen: I also looked into better types to enforce that value is an element of options, but there didn't seem to be a straightforward way to do that in function arguments

client/lib/domains/resolve-domain-status.tsx
noahtallen: My best guess is that it used to be used in the string but it wasn't removed when the string was updated. The string to translate is now `This domain expired on {{strong}}%(expiryDate)s{{/strong}}. `, so `renewCta` doesn't have any impact as far as I can tell, considering that `args` is just the string substitutions

client/me/purchases/manage-purchase/index.tsx
noahtallen: Do you have a recommendation on how to do that without creating a new string?

client/me/purchases/pre-cancellation-dialog/index.tsx
noahtallen: I decided to update this component to return null if `plan` doesn't exist. The component is about canceling your plan, which doesn't really make sense if a plan doesn't exist anyways! I also checked the consumers of the component, and they seem to only render it after checking that a plan exists.

client/my-sites/checkout/checkout-thank-you/licensing-thank-you-manual-activation-license-key.tsx
noahtallen: If this component really should work if `hasProductInfo` is false, I would recommend using that to create another message potentially

client/my-sites/email/email-non-domain-owner-message/index.tsx
noahtallen: Looking through the consuming code, I'm fairly sure a domain will always exist -- I'd recommending refining the types to reflect that if true!

client/my-sites/marketing/do-it-for-me/difm-landing.tsx
noahtallen: Added a cast 👍

client/my-sites/people/team-invite/index.tsx
noahtallen: Added

client/my-sites/plans-comparison/plans-comparison-action.tsx
noahtallen: Updated to shortcircuit when plan doesn't exist

client/signup/steps/difm-site-picker/site-delete-confirmation-dialog.tsx
noahtallen: I think this is another scenario where the component probably doesn't make sense when the site doesn't exist, so I'd recommend improving types to assert that a site does indeed exist here

client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
noahtallen: changed!

client/my-sites/checkout/composite-checkout/hooks/use-prepare-products-for-cart.ts
noahtallen: changed!

noahtallen:
I opened #78350 to explore better types for plans, which showed up a lot in the code. I think it'd be relatively easy to get a lot stricter types for the related code since we know most of the info up front

packages/wpcom-checkout/src/checkout-line-items.tsx
sirbrillig: Yeah, I wouldn't worry about it right now. This has been here for a long time and I think it either doesn't ever happen or was designed to be empty.

client/me/purchases/manage-purchase/index.tsx
tyxla: It's possible to do it with a single string, but it would become quite messy for translators, which I wouldn't recommend.  Two strings is the cleanest way IMO.

tyxla (approved):
I think this looks great! Thanks @noahtallen, great work as always. 💯 

As you recognized, there are some compromises we're making, but it's for the greater good and we're heading towards stricter types in #78350 anyway. I think we should go ahead and ship this as-is. 

I also think it would be a good idea if the pinged teams take a look at the code just in case. There could be some opportunities for improving the types or the logic to cover all those ede cases.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8058917

Hi @noahtallen, could you please edit the description of this PR and add a screenshot for our translators? Ideally it'd include all of the following strings:
- `Vulnerable WordPress version.`
- `The installed version of WordPress has a known vulnerability.`
- `Let's start your blog!`
- `this website`
- `this site`
- `Start with %(planTitle)s`

Thank you in advance!

noahtallen:
^ Unfortunately, I don't have a good way to test these flows, since I think they would only happen very rarely. It is low impact enough that I'm not very concerned about these strings

tyxla:
Agreed 👍 And what we're adding is already an improvement to what we had before, so I'm not concerned at all either.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8058917).

PR #78302, opened 2023-06-16
BackupBrowser: Fix backup browser zebra striping on sub-nodes

Initsogar:
## Proposed Changes

* Add a `isAlternate` prop to `FileBrowserNode` that will help us to decide which node will have a colored background and which not. Any time a new `FileBrowserNode` child is created, it will toggle the flag. This way, sub-nodes will start with a different background color than their parents.
* Sort items by `hasChildren` instead of by `type === 'dir'`. This way it will sort plugins and themes as well.

## Screenshots
| Change | Screenshot before | Screenshot after |
|---|---|---|
| Zebra striping | image | image |
| Sort items | image | image |

## Known issues
There might be cases where the last item in a directory will have the same background as the next upper level element.

## Testing Instructions
* Start a Jetpack Cloud live branch.
* Select one of your sites with a Jetpack VaultPress Backup plan
* Navigate to VaultPress Backup page
* Add `?flags=jetpack/backup-contents-page` at the end of the URL and refresh the page.
* Pick one date with a backup available (use the calendar and click on any date marked with a dot).
* On the daily backup card, click on `Actions (+)` and then click on `View files`.
* You should see a page similar to the one shared above.
* Ensure that the first item inside a directory starts with a background different from its parent.
* Also validate that `wp-content/plugins` and `wp-content/themes` are sorted as the image shared above.

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/backup/backup-contents-page/file-browser/file-browser-node.tsx
+++ b/client/my-sites/backup/backup-contents-page/file-browser/file-browser-node.tsx
@@ -1,5 +1,6 @@
 import { Button, Icon } from '@wordpress/components';
 import { chevronDown, chevronRight } from '@wordpress/icons';
+import classNames from 'classnames';
 import { FunctionComponent, useCallback, useState } from 'react';
 import FileTypeIcon from './file-type-icon';
 import { FileBrowserItem } from './types';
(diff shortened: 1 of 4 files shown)


jwebbdev (approved):
LGTM

> There might be cases where the last item in a directory will have the same background as the next upper level element.

Noticed this, but it's at least much better than previous.  Maybe an alternative is to not nest the nodes, but track their depth, and use that for left padding.  Then we may be able to just use the nth-child(odd) or whatever it is to do it.

Initsogar:
> Noticed this, but it's at least much better than previous. Maybe an alternative is to not nest the nodes, but track their depth, and use that for left padding. Then we may be able to just use the nth-child(odd) or whatever it is to do it.

Sounds good. Thanks for the suggestion. I will add this as an additional task to revisit later.

PR #78305, opened 2023-06-16
eslint-plugin-calypso: Fix spread operator error

tyxla:
## Proposed Changes

This PR resolves an error with the `fix/spread-element-i18n-no-variables` custom rule where it fails when attempting to lint a `translate()` call that has a spread element used in its `args`. 

## Testing Instructions

* Cherry-pick this in #78292
* Run `yarn run lint:js` and verify it works well and is able to complete the lint process.


tyxla:
This is a straightforward fix that now is confirmed to work in https://github.com/Automattic/wp-calypso/pull/78292#issuecomment-1594290635 so I'll just go ahead and ship it.

noahtallen (commented):
Looks great!

PR #78306, opened 2023-06-16
CI: tweak Jetpack VCS trigger path again.

worldomonation:
## Proposed Changes

This PR makes further tweaks to the VCS trigger path, because the Jetpack E2Es are still not triggering.

Key changes:
- remove the branch filter.
- copy the configuration as code directly from the sandbox build that does work.

## Testing Instructions

- [ ] No builds should fail.

## Pre-merge Checklist


dpasque (approved):
I'm good with this if we want to try it out! (And thank you! 😄) I do think we'll need some kind of branch filter logic there in the end since we only want to trigger on merges, but maybe we need to iterate on that later?

Edit** We can ignore the branch filter comment above. Because of how the sync happens, I think we're good and safe now starting without the branch logic, and we can add some more nuance later if really needed 👍

dpasque:
Merging and deploying for Edwin 👍

Issue #75156, opened 2023-03-31
Assembler Sidenav: adding clarity & guidance (checkmarks and premium badge)


autumnfjeld:
# Description

UPDATE on 8 May.

Per discussion below the current design to implement is this [comment](https://github.com/Automattic/wp-calypso/issues/75156#issuecomment-1515228150).  This design satisfies items 2 and 3 below.

See Dave's figma   7l53h5fxAUNjVRBQ82YFwn-fi-2579%3A61230 and reach out to @lucasmendes-design to give him a heads up with dev work starts on this.

![Image](https://user-images.githubusercontent.com/5665959/236757616-eeb72d8e-75b4-449e-bc45-bc4a6ee3b406.png)

-----

A few ideas for improving the sidenav have come up in the Colors & Fonts walkthrough call and in slack.
* pbxlJb-3EX-p2
*  p1680235303990409-slack-CRWCHQGUB

Current Colors & Fonts sidenav:

<img width="250" alt="image" src="https://user-images.githubusercontent.com/5665959/229028253-c2e721f7-c025-495c-b182-cd8040e035d7.png">

# Proposed Ideas

###  1. Spacing between Layout and Styles ✅ 
- [x]  Add subheading text, something similar to these examples
- [x] ✅ https://github.com/Automattic/wp-calypso/pull/75321
| . | . | . |
| --- | --- | --- |
|  ![Image](https://user-images.githubusercontent.com/5665959/229027373-43b058ad-712f-473e-add7-08125addc1c1.png) |   ![Image](https://user-images.githubusercontent.com/5665959/229027562-a87f3460-7f86-4fa6-9641-2289bed40597.png)  | . |

### - [ ] 2. Add premium label if user selects a premium color or font

![Image](https://user-images.githubusercontent.com/5665959/229027953-52a472e2-00a5-468e-8f37-ddbe73fc65ce.png)

### - [ ] 3. Add checkmark or similar indicator as user progresses through items

![Image](https://user-images.githubusercontent.com/5665959/229028016-b9e232b4-c95d-4e83-aa21-fa7de12489af.png)

# Design Tasks
- [ ] @davewhitley can you provide feedback and any designs as needed?
- [ ] 

# Dev Tasks
- [ ] TBD after Dave feedback

davewhitley:
This is sounding a lot like the launchpad in terms of "checklist" functionality and upgrade labels, so I'd like to borrow from that UI. I think it could work well.

[...]

PR #78313, opened 2023-06-16
Unify Design Preview: Support the font selection for the designs that don't have own style variations

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/78419

## Proposed Changes

* Introduce a feature flag, `signup/design-picker-preview-fonts`, to control this feature
* Display the font selection for the designs without style variations
* Apply the selected font variation to the preview
* Apply the selected font variation when you continue to activate the theme
* Display the “Unlock custom styles” modal when you select a font variation on your free site

**Follow-up tasks**
* The default font styles cannot be loaded as the Google API might not contain that font.
* Display another copy and modal to unlock the premium styles (colors & fonts)
* Add track events when selecting styles (colors & fonts)
* Support colors

image

https://github.com/Automattic/wp-calypso/assets/13596067/10171d65-9320-4d7b-913c-26fb385223f5

## Testing Instructions

* Preview the design without the style variation
* On the preview screen,
  * You have to see the font selection
  * The displayed fonts in the preview will be changed when you select any font
  * If the current site is a free site, the “Continue” button will become “Unlock theme” when you select any font.
  * Continue
* When you land on the launchpad, ensure the displayed fonts in the preview are the same as you select

## Pre-merge Checklist


arthur791004:
@lucasmendes-design You can preview the design with the font options by creating a new site and heading to the Design Picker from [Calypso Live link](https://github.com/Automattic/wp-calypso/pull/78313#issuecomment-1594452403). We will continue to split the free & premium fonts and update the unlock modals in the follow-up PRs 🙂

lucasmendes-design:
Hi @arthur791004, fast iteration on this one 🚀

Some notes: 

- Can we add the white background here as well? You can see the comparison below: 
image
image

- I noticed that you used "Unlock theme". But I think we can have the same behavior as SA. The "Continue" button and the modal. 
image

arthur791004:
> Can we add the white background here as well? You can see the comparison below:

Sure! Here is the result:

image

> I noticed that you used "Unlock theme". But I think we can have the same behavior as SA. The "Continue" button and the modal.

Yes, I'll address this one in the follow-up PRs 🙂

packages/design-preview/src/components/sidebar.tsx
```diff
@@ -115,6 +125,18 @@ const Sidebar: React.FC< SidebarProps > = ( {
 						</div>
 					</div>
 				) }
+				{ variations &&
```
arthur791004, line 128: The `variations` is `undefined` when we're still loading the theme details. So, we have to check it first to avoid displaying font pairings first and then changing to the style variations

miksansegundo (commented):
Code looks good 👍

### Two issues

I tested creating a new site and found these issues:

1. CSS overlapping issue with the "Unlock theme" button when scrolling
It happens when a font is selected or hovered so it has a border.

Screenshot 2566-06-21 at 10 40 34

2. Redirection to a random theme after checkout 
I chose the theme "Link in Bio: Mesh", chose a premium font, upgrade from the modal, used credits in the checkout, and the redirection took me back to the theme detail for another theme, seems random because in the first test was "Contact" and in a second test "Artfolio Detail".

https://github.com/Automattic/wp-calypso/assets/1881481/518b4675-5175-4ed2-9225-386521fa8a81


### About the follow-up tasks

> The default font styles cannot be loaded as the Google API might not contain that font.

Is this an issue? How this affects the user?

> Display another copy and modal to unlock the premium styles (colors & fonts)

For now, this PR makes the theme premium when choosing a premium font style. I think this is okay, not too much confusing. cc: @taipeicoder 

As a user:

**The theme I choose is Free**

See screenshot:

Screenshot 2566-06-21 at 10 27 54

**The font style I choose is Premium**

See screenshot:

Screenshot 2566-06-21 at 10 28 26

**The theme becomes Premium**
See screenshot:

Screenshot 2566-06-21 at 10 24 32

taipeicoder:
Feedback on my end:

- Virtual themes inherit the BC description, and it's odd. I would suggest removing it:
Screenshot 2023-06-21 at 2 33 38 PM

- There are Virtual themes that land in the Assembler, but this PR removes that behavior, such as "Image Portfolio" or "Coming Soon". I suggest we keep the original behavior and take those to the Assembler.

- The keyboard navigation for the font list also needs improvement. But that's also the case for the Assembler and can be handled later 🙂

taipeicoder:
Another issue: The custom style modal is no longer showing when selecting a non-default style variation for themes with style variations!

packages/design-preview/src/components/index.tsx
```diff
@@ -19,40 +26,68 @@ interface PreviewProps {
...
+	onGlobalStylesChange: ( globalStyles: GlobalStylesObject | null ) => void;
 }
 
 const INJECTED_CSS = `body{ transition: background-color 0.2s linear, color 0.2s linear; };`;
 
 const getVariationBySlug = ( variations: StyleVariation[], slug: string ) =>
 	variations.find( ( variation ) => variation.slug === slug );
 
-const Preview: React.FC< PreviewProps > = ( {
+// @todo Get the style variations of theme, and then combine the selected one with colors & fonts for consistency
```
fushar, line 41: @arthur791004 Can you explain more about this `@todo`? I'm interested (maybe add this in the follow-up task section in this PR description)

fushar:
Not sure if this has been discussed, but if the theme already requires a plan, should we not split free and premium fonts? As the premium fonts will already be included in the plan, and there is no way for the users to only be able to use the free font? Or this understanding is wrong? 😄 

image

packages/design-preview/src/components/index.tsx
```diff
@@ -19,40 +26,68 @@ interface PreviewProps {
...
 	previewUrl,
 	title,
 	author,
 	categories = [],
 	description,
 	shortDescription,
 	pricingBadge,
-	variations = [],
+	variations,
 	selectedVariation,
```
fushar, line 51: Nit: as we now have style and font variation, maybe rename this one to `selectedStyleVariation`? Or will than explode the diff of this PR? 😄

packages/design-preview/src/components/index.tsx
arthur791004: Currently, the technical paths between style variations and Colors & Fonts are different.
* For style variations, we get the inline CSS from the server side and inject it into the iframe for the preview. And then we get the config (settings, styles, etc) of the selected style variation by the slug when we want to activate the theme with that style.
* For Colors & Fonts, we get the config of the Colors & Fonts from the server side and transform them into CSS via [transformStyles](https://github.com/WordPress/gutenberg/blob/4d6900f7694d6c0d43b764466deac935daa6468e/packages/block-editor/src/utils/transform-styles/index.js#L20) to inject theme into the iframe for the preview. And then we can activate the theme with the config of the selected Colors & Fonts directly.

Hence, I suppose we can simply use the config of the style variations as well, and we don't need to get the data of the selected style variation from the server side when we want to activate the theme with that style. With this, we can align the technical path for both to simplify the codebase.

If we want to display Colors & Fonts for the themes with style variations in the future, it would be easier to do that as the implementation details are aligned.

packages/design-preview/src/components/index.tsx
arthur791004: I'll rename it in the follow-up PRs 👍

fushar: Sounds good!

okmttdhr (commented):
I got an overview of how Font Pairing works and where it's located. I need to look into and test it in more detail tomorrow, but non-blocking! Thank you for your PR!

packages/design-preview/src/components/index.tsx
fushar: I see, super clear, thanks for the explanation! 😄

packages/data-stores/src/site/actions.ts
```diff
@@ -402,6 +399,11 @@ export function createActions( clientCreds: WpcomClientCredentials ) {
 				);
 			}
 		}
+
+		if ( globalStyles ) {
+			yield setGlobalStyles( siteSlug, activatedTheme.stylesheet, globalStyles, activatedTheme );
```
fushar, line 404: Is it possible to merge `globalStyles` with `currentVariation`, so that we only call the endpoint in`setGlobalStyles()` once instead of twice? 

Or I guess it won't matter (for now) because it is not possible to select both style and font variations?

miksansegundo:
> Not sure if this has been discussed, but if the theme already requires a plan, should we not split free and premium fonts?

It has been discussed in https://github.com/Automattic/wp-calypso/issues/75156. We followed Lucas' suggestion to keep the same UI for both, but we still can add more clarity by adding copy under the "Premium fonts" and "Premium colors" labels to explain that the site plan includes the premium options or if it doesn't, they can be tried out for free. We still haven't created an issue for it.

arthur791004:
> Not sure if this has been discussed, but if the theme already requires a plan, should we not split free and premium fonts? As the premium fonts will already be included in the plan, and there is no way for the users to only be able to use the free font? Or this understanding is wrong? 😄

@ashar We discussed in https://github.com/Automattic/wp-calypso/issues/75156#issuecomment-1595030891 🙂

packages/data-stores/src/site/actions.ts
arthur791004: > Or I guess it won't matter (for now) because it is not possible to select both style and font variations?

Yep, you're right!

arthur791004:
> Virtual themes inherit the BC description, and it's odd. I would suggest removing it:
Screenshot 2023-06-21 at 2 33 38 PM

Sure!

> There are Virtual themes that land in the Assembler, but this PR removes that behavior, such as "Image Portfolio" or "Coming Soon". I suggest we keep the original behavior and take those to the Assembler.

I tested the themes you mentioned, and I saw they still took users to the Assembler. Could you help to confirm again? Thanks!

> The keyboard navigation for the font list also needs improvement. But that's also the case for the Assembler and can be handled later 🙂

Nice catching! I'll handle the keyboard navigation in the follow-up PRs 👍

arthur791004:
> CSS overlapping issue with the "Unlock theme" button when scrolling
It happens when a font is selected or hovered so it has a border.

Nice catch! I've updated styles 👍

image

> Redirection to a random theme after checkout
I chose the theme "Link in Bio: Mesh", chose a premium font, upgrade from the modal, used credits in the checkout, and the redirection took me back to the theme detail for another theme, seems random because in the first test was "Contact" and in a second test "Artfolio Detail".

Hmmm...the virtual themes use `blank-canvas` as the default theme, so it's the existing behavior. However, those themes are free without styles and users won't encounter this issue before... Let me think about how to fix it 🤔

> > The default font styles cannot be loaded as the Google API might not contain that font.
>
> Is this an issue? How this affects the user?

It made the preview of the default fonts looks weird and not the same as the displayed fonts on the large preview

taipeicoder:
> I tested the themes you mentioned, and I saw they still took users to the Assembler. Could you help to confirm again? Thanks!

I can't reproduce it on my end, either 🙂

arthur791004:
@miksansegundo I made a change to preserve the selected design and styles (style variation, font variation) via the URL, so users are able to see the correct virtual theme and styles when they redirect back to the design preview after the checkout. Besides, this change also allows people to keep previewing the design after they refresh the page!

taipeicoder:
> Another issue: The custom style modal is no longer showing when selecting a non-default style variation for themes with style variations!

@arthur791004 were you able to reproduce this?

arthur791004:
> Another issue: The custom style modal is no longer showing when selecting a non-default style variation for themes with style variations!
>
> @arthur791004 were you able to reproduce this?

Yes! I made the incorrect condition inside the `selectStyle` function and it leads the modal not showing. Fixed it via [2e63d60](https://github.com/Automattic/wp-calypso/pull/78313/commits/2e63d608ad56a747a94b6c480f254e6957d95d20)

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -604,10 +517,11 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow, stepName } ) => {
...
-			if (
-				shouldLimitGlobalStyles &&
-				! isDefaultGlobalStylesVariationSlug( selectedStyleVariation?.slug )
-			) {
+			const hasSelectedGlobalStyles =
+				! isDefaultGlobalStylesVariationSlug( selectedStyleVariation?.slug ) ||
+				!! selectedFontVariation;
+
+			if ( shouldLimitGlobalStyles && hasSelectedGlobalStyles ) {
 				unlockPremiumGlobalStyles();
```
fushar, line 527: I think we need to modify this function so that it will show the modal when a premium font is selected in a free site? Otherwise, the `Continue` button will do nothing and the flow cannot progress.

arthur791004: Oops...I only tested on style variations 🤦‍♂️

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/hooks/use-recipe.ts
```diff
@@ -0,0 +1,197 @@
...
+		stylesheet,
+		! preselectedFontVariationTitle
+	);
+
+	const handleSelectedDesignChange = ( design?: Design ) => {
+		setSelectedDesign( design );
+		setSearchParams( ( currentSearchParams ) => {
+			if ( design && design.is_virtual && design.recipe?.slug ) {
+				currentSearchParams.set( 'theme', design.recipe?.slug );
+				currentSearchParams.set( 'is_virtual', 'true' );
```
fushar: Where are we using / consuming this `is_virtual` query param?

arthur791004: I wanted to use it to determine which slug we have to use but finally I simply checked both as followed:
```ts
design.recipe?.slug === preselectedTheme || design.slug === preselectedTheme
```

I'll remove it since it's no longer used

packages/global-styles/src/hooks/use-font-pairing-variations.ts
```diff
@@ -2,7 +2,11 @@ import { useQuery } from '@tanstack/react-query';
 import wpcomRequest from 'wpcom-proxy-request';
 import type { GlobalStylesObject } from '../types';
 
-const useFontPairingVariations = ( siteId: number | string, stylesheet: string ) => {
+const useFontPairingVariations = (
+	siteId: number | string,
+	stylesheet: string,
+	disabled = false
```
fushar: As someone that doesn't have context for 5 months, reading the caller of this function is a bit confusing.

We need to be careful adding positional param to this as this lives in global packages, not just ours (onboarding / assembler steps).

What if we introduce the last param as options object? Something like this:

```
const useFontPairingVariations = (
	siteId: number | string,
	stylesheet: string,
	{ enabled = true }: { enabled: boolean }
)

...

enabled: !! siteId && !! stylesheet && enabled
```

then the caller is more readable:

```
const fontVariations = useFontPairingVariations(
		siteId,
		stylesheet,
		{ enabled: !!preselectedFontVariationTitle }
	);
```
```

arthur791004: Addressed via [cafe67d](https://github.com/Automattic/wp-calypso/pull/78313/commits/cafe67d17c171461b628fe0f86d046463a71dc1e)

fushar (approved):
Tested this multiple times. I think all functionalities are working.

Thanks for the refactor! Very good abstraction which leads to cleaner code. 😄

taipeicoder (approved):
Tested and working as intended, awesome PR!

There are some issues I found that we can address later on.

1. Disco has a flash of yellow background:
Screen Capture on 2023-06-27 at 17-26-08

2. Spearhead Blocks has some fonts that don't seem to change:
Screen Capture on 2023-06-27 at 17-25-22

arthur791004:
> There are some issues I found that we can address later on.
> 1. Disco has a flash of yellow background
> 2. Spearhead Blocks has some fonts that don't seem to change

Nice catch! I'll fix them in the follow-up PRs 🙏

PR #77406, opened 2023-05-25
Anchor cleanup: remove from stepper

andres-blanco:
## Proposed Changes
Removes all Anchor related code from stepper.

## Testing Instructions
* Look at the code
* Verify tests run properly
* Go through some stepper flows to verify it's working correctly

Closes https://github.com/Automattic/wp-calypso/issues/75387

PR #78315, opened 2023-06-16
Revert "Anchor cleanup: remove from stepper"

simison:
Reverts Automattic/wp-calypso#77406

Caused issues at `/setup/newsletter` flow:

- Color dropdown cut
- Font in description input
- On localhost also page isn't centered anymore

**Prod**

**Localhost**
image

**Reverting fixes:**

Screenshot 2023-06-16 at 13 41 36

### DIFF
--- a/client/landing/stepper/declarative-flow/anchor-fm-flow.ts
+++ b/client/landing/stepper/declarative-flow/anchor-fm-flow.ts
@@ -1,0 +1,117 @@
+import { isEnabled } from '@automattic/calypso-config';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { useI18n } from '@wordpress/react-i18n';
+import { SITE_STORE } from 'calypso/landing/stepper/stores';
+import { useSiteSlugParam } from '../hooks/use-site-slug-param';
+import { recordSubmitStep } from './internals/analytics/record-submit-step';
+import DesignSetup from './internals/steps-repository/design-setup';
+import ErrorStep from './internals/steps-repository/error-step';
+import { redirect } from './internals/steps-repository/import/util';
+import LoginStep from './internals/steps-repository/login';
+import PodcastTitleStep from './internals/steps-repository/podcast-title';
+import ProcessingStep from './internals/steps-repository/processing-step';
+import {
+	AssertConditionState,
+	type AssertConditionResult,
+	type Flow,
+	type ProvidedDependencies,
+} from './internals/types';
+import type { SiteSelect } from '@automattic/data-stores';
+
+export function isAnchorFmFlow() {
+	const sanitizePodcast = ( id: string ) => id.replace( /[^a-zA-Z0-9]/g, '' );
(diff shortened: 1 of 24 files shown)


javierarce:
The fix looks good to me, except for the button's color. If I'm not mistaken, or we have introduced a change recently, we always use `var(--studio-blue-50)` even if the input fields are empty.

simison:
> The fix looks good to me, except for the button's color. If I'm not mistaken, or we have introduced a change recently, we always use `var(--studio-blue-50)` even if the input fields are empty.

@javierarce thanks! That's how the design is currently implemented so I'll leave that to @crisbusquets to ponder.

javierarce:
> @javierarce thanks! That's how the design is currently implemented so I'll leave that to @crisbusquets to ponder.

Ok! 

I just noticed something else: the colors of the input placeholders are different. One is `studio-gray-30` and the other is `studio-gray-50`.

Screenshot 2023-06-16 at 13 35 27

simison:
Tested:

- `/setup/videopress`
- `/setup/newsletter`
- `/setup/free`
- `/setup/link-in-bio`

I'll deploy since this fixes bugs and Serenity team who deployed original is in different timezones.

crisbusquets:
> > The fix looks good to me, except for the button's color. If I'm not mistaken, or we have introduced a change recently, we always use `var(--studio-blue-50)` even if the input fields are empty.
> 
> @javierarce thanks! That's how the design is currently implemented so I'll leave that to @crisbusquets to ponder.

Hi!

This button is disabled until the user fills in the required inputs. It is `--studio-blue-50`, but with 50% opacity.
The launchpad has the same behavior (cc @nuriapenya to confirm).

Also, I didn't use `disabled` button style that uses `gray-5` ([link](eqlepL7MHifK8Bgr1OWqLr-fi-8758_29164)) because it's hard to read.

javierarce:
> This button is disabled until the user fills in the required inputs. It is --studio-blue-50, but with 50% opacity.
The launchpad has the same behavior (cc @nuriapenya to confirm).

Ok, but we should document it somewhere because in every other onboarding we don't disable the buttons that way. But that say, I think we should use the same solution on every form and not introduce a new divergence :/ 

> Also, I didn't use disabled button style that uses gray-5 ([link](https://github.com/Automattic/wp-calypso/pull/eqlepL7MHifK8Bgr1OWqLr-fi-8758_29164)) because it's hard to read.

Where does the link point to? 🤔 

And regarding the placeholder color I mentioned [here](https://github.com/Automattic/wp-calypso/pull/78315#issuecomment-1594539623), we should open a new issue.

crisbusquets:
@javierarce 

> Where does the link point to? 🤔

The link should point to this in Figma: `Dotcom Design Library > Buttons > Onboarding / Button > Disabled` 😅

> And regarding the placeholder color I mentioned https://github.com/Automattic/wp-calypso/pull/78315#issuecomment-1594539623, we should open a new issue.

Done: https://github.com/Automattic/wp-calypso/issues/78336

PR #78317, opened 2023-06-16
Plans: Remove dependency on HappychatButton in HappinessSupport

klimeryk:
Switch to using SupportButton instead which opens Help Center.

## Proposed Changes

* Update `HappinessSupport` component to not depend on `HappychatButton`. Instead, it should depend on Help Center via`SupportButton`.
* This also simplifies the setup - the clients should not check themselves if chat is available, if the chat option should be enabled, etc.

## Testing Instructions

Test in `/plans/my-plan`:

Screenshot 2023-06-16 at 10 47 24

Verify the links work as expected:
 - Ask a question should open Help Center with the contact options.
 - Support documentation should open a new tab with support docs.

Test in checkout thank-you page (it shows up at the very end):
Screenshot 2023-06-16 at 10 48 19

Same as above.

Make sure you also test with a Jetpack plan/purchase. In which case, the Ask a question link should go to `https://jetpack.com/contact-support/?rel=support`. And the support docs one to `https://jetpack.com/support/`.

### DIFF
--- a/client/components/support-button/index.tsx
+++ b/client/components/support-button/index.tsx
@@ -1,6 +1,7 @@
 import { recordTracksEvent } from '@automattic/calypso-analytics';
 import { Button } from '@automattic/components';
 import { HelpCenter } from '@automattic/data-stores';
+import { useStillNeedHelpURL } from '@automattic/help-center/src/hooks';
 import { useDispatch as useDataStoreDispatch } from '@wordpress/data';
 import { useI18n } from '@wordpress/react-i18n';
 import type { FC } from 'react';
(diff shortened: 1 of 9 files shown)


client/components/support-button/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+	isLink = true,
+	children,
+	skipToContactOptions = false,
+} ) => {
+	const { __ } = useI18n();
+	const { setInitialRoute, setShowHelpCenter } = useDataStoreDispatch( HELP_CENTER_STORE );
+
+	function handleClick() {
+		if ( skipToContactOptions ) {
+			setInitialRoute( '/contact-options' );
```
klimeryk: TODO: use the `useStillNeedHelpURL` hook instead of hardcoding the path to avoid the same issue that was fixed in https://github.com/Automattic/wp-calypso/pull/78337

atrumgeost:
I ran a test on a WordPress.com and a Jetpack site. It works as expected and it is a better experience. LGTM!

client/components/support-button/index.tsx
klimeryk: Done: https://github.com/Automattic/wp-calypso/pull/78317/commits/3971b142b6a3c51b00639ef96b250a7cccf3a376

PR #78323, opened 2023-06-16
Stepper: Extract the login URL constructing code into an utility function

southp:
Related to Automattic/growth-foundations#70 , which is a refactor that will make the subsequent works easier to maintain.

## Proposed Changes

This PR extracts the repetitive login URL constructing code into an utility function so. Also, it applies `useLocalizeUrl()` hook to alter the URL according to the locale in a more standardized way(see https://github.com/Automattic/wp-calypso/blob/trunk/packages/wpcom-template-parts/src/universal-header-navigation/index.tsx#L24 for an example usage).

The PR chooses the Newsletter flow as the 1st use case. Once it is approved, the same will be propagated into all the other tailored flows.

## Testing Instructions

The Newsletter flow should work as expected.

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/utils/path.ts
+++ b/client/landing/stepper/utils/path.ts
@@ -44,23 +44,28 @@ export function useLangRouteParam() {
 	return match?.params.lang;
 }
 
-export const useLoginUrl = ( {
-	flowName,
-	redirectTo,
-	pageTitle,
-}: {
-	flowName: string;
-	redirectTo: string;
-	pageTitle: string;
+export const useLoginUrl = ( params: {
+	flowName?: string;
+	redirectTo?: string;
+	pageTitle?: string;
 } ): string => {
 	const locale = useLocale();
 
 	const loginPath =
 		locale && locale !== 'en' ? `/start/account/user/${ locale }` : `/start/account/user`;
 
-	return addQueryArgs( loginPath, {
-		redirect_to: redirectTo,
-		variationName: flowName,
-		pageTitle: pageTitle,
-	} );
+	const nonEmptyQueryParameters = Object.entries( params )
+		.filter( ( [ , value ] ) => value )
+		.map( ( [ key, value ] ) => {
+			switch ( key ) {
+				case 'redirectTo':
+					return [ 'redirect_to', value ];
+				case 'flowName':
+					return [ 'variationName', value ];
+				default:
+					return [ key, value ];
+			}
+		} );
+
+	return addQueryArgs( loginPath, Object.fromEntries( nonEmptyQueryParameters ) );
 };
--- a/client/landing/stepper/utils/path.ts
+++ b/client/landing/stepper/utils/path.ts
@@ -1,5 +1,6 @@
 import { Plans } from '@automattic/data-stores';
 import languages from '@automattic/languages';
+import { addQueryArgs } from '@wordpress/url';
 import { useMatch } from 'react-router-dom';
 
 const plansPaths = Plans.plansSlugs;
@@ -41,3 +42,16 @@ export function useLangRouteParam() {
 	const match = useMatch( path );
 	return match?.params.lang;
 }
+
+export const getLoginPath = (
+	redirectTo: string,
+	pageTitle = '',
+	variationName = '',
+	loginPath = '/start/account/user'
+): string => {
+	return addQueryArgs( loginPath, {
+		redirect_to: redirectTo,
+		variationName,
+		pageTitle,
+	} );
+};
--- a/client/landing/stepper/utils/path.ts
+++ b/client/landing/stepper/utils/path.ts
@@ -1,4 +1,5 @@
 import { Plans } from '@automattic/data-stores';
+import { useLocale } from '@automattic/i18n-utils';
 import languages from '@automattic/languages';
 import { addQueryArgs } from '@wordpress/url';
 import { useMatch } from 'react-router-dom';
@@ -43,15 +44,23 @@ export function useLangRouteParam() {
 	return match?.params.lang;
 }
 
-export const getLoginPath = (
-	redirectTo: string,
-	pageTitle = '',
-	variationName = '',
-	loginPath = '/start/account/user'
-): string => {
+export const useLoginUrl = ( {
+	flowName,
+	redirectTo,
+	pageTitle,
+}: {
+	flowName: string;
+	redirectTo: string;
+	pageTitle: string;
+} ): string => {
+	const locale = useLocale();
+
+	const loginPath =
+		locale && locale !== 'en' ? `/start/account/user/${ locale }` : `/start/account/user`;
+
 	return addQueryArgs( loginPath, {
 		redirect_to: redirectTo,
-		variationName,
-		pageTitle,
+		variationName: flowName,
+		pageTitle: pageTitle,
 	} );
 };
--- a/client/landing/stepper/declarative-flow/newsletter.ts
+++ b/client/landing/stepper/declarative-flow/newsletter.ts
@@ -1,4 +1,3 @@
-import { useLocale, useLocalizeUrl } from '@automattic/i18n-utils';
 import { useFlowProgress, NEWSLETTER_FLOW } from '@automattic/onboarding';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { addQueryArgs } from '@wordpress/url';
@@ -14,7 +13,7 @@ import {
 } from 'calypso/signup/storageUtils';
 import { useSiteSlug } from '../hooks/use-site-slug';
 import { ONBOARD_STORE, USER_STORE } from '../stores';
-import { getLoginPath } from '../utils/path';
+import { useLoginUrl } from '../utils/path';
 import { recordSubmitStep } from './internals/analytics/record-submit-step';
 import { ProvidedDependencies } from './internals/types';
 import type { Flow } from './internals/types';
@@ -85,14 +84,11 @@ const newsletter: Flow = {
 			flowName,
 		} );
 		setStepProgress( flowProgress );
-		const locale = useLocale();
-		const localizeUrl = useLocalizeUrl();
-
-		const logInUrl = localizeUrl(
-			getLoginPath( `/setup/${ flowName }/newsletterSetup`, 'Newsletter', flowName ),
-			locale,
-			userIsLoggedIn
-		);
+		const logInUrl = useLoginUrl( {
+			flowName,
+			redirectTo: `/setup/${ flowName }/newsletterSetup`,
+			pageTitle: 'Newsletter',
+		} );
 
 		// Unless showing intro step, send non-logged-in users to account screen.
 		if ( ! isLoadingIntroScreen && ! userIsLoggedIn ) {
--- a/client/landing/stepper/declarative-flow/newsletter.ts
+++ b/client/landing/stepper/declarative-flow/newsletter.ts
@@ -1,4 +1,4 @@
-import { useLocale } from '@automattic/i18n-utils';
+import { useLocale, useLocalizeUrl } from '@automattic/i18n-utils';
 import { useFlowProgress, NEWSLETTER_FLOW } from '@automattic/onboarding';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { addQueryArgs } from '@wordpress/url';
@@ -14,6 +14,7 @@ import {
 } from 'calypso/signup/storageUtils';
 import { useSiteSlug } from '../hooks/use-site-slug';
 import { ONBOARD_STORE, USER_STORE } from '../stores';
+import { getLoginPath } from '../utils/path';
 import { recordSubmitStep } from './internals/analytics/record-submit-step';
 import { ProvidedDependencies } from './internals/types';
 import type { Flow } from './internals/types';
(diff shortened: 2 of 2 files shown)


client/landing/stepper/utils/path.ts
```diff
@@ -41,3 +42,16 @@ export function useLangRouteParam() {
 	const match = useMatch( path );
 	return match?.params.lang;
 }
+
+export const getLoginPath = (
```
chriskmnds: We could turn/rename this into a hook instead (`useLoginPath`) if it's only used in React functions.

chriskmnds (commented):
Nice refactor. I tested with `en` locale and redirection works. 

However, not sure how to test the `localizeUrl` part properly. I tried forcing [locale](https://github.com/Automattic/wp-calypso/pull/78323/files#diff-8770a92d677c364e1854ef215b82a7c1b4ff0bd155ddf4b22f1e81b504f82ef7R93) to `es` and I don't reach the translated page i.e. path not changing to `/start/account/user/es`. I wonder if this actually only works with full URL (not path) e.g. it does log "invalid" [here](https://github.com/Automattic/wp-calypso/blob/trunk/packages/i18n-utils/src/localize-url.tsx#L199) if I put a logger.

client/landing/stepper/utils/path.ts
ddc22: Is there an added advantage (perhaps in optimisation?) to using hooks instead of a simple function, in instances like this?

skim1220:
👋 @southp & @chriskmnds,

I am not entirely sure what this PR aims to do, but the title makes me think that it is related to how we will redirect users from the landing page to the account creation step (i.e. “Let’s get started” page) which is the first step of signup flow.  

So, I focused on **_reviewing what type of data is being collected as we move from the landing page to “Let’s get started” page_**. I particularly focused on how the following three Tracks events are logged on “Let’s get started” page as they are the most important events in our signup analysis. 

• `calypso_signup_start`
• `calypso_signup_step_start`
• `calypso_page_view`

I compared the records from the test flow in this PR to those from LOHP,  `/website-builder` , `/create-blog` . The details are in [this Google Sheet](https://docs.google.com/spreadsheets/d/1QbKXM5ZqfxT1miOvMbJLHyX73c3YnCFCRQ2IoZrkw9E/edit?usp=sharing), and please find some observations/questions below. 

(1) It seems that the flow in this PR is created as `account` flow.
- I was able to confirm that the url changes from `setup/newsletter` on the landing page to `/start/account/user` on the “Let’s get started" page. 
- I was also able to confirm that the specific event properties that we are interested in with three Tracks events have the same values as seen in the `account` flow initiated from LOHP,  `/website-builder` ,  `/create-blog`.
 
(2) I am probably missing some of prior discussion and details, but is it possible to implement “Let’s get started” page in our new signup flow as `onboarding` flow and `/start/user` step? Just like how it is currently done for LOHP and  `/website-builder`? 

Please refer to Row 8, Row 9 records in [the Google Sheet](https://docs.google.com/spreadsheets/d/1QbKXM5ZqfxT1miOvMbJLHyX73c3YnCFCRQ2IoZrkw9E/edit?usp=sharing), and compare them to Row 7 (Newsletter test flow in this PR) records.

  I am asking this because 
- This matches how it is done and what is being recorded in the current onboarding flow. 
- I find that ref property of calypso_signup_start event very helpful to see the source landing page of signup start. This property is not available when calypso_signup_start is fired from account flow for some reason. 

(3) I noticed that signup start events (`calypso_signup_start`, `calypso_singup_step_start`) are fired on the landing page in this test flow (see the red-highlighted row in [the Google Sheet](https://docs.google.com/spreadsheets/d/1QbKXM5ZqfxT1miOvMbJLHyX73c3YnCFCRQ2IoZrkw9E/edit?usp=sharing)). AFAIK, _**this should not happen**_. These Tracks events should be triggered when users land on “Let’s get started” page, the first step of signup process. 

(4) I noticed that “Get Started” button on the `/create-blog` page directs users to the `onboarding` flow whereas the other CTA buttons redirect them to the `account` flow. Is this expected? 

Let me know if anything is unclear, happy to elaborate further.

client/landing/stepper/utils/path.ts
ddc22: I had a deeper look at this and I agree with @chriskmnds  💯 
If we convert this to a hook, we can possibly encapsulate the entire localisation logic inside the new hook by calling the localisation related hooks inside it.

https://github.com/Automattic/wp-calypso/blob/5eddb0b777ee6594b00fddd4e6626afd350dce07/client/landing/stepper/declarative-flow/newsletter.ts#L88-L95

ddc22 (approved):
The flow works well. I am not sure exactly how the localize url works though. I tried to test it with different locales but got the same result. But that shouldn't be related to this change.

I have left a comment below but shouldn't be a blocker.

LGTM 👍

chriskmnds:
> The flow works well. I am not sure exactly how the localize url works though. I tried to test it with different locales but got the same result. But that shouldn't be related to this change.

@jdc91 I'm confused. So this _doesn't_ work is what you are saying? We need ensure localisation works before shipping, that's the main change happening here.

client/landing/stepper/utils/path.ts
chriskmnds: Hooks for everything :-)

southp:
Hey @skim1220 , thanks for taking a look.
This PR is merely a refactor, which means there shouldn't be any functional-wise change. However, your investigation and note is very helpful for our follow-up works, i.e. 

* https://github.com/Automattic/wp-calypso/pull/78333 for completing the signup completion event.
* https://github.com/Automattic/wp-calypso/pull/77801 for the paid media flow

Before going into your questions, I'd like to clarify some terminology –– we devs obviously don't do a very good job on naming here, so please bear with me :)  
First of all, `/setup/newsletter` is not a landing page even though it looks like one :) Everything under `/setup/` path is a sign-up flow powered by the Stepper framework. Thus, that "Launch your Newsletter" page you saw is actually the first step of the Newsletter flow. Landing pages like `/create-blog` and `/website-builder` are different: they are not part of calypso, but WordPress pages that live on WordPress.com. LOHP(logged-out homepage) is powered by yet another tech stack, but that's a different story. 

Now back to your question:

> I am probably missing some of prior discussion and details, but is it possible to implement “Let’s get started” page in our new signup flow as onboarding flow and /start/user step? Just like how it is currently done for LOHP and /website-builder?

The short answer is yes :)

Our target paid media flow(WIP in https://github.com/Automattic/wp-calypso/pull/77801) will be directly linked by the paid media landing pages documented in Automattic/growth-foundations#63, which means the new paid media flow will work just like the main onboarding flow (`/start/`, which the 1st step is `/start/user`). 


> I noticed that signup start events (calypso_signup_start, calypso_singup_step_start) are fired on the landing page in this test flow (see the red-highlighted row in [the Google Sheet](https://docs.google.com/spreadsheets/d/1QbKXM5ZqfxT1miOvMbJLHyX73c3YnCFCRQ2IoZrkw9E/edit?usp=sharing)). AFAIK, this should not happen. These Tracks events should be triggered when users land on “Let’s get started” page, the first step of signup process.

As explained above, that "Launch your Newsletter" page is actually the first step of the flow, so it's actually expected behavior. However, it won't happen on the new paid media onboarding flow, so there shouldn't be anything to worry :)

> I noticed that “Get Started” button on the /create-blog page directs users to the onboarding flow whereas the other CTA buttons redirect them to the account flow. Is this expected?

I think it's opposite? ;) And, the answer is yes.
 `/create-blog` page links to `/setup/blog` flow, which the first step will be `/start/account/user`, while `website-builder` links to the main onboarding flow(`/start/`)

Does it make sense to you? Please feel free to let me know if there is any other question :)

ddc22:
@chriskmnds you are correct 👍  TIL that the localizeUrl function does not work for relative paths. And fails silently 😬 as you pointed out. 

https://github.com/Automattic/wp-calypso/blob/ad3b4ad130802a57cdfc6094302c753234f5334b/packages/i18n-utils/src/localize-url.tsx#L197-L200
image

---

@southp looking at the existing usage seems like we have to provide the full absolute URL for localize url to work.

image

On a side note, Should we maybe fail loudly  ( in dev at least ) when this happens ?

southp:
Wow. I'm surprised. Thanks for the catch, @jdc91 , @chriskmnds . I thought it worked when I first introduced it because the user step was localized for me. However, after rechecking, it was localized not because of `localizeUrl()`, but because some i18n code work in the background. I'm feeling we'd be digging another rabbit hole if we continue this way.

@jdc91 Could you help implement the idea of [the login path hook above](https://github.com/Automattic/wp-calypso/pull/78323#discussion_r1236010395) and revert the localizing code into the old way? That way, we won't touch the localization issue here and can move on, and will have a consolidated place to address it separately in the near future.

ddc22:
@chriskmnds @southp 
After studying this a bit further seems like we have a specific set of hosts that we localize.

https://github.com/Automattic/wp-calypso/blob/ad3b4ad130802a57cdfc6094302c753234f5334b/packages/i18n-utils/src/localize-url.tsx#L114-L173

And the host of the url that we want to localize is not absolute its dynamic and changes based on the environment. It should work fluidly within the stepper framework even when running on dev, live, staging or production. But the host of the login url does change in each of the environments ( calypso.localhost:3000, calypso.live, wpcalypso.wordpress.com ). We cannot use an absolute url ( wordpress.com/setup ... ) as well because when we do this the stepper framework context will be broken when testing the flow in dev environments.

I am still not sure how exactly to solve for this 🤔

ddc22:
Thanks for the feedback @southp I have gone ahead and reverted to the previous approach and done the hook refactor.

@chriskmnds can you have another look, the localization should work now 👍

skim1220:
👋 @southp,

> which means the new paid media flow will work just like the main onboarding flow (`/start/`, which the 1st step is `/start/user`).

Perfect, thanks for all the clarfications 🙏 

> I think it's opposite? ;) And, the answer is yes.
> /create-blog page links to /setup/blog flow, which the first step will be /start/account/user, while website-builder links to the main onboarding flow(/start/)

Oh, I meant two different CTA buttons on `/create-blog` page. [This recording](https://d.pr/v/s4BPXy) should help us communicate. 

- When I click "Start a blog", it redirects me to `/setup/blog` and `calypso_signup_start` event has `flow` = `account`. 
- When I click "Get Started", it sends me to `/start/user` and `calypso_signup_start` event has `flow` = `onboarding`. 

Is this expected? If so, I wonder why we send users on this page to two different flows? 🤔

southp:
@skim1220 Re:

> Oh, I meant two different CTA buttons on /create-blog page. [This recording](https://d.pr/v/s4BPXy) should help us communicate.
> When I click "Start a blog", it redirects me to /setup/blog and calypso_signup_start event has flow = account.
> When I click "Get Started", it sends me to /start/user and calypso_signup_start event has flow = onboarding.
> Is this expected? If so, I wonder why we send users on this page to two different flows? 🤔

Ohhh, thanks for the recording. It surely helps :) Yes, this is an expected behavior. Currently that Get Started is from a shared "bar" component which doesn't change its behvior according to the page context it is displayed. I believe there was a discussion about making it overridable, but there wasn't clear conclusion about it and we also didn't have capacity to address it. @niranjan-uma-shankar Do you recall anything about it?

southp:
Thanks for implementing the hook, @jdc91 . Regarding localization, in my case it still doesn't work:

https://github.com/Automattic/wp-calypso/assets/1842898/24e7d024-cb23-40d2-a43d-9d9189d1f2e2

However, that's expected and exactly why I proposed to just revert it to the old way, since it should be handled separately. I've documented it in https://github.com/Automattic/dotcom-forge/issues/3547 and will go ahead and deploy this PR.

client/landing/stepper/utils/path.ts
```diff
@@ -41,3 +43,29 @@ export function useLangRouteParam() {
...
+		.map( ( [ key, value ] ) => {
+			switch ( key ) {
+				case 'redirectTo':
+					return [ 'redirect_to', value ];
+				case 'flowName':
+					return [ 'variationName', value ];
+				default:
+					return [ key, value ];
+			}
+		} );
```
chriskmnds, line 68: Did we need all of this complexity? It's not even clear to me what it's trying to do and whether even necessary

with filtering and mapping... something we couldn't spread out instead?

ddc22: This code tries to 
1 - filter - ignore parameters that are empty
2 - map transforms existing variable names to query params for the login escape path

And then generates a brand new object from it to feed to the `addQueryArgs` function.

Happy to go with a simpler method 👍 . Is there any particular approach you have in mind?

chriskmnds: Spread them?

ddc22: Is there a better way to implement this entire logic with only the spread operator? 🤯. How do I do that 🤔 ?

jeyip: Okay now I'm curious too. I'm preparing to 🤯 from Christos's secret magic

chriskmnds: This logic looks very primitive, pre-ES6 stuff. I wonder if we have these things laying around in Calypso somewhere. Is it something that can't be done with simple conditional spreading?

```
{
    ...( flowName && { variationName: flowName } ),
    ...( redirectTo && { redirect_to: redirectTo } ),
    ...( pageTitle && { pageTitle } ),
}
```

chriskmnds: > Is it something that can't be done with simple conditional spreading?

I haven't looked much at the code beyond the signature, so this would be my immediate "reach" for something like this. that we resolve to logic that pertains to `convert to array -> filter -> map -> switch -> convert back to object` ought to be quite intentional, so probably missing something in the rest of it 🤔

ddc22: Thanks @chriskmnds that looks definitely much better and cleaner than my approach 🙌 . The `...( flowName && ` part did not immediately resonate in my brain but is a fantastic use of spread.  

> ought to be quite intentional

It wasn't that intentional the other approach on my mind was to use if conditions check for variables and then add them to an empty object. I just felt this looked much cleaner.

jeyip: Very cool! Thanks for sharing @chriskmnds 🙏  TIL spreading falsey values is equivalent to spreading an empty object

chriskmnds: Can you submit a PR to update this, please? It's a little superfluous and imperative pattern that shouldn't be setting a precedent. :/

PR #78325, opened 2023-06-16
Reader: Fix YouTube autoplay not triggering when an autoplay iframe is inserted as a response to a thumbnail click

atanas-dev:
Related to p1HpG7-lfW-p2#comment-64019

Clicking on a featured video thumbnail in reader injects an autoplay iframe of the video but at some point YouTube has started requiring that `allow="autoplay"` is present on the iframe for autoplay to work. This means that for YouTube videos, 2 clicks are required in order to start playing a video (see the before and after at the bottom).

## Proposed Changes

* Reader: Fix YouTube autoplay not triggering when an autoplay iframe is inserted as a response to a thumbnail click.

## Testing Instructions

- Follow this blog so it shows up in your reader: https://atanasvideofeedblog.wordpress.com/
- Open up the blog's feed before checking out this patch: http://calypso.localhost:3000/read/feeds/143404674
- Click on the first YouTube post's featured video thumbnail with a play button - it should be replaced with a YouTube embed but that embed will require a second click for it to start playing.
- Checkout this PR's branch and refresh - the second click should no longer be required in order to play the video.

Before:

https://github.com/Automattic/wp-calypso/assets/22746396/abe09f37-e16e-4e58-b4fc-cc1175fc11a2

After:

https://github.com/Automattic/wp-calypso/assets/22746396/f89f3ad4-2566-4bce-854a-491fc0a45270

## Pre-merge Checklist

### DIFF
--- a/client/lib/post-normalizer/rule-content-detect-media.js
+++ b/client/lib/post-normalizer/rule-content-detect-media.js
@@ -78,6 +78,12 @@ const getAutoplayIframe = ( iframe ) => {
 		} else {
 			autoplayIframe.src += '&autoplay=1';
 		}
+
+		// ?autoplay=1 is no longer sufficient for YouTube - we also need to add autoplay to the allow attribute.
+		const allow = ( autoplayIframe.allow || '' ).split( /\s*;\s*/g );
+		allow.push( 'autoplay' );
+		autoplayIframe.setAttribute( 'allow', allow.filter( ( s ) => s.length > 0 ).join( '; ' ) );
+
 		return autoplayIframe.outerHTML;
 	}
 	return null;


atanas-dev:

@tyxla this doesn't sound right - any insight you could share here? I've never dealt with the ETK before.

client/lib/post-normalizer/rule-content-detect-media.js
```diff
@@ -78,6 +78,12 @@ const getAutoplayIframe = ( iframe ) => {
 		} else {
 			autoplayIframe.src += '&autoplay=1';
 		}
+
+		// ?autoplay=1 is no longer sufficient for YouTube - we also need to add autoplay to the allow attribute.
+		const allow = ( autoplayIframe.allow || '' ).split( /\s*;\s*/g );
+		allow.push( 'autoplay' );
+		autoplayIframe.setAttribute( 'allow', allow.filter( ( s ) => s.length > 0 ).join( '; ' ) );
```
tyxla, line 85: Do we just always want to autoplay? Should this be something that depends on how the video is added to the content (with or without autoplay)?

tyxla (approved):
Code-wise this looks good, so I'm approving 👍 

I just wanted to suggest reconsidering always auto-playing all YouTube videos. It's generally considered against the best accessibility practices (see [this article](https://www.a11yproject.com/posts/never-use-auto-play/)) and not something all users would be happy about. Perhaps this could be good material for a design discussion.

tyxla:
> > This PR modifies the release build for editing-toolkit
> 
> @tyxla this doesn't sound right - any insight you could share here? I've never dealt with the ETK before.

There are a bunch of imports from Calypso's client inside ETK, so it's possible that the file you're touching gets included - I haven't checked, but I expect it likely happens through Calypso's state. Since the change is targeted at the reader, I think you can disregard the message.

atanas-dev:
> I just wanted to suggest reconsidering always auto-playing all YouTube videos.

Just to be clear, the goal of this particular iframe is to autoplay and all other video embeds we have in the same card preview component do already autoplay - the thing is that we don't render the iframe immediately. We render a thumbnail which when clicked renders the autoplay iframe so in reality it's not autoplaying when you load the page, we're just working around making the video play the moment you click on its thumbnail (rather than having to click twice, which is something you don't need to do for VideoPress or Vimeo iframes, for example).

PR #78342, opened 2023-06-16
Hosting: Add is_in_hosting property to calypso_signup_step_start event in stepper

zaguiini:
## Proposed Changes

For some reason, we have only added the `is_in_hosting_flow` property to `/start/hosting`. On `/setup/(new|import)-hosted-site` flows that property was never passed.

## Testing Instructions

Check that `calypso_signup_step_start` is being dispatched with `is_in_hosting_flow` as a boolean depending whether the `hosting-flow` query parameter is present.


vykes-mac (approved):
g2g

PR #78383, opened 2023-06-19
Theme showcase: Consider filters when searching WP.org themes

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2763

## Proposed Changes

Add the entered filters in the theme showcase to the search term used to query WP.org themes.

Screenshot 2023-06-19 at 13 22 31

We cannot use the filters as tags in the query because the WP.com theme filters don't match the WP.org ones, so that's why we add them to the search term as a fallback.

## Testing Instructions

- Go to Appearance > Themes.
- Open the Network tab of your browser devtools and observe the calls to `api.wordpress.org/themes`.
- Enter a search term.
- Select some filters.
- Make sure the filters have been appended to the `request[search]` param of the API call.

### DIFF
--- a/client/my-sites/themes/themes-selection.jsx
+++ b/client/my-sites/themes/themes-selection.jsx
@@ -343,7 +343,13 @@ export const ConnectedThemesSelection = connect(
 		const themes = getThemesForQueryIgnoringPage( state, sourceSiteId, query ) || [];
 
 		const shouldFetchWpOrgThemes = forceWpOrgSearch && sourceSiteId !== 'wporg' && !! search; // Only fetch WP.org themes when searching a term.
-		const wpOrgQuery = { ...query, page: 1 }; // We limit the WP.org themes to one page only.
+		const wpOrgQuery = {
+			...query,
+			// We limit the WP.org themes to one page only.
+			page: 1,
+			// WP.com theme filters don't match WP.org ones, so we add them to the search term.
+			search: filter ? search + ' ' + filter.replaceAll( '+', ' ' ).replaceAll( '-', ' ' ) : search,
+		};
 		const wpOrgThemes = shouldFetchWpOrgThemes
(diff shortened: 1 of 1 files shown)


client/my-sites/themes/themes-selection.jsx
```diff
@@ -343,7 +343,13 @@ export const ConnectedThemesSelection = connect(
 		const themes = getThemesForQueryIgnoringPage( state, sourceSiteId, query ) || [];
 
 		const shouldFetchWpOrgThemes = forceWpOrgSearch && sourceSiteId !== 'wporg' && !! search; // Only fetch WP.org themes when searching a term.
-		const wpOrgQuery = { ...query, page: 1 }; // We limit the WP.org themes to one page only.
+		const wpOrgQuery = {
+			...query,
+			// We limit the WP.org themes to one page only.
+			page: 1,
+			// WP.com theme filters don't match WP.org ones, so we add them to the search term.
+			search: filter ? search + ' ' + filter.replaceAll( '+', ' ' ).replaceAll( '-', ' ' ) : search,
```
dsas: You could do both replaces in one swoop by passing a regex like `/[+-]/g` rather than the symbol. Maybe not as obvious though.

dsas (approved):

client/my-sites/themes/themes-selection.jsx
rcrdortiz: I think a regex would be more readable, and it's also a bit more efficient than searching the string twice.
```suggestion
			search: filter ? `${search} ${filter.replace(/[+-]/g, ' ')}` : search,
```

mmtr: Oh, regex, my old nemesis, we meet again! But I agree it looks more readable, so I changed it. Thanks!

PR #78386, opened 2023-06-19
Theme showcase: Only show WP.org themes when the tier is not set

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2764

## Proposed Changes

Excludes the WP.org themes from the showcase when a tier is set, since they are not good for any of the tiers:
- They aren't "Free" as they need a Business plan.
- They aren't "Premium" as they are not included in the Premium plan.
- They aren't "Paid" as users don’t need a separate payment to get them.

Screenshot 2023-06-19 at 13 40 47

## Testing Instructions

- Go to Appearance > Themes.
- Select either the "Free", "Premium", or "Paid" tier.
- Enter a search term.
- Make sure WP.org themes don't show up.


dsas (approved):
wfm!

PR #73904, opened 2023-02-28
Launchpad: Add Next Steps button at the end of domain flow

sharpirate:
Related to https://github.com/Automattic/wp-calypso/issues/73869

#### Time Estimate
Review: Short
Test: Mid

#### Proposed Changes
Add a new "Launch your site" section at the end of the domain purchase flow to allow the user to go back to Launchpad

image

#### Testing
1. Checkout this branch
2. Create a new general onboarding site http://calypso.localhost:3000/start
3. Choose a free domain and free plan
4. Once you arrive at Launchpad, click the Choose a domain task and go through the flow to upgrade your domain
5. Once you purchase your domain and arrive at the Next Steps screen you should see the new "Launch your site" section
6. The Next Steps buttons should redirect you back to Launchpad

PR #56949, opened 2021-10-12
Emails: Thank you page for Professional Email when buying domain

AllTerrainDeveloper:
#### Changes proposed in this Pull Request

This Pull Request adds the necessary code to show Thank You component after checking out a Domain with a Professional Email.

#### Testing instructions

**Test 1**

1. Go to Upgrades -> Domains
2. Click on the button Add a domain for this site -> Search for a domain
3. When asked to add an Email, fill in necessary fields and click "Add"
4. Buy with credits

**Expected result**
You should see a Thank you page, congratulating the purchase of the domain with CTA's to go to Inbox.
image

**Test 2**

1. Go to Upgrades -> Domains
2. Click on the button Add a domain for this site -> Search for a domain
3. When asked to add an Email, press the SKIP button
4. Buy with credits

**Expected result**
You should see a Thank you page, congratulating the purchase of the domain with CTA's to add professional email.
image

**Test 3**

0. Have an external domain on a domain provider different than WPCOM.
1. Go to Upgrades -> Domains
2. Click on the button Add a domain for this site -> Use a domain I own
3. Buy with credits

**REMARKS**
If you are not allowed to map domains to the site, you need to sandbox and change the function: `function is_exempt_from_domains_with_plans_only( $current_user = false )` to return always true

**Expected result**
You should see a Thank you page, congratulating the purchase of the domain mapping with CTA's to add professional email.
image

**Test 4, regression test**
1. Go to Upgrades -> domains
2. Click on the button Add a domain for this site -> Use a domain I own
3. Type the domain you own and start a transfer domain
4. But with credits

**Expected result**
You should see a thank you page, congratulating about the domain being transferred like this:
image

**Test 5**
Repeat those tests where you have bought Professional Email, but this time buy a Google Workspace, then the thank you page shouldn't have the step to go to inbox.

Example:
image

**REMARKS**
Thank you component shown in Test 1 & 2 will replace the old one below
It will replace this thank you page:
image

Related to {1200182182542585-as-1201102838390955}

PR #78394, opened 2023-06-19
Domains Thank You: move hooks from `domainRegistrationThankYouProps` to parent

### LINKED ISSUE #78378: Purchasing domain on Woo Express, ends up on white screen of death on thank you page
### Quick summary

At the final step, (thank you - https://wordpress.com/checkout/thank-you/woo-xxxxxxxxxxxxx.wpcomstaging.com/81332962) the order completes successfully, then there is a flash of content, an error appears, in the console and the page turns white.

The file that is causing the issue is https://wordpress.com/calypso/evergreen/87013.fd58d19601e1b073201b.min.js - I deactivated page optimize, the issue persists

### Steps to reproduce

1. Create a WooCommerce Free trial site
2. Purchase a domain
3. On the thank-you page, there is a flash of content, white page and an error in console
4. The domain purchase finishes successfully, no problem there

On the thank-you page, there shouldn't be any errors

### What actually happened

On the thank-you page, there is a flash of content, white page and an error in console 

### Impact

One

### Available workarounds?

Atomic

### Logs or notes

https://github.com/Automattic/wp-calypso/assets/2484390/4dbadae4-c605-4be6-a73e-a36b1f069b25

michaeldcain:
At some point, hooks were added to `domainRegistrationThankYouProps()` (and the naming updated) to retrieve the `site_intent` and `launchpadScreen` options, even though the function wasn't a component and is used in hooks elsewhere in the code. 

This PR pulls the hooks out of `domainRegistrationThankYouProps()` and passes the data through its caller in the `DomainThankYou` component.

Fixes: https://github.com/Automattic/wp-calypso/issues/78378

**To test:**
- on an Atomic site, purchase a domain
- verify that the Thank You page renders and doesn't WSOD

### DIFF
--- a/client/my-sites/checkout/checkout-thank-you/domains/domain-thank-you.tsx
+++ b/client/my-sites/checkout/checkout-thank-you/domains/domain-thank-you.tsx
@@ -34,6 +34,8 @@ const DomainThankYou: React.FC< DomainThankYouContainerProps > = ( {
 	hideProfessionalEmailStep,
 	type,
 } ) => {
+	const launchpadScreen = useSiteOption( 'launchpad_screen' );
+	const siteIntent = useSiteOption( 'site_intent' );
 	const thankYouProps = useMemo< DomainThankYouProps >( () => {
 		const propsGetter = domainThankYouContent[ type ];
 		return propsGetter( {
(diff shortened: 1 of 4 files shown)


sirbrillig:
For reference, this was a multi-tier bug. The cause appears to have been originally introduced in https://github.com/Automattic/wp-calypso/pull/56949 which added the poor naming that allowed the hooks that broke the page to be added in https://github.com/Automattic/wp-calypso/pull/73904. It's not clear why we didn't hear about the issue until now.

sirbrillig (approved):
This is much less confusing than what was there before. Thank you!

PR #78395, opened 2023-06-19
Disable polling for discover stream

eoigal:
Small PR to fix 2 issues with Discover page

* Sidebar width changes depending on width of sidebar site description - this PR sets a max width
* The discover stream polls, which I don't think it should - it doesn't work and just fails silently - this PR disables it for this streamType.
* Fix issue with undefined site description breaking when using `unescape` - picked up by sentry
*

### DIFF
--- a/client/reader/stream/style.scss
+++ b/client/reader/stream/style.scss
@@ -14,6 +14,19 @@
 		.tag-stream__header {
 			max-width: 600px;
 		}
+
+		.reader-tag-sidebar-recommended-sites {
+			max-width: 270px;
+
+			.reader-subscription-list-item .follow-button__label,
+			.reader-subscription-list-item__settings-label {
+				display: none;
+			}
+
+			.reader-site-notification-settings__button-label {
+				display: none;
+			}
+		}
 	}
 }
 
@@ -874,3 +887,4 @@
 		border-radius: 6px; /* stylelint-disable-line scales/radii */
 	}
 }
+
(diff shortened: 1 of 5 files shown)


client/state/data-layer/wpcom/read/streams/index.js
```diff
@@ -167,7 +167,6 @@ export const SITE_LIMITER_FIELDS = [
 	'feed_item_ID',
 	'global_ID',
 	'metadata',
-	'tags',
```
Addison-Stavlo, line 170: what is this doing?

client/reader/stream/style.scss
```diff
@@ -14,6 +14,19 @@
...
+
+			.reader-subscription-list-item .follow-button__label,
+			.reader-subscription-list-item__settings-label {
+				display: none;
+			}
+
+			.reader-site-notification-settings__button-label {
+				display: none;
+			}
+		}
```
Addison-Stavlo: are there any places this might cause regressions due to the general level of selectors? Should we be limiting anything to the discover stream or discover stream sidebar components here?

client/state/data-layer/wpcom/read/streams/index.js
eoigal: I added this in original [PR](https://github.com/Automattic/wp-calypso/pull/78137) but it's not needed - this array is used when `getQueryStringForPoll` is used, which it isn't for `discover` stream type.

client/reader/stream/style.scss
eoigal: Yeah, perhaps.. I'll try to limit this to just discover streams.

Addison-Stavlo (approved):

PR #78402, opened 2023-06-19
Reader: remove follow button from PostByline component

creativecoder:
## Proposed Changes

* Removes the FollowButton from PostByline
* Removes the `showPrimaryFollowButton` prop from several places in Reader that was used as a flag to show the follow button in the byline

The follow button is now shown in `ReaderPostActions` or separately in the `RelatedPostCard`. I couldn't find a remaining instance of it showing in the PostByline, so am removing it in this PR.

This cleanup is part of prep work for refreshing the Site feed page, in a similar style to what we've already done for the tag feed.

## Testing Instructions

- Spot check various reader pages (Following, Like, Conversations, Tags, etc) against production for any regressions

### DIFF
--- a/client/reader/tag-stream/main.jsx
+++ b/client/reader/tag-stream/main.jsx
@@ -26,7 +26,6 @@ class TagStream extends Component {
 	static propTypes = {
 		encodedTagSlug: PropTypes.string,
 		decodedTagSlug: PropTypes.string,
-		followSource: PropTypes.string.isRequired,
 	};
 
 	state = {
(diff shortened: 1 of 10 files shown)


bluefuton:
> I couldn't find a remaining instance of it showing in the PostByline, so am removing it in this PR.

Nice cleanup 👏 I think this byline follow button was originally used for the previous Discover (around 2015-16):

image

(from p5PDj3-1kk-p2)

client/reader/tag-stream/main.jsx
```diff
@@ -26,7 +26,6 @@ class TagStream extends Component {
 	static propTypes = {
 		encodedTagSlug: PropTypes.string,
 		decodedTagSlug: PropTypes.string,
-		followSource: PropTypes.string.isRequired,
```
creativecoder, line 29: This was causing a PropType warning in development, and requiring this prop doesn't seem to be relevant any more.

PR #78406, opened 2023-06-19
Tax Information: Update the sidebar of the tax information form 

JessBoctor:
Related to # https://github.com/Automattic/payments-shilling/issues/1747

Previously, we only provided `VAT Invoices` to users who were registered as a European Business. Since both the scope of who we charge tax-exclusive pricing and how we display our tax vendor information on receipts have changed, this is no longer accurate.

FG-AddTaxForm 

## Proposed Changes

* Remove the warning that we only provide tax receipts (VAT Invoices) to certain users
* Note that not all countries can be validated through this form and direct customers who are in other countries to contact support
* Add a link to tax documentation

## Testing Instructions

If you go to http://calypso.localhost:3000/me/purchases/vat-details, you should see the following screen: 

**BEFORE**

GH78406-Before-Screenshot 2023-07-11 at 15 44 49

Next, apply this patch and go to http://calypso.localhost:3000/me/purchases/vat-details

* You should see the new sidebar information: 

**AFTER**

GH78406-After-Screenshot 2023-07-11 at 15 54 57

* Click on the `please contact support` link and make sure it directs you to `http://calypso.localhost:3000/help`

GHPR78406-TaxInformationForm-clickSupportLink

* Click on the `learn more` link. It should open a dialogue displaying the support documentation from https://wordpress.com/support/vat-gst-other-taxes/
* Click on `Close` and make sure the dialogue closes
* Open the dialogue again and click on `Visit Article`, make sure the support article linked above is opened in a new tab

GHPR78406-TaxInformationForm-clickLearnMoreLink

* On http://calypso.localhost:3000/me/purchases/vat-details and select a country from the form fields
* Make sure the tax name (e.g. "tax (VAT/GST/CT)") changes to reflect the tax name for the selected country

GH78406-ChangeCountry-2023-07-11 15 55 16

## To dos:

- [x] Localized the URL in the `<InlineSupportLink>` component 
- [x] Make the support documentation pop-up in a modal (this already exists, just need to apply it)

## Pre-merge Checklist

### DIFF
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -88,7 +88,7 @@ export default function VatInfoPage() {
 							/* translators: %s is the name of taxes in the country (eg: "VAT" or "GST") or a generic fallback string of tax names */
 							'The %(taxName)s details saved on this page will be applied to all receipts in your account. Only supported countries appear in the dropdown. For more information about taxes, {{learnMoreLink}}click here{{/learnMoreLink}}.',
 							{
-								args: { taxName: taxName ?? translate( 'Tax (VAT/GST/CT)', { textOnly: true } ) },
+								args: { taxName: taxName ?? translate( 'tax (VAT/GST/CT)', { textOnly: true } ) },
 								components: {
 									learnMoreLink: (
 										<InlineSupportLink
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -1,4 +1,5 @@
 import { CompactCard, Button, Card } from '@automattic/components';
+import { localizeUrl } from '@automattic/i18n-utils';
 import i18n, { getLocaleSlug, useTranslate } from 'i18n-calypso';
 import { useEffect, useState, useRef, useMemo } from 'react';
 import CardHeading from 'calypso/components/card-heading';
@@ -33,6 +34,7 @@ export default function VatInfoPage() {
 	const taxName = useTaxName(
 		currentVatDetails.country ?? vatDetails.country ?? geoData?.country_short ?? 'GB'
 	);
+	const taxSupportPageURL = localizeUrl( 'https://wordpress.com/support/vat-gst-other-taxes/' );
 
 	useRecordVatEvents( { fetchError } );
 
@@ -92,9 +94,11 @@ export default function VatInfoPage() {
 								components: {
 									learnMoreLink: (
 										<InlineSupportLink
-											supportLink="https://wordpress.com/support/vat-gst-other-taxes/"
+											supportLink={ taxSupportPageURL }
 											showText={ true }
 											showIcon={ false }
+											supportPostId={ 234670 } //This is what makes it a modal dialogue
+											linkTitle="VAT, GST, and other taxes"
 										/>
 									),
 								},
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -109,10 +109,10 @@ export default function VatInfoPage() {
 						<br />
 						<br />
 						{ translate(
-							/* translators: %s is the name of taxes in the country (eg: "VAT" or "GST") or a generic fallback string of tax names */
 							/* translators: This is a list of tax-related reasons a customer might need to contact support */
 							'If you:' +
 								'{{ul}}' +
+								/* translators: %(taxName)s is the name of taxes in the country (eg: "VAT" or "GST") or a generic fallback string of tax names */
 								'{{li}}Need to update existing %(taxName)s details{{/li}}' +
 								'{{li}}Have been charged taxes as a business subject to reverse charges{{/li}}' +
 								'{{li}}Do not see your country listed in this form{{/li}}' +
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -85,7 +85,7 @@ export default function VatInfoPage() {
 					<p className="vat-info__sidebar-paragraph">
 						{ translate(
 							/* translators: %s is the name of taxes in the country (eg: "VAT" or "GST"). */
-							"We currently only provide %(taxName)s invoices to users who are properly registered. %(taxName)s information saved on this page will be applied to all of your account's receipts.",
+							"%(taxName)s details saved on this page will be applied to all of your account's receipts. Only supported countries appear in the dropdown. For more information about taxes, {click here}.",
 							{
 								textOnly: true,
 								args: { taxName: taxName ?? translate( 'VAT', { textOnly: true } ) },
--- a/client/me/purchases/vat-info/index.tsx
+++ b/client/me/purchases/vat-info/index.tsx
@@ -36,6 +36,12 @@ export default function VatInfoPage() {
 	);
 	const taxSupportPageURL = localizeUrl( 'https://wordpress.com/support/vat-gst-other-taxes/' );
 
+	const reduxDispatch = useDispatch();
+
+	const clickSupport = () => {
+		reduxDispatch( recordTracksEvent( 'calypso_vat_details_support_click' ) );
+	};
+
 	useRecordVatEvents( { fetchError } );
 
 	if ( fetchError ) {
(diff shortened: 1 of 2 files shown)


JessBoctor:
@michaeldcain and I went over this PR in a call. We decided on some improvements to the information provided.

This should make the reasons to contact Happiness more explicit and provide proper translations for link titles. Also, I learned how to add HTML elements into translation strings. 

Here is the expected layout now:

Screenshot 2023-06-21 at 16 24 50

JessBoctor:
Once this has a final review, I will put a string freeze on it.

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
 							{
-								textOnly: true,
-								args: { taxName: taxName ?? translate( 'VAT', { textOnly: true } ) },
+								args: { taxName: taxName ?? translate( 'tax (VAT/GST/CT)', { textOnly: true } ) },
+							}
+						) }
+						<br />
+						<br />
+						{ translate(
+							/* translators: %s is the name of taxes in the country (eg: "VAT" or "GST") or a generic fallback string of tax names */
```
sirbrillig: I'll be curious to see if this works or if it reports that `taxName` has no description. My understanding is that `translators` comments need to be on the line immediately preceding the string with the placeholder, but in this case there's two differences and I don't know how they will affect this: there are two `translators` comments one after another, and the string is broken up into multiple strings. 🤷  It might be fine.

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
+							/* translators: This is a list of tax-related reasons a customer might need to contact support */
+							'If you:' +
```
sirbrillig, line 113: I understand the intent of this UI but I do wonder if it might be hard to translate well into other languages. The construction of `If you: X / Y / X` feels like an English thing to me. I'm not sure how I'd go about translating it into Japanese [which uses verb conjugations to convey conditionals](https://storylearning.com/learn/japanese/japanese-tips/japanese-conditional-form). 🤔 

When faced with complexity like this I usually try to reconsider what I actually want to say and see if there's a simpler sentence structure that would work for the same idea.

In this case, what about something like, `If you need to change saved %(taxName) details, if you have been charged taxes as a business subject to reverse charges, or if you do not see your country listed in this form, {{contactSupportLink}}contact our Happiness Engineers{{/contactSupportLink}}.` then followed by a separate string `Include your %(taxName)s number and country code when you contact us.`?

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
+								'{{ul}}' +
+								'{{li}}Need to update existing Tax ID details{{/li}}' +
```
sirbrillig: Should this be something like `Need to update existing %(taxName)s details`?

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
+								'{{li}}Have been charged VAT as a VAT-Registered Business subject to reverse charges{{/li}}' +
```
sirbrillig: Should this use `%(taxName)s` also? I think this could actually be re-worded to avoid mentioning VAT at all if you write something like `if you have been charged taxes as a business subject to reverse charges`.

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
+								'{{li}}Do not see your country listed in this form{{/li}}' +
+								'{{/ul}}' +
+								'{{contactSupportLink}}Contact our Happiness Engineers{{/contactSupportLink}}. Include your %(taxName)s number and country code when you contact us.',
+							{
+								args: { taxName: taxName ?? translate( 'tax', { textOnly: true } ) },
```
sirbrillig: I don't mind it but is there a reason to use `tax` as the fallback here instead of `tax (VAT/GST/CT)` like other places?

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,58 @@ export default function VatInfoPage() {
...
+									),
+								},
+							}
+						) }
+						<br />
+						<br />
+						{ translate(
+							'For more information about taxes, {{learnMoreLink}}click here{{/learnMoreLink}}.',
+							{
+								args: { taxName: taxName ?? translate( 'tax (VAT/GST/CT)', { textOnly: true } ) },
```
sirbrillig: This string doesn't have any string interpolation so we don't need `args`.

client/me/purchases/vat-info/index.tsx
JessBoctor: Thanks for the insight! We actually pulled this from our support documentation. A similar version has already been translated into Japanese here: https://wordpress.com/ja/support/vat-gst-other-taxes/#i-cant-add-my-vat-number-to-my-account

```
もし

- 既存のビジネス納税者番号情報を更新する必要がある場合
- リバースチャージの対象となる VAT 登録事業者として VAT を請求されたことがある場合
- 購入または更新を完了できない場合
- ご質問がある場合は…
```

I would imagine the string starting at line 114 would be translated similarly. Do you think the version above makes sense?

client/me/purchases/vat-info/index.tsx
JessBoctor: I thought so at one point, too. 

However, it seems that the `Reverse chargeback` is only a thing in the EU. This is something specific that is separate from refunding tax that was charged to an exempt organization. 

Maybe we don't need this anymore, since all of our receipts have our vendor information on them. I couldn't quite tell from the description I found: https://www.revenue.ie/en/vat/what-is-vat/reverse-charge-self-accounting.aspx

client/me/purchases/vat-info/index.tsx
JessBoctor: Fixed! Thanks for catching that.

client/me/purchases/vat-info/index.tsx
JessBoctor: >It might be fine.

😬 If you were to do something different, what would it be?

client/me/purchases/vat-info/index.tsx
JessBoctor: Another thought might be to only show it is the user is in the EU? Not sure if that makes it overly complicated.

client/me/purchases/vat-info/index.tsx
JessBoctor: Fixed!

client/me/purchases/vat-info/index.tsx
JessBoctor: At one point, I had replaced the first `tax (VAT/GST/CT)` with just `tax` because it looked strange to have it directly under the card title. I must have copied this over from that. 

I have changed all of the translation strings to `fallbackTaxName`. This way we only have to translate it once.

client/me/purchases/vat-info/index.tsx
sirbrillig: Oh, interesting. Translated back to English, that looks something like what I suggested above, actually: making each bullet point its own full "if" sentence.

Screenshot 2023-06-27 at 10 16 17 AM

I suppose the translators just figured out a way to do it. I still feel that this is a little awkward to localize but I won't consider it a blocker.

client/me/purchases/vat-info/index.tsx
sirbrillig: That would not be easy to do, I think. Alright, if this is what's most accurate, then I'm ok with it.

client/me/purchases/vat-info/index.tsx
sirbrillig: I would put this comment directly above the line where `%s` is used (actually it looks like `%s` is never used in this translation? maybe you meant `%(taxName)s`?).

client/me/purchases/vat-info/index.tsx
JessBoctor: Got some more feedback from the tax team here: p1688673851881119-slack-payments-tax-engine.

Will update to match the suggested `if you have been charged taxes as a business subject to reverse charges`

JessBoctor: Fixed.

client/me/purchases/vat-info/index.tsx
JessBoctor: Done! Thanks!

client/me/purchases/vat-info/index.tsx
```diff
@@ -84,11 +100,57 @@ export default function VatInfoPage() {
...
+						{ translate(
+							'For more information about taxes, {{learnMoreLink}}click here{{/learnMoreLink}}.',
+							{
+								components: {
+									learnMoreLink: (
+										<InlineSupportLink
+											supportLink={ taxSupportPageURL }
+											showText={ true }
+											showIcon={ false }
+											supportPostId={ 234670 } //This is what makes the document appear in a dialogue
```
sirbrillig, line 149: I'm curious where you got this from? 

IIRC `InlineSupportLink` always displays help in a dialog (but maybe I'm wrong)? [Here is one place](https://github.com/Automattic/wp-calypso/blob/21125523f7e98d17f72614cf350b3062604fb564/client/my-sites/checkout/composite-checkout/payment-methods/credit-card/assign-to-all-payment-methods.tsx#L45) that uses it in checkout. Oh, I guess that prop it overrides the `supportLink`. That's fine, but this comment is a little misleading I think?

sirbrillig (approved):
This looks good to me! Thanks!

client/me/purchases/vat-info/index.tsx
JessBoctor: >I'm curious where you got this from?

Trial and error 😂 
Also, looked back to one of the PR's I worked on with Cain. 

If you don't set the post ID with the ID for the support page, the InlineSupportLink redirects to the page. If you set the `supportPostId` parameter, then it uses the dialogue.

a8ci18n:
This Pull Request is now available for translation here: https://translate.wordpress.com/deliverables/8362387

Thank you @JessBoctor for including a screenshot in the description! This is really helpful for our translators.

a8ci18n:
Translation for this Pull Request [has now been finished](https://translate.wordpress.com/deliverables/overview/8362387).

JessBoctor:
Deployed! 🎉

JessBoctor:
Screenshot 2023-07-25 at 14 28 05

PR #78411, opened 2023-06-20
Reader: combine compact and tag post cards

creativecoder:
## Proposed Changes

* Combined the compact and tag post reader post card layouts.
* Uses the more compact layout on the conversations page, where the focus is following the conversation, rather that browsing post content
* Moved the byline and ellipsis to better fit into overall card layout
* Added action icons (follow, like, etc) to the bottom of tag post cards (pe7F0s-YE-p2)
* Remove references to "is tag post" in code so the layout can be used in other places without confusion
* Fixed regression from a previous update that narrows the content cards

This is part of some prep work to prepare for refreshing the site feed page, which will use the same post card layout.

## Testing Instructions

* The main pages affected are /tags/{tag} and /read/conversations. Compare those pages with production, making sure the layout looks correct and clicks have the same functionality

### Tags

| **Before** | **After** |
| - | - |
| image | image |
| image | image |
| image | image |

### Conversations

| **Before** | **After** |
| - | - |
| image | image |


client/blocks/reader-post-card/byline.jsx
```diff
@@ -106,7 +108,7 @@ class PostByline extends Component {
 									target="_blank"
 									rel="noopener noreferrer"
 								>
-									{ siteSlug }
+									{ siteName && ( compact || ! showSiteName ) ? siteName : siteSlug }
```
roo2: this conditional doesn't look right to me, shouldn't showSiteName need to be true in order to display the `siteName`?

client/blocks/reader-post-card/index.jsx
```diff
@@ -152,7 +152,6 @@ class ReaderPostCard extends Component {
 			'is-seen': isSeen,
 			'is-expanded-video': isVideo && isExpanded,
 			'is-compact': compact,
-			'is-tag-post': isReaderTagPage,
```
roo2: nit: there's a check in client/blocks/reader-post-card/style.scss:62 that's no longer neccessary now that this is taken out

roo2 (approved):
LGTM. Just noting that it makes the posts section of the /tag/$tag pages a bit wider but this still works well across different screen sizes

creativecoder:
Given the discussion in pe7F0s-YE-p2, I plan on adding the follow button to the layout in this PR for the tags page.

client/blocks/reader-post-card/byline.jsx
creativecoder: It's using the site name in this position (rather than url) if it's not otherwise showing in the layout. I'll add a comment to explain.

client/blocks/reader-post-card/index.jsx
creativecoder: Thanks!

client/blocks/reader-post-card/byline.jsx
roo2: ahh thanks

client/reader/stream/index.jsx
```diff
@@ -516,7 +515,6 @@ class ReaderStream extends Component {
 					renderItem={ this.renderPost }
 					renderLoadingPlaceholders={ this.renderLoadingPlaceholders }
 					className="stream__list"
-					tags={ tags }
```
creativecoder, line 519: This isn't specifically related to the PR, but was inserting `tags="[object Object]"` into the dom and doesn't seem to have a specific purpose, so I removed it.

client/reader/tag-stream/main.jsx
```diff
@@ -149,6 +149,8 @@ class TagStream extends Component {
 				showFollowInHeader={ true }
 				forcePlaceholders={ ! tag } // if tag has not loaded yet, then make everything a placeholder
 				streamHeader={ tagHeader }
+				showSiteNameOnCards={ false }
+				useCompactCards={ true }
```
creativecoder, line 153: @Addison-Stavlo @eoigal You should be able to add the `useCompactCards={ true }` prop to the `Stream` component you are using for the Discover page to adopt this same layout.

eoigal (approved):
LGTM! Great job Grant, this cleans things up nicely.

Issue #77645, opened 2023-06-01
Assembler - make sure only one snackbar of same type displays at once


miksansegundo:
## What

Stacking multiple notices obscures the interface in the assembler.

![Image](https://user-images.githubusercontent.com/1881481/242487750-3c14dec6-d135-4c01-9cc6-4e6bea85f5b8.png)

## Why

For better visibility of the canvas. Following the same behavior as in the editor. See https://github.com/WordPress/gutenberg/pull/50625

## How

Let's show only one notice of the same type (add/replace/remove) at a time.

fushar:
This is shipped!

PR #78421, opened 2023-06-20
Site Assembler: ensure only one snackbar notice is shown per unique action

fushar:
## Fixes:

- #77645

## Proposed Changes

Currently, each action (add/replace/remove) in the pattern inserter in the Site Assembler flow triggers a snackbar notice in the bottom-right corner of the preview panel on the right. When we do many actions in a short time interval (e.g. trying out headers), this will generate noisy notices and somewhat block the preview.

This PR fixes this problem by only showing one notice per action (add/replace/remove). This is done by assigning the action itself as the ID of the notice, so that we can first remove the previous notice with the same ID before creating a new one.

## Testing Instructions

1. Go to `/setup/with-theme-assembler/patternAssembler?siteSlug=<site slug>`.
2. Try inserting and replacing header/footer many times; make sure that only one `Block pattern ... replaced` notice is shown.
3. Try inserting and removing homepage patterns many times; make sure that only one `Block pattern ... added` / `Block pattern ... `removed` notice is shown (for each add/remove).

Illustrative videos:

|Before|After|
|-|-|
|<video src="https://github.com/Automattic/wp-calypso/assets/1525580/50f9ffac-bbe8-4175-a387-62a3345a3c7a">|<video src="https://github.com/Automattic/wp-calypso/assets/1525580/b8db0eb7-6f91-46f4-b682-c0f173b99207">|

## Pre-merge Checklist

### DIFF
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -213,7 +213,8 @@ const PatternAssembler = ( {
 	};
 
 	const showNotice = ( action: string, pattern: Pattern ) => {
-		noticeOperations.createNotice( { content: getNoticeContent( action, pattern ) } );
+		noticeOperations.removeNotice( action );
+		noticeOperations.createNotice( { id: action, content: getNoticeContent( action, pattern ) } );
 	};
 
 	const getDesign = () =>
--- a/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
+++ b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
@@ -6,7 +6,6 @@ import {
 	__experimentalNavigatorProvider as NavigatorProvider,
 	__experimentalNavigatorScreen as NavigatorScreen,
 	__experimentalUseNavigator as useNavigator,
-	withNotices,
 } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { useDispatch, useSelect } from '@wordpress/data';
@@ -33,7 +32,7 @@ import usePatternCategories from './hooks/use-pattern-categories';
 import usePatternsMapByCategory from './hooks/use-patterns-map-by-category';
 import { usePrefetchImages } from './hooks/use-prefetch-images';
 import useRecipe from './hooks/use-recipe';
-import Notices, { getNoticeContent } from './notices/notices';
+import withNotices, { NoticesProps } from './notices/notices';
 import PatternAssemblerContainer from './pattern-assembler-container';
 import PatternLargePreview from './pattern-large-preview';
 import ScreenActivation from './screen-activation';
@@ -61,9 +60,9 @@ const PatternAssembler = ( {
 	navigation,
 	flow,
 	stepName,
-	noticeList,
 	noticeOperations,
-}: StepProps & withNotices.Props ) => {
+	noticeUI,
+}: StepProps & NoticesProps ) => {
 	const translate = useTranslate();
 	const navigator = useNavigator();
 	const [ sectionPosition, setSectionPosition ] = useState< number | null >( null );
@@ -212,11 +211,6 @@ const PatternAssembler = ( {
 		} );
 	};
 
-	const showNotice = ( action: string, pattern: Pattern ) => {
-		noticeOperations.removeNotice( action );
-		noticeOperations.createNotice( { id: action, content: getNoticeContent( action, pattern ) } );
-	};
-
 	const getDesign = () =>
 		( {
 			...selectedDesign,
@@ -238,12 +232,12 @@ const PatternAssembler = ( {
 		updateActivePatternPosition( -1 );
 		if ( pattern ) {
 			if ( header ) {
-				showNotice( 'replace', pattern );
+				noticeOperations.showPatternReplacedNotice( pattern );
 			} else {
-				showNotice( 'add', pattern );
+				noticeOperations.showPatternInsertedNotice( pattern );
 			}
 		} else if ( header ) {
-			showNotice( 'remove', header );
+			noticeOperations.showPatternRemovedNotice( header );
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
```diff
@@ -213,7 +213,8 @@ const PatternAssembler = ( {
 	};
 
 	const showNotice = ( action: string, pattern: Pattern ) => {
-		noticeOperations.createNotice( { content: getNoticeContent( action, pattern ) } );
+		noticeOperations.removeNotice( action );
+		noticeOperations.createNotice( { id: action, content: getNoticeContent( action, pattern ) } );
```
miksansegundo: I think we have to find a way to restart the [timer](https://github.com/Automattic/wp-calypso/blob/1f99ec0c229b9a9d31916c606958e1d1730a7c7b/client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/notices/notices.tsx#L35) so we make sure new notices stay on the screen for 5 seconds.

fushar: Thanks, I missed this!

However, it turned out that it's not as easy as it sounds. What I am doing here is as follows:

1. Establish a mapping between notice content => 5-second timer to remove the notice
   * we cannot use notice id as it will be the same / we cannot detect whether this notice of this id is newer
1. When we receive a new `noticeList` in the `useEffect()` hook:
   * clear and remove the timers of old notices that are already gone (dismissed or replaced because it has the same id as the new notice)
   * trigger a new timer for new notice; which is a notice with content that is not in the map yet.
  
Note that this logic has one edge case: when a user repeatedly pick the same pattern, the timer will not be reset, as the content is the same :( However, I think this is a very rare case. The code is complicated already and I am reluctant to handle this, as this will completely change the implementation 😭 

Let me know what you think!

fushar: One thing that adds to the complexity of this solution: The following two lines:

```
noticeOperations.removeNotice( action );
noticeOperations.createNotice( { id: action, content: getNoticeContent( action, pattern 
```

trigger two state changes, but React somehow batches the updates, so they only trigger one rerender and consequently one `useEffect()` call will be fired (not two). So, we only receive the final `nodeList`; we don't get the chance the see the `nodeList` version when the replaced notice is removed. (If we did, we would have had a way to clear the timer first, for example)

arthur791004: Yep, react will [batch state updates](https://react.dev/learn/queueing-a-series-of-state-updates#react-batches-state-updates) to improve the performance. If you want, you can use [flushSync](https://react.dev/reference/react-dom/flushSync) to flush any updates synchronously but it might have a bad effect on performance.

arthur791004: A simple way is to clear the timer here as we're able to get the `noticeList`. Or we can also get rid of [withNotices](https://github.com/WordPress/gutenberg/blob/c819200904335a4ce4c67cf48ee64a7427e0a09c/packages/components/src/higher-order/with-notices/index.tsx#L52) as it seems not to meet our requirements.

fushar: Interesting idea, let me try real quick...

miksansegundo: Feel free to completely change the implementation to reduce the complexity and... Enjoy the challenge! Happy to have you back @fushar.

miksansegundo (approved):
The UX LGTM 👍 

https://github.com/Automattic/wp-calypso/assets/1881481/a5a4bf36-49fc-4dca-a26b-4a90dbe0f09f

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/index.tsx
fushar: @arthur791004 @miksansegundo I've kind of reimplemented `withNotices` for our requirements. Now I've removed the `useEffect()` "hack" and I feel this logic is more readable and maintainable. Please check! 😄 

@miksansegundo : I'm happy to be back, too!

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/notices/notices.tsx
```diff
@@ -10,47 +11,91 @@ type Notice = NoticeList.Notice & {
...
+					showPatternRemovedNotice: ( pattern: Pattern ) => {
+						createNotice(
+							'pattern-removed',
+							i18n.translate( 'Block pattern "%(patternName)s" removed.', {
+								args: { patternName: pattern.title },
+							} )
+						);
+					},
+				};
+			}, [ noticeList, removeNotice ] );
```
arthur791004: The memo might not be very useful as the `noticeList` would be updated all the time. Maybe we can remove `useMemo` or use `useRef` to keep the reference of the `noticeList`.

arthur791004 (approved):
Tested and looks good to me 🎉

client/landing/stepper/declarative-flow/internals/steps-repository/pattern-assembler/notices/notices.tsx
fushar: Ah, you're right 🤦 I've removed the memoization, thanks 😄

miksansegundo:
@fushar, tested again and looks ready to ship 👍

https://github.com/Automattic/wp-calypso/assets/1881481/b627cbf5-0d3c-4f81-a441-d8a274f0a366

Some friendly reminders:

- we mark all the boxes from the Pre-merge Checklist, even if they are not done but to acknowledge that they are not required. 
- after shipped, add it to Marvel Ship Table

fushar:
@miksansegundo Thanks for testing it again and the reminders 😄 shipping... 🚀

PR #78438, opened 2023-06-20
Reader: Update discover tags stream layout

eoigal:
This PR fixes the Reader discover tags stream layout to copy what we use on the tag pages.

### Before
Screenshot 2023-06-20 at 14 04 12

### After
Screenshot 2023-06-20 at 16 41 24

### Scope
 
I'm not sure if this PR should update the `Recommended` tab post stream to have the same layout as the tags post stream. 

Screenshot 2023-06-20 at 16 44 19

The post stream use different post components.

* The `Recommended` tab uses the `RelatedPostCard` component
* The tags tabs use the `ReaderPostCard` component

### DIFF
--- a/client/blocks/reader-post-card/index.jsx
+++ b/client/blocks/reader-post-card/index.jsx
@@ -20,6 +20,7 @@ import isFeedWPForTeams from 'calypso/state/selectors/is-feed-wpforteams';
 import isReaderCardExpanded from 'calypso/state/selectors/is-reader-card-expanded';
 import isSiteWPForTeams from 'calypso/state/selectors/is-site-wpforteams';
 import { getReaderTeams } from 'calypso/state/teams/selectors';
+import PostByline from './byline';
 import ConversationPost from './conversation-post';
 import GalleryPost from './gallery';
 import PhotoPost from './photo';
@@ -115,12 +116,13 @@ class ReaderPostCard extends Component {
 			currentRoute,
 			post,
 			discoverPost,
-			discoverSite,
 			site,
 			feed,
 			onCommentClick,
+			showPrimaryFollowButton,
 			isSelected,
 			showSiteName,
+			followSource,
 			isDiscoverStream,
 			postKey,
 			isExpanded,
@@ -141,7 +143,6 @@ class ReaderPostCard extends Component {
 		const isDiscover = post.is_discover;
 		const title = truncate( post.title, { length: 140, separator: /,? +/ } );
 		const isReaderTagPage = currentRoute.startsWith( '/tag/' );
-		const isReaderDiscoverPage = currentRoute.startsWith( '/discover' );
 		const isReaderSearchPage = currentRoute.startsWith( '/read/search' );
 		const classes = classnames( 'reader-post-card', {
 			'has-thumbnail': !! post.canonical_media,
@@ -152,7 +153,7 @@ class ReaderPostCard extends Component {
 			'is-seen': isSeen,
 			'is-expanded-video': isVideo && isExpanded,
 			'is-compact': compact,
-			'is-tag-post': isReaderTagPage || isReaderDiscoverPage,
+			'is-tag-post': isReaderTagPage || isDiscoverStream,
 		} );
 
 		/* eslint-disable wpcalypso/jsx-classname-namespace */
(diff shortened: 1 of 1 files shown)


client/blocks/reader-post-card/index.jsx
```diff
@@ -146,6 +139,7 @@ class ReaderPostCard extends Component {
 		const isDiscover = post.is_discover;
 		const title = truncate( post.title, { length: 140, separator: /,? +/ } );
 		const isReaderTagPage = currentRoute.startsWith( '/tag/' );
+		const isReaderDiscoverPage = currentRoute.startsWith( '/discover' );
```
Addison-Stavlo: i think we already have a `isDiscoverStream` prop passed down from the discover controller we could alternatively use here.

client/blocks/reader-post-card/index.jsx
```diff
@@ -177,43 +171,6 @@ class ReaderPostCard extends Component {
...
-					feed={ feed }
-					showSiteName={ showSiteName || isDiscover }
-					showAvatar={ ! compact }
-					teams={ teams }
-					showFollow={ ! isDiscover }
-					showPrimaryFollowButton={ showPrimaryFollowButton }
-					followSource={ followSource }
-				/>
-			);
-		}
```
Addison-Stavlo: Is removing this `postByline` for the `else` case going to effect other pages or streams unexpectedly?

If we are looking to remove this for discover only, we could make this `null` in the conditional above, or just define it like this (the else condition) all the time and add a `! isDiscoverStream` to where we bring it in to render?

client/blocks/reader-post-card/index.jsx
```diff
@@ -287,7 +243,7 @@ class ReaderPostCard extends Component {
 		const onClick = ! isPhotoPost && ! compact ? this.handleCardClick : noop;
 		return (
 			<Card className={ classes } onClick={ onClick } tagName="article">
-				{ ! compact && ! isReaderTagPage && postByline }
+				{ ! compact && ! isReaderTagPage }
```
Addison-Stavlo: if we really are going to remove the postByline for all cases, we could remove this entire line.

client/blocks/reader-post-card/index.jsx
Addison-Stavlo: > going to effect other pages or streams unexpectedly?

It looks like there is a regression on the following stream (posts in the feed are missing the top section that is probably this byline)

Addison-Stavlo:
These tags pages look a lot better. A couple comments above.

Aside from that, im confused about this one section in the tags sidebar where it says "posts" and "sites" at the top.  In your screenshot above it shows them with numbers so they look like they are working as expected:

Screenshot 2023-06-20 at 12 51 49 PM

For the last few days (on trunk, production, this branch, all other branches) I have seen them without those numbers both in our discover tags sidebar as well as the normal tags pages. Do these seem to be working right for you?

Screenshot 2023-06-20 at 12 52 47 PM

davemart-in:
> I'm not sure if this PR should update the Recommended tab post stream to have the same layout as the tags post stream.

Happy to do it in this PR or a separate one, but yeah, let's make sure the stream cards are identical. It would probably be weird if the format was different between the recommended tab and the other "tag" tabs.

Note: @creativecoder is working on a PR here to consolidate some of these into a single reusable component in #78411

client/blocks/reader-post-card/index.jsx
eoigal: Yep, that will work.

client/blocks/reader-post-card/index.jsx
eoigal: I got a little carried away here! I've returned it and added condition below to skip including byline for discover stream posts.

eoigal:
> It would probably be weird if the format was different between the recommended tab and the other "tag" tabs.

I agree.

> Note: @creativecoder is working on a PR here to consolidate some of these into a single reusable component in #78411

This is great. The discover page can use this compact component when done. Might as well do that in a follow up PR though.

eoigal:
> For the last few days (on trunk, production, this branch, all other branches) I have seen them without those numbers both in our discover tags sidebar as well as the normal tags pages. Do these seem to be working right for you?

Yeah, these numbers are working ok for me in production.
Screenshot 2023-06-21 at 10 16 21

Addison-Stavlo (approved):
WFM! 😁 

Agreed we can update to the refactored component in a followup.

PR #78454, opened 2023-06-21
[E2E][AT][GB] Announce nightly GB E2E results on Slack in the corresponding version announcement thread, ping relevant codeowners

fullofcaffeine:
Project: p4TIVU-aAZ-p2

## Proposed Changes

Change the `E2EBuildType` create by `gutenbergPlaywrightBuildType(...)` to have two additional steps that will call a shell script in `./bin`, that in turns will use the Slack API to post a threaded message with the e2e test results to the corresponding version announcement message for the nightly version associated with the build.

For that to work, a few new parameters were added to the build, but *only* if it's a `nightly` build (for now):

* `GB_E2E_ANNOUNCEMENT_SLACK_API_TOKEN`: The Slack API token used to access the Slack API. The encrypted value has been added to TC, and is referenced here.
* `GB_E2E_ANNOUNCEMENT_SLACK_CHANNEL_ID_TEST`: The channel ID where the ann. message is posted to. Notice the `_TEST` suffix. This is because for now this points to a test channel. After tests confirm this works fine, we'll create another encrypted token and change the param to `GB_E2E_ANNOUNCEMENT_SLACK_CHANNEL_ID` and reference the new token.
* `GB_E2E_ANNOUNCEMENT_THREAD_TS`: This is the `thread_ts` param that the Slack API returns for a message after it's posted by the API. This is effectively the unique key for a message on Slack. Gutenbot will post the announcement message and then forward this data to TC when the it will trigger the nightly tests (this change is still pending)

Depending on how well it works, we might extend this flow to regular core/RC releases later.

## Testing Instructions

* Checks should pass;
* Merge and 🤞🏻

## TODO

- [x] Change Gutenbot to forward the `thread_ts` for the announcement message posted for a nightly build when triggering the builds for nightlies. This will be a bit tricky as the E2Es are triggered before announcements, so we'll need to change it so that the announcement message is posted before the E2E is triggered in order to get and pass the `thread_ts` value for it to the TC builds D114199-code


.teamcity/_self/projects/WPComTests.kt
```diff
@@ -107,20 +111,37 @@ fun gutenbergPlaywrightBuildType( targetDevice: String, buildUuid: String, atomi
...
-			// to setup a new nightly build for simple in the future. Nightly+edge doesn't make sense,
-			// but I don't think guarding against that here is worth the additional logic, as this is
-			// a factory function and a mistake like that should be clear by checking the arguments.
 			if (nightly) {
 				param("env.GUTENBERG_NIGHTLY", "true");
+				password("GB_E2E_ANNOUNCEMENT_SLACK_API_TOKEN", "credentialsJSON:8196e9b8-cf0a-4ab5-9547-95145134f04a", display = ParameterDisplay.HIDDEN);
+				// Once we move back to #gutenberg-on-dotcom in Gutenbot, change this to `GB_ANNOUNCEMENT_SLACK_CHANNEL_ID` that points
+				// to a token that has the corresponding channel id.
+				password("GB_E2E_ANNOUNCEMENT_SLACK_CHANNEL_ID_TEST", "credentialsJSON:180d1bb6-a28e-4985-bf9a-8acba63bb90c", display = ParameterDisplay.HIDDEN);
+				text("GB_E2E_ANNOUNCEMENT_THREAD_TS", value = "", allowEmpty = true, display = ParameterDisplay.HIDDEN);
```
worldomonation, line 123: **question:non-blocking** - these two lines are meant to obfuscate the Slack channel name? Not sure what the "TS" in `GB_E2E_ANNOUNCEMENT_THREAD_TS` is referring to. 

Since I don't think it's possible to view the data contained in "credentialsJSON:..." again, it might make sense to document as much as feasible in comments.

worldomonation (approved):

.teamcity/_self/projects/WPComTests.kt
fullofcaffeine: Hi Edwin 👋🏻, thanks for the review! Yes, I was extra ~paranoid~ careful here and obfuscated the Slack channel ID. The TS is a timestamp that's returned by the Slack message posting API endpoint and is used as a unique key to find that message and post other threaded messages to it. I'll add some comments there to clarify.

fullofcaffeine: Ah, I already documented it in the PR description, actually: https://github.com/Automattic/wp-calypso/pull/78454#issue-1766525701.

fullofcaffeine:
I'll merge this as I need these changes to test the whole flow (see: D114199-code). I'll deal with any issues - if any - as they arise in subsequent PRs.

PR #78460, opened 2023-06-21
CI: do not trigger WebApp and WPComTests builds when non-code files are updated.

worldomonation:
## Proposed Changes

This PR adds some trigger conditions to builds that we know clogs up the queue.

While the WPComPlugin refactor project is on hold, this is a small change we can make that should be safe and helps reduce the number of builds in the queue somewhat.

Key changes:
- do not trigger Plugin and E2E builds on document (`.md`) changes.
- similar to above, do not trigger Plugin builds on changes that include the term "stress test" (this is something I often do while fixing E2Es).

If there are other clauses that are safe to add, please let me know.

## Testing Instructions

TeamCity config should should work fine.

## Pre-merge Checklist

### DIFF
--- a/.teamcity/_self/lib/wpcom/WPComPluginBuild.kt
+++ b/.teamcity/_self/lib/wpcom/WPComPluginBuild.kt
@@ -26,7 +26,6 @@ open class WPComPluginBuild(
 	var buildSteps: BuildSteps.() -> Unit = {},
 	var buildParams: ParametrizedWithType.() -> Unit = {},
 ) : BuildType() {
- the
 	init {
 		// This block allows us to use variable names without having to prefix them with `this.@WPComPluginBuild.`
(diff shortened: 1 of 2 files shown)


dpasque (approved):
I'll defer to the platform folks if there's any weird gotchas with the plugin builds, but this LGTM for E2E builds! 👍

noahtallen (approved):
I think this will work fine! The big thing to avoid would be anything breaking the assumption of a new build for each trunk commit, but I don't think this would break that

worldomonation:
> I think this will work fine! The big thing to avoid would be anything breaking the assumption of a new build for each trunk commit, but I don't think this would break that

Yes, this should have minimal impact because the scoping is so tight as described in the original comment.

If this work out, I'd to proceed to a future where only really necessary builds get triggered, on top of the paused project to combine the Plugin builds.

PR #78461, opened 2023-06-21
Site Transfers: Support going back from site ownership transfer 2nd to 1st step

wojtekn:
Related to https://github.com/Automattic/dotcom-forge/issues/2794

## Proposed Changes

In this PR, I propose to change "Back" button behavior to point from the 2nd to 1st step, and not to General Settings

## Testing Instructions

1. Click “Transfer your site”
2. Choose administrator on the list, click “Continue”
3. Read the transfer confirmation text and change your mind about the new site owner
4. Click “Back”
5. Confirm that you see the first step of site ownership transfer and you can choose another administrator

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/site-settings/site-owner-transfer/site-owner-transfer.tsx
+++ b/client/my-sites/site-settings/site-owner-transfer/site-owner-transfer.tsx
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled';
 import { useTranslate } from 'i18n-calypso';
+import page from 'page';
 import { useState } from 'react';
 import { useQueryUserPurchases } from 'calypso/components/data/query-user-purchases';
 import { ResponseDomain } from 'calypso/lib/domains/types';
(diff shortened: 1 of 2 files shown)


kozer (changes requested):
Nice work @wojtekn !

I've found 2 small issues:

### Issue 1: Strange UI flow

The first one is kind a personal preference I guess, but I don't like it much.
When the user finishes the process, and get's the message that the email was sent, 
pressing back, navigates him to the first step, with no clear way to move back to settings.
He needs to navigate to the sidebar and select the menu from there, that requires extra effort.

Maybe adding a button to end the process in the last screen will help?

### Issue 2: Incorrect flow when we have more than 1 user.
This is how to replicate it:

1. Create 2 users, and invite them to your site
2. Start transfer for the first one.
3. Navigate through the end, and make sure you see the "email sent" message
4. Press back
5. You are now on the first screen to select a user. Now, select the other user.
6. You are navigated to the "Email sent" screen, which is incorrect.

katinthehatsite:
I agree with what @kozer mentioned for the first point. The UX is a bit strange if you are on the very last step and the email was sent and then press the `Back` button. One option that I can think of is maybe for that last `Back` button to go to something like `https://container-unruffled-mcnulty.calypso.live/settings/general/` ?

katinthehatsite (commented):
I could not find any other suggestions beside what @kozer mentioned - everything else worked as expected.

kozer:
> I agree with what @kozer mentioned for the first point. The UX is a bit strange if you are on the very last step and the email was sent and then press the `Back` button. One option that I can think of is maybe for that last `Back` button to go to something like `https://container-unruffled-mcnulty.calypso.live/settings/general/` ?

This actually sounds like it's a good candidate for a "wizard" component.
I don't know if we can reuse the "stepper" for that, or some portions of it.

It was something I thought about when I worked on the various steps. However it would increase the complexity, so I decided to make back button go straight to general.

cc: @wojtekn

wojtekn:
@katinthehatsite @kozer, thanks for giving it a test! I've modified conditions to go back to the first step only in one case and to go back to General Settings in all other cases.

> This actually sounds like it's a good candidate for a "wizard" component.
> I don't know if we can reuse the "stepper" for that, or some portions of it.

Probably it would be a good idea, but at this point, I think it's better to go with the current component.

kozer (approved):

kozer:
Works as expected now, @wojtekn !
Nice work!

katinthehatsite:
Works as expected for me as well!

PR #78476, opened 2023-06-21
Fix traffic breakdown margin for BlazePress' Campaign details

timur987:
Related to SELFSERVE-561

## Proposed Changes

* Adjust padding for left and right to `8px`,

image

* Make sure we do not consider negative values while calculating total.

image

## Testing Instructions

* Checkout this branch in your local dev. environment
* Build assets (`yarn && yarn start`)
* Make up Ad/Organic data [here](https://github.com/Automattic/wp-calypso/blob/trunk/client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx#L161-L170) like:
image

* Open a campaign details page
* Ensure your margins are 8px from each side
Screenshot 2023-06-21 at 13 31 06

* Also, check that the percentage is calculated correct even if backend provided us with negative values (in particular, if you have `327` clicks for **Ad** and a negative value for **Organic**, the former should have `100%` fill as on the image below):
Screenshot 2023-06-21 at 13 39 28

* For the case Ad/Organic ratio of `327/110`, we should have fill percent as `74.82...%` (`327/(327+110)*100`):
Screenshot 2023-06-21 at 13 44 29

## Pre-merge Checklist

### DIFF
--- a/client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
+++ b/client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
@@ -168,7 +168,9 @@ export default function CampaignItemDetails( props: Props ) {
 			value: views_organic || 0,
 		},
 	];
-	const databarTotal = databars.reduce( ( total, item ) => total + item.value, 0 );
+	const databarTotal = databars
+		.map( ( item ) => ( item.value > 0 ? item.value : 0 ) ) // Get only positive values.
+		.reduce( ( total, value ) => total + value, 0 );
 
 	const cancelCampaignButtonText =
 		status === 'active' ? __( 'Stop campaign' ) : __( 'Cancel campaign' );
--- a/client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
+++ b/client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
@@ -168,9 +168,11 @@ export default function CampaignItemDetails( props: Props ) {
 			value: views_organic || 0,
 		},
 	];
-	const databarTotal = databars
-		.map( ( item ) => ( item.value > 0 ? item.value : 0 ) ) // Get only positive values.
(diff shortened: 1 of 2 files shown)


client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
```diff
@@ -168,7 +168,9 @@ export default function CampaignItemDetails( props: Props ) {
 			value: views_organic || 0,
 		},
 	];
-	const databarTotal = databars.reduce( ( total, item ) => total + item.value, 0 );
+	const databarTotal = databars
+		.map( ( item ) => ( item.value > 0 ? item.value : 0 ) ) // Get only positive values.
```
sbarbosa: It's easier to have this code in one iteration. For example, you could move the check in the reduce like this:
```
.reduce( ( total, value ) => value > 0 ?  total + value : total, 0 );
```
Its the same as before, but without that extra iteration that the map function adds. 

Nonetheless, this is a nitpick of mine, there isn't any performance consideration in a two elements array. Fell free to ignore this.

sbarbosa (approved):
LGTM 👍. 

Left a nitpick comment, but is not a blocker for the PR. Feel free to ignore it 🙂

client/my-sites/promote-post-i2/components/campaign-item-details/index.tsx
therocket-gr: I think that @jjolmo has made some changes in the backend to not result in negative numbers.. but having a check in the front is also useful

timur987: Initially I thought that ternary operator will make it look more complex, but it look good. Thanks for the suggestion, @sbarbosa. Applied in  https://github.com/Automattic/wp-calypso/pull/78476/commits/39a2bbdc5ab260c28b4eec3cfd3bcead49774e24.

PR #78480, opened 2023-06-21
eslint-plugin-wpcalypso: Add i18n-unlocalized-url rule

yuliyan:
Related to #

## Proposed Changes

Add new rule `i18n-unlocalized-url` in `eslint-plugin-wpcalypso` to test whether the URL string should be wrapped in `localizeUrl` function call if it matches a list of URL patterns that we consider to be localizable.

## Testing Instructions

* Code review.
* Run rule unit tests `yarn run test-packages packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js`.
* Run `yarn lint:js` and confirm whether the errors for the new rule `wpcalypso/i18n-unlocalized-url` in the existing code base make sense.

## Pre-merge Checklist

### DIFF
--- a/packages/eslint-plugin-wpcalypso/.eslintrc.js
+++ b/packages/eslint-plugin-wpcalypso/.eslintrc.js
@@ -13,7 +13,7 @@ module.exports = {
 				'wpcalypso/i18n-no-this-translate': 'off',
 				'wpcalypso/i18n-no-variables': 'off',
 				'wpcalypso/i18n-translate-identifier': 'off',
-				'wpcalypso/i18n-unloclaized-url': 'off',
+				'wpcalypso/i18n-unlocalized-url': 'off',
 				'wpcalypso/redux-no-bound-selectors': 'off',
 				'wpcalypso/jsx-gridicon-size': 'off',
 				'no-restricted-imports': 'off',
--- a/packages/eslint-plugin-wpcalypso/.eslintrc.js
+++ b/packages/eslint-plugin-wpcalypso/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
 				'wpcalypso/i18n-no-this-translate': 'off',
 				'wpcalypso/i18n-no-variables': 'off',
 				'wpcalypso/i18n-translate-identifier': 'off',
+				'wpcalypso/i18n-unloclaized-url': 'off',
 				'wpcalypso/redux-no-bound-selectors': 'off',
 				'wpcalypso/jsx-gridicon-size': 'off',
 				'no-restricted-imports': 'off',
--- a/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
+++ b/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
@@ -164,7 +164,7 @@ module.exports = {
 	},
 	overrides: [
 		{
-			files: [ '**/test/*', '*.json' ],
+			files: [ '**/test/*', '*.json', '*.md' ],
 			rules: {
 				'wpcalypso/i18n-unlocalized-url': 'off',
 			},
--- a/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
+++ b/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
@@ -151,6 +151,7 @@ module.exports = {
 		'wpcalypso/i18n-mismatched-placeholders': 'error',
 		'wpcalypso/i18n-named-placeholders': 'error',
 		'wpcalypso/i18n-translate-identifier': 'error',
+		'wpcalypso/i18n-unlocalized-url': 'error',
 		'wpcalypso/jsx-gridicon-size': 'error',
 		'wpcalypso/jsx-classname-namespace': 'error',
 		'wpcalypso/redux-no-bound-selectors': 'error',
--- a/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
+++ b/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
@@ -164,7 +164,7 @@ module.exports = {
 	},
 	overrides: [
 		{
-			files: '**/test/*',
+			files: [ '**/test/*', '*.json' ],
 			rules: {
 				'wpcalypso/i18n-unlocalized-url': 'off',
 			},
--- a/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
+++ b/packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
@@ -162,4 +162,12 @@ module.exports = {
 		// Ensure our codebases use inclusive language
 		'inclusive-language/use-inclusive-words': 'error',
 	},
+	overrides: [
+		{
+			files: '**/test/*',
+			rules: {
+				'wpcalypso/i18n-unlocalized-url': 'off',
+			},
+		},
+	],
 };
--- a/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
+++ b/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
@@ -13,14 +13,10 @@ const LOCALIZABLE_URLS = [
 	'apps.wordpress.com',
 	'automattic.com/cookies(/|$)',
 	'automattic.com/privacy(/|$)',
-	'jetpack.com/?$',
-	'wordpress.com/?$',
 	'wordpress.com/blog(/|$)',
 	'wordpress.com/forums(/|$)',
 	'wordpress.com/go(/|$)',
 	'wordpress.com/support(/|$)',
-	'wordpress.com/theme(/|$)',
-	'wordpress.com/themes(/|$)',
 	'wordpress.com/tos(/|$)',
 ];
 
--- a/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
+++ b/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
@@ -1,5 +1,5 @@
 /**
- * @file @todo
+ * @file Disallow using unlocalized URL strings
  * @author Automattic
  * @copyright 2023 Automattic. All rights reserved.
  * See LICENSE.md file in root directory for full license.
@@ -9,7 +9,6 @@
 // Constants
 //------------------------------------------------------------------------------
 
-// @todo: should be replaced by url mapping from @automattic/i18n-calypso.
 const LOCALIZABLE_URLS = [
 	'apps.wordpress.com',
 	'automattic.com/cookies(/|$)',
--- a/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
+++ b/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
@@ -41,6 +41,21 @@ const LOCALIZABLE_URLS = [
 
 const getCallee = require( '../util/get-callee' );
 
+/**
+ * A helper funciton that would return the parent node which is relevant to the specific rule checks.
+ *
+ * @param   {Object} node
+ * @returns {Object}
+ */
+function getRelevantNodeParent( node ) {
+	// In case the node is operand in ternary or logical operator, return the grand parent.
+	if ( [ 'ConditionalExpression', 'LogicalExpression' ].includes( node.parent.type ) ) {
+		return getRelevantNodeParent( node.parent );
+	}
+
+	return node.parent;
+}
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -58,7 +73,7 @@ const rule = ( module.exports = function ( context ) {
 	 */
 	function handleUnlocalizedUrls( node, nodeValueString ) {
 		// Node is wrapped in localizeUrl, therefore we can assume it's localized and we don't need to do any further checks.
-		if ( getCallee( node.parent ).name === 'localizeUrl' ) {
+		if ( getCallee( getRelevantNodeParent( node ) ).name === 'localizeUrl' ) {
 			return;
 		}
 
@@ -68,7 +83,7 @@ const rule = ( module.exports = function ( context ) {
 		}
 
 		// URL string is assigned to a variable and variable is possibly later used in a localizeUrl call;
-		if ( node.parent.type === 'VariableDeclarator' ) {
+		if ( getRelevantNodeParent( node ).type === 'VariableDeclarator' ) {
 			variableDeclarationValues.push( node );
 		} else {
 			// Report unlocalized url.
@@ -82,7 +97,7 @@ const rule = ( module.exports = function ( context ) {
 	return {
 		'Program:exit': function () {
 			for ( const node of variableDeclarationValues ) {
-				if ( ! urlVariables.has( node.parent?.id?.name ) ) {
+				if ( ! urlVariables.has( getRelevantNodeParent( node )?.id?.name ) ) {
 					// Report unlocalized url.
 					context.report( {
 						node,
--- a/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
+++ b/packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
@@ -10,13 +10,28 @@
 //------------------------------------------------------------------------------
 
 const LOCALIZABLE_URLS = [
+	// apps.wordpress.com/**
 	/^(https?:)?\/\/apps\.wordpress\.com/i,
+
+	// automattic.com/cookies/
 	/^(https?:)?\/\/automattic\.com\/cookies\/?$/i,
+
+	// automattic.com/privacy/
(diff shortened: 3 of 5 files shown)


packages/eslint-plugin-wpcalypso/.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
 				'wpcalypso/i18n-no-this-translate': 'off',
 				'wpcalypso/i18n-no-variables': 'off',
 				'wpcalypso/i18n-translate-identifier': 'off',
+				'wpcalypso/i18n-unloclaized-url': 'off',
```
tyxla:
```suggestion
				'wpcalypso/i18n-unlocalized-url': 'off',
```

packages/eslint-plugin-wpcalypso/lib/configs/recommended.js
```diff
@@ -161,4 +162,12 @@ module.exports = {
 		// Ensure our codebases use inclusive language
 		'inclusive-language/use-inclusive-words': 'error',
 	},
+	overrides: [
+		{
+			files: [ '**/test/*', '*.json' ],
```
tyxla: Maybe we should also exclude markdown files?

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+	);
+}
+
+/**
+ * Check whether the provided node type is variable declarator.
+ *
+ * @param   {string}  type Node type
+ * @returns {boolean}
+ */
+function isVariableDeclarationValue( type ) {
```
tyxla: I'd totally inline this check since it's quite concise and used only once.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Check whether the provided string is localizable string.
+ *
+ * @param   {string}  string String to be tested
+ * @returns {boolean}
+ */
+function isLocalizableUrlString( string ) {
+	return LOCALIZABLE_URLS.some( ( url ) =>
+		new RegExp( `^(https?:)?//${ url.replace( '.', '\\.' ) }`, 'i' ).test( string )
```
tyxla: It may be safer to escape all regex characters, otherwise someone could add a URL with another special character and break the regex. 

I'd suggest either using a custom version like:

```
function escapeRegex(string) {
    return string.replace(/[/\-\\^$*+?.()|[\]{}]/g, '\\$&');
}
```

or if you **really have to**, use the Lodash `escapeRegExp` function.

tyxla: Perf nit: we create a new regex every time. I wonder if it's more straightforward to just create the regex objects inside `LOCALIZABLE_URLS` and use them directly. That's related to my comment above.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+ * See LICENSE.md file in root directory for full license.
+ */
+
+//------------------------------------------------------------------------------
+// Constants
+//------------------------------------------------------------------------------
+
+const LOCALIZABLE_URLS = [
+	'apps.wordpress.com',
+	'automattic.com/cookies(/|$)',
```
tyxla: It's ambiguous and unexpected that we're going to escape the dots, but we're going to keep the `|` and `$` literally inside the regex. May be more readable to declare the entire regex object right here.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const rule = ( module.exports = function ( context ) {
+	const variableDeclarationValues = [];
+	const urlVariables = new Set();
+
+	/**
+	 * Check whether the node is unlocalized url.
```
tyxla:
```suggestion
	 * Check whether the node is an unlocalized URL.
```

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+	 *
+	 * @param   {Object} node
+	 * @param   {string} nodeValueString
+	 * @returns {void}
+	 */
+	function handleUnlocalizedUrls( node, nodeValueString ) {
+		// Node is wrapped in localizeUrl, therefore we can assume it's localized and we don't need to do any further checks.
+		if ( getCallee( node.parent ).name === 'localizeUrl' ) {
```
tyxla: Does this work with aliases? There is at least one of those cases in the codebase:

https://github.com/Automattic/wp-calypso/blob/ce985467adb53ece4d878bfcfacbd120c1881ac1/packages/wpcom-template-parts/src/universal-footer-navigation/index.tsx#L3

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,123 @@
...
+			handleUnlocalizedUrls( node, node.value );
+		},
+		TemplateElement: function ( node ) {
+			const templateLiteralNode = node.parent;
+			handleUnlocalizedUrls( templateLiteralNode, node.value.raw );
+		},
+	};
+} );
+
+rule.ERROR_MESSAGE = "Url string should be wrapped in 'localizeUrl' function call";
```
tyxla:
```suggestion
rule.ERROR_MESSAGE = "URL string should be wrapped in a 'localizeUrl' function call";
```

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,68 @@
...
+//------------------------------------------------------------------------------
+
+const localizableUrls = [
+	'https://apps.wordpress.com/',
+	'https://automattic.com/cookies/',
+	'https://automattic.com/privacy/',
+	'https://wordpress.com/support/test-article/',
+	'https://wordpress.com/forums/test-thread/',
+	'https://wordpress.com/tos/',
+	'https://wordpress.com/tos/',
```
tyxla: Did we mean to duplicate this?

tyxla (commented):
Great work 👍  Thanks @yuliyan 🙌 

I've left some suggestions and some minor nits, let me know what you think.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
yuliyan: Aliases are currently used during the check. I came across the same use case during the development and thought about covering the `localizeUrl` aliases as well, but decided not to as it seems like a very rare use cases that could just be inline ignored if needed and also for the sake of keeping the implementation not to overcomplicated.

However, iterate on it later if it happens to be a more common case.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
yuliyan: I ended up using regex objects directly in the `LOCALIZABLE_URLS` definition.

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
yuliyan: Nope, that was an oversight. Removed it and updated the tests.

yuliyan:
Wow, thanks for the detailed review @tyxla 🙇I've addressed the feedback and updated the branch.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,116 @@
...
+ * @copyright 2023 Automattic. All rights reserved.
+ * See LICENSE.md file in root directory for full license.
+ */
+
+//------------------------------------------------------------------------------
+// Constants
+//------------------------------------------------------------------------------
+
+const LOCALIZABLE_URLS = [
+	// apps.wordpress.com/**
```
tyxla, line 13: Thanks for adding comments 👍

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,67 @@
...
+
+const RuleTester = require( 'eslint' ).RuleTester;
+const rule = require( '../i18n-unlocalized-url' );
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const localizableUrls = [
+	'https://apps.wordpress.com/',
```
tyxla: Care to add a few more fixtures to cover all scenarios? 

Some examples I can think of are:

* https://apps.wordpress.com/mobile/
* https://automattic.com/cookies (without the trailing slash)
* all other URLs that are intended to have a trailing slash, but will work without one too due to the [WordPress canonical redirects](https://developer.wordpress.org/reference/functions/redirect_canonical/).

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
tyxla: Sounds good 👍

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,67 @@
...
+	'https://automattic.com/cookies/',
+	'https://automattic.com/privacy/',
+	'https://wordpress.com/support/test-article/',
+	'https://wordpress.com/forums/',
+	'https://wordpress.com/tos/',
+];
+
+const nonLocalizableUrls = [ 'https://example.com', 'https://wordpress.com/forums/some-thread' ];
```
tyxla: Analogically to the localizable URLs, perhaps we can add a few more scenarios for the URLs that don't expect further paths, like `https://automattic.com/privacy`.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,116 @@
...
+	/^(https?:)?\/\/apps\.wordpress\.com/i,
+
+	// automattic.com/cookies/
+	/^(https?:)?\/\/automattic\.com\/cookies\/?$/i,
+
+	// automattic.com/privacy/
+	/^(https?:)?\/\/automattic\.com\/privacy\/?$/i,
```
tyxla: Should we support hashes? For some pages, we might want to lead to a particular section, for example:

https://automattic.com/privacy/#information-we-collect

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,116 @@
...
+
+	// wordpress.com/tos/
+	/^(https?:)?\/\/wordpress\.com\/tos\/?$/i,
+
+	// wordpress.com/blog/
+	/^(https?:)?\/\/wordpress\.com\/blog\/?$/i,
+
+	// wordpress.com/forums/
+	/^(https?:)?\/\/wordpress\.com\/forums\/?$/i,
```
tyxla: I guess we intentionally don't expect anyone to link to a particular forum thread? Perhaps we should envision that and add support for it?

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,116 @@
...
+	/^(https?:)?\/\/automattic\.com\/cookies\/?$/i,
+
+	// automattic.com/privacy/
+	/^(https?:)?\/\/automattic\.com\/privacy\/?$/i,
+
+	// wordpress.com/tos/
+	/^(https?:)?\/\/wordpress\.com\/tos\/?$/i,
+
+	// wordpress.com/blog/
+	/^(https?:)?\/\/wordpress\.com\/blog\/?$/i,
```
tyxla: Should we support particular blog links? I'm not sure if that makes sense though, since I assume each language of the blog has its own articles. Worth double-checking of course.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,116 @@
...
+
+	// wordpress.com/forums/
+	/^(https?:)?\/\/wordpress\.com\/forums\/?$/i,
+
+	// wordpress.com/go/**
+	/^(https?:)?\/\/wordpress\.com\/go($|\/.*)/i,
```
tyxla: Same as for `/blog`.

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
```diff
@@ -0,0 +1,67 @@
...
+	},
+} ).run( 'i18n-unlocalized-url', rule, {
+	valid: [].concat(
+		localizableUrls.reduce( ( cases, url ) => {
+			return cases.concat( [
+				`const url = localizeUrl( '${ url }' )`,
+				`const url = '${ url }'; localizeUrl( url );`,
+				`const link = <a href={ localizeUrl( '${ url }' ) }></a>;`,
+				`const url = '${ url }'; const link = <a href={ localizeUrl( url ) }></a>;`,
+				`const url = localizeUrl( \`${ url }?param=test\` );`,
```
tyxla, line 66: Do we support this ternary syntax? Should we add a test case for it?

https://github.com/Automattic/wp-calypso/blob/3c76a54cb4d25de3d3d7ee4ea5a925d54e168bf0/client/blocks/product-purchase-features-list/sell-online-paypal.jsx#L7-L11

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
tyxla: Looks good and the added comments make it extra readable - thanks!

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
yuliyan: Thanks for the suggestion. Added more fixtures in https://github.com/Automattic/wp-calypso/pull/78480/commits/2489f7a3be20ad3d3c8533b028a257e0972d013c which should cover pretty much all URLs that we match.

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
yuliyan: Added some more in https://github.com/Automattic/wp-calypso/pull/78480/commits/2489f7a3be20ad3d3c8533b028a257e0972d013c.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
yuliyan: Updated the regexes in https://github.com/Automattic/wp-calypso/pull/78480/commits/239d59532c243d75c104635a93a0a9749ce1faf1 to match hash and query strings.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
yuliyan: Forum threads only exist in the specific forums site (i.e. language) they were created in. Therefore, we only want to localize the root URL of the forums sites.

packages/eslint-plugin-wpcalypso/lib/rules/i18n-unlocalized-url.js
yuliyan: Blog articles URLs are non-localizable. Similarly to the forums, we only want to localize the home url of the blog site.

packages/eslint-plugin-wpcalypso/lib/rules/test/i18n-unlocalized-url.js
yuliyan: Good catch! I've added the handling for ternary and logical operators in https://github.com/Automattic/wp-calypso/pull/78480/commits/2805a428bed4a53322da2220c445e13b00efb1cd.

tyxla (approved):
Great work 🙌 thanks for addressing all the feedback 🚀 LGTM 🚢

PR #78491, opened 2023-06-21
Limited Global Styles: Fix notice in view canvas mode

mmtr:
## Proposed Changes

Fixes the Limited Global Styles notice displayed in the view canvas mode of the site editor, which was broken as of [Gutenberg 16.0](https://github.com/WordPress/gutenberg/pull/50567/files#diff-0cf5246119f61f0f26cb94918697d2cb3a0f40051045d24dd01c98241e79bde7) due to a change in the DOM.

Before | After
--- | ---
Screenshot 2023-06-21 at 14 05 24 | Screenshot 2023-06-21 at 15 18 57

## Testing Instructions

- Open the site editor.
- Make sure there is a notice at the bottom of the "view canvas mode" sidebar.
- Make sure the notice goes away when using the default style.

### DIFF
--- a/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.scss
+++ b/apps/editing-toolkit/editing-toolkit-plugin/wpcom-global-styles/notice.scss
@@ -1,5 +1,15 @@
+.edit-site-save-hub {
+	border-top: none;
+	padding-top: 0;
+}
+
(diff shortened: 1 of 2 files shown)


rcrdortiz (approved):
Works as expected. I would feel more confident if we could have a test or something so that we can see instantly when this component is going to not render the notice, but we can handle that in a follow up if needed.

PR #78495, opened 2023-06-21
New hosted site: dispatch Tracks event on suggestion fetch click

zaguiini:
## Proposed Changes

Let's see how many users click on the "Generate random name" button. Let's also understand whether the user has tried a custom name before clicking the button. This could be useful to decide if we want to keep the endpoint in the future, and even extend it to other flows that require a site name.

## Testing Instructions

Open `/setup/new-hosted-site` and pick any plan. Enable event logging in the DevTools.

Click the "Generate" button without changing the site name input

You should see the `calypso_signup_site_options_fetch_suggestion_click` event being dispatched with the following props:

```
source: site_title
has_changed_title: false
```

Change the input (e.g. add a character) and click "Generate" again. You should see the same event being dispatched with different prop values:

```
source: site_title
has_changed_title: true
```


mpkelly:
Completed test plan successfully 👍

PR #78496, opened 2023-06-21
Reader: Update popular sites sidebar component

eoigal:
This PR refactors the `ReaderSearchSidebar` component into `ReaderPopularSitesSidebar` component.

The main reason for this is that the component is used in Discover page and it doesn't make sense to be named with `search` so rename makes sense. Also we need a way to differentiate the follows from the sidebar on search and discover pages. This allows us to pass a `followSource` property.

### Test
Apply PR locally
Confirm the sidebars load as expected on Discover and Search page


Addison-Stavlo (approved):
works well and looks good on my end.

PR #78507, opened 2023-06-21
desktop: bump version to v8.0.0

### LINKED ISSUE #75452: Linux Desktop App: Update dependencies to remove gconf2
### Quick summary

A user pointed out that when they tried to install the Linux Desktop app, they were prompted about dependencies for gconf2, which has been deprecated and replaced with dconf. They're asking if we can update dependencies accordingly.

### Steps to reproduce

1. Install Linux Desktop 
2. Note the dependencies
3. Note gconf2 has been replaced and should be dconf instead if the app still needs that

Any dependencies should be current

### What actually happened

The app wants to install gconf2, and shouldn't because that's been replaced by dconf.

### Impact

Some (< 50%)

### Available workarounds?

No and the platform is unusable

Bug scrubbers: I'm not actually sure whether this makes the app unusable, but the user [in this thread](https://wordpress.com/forums/topic/linux-desktop-dependency-gconf-moved-renamed-at-gnome-org/) said they couldn't install the app.

Simple, Atomic, Self-hosted

### Logs or notes

Present in Linux Desktop App.

@mreishus  pointed out it was probably just added for Electron's sake and needs to be updated. @sandymcfadden  asked me to ping him and @nsakaimbo too  👋 

Apologies if this is in the wrong repo, entering it here anyway using our tooling, as it seemed like it might be related to [this Calypso code](fbhepr%2Suvfgbel%2Spnylcfb%2Sqrfxgbc%2Sryrpgeba%2Qohvyqre.wfba-og). Likely similar to [this issue](https://github.com/Automattic/simplenote-electron/issues/3048).

nsakaimbo:
## Proposed Changes

Prepare desktop app for release - bump major version.

Fixes #75452

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -429,7 +429,7 @@ jobs:
   # GitHub release and generate the release notes using the Git commits history
   wp-desktop-publish:
     docker:
-      - image: circleci/golang:1.12-node
+      - image: cimg/go:1.20-node
     working_directory: /home/circleci/wp-calypso
     environment:
       VERSION: << pipeline.git.tag >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,7 +439,7 @@ jobs:
           at: /home/circleci/wp-calypso
       - run:
           name: Install Dependencies
-          command: go get github.com/tcnksm/ghr
+          command: go install github.com/tcnksm/ghr@v0.16.0
       - run:
           name: Update wp-desktop repo README
           command: |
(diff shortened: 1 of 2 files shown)


worldomonation (approved):
Nice!

danielbachhuber:
Thanks @nsakaimbo !

worldomonation:
@nsakaimbo I have been informed in p1688130305943999-slack-C08GLCXMZ that a user on MacBook Air M1 2020 can't run the official 8.0.0 build. I was able to confirm on my M1 MacBook Pro of the same, and @Robertght also ran into the same thing on his M1 Pro (specific model unknown).

I'll take a look at why this might be this week, but it's looking like we might need to do another release.

PR #78522, opened 2023-06-22
Signup: skip the plans step in the domain transfer flow

aneeshd16:
Related to p1687402071427169-slack-C05CT832K2T

## Proposed Changes

* If the user is the `domain-transfer` signup flow, and selects the "Transfer" option, then skip the plans step.

## Testing Instructions

**Testing the transfer flow**
1. Go to `/start/domain-transfer`.
2. Enter any existing domain name. For the purposes of testing, you do not need to own that domain.
3. On the next step, select Transfer.
4. Confirm that the Plans step **is not** shown.
5. Confirm that you can reach Checkout and complete the purchase.

**Testing the connect flow**
1. Go to `/start/domain-transfer`.
2. Enter any existing domain name. For the purposes of testing, you do not need to own that domain.
3. On the next step, select Connect.
4. Confirm that the Plans step **is** shown.
5. Confirm that you can reach Checkout and complete the purchase.

## Pre-merge Checklist

### DIFF
--- a/client/lib/signup/step-actions/index.js
+++ b/client/lib/signup/step-actions/index.js
@@ -4,6 +4,7 @@ import {
 	planHasFeature,
 	FEATURE_UPLOAD_THEMES_PLUGINS,
 	isEcommerce,
+	isDomainTransfer,
 } from '@automattic/calypso-products';
 import { getUrlParts } from '@automattic/calypso-url';
 import { Site } from '@automattic/data-stores';
@@ -1180,7 +1181,7 @@ export function isAddOnsFulfilled( stepName, defaultDependencies, nextProps ) {
 }
 
 export function isPlanFulfilled( stepName, defaultDependencies, nextProps ) {
-	const { isPaidPlan, sitePlanSlug, submitSignupStep } = nextProps;
+	const { isPaidPlan, sitePlanSlug, submitSignupStep, flowName, signupDependencies } = nextProps;
 	const fulfilledDependencies = [];
 	const dependenciesFromDefaults = {};
 
@@ -1189,6 +1190,11 @@ export function isPlanFulfilled( stepName, defaultDependencies, nextProps ) {
 		fulfilledDependencies.push( 'themeSlugWithRepo' );
 		dependenciesFromDefaults.themeSlugWithRepo = defaultDependencies.themeSlugWithRepo;
 	}
+
+	const isTransferSelectedInDomainTransferFlow =
+		'domain-transfer' === flowName &&
+		signupDependencies?.domainItem &&
+		isDomainTransfer( signupDependencies.domainItem );
 
 	if ( isPaidPlan ) {
 		const cartItem = undefined;
@@ -1205,6 +1211,14 @@ export function isPlanFulfilled( stepName, defaultDependencies, nextProps ) {
 			{ cartItem, ...dependenciesFromDefaults }
 		);
 		recordExcludeStepEvent( stepName, defaultDependencies.cartItem );
+		fulfilledDependencies.push( 'cartItem' );
+	} else if ( isTransferSelectedInDomainTransferFlow ) {
+		const cartItem = null;
+		submitSignupStep(
+			{ stepName, cartItem, wasSkipped: true },
+			{ cartItem, ...dependenciesFromDefaults }
+		);
+		recordExcludeStepEvent( stepName, sitePlanSlug );
 		fulfilledDependencies.push( 'cartItem' );
 	}
 
--- a/client/signup/steps/domains/index.jsx
+++ b/client/signup/steps/domains/index.jsx
@@ -380,11 +380,6 @@ class DomainsStep extends Component {
 			Object.assign( { domainItem }, useThemeHeadstartItem )
 		);
 
-		// Skip the plans step if we are in the `domain-transfer` flow.
-		if ( 'domain-transfer' === this.props.flowName ) {
-			this.props.submitSignupStep( { stepName: 'plans', wasSkipped: true }, { cartItem: null } );
-		}
-
 		this.props.goToNextStep();
 	};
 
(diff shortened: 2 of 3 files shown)


client/signup/steps/domains/index.jsx
```diff
@@ -380,6 +380,11 @@ class DomainsStep extends Component {
 			Object.assign( { domainItem }, useThemeHeadstartItem )
 		);
 
+		// Skip the plans step if we are in the `domain-transfer` flow.
```
chriskmnds: Not sure how accurate this description is. Both "transfer" and "connect" look like are part of the same `domain-transfer` flow, one shows the plans while the other doesn't.

client/signup/steps/domains/index.jsx
```diff
@@ -380,6 +380,11 @@ class DomainsStep extends Component {
...
+		if ( 'domain-transfer' === this.props.flowName ) {
```
chriskmnds: Similarly, something feels odd here. Is this condition even necessary?

Also, I recall "skipping a step" in this framework was more complicated than just submitting the step from a previous step. I feel this needs more testing. What happens if you go into `/start/domain-transfer/plans` directly or in any sort of clickthrough fashion ("go back" from browser, etc). I recall we needed to create a different/sibling flow and redirect to it on a couple of occasions for similar "skip step" scenario (I can find some links if it comes to it).

chriskmnds (commented):
I tested but I'm not sure if everything works clearly, and what is relevant to these changes. I can confirm plans step is skipped in Transfer if coming out of the domains step, if that's enough.

> Testing the **connect** flow

Does it make sense to show the Free plan in this case? It also looks broken to me: selecting Free still adds the Premium plan to the cart in the end every time (tried clearing the cart, went back, selected Free, still Premium in the end).

> Testing the transfer flow

I tried once and got an error here (not sure why, but something about "invalid"). Then tried again and now see:

Screenshot 2023-06-22 at 12 43 42 PM

It seems this needs fixing too. It basically blocks domain transfer in case of error (that may not even be originating from domain owner)?

----------------------

> It seems this needs fixing too. It basically blocks domain transfer in case of error (that may not even be originating from domain owner)?

^ I guess user expected to check status and contact support in that case. I guess all good in that case.

chriskmnds:
Also took me ages to find a domain to use for "transfer", and I have run out of them by now. I can't test this anymore it seems.

client/signup/steps/domains/index.jsx
chriskmnds: I think I kinda get it now. This callback could fire from other flows, right? All good in that case.

aneeshd16: I skipped mentioning the "connect" flow in this comment since we are inside the `handleAddTransfer` method, which handles the transfer part only.

client/signup/steps/domains/index.jsx
aneeshd16: I did recall the same thing, also having used `useBranchSteps` before. However, in this case, I took inspiration from `maybeExcludeEmailsStep`: https://github.com/Automattic/wp-calypso/blob/90882ceb39d4284a278d24af3d85a7e4428de6b5/client/lib/signup/step-actions/index.js#L1053-L1086

This function also includes this two extra lines which I missed adding here:
```
	recordExcludeStepEvent( stepName, siteUrl );

	flows.excludeStep( stepName );
```

I'll dig into what these do and add them if necessary.

aneeshd16:
Thanks for the review @chriskmnds!

> Does it make sense to show the Free plan in this case? It also looks broken to me: selecting Free still adds the Premium plan to the cart in the end every time (tried clearing the cart, went back, selected Free, still Premium in the end).

Good catch! I could reproduce this as well. Let's look into this issue separately.

Apart from the next steps listed here (https://github.com/Automattic/wp-calypso/pull/78522#discussion_r1238346528), do you think there is anything else we should tackle before merging this PR?

chriskmnds (approved):
Thanks for addressing comments @aneeshd16 

> Apart from the next steps listed here (https://github.com/Automattic/wp-calypso/pull/78522#discussion_r1238346528), do you think there is anything else we should tackle before merging this PR?

I think all good. You seem confident about the "skip step" part. I can't quite connect with what we needed to do previously that seemed that much more difficult - I believe it was an experiment variation without a step, and not in response to user input? 🤔 

Nothing else seems related to these changes.

chriskmnds:
> Good catch! I could reproduce this as well. Let's look into this issue separately.

Totally. Adding Premium to cart might even be intentional there. Definitely worth raising up separately 😬

aneeshd16:
Hi @chriskmnds, I found a more robust and testable way to skip the plans step, can you please review it again? Sorry for not finding this earlier!

chriskmnds:
> Hi @chriskmnds, I found a more robust and testable way to skip the plans step, can you please review it again? Sorry for not finding this earlier!

argh I'm spending more time looking for a domain to use :-) 
I keep hitting:

Screenshot 2023-06-22 at 2 59 16 PM

client/lib/signup/step-actions/index.js
```diff
@@ -1206,6 +1212,14 @@ export function isPlanFulfilled( stepName, defaultDependencies, nextProps ) {
...
 		recordExcludeStepEvent( stepName, defaultDependencies.cartItem );
 		fulfilledDependencies.push( 'cartItem' );
+	} else if ( isTransferSelectedInDomainTransferFlow ) {
+		const cartItem = null;
+		submitSignupStep(
+			{ stepName, cartItem, wasSkipped: true },
+			{ cartItem, ...dependenciesFromDefaults }
+		);
+		recordExcludeStepEvent( stepName, sitePlanSlug );
+		fulfilledDependencies.push( 'cartItem' );
```
chriskmnds, line 1222: Looks more robust indeed. A lot of things to be put together. Makes me wonder how different/simpler such scenario would be in Stepper :-)

PR #78528, opened 2023-06-22
CI: add retry to Calypso E2E and Gutenberg E2E specs.

worldomonation:
Related to pciE2j-2cY-p2.

## Proposed Changes

This PR adds a single retry of only failed tests.

## Testing Instructions

Locally, I modified an existing test to fail intentionally on first pass by throwing an exception halfway through.
If the environment variable `RETRY` is defined, then it will not throw an exception, allowing the step to pass.

I ran the first pass and it failed as expected.
Then, on the second pass, I add the `--onlyFailures` flag and defined the env var `RETRY` to allow the step to pass.
The second pass:
- only ran the failing test suite, as expected
- passed all steps

<details>
    <summary>Logs</summary>

```
➜  ~/workspace/wp-calypso-second/test/e2e git:(ci/e2e-add-retry) ✗ yarn jest test/e2e/specs/infrastructure/infrastructure__ssr-enabled.ts test/e2e/specs/me/me__smoke.ts
  console.log
    Intentionally failing...

      at Object.<anonymous> (specs/infrastructure/infrastructure__ssr-enabled.ts:23:15)

Artifacts for infrastructure__ssr-enabled: /Users/edwintakahashi/workspace/wp-calypso-second/test/e2e/results/infrastructure__ssr-enabled__2023-06-22_09-08-44-745-GmNF2y
 FAIL  specs/infrastructure/infrastructure__ssr-enabled.ts
  Server-side Rendering
    Check SSR endoint: log-in
      ✓ Check SSR: log-in (2378 ms)
    Check SSR endoint: themes
      ✕ Check SSR: themes (143 ms)
    Check SSR endoint: theme/twentytwentythree
      ○ skipped Check SSR: theme/twentytwentythree

  ● Server-side Rendering › Check SSR endoint: themes › Check SSR: themes



      22 | 					if ( endpoint === 'themes' ) {
      23 | 						console.log( 'Intentionally failing...' );
    > 24 | 						throw new Error();
         | 						      ^
      25 | 					}
      26 | 				}
      27 | 				await page.goto( DataHelper.getCalypsoURL( endpoint ), { timeout: 20 * 1000 } );

      at Object.<anonymous> (specs/infrastructure/infrastructure__ssr-enabled.ts:24:13)

 PASS  specs/me/me__smoke.ts (7.452 s)
  Me: Smoke Test
    ✓ Navigate to /me (809 ms)
    ✓ Navigate to Me > Account Settings (79 ms)
    ✓ Navigate to Me > Purchases (164 ms)
    ✓ Navigate to Me > Security (248 ms)
    ✓ Navigate to Me > Privacy (85 ms)
    ✓ Navigate to Me > Notification Settings (114 ms)
    ✓ Navigate to Me > Blocked Sites (264 ms)
    ✓ Navigate to Me > Apps (129 ms)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 1 skipped, 9 passed, 11 total
Snapshots:   0 total
Time:        8.215 s, estimated 11 s
Ran all test suites matching /test\/e2e\/specs\/infrastructure\/infrastructure__ssr-enabled.ts|test\/e2e\/specs\/me\/me__smoke.ts/i.
➜  ~/workspace/wp-calypso-second/test/e2e git:(ci/e2e-add-retry) ✗ RETRY=1 yarn jest test/e2e/specs/infrastructure/infrastructure__ssr-enabled.ts test/e2e/specs/me/me__smoke.ts --onlyFailures
 PASS  specs/infrastructure/infrastructure__ssr-enabled.ts (8.72 s)
  Server-side Rendering
    Check SSR endoint: log-in
      ✓ Check SSR: log-in (1971 ms)
    Check SSR endoint: themes
      ✓ Check SSR: themes (2326 ms)
    Check SSR endoint: theme/twentytwentythree
      ✓ Check SSR: theme/twentytwentythree (3635 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        8.972 s
Ran all test suites matching /test\/e2e\/specs\/infrastructure\/infrastructure__ssr-enabled.ts|test\/e2e\/specs\/me\/me__smoke.ts/i.
```
</details>

## Pre-merge Checklist

### DIFF
--- a/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
+++ b/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
@@ -204,6 +204,9 @@ open class E2EBuildType(
 			// Don't fail if the runner exists with a non zero code. This allows a build to pass if the failed tests have been muted previously.
 			nonZeroExitCode = false
 
+			// Support retries using the --onlyFailures flag in Jest.
+			supportTestRetry = true
+
 			// Fail if the number of passing tests is 50% or less than the last build. This will catch the case where the test runner crashes and no tests are run.
 			failOnMetricChange {
 				metric = BuildFailureOnMetric.MetricType.PASSED_TEST_COUNT
--- a/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
+++ b/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
@@ -145,8 +145,14 @@ open class E2EBuildType(
 					cd test/e2e
 					mkdir temp
 
+					# Disable exit on error to support retries.
+					set +o errexit
+
 					# Run suite.
 					xvfb-run yarn jest --reporters=jest-teamcity --reporters=default --maxWorkers=%JEST_E2E_WORKERS% --group=$testGroup
+
+					# Restore exit on error.
+					set -o errexit
 
 					# Retry failed tests only.
 					xvfb-run yarn jest --reporters=jest-teamcity --reporters=default --maxWorkers=%JEST_E2E_WORKERS% --group=$testGroup --onlyFailures
--- a/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
+++ b/.teamcity/_self/lib/customBuildType/E2EBuildType.kt
@@ -147,6 +147,9 @@ open class E2EBuildType(
 
 					# Run suite.
 					xvfb-run yarn jest --reporters=jest-teamcity --reporters=default --maxWorkers=%JEST_E2E_WORKERS% --group=$testGroup
+
+					# Retry failed tests only.
+					xvfb-run yarn jest --reporters=jest-teamcity --reporters=default --maxWorkers=%JEST_E2E_WORKERS% --group=$testGroup --onlyFailures
 				"""
 				dockerImage = "%docker_image_e2e%"
 				dockerRunParameters = "-u %env.UID% --shm-size=4g"
--- a/packages/calypso-e2e/src/jest-playwright-config/environment.ts
+++ b/packages/calypso-e2e/src/jest-playwright-config/environment.ts
@@ -152,6 +152,11 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
 				let contextIndex = 1;
 
 				const artifactFilename = `${ this.testFilename }__${ sanitizeString( this.failure.name ) }`;
+				const datetime = new Date()
+					.toISOString()
+					.replace( /T/, '_' )
+					.replace( /\..+/, '' )
+					.replace( /:/g, '-' );
 
 				for await ( const context of contexts ) {
 					let pageIndex = 1;
@@ -164,7 +169,7 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
 					await context.tracing.stop( { path: traceFilePath } );
 
 					for await ( const page of context.pages() ) {
-						const pageName = `${ artifactFilename }__${ contextIndex }-${ pageIndex }`;
+						const pageName = `${ artifactFilename }__${ datetime }__${ contextIndex }-${ pageIndex }`;
 						// Define artifact filename.
 						const mediaFilePath = path.join( this.testArtifactsPath, pageName );
 
--- a/packages/calypso-e2e/src/jest-playwright-config/environment.ts
+++ b/packages/calypso-e2e/src/jest-playwright-config/environment.ts
@@ -110,20 +110,10 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
 		this.testArtifactsPath = await fs.mkdtemp(
 			path.join( env.ARTIFACTS_PATH, `${ this.testFilename }__${ date }-` )
 		);
-		const logFilePath = path.join( this.testArtifactsPath, `${ this.testFilename }.log` );
 
 		// Start the browser.
 		const browser = await browserType.launch( {
 			...config.launchOptions,
-			logger: {
-				log: async ( name: string, severity: string, message: string ) => {
-					await fs.appendFile(
-						logFilePath,
-						`${ new Date().toISOString() } ${ process.pid } ${ name } ${ severity }: ${ message }\n`
-					);
-				},
-				isEnabled: ( name ) => name === 'api',
-			},
 		} );
 
 		// Set up the proxy trap.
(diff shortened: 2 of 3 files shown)


WunderBart:
That was fast! ⚡ It looks good to me. I'm assuming that if all tests passed in the first run, the second attempt would say `nothing to test` or something of that sort?

dpasque:
@WunderBart 

> I'm assuming that if all tests passed in the first run, the second attempt would say nothing to test or something of that sort?

Yep! 🙂  It looks like this:
```
No failed test found.
Run Jest without `--onlyFailures` or with `--all` to run all tests.
```

dpasque:
@worldomonation Awesome!! Thank you SO much! 😄 This was exactly what I was thinking, and I really appreciate you knocking this out!

I think the one thing we might be missing is in the Build definition --> failureConditions --> We need to set `supportTestRetry` to `true` -- this helps the builds run green if multiple tests are run and have different outputs:
https://teamcity.jetbrains.com/app/dsl-documentation/root/failure-conditions/support-test-retry.html

worldomonation:
> @worldomonation Awesome!! Thank you SO much! 😄 This was exactly what I was thinking, and I really appreciate you knocking this out!
> 
> I think the one thing we might be missing is in the Build definition --> failureConditions --> We need to set `supportTestRetry` to `true` -- this helps the builds run green if multiple tests are run and have different outputs: https://teamcity.jetbrains.com/app/dsl-documentation/root/failure-conditions/support-test-retry.html

Thanks for the heads up, I will try adding that into the conditions. Once ready I'll ask for a re-review.

dpasque:
@worldomonation -- I took a look at the E2E runs in CI, and I noticed that it _should_ have done a retry, but it didn't!

I didn't realize this before, but after poking around in the Admin view, it looks like all TeamCity bash scripts get `set -o errexit` set by default, so the script will quit execution anytime it hits a non-zero code.

So, I think we have to put the second rerun into its own bash script step, right? Otherwise it will never run! Because if the first run fails, it will return a 1, which will exit the script before the rerun happens.

Sorry I missed that initially! 🙇

worldomonation:
> @worldomonation -- I took a look at the E2E runs in CI, and I noticed that it _should_ have done a retry, but it didn't!
> 
> I didn't realize this before, but after poking around in the Admin view, it looks like all TeamCity bash scripts get `set -o errexit` set by default, so the script will quit execution anytime it hits a non-zero code.
> 
> So, I think we have to put the second rerun into its own bash script step, right? Otherwise it will never run! Because if the first run fails, it will return a 1, which will exit the script before the rerun happens.
> 
> Sorry I missed that initially! 🙇

Hmm, that makes sense. 

Either we'd need a new step, or to `set +e` before the `yarn jest ...` part. 

[Reference](https://intoli.com/blog/exit-on-errors-in-bash-scripts/)

.teamcity/_self/lib/customBuildType/E2EBuildType.kt
```diff
@@ -145,8 +145,17 @@ open class E2EBuildType(
 					cd test/e2e
 					mkdir temp
 
+					# Disable exit on error to support retries.
+					set +o errexit
```
worldomonation, line 149: @noahtallen @WunderBart @dpasque **comment:blocking** - while I understand there's nothing more permanent than a temporary solution, I am not comfortable adding a build step (and the associated overhead, however minor) to all Pre-Release Tests and Calypso E2E builds. 

What do you think? Would you both prefer to have a separate step for clear separation?

worldomonation:
Looks like the retry is working as of the most recent revision:

image

[Check out this build for an example](https://teamcity.a8c.com/buildConfiguration/calypso_calypso_WebApp_Calypso_E2E_Playwright_desktop/10270345?buildTab=log&focusLine=0&linesState=245.855.856.857.860.923.924.927.933.1289.1290.1291.1343.1344.1345.1348&logView=flowAware).

worldomonation:
However, there is a new issue: if the same spec fails twice in a row, the artifact name from Run 2 clashes with the one from Run 1, causing `xargs -r -0 mv -t screenshots` to fail. This in turn results in the trace and log files not being copied.

image

Looking at the web, it seems this is a method to safe the user from self-destruction: https://stackoverflow.com/q/4669420

I reproduced the issue locally by making a spec intentionally fail and then re-trying:
image

----

Seeing that there is no apparent workaround on the web for this issue, the best approach is to either
1. change the structure of the artifact (instead of throwing everything into the top-level folder Artifacts > Screenshots, allow it to  be in individual subfolders);
2. add timestamp to the file name, so it doesn't clash;
3. `set +e` on the artifact collection step so that errors don't stop the process.

1 is most involved solution for something that may not stick around for too long, given we've sort-of-agreed on moving to PTR.
2 is a simple change and if we use date and time strings instead of a timestamp, it will be clear when it was generated.
3 is a hacky solution that would be easy to implement.

.teamcity/_self/lib/customBuildType/E2EBuildType.kt
noahtallen: My main question is just whether there is all that much overhead in an extra step -- it doesn't seem to take that much time in my experience, unless there is a different issue here.

That said, I don't think it matters that much. It's only really useful to have an extra step for people who debug the test suites a lot (e.g. us), so if you don't think it makes a big difference, I don't think we need to care about it too much

dpasque: I'm fine either way too!

> it doesn't seem to take that much time in my experience, unless there is a different issue here.

This was my experience too, but again, I don't think it really matters whether we make a separate build step or toggle the flags like you have it 😄

dpasque:
> However, there is a new issue: if the same spec fails twice in a row, the artifact name from Run 2 clashes with the one from Run 1, causing xargs -r -0 mv -t screenshots to fail. This in turn results in the trace and log files not being copied.

Oh shoot, I hadn't thought of that!

I think of the options you presented, I like number 2.

Really, I think what we want is just a way to...
1. Record artifacts for all failures, even if they pass in subsequent reruns (this helps with flaky debugging).
2. Easily distinguish for failing tests the artifacts from the first or second run.

I think time string is fine, or we could honestly just do something like have an ENV var called "RETRY_NUMBER" and when we do the rerun command, set `RETRY_NUMBER=1`, and then we could use that value in the artifact name (maybe append 'retry-1' or something). I think that would be pretty simple too, and would maybe make it a bit clearer when looking at the artifacts which is which. But that said, I think the time string is fine too to just get this rolling! 😄

worldomonation:
@dpasque Implemented fix in https://github.com/Automattic/wp-calypso/pull/78528/files#diff-d33147995e808c2f248624314bf232597704a415a95e3df79f653ea40f626a41R172.

worldomonation:
Yet another problem: I think it would be nice to denote that a build had failures, but passed on retry. How could I do that?

Solution: a build tag is useful because it's entirely contained within TeamCity, is flexible and can be applied to any build. 

However, TeamCity being TeamCity, there isn't a way to do this unless one hacks a REST API call from a build step. You can see that I was able to post a tag on a still-running build:
image

Sandbox build: https://teamcity.a8c.com/admin/editBuildRunners.html?id=buildType%3ATeamcitySandbox_EdwinTest

**What's the issue then?**
For prototyping, I think we can use a CLI token tied to my TeamCity username that I will generate. We don't know if this is required once Playwright Test Runner is in the picture, because that test runner likely does its own tracking of retries.

But, should my account on TeamCity be terminated for any reason, then the step will stop working until the access token is replaced with another one. So if possible, it would be best to create an access token using a Systems-managed bot user.

packages/calypso-e2e/src/jest-playwright-config/environment.ts
```diff
@@ -110,20 +110,10 @@ class JestEnvironmentPlaywright extends NodeEnvironment {
...
 			...config.launchOptions,
-			logger: {
-				log: async ( name: string, severity: string, message: string ) => {
-					await fs.appendFile(
-						logFilePath,
-						`${ new Date().toISOString() } ${ process.pid } ${ name } ${ severity }: ${ message }\n`
-					);
-				},
-				isEnabled: ( name ) => name === 'api',
-			},
```
worldomonation, line 126: I've chosen to remove the capturing of `.log` files due to a persisting clash in the log filenames _if there was a retry_.

1. because the file capture is started as part of the browser launch, it has no context on whether the test step will pass/fail. 
2. if the build required a retry, it will generate a new log file with the same name.
3. content captured in the log file isn't very useful due to https://github.com/microsoft/playwright/issues/5871. Besides, trace files have supplanted debugging needs.

dpasque (approved):
Btw - dropping an approval here so you can merge when you feel fit! 👍 I'm good with all the changes and approaches you've laid out, and don't think we need to worry about tagging builds for the moment. I think we can just get something simple going that starts turning more builds green! 😄

PR #78531, opened 2023-06-22
Reader: Update discover stream to use compact cards

eoigal:
This is a follow up to https://github.com/Automattic/wp-calypso/pull/78411 that makes it easier to use compact cards (that also replace the post cards on tag pages).

This PR sets the `useCompactCards` property and updates the Recommended post stream to use the `ReaderPostCardAdapter` component rather than the `EmptySearchRecommendedPost` component that was been used up to now.

### Test
* Apply patch locally
* Confirm the post stream for the `Recommended` and other tag look the same.


Addison-Stavlo (approved):
Looks good on my end!

PR #78533, opened 2023-06-22
Ensure Credits tab has its right margin

timur987:
Related to SELFSERVE-553
Discussion: p1686559561135739-slack-C053FAERCDP

We want to make Tabs Navigation within Tools > Advertising accessible from mobile devices.

## Proposed Changes

* Hide dropdown menu (it isn't used but might be shown),
* Make sure the Credits tab is aligned to the right for mobile screen too,
* For small mobile screens (less than 364px; the number was chosen according to the current tab's content) make sure we can slide to tabs that no fully visible due to small screen size:

## Testing Instructions

* Check out to this branch and build assets
* Open Tools > Advertising
* Try various screen widths from `1300px` down to `364px` and make sure tabs are fully visible
* Check screen width of `364px` (this screen width should have all tabs visible
Screenshot 2023-06-22 at 15 53 47

* Make screen smaller (e.g. `320px`)
* Ensure clicking to "Credits" tabs leaves `16px` margin to the right from it:

https://github.com/Automattic/wp-calypso/assets/115007291/f32a0922-ede7-42ee-ae4a-39d0e25afa84

## Pre-merge Checklist


client/my-sites/promote-post-i2/style.scss
```diff
@@ -256,7 +265,7 @@ body.is-section-promote-post-i2 {
 			margin-right: 0;
 
 			@media screen and (max-width: $break-medium) {
-				margin: 0;
+				margin: 0 0 0 auto; // Align Credits tab to the right
```
sbarbosa, line 274: I checked in the designs, and the mobile version of the tabs shows the credits next to the other two. 
Maybe we can start a conversation with Bart regarding this? I am not sure what alignment is correct here.

client/my-sites/promote-post-i2/style.scss
```diff
@@ -471,6 +480,29 @@ $break-medium-expanded-menu: $break-medium + 272px;
 	}
 }
 
+// Tabs tweak for mobile screens smaller than 364px (the mentioned width can contain all tabs and margins)
+@media screen and (max-width: 364px) {
+	body.is-section-promote-post-i2 .layout__content .main.is-wide-layout {
+		.section-nav-tabs {
+			.section-nav-tabs__list {
+				width: 348px; // Equal to 364px - left margin of 16px to make sure it will add the right margin
```
sbarbosa: I would try to go with an approach using margins or padding. If we use width here, it won't work as we expect in languages (or if the person has a lot of credits/posts/campaigns).

sbarbosa (commented):
Hi! I tested the PR, but It didn't work for me, I wasn't able to see extra space next to the credits section.



I left some comments; I think we may need to find a different solution without using width here.

client/my-sites/promote-post-i2/style.scss
timur987: Yep, you're right. I'll try to find another solution then.

timur987: Thanks to @bak's  suggestion (within the thread of p1687458785783789-slack-C053FAERCDP) the solution is much more easier than I thought before. Please check it out, @sbarbosa.

sbarbosa: I tested the solution and worked fine. But with that last addition of the overflow hidden, we lost the possibility to do manual scrolls.

But, checking on where the overflow was, I think the initial solution would work nice if we just moved it up a level.
Before, the overflow was at a `.section-nav` level. But, if we move that to the same place you put your overflow, now the container already has the correct padding, and the scroll works correctly.

Let me give you a patch with the change:
[Scroll.patch](https://github.com/Automattic/wp-calypso/files/11884579/Scroll.patch)

It looks like this (in the same file, under the section-nav class, that code is found):
Screenshot 2023-06-27 at 12 39 19

With that, I think I got the correct behavior:

https://github.com/Automattic/wp-calypso/assets/1258162/9c7a1c8a-818b-4fe9-9361-b872d862d77f

Let me know what you think. We can go with your proposed solution if we don't want the manual scroll. But I think it could be nice to have both of them (in case some languages have really long labels for the navigation).

timur987: Thanks @sbarbosa! I've applied your patch and it works like a charm!

sbarbosa (approved):
Nice! It worked like a charm! Thanks!

PR #78537, opened 2023-06-22
Theme Showcase: Hide the Activate Button for Dotorg Themes on Sites on Lower Plans

Copons:
## Proposed Changes

* Hide the Dotorg themes activate button on Free, Personal, Premium sites for logged-in users.
* Logged-out users will still see the Activate button, pointing to the login form.

This is a tentative fix for a regression, possibly introduced in #76837, which extended the primary CTA for Dotorg themes to logged-out users.

It seems that the change leaked to logged-in users, where the button would unsuccessfully attempt to start the theme installation process.

| Before | After |
|--------|--------|
| Screenshot 2023-06-22 at 16 20 25 | Screenshot 2023-06-22 at 16 20 22 | 

### Note

I'd love a review from both @Automattic/t-rex and @Automattic/lego to ensure this is actually what we want, and I haven't misinterpreted the situation.

cc @retnonindya who found the bug in the first place.

## Testing Instructions

* Logged-in and with a Free site selected...
  * Navigate to the Theme Showcase and search for a Dotorg theme (e.g. Astra).
  * Ensure the Activate button is not displayed.
* Pick a Business site.
  * Search for a Dotorg theme again.
  * Ensure the Activate button is displayed without regressions.
* Logged-out...
  * Navigate to the Theme Showcase and search for a Dotorg theme (e.g. Astra).
  * Ensure the Activate button is displayed without regressions.

### DIFF
--- a/client/my-sites/theme/main.jsx
+++ b/client/my-sites/theme/main.jsx
@@ -635,11 +635,16 @@ class ThemeSheet extends Component {
 			showTryAndCustomize,
 			isExternallyManagedTheme,
 			isWporg,
+			isLoggedIn,
 		} = this.props;
 		const placeholder = <span className="theme__sheet-placeholder">loading.....</span>;
 		const title = name || placeholder;
 		const tag = author ? translate( 'by %(author)s', { args: { author: author } } ) : placeholder;
-		const shouldRenderButton = ! retired && ! isWPForTeamsSite && ! this.shouldRenderForStaging();
+		const shouldRenderButton =
+			! retired &&
+			! isWPForTeamsSite &&
+			! this.shouldRenderForStaging() &&
+			( ! this.hasWpOrgThemeUpsellBanner() || ! isLoggedIn );
 
 		return (
 			<div className="theme__sheet-header">


client/my-sites/theme/main.jsx
```diff
@@ -635,11 +635,16 @@ class ThemeSheet extends Component {
...
 		} = this.props;
 		const placeholder = <span className="theme__sheet-placeholder">loading.....</span>;
 		const title = name || placeholder;
 		const tag = author ? translate( 'by %(author)s', { args: { author: author } } ) : placeholder;
-		const shouldRenderButton = ! retired && ! isWPForTeamsSite && ! this.shouldRenderForStaging();
+		const shouldRenderButton =
+			! retired &&
+			! isWPForTeamsSite &&
+			! this.shouldRenderForStaging() &&
+			( ! this.hasWpOrgThemeUpsellBanner() || ! isLoggedIn );
```
okmttdhr, line 647: badge

May I ask why `hasWpOrgThemeUpsellBanner` is needed also?

rcrdortiz, line 647: Wouldn't it be semantically more accurate to use a method/function like isUserOnFreePlan in this context? The current implementation doesn't make it immediately clear that showing the button depends on whether the user is on a free plan. Instead, it gives an impression that it's based on whether we're displaying the org upsell banner or not. Could we clarify this?

rcrdortiz (approved):
Works as advertised, I'll approve this PR but you have a few non-blocking comments.

client/my-sites/theme/main.jsx
mmtr: That controls the presence of the "Access this third-party theme with the Business plan!" banner:

Screenshot 2023-06-23 at 09 36 47

We want to hide the button when that banner is visible.

okmttdhr: Ahh, a logged-out user also has that banner, so we combined `( ! this.hasWpOrgThemeUpsellBanner() || ! isLoggedIn )`. Thank you for your comment!

My concern was described in [another comment](https://github.com/Automattic/wp-calypso/pull/78537#discussion_r1239460031), but non-blocking!

okmttdhr (approved):

client/my-sites/theme/main.jsx
mmtr: > it gives an impression that it's based on whether we're displaying the org upsell banner or not.

That's precisely the intent. The button needs to be hidden for WP.org themes if the site has a Free, Personal, or Premium plan, which are the conditions needed for the WP.org banner to be displayed.

rcrdortiz: The proposed changes in this PR only reference the FREE plan.
> Hide the Dotorg themes activate button on free sites for logged-in users.

I can confirm that the implemented behaviour is that the button is being displayed if the user has a paid plan and is only hidden when the user has a free plan. Which makes sense since if you are on a paid plan you can install the theme.

mmtr: > Which makes sense since if you are on a paid plan you can install the theme.

Only on the Business or eCommerce plans 🤔 . WP.org themes cannot be installed if the site has a Personal or a Premium plan. If the button is visible for those plans, that's definitely unintended.

rcrdortiz: I see, thanks for the clarification! The button is only visible on plans that can install themes.

Copons: Ugh, apologies for the confusion.

I'll rename the PR title, because, as you've already noticed, the Activate button is not supposed to show on Free/Personal/Premium, and only on Business/Ecommerce.

Copons: Also FWIW, I agree with @rcrdortiz that the function is poorly named — or, it was well-named, but then it was used for counter-intuitive reasons.

I've chosen not to proceed with a rename to keep this short and to restore the original functionality (regressed in #76837).

PR #76535, opened 2023-05-03
WooExpress: Add specific plan types for Essential and Performance plan

valterlorran:
Related to #76205

## Proposed Changes

The current implementation of WooExpress plans uses the `TYPE_ECOMMERCE` attribute in the plan information. However, this has caused an issue with the `getPlanSlugForTermVariant` function, which is always returning the term for the Performance plan even for the Essential plan. Moreover, it's recommended that the plans have their own definition of a type, rather than relying on a single type for different plans.

* Adds the **TYPE_WOOEXPRESS_SMALL** type for the Essential plan
* Adds the **TYPE_WOOEXPRESS_MEDIUM** type for the Performance plan

We are working on the impacts of this change here: paYKcK-2U9-p2

## Testing Instructions

* Run the tests for the `plan-lookups.js` file:
```bash
yarn jest 'packages/calypso-products/test/plan-lookups.js' -c 'packages/calypso-products/jest.config.js'
```

#### My Plan page
* Navigate to `/plans/my-plan/{SITE-SLUG}` on a site with the Essential or the Performance plan.
* Make sure the plan features are displayed
* Make sure the text is displayed in the `Video and audio posts` section. The storage limit and the name of the plan will be wrong, but this issue is being addressed here: #76515

#### Analytics
Run the unit tests and verify that it passes:
```bash
yarn run test-client client/lib
```

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
[...]

PR #78540, opened 2023-06-22
Checkout: Show ecommerce features for Woo Express plans

michaeldcain:
In #76535, new product types were added for the Woo Express plans. This caused Checkout to only display the default plan features, since previously Woo Express plans passed the `isWpComEcommercePlan()` check.

This PR adds a new check for `isWooExpressPlan()` to the Checkout features, in order to return the same list as other ecommerce plans.

See: pbOQVh-3sY-p2#comment-4146

Screenshot 2023-06-22 at 12 53 06 PM

**To test:**
- visit Checkout with a Woo Express plan `/checkout/{site_slug}/wooexpress-medium-monthly`
- verify that you see ecommerce features in the "Included with your purchase" list

### DIFF
--- a/client/my-sites/checkout/composite-checkout/lib/get-plan-features.ts
+++ b/client/my-sites/checkout/composite-checkout/lib/get-plan-features.ts
@@ -1,6 +1,7 @@
 import { isEnabled } from '@automattic/calypso-config';
 import {
 	isMonthly,
+	isWooExpressPlan,
 	isWpComProPlan,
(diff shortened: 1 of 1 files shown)


martian77:
Screenshot 2023-06-23 at 14 48 07

I think I only get a 4-day refund because of the domain being added (I forgot I had something in the cart for this site already!). Once I removed that, it matches the expected list. 👍

martian77 (approved):

PR #78554, opened 2023-06-22
Checkout: Convert RegionAddressFieldsets to TypeScript

sirbrillig:
## Proposed Changes

This converts the `RegionAddressFieldsets` component to TypeScript.

## Testing Instructions

TBD

### DIFF
--- a/client/components/domains/contact-details-form-fields/custom-form-fieldsets/region-address-fieldsets.tsx
+++ b/client/components/domains/contact-details-form-fields/custom-form-fieldsets/region-address-fieldsets.tsx
@@ -1,4 +1,4 @@
-import { localize } from 'i18n-calypso';
+import { localize, LocalizeProps } from 'i18n-calypso';
 import { includes } from 'lodash';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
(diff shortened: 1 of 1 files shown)


sirreal:
I cherry-picked https://github.com/Automattic/wp-calypso/pull/78571 here, it should demonstrate that the fix there works.

client/components/domains/contact-details-form-fields/custom-form-fieldsets/region-address-fieldsets.tsx
```diff
@@ -37,7 +40,7 @@ export class RegionAddressFieldsets extends Component {
 		hasCountryStates: false,
 	};
```
jsnajdr, line 43: The issue with `localize` is not caused by bad HOC types per se, but there seems to be some weird interaction between the TypeScript type for the component props (`RegionAddressFieldsetsProps`) and the `defaultProps` declaration. The React types try to infer the component type also from the `propTypes` and `defaultProps` fields. There's a type called `ReactManagedAttributes` for that. And it's the mixture of these inferred types and the declared `RegionAddressFieldsetsProps` that leads to these `never`'s. The `RegionAddressFieldsets` component doesn't declare `propTypes`, only `defaultProps`, and that's probably what confuses the inference helper. @sirreal's patch removes the confusion, but I have no idea why and would have to spend hours to figure it out.

I think the very best solution would be to rewrite `RegionAddressFieldsets` to a function component with hooks. Get rid of the legacy `defaultProps`, and also of the `localize` HOC. HOCs are notoriously difficult, if not impossible, to type. Typing the function component is going to be very easy.

sirbrillig: Thanks for the insight! I'll see if we can convert it. I'm going to assign this to @spraveenitpro for now.

PR #78555, opened 2023-06-22
Reader: fix RSS feed lookup

eoigal:
Fixes https://github.com/Automattic/loop/issues/108

This PR tries to handle RSS feed lookups better in the Reader search page.

The issue above looks up the following URLs (that should have feeds);

* nealmueller.com
* daringfireball.net

The endpoint doesn't handle this URLs properly and either no results are found or the results are only related because that URL was found referenced in a post on a site elsewhere.

There are a few things to note here;

* The search query returns a feed if a feed is parsed from the query URL
* The feed may not already exist on our backend (no feed ID found) but the user should still be able to follow this feed
* If the search query finds a feed, we will show the `SearchFollowButton` component 
Screenshot 2023-06-23 at 00 10 32
* if the search query doesn't find a feed but it still looks like a feed URL, we'll still show the `SearchFollowButton` component
* If no feed found and the search query doesn't look like a feed URL, then we don't show the `SearchFollowButton` 

### Before

https://github.com/Automattic/wp-calypso/assets/5560595/370752ef-c44a-4ea0-b494-f44729b152f4

### After

https://github.com/Automattic/wp-calypso/assets/5560595/803e6bde-ed8a-4c26-ab56-69fedcd2057d

### Testing
* Apply PR locally
* Apply patch code-D114309 on sandbox
* Go to http://calypso.localhost:3000/read/search and lookup the 2 URLs above - confirm the UX works as you might expect.
* Test that the search works with other URLs 
* Test search works with non URL queries
* Anything else you can think of!

### DIFF
--- a/client/reader/search-stream/search-follow-button.jsx
+++ b/client/reader/search-stream/search-follow-button.jsx
@@ -3,7 +3,7 @@ import { localize } from 'i18n-calypso';
 import { some } from 'lodash';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
-import { resemblesUrl, withoutHttp, addSchemeIfMissing } from 'calypso/lib/url';
+import { resemblesUrl, withoutHttp, addSchemeIfMissing, urlToDomainAndPath } from 'calypso/lib/url';
 import ReaderFollowFeedIcon from 'calypso/reader/components/icons/follow-feed-icon';
 import ReaderFollowingFeedIcon from 'calypso/reader/components/icons/following-feed-icon';
 import FollowButton from 'calypso/reader/follow-button';
@@ -63,7 +63,7 @@ class SearchFollowButton extends Component {
 		// If we find a feed then set the feed object
 		let feed;
 		if ( resemblesUrl( query ) ) {
-			feed = feeds?.find( ( f ) => f.feed_URL.includes( query ) );
+			feed = feeds?.find( ( f ) => f.feed_URL.includes( urlToDomainAndPath( query ) ) );
 		}
 
 		// If no feed found, then don't show the follow button
@@ -71,10 +71,8 @@ class SearchFollowButton extends Component {
 			return null;
 		}
 
-		// We can use the feed to create a follow button
-
 		// If already following this feed then don't show the follow button
-		if ( feed.is_following !== undefined && feed.is_following === true ) {
+		if ( feed.is_following === true ) {
 			return null;
 		}
 
--- a/client/reader/search-stream/search-follow-button.jsx
+++ b/client/reader/search-stream/search-follow-button.jsx
@@ -3,19 +3,17 @@ import { localize } from 'i18n-calypso';
 import { some } from 'lodash';
 import PropTypes from 'prop-types';
 import { Component } from 'react';
-import { connect } from 'react-redux';
 import { resemblesUrl, withoutHttp, addSchemeIfMissing } from 'calypso/lib/url';
 import ReaderFollowFeedIcon from 'calypso/reader/components/icons/follow-feed-icon';
 import ReaderFollowingFeedIcon from 'calypso/reader/components/icons/following-feed-icon';
 import FollowButton from 'calypso/reader/follow-button';
-import { getReaderAliasedFollowFeedUrl } from 'calypso/state/reader/follows/selectors';
 import { commonExtensions } from 'calypso/state/reader/follows/selectors/get-reader-aliased-follow-feed-url';
 import './style.scss';
 
 class SearchFollowButton extends Component {
 	static propTypes = {
 		query: PropTypes.string,
-		feed: PropTypes.object,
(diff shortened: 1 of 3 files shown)


roo2:
I think this is caused by D114309-code, when I search for things that aren't URLs I'm still getting suggestions e.g.

I suspect that maybe "prepare_url" is working too well? 😆

Screen Shot 2023-06-23 at 11 50 44 am

client/reader/search-stream/site-results.jsx
```diff
@@ -82,11 +82,14 @@ export default connect(
 		} );
 
 		// Check if searchResults has one item and if it has a feed_ID
-		if ( searchResults && searchResults.length === 1 ) {
+		if ( searchResults && searchResults.length > 0 ) {
```
Addison-Stavlo: In general i feel like we are making some very subtle changes in these files. This seems like the most substantial? Is this where we think the regression happened: a change to the backend resulting in more than 1 result in these cases causing this to get bypassed?

Also note the comment above could be updated now.

Addison-Stavlo:
This is testing pretty well on my end now.  There is still an awkward stage of showing "no results found" for a few seconds when searching a URL like in the examples before it updates to show the feed and follow button, unsure how we should handle that currently.

client/reader/search-stream/site-results.jsx
```diff
@@ -81,14 +81,42 @@ export default connect(
...
-				feed = getFeed( state, feed.feed_ID );
-			}
-			ownProps.onReceiveSearchResults( feed );
+		// Check if searchResults has any feeds
+		if ( searchResults && searchResults.length > 0 ) {
+			const feeds = searchResults;
+			// We want to create a list of unique feeds based on searchResults
+			// We need to do this because the search results may contain duplicate feeds URLs with different http or https schemes
+			// eslint-disable-next-line wpcalypso/redux-no-bound-selectors
+			const feedResults = feeds.reduce( ( uniqueFeeds, feed ) => {
```
roo2, line 90: hmm I wonder if there are other areas where this will be a problem? It looks like the "manage following" page also uses the feed search, as does the edit page for custom lists. I think we may need to move this logic either to the back end or somewhere it can be applied for all of these cases

client/reader/search-stream/site-results.jsx
```diff
@@ -81,14 +81,42 @@ export default connect(
...
+				// Strip out the URL scheme for subscribe_URL
+				const strippedSubscribeURL = feed.subscribe_URL.replace( /^https?:\/\//, '' );
+
+				// Check if the array already has an item with the same strippedSubscribeURL or feed_ID
+				const foundItem = uniqueFeeds.find(
+					// eslint-disable-next-line wpcalypso/redux-no-bound-selectors
+					( uniqueFeed ) =>
+						( ! feed.feed_ID &&
+							uniqueFeed.subscribe_URL.replace( /^https?:\/\//, '' ) === strippedSubscribeURL ) ||
```
roo2: I was able to get a NPE because subscribe_URL was null. This was when I searched for "longreads", I believe longreads.com does have both an HTTP and HTTPS version

roo2 (approved):

roo2:
I added a small fix to support searches with the full http**s** protocol. Previously, the _s_ in https was causing the feed_URL not to be matched

client/reader/search-stream/search-follow-button.jsx
```diff
@@ -44,22 +47,55 @@ class SearchFollowButton extends Component {
...
+		// If no feed found, then don't show the follow button
+		if ( ! feed ) {
+			return null;
 		}
 
+		// We can use the feed to create a follow button
+
 		// If already following this feed then don't show the follow button
-		if ( isFollowing ) {
+		if ( feed.is_following !== undefined && feed.is_following === true ) {
```
roo2: I think we should remove this logic. As it currently stands, if you search for a feed that you are already following, you will see neither the site result with the follow button nor the normal search results.

client/reader/search-stream/search-follow-button.jsx
```diff
@@ -44,22 +47,53 @@ class SearchFollowButton extends Component {
...
+		if ( resemblesUrl( query ) ) {
+			feed = feeds?.find( ( f ) => f.feed_URL.includes( urlToDomainAndPath( query ) ) );
+		}
 
-		if ( feed && feed.name?.length ) {
-			isFollowing = feed.is_following;
-			followTitle = feed.name;
+		// If no feed found, then don't show the follow button
+		if ( ! feed ) {
+			return null;
```
roo2, line 71: In the future it would be good to have a "no results found" state returned by this component. If you search for a non-existent *.wordpress.com blog, the backend returns `https://wordpress.com/blog/feed/` which results in a state like this:

Screen Shot 2023-06-26 at 4 31 28 pm

PR #69242, opened 2022-10-19
Global Styles: Show notice in privacy settings

rcrdortiz:
This PR implements the UI changes necessary for displaying a notice in the privacy settings when our users are using Global Styles and are required to upgrade their plan in order to make the customization changes they've made available publicly.

I've decided not to show the notice when the site is not launched because of two main reasons. This isn't specified in the design and I think this wasn't taken into consideration, the second reason is to reduce friction when launching a site, I experimented a bit with the UI and it becomes cluttered with buttons and we aren't providing a clear path that our users should follow, which could lead to a decrease in the number of sites that get launched. 

Also, when a user launches a free site, he will already get prompted to purchase a domain and a plan.

Captura de Pantalla 2022-10-19 a las 10 12 03

#### Proposed Changes

* We now check if the user has GS in use and if we should show a notice.
* If the user qualifies, we show a notice that includes a button so that the user can upgrade his plan.

#### Testing Instructions

- [MAKE SURE TO INSTALL THE ETK OF THIS BRANCH](https://github.com/Automattic/wp-calypso/pull/69242#issuecomment-1285541319).
- MAKE SURE YOU ADD THE STICKER BLOG TO THE SITES YOU'LL WORK WITH.

#### Happy Case
* Go to the general settings screen, privacy section. http://calypso.localhost:3000/settings/general/{siteSlug}
* Make sure your site is `FREE` and that you have some GS customizations applied.
   * When the site is not launched you should not see a notice
   * When the site is launched you should see a notice.
      * Make sure the buttons on the notice work as expected.
* Remove or reset your customizations.
* You should not see the notice.

#### Mobile devices
* Repeat the happy case using a mobile viewport (just the part where you interact with the notice).

#### Test un-launched site
* Create a site and make sure to not launch it.
* Make sure it has the corresponding blog sticker.
* Go to the general settings screen, privacy section.
* You should not see the notice.

#### Test tracking event when clicking on the upgrade button.
* Go through the Happy Case
* Open tracks and search for `calypso_global_styles_gating_%`
* You should see some events named `calypso_global_styles_gating_settings_notice_upgrade_click`

#### Test network error when fetching global styles info
* Block the global styles info request in Chrome's inspector and reload.
* You should see no console errors.
* You should see no errors on the screen.
* You should see the last value cached by `useQuery`.

#### Test for regressions
* Accent color should continue to work as before. See this [PR](https://github.com/Automattic/wp-calypso/pull/68841) for testing instructions

### Test evidence
[...]

PR #78576, opened 2023-06-23
Limited Global Styles: Generalize badges for custom styles

### LINKED ISSUE #78638: Clicking a style variation disc in the showcase doesn't add premium label
### Quick summary

Clicking a style variation in the design picker adds the “★ Premium” label to show it's a paid-for feature, while in the logged-in theme showcase it doesn’t do anything.

### Steps to reproduce

1. Go to wordpress.com/themes
2. Find a theme that has style variations
3. Click on one of the little discs underneath

The thumbnail to change
A  “★ Premium” label to appear as happens in the onboarding design picker

### What actually happened

The thumbnail changed but the label remained as "Free"

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

https://github.com/Automattic/wp-calypso/assets/93301/3997ae34-853f-4b94-8379-25e0d74b331a

mmtr:
Fixes https://github.com/Automattic/dotcom-forge/issues/2856
Fixes https://github.com/Automattic/wp-calypso/issues/78638

## Proposed Changes

- Changes the theme type badges so _locked_ style variations are labeled with "★ Upgrade" instead of "★ Premium" which might incorrectly infer they need a Premium plan.
- For the same reason, the "Premium style" title has been removed from the tooltip, since just the tooltip message should be enough.
- Additionally, this also fixes the badges in the Theme showcase so _locked_ style variations are correctly labeled.

Section | Before | After
--- | --- | ---
Design picker | Screenshot 2023-06-23 at 13 11 33 | Screenshot 2023-06-28 at 10 55 54
Theme showcase | Screenshot 2023-06-28 at 10 56 49 | Screenshot 2023-06-28 at 10 57 16

## Testing Instructions

- Find a free theme with style variations.
- Click on a disc that represents a custom style variation.
- For sites with Limited Global Styles, make sure the badge changes to "★ Upgrade" 
- For sites with full Global Styles, make sure it remains as "Free".
- For sites with Limited Global Styles, make sure the tooltip no longer displays "Premium style" in the title.
- For sites with full Global Styles, make sure there is no tooltip.
- Find now a premium theme with style variations.
- Click on a disc that represents a custom style variation.
- Make sure the badge is still "★ Premium".
- Make sure the tooltip title still says "Premium theme".
- Repeat the steps in the theme showcase (both while logged-in and logged-out, but note that you cannot full GS if logged-out).


Copons:
Thanks for the proactivity @mmtr, but let's wait for feedback on the audit (paYJgx-3wz-p2). 🙂

client/components/theme-type-badge/index.tsx
```diff
@@ -19,56 +19,51 @@ import './style.scss';
 
 interface Props {
 	canGoToCheckout?: boolean;
-	forcePremium?: boolean;
+	isLockedStyleVariation?: boolean;
```
rcrdortiz, line 22: Maybe we could change the name of this variable so that it's more similar to our product wording. I think a good name could be `isUpgradeRequiredForStyleVariation `

client/components/theme-type-badge/index.tsx
```diff
@@ -19,56 +19,51 @@ import './style.scss';
...
 
 	useEffect( () => {
-		if ( type === FREE_THEME ) {
+		if ( type === FREE_THEME && ! isLockedStyleVariation ) {
 			return;
 		}
 		recordTracksEvent( 'calypso_upgrade_nudge_impression', {
 			cta_name: 'theme-upsell',
 			theme: themeId,
 		} );
```
rcrdortiz, line 45: I found this part a bit odd, I know it's not strictly part of this review, but wouldn't it be more simple to:
```suggestion
		if ( type !== FREE_THEME && isLockedStyleVariation ) {
			recordTracksEvent( 'calypso_upgrade_nudge_impression', {
				cta_name: 'theme-upsell',
				theme: themeId,
			} );
		}
```

client/components/theme/index.jsx
```diff
@@ -344,3 +350,8 @@ export default connect(
 	},
 	{ recordTracksEvent, setThemesBookmark, updateThemes }
 )( localize( Theme ) );
+
+export default ( props ) => {
+	const { shouldLimitGlobalStyles } = usePremiumGlobalStyles( props.siteId );
+	return <ConnectedTheme { ...props } shouldLimitGlobalStyles={ shouldLimitGlobalStyles } />;
+};
```
rcrdortiz, line 357: Is there a reason why we're not including the `shouldLimitGlobalStyle` prop in the `connect` call?

Wouldn't the following DIFF also allow us to add the `shouldLimitGlobalStyle` prop without introducing an additional level of indirection? Or am I missing something?

```DIFF
diff --git a/client/components/theme/index.jsx b/client/components/theme/index.jsx
index 8a94fb6cab..7fc2f0e850 100644
--- a/client/components/theme/index.jsx
+++ b/client/components/theme/index.jsx
@@ -332,12 +332,13 @@ export class Theme extends Component {
        }
 }
 
-const ConnectedTheme = connect(
+export default connect(
        ( state, { theme, siteId } ) => {
                const {
                        themes: { themesUpdate },
                } = state;
                const { themesUpdateFailed, themesUpdating, themesUpdated } = themesUpdate;
+               const { shouldLimitGlobalStyles } = usePremiumGlobalStyles( siteId );
                const isExternallyManagedTheme = getIsExternallyManagedTheme( state, theme.id );
 
                return {
@@ -346,12 +347,8 @@ const ConnectedTheme = connect(
                        isUpdated: themesUpdated && themesUpdated.indexOf( theme.id ) > -1,
                        isExternallyManagedTheme,
                        siteSlug: getSiteSlug( state, siteId ),
+                       shouldLimitGlobalStyles,
                };
        },
        { recordTracksEvent, setThemesBookmark, updateThemes }
 )( localize( Theme ) );
-
-export default ( props ) => {
-       const { shouldLimitGlobalStyles } = usePremiumGlobalStyles( props.siteId );
-       return <ConnectedTheme { ...props } shouldLimitGlobalStyles={ shouldLimitGlobalStyles } />;
-};
(END)

```

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
```diff
@@ -278,20 +278,13 @@ const UnifiedDesignPickerStep: Step = ( { navigation, flow, stepName } ) => {
...
-		themeId: string,
-		forcePremium: boolean,
-		tooltipHeader: string,
-		tooltipMessage: string
-	) => (
+	const getBadge = ( themeId: string, isLockedStyleVariation: boolean ) => (
 		<ThemeTypeBadge
 			canGoToCheckout={ false }
-			forcePremium={ forcePremium }
+			isLockedStyleVariation={ isLockedStyleVariation }
```
rcrdortiz, line 284: There is a high level of property propagation in the current architecture. The unified-design-picker resolves the isLockedStyleVariation and then passes it to getBadge -> ThemeTypeBadge -> tooltip, which causes a tight level of coupling and increases the maintenance costs since, if/when the property needs to be changed, we will need to modify more than one part of the codebase, which goes against the Single Responsibility principle.

I know it's a bit out of scope since it isn't a product change, but a better approach would be to have individual badge components for each type: wpcom upgrade, wpcom, and wporg. We can convert the getBadge function from a builder method to a factory method that returns the corresponding badge. This way, we can reduce the overall complexity of the architecture, improve readability since the badge wouldn't need to perform as much logic, and if the conditions for showing one badge or another change, we wouldn't need to modify more than one file across the stack.

I'm not expecting immediate action on this, especially since all you did was change the name of the variable haha. But even then, just look at the number of places you needed to change the name. This comment just aims to surface and raise awareness of the current issues we have with the architecture of our codebase and why we have high maintenance costs.

client/components/theme-type-badge/index.tsx
mmtr: I considered that too, but couldn't think of a better name than `isLockedStyleVariation`.

To me `isUpgradeRequiredForStyleVariation` infers that the site needs an upgrade to use some style variations, but it's not clear whether the badge is currently used for those variations or not and if you'd need to check something else. A more clear name would be `isStyleVariationThatRequiresUpgrade` but I prefer `isLockedStyleVariation` because it's shorter.

By using `isLockedStyleVariation` as name, I'm explicitly telling the component that the badge will be for a style variation that needs to be unlocked by upgrading. And we're using the lock/unlock terminology in the UX, so it's aligned to some extent.

client/components/theme-type-badge/index.tsx
mmtr: I always find harder to read code that is indented due to `if` statements, so (if possible) I have the preference of inverting the condition and return early to keep the indentation shallow.

See pbOQVh-29D-p2#keep-indentation-shallow-and-avoid-else

client/components/theme/index.jsx
mmtr: Reacts hooks can be used only on functional components. You actually had to do something similar in https://github.com/Automattic/wp-calypso/pull/69242/files#diff-bc79a829db32a21a70056887cf8a6bd4ebce6890c1a8fd3091258dba0610b680R809 😄

client/components/theme-type-badge/index.tsx
rcrdortiz: That's usually the case when more than one thing is being checked. In this specific case, there is only one condition/outcome.

client/components/theme/index.jsx
rcrdortiz: Then we could add the hook inside the actual component, right?

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
mmtr: > The unified-design-picker resolves the isLockedStyleVariation and then passes it to getBadge -> ThemeTypeBadge -> tooltip, which causes a tight level of coupling and increases the maintenance costs

Totally agree.

> a better approach would be to have individual badge components for each type: wpcom upgrade, wpcom, and wporg

Actually, I think a simpler approach, which aligns better with how React apps are typically developed, would be to store the current selected style variation in the state, and read that state from the tooltip component, so we don't need to pass the info as a prop.

client/components/theme/index.jsx
mmtr: Nope, because it's a class component, not a functional component.

client/components/theme-type-badge/index.tsx
mmtr: Yeah, the problem is more obvious where there are more than one check, but even with just one I still prefer an early return and avoid the extra indentation. I find it easier to read.

client/landing/stepper/declarative-flow/internals/steps-repository/design-setup/unified-design-picker.tsx
rcrdortiz: > Actually, I think a simpler approach, which aligns better with how React apps are typically developed, would be to store the current selected style variation in the state, and read that state from the tooltip component, so we don't need to pass the info a prop.

These suggestions are not mutually exclusive and could potentially be combined. We could implement the factory method for creating badge components and also lift the state up to reduce prop drilling. This would reduce complexity and increase the maintainability of the codebase.

rcrdortiz (approved):
Tests well.

PR #78586, opened 2023-06-23
Domains: add link to top bar

zaguiini:
Related to p1687387772424839-slack-C029GN3KD

## Proposed Changes

This PR adds the "Domains" link alongside "My Sites" in the top bar.

| Desktop | Mobile |
| ---------| -------|
| image | image |

## Testing Instructions

1. Open Calypso and verify that clicking the "Domains" link sends the user to `/domains/manage`;
2. Verify that the icon is showing without the label on mobile;
3. Verify that the section becomes active (different, lighter background color) in the top bar.

### DIFF
--- a/client/sections.js
+++ b/client/sections.js
@@ -236,7 +236,7 @@ const sections = [
 		name: 'domains',
 		paths: [ '/domains' ],
 		module: 'calypso/my-sites/domains',
-		group: 'sites',
+		group: 'domains',
 	},
 	{
 		name: 'incoming-redirect',
--- a/client/layout/masterbar/logged-in.jsx
+++ b/client/layout/masterbar/logged-in.jsx
@@ -152,6 +152,10 @@ class MasterbarLoggedIn extends Component {
 		}
 	};
 
+	clickDomains = () => {
+		this.props.setNextLayoutFocus( 'content' );
+	};
+
 	clickReader = () => {
 		this.props.recordTracksEvent( 'calypso_masterbar_reader_clicked' );
 		this.handleLayoutFocus( 'reader' );
@@ -175,6 +179,10 @@ class MasterbarLoggedIn extends Component {
 
 	preloadMySites = () => {
 		preload( this.props.domainOnlySite ? 'domains' : 'stats' );
+	};
+
+	preloadDomains = () => {
+		preload( 'domains' );
 	};
 
 	preloadReader = () => {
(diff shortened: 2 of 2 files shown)


client/sections.js
```diff
@@ -236,7 +236,7 @@ const sections = [
 		name: 'domains',
 		paths: [ '/domains' ],
 		module: 'calypso/my-sites/domains',
-		group: 'sites',
+		group: 'domains',
```
zaguiini, line 239: This change was necessary otherwise `section` and `sectionGroup` are always `sites`, making it more difficult to add the active background color to the link.

Changing this causes the `sites` group not to be loaded (not exactly a bad thing if the user is coming from /sites) but introduces some delay if the user is already browsing a site in Calypso.

Let me know if we are against this change, and give me a potential solution 😄

mpkelly: An alternative would be to use the section `name` instead, but we'd have to update the other items, as without this change, the `My Sites` is active even when I am at `/domains/manage`.  

But given `domains` is now top-level, I think it makes sense that it is its own `group`. So this change is probably fine, but let's see what the other reviewers think.

mpkelly:
Tested on desktop and mobile. It's working fine for me, but this PR needs multiple reviews, so I will let others review it.

client/sections.js
p-jackson: The other thing is going to do is remove the `.is-group-sites` class from the `<body>` and replace it with `.is-group-domains`. These group classes are used to apply styles just for a particular group of sections in the app.

Afaict we aren't depending on any of the `.is-group-sites` styles so it should be fine to move.

However I'm think we should actually keep `/domains` in the `sites` group for a different reason. I'll put together a video to demonstrate what I mean.

p-jackson:
Personally I think if you navigate to the domains page via a site's sidebar (so you're just seeing a site-specific domain page) then the Domains link in the toolbar shouldn't become active. It's confusing because when the Domains link is active sometimes you see a full sidebar and sometimes you only see the "all sites" sidebar.

I've got an alternative approach which doesn't split Domains out into it's own section group, and instead the link in the toolbar is really just a hard coded special link: https://github.com/Automattic/wp-calypso/pull/78611

dzver:
Hey, I would like to learn more about the purpose of the change?

Wouldn't it be better for such a high-visibility link, if approved, to point to the new domain search instead? The all-sites domain management can be exposed on /me, and within /domains/add, if needed.

zaguiini:
This PR is probably going to be superseded by https://github.com/Automattic/wp-calypso/pull/78611.

PR #78501, opened 2023-06-21
Hide long data format via tooltip

timur987:
Related to SELFSERVE-613

## Proposed Changes

* Make the date format shorted, and
* Hide the long version under a tooltip.

## Testing Instructions

* Run Calypso locally and opens Tools > Advertising.
* Ensure date format looks the same as in [Figma](https://www.figma.com/file/uvh2KtY65b7hGBHI9exbJ1/BlazePress?type=design&node-id=8572-76417&t=vSRZCWOBuy4WLzwg-0).

date-tooltip

* (Optional) To limit unit tests to run only the new tests you can adjust [`package.json:116`](https://github.com/Automattic/wp-calypso/blob/trunk/package.json#L116) to become:
```json
"test-client": "TZ=UTC jest -c=test/client/jest.config.js -i client/my-sites/promote-post-i2/test/utils.test.tsx",
```

* Run unit tests to make sure the new test is working fine by running (if there's a way to run specified unit tests without changing `package.json` please let me know):
```sh
yarn run test-client
``` 

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [x] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
- [ ] Have you used memoizing on expensive computations? More info in [Memoizing with create-selector](https://github.com/Automattic/wp-calypso/blob/trunk/packages/state-utils/src/create-selector/README.md) and [Using memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors) and [Our Approach to Data](https://github.com/Automattic/wp-calypso/blob/trunk/docs/our-approach-to-data.md)
- [ ] Have we added the "[Status] String Freeze" label as soon as any new strings were ready for translation (p4TIVU-5Jq-p2)?
- [ ] For changes affecting Jetpack: Have we added the "[Status] Needs Privacy Updates" label if this pull request changes what data or activity we track or use (p4TIVU-ajp-p2)?

PR #78596, opened 2023-06-23
Fixes the relative date shown in Promote post

sbarbosa:
Related to SELFSERVE-628

## Proposed Changes

This PR fixes the relative time shown on the Advertising page. It uses a similar function to the one used on the posts page.
This way, it will feel normal to the users.

**NOTE:** This PR needs to be merged after this diff: D114400. If not, it will show incorrect relative times. 
We already are showing an incorrect time, but this change it make it more noticeable,

Screenshot 2023-08-10 at 14 25 27

Screenshot 2023-08-10 at 14 19 31

## Testing Instructions

* Open the live preview and navigate to Tools->Advertising
* Verify that the correct relative time is shown when the difference is less than a week

## Pre-merge Checklist


timur987:
Thanks for the changes, @sbarbosa. They work as advertised but I doubt about the way on how we describe the age of the post. Last time I used the same approach as you but then I got a [comment](https://github.com/Automattic/wp-calypso/pull/78501#issuecomment-1601882470) from @kthai15 with a suggestion to use the same approach for dashboard as in the widget (not as it was done in Calypso before). I see you returned back showing the age as it was in Calypso.

That's why I suggest to be on the same page on how we should show the age of the post.

jjolmo (approved):
LGTM! Tested locally
image

kthai15:
What I was trying to say what that the relative date should be consistent between dashboard and widget. I wasn't sure of the exact logic of how it constructs this date in the Calypso dashboard, so I invented some arbitrary logic in the widget.

```
// dateDiff is in days
{dateDiff === 0 && translate("hours ago")}
{dateDiff === 1 && translate("1 day ago")}
{dateDiff > 1 && isWithinCurrentYear && moment(date).format("MMM DD")}
{dateDiff > 1 && !isWithinCurrentYear && moment(date).format("MMM DD, YYYY")}
```

I'm sure this differs from the logic currently in the dashboard, so let us know what we need to change in the widget for it to behave in the same manner. 😄

sbarbosa:
Thanks @timur987 for letting me know about that discussion. 
I will start a slack discussion related to this, and we can come up with a solution on both ends.

I created this bug/PR because the "hours ago" didn't sound correct for me, but I may be wrong. 
I prefer to be a little more specific, like 1 hour ago, 1 minutes ago. Maybe we can get a common ground and don't go past minutes. 

But I think we use moment in both applications, and the fromNow method already solves this problem, so it may be a good idea to apply this same change in the widget.

I will leave this PR open for now until we decide how the final solution would be.

jjolmo:
I've created a purposal in the Widget side using this strategy and the last notes added by Bart: check PR no 331

sbarbosa:
To match the changes made in the widget, I added a change to show `Now` instead of `a few seconds ago`.

PR #78580, opened 2023-06-23
Subscription Management: Rework tracks foundation, add tracks event for subscribe, unsubscribe, site_url/icon/title clicked.

yansern:
Fixes https://github.com/Automattic/wp-calypso/issues/78487

## Proposed Changes
* Introduce a base tracks event function:
  * recordSubscriptionsTracksEvent( name, props )
* All functions to record tracks event are hooks.
  * The reason for this design is to be able to fetch data returned by hooks.
* All tracks event have the following props (for closer parity with Reader's version):
  * portal = identifies the portal where the event is recorded, e.g. `reader` or `subscriptions`
  * source = (optional) identifies the ui where the event is recorded, e.g. `recommended-site-dismiss-button`, `subscriptions-site-list`
  * subscription_count = user's total subscription count (seems to exist in all reader tracks event, so we're keeping it here too).
* Adds the following tracks event:
  * calypso_subscriptions_site_subscribed
  * calypso_subscriptions_site_unsubscribed
  * calypso_subscriptions_site_resubscribed
  * calypso_subscriptions_site_url_clicked
  * calypso_subscriptions_site_icon_clicked
  * calypso_subscriptions_site_title_clicked
* Reworked previous tracks events:
  * calypso_subscriptions_notifications_toggle_on
  * calypso_subscriptions_notifications_toggle_off
  * calypso_subscriptions_post_emails_toggle_on
  * calypso_subscriptions_post_emails_toggle_off
  * calypso_subscriptions_comment_emails_toggle_on
  * calypso_subscriptions_comment_emails_toggle_off
  * calypso_subscriptions_post_emails_set_frequency
* Some tracks event contains additional reader-related operations (bumpStats, recordGaEvent, etc.)

## Testing Instructions

* Open up `/subscriptions`.
* Open up network panel and filter to `*.gif`.
* Clicking on site title, site icon & site url should log tracks event.
* Changing all site subscription settings should log tracks event.
* Subscribing, unsubscribing and resubscribing should log tracks event.

## Pre-merge Checklist

- [ ] Has the general commit checklist been followed? (PCYsg-hS-p2)
- [ ] [Have you written new tests](https://wpcalypso.wordpress.com/devdocs/docs/testing/index.md) for your changes?
- [ ] Have you tested the feature in Simple (P9HQHe-k8-p2), Atomic (P9HQHe-jW-p2), and self-hosted Jetpack sites (PCYsg-g6b-p2)?
- [ ] Have you checked for TypeScript, React or other console errors?
[...]

Issue #78372, opened 2023-06-19
Subscription Management: Record tracks event when subscribing and unsubscribing from sites.


yansern:
When user (re)subscribes or unsubscribes from a site, we should record the following tracks event:
- calypso_subscriptions_site_subscribed
- calypso_subscriptions_site_unsubscribed

On subscriptions portal, It should contain the following tracks properties:
- portal: reader | subscriptions
- url

On reader portal, it should additionally these properties:
<img width="302" alt="image" src="https://github.com/Automattic/wp-calypso/assets/1287077/dbc7d0b1-ac13-4b3f-a864-ff18e9917aa7">

The finer details needs to be sorted out here later.

PR #78601, opened 2023-06-24
Subscription Management: Rework recommended site tracks implementation.

### LINKED ISSUE #78487: Subscription Management: Record tracks event when clicking on site title, site icon and site url.

yansern:
Fixes https://github.com/Automattic/wp-calypso/issues/78487 https://github.com/Automattic/wp-calypso/issues/78372

This PR is branched off `add/subscriptions-tracks-base` from the following PR. https://github.com/Automattic/wp-calypso/pull/78580

## Proposed Changes

* No more relying on reader's `stats.js` library as that library is code soup.
* Migrate over to the official functions `recordTrainTracksRender` and `recordTrainTracksInteract` function from `@automattic/calypso-analytics`.
* All TrainTracksRender & TrainTracksInteract track event properties are cleaned-up to ensure they stick to the proper usage guideline set out on the fieldguide.
* Introduced the following tracks events:
  * calypso-subscriptions-recommended-site-subscribed
  * calypso-subscriptions-recommended-site-unsubscribed (unused)
  * calypso-subscriptions-recommended-site-dismissed
* Ensure additional `bumpStats` and `gaRecordEvent` still continues to execute for parity with Reader.

## Testing Instructions

**Setup:**
* Set `localStorage.setItem( 'debug', 'calypso:analytics*' );` on your console.
* On your network panel, filter to `.gif`.

**Test:**
* Go to `/read/subscriptions`.
* Try the following actions on the recommended site:
  * Subscribe a recommended site
  * Dismiss a recommended site
  * Click on the site title
  * Click on the site icon
  * Click on the site url
* Monitor the tracks event on the console log or the network panel.

## Pre-merge Checklist

### DIFF
--- a/client/landing/subscriptions/components/site-subscriptions-manager/site-row.tsx
+++ b/client/landing/subscriptions/components/site-subscriptions-manager/site-row.tsx
@@ -16,6 +16,8 @@ import {
 	useRecordPostEmailsToggle,
 	useRecordCommentEmailsToggle,
 	useRecordPostEmailsSetFrequency,
+	SOURCE_SUBSCRIPTIONS_SITE_LIST,
+	SOURCE_SUBSCRIPTIONS_UNSUBSCRIBED_NOTICE,
 } from '../../tracks';
 import { Link } from '../link';
 import { SiteSettingsPopover } from '../settings';
@@ -74,8 +76,8 @@ type SiteRowProps = Reader.SiteSubscription & {
 };
 
 const SiteRow = ( {
-	blog_ID,
-	feed_ID,
+	blog_ID: blog_id,
+	feed_ID: feed_id,
 	name,
 	site_icon,
 	URL: url,
@@ -123,11 +125,6 @@ const SiteRow = ( {
 	const recordSiteUnsubscribed = useRecordSiteUnsubscribed();
 	const recordSiteResubscribed = useRecordSiteResubscribed();
 
-	// Make object assignment a little easier
-	const SOURCE_SUBSCRIPTIONS_SITE_LIST = 'subscriptions-site-list';
-	const blog_id = blog_ID;
-	const feed_id = feed_ID;
-
 	const unsubscribeSuccessCallback = () => {
 		recordSiteUnsubscribed( { blog_id, url, source: SOURCE_SUBSCRIPTIONS_SITE_LIST } );
 		dispatch(
@@ -141,7 +138,7 @@ const SiteRow = ( {
 						recordSiteResubscribed( {
 							blog_id,
 							url,
-							source: 'subscriptions-unsubscribed-notice',
+							source: SOURCE_SUBSCRIPTIONS_UNSUBSCRIBED_NOTICE,
 						} );
 					},
 				}
@@ -153,13 +150,13 @@ const SiteRow = ( {
 
 	const siteTitleUrl = useMemo( () => {
 		if ( portal === ReaderPortal ) {
-			return `/read/feeds/${ feed_ID }`;
+			return `/read/feeds/${ feed_id }`;
 		}
 
 		if ( portal === SubscriptionsPortal ) {
-			return `/subscriptions/site/${ blog_ID }`;
+			return `/subscriptions/site/${ blog_id }`;
 		}
-	}, [ blog_ID, feed_ID, portal ] );
+	}, [ blog_id, feed_id, portal ] );
 
 	const handleNotifyMeOfNewPostsChange = ( send_posts: boolean ) => {
 		// Update post notification settings
@@ -285,7 +282,7 @@ const SiteRow = ( {
 					updatingEmailMeNewComments={ updatingEmailMeNewComments }
 					onUnsubscribe={ () =>
 						unsubscribe(
-							{ blog_id: blog_ID, url, doNotInvalidateSiteSubscriptions: true },
+							{ blog_id, url, doNotInvalidateSiteSubscriptions: true },
 							{ onSuccess: unsubscribeSuccessCallback }
 						)
 					}
--- a/client/landing/subscriptions/tracks/index.ts
+++ b/client/landing/subscriptions/tracks/index.ts
@@ -9,6 +9,5 @@ export { default as useRecordCommentEmailsToggle } from './use-record-comment-em
 export { default as useRecordNotificationsToggle } from './use-record-notifications-toggle';
 export { default as useRecordPostEmailsSetFrequency } from './use-record-post-emails-set-frequency';
 export { default as useRecordPostEmailsToggle } from './use-record-post-emails-toggle';
-// export { default as useRecordRecommendedSiteSubscribed } from './use-record-recommended-site-subscribed';
-// export { default as useRecordRecommendedSiteUnsubscribed } from './use-record-recommended-site-unsubscribed';
-// export { default as useRecordRecommendedSiteDismissed } from './use-record-recommended-site-dismissed';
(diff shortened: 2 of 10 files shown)


client/landing/subscriptions/components/site-subscriptions-manager/site-row.tsx
```diff
@@ -74,8 +76,8 @@ type SiteRowProps = Reader.SiteSubscription & {
 };
 
 const SiteRow = ( {
-	blog_ID,
-	feed_ID,
+	blog_ID: blog_id,
+	feed_ID: feed_id,
```
mashikag, line 80: Thanks for addressing those. 🙂

client/landing/subscriptions/tracks/index.ts
```diff
@@ -1,3 +1,5 @@
+export const SOURCE_SUBSCRIPTIONS_SITE_LIST = 'subscriptions-site-list';
```
mashikag: This looks a bit dangerous to me. Using the constants inside any of the exported hooks could lead to a circular dependency. I realize this is something that we are probably not doing ATM, but it is potentially dangerous pattern. The safest option would be to have a separate constants file.

client/reader/recommended-sites/recommended-site.tsx
```diff
@@ -24,20 +27,19 @@ import {
...
 import { getStreamUrl } from 'calypso/reader/route';
-import { recordFollow } from 'calypso/reader/stats';
 import { Feed } from 'calypso/state/data-layer/wpcom/read/feed/types';
 import { Site } from 'calypso/state/data-layer/wpcom/read/sites/types';
-import { recordReaderTracksEvent } from 'calypso/state/reader/analytics/actions';
 import { follow } from 'calypso/state/reader/follows/actions';
 import isReaderFollowFeedLoading from 'calypso/state/reader/follows/selectors/is-reader-follow-feed-loading';
 import { dismissSite } from 'calypso/state/reader/site-dismissals/actions';
-import { READER_FOLLOWING_MANAGE_RECOMMENDATION } from '../follow-sources';
+import { useSubscriptionManagerContext } from '../../landing/subscriptions/components/subscription-manager-context';
```
mashikag: We try to stick to the absolute import paths, if the imported module is two or more dir levels up. This import could be re-written as `calypso/landing/subscriptions/components/subscription-manager-context`

client/reader/recommended-sites/recommended-site.tsx
```diff
@@ -108,25 +101,102 @@ const RecommendedSite = ( {
 
 	useEffect( () => {
 		if ( railcar ) {
-			recordTracksRailcarRender( 'recommended_site', railcar, {
-				ui_algo: 'following_manage_recommended_site',
-				ui_position: uiPosition,
+			// reader: railcar, ui_algo: following_manage_recommended_site, ui_position, fetch_algo, fetch_position, rec_blog_id (incorrect: fetch_lang, action)
+			// submain: railcar, ui_algo: subscriptions_recommended_site, ui_position, fetch_algo, fetch_position, rec_blog_id
```
mashikag: I believe you have a typo here. I think you intended it to write it as: `// subman: ...`. 
However, I think the abbreviation is not very obvious, perhaps spelling it out fully as `// subscriptions management: ...` could make the comment more understandable? 😁

client/reader/recommended-sites/recommended-site.tsx
```diff
@@ -108,25 +101,102 @@ const RecommendedSite = ( {
...
-	}, [ railcar, uiPosition ] );
+	}, [ railcar, railcarId, uiPosition ] );
 
-	const recordEvent = ( eventName: ReaderEvent ) => {
-		const eventProps = {
-			blog_id: siteId,
-			feed_id: feedId,
-			source: READER_FOLLOWING_MANAGE_RECOMMENDATION,
-		};
+	const { isReaderPortal } = useSubscriptionManagerContext();
```
mashikag, line 119: Hmm, this seems to me like a bit of over engineering for now. The component is defined in, and used only, in the reader. The `isReaderPortal` will always be `true`. One alternative implementation, that comes to my mind, would be to define reader specific tracks hooks in the reader space, perhaps in: `calypso/reader/stats`.

client/reader/recommended-sites/recommended-site.tsx
```diff
@@ -99,6 +91,7 @@ const RecommendedSite = ( {
 }: RecommendedSiteProps ) => {
 	const translate = useTranslate();
 	const dispatch = useDispatch();
+	const railcarId = railcar?.railcar as string;
```
mashikag: There should be no need to use `as` here. Because you are assuming that `railcar?.railcar` is a string, this means that we should expect the `railcar` object to always be there (i.e. make it required), and it should have a required `railcar` prop of type `string`. 🤔 
Unless the railcar object may actually be undefined. Then this should be handled accordingly too, without the need to override the type. I can help you look at this, if you'd like to.

yansern: Yes please. :)

client/reader/recommended-sites/recommended-site.tsx
yansern: It's just to wrap the additional bumpStats/gaEvent codes in what is considered a Reader-specific stat tracking. Sure, I could use just `// inline comments` instead but by wrapping it in actual `isReaderPortal` logic now, the Recommended Sites section can be enabled on the standalone subscriptions portal with less work.

client/reader/recommended-sites/recommended-site.tsx
```diff
@@ -108,25 +101,102 @@ const RecommendedSite = ( {
...
-		const eventProps = {
-			blog_id: siteId,
-			feed_id: feedId,
-			source: READER_FOLLOWING_MANAGE_RECOMMENDATION,
-		};
+	const { isReaderPortal } = useSubscriptionManagerContext();
 
-		dispatch( recordReaderTracksEvent( eventName, eventProps ) as unknown as AnyAction );
+	const blog_id = siteId as unknown as string;
+	const feed_id = feedId as unknown as string;
```
mashikag: This is not going to ensure the actual type to be `string`. Those are only runtime checks. Instead you should:

```
const blog_id = String( siteId );
const feed_id = feedId !== undefined ? String( feedId ) : undefined;
```
However, it looks like `recordSiteIconClicked` requires the `feed_id` to be defined. This means that we should either return some default string value, when it is `undefined`, or we should make the `feed_id` optional for  `recordSiteIconClicked`, if this is really the case, or perhaps the `RecommendedSiteProps` should have the `feedId` as required.

client/reader/recommended-sites/recommended-site.tsx
yansern: Done. https://github.com/Automattic/wp-calypso/pull/78601/commits/476b2557c783294410a69241444e5b4f03f558b1

client/reader/recommended-sites/recommended-site.tsx
yansern: Done. https://github.com/Automattic/wp-calypso/pull/78601/commits/476b2557c783294410a69241444e5b4f03f558b1

client/landing/subscriptions/tracks/index.ts
yansern: Done. https://github.com/Automattic/wp-calypso/pull/78601/commits/29a8d90fcf167337534e8201dec86763ceb27f70

